PR #4330, opened 2022-05-05
change(doc): add transaction index diagram to RFC-0005

conradoplg:
## Motivation

The transaction index might be a bit difficult to understand, this adds a Mermaid diagram to illustrate it.

[Rendered](https://github.com/ZcashFoundation/zebra/blob/b4300f36c329c74096176bef6dbe6d7ec233449a/book/src/dev/rfcs/0005-state-updates.md#rocksdb-data-structures) (scroll down to see diagram)

## Review

Anyone can review, it's not urgent.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #4359, opened 2022-05-09
fix(doc): add support for Mermaid to render graphs

conradoplg:
## Motivation

https://github.com/ZcashFoundation/zebra/pull/4330 added a Mermaid diagram that, while supported natively in GitHub, is not supported natively in mdbook; making it [fail to render](https://zebra.zfnd.org/dev/rfcs/0005-state-updates.html#rocksdb-data-structures) in the website.

## Solution

Use [mdbook-mermaid](https://github.com/badboy/mdbook-mermaid) to add support to mermaid.js to render it on-the-fly.

Another approach would be to render the graphs and upload them, but this seems easier.

Should render like this

image

## Review

Not urgent, anyone can review.

### Reviewer Checklist

  - [ ] Tests for Errors


dconnolly (approved):
🧜‍♀️

PR #4369, opened 2022-05-10
fix(build): put gRPC tests behind an optional feature flag to fix production build issues

teor2345:
## Motivation

This fixes a production build failure introduced by the gRPC tests in #3512.

Closes #4367
Depends-On: #4370, #4372

### Specifications

https://doc.rust-lang.org/cargo/reference/features.html
https://doc.rust-lang.org/cargo/reference/build-scripts.html

### Designs

Put the gRPC tests and dependencies behind an optional `lightwalletd-grpc-tests` feature, so they don't get compiled in production builds.

## Solution

Fixes:
- Make lightwalletd gRPC tests depend on a new lightwalletd-grpc-tests feature
- Activate lightwalletd-grpc-tests feature for gRPC tests
- Document Zebra's optional production and test feature flags

Performance:
- Activate lightwalletd-grpc-tests feature to avoid rebuilds
- Remove redundant cmake and protobuf-compiler dependencies
- Minimise dependencies in Dockerfiles

Dependencies:
- PR #4370 
- PR #4372
- Cherry-pick "remove enable-sentry feature from tests" from PR #4268, and resolve conflicts

### Testing

The "Install zebrad from lockfile without cache" CI job runs without these dependencies, and the gRPC tests run with them. So this change is already covered by existing tests.

## Review

Gustavo and Deirdre should review this PR.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #4375, opened 2022-05-11
fix(ci): make test selection and logging consistent

teor2345:
## Motivation

We call tests in a few different ways, we should use the same test selection and logging arguments every time.

Depends-On: #4369

## Solution

Call tests consistently:
- select the specific package and test binary for each test
- when selecting a specific test, use `--include-ignored` to search all tests, so `#[ignored]` is not a breaking change
- use `--nocapture` to see test logs
- use the same order of arguments
- remove redundant argument quoting

Related changes:
- remove a redundant build step
- handle more errors in entrypoint.sh

## Review

I think Gustavo will probably want to review this PR.

### Reviewer Checklist

  - [ ] Tests for Errors


.github/workflows/continous-integration-os.yml
```diff
@@ -264,17 +265,6 @@ jobs:
...
-      - name: cargo fetch
-        uses: actions-rs/cargo@v1.0.3
-        with:
-          command: fetch
-
-      - name: Build
-        uses: actions-rs/cargo@v1.0.3
-        with:
-          command: build
-          args: --verbose --release
```
dconnolly, line 276: Why remove?

gustavovalverde (dismissed):
LGTM

.github/workflows/continous-integration-os.yml
gustavovalverde: I can't talk on Teor's behalf, but when I saw this being removed, and checking the previous steps, it thought this was done as this is not doing any integration test nor any different than what's being validated in previous steps (redundancy) or the Dockerfile, but to be consistent then the name should also be changed:

https://github.com/ZcashFoundation/zebra/blob/f2a3919d2df0c1db857178f3d41a4f199280dc37/.github/workflows/continous-integration-os.yml#L244

teor2345: We already do similar builds in:

OS:
- Install zebrad from lockfile without cache on ubuntu-latest
- Run tests (because we build the tests)

Docker:
- Both Docker builds (zcash-params and docker/Dockerfile)

So there's no need for a 5th overlapping build step.

dconnolly: OK cool, looks good

dconnolly (approved):
👍

PR #4352, opened 2022-05-07
tests(rpc): add snapshot tests for rpc methods responses

oxarbitrage:
## Motivation

We want to lock the responses of the rpc methods so if anything change, tests will fail.

Close https://github.com/ZcashFoundation/zebra/issues/4131

## Solution

Add snapshots for each rpc method.

## Review

@teor2345 introduced the snapshots and worked in most of them so it can be a good candidate, however anyone who had worked in [this file](https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/src/methods/tests/vectors.rs) before will find this tests easy to understand (with the addition of the snapshot).

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #4384, opened 2022-05-12
fix(rpc): fixes for RPC snapshot tests

teor2345:
## Motivation

These fixes should help with PR #4352.

Some of them are needed to pass CI.

## Solution

Test fixes:
- Snapshot testnet state, RPCs, and transparent addresses (before this fix, we were always snapshotting Mainnet in some tests)

CI fixes:
- Ignore some duplicated dependencies we don't control 

Refactors:
- Create the snapshot settings once

## Review

@oxarbitrage wrote the original PR.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -57,8 +57,12 @@ skip-tree = [
     # wait for halo2 to upgrade
     { name = "blake2b_simd", version = "=0.5.11" },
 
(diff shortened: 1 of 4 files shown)


oxarbitrage (approved):
Thanks, these all look good.

PR #4355, opened 2022-05-09
2. fix(state): index spending transaction IDs for each address

teor2345:
## Motivation

When I compare `zebrad` and `zcashd` RPC responses for `getaddresstxids`, using this command:
```sh
zebra-utils/zcash-rpc-diff 28232 'getaddresstxids { "addresses": ["t3dvVE3SQEi7kqNzwrfNePxZ1d4hUyztBA1"], "start": 1, "end": 1661134 }'
```

All the spending transaction IDs are missing.

Depends-On: #4385

### Specifications

It's not explicitly listed in the docs, but `zcashd` includes spending transaction IDs:
https://zcash.github.io/rpc/getaddresstxids.html

## Solution

- Correctly populate spending address transaction ID indexes
- Increment the database format to rebuild address tx ID indexes

Related changes:
- Split writing transaction indexes into transparent and shielded
- Split writing transparent indexes into created and spent
- Update non-finalized docs to prevent similar bugs 

@gustavovalverde @dconnolly this PR might cause DevOps bugs or slow builds, because it increments the state version.

## Review

@conradoplg and I wrote this code.

I've marked this PR as a high priority, because it changes the state version.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

It's really tricky to test this code, because transparent outputs can't be spent in the first 100 blocks, due to the coinbase maturity consensus rule. I'm testing it locally by doing a full sync, then re-running the manual comparison with `zcashd`.

PR #4385, opened 2022-05-12
1. fix(ci): rebuild cached state before running cached state tests

### LINKED ISSUE #4153: Re-use cached state with the same version from any branch
## Motivation

In PR #4074, we check for unexpected state versions before running Google Cloud tests.

But we want tests to re-use the previous cached state with the same version.

### Designs

We want to re-use the cached state from any branch, as long as:
- the state version is the same, and
- the state kind (checkpoint or tip) is the same.

If there is a state with that version on `main`, and on another branch, use the one from `main`.

If a cached state with that version can't be found, the state should be rebuilt.
(We currently check for changes to the state version `constants.rs` Rust file, but we should check for a matching cached state version instead.)

And then the tests should be run with that new state.

### Out of Scope

Old cached state versions should be cleaned up:
- when the `main` branch has a higher state version.

I'm not sure what we should do if a previous cached state version has a bug. But we can make this decision later, because we only have one state version change planned, and it is a low priority.

As a workaround, we can change the state version in the code, which will regenerate a new cached state.

## Related Work

- #3881 
- #4074

teor2345:
## Motivation

In PR #4355, if the state version changes, we want to rebuild the cached state before running tests.
In #4153, we want to re-use cached state with the same version from any branch.

It turns out these are the same issue.

Depends-On: #4391
Closes #4153.

### Designs

As specified in #4153:

If a cached state with that version can't be found, the state should be rebuilt.
(And then the tests should be run with that new state.)

## Solution

If the PR changes the state version:
- run the cached state rebuilds
- wait for them to finish
- run tests that depend on cached state

If the PR does not change the state version:
- skip the cached state rebuilds
- run tests that depend on cached state

Related fixes:
- Add a missing full-sync -> changed-files job dependency

## Review

This PR blocks PR #4355, so it is urgent.

@gustavovalverde probably understands this part of the code the best.

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

Do we need tickets for these?
- automatically test the cached state rebuild when its workflow files change
    - split the cached state rebuild workflow file from the tests, so test changes don't rebuild cached state
- old cached state versions should be cleaned up when the main branch has a higher state version


teor2345:
There's a bug in this PR - it should have regenerated the cached state disks, but it didn't.

.github/workflows/continous-integration-docker.yml
```diff
@@ -205,8 +208,10 @@ jobs:
       height_grep_text: 'flushing database to disk height=Height'
 
   # Test that Zebra syncs and fully validates a few thousand blocks from a cached mandatory checkpoint disk
+  #
+  # If the state version has changed, waits for the new cached state to be created.
   test-stateful-sync:
-    needs: build
+    needs: [build, regenerate-stateful-disks]
```
teor2345: I think the condition I want here is:
- build has succeeded
- regenerate has succeeded or skipped

(but I think the default is to require both to succeed)

.github/workflows/continous-integration-docker.yml
```diff
@@ -83,6 +83,8 @@ jobs:
         with:
           files: |
             zebra-state/**/constants.rs
+            .github/workflows/continous-integration-docker.yml
+            .github/workflows/deploy-gcp-tests.yml
```
gustavovalverde: I don't think we want disks being rebuilt when changing this files, we can accomplish that outcome (if needed) by doing a manual `workflow_dispatch`. Most changes to this files won't require a new disk.

```suggestion
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -239,8 +246,10 @@ jobs:
       height_grep_text: 'finished initial sync to chain tip, using gossiped blocks sync_percent=100.* current_height=Height'
 
   # Test that Zebra can answer a synthetic RPC call, using a cached Zebra tip state
+  #
+  # If the state version has changed, waits for the new cached state to be created.
   lightwalletd-rpc-test:
-    needs: build
+    needs: [build, test-full-sync]
```
gustavovalverde: `lightwalletd-rpc-test` running after a `test-full-sync` will mostly fail as most of the times the full sync is timing out, and then this test will also fail.

Is this a hard dependency or a soft dependency? If this is a soft dependency (it doesn't require to run from the `latest` cached state everytime) then we should make a PR to create a job (`use-cached-state`) to analyze this requirements and depend on that one.


```suggestion
    needs: [build]
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -256,8 +265,10 @@ jobs:
 
   # Test that Zebra can handle a lightwalletd send transaction RPC call, using a cached Zebra tip state
   # TODO: use a cached lightwalletd tip state to speed up the test (#4303)
+  #
+  # If the state version has changed, waits for the new cached states to be created.
   lightwalletd-transactions-test:
-    needs: build
+    needs: [build, test-full-sync]
```
gustavovalverde: Same as previous

```suggestion
    needs: [build]
```

gustavovalverde (changes requested):
In a broad sense, I feel this PR has two different (and opposing [?]) objectives:
1- `.github/workflows/deploy-gcp-tests.yml` is everything we need to solve #4153, searching for an available cached state in `main` and if not available then on other branches, including the actual. ✅ 

2- Changes in `.github/workflows/continous-integration-docker.yml` are restricting the above approach to only run when a cached state is generated from a `full sync` or `disk rebuild` (as the tests are depending on this jobs) ❓ 

From my POV we should move all changes being done in `.github/workflows/continous-integration-docker.yml` to another PR and merge the ones in `.github/workflows/deploy-gcp-tests.yml` as we need those for `lightwalletd`

teor2345:
> In a broad sense, I feel this PR has two different (and opposing [?]) objectives:

This PR implements all of ticket #4153, which is needed to get the tests to run correctly when we change the state version in a PR. (The objectives aren't opposing, they are complementary.)

> 1- `.github/workflows/deploy-gcp-tests.yml` is everything we need to solve #4153, searching for an available cached state in `main` and if not available then on other branches, including the actual. ✅

The objectives are:

We want to re-use the cached state from any branch, as long as:
- the state version is the same, and
- the state kind is the same (checkpoint or tip, and zebra or lwd)

If there is a state with that version on main, and on another branch, use the one from main.

> 2- Changes in `.github/workflows/continous-integration-docker.yml` are restricting the above approach to only run when a cached state is generated from a `full sync` or `disk rebuild` (as the tests are depending on this jobs) ❓

The objective is:

If a cached state with that version can't be found, the state should be rebuilt.

And then the tests should be run using the newly rebuilt state.

But there's a dependency bug in the workflow logic.

> From my POV we should move all changes being done in `.github/workflows/continous-integration-docker.yml` to another PR and merge the ones in `.github/workflows/deploy-gcp-tests.yml` as we need those for `lightwalletd`

Sure, that makes sense, give me a few minutes.

.github/workflows/continous-integration-docker.yml
teor2345: We want to run more tests automatically, so we don't accidentally merge bugs to the `main` branch.
Is this one of those tests we need to run automatically?

(I think we've merged a few bugs in this workflow to `main` recently.)

teor2345:
> > From my POV we should move all changes being done in `.github/workflows/continous-integration-docker.yml` to another PR and merge the ones in `.github/workflows/deploy-gcp-tests.yml` as we need those for `lightwalletd`
> 
> Sure, that makes sense, give me a few minutes.

This is now split into PR #4391.

.github/workflows/continous-integration-docker.yml
teor2345: This was fixed in commit 53989f9

.github/workflows/continous-integration-docker.yml
teor2345: The goal of this change is to implement ticket #4153:
> If a cached state with that version can't be found, the state should be rebuilt.
> (And then the tests should be run with that new state.)

> `lightwalletd-rpc-test` running after a `test-full-sync` will mostly fail as most of the times the full sync is timing out, and then this test will also fail.

Full sync timeouts are a separate bug, I'd like to deal with them in a separate issue.

And they seem to work fine sometimes:
https://github.com/ZcashFoundation/zebra/runs/6401961075?check_suite_focus=true

> Is this a hard dependency or a soft dependency? If this is a soft dependency (it doesn't require to run from the `latest` cached state everytime) then we should make a PR to create a job (`use-cached-state`) to analyze this requirements and depend on that one.

This is a conditional hard dependency.

If the state version has changed, a state rebuild is required, before we run tests that use cached state.
- Before PR #4391, each test does a checkpoint sync or a full sync, because the state version doesn't match (or some tests fail because the new state version is not in the cache)
- After PR #4391, the test fails because a cached state disk with a matching version is not available

If the state version hasn't changed, the state rebuild is skipped, and the tests use a previously cached state.

teor2345:
@gustavovalverde this is now ready for re-review after the split.

The workflows that were changed in this PR won't run until PR #4391 merges.

.github/workflows/continous-integration-docker.yml
```diff
@@ -239,10 +246,13 @@ jobs:
...
   # Test that Zebra can answer a synthetic RPC call, using a cached Zebra tip state
+  #
+  # If the state version has changed, waits for the new cached state to be created.
+  # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
   lightwalletd-rpc-test:
-    needs: build
+    needs: test-full-sync
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    if: ${{ github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' }}
+    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' }}
```
gustavovalverde, line 310: I'm not sure on the expected behavior from `!cancelled() && !failure()`

Shouldn't be this something like `needs.regenerate-stateful-disks.outputs.conclusion != 'cancelled'` ?

As I'm supposing this is trying to get the previous job conclusion status, and based on that then decide if this job will run or not.

gustavovalverde (changes requested):

teor2345:
This PR doesn't quite work.

Right now the cached state rebuilds only run on `main` branch pushes.
I want them to also run on PR pushes, but only if there isn't a cached state disk already.

This means I need to move the disk search job into another workflow file, and call it before I run any cached state rebuild workflows.

teor2345:
I also need to fix the dependencies of the `lightwalletd-full-sync` job when it merges, and any other new jobs.

.github/workflows/continous-integration-docker.yml
teor2345: > I'm not sure on the expected behavior from `!cancelled() && !failure()`

Here's the behaviour I want:
https://github.com/ZcashFoundation/zebra/blob/b841fe71b2be706f8dcb18a953127bdca85a8773/.github/workflows/continous-integration-docker.yml#L250-L251

But I can't write `success() || skipped()`, because there is no `skipped()`.
So instead I have to write `!cancelled() && !failure()`.

`cancelled()`:
> > Returns true if the workflow was canceled.

`failure()`:
> > Returns true when any previous step of a job fails. If you have a chain of dependent jobs, failure() returns true if any ancestor job fails.

https://docs.github.com/en/actions/learn-github-actions/expressions#cancelled

> Shouldn't be this something like `needs.regenerate-stateful-disks.outputs.conclusion != 'cancelled'` ?

This does the same thing as `!cancelled()`.

> As I'm supposing this is trying to get the previous job conclusion status, and based on that then decide if this job will run or not.

Yes, and that works, but the disk rebuild condition doesn't:
https://github.com/ZcashFoundation/zebra/pull/4385#issuecomment-1125623757

gustavovalverde: Conditionals reference: https://github.com/actions/runner/issues/491#issuecomment-926924523

gustavovalverde:
Keeping this here as a reference to continue working on this later: https://github.com/actions/runner/issues/491#issuecomment-926924523

teor2345:
Here is my plan for finishing this PR:

A new refactor PR:
- split the `get-disk-name` step into its own sub-workflow
- replace `changed-files` with 3 calls to `get-disk-name`:
    - zebrad/checkpoint
    - zebrad/tip
    - lwd/tip
- pass the correct disk name in each gcp deploy workflow call
- remove unused workflow parameters

This PR:
- if a disk name is missing, do the corresponding rebuild, otherwise, skip the rebuild
- after the disk is available from the cache or rebuild, run the jobs that depend on it

Edit: I haven't started the refactor, so if @gustavovalverde wanted to do it, that would be ok. But no pressure!

gustavovalverde:
After evaluating between `workflow_run` and `workflow_call` I think using `workflow_run` is a best approach for this specific use case. I'll soon push this changes with proposals from https://github.com/ZcashFoundation/zebra/pull/4385#issuecomment-1127318235

Reference to `workflow_run` https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

gustavovalverde:
Just forget about the previous comment as a caller workflow does not support variables as an input 
image

We can do it using secrets, but that's just obfuscating information that should not be obfuscated

gustavovalverde:
@teor2345 could you confirm me this assumptions?

- We always want to search for a disk matching the actual version from `constants.rs`
- As far as the previous statement is true we don't need to validate if `constants.rs` was updated or not
- If no disk is found with this version (in `main` or another `branch`) then we want to rebuild

We can have a job that makes this validations for `lwd` and `zebra`, if it returns false it builds the corresponding cached state, if it returns true it continues as expected.

I think this is the less invasive implementation, just want to confirm I can continue with those assumptions.

teor2345:
> * We always want to search for a disk matching the actual version from `constants.rs`
> * As far as the previous statement is true we don't need to validate if `constants.rs` was updated or not

Yes, checking the exact version is enough.

And we need to remove the `changed-files` check, because it will trigger an unnecessary rebuild on each push to the PR. (If the PR previously did a rebuild, then the next run should be using that cache.)

> * If no disk is found with this version (in `main` or another `branch`) then we want to rebuild

This all looks good.

> We can have a job that makes this validations for `lwd`

`lwd` does not use the Zebra state version, so it does not need to be rebuilt when the Zebra state version changes.
(`lwd` currently gets rebuilt on every PR, which is what we need for our testing.)

> and `zebra`, if it returns false it builds the corresponding cached state, if it returns true it continues as expected.

Only the Zebra cache needs to be rebuilt if the Zebra state version changes.

Edit: if there is a Zebra state version rebuild, all the jobs that use the cached Zebra state need to wait, and run after the rebuild is successful.

gustavovalverde:
This could be prettier (?) https://github.com/ZcashFoundation/zebra/pull/4385/files#diff-4dc8170bbb198e5ddb8c8d186aa2cf20ee5117960f71ad9197b2a800ad878bceR104 but in the meanwhile is functional

cc: @dconnolly to have a look at it

.github/workflows/continous-integration-docker.yml
```diff
@@ -186,15 +232,19 @@ jobs:
...
-    needs: [ build, changed-files]
+    needs: [ build, get-available-disks ]
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    # Only run this job if the database format version has (likely) changed.
-    if: ${{ (needs.changed-files.outputs.rebuild == 'true' && github.event_name == 'push') || github.event.inputs.regenerate-disks == 'true' }}
+    # Only run this job if there's no available disk matching the actual state version from constants.rs,
+    # or if we manually execute a workflow_dispatch using regenerate-disks.
+    #
+    # Note: the output from get-available-disks should match with the caller workflow inputs
+    if: ${{ (fromJSON(needs.get-available-disks.outputs.zebra_checkpoint_disk) && github.event_name == 'push') || github.event.inputs.regenerate-disks == 'true' }}
```
teor2345: Question: I'm not sure if we need the `&& github.event_name == 'push'` filter here. Why are we excluding `pull_request` events?

```suggestion
    if: ${{ fromJSON(needs.get-available-disks.outputs.zebra_checkpoint_disk) || github.event.inputs.regenerate-disks == 'true' }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -217,18 +270,21 @@ jobs:
...
-    needs: build
+    needs: [ build, get-available-disks ]
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    # TODO change `github.ref_name == 'main'` to `startsWith(github.head_ref, 'mergify/merge-queue/')`
-    # to only run on Mergify head branches, and on manual dispatch:
+    # to also run on Mergify head branches,
+    # add `|| (github.event_name == 'push' && startsWith(github.head_ref, 'mergify/merge-queue/'))`:
     # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request-1
-    if: ${{ (github.event_name == 'push' && github.ref_name == 'main') || (needs.changed-files.outputs.rebuild == 'true' && github.event_name == 'push') || github.event.inputs.run-full-sync == 'true' }}
+    if: ${{ (github.event_name == 'push' && github.ref_name == 'main') || (fromJSON(needs.get-available-disks.outputs.zebra_tip_disk) && github.event_name == 'push') || github.event.inputs.run-full-sync == 'true' }}
```
```suggestion
    if: ${{ (github.event_name == 'push' && github.ref_name == 'main') || fromJSON(needs.get-available-disks.outputs.zebra_tip_disk) || github.event.inputs.run-full-sync == 'true' }}
```

teor2345 (commented):
Looks good, I really like how you've done this!

I just need to check if we want to tag beta.10 before we merge this risky CI change.

.github/workflows/continous-integration-docker.yml
```diff
@@ -255,10 +318,18 @@ jobs:
...
+  # Runs:
+  # - after every PR is merged to `main`
+  # - on every PR update
+  #
+  # If the state version has changed, waits for the new cached states to be created.
+  # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
+  # TODO: move this job under lightwalletd-full-sync to have a sequential logic
   lightwalletd-transactions-test:
-    needs: build
+    needs: [ test-full-sync, lightwalletd-full-sync ]
```
teor2345: nitpick: this dependency is redundant, because `lightwalletd-full-sync` depends on `test-full-sync`

```suggestion
    needs: lightwalletd-full-sync
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -292,10 +370,16 @@ jobs:
...
   # Test update sync of lightwalletd with a lightwalletd and Zebra tip state
+  # Runs:
+  # - after every PR is merged to `main`
+  # - on every PR update
+  #
+  # If the state version has changed, waits for the new cached states to be created.
+  # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
   lightwalletd-update-sync:
-    needs: build
+    needs: [ test-full-sync, lightwalletd-full-sync ]
```
```suggestion
    needs: lightwalletd-full-sync
```

teor2345 (commented):
This seems to work well without a state version change (this PR), but fail with a state version change (PR #4355).

Here is what I'm seeing:
* test-full-sync doesn't run, but it needs to, which causes disk search failures in:
    * fully-synced-rpc: https://github.com/ZcashFoundation/zebra/runs/6498847914?check_suite_focus=true
    * lwd-send-transactions: https://github.com/ZcashFoundation/zebra/runs/6498851161?check_suite_focus=true
* regenerate-stateful-disks doesn't run, which causes disk search failures in:
    * sync-past-checkpoint: https://github.com/ZcashFoundation/zebra/runs/6498848031?check_suite_focus=true
*  lwd-full-sync runs, but looks for the wrong disk (should be zebrad-cache)
    * https://github.com/ZcashFoundation/zebra/runs/6498847837?check_suite_focus=true
    * which causes disk search failures in lwd-update-sync: https://github.com/ZcashFoundation/zebra/runs/6498851042?check_suite_focus=true

.github/workflows/continous-integration-docker.yml
gustavovalverde: We added this as the PR was building the cached state, and then mergify was building the cached state once again.


~When a commit is pushed to an opened PR (sync event) this still applies, so we're not excluding PRs as a whole. But when mergify opens a PR, it's (commonly) not adding new commits. I think we don't want to rebuild (as it takes hours) once again on the mergify queue.~

gustavovalverde: We added this as the PR was building the cached state, and then mergify was building the cached state once again.


When a commit is pushed to an opened PR (sync event) this still applies, so we're not excluding PRs as a whole. But when mergify opens a PR, it's (commonly) not adding new commits. I think we don't want to rebuild (as it takes hours) once again on the mergify queue.

gustavovalverde: Nevermind, you were right.

gustavovalverde (dismissed):
I've included @teor2345 suggestions, which also fixed the conditions

teor2345 (commented):
This PR needs a few changes:
- keep the patch file in sync: https://github.com/ZcashFoundation/zebra/pull/4355#discussion_r877150370
- resolve conflicts with PR #4355 
- make sure there are no new CI changes in  PR #4355's diff (they should all be in this PR's diff)

gustavovalverde (approved):
Remaining suggestions applied

PR #4393, opened 2022-05-13
fix(rpc): redact 6 kB of block data in snapshots

teor2345:
## Motivation

PR #4352 adds 6 kB of block data in snapshot files.

Depends-On: #4352

## Solution

Instead, we can assert the data matches the block data, and then redact the data in the snapshot.

## Review

@oxarbitrage wrote PR #4352.

### Reviewer Checklist

  - [x] Code implements Specs and Designs
  - [x] Tests for Expected Behaviour
  - [ ] Tests for Errors


zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -183,37 +184,48 @@ fn snapshot_rpc_getblockchaininfo(info: GetBlockChainInfo, settings: &insta::Set
...
 fn snapshot_rpc_getaddressbalance(address_balance: AddressBalance, settings: &insta::Settings) {
     settings.bind(|| insta::assert_json_snapshot!("get_address_balance", address_balance));
 }
 
-/// Snapshot `getblock` response, using `cargo insta` and RON serialization.
-fn snapshot_rpc_getblock(block: GetBlock, settings: &insta::Settings) {
-    settings.bind(|| insta::assert_json_snapshot!("get_block", block));
+/// Check `getblock` response, using `cargo insta`, JSON serialization, and block test vectors.
+///
+/// Does not create a snapshot file.
```
teor2345:
```suggestion
///
/// The snapshot file does not contain any data, but it does enforce the response format.
```

dconnolly (approved):

zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -183,37 +184,48 @@ fn snapshot_rpc_getblockchaininfo(info: GetBlockChainInfo, settings: &insta::Set
...
+/// The snapshot file does not contain any data, but it does enforce the response format.
+fn snapshot_rpc_getblock(block: GetBlock, block_data: &[u8], settings: &insta::Settings) {
+    let block_data = hex::encode(block_data);
+
+    settings.bind(||
+                  insta::assert_json_snapshot!("get_block", block, { "." => insta::dynamic_redaction(move |value, _path| {
+                      // assert that the block data matches, without creating a 1.5 kB snapshot file
+                      assert_eq!(value.as_str().unwrap(), block_data);
+
+                      "[block_data]"
```
oxarbitrage:
```suggestion
                      "[BlockData]"
```

I will like to be consistent with other redactions like `Height` or `SubVersion`. We might change thoses or this one to lower case.

zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -183,37 +184,48 @@ fn snapshot_rpc_getblockchaininfo(info: GetBlockChainInfo, settings: &insta::Set
...
+                  }),})
```
oxarbitrage:
```suggestion
                  }),
              })
```
Not sure if rustfmt will allow it but i prefer to have this in separated lines of code. (untested)

teor2345: Fixed in 91e14bb

PR #4397, opened 2022-05-14
Fix typos

rex4539:


teor2345 (changes requested):
Thanks for these fixes!

We can't accept any changes to `book/mermaid.min.js`, because it is a generated file from an external source.

All the other changes are ok.

teor2345 (approved):

PR #3990, opened 2022-03-28
feat(rpc): Implement `z_gettreestate` RPC

upbqdn:
## Motivation

To support `lightwalletd`, Zebra needs to implement the `z_gettreestate` RPC documented here:  https://zcash.github.io/rpc/z_gettreestate.html.

## Solution

This PR implements the `z_gettreestate` RPC so that Sapling & Orchard commitment trees in the response are compatible with the ones returned by `zcashd`, and we ignore Sprout. The trees in the JSON response are represented as hex strings under the key `"finalState"`. Zebra stores the trees internally in a so-called "dense" format, while `lightwalletd` expects the trees in a "sparse" format. For this reason, we convert the trees into the sparse format before the serialization. However, Orchard trees returned by `zcashd` contain empty trailing parent nodes, while Sapling trees don't. `librustzcash` supports a conversion that includes the empty trailing parent nodes, so we utilize it for Orchard; and to ensure compatibility with `zcashd`, we implement our own conversion for Sapling so that the returned trees don't contain the empty trailing parent nodes.

Closes #3156.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs

PR #4401, opened 2022-05-16
tests(rpc): Add `z_get_treestate` snapshot test

### LINKED ISSUE #4379: Add `z_gettreestate` response snapshot test
In https://github.com/ZcashFoundation/zebra/pull/4352 we added snapshot tests for rpc methods responses however, the `z_gettreestate` method is not in the main branch at the moment of writing so it needs to be added after https://github.com/ZcashFoundation/zebra/pull/3990 is in place.

oxarbitrage:
## Motivation

In https://github.com/ZcashFoundation/zebra/pull/4352 we added snapthot tests for rpc methods responses. We left `z_get_treestate` out because it was work in progress at the moment of creating the tests. Now that https://github.com/ZcashFoundation/zebra/pull/3990 is merged we can add the missing test.

Close https://github.com/ZcashFoundation/zebra/issues/4379

## Solution

Imnplement a very basic tests where block height 1 is queried for tree states.

## Review

Anyone can review.

### DIFF
--- a/zebra-rpc/src/methods/tests/snapshots/z_get_treestate@mainnet_10.snap
+++ b/zebra-rpc/src/methods/tests/snapshots/z_get_treestate@mainnet_10.snap
@@ -1,0 +1,10 @@
+---
+source: zebra-rpc/src/methods/tests/snapshot.rs
+assertion_line: 226
+expression: tree_state
+---
+{
+  "hash": "0007bc227e1c57a4a70e237cad00e7b7ce565155ab49166bc57397a26d339283",
+  "height": 1,
+  "time": 1477671596
+}
--- a/zebra-rpc/src/methods/tests/snapshot.rs
+++ b/zebra-rpc/src/methods/tests/snapshot.rs
@@ -16,9 +16,6 @@ use zebra_test::mock_service::MockService;
 use super::super::*;
 
 /// Snapshot test for RPC methods responses.
-///
-/// TODO:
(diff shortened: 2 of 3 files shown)


zebra-rpc/src/methods/tests/snapshots/z_get_treestate@mainnet_10.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: zebra-rpc/src/methods/tests/snapshot.rs
+assertion_line: 226
+expression: tree_state
+---
+{
+  "hash": "0007bc227e1c57a4a70e237cad00e7b7ce565155ab49166bc57397a26d339283",
+  "height": 1,
+  "time": 1477671596
```
teor2345, line 9: @upbqdn @conradoplg is this the tree state we expect to see at height 1?

teor2345 (commented):
Looks good, just wanted to double-check the exact format.

zebra-rpc/src/methods/tests/snapshots/z_get_treestate@mainnet_10.snap
conradoplg: Yep, no Sapling nor Orchard data at that point so it looks right

conradoplg (approved):
Looks good, thanks! It would be nice to add another test with a higher block so that we test the Sapling and Orchard data, but we can do that after the cached state issues are sorted out

Issue #4340, opened 2022-05-06
ci: Require successful lightwalletd integration tests for PRs to merge


teor2345:
## Motivation

We're running a lot of lightwalletd integration tests in CI, but most of them are not required to merge.

While we are actively working on lightwalletd, we should require all the tests to pass before merging.

## Designs

- We'll need a patch job for each test
- Then we need to add the tests to the `main` branch protection rule

### List of Remaining Tests

Zebra:
- [x] Zebra checkpoint / Run sync-to-checkpoint test
- [x] Zebra checkpoint / Run sync-past-checkpoint test
- [x] Zebra tip / Run full-sync-to-tip test

Lightwalletd:
- [x] lightwalletd tip / Run lwd-full-sync test
- [x] lightwalletd tip / Run send-transactions test (sp?)
- [x] lightwalletd tip / Run lwd-update-tip test (sp?)
- [x] lightwalletd tip / Run lwd-grpc test (sp?)

### Non-Lightwalletd Tests

- [x] Coverage on stable

ftm1000:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/4340?planning-poker) with ZenHub @conradoplg @dconnolly @oxarbitrage @teor2345 @upbqdn

teor2345:
@gustavovalverde I can do this ticket if you'd like. Are there any blockers?

I can start on the tests that are already in CI, and do new ones as they are added.

gustavovalverde:
You can go ahead if you'd like. This one has to be merged first, as it's causing the tests not to run:
- https://github.com/ZcashFoundation/zebra/pull/4364

teor2345:
This is now done in the GitHub settings for Zebra.

PR #4411, opened 2022-05-17
fix(ci): fix errors in Docker entrypoint changes

teor2345:
## Motivation

PR #4409 had a quoting bug.

Our tests should have stopped that PR merging, but we still need to do #4340.

## Solution

- Move shell globs outside quotes
- Print errors and do fallback directory listings
- Limit the length of full directory tree listings
- Remove `exec` and `exit 1` so that we can run commands after `cargo`

## Review

I'm probably going to admin-merge this, because we need some diagnostics before other people start work.


docker/entrypoint.sh
```diff
@@ -21,52 +21,56 @@ case "$1" in
...
         # For these tests, we activate the gRPC feature to avoid recompiling `zebrad`,
         # but we don't actually run any gRPC tests.
         elif [[ "$TEST_FULL_SYNC" -eq "1" ]]; then
             # Run a Zebra full sync test.
-            exec cargo test --locked --release --features lightwalletd-grpc-tests --package zebrad --test acceptance -- --nocapture --include-ignored full_sync_mainnet
+            cargo test --locked --release --features lightwalletd-grpc-tests --package zebrad --test acceptance -- --nocapture --include-ignored full_sync_mainnet
+            # List directory generated by test
             # TODO: replace with $ZEBRA_CACHED_STATE_DIR in Rust and workflows
-            ls -lh "/zebrad-cache/*/*"
+            ls -lh "/zebrad-cache"/*/* || (echo "No /zebrad-cache/*/*"; ls -lhR "/zebrad-cache" || echo "No /zebrad-cache directory")
```
dconnolly: oof

dconnolly (dismissed):

conradoplg (dismissed):
Removing `exec` makes the script fail because it ends in `exit 1`.

To save a bit of time I'll go ahead and remove it just to check if the rest works, but feel free to fix it differently

gustavovalverde:
I think the `ls` command should consider the specific structure required by zebra and lightwalletd cached state for this to work as expected.

This output should fail as it's not the expected structure https://github.com/ZcashFoundation/zebra/runs/6477434066?check_suite_focus=true#step:9:239

P.D.: Seems like @teor2345 were right with their concern on the disk mounting being unreliable. If correct, it's almost certain the workflow will require two separate disks for each cached state (I'll double check this).

teor2345:
> Removing `exec` makes the script fail because it ends in `exit 1`.
> 
> To save a bit of time I'll go ahead and remove it just to check if the rest works, but feel free to fix it differently

This seems to have worked, we're back to the same two failures we were expecting.

teor2345:
> I think the `ls` command should consider the specific structure required by zebra and lightwalletd cached state for this to work as expected.
> 
> This output should fail as it's not the expected structure https://github.com/ZcashFoundation/zebra/runs/6477434066?check_suite_focus=true#step:9:239

The Rust tests already validate the structure, the `ls` command is just designed to list what's there.
(We avoid doing validation in two different places, because then we have to keep them in sync.)

> P.D.: Seems like @teor2345 were right with their concern on the disk mounting being unreliable. If correct, it's almost certain the workflow will require two separate disks for each cached state (I'll double check this).

👍

teor2345:
I admin-merged this PR, so we can get on with merging NU5 mainnet changes.

Issue #4168, opened 2022-04-21
ci: call lightwalletd wallet queries


conradoplg:
## Motivation

We need to call the `lightwalletd_wallet_grpc_tests` tests from #3655 in CI, using the lightwalletd cached state from #4165.

(PR #4267 already added the tests from #3512.)

teor2345:
@gustavovalverde I re-opened this ticket, because we still need to call the `lightwalletd_wallet_grpc_tests` in CI.

PR #4424, opened 2022-05-18
fix(rpc) Use the Sapling activation height in gRPC tests

upbqdn:
## Motivation

`lightwalletd` only supports post-Sapling blocks, so we can't use pre-Sapling blocks in our tests.  

## Solution

This PR uses the Sapling activation height in the gRPC tests, and it also fixes an off-by-one error in counter incrementation.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
+++ b/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
@@ -31,7 +31,9 @@
 use color_eyre::eyre::Result;
 
 use zebra_chain::{
-    block::Block, parameters::Network, parameters::NetworkUpgrade::Canopy,
+    block::Block,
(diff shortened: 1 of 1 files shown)


upbqdn:
Here's how I run the tests:
```bash
export ZEBRA_CACHED_STATE_DIR="~/.cache/zebra"
export LIGHTWALLETD_DATA_DIR="~/zcash/lightwalletd"
export ZEBRA_TEST_LIGHTWALLETD="true"
cargo t --features lightwalletd-grpc-tests -- --ignored --nocapture lightwalletd_wallet_grpc
```

It takes a bit of effort to run the tests.

teor2345 (approved):
Looks good, we'll put these tests in CI in #4168

Issue #4433, opened 2022-05-19
Installation error


ruzcash:
## Description

Hello! I'm trying to install Zebra and I'm getting the error at the very end of compilation:

Compiling zebra-consensus v1.0.0-beta.9 (/home/ak/.cargo/git/checkouts/zebra-d0d93d7ded255019/c190e30/zebra-consensus)
Compiling zebra-rpc v1.0.0-beta.9 (/home/ak/.cargo/git/checkouts/zebra-d0d93d7ded255019/c190e30/zebra-rpc)
error: there is no argument named `address`
   --> zebra-rpc/src/methods.rs:829:69
    |
829 |                     Error::invalid_params(&format!("invalid address {address:?}: {error}"))
    |                                                                     ^^^^^^^^^^^

error: there is no argument named `error`
   --> zebra-rpc/src/methods.rs:829:82
    |
829 |                     Error::invalid_params(&format!("invalid address {address:?}: {error}"))
    |                                                                                  ^^^^^^^

error: there is no argument named `method`
  --> zebra-rpc/src/server/tracing_middleware.rs:44:37
   |
44 |                 format!(r#"method = {method:?}, params = {params:?}"#)
   |                                     ^^^^^^^^^^

error: there is no argument named `params`
  --> zebra-rpc/src/server/tracing_middleware.rs:44:58
   |
44 |                 format!(r#"method = {method:?}, params = {params:?}"#)
   |                                                          ^^^^^^^^^^

error: there is no argument named `method`
  --> zebra-rpc/src/server/tracing_middleware.rs:47:43
   |
47 |                 format!(r#"notification = {method:?}, params = {params:?}"#)
   |                                           ^^^^^^^^^^

error: there is no argument named `params`
  --> zebra-rpc/src/server/tracing_middleware.rs:47:64
   |
47 |                 format!(r#"notification = {method:?}, params = {params:?}"#)
   |                                                                ^^^^^^^^^^

error: there is no argument named `call_description`
  --> zebra-rpc/src/server/tracing_middleware.rs:66:64
   |
66 |             tracing::warn!("Received unrecognized RPC request: {call_description}");
   |                                                                ^^^^^^^^^^^^^^^^^^

error: could not compile `zebra-rpc` due to 7 previous errors
warning: build failed, waiting for other jobs to finish...
[...]

PR #4435, opened 2022-05-19
change: fail early when using Rust 1.57 and earlier, because they don't work

conradoplg:
We recently adopted formatting features introduced in Rust 1.58, but we're not enforcing the minimum version, which caused [issues for people compiling with Ubuntu 20](https://github.com/ZcashFoundation/zebra/issues/4433) (which ships 1.57).

## Solution

Use `rust-version` for `zebrad` to enforce the version and document it.

## Review

Urgent, blocking beta.10 release.

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

I'll create an issue to address this in CI to make sure we always comply with the minimum version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,9 +100,9 @@ install mechanism. To run `zebrad`, follow the instructions to compile `zebrad`
 for your platform:
 
 1. Install [`cargo` and `rustc`](https://www.rust-lang.org/tools/install).
-     - You will need version 1.58 or later; if you're using the version shipped with your OS it
-       might be outdated. Run `rustc --version` to check. If it's outdated, install it following
-       the instructions in the link. We always test with the latest `stable` version.
+     - Zebra is tested with the latest `stable` Rust version.
+        Earlier versions are not supported or tested, but they might work.
+        (Rust 1.57 and earlier are not supported, due to missing features.)
 2. Install Zebra's build dependencies:
      - **libclang:** the `libclang`, `libclang-dev`, `llvm`, or `llvm-dev` packages, depending on your package manager
      - **clang** or another C++ compiler: `g++`, `Xcode`, or `MSVC`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,6 +100,9 @@ install mechanism. To run `zebrad`, follow the instructions to compile `zebrad`
 for your platform:
 
 1. Install [`cargo` and `rustc`](https://www.rust-lang.org/tools/install).
+     - You will need version 1.58 or later; if you're using the version shipped with your OS it
+       might be outdated. Run `rustc --version` to check. If it's outdated, install it following
+       the instructions in the link. We always test with the latest version.
```
dconnolly:
```suggestion
       the instructions in the link. We always test with the latest `stable` version.
```

dconnolly (dismissed):

README.md
```diff
@@ -100,6 +100,9 @@ install mechanism. To run `zebrad`, follow the instructions to compile `zebrad`
 for your platform:
 
 1. Install [`cargo` and `rustc`](https://www.rust-lang.org/tools/install).
+     - You will need version 1.58 or later; if you're using the version shipped with your OS it
+       might be outdated. Run `rustc --version` to check. If it's outdated, install it following
+       the instructions in the link. We always test with the latest `stable` version.
```
teor2345: I don't think we should commit to supporting rust 1.58 unless we have it in CI.

Instead, can we say:
```suggestion
     - Zebra is tested with the latest `stable` Rust version.
        Earlier versions are not supported or tested, but they might work.
        (Rust 1.57 and earlier are not supported, due to missing features.)
```

zebrad/Cargo.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Zcash Foundation <zebra@zfnd.org>"]
 license = "MIT OR Apache-2.0"
 version = "1.0.0-beta.10"
 edition = "2021"
+rust-version = "1.58"
```
teor2345, line 9:
```suggestion
# Zebra is only supported on the latest stable Rust version. Some earlier versions might work.
# Zebra uses features introduced in Rust 1.58.
rust-version = "1.58"
```

dconnolly (approved):

PR #4453, opened 2022-05-21
CI(rpc): Add grpc tests to CI 

### LINKED ISSUE #4168: ci: call lightwalletd wallet queries

oxarbitrage:
## Motivation

We want to add the wallet grpc tests to the CI, similar to how the sending transactions were done.

Close https://github.com/ZcashFoundation/zebra/issues/4168
Depends-On: #4457

## Solution

Add workflow and docker for this tests in https://github.com/ZcashFoundation/zebra/commit/d15f6fcea1d14b5e43a9c0ba5ccc1ca8e651f06c

Moved jobs to fix a `TODO` in https://github.com/ZcashFoundation/zebra/commit/8e002d65aadc5883fffa93628e166447da724a4f

## Review

I don't know if it works so the PR is a draft by now. I will like @gustavovalverde to take a look and help with the testing.


gustavovalverde:
Can we reduce the diff so it's easier to review?

oxarbitrage:
> Can we reduce the diff so it's easier to review?

There are two commits. The first one is where the new stuff is added: https://github.com/ZcashFoundation/zebra/pull/4453/commits/d15f6fcea1d14b5e43a9c0ba5ccc1ca8e651f06c

The second one moves the jobs lower in the file, you can ignore that by now if you want.

gustavovalverde:
Yep, this is the correct way to implement this 👍🏻 

The only thing we should have to confirm this is running as expected is a full sync with the new state version, but since the upgrade all full syncs are timing out. I'll re-run the ones in main once again to see if we're lucky.

cc: @teor2345 we might have to prioritize a long term fix for the full sync (even if that's using a whole new approach)

.github/workflows/continous-integration-docker.yml
```diff
@@ -319,78 +319,102 @@ jobs:
...
-      test_id: lwd-update-sync
-      test_description: Test lightwalletd update sync with both states
-      test_variables: '-e TEST_LWD_UPDATE_SYNC=1 -e ZEBRA_TEST_LIGHTWALLETD=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache -e LIGHTWALLETD_DATA_DIR=/var/cache/lwd-cache'
+      test_id: lwd-grpc-wallet
+      test_description: Test gRPC calls via lightwalletd
+      test_variables: '-e TEST_LWD_GRPC=1 -e ZEBRA_TEST_LIGHTWALLETD=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache -e LIGHTWALLETD_DATA_DIR=/var/cache/lwd-cache'
       needs_zebra_state: true
       needs_lwd_state: true
       saves_to_disk: false
-      disk_prefix: lwd-cache
```
teor2345: Review Diff: Can you please re-order the jobs in a separate PR?

If a PR changes the order and changes the content, it makes it hard to work out what has actually changed.

teor2345:
> cc: @teor2345 we might have to prioritize a long term fix for the full sync (even if that's using a whole new approach)

I opened ticket #4456 for the fix, and PR #4457 as a temporary workaround.

There's still one quick fix left - testing sync speed in different Google Cloud regions.
(After that, we need to investigate the cause of the slow syncs.)

.github/workflows/continous-integration-docker.yml
```diff
@@ -319,78 +319,102 @@ jobs:
...
+  # Test that Zebra can handle gRPC wallet calls, using a cached Zebra tip state
+  #
   # Runs:
   # - after every PR is merged to `main`
   # - on every PR update
   #
   # If the state version has changed, waits for the new cached states to be created.
   # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
-  lightwalletd-update-sync:
+  lightwalletd-grpc-test:
```
teor2345, line 430: This is blocked by PR #4457, which should enable these tests to actually run.

Then we will know if we can merge this PR, or we need to fix the gRPC tests first.

gustavovalverde:
@oxarbitrage it seems like the output from your test is the expected result

teor2345 (approved):
Thanks, looks good!

teor2345:
Looks like a temporary or unrelated failure, I'll retry merging

```
   0: stdout of command did not contain any matches for the given regex

Command:
   "lightwalletd" "--zcash-conf-path" "/tmp/zebrad_testsaMWaXu/lightwalletd-zcash.conf" "--data-dir" "/var/cache/lwd-cache" "--log-file" "/dev/stdout" "--grpc-bind-addr" "127.0.0.1:0" "--http-bind-addr" "127.0.0.1:0" "--no-tls-very-insecure"

Match Regex:
   [
       "Adding new mempool txid",
   ]
```

https://github.com/ZcashFoundation/zebra/runs/6601527110?check_suite_focus=true#step:9:1223

teor2345:
@mergifyio update

teor2345:
@mergifyio update

Issue #4456, opened 2022-05-22
Full sync test takes slightly longer than CI timeout


teor2345:
## Motivation

The full sync test is timing out in CI, but we need it to work to run lightwalletd tests.

Example:
https://github.com/ZcashFoundation/zebra/runs/6539522740?check_suite_focus=true#step:7:11496

### Priority

This is a high-priority, urgent change - we can't do reliable lightwalletd testing without it.

## Alternative Solutions

- [x] #4459

Needs separate scheduling:
- profiling to see where Zebra is waiting - tokio-console - #2112

### Rejected Solutions

We tried these in #4155, but they didn't seem to help:
- move large database reads to a blocking thread, so they don't block network futures
- move hash calculation and signature verification to a blocking thread, so they don't block network futures

## Follow Up Work

- revert #4457

teor2345:
Obsoleted by ticket #4661.

PR #4457, opened 2022-05-22
fix(ci): Temporarily finish full sync at 99%

teor2345:
## Motivation

As a workaround for #4456, we can just finish the full sync slightly early.

This removes coverage for:
- finishing sync
- gossiped blocks
- mempool

But allows us to test lightwalletd.

## Solution

- finish full sync at 99% in CI
- finish full sync at 99% in Rust

## Review

This PR is urgent because it blocks lightwalletd testing.

### Reviewer Checklist

  - [ ] Full sync doesn't time out

## Follow Up Work

We should revert this change once #4456 is fixed.

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -294,7 +294,7 @@ jobs:
       needs_zebra_state: false
       saves_to_disk: true
       disk_suffix: tip
-      height_grep_text: 'estimated progress to chain tip sync_percent=99.2.* current_height=Height'
+      height_grep_text: 'estimated progress to chain tip sync_percent=99.* current_height=Height'
 
   # Test that Zebra can answer a synthetic RPC call, using a cached Zebra tip state
   #
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -294,7 +294,7 @@ jobs:
       needs_zebra_state: false
       saves_to_disk: true
       disk_suffix: tip
-      height_grep_text: 'estimated progress to chain tip sync_percent=99.5.* current_height=Height'
+      height_grep_text: 'estimated progress to chain tip sync_percent=99.2.* current_height=Height'
 
   # Test that Zebra can answer a synthetic RPC call, using a cached Zebra tip state
   #
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -294,7 +294,7 @@ jobs:
       needs_zebra_state: false
       saves_to_disk: true
       disk_suffix: tip
-      height_grep_text: 'finished initial sync to chain tip, using gossiped blocks sync_percent=100.* current_height=Height'
+      height_grep_text: 'estimated progress to chain tip sync_percent=99.5.* current_height=Height'
 
   # Test that Zebra can answer a synthetic RPC call, using a cached Zebra tip state
   #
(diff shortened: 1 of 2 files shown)


conradoplg:
[It failed by a few minutes](https://github.com/ZcashFoundation/zebra/runs/6547196514?check_suite_focus=true). I'm retrying it in case we got bad luck, but it seems we'll need to make it a bit faster, or change the percentage.

gustavovalverde:
The full sync completes, but the next step (`Create image from the state disk`) is also included in the 6 hours timeout. I'm thinking on how to solve this.

I was thinking on decoupling this task from the job, but then the instance will be kept alive indefinitely if the _deletion_ step is not executed. I'm seeing if we can put a timer on the instance itself to auto-delete after 6h:30m

.github/workflows/continous-integration-docker.yml
```diff
@@ -294,7 +294,7 @@ jobs:
       needs_zebra_state: false
       saves_to_disk: true
       disk_suffix: tip
-      height_grep_text: 'finished initial sync to chain tip, using gossiped blocks sync_percent=100.* current_height=Height'
+      height_grep_text: 'estimated progress to chain tip sync_percent=99.5.* current_height=Height'
```
```suggestion
      height_grep_text: 'estimated progress to chain tip sync_percent=99.2.* current_height=Height'
```

zebrad/tests/common/sync.rs
```diff
@@ -39,7 +39,10 @@ pub const STOP_AT_HEIGHT_REGEX: &str = "stopping at configured height";
 /// - we have synced all known checkpoints,
 /// - the syncer has stopped downloading lots of blocks, and
 /// - we are regularly downloading some blocks via the syncer or block gossip.
-pub const SYNC_FINISHED_REGEX: &str = "finished initial sync to chain tip, using gossiped blocks";
+///
+/// Temporary workaround for slow syncs - stop at 99.5%.
+/// TODO: revert this change (#4456)
+pub const SYNC_FINISHED_REGEX: &str = "estimated progress to chain tip sync_percent=99.5";
```
teor2345:
```suggestion
pub const SYNC_FINISHED_REGEX: &str = "estimated progress to chain tip sync_percent=99.2";
```

teor2345:
> [It failed by a few minutes](https://github.com/ZcashFoundation/zebra/runs/6547196514?check_suite_focus=true). I'm retrying it in case we got bad luck, but it seems we'll need to make it a bit faster, or change the percentage.

I changed it to 99.2% for now

.github/workflows/continous-integration-docker.yml
```diff
@@ -294,7 +294,7 @@ jobs:
       needs_zebra_state: false
       saves_to_disk: true
       disk_suffix: tip
-      height_grep_text: 'finished initial sync to chain tip, using gossiped blocks sync_percent=100.* current_height=Height'
+      height_grep_text: 'estimated progress to chain tip sync_percent=99.2.* current_height=Height'
```
```suggestion
      height_grep_text: 'estimated progress to chain tip sync_percent=99.* current_height=Height'
```

zebrad/tests/common/sync.rs
```diff
@@ -39,7 +39,10 @@ pub const STOP_AT_HEIGHT_REGEX: &str = "stopping at configured height";
 /// - we have synced all known checkpoints,
 /// - the syncer has stopped downloading lots of blocks, and
 /// - we are regularly downloading some blocks via the syncer or block gossip.
-pub const SYNC_FINISHED_REGEX: &str = "finished initial sync to chain tip, using gossiped blocks";
+///
+/// Temporary workaround for slow syncs - stop at 99.5%.
+/// TODO: revert this change (#4456)
+pub const SYNC_FINISHED_REGEX: &str = "estimated progress to chain tip sync_percent=99.2";
```
teor2345:
```suggestion
pub const SYNC_FINISHED_REGEX: &str = "estimated progress to chain tip sync_percent=99";
```

teor2345:
This works, and it gives us the temporary speed increase we need, so I'm just going to merge it to unblock other PRs.

We can revert it later.

PR #4466, opened 2022-05-24
fix(ci): Make reusable workflow job names match patch job names

teor2345:
## Motivation

To implement #4340, we need the real and patch workflow names to match.

## Solution

- Name each calling job, so we don't see job IDs in GitHub status reports
- Name each patch job after both the calling job name and the re-usable workflow job name

## Review

Anyone can review this PR.

### Reviewer Checklist

  - [ ] Real and patch job names match in GitHub status reports on PRs

## Follow Up Work

- Add these workflows to GitHub settings


gustavovalverde (approved):
LGTM!

PR #4467, opened 2022-05-24
do not merge: Try removing reusable workflow names

teor2345:
### Reviewer Checklist

  - [ ] Tests for Errors


teor2345:
This doesn't work, both the calling job and re-usable workflow job names are always included.

PR #4469, opened 2022-05-24
refactor(ci): allow more time for tests to end gracefully

gustavovalverde:
## Motivation

When running a full sync or any other test which takes almost 5 hours, having those jobs running with other actions that might take several minutes, also reduces the overall time from the job_id.

Relates to: https://github.com/ZcashFoundation/zebra/issues/4456

Depends-On: #4483, #4496, #4504

## Solution

- We use a separate job for image creation and deletion to handle this cases, as this would allow test jobs to run for a few more minutes.

## Review

@teor2345 or @dconnolly can review this


.github/workflows/deploy-gcp-tests.yml
```diff
@@ -357,6 +310,72 @@ jobs:
           --mount type=volume,src=${{ inputs.disk_prefix }}-${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }},dst=${{ inputs.root_state_path }}/${{ inputs.lwd_state_dir }} \
           ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}"
 
+  create-state-image:
+    name: Create ${{ inputs.test_id }} cached state image
```
teor2345, line 320: Question: I'd like to block merging if this job fails. Is there an easy way to do that?

(I think we might have to add a patch entry for each test that calls this re-usable workflow.)

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -375,9 +393,34 @@ jobs:
           --storage-location=us \
           --description="Created from commit ${{ env.GITHUB_SHA_SHORT }} with height ${{ env.SYNC_HEIGHT }}"
 
+  delete-instance:
+    name: Delete ${{ inputs.test_id }} instance
```
teor2345, line 403: I'd also like to block merging if this job fails, but I'm much less concerned about it, because it is a simple cleanup step.

teor2345 (commented):
This looks good, but I'm wondering how we check for failures.

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: The only scenario where the deletion would fail is if the instance was not created as a result from a failure, but we anyways run the delete job just to ensure any odd scenario is being considered.

We also don't want to fail this job as we've always done, so I've added the `continue-on-error: true` this step had before becoming a job (this also keeps the same behavior we've always had)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: We're using `needs: [ test-without-cached-state, test-with-cached-state ]` in this step now, so we shouldn't worry about this step but the ones this step _needs_.

`gcloud compute images create` only fails if the instance does not exist, which won't be the case if we depend on the instance being created and running successfully.

gustavovalverde:
@teor2345 I've had to remove a few required actions, as other PRs were stalled. As you can see in this one `Zebra tip JSON-RPC / Run fully-synced-rpc test` is waiting for the status to be reported,  and it's happening as the action was skipped and then the name is now just `Zebra tip JSON-RPC`, as the reusable workflow was never called.

teor2345:
> @teor2345 I've had to remove a few required actions, as other PRs were stalled. As you can see in this one `Zebra tip JSON-RPC / Run fully-synced-rpc test` is waiting for the status to be reported, and it's happening as the action was skipped and then the name is now just `Zebra tip JSON-RPC`, as the reusable workflow was never called.

Blocking this PR from merging is the correct behaviour, because:
- `CI Docker / Zebra tip / Run full-sync-to-tip test` failed, which should block merging
- it doesn't matter if the dependent workflows fail or wait, because this PR has already failed

teor2345:
> * `CI Docker / Zebra tip / Run full-sync-to-tip test` failed, which should block merging

I opened PR #4483 to fix this, and made this PR depend on it.

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345 (commented):
I'm happy for this to merge if CI passes.

teor2345 (approved):
> I'm happy for this to merge if CI passes.

I've turned on all the required checks in the branch protection rules, so we shouldn't need to check anything manually here.

teor2345:
@mergifyio update

teor2345:
@mergifyio upgrade

teor2345:
@mergifyio update

teor2345:
@mergifyio update

PR #4474, opened 2022-05-24
fix(ci): Require full sync to reach 100%

teor2345:
## Motivation

PRs #4457 and #4483 were a temporary fix, we want to restore full test coverage after we find another way to speed up the sync.

### Scheduling

Depends-On: #4456 

## Solution

Reverts ZcashFoundation/zebra#4457
Reverts ZcashFoundation/zebra#4483


dconnolly (dismissed):

teor2345:
We'll re-do this change as part of ticket #4661.

Issue #2112, opened 2021-05-05
Integrate tokio-console


teor2345:
**Scheduling**

This ticket is only required if we've fixed all the known hang bugs in Zebra, and Zebra still has frequent hangs.

**Is your feature request related to a problem? Please describe.**

It's hard to see what tokio is doing inside Zebra.

> another cool thing the console does now is its ⚠️warnings system. this is like a linter, but for your program’s runtime behavior. by monitoring the tasks in the application, we can detect behavior patterns that might indicate bugs or performance issues.

https://twitter.com/mycoliza/status/1471880880289366022

**Describe the solution you'd like**

- [ ] Integrate tokio-console into Zebra: https://github.com/tokio-rs/console#using-it
- [ ] Manually test that it works

**Describe alternatives you've considered**

Continue trying to debug tokio using metrics and other indirect methods

teor2345:
I've marked this as a medium priority, because we still have hang and performance bugs in Zebra that are hard to diagnose and fix.

ftm1000:
redistributing issues that can be separately worked from this epic #2311

teor2345:
We don't need to do this unless there are some performance or hang issues we need to fix.

teor2345:
Re-opening this ticket, because we might want to do it this sprint.

PR #4475, opened 2022-05-24
change(deps): Turn on tokio's internal parking_lot feature

teor2345:
## Motivation

Using the `parking_lot` concurrency crate could make Zebra faster.

It is also recommended as part of #2112:
https://github.com/tokio-rs/console/blob/main/console-subscriber/README.md#crate-feature-flags

## Solution

Enable the `parking_lot` feature in:
- tokio
- tracing-subscriber

This should be compatible with Zebra's existing locking.

## Review

Anyone can review this PR, but we only want to merge it if it's faster than the alternative.

### Reviewer Checklist

  - [ ] Existing tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5611,6 +5611,7 @@ dependencies = [
  "chrono",
  "lazy_static",
  "matchers",
+ "parking_lot 0.11.2",
  "regex",
  "serde",
  "serde_json",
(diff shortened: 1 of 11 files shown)


teor2345:
I started a manual full sync run here:
https://github.com/ZcashFoundation/zebra/actions/runs/2376561875

Currently the time to beat is around 5h50m to 6h.

teor2345:
This is actually slower (5h53m vs 5h49m), so we won't do it.

PR #4392, opened 2022-05-13
3. Require network names in cached state disk names

teor2345:
## Motivation

As part of ticket #4153, we want to match cached states with the correct version and network.

In PR #4391, we fixed a bug where cached state disks were missing the network name.
Now that PR has merged to `main`, and its state is rebuilt, we can require the network in disk names.

Depends-On: #4385

## Solution

Search for the exact network name in cached state disk names.

## Review

Anyone can review this PR, it's not urgent.

### Reviewer Checklist

  - [x] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #4477, opened 2022-05-24
fix(ci): set a network env variable to be used in `get-available-disks`

gustavovalverde:
## Motivation

The blockchain Network name is being used in the `get-available-disks` job, but this variable was not being passed to this job.

When merging https://github.com/ZcashFoundation/zebra/pull/4392 we didn't considered this environment variable was no longer available in this workflow

## Solution

- Add `Network` as a global env variable
- Downcase the `Network` to be used by disk image search (as images are in lowercase)

## Review

Anyone from @ZcashFoundation/devops-reviewers, but this is urgent as a bunch of VMs are being triggered and most jobs are not finding the appropriate image, and I might admin merge it if it runs successfully 

## Follow Up Work

Check is this was also causing the cached state issues


conradoplg (approved):
Looks good

PR #4482, opened 2022-05-24
feat(ci): add a codespell linting action

### LINKED ISSUE #3318: Consider adding codespell to the CI
## Motivation

It seems @rex4539 is a regular contributor who fixes our typos by running [codespell](https://github.com/codespell-project/codespell) over the codebase. As @oxarbitrage  mentioned, it might be possible to integrate a spell-checker like [codespell](https://github.com/codespell-project/codespell) into our CI.

### Specifications

Here's a previous discussion about spell-checking with @teor2345:

> The tricky part of this change is adding a dictionary with Zcash-specific words, Rust-specific words, and engineering-specific words. Otherwise the spelling CI might fail every time we add a new feature or design.

> > Codespell looks for common (and a lot of) misspellings instead of checking dictionary membership. Here's the relevant description: It does not check for word membership in a complete dictionary, but instead looks for a set of common misspellings. Therefore it should catch errors like "adn", but it will not catch "adnasdfasdf". This also means it shouldn't generate false-positives when you use a niche term it doesn't know about.

> > Here are the misspellings https://github.com/codespell-project/codespell/tree/master/codespell_lib/data.
We could even add ours if we find a common one.

gustavovalverde:
## Motivation

Codespell looks for common (and a lot of) misspellings instead of checking dictionary membership. It does not check for word membership in a complete dictionary, but instead looks for a set of common misspellings. Therefore it should catch errors like "adn", but it will not catch "adnasdfasdf". This also means it shouldn't generate false-positives when you use a niche term it doesn't know about.

Fixes #3318

## Solution

- Add a codespell action to to the `Lint` workflow and activate it when a rust file is changed

## Review

@upbqdn or anyone from @ZcashFoundation/devops-reviewers 

### Reviewer Checklist

  - [ ] Codespell is not generating false-positives
  - [ ] Tests for Errors

## Follow Up Work

- Validate if we can fix the amount of errors now
- If not, add a `continue-on-error` on the job

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,2 +1,3 @@
 [codespell]
 ignore-words-list=crate
+exclude-file=mermaid.min.js
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,2 @@
+[codespell]
+ignore-words-list=crate
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
-ignore-words-list=crate
+ignore-words-list=crate,Sur
 exclude-file=book/mermaid.min.js
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 ignore-words-list=crate
-exclude-file=mermaid.min.js
+exclude-file=book/mermaid.min.js
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,6 +33,7 @@ jobs:
             **/Cargo.lock
             clippy.toml
             .cargo/config.toml
+            .github/workflows/lint.yml
 
       - name: Workflow files
         id: changed-files-workflows
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -122,7 +122,6 @@ jobs:
 
   codespell:
     runs-on: ubuntu-latest
-    if: ${{ needs.changed-files.outputs.rust == 'true' }}
     needs: changed-files
     steps:
       - uses: actions/checkout@v3.0.2
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -129,3 +129,4 @@ jobs:
       - uses: plettich/action-codespell@master
         with:
           github_token: ${{ secrets.github_token }}
+          level: warning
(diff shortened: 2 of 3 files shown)


teor2345 (changes requested):
I'm not sure if we actually want this check in CI.

Here are the benefits I see:
- Documentation is spelt correctly

Here are the drawbacks I see:
- We block merges for a single misspelling, even if the meaning is clear, or
- We get red check marks on every PR if we accidentally merge a misspelling

This seems like a negative experience, particularly for new contributors.

What does everyone think?

gustavovalverde:
> Here are the drawbacks I see:
> 
> * We block merges for a single misspelling, even if the meaning is clear, or
> * We get red check marks on every PR if we accidentally merge a misspelling

- Merges won't get block using `warning` as the maximum level
- This only validates the current changes in the PR so we should not worry on existing misspellings as those will be filtered and won't raise a warning
  - Example: https://github.com/ZcashFoundation/zebra/runs/6583671800

teor2345:
> > Here are the drawbacks I see:
> > 
> > * We block merges for a single misspelling, even if the meaning is clear, or
> > * We get red check marks on every PR if we accidentally merge a misspelling
> 
> * Merges won't get block using `warning` as the maximum level
> * This only validates the current changes in the PR so we should not worry on existing misspellings as those will be filtered and won't raise a warning      
>   * Example: https://github.com/ZcashFoundation/zebra/runs/6583671800

Can you show me an example of what a misspelling in a PR's changes looks like?

upbqdn:
I personally think the benefits of this check outweigh the drawbacks. I think this because of the reasons gustavovalverde stated: PRs don't get blocked and only edited files are checked.
I agree that it can be a negative experience, but not to the extend that it would discourage people from contributing (I'm usually happy when a machine checks my work for free, and gives me suggestions). The biggest drawback I see is the maintenance of `ignore-words-list`.

I don't think it's complicated to deprecate this check, so if we start using it, and we don't like it, we can easily get rid of it.

gustavovalverde:
Result from the linter. I'm reverting this test commit
image

conradoplg (commented):
This looks good, but I don't get this: where is it specified that it should run only on changed files? Or is that the default behavior?

teor2345:
> This looks good, but I don't get this: where is it specified that it should run only on changed files? Or is that the default behavior?

Review Dog does this kind of filtering automatically.

teor2345 (approved):
Looks great, thanks for the explanation!

PR #4486, opened 2022-05-25
fix(ci): mount the lwd-cache dir to the `lightwalletd-full-sync`

### LINKED ISSUE #4415: Fix lightwalletd cached state disks in CI
## Motivation

Currently, Zebra's CI mounts a single disk to two docker paths for the Zebra and lightwalletd state.

But this is unreliable, it is currently causing the following tests to fail:
* lightwalletd update sync
* lightwalletd send transactions

### Designs

One possible fix is:
- use one disk for Zebra cached state
- use a separate disk for lightwalletd cached state
- mount the required disks in each docker container

gustavovalverde:
## Motivation

Tests which needs a lightwalletd cached state have been failing for a while, this issue was introduced in a refactor where we wrongly stoped mounting the `lwd_state_dir` to the `lightwalletd-full-sync` test.

Fixes #4415 

### Designs

We need a condition that consider this type of scenarios, in the meanwhile we're adding this test id as an exception, until another refactor is done in this file.

## Solution

- Add `lightwalletd-full-sync` as an exception when just a zebra cached state is needed and run it when a lwd cached state is needed, even if this specific test does not need one, as we need to mount the path to save the content.
- Run this test just if a zebra full sync is run or if there's no lwd cache available. As a result this will always run in `main`, when a state upgrade happens and if a manual full sync is triggered.

## Review

@teor2345 can review this, or anyone from @ZcashFoundation/devops-reviewers if every test runs as expected

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

- Improve this conditions

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -300,7 +300,7 @@ jobs:
         # lightwalletd-full-sync test is an exception to this rule, as it does not need a lwd cached state,
         # but it does saves a lwd cached state
         # TODO: we should find a better logic for this use cases
-        if: ${{ inputs.needs_zebra_state && !inputs.needs_lwd_state && inputs.test_id != "lightwalletd-full-sync" }}
+        if: ${{ inputs.needs_zebra_state && !inputs.needs_lwd_state && inputs.test_id != 'lightwalletd-full-sync' }}
         run: |
           gcloud compute ssh \
           ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
@@ -346,7 +346,7 @@ jobs:
         # lightwalletd-full-sync test is an exception to this rule, as it does not need a lwd cached state,
         # but it does saves a lwd cached state
         # TODO: we should find a better logic for this use cases
-        if: ${{ inputs.needs_zebra_state && inputs.needs_lwd_state || inputs.test_id == "lightwalletd-full-sync" }}
+        if: ${{ inputs.needs_zebra_state && inputs.needs_lwd_state || inputs.test_id == 'lightwalletd-full-sync' }}
         run: |
           gcloud compute ssh \
           ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-gcp-tests.yml
```diff
@@ -297,7 +297,10 @@ jobs:
       - name: Run ${{ inputs.test_id }} test
         # This step mounts the volume only when a single cached state is needed, in this case
         # the cached state from Zebra.
-        if: ${{ inputs.needs_zebra_state && !inputs.needs_lwd_state }}
+        # lightwalletd-full-sync test is an exception to this rule, as it does not need a lwd cached state,
+        # but it does saves a lwd cached state
+        # TODO: we should find a better logic for this use cases
+        if: ${{ inputs.needs_zebra_state && !inputs.needs_lwd_state && inputs.test_id != lightwalletd-full-sync }}
```
teor2345: Nice catch!

But I think the syntax needs a tweak:
```suggestion
        # lightwalletd-full-sync test is an exception to this rule, as it does not need a lwd cached state,
        # but it does saves a lwd cached state
        # TODO: we should find a better logic for this use cases
        if: ${{ inputs.needs_zebra_state && !inputs.needs_lwd_state && inputs.test_id != "lightwalletd-full-sync" }}
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -340,7 +343,10 @@ jobs:
       - name: Run ${{ inputs.test_id }} test
         # This step mounts the volume only when both cached states are needed, in this case
         # the cached state from Zebra and Lightwalletd
-        if: ${{ inputs.needs_zebra_state && inputs.needs_lwd_state }}
+        # lightwalletd-full-sync test is an exception to this rule, as it does not need a lwd cached state,
+        # but it does saves a lwd cached state
+        # TODO: we should find a better logic for this use cases
+        if: ${{ inputs.needs_zebra_state && inputs.needs_lwd_state || inputs.test_id == lightwalletd-full-sync }}
```
```suggestion
        if: ${{ inputs.needs_zebra_state && inputs.needs_lwd_state || inputs.test_id == "lightwalletd-full-sync" }}
```

teor2345 (commented):
I think there's a minor syntax error here

gustavovalverde:
Admin merging as further work is halted, and some PRs are also halted. The global behavior was reviewed, and any additional changes can be added in a future PR

PR #4416, opened 2022-05-18
fix(ci): stop failing the send transaction test

teor2345:
## Motivation

In PR #4303, we required cached lightwalletd state for the send transaction test.
But that state is currently unreliable.

Depends-On: #4411

## Solution

- revert to the previous behaviour, were cached state is allowed, but not required

## Review

Anyone can review this PR, it's a high priority because it fixes a CI failure.

We expect the update lightwalletd sync test to fail, that's ticket #4415.

### Reviewer Checklist

  - [ ] send transaction test passes CI

## Follow Up Tasks

Revert this PR once #4415 is implemented.

PR #4487, opened 2022-05-25
fix(ci): Require cached state for the send transactions test

teor2345:
## Motivation

We want the send transactions test to always be fast.

Depends-On: #4486

## Solution

Reverts ZcashFoundation/zebra#4416


gustavovalverde (approved):

gustavovalverde:
@Mergifyio update

PR #4494, opened 2022-05-25
fix(doc): Fix the syntax of links in comments

upbqdn:
## Motivation

Running `cargo doc` currently produces a lot of warnings. Many of these warnings are due to an incorrect use of links in comments.

To avoid merge conflicts:
Depends-On: #4453

## Solution

This PR adds `<` and `>` around each link where needed. It reduces the number of warnings from 479 to 345.

This PR will make the refactoring of our docs easier since the compilation won't be so noisy.


upbqdn:
I can fix the rest of the warnings produced by `cargo doc`, but I think it would be much easier to review it as a separate PR that would follow after this is merged.

`cargo doc` already produces much less warnings with the fixed links.

zebra-chain/src/orchard/sinsemilla.rs
```diff
@@ -191,8 +191,8 @@ pub fn sinsemilla_short_commit(
 }
 
 // TODO: test the above correctness and compatibility with the zcash-hackworks test vectors
-// https://github.com/ZcashFoundation/zebra/issues/2079
-// https://github.com/zcash-hackworks/zcash-test-vectors/pulls
+// <https://github.com/ZcashFoundation/zebra/issues/2079>
```
oxarbitrage: Can you confirm that the link in docs that start with `//` like this one instead of `///` or `//!` also produce warnings ?

upbqdn: Comments starting with `//` don't produce warnings, but I think it's useful to have them in the correct format in case we add a `/` when we want to propagate them to the docs.

zebra-chain/src/sapling/keys.rs
```diff
@@ -5,8 +5,8 @@
 //! diversified payment address addr_d = (d, pk_d ) are derived from sk,
 //! as described in [Sapling Key Components][ps]." - [§3.1][3.1]
 //!
-//! [ps]: https://zips.z.cash/protocol/protocol.pdf#saplingkeycomponents
-//! [3.1]: https://zips.z.cash/protocol/protocol.pdf#addressesandkeys
+//! [ps]: <https://zips.z.cash/protocol/protocol.pdf#saplingkeycomponents>
+//! [3.1]: <https://zips.z.cash/protocol/protocol.pdf#addressesandkeys>
```
oxarbitrage: Can you also confirm if this ones also produces warnings. I think the `<` and `>` purpose is to make the links clickeable however this ones are not displayed as `http://`.

Check for example https://doc.zebra.zfnd.org/zebra_chain/sapling/keys/index.html

Thanks.

upbqdn: This is a good point. That doesn't produce a warning. However, it is not wrong to have the angle brackets there either. See here https://www.markdownguide.org/basic-syntax/#formatting-the-second-part-of-the-link. Angle brackets are optional in this case. It think it might be good to use them here as well for consistency.

teor2345 (commented):
Is there an automated script we can run to update or check this PR?

It's going to conflict with some other PRs we're working on.

upbqdn:
> Is there an automated script we can run to update or check this PR?

I'm doing it semi-automatically with regexes in Emacs. I think the easiest way would be to wait until those PRs are merged, then I can update this PR.

zebrad/src/commands/generate.rs
```diff
@@ -30,7 +30,7 @@ impl Runnable for GenerateCmd {
 #
 # The config format (including a complete list of sections and fields) is
 # documented here:
-# https://doc.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html
+# <https://doc.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html>
```
oxarbitrage: We probably don't want to do this one as it is not a rustdoc comment.

teor2345:
> > Is there an automated script we can run to update or check this PR?
> 
> I'm doing it semi-automatically with regexes in Emacs. I think the easiest way would be to wait until those PRs are merged, then I can update this PR.

The only PR I can see this conflicting with is PR #4453, so let's update and merge after it merges.

zebra-chain/src/transaction/tests/vectors.rs
```diff
@@ -145,7 +145,7 @@ fn doesnt_deserialize_transaction_with_invalid_value_balance() {
     // Set value balance to non-zero
     // There are 4 * 4 byte fields and 2 * 1 byte compact sizes = 18 bytes before the 8 byte amount
     // (Zcash is little-endian unless otherwise specified:
-    // https://zips.z.cash/protocol/nu5.pdf#endian)
+    // <https://zips.z.cash/protocol/nu5.pdf#endian)>
```
oxarbitrage:
```suggestion
    // <https://zips.z.cash/protocol/nu5.pdf#endian>)
```

zebra-chain/src/orchard/sinsemilla.rs
oxarbitrage: Yea, well, i guess is the same, i was just thinking that from now on we should block any PR that uses a link without `<` and `>` in any comment and not just in rustdoc comments. Should we write this somewhere as a contributor guideline ?

upbqdn: I also think it would be a good practice to use `<` `>` with every link. We can even embed `cargo doc`s output to our CI to check for any warnings once I clear all the legacy ones.

zebra-chain/src/sapling/keys.rs
oxarbitrage: I don't have a strong opinion but i slightly prefer to don't add the extra characters if they are not needed. I see the extra characters as a way to make the link clickable and not that much as the correct format. In this context making them clickable don't have much sense as they are not visible.

zebra-chain/src/transaction/tests/vectors.rs
upbqdn: Thanks for checking them all! This one slipped my attention.

zebra-chain/src/sapling/keys.rs
upbqdn: Resolved in [8b091aa](https://github.com/ZcashFoundation/zebra/pull/4494/commits/8b091aa9fab453e7d3559a5d474e0879183b9bfb).

upbqdn: Actually, the references also require the angle brackets because `cargo doc` does produce warnings if the angle brackets are missing. I'm going to revert the reversion.

dconnolly:
Have you checked how these changes render?

upbqdn:
> Have you checked how these changes render?

They render well. I checked around 5 specimens. The links became clickable in the docs.

tower-fallback/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 //!
 //! TODO: compare with similar code in linkerd.
 //!
-//! [aws-fallback]: https://aws.amazon.com/builders-library/avoiding-fallback-in-distributed-systems/
+//! [aws-fallback]: <https://aws.amazon.com/builders-library/avoiding-fallback-in-distributed-systems/>
```
dconnolly: These are interesting because they should just link in the reference citing them and not render the full url itself, is rustdoc complaining about this?

zebra-chain/src/orchard/sinsemilla.rs
oxarbitrage: I am more in the side of don't make the change or force users to do it certain way if it is not needed but i am ok either way. @dconnolly do you have an opinion?

tower-fallback/src/lib.rs
upbqdn: I did more investigation, and rustdoc actually doesn't complain here. This is what confused me:

image

When I saw this error, I thought such links have to be enclosed in `<` `>` as well. Here's the code: 
```rust
    /// Check if the `network_upgrade` fields from each transaction in the block matches
    /// the network upgrade calculated from the `network` and block height.
    ///
    /// # Consensus
    ///
    /// > [NU5 onward] The nConsensusBranchId field MUST match the consensus branch ID used
    /// > for SIGHASH transaction hashes, as specified in [ZIP-244].
    ///
    /// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
    /// [ZIP-244]: https://zips.z.cash/zip-0244
```
The problem was that there wasn't a newline between the two links at the bottom. 

I'm going to revert the reversion of the reversion.

zebra-chain/src/sapling/keys.rs
oxarbitrage: Hmm, so it seems the lint checks every link inside rustdoc comment whatever is visible or not. I think this might be a bug in the lint or they will have reasons. Anyway, if it warns we should do the change. Thanks.

zebra-chain/src/orchard/sinsemilla.rs
oxarbitrage: I was doing a quick check around and just found that the `cargo` rust project does not use the `<link>` format in normal comments: 

- https://github.com/rust-lang/cargo/blob/master/src/cargo/core/features.rs#L134
- https://github.com/rust-lang/cargo/blob/master/src/cargo/core/profiles.rs#L274

But it definitly do so in rustdocs:

- https://github.com/rust-lang/cargo/blob/master/src/cargo/core/features.rs#L22

zebra-chain/src/sapling/keys.rs
upbqdn: See https://github.com/ZcashFoundation/zebra/pull/4494#discussion_r882650766. `cargo doc` doesn't actually produce warnings in this case.

oxarbitrage: Thank you, that makes much more sense to me.

tower-fallback/src/lib.rs
upbqdn: Done in [a26e079](https://github.com/ZcashFoundation/zebra/pull/4494/commits/a26e0797b7fe5fd3ba2fbb50f4550daed654d918).

conradoplg (dismissed):

zebra-chain/src/orchard/sinsemilla.rs
conradoplg: @oxarbitrage do you think we should remove the format from normal comments, or can we mark this as solved? I don't mind either way. It seems easier to merge this as is, and when we add the check to CI the normal comments won't be enforced anyway, so it won't block if people use the format in those.

oxarbitrage: I think we should not use this format in normal comments so that change on my opinion needs to be reversed. 

I am also unsure on how is the workflow between this PR and https://github.com/ZcashFoundation/zebra/pull/4514

upbqdn: > I think we should not use this format in normal comments so that change on my opinion needs to be reversed.

OK, I will reverse it.

> I am also unsure on how is the workflow between this PR and https://github.com/ZcashFoundation/zebra/pull/4514

I thought I'd rebase that PR once this is merged. That PR will be much easier to review then.

upbqdn: While reversing the normal comments, I noticed we use Markdown syntax for other parts of the comments, even though the comments are not rendered in the docs. A part of the Markdown syntax is to use `<` and `>` for links: https://www.markdownguide.org/basic-syntax/#urls-and-email-addresses. Don't we create an inconsistency if we don't use the angle brackets?

oxarbitrage: I dont think we do because normal comments don't have to be markdown at all.

upbqdn: Done in [f567869](https://github.com/ZcashFoundation/zebra/pull/4494/commits/f567869478277cb213d8462a327d2ecdc3ee7d39).

zebra-chain/src/orchard/keys.rs
```diff
@@ -982,7 +982,7 @@ impl PartialEq<[u8; 32]> for TransmissionKey {
 /// An _outgoing cipher key_ for Orchard note encryption/decryption.
 ///
 /// <https://zips.z.cash/protocol/nu5.pdf#saplingandorchardencrypt>
-// TODO: derive `OutgoingCipherKey`: https://github.com/ZcashFoundation/zebra/issues/2041
+// TODO: derive `OutgoingCipherKey`: <https://github.com/ZcashFoundation/zebra/issues/2041>
```
oxarbitrage: Optional: Not needed.
```suggestion
// TODO: derive `OutgoingCipherKey`: https://github.com/ZcashFoundation/zebra/issues/2041
```

zebra-chain/src/orchard/note/ciphertexts.rs
```diff
@@ -11,7 +11,7 @@ use crate::serialization::{SerializationError, ZcashDeserialize, ZcashSerialize}
 pub struct EncryptedNote(#[serde(with = "BigArray")] pub(crate) [u8; 580]);
 
 // These impls all only exist because of array length restrictions.
-// TODO: use const generics https://github.com/ZcashFoundation/zebra/issues/2042
+// TODO: use const generics <https://github.com/ZcashFoundation/zebra/issues/2042>
```
oxarbitrage: Optional: not needed.
```suggestion
// TODO: use const generics https://github.com/ZcashFoundation/zebra/issues/2042
```

zebra-chain/src/orchard/tests/vectors/group_hash.rs
```diff
@@ -6,7 +6,7 @@ pub struct TestVector {
     pub(crate) point: [u8; 32],
 }
 
-// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_group_hash.py
+// From <https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_group_hash.py>
```
oxarbitrage: Optional: not needed.
```suggestion
// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_group_hash.py
```

zebra-chain/src/orchard/tests/vectors/sinsemilla.rs
```diff
@@ -7,7 +7,7 @@ pub struct TestVector {
     pub(crate) hash: [u8; 32],
 }
 
-// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_sinsemilla.py
+// From <https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_sinsemilla.py>
```
oxarbitrage: Optional: not needed.,
```suggestion
// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_sinsemilla.py
```

oxarbitrage:
It seems there are still a lot of normal comments that were changed to the format. I listed a few but there are more as i go lower in the review. I think the intentio nwas to reverse all those in https://github.com/ZcashFoundation/zebra/pull/4494/commits/f567869478277cb213d8462a327d2ecdc3ee7d39 but it seems not all of them were catched.

oxarbitrage (changes requested):
Blocking until we reverse all the format changes to the normal (`//`) comments.

zebra-chain/src/orchard/tests/vectors/sinsemilla.rs
upbqdn: I fixed this one in [36fc845](https://github.com/ZcashFoundation/zebra/pull/4494/commits/36fc8453d4239291deb6e3f692a0d6b9253f17e8).

zebra-chain/src/orchard/tests/vectors/group_hash.rs
upbqdn: I fixed this one in [36fc845](https://github.com/ZcashFoundation/zebra/pull/4494/commits/36fc8453d4239291deb6e3f692a0d6b9253f17e8).

upbqdn:
> It seems there are still a lot of normal comments that were changed to the format. I listed a few but there are more as i go lower in the review. I think the intentio nwas to reverse all those in https://github.com/ZcashFoundation/zebra/commit/f567869478277cb213d8462a327d2ecdc3ee7d39 but it seems not all of them were watched.

Thanks so much for noticing this. These comments were at the beginnings of lines, and my regex didn't catch them. It's fixed in [36fc845](https://github.com/ZcashFoundation/zebra/pull/4494/commits/36fc8453d4239291deb6e3f692a0d6b9253f17e8) now.

zebra-chain/src/orchard/note/ciphertexts.rs
upbqdn: Resolved in [36fc845](https://github.com/ZcashFoundation/zebra/pull/4494/commits/36fc8453d4239291deb6e3f692a0d6b9253f17e8).

zebra-chain/src/orchard/keys.rs
upbqdn: Resolved in [36fc845](https://github.com/ZcashFoundation/zebra/pull/4494/commits/36fc8453d4239291deb6e3f692a0d6b9253f17e8).

zebra-state/src/service/check/anchors.rs
```diff
@@ -146,7 +146,7 @@ pub(crate) fn anchors_refer_to_earlier_treestates(
         // > (i.e. when there are any Action descriptions), MUST refer to some
         // > earlier block’s final Orchard treestate.
         //
-        // <https://zips.z.cash/protocol/protocol.pdf#actions>
+        // https://zips.z.cash/protocol/protocol.pdf#actions
```
oxarbitrage: Minor/Optional: We don't need to change the links in normal comments that use `<`, `>`. The idea of normal comments is that the person writing it can do whatever they want. In this case, this are consensus comments so maybe we should keep them untouched ?

oxarbitrage: I see the writer made markdown in this kind of consensus comments. Even if they are not rendered now because they are normal comments, they could be visible in the future.

upbqdn: I think the easiest way would be to close this PR without merging, and open a new one that edits only comments starting with `///` or `//!`?

conradoplg: I think we can just merge as is. The reason why brackets were added in those comments was because we thought we would want to add them in all instances, in the future. Since we decided to not make them mandatory in regular comments, this seems fine now

oxarbitrage: Yea, well, i don't think is very important but here is my reasoning about it. I think is ok to keep the formatted links here because this specific comments already uses `> Quote` for quoting text from the spec. This reveals markdown intention. If the text is markdown then it is suggested to use https://www.markdownguide.org/basic-syntax/#urls-and-email-addresses

oxarbitrage (dismissed):
There is some open discussion but you can ignore it.

I am ok if any of this are done:
- merge as is.
- revert the changes discussed in https://github.com/ZcashFoundation/zebra/pull/4494#discussion_r884951735 in this PR
- start a new PR that edits only comments starting with `///` or `//!`

zebra-state/src/service/check/anchors.rs
upbqdn: I reverted some changes again in [398bebb](https://github.com/ZcashFoundation/zebra/pull/4494/commits/398bebb8b6a0b3a1682668a37292629a33fbd58d).

oxarbitrage (approved):

zebra-state/src/service/check/anchors.rs
oxarbitrage: Thanks!

PR #4501, opened 2022-05-25
change(sync): Activate the mempool after 2 syncer runs at the chain tip, rather than 4

teor2345:
## Motivation

We want to speed up the full sync, as part of #4456.

Activating the mempool sooner also helps with usability and wallet performance. Users might be confused or frustrated if it takes 4-6 minutes for the mempool to activate after reaching the tip.

## Solution

- Average recently synced blocks over 2 syncer runs, rather than 4

This makes spurious mempool activation slightly more likely, but the consistent speed increase is worth it.

## Review

Anyone can review this PR, it's urgent because slow syncs block other work.

### Reviewer Checklist

  - [ ] Existing tests pass

### DIFF
--- a/zebrad/src/components/sync/recent_sync_lengths.rs
+++ b/zebrad/src/components/sync/recent_sync_lengths.rs
@@ -31,7 +31,7 @@ impl RecentSyncLengths {
     /// * clearing temporary errors and temporary syncs quickly
     /// * distinguishing between temporary and sustained syncs/errors
     /// * activating the syncer shortly after reaching the chain tip
-    pub const MAX_RECENT_LENGTHS: usize = 4;
+    pub const MAX_RECENT_LENGTHS: usize = 2;
 
     /// Create a new instance of [`RecentSyncLengths`]
     /// and a [`watch::Receiver`] endpoint for receiving recent sync lengths.


teor2345:
@mergifyio update

conradoplg (approved):
Looks good

upbqdn:
The reason we stuck with 4 syncer runs before was to prevent activating the mempool too early because we sometimes got a sync length close to zero even though we were far away from the tip. If this happens twice in a row, Zebra will now activate it's mempool too early. Back then, we estimated that the probability of getting four such lengths in a row was sufficiently low. I don't know if the occurrence of almost-zero lengths lowered during syncing. If it didn't, I think we should test this PR. CC @conradoplg @teor2345.

teor2345:
> The reason we stuck with 4 syncer runs before was to prevent activating the mempool too early because we sometimes got a sync length close to zero even though we were far away from the tip. If this happens twice in a row, Zebra will now activate it's mempool too early. Back then, we estimated that the probability of getting four such lengths in a row was sufficiently low. I don't know if the occurrence of almost-zero lengths lowered during syncing. If it didn't, I think we should test this PR. CC @conradoplg @teor2345.

I tested it locally, and it didn't seem to make much of a difference.

There is a full sync running here, I'll check to see if it has any speed issues, and I'll also check for spurious activations:
https://github.com/ZcashFoundation/zebra/runs/6615351376?check_suite_focus=true

PR #4504, opened 2022-05-26
fix(ci): Permanently fix unreliable sync finished log regex

teor2345:
## Motivation

The full sync to tip regex is unreliable, and causing some tests to sometimes fail.

## Solution

Use a regex that Zebra is guaranteed to log eventually.

Unlike the other recent changes, this is a permanent fix.

## Review

This is an urgent fix to stop CI and merges sometimes failing.

### Reviewer Checklist

  - [ ] CI passes


teor2345:
The tests passed, and this is a trivial fix but a blocker for a lot of PRs. So I admin-merged it.

PR #4514, opened 2022-05-26
fix(doc): Fix various doc warnings, part 1

upbqdn:
## Motivation

`cargo doc` currently produces a lot of warnings. We would like to incorporate `cargo doc` into our CI, so we should fix these warnings.

Previous PR:
- #4494

## Solution

I manually fixed various issues that were produced by `cargo doc --document-private-items`.

This PR doesn't fix all warnings yet because it would be too cumbersome to review it. There will be upcoming PRs that will deal with the remaining warnings.


conradoplg (approved):
Looks good, thanks!

upbqdn:
@mergifyio refresh

PR #4518, opened 2022-05-27
fix(ci): Add lightwalletd gRPC, clippy, rustfmt patch jobs

teor2345:
## Motivation

We forgot to add a patch job for the:
- lightwalletd GRPC tests
- clippy lint
- rustfmt lint

Some PRs will show as pending until this merges.

We're also running some patch jobs when we don't need to.

## Review

I might just admin-merge this.


teor2345:
I admin-merged because this CI bug is blocking other PRs from merging.

PR #4523, opened 2022-05-27
doc(tests): full sync and lightwalletd tests

### LINKED ISSUE #4363: Summarise the environmental variables used to activate `zebrad` acceptance tests
## Motivation

Some `zebrad` acceptance tests need special state to run, they get skipped when it is not available.

We want to document:
- the state requirements for each group of tests (lightwalletd binaries, zebra or lightwalletd cached state)
- the environmental variables (or Rust features) used to run each test
- the environmental variables and command line flags used to see test messages (`RUST_LOG=info cargo test -- --nocapture`)
- the required lightwalletd binary - https://github.com/adityapk00/lightwalletd

We can add these docs to the top of `acceptance.rs`.

### Designs

If the environmental variables aren't set, most tests skips themselves, and logs a message.

Some tests use a Rust feature instead, but we're trying to move away from that.

oxarbitrage:
## Motivation

Running the full sync and lightwalletd acceptance tests is hard by the different configuration options you need to have available in the operationg system to run each test. We want to document the process making it easier for anyone to be able to run this tests.

Close https://github.com/ZcashFoundation/zebra/issues/4363

## Solution

I think there is room for normalization in regards to the tests but this PR just document the process as it is right now without making changes to how the different tests are run.

## Review

Anyone can review.

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

Unfortunately, currently this new documentation will not be visible anywhere outside the source code file itself.


oxarbitrage:
Acceptance test: runs zebrad as a subprocess and asserts its
output for given argument combinations matches what is expected.

## Note on port conflict

If the test child has a cache or port conflict with another test, or a
running zebrad or zcashd, then it will panic. But the acceptance tests
expect it to run until it is killed.

If these conflicts cause test failures:
  - run the tests in an isolated environment,
  - run zebrad on a custom cache path and port,
  - run zcashd on a custom port.

## Failures due to Configured Network Interfaces or Network Connectivity

If your test environment does not have any IPv6 interfaces configured, skip IPv6 tests
by setting the `ZEBRA_SKIP_IPV6_TESTS` environmental variable.

If it does not have any IPv4 interfaces, IPv4 localhost is not on `127.0.0.1`,
or you have poor network connectivity,
skip all the network tests by setting the `ZEBRA_SKIP_NETWORK_TESTS` environmental variable.

## Large/full sync tests

This file has sync tests that are marked as ignored because they take too much time to run.
Some of them require environment variables or directories to be present:

- `FULL_SYNC_MAINNET_TIMEOUT_MINUTES` env variable: The total number of minutes we
will allow this test to run or give up. Value for the Mainnet full sync tests.
- `FULL_SYNC_TESTNET_TIMEOUT_MINUTES` env variable: The total number of minutes we
will allow this test to run or give up. Value for the Testnet ful  sync tests.
- `/zebrad-cache` directory: For some sync tests, this needs to be created in
the file system, the created directory should have write permissions.

Here are some examples on how to run each of the tests:

```console
$ cargo test sync_large_checkpoints_mainnet -- --ignored --nocapture

$ cargo test sync_large_checkpoints_mempool_mainnet -- --ignored --nocapture

$ sudo mkdir /zebrad-cache
$ sudo chmod 777 /zebrad-cache
$ export FULL_SYNC_MAINNET_TIMEOUT_MINUTES=600
$ cargo test full_sync_mainnet -- --ignored --nocapture

$ sudo mkdir /zebrad-cache
$ sudo chmod 777 /zebrad-cache
$ export FULL_SYNC_TESTNET_TIMEOUT_MINUTES=600
$ cargo test full_sync_testnet -- --ignored --nocapture
```

Please refer to the documentation of each test for more information.

## Lightwalletd tests

The lightwalletd software is an interface service that uses zebrad or zcashd RPC methods to serve wallets or other applications with blockchain content in an efficient manner.
There are several versions of lightwalled in the form of different forks. The original
repo is <https://github.com/zcash/lightwalletd> but Zecwallet Lite uses a custom fork: <https://github.com/adityapk00/lightwalletd>.
The custom fork from adityapk00 is the one Zebra use for this tests:

Zebra lightwalletd tests are not all marked as ignored but none will run unless
at least the `ZEBRA_TEST_LIGHTWALLETD` environment variable is present:

- `ZEBRA_TEST_LIGHTWALLETD` env variable: Needs to be present to run any of the lightwalletd tests.
- `ZEBRA_CACHED_STATE_DIR` env var: The path to a zebra blockchain database.
- `LIGHTWALLETD_DATA_DIR` env variable. The path to a lightwalletd database.
- `--features lightwalletd-grpc-tests` cargo flag. The flag given to cargo to build the source code of the running test.

Here are some examples of running each test:

```console
$ export ZEBRA_TEST_LIGHTWALLETD=true
$ cargo test lightwalletd_integration -- --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/chain"
$ export LIGHTWALLETD_DATA_DIR="/path/to/lightwalletd/database"
$ cargo test lightwalletd_update_sync -- --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/chain"
$ cargo test lightwalletd_full_sync -- --ignored --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ cargo test lightwalletd_test_suite -- --ignored --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/chain"
$ cargo test fully_synced_rpc_test -- --ignored --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/chain"
$ export LIGHTWALLETD_DATA_DIR="/path/to/lightwalletd/database"
$ cargo test sending_transactions_using_lightwalletd --features lightwalletd-grpc-tests -- --ignored --nocapture

$ export ZEBRA_TEST_LIGHTWALLETD=true
$ export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/chain"
$ export LIGHTWALLETD_DATA_DIR="/path/to/lightwalletd/database"
$ cargo test lightwalletd_wallet_grpc_tests --features lightwalletd-grpc-tests -- --ignored --nocapture
```

Please refer to the documentation of each test for more information.

zebrad/tests/acceptance.rs
```diff
@@ -20,7 +20,91 @@
...
+//!
+//! ### Notes
+//!
+//! - `ZEBRA_TEST_LIGHTWALLETD` env variable: Needs to be present to run any of the lightwalletd tests.
+//! - `ZEBRA_CACHED_STATE_DIR` env var: The path to a zebra blockchain database.
+//! - `LIGHTWALLETD_DATA_DIR` env variable. The path to a lightwalletd database.
+//! - `--features lightwalletd-grpc-tests` cargo flag. The flag given to cargo to build the source code of the running test.
+//!
+//! Please refer to the documentation of each test for more information.
+//!
```
teor2345: Nitpick:

```suggestion

```

zebrad/tests/acceptance.rs
```diff
@@ -695,11 +780,10 @@ fn full_sync_test(network: Network, timeout_argument_name: &str) -> Result<()> {
...
     } else {
-        tracing::info!(
-            ?network,
-            "skipped full sync test, \
-             set the {:?} environmental variable to run the test",
-            timeout_argument_name,
+        eprintln!(
+            "Skipped full sync test for {}, \
+            set the {:?} environmental variable to run the test",
+            network, timeout_argument_name,
```
teor2345, line 787: Consistency:

Can you please also fix:
https://github.com/ZcashFoundation/zebra/blob/be91ab29eeabe1b6d4fdc64653b954f260e76f1b/zebrad/tests/acceptance.rs#L1488

https://github.com/ZcashFoundation/zebra/blob/d0d611ef1986c20fa9a8f77bc8602cc45c57d805/zebrad/tests/common/lightwalletd.rs#L273-L274

zebrad/tests/acceptance.rs
```diff
@@ -20,7 +20,91 @@
...
+//! $ export LIGHTWALLETD_DATA_DIR="/path/to/lightwalletd/database"
+//! $ cargo test lightwalletd_wallet_grpc_tests --features lightwalletd-grpc-tests -- --ignored --nocapture
+//! ```
+//!
+//! ### Notes
+//!
+//! - `ZEBRA_TEST_LIGHTWALLETD` env variable: Needs to be present to run any of the lightwalletd tests.
+//! - `ZEBRA_CACHED_STATE_DIR` env var: The path to a zebra blockchain database.
+//! - `LIGHTWALLETD_DATA_DIR` env variable. The path to a lightwalletd database.
+//! - `--features lightwalletd-grpc-tests` cargo flag. The flag given to cargo to build the source code of the running test.
```
teor2345: Priority: The list of environmental variables is at the bottom, and developers have to read through a lot of repetitive scripts to get to them. This makes it less useful as a quick reference.

zebrad/tests/acceptance.rs
```diff
@@ -20,7 +20,91 @@
...
+//! repo is <https://github.com/zcash/lightwalletd> but Zecwallet Lite uses a custom fork: <https://github.com/adityapk00/lightwalletd>.
+//! The custom fork from adityapk00 is the one Zebra use for this tests:
+//!
+//! Zebra lightwalletd tests are not all marked as ignored but none will not run unless
+//! at least the `ZEBRA_TEST_LIGHTWALLETD` environment variable is present.
+//! Here are some examples of running each test:
+//!
+//! ```console
+//! $ export ZEBRA_TEST_LIGHTWALLETD=true
+//! $ cargo test lightwalletd_integration -- --nocapture
```
teor2345, line 75: Reliability: The scripts here work right now, but there is no guarantee they will continue to work as we make changes to the tests.

Maintenance: We might forget to update this documentation when we add new tests.

teor2345 (commented):
This looks like it would work, but I'm not sure if it is the best way to document the environmental variables.

Instead, we could document how to call tests using the existing `docker/entrypoint.sh` script:
- give one example of calling the `docker/entrypoint.sh` script
- summarise the common environmental variables used in multiple tests
- list all the tests, and which environmental variables they use, in a compact table
- describe any tricky setup steps, such as `/zebrad-cache`

I think that would keep the documentation focused, and help fix some of the issues with the scripts in this PR.

zebrad/tests/acceptance.rs
```diff
@@ -20,7 +20,91 @@
...
-
+//!
+//! ## Large/full sync tests
+//!
+//! This file has sync tests that are marked as ignored because they take too much time to run.
+//! Some of them require environment variables or directories to be present.
+//! Here are some examples on how to run each of them:
+//!
+//! ```console
+//! $ cargo test sync_large_checkpoints_mainnet -- --ignored --nocapture
```
teor2345, line 39: Consistency/Reliability: The scripts here are slightly different to what we run in CI. These versions are more fragile.

oxarbitrage: this comment is not clear. can you put a link to what you mean ?

zebrad/tests/acceptance.rs
oxarbitrage: this comment is not clear

zebrad/tests/acceptance.rs
oxarbitrage: i dont understand the keywords you are using. what "Priority" means here ?

oxarbitrage: I am wondering what is the value of this kind of comments if the final review( https://github.com/ZcashFoundation/zebra/pull/4523#pullrequestreview-988669353) is "do it again but different".

zebrad/tests/acceptance.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/4523/commits/0d232f2fe64c1dda1a3d2c010c939a85a874ac07

zebrad/tests/acceptance.rs
conradoplg: In CI (`.github/workflows/continous-integration-os.yml`) they seem to be run as `--verbose --package zebrad --test acceptance -- --nocapture --include-ignored sync_large_checkpoints_`. Specifying the package and test saves some time since `cargo test` does not need to check every test to see if it matches the name. But otherwise this seem fine. Is that the reason @teor2345 ?

zebrad/tests/acceptance.rs
conradoplg: My understading is: the way we call these tests is by calling `entrypoint.sh`.

If we add the explicit command here, and later change `entrypoint.sh`, then they could become out of sync.

One alternative would be to change the command above, from an explicit `cargo test`, to a call to `entrypoint.sh`. I wonder if that would be a bit confusing though.

So my take is that this is fine to leave as is for now, and we can improve it later if it causes problems.

zebrad/tests/acceptance.rs
conradoplg: I think the suggestion it to simply move the description of the variables closer to the top of the comment. I agree that seems a good idea.

conradoplg (dismissed):
I'm fine with the current state of this, but moving the description of the variables closer to the top is a simple change that will be helpful for readers.

@teor2345 's suggestion of improving the overall structure of the documentation is helpful, but I think the current form is good enough for the purpose of the ticket (with the change mentioned), and we can revisit this in the future if we find that people are still confused about the tests and variables.

zebrad/tests/acceptance.rs
```diff
@@ -20,6 +20,88 @@
...
+//! ```
+//!
+//! Please refer to the documentation of each test for more information.
+//!
+//! ## Lightwalletd tests
+//!
+//! The lightwalletd software is an interface service that uses zebrad or zcashd RPC methods to serve wallets or other applications with blockchain content in an efficient manner.
+//! There are several versions of lightwalled in the form of different forks. The original
+//! repo is <https://github.com/zcash/lightwalletd> but Zecwallet Lite uses a custom fork: <https://github.com/adityapk00/lightwalletd>.
+//! The custom fork from adityapk00 is the one Zebra use for this tests:
```
oxarbitrage, line 61: This might not be true after nu5. It seems we should change this before merging. We also need to open a PR to change https://github.com/ZcashFoundation/zebra/blob/main/book/src/user/lightwalletd.md#download-and-build-lightwalletd

```suggestion
//! The lightwalletd implementation Zebra use for this tests is the original one: <https://github.com/zcash/lightwalletd> 
```

zebrad/tests/acceptance.rs
oxarbitrage: Thanks, lets leave it as it is by now.

zebrad/tests/acceptance.rs
oxarbitrage: Was done at https://github.com/ZcashFoundation/zebra/pull/4523/commits/407e95e2b9d9eb01931b6e9e8e9b06de8ce32ac1

zebrad/tests/acceptance.rs
teor2345: We've had problems when we make small changes to test names or flags, and they get skipped. Because we can't really detect that in CI. That's why the `entrypoint.sh` script uses `--include-ignored` for all tests.

`--ignored` means "only run the test if it is marked `#[ignored]` in Rust".

`--include-ignored` is more reliable, because it means "run the test with this name, it doesn't matter if it is `#[ignored]` or not".

Conrado's point about compile speed is also a good one, but that is just an efficiency thing. It won't give us the wrong test results.

zebrad/tests/acceptance.rs
teor2345: Zebra was changed to use the original fork in PR #4553. We guarantee that the original fork will work, but our tests should work with both.

We might want to keep the sentence that says there are multiple forks, because it is very confusing if you don't know that.

teor2345:
@mergifyio update

teor2345 (approved):
I think this is fine for now, we can fix up any minor documentation issues later.

PR #4577, opened 2022-06-07
change(test): add an identifiable suffix to zcash-rpc-diff temp directories

teor2345:
## Motivation

Some `zcash-rpc-diff` queries can produce very large outputs, so developers might need to delete its temporary directories.

But they can be hard to find on disk.

## Solution

- add a `.rpc-diff` suffix to temporary directories
- fall back to a generic name if the suffix doesn't work

## Review

Anyone can review this low priority PR.

### Reviewer Checklist

  - [ ] manual tests still work

### DIFF
--- a/zebra-utils/zcash-rpc-diff
+++ b/zebra-utils/zcash-rpc-diff
@@ -27,7 +27,9 @@ fi
 ZEBRAD_RPC_PORT=$1
 shift
 
-ZCASH_RPC_TMP_DIR=$(mktemp -d)
+# Use an easily identified temp directory name,
+# but fall back to a generic name if `mktemp` is outdated.
+ZCASH_RPC_TMP_DIR=$(mktemp --suffix=.rpc-diff -d 2>/dev/null || mktemp -d)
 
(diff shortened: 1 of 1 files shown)


teor2345:
I marked it as "do not merge", because we might want to wait to merge it until after the demo.

zebra-utils/zcash-rpc-diff
```diff
@@ -27,7 +27,9 @@ fi
 ZEBRAD_RPC_PORT=$1
 shift
 
-ZCASH_RPC_TMP_DIR=$(mktemp -d)
+# Use an easily identified temp directory name,
+# but fall back to a generic name if `mktemp` is outdated.
```
oxarbitrage: How can mktemp be outdated and what does this really means ?

teor2345: Older versions of the `mktemp` command might not support the `--suffix` argument.

```suggestion
# but fall back to the default temp name if `mktemp` does not understand `--suffix`.
```

teor2345:
@mergifyio update

gustavovalverde (approved):
LGTM

PR #4580, opened 2022-06-07
fix(ci): Turn off --nocapture for test-all to prevent GitHub hangs and blank logs

teor2345:
## Motivation

The test-all test is currently hanging on GitHub actions, across multiple PRs.

Also, most of its log output doesn't render in my browser, and the logs are very long.

## Solution

- disable --nocapture just for test-all, so that there is less test output

## Review

If this PR fixes the rendering or the hangs, we should merge it. It's critical either way.

If it doesn't fix anything, please just close it.

### Reviewer Checklist

  - [ ] Test all passes
  - [ ] Test all logs are all visible


dconnolly:
No effect :/

gustavovalverde:
@Mergifyio update

teor2345:
This doesn't actually fix the problem, so we don't want to merge it.

PR #4584, opened 2022-06-08
increase lightwalletd timeout, remove testnet tests

### LINKED ISSUE #4551: ci: lightwalletd tests fail with "could not read ScriptSig"
## Motivation

Zebra's lightwalletd tests are currently failing with:
> {"app":"lightwalletd","error":"error parsing block: parsing transaction 0: while parsing transparent input: could not read ScriptSig","height":1687104,"level":"warning","msg":"error zcashd getblock rpc","time":"2022-06-01T02:00:00Z"}
> {"app":"lightwalletd","level":"fatal","msg":"unable to issue RPC call to zcashd node","time":"2022-06-01T02:00:00Z","timeouts":11}
https://github.com/ZcashFoundation/zebra/runs/6681266038?check_suite_focus=true#step:9:1437

It looks like https://github.com/adityapk00/lightwalletd has not been updated for NU5, so we need to switch our tests to another `lightwalletd`.

### Designs

We could switch to the most up-to-date `lightwalletd` fork:
https://github.com/zcash/lightwalletd

Or we could temporarily disable these error messages.

conradoplg:
## Motivation

While trying to fix #4551 I managed to make it use the upstream lightwalletd, but for some reason it takes much more time to sync.

This increases the timeout to make it pass.

It also reverts to Aditya's fork because upstream seems to be deadlocking. (Increasing the timeout is still required because the recent Aditya's is also slower)

It also remove network tests that used the testnet, because it's not currently reliable (seeders are not returning any proper peers).

## Solution

- Increase timeout
- Revert to Aditya's fork (it synced with upstream so now it supports NU5)
- Disable testnet tests

## Review

It's a draft for now because I want to confirm it it works first.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/.github/workflows/zcash-lightwalletd.yml
+++ b/.github/workflows/zcash-lightwalletd.yml
@@ -11,7 +11,7 @@ on:
       # rebuild lightwalletd whenever the related Zebra code changes
       #
       # TODO: this code isn't compiled in this docker image
-      #       rebuild whenever the actual code at zcash/lightwalletd/master changes
+      #       rebuild whenever the actual code at lightwalletd/master changes
       - 'zebra-rpc/**'
       - 'zebrad/tests/acceptance.rs'
       - 'zebrad/src/config.rs'
--- a/.github/workflows/zcash-lightwalletd.yml
+++ b/.github/workflows/zcash-lightwalletd.yml
@@ -50,7 +50,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3.0.2
         with:
-          repository: zcash/lightwalletd
+          repository: adityapk00/lightwalletd
           ref: 'master'
           persist-credentials: false
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -92,7 +92,7 @@ FROM deps AS tests
 # TODO: do not hardcode the user /root/ even though is a safe assumption
 # Pre-download Zcash Sprout, Sapling parameters and Lightwalletd binary
 COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/zcash-params /root/.zcash-params /root/.zcash-params
-COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/lightwalletd:pr-4584 /lightwalletd /usr/local/bin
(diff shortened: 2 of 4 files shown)


.github/workflows/zcash-lightwalletd.yml
```diff
@@ -11,7 +11,7 @@ on:
       # rebuild lightwalletd whenever the related Zebra code changes
       #
       # TODO: this code isn't compiled in this docker image
-      #       rebuild whenever the actual code at zcash/lightwalletd/master changes
+      #       rebuild whenever the actual code at lightwalletd/master changes
```
upbqdn, line 14: Shouldn't this be `adityapk00/lightwalletd/master`?

upbqdn (approved):
I have one question, everything else looks good.

.github/workflows/zcash-lightwalletd.yml
dconnolly, line 14:
```suggestion
      #       rebuild whenever the actual code at adityapk00/lightwalletd/master changes
```

conradoplg: (I had added a comment about this but didn't publish it :smiling_face_with_tear: )

My idea was to avoid having unneeded repetition so that we don't need to change it if we ever change the repository again.

conradoplg: My idea was to avoid having unneeded repetition so that we don't need to change it if we ever change the repository again.

upbqdn: I see, that makes sense then.

conradoplg: I'll mark this as solved in order to fix CI since it's blocking other PRs. @dconnolly let me know if you'd like to really make this change, I can make it in another PR

dconnolly: No biggie 👍

gustavovalverde (approved):

conradoplg:
Mergify test failed on an [apparent network issue](https://github.com/ZcashFoundation/zebra/runs/6864123644?check_suite_focus=true) while downloading params. Synced with main to try again.

PR #4586, opened 2022-06-08
feat(state): delete old database directories

### LINKED ISSUE #1213: Detect when an older state version is present, and delete it
**Is your feature request related to a problem? Please describe.**

The plan for future database format changes is to blow away the database and start over.  Because the database is a cache of chain state, we don't lose any information by doing so, and it's much simpler and more reliable than trying to perform a migration.

_Originally posted by @hdevalence in https://github.com/ZcashFoundation/zebra/pull/1172#issuecomment-714634746_

**Describe the solution you'd like**

Delete any previous versions on database load:
- [ ] {cache_dir}/zebra/v{number}/{network}
    - [ ] If Zebra is running version `n`, we must not delete versions greater than `n`
    - [ ] If possible, restrict the deletion to one filesystem, and don't follow symlinks

Optional:
- [ ] Add a debug config that turns this behaviour off, so developers can keep old states around

Note: number starts at 0

**Scheduling**

This ticket is only required for the first zebra-state database format change after the Zebra stable release.

oxarbitrage:
## Motivation

When the database format change, a new full blockchain will be downloaded when zebra starts. The old blockchain will remain in the file system occupying a lot of hard drive space.

We want to delete this old versions of the blockchain by default when zebra starts, but we also want a config option to not do it if the user want to keep the different versions for whatever reason.

This PR will close https://github.com/ZcashFoundation/zebra/issues/1213 if it gets merged.

## Solution

When zebra start and if the `delete_old_database` config option is set to `true`, the code added here will check the database directories and delete the ones that are considered all. Directory will be deleted will all the content inside them. 

## Review

I think anyone can review, open for improvements, probably some code in the `check_and_delete_old_databases` can be simplified a bit.

No tests were made, i do tested manually. `check_and_delete_old_databases` function could be tested with different scenarios but i am unsure if it will worth to do it.

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -128,24 +128,25 @@ impl Default for Config {
 /// Iterate over the files and directories in the databases folder and delete if:
 /// - The state directory exists.
 /// - The entry is a directory.
-/// - The directory name has a lenght of at least 2 characters.
+/// - The directory name has a length of at least 2 characters.
 /// - The directory name has a prefix `v`.
 /// - The directory name without the prefix can be parsed as an unsigned number.
 /// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
 pub async fn check_and_delete_old_databases(config: Config) {
-    if !config.ephemeral && config.delete_old_database {
-        info!("checking old database versions");
+    if config.ephemeral || !config.delete_old_database {
+        return;
+    }
 
-        let cache_dir = config.cache_dir.join("state");
-        if let Some(read_dir) = read_dir(cache_dir.clone()) {
-            for entry in read_dir.flatten() {
-                if let Some(dir_name) = parse_dir_name(entry) {
-                    if let Some(version_number) = parse_version_number(dir_name.clone()) {
-                        if version_number < crate::constants::DATABASE_FORMAT_VERSION {
-                            let delete_path = cache_dir.join(dir_name);
-                            if remove_dir_all(delete_path.clone()).is_ok() {
-                                info!("deleted outdated state directory {:?}", delete_path);
-                            }
+    info!("checking old database versions");
+    let cache_dir = config.cache_dir.join("state");
+    if let Some(read_dir) = read_dir(cache_dir.clone()) {
+        for entry in read_dir.flatten() {
+            if let Some(dir_name) = parse_dir_name(entry) {
+                if let Some(version_number) = parse_version_number(dir_name.clone()) {
+                    if version_number < crate::constants::DATABASE_FORMAT_VERSION {
+                        let delete_path = cache_dir.join(dir_name);
+                        if remove_dir_all(delete_path.clone()).is_ok() {
+                            info!("deleted outdated state directory {:?}", delete_path);
                         }
                     }
                 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -1,4 +1,7 @@
-use std::path::PathBuf;
+use std::{
+    fs::{remove_dir_all, DirEntry, ReadDir},
+    path::PathBuf,
+};
 
 use serde::{Deserialize, Serialize};
 
@@ -118,4 +121,65 @@ impl Default for Config {
             delete_old_database: true,
         }
     }
+}
+
+/// Check if there are old database folders and delete them from the filesystem.
+///
+/// Iterate over the files and directories in the databases folder and delete if:
+/// - The state directory exists.
+/// - The entry is a directory.
+/// - The directory name has a lenght of at least 2 characters.
+/// - The directory name has a prefix `v`.
+/// - The directory name without the prefix can be parsed as an unsigned number.
+/// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
+pub fn check_and_delete_old_databases(cache_dir: PathBuf) {
+    let cache_dir = cache_dir.join("state");
+    if let Some(read_dir) = read_dir(cache_dir.clone()) {
+        for entry in read_dir.flatten() {
+            if let Some(dir_name) = parse_dir_name(entry) {
+                if let Some(version_number) = parse_version_number(dir_name.clone()) {
+                    if version_number < crate::constants::DATABASE_FORMAT_VERSION {
+                        let delete_path = cache_dir.join(dir_name);
+                        if remove_dir_all(delete_path.clone()).is_ok() {
+                            info!("deleted outdated state directory {:?}", delete_path);
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+fn read_dir(cache_dir: PathBuf) -> Option<ReadDir> {
+    if cache_dir.exists() {
+        if let Ok(read_dir) = cache_dir.read_dir() {
+            return Some(read_dir);
+        }
+    }
+    None
+}
+
+fn parse_dir_name(entry: DirEntry) -> Option<String> {
+    if let Ok(file_type) = entry.file_type() {
+        if file_type.is_dir() {
+            if let Ok(dir_name) = entry.file_name().into_string() {
+                return Some(dir_name);
+            }
+        }
+    }
+    None
+}
+
+fn parse_version_number(dir_name: String) -> Option<u32> {
+    if dir_name.len() >= 2 && dir_name.starts_with('v') {
+        if let Some(potential_version_number) = dir_name.strip_prefix('v') {
+            return Some(
+                potential_version_number
+                    .to_string()
+                    .parse()
+                    .unwrap_or(u32::MAX),
+            );
+        }
+    }
+    None
 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -57,6 +57,13 @@ pub struct Config {
     ///
     /// Set to `None` by default: Zebra continues syncing indefinitely.
     pub debug_stop_at_height: Option<u32>,
+
+    /// Whether to delete the old database directories when present.
+    ///
+    /// Set to `true` by default. If this is set to `false`,
+    /// no check for old database versions will be made and nothing will be
+    /// deleted.
+    pub delete_old_database: bool,
 }
 
 fn gen_temp_path(prefix: &str) -> PathBuf {
@@ -108,6 +115,7 @@ impl Default for Config {
             cache_dir,
             ephemeral: false,
             debug_stop_at_height: None,
+            delete_old_database: true,
         }
     }
 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -1,9 +1,13 @@
+//! Cached state configuration for Zebra.
+
 use std::{
-    fs::{remove_dir_all, DirEntry, ReadDir},
-    path::PathBuf,
+    fs::{canonicalize, remove_dir_all, DirEntry, ReadDir},
+    path::{Path, PathBuf},
 };
 
 use serde::{Deserialize, Serialize};
+use tokio::task::{spawn_blocking, JoinHandle};
+use tracing::Span;
 
 use zebra_chain::parameters::Network;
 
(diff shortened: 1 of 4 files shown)


zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,43 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+
+/// Check if there are old database folders and delete them from the filesystem.
+///
+/// Iterate over the files and directories in the databases folder and delete if:
+/// - The entry is a directory.
+/// - The directory name has a lenght of at least 2 characters.
+/// - The directory name has a prefix `v`.
+/// - The directory name without the prefix can be parsed as an unsigned number.
+/// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
+fn check_and_delete_old_databases(cache_dir: PathBuf) -> Result<(), Report> {
```
oxarbitrage: I was thinking on putting this somewhere else but not sure where so ended up here.

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,43 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+                .file_name()
+                .into_string()
+                .expect("conversion from osstring to string should work.");
+            if dir_name.len() >= 2 && dir_name.starts_with('v') {
+                let potential_version_number_string = dir_name
+                    .strip_prefix('v')
+                    .expect("should not panic as we know there is a `v` prefix.")
+                    .to_string();
+                let version_number: u32 =
+                    potential_version_number_string.parse().unwrap_or(u32::MAX);
```
oxarbitrage: When the string after `v` can't be parsed as a number we add the max u32 value so the line below will not be able to delete something as `vwhatever`. I think we only want to delete directories that are blockchain databases and not anything else that could be there for whatever reason.

zebrad/src/commands/start.rs
teor2345: Optional: We could move it to `zebra_state::check_and_delete_old_databases()`?
(It could go in a finalized state or config submodule, then get a `pub use` in `lib.rs`.)

It seems like it's very state-specific, so it probably belongs in that crate. And if we keep it separate from `zebra_state::init()`, then we can run it in a different task.

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+///
+/// Iterate over the files and directories in the databases folder and delete if:
+/// - The state directory exists.
+/// - The entry is a directory.
+/// - The directory name has a lenght of at least 2 characters.
+/// - The directory name has a prefix `v`.
+/// - The directory name without the prefix can be parsed as an unsigned number.
+/// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
+fn check_and_delete_old_databases(cache_dir: PathBuf) -> Result<(), Report> {
+    let cache_dir = cache_dir.join("state");
```
teor2345: This path won't work for ephemeral databases - their path format is different, and they are supposed to delete themselves anyway.

Can we skip this whole function if `config.ephemeral` is set?

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+    if cache_dir.exists() {
+        for entry in (cache_dir.read_dir()?).flatten() {
+            if entry.file_type()?.is_dir() {
+                let dir_name = entry
+                    .file_name()
+                    .into_string()
+                    .expect("conversion from osstring to string should work.");
```
teor2345: This conversion can fail on some filesystems, depending on the directory name.

We should continue on error, because `v{number}` is guaranteed to be a valid UTF-8 string. So if conversion fails, we would end up skipping that directory anyway. (But maybe going on to delete other valid directories.)

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+                if dir_name.len() >= 2 && dir_name.starts_with('v') {
+                    let potential_version_number_string = dir_name
+                        .strip_prefix('v')
+                        .expect("should not panic as we know there is a `v` prefix.")
+                        .to_string();
```
teor2345: Optional: `strip_prefix`, `starts_with`, and the length check can be combined into something like:

```rust
if let Some(version_number) = dir_name.strip_prefix('v').and_then(u32::parse) { ... }
```

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+                    let version_number: u32 =
+                        potential_version_number_string.parse().unwrap_or(u32::MAX);
+                    if version_number < zebra_state::constants::DATABASE_FORMAT_VERSION {
+                        let delete_path = cache_dir.join(dir_name);
+                        info!(
+                            "deleting {}",
+                            delete_path
+                                .to_str()
+                                .expect("path is valid, should not panic")
```
teor2345: This conversion can fail on some filesystems, depending on the home directory name:
```suggestion
                            "deleting outdated state directory {:?}",
                            delete_path,
```

zebrad/src/commands/start.rs
```diff
@@ -105,6 +105,11 @@ impl StartCmd {
         let config = app_config().clone();
         info!(?config);
 
+        if config.state.delete_old_database {
+            info!("checking old database versions");
+            check_and_delete_old_databases(config.state.cache_dir.clone())?;
+        }
```
teor2345: Optional: Deleting old database files can take a few minutes. Can we use `spawn_blocking` to delete files, and turn it into a task like `groth16_download`?

That way, the old state gets deleted while the rest of Zebra is launching.

teor2345:
This looks good, and it looks like it should work.

If we wanted to test it, we'd need to pass the current state version as a function argument. And I'm not sure it's worth it. Our ephemeral database code is very similar, and it works fine.

I have some suggestions about edge cases, simplifications, and module structure.

Did you want to catch up to talk through them, or I could make a PR, or I could write some comments?
(I'm happy to just do edge cases to start with, if that would help.)

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+                        potential_version_number_string.parse().unwrap_or(u32::MAX);
+                    if version_number < zebra_state::constants::DATABASE_FORMAT_VERSION {
+                        let delete_path = cache_dir.join(dir_name);
+                        info!(
+                            "deleting {}",
+                            delete_path
+                                .to_str()
+                                .expect("path is valid, should not panic")
+                        );
+                        remove_dir_all(delete_path)?;
```
teor2345: We probably want to continue on error, rather than skipping the rest of the directories we could delete.

zebrad/src/commands/start.rs
```diff
@@ -601,3 +606,46 @@ impl config::Override<ZebradConfig> for StartCmd {
...
+/// - The entry is a directory.
+/// - The directory name has a lenght of at least 2 characters.
+/// - The directory name has a prefix `v`.
+/// - The directory name without the prefix can be parsed as an unsigned number.
+/// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
+fn check_and_delete_old_databases(cache_dir: PathBuf) -> Result<(), Report> {
+    let cache_dir = cache_dir.join("state");
+    if cache_dir.exists() {
+        for entry in (cache_dir.read_dir()?).flatten() {
+            if entry.file_type()?.is_dir() {
```
teor2345: We probably want to continue on subdirectory errors, rather than skipping the remaining subdirectories.

oxarbitrage:
Comments or PR are welcome (whatever is easier for you). Thanks.

teor2345 (commented):
Filesystem error and path handling is a bit tricky in Rust, I made some suggestions to avoid panics.

I have some other suggestions about performance and module structure, but they are optional.

zebrad/src/commands/start.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/4586/commits/fd02f55cb0cfdf3dacbd37a77609275b96aa24c5

teor2345 (dismissed):
The premature returns and panics are all fixed, so we can merge this if you want.

(The other changes can be made in other PRs, but they are all optional.)

zebra-state/src/config.rs
```diff
@@ -1,4 +1,7 @@
-use std::path::PathBuf;
+use std::{
+    fs::{remove_dir_all, DirEntry, ReadDir},
+    path::PathBuf,
+};
 
 use serde::{Deserialize, Serialize};
```
teor2345: In async functions, we should use async filesystem operations.

`std` filesystem operations will block all the async tasks on the same thread, which is what we're trying to avoid.

```suggestion
use std::path::PathBuf;

use serde::{Deserialize, Serialize};
use tokio::fs::{remove_dir_all, DirEntry, ReadDir};
```

oxarbitrage: If we do this change there will be a lot of additional work to be done:

```
oxarbitrage@oxarbitrage-Lenovo-ideapad-320S-14IKB:~/zebra/issue1213/zebra$ cargo clippy --all-features --all-targets
   Compiling zebrad v1.0.0-beta.11 (/home/oxarbitrage/zebra/issue1213/zebra/zebrad)
    Checking zebra-state v1.0.0-beta.11 (/home/oxarbitrage/zebra/issue1213/zebra/zebra-state)
error[E0599]: the method `flatten` exists for struct `tokio::fs::ReadDir`, but its trait bounds were not satisfied
   --> zebra-state/src/config.rs:140:31
    |
140 |         for entry in read_dir.flatten() {
    |                               ^^^^^^^ method cannot be called on `tokio::fs::ReadDir` due to unsatisfied trait bounds
    |
   ::: /home/oxarbitrage/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.18.2/src/fs/read_dir.rs:57:1
    |
57  | pub struct ReadDir(State);
    | -------------------------- doesn't satisfy `tokio::fs::ReadDir: std::iter::Iterator`
    |
note: the following trait bounds were not satisfied because of the requirements of the implementation of `std::iter::Iterator` for `_`:
      `tokio::fs::ReadDir: std::iter::Iterator`
   --> zebra-state/src/service/block_iter.rs:70:6
    |
70  | impl Iterator for Iter<'_> {
    |      ^^^^^^^^     ^^^^^^^^

error[E0599]: no method named `is_ok` found for opaque type `impl futures::Future<Output = std::result::Result<(), std::io::Error>>` in the current scope
   --> zebra-state/src/config.rs:145:64
    |
145 |                         if remove_dir_all(delete_path.clone()).is_ok() {
    |                                                                ^^^^^ method not found in `impl futures::Future<Output = std::result::Result<(), std::io::Error>>`
    |
help: consider `await`ing on the `Future` and calling the method on its `Output`
    |
145 |                         if remove_dir_all(delete_path.clone()).await.is_ok() {
    |                                                                ++++++

error[E0308]: mismatched types
   --> zebra-state/src/config.rs:158:25
    |
158 |             return Some(read_dir);
    |                         ^^^^^^^^ expected struct `tokio::fs::ReadDir`, found struct `std::fs::ReadDir`

error[E0308]: mismatched types
   --> zebra-state/src/config.rs:165:12
    |
165 |     if let Ok(file_type) = entry.file_type() {
    |            ^^^^^^^^^^^^^ expected opaque type, found enum `std::result::Result`
    |
note: while checking the return type of the `async fn`
   --> /home/oxarbitrage/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.18.2/src/fs/read_dir.rs:285:38
    |
285 |     pub async fn file_type(&self) -> io::Result<FileType> {
    |                                      ^^^^^^^^^^^^^^^^^^^^ checked the `Output` of this `async fn`, expected opaque type
    = note: expected opaque type `impl futures::Future<Output = std::result::Result<std::fs::FileType, std::io::Error>>`
                      found enum `std::result::Result<_, _>`
help: consider `await`ing on the `Future`
    |
165 |     if let Ok(file_type) = entry.file_type().await {
    |                                             ++++++

Some errors have detailed explanations: E0308, E0599.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `zebra-state` due to 4 previous errors
warning: build failed, waiting for other jobs to finish...
^C  Building [=====================> ] 1193/1214: zebra-state(test)                                                                                                    
oxarbitrage@oxarbitrage-Lenovo-ideapad-320S-14IKB:~/zebra/issue1213/zebra$ 
```

zebra-state/src/config.rs
```diff
@@ -108,6 +118,73 @@ impl Default for Config {
...
+
+    info!("checking old database versions");
+    let cache_dir = config.cache_dir.join("state");
+    if let Some(read_dir) = read_dir(cache_dir.clone()) {
+        for entry in read_dir.flatten() {
+            if let Some(dir_name) = parse_dir_name(entry) {
+                if let Some(version_number) = parse_version_number(dir_name.clone()) {
+                    if version_number < crate::constants::DATABASE_FORMAT_VERSION {
+                        let delete_path = cache_dir.join(dir_name);
+                        if remove_dir_all(delete_path.clone()).is_ok() {
```
teor2345: If we get the `delete_path` wrong, we could delete the user's entire home directory. (Or the entire disk, if Zebra is running as the root user.)

This might also happen if they have a weird cache set up using symlinks.

Here are some things we could do:
- a runtime check that the path is inside the cache directory ([canonicalize](https://docs.rs/tokio/latest/tokio/fs/fn.canonicalize.html), [starts_with](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.starts_with))
- an acceptance test that makes sure we delete an old directory, but don't delete files outside the cache directory

What do you think? I can help write these if you want.

zebrad/src/commands/start.rs
oxarbitrage: I tried to do this without luck. I think it is not that bad as it is specially now that we moved the parse code into its own function however feel free to send a suggestion into the `parse_version_number()`.

zebra-state/src/config.rs
oxarbitrage: I dont think i will add more to this PR today so feel free to add this.

zebrad/src/commands/start.rs
teor2345: I did this in a651377daf70f936c045d565fc1293ad202628de, my original suggestion didn't quite work.

zebrad/src/commands/start.rs
teor2345: I used `spawn_blocking` to do this in 670533f38719a51d351d907a5101e03745dfd2a4.

zebra-state/src/config.rs
teor2345: I tried making some of those changes, but it seems like `spawn_blocking` is easier.

So I just launched the whole delete task in a separate thread in 670533f38719a51d351d907a5101e03745dfd2a4.

zebra-state/src/config.rs
teor2345: I wrote the runtime check in 4249029734c3b144184bc05877b9402fa02c28dd.

I didn't have time to write the tests, there are some other things I need to fix today. But I'm happy to help tomorrow.

Did you want to do a simple test?
We could create a fake state directory, there doesn't need to be anything in it.

zebra-state/src/config.rs
oxarbitrage: I think it will be nice to test some suggestions before making them. I understand that this is a lot more work than just adding a suggestion of code change without seeing anything else but i think it worth it sometimes. This is one of the things that @jvff used to do. If you realize the suggestion is too complicated then you don't even add it, if it is how you expected then we have a suggestion that can be implemented with the click of a button.

zebra-state/src/config.rs
oxarbitrage: I am not sure of what exactly the test should do. Create a state directory and then run zebra ? Then make sure the directory is not deleted ? Add some old version directories inside and make sure they do get deleted ?

teor2345: > Create a state directory and then run zebra ? Then make sure the directory is not deleted ?

The restart/relaunch tests already do this, so we don't need to add any extra tests for this.

> Add some old version directories inside and make sure they do get deleted ?

This would be good, and also make sure we don't delete files or directories outside the cache directory.

oxarbitrage: How do you know no file in the filesystem was deleted ?

oxarbitrage: Maybe put a `v1` dir anywhere and make sure it was not deleted ?

teor2345: Sounds good!

Maybe put it next to the `state` directory?
(A `v1` directory inside the `state` directory gets deleted. But if it's just one directory lower, it should stay there.)

zebra-state/src/config.rs
teor2345: Thanks for that feedback, I will try to only make suggestions that I am sure will work.

zebra-state/src/config.rs
oxarbitrage: added test at https://github.com/ZcashFoundation/zebra/pull/4586/commits/5a707e1fad1028dc3de6236fb56970f307bb65c4

teor2345:
I'm going to merge `main` into this PR, to check if the CI failures are actually in this PR.
(I thought we had fixed the macOS failure with PR #4613.)

teor2345:
@mergifyio update

zebrad/tests/acceptance.rs
```diff
@@ -1642,6 +1642,66 @@ async fn fully_synced_rpc_test() -> Result<()> {
...
+    create_dir(&outside_dir)?;
+    assert!(outside_dir.as_path().exists());
+
+    // create a `v1` dir insidecache dir that should be deleted
+    let inside_dir = cache_dir.join("v1");
+    create_dir(&inside_dir)?;
+    assert!(inside_dir.as_path().exists());
+
+    // modify config with our cache dir and not ephimeral configuration
+    // (delete old databases function will not run when epehemeral = true)
```
teor2345: Nit: There are a few typos here, but `ephemeral` is a tricky word to spell. We should have just called the config `temp`.

```suggestion
    // create a `v1` dir inside cache dir that should be deleted
    let inside_dir = cache_dir.join("v1");
    create_dir(&inside_dir)?;
    assert!(inside_dir.as_path().exists());

    // modify config with our cache dir and not ephemeral configuration
    // (delete old databases function will not run when ephemeral = true)
```

zebrad/tests/acceptance.rs
```diff
@@ -1642,6 +1642,66 @@ async fn fully_synced_rpc_test() -> Result<()> {
...
+    let mut child = run_dir
+        .with_config(&mut config)?
+        .spawn_child(args!["start"])?;
+
+    // delete checker running
+    child.expect_stdout_line_matches("checking for old database versions".to_string())?;
+
+    // inside dir was deleted
+    child.expect_stdout_line_matches(format!(
+        "deleted outdated state directory deleted_state={:?}",
```
teor2345, line 1684: It looks like the new test never stops on macOS:
> test delete_old_databases has been running for over 60 seconds

https://github.com/ZcashFoundation/zebra/runs/6939346214?check_suite_focus=true#step:13:54440

It seems like the code runs correctly:
> zebra_state::config: deleted outdated state directory deleted_state="/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebrad_testsfBKR8x/state/v1"

https://github.com/ZcashFoundation/zebra/runs/6939346214?check_suite_focus=true#step:13:54350

But the regex in the test doesn't match.

This is probably because the code uses `canonicalize` on the path, but the test doesn't. (macOS `/tmp` is a symlink to its actual temporary directory.)

zebrad/tests/acceptance.rs
```diff
@@ -1642,6 +1642,66 @@ async fn fully_synced_rpc_test() -> Result<()> {
...
+        inside_dir
+    ))?;
+    assert!(!inside_dir.as_path().exists());
+
```
teor2345, line 1688: Can we check for "finished old database version cleanup task" before checking that the directory was not deleted?

This makes sure:
- the task doesn't hang
- the task has finished deleting all the directories it wants to delete

teor2345 (commented):
This looks good, we just need to fix a non-canonical path on macOS.

zebrad/tests/acceptance.rs
oxarbitrage: My bad, didn't pay enough attention.

teor2345 (approved):
Thanks, this looks great, and thanks for the tests!

PR #4587, opened 2022-06-09
fix(config): Duration fields

### LINKED ISSUE #2847: Make config `Duration` easier to write in the toml file
## Motivation

In the network and in the mempool config sections we use `Duration` type for some variables. This are not very friendly for the user so we might change them in the future. 

Some hints on what to do can be found in https://github.com/ZcashFoundation/zebra/pull/2845#discussion_r724543181

This is very low priority at this moment. 

## Potential APIs

We could use `humantime_serde` like `arti` does.

oxarbitrage:
## Motivation

We have a few fields of the Zebra config that are of type `Duration`. This are at the moment `network.crawl_new_peer_interval` and `eviction_memory_time`. 

This fields are displayed as follows in the configuration file:

```
[mempool.eviction_memory_time]
nanos = 0
secs = 3600
```

We want a more user friendly way for this entries of the config. Close https://github.com/ZcashFoundation/zebra/issues/2847

## Solution

Added [humantime-serde](https://crates.io/crates/humantime-serde) to this fields as suggested in the ticket as it seems the best option. Now the fields with the problem will look as:

```
eviction_memory_time = '1h'
```

Which is a lot better. The change also allowed to move the debug fields to the end of the config section.

## Review

I think anyone can review this.


zebrad/src/components/mempool/config.rs
```diff
@@ -39,10 +28,18 @@ pub struct Config {
...
+
+    /// If the state's best chain tip has reached this height, always enable the mempool,
+    /// regardless of Zebra's sync status.
+    ///
+    /// Set to `None` by default: Zebra always checks the sync status before enabling the mempool.
+    //
+    // TODO:
+    // - allow the mempool to be enabled before the genesis block is committed?
+    //   we could replace `Option` with an enum that has an `AlwaysEnable` variant
+    pub debug_enable_at_height: Option<u32>,
```
upbqdn, line 42: It's not clear to me why you moved the field to the bottom.

upbqdn (approved):
All looks good.

zebrad/src/components/mempool/config.rs
oxarbitrage: There was a `TODO` in line 19 of this file that was saying:
```
// - move debug configs last (needs #2847)
```

So i think we want to have the debug options at the end of each section as the other options are more important. That is the reason i moved, i could revert that part if it is not a good idea anymore, the `TODO` comment may be pretty old.

upbqdn: Makes sense, thank you for explaining. I think it's good the way it is.

teor2345:
@mergifyio update

PR #4598, opened 2022-06-13
feat(actions): delete old GCP resources

### LINKED ISSUE #2090: Auto-delete old Google Cloud instance templates
**Is your feature request related to a problem? Please describe.**

Every few months, we run out of Google Cloud instance templates:
> ERROR: (gcloud.compute.instance-templates.create-with-container) Could not fetch resource:
> - Quota 'INSTANCE_TEMPLATES' exceeded.  Limit: 300.0 globally.

**Describe the solution you'd like**

Instead, we should auto-delete templates that are more than `30` days old, before creating a new template.

**Describe alternatives you've considered**

Just keep on manually deleting them.

**Additional context**

We expect this to next become a problem some time in June, so I've scheduled it for the last sprint in May.

gustavovalverde:
## Motivation

We'd like to delete old resources, in this specific case we're starting with instance templates, as those resources commonly hit GCP's quota limit.

Fixes #2090 

### Designs

- Create a workflow specifically to delete resources based on an scheduled action (monthly)
- This action should be generic, and be able to receive multiple resources not just the ones specified in this PR

## Solution

- Create `delete-gcp-resources.yml` action
- Add a `Delete instance templates` step which list instances with 30+ days of creation and delete them all

## Review

Not ready for review

### DIFF
--- a/.github/workflows/delete-gcp-resources.yml
+++ b/.github/workflows/delete-gcp-resources.yml
@@ -37,5 +37,5 @@ jobs:
 
           for TEMPLATE in $TEMPLATES
           do
-            gcloud compute instance-templates ${TEMPLATE}
+            gcloud compute instance-templates delete ${TEMPLATE}
           done
--- a/.github/workflows/delete-gcp-resources.yml
+++ b/.github/workflows/delete-gcp-resources.yml
@@ -5,8 +5,8 @@ on:
     - cron: "0 0 1 * *"
   workflow_dispatch:
   push:
-    branches:
-      - "!main"
+    # branches:
+    #   - "!main"
 
 jobs:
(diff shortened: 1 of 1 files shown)


teor2345 (commented):
Looks good, is there any particular reason to do this monthly, rather than weekly?

I'm wondering if we might have a busy month, and reach the quota within that month.

gustavovalverde:
> Looks good, is there any particular reason to do this monthly, rather than weekly?

We get near that quota (700 templates) every 4-6 months, based on this quarter rhythm (~100 PRs/month) it should be safe enough even if a month gets very busy

teor2345 (approved):

PR #4599, opened 2022-06-13
ci(lightwalletd): Zebra's Dockerfile needs `latest` lwd image

### LINKED ISSUE #4590: CI: correct lightwalletd image not being fetched
## Motivation

In CI, when the `Dockerfile` is built by `build-docker-image.yml`, it does not specify a tag in `COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/lightwalletd /lightwalletd /usr/local/bin`, which makes it fetch the `latest` tag.

However, the `latest` tag does not seem to be applied currently, which make it point to a old image. (I temporarily worked around that by tagging manually).

We should change it so that the correct image is used.

### Designs

I'm not sure what is the best approach here. We probably need something similar to the logic used by cached disks: use the image from the same PR/branch, or fall back to the last `main` image.

gustavovalverde:
## Motivation

We need to add this condition to lightwalletd docker build, to always create a `latest` tag when merging to the default branch: `main`, as we're fetching this tag from the Zebra Dockerfile build, but this tag is not being added when building lightwalletd

Fixes #4590 

### Designs

By default `docker/metadata-action` creates tags based on the event being triggered, but the latest tag is reserved for releases. As we can't wait releases for this specific use case, we're building the `latest` when merging to the default branch.

## Solution

- Add this condition to `docker/metadata-action` --> `type=raw,value=latest,enable={{is_default_branch}}`

## Review

Anyone from @ZcashFoundation/devops-reviewers or @conradoplg

### DIFF
--- a/.github/workflows/zcash-lightwalletd.yml
+++ b/.github/workflows/zcash-lightwalletd.yml
@@ -73,6 +73,7 @@ jobs:
           images: |
             ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}
           # generate Docker tags based on the following events/attributes
+          # set latest tag for default branch
           tags: |
             type=schedule
             type=ref,event=branch
@@ -81,6 +82,7 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
             type=sha
+            type=raw,value=latest,enable={{is_default_branch}}
 
       - name: Set up QEMU
         id: qemu


dconnolly (approved):

conradoplg:
I'm not sure if this completely solves the problem. What happens if we need to update the image? The PR that changes it will use the `latest` image which will not be the updated one and it will fail CI.

teor2345:
> I'm not sure if this completely solves the problem. What happens if we need to update the image? The PR that changes it will use the `latest` image which will not be the updated one and it will fail CI.

I had the same thought, it makes testing a bit tricky.

We might want to use the same algorithm we use for the state:
- wait for the image from the current PR, if it is being rebuilt, or
- use the image from `main`

upbqdn:
I can't figure out why GitHub says this: 
image
above Teor's comment.

teor2345:
> I can't figure out why GitHub says this: image above Teor's comment.

The commit text contains `#4599`, so GitHub automatically mentions that commit on PR #4599.

Issue #2297, opened 2021-06-14
Enforce unwrap-pertinent Clippy lints


dconnolly:
## Motivation

We have a few bugs that have crept in due to `unwrap()`s being trigger-able in production. 

## Solution

Enforce these Clippy lints in production (i.e. non-test code), by either adding `clippy:(allow|deny)` annotations at the top of each crate or in each module.

https://rust-lang.github.io/rust-clippy/master/#fallible_impl_from
> Checks for impls of From<..> that contain panic!() or unwrap()

https://rust-lang.github.io/rust-clippy/master/#unwrap_in_result
> Checks for functions of type Result that contain expect() or unwrap()

This might be extra pedantic:
https://rust-lang.github.io/rust-clippy/master/#unwrap_used
>Checks for impls of From<..> that contain panic!() or unwrap() 

### Alternatives

Try to enforce these rules at review time, by a human.
Try to write a custom Clippy lint to enforce these rules: https://github.com/trailofbits/dylint (why, if Clippy proper has the above lints?)

teor2345:
I think we should enforce `fallible_impl_from` in test code as well.

We're unlikely to have `From` impls in test code, and if we do, we might as well get them right.

@dconnolly what do you think?

teor2345:
This is not a priority for Zebra right now.

PR #4609, opened 2022-06-14
feat(clippy): Add `fallible_impl_from` lint

oxarbitrage:
## Motivation

In https://github.com/ZcashFoundation/zebra/issues/2297 we want to add some panic lints to zebra. This is part 1 of 3 where the `fallible_impl_from` is added. 

## Solution

Add the lint and fix the code where needed.

## Review

Please review carefully, there are error messages that might be changed. I think @dconnolly wrote most of the changed code in orchard and sapling keys but anyone that can take a look is welcome.

### Reviewer Checklist

  - [ ] Current test pass
  - [ ] Errors strings added are correct

## Follow Up Work

We want to add 2 more related lints but i think it was too much to add them in a single PR so i splitted into 3.

### DIFF
--- a/zebra-chain/src/orchard/keys.rs
+++ b/zebra-chain/src/orchard/keys.rs
@@ -1,7 +1,7 @@
 //! Orchard key types.
 //!
 //! <https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents>
-
+#![allow(clippy::fallible_impl_from)]
 #![allow(dead_code)]
 
 #[cfg(test)]
@@ -308,7 +308,9 @@ impl From<SpendValidatingKey> for [u8; 32] {
 
 impl From<SpendAuthorizingKey> for SpendValidatingKey {
     fn from(ask: SpendAuthorizingKey) -> Self {
-        let sk = redpallas::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).unwrap();
+        let sk = redpallas::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).expect(
+            "a scalar converted to byte array and then converted back to a scalar should not fail",
+        );
 
         Self(redpallas::VerificationKey::from(&sk))
     }
--- a/zebra-chain/src/sapling/keys.rs
+++ b/zebra-chain/src/sapling/keys.rs
@@ -8,6 +8,7 @@
 //! [ps]: https://zips.z.cash/protocol/protocol.pdf#saplingkeycomponents
 //! [3.1]: https://zips.z.cash/protocol/protocol.pdf#addressesandkeys
 #![allow(clippy::unit_arg)]
+#![allow(clippy::fallible_impl_from)]
 #![allow(dead_code)]
 
 #[cfg(test)]
@@ -507,7 +508,9 @@ impl From<AuthorizingKey> for [u8; 32] {
 
 impl From<SpendAuthorizingKey> for AuthorizingKey {
     fn from(ask: SpendAuthorizingKey) -> Self {
-        let sk = redjubjub::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).unwrap();
+        let sk = redjubjub::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).expect(
+            "a scalar converted to byte array and then converted back to a scalar should not fail",
+        );
         Self(redjubjub::VerificationKey::from(&sk))
     }
 }
--- a/zebra-network/src/protocol/external/message.rs
+++ b/zebra-network/src/protocol/external/message.rs
@@ -352,7 +352,11 @@ where
             // use specific varieties of `RejectReason`.
             ccode: RejectReason::Other,
 
-            reason: e.source().unwrap().to_string(),
+            reason: if let Some(reason) = e.source() {
+                reason.to_string()
+            } else {
+                String::from("")
+            },
 
             // Allow this to be overridden but not populated by default, methinks.
             data: None,
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -50,7 +50,10 @@ rustflags = [
     "-Aclippy::try_err",
 
     # Links in public docs can point to private items.
-    "-Arustdoc::private_intra_doc_links"
+    "-Arustdoc::private_intra_doc_links",
+
+    # Panics
+    "-Wclippy::fallible_impl_from",
 
     # TODOs:
     # `cargo fix` might help do these fixes,
--- a/zebra-chain/src/work/u256.rs
+++ b/zebra-chain/src/work/u256.rs
@@ -3,6 +3,7 @@
 // it raises a lot of issues in the macro.
 #![allow(clippy::all)]
 #![allow(clippy::range_plus_one)]
+#![allow(clippy::fallible_impl_from)]
 
 use uint::construct_uint;
 


zebra-network/src/protocol/external/message.rs
```diff
@@ -352,7 +352,11 @@ where
             // use specific varieties of `RejectReason`.
             ccode: RejectReason::Other,
 
-            reason: e.source().unwrap().to_string(),
+            reason: if let Some(reason) = e.source() {
+                reason.to_string()
+            } else {
+                String::from("No reason detail")
```
teor2345: Since this is a bitcoin network message, we might want to just send an empty string here.

Can you check what `zcashd` does, or what the bitcoin reference says?

zebra-chain/src/orchard/keys.rs
```diff
@@ -564,18 +580,25 @@ impl From<FullViewingKey> for [u8; 96] {
...
+                    nullifier_deriving_key: NullifierDerivingKey::from(sk),
+                    ivk_commit_randomness: IvkCommitRandomness::from(sk),
+                })
+            } else {
+                Err("Invalid spend validation key")
+            }
+        } else {
+            Err("Invalid spend authorizing key")
         }
     }
```
teor2345: Optional: Can this be simplified by using `?` to return the original errors?

teor2345 (commented):
I'd like to check the changes to the network protocol, everything else is optional.

I did not review the cryptography in detail.

teor2345:
I'm updating this PR, so I can see if the CI failures are from an old `main` branch, or if they are new in this PR.

teor2345:
@mergifyio update

zebra-chain/src/orchard/keys.rs
oxarbitrage: Nice one, i did this at https://github.com/ZcashFoundation/zebra/pull/4609/commits/8948f22aff823ff0b2738d1c80be32b545c7c796

zebra-network/src/protocol/external/message.rs
oxarbitrage: The bitcoin reference does not seems to say anything about what to do when there is no reason text: https://en.bitcoin.it/wiki/Protocol_documentation#reject

In zcash i am not sure of what they are doing either: https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/main.cpp#L7334-L7338

I think an empty string can be a bit better than what we have now however i don't have any proof for one side or the other.

oxarbitrage: Changed to empty string in https://github.com/ZcashFoundation/zebra/pull/4609/commits/8a571e5791e77710c067dff8acc3275c05f11fe0

zebra-chain/src/orchard/keys.rs
```diff
@@ -294,23 +302,31 @@ pub struct SpendValidatingKey(pub(crate) redpallas::VerificationKey<SpendAuth>);
...
-
 impl From<SpendValidatingKey> for [u8; 32] {
     fn from(ak: SpendValidatingKey) -> [u8; 32] {
         ak.0.into()
     }
 }
 
-impl From<SpendAuthorizingKey> for SpendValidatingKey {
-    fn from(ask: SpendAuthorizingKey) -> Self {
-        let sk = redpallas::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).unwrap();
```
conradoplg, line 311: This `unwrap()` will never fail, because `ask`, a scalar, is converted to a byte array and then converted back to a scalar inside `redpallas::SigningKey::<SpendAuth>::try_from`.

Which would make the fix to be: changing it to `expect()` with the reason why it's OK, and manually allowing it here if it still triggers the lint.

I realize this will require changing the entire PR (though it will be much simpler), sorry about that :sweat:

conradoplg (dismissed):

zebra-chain/src/orchard/keys.rs
oxarbitrage: What is the downside of leaving this as a try_from even if it will never fail ?

conradoplg: It just requires more code that is not strictly required. It also makes more confusing to use, since users of the crate will have to handle the error and they won't know that the error will never be returned in the first place.

oxarbitrage: It seems that `redpallas::SigningKey::<SpendAuth>::try_from` should be a `redpallas::SigningKey::<SpendAuth>::from` if it can't never fail. Do you think we should fix this there ?

oxarbitrage: Anyway, maybe for another pull request. Ill add a `expect` as suggested.

conradoplg: It can fail, because not every 32-byte array represents a valid scalar. It's just that this particular 32-byte array we know must be the valid representation of a scalar because it was just converted from one (with `<[u8; 32]>::from(ask)`), thus we know this particular call to `try_from` will not fail. (We are basically converting scalar -> byte array -> scalar)

oxarbitrage: Ok, understood, thank you for the details.

zebra-chain/src/sapling/keys.rs
```diff
@@ -505,10 +505,15 @@ impl From<AuthorizingKey> for [u8; 32] {
     }
 }
 
-impl From<SpendAuthorizingKey> for AuthorizingKey {
-    fn from(ask: SpendAuthorizingKey) -> Self {
-        let sk = redjubjub::SigningKey::<SpendAuth>::try_from(<[u8; 32]>::from(ask)).unwrap();
```
oxarbitrage, line 510: @conradoplg do you think this is the same as https://github.com/ZcashFoundation/zebra/pull/4609/files#r899478519 ?

conradoplg: Yep, it's the same

zebra-network/src/protocol/external/message.rs
teor2345: Thanks for checking this, I think it's fine for now.

zebra-chain/src/sapling/keys.rs
oxarbitrage: So i think this is what we will end up doing, let me know how do you see it:

- Leave orchard and sapling files as they are but add `expect` instead of unwrap in those 2 lines that we discussed. This will also leave unchanged the address and the test files.
- Allow the lint in the above files.
- We can keep the lint active in protocol messages.
- We can keep the exception in u256.

So basically we will be introducing a lint but allowing almost all the code that the lint forbid. Will this worth ?

teor2345: I think the lint is still useful, because it will make us carefully review any panicking `From` impls that we add to Zebra.

(Sometimes the reviews will show that they are ok, and we'll add a comment and allow the lint. That is also a good outcome.)

oxarbitrage: Sounds good.

conradoplg: Agreed!

oxarbitrage:
Force pushed, this is now just https://github.com/ZcashFoundation/zebra/pull/4609/commits/c0c30edf57111876c6ed23c8cb3a84e316a277a8

teor2345 (approved):
Thanks, looks good!

teor2345:
@mergifyio update

PR #4561, opened 2022-06-02
fix(doc): Fix various doc warnings, part 2

upbqdn:
## Motivation

`cargo doc` currently produces a lot of warnings. We would like to incorporate cargo doc into our CI, so we should fix these warnings.

Previous PRs:
- #4514
- #4494

## Solution

I manually fixed various issues that were produced by `cargo doc --document-private-items`.

This PR doesn't fix all warnings yet because it would be too cumbersome to review it. There will be probably two more PRs that will deal with the remaining warnings.

PR #4611, opened 2022-06-14
fix(doc): Fix various doc warnings, part 3

### LINKED ISSUE #2892: Go over all `cargo doc` warnings
When running `cargo doc` documentation for the zebra project is generated however there are a lot of warnings that we should go over and try to fix. Most of them are suggestions to use auto link and missing paths. This is low priority as the the documentation is currently generated with this small issues.

upbqdn:
## Motivation

This PR wraps up the process of fixing the warnings produced by `cargo doc`.

Previous PRs:
- #4561
- #4514
- #4494

## Solution

This PR fixes all remaining unresolved links. However, some public links point to private items, so this PR also allows pointing to private items by adding the inner attribute
```rust
 #![allow(rustdoc::private_intra_doc_links)]
```
to all crates that require it. The docs must then be built with the `--document-private-items` flag in order for the attribute to apply.

Some warnings are produced due to duplicates in dependencies. For this reason, it is necessary to build the docs with the `--no-deps` flag. We can omit this flag once we stop skipping the crates listed here https://github.com/ZcashFoundation/zebra/blob/fix-doc-warnings/deny.toml#L26.

Based on the description above, the command to build the docs becomes:
```bash
```
The PR also adds line-wraps on doc comments that it fixes.

---

Closes #2892.

### Reviewer Checklist

  - [ ] There are no warnings produced by running the command specified above.


upbqdn:
I labeled this PR as high-priority so that it gets merged quickly and other PRs don't introduce new warnings.
Also, it will be better to review this PR by looking at commits that were created on Jun 12 onward. I was reusing the old branch from the previous PRs instead of forking a new branch off `main` for each PR. The consequence is that commits that were already merged into `main` show up in this PR. I was doing it this way because I didn't want to see warnings that were fixed but weren't merged into `main` yet. I wonder if there's a better way to something like this.

oxarbitrage:
For the commits one option is to create a new branch locally based on `main`, cherry pick, and then force push. Should be something as:

```
git clone https://github.com/ZcashFoundation/zebra
cd zebra
git checkout -b fix-doc-warnings

# cherry pick only the commits that are for this PR
git cherry-pick [commit hash]
...

git push -f origin fix-doc-warnings
```

Start fro the oldest commit that need to be included and then go down the list. Don't cherry pick any unrelated commit or merges to main.
I think there should be no conflicts but if there are, the git command output will tell you what to do after fixed.

teor2345 (commented):
> This PR fixes all remaining unresolved links. However, some public links point to private items, so this PR also allows pointing to private items by adding the inner attribute
> 
> ```rust
>  #![allow(rustdoc::private_intra_doc_links)]
> ```

We use a cargo config for lints, does it work for `rustdoc` lints?
https://github.com/ZcashFoundation/zebra/blob/main/.cargo/config.toml

teor2345 (commented):
Looks good, thanks for doing all this work!

Now we just have to get it in CI, so you're not the one fixing them up all the time 😅

teor2345:
@mergifyio update

gustavovalverde (approved):

upbqdn:
> We use a cargo config for lints, does it work for rustdoc lints?

It does. I opened #4627 to fix that.

PR #4616, opened 2022-06-15
ci(Mergify): configuration update

dconnolly:
This change has been made by @dconnolly from https://mergify.com config editor.


dconnolly:
Whoops

PR #4618, opened 2022-06-15
cleanup(net): Group PeerSet fields into themes

teor2345:
## Motivation

I did this work as a cleanup for some peer set changes, but I don't know if we need them any more.

So I just want to merge the cleanup separately.

## Solution

- Group peer set fields into themes
- Add missing module comments

## Review

Anyone can review this optional PR.

### Reviewer Checklist

  - [ ] Existing tests pass

### DIFF
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -164,14 +164,23 @@ where
     D::Error: Into<BoxError>,
     C: ChainTip,
 {
+    // Peer Tracking: New Peers
+    //
     /// Provides new and deleted peer [`Change`]s to the peer set,
     /// via the [`Discover`] trait implementation.
     discover: D,
 
+    /// A channel that asks the peer crawler task to connect to more peers.
+    demand_signal: mpsc::Sender<MorePeers>,
+
+    // Peer Tracking: Ready Peers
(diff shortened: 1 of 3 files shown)


teor2345:
@mergifyio update

zebra-network/src/peer_set/set.rs
```diff
@@ -170,14 +170,23 @@ where
     D::Error: Into<BoxError>,
     C: ChainTip,
 {
+    // Peer Tracking: New Peers
+    //
```
oxarbitrage, line 174:
```suggestion
    // Peer Tracking: New Peers

```
I am not sure if this is better but i think it can be a bit confusing to realize about the themes as it is now. Id this is changed then it should be done for all the other section headers as well.

oxarbitrage (approved):
It looks good, i just made a small comment about the header text of each group but it is optional.

zebra-network/src/peer_set/set.rs
teor2345: Rustfmt will remove the empty line if we do this change 😅

oxarbitrage: Sorry, I should have tried it.

PR #4627, opened 2022-06-15
Use `config.toml` instead of inner attributes for private links

upbqdn:
## Motivation & Solution

This PR addresses https://github.com/ZcashFoundation/zebra/pull/4611#pullrequestreview-1006677722.

The PR removes the inner attributes `#![allow(rustdoc::private_intra_doc_links)]` from the `lib.rs` files of all crates that contained this attribute, and adds a corresponding rustflag to `.cargo/config.toml` instead.

### Reviewer Checklist

- [ ] Running `cargo doc --document-private-items --all-features --no-deps` doesn't produce any warnnings.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -49,6 +49,9 @@ rustflags = [
     # Code styles we want to accept
     "-Aclippy::try_err",
 
+    # Links in public docs can point to private items.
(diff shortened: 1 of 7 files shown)


teor2345 (approved):
Looks good to me, but I didn't test it.

(We'll test this in CI soon.)

PR #4631, opened 2022-06-16
fix(state): only delete old database directories inside the cache directory

teor2345:
## Motivation

Users might have a weird Zebra state setup, where they have symlinked an outdated database path to another directory containing important files.

In this edge case, we want to skip deleting that directory.

Depends-On: #4586

## Solution

- Check that the canonical path of the outdated directory is inside the canonical path of the cache directory

Related Fixes:
- Simplify version parsing
- Use spawn_blocking to launch the task on a separate thread
- Abort the cleanup task when Zebra exits

Related Cleanups:
- Split and rename some functions, so it's easy to see that we are deleting the right directory
- Add references to make ownership easier

## Review

@oxarbitrage wrote the original PR #4586.

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

This PR doesn't have any tests.

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -184,15 +184,7 @@ fn parse_dir_name(entry: DirEntry) -> Option<String> {
 ///
 /// Returns `None` if parsing fails, or the directory name is not in the expected format.
 fn parse_version_number(dir_name: String) -> Option<u32> {
-    if dir_name.len() >= 2 && dir_name.starts_with('v') {
-        if let Some(potential_version_number) = dir_name.strip_prefix('v') {
-            return Some(
-                potential_version_number
-                    .to_string()
-                    .parse()
-                    .unwrap_or(u32::MAX),
-            );
-        }
-    }
-    None
+    dir_name
+        .strip_prefix('v')
+        .and_then(|version| version.parse().ok())
 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -127,17 +127,16 @@ impl Default for Config {
 ///
 /// Iterate over the files and directories in the databases folder and delete if:
 /// - The state directory exists.
-/// - The entry is a directory.
-/// - The directory name has a length of at least 2 characters.
-/// - The directory name has a prefix `v`.
-/// - The directory name without the prefix can be parsed as an unsigned number.
+/// - It has one or more sub-directories.
+/// - The sub-directory name has a prefix `v`.
+/// - The sub-directory name without the prefix can be parsed as a [`u32`].
 /// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
 pub async fn check_and_delete_old_databases(config: Config) {
     if config.ephemeral || !config.delete_old_database {
         return;
     }
 
-    info!("checking old database versions");
+    info!("checking for old database versions");
     let cache_dir = config.cache_dir.join("state");
     if let Some(read_dir) = read_dir(cache_dir.clone()) {
         for entry in read_dir.flatten() {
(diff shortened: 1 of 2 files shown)


zebra-state/src/config.rs
```diff
@@ -123,48 +127,105 @@ impl Default for Config {
...
 /// - The entry is a directory.
-/// - The directory name has a length of at least 2 characters.
 /// - The directory name has a prefix `v`.
 /// - The directory name without the prefix can be parsed as an unsigned number.
 /// - The parsed number is lower than the hardcoded `DATABASE_FORMAT_VERSION`.
-pub async fn check_and_delete_old_databases(config: Config) {
+pub fn check_and_delete_old_databases(config: Config) -> JoinHandle<()> {
+    let current_span = Span::current();
+
+    spawn_blocking(move || {
```
teor2345, line 144: This runs blocking filesystem operations in their own thread.

zebra-state/src/config.rs
```diff
@@ -123,48 +127,105 @@ impl Default for Config {
...
-                        let delete_path = cache_dir.join(dir_name);
-                        if remove_dir_all(delete_path.clone()).is_ok() {
-                            info!("deleted outdated state directory {:?}", delete_path);
-                        }
-                    }
-                }
+    info!("checking for old database versions");
+
+    let state_dir = config.cache_dir.join("state");
+    if let Some(state_dir) = read_dir(&state_dir) {
```
teor2345, line 163: I was a bit confused by the different paths both named `cache_dir`, so I renamed one of them.

zebra-state/src/config.rs
```diff
@@ -123,48 +127,105 @@ impl Default for Config {
...
     }
 }
 
-fn read_dir(cache_dir: PathBuf) -> Option<ReadDir> {
-    if cache_dir.exists() {
-        if let Ok(read_dir) = cache_dir.read_dir() {
+/// Return a `ReadDir` for `dir`, after checking that `dir` exists and can be read.
+///
+/// Returns `None` if any operation fails.
+fn read_dir(dir: &Path) -> Option<ReadDir> {
```
teor2345, line 177: This actually reads `state_dir` now, so I renamed the argument.

zebra-state/src/config.rs
```diff
@@ -175,16 +236,11 @@ fn parse_dir_name(entry: DirEntry) -> Option<String> {
...
-        }
-    }
-    None
+/// Parse the state version number from `dir_name`.
+///
+/// Returns `None` if parsing fails, or the directory name is not in the expected format.
+fn parse_version_number(dir_name: &str) -> Option<u32> {
+    dir_name
+        .strip_prefix('v')
+        .and_then(|version| version.parse().ok())
```
teor2345, line 245: My initial suggestion was wrong, `parse` is a `str` method, not a `u32` method.

zebra-state/src/config.rs
```diff
@@ -123,48 +127,105 @@ impl Default for Config {
...
+
+    if !outdated_path.starts_with(&cache_path) {
+        info!(
+            skipped_path = ?outdated_path,
+            ?cache_path,
+            "skipped cleanup of outdated state directory: state is outside cache directory",
+        );
+
+        return None;
+    }
```
teor2345, line 219: This check makes sure we only delete things inside the cache directory.
I also flattened out all the `if` statements, so it's obvious what each check is doing.

Do you think this check should be in a separate function?

zebra-state/src/config.rs
oxarbitrage: Yea, the join was in the start.rs file in a previous version but had to move it when spanwing in a new task. Forgot to rename one of those variables.

zebra-state/src/config.rs
oxarbitrage: I think this might be a good idea but i am going to leave it as it is here and try in the original PR with this code merged. Is optional.

oxarbitrage (approved):

PR #4635, opened 2022-06-16
ci(docs): test `cargo doc` warnings and errors

### LINKED ISSUE #4524: Add `cargo doc` warnings/errors to CI
## Motivation

After #2892 is done, we'll want to make sure we don't introduce new errors. To do that, we should test it in CI.

### Designs

- [ ] Add `cargo doc` to CI to make sure no errors/warnings are thrown
- [ ] Check for warnings and publish docs with
  ```bash
  ```

gustavovalverde:
## Motivation

We want to make sure we don't introduce new errors in documentation. To do that, we have to test with `cargo doc` in CI.

Fixes #4524 

## Solution

- Add a new job to the `Lint` workflow including `cargo doc`
- Use the following arguments for cargo doc: `--no-deps --document-private-items --all-features`

## Review

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -95,7 +95,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: actions-rs/toolchain@v1.0.1
+      - uses: actions-rs/toolchain@v1.0.6
         with:
           toolchain: stable
           components: rustfmt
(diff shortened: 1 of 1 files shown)


teor2345 (commented):
Do we want to require the doc build to pass CI before we merge PRs?

I think the answer is "yes, we should fix doc issues before PRs merge".

(Have we tested that broken docs fail CI?)

upbqdn:
Looks good to me.

> Do we want to require the doc build to pass CI before we merge PRs?

My answer is also yes.

Suggestion: would it be worth splitting `changed-files.outputs.rust` into
- `changed-files.outputs.rust-src` and
- `changed-files.outputs.rust-config`,

to add more granularity?

Some jobs, such as `docs`, don't have to be triggered when we edit Rust's config files.

gustavovalverde:
@upbqdn I'm not sure is worth splitting, as if the config files get changed other workflows which take longer also get triggered. But we can surely improve this later (globally)

teor2345 (approved):
We want this merged soon, so we don't accidentally introduce doc bugs

PR #4636, opened 2022-06-17
imp(docs): manage`cargo-mdbook` as a GitHub action

### LINKED ISSUE #4396: Manage `cargo-mdbook` as a GitHub action
## Motivation

Hardcoding versions (or just 'latest') of build tools with no notifications of updates from Dependabot is not ideal.

We should use https://github.com/peaceiris/actions-mdbook  instead

gustavovalverde:
## Motivation

Hardcoding versions (or just 'latest') of build tools with no notifications of updates from Dependabot is not ideal.

Fixes #4396

## Solution

- Use https://github.com/peaceiris/actions-mdbook instead

## Review

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,10 +37,10 @@ jobs:
 
       - uses: Swatinem/rust-cache@v1
 
-      - name: Install mdbook
-        run: |
-          cargo install mdbook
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Looks great, thanks!

PR #4637, opened 2022-06-17
fix(build): add `beta` rust to CI

### LINKED ISSUE #2541: Re-enable beta Rust tests in CI
## Motivation

We want to test Zebra with the next Rust release candidate, so we can report any compilation bugs and get them fixed.

On 29 July, Zebra's beta Rust tests started failing, due to a beta rust bug (https://github.com/rust-lang/rust/pull/87483).

But the stable CI tests and beta zebra-chain build were fine.

## Actions

We disabled the beta Rust tests in PR #2542.

We should re-enable beta Rust:
* after the fix is backported to beta
    * https://github.com/rust-lang/rust/pull/87483
    * https://github.com/rust-lang/rust/pull/88190
* do we need to wait for a beta point release, or are we using daily beta builds?

gustavovalverde:
## Motivation

On 29 July 2021, Zebra's tests started failing on beta Rust (https://github.com/rust-lang/rust/pull/87483). This Rust PR fixed the test build failure we were seeing: https://github.com/rust-lang/rust/pull/87483 and it was also backported.

Fixes #2541

Depends-On: #4668

## Solution

- Re-enable the Rust beta tests in the CI matrix.

## Review

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -153,47 +153,6 @@ jobs:
           # Note: this only runs the zebrad acceptance tests, because re-running all the test binaries is slow on Windows
           args: --verbose --package zebrad --test acceptance -- --nocapture --include-ignored sync_large_checkpoints_
 
-  test-fake-activation-heights:
-    name: Test ${{ matrix.rust }} zebra-state with fake activation heights on ubuntu-latest
-    timeout-minutes: 60
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        rust: [stable, beta]
-
-    steps:
-      - uses: actions/checkout@v3.0.2
-        with:
-          persist-credentials: false
-
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          profile: minimal
-          override: true
-
-      - uses: Swatinem/rust-cache@v1
-
-      - name: cargo fetch
-        uses: actions-rs/cargo@v1.0.3
-        with:
-          command: fetch
-
-      # This test changes zebra-chain's activation heights,
-      # which can recompile all the Zebra crates,
-      # so we want its build products to be cached separately.
-      #
-      # Also, we don't want to accidentally use the fake heights in other tests.
-      - name: Run tests with fake activation heights
-        uses: actions-rs/cargo@v1.0.3
-        env:
-          TEST_FAKE_ACTIVATION_HEIGHTS: ''
-        with:
-          command: test
-          # Note: this only runs the zebra-state crate tests,
-          # because re-running all the test binaries can be slow
-          args: --verbose --package zebra-state --lib -- --nocapture --include-ignored with_fake_activation_heights
-
   build-chain-no-features:
     name: Build ${{ matrix.rust }} zebra-chain w/o features on ubuntu-latest
     timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/continous-integration-os.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       matrix:
         # TODO: Windows was removed for now, see https://github.com/ZcashFoundation/zebra/issues/3801
         os: [ubuntu-latest, macos-latest]
-        rust: [stable]
+        rust: [stable, beta]
```
teor2345, line 56: I'd like to only run beta Rust on Linux:
- the concurrent macOS runner limit is much lower than the Linux limit
- macOS is slower than Linux
- macOS is a second-tier Zebra support platform
- there shouldn't be much difference between the macOS and Linux build or test results

teor2345 (changes requested):
I think running beta Rust on macOS will be slow, and it seems pretty redundant

teor2345 (commented):
Do we want to require the beta tests to pass before we merge PRs?

I think the answer is "yes, but we should disable them and open a ticket if they fail".

gustavovalverde:
> Do we want to require the beta tests to pass before we merge PRs?

Yes. This shouldn't hurt

.github/workflows/continous-integration-os.yml
gustavovalverde: Fixed by [d79e10f](https://github.com/ZcashFoundation/zebra/pull/4637/commits/d79e10f180a6611a3d07298a621beb51283127e2)

teor2345 (dismissed):
> > Do we want to require the beta tests to pass before we merge PRs?
> 
> Yes. This shouldn't hurt

Can you please add patch jobs to this PR, so we can require these jobs in GitHub branch protection settings?

teor2345:
@mergifyio update

gustavovalverde:
> Can you please add patch jobs to this PR, so we can require these jobs in GitHub branch protection settings?

A patch already exists for this job, we're just adding extra arguments to it. This new arguments are already available as a required check.

.github/workflows/continous-integration-os.yml
```diff
@@ -152,7 +159,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable]
+        rust: [stable, beta]
```
teor2345: The fake activation heights test results shouldn't depend on the OS and Rust release.

Can we remove this job entirely, and just use "Test with fake activation heights" (Docker) instead?
(I removed "Test stable zebra-state with fake activation heights on ubuntu-latest" from the branch protection rules.)

teor2345:
> > Can you please add patch jobs to this PR, so we can require these jobs in GitHub branch protection settings?
> 
> A patch already exists for this job, we're just adding extra arguments to it. This new arguments are already available as a required check.

If we want to require that "Test beta on ubuntu-latest" passes, we need a patch job for it.

I opened PR #4668 to add these patch jobs, and fix up some we missed from PR #4640.
(In general, I'd like to keep patch jobs in sync with real jobs, so we can quickly require jobs if we need to.)

.github/workflows/continous-integration-os.yml
teor2345: I'm just going to go ahead and make this change, so we can merge.

I'm happy to revert later if you'd like.

teor2345:
@mergifyio update

teor2345 (approved):
Looks good, I can do the branch protection rules after it merges

Issue #4150, opened 2022-04-20
Update column family names to match Zebra's database design


teor2345:
## Motivation

To speed up development and minimise diffs, we kept the same column family names in Zebra's database.

But we should update them to match the design. This requires a state version increment.

### Designs

The column family names should match:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/rfcs/0005-state-updates.md#rocksdb-data-structures

teor2345:
PR #4355 contains another state version increment which fixes a bug.

upbqdn:
This issue is resolved by PR #4639. I used the string `Addresses #4150` in the description of the PR, but GitHub apparently didn't recognize it and didn't automatically close this issue. It turns out one has to use one of these keywords in order to trigger the automatic closing:
```
close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved
```

PR #4639, opened 2022-06-17
fix(state): Update column family names to match Zebra's database design

upbqdn:
## Motivation

The documentation at https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/rfcs/0005-state-updates.md#rocksdb-data-structures does not currently match the source code. This PR makes the source code consistent with the docs.

Depends-On: #4586
Depends-On: #4669
Depends-On: #4704

## Solution
This PR:
- Renames all column families that didn't match the docs so that they match the docs. It also renames some variables in the non-finalized state to match the new names.
- Reorders the column families so that the order matches the docs.
- Updates `struct Chain` in the docs for the non-finalized state.
- Increments `DATABASE_FORMAT_VERSION` to 25.
- Removes obsolete docs.
- Deletes unused snapshots.

Addresses #4150.

## Review

I have not ran all tests yet, and some tests were failing, so I'm labelling this PR as a draft until I make the tests pass.

### Reviewer Checklist


teor2345:
We want to use this state version change to test old database deletion in PR #4586, so I've marked that as a dependency.

book/src/dev/rfcs/0004-asynchronous-script-verification.md
```diff
@@ -426,10 +426,12 @@ ancestors are downloaded.
...
+<!-- TODO: address the three items below since #2455 is closed. -->
+
 - Should we stop storing heights for non-coinbase UTXOs? (#2455)
 
 - Should we avoid storing any extra data for UTXOs, and just lookup the coinbase
-  flag and height using `outpoint.hash` and `tx_by_hash`? (#2455)
+  flag and height using `outpoint.hash` and `tx_loc_by_hash`? (#2455)
 
 - The maturity check can be skipped for UTXOs from the finalized state,
 because Zebra only finalizes mature UTXOs. We could implement this
```
teor2345: I think we can delete all these.

We only store `transparent::Output`s in `utxo_by_out_loc`, so the height and extra data changes are either done or obsolete:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/rfcs/0005-state-updates.md#rocksdb-data-structures

The maturity check can't be skipped, because:
1. The non-finalized state starts at less than 100 blocks from the finalized tip, and
2. Zebra can roll back the non-finalized state to a tip that is less than 100 blocks from the finalized tip

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -368,27 +368,24 @@ impl DiskDb {
...
-            //rocksdb::ColumnFamilyDescriptor::new("tx_by_transparent_addr_loc", db_options.clone()),
-            // TODO: rename to utxo_by_out_loc (#3952)
-            rocksdb::ColumnFamilyDescriptor::new("utxo_by_outpoint", db_options.clone()),
             rocksdb::ColumnFamilyDescriptor::new(
-                "utxo_loc_by_transparent_addr_loc",
+                "tx_loc_by_transparent_addr_loc",
                 db_options.clone(),
             ),
+            // TODO: #3951
+            //rocksdb::ColumnFamilyDescriptor::new("tx_by_transparent_addr_loc", db_options.clone()),
```
teor2345: I think this TODO is also done or obsolete.
```suggestion
```

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/block_header_by_height_raw_data@mainnet_0.snap
```diff
@@ -0,0 +1,11 @@
+---
```
teor2345, line 1: Can we delete the unused snapshot files as well?

If you re-run the tests with:
```sh
cargo insta test --review --delete-unreferenced-snapshots
```

That should delete the old snapshots that we aren't using any more, and turn the git diff into a move.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/empty_column_families@mainnet_0.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: zebra-state/src/service/finalized_state/disk_format/tests/snapshot.rs
+assertion_line: 154
```
teor2345, line 3: Nit: If you update your version of `cargo insta`, it won't add the assertion line any more.

teor2345 (commented):
These all look pretty good!

I have a few suggestions for cleaning things up a bit.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/block_header_by_height_raw_data@mainnet_0.snap
upbqdn: I ran the command, and it deleted unused snapshots. I committed the deletions at d99d7bfb4a507cff9fc808c8871835e276f1a01f. There were no moves though.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/empty_column_families@mainnet_0.snap
upbqdn: I updated `cargo insta`, then ran `cargo insta test --review`, but nothing changed. Should I remove the lines by hand?

upbqdn:
I updated the description of the PR to match the newly added commits. I'll wait for PR #4586, and then increment `DATABASE_FORMAT_VERSION` to 26.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/block_header_by_height_raw_data@mainnet_0.snap
teor2345: In the full PR diff, git shows those files as moved, because they are very similar to the deleted & added files.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/empty_column_families@mainnet_0.snap
teor2345: That's weird, `cargo insta` versions 0.13 and later should stop adding those lines:
https://github.com/mitsuhiko/insta/blob/master/CHANGELOG.md#1130

It doesn't really matter either way, I don't think we'll change these files for a while.

teor2345:
> I updated the description of the PR to match the newly added commits. I'll wait for PR #4586, and then increment `DATABASE_FORMAT_VERSION` to 26.

PR #4586 doesn't change the database format, it just changes how we deal with outdated versions that are left behind on disk.

So this PR can stay at version 25.

teor2345 (approved):
PR #4586 is about to merge, so we can merge this one next if you want.

teor2345:
@mergifyio update

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/block_header_by_height_raw_data@mainnet_0.snap
upbqdn: I see now.

teor2345:
> No cached state disk available
> Expected zebrad-cache-(branch)-[0-9a-f]+-v25-mainnet-checkpoint

https://github.com/ZcashFoundation/zebra/runs/6991264216?check_suite_focus=true#step:6:97

It looks like we skipped the "create state image" job, but we should have run it when the state version was updated:
https://github.com/ZcashFoundation/zebra/runs/6991230027?check_suite_focus=true

I'll see if I can fix this in another PR.

teor2345:
We haven't been creating cached state disks for about a month, I think that's fixed by:
- #4669

upbqdn:
@mergifyio update.

upbqdn:
@Mergifyio update

PR #1365, opened 2020-11-24
Fix individual crate compilation failures

teor2345:
## Motivation

Some Zebra crates don't compile individually due to missing features in their dependencies. 

## Solution

Add those features to each crate's dependency list.

## Review

@hdevalence can review this change whenever he likes.

It would be nice to get it fixed before the first alpha.

## Follow Up Work

Detect these kind of dependency bugs in CI #1364

PR #4640, opened 2022-06-17
feat(ci): build each crate individually

### LINKED ISSUE #1364: Add a CI job which compiles each Zebra crate individually
## Motivation

Sometimes we add code that needs a feature to a Zebra crate, then activate that feature in `zebrad`. But we forget to activate the feature in the crate that actually contains the code.

For an example, see PR #1365, where individual crates use features in their dependencies, but don't declare those features in their `Cargo.toml`.

## Design

- [ ] Build each Zebra crate individually in a CI job, including Zebra's custom Tower crates
- [ ] Fail the PR if any build fails.

We should add a new CI job, so that we don't slow down existing jobs.

Example script:
```sh
for crate in tower* zebra*; do
    pushd "$crate"
    cargo build
    popd
done
```

## Related Work

- #4550

## Alternatives

<details>

Continue to occasionally break crates when they're used individually.

We can implement theses checks using `cargo-hack --workspace`:
- Build all targets (library, binary, example, test, and benchmark)
- Build the `--feature-powerset`

</details>

## Follow Up Tasks

<details>

- [ ] Also test if the crate is packaged succesfully (`cargo package`)
- [ ] Also test if it compiles/packages with the MSRV (and specify one if it doesn't)

</details>

### LINKED ISSUE #4550: Add all-features and no-default-features builds to CI
## Motivation

We have added a bunch of features to Zebra, and we want to make sure we don't accidentally break them.

### Designs

Build each Zebra crate with:
- `--all-features`
- `--no-default-features`

Fail the PR if any build fails. There's no need to run tests.

Run `cargo deny check bans` with:
- `--all-features`
- `--no-default-features`

Fail the PR if any check fails.

## Related Work

- #1364

gustavovalverde:
## Motivation

Sometimes we add code that needs a feature to a Zebra crate, then activate that feature in `zebrad`. But we forget to activate the feature in the crate that actually contains the code.

For an example, see PR #1365, where individual crates use features in their dependencies, but don't declare those features in their `Cargo.toml`.

Fixes #1364
Fixes #4550

### Designs

- Build each Zebra crate individually in a CI job, including Zebra's custom Tower crates
- Fail the PR if any build fails.

## Solution

- Create a specific workflow to accomplish this task
- Use a matrix strategy to have a separate job for each crate, with individual logs
- Keep a `fail-fast` strategy, as we don't want to keep consuming minutes if a crate build failed

## Review

Not ready

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

It would be better to generate the crate names automatically, but this task would need further analysis to be able to generate a file with the needed GHA format for a matrix.


.github/workflows/build-crates-individually.yml
```diff
@@ -0,0 +1,103 @@
...
+    branches:
+      - main
+    paths:
+      # production code and test code
+      - '**/*.rs'
+      # dependencies
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      # workflow definitions
+      - '.github/workflows/build-crates-individually.yml'
```
teor2345, line 15: Changes to these files can also affect the results, because they change warnings, and some code is compiled differently in individual crates:
- https://github.com/ZcashFoundation/zebra/blob/main/.cargo/config.toml
- https://github.com/ZcashFoundation/zebra/blob/main/clippy.toml

I think we might want to open a ticket to add these files to all our CI.

.github/workflows/build-crates-individually.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo $matrix | json2yaml
+
+  build:
+    name: Build ${{ matrix.crates.crate }} crate
+    needs: [ matrix, check-matrix ]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        features: ['', '--all-features', '--no-default-features']
```
teor2345: It might be faster to run these commands sequentially, so they can re-use the local cargo cache. Some Zebra crates do not have any features, and most don't have any default features.

The most efficient order starts with the smallest number of features:
```sh
cargo build --no-default-features
cargo build
cargo build --all-features
```

teor2345 (commented):
Just some things I noticed, happy to leave someone else to do the full review.

Are we going to make all these jobs required in CI?
That seems like a lot of jobs to add, we might want to think about making a final job that depends on all of them.

.github/workflows/build-crates-individually.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: actions-rs/cargo@v1.0.3
+      - id: set-matrix
+        name: Dinamically build crates JSON
+        run: |
+          TEMP_DIR=$(mktemp -d)
```
teor2345, line 60: There are a bunch of shellcheck and typo warnings here.

We can ignore the glob warnings, but the other ones should be fixed.

.github/workflows/build-crates-individually.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "$( cargo tree --depth 0 --edges no-normal,no-dev,no-build,no-proc-macro --format {p} --prefix none | cut -d ' ' -f1 | sed '/^$/d' | awk '{ printf "\"%s\",\n", $0 }' | sed '$ s/.$//')" > $TEMP_DIR/crates.txt
+          JSON_CRATES=$(cat $TEMP_DIR/crates.txt)
+          MATRIX=$((
```
teor2345:
```suggestion
          MATRIX=$($(
```

gustavovalverde: Fixed by: [c14c51d](https://github.com/ZcashFoundation/zebra/pull/4640/commits/c14c51d1e3a7e76208d9d1594bd81774e46e0eb2)

.github/workflows/build-crates-individually.yml
gustavovalverde: Fixed by: [c14c51d](https://github.com/ZcashFoundation/zebra/pull/4640/commits/c14c51d1e3a7e76208d9d1594bd81774e46e0eb2)

.github/workflows/build-crates-individually.yml
teor2345: This is now ticket #4652.

teor2345 (approved):
Looks good, thanks for this!

teor2345:
I have this list of tasks ready to go in the `main` branch protection rules, as soon as this PR merges.

teor2345 (dismissed):
Can you please add patch jobs to this PR?

We need them if we want to require these jobs in CI.

gustavovalverde:
patch file added

.github/workflows/build-crates-individually.patch.yml
```diff
@@ -0,0 +1,76 @@
...
+      - '**/Cargo.lock'
+      # workflow definitions
+      - '.github/workflows/build-crates-individually.yml'
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUST_BACKTRACE: full
+  RUST_LIB_BACKTRACE: full
+  COLORBT_SHOW_HIDDEN: '1'
+
```
teor2345: I don't think these env vars are needed, because we don't actually build anything
```suggestion
```

.github/workflows/build-crates-individually.patch.yml
```diff
@@ -0,0 +1,76 @@
...
+    needs: [ matrix ]
+    steps:
+      - run: 'echo "No job required"'
+
+  build:
+    name: Build ${{ matrix.crate }} crate
+    needs: [ matrix, check-matrix ]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
```
teor2345: These jobs should always succeed
```suggestion
```

teor2345 (commented):
I think we can delete some redundant config lines, but it doesn't really matter either way

teor2345 (approved):
Looks good, let's merge and see how we go

Issue #4576, opened 2022-06-07
See if registry tracing stack is faster than FmtSubscriber stack


teor2345:
## Motivation

After PRs #4519 and #4539, we have two alternate tracing layer stacks:
- `FmtSubscriber`
- tracing `registry`

The default is `FmtSubscriber`, but tracing `registry` is supposed to be faster.

### Designs

Check if a full sync with a tracing `registry` stack is faster.

Development:
- split the `tracing-registry` feature out of the `tokio-console` feature

Testing:
- make the `tracing-registry` feature the default
- run a full sync test
- make `FmtSubscriber` or `tracing-registry` the default, depending on which is faster

teor2345:
I don't think we can make this change, because the filters in the registry stack are always showing trace and debug logs, but only for some tasks.

I spent a few hours looking for this bug, and I couldn't work out the cause.

This might be a bug in the tracing registry, or it might be an obscure bug in Zebra.

PR #4660, opened 2022-06-21
fix(log): Add some missing tracing spans

teor2345:
## Motivation

Some zebra-network tasks are missing the global tracing span.

## Review

Anyone can review this low-priority PR.

### Reviewer Checklist

  - [x] Address book and next peer tasks have global spans

## Follow Up Work

This might be related to the bug blocking ticket #4576, but there were also spans missing in `try_to_sync`.


upbqdn (approved):
All looks good. I checked that the address book and next peer tasks have global spans.

teor2345:
@mergifyio update

PR #4662, opened 2022-06-21
fix(panic): Reduce the minimum lookahead limit, and fix a config panic

teor2345:
## Motivation

We're seeing a lot of Orchard actions on the network, so we might want to reduce the concurrency and lookahead limits.

## Solution

- Reduce the minimum lookahead limit to 400 blocks (a single checkpoint)
- If the user configures a limit that's too low, don't panic, just warn and fix it

## Review

This might help fix some of the syncing issues, so it's a high priority.

### Reviewer Checklist

  - [ ] `lookahead_limit = 400` works in the config

### DIFF
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -57,8 +57,8 @@ const BLOCK_DOWNLOAD_RETRY_LIMIT: usize = 3;
 
 /// A lower bound on the user-specified lookahead limit.
 ///
-/// Set to two checkpoint intervals, so that we're sure that the lookahead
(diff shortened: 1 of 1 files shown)


oxarbitrage (commented):
I think the change looks good but i am unsure if it will help with the recent issues, as far as i know by what was reported, Zebra is not making more progress near the tip but not panicking. 
Did you saw this panic during sync recently ?

oxarbitrage (approved):
Ok, nevermind, i found some more context in discord: https://discord.com/channels/676527656170160146/716086297210650634/988679758877446147

Thanks.

Issue #4600, opened 2022-06-13
ci: re-enable zebra & lightwalletd status checks


gustavovalverde:
## Motivation

We've recently disabled most lighwalletd CI tests for as required status checks, with recent changes on CI we should periodically check if this checks can be re-enabled, so we don't merge faulty code

### Designs

Re-enable the following checks:
- [x] `Zebra checkpoint / Run sync-to-checkpoint test`
- [x] `Zebra tip / Run full-sync-to-tip test`
- [x] `lightwalletd tip / Run lwd-full-sync test`
  - [x] `lightwalletd tip / Run lwd-grpc-wallet test`
  - [x] `lightwalletd tip / Run lwd-send-transactions test`
  - [x] `lightwalletd tip / Run lwd-update-sync test`

These depend on PRs:
- #4669
- #4663
- maybe #4670

Enable new checks:
- [x] `Rust doc` #4635
- [x] Build all crates individually (13 crate jobs + check matrix job) #4640 
- [x] Beta Rust #4637
- [x] `cargo deny` with `--no-default-features` and `--all-features` #4640

teor2345:
I just re-enabled all these branch protection rules.

PR #4537, opened 2022-05-30
test(grpc): `GetMempoolTx` and `GetMempoolStream` test

oxarbitrage:
## Motivation

We have this 2 calls missing from the grpc tests.

Close https://github.com/ZcashFoundation/zebra/issues/4350

Depends-On: #4663
Depends-On: #4704
Depends-On: #4848

## Solution

One of the calls seems to be not working.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

Issue #4661, opened 2022-06-21
Split full sync test into checkpoint and full validation jobs


teor2345:
## Motivation

Zebra takes around 6 hours to fully sync, and that's also the GitHub job timeout.

But if we split the sync into multiple jobs, we won't get CI failures when the sync takes a little bit longer.

### Designs

- [ ] Run a checkpoint job until Zebra finishes checkpoint syncing
- [ ] Run a full validation job until Zebra reaches the tip

PR #4663, opened 2022-06-21
fix(ci): Ignore lightwalletd hangs for now

teor2345:
## Motivation

Sometimes `lightwalletd` hangs when it gets near the tip.

We can't really fix that, so we just check that it has synced some blocks instead.

Depends-On: #4669, #4675

## Solution

- Allow "adding block to cache" as an alternative tip log, where we previously required "waiting for block"

## Review

This is pretty urgent, it's blocking some lightwalletd tests.

### Reviewer Checklist

  - [ ] CI passes

## Follow Up Tasks

- #4600 
- #4537

### DIFF
--- a/zebra-chain/src/transaction/tests/vectors.rs
+++ b/zebra-chain/src/transaction/tests/vectors.rs
@@ -838,14 +838,15 @@ fn zip143_sighash() -> Result<()> {
             Some(output) => mock_pre_v5_output_list(output, input_index.unwrap()),
             None => vec![],
         };
-        let result =
-            hex::encode(transaction.sighash(
+        let result = hex::encode(
+            transaction.sighash(
                 NetworkUpgrade::from_branch_id(test.consensus_branch_id)
                     .expect("must be a valid branch ID"),
                 HashType::from_bits(test.hash_type).expect("must be a valid HashType"),
                 &all_previous_outputs,
                 input_index,
-            ));
+            ),
+        );
         let expected = hex::encode(test.sighash);
         assert_eq!(expected, result, "test #{}: sighash does not match", i);
     }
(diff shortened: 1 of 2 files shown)


gustavovalverde:
Was reviewing to approve, but test is still failing even though it shows a matching regex
```
{"app":"lightwalletd","level":"info","msg":"Adding block to cache 1020427 000000000049429cfbc02c7d798e23e5db748310eda499977569a5484116ab46","time":"2022-06-21T07:25:34Z"}
```

But the actual regex is expecting a different output:
```
Match Regex:
   [
       "finished initial sync to chain tip, using gossiped blocks .*sync_percent.*=.*100\\.",
   ]
```

FMT is also complaining

Maybe @conradoplg can have a look at this one.

conradoplg:
@Mergifyio update

conradoplg:
> Maybe @conradoplg can have a look at this one.

That's the Zebra sync failure which was hopefully solved by #4662 . I've updated this PR to pick up the fix.

teor2345:
The sync failure in this PR should get better after #4669, and whatever fixes we need to fix the sync speed.

gustavovalverde:
I'm re-testing https://github.com/ZcashFoundation/zebra/pull/4669. Let's wait for the result, merge it to `main` and then update this PR to confirm all checks pass

teor2345:
This is blocking a bunch of work, so I made it "Critical" priority.

teor2345:
> I'm re-testing #4669. Let's wait for the result, merge it to `main` and then update this PR to confirm all checks pass

#4669 passed, I manually rebased on that branch and the latest `main`.

So the diff will contain PR #4669 as well, but the CI results will be correct.

teor2345:
This is still failing, it might need:
- all of #4661 
    - PR #4675, and
    - splitting checkpoint sync and full sync into separate jobs based on log output
- PR #4679
- a successful full sync on `main`

upbqdn:
@Mergifyio update

upbqdn (approved):

PR #4667, opened 2022-06-21
lint(clippy): add `unwrap_in_result` lint

oxarbitrage:
## Motivation

In https://github.com/ZcashFoundation/zebra/issues/2297 we want to add a few panic related lints. This PR adds `unwrap_in_result` and add a bunch of exceptions.

## Solution

I followed a few simple rules to do this:

- When `?` is possible, use it.
- Change `unwrap()` to `expect()` with a message where possible and add an exception.

The result is a very intrusive change.

## Review

This can be painful to review so i have it as a draft because i am not sure if the change will worth given the amount of exceptions we are adding.

### DIFF
--- a/zebra-chain/src/orchard/commitment.rs
+++ b/zebra-chain/src/orchard/commitment.rs
@@ -141,10 +141,11 @@ impl NoteCommitment {
 
         let rcm = CommitmentRandomness::from(note.rseed);
 
-        Some(NoteCommitment::from(
-            sinsemilla_commit(rcm.0, b"z.cash:Orchard-NoteCommit", &s)
-                .expect("valid orchard note commitment, not ⊥ "),
-        ))
+        Some(NoteCommitment::from(sinsemilla_commit(
+            rcm.0,
+            b"z.cash:Orchard-NoteCommit",
+            &s,
+        )?))
     }
 
     /// Extract the x coordinate of the note commitment.
--- a/zebra-chain/src/orchard/commitment.rs
+++ b/zebra-chain/src/orchard/commitment.rs
@@ -113,6 +113,7 @@ impl NoteCommitment {
     ///
     /// <https://zips.z.cash/protocol/nu5.pdf#concretewindowedcommit>
     #[allow(non_snake_case)]
+    #[allow(clippy::unwrap_in_result)]
     pub fn new(note: Note) -> Option<Self> {
         // s as in the argument name for WindowedPedersenCommit_r(s)
         let mut s: BitVec<u8, Lsb0> = BitVec::new();
@@ -141,11 +142,10 @@ impl NoteCommitment {
 
         let rcm = CommitmentRandomness::from(note.rseed);
 
-        Some(NoteCommitment::from(sinsemilla_commit(
-            rcm.0,
-            b"z.cash:Orchard-NoteCommit",
-            &s,
-        )?))
+        Some(NoteCommitment::from(
+            sinsemilla_commit(rcm.0, b"z.cash:Orchard-NoteCommit", &s)
+                .expect("valid orchard note commitment, not ⊥ "),
+        ))
     }
 
     /// Extract the x coordinate of the note commitment.
--- a/zebra-chain/src/sapling/address.rs
+++ b/zebra-chain/src/sapling/address.rs
@@ -73,7 +73,7 @@ impl std::str::FromStr for Address {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
                 let mut decoded_bytes = io::Cursor::new(
-                    Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"),
+                    Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32"),
                 );
 
                 let mut diversifier_bytes = [0; 11];
--- a/zebra-chain/src/sapling/address.rs
+++ b/zebra-chain/src/sapling/address.rs
@@ -48,6 +48,7 @@ impl fmt::Debug for Address {
 }
 
 impl fmt::Display for Address {
+    #[allow(clippy::unwrap_in_result)]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut bytes = io::Cursor::new(Vec::new());
 
@@ -59,17 +60,18 @@ impl fmt::Display for Address {
             _ => human_readable_parts::TESTNET,
         };
 
-        bech32::encode_to_fmt(f, hrp, bytes.get_ref().to_base32(), Variant::Bech32).unwrap()
+        bech32::encode_to_fmt(f, hrp, bytes.get_ref().to_base32(), Variant::Bech32).expect("hrp is valid")
     }
 }
 
 impl std::str::FromStr for Address {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let mut decoded_bytes = io::Cursor::new(Vec::<u8>::from_base32(&bytes).unwrap());
+                let mut decoded_bytes = io::Cursor::new(Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"));
 
                 let mut diversifier_bytes = [0; 11];
                 decoded_bytes.read_exact(&mut diversifier_bytes)?;
@@ -83,7 +85,7 @@ impl std::str::FromStr for Address {
                     },
                     diversifier: keys::Diversifier::from(diversifier_bytes),
                     transmission_key: keys::TransmissionKey::try_from(transmission_key_bytes)
-                        .unwrap(),
+                        .expect("bytes are valid transmission key"),
                 })
             }
             _ => Err(SerializationError::Parse("bech32 decoding error")),
--- a/zebra-chain/src/sapling/address.rs
+++ b/zebra-chain/src/sapling/address.rs
@@ -88,9 +88,7 @@ impl std::str::FromStr for Address {
                     },
                     diversifier: keys::Diversifier::from(diversifier_bytes),
                     transmission_key: keys::TransmissionKey::try_from(transmission_key_bytes)
-                        .map_err(|_| {
-                            SerializationError::Parse("invalid transmission key bytes")
-                        })?,
+                        .map_err(|_| SerializationError::Parse("invalid transmission key bytes"))?,
                 })
             }
             _ => Err(SerializationError::Parse("bech32 decoding error")),
--- a/zebra-chain/src/sapling/address.rs
+++ b/zebra-chain/src/sapling/address.rs
@@ -60,7 +60,8 @@ impl fmt::Display for Address {
             _ => human_readable_parts::TESTNET,
         };
 
-        bech32::encode_to_fmt(f, hrp, bytes.get_ref().to_base32(), Variant::Bech32).expect("hrp is valid")
+        bech32::encode_to_fmt(f, hrp, bytes.get_ref().to_base32(), Variant::Bech32)
+            .expect("hrp is valid")
     }
 }
 
@@ -71,7 +72,9 @@ impl std::str::FromStr for Address {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let mut decoded_bytes = io::Cursor::new(Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"));
+                let mut decoded_bytes = io::Cursor::new(
+                    Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"),
+                );
 
                 let mut diversifier_bytes = [0; 11];
                 decoded_bytes.read_exact(&mut diversifier_bytes)?;
--- a/zebra-chain/src/sapling/address.rs
+++ b/zebra-chain/src/sapling/address.rs
@@ -68,13 +68,13 @@ impl fmt::Display for Address {
 impl std::str::FromStr for Address {
     type Err = SerializationError;
 
-    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let mut decoded_bytes = io::Cursor::new(
-                    Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32"),
-                );
+                let mut decoded_bytes =
+                    io::Cursor::new(Vec::<u8>::from_base32(&bytes).map_err(|_| {
+                        SerializationError::Parse("bech32::decode guarantees valid base32")
+                    })?);
 
                 let mut diversifier_bytes = [0; 11];
                 decoded_bytes.read_exact(&mut diversifier_bytes)?;
@@ -88,7 +88,9 @@ impl std::str::FromStr for Address {
                     },
                     diversifier: keys::Diversifier::from(diversifier_bytes),
                     transmission_key: keys::TransmissionKey::try_from(transmission_key_bytes)
-                        .expect("bytes are valid transmission key"),
+                        .map_err(|_| {
+                            SerializationError::Parse("invalid transmission key bytes")
+                        })?,
                 })
             }
             _ => Err(SerializationError::Parse("bech32 decoding error")),
--- a/zebra-chain/src/sapling/keys.rs
+++ b/zebra-chain/src/sapling/keys.rs
@@ -273,7 +273,8 @@ impl FromStr for SpendingKey {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let decoded = Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32");
+                let decoded =
+                    Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32");
 
                 let mut decoded_bytes = [0u8; 32];
                 decoded_bytes[..].copy_from_slice(&decoded[0..32]);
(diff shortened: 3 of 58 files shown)


teor2345 (commented):
This does seem like a big change.

It also changes the behaviour of some of our code, so we'll need to review it carefully:
- we use `expect()` or `unwrap()` when the error is a bug we can't recover from
- we use `?` when the error is something we can retry or recover from
- when this PR changes from panics to `?`, we need to check that all those errors are recoverable

When I'm trying to decide if a big change is worth it, I ask myself:
1. What is the future benefit of this change?
2. What will it cost us in future?
3. What is the cost right now?

Here are some initial thoughts:
1. It will make us carefully review panics in code that returns `Result`
2. It makes the code harder to maintain, because we've added a bunch of lint exceptions
3. We need to draft this PR (sunk cost), carefully review it, and revise it where needed

I'd like to get another reviewer's opinion before we decide anything.

zebra-chain/src/orchard/commitment.rs
```diff
@@ -141,10 +141,11 @@ impl NoteCommitment {
...
 
-        Some(NoteCommitment::from(
-            sinsemilla_commit(rcm.0, b"z.cash:Orchard-NoteCommit", &s)
-                .expect("valid orchard note commitment, not ⊥ "),
-        ))
+        Some(NoteCommitment::from(sinsemilla_commit(
+            rcm.0,
+            b"z.cash:Orchard-NoteCommit",
+            &s,
+        )?))
```
daira: [Section 4.1.8](https://zips.z.cash/protocol/protocol.pdf#abstractcommit) of the protocol spec says:
image

It appears as though with this change, the case of an invalid diversifier (the `?` on line 127) is being conflated with a $\bot$ or 0 return from Commit<sup>ivk</sup>. ($\bot$ and 0 as outputs of Commit<sup>ivk</sup> are not distinguished from each other in [section 4.2.3](https://zips.z.cash/protocol/protocol.pdf#orchardkeycomponents).)

It is ok to panic on cases that have insignificant probability even under adversarial input, which is the case here (see [Theorem 5.4.4](https://zips.z.cash/protocol/protocol.pdf#thmsinsemillaex) for the $\bot$ case; the 0 case is similar because the adversary would have to find $r$ such that SinsemillaCommit<sub>$r$</sub> returns $\mathcal{O}_{\mathbb{P}}$ which gives a nontrivial discrete log relation).

zebra-chain/src/transaction/txid.rs
```diff
@@ -36,9 +36,7 @@ impl<'a> TxIdBuilder<'a> {
     /// In these cases it's simply the hash of the serialized transaction.
     fn txid_v1_to_v4(self) -> Result<Hash, io::Error> {
         let mut hash_writer = sha256d::Writer::default();
-        self.trans
-            .zcash_serialize(&mut hash_writer)
-            .expect("Transactions must serialize into the hash.");
+        self.trans.zcash_serialize(&mut hash_writer)?;
```
daira: This is an example of a change that I would suggest should probably not be made. The code before this PR assumed that failure of `zcash_serialize` in this case would indicate a programming error, which is not the same as an I/O error. The handling of the latter would likely obscure the programming error (if the assumption was in fact correct).

teor2345:
@oxarbitrage this PR is named as a feature, but it's actually a developer lint.
Do you mind if I rename it?

oxarbitrage:
> Do you mind if I rename it?

please go ahead.

conradoplg:
I agree it seems risky to change the behavior of functions to make panics into errors.

We could turn this PR into something that just enables the lint and disables it in every usage, but I'm not sure if it's worth doing it. The upside would be to force us to consider the usage of the pattern since it would warn of us if we write new code with that pattern. But that sounds pretty common in our codebase. (I would not be against it though since most of the work was already done in the PR)

I feel like it's more important to correctly document all our panics (i.e. turn all `unwrap()` into `expect()` with a good explanation), not sure what is the lint for that.

conradoplg:
That would be https://rust-lang.github.io/rust-clippy/master/#unwrap_used

oxarbitrage:
Yea, that is the third one in the ticket https://github.com/ZcashFoundation/zebra/issues/2297, should be done in a different PR.
The good thing about that one is that it haves a flag to disable it in test code but we will still have to add exceptions. For example `CtOption` types do not have `.except()` and they are used in our codebase. 

In regards to this one, the suggestion of revert the `?` changes and go back to `except` and `unwrap` sound as something we could do without too much effort. Don't use the lint at all and close the PR is the other option.

zebra-chain/src/transaction/txid.rs
oxarbitrage: Thank you @daira , this changes were reverted in https://github.com/ZcashFoundation/zebra/pull/4667/commits/984ee1ce82b1144210c697444797555f2fa62a64 , so we keep the panic and added an exception here and in the rest of them.

zebra-chain/src/orchard/commitment.rs
oxarbitrage: reverted in https://github.com/ZcashFoundation/zebra/pull/4667/commits/984ee1ce82b1144210c697444797555f2fa62a64

zebra-chain/src/transaction.rs
```diff
@@ -15,6 +15,7 @@ mod txid;
 mod unmined;
 
 #[cfg(any(test, feature = "proptest-impl"))]
+#[allow(clippy::unwrap_in_result)]
```
teor2345, line 18: Can you put this in the module file instead?

zebrad/tests/acceptance.rs
```diff
@@ -119,6 +119,7 @@ use zebra_state::constants::LOCK_FILE_ERROR;
 
 use zebra_test::{args, command::ContextFrom, net::random_known_port, prelude::*};
 
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this inside the module please?

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -148,7 +149,7 @@ impl VerifiedSet {
                 .map(|tx| tx.clone().eviction_weight())
                 .collect();
 
-            let dist = WeightedIndex::new(weights).unwrap();
+            let dist = WeightedIndex::new(weights).expect("iterator is not empty");
```
teor2345: It's a bit more complicated than that:
https://docs.rs/rand/latest/rand/distributions/struct.WeightedIndex.html#method.new

```suggestion
            let dist = WeightedIndex::new(weights).expect("there is at least one weight and all weights are valid");
```

zebra-test/src/lib.rs
```diff
@@ -13,6 +13,7 @@ use tracing_subscriber::{fmt, prelude::*, EnvFilter};
 use std::sync::Once;
 
 #[allow(missing_docs)]
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this inside the module?

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -189,6 +195,7 @@ impl
     }
 }
 
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on the specific method which needs the lint exception?

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -166,6 +171,7 @@ impl
         ),
         _position: RevertPosition,
     ) {
+        // TODO: change this unwraps to expects or explain
         self.balance =
             (self.balance + spent_output.utxo.output.value().constrain().unwrap()).unwrap();
```
teor2345: Here are some explanations, but you might also need `rustfmt`

```rust
        self.balance =
            (self.balance + spent_output.utxo.output.value().constrain().expect("NonNegative values are always valid NegativeAllowed values")).expect("reversing previous balance changes is always valid");
```

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -139,6 +143,7 @@ impl
         ),
     ) -> Result<(), ValidateContextError> {
         // Spending a UTXO subtracts value from the balance
+        // TODO: change this unwraps to expects or explain
         self.balance =
             (self.balance - spent_output.utxo.output.value().constrain().unwrap()).unwrap();
```
teor2345: Here are some explanations, but you might also need `rustfmt`

```rust
        self.balance =
            (self.balance - spent_output.utxo.output.value().constrain().expect("NonNegative values are always valid NegativeAllowed values")).expect("total UTXO value has already been checked");
```

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -57,6 +57,7 @@ pub struct TransparentTransfers {
 // A created UTXO
 //
 // TODO: replace arguments with a struct
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on methods instead?

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -118,6 +121,7 @@ impl
 // A transparent input
 //
 // TODO: replace arguments with a struct
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on methods instead?

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -94,6 +96,7 @@ impl
         &(outpoint, created_utxo): &(&transparent::OutPoint, &transparent::OrderedUtxo),
         _position: RevertPosition,
     ) {
+        // TODO: change this unwraps to expects or explain
         self.balance =
             (self.balance - created_utxo.utxo.output.value().constrain().unwrap()).unwrap();
```
teor2345: Here are some explanations, but you might also need `rustfmt`

```rust
        self.balance =
            (self.balance - created_utxo.utxo.output.value().constrain().expect("NonNegative values are always valid NegativeAllowed values")).expect("reversing previous balance changes is always valid");
```

zebra-state/src/service/non_finalized_state/chain/index.rs
```diff
@@ -70,6 +71,7 @@ impl
         &mut self,
         &(outpoint, created_utxo): &(&transparent::OutPoint, &transparent::OrderedUtxo),
     ) -> Result<(), ValidateContextError> {
+        // TODO: change this unwraps to expects or explain
         self.balance =
             (self.balance + created_utxo.utxo.output.value().constrain().unwrap()).unwrap();
```
teor2345: Here are some explanations, but you might also need `rustfmt`

```rust
        self.balance =
            (self.balance + created_utxo.utxo.output.value().constrain().expect("NonNegative values are always valid NegativeAllowed values")).expect("total UTXO value has already been checked");
```

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -353,6 +354,7 @@ impl ZebraDb {
     }
 }
 
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on the individual methods instead?

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -38,6 +38,7 @@ use crate::{
     BoxError, FinalizedBlock,
 };
 
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on the individual methods instead?

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -45,6 +45,7 @@ use crate::{
 #[cfg(test)]
 mod tests;
 
+#[allow(clippy::unwrap_in_result)]
```
teor2345: Can you put this on the methods instead?

zebra-script/src/lib.rs
```diff
@@ -162,7 +163,6 @@ impl CachedFfiTransaction {
         let script_pub_key: &[u8] = lock_script.as_raw_bytes();
 
         // This conversion is useful on some platforms, but not others.
-        #[allow(clippy::useless_conversion)]
```
teor2345: Please don't remove this lint allow, it is required for Windows compatibility

zebra-script/src/lib.rs
```diff
@@ -176,6 +176,7 @@ impl CachedFfiTransaction {
             | zcash_script::zcash_script_SCRIPT_FLAGS_VERIFY_CHECKLOCKTIMEVERIFY;
         // This conversion is useful on some platforms, but not others.
         #[allow(clippy::useless_conversion)]
+        #[allow(clippy::unwrap_in_result)]
```
teor2345: This seems redundant, the lint is already allowed on the method

```suggestion
```

zebra-chain/src/sapling/address.rs
```diff
@@ -59,17 +60,21 @@ impl fmt::Display for Address {
...
 impl std::str::FromStr for Address {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let mut decoded_bytes = io::Cursor::new(Vec::<u8>::from_base32(&bytes).unwrap());
+                let mut decoded_bytes = io::Cursor::new(
+                    Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"),
```
teor2345: This reason needs to explain why the code will never panic

```suggestion
                    Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32"),
```

zebra-chain/src/sapling/address.rs
```diff
@@ -59,17 +60,21 @@ impl fmt::Display for Address {
...
     }
 }
 
 impl std::str::FromStr for Address {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
```
teor2345, line 73: The lint has discovered a coding error here.
This code handles untrusted data, so it must not panic.

But if the `TransmissionKey` is invalid, it panics.
It should return an error instead.

zebra-chain/src/sapling/keys.rs
```diff
@@ -254,23 +254,26 @@ impl From<[u8; 32]> for SpendingKey {
...
 
 impl FromStr for SpendingKey {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let decoded = Vec::<u8>::from_base32(&bytes).unwrap();
+                let decoded = Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32");
```
teor2345: This code parses untrusted data, so the reason needs to explain why the code will never panic

```suggestion
                let decoded = Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32");
```

zebra-chain/src/sapling/keys.rs
```diff
@@ -690,10 +695,11 @@ impl From<(AuthorizingKey, NullifierDerivingKey)> for IncomingViewingKey {
 impl FromStr for IncomingViewingKey {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let decoded = Vec::<u8>::from_base32(&bytes).unwrap();
+                let decoded = Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32");
```
teor2345: This code parses untrusted data, so the reason needs to explain why the code will never panic

```suggestion
                let decoded = Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32");
```

zebra-chain/src/sapling/keys.rs
```diff
@@ -970,17 +977,21 @@ impl fmt::Display for FullViewingKey {
...
 impl FromStr for FullViewingKey {
     type Err = SerializationError;
 
+    #[allow(clippy::unwrap_in_result)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match bech32::decode(s) {
             Ok((hrp, bytes, Variant::Bech32)) => {
-                let mut decoded_bytes = io::Cursor::new(Vec::<u8>::from_base32(&bytes).unwrap());
+                let mut decoded_bytes = io::Cursor::new(
+                    Vec::<u8>::from_base32(&bytes).expect("bytes are valid base32"),
```
teor2345: This code parses untrusted data, so the reason needs to explain why the code will never panic

```suggestion
                    Vec::<u8>::from_base32(&bytes).expect("bech32::decode guarantees valid base32"),
```

teor2345 (commented):
Thanks, this looks pretty good now. I appreciate all the work you put in here.

I have checked each change, and made a bunch of suggestions.

Overall, the lint:
- discovered one incorrect panic - it should be an error
- helped discover bugs #4691 and #4692 (they are full wallet only bugs)
- helped document why many `expect`s can never panic
- added a large number of lint `allow`s

So I think it is valuable enough to keep.

I think we should keep doing what you're doing with test modules - the lint is only really useful for production code.

zebrad/src/application.rs
```diff
@@ -393,14 +394,15 @@ impl Application for ZebradApp {
...
+    #[allow(clippy::unwrap_in_result)]
     fn init(&mut self, command: &Self::Cmd) -> Result<(), FrameworkError> {
         // Create and register components with the application.
         // We do this first to calculate a proper dependency ordering before
         // application configuration is processed
         self.register_components(command)?;
 
         // Fire callback to signal state in the application lifecycle
-        let config = self.config.take().unwrap();
+        let config = self.config.take().expect("we should have a config object");
```
teor2345: We need to explain why the panic never happens.

In this case, we are using `abscissa` in an unusual way, so we should explain exactly what we're doing.

```suggestion
        let config = self.config.take().expect("register_components always populates the config");
```

teor2345 (dismissed):
Thanks for this, it looks like a lot of work!

Let's get it merged before there are any conflicts.

teor2345 (approved):

PR #4670, opened 2022-06-22
fix(sync): change default sync config to improve reliability

teor2345:
## Motivation

Zebra is a bit slow to verify orchard blocks with lots of transactions, which can cause a failure loop.

This PR should reduce the concurrency a bit, allowing it to sync more reliably.

## Solution

Concurrency:
- Decrease the default lookahead limit from 2000 to 400
- Increase the block verification timeout from 6 to 10 minutes
- Decrease the default concurrent downloads config from 50 to 25

Priority:
- Try to run the spawned download task before queueing the next download

Cancellation:
- Allow verification to be cancelled if the verifier is busy 

## Review

This is pretty urgent, it seems to be blocking a lot of other fixes and tests.

### Reviewer Checklist

  - [ ] Full sync works more reliably

### DIFF
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -57,7 +57,7 @@ const BLOCK_DOWNLOAD_RETRY_LIMIT: usize = 3;
 
 /// A lower bound on the user-specified lookahead limit.
 ///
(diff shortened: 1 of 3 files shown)


conradoplg (approved):
Looks good

PR #4676, opened 2022-06-22
tests(config): Add tests for old configs

### LINKED ISSUE #4625: Test that old zebrad.toml files still work
## Motivation

We want to make sure that we don't accidentally break the config format.

### Designs

- [ ] Commit a copy of the `zebrad generate` output to a test folder in the Zebra git repository
- [ ] Add an acceptance test that makes sure the config is up to date in the Zebra git repository
- [ ] Add an acceptance test that launches `zebrad start` using the config from the Zebra git repository

We already have a test that launches using a freshly generated config.

oxarbitrage:
## Motivation

We want to add tests for breaking changes in our config. Closes https://github.com/ZcashFoundation/zebra/issues/4625

## Solution

I had to change the type of the peer set fields in the config to `IndexSet` , so the peers are always displayed in the same order. I eventually compare two configs to be equal. This work is done in  https://github.com/ZcashFoundation/zebra/commit/7adada806b8b0ebeccccde91d239d090ded0ce10

Tests are static config file intro is done in https://github.com/ZcashFoundation/zebra/commit/f7053b0fc5f2bfc6c00171dcf81daef42b7bac3b

## Review

I think anyone can review.

### Reviewer Checklist

  - [ ] Added tests make sense and pass


zebrad/tests/common/config.rs
```diff
@@ -80,3 +84,15 @@ pub fn testdir() -> Result<TempDir> {
...
+/// Get stored config path
+pub fn stored_config_path() -> PathBuf {
+    Path::new(env!("CARGO_BIN_EXE_zebrad"))
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join("zebrad/tests/common/config.toml")
```
oxarbitrage: This can be fragile, not to mention ugly. Open for a better way to do it.

zebrad/tests/acceptance.rs
```diff
@@ -1706,6 +1706,57 @@ async fn delete_old_databases() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn stored_config_works() -> Result<()> {
```
teor2345: The ports and paths in this test will conflict with valid_generated_config.

To avoid the conflict, call this function in valid_generated_config_test after valid_generated_config finishes.

zebrad/tests/acceptance.rs
```diff
@@ -1706,6 +1706,57 @@ async fn delete_old_databases() -> Result<()> {
...
+    let child =
+        run_dir.spawn_child(args!["generate", "-o": generated_config_path.to_str().unwrap()])?;
+
+    let output = child.wait_with_output()?;
+    let _output = output.assert_success()?;
+
+    let contents_generated = std::fs::read_to_string(generated_config_path).unwrap();
+    let contents_stored = std::fs::read_to_string(stored_config_path).unwrap();
+
+    assert_eq!(contents_generated, contents_stored);
```
teor2345: We might want to tell developers why the test failed, and what command they need to run to update the stored config file.

zebrad/tests/common/config.rs
teor2345: Here's one way to avoid using the file path:
1. Use include_str!() to turn it into a string variable: https://doc.rust-lang.org/std/macro.include_str.html
2. Parse the string as a config, if parsing fails, fail the test
3. Return the config from this function
4. Launch zebrad with the config using zebrad.with_config()

You could skip parsing the config, but then you'd have to set up the temp path and process arguments yourself.

teor2345 (commented):
Looks good so far, I think I found a way to avoid using the git checkout paths.

zebrad/tests/acceptance.rs
```diff
@@ -1706,6 +1706,57 @@ async fn delete_old_databases() -> Result<()> {
...
+    // [Note on port conflict](#Note on port conflict)
+    output
+        .assert_was_killed()
+        .wrap_err("Possible port conflict. Are there other acceptance tests running?")?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn stored_config_is_newest() -> Result<()> {
```
teor2345: I don't think this needs to be an async function.

zebrad/tests/acceptance.rs
teor2345: It will be easier to call this function if it's not async.
(I don't think it needs to be async.)

zebrad/tests/acceptance.rs
oxarbitrage: Thanks, they don't need to be async at all. https://github.com/ZcashFoundation/zebra/pull/4676/commits/2483cea8a02eacebcba3703f0bb2b1573f92effc

zebrad/tests/common/config.rs
teor2345: (This is pretty much what `cargo insta` does with its snapshots, or it handles the different paths internally.)

zebrad/tests/acceptance.rs
```diff
@@ -1706,6 +1706,76 @@ async fn delete_old_databases() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn stored_config_works() -> Result<()> {
+    zebra_test::init();
+
+    // only run the test if we found a stored config
+    if let Some(stored_config_path) = stored_config_path() {
```
teor2345: This check can always skip the test. If that happens, we could break the config, and CI would still pass.

Can we require that the config is always available?
Did you want some help with the paths?

oxarbitrage: It will skip it only for some CI jobs (only one actually). We just need one test that finds the config to trigger the error and alert of the problem.

This works ok as it is but feel free to add whatever you want.

zebrad/tests/acceptance.rs
oxarbitrage: Not needed if our tests are not async ?

zebrad/tests/acceptance.rs
oxarbitrage: This is a good one, thanks. Added in https://github.com/ZcashFoundation/zebra/pull/4676/commits/f423afbcf3be61989e29bcaccc833042c1ba967a

zebrad/tests/common/config.rs
oxarbitrage: Hard to follow what to do and to me, unnecessary. I did some changes to this function but feel free to  change it to what you want.

teor2345:
There might be some ways to break a config that aren't covered by the design in the ticket.
(This isn't anyone's fault, sometimes we only find out what works when we start writing code.)

At the moment, if we change a field format:
- `stored_config_is_newest` makes us update the stored config in the PR
- the updated stored config in the PR will pass `stored_config_works` (because it's testing the same config as `valid_generated_config`)

But what we want is:
- `stored_config_works` checks that older versions of the config work with the current `zebrad`
- `stored_config_is_newest` makes sure new config fields get checked with future versions of `zebrad`

We might need to keep a copy of every generated config change, and test them all.
(We can delete old configs when we do a breaking release, like at a network upgrade.)

I'm sorry this is more complicated than the ticket.
Let me know what you think?

zebrad/tests/acceptance.rs
teor2345: `cargo test` runs tests in parallel in different threads:
https://doc.rust-lang.org/cargo/commands/cargo-test.html#description

If the tests are async, they get launched on the async executor, but either way they run concurrently.

zebrad/tests/acceptance.rs
teor2345: We've had CI bugs where we always skip a test, and we never notice.
How can we make sure at least one test finds the config?

I think we can use `CARGO_MANIFEST_DIR` to find the path, it is `zebrad/Cargo.toml`:
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

teor2345: I just remembered something:
- `env()` (the function) gets the environment when the tests run
- `env!()` (the macro) gets the environment when the tests are compiled

You might need `env!()` (the macro) to get `CARGO_MANIFEST_DIR`, because it might only be set when the tests are built.

oxarbitrage: `env!()` is what we are using and it seems to be working ok locally and in CI.

teor2345 (approved):
Thanks, looks good!

PR #4679, opened 2022-06-23
fix(sync): Slightly increase some syncer defaults

teor2345:
## Motivation

We changed the syncer defaults to make full verification more reliable, but that slowed down syncing by about 15%.

This is a partial revert of PR #4670.

## Solution

- Increase the syncer defaults to about halfway between the previous and current defaults

This is a temporary fix until we get orchard batch verification.

## Review

This PR needs a full sync test before it gets merged.

### Reviewer Checklist

  - [ ] Full sync is greater than 85% at 6 hours


teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2546930195

We want it to get higher than 85% at 6 hours.

conradoplg:
It timed out at 96.781%, so it seems good?

teor2345:
> It timed out at 96.781%, so it seems good?

Yep that's much better, we'll fix the timeout with #4661

gustavovalverde (approved):
This is working as expected (sync is faster in the tested workflow). As it's a very straightforward change I'll approve it so we can test PRs being dependent on this one.

PR #4687, opened 2022-06-24
fix(network messages): add limits to rejection message and reason

### LINKED ISSUE #4632: Limit the size of reject network message strings
## Motivation

We want to avoid sending large error messages on the network.

`zcashd` does this with a limit of `111`, we might want ours to be slightly smaller:
https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/main.cpp#L7334-L7335

### Specifications

Bitcoin doesn't say much about limits:
https://en.bitcoin.it/wiki/Protocol_documentation#reject

oxarbitrage:
## Motivation

Zcash limits the rejection  message and the reason to specific constants. We want to do similar for Zebra.

Close https://github.com/ZcashFoundation/zebra/issues/4632

Depends-On: #4714

## Solution

Add constants and truncate `Mesage::Reject::message` to `MAX_REJECT_MESSAGE_LENGTH` and `Mesage::Reject::reason` to `MAX_REJECT_REASON_LENGTH` 

## Review

I think anyone can review, i left some comments with references to zcashd.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebra-network/src/protocol/external/message.rs
+++ b/zebra-network/src/protocol/external/message.rs
@@ -6,7 +6,6 @@ use chrono::{DateTime, Utc};
 
 use zebra_chain::{
     block::{self, Block},
-    serialization::ZcashSerialize,
     transaction::UnminedTx,
 };
 
@@ -351,40 +350,37 @@ const MAX_REJECT_MESSAGE_LENGTH: usize = 12;
 /// This is equivalent to `MAX_REJECT_MESSAGE_LENGTH` in zcashd.
 const MAX_REJECT_REASON_LENGTH: usize = 111;
 
+// TODO: add tests for Error conversion and Reject message serialization (#4633)
+// (Zebra does not currently send reject messages, and it ignores received reject messages.)
 impl<E> From<E> for Message
 where
     E: Error,
 {
     fn from(e: E) -> Self {
-        let message_bytes = e.to_string().zcash_serialize_to_vec().unwrap_or_default();
+        let message = e
+            .to_string()
+            .escape_default()
+            .take(MAX_REJECT_MESSAGE_LENGTH)
+            .collect();
+        let reason = e
+            .source()
+            .map(ToString::to_string)
+            .unwrap_or_default()
+            .escape_default()
+            .take(MAX_REJECT_REASON_LENGTH)
+            .collect();
 
         Message::Reject {
-            message: String::from_utf8(
-                message_bytes[0..std::cmp::min(MAX_REJECT_MESSAGE_LENGTH, message_bytes.len())]
-                    .to_vec(),
-            )
-            .unwrap_or_else(|_| String::from("")),
+            message,
 
             // The generic case, impls for specific error types should
             // use specific varieties of `RejectReason`.
             ccode: RejectReason::Other,
 
-            reason: if let Some(reason) = e.source() {
-                let reason_bytes = reason
-                    .to_string()
-                    .zcash_serialize_to_vec()
-                    .unwrap_or_default();
+            reason,
 
-                String::from_utf8(
-                    reason_bytes[0..std::cmp::min(MAX_REJECT_REASON_LENGTH, reason_bytes.len())]
-                        .to_vec(),
-                )
-                .unwrap_or_else(|_| String::from(""))
-            } else {
-                String::from("")
-            },
-
-            // Allow this to be overridden but not populated by default, methinks.
+            // The hash of the rejected block or transaction.
+            // We don't have that data here, so the caller needs to fill it in later.
             data: None,
         }
     }
--- a/zebra-network/src/protocol/external/message.rs
+++ b/zebra-network/src/protocol/external/message.rs
@@ -356,10 +356,11 @@ where
     E: Error,
 {
     fn from(e: E) -> Self {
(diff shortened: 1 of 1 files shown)


zebra-network/src/protocol/external/message.rs
```diff
@@ -340,20 +340,29 @@ pub enum Message {
     FilterClear,
 }
 
+/// The maximum size of the rejection message
+const MAX_REJECT_MESSAGE_LENGTH: usize = 12;
```
oxarbitrage, line 346: From https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/main.cpp#L7334-L7335 we know zcashd uses `COMMAND_SIZE` for the message field. This is equal to 12:  https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/protocol.h#L33

zebra-network/src/protocol/external/message.rs
```diff
@@ -340,20 +340,29 @@ pub enum Message {
...
+/// The maximum size of the rejection reason
+const MAX_REJECT_REASON_LENGTH: usize = 100;
```
oxarbitrage: From https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/main.cpp#L7334-L7335 we know zcashd uses `MAX_REJECT_MESSAGE_LENGTH` for the message field. This is equal to 111: https://github.com/zcash/zcash/blob/1b5ab4a06e52c23aca48afb1ebb72a25ef5dd3ca/src/main.h#L117

I added `100` here instead of `111` as we want a  slightly smaller value: https://github.com/ZcashFoundation/zebra/issues/4632#issue-1274130910

zebra-network/src/protocol/external/message.rs
teor2345: Can you put this explanation in a doc comment?

It will help us understand why we chose `12` when we look at this code later.

zebra-network/src/protocol/external/message.rs
teor2345: Can you put this explanation in a doc comment?

It will help us understand why we chose `100` when we look at this code later.

zebra-network/src/protocol/external/message.rs
```diff
@@ -340,20 +340,29 @@ pub enum Message {
...
+
 impl<E> From<E> for Message
 where
     E: Error,
 {
     fn from(e: E) -> Self {
+        let mut message = e.to_string();
+        message.truncate(MAX_REJECT_MESSAGE_LENGTH);
```
teor2345: Can we call `shrink_to_fit` on both strings after we truncate them?
https://doc.rust-lang.org/std/string/struct.String.html#method.shrink_to_fit

If we have a lot of very large error strings, they will take up a lot of memory. That memory isn't freed by `truncate`, we need to call `shrink_to_fit` as well.

teor2345 (commented):
Looks great!

We could merge this like it is, but it increases the risk of memory denial of service, and there is an easy fix for that.

zebra-network/src/protocol/external/message.rs
```diff
@@ -340,20 +340,29 @@ pub enum Message {
...
+            message,
 
             // The generic case, impls for specific error types should
             // use specific varieties of `RejectReason`.
             ccode: RejectReason::Other,
 
             reason: if let Some(reason) = e.source() {
-                reason.to_string()
+                let mut reason = reason.to_string();
+                reason.truncate(MAX_REJECT_REASON_LENGTH);
```
teor2345: Can we escape any weird characters before truncating the string?
https://doc.rust-lang.org/std/primitive.str.html#method.escape_default

Error message strings might contain weird characters, because they are generated from untrusted peer data. So we should do something to sanitize them.

If we don't escape weird characters, our limits might not work like we expect. Because `truncate()` limits the length in UTF-8 characters, but `zcashd` limits the length in bytes. (Some UTF-8 characters can take up multiple bytes.)

Here is how we convert the string to bytes when we serialize the message:
https://github.com/ZcashFoundation/zebra/blob/e7094201604c2f473145169a10c23a703a57c96d/zebra-chain/src/serialization/zcash_serialize.rs#L163-L169

Here are some examples of multi-byte UTF-8 characters:
https://doc.rust-lang.org/std/str/fn.from_utf8.html#examples

zebra-network/src/protocol/external/message.rs
```diff
@@ -340,25 +340,47 @@ pub enum Message {
...
 impl<E> From<E> for Message
 where
     E: Error,
 {
     fn from(e: E) -> Self {
+        let message = e
+            .to_string()
+            .escape_default()
+            .take(MAX_REJECT_MESSAGE_LENGTH)
+            .collect();
```
teor2345, line 364: I didn't realise that `escape_*` returned an iterator.

This allocates the same amount of memory for the temporary string, but it's dropped right after creating the smaller truncated string.

oxarbitrage: lookg good. thanks.

teor2345 (approved):
Let's merge after the release

PR #4689, opened 2022-06-26
Warn the user when Zebra doesn't recognize the format of `zebrad.toml`

### LINKED ISSUE #4665: Warn the user when Zebra doesn't recognize the format of `zebrad.toml`
## Motivation

When the format of `zebrad.toml` changes, Zebra doesn't recognize it appropriately and returns an unclear error saying that it couldn't read some data. This becomes an issue when users install a new version of Zebra that uses a new format of `zebrad.toml`, but keep their old `zebrad.toml` file.

### Suggested solution

Zebra should provide a message saying that it didn't recognize some items in the `zebrad.toml` file, and suggest that the file might need to be updated. 

## Related Work

This issue depends on issue #4625.

upbqdn:
## Motivation

When the format of zebrad.toml changes, Zebra doesn't recognize it appropriately and returns an unclear error saying that it couldn't read some data. This becomes an issue when users install a new version of Zebra that uses a new format of zebrad.toml, but keep their old `zebrad.toml` file.

## Solution

Zebra now provides a message saying that it could not parse the provided config file, and suggests that the user might be using a deprecated format of the file.

Close #4665

### Reviewer Checklist

### DIFF
--- a/zebrad/src/application.rs
+++ b/zebrad/src/application.rs
@@ -207,7 +207,7 @@ impl Application for ZebradApp {
             Some(path) => match self.load_config(&path) {
                 Ok(config) => config,
                 Err(e) => {
-                    status_err!("Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file.");
+                    status_err!("Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file. You can generate a valid config by running \"zebrad generate\", and diff it against yours to examine any format inconsistencies.");
                     return Err(e);
                 }
             },
--- a/zebrad/src/application.rs
+++ b/zebrad/src/application.rs
@@ -5,6 +5,7 @@ use std::{fmt::Write as _, io::Write as _, process};
 use abscissa_core::{
     application::{self, fatal_error, AppCell},
     config::{self, Configurable},
+    status_err,
     terminal::{component::Terminal, stderr, stdout, ColorChoice},
     Application, Component, EntryPoint, FrameworkError, Shutdown, StandardPaths, Version,
 };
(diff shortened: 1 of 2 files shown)


zebrad/tests/acceptance.rs
```diff
@@ -561,6 +562,72 @@ fn valid_generated_config(command: &str, expect_stdout_line_contains: &str) -> R
...
+    // Run Zebra in a temp dir so that it loads the config, and kill it after
+    // two seconds.
+    //
+    // Note: We should actually not need to kill Zebra because it should
+    // instantly terminate itself after the start since the config file is
+    // invalid. Two seconds should be sufficient for Zebra to read the config,
+    // and terminate with an error message.
+    let mut child = testdir.spawn_child(args!["start"])?;
+    std::thread::sleep(Duration::from_secs(2));
+    child.kill()?;
```
upbqdn, line 622: I tried running the test without `child.kill()?`, and it worked well. I was trying to figure out a way to return an error when Zebra runs for more than two seconds, but without success.

zebrad/src/application.rs
```diff
@@ -202,11 +203,16 @@ impl Application for ZebradApp {
...
-            .unwrap_or_default();
+        let config = match command.config_path() {
+            Some(path) => match self.load_config(&path) {
+                Ok(config) => config,
+                Err(e) => {
+                    status_err!("Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file.");
+                    return Err(e);
+                }
+            },
+            None => ZebradConfig::default(),
```
upbqdn, line 214: Tracking is not active here yet, so I couldn't use `info!`. I wanted to say that `No config file provided. Using default options.`, but left it out because I only found the `status_ok!` macro provided by Abscissa that uses `stdout`, so it would start sticking out once we redirect the output to a regular log file.

teor2345: Can we suggest how to fix the config in the `status_err!()` message?

teor2345 (commented):
Looks good!

I did a merge with PR #4676 and added some comments.

zebrad/src/application.rs
upbqdn: Done in [b30b2ad](https://github.com/ZcashFoundation/zebra/pull/4689/commits/b30b2addd02b4b78dad35e931dc35da76ac5af8c).

upbqdn: Now I'm not sure if I should add any newline characters to the message since it's long.

teor2345: I think either way is fine.

teor2345 (approved):
Thanks for the quick fixes!

Issue #4672, opened 2022-06-22
Add support for Orchard proof batch verification


conradoplg:
## Motivation

The `halo2_proofs` crate ~~does not yet support batch verification~~ (I think it does but we don't use it yet, but in any case its performance was improved recently). When it does, we should use it, since there are some blocks with many Orchard actions that are taking seconds to validate.

### Designs

- [ ] Use batch verification in Zebra (see how we handle Sapling batch verification)
- [ ] Revert the relevant changes added in https://github.com/ZcashFoundation/zebra/pull/4670/ if applicable

dconnolly:
Looking at the changes in https://github.com/zcash/orchard/pull/338/files, we will probably pull in `halo2_proofs::plonk::BatchVerifier` and slot that into our `tower::Service`

teor2345:
Note commitment tree updates are slower than proofs, and single-threaded, so they are a higher priority than orchard batching.

teor2345:
Performance improvements are a low priority right now.

PR #4699, opened 2022-06-27
deps: upgrade ECC dependencies

conradoplg:
## Motivation

ECC upgraded some of dependencies, and that includes improved batched halo2 verification which we want.

## Solution

- Upgrade deps
- Sync our zcash_proofs fork with upstream and point to that
- Had to add some entries to `deny.toml` because those crates haven't been bumped in `zcash` itself, and that's what we need to use for `zcash_script`.

## Review

With this we should be able to improve sync performance, so it's important.

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

- Actually use the batch API: https://github.com/ZcashFoundation/zebra/issues/4672

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1487,8 +1487,18 @@ dependencies = [
 
 [[package]]
 name = "equihash"
-version = "0.1.0"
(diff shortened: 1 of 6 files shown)


teor2345 (approved):
Looks great, thanks for this!

PR #4694, opened 2022-06-27
feat(log): Show the current network upgrade in progress logs

teor2345:
## Motivation

It takes almost 6 hours for Zebra to sync all its checkpoints, so we need to be able to split CI jobs by network upgrade as well as checkpoint/full validation.

This is part of ticket #4661, it also includes a refactor I did as part of loading blocks from a file.

## Solution

Features:
- Show the network upgrade in progress logs
- Log when Zebra verifies the final checkpoint

Fixes:
- Log elapsed time using `humantime` truncated to whole seconds, rather than raw seconds and microseconds
- Fix incorrect negative sign on elapsed time
- Remove the space before the percentage sign when showing percent progress

Refactors:
- Move the progress task into its own module

## Review

This is a high priority because I need it for ticket #4661.

### Reviewer Checklist

  - [ ] Progress logs contain network upgrade
  - [ ] Time is formatted nicely
  - [ ] Existing tests pass

PR #4690, opened 2022-06-27
refactor(ci): Split `docker run` into launch, `logs`, and `wait`

teor2345:
## Motivation

This PR further splits up the `docker run` job.

It is part of ticket #4661.

## Solution

This PR splits the `docker run` jobs into 3 different jobs:
- launch: `docker run --detach`
- logs: `docker logs --tail all --follow`
- result: `docker wait`

It also fixes up some comments and formatting.

## Review

This might be one for Gustavo, but anyone who understands GitHub workflows and docker should be able to review it.

### Reviewer Checklist

  - [ ] All GCP workflow jobs have launch, logs, and result steps
  - [ ] Failing jobs still fail

## Follow Up Work

- split the logs into checkpoint (sprout, sapling, orchard?) and full validation (sapling?, orchard)
- require that full sync goes to 100%

PR #4704, opened 2022-06-28
fix(ci): Split Docker logs into sprout, other checkpoints, and full validation

### LINKED ISSUE #4661: Split full sync test into checkpoint and full validation jobs

teor2345:
## Motivation

We want to avoid the 6 hour job limit for GitHub actions by splitting CI jobs.

Depends-On: #4690, #4694

Close #4661

## Solution

Split the log following CI job into:
- sprout checkpoints (ends on sapling activation)
- sapling/orchard checkpoints (ends at the final checkpoint)
- full validation (ends at the end of the test)

Some tests don't do a full sync, so all of the log jobs also end if the test has ended.

## Review

This PR blocks a bunch of other PRs, so I've marked it as a critical hotfix.

The diff on this PR will show extra code until PRs #4690 and #4694 merge.

### Reviewer Checklist

  - [ ] Test job structure is correct
  - [ ] Full sync consistently passes without timing out
  - [ ] Failing jobs still fail

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -672,9 +672,10 @@ jobs:
           --zone ${{ env.ZONE }} \
           --quiet \
           --ssh-flag="-o ServerAliveInterval=5" \
-          --command="docker wait ${{ inputs.test_id }}")
-
-          exit ${EXIT_CODE}
+          --command="docker wait ${{ inputs.test_id }}" \
+          )
+          echo "$EXIT_CODE"
+          exit "$EXIT_CODE"
 
 
   # create a state image from the instance's state disk, if requested by the caller
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -662,20 +662,28 @@ jobs:
 
       # Wait for the container to finish, then exit with the test's exit status.
       #
-      # `docker wait` prints the container exit status as a string, but we need to exit `ssh` with that status.
-      # `docker wait` can also wait for multiple containers, but we only ever wait for a single container.
+      # If the container has already finished, `docker wait` should return its status.
+      # But sometimes this doesn't work, so we use `docker inspect` as a fallback.
+      #
+      # `docker wait` prints the container exit status as a string, but we need to exit the `ssh` command
+      # with that status.
+      # (`docker wait` can also wait for multiple containers, but we only ever wait for a single container.)
       - name: Result of ${{ inputs.test_id }} test
         run: |
-          EXIT_CODE=$(\
           gcloud compute ssh \
           ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
           --zone ${{ env.ZONE }} \
           --quiet \
           --ssh-flag="-o ServerAliveInterval=5" \
-          --command="docker wait ${{ inputs.test_id }}" \
-          )
-          echo "$EXIT_CODE"
-          exit "$EXIT_CODE"
+          --command=' \
+          EXIT_STATUS=$( \
+          docker wait ${{ inputs.test_id }} || \
+          docker inspect --format "{{.State.ExitCode}}" ${{ inputs.test_id }} || \
+          echo "missing container, or missing exit status for container" \
+          ); \
+          echo "docker exit status: $EXIT_STATUS"; \
+          exit "$EXIT_STATUS" \
+          '
 
 
   # create a state image from the instance's state disk, if requested by the caller
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -100,8 +100,6 @@ jobs:
       contents: 'read'
       id-token: 'write'
     steps:
-      # TODO: can we delete this step and set create_credentials_file to false in Google Cloud?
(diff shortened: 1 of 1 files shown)


teor2345:
I'm running a full sync and checkpoint sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2572944770

The job structure is correct, but we need to check that all the log jobs run, finish, and succeed.

teor2345:
The command to show the output and `grep` it at the same time was wrong.

Here's a new full sync and checkpoint job:
https://github.com/ZcashFoundation/zebra/actions/runs/2573686935

We need to check that all the log jobs run, finish, and succeed.

conradoplg:
> The command to show the output and `grep` it at the same time was wrong.
> 
> Here's a new full sync and checkpoint job: https://github.com/ZcashFoundation/zebra/actions/runs/2573686935
> 
> We need to check that all the log jobs run, finish, and succeed.

That [failed](https://github.com/ZcashFoundation/zebra/runs/7086099497?check_suite_focus=true) with `Process completed with exit code 141.`. I think that when `grep` quits, `tee` attempts to [write to a closed pipe](https://stackoverflow.com/questions/19120263/why-exit-code-141-with-grep-q) and finishes with an error, making the entire script fail because of the `pipefail`. I remove the `pipefail`s in  a7ee37b and I'm trying a full sync in https://github.com/ZcashFoundation/zebra/actions/runs/2576925253

conradoplg:
I think I launched that incorrectly... trying again in https://github.com/ZcashFoundation/zebra/actions/runs/2577580881

teor2345:
> > The command to show the output and `grep` it at the same time was wrong.
> > Here's a new full sync and checkpoint job: https://github.com/ZcashFoundation/zebra/actions/runs/2573686935
> > We need to check that all the log jobs run, finish, and succeed.
> 
> That [failed](https://github.com/ZcashFoundation/zebra/runs/7086099497?check_suite_focus=true) with `Process completed with exit code 141.`. I think that when `grep` quits, `tee` attempts to [write to a closed pipe](https://stackoverflow.com/questions/19120263/why-exit-code-141-with-grep-q) and finishes with an error, making the entire script fail because of the `pipefail`. I remove the `pipefail`s in [a7ee37b](https://github.com/ZcashFoundation/zebra/commit/a7ee37bebdc107a4215e7dd307b189d925969234) and I'm trying a full sync in https://github.com/ZcashFoundation/zebra/actions/runs/2576925253

Thanks for fixing this up!

I'd like to keep `pipefail`, so that the test fails if the `docker logs` command fails.

`tee --output-error=exit-nopipe` should ignore pipe errors, but fail writing to `/dev/stderr`.
(Just in case we get that path wrong, or writing fails.)

teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2580036755

The last one had the right structure and log job split, so we just need to make sure this one passes.

teor2345:
Turns out that any broken pipe leads to a non-zero exit status, so I had to remove `pipefail`, and use `tee --output-error=exit`.

We'll catch any test failures in the final step that checks the container's exit status.

Here's a full and checkpoint sync run:
https://github.com/ZcashFoundation/zebra/actions/runs/2580386321

gustavovalverde:
I'm reviewing this, I'm not completely sure yet. I'll block it until I'm sure of the process and outcome

teor2345:
We're running:
```
exit "Generating ssh key ..."
```

Which causes an error.

I've just pushed a change to print that output, so I can see which parts of it are the exit code.

teor2345:
I think we can fall back to `docker inspect` if `docker wait` doesn't work.
(`docker wait` has known bugs.)

If that doesn't work, we could also try `docker events`, but it only keeps the last 1000 events.

Currently I'm failing the test if `docker` can't find the container.
But we could also make the test succeed for now, and rely on PR #4710 to check the test logs for failures.

Here's a full tip and checkpoint sync to test it all out:
https://github.com/ZcashFoundation/zebra/actions/runs/2587061833

teor2345:
For some reason, the job status isn't showing on this PR, but you can see it here:
https://github.com/ZcashFoundation/zebra/actions/runs/2587046986

teor2345:
I think we have the behaviour we want here.

Passing tests pass the workflow:
https://github.com/ZcashFoundation/zebra/runs/7123703714

Failing tests fail the workflow, and dependent jobs:
https://github.com/ZcashFoundation/zebra/runs/7124420969
(The sync bug causing this failure will be fixed by ticket #4715.)

Cancellations stop most jobs, but still do cleanup jobs:
https://github.com/ZcashFoundation/zebra/actions/runs/2586943058
(I cancelled the lightwalletd tip job.)

And the tip and checkpoint images get created:
https://github.com/ZcashFoundation/zebra/actions/runs/2587061833

I haven't been able to confirm that `docker wait || docker inspect` works, because I haven't seen `docker wait` fail on the most recent commit. If the fix in this PR doesn't work, PR #4710 checks the test status using the test logs (and we can just ignore the unreliable `docker wait` status).

I'd like to do any other fixes in separate PRs.

teor2345:
@mergifyio update

gustavovalverde (approved):

PR #4708, opened 2022-06-28
change(checkpoints): update mainnet and testnet checkpoints

teor2345:
## Motivation

Updating the checkpoints will speed up Zebra for users and in CI.

## Review

This is a routine change, anyone can review it.

### Reviewer Checklist

  - [ ] Existing tests pass


oxarbitrage (approved):
Always good to do this. should we do this with every release ?

teor2345:
> Always good to do this. should we do this with every release ?

It doesn't really change performance that much, so every few months should be fine.
But if the large Orchard blocks continue, we might want to do it more often.

Issue #2653, opened 2021-08-20
Make sure tests are actually run when some of them are selected


conradoplg:
## Motivation

In some cases we select one or a group of tests to be run. For example, in the .yml files to regenerate the cached state, and in https://github.com/ZcashFoundation/zebra/pull/2633/ to run some tests with fake activation heights.

However, it's possible that renaming these tests can cause the tests to be no longer run, and we could not notice that (since `cargo test` will happily return 0 indicating success even if no tests are run).

Come up with a way to detect if that happens and fail. (For example, we could `grep` the output and check if something has actually run)

### Examples

Here are some example tickets:
- #3673 
- #2402

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/2653?planning-poker) with ZenHub @conradoplg @dconnolly @jvff @oxarbitrage @teor2345 @upbqdn

teor2345:
We don't know how to do this easily, we should fix missing tests when they come up.

We could search for "run 0 tests" in the logs.

teor2345:
PR #4710 fixes part of this ticket, but only for a single workflow

PR #4710, opened 2022-06-28
fix(ci): make sure Rust tests actually run in deploy-gcp-tests.yml

teor2345:
## Motivation

It's easy to accidentally skip a test in CI, and not notice.
We want to check that we're actually running tests, and that they are succeeding.

Part of #2653.

Depends-On: #4704

## Solution

In `deploy-gcp-tests.yml`:
- before checking the exit status, check for "test result: 1 passed" in the logs
- use a regex to allow multiple tests, even though we only currently run one
- fail if the test logs don't show a success, or the test command returns a failing exit status

This PR does not check tests in any other workflows.

## Review

This is a low priority, but it seemed easy to fix while I was in that file.

### Reviewer Checklist

  - [ ] Existing tests pass


teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
@mergifyio update

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -834,7 +834,9 @@ jobs:
           service_account: 'github-service-account@zealous-zebra.iam.gserviceaccount.com'
           token_format: 'access_token'
 
-      # Wait for the container to finish, then exit with the test's exit status.
+      # Check that the container executed at least 1 Rust test harness test, and that all tests passed.
```
dconnolly, line 837: 👍

dconnolly (approved):

teor2345:
@mergifyio refresh

PR #4726, opened 2022-06-30
fix(sync): Temporarily set full verification concurrency to 30 blocks

### LINKED ISSUE #4650: Zebra is slow near the tip
## Description

Recently Zebra became very slow when near the tip (~1000 blocks maybe?).

While in that state it consume half the total CPU and lots of memory.

This showed up while testing https://github.com/ZcashFoundation/zebra/pull/4626. See attempts 2 and 4 in CI (attempt 3 was cancelled) https://github.com/ZcashFoundation/zebra/runs/6937801381?check_suite_focus=true IIRC the code didn't change between those (just the lightwalletd image). Attempt 2 failed due to the lightwalletd stuck bug (which I was investigating) but attempt 4 failed due to Zebra not syncing in time.

@upbqdn also was able to reproduce this, and I also reproduced this in my machine.

### Steps to Reproduce

Try to sync Zebra from a near-tip state.

### Expected Behaviour

It should sync.

### Actual Behaviour

It takes a long time to sync (hours?), or does not sync at all

### Zebra Logs

I can get some logs if needed

### Zebra Version

Both in `main` and I also tested it in `beta.11`, same results.

### Operating System

Ubuntu 20.04 (my machine)

### LINKED ISSUE #4715: Add more backpressure from full validation to syncing
## Motivation

When Zebra's full validation slows down:
- it uses a lot of RAM and CPU verifying blocks
- any extra verifications are slower because RAM or CPU is fully used
- the syncer keeps downloading and queueing large numbers of blocks for verification, taking more RAM
- when blocks timeout, any uncommitted blocks get cancelled, and they have to be downloaded and verified again

We've partially fixed this issue by decreasing the syncer lookahead in PRs #4662, #4670, #4679.
We also improved `halo2` verification speed in PR #4699.

### Designs

When we impose backpressure on the syncer, we need to verify the lowest blocks first.
This means that the backpressure has to be implemented in the syncer task.
(After that, the individual block download and verify tasks can run out of order.)

Here is one possible fix:
- when doing checkpoint verification, let the syncer look ahead 400-1200 blocks
- when doing full verification, let the syncer look ahead 1-`max_concurrent_block_requests` blocks

400 is the minimum for checkpointing, because the maximum checkpoint gap is 400 blocks.
The `max_concurrent_block_requests` default might need to be tuned for good full verification performance.

Here is where the syncer limit is implemented:
https://github.com/ZcashFoundation/zebra/blob/54efbe9d2d0003142e0e3a08772722e24eb34127/zebrad/src/components/sync.rs#L425-L436

### LINKED ISSUE #4729: Stop batch verification lagging and failing entire batches
## Motivation

The batch verifier can lag and fail entire blocks, even if the proofs and signatures are valid. We should re-design it so it can't lag.

If we can't do that, the dropped verifications should fail immediately, rather than waiting for the block verification timeout:

```
2022-06-30T17:51:02.943983ZERROR{net="Main"}:sync:try_to_sync:extend_tips:zebra_consensus::primitives::redpallas: batch verification receiver lagged and lost verification results
2022-06-30T17:51:57.423166Z INFO{net="Main"}:zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.819% current_height=Height(1718380) network_upgrade=Nu5 remaining_sync_blocks=3119 time_since_last_state_block=1m
...
2022-06-30T17:54:57.425405Z INFO{net="Main"}:zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.819% current_height=Height(1718380) network_upgrade=Nu5 remaining_sync_blocks=3122 time_since_last_state_block=4m
2022-06-30T17:55:57.426619Z INFO{net="Main"}:zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.819% current_height=Height(1718381) network_upgrade=Nu5 remaining_sync_blocks=3121 time_since_last_state_block=0s
2022-06-30T17:56:47.835664Z WARN{net="Main"}:sync:try_to_sync:zebrad::components::sync: error downloading and verifying block e=Invalid { error: Block(Transaction(InternalDowncastError("downcast to known transaction error type failed, original error: Elapsed(())"))), height: Height(1718445), hash: block::Hash("0000000000f0f61e6f42984784ad367711c0b3e704e840797606314426dc2a90") }
```

https://github.com/ZcashFoundation/zebra/runs/7135923149?check_suite_focus=true#step:6:644

## Designs

We can either:
- replace the batch verifier with a watch channel, and create a new watch channel for each batch
- create a new broadcast channel for each batch, so it only ever has one result in it, and make the channel size 1

teor2345:
## Motivation

We're seeing a lot of sync failures in CI and on developer machines, because some large orchard blocks take a long time to verify.

Depends-On: #4752
Close #4715
Close #4729 
Close #4650

## Solution

Until we implement orchard batching, temporarily limit full verification to 30 blocks in parallel.

Syncer changes:
- Limit the number of blocks submitted to download and verify
    - By the syncer
    - By the inbound service
- Keep unused extra hashes and submit them to the sync downloader later
- Split the checkpoint and full verify concurrency configs
    - Adjust lookahead limits when transitioning to full verification
    - Use a lower concurrency limit during full verification
    - Decrease full verification concurrency to 5 blocks

Related changes:
- Return the maximum checkpoint height from the chain verifier
- Return the verified block height from the sync downloader (actually not needed, I can remove these changes if we want)
- Remove a verbose log

## Review

This bug causes a lot of CI workflows to run slow or fail, so it's a high priority.

I'm still testing it locally, it also needs a full sync test.

### Reviewer Checklist

  - [ ] Existing CI sync tests pass

### DIFF
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -2,7 +2,7 @@
 //!
 //! It is used when Zebra is a long way behind the current chain tip.
 
-use std::{collections::HashSet, pin::Pin, sync::Arc, task::Poll, time::Duration};
+use std::{cmp::max, collections::HashSet, pin::Pin, sync::Arc, task::Poll, time::Duration};
 
 use color_eyre::eyre::{eyre, Report};
 use futures::stream::{FuturesUnordered, StreamExt};
@@ -251,6 +251,9 @@ where
 
     /// The lengths of recent sync responses.
     recent_syncs: RecentSyncLengths,
+
+    /// The highest block this syncer has verified.
+    verified_height: Option<Height>,
 }
 
 /// Polls the network to determine whether further blocks are available and
@@ -350,6 +353,7 @@ where
             latest_chain_tip,
             prospective_tips: HashSet::new(),
             recent_syncs,
+            verified_height: None,
         };
 
         (new_syncer, sync_status)
@@ -766,7 +770,9 @@ where
             let response = self.downloads.next().await.expect("downloads is nonempty");
 
             match response {
-                Ok(hash) => trace!(?hash, "verified and committed block to state"),
+                Ok(response) => self
+                    .handle_block_response(Ok(response))
+                    .expect("never returns Err for Ok"),
                 Err(error) => {
                     // TODO: exit syncer on permanent service errors (NetworkError, VerifierError)
                     if Self::should_restart_sync(&error) {
@@ -813,7 +819,12 @@ where
         response: Result<(Height, block::Hash), BlockDownloadVerifyError>,
     ) -> Result<(), BlockDownloadVerifyError> {
         match response {
-            Ok(hash) => trace!(?hash, "verified and committed block to state"),
+            Ok((height, hash)) => {
+                trace!(?height, ?hash, "verified and committed block to state");
+
+                // Blocks commit to the state in order, but their results can return out of order
+                self.verified_height = Some(max(height, self.verified_height.unwrap_or(height)));
+            }
             Err(_) => return Self::handle_response(response.map(|_| ())),
         }
 
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -2,7 +2,7 @@
 //!
 //! It is used when Zebra is a long way behind the current chain tip.
 
-use std::{cmp::max, collections::HashSet, pin::Pin, sync::Arc, task::Poll, time::Duration};
+use std::{collections::HashSet, pin::Pin, sync::Arc, task::Poll, time::Duration};
 
 use color_eyre::eyre::{eyre, Report};
 use futures::stream::{FuturesUnordered, StreamExt};
@@ -265,9 +265,6 @@ where
 
     /// The lengths of recent sync responses.
     recent_syncs: RecentSyncLengths,
-
-    /// The highest block this syncer has verified.
-    verified_height: Option<Height>,
 }
 
 /// Polls the network to determine whether further blocks are available and
@@ -378,7 +375,6 @@ where
             latest_chain_tip,
             prospective_tips: HashSet::new(),
             recent_syncs,
-            verified_height: None,
         };
 
         (new_syncer, sync_status)
@@ -387,8 +383,6 @@ where
     /// Runs the syncer to synchronize the chain and keep it synchronized.
     #[instrument(skip(self))]
     pub async fn sync(mut self) -> Result<(), Report> {
-        self.verified_height = self.state_tip().await?;
-
         // We can't download the genesis block using our normal algorithm,
         // due to protocol limitations
         self.request_genesis().await?;
@@ -796,10 +790,6 @@ where
     /// Download and verify the genesis block, if it isn't currently known to
     /// our node.
     async fn request_genesis(&mut self) -> Result<(), Report> {
-        if self.verified_height.is_some() {
-            return Ok(());
-        }
-
         // Due to Bitcoin protocol limitations, we can't request the genesis
         // block using our standard tip-following algorithm:
         //  - getblocks requires at least one hash
@@ -807,7 +797,7 @@ where
         //  - the genesis hash is used as a placeholder for "no matches".
         //
         // So we just download and verify the genesis block here.
-        while self.state_tip().await?.is_none() {
+        while !self.state_contains(self.genesis_hash).await? {
             info!("starting genesis block download and verify");
 
             let response = self.downloads.download_and_verify(self.genesis_hash).await;
@@ -880,7 +870,8 @@ where
 
         // When the state is empty, we want to verify using checkpoints
         let verified_height: usize = self
-            .verified_height
+            .latest_chain_tip
+            .best_tip_height()
             .unwrap_or(Height(0))
             .0
             .try_into()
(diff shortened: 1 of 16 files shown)


teor2345:
Block 1718686 (or nearby blocks) might be a problem, it takes about 5 minutes to verify on a 36 core machine, even with a 5 block concurrency limit.

But I'd be happy to merge this PR, and tune the default concurrency limit later.

zebrad/src/config.rs
```diff
@@ -174,16 +174,15 @@ impl Default for MetricsSection {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields, default)]
 pub struct SyncSection {
-    /// The maximum number of concurrent block download requests during sync.
+    /// The number of parallel block download requests.
```
teor2345, line 177: The config changes document the goals of this ticket - split the download, checkpoint verify, and full verify concurrency limits.

zebrad/src/components/sync.rs
```diff
@@ -789,34 +834,92 @@ where
...
             self.downloads.download_and_verify(hash).await?;
         }
 
-        Ok(())
+        Ok(extra_hashes)
+    }
+
+    /// The configured lookahead limit, based on the currently verified height,
+    /// and the number of hashes we haven't queued yet..
+    fn lookahead_limit(&self, new_hashes: usize) -> usize {
```
teor2345, line 867: Dynamically calculate different lookahead limits for full and checkpoint verification.

zebrad/src/components/sync.rs
```diff
@@ -789,34 +834,92 @@ where
...
+            hashes.len = hashes.len(),
+            ?lookahead_limit,
+            "requesting blocks",
+        );
+
+        let extra_hashes = if hashes.len() > lookahead_limit {
+            hashes.split_off(lookahead_limit)
+        } else {
+            IndexSet::new()
+        };
```
teor2345, line 856: Return any excess block hashes above the lookahead limit.

zebrad/src/components/sync.rs
```diff
@@ -397,58 +438,62 @@ where
...
+                    in_flight = self.downloads.in_flight(),
+                    extra_hashes = extra_hashes.len(),
+                    lookahead_limit = self.lookahead_limit(extra_hashes.len()),
+                    state_tip = ?self.latest_chain_tip.best_tip_height(),
+                    "requesting more blocks",
+                );
+
+                let response = self.request_blocks(extra_hashes).await;
+                extra_hashes = Self::handle_hash_response(response)?;
+            } else {
```
teor2345, line 483: When we run out of pending blocks, submit some extra hashes.

zebra-consensus/src/primitives/halo2.rs
```diff
@@ -249,12 +251,16 @@ impl Service<BatchControl<Item>> for Verifier {
...
-                        .verify(thread_rng(), self.vk)
-                        .map_err(Halo2Error::from),
-                );
+                // # Correctness
+                //
+                // Do CPU-intensive work on a dedicated thread, to avoid blocking other futures.
+                //
+                // TODO: use spawn_blocking to avoid blocking code running concurrently in this task
+                let result = tokio::task::block_in_place(|| batch.verify(thread_rng(), self.vk));
+                let _ = self.tx.send(result.map_err(Halo2Error::from));
```
teor2345: This stops Orchard proofs blocking other futures.

zebra-consensus/src/primitives/halo2.rs
```diff
@@ -152,7 +152,9 @@ pub static VERIFIER: Lazy<
     Fallback::new(
         Batch::new(
             Verifier::new(&VERIFYING_KEY),
-            super::MAX_BATCH_SIZE,
+            // TODO: replace with MAX_BATCH_ITEMS once batch verification is implemented (#4672)
+            // (the latency is ignored with only one item)
+            1,
```
teor2345: Verify proofs immediately, rather than waiting for their (fake) batch to finish.

zebra-consensus/src/primitives/redpallas.rs
```diff
@@ -112,8 +112,16 @@ impl Service<BatchControl<Item>> for Verifier {
...
+
                 let batch = mem::take(&mut self.batch);
-                let _ = self.tx.send(batch.verify(thread_rng()));
+                // # Correctness
+                //
+                // Do CPU-intensive work on a dedicated thread, to avoid blocking other futures.
+                //
+                // TODO: use spawn_blocking to avoid blocking code running concurrently in this task
+                let result = tokio::task::block_in_place(|| batch.verify(thread_rng()));
+                let _ = self.tx.send(result);
```
teor2345: We were also seeing errors in the redpallas verifier, so it moved it to a blocking thread as well.

conradoplg:
FYI I tried this out and my node still gets stuck at height 1719629

oxarbitrage (dismissed):
The code looks good to me however i am not sure if it brings too much practical improvements. Teor reported some improvements in the sync using this PR but further (unrelated?) problems a bit further. Conrado reported no benefits.

With that said, feel free to merge.

teor2345:
It looks like this isn't actually the underlying issue, so I'd like to split this PR up before we merge.

teor2345:
I'm running a full sync test here, to check if this solves the remaining syncer stalls:
https://github.com/ZcashFoundation/zebra/actions/runs/2619418857

It should be faster than PR #4752.

teor2345:
(I removed the batch verifier changes from this PR, I'll submit them in another PR.)

teor2345:
Using a partly-synced local Zebra instance, I can get Zebra to sync to the tip with:
- #4752 
- #4726

I'm running full syncs to confirm, but we might want to merge them anyway, so we can move on with other work.

oxarbitrage (approved):

gustavovalverde:
Admin merged, as the full sync got stuck and we need the fix anyways.

PR #4730, opened 2022-06-30
fix(ci): Split out Canopy logs into a separate job

teor2345:
## Motivation

We've seen timeouts when checkpointing 1.2 million blocks from Sapling to Canopy in a single GitHub Actions job.

## Solution

- Split into Sprout-Heartwood and Canopy-Nu5 jobs (about 600K blocks each)

## Review

This is urgent because it stops CI failures and timeouts.

### Reviewer Checklist

  - [ ] Existing tests pass
  - [ ] Full sync passes

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -510,13 +510,12 @@ jobs:
           ${{ inputs.test_id }} | \
           tee --output-error=exit /dev/stderr | \
           grep --max-count=1 --extended-regexp --color=always \
-          '(estimated progress.*network_upgrade.*=.*Sapling)|(test result:.*finished in)' \
+          '(estimated progress.*network_upgrade.*=.*Sapling)|(estimated progress.*network_upgrade.*=.*Blossom)|(estimated progress.*network_upgrade.*=.*Heartwood)|(estimated progress.*network_upgrade.*=.*Canopy)|(estimated progress.*network_upgrade.*=.*Nu5)|(test result:.*finished in)' \
           "
 
-  # follow the logs of the test we just launched, up to the last checkpoint (or the test finishing)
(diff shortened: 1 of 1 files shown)


teor2345:
I'm running a full sync and checkpoint sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2593382789

The job structure is correct, so it will probably work.

teor2345:
> I'm running a full sync and checkpoint sync here: https://github.com/ZcashFoundation/zebra/actions/runs/2593382789
> 
> The job structure is correct, so it will probably work.

The split succeeded, but one job failed due to an unrelated bug.

teor2345:
I'm not sure if we need this, we might have found the bug causing slow syncs.

teor2345:
We need this fix, the Sapling-Nu5 checkpoints are taking longer than 6 hours and timing out:
https://github.com/ZcashFoundation/zebra/runs/7218894297

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -510,14 +510,67 @@ jobs:
...
+          --ssh-flag="-o ServerAliveInterval=5" \
+          --command \
+          "\
+          docker logs \
+          --tail all \
+          --follow \
+          ${{ inputs.test_id }} | \
+          tee --output-error=exit /dev/stderr | \
+          grep --max-count=1 --extended-regexp --color=always \
+          '(estimated progress.*network_upgrade.*=.*Canopy)|(estimated progress.*network_upgrade.*=.*Nu5)|(test result:.*finished in)' \
```
dconnolly, line 567: Is this regex the one we want?

dconnolly: Seems yes, approving

dconnolly (approved):

PR #4752, opened 2022-07-05
fix(verify): Only verify halo2 proofs once per transaction

teor2345:
## Motivation

There is only one aggregated halo2 proof per transaction, but Zebra is verifying the same proof for each action.

This uses a lot of RAM and CPU, and stops Zebra syncing to the tip.

## Solution

Only verify the halo2 proof once per transaction.

## Review

This bug is blocking almost all other work, it's urgent.

### Reviewer Checklist

  - [x] All Halo2 proofs and Orchard Actions are being verified
  - [ ] CI passes
  - [ ] Full sync passes

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -909,25 +909,25 @@ where
         let mut async_checks = AsyncChecks::new();
 
         if let Some(orchard_shielded_data) = orchard_shielded_data {
+            // # Consensus
+            //
+            // > The proof 𝜋 MUST be valid given a primary input (cv, rt^{Orchard},
+            // > nf, rk, cm_x, enableSpends, enableOutputs)
+            //
+            // https://zips.z.cash/protocol/protocol.pdf#actiondesc
+            //
+            // Queue the verification of the Halo2 proof for each Action
+            // description while adding the resulting future to our
+            // collection of async checks that (at a minimum) must pass for
+            // the transaction to verify.
+            async_checks.push(
+                primitives::halo2::VERIFIER
+                    .clone()
+                    .oneshot(primitives::halo2::Item::from(orchard_shielded_data)),
+            );
+
             for authorized_action in orchard_shielded_data.actions.iter().cloned() {
                 let (action, spend_auth_sig) = authorized_action.into_parts();
-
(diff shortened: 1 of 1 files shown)


teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2618944705

It should finish in around 6 hours.

oxarbitrage (dismissed):
Definitely good, no need to do this inside the loop. Nice find.

oxarbitrage:
> It should finish in around 6 hours.

It seems this should be merged anyway even if it does not make the difference in time right now, it is a performance fix.

zebra-consensus/src/transaction.rs
```diff
@@ -909,26 +909,26 @@ where
...
+            //
+            // > The proof 𝜋 MUST be valid given a primary input (cv, rt^{Orchard},
+            // > nf, rk, cm_x, enableSpends, enableOutputs)
+            //
+            // https://zips.z.cash/protocol/protocol.pdf#actiondesc
+            //
+            // Queue the verification of the Halo2 proof for each Action
+            // description while adding the resulting future to our
+            // collection of async checks that (at a minimum) must pass for
+            // the transaction to verify.
```
upbqdn: I think we don't queue the proof for each Action anymore. Maybe something like this?

```suggestion
            // Unlike Sapling, Orchard shielded transactions have a single 
            // aggregated Halo2 proof per transaction, even with multiple 
            // Actions in one transaction. So we queue it for verification 
            // only once instead of queuing it up for every Action description.
```

upbqdn (commented):
That's a great find!

upbqdn (dismissed):
All looks good.

dconnolly (dismissed):
Double-checked the spec and the existing Halo2 verifier service, looks good!

When we push a new `halo2::Item` to the batch verifier, it produces `orchard::circuit::Instance`s from each Action in the transacton, and then the whole set of `Instance`s is passed as input to `verify()` a single `Halo2` aggregate proof, so this change is correct.

dconnolly (approved):

gustavovalverde:
Admin merged, as the full sync got stuck and we need the fix anyways.

PR #4756, opened 2022-07-06
return errors from `send_periodic_heartbeats_with_shutdown_handle`

### LINKED ISSUE #4733: Return connection errors from the heartbeat task to the `Client`
## Motivation

We're seeing  "failed servers must set their error slot" panics when Zebra is overloaded.

This panic is caused by the design of the `Client` connection code. Currently, the `Client` tasks share an `ErrorSlot` mutex, which must be updated before they exit.

This ticket stops using the error slot in the heartbeat task.

### Designs

Instead of using the error slot, `Client` tasks should exit and return an error to the client instance.
The `Client` should exit when any task errors or stops.

We could try something like this:
(click for details)

<details>

Replace `Client.error_slot` with returning a `PeerError`:
- end the task when there is a permanent connection error, returning `Result<(), PeerError>’
- check the error via the `JoinHandle` in the `Client`
- shut down the `Client` when the heartbeat task exits or errors

</details>

## Related Work

This is part of:
- #3263

oxarbitrage:
## Motivation

We want to return errors from `send_periodic_heartbeats_with_shutdown_handle`. When finished this should close https://github.com/ZcashFoundation/zebra/issues/4733

Right now, this is a draft to ask @teor2345 some questions as i am unsure if this is complete. I think we can't remove `error_slot` yet as it is still used by the connection task.

## Solution

Return `Result` from `send_periodic_heartbeats_with_shutdown_handle`, then follow compiler to make it work.

## Review

@teor2345 , i am unsure if this is complete, that could be missing or if there are any problems. Will like you to take an initial look when you can. 

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -25,6 +25,7 @@ use crate::{
         external::{types::Version, InventoryHash},
         internal::{Request, Response},
     },
+    BoxError,
 };
 
 use super::{ErrorSlot, PeerError, SharedPeerError};
@@ -60,7 +61,7 @@ pub struct Client {
     pub(crate) connection_task: JoinHandle<()>,
 
     /// A handle to the task responsible for sending periodic heartbeats.
-    pub(crate) heartbeat_task: JoinHandle<()>,
+    pub(crate) heartbeat_task: JoinHandle<Result<(), BoxError>>,
 }
 
 /// A signal sent by the [`Client`] half of a peer connection,
@@ -438,7 +439,7 @@ impl Client {
                 // Heartbeat task is still running.
                 Ok(())
             }
-            Poll::Ready(Ok(())) => {
+            Poll::Ready(Ok(_)) => {
                 // Heartbeat task stopped unexpectedly, without panicking.
                 self.set_task_exited_error("heartbeat", PeerError::HeartbeatTaskExited)
             }
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -431,7 +431,10 @@ impl Client {
             .is_ready();
 
         if is_canceled {
-            return self.set_task_exited_error("heartbeat", PeerError::HeartbeatTaskExited);
+            return self.set_task_exited_error(
+                "heartbeat",
+                PeerError::HeartbeatTaskExited("Task was cancelled".to_string()),
+            );
         }
 
         match self.heartbeat_task.poll_unpin(cx) {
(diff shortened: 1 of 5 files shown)


zebra-network/src/peer/client.rs
```diff
@@ -438,7 +439,7 @@ impl Client {
                 // Heartbeat task is still running.
                 Ok(())
             }
-            Poll::Ready(Ok(())) => {
+            Poll::Ready(Ok(_)) => {
```
oxarbitrage: This and the panic below it generates me some doubts. Do you think we should return `Ok` here and `PeerError` below to avoid the panic ?

teor2345: Good catch, I think this is partly an existing bug.

Since we added a result to the JoinHandle, the type is now `Poll<Result<Result<(), PeerError>, JoinError>`.

There's a few different cases we need to handle:
- if the `Err` is [`JoinError::is_cancelled()`](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#method.is_cancelled), then some other task told the heartbeat task to exit, so we can return a `PeerError` (feel free to make a new variant)
- if the `Err` is `JoinError::is_panic()`, then we need to re-panic here, we can't recover from a panic. There's [a good example of how to do that in the `JoinError::try_into_panic()` docs](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#method.try_into_panic)
- if the inner result is a `Ok(Err(PeerError))`, we can return that error
- if the inner result is `Ok(Ok(()))`, we can return `PeerError::HeartbeatTaskExited`
- otherwise the `Poll::Pending` code stays the same

We missed the cancelled case when we last changed this code. Or it might be a new tokio feature, I'm not sure.

teor2345 (commented):
This looks really good, it just needs a quick fix for the panic handling, as you said.

Good pickup on that bug, it seems like it's been there for a while.

zebra-network/src/peer/client.rs
oxarbitrage: this is just an idea, i am ok on doing it  different https://github.com/ZcashFoundation/zebra/pull/4756/commits/90526e31fe78fbdd3dd43eab2cc4a5315fc36f89

Basically added an argument to `HeartbeatTaskExited`. Let me know how do you see it.

teor2345 (approved):
Thanks this looks great!

zebra-network/src/peer/client.rs
teor2345: That's a great idea, it makes it clear that the heartbeat task stopped, whatever the error actually was.

oxarbitrage:
@mergify update

teor2345:
There are other tasks that are higher priority, so I moved this one to medium.

PR #4199, opened 2022-04-26
2. fix(perf): When writing blocks to disk, don't block other async tasks

teor2345:
## Motivation

We want to improve Zebra's sync speed to avoid test timeouts in CI (#4155).

One way to speed up the sync is move blocking tasks to dedicated threads, so they don't stop other async tasks on the same thread from running. This PR moves disk writes to a `tokio` blocking thread.
Close #4779

Partially addresses #2188

Depends-On: #4792

### Specifications

https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html

## Solution

- Move CommitBlock and CommitFinalizedBlock requests to a `tokio` block-in-place thread

Related changes:
- Remove mempool downloader requirement for a populated state

Test fixes:
- Add blocking tokio threads to tests that now need them
    - Remove `tokio::time::pause()` from tests that don't need it (it doesn't work with blocking threads)
- In tests that need `tokio::time::pause()`, stop writing blocks to the state
    - Stop requiring a populated state in the mempool downloader
- Improve failure debug output in some tests
- Improve reliability of some tests with multithreaded executors

## Review

Anyone can review this PR, it's urgent because CI is failing, and it blocks a release.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

## Follow Up Work

- #4788

PR #4757, opened 2022-07-06
change(state): Wrap commitment trees into `Arc`

### LINKED ISSUE #4766: Store note commitment trees in `Arc`s in the non-finalized state
This is a part of #3757.

Storing note commitment trees in `Arc`s saves memory and makes cloning cheaper.

upbqdn:
## Motivation

Zebra maintains a copy of the same note commitment tree with each non-finalized state when handling multiple non-finalized states. It would be better to maintain only `Arc`s, and use copy-on-write.

## Solution

This PR wraps Sprout, Sapling and Orchard note commitment trees into `Arc`s in both finalized and non-finalized states.

Closes #4766.

## Review

We should check if the changes affect syncing speed.

### DIFF
--- a/zebra-state/src/service/check/anchors.rs
+++ b/zebra-state/src/service/check/anchors.rs
@@ -1,7 +1,7 @@
 //! Checks for whether cited anchors are previously-computed note commitment
 //! tree roots.
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
 use zebra_chain::sprout;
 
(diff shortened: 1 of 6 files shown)


teor2345 (commented):
Looks good!

teor2345:
It looks like these changes might help with some state hangs, but we need to test them after PR #4776 merges.

teor2345 (approved):
Looks great, thanks for this!

teor2345:
I admin-merged this PR because:
- it should improve performance, and
- I want to rebase PR #4199 on top of it

upbqdn:
I thought I marked this PR as ready for review more than a week ago, but apparently, I didn't. Sorry for that.

Issue #3263, opened 2021-12-20
Security: Remove the ErrorSlot from the heartbeat and connection tasks


teor2345:
## Motivation

The error handling for the `Connection` and heartbeat tasks is really messy. It has been really buggy in the past, and it's hard to understand or modify. The way the code is written can cause hangs or panics.

At the moment, it's causing a "failed servers must set their error slot" panic when Zebra is overloaded.

### Scheduling

This is a denial of service risk which might be triggerable remotely.

This is high-risk hang code with multiple mutexes, so it should be prioritised over other potential hang or panic bugs.

### Sub-Tasks

High-priority ongoing connection panics:
- [ ] #4770
- [x] #4733
- [x] #4734
- [x] #4735

Other panics that are very rare:
- [x] #4736

Cleanups:
- [x] #4737

This will give us a much stronger guarantee that the connection task will exit when there's an error.

## Related Work

- #3254

ftm1000:
redistributing issues that can be separately worked from Epic #2311

ftm1000:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/3263?planning-poker) with ZenHub @conradoplg @dconnolly @gustavovalverde @jvff @oxarbitrage @teor2345 @upbqdn

teor2345:
This cleanup isn't needed any time soon.

teor2345:
I got a panic in this code, so I'm re-opening this ticket.

I'll add the panic info to the ticket, and the details in a comment.

teor2345:
Here is the panic I saw recently:

```
Message:  failed servers must set their error slot
Location: zebra-network/src/peer/client.rs:496
```

It looks like it might have been triggered by switching to another chain fork.

Here are the detailed logs:

<details>

```
2022-06-24T14:24:44.018156Z  INFO {zebrad="135aedc" net="Main"}: zebrad::components::mempool::gossip: sending mempool transaction broadcast request=AdvertiseTransactionIds({Witnessed(WtxId { id: transaction::Hash("3fa349d6b60b6a70e23af69513faceafc52ce13061a0538de364691cac2a8252"), auth_digest: AuthDigest("9a830fe29ab5db38201f44e662aa85edd146f943f687b467ad5bfd270888c5e5") })})
[...]

PR #4770, opened 2022-07-11
fix(panic): Stop panicking when the connection error slot is not set

teor2345:
## Motivation

Every few days, my Zebra instances panic with "failed servers must set their error slot".
This PR fixes the panic in ticket #3263 by removing it.

We still need to clean up the code to avoid deadlocks, but that is less urgent.

## Solution

- Remove the `expect()` and use a `ConnectionTaskExited` error instead of panicking

## Review

I'd like to get this fix in this release, so reviewing this PR is urgent.

### Reviewer Checklist

  - [ ] Existing tests pass

## Follow Up Work

The rest of #3263.

### DIFF
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -19,15 +19,13 @@ use tokio::{sync::broadcast, task::JoinHandle};
 use tower::Service;
 
 use crate::{
-    peer::error::AlreadyErrored,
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
I think this is good enough for now, i was working on a more complete attempt at https://github.com/ZcashFoundation/zebra/pull/4771 but it will take some more time.

teor2345:
This fixes a panic we're seeing every few days on `main`, so we're ok to merge without all the tests passing.

Issue #4581, opened 2022-06-08
Investigate "tasks have lost their waker" and fix if needed


conradoplg:
## Motivation

`tokio-console`, for which we added support in https://github.com/ZcashFoundation/zebra/pull/4519, is reporting a bunch os "tasks have lost their waker". We should investigate.

Most of them seem to be from the handshake code:
https://github.com/ZcashFoundation/zebra/blob/ca0520b2e8e7c40fa4c24da45896ff9430992a9d/zebra-network/src/peer_set/initialize.rs#L298-L299

https://github.com/ZcashFoundation/zebra/blob/766dd9357b780e49d95398c85259ac1ff4290fa5/zebra-network/src/peer_set/initialize.rs#L757-L758

https://github.com/ZcashFoundation/zebra/blob/766dd9357b780e49d95398c85259ac1ff4290fa5/zebra-network/src/peer/handshake.rs#L1016-L1020

![image](https://user-images.githubusercontent.com/35766/172649593-f8dcfe38-47e6-4cf4-8520-c1a32a1408fa.png)

### Designs

- [ ] Investigate the cause of the issue
- [ ] If it's an issue, fix it if possible. If it's too complex, create another issue for it.
- [ ] If it's a false positive, consider reporting it to `tokio-console`

teor2345:
I've had a look at all the lost wakers, and they seem pretty harmless.

If we're just dropping wakers for handshakes we're not using any more, that's probably a task cleanup bug. And we make new handshakes as needed anyway.

teor2345:
The warning might also be incorrect or confusing, I'm still seeing the poll number for those tasks going up, even after the lost waker warning.

But handshakes should only last for 4 seconds, so there's definitely a bug here.

teor2345:
This appears to be a known bug in tokio-console, related to tasks spawning other tasks:
https://github.com/tokio-rs/console/issues/345

PR #4772, opened 2022-07-12
fix(net): Fix handshake timing and error handling

teor2345:
## Motivation

Zebra's connection handshake code might be able to hang, it can panic silently, and initial connections are all happening at the same time.

I found these when investigating ticket #4581.

## Solution

- Actually add a delay between spawning initial handshakes 
- Forward panics in the handshake code to the calling task
- Wrap the entire handshake in a timeout, not just some network communication

## Review

These bug fixes can go in whenever we're ready, they're not urgent.

### Reviewer Checklist

  - [ ] Tests for Errors


zebra-network/src/peer_set/initialize.rs
```diff
@@ -289,21 +289,24 @@ where
...
-                task
-            })
+            let outbound_connector = outbound_connector.clone();
+
+            // Spawn a new task to make the outbound connection.
+            tokio::spawn(
+                async move {
+                    // Only spawn one outbound connector per `MIN_PEER_CONNECTION_INTERVAL`,
+                    // sleeping for an interval according to its index in the list.
+                    sleep(constants::MIN_PEER_CONNECTION_INTERVAL.saturating_mul(i as u32)).await;
```
upbqdn, line 299: Good catch.

upbqdn (approved):
All looks good. It was worth following the individual commits.

teor2345:
@mergifyio update

PR #4776, opened 2022-07-13
feat(verify): Concurrently verify proof and signature batches

teor2345:
## Motivation

Currently, Zebra only verifies one batch of each type of cryptography.
This PR verifies multiple batches in parallel, up to the number of available CPUs.

In practice, I've seen it use all the physical cores on my machine for a few seconds. To see this, I needed to set the number of concurrent blocks to 100.

After the blocks all get verified, Zebra waits for other CPU-bound tasks, or writing the blocks to disk.

(We can improve performance more by moving other CPU-bound tasks to `rayon`, and disk-bound tasks to blocking `tokio` threads. But that belongs in a separate ticket.)

### Specifications

`tokio`'s warning about CPU-bound tasks:
https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code

`rayon::scope_fifo()`:
https://docs.rs/rayon/latest/rayon/fn.scope_fifo.html

## Solution

Increase the batch item queue size:
- Queue enough batch items so there's a batch ready for each available CPU
- During full verification, verify 20 blocks concurrently

Run all verifier cryptography on blocking CPU-bound threads:
- Initialize the rayon thread pool with a configurable number of threads
- Spawn on a rayon thread pool inside `tokio::spawn_blocking()`
- Spawn multiple batches concurrently
- Store a list of concurrently running batches
- Clear pending batches once they finish
- Stop accepting new batch items once we're at the concurrent batch limit

Related cleanups:
- Simplify batch worker loop
- Improve tracing output
- Improve error handling

Related fixes:
- Run docker tests on PR series
 
### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] CI passes

PR #4777, opened 2022-07-13
chore(sync): Update Zebra mainnet and testnet checkpoints - 13 July 2022

teor2345:
## Motivation

Updating the checkpoints speeds up Zebra's syncing.

## Solution

- Update Zebra mainnet and testnet checkpoints on 13 July 2022

## Review

This is urgent because we want it in the next release.

### Reviewer Checklist

  - [ ] CI might not pass, because syncing needs other fixes

## Follow Up Work

- #4776 is also needed to make syncing work


oxarbitrage (approved):
Thanks @teor2345

oxarbitrage:
>  #4776 is also needed to make syncing work

Not really from here, i synchronized the mainnet with the main branch locally. It was a slow process at the end (toke around 14 hours) but never hanged or crashed. I think the checkpoint update will help to make sync faster given the current block sizes we are seeing in the network.

Anyways, everything seems to point to a problem in the batches so i think your PR will maybe fix the underlying problem. I will take a look into it now.

teor2345:
> > #4776 is also needed to make syncing work
> 
> Not really from here, i synchronized the mainnet with the main branch locally. It was a slow process at the end (toke around 14 hours) but never hanged or crashed. I think the checkpoint update will help to make sync faster given the current block sizes we are seeing in the network.
> 
> Anyways, everything seems to point to a problem in the batches so i think your PR will maybe fix the underlying problem. I will take a look into it now.

There's a sync problem that only seems to happen on some machines or networks, so it's tricky to reproduce and fix.
We see it in CI more than in other places.

teor2345:
This passed the CI we expect it to pass, so I admin-merged it.

PR #4786, opened 2022-07-18
refactor(build): use FHS for deployments and artifacts

### LINKED ISSUE #3432: Use the Filesystem Hierarchy Standard (FHS) for deployments and artifacts
## Motivation

Linux has its own filesystem convention, which is standard and considered by most sysadmins when creating products, OS-level restrictions, and even when taking security measures.

We should respect this convention for compatibility with different deployment mechanisms, and when high-security standards are being in placed (most directories are being restricted)

A good example is the Container-Optimized OS used by Google: https://cloud.google.com/container-optimized-os/docs/concepts/disks-and-filesystem#working_with_the_file_system

### Specifications

Filesystem Hierarchy Standard - v3.0: https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html
Linux man for `hier`: https://man7.org/linux/man-pages/man7/hier.7.html

## Related Work
#3344

gustavovalverde:
## Motivation

Linux has its own filesystem convention, which is standard and considered by most sysadmins when creating products, OS-level restrictions, and even when taking security measures.

We should use this convention in our deployments and artifacts, so we don't get constrained in platforms which had restrictions based on this convention.

Fixes #3432

### Specifications

Filesystem Hierarchy Standard - v3.0: https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html
Linux man for `hier`: https://man7.org/linux/man-pages/man7/hier.7.html

## Solution

Do not use the `/app` directory to keep build binaries, as this does not comply with FHS. The correct place is `/opt/<app-name>`

## Review

### Reviewer Checklist

  - [ ] Tests for Errors


docker/zcash-lightwalletd/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN go mod download
 COPY . ./
 
 # Build and install the binary.
-RUN go build -v -o /lightwalletd
+RUN go build -v -o /opt/lightwalletd
```
teor2345, line 27: After this change, will these entrypoints still work?
Why are they different?

https://github.com/ZcashFoundation/zebra/blob/54e17d40acf79308f8bd0cb79e0545d1b1fca114/docker/zcash-lightwalletd/Dockerfile#L39
https://github.com/ZcashFoundation/zebra/blob/54e17d40acf79308f8bd0cb79e0545d1b1fca114/docker/zcash-lightwalletd/Dockerfile#L60

teor2345:
@mergifyio update

docker/zcash-lightwalletd/Dockerfile
gustavovalverde: As we do the following `COPY --from=build /opt/lightwalletd /usr/local/bin` in the `runtime` stage

We're, in the end, using `/usr/local/bin`, making the `lightwalletd` binary available as a command for execution to the user, just like any other linux command.

I removed the `/` in the final entrypoint as in this stage (`runtime`) the binary is only available in `/usr/local/bin`, not in the actual `WORKDIR`

teor2345 (approved):
Looks good!

The zcash-params job doesn't actually run until we merge to main, so can you please check main and the next PR that builds off the new zcash-params image?

(We did it this way because we didn't want to overload the parameter download server and get rate-limited.)

Issue #4788, opened 2022-07-18
Move database block and transaction fetches to a dedicated blocking and CPU-heavy thread


teor2345:
## Motivation

Zebra's sync can be slow or fail, because transaction deserialization involves some CPU-heavy cryptography.

So we want to move transaction deserialization into a `rayon` CPU thread pool `scope`.

### API Reference

Blocking threads:
https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

CPU-heavy threads:
https://docs.rs/rayon/latest/rayon/fn.scope_fifo.html

### Detailed Analysis

See https://github.com/ZcashFoundation/zebra/issues/4583#issuecomment-1186662862

### Designs

In `zebra-state`:
- move block and transaction response futures into `tokio` blocking threads
- move block or transaction deserialization into the `rayon` thread pool

Example code for `rayon` in `tokio` blocking threads:
https://github.com/ZcashFoundation/zebra/blob/81727d7df0ee57e6d4f2172c287e52e414c4e6a0/zebra-consensus/src/primitives/ed25519.rs#L121-L134

## Related Work

We might want to use the `ReadStateService` to parallelise these fetches, but that seems like a separate ticket.

teor2345:
This was fixed but then reverted, because it didn't actually speed things up.

PR #4792, opened 2022-07-19
1. fix(state): when sending `headers` to peers, only deserialize the header data from disk

### LINKED ISSUE #1924: Stop sending incorrect transaction counts in CountedHeader messages
## Motivation

Zebra supplies the number of transactions in the original block in a `CountedHeader` struct.

But that doesn't match the Bitcoin spec, which always sends zero.

### Specification

The Bitcoin wiki says:
> Number of transaction entries, this value is always 0

https://en.bitcoin.it/wiki/Protocol_documentation#Block_Headers

### Implementation in `zcashd`

`zcashd` always populates this field with zero:
https://github.com/zcash/zcash/blob/420f8dfe38fd6b2465a665324366c2ae14aa98f4/src/main.cpp#L6864-L6874

`zcashd` ignores this field from other nodes:
https://github.com/zcash/zcash/blob/420f8dfe38fd6b2465a665324366c2ae14aa98f4/src/main.cpp#L7046-L7050

teor2345:
## Motivation

When Zebra sends block headers to peers, it fetches the entire block from disk, then throws away the transactions.

This is slow for two reasons:
- fetching extra data costs disk reads
- deserializing transaction data cost CPU time

If we stop deserializing the transactions, it helps with bug #4788, because we avoid some block fetches entirely.

Close #1924, since we were removing transaction counts anyway.

## Solution

- When sending `headers` to peers, only deserialize the header data from disk
    - Call the new `block_header()` method from the existing `block()` method
- Wrap the block header in an `Arc`, so a headers response can be sent to the network without copying 200 kB of data
    - Implement `block::Hash` from `Arc<Header>`
    - Update a lot of tests to use `Arc<Header>`
- Remove incorrect counts from `CountedHeader`

## Review

Anyone can review this PR, it's urgent because it improves sync speed.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebra-chain/src/block.rs
+++ b/zebra-chain/src/block.rs
@@ -1,6 +1,6 @@
 //! Blocks and block-related structures (heights, headers, etc.)
 
-use std::{collections::HashMap, fmt, ops::Neg};
+use std::{collections::HashMap, fmt, ops::Neg, sync::Arc};
 
 use crate::{
     amount::NegativeAllowed,
@@ -46,9 +46,9 @@ pub use arbitrary::LedgerState;
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Serialize))]
 pub struct Block {
     /// The block header, containing block metadata.
-    pub header: Header,
+    pub header: Arc<Header>,
     /// The block transactions.
-    pub transactions: Vec<std::sync::Arc<Transaction>>,
+    pub transactions: Vec<Arc<Transaction>>,
 }
 
 impl fmt::Display for Block {
@@ -219,7 +219,7 @@ impl Block {
 
 impl<'a> From<&'a Block> for Hash {
     fn from(block: &'a Block) -> Hash {
-        block.header.into()
+        block.header.as_ref().into()
     }
 }
 
--- a/zebra-chain/src/block/header.rs
+++ b/zebra-chain/src/block/header.rs
@@ -1,12 +1,12 @@
 //! The block header.
 
-use std::usize;
+use std::sync::Arc;
 
 use chrono::{DateTime, Duration, Utc};
 use thiserror::Error;
 
 use crate::{
-    serialization::{CompactSizeMessage, TrustedPreallocate, MAX_PROTOCOL_MESSAGE_LEN},
+    serialization::{TrustedPreallocate, MAX_PROTOCOL_MESSAGE_LEN},
     work::{difficulty::CompactDifficulty, equihash::Solution},
 };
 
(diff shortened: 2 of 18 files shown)


zebra-chain/src/block.rs
```diff
@@ -46,9 +46,9 @@ pub use arbitrary::LedgerState;
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Serialize))]
 pub struct Block {
     /// The block header, containing block metadata.
-    pub header: Header,
+    pub header: Arc<Header>,
```
teor2345, line 49: Using `Arc<Header>` reduces memory copies and memory usage, but it needs updates to a bunch of tests.

zebra-chain/src/block/header.rs
```diff
@@ -125,18 +125,14 @@ impl Header {
...
 pub struct CountedHeader {
     /// The header for a block
-    pub header: Header,
-
-    /// The number of transactions that come after the header
-    ///
-    /// TODO: should this always be zero? (#1924)
-    pub transaction_count: CompactSizeMessage,
+    pub header: Arc<Header>,
 }
```
teor2345, line 136: We can't get the transaction count any more, and it was always wrong to send it, so it's easiest just to remove it.

zebra-state/src/service.rs
```diff
@@ -918,17 +927,10 @@ impl Service<Request> for StateService {
...
-                                .transactions
-                                .len()
-                                .try_into()
-                                .expect("transaction count has already been validated"),
-                            header: block.header,
-                        }
+                        let header = self
+                            .best_block_header(hash.into())
+                            .expect("block header for found hash is in the best chain");
+                        block::CountedHeader { header }
```
teor2345, line 932: This is the core of the bug fix: get headers from the database, rather than entire blocks.

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -79,20 +79,31 @@ impl ZebraDb {
...
+    //
+    // TODO: move this method to the start of the section
+    #[allow(clippy::unwrap_in_result)]
+    pub fn block(&self, hash_or_height: HashOrHeight) -> Option<Arc<Block>> {
+        // Block
+        let height = hash_or_height.height_or_else(|hash| self.height(hash))?;
+        let header = self.block_header(height.into())?;
+
         // Transactions
         let tx_by_loc = self.db.cf_handle("tx_by_loc").unwrap();
```
teor2345, line 108: This is where we get the block header from the database, rather than the block.

zebra-state/src/service/read.rs
```diff
@@ -77,6 +77,33 @@ where
         .or_else(|| db.block(hash_or_height))
 }
 
+/// Returns the [`block:Header`] with [`block::Hash`](zebra_chain::block::Hash) or
```
oxarbitrage, line 80:
```suggestion
/// Returns the [`block::Header`](zebra_chain::block::Header) with [`block::Hash`](zebra_chain::block::Hash) or
```

oxarbitrage (approved):
It looks great, thanks @teor2345

teor2345:
I admin-merged this PR because it is an urgent sync fix.

PR #4799, opened 2022-07-20
perf(commitments): generate sapling point outside the method

### LINKED ISSUE #4783: Avoid calling `find_group_hash` with every new `ValueCommitment`
`verify_sapling_shielded_data` is a hot task in Zebra when we have too many sapling transactions. Any improvements we can make there can help the sync performance.

There is one potential improvement we can make here:
https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/commitment.rs#L283-L286

There are two things we need to discuss first:

1 - If we don't generate these points with each new commitment then where do we do it. 

One possible option is to add them as fields of the transaction verifier struct: https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/transaction.rs#L64 

2- `find_group_hash` calls `jubjub_group_hash` which produces a random, always different, point in the curve. If we do this change we will start using the same generated point for all created note commitments. I think this will be ok but i want to make sure.

oxarbitrage:
## Motivation

Fix https://github.com/ZcashFoundation/zebra/issues/4783

## Solution

Generate the sapling points once as a global for sapling (https://github.com/ZcashFoundation/zebra/commit/f79c438ca4096d08ce27d720ee464631db934597).

Orchard is already doing it buy the lazy_static is inside the method. Maybe this is ok but i changed it anyway for consistency (https://github.com/ZcashFoundation/zebra/commit/4cbe1fcd37f3d6035890c548ec47bfb6f946dace).

## Review

Anyone can review.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebra-chain/src/orchard/commitment.rs
+++ b/zebra-chain/src/orchard/commitment.rs
@@ -301,15 +301,14 @@ impl ValueCommitment {
     /// <https://zips.z.cash/protocol/nu5.pdf#concretehomomorphiccommit>
     #[allow(non_snake_case)]
     pub fn new(rcv: pallas::Scalar, value: Amount) -> Self {
-        lazy_static! {
-            static ref V: pallas::Point = pallas_group_hash(b"z.cash:Orchard-cv", b"v");
-            static ref R: pallas::Point = pallas_group_hash(b"z.cash:Orchard-cv", b"r");
-        }
-
         let v = pallas::Scalar::from(value);
-
         Self::from(*V * v + *R * rcv)
     }
+}
+
+lazy_static! {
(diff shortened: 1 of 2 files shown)


upbqdn (approved):
All looks good.

A suggestion regarding maintenance:

It is often desirable to keep variables close to where they are used. As described here https://users.rust-lang.org/t/lazy-static-inside-a-functions-scope/31700/2, statics are not reinitialized on subsequent function calls, so having `V` and `R` inside the function helps to keep the scope of the module a bit cleaner.

teor2345:
I admin-merged this PR because it is an urgent sync fix.

PR #4804, opened 2022-07-22
fix(consensus): Increase the number of blocks checked for legacy transactions

teor2345:
## Motivation

In NU5 we added a check for previous Zebra versions that had followed an incorrect chain.

But we only check 100 blocks for legacy transactions. Sometimes testnet has no transactions for 100 blocks (2 hours), and fails the check.

## Solution

- Increase the number of blocks checked to 1000
- Improve logging

## Review

Anyone can review this PR.
This bug needs to be fixed eventually, but at the moment I've only seen it happen rarely, and only on testnet.

### Reviewer Checklist

  - [ ] Test pass

### DIFF
--- a/zebra-state/src/service/tests.rs
+++ b/zebra-state/src/service/tests.rs
@@ -314,12 +314,20 @@ proptest! {
     fn no_transaction_with_network_upgrade(
         (network, nu_activation_height, chain) in partial_nu5_chain_strategy(4, true, OVER_LEGACY_CHAIN_LIMIT, NetworkUpgrade::Canopy)
     ) {
+        let tip_height = chain
+            .last()
+            .expect("chain contains at least one block")
+            .coinbase_height()
+            .expect("chain contains valid blocks");
+
         let response = crate::service::check::legacy_chain(nu_activation_height, chain.into_iter().rev(), network)
             .map_err(|error| error.to_string());
 
         prop_assert_eq!(
             response,
-            Err("giving up after checking too many blocks".into())
+            Err(format!(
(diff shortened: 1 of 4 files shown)


teor2345:
~~I've seen this happen twice on testnet, so I think we should include it in the release, to avoid user confusion.~~

Edit: Actually testnet users should be using the `main` branch

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
@mergifyio update

oxarbitrage:
@mergifyio update

oxarbitrage (dismissed):

teor2345:
@mergifyio update

teor2345:
@mergifyio update

zebra-state/src/service/tests.rs
```diff
@@ -314,12 +314,20 @@ proptest! {
...
+            .expect("chain contains valid blocks");
+
         let response = crate::service::check::legacy_chain(nu_activation_height, chain.into_iter().rev(), network)
             .map_err(|error| error.to_string());
 
         prop_assert_eq!(
             response,
-            Err("giving up after checking too many blocks".into())
+            Err(format!(
+                "could not find any transactions in recent blocks: checked 1000 blocks back from {tip_height:?}",
```
teor2345: Arya suggested this on another PR, but it needs an import as well, I'll do that tomorrow:

```suggestion
                "could not find any transactions in recent blocks: checked {MAX_LEGACY_CHAIN_BLOCKS} blocks back from {tip_height:?}",
```

oxarbitrage (approved):

PR #4802, opened 2022-07-21
1. fix(perf): Run CPU-intensive state updates in parallel rayon threads

teor2345:
## Motivation

We want to improve Zebra's sync speed to avoid test timeouts in CI (#4155).

This PR moves CPU-intensive verification in state updates to the `rayon` thread pool.

Close #4790

Partially addresses #4583

Depends-On: #4199
Depends-On: #4795

### Specifications

https://docs.rs/rayon/latest/rayon/fn.in_place_scope_fifo.html

## Solution

- Run CPU-intensive chain validation and updates in parallel `rayon` threads
    - Run non-finalized validation and `Chain` updates in parallel
    - Run non-finalized sprout, sapling, and orchard tree updates and roots in parallel
    - After a non-finalized chain fork, do note commitment and history tree rebuilds in parallel
    - Run finalized sprout, sapling, and orchard tree updates and roots in parallel

Other performance improvements:
- Wrap `HistoryTree` in an `Arc` in the state 
- Skip generating unused interstitial Sprout treestates

Related changes:
- Remove redundant checks for empty shielded data

Test fixes:
- Improve anchor validation debugging and error messages
- Work around a test data bug, and save some CPU during tests

## Review

Anyone can review this PR, it's urgent because CI is failing, and it blocks a release.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

## Follow Up Work

- #4788

Issue #4583, opened 2022-06-08
Investigate busiest tasks per tokio-console


conradoplg:
## Motivation

tokio-console, for which we added support in https://github.com/ZcashFoundation/zebra/pull/4519, is reporting a bunch of busy tasks but it's not clear what they are. We should investigate it.

On a syncing-pre-checkpoint node, it's looking like this in my machine:

![image](https://user-images.githubusercontent.com/35766/172650585-34b311f4-2e9f-4788-831b-60cb6ff048bb.png)

The target is `tokio::task::blocking` and fields are

```
│fn=tokio::runtime::thread_pool::worker::Launch::launch::{{closure}}                                                                                                                                              │
│kind=blocking                                                                                                                                                                                                    │
│task.id=2                                                                                                                                                                                                        │
│spawn.location=<cargo>/tokio-1.18.2/src/runtime/thread_pool/worker.rs:358:13 
```

CPU usage seems to be at 200% which does not seem to reflect that. There are 24 of those tasks and my CPU has 24 cores, so it feels like CPU usage should be much higher? Not sure.

### Specifications

<!--
If this change is based on consensus rules, quote them, and link to the Zcash spec or ZIP:
https://zips.z.cash/#nu5-zips
If this changes network behaviour, quote and link to the Bitcoin network reference:
https://developer.bitcoin.org/reference/p2p_networking.html
-->

### Investigations

- [ ] Investigate if it's possible to make tokio-console report what exactly are those tasks, or find out what they are
- [ ] Investigate if the reported results make sense with the expected behaviour, and open tickets if any issues are found
- [ ] Check what function is actually running at the top of each thread's stack:
    - https://unix.stackexchange.com/questions/166541/how-to-know-where-a-program-is-stuck-in-linux
    - https://fasterthanli.me/articles/when-rustc-explodes#profiling-with-not-perf-nperf (or the `perf` section)

### Missing `tokio-console` features

- Tasks that might be blocking: https://github.com/tokio-rs/console/issues/150
- Tasks that run for a long time: https://github.com/tokio-rs/console/issues/203

## Related Work

<!--
Is this change related to other features or tickets?
-->

oxarbitrage:
> CPU usage seems to be at 200% which does not seem to reflect that. There are 24 of those tasks and my CPU has 24 cores, so it feels like CPU usage should be much higher? Not sure.

There is a typo and this should say 20% ?

conradoplg:
[...]

Issue #4155, opened 2022-04-20
Full synchronization test times out in CI


jvff:
## Motivation

The full chain synchronization test is currently timing out in CI. We should find a way to speed up synchronization so that it doesn't time out.

jvff:
One possible solution is to:
- [x] increase the checkpoint height, so that those blocks get validated and finalized more quickly - #4183

teor2345:
Some other possible solutions are:

#### Quick tweaks, likely to help

- [x] optimise Rust code at link time - #4184
- [x] increase (or maybe decrease) the number of block download retries - #4185
- [x] move database writes to a blocking thread, so they don't block network futures - #4199
- [ ] move large database reads to a blocking thread, so they don't block network futures

#### Quick tweaks, likely to help, might cause bugs on slower networks

- [x] decrease (or maybe increase) the handshake timeout - #4212
- [x] decrease (or maybe increase) the network request timeout - #4213

#### Quick tweaks, might help

- [x] reduce the default hedge to 15 seconds, to increase the initial sync retry speed - covered by #4213
- [ ] launch the full sync test in another cloud region that's closer to more Zcash peers

#### Out of scope

Needs separate tickets, priorities, scheduling:

<details>

#### Larger changes, likely to help

- move hash calculation and signature verification to a blocking thread, so they don't block network futures
    - we should only do this if the similar disk writes change helps
- profiling to see where Zebra is waiting - tokio-console #2112 
 
</details>

PR #4805, opened 2022-07-22
2. fix(perf): Run CPU-intensive state reads in parallel rayon threads

### LINKED ISSUE #4788: Move database block and transaction fetches to a dedicated blocking and CPU-heavy thread

teor2345:
## Motivation

We want to improve Zebra's sync speed to avoid test timeouts in CI (#4155).

This PR moves CPU-intensive cryptography in database transaction reads to the `rayon` thread pool.
It also moves most large state reads into blocking tokio threads.

Close #4788

Partially addresses #4583

Depends-On: #4802

## Solution

- Spawn large database reads into blocking `tokio` threads
  - Concurrently read all blocks, headers, and transactions from disk, even in the `StateService`
- Run zebra-state transaction deserialization on a `rayon` thread 

## Review

This is an urgent sync issues fix.

### Reviewer Checklist

  - [ ] Tests for Errors


oxarbitrage (approved):
looks great, thanks!

PR #4806, opened 2022-07-22
fix(rpc): Isolate RPC queries from the rest of Zebra, to improve performance

teor2345:
## Motivation

We want to use a separate tokio executor for RPCs, so that RPC queries don't slow down syncing. (And syncing doesn't slow down RPCs.)

### API Reference

This is the current behaviour of the RPC server:

> Utilize existing event loop executor to poll RPC results.

https://docs.rs/jsonrpc-http-server/18.0.0/jsonrpc_http_server/struct.ServerBuilder.html#method.event_loop_executor

But if we don't supply an executor, all threads will spawn a new runtime:

> The first thread will use provided Executor instance and all other threads will use UninitializedExecutor to spawn a new runtime for futures.

https://docs.rs/jsonrpc-http-server/18.0.0/jsonrpc_http_server/struct.ServerBuilder.html#method.threads

## Solution

- Stop sharing the tokio executor with RPCs
- Move RPC server startup and shutdown into a `std::thread`, so the RPC executor can block while it shuts down
- Add tests for RPC startup and port conflicts

## Review

This is a routine performance/reliability fix, it can be reviewed whenever we have time.

### Reviewer Checklist

  - [ ] `lightwalletd` and RPC tests pass

### DIFF
--- a/zebra-rpc/src/server.rs
+++ b/zebra-rpc/src/server.rs
@@ -28,6 +28,9 @@ use crate::{
 pub mod compatibility;
 mod tracing_middleware;
 
+#[cfg(test)]
+mod tests;
+
 /// Zebra RPC Server
 #[derive(Clone, Debug)]
 pub struct RpcServer;
--- a/zebra-rpc/src/server.rs
+++ b/zebra-rpc/src/server.rs
@@ -7,6 +7,8 @@
 //! See the full list of
 //! [Differences between JSON-RPC 1.0 and 2.0.](https://www.simple-is-better.org/rpc/#differences-between-1-0-and-2-0)
 
+use std::panic;
+
 use jsonrpc_core::{Compatibility, MetaIoHandler};
 use jsonrpc_http_server::ServerBuilder;
 use tokio::task::JoinHandle;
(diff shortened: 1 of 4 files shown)


teor2345:
This doesn't seem to help (yet!), and it needs a test fix for a concurrency bug.

Let's re-test after we've merged some other fixes.

teor2345:
It looks like this change might let different Zebra instances share the same RPC port, which isn't what we want.

Reducing the number of threads to 1 might avoid that, I'll have to check.

It's also possible that it's a spurious test failure.

teor2345:
@mergifyio update

teor2345:
Let's see if this works on top of the latest fixes.

teor2345:
@mergifyio update

teor2345:
When the RPC server spawn() fails, it drops the tokio executor it has just created. But that needs to happen in a non-async context, so the server can block until all tasks finish.

`tokio` doesn't allow executors to be dropped inside spawn_blocking(), even though it allows blocking tasks. So we might have to escape from spawn_blocking() using a std::thread. 😞

teor2345:
CI is failing due to slow syncs, so existing performance fixes are a high priority.

oxarbitrage (dismissed):
Makes sense to me. It seems to be the right thing to do.

teor2345:
@mergifyio update

oxarbitrage (approved):

upbqdn (commented):
It's still not clear to me why we have to use `std::thread::spawn` to spawn the server inside `tokio::task::spawn_blocking`.

PR #4811, opened 2022-07-22
fix(build): Remove duplicate `rayon` dependency from zebra-chain

teor2345:
## Motivation

PR #4805 got merged without any build checks, causing a duplicate `rayon` dependency in `zebra-chain/Cargo.toml` on the `main` branch.
This breaks all the builds.

## Solution

Remove the duplicate dependency.

## Review

This is a critical build fix.

### Reviewer Checklist

  - [ ] CI passes, except for the lightwalletd sync test


oxarbitrage (approved):

teor2345:
I admin-merged this today, so that the full sync is finished by the time I start work tomorrow.

It looks like our recent fixes make Zebra work sometimes during the lightwalletd sync, but it still fails sometimes.
So this is progress!

PR #4813, opened 2022-07-24
1. add(ci): Add a Zebra cached state update test, fix lightwalletd tests

teor2345:
## Motivation

We're seeing failures when syncing `zebrad` to run `lightwalletd`.
But we don't know if those failures are caused by `lightwalletd`, the RPCs, or the rest of `zebrad`.

This test runs `zebrad` the same way as the `lightwalletd` tests, but without actually launching `lightwalletd` or opening an RPC port.

## Solution

Add `zebrad` test:
- Add a `zebrad_update_sync()` test, that update syncs Zebra without `lightwalletd`
- Run the `zebrad_update_sync()` test in CI

Fix `lightwalletd` test:
- Increase `lightwalletd` test timeouts for `zebrad` slowness
- Give `zebrad` extra time to finish after ignoring a `lightwalletd` failure

Clippy:
- Fix `clippy::let_and_return`

## Review

This is an urgent diagnostic for the current CI failures.

We might want to merge this test even if it doesn't pass CI yet, so we can work out when the `zebrad` part of the tests is fixed.

### Reviewer Checklist

  - [ ] Test code makes sense

## Follow Up Work

- #4814

Issue #4747, opened 2022-07-04
Mini-Epic: Stop tokio tasks running for a long time and blocking other tasks


teor2345:
## Motivation

At the moment, Zebra can't sync all the way to the tip, because some tokio tasks run for a long time, and block other tasks.
(It's also possible there are some deadlocks, livelocks, or missed task exits.)

We should discover the specific bugs using `tokio-console`, and then open a ticket for each one.

### Issues that need investigation

- [x] #4583 
- [x] #4823 and then speed them up
- [ ] Add a Future wrapper that times each poll, and logs long polls

### CPU usage analysis

- [x] #4825 & #4826

- [x] #4779

Deserialization (in `zebra-network` or `zebra-state`):
- [x] `sapling::output::OutputPrefixInTransactionV5::zcash_deserialize()` (#4787)
    - `sapling::committment::ValueCommitment::try_from::<[u8; 32]>()`
    - `bellman::groth16::Proof::read()`
    - `jubjub::AffinePoint::from_bytes_inner()`
    - `bls12_381::scalar::square()`
    - `bls12_381::scalar::sqrt()`
- [x] `finalized_state::ZebraDb::block()` (#4788)
    - partially fixed by #4792

Verification (in `zebra-consensus`):
- [ ] `groth16::DescriptionWrapper::try_from()` (#4789)
    - `transaction::Verifier::verify_v5_transaction()`

Note commitment tree updates (in `zebra-state`, either finalized or non-finalized):
- [x] Note commitment tree append and root (#4790)
    - `non_finalized_state::chain::UpdateWith`
    - `sapling::tree::merkle_crh_sapling()`
    - `sapling::commitment::pedersen_hashes::pedersen_hash()`
    - `sapling::commitment::pedersen_hashes::pedersen_hash_to_point()`
    - `incrementalmerkletree::bridgetree::Frontier::append()`
- [x] #4721 should also help with non-finalized blocks

## Fixed Issues

<details>

- [x] #4581

Fixed by #4750:
- [x] #4738
- [x] #4740
- [x] #4729
- [x] Replace batch verifier broadcast channels with watch channels (cleanup only)

Fixed by #4752 and #4726:
- [x] #4650 
- [x] Lagged inventory advertisements (might also need #4750)
- [x] Verification failure after block 1719629 on mainnet (also block 1944652 on testnet is very slow)

</details>

teor2345:
We've made enough performance improvements for now.

PR #4815, opened 2022-07-25
2. add(log): Log when state requests take a long time

teor2345:
## Motivation

To work out why Zebra is slow, we need to know which state queries are slow.

Diagnostics for #4747.

Depends-On: #4813 

## Solution

- Time state init and each state request, log when it takes too long
  - Add a CodeTimer diagnostic struct for long-running code

Related Fixes:
- Instrument state futures and functions with tracing spans

## Review

I think we'll want this PR in `main` soon, so we can fix the remaining state bugs.

### Reviewer Checklist

  - [ ] Existing tests pass

### DIFF
--- a/.github/workflows/continous-integration-docker.patch.yml
+++ b/.github/workflows/continous-integration-docker.patch.yml
@@ -63,6 +63,12 @@ jobs:
     steps:
       - run: 'echo "No build required"'
 
+  test-update-sync:
+    name: Zebra tip update / Run update-to-tip test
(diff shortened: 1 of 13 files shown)


oxarbitrage (approved):
I think this looks really good. Unsure about the numbers used but we will have to try that in a real sync to really know. I am very keen to try this out.

teor2345:
I admin-merged this PR, because it contains urgent diagnostics to finish off this release.

PR #4825, opened 2022-07-26
1. fix(state): Stop reading redundant blocks for every FindHashes and FindHeaders request

teor2345:
## Motivation

Zebra reads 160-500 redundant blocks for each FindHashes or FindHeaders request.
The FindHeaders requests regularly take 5-10 seconds to answer.

This is a security issue because remote nodes can send these requests.

## Solution

- Find a range of heights, rather than a list of hashes
- Stop reading redundant blocks and calculating redundant hashes
- Remove redundant state queries
- Allow some inconsistencies that can happen during concurrent writes, but don't do concurrent queries yet
- Simplify some of the code

This PR contains a regression - the header queries used to be partially concurrent, now they block the rest of the state.
The next PR will implement full concurrency.

## Review

This is an urgent PR to fix Zebra sync performance.

### Reviewer Checklist

  - [ ] Tests for Errors

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -18,6 +18,7 @@
 use std::{
     convert,
     future::Future,
+    ops::{RangeBounds, RangeInclusive},
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
looks good to me

teor2345:
Thanks for the review!

All the tests passed except for lightwalletd, which is fixed by PR #4828.
So I'm going to admin-merge this.

PR #4826, opened 2022-07-26
2. fix(state): Make FindHeaders and FindHashes run concurrently with state updates

teor2345:
## Motivation

Zebra reads 160 headers for each FindHeaders request.
The FindHeaders requests used to take 5-10 seconds to answer, they might still be a bit slow.

So we want to run them concurrently with state updates.

Depends-On: #4825

## Solution

- Make FindHeaders and FindHashes run concurrently with state updates
    - Use the new methods in existing state methods
- Make some state queries more efficient

## Review

This is an urgent PR to fix Zebra sync performance.

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #4828, opened 2022-07-26
fix(ci): fix hangs in lightwalletd tests by checking concurrent process output in different threads

### LINKED ISSUE #4820: Fix `lightwalletd` test hangs by checking process output in dedicated threads
## Motivation

Sometimes `lightwalletd` hangs, but we make the Zebra integration test pass anyway.

When this happens, we update the `lightwalletd` cached state with a small number of blocks, so future tests are slow, or they fail because they don't have enough blocks.

### Alternative Fixes

Fix `lightwalletd` hangs by giving each child process a dedicated test thread to check logs.

_*OR*_

When `lightwalletd` hangs, restart it.

_*OR*_

When `lightwalletd` hangs, don't update the cached `lightwalletd` state

_*OR*_

In the Rust test harnesses:
- don't check that the `lightwalletd` cached state contains a minimum number of blocks
- increase the test timeouts to allow a full `lightwalletd` sync

teor2345:
## Motivation

lightwalletd is unstable in our CI, because `zebrad` hangs when its log buffer fills up.

Close #4820 
Depends-On: #4826

## Solution

- Clear the `zebrad` and `lightwalletd` logs in different threads, to avoid deadlocks due to the log buffer filling up
- Increase test timeouts

This increases the sync test time, but we can fix that in later PRs.

Related fixes:
- Only copy the state once in the send transactions test, to avoid a "disk full" error
- Wait longer for lightwalletd gRPC server startup, to avoid a "connection refused" error
- Look for cached state disks for this commit and branch, then the `main` branch, then any branch
- Do ZK parameter preloads in the lightwalletd tests that need them
- Make code execution time logs shorter

## Review

This is a high-priority sync fix.

### Reviewer Checklist

  - [ ] CI passes

## Follow Up Tasks

- #4834

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -1456,7 +1456,7 @@ fn lightwalletd_integration_test(test_type: LightwalletdTestType) -> Result<()> 
         });
 
         // Retrieve the child process handles from the threads
-        let (zebrad_thread, mut lightwalletd) = zebrad_thread_and_lightwalletd
+        let (zebrad_thread, lightwalletd) = zebrad_thread_and_lightwalletd
             .join()
             .unwrap_or_else(|panic_object| panic::resume_unwind(panic_object))?;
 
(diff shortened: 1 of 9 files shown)


teor2345:
I ran 77 of these tests locally overnight, and they all succeeded.
(CI could be different, but it's still a good sign the restart seems to be working reliably.)

I just need to fix a bug in the GitHub actions where we're using an old cached state.

teor2345:
This should all work now, we're just waiting for CI.

zebrad/tests/common/launch.rs
```diff
@@ -44,26 +44,27 @@ pub const BETWEEN_NODES_DELAY: Duration = Duration::from_secs(2);
...
 /// TODO: reduce to 20 minutes when `zebrad` sync performance improves
-pub const LIGHTWALLETD_UPDATE_TIP_DELAY: Duration = LIGHTWALLETD_FULL_SYNC_TIP_DELAY;
+pub const LIGHTWALLETD_UPDATE_TIP_DELAY: Duration = Duration::from_secs(60 * 60);
 
 /// The amount of time we wait for lightwalletd to do a full sync to the tip.
 ///
-/// `lightwalletd` takes about an hour to fully sync,
-/// and Zebra needs time to activate its mempool.
+/// See [`LIGHTWALLETD_UPDATE_TIP_DELAY`] for details.
 pub const LIGHTWALLETD_FULL_SYNC_TIP_DELAY: Duration = Duration::from_secs(90 * 60);
```
teor2345:
```suggestion
pub const LIGHTWALLETD_FULL_SYNC_TIP_DELAY: Duration = Duration::from_secs(150 * 60);
```

teor2345:
I just increased the full sync timeout.

It looks like the `getblock` RPC is a bit slower than it used to be, I opened ticket #4831 to follow this up in the next sprint.

oxarbitrage (dismissed):

teor2345:
I added a commit to clear the `zebrad` logs regularly during tests.
We might be getting deadlocks due to the log buffer filling up.

I also opened ticket #4834 for a permanent fix.

teor2345:
I'm running a full sync test for this branch here:
https://github.com/ZcashFoundation/zebra/actions/runs/2743858563

Acceptable times are 12-16 hours.

teor2345:
I needed to re-do the fix, because the problem was actually deadlocks from not reading the `zebrad` logs.

zebrad/tests/acceptance.rs
```diff
@@ -1432,60 +1432,64 @@ fn lightwalletd_integration_test(test_type: LightwalletdTestType) -> Result<()>
...
-        //
-        // TODO: re-enable this code when lightwalletd hangs are fixed
-        if cfg!(lightwalletd_hang_fix) {
-            if let Some(ref mut lightwalletd) = lightwalletd {
-                lightwalletd.expect_stdout_line_matches(regex::escape(
-                    "Block hash changed, clearing mempool clients",
-                ))?;
-                lightwalletd
-                    .expect_stdout_line_matches(regex::escape("Adding new mempool txid"))?;
-            }
```
teor2345, line 1486: I deleted these checks, because we can't guarantee there will be any transactions or chain forks on `mainnet` during the test.

teor2345:
There was a google cloud SSH failure:
```
Timeout, server 35.222.36.92 not responding.

Recommendation: To check for possible causes of SSH connectivity issues and get
recommendations, rerun the ssh command with the --troubleshoot option. Example:

gcloud None compute ssh example-instance --zone=us-central1-a --troubleshoot

Or, to investigate an IAP tunneling issue:

gcloud None compute ssh example-instance --zone=us-central1-a --tunnel-through-iap --troubleshoot

ERROR: (gcloud.compute.ssh) [/usr/bin/ssh] exited with return code [255].
```
https://github.com/ZcashFoundation/zebra/runs/7553151823?check_suite_focus=true#step:6:332

I've never seen one of these before, so I restarted those jobs.

oxarbitrage (approved):

teor2345:
Admin-merging because the one failing test is going to be fixed soon.

There's just one issue with a regex in the send transactions test, I'll fix it in PR #4840.

PR #4838, opened 2022-07-27
fix(ci): split canopy sync into a separate GitHub actions job

teor2345:
## Motivation

With the new checkpoints, the checkpoint sync job takes close to 6 hours.

Also, when the cached state disks on the main branch are broken, they cause failures in PRs, even if the disks from the PR are working.

## Solution

- Split the checkpoint job into canopy and NU5 checkpoints
- Look for cached state disks for this commit and branch, then the main branch, then any branch

I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2749604277

## Review

This is an urgent fix for this release, maybe @gustavovalverde can review it?

### Reviewer Checklist

  - [ ] Most of CI passes
  - [ ] Full sync jobs work as expected (they might not pass, that depends on other fixes)

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -256,7 +256,7 @@ jobs:
       # - To ${{ inputs.zebra_state_dir || inputs.disk_prefix }} if not
       #
       # If there are multiple disks:
-      # - prefer images generated from the `main` branch, then any other branch
+      # - prefer images generated from this branch, then the `main` branch, then any other branch
       # - prefer newer images to older images
       #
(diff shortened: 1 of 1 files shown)


conradoplg (approved):
Looks good

teor2345:
The `lightwalletd` failure should be fixed by PR #4828.

teor2345:
The checkpoint sync passed, I'm just waiting for the full sync before I admin-merge.

teor2345:
We need to do another job split, I'll do that in PR #4840.

Admin-merging because we have plans to fix the failing job.

PR #4840, opened 2022-07-28
fix(ci): split NU5 sync into two GitHub actions jobs

teor2345:
## Motivation

We need to temporarily split the NU5 sync job, until we improve sync speeds.

There's also a bug in the send transaction test "lightwalletd sync is finished" regex.

## Solution

- split the NU5 sync at block 1,740,000, which is currently about 5 hours into the NU5 sync
- to see if it has finished syncing, check when lightwalletd is near the tip height

## Review

This is hopefully the last urgent fix, it should pass without admin-merging.

### Reviewer Checklist

  - [ ] CI passes
  - [ ] full sync passes

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -683,7 +683,7 @@ jobs:
           ${{ inputs.test_id }} | \
           tee --output-error=exit /dev/stderr | \
           grep --max-count=1 --extended-regexp --color=always \
-          '(estimated progress.*174[0-9][0-9][0-9][0-9])|(test result:.*finished in)' \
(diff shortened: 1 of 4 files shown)


teor2345:
I started a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/2752398109

gustavovalverde (approved):

teor2345:
I made a new full sync here with some fixes:
https://github.com/ZcashFoundation/zebra/actions/runs/2756819466

teor2345:
> I started a full sync here: https://github.com/ZcashFoundation/zebra/actions/runs/2752398109

This sync passed, but one job was a bit close to 6 hours, and there was a bug in the split regex.
So I fixed the regex and re-ran the full sync.

We can merge this PR if it succeeds, we don't need to wait for the new full sync.

dconnolly (approved):

PR #4848, opened 2022-07-31
fix(test): truncate the number of transactions in send transaction test

### LINKED ISSUE #4847: Limit the number of transactions sent in the lightwalletd send transaction test
## Motivation

We need to limit the number of transactions in the send transaction test, so it is less than the queue limit:

> Message: assertion failed: `(left == right)` 
> left: `SendResponse { error_code: 0, error_message: "transaction dropped because the queue is full" }`, 
> right: `SendResponse { error_code: 0, error_message: "\"c900ca01bee9a4b1e4868a9e1b208d7b8cbc470436e3d5b164b4f22b5a0c5e9d\"" }` 
> Location: zebrad/tests/common/lightwalletd/send_transaction_test.rs:123

https://github.com/ZcashFoundation/zebra/runs/7587005032?check_suite_focus=true#step:6:193

oxarbitrage:
## Motivation

We are seeing some full queue errors in some runs of the send transaction test. Close https://github.com/ZcashFoundation/zebra/issues/4847

## Solution

Truncate the transactions sent in this test. I think this should work but i need to test in CI.

## Review

Anyone can review.

### Reviewer Checklist

  - [ ] `send_transaction_test` pass
  - [ ] All the other tests pass

### DIFF
--- a/zebra-rpc/src/queue.rs
+++ b/zebra-rpc/src/queue.rs
@@ -39,7 +39,7 @@ mod tests;
 const NUMBER_OF_BLOCKS_TO_EXPIRE: i64 = 5;
 
 /// Size of the queue and channel.
-const CHANNEL_AND_QUEUE_CAPACITY: usize = 20;
+pub const CHANNEL_AND_QUEUE_CAPACITY: usize = 20;
 
 /// The height to use in spacing calculation if we don't have a chain tip.
 const NO_CHAIN_TIP_HEIGHT: Height = Height(1);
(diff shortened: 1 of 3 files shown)


teor2345 (approved):
The error message:
> transaction dropped because the queue is full

comes from the RPC queue, not the mempool queue.

So I think we should limit it based on both queues.
I pushed a commit cfe3ebf96700b9d50475738f35b4edc73cd7b2f1.

PR #4849, opened 2022-07-31
fix(test): Skip some RPC tests when ZEBRA_SKIP_NETWORK_TESTS is set

teor2345:
## Motivation

We forgot to check ZEBRA_SKIP_NETWORK_TESTS in some RPC tests.

## Review

If we're developing without the network over the next few weeks, we might want this fix.

### Reviewer Checklist

  - [ ] Tests for Errors


oxarbitrage (approved):
looks good.

Issue #4562, opened 2022-06-02
Deploy long-running Zebra nodes after each tagged release


conradoplg:
## Motivation

We want to have long-running Zebra nodes to:

- Use for testing (check if it keeps syncing correctly)
- Make it more likely to be found by seeders and block explorers

### Designs

- [x] Add a CI workflow that deploys a Zebra node after each tag is published
  - [ ] Also manually if needed
  - [x] Each deploy overwrites the previous one?
  - [ ] Use a static IP?
  - [ ] Also deploy to testnet?

Issue #4546, opened 2022-05-31
Fix automatic Zebra deployments to Google Cloud


conradoplg:
## Motivation

All node deployments seem to be failing in GCP with messages like:

```
2022/05/31 22:15:00 Error: Failed to start container: Error response from daemon: {"message":"manifest for us-docker.pkg.dev/zealous-zebra/zebra/main:bfa319d not found: manifest unknown: Failed to fetch \"bfa319d\""}
```

### Designs

- [ ] Fix the issue
- [ ] Think about a way to get notified somehow if a deployment fails and implement it if possible, or create a new ticket for it if it's complex

gustavovalverde:
We'll be notifying this instead of "fixing" it, as we have to change the way we deploy, and Google Instance Groups are not flexible enough to accomplish this. I'll be tracking this as a general implementation for infrastructure.

Issue #1904, opened 2021-03-15
Some DNS seeders and block explorers don't show Zebra nodes


teor2345:
## Motivation

In #1844, we deployed a few `zcashd` instances on testnet, to make Zebra's testnet tests more reliable. But those nodes aren't showing up in the ECC's or the Foundation's DNS seeders.

As far as we can tell, the nodes are configured correctly, they are synced to tip, their listener ports are open, and they are advertising their external IPv4 addresses. But they don't show up in the DNS of either seeder implementation.

Also, Zebra nodes don't show up in some other DNS seeders or blockchain explorers.

### Bugs to fix

- [x] #4875

### Things to investigate

- [ ] Try pointing a seeder at a long-lived Zebra instance https://github.com/ZcashFoundation/zebra/pull/4870#issuecomment-1204593184

- [ ] `zcashd` seems to ignore peers from `zebrad`: https://github.com/zcash/zcash/issues/3881#issuecomment-1140146874

Other Protocol Mismatches:
- [ ] `version.addr_from` is typically dummy bytes, but Zebra sends the actual details: https://en.bitcoin.it/wiki/Protocol_documentation#version

### List of Block Explorers

Some block explorers show up to 250 nodes.

There are about 150-200 nodes on mainnet that Zebra instances will connect to.
(Recent Zebra versions are limited to 75 outbound and 50 inbound connections, but this is configurable.)

Zebra might connect to fewer nodes because it has strict latency limits.

#### Zebra Nodes

This block explorer shows about 65 nodes, but it includes Zebra:
* https://explorer.zecmate.com/network

#### No Zebra Nodes

This block explorer shows about 250 nodes:
* https://blockchair.com/zcash/nodes (click "Load More" to see all user agents)

These block explorers show connected nodes only:
* https://zcashblockexplorer.com/nodes

Currently down:
* https://explorer.testnet.z.cash/

### Completed

Make Zebra have long-term connections to peers:
- #3120

DNS seeder bugs:
- [x] work out why some DNS seeders return peers with outdated network protocol versions
  - this is a bug in the Foundation's DNS seeder: https://github.com/ZcashFoundation/dnsseeder/issues/10

Metrics:
- [x] #2871

### Optional

We could investigate why DNS seeders ignore `zebrad` nodes:

<details>

Infrastructure:
- [ ] try running `zcashd` on a different network, to see if Google is blocking cryptocurrency traffic
[...]

PR #4870, opened 2022-08-02
feat(net): return peer metadata from `connect_isolated` functions

### LINKED ISSUE #4678: Return peer metadata from `connect_isolated` functions
## Motivation

We want to support development of an alternate DNS seeder based on `zebra-network`.

## Requirements

There's some peer metadata which we need access to (from the `connect_isolated`-style functions) in order to reproduce the [seeder filtering algorithms](https://github.com/zcash/zcash/issues/3881#issuecomment-1145555036) and I'm unsure how to make that happen:

* The value of the `PeerServices` bitset
* The numeric protocol version number (the stuff that looks like `170100`)
* The textual user-agent string (the stuff that looks like `/MagicBean:5.0.0/`)

I see there's a `negotiate_version` in `handshake.rs` that seems to do the version handshaking and it returns a tuple `(remote_version, remote_services, remote_canonical_addr)` (which is almost all we want, it's missing the text user-agent string).

The place where `negotiate_version` is called is `impl<S, PeerTransport, C> Service<HandshakeRequest<PeerTransport>> for Handshake<S, C>`; but I'm not sure how the latter gets called exactly, and it doesn't return/pass through the remote version/services info.

_Originally posted by @superbaud in https://github.com/zcash/zcash/issues/3881#issuecomment-1162096689_

teor2345:
## Motivation

We want to support development of an alternate DNS seeder based on `zebra-network`.

Depends-On: #4869

Closes #4678

### Designs

See the detailed design in ticket #4678.

## Solution

Interface changes:
- Return a Client instance from connect_isolated_* functions
- Export the Client type from zebra-network
- Add a copy of the Version message to the ConnectionInfo struct
- Export the types required to use the new APIs

Refactors & improvements:
- Move LoadTrackedCient version into a ConnectionInfo struct
- Add negotiated version to ConnectionInfo
- Add the peer address to ConnectionInfo, add ConnectionInfo to Connection
- Add peer services and user agent to ConnectionInfo

Related fixes:
- Add and improve debug formatting

## Review

Zebra developers should feel free to review this, but we need to check if it works for the zebra-network users first.

### Reviewer Checklist

  - [ ] Code implements Design
  - [ ] Tests for Errors

## Follow Up Tasks

- Add a copy of the `ConnectionInfo` to `MetaAddr`
- Write tests that make sure remote peer metadata gets turned into the correct `ConnectionInfo` fields
    - peer set connections
    - isolated connections

### DIFF
--- a/zebra-network/src/lib.rs
+++ b/zebra-network/src/lib.rs
@@ -168,10 +168,14 @@ pub use crate::{
     config::Config,
     isolated::{connect_isolated, connect_isolated_tcp_direct},
     meta_addr::PeerAddrState,
-    peer::{Client, HandshakeError, PeerError, SharedPeerError},
+    peer::{Client, ConnectedAddr, ConnectionInfo, HandshakeError, PeerError, SharedPeerError},
     peer_set::init,
     policies::RetryLimit,
-    protocol::internal::{InventoryResponse, Request, Response},
+    protocol::{
+        external::Version,
+        internal::{InventoryResponse, Request, Response},
+        types::PeerServices,
+    },
 };
 
 /// Types used in the definition of [`Request`] and [`Response`] messages.
--- a/zebra-network/src/lib.rs
+++ b/zebra-network/src/lib.rs
@@ -168,7 +168,7 @@ pub use crate::{
     config::Config,
     isolated::{connect_isolated, connect_isolated_tcp_direct},
     meta_addr::PeerAddrState,
-    peer::{HandshakeError, PeerError, SharedPeerError},
+    peer::{Client, HandshakeError, PeerError, SharedPeerError},
     peer_set::init,
     policies::RetryLimit,
     protocol::internal::{InventoryResponse, Request, Response},
--- a/zebra-network/src/lib.rs
+++ b/zebra-network/src/lib.rs
@@ -177,7 +177,7 @@ pub use crate::{
     },
 };
 
-/// Types used in the definition of [`Request`], [`Response`], and [`Message::Version`] messages.
+/// Types used in the definition of [`Request`], [`Response`], and [`VersionMessage`].
 pub mod types {
     pub use crate::{
         meta_addr::MetaAddr,
--- a/zebra-network/src/lib.rs
+++ b/zebra-network/src/lib.rs
@@ -172,7 +172,7 @@ pub use crate::{
     peer_set::init,
     policies::RetryLimit,
     protocol::{
-        external::Version,
+        external::{Version, VersionMessage},
         internal::{InventoryResponse, Request, Response},
         types::PeerServices,
     },
--- a/zebra-network/src/lib.rs
+++ b/zebra-network/src/lib.rs
@@ -174,13 +174,18 @@ pub use crate::{
     protocol::{
         external::{Version, VersionMessage},
         internal::{InventoryResponse, Request, Response},
-        types::PeerServices,
     },
 };
 
-/// Types used in the definition of [`Request`] and [`Response`] messages.
+/// Types used in the definition of [`Request`], [`Response`], and [`Message::Version`] messages.
 pub mod types {
-    pub use crate::{meta_addr::MetaAddr, protocol::types::PeerServices};
+    pub use crate::{
+        meta_addr::MetaAddr,
+        protocol::{
+            external::{AddrInVersion, Nonce},
+            types::PeerServices,
+        },
+    };
 
     #[cfg(any(test, feature = "proptest-impl"))]
     pub use crate::protocol::external::InventoryHash;
--- a/zebra-network/src/peer/handshake.rs
+++ b/zebra-network/src/peer/handshake.rs
@@ -127,7 +127,7 @@ pub struct ConnectionInfo {
     /// which will appear as the connected address to the OS and Zebra.
     pub connected_addr: ConnectedAddr,
 
-    /// The network protocol [`Message::Version`](crate::Message) sent by the remote peer.
+    /// The network protocol [`VersionMessage`](crate::VersionMessage) sent by the remote peer.
     pub remote: VersionMessage,
 
     /// The network protocol version negotiated with the remote peer.
@@ -566,7 +566,7 @@ where
 /// We split `Handshake` into its components before calling this function,
 /// to avoid infectious `Sync` bounds on the returned future.
 ///
-/// Returns the [`Message::Version`](crate::Message) sent by the remote peer.
+/// Returns the [`VersionMessage`](crate::VersionMessage) sent by the remote peer.
 #[allow(clippy::too_many_arguments)]
 pub async fn negotiate_version<PeerTransport>(
     peer_conn: &mut Framed<PeerTransport, Codec>,
--- a/zebra-network/src/peer/handshake.rs
+++ b/zebra-network/src/peer/handshake.rs
@@ -98,6 +98,13 @@ where
     }
 }
 
+/// The metadata for a peer connection.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ConnectionInfo {
+    /// The network protocol version reported by the remote peer.
+    pub version: Version,
+}
+
 /// The peer address that we are handshaking with.
 ///
 /// Typically, we can rely on outbound addresses, but inbound addresses don't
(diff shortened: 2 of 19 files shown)


softminus:
(I'm writing a seeder in Rust that uses the `zebra_network` crate)

My bad for not realizing we'd need this as well, but we definitely need access to `start_height` (the current C++ seeder logic uses that as one of the signals to classify peers) as well and the `relay` field too (the current seeder doesn't use it but it might be useful) from the `Version` message: https://doc-internal.zebra.zfnd.org/zebra_network/protocol/external/message/enum.Message.html#variant.Version

I'm wondering if it'd be possible to just include a deserialized copy of the `Version` message we received from the other end?

But besides that it seems OK, I'm going to try seeing if I can aim my `Cargo.toml` at your forked branch and get my in-progress seeder code working with it!

softminus:
Also I think we need to make all the types for the fields of `connection_info` accessible from the outside, when I do `use zebra_network::protocol::external::types::Version;` i get:
```
sasha_work@seasonal-dream zcash-cotyledon % cargo run
   Compiling zcash-cotyledon v0.1.0 (/Users/sasha_work/remote_vm_repos/zcash-cotyledon)
error[E0603]: module `protocol` is private
   --> src/main.rs:20:20
    |
20  | use zebra_network::protocol::external::types::Version;
    |                    ^^^^^^^^ private module
    |
note: the module `protocol` is defined here
   --> /Users/sasha_work/repos/zebra/zebra-network/src/lib.rs:153:1
    |
153 | mod protocol;
    | ^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0603`.
error: could not compile `zcash-cotyledon` due to previous error
sasha_work@seasonal-dream zcash-cotyledon % 
```

and likewise for `connected_addr` and `peer_services`: 
```
sasha_work@seasonal-dream zcash-cotyledon % cargo run
   Compiling zcash-cotyledon v0.1.0 (/Users/sasha_work/remote_vm_repos/zcash-cotyledon)
error[E0603]: module `peer` is private
   --> src/main.rs:10:21
    |
10  | use  zebra_network::peer::handshake;
    |                     ^^^^ private module
    |
note: the module `peer` is defined here
   --> /Users/sasha_work/repos/zebra/zebra-network/src/lib.rs:150:1
    |
150 | mod peer;
    | ^^^^^^^^^

error[E0603]: module `protocol` is private
   --> src/main.rs:11:20
    |
11  | use zebra_network::protocol::external::types;
    |                    ^^^^^^^^ private module
    |
note: the module `protocol` is defined here
   --> /Users/sasha_work/repos/zebra/zebra-network/src/lib.rs:153:1
    |
153 | mod protocol;
    | ^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0603`.
error: could not compile `zcash-cotyledon` due to 2 previous errors
sasha_work@seasonal-dream zcash-cotyledon % 
```

teor2345:
> My bad for not realizing we'd need this as well, but we definitely need access to `start_height` (the current C++ seeder logic uses that as one of the signals to classify peers)

Oh, that's probably the cause of Zebra bug #1904 (which is unrelated to what you're doing, but easy for us to fix).

> as well and the `relay` field too (the current seeder doesn't use it but it might be useful) from the `Version` message: https://doc-internal.zebra.zfnd.org/zebra_network/protocol/external/message/enum.Message.html#variant.Version
> 
> I'm wondering if it'd be possible to just include a deserialized copy of the `Version` message we received from the other end?

Yep, that's probably fine, we'll wrap it in an `Arc` to save memory in the address book. (Otherwise we'll have about 5 copies of each Version message.)

> But besides that it seems OK, I'm going to try seeing if I can aim my `Cargo.toml` at your forked branch and get my in-progress seeder code working with it!

Great!

teor2345:
> Also I think we need to make all the types for the fields of `connection_info` accessible from the outside

Some of those types might be exposed at a higher module level. But if not, that's an easy fix, feel free to submit a PR, or I'll get to it later this week.

softminus:
> > My bad for not realizing we'd need this as well, but we definitely need access to `start_height` (the current C++ seeder logic uses that as one of the signals to classify peers)
> 
> Oh, that's probably the cause of Zebra bug #1904 (which is unrelated to what you're doing, but easy for us to fix).

I think that that code is commented out in the version that's actually running:
```
static inline int GetRequireHeight(const bool testnet = fTestNet)
{
    // return testnet ? 500000 : 350000;
    return 0;
}
```
and I'm not quite sure why, but I think that makes it not the cause.

If it'd help, I definitely could try aiming the extant C++ seeder at a zebra node and seeing why it's not showing up. Do you have an IP address / DNS name of a long-running zebra node I could use for that?

teor2345:
> If it'd help, I definitely could try aiming the extant C++ seeder at a zebra node and seeing why it's not showing up. Do you have an IP address / DNS name of a long-running zebra node I could use for that?

That would be great!

We can ask around for one, all the ones I know about are transient IPs, nodes, or firewalled. Or @gustavovalverde will have a long-term seeder IP once we fix deployments in tickets #4546 and #4562.

I'll leave it to you both to work that out - we can open a Zebra ticket for the testing if it would help.

teor2345:
@mergifyio update

teor2345:
> > as well and the `relay` field too (the current seeder doesn't use it but it might be useful) from the `Version` message: https://doc-internal.zebra.zfnd.org/zebra_network/protocol/external/message/enum.Message.html#variant.Version
> > I'm wondering if it'd be possible to just include a deserialized copy of the `Version` message we received from the other end?
> 
> Yep, that's probably fine, we'll wrap it in an `Arc` to save memory in the address book. (Otherwise we'll have about 5 copies of each Version message.)

I split out a `VersionMessage` struct, and added it to the `Client` struct.

I also fixed up the `zebra_network` exports, and added a test to make sure that any future exports happen.

teor2345:
@superbaud just wondering what you thought of these new changes?

superbaud:
> @superbaud just wondering what you thought of these new changes?

I'm gonna take a look at them today and tomorrow, sorry for the delay!

softminus:
> @superbaud just wondering what you thought of these new changes?

I'm gonna take a look at them today and tomorrow, sorry for the delay and thanks for working on this!

teor2345:
Oops, I made a mistake with the rebase, here's the correct one.

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
Let's merge this new code to avoid code conflicts. If we need to make more changes to support DNS seeders, we can do them later.

teor2345:
@mergifyio update

zebra-network/src/peer/handshake.rs
```diff
@@ -98,6 +117,26 @@ where
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ConnectionInfo {
+    /// The connected peer address, if known.
+    /// This address might not be valid for outbound connections.
+    ///
+    /// Peers can be connected via a transient inbound or proxy address,
+    /// which will appear as the connected address to the OS and Zebra.
+    pub connected_addr: ConnectedAddr,
+
+    /// The network protocol [`Message::Version`](crate::Message) sent by the remote peer.
```
teor2345: Fix this doc link

zebra-network/src/lib.rs
```diff
@@ -168,15 +168,24 @@ pub use crate::{
...
     policies::RetryLimit,
-    protocol::internal::{InventoryResponse, Request, Response},
+    protocol::{
+        external::{Version, VersionMessage},
+        internal::{InventoryResponse, Request, Response},
+    },
 };
 
-/// Types used in the definition of [`Request`] and [`Response`] messages.
+/// Types used in the definition of [`Request`], [`Response`], and [`Message::Version`] messages.
```
teor2345: Fix link

zebra-network/src/peer/handshake.rs
teor2345: Fixed in df9b50b.

zebra-network/src/lib.rs
teor2345: Fixed in df9b50b.

oxarbitrage (approved):

PR #4721, opened 2022-07-25
feat(state): Send treestate from non-finalized state to finalized state

upbqdn:
## Motivation

When Zebra commits a finalized block, it re-calculates all the note commitment and history trees. This uses a lot of extra CPU. Zebra also fetches these structures from the database for each block coming from the non-finalized state.

The non-finalized state computes the treestate for each block, and it also stores it in RAM. It could pass the treestate over to the finalized state.

## Solution

I created a new structure `FinalizedWithTrees` that wraps the
treestate and the finalized block. I did that because the original
`FinalizedBlock` is `Eq`, but `HistoryTree` can't be `Eq`.

This PR makes Zebra faster because:
1. The finalized state doesn't retrieve the treestate from the disk if
the non-finalized state supplies it.

2. The finalized state doesn't recompute the treestate if the
non-finalized state supplies it.

Close #4824.

PR #4904, opened 2022-08-09
fix(network): send height to peers

### LINKED ISSUE #4875: Send the state tip height in Version messages
## Motivation

Some DNS seeders use the tip height to classify peers, but Zebra always sends a zero height.

For details, see: https://github.com/ZcashFoundation/zebra/pull/4870#issuecomment-1203241525
and
https://github.com/ZcashFoundation/zebra/pull/4870#issuecomment-1204593184

## Design

We should have a `ChainTip` somewhere in the handshaker, we can use it to get the height. (It might be wrapped inside `MinimumPeerVersion`.)

## Related Issues

This might not fix #1904, because some seeders have the height check commented out.

oxarbitrage:
## Motivation

Zebra currently send height 0 in the `start_height` field of the `Message::Version`. This works protocol wise but DNS seeders expect a height here to classify peers. Resolves https://github.com/ZcashFoundation/zebra/issues/4875

## Solution

Pretty simple solution, we already have everything available to end the height when we are building the version message.

## Review

Anyone can review, i could be missing something as i thought this were going to be more complex. Local tests pass and i know this works while checkpointing, however i need to check a full sync. I expect the CI will assist me there.

### Reviewer Checklist

  - [ ] CI pass including full sync tests


teor2345 (commented):
Looks great!

It seems like syncing is slow enough to fail CI, but I don't think that's because of the changes in this PR. But we'll need to fix the sync bugs before we merge any more PRs.

I'll try re-running those jobs, in case it was temporary network slowness.

teor2345 (approved):
We can merge this as soon as we fix the unrelated sync failures.

teor2345:
You could try:
- increasing the timeouts on the failing tests
- updating the Zebra mainnet checkpoints (until #4721 gets merged, the full verifier will do twice as much note commitment work as the checkpoint verifier)

oxarbitrage:
@mergifyio update

teor2345:
It seems like Zebra appears in block explorers and nodes without this fix, so I'm reducing the priority to make room for other PRs in the queue.

oxarbitrage:
@mergifyio update

gustavovalverde:
@Mergifyio update

PR #4927, opened 2022-08-22
Fix clippy::derive_partial_eq_without_eq in generated gRPC test code

### LINKED ISSUE #4921: Fix clippy warnings in generated prost code
## Motivation

CI fails with clippy warnings in generated gRPC code.

## Background

> There is an additional problem in the CI related to clippy and the last version of it https://github.com/ZcashFoundation/zebra/pull/4918/checks?check_run_id=7885954690
> 
> It seems the problem is in the `prost` crate. Here someone tries to fix it but didn't worked out: [tokio-rs/prost#704](https://github.com/tokio-rs/prost/pull/704) I think they ended up removing clippy at [tokio-rs/prost#702](https://github.com/tokio-rs/prost/pull/702) however is hard to tell as they dont explain anything on their pull requests.

They gave up on making clippy pass in their CI, which pushes the work onto downstream users.

But we still want clippy to pass in our CI, so I'll open a PR to fix it tomorrow.
(This is a different bug to CI timeouts, so it should get a different PR.)

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/4910#issuecomment-1222064386_

teor2345:
## Motivation

`tonic`'s generated code causes `clippy::derive_partial_eq_without_eq` warnings, but Zebra's CI requires our test builds to have no clippy warnings.

The `tonic` developers removed clippy from their CI, so we have to fix it ourselves. (The standard approach is to silence clippy warnings in generated code.)

Close #4921.

## Solution

Since the lightwalletd gRPC types can all be compared for total equality, we can tell `tonic` to derive `Eq` for all those types.

## Review

This is critical priority because it is stopping other PRs merging.

### Reviewer Checklist

  - [x] Clippy CI passes

### DIFF
--- a/zebrad/build.rs
+++ b/zebrad/build.rs
@@ -60,6 +60,10 @@ fn main() {
     tonic_build::configure()
         .build_client(true)
         .build_server(false)
+        // The lightwalletd gRPC types don't use floats or complex collections,
+        // so we can derive `Eq` as well as the default generated `PartialEq` derive.
+        // This fixes `clippy::derive_partial_eq_without_eq` warnings.
+        .type_attribute(".", "#[derive(Eq)]")
         .compile(
             &["tests/common/lightwalletd/proto/service.proto"],
             &["tests/common/lightwalletd/proto"],


arya2 (approved):
**Test beta on ubuntu-latest** hit an `IoError(Custom { kind: Other, error: "sapling-spend.params failed` but the same step on **Test stable on ubuntu-latest** succeeded and a minor rustc update should be irrelevant to this change.

Looks good to me if there's no chance lightwalletd gRPC types might use floats or complex collections in the future.

teor2345:
> **Test beta on ubuntu-latest** hit an `IoError(Custom { kind: Other, error: "sapling-spend.params failed` but the same step on **Test stable on ubuntu-latest** succeeded and a minor rustc update should be irrelevant to this change.

Alfredo and I were just talking about that, it's weird that it only happens on beta Rust. It might be a performance issue.

I'll open a ticket with more logging, and turning on performance optimisations for the download crates.

> Looks good to me if there's no chance lightwalletd gRPC types might use floats or complex collections in the future.

If we try to add something incompatible with `Eq` to our gRPC types, our tests will fail on that PR. Then we can go chat to the `lightwalletd` developers about why it was added.

teor2345:
I admin-merged this PR because it's low risk, and multiple CI failures are blocking Mergify.

PR #4938, opened 2022-08-23
Update supported Rust versions in README

teor2345:
## Motivation

We just discovered that an older Rust version definitely isn't supported.

## Review

This is a routine documentation update.

### Reviewer Checklist

  - [ ] Docs make sense


oxarbitrage (approved):
looks good.

Issue #2564, opened 2021-08-03
Include git metadata in docker logs and Google Cloud panic reports


teor2345:
## Motivation

Currently,
- Zebra's Docker logs don't show the git commit
- Zebra's Google Cloud and Docker panic reports only show the git commit, passed through an environmental variable.

But we'd like to see the branch name and other detailed git information via vergen.

## Tasks

The vergen git metadata is likely being disabled due to git security rules.

The fix is here:
https://github.com/rustyhorde/vergen#notes-about-the-optional-git2-015-dependency

If that doesn't work, we should also check:
* there is a `git` binary installed in Docker,
* there is a `.git` directory in Google cloud builds,
* the path to that directory is where vergen or the `zebrad` build script expect it to be.

We could try changing to the `zebra` source directory before building.

### API Reference

https://docs.rs/vergen/5.1.14/vergen/index.html

teor2345:
PR #4939 removes the git commit from some Google Cloud objects, so this ticket needs to be done soon.

teor2345:
This is useful, and we should do it eventually, but it hasn't actually been a problem yet.

teor2345:
We're doing fine diagnosing bugs without this

teor2345:
Vergen just added documentation about how to fix this issue in Docker to its readme.

PR #4939, opened 2022-08-23
ci(build): deploy long running node instances on release

### LINKED ISSUE #4562: Deploy long-running Zebra nodes after each tagged release

gustavovalverde:
## Previous behavior:
Each time we merged to main new nodes would be deployed, this is an
expected behavior as we need to ensure nodes get deployed and run
without issues, but this could also replace nodes very hastily.

## Expected behavior:
We want instances which would run for a longer time, to allow us to
troubleshoot issues or inspect the behavior of this instances for longer
periods of time (2+ weeks)

## Applied solution:
Deploy a versioned manage instance group (MiG) using the major version
of the release semver. We just use the first part of the version to
replace old instances, and change it when a major version is released
to keep a segregation between new and old versions.

Fixes #4562

## Review

This is still a draft until a versioned release is tested

### Reviewer Checklist

  - [ ] Deployment works
  - [ ] Deployment is versioned
  - [ ] A major version is used from semver as the identifier (v1, v2, etc)

## Follow Up Work

None identified so far


.github/workflows/build-docker-image.yml
```diff
@@ -67,12 +74,13 @@ jobs:
...
             type=schedule
+            type=sha
             type=ref,event=branch
             type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
-            type=sha
+            # disabled if major zero
+            type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}
```
teor2345: I'm not sure about skipping major version zero.

Zebra is currently at major version 1, so this condition will never apply?
Also this isn't listed in the PR description.

teor2345 (commented):
Thanks for doing this, I think we've all been looking forward to it!

I'm not sure if I understand this PR, would you mind explaining it to me?

1. I'd like to know the old and new behaviour:
* when we merge to main
* when we tag a release
* on mainnet and testnet

2. Ticket #4562 is for nodes to be deployed every time we tag a release, is that part of this PR?

3. I'm a bit confused about this comment:
> This is still a draft until a versioned release is tested

If we keep this PR in draft, it will never merge, so we'll never test what happens when a release is tagged.

4. Can you add or update a document explaining Zebra's continuous deployment as part of this PR? It only needs to be short, 3-5 sentences should be enough for now.

.github/workflows/build-docker-image.yml
gustavovalverde: This is practically just skipping major versions if the tags are, for example `v0.1.0` which we might use for testing or other purposes. It's basically skipping every tag starting with `v0.`

gustavovalverde:
> 1. I'd like to know the old and new behaviour:
> 
> * when we merge to main
> * when we tag a release
> * on mainnet and testnet

- The actual behavior is not changed. We're deploying instances when we're merging/pushing to `main` and each new commit replaces the previous version.
- When we tag a release, the same happens but using a different reference as a name, including the actual major version being released `v1`, which would result in an instance named `zebrad-v0-9fb8742`
- We're not deploying to Testnet in the whole workflow, that should be another ticket as builds are neither considering Testnet as an option.



> * Ticket #4562 is for nodes to be deployed every time we tag a release, is that part of this PR?
> *
Yes, the main logic is considered here:
https://github.com/ZcashFoundation/zebra/blob/7a385146b27757914a05a2fff929ca59f5b66aca/.github/workflows/continous-delivery.yml#L17-L19

https://github.com/ZcashFoundation/zebra/blob/7a385146b27757914a05a2fff929ca59f5b66aca/.github/workflows/continous-delivery.yml#L38-L53



> 3\. I'm a bit confused about this comment:
> 
> > This is still a draft until a versioned release is tested

Yeah, I have to create a test release so I'm able to test this, I'll be doing this after adding the final comments to the PR. I just need to confirm with you If I can create a new release to test this...something like `v1.0.0-test.1`, with it's related tag. We can later delete the release...but there's no other way to test this, sadly.

> 4\. Can you add or update a document explaining Zebra's continuous deployment as part of this PR? It only needs to be short, 3-5 sentences should be enough for now.

Sure

.github/workflows/build-docker-image.yml
teor2345: This doesn't match Rust conventions, it's common for Rust crates to stay at `v0.*` for years.

I think it would be very surprising to have this skipped, particularly if it is undocumented.

.github/workflows/continous-delivery.yml
```diff
@@ -23,6 +26,32 @@ env:
   MACHINE_TYPE: c2-standard-4
 
 jobs:
+  # If a release was made we want to substract the first part of the semver from the
```
teor2345:
```suggestion
  # If a release was made we want to extract the first part of the semver from the
```

.github/workflows/continous-delivery.yml
```diff
@@ -23,6 +26,32 @@ env:
...
+  # - If our semver is v1.3.0 the resulting output from this job would be v1
+  # 
+  # Note: We just use the first part of the version to replace old instances, and change
+  # it when a major version is released, to keep a segregation between new and old 
+  # versions.
+  versioning:
+    name: Versioning
+    runs-on: ubuntu-latest
+    outputs:
+      api_version: ${{ steps.set.outputs.api_version }}
```
teor2345: On semver.org, this is called the "major" version, and that's the convention for Rust projects.

teor2345:
Thanks for the explanations, I think I understand better now.

> * We're not deploying to Testnet in the whole workflow, that should be another ticket as builds are neither considering Testnet as an option.

Can you open a ticket for testnet deployments?
It can be a low priority for now, but we will probably want it before NU6 testnet activation.

> > I'm a bit confused about this comment:
> > > This is still a draft until a versioned release is tested
> 
> Yeah, I have to create a test release so I'm able to test this, I'll be doing this after adding the final comments to the PR. I just need to confirm with you If I can create a new release to test this...something like `v1.0.0-test.1`, with it's related tag. We can later delete the release...but there's no other way to test this, sadly.

That's ok, but we need to say that the release is unofficial.
So please tag it something like `v1.0.0-unofficial-test.1`, and mark it as a pre-release on GitHub.

Also please let the team know, so they don't get confused!

teor2345:
How many nodes are we deploying at the same time?

There's a cost vs test coverage tradeoff here.

But in general, since we're running around 0-3 test nodes that are syncing blocks from the network, it is good practice to run a similar number of nodes that are serving blocks to the network. So 1 per-commit node and 1 per-release node should be enough.

.github/workflows/build-docker-image.yml
gustavovalverde: Fixed by [ci(build): allow v0 as a major version tag](https://github.com/ZcashFoundation/zebra/pull/4939/commits/6c3e27877f5754fdcb8096126bdb2e3e6fd833e8)

.github/workflows/continous-delivery.yml
gustavovalverde: Fixed by [fix(build): use rust conventions for versioning](https://github.com/ZcashFoundation/zebra/pull/4939/commits/98c9b87d96071bf12307b41f3a1a6fdf36804a75)

teor2345 (commented):
Thanks, looks good, hope the testing goes well!

teor2345 (changes requested):
Thanks, this looks good, just waiting on a short document that explains our continuous deployment.

Since this PR creates some Google Cloud objects without git commit info, I've added ticket #2564 to the next sprint.

gustavovalverde:
> 4\. Can you add or update a document explaining Zebra's continuous deployment as part of this PR? It only needs to be short, 3-5 sentences should be enough for now.

My bad @teor2345 I read this wrongly and understood it as adding documentation to the file. Please recommend me a place or existing document where you see fit to copy this information.

teor2345:
> > 4. Can you add or update a document explaining Zebra's continuous deployment as part of this PR? It only needs to be short, 3-5 sentences should be enough for now.
> 
> My bad @teor2345 I read this wrongly and understood it as adding documentation to the file. Please recommend me a place or existing document where you see fit to copy this information.

No worries!

We can add developer documentation to:
https://github.com/ZcashFoundation/zebra/tree/main/book/src/dev

We also need to document how other parts of CI work, but we can add those documents in a future PR that changes how CI works.

teor2345 (approved):
Thanks, let's see how this goes!

teor2345:
@mergifyio update

PR #4942, opened 2022-08-24
fix(test): Show full Zebra test panic details in CI logs

teor2345:
## Motivation

We're seeing some Zebra panics in CI. Our custom test harness just reports "The application panicked (crashed).", with no details.

Example logs:
https://github.com/ZcashFoundation/zebra/runs/7981879420?check_suite_focus=true#step:6:1103

Depends-On: #4945

## Solution

- Add output logs to test context, and add test harness tests to check it works
- Let empty test child logs be read again (and produce empty output

Related refactors:
- Add an `ignore_exited` argument to test child process `kill()`
- Handle test failure regexes using Result::Err, rather than panicking

Related changes:
- Log the command timeout when an acceptance test fails

## Review

This is an urgent fix, so we can see panic details in our CI logs.

### Reviewer Checklist

  - [x] CI passes

### DIFF
--- a/zebra-test/src/command.rs
+++ b/zebra-test/src/command.rs
@@ -523,7 +523,7 @@ impl<T> TestChild<T> {
 
             if wrote_lines {
                 // Write an empty line, to make output more readable
-                self.write_to_test_logs("");
+                Self::write_to_test_logs("", self.bypass_test_capture);
             }
         }
 
@@ -533,7 +533,7 @@ impl<T> TestChild<T> {
             while self.wait_for_stderr_line(None) {}
 
             if wrote_lines {
-                self.write_to_test_logs("");
+                Self::write_to_test_logs("", self.bypass_test_capture);
             }
         }
 
@@ -554,11 +554,17 @@ impl<T> TestChild<T> {
         self.apply_failure_regexes_to_outputs();
 
         if let Some(line) = self.stdout.as_mut().and_then(|iter| iter.next()) {
+            let bypass_test_capture = self.bypass_test_capture;
+
             if let Some(write_context) = write_context.into() {
-                self.write_to_test_logs(write_context);
+                Self::write_to_test_logs(write_context, bypass_test_capture);
             }
 
-            self.write_to_test_logs(line.expect("failure reading test process logs"));
+            Self::write_to_test_logs(
+                line.context_from(self)
+                    .expect("failure reading test process logs"),
+                bypass_test_capture,
+            );
 
             return true;
         }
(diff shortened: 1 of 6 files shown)


zebra-test/src/command.rs
```diff
@@ -526,12 +553,12 @@ impl<T> TestChild<T> {
     {
         self.apply_failure_regexes_to_outputs();
 
-        if let Some(Ok(line)) = self.stdout.as_mut().and_then(|iter| iter.next()) {
+        if let Some(line) = self.stdout.as_mut().and_then(|iter| iter.next()) {
```
arya2: small nit: `line_result` or `line_res` might be more descriptive than `line` here and on line 602

teor2345:
Now we know what the panic bug is, this isn't a critical priority any more.

arya2 (dismissed):

zebra-test/src/command.rs
```diff
@@ -771,16 +834,15 @@ impl<T> TestChild<T> {
 
         if self.is_running() {
             // If the process exits between is_running and kill, we will see
-            // spurious errors here. If that happens, ignore "no such process"
+            // spurious errors here. So we want to ignore "no such process"
             // errors from kill.
-            self.kill()?;
+            self.kill(true).context_from(self)?;
```
teor2345: Revert this - we're getting the context higher up

zebrad/tests/common/sync.rs
```diff
@@ -248,7 +248,7 @@ pub fn sync_until(
 
         // make sure the child process is dead
         // if it has already exited, ignore that error
-        let _ = child.kill();
+        let _ = child.kill(true);
```
teor2345: Stop ignoring other kill errors here.

zebra-test/src/command.rs
```diff
@@ -3,7 +3,7 @@
 use std::{
```
teor2345, line 3: Rebase on PR #4945 before pushing

zebra-test/src/command.rs
teor2345: Fixed in commit 6c0ec540d7

zebra-test/src/command.rs
teor2345: Fixed in 9074205b29

zebrad/tests/common/sync.rs
teor2345: Fixed in 92afab915e

zebra-test/src/command.rs
teor2345: Done!

teor2345:
Marking this as critical because we're still seeing panics, and we need to know what they are. So we want to merge this first.

teor2345:
@mergifyio update

teor2345:
I added some extra test debug output in commit 9d4c709d71. Alfredo and I talked about it earlier this week to make it easier to diagnose test failures.

I was going to put it in PR #4958, but that might have conflicted with this PR, so I put it here instead.

teor2345:
@mergifyio update

gustavovalverde (approved):
Approving as this is critical and CI is passing.

PR #4945, opened 2022-08-24
fix(ci): Increase CI disk size to 200GB

teor2345:
## Motivation

Some tests are failing because they run out of disk space.

We can limit the size of the database by changing how it is configured in Zebra, but right now we just need to make CI pass.

### Logs

> Unread Stderr: 
> ...
>  Message: expected that errors would not occur when writing to disk or updating note commitment and history trees: Error { message: "IO error: No space left on device: While appending to file: /var/cache/zebrad-cache/state/v25/mainnet/017386.log: No space left on device" } 
> Location: zebra-state/src/service.rs:301

https://github.com/ZcashFoundation/zebra/runs/8002360538?check_suite_focus=true#step:6:1216

## Solution

- Increase the size of cached state disks to 200GB
- Increase the size of boot disks to 200GB (the send transaction test copies the state to `/tmp`)

## Review

This is an urgent fix to CI, anyone can review it.

### Reviewer Checklist

  - [ ] CI passes

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -130,9 +130,9 @@ jobs:
         id: create-instance
         run: |
           gcloud compute instances create-with-container "${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}" \
-          --boot-disk-size 100GB \
+          --boot-disk-size 200GB \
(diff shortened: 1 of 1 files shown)


conradoplg (approved):

gustavovalverde (approved):
LGTM

teor2345:
The 200GB disk size only applies to newly created disks, so we need to:
- admin-merge this PR
- wait for a full sync to run on `main`

Then other PRs will start passing.

PR #4948, opened 2022-08-25
change(doc): Increase disk and network requirements for long-term deployment

teor2345:
## Motivation

We increased our CI disk sizes in PR #4945, so we should also increase our recommended disk and network capacity.

## Solution

- recommend 300 GB disk to allow for long-term growth
- recommend 100+ GB network uploads and downloads per month, to allow for initial sync and long-term growth

## Review

Anyone can review this PR, it's not urgent.

### Reviewer Checklist

  - [ ] Docs make sense


gustavovalverde (approved):
LGTM

PR #4949, opened 2022-08-25
fix(ci): Write cached state images after update syncs, and use the latest image from any commit

### LINKED ISSUE #4842: Replace cached state from the same commit, rather than failing the job
## Motivation

If a cached state job is run twice on the same commit on the same branch, the job fails.

If other required  jobs depend on that cached state job, they are skipped, and the PR won't merge.

> The resource 'projects/zealous-zebra/global/images/lwd-cache-4765-merge-59532a1-v25-mainnet-tip' already exists
> Error: Process completed with exit code 1.

https://github.com/ZcashFoundation/zebra/runs/7588987700?check_suite_focus=true#step:8:72

### Suggested Solution

Replace the image, rather than failing the job.

If that's not possible, it's also ok to skip creating the image. But the job still has to run successfully. That's not ideal, because it leaves an older (and possibly buggy) cached state image behind.

teor2345:
## Motivation

Zebra's CI is slow, but we can speed it up by:
- writing images after update syncs
- using the latest image from any commit on any branch

We haven't seen any problems from sharing images between branches, so test speed is a higher priority.

Also close #4842

## Solution

- Save cached state on full syncs and updates
- Add an -update suffix to CI images created by updating cached state
- Make disk image names unique by adding a time suffix
- Use the latest image from any branch, but prefer the current commit
- Document Zebra's continuous integration tests

## Review

This is an urgent CI speed fix, but it might depend on a successful full sync creating new images on at least one branch.

### Reviewer Checklist

  - [ ] CI passes and is faster after the first run

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -330,11 +330,10 @@ jobs:
       test_description: Test syncing to tip with a Zebra tip state
       test_variables: '-e TEST_UPDATE_SYNC=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache'
       needs_zebra_state: true
-      # TODO: do we want to update the disk on every PR, to increase CI speed?
-      saves_to_disk: false
+      # update the disk on every PR, to increase CI speed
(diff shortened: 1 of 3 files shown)


gustavovalverde (approved):
LGTM

PR #4950, opened 2022-08-25
fix(ci): Fix a regular expression typo in a full sync job

teor2345:
## Motivation

There's a mistaken `||` in a regular expression in a full sync job, which causes the job to end really quickly, so the next job times out.

This only impacts full syncs, the update syncs are short enough that the timeout doesn't happen.

## Review

Anyone can review this PR, it's needed to fully apply the fix in #4945 to the cached state disks.
But it doesn't block other PRs merging.

### Reviewer Checklist

  - [ ] Fix makes sense

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -716,7 +716,7 @@ jobs:
           ${{ inputs.test_id }} | \
           tee --output-error=exit /dev/stderr | \
           grep --max-count=1 --extended-regexp --color=always \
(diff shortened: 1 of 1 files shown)


arya2 (approved):
So it was skipping alternatives following the ||, makes sense, nice find

teor2345:
It was my mistake originally!

(For people who need a reminder, `|` separates two alternative regular expressions, so `A||B` means "`A` or match everything or `B`", because an empty regular expression matches everything.)

teor2345:
I admin-merged this PR because it was failing with an "out of disk space" error.
That error is fixed on the `main` branch, but we need this fix to make the `main` branch CI run correctly.

PR #4952, opened 2022-08-25
Full sync with logging for very slow state block commits

### LINKED ISSUE #4932: Reduce the frequency of Zebra performance logs
## Motivation

As part of #4747, we added some info-level performance logs to Zebra.

But performance is good enough now, so we can reduce the frequency of those logs.

## Tasks

We could downgrade all the logs to debug level, but it might still be useful to log them for very long tasks.

For example:
- the state can take up to a minute to commit a block, so we want to log that at trace or debug level
- rebuilding non-finalized note commitment trees can take 1-2 minutes, so we could log that at info level
- blocks timeout after 10 minutes, so we might want to log a warning before 10 minutes

The performance logging is implemented in:
https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/diagnostic.rs

### LINKED ISSUE #4937: Tracking: Commit blocks to state using a separate task
## Motivation

Zebra takes 10-15 minutes to commit some blocks to the state while checkpointing, around blocks 1,718,00 to 1,772,000.

The slow blocks are different on different runs.

This is unacceptable performance, because:
* it's much slower than `zcashd`
* Zebra will appear to hang for 15 minutes, which is a usability and security issue
* it causes warnings in the Zebra logs
* if it's remotely triggerable, it could be a denial of service risk

## Diagnosis

Zebra queues up to 1200 blocks, then commits them all in the same state request, after the missing block arrives. This can take up to 10 seconds per block.

### Design

Add a block commit task to the state, which runs in a separate thread. The task should be between the block queue and the block verifier.

We'll need to move the shared mutable chain state into the block commit task, so we will also need to redirect `StateService` read requests to the concurrent `ReadStateService`.

Here is a diagram of the new state design:
https://docs.google.com/drawings/d/1FXpAUlenDAjl8nkftrypdAPsj0jr-Ut9gZlSP57nuyc/edit

#### Stop Accessing Mutable Chain State

- [x] #5102

#### Setup Block Commit task

- [x] Add a new block commit task with unused channels

#### Add channels to send blocks to the task

- [x] Add a channel that handles finalized state `CommitFinalizedBlock` requests
- [x] Add a channel that handles non-finalized state `CommitBlock` requests
- [x] We want two channels so we can wait for the last finalized block before committing the first non-finalized block (by height)
    - The current implementation of this has a bug: #5125

#### Error Handling & Testing

- [x] Handle panics in the block commit task by panicking in the service
- [x] Testing - what new tests do we want
    - [x] #5315
    - [x] #2654

Optional tasks:

<details>

#### Optional Cleanup Tasks

Bug fixes:
- [x] #5126

Refactors:
- Make pending_utxos.respond() async using a channel, so we can use `ReadRequest::ChainUtxo` in `AwaitUtxo`

Renames & Formatting:
- Rename every instance of address `*` or `transparent_*` to `address_*`
- Put the `Request` and `Response` enums in a consistent order

</details>

## In Scope

- Non-finalized state
- Finalized state
- Running the task in a separate thread

## Out of Scope

We don't think we'll need to make these changes as part of this change:

<details>

- [x] #5101
- [x] #5103
  (this reduces the number of state requests from the syncer)

</details>

These are definitely out of scope:
- Other state refactors
- Other performance improvements
- Note commitment tree performance improvements

teor2345:
## Motivation

We want to find out which parts of state block commits are slow in CI, because we can't reproduce them locally.

These are diagnostics for ticket #4937, where some blocks take 15 minutes to write to the state.

## Solution

- change info threshold to 5 minutes and warn to 9 minutes
- add slow code logging for block commits, block writes, and note commitment tree updates

## Review

This PR doesn't need review.

### Reviewer Checklist

  - [ ] Logging works

## Follow Up Work

Read the logs to work out why #4937 is happening

### DIFF
--- a/zebra-chain/src/diagnostic.rs
+++ b/zebra-chain/src/diagnostic.rs
@@ -66,6 +66,11 @@ impl CodeTimer {
         self.finish_inner(Some(module_path), Some(line), description);
     }
 
+    /// Ignore this timer: it will not check the elapsed time or log any warnings.
+    pub fn ignore(mut self) {
+        self.has_finished = true;
+    }
+
     /// Finish timing the execution of a function, method, or other code region.
     ///
     /// This private method can be called from [`CodeTimer::finish()`] or `drop()`.
--- a/zebra-chain/src/diagnostic.rs
+++ b/zebra-chain/src/diagnostic.rs
@@ -7,10 +7,14 @@ use std::time::{Duration, Instant};
 use crate::fmt::duration_short;
 
 /// The default minimum info-level message time.
-pub const DEFAULT_MIN_INFO_TIME: Duration = Duration::from_secs(5);
+///
+/// This is high enough to ignore most slow code.
+pub const DEFAULT_MIN_INFO_TIME: Duration = Duration::from_secs(5 * 60);
 
 /// The default minimum warning message time.
-pub const DEFAULT_MIN_WARN_TIME: Duration = Duration::from_secs(20);
+///
+/// This is a little lower than the block verify timeout.
+pub const DEFAULT_MIN_WARN_TIME: Duration = Duration::from_secs(9 * 60);
 
 /// A guard that logs code execution time when dropped.
 #[derive(Debug)]
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -6,6 +6,7 @@ use thiserror::Error;
 
 use crate::{
     block::{Block, Height},
+    diagnostic::CodeTimer,
     orchard, sapling, sprout,
 };
 
@@ -49,7 +50,8 @@ impl NoteCommitmentTrees {
         &mut self,
         block: &Arc<Block>,
     ) -> Result<(), NoteCommitmentTreeError> {
-        self.update_trees_parallel_list(
+        let timer = CodeTimer::start();
+        let res = self.update_trees_parallel_list(
             [(
                 block
                     .coinbase_height()
@@ -58,7 +60,9 @@ impl NoteCommitmentTrees {
             )]
             .into_iter()
             .collect(),
-        )
+        );
+        timer.finish(module_path!(), line!(), "update_trees_parallel");
+        res
     }
 
     /// Updates the note commitment trees using the transactions in `block`,
@@ -70,6 +74,7 @@ impl NoteCommitmentTrees {
         &mut self,
         block_list: BTreeMap<Height, Arc<Block>>,
     ) -> Result<(), NoteCommitmentTreeError> {
+        let timer = CodeTimer::start();
         // Prepare arguments for parallel threads
         let NoteCommitmentTrees {
             sprout,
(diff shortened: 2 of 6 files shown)


zebra-state/src/service/tests.rs
```diff
@@ -314,12 +314,20 @@ proptest! {
...
+            .expect("chain contains valid blocks");
+
         let response = crate::service::check::legacy_chain(nu_activation_height, chain.into_iter().rev(), network)
             .map_err(|error| error.to_string());
 
         prop_assert_eq!(
             response,
-            Err("giving up after checking too many blocks".into())
+            Err(format!(
+                "could not find any transactions in recent blocks: checked 1000 blocks back from {tip_height:?}",
```
arya2: 1000 could be replaced with {MAX_LEGACY_CHAIN_BLOCKS} if we want this test to succeed when that's changed

arya2 (dismissed):
Looks good to me

zebra-state/src/service/tests.rs
teor2345: I accidentally pushed two PRs to this branch, this is actually from PR #4804. I'll sort it out tomorrow.

teor2345: Fixed in c334ba465 in PR #4804

oxarbitrage (commented):
The code looks good for sure but i dont think we should be merging this things into the main branch.

This seems more suitable for a local test or for a specific branch outside main.

teor2345:
> The code looks good for sure but i dont think we should be merging this things into the main branch.
> 
> This seems more suitable for a local test or for a specific branch outside main.

Unfortunately, I can't reproduce this issue on my local machine, so I need to add this debugging to the `main` branch.
Once we know what the issue is, we can remove it.

oxarbitrage:
Sorry, still dont understand why you need to add it to the main branch. Can't you run the CI without merging ?

teor2345:
> Sorry, still dont understand why you need to add it to the main branch. Can't you run the CI without merging ?

I'm not sure if it will work, but I can try.

I'll open a new PR to disable the logs for #4932, because that needs to go into `main`. Then I can make this PR and draft and do a full sync on it.

teor2345:
I'm running a full sync here with these extra logs:
https://github.com/ZcashFoundation/zebra/actions/runs/2953663804

teor2345:
@mergifyio update

teor2345:
Let's try that full sync again, after the concurrency fixes in PR #5000:
https://github.com/ZcashFoundation/zebra/actions/runs/2959955933

teor2345:
The logs I added didn't actually show any extra information, which is a bit of a surprise. It's possible we're just committing a large number of slow blocks at a time.

I added some more logging and started another full sync:
https://github.com/ZcashFoundation/zebra/actions/runs/2968581615

teor2345:
It appears that committing blocks hangs because we sometimes commit over 1000 blocks at the same time, then return the result to the state caller (checkpointer or non-finalized state block committer):
> 2022-09-01T22:50:25.650944Z  WARN state: run time: very long queue_and_commit_finalized: 1774457..=1775638, 1181 blocks time=11m 24s module=zebra_state::service::finalized_state line=173
> 2022-09-01T22:50:25.650981Z  WARN state: run time: very long CommitFinalizedBlock time=11m 24s module=zebra_state::service line=697

https://github.com/ZcashFoundation/zebra/runs/8147047219?check_suite_focus=true#step:6:165

I'll update ticket #4937 with the details.

teor2345:
This PR has achieved its goals, but we'll need the extra logging to test any fixes.

teor2345:
This PR was only used for testing slow block commits, we can re-open it if we need to re-test.

PR #4957, opened 2022-08-26
adds start as default subcommand for zebrad

arya2:
## Motivation

We want to run `zebrad start` by default when users run `zebrad` so it behaves more similarly to zcashd and bitcoind.

## Solution

  - add default_expr on command field of EntryPoint

## Review

Anyone can review this PR

### Reviewer Checklist

  - [ ] runs the start command by default when no subcommand is provided

## Follow Up Work

Additional CLI args, e.g. tracing filters which override the zebrad.toml config, aren't provided.


zebrad/src/application.rs
```diff
@@ -37,6 +43,98 @@ pub fn app_config() -> config::Reader<ZebradApp> {
     config::Reader::new(&APPLICATION)
 }
 
+/// Toplevel entrypoint command.
+///
+/// Handles obtaining toplevel help as well as verbosity settings.
+#[derive(Debug, Options)]
+pub struct EntryPoint {
```
teor2345: Can you put this struct in a submodule of `application`?
Can you link to the `abscissa` code that you copied?

The documentation will help us when we upgrade abscissa to the latest version.
(Linking to copied code is required for licensing reasons. Having a separate module is useful but optional.)

zebrad/tests/acceptance.rs
```diff
@@ -445,21 +445,6 @@ fn ephemeral(cache_dir_config: EphemeralConfig, cache_dir_check: EphemeralCheck)
     Ok(())
 }
 
-#[test]
-fn app_no_args() -> Result<()> {
```
teor2345, line 450: Can you replace this test with a test that makes sure Zebra runs the start command?
New features need a test to make sure they keep working.

The test can be similar to `start_no_args()`, but it might need to be called from `config_test()` to avoid port conflicts.

teor2345 (changes requested):
Looks good, I wouldn't have thought of doing it that way!

I just have a few suggestions to follow our development style.

teor2345:
It looks like the rustfmt check failed, you might want to set up your editor to format files on save. Or you could set up a git commit hook that runs `cargo fmt --all`.

zebrad/tests/acceptance.rs
arya2: Calling it from `config_test()` to avoid port conflicts makes sense to me, `zebra_zcash_listener_conflict()` still failed so I added `start_no_args()` and `start_args()` there too.

zebrad/src/application/entry_point.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Subcommand to execute.
+    ///
+    /// The `command` option will delegate option parsing to the command type,
+    /// starting at the first free argument. Defaults to start.
+    #[options(command, default_expr = "Some(ZebradCmd::Start(StartCmd::default()))")]
+    pub command: Option<ZebradCmd>,
+}
+
+impl EntryPoint {
+    /// Borrow the underlying command type or print usage info and exit
```
arya2:
```suggestion
    /// Borrow the underlying command type
```

zebrad/tests/acceptance.rs
teor2345: > `zebra_zcash_listener_conflict()` still failed so I added `start_no_args()` and `start_args()` there too.

I'm not sure why those failures are happening for these tests, because:

They are configured to run on separate ports:
https://github.com/ZcashFoundation/zebra/blob/87f4308caf0734a0cb163bba40107b85026a0ed1/zebrad/tests/common/config.rs#L31-L32

Running tests on different ports in parallel makes sure that the config is actually being applied. If the config isn't applied, all the Zebra instances run on the default ports, causing a port conflict and test failure.

This is passing on Linux and macOS in CI. So I'm wondering if it is platform-specific.
(Zebra doesn't support Windows or WSL and the moment, and we have minimal support for macOS.)

Can you show me the info-level logs from each test failure?

You might need to do:
```sh
export RUST_LOG=info
```
Before running the tests.

arya2: After re-running the tests with the same code, this was almost certainly a different mistake on my end before committing. `zebra_zcash_listener_conflict()` succeeds consistently and the info logs are what they normally are when we run zebrad.

teor2345 (dismissed):
Thanks!

@upbqdn this is a breaking change, can you mention it at the top of the release notes?

teor2345:
I merged a conflicting dependency update from `main` into this PR.

teor2345 (dismissed):

zebrad/tests/acceptance.rs
```diff
@@ -517,6 +502,37 @@ fn config_test() -> Result<()> {
...
+    let _init_guard = zebra_test::init();
+
+    // start caches state, so run one of the start tests with persistent state
+    let testdir = testdir()?.with_config(&mut persistent_test_config()?)?;
+
+    let mut child = testdir.spawn_child(args![])?;
+
+    // Run the program and kill it after a few seconds
+    std::thread::sleep(LAUNCH_DELAY);
+    child.kill()?;
```
teor2345: Update based on test API changes from another PR

```suggestion
    child.kill(true)?;
```

teor2345 (approved):

teor2345:
This failed with:
>  - The resource 'projects/zealous-zebra/global/images/zebrad-cache-4721-merge-9be8664-v25-mainnet-tip-u-0830011712' is not ready

https://github.com/ZcashFoundation/zebra/runs/8082952279?check_suite_focus=true#step:7:78

I'm going to try again and see if that works.

teor2345:
@mergifyio update

PR #4961, opened 2022-08-27
fix(ci): Increase full sync timeouts for longer syncs

teor2345:
## Motivation

A full Zebra sync takes about 24 hours right now, so we need to increase the Rust test timeout and the number of sync jobs.

## Solution

- Split a GitHub actions job that is timing out
- Increase the Rust test timeout to 24 hours

Related refactors:
- Split alternative regular expressions onto separate lines for readability

## Review

This is urgent because the sync tests won't pass until it merges to `main` and finishes a full sync.
We can only really test this PR by merging it to `main`.

Update sync CI probably won't pass until PR #4962 gets merged.

### Reviewer Checklist

  - [ ] Changes make sense
  - [ ] CI fails with a disk full panic (fixed by PR #4962)

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -733,7 +733,7 @@ jobs:
   # (or the test finishing)
   #
   # We chose this height because it was about 9 hours into the NU5 sync, at the end of August 2022.
-  logs-1757k:
+  logs-1760k:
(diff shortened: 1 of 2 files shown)


gustavovalverde (approved):
LGTM

teor2345:
I admin-merged this PR because it is failing due to bugs fixed in PR #4962 (for old cached states) or bugs fixed on `main` (for newly generated cached states).

PR #4962, opened 2022-08-27
fix(ci): Expand cached state disks before running tests

### LINKED ISSUE #4295: Truncate branch name when used in Google Cloud names
## Motivation

Zebra's DevOps workflows fail on long branch names:
> ERROR: (gcloud.compute.images.create) Could not fetch resource:
> - Invalid value for field 'resource.name': 'zebrad-cache-increase-block-validation-timeouts-0eb6a83-v23-mainnet-tip'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'

https://github.com/ZcashFoundation/zebra/runs/6283701351?check_suite_focus=true#step:10:69

It would be good to fix this eventually, but it's not urgent.

### Designs

- truncate branch names before using them in length-limited contexts (in this case, we can only use 24 characters)

teor2345:
## Motivation

In PR #4945, we increased the cached state disk size to 200GB. But that only applies to newly created cached states. (Which take 24 hours to run, and still need fixes like PR #4961.)

So we want to expand older cached state disks to the latest size before using them.

Also fixes #4295.

## Solution

- Grow the disk partition to the full disk size
- Resize the filesystem to the full partition size
- Actually run the test

Related fixes:
- Keep cached state image names within the 63-character google cloud limit (bugfix on PR #4949)
- Adapt cached state image search for short branch names by ignoring branch names (just use short commit hashes)

## Review

This is urgent because it's stopping all the other PRs from passing.

### Reviewer Checklist

  - [ ] CI passes

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -367,7 +367,7 @@ jobs:
           --ssh-flag="-o ServerAliveInterval=5" \
           --command \
           "\
-          sudo e2fsck -v -f /dev/sdb \
+          sudo e2fsck -v -f -p /dev/sdb \
           && \
           sudo resize2fs -p /dev/sdb \
           && \
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -367,6 +367,11 @@ jobs:
           --ssh-flag="-o ServerAliveInterval=5" \
           --command \
           "\
+          sudo apt-get --quiet update --yes \
+          && \
+          sudo apt-get --quiet install --yes --no-install-recommends \
+          cloud-guest-utils \
+          && \
           sudo growpart --verbose /dev/sdb 1 \
           && \
           sudo resize2fs -p /dev/sdb \
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -367,12 +367,7 @@ jobs:
           --ssh-flag="-o ServerAliveInterval=5" \
           --command \
           "\
-          sudo apt-get --quiet update --yes \
-          && \
-          sudo apt-get --quiet install --yes --no-install-recommends \
-          cloud-guest-utils \
-          && \
-          sudo growpart --verbose /dev/sdb 1 \
+          sudo e2fsck -v -f /dev/sdb \
           && \
           sudo resize2fs -p /dev/sdb \
           && \
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -938,14 +938,22 @@ jobs:
         with:
           short-length: 7
 
-      # Disk images in GCP are required to be in lowercase, but the blockchain network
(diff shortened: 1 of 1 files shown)


gustavovalverde (dismissed):
This was a great finding! Thank you!

gustavovalverde:
This PR is failing as the disk name is too long. I'll be fixing this so we're able to continue with this merge https://github.com/ZcashFoundation/zebra/runs/8050845710?check_suite_focus=true#step:9:75

teor2345:
> This PR is failing as the disk name is too long. I'll be fixing this so we're able to continue with this merge https://github.com/ZcashFoundation/zebra/runs/8050845710?check_suite_focus=true#step:9:75

@gustavovalverde I'm happy to fix it, I have a format that will come in under the 63 character limit, but still be unique.

gustavovalverde:
Oh, I think we were working on the disk name solution in parallel. I have a PR with a similar solution here: https://github.com/ZcashFoundation/zebra/pull/4966

I don't mind if your prefer your solution, but please take into consideration some of the examples I made there as, following the actual name reference in this PR (`${{ inputs.disk_prefix }}-${SHORT_GITHUB_REF}-${{ env.GITHUB_SHA_SHORT }}-v${{ env.STATE_VERSION }}-${{ env.NETWORK }}-${{ inputs.disk_suffix }}${UPDATE_SUFFIX}-${TIME_SUFFIX}`) a disk exceeding 63 characters would be possible, for example:
- `zebrad-cache-4962-merge-e098efa-v25-mainnet-checkpoint-update-2022-08-27-14-04-40` (81 characters)

Also, keeping the commit does not seem necessary as explained in https://github.com/ZcashFoundation/zebra/pull/4966 as the GITHUB_REF should be enough to find the latest available disk from the PR/branch

teor2345:
> Oh, I think we were working on the disk name solution in parallel. I have a PR with a similar solution here: #4966

Sorry about that, but I'm happy to be responsible for fixing this one.

> I don't mind if your prefer your solution, but please take into consideration some of the examples I made there as, following the actual name reference in this PR (`${{ inputs.disk_prefix }}-${SHORT_GITHUB_REF}-${{ env.GITHUB_SHA_SHORT }}-v${{ env.STATE_VERSION }}-${{ env.NETWORK }}-${{ inputs.disk_suffix }}${UPDATE_SUFFIX}-${TIME_SUFFIX}`) a disk exceeding 63 characters would be possible, for example:
> 
> * `zebrad-cache-4962-merge-e098efa-v25-mainnet-checkpoint-update-2022-08-27-14-04-40` (81 characters)

This doesn't actually happen with the latest code in this PR, because:
- the checkpoint is at a fixed height, so it is never updated
- the update suffix is "-u"
- the date doesn't have a year or hyphens

So the actual longest branch name is 63 characters:
- `zebrad-cache-4962-merge123-e098efa-v25-mainnet-tip-u-0827140440`

> Also, keeping the commit does not seem necessary as explained in #4966 as the GITHUB_REF should be enough to find the latest available disk from the PR/branch

Since the branch name is truncated, it's no longer unique, but bug #4842 and accurate search require it to be unique.

gustavovalverde (approved):

PR #4963, opened 2022-08-27
fix(doc): Update disk usage based on recent data

teor2345:
## Motivation

The initial sync size in our docs was too low, it's actually at least 85 GB.

## Solution

- Make initial sync at least 100 GB
- Update some related docs

## Review

This is a routine doc update.

### Reviewer Checklist

  - [ ] Docs make sense


gustavovalverde (approved):
LGTM

Issue #4965, opened 2022-08-27
ci: delete old cached disk images from GCP


gustavovalverde:
## Motivation

Disk images are being accumulated in GCP for a few years, but this generated unneeded costs as we're not using images older than 1-2 weeks. 

### Designs

Even though we could delete images based on a timeframe, we must keep at least the 2 latest images of each type (zebra checkpoint, zebra tip, lwd tip). Once we've excluded those images, we can delete any older images after 1 week. This gives us time to inspect the content of failing images, and time to run a full sync on `main`, while the existing `main` image is still available to be used as a base.

## Related Work

None so far

Issue #4964, opened 2022-08-27
ci: If a cached disk image is manually triggered it won't be used by the opened PR


gustavovalverde:
## Motivation

When triggering a full sync or checkpoint sync from a branch, we might expect that a successful build of the cached disk image would be available for following workflow runs in the same branch (this will mostly be a PR in practice). But this won't happen as the GitHub reference for manual trigger uses the `branch name` and the automatic workflow runs on a PR uses the `PR reference` which is `merge-<pr_id>`

### Designs

When searching for disks in a PR, also search for disks having the branch name, and not just the PR reference

## Related Work

This was found while working with https://github.com/ZcashFoundation/zebra/issues/4295

teor2345:
We don't want to prioritise previous commits from the same PR, because we want to be testing with the output of the current commit.

We've had bugs where the cached state created by previous commits worked, but the current commit was failing. While we can't avoid them entirely due to performance reasons, I don't want to make them more likely.

gustavovalverde:
Just to be clear on this one, if you manually trigger a full sync and/or checkpoint sync from a branch/PR, the disk image will just be created but never used afterwards.

We either allow the branch to use this image, which might have the actual commit if a re-run is triggered, or never create it as it won't be used anyways.

PR #4966, opened 2022-08-27
ci(disks): truncate disk names and improve disk name search in PRs

### LINKED ISSUE #4295: Truncate branch name when used in Google Cloud names

gustavovalverde:
## Previous behavior:
Some tests were not able to create a disk image as the following error was being raised
in GCP: Invalid value for field 'resource.name'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'

## Expected behavior:
Disk image names in GCP have a 61 characters max length constraint, and disks names
are being created with a length between 71 and 84 chars.

## Solution:
This refactors the disk naming and the way we find the latest cached disk image.
Instead of searching the commit SHA of the current "running" commit, the first
search is focused on finding a disk from the actual PR, as matching the actual
commit will rarely be a match unless re-running a job, and even in this scenario
we can search in the PR and the latest disk will be returned from a successful run

Here's a disk name example with the longest name from a PR after this changes are
applied: `zebrad-cache-4962-merge-v25-mainnet-checkpoint-update` (53 chars long).
If a disk is manually triggered using `workflow_dispatch` the branch name must
be <= 18 characters so we truncate the branch name at disk creation and while
searching for an available disk name for the same branch.

Fixes https://github.com/ZcashFoundation/zebra/issues/4295

## Review

This PR should be reviewed by @teor2345 

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

- https://github.com/ZcashFoundation/zebra/issues/4964
- https://github.com/ZcashFoundation/zebra/issues/4965

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -265,8 +265,8 @@ jobs:
       # - To ${{ inputs.zebra_state_dir || inputs.disk_prefix }} if not
       #
       # If there are multiple disks:
-      # - prefer images generated from this branch and commit, then
-      # - if prefer_main_cached_state is true, prefer images from the `main` branch, then
+      # - prefer images generated from this branch, then
+      # - if prefer_main_cached_state is true, prefer images from the `main` branch, else
       # - use images from any other branch.
       # Within each of these categories:
       # - prefer newer images to older images
@@ -285,13 +285,13 @@ jobs:
               DISK_PREFIX=${{ inputs.zebra_state_dir || inputs.disk_prefix }}
           fi
 
-          # Try to find an image generated from this branch and commit
+          # Try to find an image generated from this branch
           # Fields are listed in the "Create image from state disk" step
-          COMMIT_DISK_PREFIX="${DISK_PREFIX}-${GITHUB_REF_SLUG_URL}-${GITHUB_SHA_SHORT}-v${LOCAL_STATE_VERSION}-${NETWORK}-${{ inputs.disk_suffix }}"
-          COMMIT_CACHED_DISK_NAME=$(gcloud compute images list --filter="name~${COMMIT_DISK_PREFIX}" --format="value(NAME)" --sort-by=~creationTimestamp --limit=1)
-          echo "${GITHUB_REF_SLUG_URL}-${GITHUB_SHA_SHORT} Disk: $COMMIT_CACHED_DISK_NAME"
-          if [[ -n "$COMMIT_CACHED_DISK_NAME" ]]; then
-              echo "Description: $(gcloud compute images describe $COMMIT_CACHED_DISK_NAME --format='value(DESCRIPTION)')"
+          PR_DISK_PREFIX="${DISK_PREFIX}-${GITHUB_REF_SLUG_URL:0:18}-v${LOCAL_STATE_VERSION}-${NETWORK}-${{ inputs.disk_suffix }}"
+          PR_CACHED_DISK_NAME=$(gcloud compute images list --filter="name~${PR_DISK_PREFIX}" --format="value(NAME)" --sort-by=~creationTimestamp --limit=1)
+          echo "${GITHUB_REF_SLUG_URL} Disk: $PR_CACHED_DISK_NAME"
+          if [[ -n "$PR_CACHED_DISK_NAME" ]]; then
+              echo "Description: $(gcloud compute images describe $PR_CACHED_DISK_NAME --format='value(DESCRIPTION)')"
           fi
 
           # Try to find an image generated from the main branch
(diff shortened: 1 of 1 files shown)


teor2345:
> This refactors the disk naming and the way we find the latest cached disk image. Instead of searching the commit SHA of the current "running" commit, the first search is focused on finding a disk from the actual PR, as matching the actual commit will rarely be a match unless re-running a job, and even in this scenario we can search in the PR and the latest disk will be returned from a successful run

Unfortunately, the branch names in this PR will cause a regression to bug #4842, because the disk names are not unique. To be unique, the disk names need to contain the commit and timestamp.

We also don't want to prioritise previous commits from the same PR, because we want to be testing with the output of the current commit. (We've had bugs where the cached state created by previous commits worked, but the current commit was failing. While we can't avoid them entirely due to performance reasons, I don't want to make them more likely.)

gustavovalverde:
> We also don't want to prioritise previous commits from the same PR

I don't think we're prioritizing "previous commits" with this approach, we're just prioritizing the latest/actual commit from the same PR/branch, as we're filtering images by creation date.

> Unfortunately, the branch names in this PR will cause a regression to bug https://github.com/ZcashFoundation/zebra/issues/4842

To workaround https://github.com/ZcashFoundation/zebra/issues/4842 we can do as with templates, delete it if exists and create a new one. But as this error just happens when re-running on the **same commit**, the previous cached state and the actual cached state should be the same, we're just making the jobs run longer as we're not considering a potentially healthy image and rebuilding it again (with the same code). As far as I'm concern, we don't expect Zebra to generate different cached states with the exact same codebase.

teor2345:
> > We also don't want to prioritise previous commits from the same PR
> 
> I don't think we're prioritizing "previous commits" with this approach, we're just prioritizing the latest/actual commit from the same PR/branch, as we're filtering images by creation date.

That's not quite how it works, because searching for the branch name will choose either:
1. an updated state from the previous commit on the same branch, or
2. an updated state from the current commit on the same branch

Option 1 has caused bugs to be merged to main before, because the previous commit created a working state, but the current commit (which is what gets merged) didn't.

We can't avoid this situation entirely without sacrificing CI performance, because that would mean always building from a main branch cached state, which is slow.

But if we only prioritise the current commit when searching for a cache state, merging previous buggy states to main should be rare enough that it doesn't matter. That's why I just search for the commit in PR #4962.

Searching for the commit also ignores the different GITHUB_REFs of manually triggered jobs, pushed branches, and PR merge branches. So it helps with that too.

> > Unfortunately, the branch names in this PR will cause a regression to bug #4842
> 
> To workaround #4842 we can do as with templates, delete it if exists and create a new one. 

I don't want to delete the cached state, because that creates a race condition - there are a few minutes where the old cached state has been deleted, but the new image hasn't been created yet. This means any jobs that start during that time will have different behaviour. (They'll definitely be slower, and they might behave differently or fail.)

There's also a complex interaction with #4965 that I don't really want to have to think through, I'd prefer to only delete old cached states in one place.

Instead, I made the image names unique in PR #4962 by adding a timestamp. That way there's no gap with a missing image.

> But as this error just happens when re-running on the **same commit**, the previous cached state and the actual cached state should be the same, we're just making the jobs run longer

Future jobs will actually run shorter, because:
- they'll use the most recent state for this commit, which is likely to be close to the tip
- the next job will use the updated state from this repeated run, which will be even closer to the tip

> as we're not considering a potentially healthy image and rebuilding it again (with the same code). As far as I'm concern, we don't expect Zebra to generate different cached states with the exact same codebase.

Since the chain is always adding new blocks, the same code creates different cached states at different times - the later ones have a higher block height.

But there's shouldn't be any other differences (unless there's a bug in the PR).

gustavovalverde:
Ok, now I'm understanding the edge cases better. Thank you for taking the time for explaining it. I'd like to add this same answer somewhere in a document, maybe the same one you created for continuous integration

teor2345:
> Ok, now I'm understanding the edge cases better. Thank you for taking the time for explaining it. I'd like to add this same answer somewhere in a document, maybe the same one you created for continuous integration

That sounds good, maybe we want to have the details next to the code, so we don't accidentally break things. I'll think about it later this week, and do an update to the workflow files. Once I've done that, I'll have a better idea of how I want to summarise it in the developer docs.

teor2345:
These edge cases are documented in PRs #4972 and #4973.

PR #4970, opened 2022-08-28
ci(deploy): do not run `versioning` job when pushing to `main`

gustavovalverde:
## Previous behavior:
When a push was detected in the `main` branch, the workflow would run the
`versioning` job and crash trying to detect the version being deployed as
there was none.

## Expected behavior:
Do not fail the `versioning` job when pushing to `main`

## Solution:
Limit the `versioning` job to only run when a release event is triggered
and allow the `deploy-nodes` job to run even if `versioning` is skipped

## Review

This is an urgent PR to anyone from the @ZcashFoundation/devops-reviewers to review


teor2345 (approved):
Thanks!

PR #4973, opened 2022-08-28
doc(ci): Document edge cases in sync workflows

teor2345:
## Motivation

We've discovered a bunch of stuff about CI workflows and cached state over the past 2 months, we should write it down next to the workflow code.

## Review

These docs should get merged before we make the next workflow changes.

### Reviewer Checklist

  - [ ] Docs make sense


gustavovalverde (approved):
Thank you for this!

PR #4980, opened 2022-08-29
ci(build): allow to use lightwalletd images from PRs

### LINKED ISSUE #4612: ci: Wait for lightwalletd image rebuild if it has changed
## Motivation

If we change the `lightwalletd` Docker image in a PR, we want to see those changes in the PR's tests. Otherwise we won't know if our fixes work.

If we don't change the image, the latest `main` branch image is fine.

## Design

Before starting this PR, we want to add the job to the branch protection rules, so we don't merge broken code:
- #4882

We might want to use the same dependency algorithm we use for the state cache:
- wait for the image from the current PR and commit, if it is being rebuilt, then use it, or
- use the image from `main`

## Original Conversation

> I'm not sure if this completely solves the problem. What happens if we need to update the image? The PR that changes it will use the `latest` image which will not be the updated one and it will fail CI.

I had the same thought, it makes testing a bit tricky.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/4599#issuecomment-1155790695_

gustavovalverde:
## Previous behavior:
If the lightwalledt Dockerfile was changed in a PR, we couldn't test the
changes from the actual branch as the image being pulled was the one from
the `main` branch.

## Expected behavior:

If we change the lightwalletd Docker image in a PR, those changes in the
PR should be reflected in tests. Otherwise we won't know if our fixes work.

If the image is not changed, the latest `main` branch image is fine.

## Solution:
When building Zebra, allow to build using a lightwalletd image built from
the actual working branch/PR if lightwalletd code has likely changed

Fixes https://github.com/ZcashFoundation/zebra/issues/4612

## Review

Dependencies are not right here, this requires further work. No review is required yet.

### Reviewer Checklist
  - [ ] Tests for Errors

## Follow Up Work

This might need a bigger refactor, as we shouldn't build Zebra before lightwalletd.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -126,6 +126,7 @@ jobs:
             ZEBRA_SKIP_IPV6_TESTS=${{ inputs.zebra_skip_ipv6_tests }}
             CHECKPOINT_SYNC=${{ inputs.checkpoint_sync }}
             RUST_LOG=${{ inputs.rust_log }}
+            LWD_VERSION=${{ env.GITHUB_REF_SLUG_URL }}
           push: true
           # To improve build speeds, for each branch we push an additional image to the registry,
           # to be used as the caching layer, using the `max` caching mode.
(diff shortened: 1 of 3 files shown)


teor2345:
@gustavovalverde we've taken ticket #4612 out of the sprint, because it's not causing any problems right now. 

Did you want to close this PR, and add it to the ticket so we have a draft to start from?

gustavovalverde:
Linked to https://github.com/ZcashFoundation/zebra/issues/4612 until it becomes a priority once again

Issue #4976, opened 2022-08-29
Move continuous deployment to another Google Cloud region


teor2345:
## Motivation

In ticket #4857, we're limiting the number of concurrent workflow runs to avoid CI failures due to Google Cloud CPU quotas.

If we deploy our release and `main` branch instances to another region, that will reduce how much quota we're using in `us-central1`.

### Tasks

- Move managed instance groups to another region, but still in the US for good performance
- Manually delete the old managed instance groups

Optional:
- Change Zebra build instances to another region
- Change the `zcashd` instances to another region
- Consider changing the default region for instances manually created by developers, if possible

teor2345:
This seems ok for now, we haven't seen these errors in a while.

teor2345:
This doesn't seem to be an issue right now.

PR #4981, opened 2022-08-29
ci(concurrency)!: run a single CI workflow as required

### LINKED ISSUE #1785: Deploy mainnet nodes CD error: resource is not ready
## Motivation

There was a single mainnet deployment failure with the error:
> The resource 'projects/zealous-zebra/regions/us-east1/instanceGroupManagers/zebrad-main' is not ready

This error happened when two PRs were merged less than a minute apart.

It seems like a rare situation, so I'm marking this `P-Medium`.

## Solution

We should skip the mainnet deploy job if it is already running:
https://github.com/marketplace/actions/skip-duplicate-actions

I think we'll want a config that cancels the newer jobs, because we can't cancel a deployment that's halfway done:
```yaml
concurrent_skipping: 'always'
cancel_others: false
```

## Logs

```
Run gcloud compute instance-groups managed rolling-action start-update \
ERROR: (gcloud.compute.instance-groups.managed.rolling-action.start-update) Could not fetch resource:
 - The resource 'projects/zealous-zebra/regions/us-east1/instanceGroupManagers/zebrad-main' is not ready
```

### LINKED ISSUE #4857: Limit the number of concurrent Google Cloud jobs
## Motivation

When we're running a lot of GitHub workflows at the same time, we can hit CPU limits or other quotas in Google Cloud.

For example, one of the limits is 500 CPUs.

## Suggested Fix

- Set the concurrency limit on jobs that use Google Cloud

I think the limit calculation is roughly:
500 quota / 16 CPUs per machine - 4 CD instances = 25 job concurrency limit
500 quota / 16 CPUs per machine / 4 jobs launched simultaneously per workflow - 1 CD "workflow" = 6 workflow concurrency limit

But the limits might only be 12 jobs or 3 workflows, if each core counts as 2 quota CPUs.

### LINKED ISSUE #4977: Only run one full sync on main at a time
## Motivation

There isn't much point in running multiple Mainnet full syncs on the `main` branch at the same time.

It slows down the Zcash network, costs the Foundation more money, and it doesn't provide much more diagnostic information.

### Designs

We could skip or cancel full syncs if there is already one running on `main`.
(But still run all the dependent jobs, if possible.)

We don't want to just limit the concurrency of jobs running on `main`, because that could queue up days worth of jobs on `main`.

gustavovalverde:
## Previous behavior:
Multiple Mainnet full syncs were able to run on the main branch at the
same time, and pushing multiple commits to the same branch would run
multiple CI workflows, when only the run from last commit was relevant

## Expected behavior:
Ensure that only a single CI workflow runs at the same time in PRs.
The latest commit should cancel any previous running workflows from the
same PR.

## Solution:
Use GitHub actions concurrency feature https://docs.github.com/en/actions/using-jobs/using-concurrency

Fixes https://github.com/ZcashFoundation/zebra/issues/4977
Fixes https://github.com/ZcashFoundation/zebra/issues/4857
Fixes https://github.com/ZcashFoundation/zebra/issues/1785

## Review

@teor2345 can review this

### Reviewer Checklist

  - [x] Re-run a workflow while running or push an empty commit
  - [ ] Tests for Errors

### DIFF
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -1,7 +1,7 @@
 name: Build crates individually
 
-# Ensures that only a single CI workflow runs at a time in Pull Requests. The latest
-# commit will cancel any previous running workflows from the same PR.
+# Ensures that only a single CI workflow runs at a time in Pull Requests. The
+# latest commit will cancel any previous running workflows from the same PR
 concurrency: 
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -1,8 +1,9 @@
 name: Build crates individually
 
-# Ensures that only a single CI workflow runs at a time in Pull Requests. The
-# latest commit will cancel any previous running workflows from the same PR
-concurrency: 
+# Ensures that only one workflow task will run at a time. Previous builds, if
+# already in process, will get cancelled. Only the latest commit will be allowed
+# to run, cancelling any workflows in between
+concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -1,5 +1,11 @@
 name: Build crates individually
 
+# Ensures that only a single CI workflow runs at a time in Pull Requests. The latest
+# commit will cancel any previous running workflows from the same PR.
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 on:
   workflow_dispatch:
   push:
--- a/.github/workflows/continous-delivery.yml
+++ b/.github/workflows/continous-delivery.yml
@@ -1,8 +1,8 @@
 name: CD
 
-# Ensures that only one deploy task will run at a time. Previous deployments, if
-# already started, won't get cancelled. Following workflow runs will get queued
-# until the previous deployment(s) finishes
+# Ensures that only one workflow task will run at a time. Previous deployments, if
+# already in process, won't get cancelled. Instead, we let the first to complete
+# then queue the latest pending workflow, cancelling any workflows in between
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: false
--- a/.github/workflows/continous-delivery.yml
+++ b/.github/workflows/continous-delivery.yml
@@ -1,5 +1,12 @@
 name: CD
 
+# Ensures that only one deploy task will run at a time. Previous deployments, if
+# already started, won't get cancelled. Following workflow runs will get queued
+# until the previous deployment(s) finishes
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 on:
   workflow_dispatch:
     inputs:
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -1,5 +1,11 @@
 name: CI Docker
 
+# Ensures that only a single CI workflow runs at a time in Pull Requests. The latest
+# commit will cancel any previous running workflows from the same PR.
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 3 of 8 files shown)


.github/workflows/build-crates-individually.yml
```diff
@@ -1,5 +1,11 @@
 name: Build crates individually
 
+# Ensures that only a single CI workflow runs at a time in Pull Requests. The
```
teor2345: This isn't a full sync on `main` or a Google Cloud job, so its concurrency is automatically limited to the correct limit for GitHub actions.

.github/workflows/continous-delivery.yml
```diff
@@ -1,5 +1,12 @@
 name: CD
 
+# Ensures that only one deploy task will run at a time. Previous deployments, if
+# already started, won't get cancelled. Following workflow runs will get queued
+# until the previous deployment(s) finishes
```
teor2345: This config has acceptable behaviour, but these comments don't quite explain how it works. Interim pending runs get still get cancelled, only the first and latest run will complete:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

.github/workflows/continous-integration-os.yml
```diff
@@ -1,5 +1,11 @@
 name: CI OSes
 
+# Ensures that only a single CI workflow runs at a time in Pull Requests. The
```
teor2345: This isn't a full sync on `main` or a Google Cloud job, so its concurrency is automatically limited to the correct limit for GitHub actions.

.github/workflows/continous-integration-docker.yml
```diff
@@ -312,6 +318,12 @@ jobs:
       saves_to_disk: true
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height'
+    # As the general workflow concurrency group just get matched in Pull Requests,
+    # it has no impact on this job, but we want to cancel running full syncs in
+    # `main` if a new PR gets merged
+    concurrency: 
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
```
teor2345: If we're having a busy week, we could keep on merging PRs, and no full syncs will ever complete, because they keep getting cancelled. Instead, it's better to let the first sync finish, then launch the latest sync.

```suggestion
    # We don't want to cancel running full syncs on `main` if a new PR gets merged,
    # because we might never finish a full sync during busy weeks. Instead, we let the
    # first sync complete, then queue the latest pending sync, cancelling any syncs in between.
    # (As the general workflow concurrency group just gets matched in Pull Requests,
    # it has no impact on this job.)
    concurrency: 
      group: ${{ github.workflow }}-${{ github.ref }}
      cancel-in-progress: false
```

teor2345 (changes requested):
Looks good, but I think it might need a few tweaks.

I'm not sure if this PR solves ticket #4857, because it's still possible to launch multiple concurrent Google Cloud jobs from different PRs. But it does fix part of the problem, so I'll mark it as a low priority, and we can see if it's still causing issues.

.github/workflows/build-crates-individually.yml
teor2345: But maybe it would be less confusing (and faster) if it works the same for GitHub actions and Google Cloud jobs.

.github/workflows/continous-integration-os.yml
teor2345: But maybe it would be less confusing (and faster) if it works the same for GitHub actions and Google Cloud jobs.

.github/workflows/continous-integration-docker.yml
teor2345: Oops I forgot to switch cancel-in-progress to false

teor2345 (changes requested):
We might also want to cancel outdated workflows on a PR in the same way, for consistency:
- coverage.yml - important, it's very slow
- lint.yml - some lints can be slow

And limit the following deployments on `main`:
- zcash-params.yml - important, the server has a rate-limit, and it can be very slow
- docs.yml - I think I've seen doc deployment conflicts before (?)

.github/workflows/continous-delivery.yml
gustavovalverde: Fixed by: [fix(concurrency): explain the behavior better & add new ones](https://github.com/ZcashFoundation/zebra/pull/4981/commits/3316cbbac32d75fe809e1363d45730c2f3740982)

gustavovalverde:
> I'm not sure if this PR solves ticket #4857

As a single PR could have several commits running (using GCP instances) at the same time + the `main` branch also having several commits running a full sync, we were having more than expected concurrent jobs in a busy week.

This will considerable reduce the concurrency + https://github.com/ZcashFoundation/zebra/issues/4976 
And we can also increase the quota a 25% more.

teor2345 (approved):
Thank you so much for this!

teor2345:
> Ensure that only a single CI workflow runs at the same time in PRs.
The latest commit should cancel any previous running workflows from the
same PR.

@gustavovalverde can you update the CI and CD docs with a short mention of the concurrency limits & cancellation behaviour?

upbqdn:
@gustavovalverde should I mention in the changelog that this PR is a breaking change? I don't think it's a breaking change for users, that's why I'm not sure about it.

gustavovalverde:
It's just for developers, not for users. It can be kept internal.

PR #4989, opened 2022-08-29
gRPC mempool queries test research

oxarbitrage:
## Motivation

After a lot of days testing i was able to make some little progress researching the issue with the gRPC mempool tests. I don’t have a proper solution yet.

This PR is just to show stuff and it is not for merging, at least not as it is now.

This PR adds 2 test vector transactions, that i got randomly from the zcash mainnet network. The only modification to the raw bytes made was to change the expiration field to 4M. This not only will make the expiration height validation check to pass but also differentiates the transaction from one already in the state.

https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebra-test/src/vectors.rs#L192-L20

None of them will get added to the mempool with the current test code, this is the problem we always had: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs#L346-L385.

- `DUMMY_TX2` is a v5 transaciton with only transaparent data. When using it from the test, the transaction can’t pass from https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebra-consensus/src/transaction.rs#L358-L359

- `DUMMY_TX3` is a transaction with no transparent data and only sapling data. This transaction will fail at the async validation of the sapling shielded data (https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebra-consensus/src/transaction.rs#L396)

This could be fine, transactions could be referring to data not available anymore. However we never get an error in the test side(client side) so we have no idea what went wrong unless we add logging in zebra validation and get the output in the zebra node.

For testing, we can remove some or all the checks from `zebra-consensus` and make the transaction pass validation. However, by removing validation the transaction still never gets inserted anyway. Basically, this is never code is never executed after success validation: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/src/components/mempool.rs#L338 but i saw this being called on validation errors: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/src/components/mempool.rs#L346

For demo purposes this PR will insert a transaction before the validation is fully done.
 
 With this modification `DUMMY_TX2` will still not make it to the mempool but `DUMMY_TX3` will do.
 
 So i think we have 2 issues here:
 
- Build a transaction that can always pass validation no matter at what height we are (but we can assume we are after nu5).
- Even with a transaction that can pass validation, it will not get inserted to the mempool from the test.

I want to highlight the second point because is important. If you have a zebra node running and you send a transaction to it by rpc (by `sendrawtransaction`), the transaction will get inserted to the mempool. The problem is in the test and not in normal zebra functionality. One possibility is that one of the futures can be blocked looking to download the transaction from peers while the tests are running with no peers at all.

So, we have a few options here:

- Try the same test with peers and with a transaction that will pass the validation. I am unsure if this will work.
- Test with peers and wait for a transaction from them gets inserted into our mempool organically. Can never happen.
- Have a debug flag that will force insert the transaction as it is shown in this PR.

The test we are after is made to check if `GetMempoolTx` replies with transactions when they are present in the mempool, how do they get there is not this test purpose so we can “inject” transactions anyway we want.


oxarbitrage:
Ill better add that config flag now as i didn't realized the code change is interfering with other tests.

zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
```diff
@@ -344,5 +343,46 @@ pub async fn run() -> Result<()> {
...
+    // Get the mempool by calling `GetMempoolStream`.
+    let mut transaction_stream = rpc_client.get_mempool_stream(Empty {}).await?.into_inner();
+
+    let mut counter = 0;
+    while let Some(_tx) = transaction_stream.message().await? {
+        counter += 1;
+    }
+
+    // This one is not working.
+    assert_eq!(counter, 0);
```
oxarbitrage, line 385: I am not sure yet if this is a problem in our end or in lightwalletd. I will check after the other call gets sorted but it could be because of https://github.com/adityapk00/lightwalletd/blob/master/frontend/service.go#L515-L517

I tried this call instead of `GetMempoolTx` just to be sure that the stream was not consumed by the first call, however this didn't worked.

teor2345: Good find!

I think it's ok for the test to expect no transactions for now. We can link to the `lightwalletd` code that disables the mempool stream, and fix the test when `lightwalletd` starts streaming transactions again.

teor2345 (commented):
It looks like this PR is nearly finished off, thanks for working out why the test was failing!

We could simplify this PR by doing these new tests during the send transaction test:
https://github.com/ZcashFoundation/zebra/blob/e5c0e108879413c80ce8c4847599087bc182f05e/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131

That way, we won't need the new transaction test vectors. And since the sent transactions are valid, we won't need a config to skip validation.

What do you think?
Did you want to give it a shot, or did you want me to push a PR with that change?

oxarbitrage:
Doing it in the send transaction test is not a good idea for the following reasons, i gave up with that in a previous research months ago:

1- You need to have double of the space of the blockchain available as the test do a full copy of the blockchain.
2- Test takes hours to run locally, so it is a real pain to debug, this test is a lot quicker.
3- There is no guarantee that the transactions inserted in the test will get into the mempool, in fact, in general they dont. The test just check that there are no errors when sending transactions: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131 but make them get inserted into the mempool is another story.

teor2345:
> 3- There is no guarantee that the transactions inserted in the test will get into the mempool, in fact, in general they dont. The test just check that there are no errors when sending transactions: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131 but make them get inserted into the mempool is another story.

That's weird, I expected them to just go straight in. Maybe that's a problem with the send transaction test that we can fix, I'll check it out.

teor2345:
> > 3- There is no guarantee that the transactions inserted in the test will get into the mempool, in fact, in general they dont. The test just check that there are no errors when sending transactions: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131 but make them get inserted into the mempool is another story.
> 
> That's weird, I expected them to just go straight in. Maybe that's a problem with the send transaction test that we can fix, I'll check it out.

There's definitely something wrong with either the send transaction test or with the mempool itself. It seems like I'm getting an `InMempool` error with an empty mempool. I'll have some better diagnostics and hopefully a PR tomorrow.

I'd still like to avoid adding another config option to Zebra, but if we need it for testing, I'm happy to do it.

teor2345:
> 1- You need to have double of the space of the blockchain available as the test do a full copy of the blockchain.
> 2- Test takes hours to run locally, so it is a real pain to debug, this test is a lot quicker.

We could fix both these test issues by getting the transactions from the non-finalized state using JSON-RPC, rather than copying the entire finalized state. I'll open a follow-up ticket if we decide to go that way.

daira:
Perhaps I'm missing something, but a transaction constructed from a valid one by changing one (effecting) field can't be valid, because the signatures will be on the wrong hash.

teor2345:
> Perhaps I'm missing something, but a transaction constructed from a valid one by changing one (effecting) field can't be valid, because the signatures will be on the wrong hash.

Yep, that's why I'm trying to avoid the modified transaction test vectors and adding an option to skip mempool validation.

teor2345:
> > > 3- There is no guarantee that the transactions inserted in the test will get into the mempool, in fact, in general they dont. The test just check that there are no errors when sending transactions: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131 but make them get inserted into the mempool is another story.
> > 
> > 
> > That's weird, I expected them to just go straight in. Maybe that's a problem with the send transaction test that we can fix, I'll check it out.
> 
> There's definitely something wrong with either the send transaction test or with the mempool itself. It seems like I'm getting an `InMempool` error with an empty mempool. I'll have some better diagnostics and hopefully a PR tomorrow.

I think the mempool is actually ok, this is caused by the `skip_full_validation` option added in this PR. The queued transactions are rejected because they are already added to the mempool by `skip_full_validation`.

I suspect that the send transaction test can't query the mempool because it's not at the estimated tip. That's something we can fix with a debug option, and it's a lot safer than skipping validation entirely.

teor2345:
Hey @oxarbitrage I got the send transactions test working with the mempool gRPCs, you can see a bunch of the details on PR #5016.

> Doing it in the send transaction test is not a good idea for the following reasons, i gave up with that in a previous research months ago:
> 
> 1- You need to have double of the space of the blockchain available as the test do a full copy of the blockchain.

This will be fixed in ticket #5015, but it's not a problem at the moment because we've almost finished the lightwalletd tests.

> 2- Test takes hours to run locally, so it is a real pain to debug, this test is a lot quicker.

It only takes about 20 minutes for me, I'm not sure if your disks are slow, or if the copy was doing too much.
I did some fixes for this in PR #5016.

> 3- There is no guarantee that the transactions inserted in the test will get into the mempool, in fact, in general they dont. The test just check that there are no errors when sending transactions: https://github.com/ZcashFoundation/zebra/blob/grpc_mempool_tests/zebrad/tests/common/lightwalletd/send_transaction_test.rs#L131 but make them get inserted into the mempool is another story.

The transactions were going into the mempool, but `lightwalletd` wasn't returning them, because Zebra wasn't at the tip. I added a debug option to fake being at the tip. (This is safer for users than having options that disable consensus rule validation, it's also better in an audit.)

Also right now `lightwalletd` doesn't return any transactions at all, so I just checked the `zebrad` logs instead.

oxarbitrage:
> It only takes about 20 minutes for me, I'm not sure if your disks are slow, or if the copy was doing too much.
I did some fixes for this in PR https://github.com/ZcashFoundation/zebra/pull/5016.

Yea, well, building in release mode helps. Thank you for that (https://github.com/ZcashFoundation/zebra/pull/5016#issuecomment-1232482447). My disk are not ssd for this test because i had to hold the chain and the copy so that also makes it slower. 

Another factor is that it depends where you are from the tip. For example my zebra node was in the tip maybe 4 days ago so now the test is trying to get there but very slow. I had started a test using https://github.com/ZcashFoundation/zebra/pull/5016 like 3 hours ago and it is trying to sync (now at 99.757%).

oxarbitrage:
> Also right now lightwalletd doesn't return any transactions at all, so I just checked the zebrad logs instead.

Actually, locally, lightwalletd returns transactions if they are in the mempool by using `GetMempoolTx` but it will not when `GetMempoolStream`

oxarbitrage:
Replaced with the real PR at https://github.com/ZcashFoundation/zebra/pull/5016

PR #4997, opened 2022-08-30
fix(log): Increase slow code log thresholds to reduce verbose logs and warnings

teor2345:
## Motivation

We've decided that Zebra's current performance is mostly ok, so we want to reduce the number of performance logs.

Closes ticket #4932

## Solution

- change info threshold to 5 minutes and warn to 9 minutes

## Review

This is a routine usability fix.

### Reviewer Checklist

  - [ ] CI passes

### DIFF
--- a/zebra-chain/src/diagnostic.rs
+++ b/zebra-chain/src/diagnostic.rs
@@ -7,10 +7,14 @@ use std::time::{Duration, Instant};
 use crate::fmt::duration_short;
 
 /// The default minimum info-level message time.
-pub const DEFAULT_MIN_INFO_TIME: Duration = Duration::from_secs(5);
+///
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):

oxarbitrage (commented):
This does not seems to close https://github.com/ZcashFoundation/zebra/issues/4932 fully but a quick fix.

Here is what i think we can do instead of adding/removing logging to/from the main branch each time we want to see the performance of something:

- https://github.com/ZcashFoundation/zebra/pull/4952#issuecomment-1231085651. If that works we can do performance checks outside main.
- Remove all codetimer code we have now in main.
- Decide then what code logs can be useful to the end user and what log level we want to use for it, if any.

PR #5001, opened 2022-08-30
fix(ci): Split a long full sync job

teor2345:
## Motivation

A full sync job is timing out, so we want to split the job.

### Designs

I followed the instructions in:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#resolving-ci-sync-timeouts

## Solution

- Add a job that stops at height 1780k, roughly 4 hours long

## Review

Anyone can review this PR, we need it to fix full syncs on `main`.

### Reviewer Checklist

  - [ ] CI passes

## Follow Up Work

We might want to find an alternative to all these extra jobs, maybe #4937 will speed up these syncs.

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -755,7 +755,7 @@ jobs:
   # follow the logs of the test we just launched, up to block 1,760,000 or later
   # (or the test finishing)
   #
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
I think it's time to think on a self hosted runner, at least. So we don't have to keep creating more jobs, and just allow a hosted to run until it's done.

PR #5009, opened 2022-08-30
build(deps): bump prost, tonic, tonic-build and console-subscriber

oxarbitrage:
## Motivation

We want to update `prost`, `tonic`, `tonic-build` and `console-subscriber` at the same time. 

Replaces https://github.com/ZcashFoundation/zebra/pull/4853

## Solution

Upgrade the 4 crates manually, clippy issues that were blocking this were resolved.

## Review

I think anyone can review.

### Reviewer Checklist

  - [ ] PR builds
  - [ ] All tests pass

### DIFF
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -110,6 +110,9 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Install last version of Protoc
+        uses: arduino/setup-protoc@v1
+
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -112,6 +112,8 @@ jobs:
 
       - name: Install last version of Protoc
         uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
 
       - uses: actions-rs/toolchain@v1
         with:
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -114,6 +114,7 @@ jobs:
         uses: arduino/setup-protoc@v1
         with:
           version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: actions-rs/toolchain@v1
         with:
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -80,6 +80,8 @@ jobs:
 
       - name: Install last version of Protoc
         uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
 
       - uses: actions-rs/toolchain@v1
         with:
@@ -204,6 +206,8 @@ jobs:
 
       - name: Install last version of Protoc
         uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
 
       - uses: actions-rs/toolchain@v1
         with:
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -82,6 +82,7 @@ jobs:
         uses: arduino/setup-protoc@v1
         with:
           version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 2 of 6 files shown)


oxarbitrage:
Hmm, this needs a new dependency, more info in https://github.com/hyperium/tonic/issues/1047

Ill try to do this locally and then try to figure out where in the CI we need to add the new dependency. Also this will need changes to the README so is a bit more work than i was expecting but ill work on it.

oxarbitrage:
Ok, so i fixed this locally. As it says in the issue i posted above this now requires `protoc` to be in your OS. I installed this by using `apt-get` in my linux but this installed `protoc 3.0` which is too old and gives an error when running clippy on zebra:

```
  thread 'main' panicked at 'Failed to generate lightwalletd gRPC files: Custom { kind: Other, error: "protoc failed: compact_formats.proto:8:8: Option \"swift_prefix\" unknown.\nservice.proto: Import \"compact_formats.proto\" was not found or had errors.\nservice.proto:143:36: \"CompactBlock\" is not defined.\nservice.proto:145:51: \"CompactBlock\" is not defined.\nservice.proto:166:47: \"CompactTx\" is not defined.\n" }', zebrad/build.rs:71:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So i installed the last version using the binaries provided at https://github.com/protocolbuffers/protobuf/releases/tag/v21.5

With this version everything looks good. 

We need to install this in the CI, what other projects are doing is adding https://github.com/arduino/setup-protoc to the workflow.

for example https://github.com/hyperium/tonic/blob/master/.github/workflows/CI.yml#L29

teor2345:
> Also this will need changes to the README so is a bit more work than i was expecting but ill work on it.

We're only using these dependencies in tests, so we can update the developer documentation and link to it from the README. (We're trying to keep the README short.)

It's also ok to just link to the line in the Dockerfile or GitHub workflow that installs dependencies, because they change pretty often.

Also this PR removes the need for cmake in CI.

teor2345:
The `cargo deny` output shows a duplicate dependency, with the older version coming via `console-subscriber`:
>    = prost v0.10.4
>      ├── console-api v0.3.0
>      │   └── console-subscriber v0.1.6
>      │       └── zebrad v1.0.0-beta.13

https://github.com/ZcashFoundation/zebra/runs/8103900323?check_suite_focus=true#step:4:24

Try upgrading `console-subscriber` in this PR as well.

upbqdn (commented):
All looks good. I tested this PR locally, and the following checks passed with the `protoc` binary installed:
```bash
cargo fmt --all -- --check
cargo deny --workspace --all-features check bans sources
cargo clippy --all-targets --release --all-features
cargo test --release
```

I looked into the book to check where we could document that Zebra devs need `protoc`, but I didn't find any suitable place where this could be mentioned. Maybe a subsection here https://zebra.zfnd.org/CONTRIBUTING.html?

oxarbitrage:
Not sure what is this one: https://github.com/ZcashFoundation/zebra/runs/8142400656?check_suite_focus=true#step:3:12

upbqdn:
> Not sure what is this one: https://github.com/ZcashFoundation/zebra/runs/8142400656?check_suite_focus=true#step:3:12

Did we really reach the rate limit?
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions

The error also says that authenticated users get higher limit.

upbqdn:
> Also this PR removes the need for cmake in CI.

Why is that so?

teor2345 (changes requested):
> > Not sure what is this one: https://github.com/ZcashFoundation/zebra/runs/8142400656?check_suite_focus=true#step:3:12
> 
> Did we really reach the rate limit? https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions
> 
> The error also says that authenticated users get higher limit.

Maybe @gustavovalverde can help, it seems like we need to fix the rate-limit before we merge this PR.

gustavovalverde:
Sorry for the delay here @oxarbitrage. The fix has been applied for the actions, just a Cargo.lock conflict is missing. I'll leave the rest to you.

> I pinned the version to `3.x` so we don't have issues in the future with this.

teor2345 (dismissed):
I bumped the `console-subscriber` version again, and resolved the `Cargo.lock` conflict.

teor2345 (approved):

PR #5013, opened 2022-08-31
Breaking: fix(rpc): Add a config for multi-threaded RPC server requests

### LINKED ISSUE #5012: Run JSON-RPC queries in multiple threads
## Motivation

ZecWallet Lite have asked us to enable multi-threaded execution in our RPC library.

Before PR #4806, the RPC library shared the same multi-threaded `tokio` executor as the rest of Zebra. #4806 splits the RPC library into into its own executor, but it's single-threaded.

This might be a performance regression, it needs to be fixed before the next Zebra release.

Depends-On: #4806 

### API Reference

This is a single-line change when starting the server:
https://docs.rs/jsonrpc-http-server/latest/jsonrpc_http_server/struct.ServerBuilder.html#method.threads

### Designs

We also need to:
- configure the number of threads
- document that multi-threaded execution disables port conflict detection for JSON-RPC, so it's possible to have multiple Zebra instances listening on the same port

teor2345:
## Changelog

See the README changes in this PR for a changelog breaking change notice.

## Motivation

Our users have asked for multi-threaded RPCs.

## Solution

- Add a config for multi-threaded RPCs
- Add tests for multi and single threaded RPC calls
- Run `lightwalletd` tests with multi-threaded RPCs

Depends-On: #4806
Fixes #5012

## Review

This is a routine bug fix.

### Reviewer Checklist

  - [x] Code implements Specs and Designs
  - [x] Tests for Expected Behaviour
  - [x] Tests for Errors

### DIFF
--- a/zebra-rpc/src/server/tests/vectors.rs
+++ b/zebra-rpc/src/server/tests/vectors.rs
@@ -37,7 +37,7 @@ fn rpc_server_spawn(parallel_cpu_threads: bool) {
     let port = zebra_test::net::random_known_port();
     let config = Config {
         listen_addr: Some(SocketAddrV4::new(Ipv4Addr::LOCALHOST, port).into()),
-        parallel_cpu_threads: 1,
+        parallel_cpu_threads: if parallel_cpu_threads { 2 } else { 1 },
     };
 
     let rt = tokio::runtime::Runtime::new().unwrap();
@@ -101,7 +101,7 @@ fn rpc_server_spawn_unallocated_port(parallel_cpu_threads: bool) {
     let port = zebra_test::net::random_unallocated_port();
     let config = Config {
         listen_addr: Some(SocketAddrV4::new(Ipv4Addr::LOCALHOST, port).into()),
-        parallel_cpu_threads: 1,
+        parallel_cpu_threads: if parallel_cpu_threads { 0 } else { 1 },
     };
 
     let rt = tokio::runtime::Runtime::new().unwrap();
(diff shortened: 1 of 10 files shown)


zebra-rpc/src/server/tests/vectors.rs
```diff
@@ -62,14 +77,31 @@ fn rpc_server_spawn() {
...
+///
+/// Set `parallel_cpu_threads` to true to auto-configure based on the number of CPU cores.
+#[tracing::instrument]
+fn rpc_server_spawn_unallocated_port(parallel_cpu_threads: bool) {
     let _init_guard = zebra_test::init();
 
     let port = zebra_test::net::random_unallocated_port();
     let config = Config {
         listen_addr: Some(SocketAddrV4::new(Ipv4Addr::LOCALHOST, port).into()),
+        parallel_cpu_threads: 1,
```
arya2:
```suggestion
        parallel_cpu_threads: if parallel_cpu_threads { 0 } else { 1 },
```

arya2 (changes requested):
Looks good, one small fix suggested for the `rpc_sever_spawn_parallel_threads()` test

zebra-rpc/src/server/tests/vectors.rs
teor2345: Thanks for catching this!

I fixed it in 0bcbcc87d26f2c1f38cc4bf9d0bc45da51df947f.
I also squashed the fixup commits (and rebased on the latest `main` branch).

arya2 (dismissed):
Looks good

arya2:
[isolate-rpcs](https://github.com/ZcashFoundation/zebra/tree/isolate-rpcs) merged, should probably target main now

oxarbitrage (approved):
Already approved before by @arya2 , i also checked the PR and looks good to me too so i am re approving so it can gets merged.

PR #5026, opened 2022-09-01
change(doc): Make the release instructions more explicit

upbqdn:
## Motivation & Solution

When doing the release, the author of the release has to _replace_ the prepopulated draft changelog with the final changelog they created instead of just _copying_ it into the release description.

This is a small and optional change that makes the checklist a bit easier to follow when doing the release.


oxarbitrage (approved):
makes sense, thanks.

PR #5032, opened 2022-09-01
fix(test) switch zebrad to a non-blocking tracing logger

### LINKED ISSUE #4834: Switch zebrad to a non-blocking tracing logger
## Motivation

Some of our tests might be deadlocking when the log buffers fill up.
This could happen to users as well.

This is a high priority, because it is causing test failures.

### API Reference

https://docs.rs/tracing-appender/0.2.2/tracing_appender/#non-blocking-writer

### Designs

- Use `tracing-appender` to make stdout writes non-blocking
- Use a lossy logger, so we drop any output we can't write, rather than blocking

arya2:
## Motivation

We want to switch zebrad to a non-blocking tracing logger to use a separate thread when writing tracing logs to stdout so that our tests won't deadlock.

Fixes #4834.

## Solution

- Provide tracing FmtSubscriber (or fmt::Layer when tokio-console feature is enabled) with a lossy [non-blocking writer](https://docs.rs/tracing-appender/0.2.2/tracing_appender/#non-blocking-writer) to use instead of stdout

## Review

Anyone can review this PR

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -354,6 +354,7 @@ impl FinalizedState {
         let _ = stderr().lock().flush();
 
         // Give some time to logger thread to flush out any remaining lines to stdout
+        // and yield so that tests pass on MacOS
         std::thread::sleep(std::time::Duration::from_secs(3));
 
         // Exits before calling drop on the WorkerGuard for the logger thread,
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -353,6 +353,9 @@ impl FinalizedState {
         let _ = stdout().lock().flush();
         let _ = stderr().lock().flush();
 
+        // Give some time to logger thread to flush out any remaining lines to stdout
+        std::thread::sleep(std::time::Duration::from_secs(3));
+
         // Exits before calling drop on the WorkerGuard for the logger thread,
         // dropping any lines that haven't already been written to stdout.
         // This is okay for now because this is test-only code
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -106,6 +106,11 @@ impl FinalizedState {
                 // So we want to drop it before we exit.
                 std::mem::drop(new_state);
 
+                // Drops tracing log output that's hasn't already been written to stdout
+                // since this exits before calling drop on the WorkerGuard for the logger thread.
+                // This is okay for now because this is test-only code
+                //
+                // TODO: Call ZebradApp.shutdown or drop its Tracing component before calling exit_process to flush logs to stdout
                 Self::exit_process();
             }
         }
@@ -307,6 +312,11 @@ impl FinalizedState {
             // We're just about to do a forced exit, so it's ok to do a forced db shutdown
             self.db.shutdown(true);
 
+            // Drops tracing log output that's hasn't already been written to stdout
+            // since this exits before calling drop on the WorkerGuard for the logger thread.
+            // This is okay for now because this is test-only code
+            //
+            // TODO: Call ZebradApp.shutdown or drop its Tracing component before calling exit_process to flush logs to stdout
             Self::exit_process();
         }
 
@@ -343,6 +353,9 @@ impl FinalizedState {
         let _ = stdout().lock().flush();
         let _ = stderr().lock().flush();
 
+        // Exits before calling drop on the WorkerGuard for the logger thread,
+        // dropping any lines that haven't already been written to stdout.
+        // This is okay for now because this is test-only code
         std::process::exit(0);
     }
 }
--- a/zebrad/src/components/tracing/component.rs
+++ b/zebrad/src/components/tracing/component.rs
@@ -47,12 +47,11 @@ impl Tracing {
         let filter = config.filter.unwrap_or_else(|| "".to_string());
         let flame_root = &config.flamegraph;
 
-        // Builds a lossy NonBlocking logger with a line limit of 8_000, lines sent to the worker past
-        // this limit (via the write method which sends it to a crossbeam::bounded channel) 
-        // will be dropped without being written to stdout
+        // Builds a lossy NonBlocking logger with a default line limit of 128_000 or an explicit buffer_limit.
+        // The write method queues lines down a bounded channel with this capacity to a worker thread that writes to stdout.
+        // Increments error_counter and drops lines when the buffer is full.
         let (non_blocking, _guard) = NonBlockingBuilder::default()
-            .lossy(true)
-            .buffered_lines_limit(8_000)
+            .buffered_lines_limit(config.buffer_limit.max(100))
             .finish(std::io::stdout());
 
         // Only use color if tracing output is being sent to a terminal or if it was explicitly
--- a/zebrad/src/components/tracing/component.rs
+++ b/zebrad/src/components/tracing/component.rs
@@ -10,7 +10,7 @@ use tracing_subscriber::{
     EnvFilter,
 };
 
-use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+use tracing_appender::non_blocking::{NonBlocking, NonBlockingBuilder, WorkerGuard};
 
 use crate::{application::app_version, config::TracingSection};
 
@@ -47,9 +47,13 @@ impl Tracing {
         let filter = config.filter.unwrap_or_else(|| "".to_string());
         let flame_root = &config.flamegraph;
 
-        // By default, the built NonBlocking will be lossy with a line limit of 128_000, lines sent to the worker past
-        // this limit will be dropped without being written to stdout
-        let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
+        // Builds a lossy NonBlocking logger with a line limit of 8_000, lines sent to the worker past
+        // this limit (via the write method which sends it to a crossbeam::bounded channel) 
+        // will be dropped without being written to stdout
+        let (non_blocking, _guard) = NonBlockingBuilder::default()
+            .lossy(true)
+            .buffered_lines_limit(8_000)
+            .finish(std::io::stdout());
 
         // Only use color if tracing output is being sent to a terminal or if it was explicitly
         // forced to.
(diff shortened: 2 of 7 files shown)


zebrad/tests/acceptance.rs
```diff
@@ -1212,6 +1212,51 @@ async fn rpc_endpoint() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn non_blocking_logger() -> Result<()> {
+    let _init_guard = zebra_test::init();
+
+    // Write a configuration that has RPC listen_addr set
+    // [Note on port conflict](#Note on port conflict)
+    let mut config = random_known_rpc_port_config()?;
```
teor2345: We might need to set `config.tracing.filter` to "debug" or "trace" to get enough log output in this test.

zebrad/tests/acceptance.rs
```diff
@@ -1212,6 +1212,51 @@ async fn rpc_endpoint() -> Result<()> {
...
+
+    // Wait until port is open.
+    child.expect_stdout_line_matches(
+        format!("Opened RPC endpoint at {}", config.rpc.listen_addr.unwrap()).as_str(),
+    )?;
+
+    // Create an http client
+    let client = reqwest::Client::new();
+
+    for _ in 0..10_000 {
```
teor2345: Is this enough to reach the default 128_000 line limit?

zebrad/src/components/tracing/component.rs
```diff
@@ -32,6 +47,9 @@ impl Tracing {
         let filter = config.filter.unwrap_or_else(|| "".to_string());
         let flame_root = &config.flamegraph;
 
+        // By default, the built NonBlocking will be lossy. (with a line limit of 128_000)
```
teor2345: What happens when the line limit is reached?
Can you document that here?

zebrad/src/components/tracing/component.rs
```diff
@@ -16,14 +22,23 @@ pub struct Tracing {
...
     /// The originally configured filter.
     initial_filter: String,
 
     /// The installed flame graph collector, if enabled.
     #[cfg(feature = "flamegraph")]
     flamegrapher: Option<flame::Grapher>,
+
+    /// Drop guard for worker thread of non-blocking logger,
+    /// responsible for flushing any remaining logs when the program terminates
+    _guard: WorkerGuard,
```
teor2345, line 41: Zebra calls `sys::exit()` in a few different places, how can we make sure the guard is dropped before that happens?

This is the standard shutdown after a Ctrl-C. I think we should have called all the destructors here, but I'm not sure. Do we need a test for that?
https://github.com/ZcashFoundation/zebra/blob/3ff56c22cda4504832d4fe57fd6357e8743965ed/zebrad/src/application.rs#L468-L475

We only call this code during tests, so we've decided not to fix the `exit()` calls here. It's ok if we lose a bit of output in the test, as long as the test still passes. CI will check this for us.
https://github.com/ZcashFoundation/zebra/blob/dc9da24c8b6cee5c05315f0d77bee52613f75fd3/zebra-state/src/service/finalized_state.rs#L298-L311

teor2345 (changes requested):
This looks great!

I just want to double-check a few edge cases. Let me know if you have any questions, or you'd like me to help with the code changes.

zebrad/tests/acceptance.rs
arya2: Definitely not, this is to test that there isn't a deadlock when there are a lot of queries in quick succession - but it should be more than the buffer limit so that it also tests that it will lose lines past the line limit rather than block - I updated the line limit and the number of iterations in the test in [`6d2d118`](https://github.com/ZcashFoundation/zebra/pull/5032/commits/6d2d1186f83176b1cef9b1187f68bb7d5dddc96b)

zebrad/src/components/tracing/component.rs
arya2: It ignores any new calls to write instead of queuing them in the channel, I added some more detail here

zebrad/src/components/tracing/component.rs
arya2: It looks like it's already calling all the destructors on self except when there's a FrameworkError in one of the `before_shutdown()` methods, neither our components that have that method nor the default implementation return a FrameworkError, but if one were added that might cause problems so I'll add a `mem::take` there as well

zebrad/src/components/tracing/component.rs
```diff
@@ -32,6 +47,13 @@ impl Tracing {
         let filter = config.filter.unwrap_or_else(|| "".to_string());
         let flame_root = &config.flamegraph;
 
+        // Builds a lossy NonBlocking logger with a default line limit of 128_000 or an explicit buffer_limit.
+        // The write method queues lines down a bounded channel with this capacity to a worker thread that writes to stdout.
+        // Increments error_counter and drops lines when the buffer is full.
+        let (non_blocking, _guard) = NonBlockingBuilder::default()
+            .buffered_lines_limit(config.buffer_limit.max(100))
```
teor2345, line 54: TODO: we'll want to increase this limit once we have the tests passing

zebrad/tests/acceptance.rs
```diff
@@ -1212,6 +1212,77 @@ async fn rpc_endpoint() -> Result<()> {
...
+    });
+
+    // Wait until the spawned task finishes
+    std::thread::sleep(Duration::from_secs(10));
+
+    rt.shutdown_timeout(Duration::from_secs(3));
+
+    match test_task_handle.now_or_never() {
+        Some(Ok(result)) => result,
+        Some(Err(_)) => Err(eyre!("join error")),
```
teor2345: The test is failing with this error, so I'm going to push a commit that prints the error:
```suggestion
        Some(Err(error)) => Err(eyre!("join error: {:?}", error)),
```

zebrad/tests/acceptance.rs
```diff
@@ -1212,6 +1212,77 @@ async fn rpc_endpoint() -> Result<()> {
...
+        let zebra_rpc_address = config.rpc.listen_addr.unwrap();
+
+        let dir = testdir()?.with_config(&mut config)?;
+        let mut child = dir.spawn_child(args!["start"])?;
+        // Wait until port is open.
+        child.expect_stdout_line_matches(
+            format!("Opened RPC endpoint at {}", config.rpc.listen_addr.unwrap()).as_str(),
+        )?;
+
+        let mut cmd2 = std::process::Command::new("echo")
```
teor2345: I'm not sure if this extra task is needed for the test. So once we have the tests passing, we might want to delete it.

zebrad/tests/acceptance.rs
```diff
@@ -1212,6 +1212,77 @@ async fn rpc_endpoint() -> Result<()> {
...
+            .assert_was_killed()
+            .wrap_err("Possible port conflict. Are there other acceptance tests running?")?;
+
+        Ok(())
+    });
+
+    // Wait until the spawned task finishes
+    std::thread::sleep(Duration::from_secs(10));
+
+    rt.shutdown_timeout(Duration::from_secs(3));
```
teor2345: The GitHub runners are slower machines, so we might need to increase the timeouts here.

We could also use a condition variable to signal when the task has finished, which would let us have a much longer timeout for slow machines:
https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait_timeout

teor2345:
I noticed that some tests were failing, so I pushed a commit that shows the error they are failing with.

zebrad/tests/acceptance.rs
arya2: It was not needed, deleted in [`590bb91`](https://github.com/ZcashFoundation/zebra/pull/5032/commits/590bb914f65ffd8fd0da57446077bf6f6f644e21)

zebrad/tests/acceptance.rs
arya2: I updated the timeout to 45 seconds with a `mpsc::channel` for notifying the main thread when it's done instead of a condvar for simplicity. It was iterating far more than needed so I reduced the number of iterations as well. The Github runner should be okay now.

zebrad/src/components/tracing/component.rs
```diff
@@ -47,6 +69,7 @@ impl Tracing {
 
             let logger = FmtSubscriber::builder()
                 .with_ansi(use_color)
+                .with_writer(non_blocking)
```
teor2345, line 72: Do we need to add the writer to the `ErrorLayer` as well, or does it use the same writer as the `FmtSubscriber`?

zebrad/src/components/tracing/component.rs
```diff
@@ -82,6 +105,7 @@ impl Tracing {
             // Using `FmtSubscriber` as the base subscriber, all the logs get filtered.
             let logger = fmt::Layer::new()
                 .with_ansi(use_color)
+                .with_writer(non_blocking)
```
teor2345, line 108: Do we need to add the writer to the `ErrorLayer` as well, or does it use the same writer as the `fmt::Layer`?

teor2345 (commented):
It looks like two other tests are failing with these changes, but only on macOS:
- restart_stop_at_height
- sync_one_checkpoint_mainnet

These are the tests that `exit()` abruptly, and don't drop the worker guard.

Here are some things to try:
- make the test-only state code sleep for a few seconds before calling `exit()`
- make the test-only state code panic instead of exiting (this could be complicated, I'm not sure if we propagate panics correctly from that far inside the state)

zebrad/src/components/tracing/component.rs
arya2: No need or ability to add the writer to the ErrorLayer, it uses the writer from the FmtSubscriber

zebrad/src/components/tracing/component.rs
arya2: Here too, every layer uses the writer from fmt::Layer

teor2345:
> Here are some things to try:

If we can't fix it quickly, it's ok to disable these tests on macOS. It's a tier 2 platform, so we don't guarantee test coverage.

zebra-state/src/service/finalized_state.rs
```diff
@@ -343,6 +353,12 @@ impl FinalizedState {
         let _ = stdout().lock().flush();
         let _ = stderr().lock().flush();
 
+        // Give some time to logger thread to flush out any remaining lines to stdout
```
arya2, line 356:
```suggestion
        // Give some time to logger thread to flush out any remaining lines to stdout
        // and yield so that tests pass on MacOS
```

teor2345 (approved):
Thanks for fixing those tests, let's go!

PR #5082, opened 2022-09-05
fix(ci): delete GCP resources, but keep some recent cached state images

teor2345:
## Motivation

- The current GCP delete command has a typo
- Some of the deletion schedules are very long, and storing terabytes of data is expensive
- We're deleting the only checkpoint disk every 60 days, forcing the next PR that runs to rebuild it

## Solution

Functional changes:
- start deleting disks (as well as images and instances)
- delete instances, disks, and images after 7 days
- preserve the last 3 images of each type
- use the correct delete command
- change the schedule to weekly, just before Teor starts their week (so there is someone around if it causes any issues)

Diagnostic changes:
- show which images and disks are deleted
- show output of failed commands
- silence the warning about GCP credential files

## Review

I think @gustavovalverde might want to review this PR.

### Reviewer Checklist

  - [ ] Manual workflow run passes

### DIFF
--- a/.github/workflows/delete-gcp-resources.yml
+++ b/.github/workflows/delete-gcp-resources.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Delete old instance templates
         run: |
           DELETE_BEFORE_DATE=$(date --date="$DELETE_AGE_DAYS days ago" '+%Y%m%d')
-          TEMPLATES=$(gcloud compute instance-templates list --sort-by=creationTimestamp  --filter="creationTimestamp < $DELETE_BEFORE_DATE" --format='value(NAME)')
+          TEMPLATES=$(gcloud compute instance-templates list --sort-by=creationTimestamp  --filter="name~-[0-9a-f]+$ AND creationTimestamp < $DELETE_BEFORE_DATE" --format='value(NAME)')
 
           for TEMPLATE in $TEMPLATES
           do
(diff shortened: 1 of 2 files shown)


.github/workflows/delete-gcp-resources.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           workload_identity_provider: 'projects/143793276228/locations/global/workloadIdentityPools/github-actions/providers/github-oidc'
           service_account: 'github-service-account@zealous-zebra.iam.gserviceaccount.com'
           token_format: 'access_token'
+          create_credentials_file: false
```
gustavovalverde:
```suggestion
```

I tested this as I recall using this approach failed before. I once included this as I saw a GHA's warning, but it doesn't work as expected.

gustavovalverde (changes requested):
A recent change makes the job fail https://github.com/ZcashFoundation/zebra/runs/8197310288?check_suite_focus=true#step:3:29

.github/workflows/delete-gcp-resources.yml
teor2345: Ok, we should probably use the checkout action to silence the warning

teor2345: Fixed by:
https://github.com/ZcashFoundation/zebra/pull/5082/files#diff-934c576d4f57156181f9078afdfef6c33b80cca73f5825f6f06ad72ed018a4efR25

teor2345:
I ran the workflow manually here:
- https://github.com/ZcashFoundation/zebra/runs/8197676596?check_suite_focus=true

teor2345:
We'll need to re-run the workflow after the current run finishes, I made a few tweaks.

teor2345:
I ran it manually and it seemed to work fine:
https://github.com/ZcashFoundation/zebra/actions/runs/2997031847

teor2345:
This still leaves us with 400 cached state disks, but I think we might want to fix that by creating fewer disks.

gustavovalverde (approved):

PR #5088, opened 2022-09-06
fix(ci): Wait 1 day before creating cached state image updates 

teor2345:
## Motivation

We've created 400 cached state images over the past week, that's about 40 TB of data.
This is about 25% of our monthly bill.

## Solution

- Skip creating updated cached state images unless the original image is at least 1 day old (1152 blocks)
  - Get the original height from the image labels
  - Fix the update flag calculation
  - Set the zebrad app name in cached state jobs
- Skip creating updated lightwalletd images - they happen right after a full lightwalletd sync
- Make the sync height step fail if there is no sync height (this is either a Zebra, lightwalletd, or CI bug)

## Review

Anyone can review this PR.

### Reviewer Checklist

  - [x] Code makes sense
  - [x] CI passes
  - [x] Some updates get skipped

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -420,7 +420,7 @@ jobs:
       test_variables: '-e TEST_LWD_UPDATE_SYNC=1 -e ZEBRA_TEST_LIGHTWALLETD=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache -e LIGHTWALLETD_DATA_DIR=/var/cache/lwd-cache'
       needs_zebra_state: true
       needs_lwd_state: true
-      # since we do a full sync in every PR, the new cached state is only be an hour new than the original one
+      # since we do a full sync in every PR, the new cached state will only be a few minutes newer than the original one
       saves_to_disk: false
       disk_prefix: lwd-cache
       disk_suffix: tip
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -169,6 +169,8 @@ jobs:
       zebra_skip_ipv6_tests: '1'
       rust_log: info
 
+  # zebrad tests without cached state
+
   # Run all the zebra tests, including tests that are ignored by default.
   # Skips tests that need a cached state disk or a lightwalletd binary.
   #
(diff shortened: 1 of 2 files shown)


.github/workflows/continous-integration-docker.yml
```diff
@@ -417,8 +420,8 @@ jobs:
       test_variables: '-e TEST_LWD_UPDATE_SYNC=1 -e ZEBRA_TEST_LIGHTWALLETD=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache -e LIGHTWALLETD_DATA_DIR=/var/cache/lwd-cache'
       needs_zebra_state: true
       needs_lwd_state: true
-      # update the disk on every PR, to increase CI speed
-      saves_to_disk: true
+      # since we do a full sync in every PR, the new cached state is only be an hour new than the original one
```
```suggestion
      # since we do a full sync in every PR, the new cached state will only be a few minutes newer than the original one
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1111,7 +1114,7 @@ jobs:
           token_format: 'access_token'
 
       # Get the state version from the local constants.rs file to be used in the image creation,
-      # as the state version is part of the disk image name.
+      # as the state version is part of the disk image.
```
```suggestion
      # as the state version is part of the disk image name.
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1120,53 +1123,95 @@ jobs:
...
 
-      # Sets the $UPDATE_SUFFIX env var to "-u" if using cached state,
-      # and the empty string otherwise.
+      # Get the original cached state height from google cloud.
       #
-      # Also sets a unique date and time suffix $TIME_SUFFIX.
-      - name: Set update and time suffixes
+      # If the height is missing from the image labels, uses zero instead.
+      # TODO: fail the job if needs_zebra_state but the height is missing
+      #       after all the old images have been deleted, this should happen around 15 September 2022
```
```suggestion
      #       we can make this change after all the old images have been deleted, this should happen around 15 September 2022
```

teor2345:
This seems to work:
> Skipped cached state update because the new sync height 1798929 was less than 1152 blocks above the original height 1798761

https://github.com/ZcashFoundation/zebra/runs/8201687557?check_suite_focus=true#step:10:83

arya2 (dismissed):
Logic and behavior looks good to me.

arya2 (approved):

PR #5100, opened 2022-09-06
add(doc): Explain how Mergify works, and when it can be used

teor2345:
## Motivation

We want to document how mergify works for developers and external contributors.


oxarbitrage (approved):
looks good, thank you for adding it.

PR #5106, opened 2022-09-07
doc(consensus): Document why we do a UTXO check that looks redundant

teor2345:
## Motivation

This UTXO check looks redundant, but it is actually needed for concurrent verification.

## Solution

- Document how this consensus rule is implemented
- Inline the call stack for the check, so it is efficient when it is not needed (most of the checkpoint sync)

## Review

Anyone can review this PR.

### Reviewer Checklist

  - [x] Documentation makes sense


teor2345:
@mergifyio update

arya2 (approved):
Makes sense

Issue #5044, opened 2022-09-02
Fix "The resource is not ready" errors in CI


teor2345:
## Motivation

We've had multiple PRs fail with:
> Run gcloud compute instances create-with-container "update-to-tip-5020-merge-5690565" \
> ERROR: (gcloud.compute.instances.create-with-container) Could not fetch resource:
> - The resource 'projects/zealous-zebra/global/images/zebrad-cache-main-bc7294f-v25-mainnet-tip-u-0901230920' is not ready

https://github.com/ZcashFoundation/zebra/runs/8146703881?check_suite_focus=true#step:7:77

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/5020#issuecomment-1234993941_

## Tasks

- [ ] When selecting images for tests, add a `status: READY` filter
- [ ] When selecting the 2 images to keep, add a `status: READY` filter
- [ ] When selecting images to delete, _*do not*_ add a `status: READY` filter, because we are already filtering out recent images

teor2345:
I haven't seen these errors recently, I wonder if they are caused by rate-limiting, or they are more common under load?

We can fix this error by searching for images with `status: READY` whenever we list images:
```sh
$ gcloud compute images describe lwd-cache-4721-merge-823b8a9-v25-mainnet-tip-u-0901181806 
archiveSizeBytes: '172968691968'
creationTimestamp: '2022-09-01T11:18:09.904-07:00'
description: 'Created from commit 823b8a9 with height '
diskSizeGb: '200'
id: '6653236955544635791'
kind: compute#image
labelFingerprint: 42WmSpB8rSM=
name: lwd-cache-4721-merge-823b8a9-v25-mainnet-tip-u-0901181806
selfLink: https://www.googleapis.com/compute/v1/projects/zealous-zebra/global/images/lwd-cache-4721-merge-823b8a9-v25-mainnet-tip-u-0901181806
sourceDisk: https://www.googleapis.com/compute/v1/projects/zealous-zebra/zones/us-central1-a/disks/lwd-full-sync-823b8a9
sourceDiskId: '4921097591623342289'
sourceType: RAW
status: READY
storageLocations:
- us
```

teor2345:
I saw this error again today:
> ERROR: (gcloud.compute.instances.create-with-container) Could not fetch resource:
[...]

PR #5107, opened 2022-09-07
2. change(state): Run AwaitUtxo read requests without shared mutable chain state

### LINKED ISSUE #5102: Make StateService requests into concurrent ReadStateService requests
## Motivation

At the moment, Zebra reads the state via two different services:
- a blocking `StateService`
- a concurrent `ReadStateService`

Most read requests are able to be run concurrently, so we could move them to the `ReadStateService`. This allows us to move mutable chain state into the block commit task.

This should also improve performance.

teor2345:
## Motivation

This PR runs `StateService::AwaitUtxo` read requests without accessing shared mutable chain state.
As a useful side-effect, this makes all `StateService` read requests concurrent.

Closes #5102.

### Designs

## Solution

Don't access shared mutable chain state in `StateService` read requests:
- Re-Implement `AwaitUtxo` without accessing shared mutable chain state, using `StateService.pending_utxos` and `ReadStateService::AnyUtxo`
  - Send the whole `NonFinalizedState` to the `ReadStateService`
  - Rename the `ChainUtxo` request to `BestChainUtxo`
  - Add an `AnyChainUtxo` request

Related documentation:
- Update the state service module documentation
- Explain how spent UTXOs are treated by the state
- Clarify how cached Chains impact state read requests

Related Cleanups:
- Move `AwaitUtxos` next to the other shared writeable state requests
- Rename `ReadResponse::Utxos` to `ReadResponse::AddressUtxos`
- Rename an `out_point` variable to `outpoint` for consistency
- Rename `transparent_utxos` to `address_utxos`
- Move the `QueuedBlock` type into the `queued_blocks` module

## Testing

We should run a full sync on this before we merge, and make sure it is not significantly slower than the `main` branch.

## Review

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

Actually create a separate task to commit blocks to the state.

### DIFF
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -549,11 +549,20 @@ pub enum ReadRequest {
     /// Looks up a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
     /// returning `None` immediately if it is unknown.
     ///
-    /// Checks verified blocks in the finalized chain and non-finalized chain.
+    /// Checks verified blocks in the finalized chain and the _best_ non-finalized chain.
     ///
     /// This request is purely informational, there is no guarantee that
     /// the UTXO remains unspent in the best chain.
     BestChainUtxo(transparent::OutPoint),
+
+    /// Looks up a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
+    /// returning `None` immediately if it is unknown.
+    ///
+    /// Checks verified blocks in the finalized chain and _all_ non-finalized chains.
+    ///
+    /// This request is purely informational, there is no guarantee that
+    /// the UTXO remains unspent in the best chain.
+    AnyChainUtxo(transparent::OutPoint),
 
     /// Computes a block locator object based on the current best chain.
     ///
@@ -687,10 +696,9 @@ impl TryFrom<Request> for ReadRequest {
                 Err("ReadService does not write blocks")
             }
 
-            Request::AwaitUtxo(_) => {
-                Err("ReadService does not track pending UTXOs. \
-                     Manually convert the request to ReadRequest::ChainUtxo, and handle pending UTXOs.")
-            }
+            Request::AwaitUtxo(_) => Err("ReadService does not track pending UTXOs. \
+                     Manually convert the request to ReadRequest::AnyChainUtxo, \
+                     and handle pending UTXOs"),
         }
     }
 }
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -445,9 +445,11 @@ pub enum Request {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
-    /// Request a UTXO identified by the given
-    /// [`OutPoint`](transparent::OutPoint), waiting until it becomes available
-    /// if it is unknown.
+    /// Request a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
+    /// waiting until it becomes available if it is unknown.
+    ///
+    /// Checks the finalized chain, non-finalized chain, queued unverified blocks,
+    /// and any blocks that arrive at the state after the request future has been created.
     ///
     /// This request is purely informational, and there are no guarantees about
     /// whether the UTXO remains unspent or is on the best chain, or any chain.
@@ -458,6 +460,8 @@ pub enum Request {
     /// UTXO requests should be wrapped in a timeout, so that
     /// out-of-order and invalid requests do not hang indefinitely. See the [`crate`]
     /// documentation for details.
+    ///
+    /// Outdated requests are pruned on a regular basis.
     AwaitUtxo(transparent::OutPoint),
 
     /// Finds the first hash that's in the peer's `known_blocks` and the local best chain.
@@ -541,6 +545,15 @@ pub enum ReadRequest {
     /// * [`ReadResponse::Transaction(Some(Arc<Transaction>))`](ReadResponse::Transaction) if the transaction is in the best chain;
     /// * [`ReadResponse::Transaction(None)`](ReadResponse::Transaction) otherwise.
     Transaction(transaction::Hash),
+
+    /// Looks up a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
+    /// returning `None` immediately if it is unknown.
+    ///
+    /// Checks verified blocks in the finalized chain and non-finalized chain.
+    ///
+    /// This request is purely informational, there is no guarantee that
+    /// the UTXO remains unspent in the best chain.
+    ChainUtxo(transparent::OutPoint),
 
     /// Computes a block locator object based on the current best chain.
     ///
@@ -662,8 +675,6 @@ impl TryFrom<Request> for ReadRequest {
             Request::Block(hash_or_height) => Ok(ReadRequest::Block(hash_or_height)),
             Request::Transaction(tx_hash) => Ok(ReadRequest::Transaction(tx_hash)),
 
-            Request::AwaitUtxo(_) => unimplemented!("use StoredUtxo here"),
-
             Request::BlockLocator => Ok(ReadRequest::BlockLocator),
             Request::FindBlockHashes { known_blocks, stop } => {
                 Ok(ReadRequest::FindBlockHashes { known_blocks, stop })
@@ -674,6 +685,11 @@ impl TryFrom<Request> for ReadRequest {
 
             Request::CommitBlock(_) | Request::CommitFinalizedBlock(_) => {
                 Err("ReadService does not write blocks")
+            }
+
+            Request::AwaitUtxo(_) => {
+                Err("ReadService does not track pending UTXOs. \
+                     Manually convert the request to ReadRequest::ChainUtxo, and handle pending UTXOs.")
             }
         }
     }
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -448,7 +448,7 @@ pub enum Request {
     /// Request a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
     /// waiting until it becomes available if it is unknown.
     ///
-    /// Checks the finalized chain, non-finalized chain, queued unverified blocks,
+    /// Checks the finalized chain, all non-finalized chains, queued unverified blocks,
     /// and any blocks that arrive at the state after the request future has been created.
     ///
     /// This request is purely informational, and there are no guarantees about
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -553,7 +553,7 @@ pub enum ReadRequest {
     ///
     /// This request is purely informational, there is no guarantee that
     /// the UTXO remains unspent in the best chain.
-    ChainUtxo(transparent::OutPoint),
+    BestChainUtxo(transparent::OutPoint),
 
     /// Computes a block locator object based on the current best chain.
     ///
--- a/zebra-state/src/response.rs
+++ b/zebra-state/src/response.rs
@@ -76,12 +76,12 @@ pub enum ReadResponse {
     /// The response to a `FindBlockHeaders` request.
     BlockHeaders(Vec<block::CountedHeader>),
 
-    /// The response to a `ChainUtxo` request, from verified blocks in the
+    /// The response to a `BestChainUtxo` request, from verified blocks in the
     /// non-finalized chain or finalized chain.
     ///
     /// This response is purely informational, there is no guarantee that
     /// the UTXO remains unspent in the best chain.
-    ChainUtxo(Option<transparent::Utxo>),
+    BestChainUtxo(Option<transparent::Utxo>),
 
     /// Response to [`ReadRequest::SaplingTree`] with the specified Sapling note commitment tree.
     SaplingTree(Option<Arc<sapling::tree::NoteCommitmentTree>>),
--- a/zebra-state/src/response.rs
+++ b/zebra-state/src/response.rs
@@ -39,7 +39,7 @@ pub enum Response {
     /// Response to [`Request::Block`] with the specified block.
     Block(Option<Arc<Block>>),
 
-    /// The response to a `AwaitUtxo` request, from the non-finalized chain, finalized chain,
+    /// The response to a `AwaitUtxo` request, from any non-finalized chains, finalized chain,
     /// pending unverified blocks, or blocks received after the request was sent.
     Utxo(transparent::Utxo),
 
--- a/zebra-state/src/response.rs
+++ b/zebra-state/src/response.rs
@@ -89,7 +89,7 @@ pub enum ReadResponse {
     AddressesTransactionIds(BTreeMap<TransactionLocation, transaction::Hash>),
 
     /// Response to [`ReadRequest::UtxosByAddresses`] with found utxos and transaction data.
-    Utxos(AddressUtxos),
+    AddressUtxos(AddressUtxos),
 }
 
 /// Conversion from read-only [`ReadResponse`]s to read-write [`Response`]s.
@@ -117,8 +117,7 @@ impl TryFrom<ReadResponse> for Response {
 
             ReadResponse::AddressBalance(_) => unimplemented!(),
             ReadResponse::AddressesTransactionIds(_) => unimplemented!(),
-            // TODO: Rename to AddressUtxos
-            ReadResponse::Utxos(_) => unimplemented!(),
+            ReadResponse::AddressUtxos(_) => unimplemented!(),
         }
     }
 }
--- a/zebra-state/src/response.rs
+++ b/zebra-state/src/response.rs
@@ -39,7 +39,8 @@ pub enum Response {
     /// Response to [`Request::Block`] with the specified block.
     Block(Option<Arc<Block>>),
 
-    /// The response to a `AwaitUtxo` request.
+    /// The response to a `AwaitUtxo` request, from the non-finalized chain, finalized chain,
+    /// pending unverified blocks, or blocks received after the request was sent.
     Utxo(transparent::Utxo),
 
     /// The response to a `FindBlockHashes` request.
@@ -75,6 +76,13 @@ pub enum ReadResponse {
     /// The response to a `FindBlockHeaders` request.
     BlockHeaders(Vec<block::CountedHeader>),
 
+    /// The response to a `ChainUtxo` request, from verified blocks in the
+    /// non-finalized chain or finalized chain.
+    ///
+    /// This response is purely informational, there is no guarantee that
+    /// the UTXO remains unspent in the best chain.
+    ChainUtxo(Option<transparent::Utxo>),
+
     /// Response to [`ReadRequest::SaplingTree`] with the specified Sapling note commitment tree.
     SaplingTree(Option<Arc<sapling::tree::NoteCommitmentTree>>),
 
@@ -108,16 +116,20 @@ impl TryFrom<ReadResponse> for Response {
                 Ok(Response::Transaction(tx_and_height.map(|(tx, _height)| tx)))
             }
 
+            ReadResponse::ChainUtxo(_) => Err("ReadService does not track pending UTXOs. \
+                                               Manually unwrap the response, and handle pending UTXOs."),
+
             ReadResponse::BlockLocator(hashes) => Ok(Response::BlockLocator(hashes)),
             ReadResponse::BlockHashes(hashes) => Ok(Response::BlockHashes(hashes)),
             ReadResponse::BlockHeaders(headers) => Ok(Response::BlockHeaders(headers)),
 
-            ReadResponse::SaplingTree(_) => unimplemented!(),
-            ReadResponse::OrchardTree(_) => unimplemented!(),
-
-            ReadResponse::AddressBalance(_) => unimplemented!(),
-            ReadResponse::AddressesTransactionIds(_) => unimplemented!(),
-            ReadResponse::AddressUtxos(_) => unimplemented!(),
+            ReadResponse::SaplingTree(_)
+            | ReadResponse::OrchardTree(_)
+            | ReadResponse::AddressBalance(_)
+            | ReadResponse::AddressesTransactionIds(_)
+            | ReadResponse::AddressUtxos(_) => {
+                Err("there is no corresponding Response for this ReadResponse")
+            }
         }
     }
 }
--- a/zebra-state/src/response.rs
+++ b/zebra-state/src/response.rs
@@ -77,11 +77,18 @@ pub enum ReadResponse {
     BlockHeaders(Vec<block::CountedHeader>),
 
     /// The response to a `BestChainUtxo` request, from verified blocks in the
-    /// non-finalized chain or finalized chain.
+    /// _best_ non-finalized chain, or the finalized chain.
     ///
     /// This response is purely informational, there is no guarantee that
     /// the UTXO remains unspent in the best chain.
     BestChainUtxo(Option<transparent::Utxo>),
+
+    /// The response to an `AnyChainUtxo` request, from verified blocks in
+    /// _any_ non-finalized chain, or the finalized chain.
+    ///
+    /// This response is purely informational, there is no guarantee that
+    /// the UTXO remains unspent in the best chain.
+    AnyChainUtxo(Option<transparent::Utxo>),
 
     /// Response to [`ReadRequest::SaplingTree`] with the specified Sapling note commitment tree.
     SaplingTree(Option<Arc<sapling::tree::NoteCommitmentTree>>),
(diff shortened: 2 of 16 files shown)


teor2345:
@mergifyio update

teor2345:
I'm running a full sync and a checkpoint sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/3013393462

teor2345:
Weird, this built on my machine. I'll push my latest changes and see what's up.

Edit: it was some newly created files I forgot to commit.

teor2345:
I'm running a full sync and a checkpoint sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/3018034862

teor2345:
I did a full sync and an update sync locally:
- both succeeded on testnet
- update sync succeeded on mainnet
- full sync is still going on mainnet, it seems to be about the same speed, there are no warnings

zebra-state/src/service.rs
```diff
@@ -89,40 +90,55 @@ pub type QueuedFinalized = (
...
+    /// [`LatestChainTip`] and [`ChainTipChange`].
     chain_tip_sender: ChainTipSender,
 
-    /// A sender channel for the current best non-finalized chain.
+    /// A sender channel used to update the current best non-finalized chain for [`ReadStateService`].
     best_chain_sender: watch::Sender<Option<Arc<Chain>>>,
+
+    /// A cloneable [`ReadStateService`], used to answer concurrent read requests.
+    ///
+    /// TODO: move concurrent read requests to [`ReadRequest`], and remove `read_service`.
```
```suggestion
    ///
    /// TODO: move users of read [`Request`]s to [`ReadStateService`], and remove `read_service`.
```

zebra-state/src/service/read/find.rs
```diff
@@ -1,28 +1,66 @@
...
+
+/// Return the depth of block `hash` from the chain tip.
+/// Searches `chain` for `hash`, then searches `db`.
+pub fn depth<C>(chain: Option<C>, db: &ZebraDb, hash: block::Hash) -> Option<u32>
+where
+    C: AsRef<Chain>,
+{
+    let chain = chain.as_ref();
+
+    let tip = tip_height(chain, db)?;
```
teor2345, line 78: Explain why doing this concurrently is ok.

zebra-state/src/service/read/find.rs
```diff
@@ -53,7 +91,52 @@ where
...
 
-/// Find the first hash that's in the peer's `known_blocks` and the chain.
+/// Create a block locator from `chain` and `db`.
+pub fn block_locator<C>(chain: Option<C>, db: &ZebraDb) -> Option<Vec<block::Hash>>
+where
+    C: AsRef<Chain>,
+{
+    let chain = chain.as_ref();
+
+    let tip_height = tip_height(chain, db)?;
```
teor2345, line 162: Explain why doing this concurrently is ok.
Is it ok to access hashes by height if the finalized state is changing?

zebra-state/src/request.rs
```diff
@@ -441,9 +442,11 @@ pub enum Request {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
-    /// Request a UTXO identified by the given
-    /// [`OutPoint`](transparent::OutPoint), waiting until it becomes available
-    /// if it is unknown.
+    /// Request a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
+    /// waiting until it becomes available if it is unknown.
+    ///
+    /// Checks the finalized chain, non-finalized chain, queued unverified blocks,
```
```suggestion
    /// Checks the finalized chain, all non-finalized chains, queued unverified blocks,
```

zebra-state/src/response.rs
```diff
@@ -42,7 +39,8 @@ pub enum Response {
     /// Response to [`Request::Block`] with the specified block.
     Block(Option<Arc<Block>>),
 
-    /// The response to a `AwaitUtxo` request.
+    /// The response to a `AwaitUtxo` request, from the non-finalized chain, finalized chain,
```
teor2345:
```suggestion
    /// The response to a `AwaitUtxo` request, from any non-finalized chains, finalized chain,
```

zebra-state/src/service.rs
```diff
@@ -508,23 +495,21 @@ impl StateService {
...
-    pub fn any_utxo(&self, outpoint: &transparent::OutPoint) -> Option<transparent::Utxo> {
-        // We ignore any UTXOs in FinalizedState.queued_by_prev_hash,
-        // because it is only used during checkpoint verification.
+    ///
+    /// Finalized UTXOs are handled via [`ReadRequest::ChainUtxo`].
+    pub fn any_non_finalized_utxo(
+        &self,
+        outpoint: &transparent::OutPoint,
+    ) -> Option<transparent::Utxo> {
         self.mem
```
teor2345: Work out how to access all the chains without using `self.mem`

zebra-state/src/service.rs
```diff
@@ -711,27 +701,105 @@ impl Service<Request> for StateService {
...
+                if let Some(utxo) = self.any_non_finalized_utxo(&outpoint) {
+                    self.pending_utxos.respond(&outpoint, utxo);
+
+                    // We're finished, the returned future gets the UTXO from the respond() channel.
+                    timer.finish(module_path!(), line!(), "AwaitUtxo/any-non-finalized");
+
+                    return response_fut;
+                }
+
+                // We ignore any UTXOs in FinalizedState.queued_by_prev_hash,
```
teor2345, line 712: Open a ticket for this?

zebra-state/src/service/read/find.rs
teor2345: Added a comment in 47011e1a52

zebra-state/src/service/read/find.rs
teor2345: Added a comment in 47011e1a52

zebra-state/src/service.rs
teor2345: I opened #5126.

zebra-state/src/service.rs
teor2345: Fixed in ecb0259246 by sending the whole `NonFinalizedState` to the `ReadStateService`.

arya2 (dismissed):
The logic in `Tip`, `Depth`, `BlockLocator`, `FindBlockHashes`, and `FindBlockHeaders` requests of `ReadStateService` looks to be the same, and the only difference I see in the `AwaitUtxo` request of `StateService` is that it iterates through `queued_blocks` before checking `any_utxo()`, but that shouldn't cause issues.

The `TryFrom` impls for `Request` and `ReadResponse` look good.

The new dead code and the unused `ChainUtxo` request in `ReadStateService` are better than potentially forgetting we had those later and re-writing similar logic.

zebra-state/src/service.rs
```diff
@@ -711,27 +682,102 @@ impl Service<Request> for StateService {
                 .instrument(span)
                 .boxed()
             }
-            Request::Depth(hash) => {
+
+            // Uses pending_utxos and queued_blocks in the StateService.
+            // Directly uses the shared NonFinalizedState in the ReadStateService.
+            Request::AwaitUtxo(outpoint) => {
+                metrics::counter!(
```
arya2, line 689: it would be nice if some of these metrics counters were moved up to cut down on repetition

zebra-state/src/service.rs
```diff
@@ -759,271 +811,414 @@ impl Service<Request> for StateService {
...
+    type Error = BoxError;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    #[instrument(name = "read_state", skip(self))]
+    fn call(&mut self, req: ReadRequest) -> Self::Future {
```
arya2: would be nice to delete the `metric::counter!()` calls in the arm below to shorten the code a bit here too

```suggestion
    fn call(&mut self, req: ReadRequest) -> Self::Future {
        metrics::counter!(
            "state.requests",
            1,
            "service" => "read_state",
            "type" => match req {
                ReadRequest::Depth(_) => "depth",
                ReadRequest::Tip => "tip",
                ReadRequest::BlockLocator => "block_locator",
                ReadRequest::Transaction(_) => "transaction",
                ReadRequest::ChainUtxo { .. } => "chain_utxo",
                ReadRequest::Block(_) => "block",
                ReadRequest::FindBlockHashes { .. } => "find_block_hashes",
                ReadRequest::FindBlockHeaders { .. } => "find_block_headers",
                ReadRequest::SaplingTree { .. } => "sapling_tree",
                ReadRequest::OrchardTree { .. } => "orchard_tree",
                ReadRequest::AddressBalance { .. } => "address_balance",
                ReadRequest::TransactionIdsByAddresses { .. } => "transaction_ids_by_addesses",
                ReadRequest::UtxosByAddresses(_) => "utxos_by_addesses"
            }
        );
```

zebra-state/src/service.rs
```diff
@@ -610,6 +576,8 @@ impl Service<Request> for StateService {
     #[instrument(name = "state", skip(self, req))]
     fn call(&mut self, req: Request) -> Self::Future {
```
arya2: Not sure if the Requests that are now redirected to the ReadStateService should be counted as both state service and read_state, but might be nicer to have it in a `.count_metric()` method than a `.name()` method

```suggestion
        metrics::counter!(
            "state.requests",
            1,
            "service" => "state",
            "type" => match req {
                Request::Depth(_) => "depth",
                Request::Tip => "tip",
                Request::BlockLocator => "block_locator",
                Request::Transaction(_) => "transaction",
                Request::Block(_) => "block",
                Request::FindBlockHashes { .. } => "find_block_hashes",
                Request::FindBlockHeaders { .. } => "find_block_headers",
                Request::CommitBlock(_) => "commit_block",
                Request::CommitFinalizedBlock(_) => "commit_finalized_block",
                Request::AwaitUtxo(_) => "await_utxo",
            }
        );

```

zebra-state/src/service.rs
```diff
@@ -711,27 +682,102 @@ impl Service<Request> for StateService {
...
+                }
+
+                // We're finished, but the returned future is waiting on the respond() channel.
+                timer.finish(module_path!(), line!(), "AwaitUtxo/waiting");
+
+                response_fut
+            }
+
+            // TODO: add a name() method to Request, and combine all the generic read requests
+            //
```
arya2: Following the suggestion to add the metrics::counter at the top of `.call()`

```suggestion
            Request::Depth(_)
            | Request::Tip
            | Request::BlockLocator
            | Request::Transaction(_)
            | Request::Block(_)
            | Request::FindBlockHashes { .. }
            | Request::FindBlockHeaders { .. } => {
                // Redirect the request to the concurrent ReadStateService
                let read_service = self.read_service.clone();
                async move {
                    let req = req
                        .try_into()
                        .expect("ReadRequest conversion should not fail");
                    let rsp = read_service.oneshot(req).await?;
                    let rsp = rsp.try_into().expect("Response conversion should not fail");
                    Ok(rsp)
                }
                .boxed()
            }

```

zebra-state/src/service.rs
teor2345: Would you mind pushing a separate PR with these changes?
They are a good cleanup, but they don't block the rest of the work in #4937, so we can merge them separately.

zebra-state/src/service.rs
teor2345: These changes can also go in that separate PR.

zebra-state/src/service.rs
teor2345: These changes can also go in that separate PR.

teor2345:
The previous full sync was about 10% slower during full verification.

I'm running a full sync again with the latest code:
https://github.com/ZcashFoundation/zebra/actions/runs/3025949651

I have some ideas to fix that, but I think they should go in a separate PR.
We might also see a significant speedup once we implement a separate task for block commits.

So I might need to temporarily increase the full sync timeout in this PR, just to make `main` pass while we're doing these other changes.

zebra-state/src/service/read/block.rs
```diff
@@ -88,3 +92,48 @@ where
         })
         .or_else(|| db.transaction(hash))
 }
+
+/// Returns the [`Utxo`] for [`transparent::OutPoint`], if it exists in the
+/// non-finalized `chain` or finalized `db`.
+///
+/// UTXOs may be returned regardless of whether they have been spent.
```
upbqdn: Consistency note: other comments say "are".
```suggestion
/// UTXOs are returned regardless of whether they have been spent.
```

zebra-state/src/service/read/block.rs
```diff
@@ -88,3 +92,48 @@ where
...
+    // memory, but `db` stores transactions on disk, with a memory cache.)
+    chain
+        .and_then(|chain| chain.as_ref().created_utxo(&outpoint))
+        .or_else(|| db.utxo(&outpoint).map(|utxo| utxo.utxo))
+}
+
+/// Returns the [`Utxo`] for [`transparent::OutPoint`], if it exists in the
+/// `non_finalized_state` or finalized `db`.
+///
+/// UTXOs may be returned regardless of whether they have been spent.
```
upbqdn: Consistency note: other comments say "are".
```suggestion
/// UTXOs are returned regardless of whether they have been spent.
```

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -340,14 +340,13 @@ impl NonFinalizedState {
...
-            if let Some(utxo) = chain.created_utxos.get(outpoint) {
-                return Some(utxo.utxo.clone());
-            }
-        }
-
-        None
+        self.chain_set
+            .iter()
+            .rev()
+            .find_map(|chain| chain.created_utxo(outpoint))
```
upbqdn, line 349: The doc comment for this function says it returns
> the [`transparent::Utxo`] pointed to by the given [`transparent::OutPoint`]

However, it returns such a UTXO from the first chain when there are multiple parallel non-finalized chains containing the same UTXO.

I checked whether this affects anything, and it doesn't seem so.

teor2345:
> The previous full sync was about 10% slower during full verification.

Actually, it seems like full syncs have been failing on `main` for a while now, so I pushed some PRs to fix that.

zebra-state/src/service/read/block.rs
```diff
@@ -88,3 +92,48 @@ where
...
+pub fn utxo<C>(chain: Option<C>, db: &ZebraDb, outpoint: transparent::OutPoint) -> Option<Utxo>
+where
+    C: AsRef<Chain>,
+{
+    // # Correctness
+    //
+    // The StateService commits blocks to the finalized state before updating
+    // the latest chain, and it can commit additional blocks after we've cloned
```
upbqdn: I'm probably missing something, but it's not clear to me what `latest chain` refers to.

zebra-state/src/service/read/find.rs
```diff
@@ -1,59 +1,213 @@
...
 where
     C: AsRef<Chain>,
 {
+    // # Correctness
+    //
+    // The StateService commits blocks to the finalized state before updating
+    // the latest chain, and it can commit additional blocks after we've cloned
+    // this `chain` variable.
+    //
+    // Finalized state updates can only add overlapping blocks, and heights are unique
```
upbqdn, line 105: I'm probably missing something, but it's not clear to me what `overlapping blocks` refers to.

zebra-state/src/service/read/find.rs
```diff
@@ -1,59 +1,213 @@
...
 where
     C: AsRef<Chain>,
 {
+    // # Correctness
+    //
+    // The StateService commits blocks to the finalized state before updating
+    // the latest chain, and it can commit additional blocks after we've cloned
+    // this `chain` variable.
+    //
+    // Finalized state updates can only add overlapping blocks, and hashes are unique.
```
upbqdn, line 125: I'm probably missing something, but it's not clear to me what `overlapping blocks` refers to.

zebra-state/src/service/read/find.rs
```diff
@@ -1,59 +1,213 @@
...
```
upbqdn, line 91: I'm probably missing something, but it's not clear to me what `overlapping blocks` refers to.

zebra-state/src/service/read/find.rs
```diff
@@ -1,59 +1,213 @@
...
+    let chain = chain.as_ref();
+
+    // # Correctness
+    //
+    // The StateService commits blocks to the finalized state before updating
+    // the latest chain, and it can commit additional blocks after we've cloned
+    // this `chain` variable.
+    //
+    // It is ok to do this lookup in two different calls. Finalized state updates
+    // can only add overlapping blocks, and hashes are unique.
```
upbqdn, line 76: I'm probably missing something, but it's not clear to me what `overlapping blocks` refers to.

zebra-state/src/service/read/find.rs
```diff
@@ -1,59 +1,213 @@
...
+    let chain = chain.as_ref();
+
+    // # Correctness
+    //
+    // The StateService commits blocks to the finalized state before updating
+    // the latest chain, and it can commit additional blocks after we've cloned
+    // this `chain` variable.
+    //
+    // It is ok to do these lookups using multiple database calls. Finalized state updates
+    // can only add overlapping blocks, and hashes are unique.
```
upbqdn, line 152: I'm probably missing something, but it's not clear to me what `overlapping blocks` refers to.

teor2345:
I split PR #5132 out of this PR, because it seems like we've finished work on most of the state requests, so they can merge. But we're still working on the `AwaitUtxos` request.

upbqdn (commented):
Looks great so far. I need a bit more time to finish reviewing the `AwaitUtxo` requests.

zebra-state/src/service.rs
teor2345: > Not sure if the Requests that are now redirected to the ReadStateService should be counted as both state service and read_state

I think it's ok for now, because some requests (`Block` and `Transaction`) are used via both `StateService` and `ReadStateService` directly. So their metrics will be different across the services.

Eventually I'd like to use `ReadStateService` directly, get rid of the read requests in the `StateService`, and then the corresponding metrics will go as well. But that's something for a future cleanup.

zebra-state/src/service/non_finalized_state.rs
teor2345: The `OutPoint` contains the transaction hash, which includes all the content of the `Utxo`. So there is only ever one `Utxo` per `OutPoint`, even across multiple chains.

zebra-state/src/service/read/block.rs
teor2345: This was actually a deliberate difference, but you're right - it's not clear what it means.

I explained what actually happens in commit 1d4668432a

zebra-state/src/service/read/block.rs
teor2345: I explained what actually happens in commit 1d4668432a

zebra-state/src/service/read/block.rs
teor2345: It's the cached `Chain` in the `chain` variable passed to this function.
Hopefully the doc changes in commit 8015f51c48 make this a bit clearer.

zebra-state/src/service/read/find.rs
teor2345: If the request is delayed, the cached `chain` can contain some of the same blocks as the shared `db` reference.
Hopefully the doc changes in commit 8015f51c48 make this a bit clearer.

zebra-state/src/service/read/find.rs
teor2345: See commit 8015f51c48.

zebra-state/src/service/read/find.rs
teor2345: See commit 8015f51c48.

zebra-state/src/service/read/find.rs
teor2345: See commit 8015f51c48.

zebra-state/src/service/read/find.rs
teor2345: See commit 8015f51c48.

teor2345:
I made `AwaitUtxo` non-blocking, that should be enough to get us started with block commit channels.

zebra-state/src/service/non_finalized_state.rs
upbqdn: I meant the scenario when there is the same transaction in two parallel chains, for example when two separate miners produce two separate blocks with some transactions being the same. Does your claim hold in this scenario as well?

upbqdn: But as I said, I don't think this scenario is important because it doesn't affect Zebra's behavior. I just wanted to understand what actually happens.

arya2 (dismissed):
Looks good, I added some cleanup in [minor cleanup of zebra-state/service.rs](https://github.com/ZcashFoundation/zebra/pull/5137) for later

zebra-state/src/service/non_finalized_state.rs
teor2345: > I meant the scenario when there is the same transaction in two parallel chains, for example when two separate miners produce two separate blocks with some transactions being the same. Does your claim hold in this scenario as well?

In this case:
- if the transactions are the same, the `OutPoint`s are the same, and they return the same `Utxo`
- if the transactions are different, the transaction hash or index in the `OutPoint`s are different, even if the `Utxo`s have the same value and spend script

https://github.com/ZcashFoundation/zebra/blob/97fb85dca9d00a717661ea0cbab7ce66b8d23977/zebra-chain/src/transparent.rs#L99-L113

teor2345:
> Run gcloud compute instances create-with-container "update-to-tip-5107-merge-7ae4b28" \
>  - The resource 'projects/zealous-zebra/global/images/zebrad-cache-5143-merge-fb82d20-v25-mainnet-tip-u-013716' is not ready

https://github.com/ZcashFoundation/zebra/actions/runs/3041636484/jobs/4899242654

This will be fixed by #5044

teor2345:
@mergifyio update

zebra-state/src/service/non_finalized_state.rs
upbqdn: Got it, thank you for taking the time to explain it.

teor2345:
I'm running a full sync here, but we don't need to wait for it before we merge this PR:
https://github.com/ZcashFoundation/zebra/actions/runs/3047965824

zebra-state/src/service/read/block.rs
```diff
@@ -1,14 +1,29 @@
 //! Shared block, header, and transaction reading code.
+//!
+//! In the functions in this module:
+//!
+//! The StateService commits blocks to the finalized state before updating
+//! `chain` or `non_finalized_state` from the latest chains. Then it can
```
upbqdn, line 6: I still have trouble with `latest chains`.

Does it refer to the most recently updated chain in the `chain_set` in the `NonFinalizedState`?

zebra-state/src/service.rs
```diff
@@ -170,11 +164,11 @@ pub struct ReadStateService {
     /// so it might include some block data that is also in `best_mem`.
     db: ZebraDb,
 
-    /// A watch channel for the current best in-memory chain.
+    /// A watch channel for a recent [`NonFinalizedState`].
```
upbqdn, line 167: Maybe we could remove `recent` since there are no other `NonFinalizedState`s?

It made me think if there are any non-recent `NonFinalizedState`s.

zebra-state/src/service.rs
```diff
@@ -704,7 +679,80 @@ impl Service<Request> for StateService {
...
+                // Prepare the AwaitUtxo future from PendingUxtos.
+                let response_fut = self.pending_utxos.queue(outpoint);
+                // Only instrument `response_fut`, the ReadStateService already
+                // instruments its requests with the same span.
+                let span = Span::current();
+                let response_fut = response_fut.instrument(span).boxed();
+
+                // Check the non-finalized block queue outside the returned future,
+                // so we can access mutable state fields.
+                if let Some(utxo) = self.queued_blocks.utxo(&outpoint) {
```
upbqdn, line 703: I wonder if this actually returns `Some` in practice.

The reason is that I'd expect almost all UTXOs to be at least a few blocks apart.

upbqdn (approved):
All looks good. I left some minor comments.

zebra-state/src/service/read/block.rs
teor2345: I'll make this change to all these files in my editor.
At the same time, I'll change `StateService` to "block commit task", because that's what it will be after this PR series.

```suggestion
//! `chain` or `non_finalized_state` with a cached copy of the non-finalized chains in `NonFinalizedState.chain_set`. Then the block commit task can
```

zebra-state/src/service.rs
teor2345:
```suggestion
    /// A watch channel with a cached copy of the [`NonFinalizedState`].
```

zebra-state/src/service.rs
teor2345: Here's one way it could return `Some`:
1. The syncer sends 20 blocks for full verification in parallel
2. The first block takes a while to verify
3. The fifth block gets verified quickly, because it doesn't depend on anything from the first four blocks, then it gets put in the queue, waiting for the first four blocks
4. The tenth block spends a UTXO from the fifth block, which is in the queue, so `AwaitUtxo` returns `Some` here

Feel free to add a debug log and see how often that happens!

zebra-state/src/service.rs
teor2345: Updated in PR #5134 in commit 22ef6561f8

zebra-state/src/service/read/block.rs
teor2345: Updated in PR #5134 in commit 22ef6561f8

zebra-state/src/service.rs
upbqdn: Makes sense!

PR #5112, opened 2022-09-07
tests(config): Upgrade config file tests 

### LINKED ISSUE #4684: Test that every old zebrad.toml config still works
## Scheduling

This test is required for the  _next_ time Zebra changes its config format.
(Last changes were in July 2022, June 2022, then in June 2021.)

## Motivation

There are some ways to break a config that aren't covered by ticket #4625.

At the moment, if we add a new field, we won't be testing that new field with every newer `zebrad` versions.

So we need to keep a copy of every generated config change, and test them all.
(We can delete old configs when we do a breaking release, like at a network upgrade.)

## Designs

We might want something like:
- `stored_config_works` checks that every older version of the config works with the current `zebrad`
- a new `stored_configs_update_newest` test, that makes sure the new config has been added to the config directory, so that its new fields get checked with future versions of `zebrad`

### Test Coverage

Zebra has had production users since the beta.12 release, so we need to test the configs generated by that release, and every release after it.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/4676#issuecomment-1164983054_

oxarbitrage:
## Motivation

Currently zebra has only one old config file stored to do compatibility checks. We want to have more config versions and test against them. Closes https://github.com/ZcashFoundation/zebra/issues/4684 

## Solution

Add a folder with different config versions and make sure zebra can start with any of them. In this PR i am including configs for v1.0.0-beta.12, v1.0.0-beta.12 and v1.0.0-beta.14. Before that (at v1.0.0-beta.11) the config don't work. This is because we changed the type of time fields to human readable form. It will not worth to remain compatible with that now but we might want to remain compatible in the future if we need to implement field type changes

## Review

Anyone can review.

### Reviewer Checklist

  - [ ] New test is good enough and pass.

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -511,6 +511,9 @@ fn config_tests() -> Result<()> {
     // Check what happens when Zebra parses an invalid config
     invalid_generated_config()?;
 
+    // Check that we have a current version of the config stored
+    last_config_is_stored()?;
+
     // Check that Zebra stored configuration works
     stored_configs_works()?;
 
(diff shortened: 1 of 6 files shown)


zebrad/tests/common/configs/v1.0.0-beta.14.toml
```diff
@@ -0,0 +1,66 @@
+# Default configuration for zebrad.
```
teor2345: Is there any difference between `v1.0.0-beta.13.toml` and `v1.0.0-beta.14.toml`?
We only need to test files if they are different.

teor2345 (changes requested):
This looks like a good test!

How will we remember to add a new config when we add new config options?

Here is one possible way to do it:
- Add a new test that runs `zebrad generate`, and checks that it matches at least one file in `configs_dir()`
- This test will fail at the moment, because we just merged a PR that adds a config option (after the last release).

It's ok to add the latest config with the name of the next release, because if it needs to be updated, the new test we just added will fail.

zebrad/tests/acceptance.rs
```diff
@@ -597,6 +602,59 @@ fn valid_generated_config(command: &str, expect_stdout_line_contains: &str) -> R
...
+        // Read stored config
+        let stored_content = fs::read_to_string(config_file_full_path(config_file.path()))
+            .expect("Should have been able to read the file");
+
+        // If any stored config is equal to the generated then we are good.
+        if stored_content.eq(&processed_generated_content) {
+            return Ok(());
+        }
+    }
+    Err(eyre!("last config is not stored"))
```
teor2345: I'll just add some instructions for how to fix the test failure:
```suggestion
    Err(eyre!(
        "latest zebrad config is not being tested for compatibility.\n\
        Run `zebrad generate -o zebrad/tests/common/configs/<next-release-tag>.toml`\n\
        and commit the latest config to Zebra's git repository"
    ))
```

teor2345 (commented):
Thanks, this is really good!

I ran the test without the latest config, and it failed, as we expected.
I added some instructions to the error in this PR, telling developers how to fix the test failure.

teor2345 (approved):

PR #5113, opened 2022-09-07
change(ci): Delete cached state images older than 2 days, and check every day

teor2345:
## Motivation

We reduced costs by about 1/3 by deleting old images, but we can do better than that by deleting more old unused images.

## Solution

- Delete cached state images older than 2 days, keeping 2 images of each type
- Run deletion every day

## Review

This is something we should do soon, but it's not critical.

### Reviewer Checklist

  - [x] Changes make sense

### DIFF
--- a/.github/workflows/delete-gcp-resources.yml
+++ b/.github/workflows/delete-gcp-resources.yml
@@ -1,18 +1,19 @@
 name: Delete GCP resources
 
 on:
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):

gustavovalverde:
Thank you very much for improving this as expected @teor2345 🎉

PR #5124, opened 2022-09-08
fix(doc): Explain how image deletion actually works

teor2345:
## Motivation

Image deletion keeps slightly more images than we want, which is ok for now, but we should document it.

## Review

Anyone can review this PR, we just need to fix the docs eventually.

### Reviewer Checklist

  - [ ] Docs make sense

### DIFF
--- a/.github/workflows/delete-gcp-resources.yml
+++ b/.github/workflows/delete-gcp-resources.yml
@@ -73,13 +73,14 @@ jobs:
 
       # Deletes cache images older than $DELETE_AGE_DAYS days.
       #
-      # Keeps an extra $KEEP_LATEST_IMAGE_COUNT images of each type:
+      # Keeps all images younger than $DELETE_AGE_DAYS.
+      # Also keeps $KEEP_LATEST_IMAGE_COUNT older images of each type:
       # - zebrad checkpoint cache
       # - zebrad tip cache
       # - lightwalletd + zebrad tip cache
       #
       # TODO:
-      # - keep the latest $KEEP_LATEST_IMAGE_COUNT, if there are that many recent images, delete all the outdated images
+      # - keep the latest $KEEP_LATEST_IMAGE_COUNT, if there are at least that many recent images, delete all the outdated images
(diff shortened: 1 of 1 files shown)


.github/workflows/delete-gcp-resources.yml
```diff
@@ -73,13 +73,15 @@ jobs:
 
       # Deletes cache images older than $DELETE_AGE_DAYS days.
       #
-      # Keeps the latest $KEEP_LATEST_IMAGE_COUNT images of each type:
+      # Keeps an extra $KEEP_LATEST_IMAGE_COUNT images of each type:
```
arya2:
```suggestion
      # Keeps an extra $KEEP_LATEST_IMAGE_COUNT of the most recent images that are older than $DELETE_AGE_DAYS of each type:
```

arya2 (commented):
Makes sense

.github/workflows/delete-gcp-resources.yml
```diff
@@ -73,13 +73,15 @@ jobs:
...
       # - zebrad checkpoint cache
       # - zebrad tip cache
       # - lightwalletd + zebrad tip cache
       #
-      # TODO: when we add testnet to the workflows, keep the latest $KEEP_LATEST_IMAGE_COUNT testnet images, 
-      #       and the latest $KEEP_LATEST_IMAGE_COUNT mainnet images.
+      # TODO:
+      # - keep the latest $KEEP_LATEST_IMAGE_COUNT, if there are that many recent images, delete all the outdated images
```
teor2345:
```suggestion
      # - keep the latest $KEEP_LATEST_IMAGE_COUNT, if there are at least that many recent images, delete all the outdated images
```

.github/workflows/delete-gcp-resources.yml
teor2345: That's helpful, but maybe we could spell out exactly what happens:
```suggestion
      # Keeps all images younger than $DELETE_AGE_DAYS.
      # Also keeps $KEEP_LATEST_IMAGE_COUNT older images of each type:
```

gustavovalverde (approved):

PR #5130, opened 2022-09-11
fix(ci): Update Zebra checkpoints

teor2345:
## Motivation

Full syncs are currently timing out on the `main` branch.

## Solution

- Update the mainnet and testnet checkpoints

## Review

This is urgent because CI is failing.
Anyone can review this PR.

### Reviewer Checklist

  - [ ] CI passes

PR #5129, opened 2022-09-11
fix(ci): Increase the full sync timeout

teor2345:
## Motivation

Full syncs are currently failing in CI at about 99.6% with a full sync timeout error.

## Solution

- Increase the timeout from 24 to 28 hours
- Make a constant for the timeout
- Update the developer documentation

## Review

This is urgent because full syncs are failing on the `main` branch.
Anyone can review this PR.

### Reviewer Checklist

  - [ ] CI passes after this PR is merged to `main`

PR #5132, opened 2022-09-11
1. change(state): Run most StateService read requests without shared mutable chain state

teor2345:
## Motivation

This PR runs most `StateService` read requests without accessing shared mutable chain state.
As a useful side-effect, this makes most `StateService` read requests concurrent.

Part of #5102.
Depends-On: #5129, #5130

### Designs

## Solution

Don't access shared mutable chain state in `StateService` read requests:
- Add a `ReadStateService` to the `StateService`
- Redirect `Block` and `Transaction` requests to the `ReadStateService`
- Re-Implement `Tip`, `Depth`, `BlockLocator`, `FindBlockHashes`, and `FindBlockHeaders` requests concurrently using the `ReadStateService`

Documentation:
- Add TODOs for splitting various state functions as part of #4937
- Document how state requests use shared state
- Turn a repeated comment into documentation
- Tidy doc links

Related Cleanups:
- Put `AddressBalance` in a consistent enum order
- Remove some redundant timers
- Move `BlockLocator` tests

## Testing

We've run a full sync on PR #5107, it was not significantly slower than the `main` branch. (Maybe about 10% slower for full verification.)

We might want to run another full sync after the upcoming timeout and checkpoint PRs merge, but they are not blockers.

## Review

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

Do the same thing with AwaitUtxo requests.
Actually create a separate task to commit blocks to the state.


upbqdn:
@oxarbitrage I have a lot of context for this PR, so I can do the review.

zebra-state/src/request.rs
```diff
@@ -522,15 +535,66 @@ pub enum ReadRequest {
...
+
+    /// Finds the first hash that's in the peer's `known_blocks` and the local best chain.
+    /// Returns a list of hashes that follow that intersection, from the best chain.
+    ///
+    /// If there is no matching hash in the best chain, starts from the genesis hash.
+    ///
+    /// Stops the list of hashes after:
+    ///   * adding the best tip,
+    ///   * adding the `stop` hash to the list, if it is in the best chain, or
+    ///   * adding 500 hashes to the list.
```
upbqdn:
```suggestion
    ///   * adding [`MAX_FIND_BLOCK_HASHES_RESULTS`] hashes to the list.
```

zebra-state/src/service.rs
```diff
@@ -89,40 +90,55 @@ pub type QueuedFinalized = (
 ///                        Zebra stores the single best chain in the finalized state,
 ///                        and re-loads it from disk when restarted.
 ///
-/// Requests to this service are processed in series,
-/// so read requests wait for all queued write requests to complete,
-/// then return their answers.
+/// Read requests to this service are processed concurrently.
```
upbqdn: I find it slightly confusing that this line says that the read requests are processed concurrently, but lines 96 and 97 say that the requests nevertheless share a queue with block write requests.

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -366,6 +366,7 @@ impl NonFinalizedState {
     }
 
     /// Returns the hash for a given `block::Height` if it is present in the best chain.
+    #[allow(dead_code)]
```
upbqdn, line 369: Why don't we delete these?

zebra-state/src/service/read/find.rs
```diff
@@ -53,7 +91,52 @@ where
...
+    for height in heights {
+        if let Some(hash) = hash_by_height(chain, db, height) {
+            hashes.push(hash);
+        }
+    }
+
+    Some(hashes)
+}
+
+/// Get the heights of the blocks for constructing a block_locator list.
```
upbqdn, line 119: The following suggestion is unrelated to the PR, but I think it's worth having it in the code since it took me a while to figure out how the function assembles the heights.

```suggestion
/// Gets the heights of the blocks for constructing a block_locator list.
///
/// The initial height in the returned `vec` is the tip height. The initial distance
/// between heights is 1. This distance increases exponentially with base 2 
/// between each subsequent height. The distance between the first and last
/// height is at most [`constants::MAX_BLOCK_REORG_HEIGHT`], and the
/// number of returned heights is approximately `log_2` of this value.
```

zebra-state/src/request.rs
upbqdn:
```suggestion
    ///   * adding [`MAX_FIND_BLOCK_HASHES_RESULTS`] hashes to the list.
```

upbqdn (dismissed):
I noticed a few nits. Otherwise, it looks great.

zebra-state/src/request.rs
upbqdn: GitHub is weird today. This suggestion is here twice, so I'm closing this one, and keeping the other one open.

zebra-state/src/service/read/find.rs
teor2345: Thanks! Some of these things are implementation details, so I might put them inside the function in comments. But the rest I'll put in the function docs.

teor2345: What do you think of my changes in f1f2230f9b ?

zebra-state/src/service/non_finalized_state.rs
teor2345: We might use them later when we change the code, and they can be a bit tricky to implement correctly.

They cost us a tiny bit of compilation and maintenance time, but they get deleted during linking by dead code elimination.

zebra-state/src/service.rs
teor2345: It's a bit confusing, I changed it to "tower::Buffer" and "ordered queue", hopefully that makes it clearer.
See b110937375.

zebra-state/src/service/read/find.rs
```diff
@@ -53,7 +91,71 @@ where
...
+///
+/// Zebra uses an exponentially decreasing list of block heights, starting at the tip.
+/// See [`block_locator()`] for details.
+pub fn block_locator_heights(tip_height: block::Height) -> Vec<block::Height> {
+    // The initial height in the returned `vec` is the tip height,
+    // and the final height is `MAX_BLOCK_REORG_HEIGHT` below the tip.
+    //
+    // The initial distance between heights is 1. This distance increases
+    // exponentially with base 2 between each subsequent height.
```
upbqdn: I even found a simpler way to put it.
```suggestion
    // The initial distance between heights is 1, and it doubles
    //  between each subsequent height.
```

zebra-state/src/service/read/find.rs
upbqdn: It's probably too much nit-picking at this point, but I think the expression _"exponentially decreasing"_ is not the right description in this case.

An exponentially decreasing quantity decreases fast when it's large and slow when it's low. I'd depict such a decrease by plotting the function e^-x:
image


However, in our case, we have the "inverse" type of decreasing --- slow when the value is large, and fast as the value gets lower. I'd depict such a decrease by plotting the function -e^x:
image
 I was trying to find a name for this type of decrease, but nothing came up. I would still rather avoid the expression _"exponentially decreasing"_ here.

zebra-state/src/service.rs
upbqdn: I think it's very clear now.

zebra-state/src/service/read/find.rs
upbqdn: It's probably too much nit-picking at this point, but I think the expression _"exponentially decreasing"_ is not the right description in this case.

An exponentially decreasing quantity decreases fast when it's large and slow when it's low. I'd depict such a decrease by plotting the function e^-x:
image


However, in our case, we have the "inverse" type of decreasing --- slow when the value is large, and fast as the value gets lower. I'd depict such a decrease by plotting the function -e^x:
image
 I was trying to find a name for this type of decrease, but nothing came up. I would still rather avoid the expression _"exponentially decreasing"_ here.

teor2345: I fixed this in 9efa75640.

Since this change isn't blocking, and isn't related to this PR, let's open another PR for any further changes?

zebra-state/src/service/read/find.rs
teor2345: I fixed this in 9efa75640.

Since this change isn't blocking, and isn't related to this PR, let's open another PR for any further changes?

upbqdn: Definitely.

zebra-state/src/service/read/find.rs
upbqdn: Definitely. Sorry for the mess.

It looks like this type of decrease is not particularly important for humanity, so there was no need to coin a term for it and rigorously define it. Maybe it's a good opportunity for us now? Inverse exponential decay?

upbqdn (approved):

Issue #5069, opened 2022-09-05
Fix "Connection closed by remote host" error on Google Cloud ssh


teor2345:
## Motivation

A few different PRs seem to be failing with this error:
> kex_exchange_identification: Connection closed by remote host
https://github.com/ZcashFoundation/zebra/runs/8161163434?check_suite_focus=true#step:7:86

teor2345:
I haven't seen these errors recently, I wonder if they are caused by rate-limiting?

We could add the `ConnectionAttempts` option to the SSH command-line to fix this.

teor2345:
This happened again today:
https://github.com/ZcashFoundation/zebra/actions/runs/3033472434/jobs/4881881486#step:8:106

teor2345:
Failed on `main`:
https://github.com/ZcashFoundation/zebra/actions/runs/3053855581/jobs/4926083103#step:6:105

teor2345:
It looks like this is now more frequent than the other Google Cloud errors.

teor2345:
Failed 3 times on `main` in the same workflow:
https://github.com/ZcashFoundation/zebra/actions/runs/3150972503/jobs/5131855566
https://github.com/ZcashFoundation/zebra/actions/runs/3150972503/jobs/5131861932
https://github.com/ZcashFoundation/zebra/actions/runs/3150972503/jobs/5131846881

gustavovalverde:
The `main` branch has been failing recently with this errors. So I'm reopening this one

gustavovalverde:
I've been able to find this error log in GCP

```
Supplied fingerprint does not match current metadata fingerprint.
```

I've also found the root cause: some `ssh-keys` in the project metadata that might have been added when testing OS Login.

This might not solve the problem, but will at least reduce the noise when searching for the actual error.

teor2345:
> I've also found the root cause: some `ssh-keys` in the project metadata that might have been added when testing OS Login.

Thanks for tracking this down!

PR #5134, opened 2022-09-12
change(state): Write finalized blocks to the state in a separate thread, to avoid network and RPC hangs

### LINKED ISSUE #5125: Avoid temporary failures verifying the first non-finalized block
## Motivation

But it has two bugs: 

### Status

## Tasks

teor2345:
## Motivation

This PR starts moving state block write requests into their own task.
When it is finished, it will make state block writes fully concurrent.

Part of #4937.
Closes #5125.

### Designs

## Solution

Add a block write task:
- Commit finalized blocks to the state in a separate task
- Add last_block_hash_sent to the state service, to avoid database accesses when draining the queue
- Add task handles to check for task exits and panics

Add block write channels:
- Finalized state
- Non-finalized state (not used in this PR)
- Close the finalized block channel when we're finished with it
- Remove a redundant `block_in_place()`

Deal with errors & shutdowns:
- Update last_block_hash_sent regardless of commit errors
- Check for panics in the block write task
- Close state channels and tasks on drop
- Drop channels and check for closed channels in the block commit task
- Work around a RocksDB shutdown bug in the tests, by removing previous code that cancelled RocksDB background work on shutdown
- Drop any leftover queued blocks when finished with the finalized state, and return an error on their channels
- Return an error on the channels of replaced duplicate blocks

Related fixes:
- Wait for the block commit task in tests, and check for errors
- Wait for chain tip updates in the RPC tests
- Improve RPC error logging
- When running a proptest that sleeps a lot, only run it once
- Explain finalized block duplicate handling and consensus rules

Related cleanups:
- Remove some duplicate fields across StateService and ReadStateService
- Rename a field to StateService.max_queued_finalized_height

## Testing

We should run a full sync on this before we merge, and make sure it is not significantly slower than the `main` branch.

## Review

### Reviewer Checklist

  - [ ] Tests for Errors

## Follow Up Work

Handle non-finalized blocks.
Handle task exits and panics correctly.
Testing.

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -809,31 +809,9 @@ impl Service<Request> for StateService {
     type Future =
         Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
 
-    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         // Check for panics in the block write task
-        //
-        // TODO: turn this into a method on the ReadStateService
-        let block_write_task = self.read_service.block_write_task.take();
-
-        if let Some(block_write_task) = block_write_task {
-            if block_write_task.is_finished() {
-                match Arc::try_unwrap(block_write_task) {
-                    // We are the last state with a reference to this task, so we can propagate any panics
-                    Ok(block_write_task_handle) => {
-                        if let Err(thread_panic) = block_write_task_handle.join() {
-                            std::panic::resume_unwind(thread_panic);
-                        }
-                    }
-                    // We're not the last state, so we need to put it back
-                    Err(arc_block_write_task) => {
-                        self.read_service.block_write_task = Some(arc_block_write_task)
-                    }
-                }
-            } else {
-                // It hasn't finished, so we need to put it back
-                self.read_service.block_write_task = Some(block_write_task);
-            }
-        }
+        let poll = self.read_service.poll_ready(cx);
 
         // Prune outdated UTXO requests
         let now = Instant::now();
@@ -862,7 +840,7 @@ impl Service<Request> for StateService {
             }
         }
 
-        Poll::Ready(Ok(()))
+        poll
     }
 
     #[instrument(name = "state", skip(self, req))]
@@ -1072,8 +1050,6 @@ impl Service<ReadRequest> for ReadStateService {
 
     fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         // Check for panics in the block write task
-        //
-        // TODO: turn this into a shared function for the StateService and ReadStateService
         let block_write_task = self.block_write_task.take();
 
         if let Some(block_write_task) = block_write_task {
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -255,12 +255,13 @@ impl Drop for StateService {
         std::mem::drop(self.finalized_block_write_sender.take());
         std::mem::drop(self.non_finalized_block_write_sender.take());
 
-        // Shut down the database (blocking):
-        // - stops the block write task if it is busy writing to the database
-        // - stops concurrent reads from the database
-        self.read_service.db.shutdown(true);
+        //// Shut down the database (blocking):
+        //// - stops the block write task if it is busy writing to the database
+        //// - stops concurrent reads from the database
+        //self.read_service.db.shutdown(true);
 
-        // Then drop self.read_service, which checks the block write task for panics.
+        // Then drop self.read_service, which checks the block write task for panics,
+        // and tries to shut down the database.
     }
 }
 
@@ -269,14 +270,14 @@ impl Drop for ReadStateService {
         // The read state service shares the state,
         // so dropping it should check if we can shut down.
 
-        // TODO: rename this to try_shutdown()?
-        self.db.shutdown(false);
-
         // Wait until the block write task finishes, then check for panics (blocking).
         //
         // TODO: move this into a function
         if let Some(block_write_task) = self.block_write_task.take() {
             if let Ok(block_write_task_handle) = Arc::try_unwrap(block_write_task) {
+                // We're the last database user, so we can tell it to shut down.
+                self.db.shutdown(true);
+
                 // We are the last state with a reference to this thread,
                 // so we can propagate any panics.
                 // (We'd also like to abort the thread, but std::thread::JoinHandle can't do that.)
@@ -287,6 +288,11 @@ impl Drop for ReadStateService {
                     info!("shutting down the state without waiting for the block write task");
                 }
             }
+        } else {
+            // Even if we're not the last database user, try shutting it down.
+            //
+            // TODO: rename this to try_shutdown()?
+            self.db.shutdown(false);
         }
     }
 }
@@ -308,7 +314,7 @@ impl StateService {
 
         let timer = CodeTimer::start();
         let initial_tip = finalized_state
-            .db()
+            .db
             .tip_block()
             .map(FinalizedBlock::from)
             .map(ChainTipBlock::from);
@@ -349,7 +355,7 @@ impl StateService {
         let queued_non_finalized_blocks = QueuedBlocks::default();
         let pending_utxos = PendingUtxos::default();
 
-        let last_block_hash_sent = finalized_state.db().finalized_tip_hash();
+        let last_block_hash_sent = finalized_state.db.finalized_tip_hash();
 
         let state = Self {
             network,
@@ -656,10 +662,10 @@ impl StateService {
         self.check_contextual_validity(&prepared)?;
         let parent_hash = prepared.block.header.previous_block_hash;
 
-        if self.disk.db().finalized_tip_hash() == parent_hash {
-            self.mem.commit_new_chain(prepared, self.disk.db())?;
+        if self.disk.db.finalized_tip_hash() == parent_hash {
+            self.mem.commit_new_chain(prepared, &self.disk.db)?;
         } else {
-            self.mem.commit_block(prepared, self.disk.db())?;
+            self.mem.commit_block(prepared, &self.disk.db)?;
         }
 
         Ok(())
@@ -732,11 +738,11 @@ impl StateService {
         check::block_is_valid_for_recent_chain(
             prepared,
             self.network,
-            self.disk.db().finalized_tip_height(),
+            self.disk.db.finalized_tip_height(),
             relevant_chain,
         )?;
 
-        check::nullifier::no_duplicates_in_finalized_chain(prepared, self.disk.db())?;
+        check::nullifier::no_duplicates_in_finalized_chain(prepared, &self.disk.db)?;
 
         Ok(())
     }
@@ -792,7 +798,7 @@ impl ReadStateService {
 
         let read_service = Self {
             network: finalized_state.network(),
-            db: finalized_state.db().clone(),
+            db: finalized_state.db.clone(),
             non_finalized_state_receiver: WatchReceiver::new(non_finalized_state_receiver),
             block_write_task: Some(block_write_task),
         };
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -166,9 +166,8 @@ pub(crate) struct StateService {
     ///
     /// Used to check for panics when writing blocks.
     //
-    // TODO: actually check for panics
-    //       work out if we need to shut it down when the service is shutting down
-    _block_write_task: Arc<std::thread::JoinHandle<()>>,
+    // TODO: check for panics in ReadStateService
+    block_write_task: Option<Arc<std::thread::JoinHandle<()>>>,
 
     // Pending UTXO Request Tracking
     //
@@ -246,9 +245,7 @@ pub struct ReadStateService {
     /// once the queues have received all their parent blocks.
     ///
     /// Used to check for panics when writing blocks.
-    //
-    // TODO: actually check for panics
-    _block_write_task: Arc<std::thread::JoinHandle<()>>,
+    block_write_task: Option<Arc<std::thread::JoinHandle<()>>>,
 }
 
 impl StateService {
@@ -321,7 +318,7 @@ impl StateService {
             finalized_block_write_sender,
             last_block_hash_sent,
             invalid_block_reset_receiver,
-            _block_write_task: block_write_task,
+            block_write_task: Some(block_write_task),
             pending_utxos,
             last_prune: Instant::now(),
             chain_tip_sender,
@@ -714,7 +711,7 @@ impl ReadStateService {
             network: finalized_state.network(),
             db: finalized_state.db().clone(),
             non_finalized_state_receiver: WatchReceiver::new(non_finalized_state_receiver),
-            _block_write_task: block_write_task,
+            block_write_task: Some(block_write_task),
         };
 
         tracing::info!("created new read-only state service");
@@ -730,6 +727,30 @@ impl Service<Request> for StateService {
         Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
 
     fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // Check for panics in the block write task
+        //
+        // TODO: turn this into a shared function for the StateService and ReadStateService
+        let block_write_task = self.block_write_task.take();
+
+        if let Some(block_write_task) = block_write_task {
+            if block_write_task.is_finished() {
+                match Arc::try_unwrap(block_write_task) {
+                    // We are the last state with a reference to this task, so we can propagate any panics
+                    Ok(block_write_task_handle) => {
+                        if let Err(thread_panic) = block_write_task_handle.join() {
+                            std::panic::resume_unwind(thread_panic);
+                        }
+                    }
+                    // We're not the last state, so we need to put it back
+                    Err(arc_block_write_task) => self.block_write_task = Some(arc_block_write_task),
+                }
+            } else {
+                // It hasn't finished, so we need to put it back
+                self.block_write_task = Some(block_write_task);
+            }
+        }
+
+        // Prune outdated UTXO requests
         let now = Instant::now();
 
         if self.last_prune + Self::PRUNE_INTERVAL < now {
@@ -789,6 +810,10 @@ impl Service<Request> for StateService {
                     span.in_scope(|| self.queue_and_commit_non_finalized(prepared))
                 });
 
+                // TODO:
+                //   - check for panics in the block write task here,
+                //     as well as in poll_ready()
+
                 // The work is all done, the future just waits on a channel for the result
                 timer.finish(module_path!(), line!(), "CommitBlock");
 
@@ -834,6 +859,10 @@ impl Service<Request> for StateService {
                 let rsp_rx = tokio::task::block_in_place(move || {
                     span.in_scope(|| self.queue_and_commit_finalized(finalized))
                 });
+
+                // TODO:
+                //   - check for panics in the block write task here,
+                //     as well as in poll_ready()
 
                 // The work is all done, the future just waits on a channel for the result
                 timer.finish(module_path!(), line!(), "CommitFinalizedBlock");
@@ -957,6 +986,29 @@ impl Service<ReadRequest> for ReadStateService {
         Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
 
     fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // Check for panics in the block write task
+        //
+        // TODO: turn this into a shared function for the StateService and ReadStateService
+        let block_write_task = self.block_write_task.take();
+
+        if let Some(block_write_task) = block_write_task {
+            if block_write_task.is_finished() {
+                match Arc::try_unwrap(block_write_task) {
+                    // We are the last state with a reference to this task, so we can propagate any panics
+                    Ok(block_write_task_handle) => {
+                        if let Err(thread_panic) = block_write_task_handle.join() {
+                            std::panic::resume_unwind(thread_panic);
+                        }
+                    }
+                    // We're not the last state, so we need to put it back
+                    Err(arc_block_write_task) => self.block_write_task = Some(arc_block_write_task),
+                }
+            } else {
+                // It hasn't finished, so we need to put it back
+                self.block_write_task = Some(block_write_task);
+            }
+        }
+
         Poll::Ready(Ok(()))
     }
 
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -424,8 +424,15 @@ impl StateService {
 
         if self.finalized_block_write_sender.is_some() {
             // We're still committing finalized blocks
-            self.queued_finalized_blocks
-                .insert(queued_prev_hash, queued);
+            if let Some(duplicate_queued) = self
+                .queued_finalized_blocks
+                .insert(queued_prev_hash, queued)
+            {
+                Self::send_finalized_block_error(
+                    duplicate_queued,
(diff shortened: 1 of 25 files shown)


teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/3050207033

If the mutex makes us loose too much performance, we might have to add the non-finalized channel to this PR before we merge it.

teor2345:
There are some clippy warnings in some unchanged code:
> changes to closure capture in Rust 2021 will affect drop order

I think they might be solved by PR #5137, otherwise we might just disable them.
(Drop order doesn't matter here, and we are on Rust 2021 edition.)

teor2345:
There are some errors in the tests, I think they might be fixed by moving the non-finalized state to the block commit task.
If not, I can add workarounds.

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,105 @@
...
+pub fn write_blocks_from_channels(
+    mut finalized_block_write_receiver: tokio::sync::mpsc::UnboundedReceiver<QueuedFinalized>,
+    mut non_finalized_block_write_receiver: tokio::sync::mpsc::UnboundedReceiver<
+        QueuedNonFinalized,
+    >,
+    mut finalized_state: FinalizedState,
+    invalid_block_reset_sender: tokio::sync::mpsc::UnboundedSender<block::Hash>,
+    chain_tip_sender: Arc<Mutex<ChainTipSender>>,
+) {
+    while let Some(ordered_block) = finalized_block_write_receiver.blocking_recv() {
```
teor2345, line 34: Close the other end of this channel when we're done: when we get a CommitBlock that is the child of the last sent hash

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,105 @@
...
+                 and clearing child blocks in the channel",
+            );
+
+            let send_result =
+                invalid_block_reset_sender.send(finalized_state.db().finalized_tip_hash());
+
+            if send_result.is_err() {
+                info!("StateService closed the block reset channel. Is Zebra shutting down?");
+                return;
+            }
```
teor2345: Don't send multiple times - if we do, then we reset even after the state has sent a valid block.

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+    }
+
+    while let Some(_block) = non_finalized_block_write_receiver.blocking_recv() {
+        // TODO:
+        // - read from the channel
+        // - commit blocks to the non-finalized state
+        // - commit blocks to the finalized state
```
teor2345:
```suggestion
        // - if there are any ready, commit blocks to the finalized state
```

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,105 @@
...
+        // - handle errors
```
teor2345:
```suggestion
        // - handle errors by sending a reset with all the block hashes in the non-finalized state, and the finalized tip
```

zebra-state/src/service/write.rs
teor2345: Fixed by cbc3d4173b

zebra-state/src/service/write.rs
teor2345: Fixed by commit 7f87c7a688

teor2345:
Hey @arya2, I tried to change a few parts of the code to get shutdown working, I hope they don't cause too many rebase conflicts!

Edit: seems like shutdown isn't quite working yet

teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/3059036752

teor2345:
I think I fixed the shutdown panics, the solution was "stop trying so hard":
https://github.com/ZcashFoundation/zebra/blob/cd9e2395623a8b5f2c55843f69c472c7558352a2/zebra-state/src/service/finalized_state/disk_db.rs#L714-L715

teor2345:
It seems like there are also some timing issues where the tests don't wait for the block commit task to finish.

For example:
https://github.com/ZcashFoundation/zebra/actions/runs/3065213251/jobs/4949092612#step:14:5326

I'll fix these on Monday.
(They pass on my machine, which is why I think they are a timing issue.)

zebra-state/src/service.rs
```diff
@@ -98,19 +100,67 @@ pub(crate) struct StateService {
...
     pub(crate) disk: FinalizedState,
 
-    /// The non-finalized chain state, including its in-memory chain forks.
-    mem: NonFinalizedState,
+    /// A channel to send blocks to the `block_write_task`,
+    /// so they can be written to the [`NonFinalizedState`].
+    //
+    // TODO: actually send blocks on this channel
+    non_finalized_block_write_sender:
+        Option<tokio::sync::mpsc::UnboundedSender<QueuedNonFinalized>>,
```
arya2, line 136: I'm just curious, why the tokio channel rather than std?

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,130 @@
...
+                if send_result.is_err() {
+                    info!("StateService closed the block reset channel. Is Zebra shutting down?");
+                    return;
+                }
+            }
+        }
+    }
+
+    // We're finished receiving finalized blocks from the state.
+    finalized_block_write_receiver.close();
```
arya2:
```suggestion
    // We're finished receiving finalized blocks from the state.
```

we should get rid of this since it would only ever reach past the blocking_recv call once the channel is already closed and empty

teor2345:
> I'm running a full sync here: https://github.com/ZcashFoundation/zebra/actions/runs/3059036752

It was about 23 minutes slower than commit 202dc43b11114761068655178d73de144d18d6e0:
https://github.com/ZcashFoundation/zebra/actions/runs/3070874944/jobs/4972408731#step:6:499

This is possibly the `Mutex`, but even if it's not, 20 minutes over 27 hours is within normal performance variations.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -715,16 +749,24 @@ impl DiskDb {
         let clone_prevention_guard = Arc::get_mut(&mut self.db);
 
         if clone_prevention_guard.is_none() && !force {
```
arya2: the `shutdown` method above returns before calling `delete_ephemeral` if `clone_prevention_guard.is_none()` and `force` is false, and that's the only place where `delete_ephemeral` is called.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -715,16 +749,24 @@ impl DiskDb {
...
-            debug!(
-                "dropping cloned DiskDb, \
-                 but keeping shared database files until the last reference is dropped",
-            );
-
-            return;
+            let path = self.path();
+            if force {
```
arya2: the if statement for this block requires `!force`

arya2 (commented):
Great work with these changes

zebra-state/src/service.rs
```diff
@@ -249,28 +399,106 @@ impl StateService {
...
+                "already finished committing finalized blocks: dropped duplicate block, \
+                 block is already committed to the state"
+                    .into(),
+            ));
+        }
+
+        if self.queued_finalized_blocks.is_empty() {
+            self.max_queued_finalized_height = f64::NAN;
+        } else if self.max_queued_finalized_height.is_nan()
+            || self.max_queued_finalized_height < queued_height.0 as f64
```
upbqdn, line 461: I assume we have to do it this way because it wouldn't be easier to deal with NaNs in code that's closer to Grafana. Is that correct?

zebra-state/src/service.rs
teor2345: `UnboundedSender` and `UnboundedReceiver` have a simpler API that did exactly what I wanted:
https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html

I think we could switch if we needed to, let me know if you need a channel method that isn't in `tokio`.

zebra-state/src/service/write.rs
teor2345: Thanks, fixed in 870fa3fae4

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: Good catch!

The sole reference check was also more complicated than it needed to be, and it was documented incorrectly.

So I fixed that all up in commit 4e4c186f47

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: Oops, removed in 048038b3bb

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,131 @@
...
+        match finalized_state.commit_finalized(ordered_block) {
+            Ok(finalized) => {
+                let tip_block = ChainTipBlock::from(finalized);
+
+                // TODO: update the chain tip sender with non-finalized blocks in this function,
+                //       and get rid of the mutex
+                chain_tip_sender
+                    .lock()
+                    .expect("unexpected panic in block commit task or state")
+                    .set_finalized_tip(tip_block);
```
teor2345, line 78: I think this `Mutex` changes the timing of the tip update, which causes some RPC tests that depend on that exact timing to fail:
- `zebra_rpc::methods::tests::snapshot::test_rpc_response_data`
- `zebra_rpc::methods::tests::vectors::rpc_getaddresstxids_response`
- `zebra_rpc::methods::tests::vectors::rpc_getbestblockhash`

This seems to happen on some test runs on some machines, but not at all on my machine 😭 

Here are the CI failures:
https://github.com/ZcashFoundation/zebra/actions/runs/3065213251/jobs/4949092612#step:14:5326
https://github.com/ZcashFoundation/zebra/actions/runs/3065213251/jobs/4949092612#step:14:5746
https://github.com/ZcashFoundation/zebra/actions/runs/3065213251/jobs/4949092612#step:14:5877
https://github.com/ZcashFoundation/zebra/actions/runs/3065213265/jobs/4949322117#step:3:4483

The fix is to wait for the chain tip update _before_ sending the RPC requests: 55e03969ba0bbef4f3f2bb7fa3569ee2668e038f

zebra-state/src/service.rs
```diff
@@ -271,75 +399,104 @@ impl StateService {
...
             "state.checkpoint.queued.block.count",
             self.queued_finalized_blocks.len() as f64,
         );
 
-        highest_queue_commit
+        rsp_rx
+    }
+
+    /// Finds queued finalized blocks to be committed to the state in order,
+    /// remove them from the queue, and sends them to the block commit task.
```
upbqdn:
```suggestion
    /// removes them from the queue, and sends them to the block commit task.
```

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use zebra_chain::block::{self, Height};
+
+use crate::service::{
+    finalized_state::FinalizedState,
+    queued_blocks::{QueuedFinalized, QueuedNonFinalized},
+    ChainTipBlock, ChainTipSender,
+};
+
+/// Reads blocks from the channels, write them to the `finalized_state`,
```
upbqdn:
```suggestion
/// Reads blocks from the channels, writes them to the `finalized_state`,
```

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,131 @@
...
+/// and updates the
```
upbqdn:
```suggestion
/// and updates the `chain_tip_sender`.
```

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,131 @@
...
+    // until the state closes the finalized block channel's sender.
+    while let Some(ordered_block) = finalized_block_write_receiver.blocking_recv() {
+        // TODO: split these checks into separate functions
+
+        if invalid_block_reset_sender.is_closed() {
+            info!("StateService closed the block reset channel. Is Zebra shutting down?");
+            return;
+        }
+
+        // Discard any children of invalid blocks in the channel
```
upbqdn, line 42: Are we actually discarding anything here?

zebra-state/src/service/tests.rs
```diff
@@ -515,6 +468,80 @@ proptest! {
     }
 }
 
+// This test sleeps for every block, so we only ever want to run it once
```
upbqdn, line 474: How is it possible that the test would be run multiple times if we didn't wrap it into `proptest`?

zebra-state/src/service/write.rs
teor2345: This seems to be causing test hangs, I'll see what I can do.

teor2345: I fixed this by adding a timeout in 770f73a76e

zebra-state/src/service.rs
```diff
@@ -716,14 +924,13 @@ impl Service<Request> for StateService {
...
-                // and written to disk.
-                //
-                // See the note in `CommitBlock` for more details.
-                let span = Span::current();
-                let rsp_rx = tokio::task::block_in_place(move || {
-                    span.in_scope(|| self.queue_and_commit_finalized(finalized))
-                });
+                // This method doesn't block, access the database, or perform CPU-intensive tasks,
+                // so we can run it directly in the tokio executor's Future threads.
+                let rsp_rx = self.queue_and_commit_finalized(finalized);
```
teor2345, line 969: I removed this redundant `block_in_place()` in 619caef95b.

zebra-state/src/service/arbitrary.rs
```diff
@@ -209,7 +213,24 @@ pub async fn populated_state(
...
-        rsp.expect("blocks should commit just fine");
+        // Wait for the block result and the chain tip update,
+        // which both happen in a separate thread from this one.
+        rsp.expect("unexpected block commit failure");
+
+        // Allow the chain tip update
+        if let Err(timeout_error) = timeout(
+            CHAIN_TIP_UPDATE_WAIT_LIMIT,
+            chain_tip_change.wait_for_tip_change(),
+        )
```
teor2345, line 224: I added this wait for tip change and timeout to fix test concurrency bugs in b890114eb4f8470a4c6b0d741d718167f894d531 and 770f73a76e28f826d7d8d34499e193cffb0eec9b.

zebra-state/src/service.rs
teor2345: `NaN`s show up as empty graphs, which is what we want when the code is empty.

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,131 @@
...
+            debug!(
+                ?next_valid_height,
+                invalid_height = ?ordered_block.0.height,
+                invalid_hash = ?ordered_block.0.hash,
+                "got a block that was too high. \
+                 Assuming a parent block failed, and dropping this block",
+            );
+
+            // We don't want to send a reset here, because it could overwrite a valid sent hash
+        }
```
teor2345, line 66: We need to drop the block here, and skip committing the block:
```suggestion
            std::mem::drop(ordered_block);
            continue;
        }
```

zebra-state/src/service/write.rs
```diff
@@ -44,17 +44,11 @@ pub fn write_blocks_from_channels(
                 ?next_valid_height,
```
teor2345, line 56: ^ Add a comment:
> `commit_finalized()` requires that we commit blocks in height order.
> If there has been a block commit error, we need to drop all the children of that block.

zebra-state/src/service/tests.rs
teor2345: It was moved from one proptest block to another.

zebra-state/src/service/write.rs
teor2345: Oops!

zebra-state/src/service.rs
```diff
@@ -191,6 +236,59 @@ pub struct ReadStateService {
...
+    /// Used to check for panics when writing blocks.
+    block_write_task: Option<Arc<std::thread::JoinHandle<()>>>,
+}
+
+impl Drop for StateService {
+    fn drop(&mut self) {
+        // The state service owns the state, tasks, and channels,
+        // so dropping it should shut down everything.
+
+        // Close the channels (non-blocking)
```
teor2345, line 252:
```suggestion
        // Close the channels (non-blocking)
        // This makes the block write thread exit the next time it checks the channels.
        // We want to do this here so we get any errors or panics from the block write task before it shuts down.
```

zebra-state/src/service/write.rs
teor2345: Done in 72ddbe3d0

teor2345:
Seems like we might have one final test timing issue on macOS 😭 

`mempool_cancel_downloads_after_network_upgrade` failed with:
> Message:  Test failed: assertion failed: `(left == right)`
>   left: `Some(Instant { t: 2398753679665 })`,
>  right: `None`: an already-evicted transaction should not be evicted again; minimal failing input: txid = Legacy(transaction::Hash("0000000000000000000000000000000000000000000000000000000000000000"))

https://github.com/ZcashFoundation/zebra/actions/runs/3094459231/jobs/5007860660#step:14:59341

I'll see if there is a quick fix tomorrow, otherwise I'll just disable this test on macOS.
(It's a second-tier support platform without guaranteed tests.)

teor2345:
I made these changes:
- wait for chain tip updates in a failing macOS test
- drop duplicate and leftover blocks, and send errors for dropped blocks
- update CommitFinalizedBlock documentation

teor2345:
Google Cloud ssh failure #5069:
> kex_exchange_identification: Connection closed by remote host

https://github.com/ZcashFoundation/zebra/actions/runs/3109728269/jobs/5041009873#step:6:105

teor2345:
Failure in inbound_block_height_lookahead_limit, looks like another block out of sequence thing:

>   left: `AdvertiseBlock(block::Hash("00040fe8ec8471911baa1db1266ea15dd06b4a8a5c453883c000b031973dce08"))`,
>  right: `AdvertiseBlock(block::Hash("0007bc227e1c57a4a70e237cad00e7b7ce565155ab49166bc57397a26d339283"))`: received an unexpected request
>  in zebra_test::mock_service::MockService<zebra_network::protocol::internal::request::Request, zebra_network::protocol::internal::response::Response, zebra_test::mock_service::PanicAssertion, alloc::boxed::Box<dyn std::error::Error+core::marker::Sync+core::marker::Send>>

https://github.com/ZcashFoundation/zebra/actions/runs/3109728269/jobs/5041101430#step:3:8517

teor2345:
> Failure in inbound_block_height_lookahead_limit, looks like another block out of sequence thing:

I think I fixed it.

arya2:
> Failure in inbound_block_height_lookahead_limit, looks like another block out of sequence thing:

The same failure happened in mempool_advertise_transaction_ids, which passed in the previous run

>   left: `AdvertiseBlock(block::Hash("00040fe8ec8471911baa1db1266ea15dd06b4a8a5c453883c000b031973dce08"))`,
>  right: `AdvertiseBlock(block::Hash("0007bc227e1c57a4a70e237cad00e7b7ce565155ab49166bc57397a26d339283"))`: received an unexpected request
>  in zebra_test::mock_service::MockService<zebra_network::protocol::internal::request::Request, zebra_network::protocol::internal::response::Response, zebra_test::mock_service::PanicAssertion, alloc::boxed::Box<dyn std::error::Error+core::marker::Sync+core::marker::Send>>

https://github.com/ZcashFoundation/zebra/actions/runs/3110257397/jobs/5041556025#step:3:8863

teor2345:
> > Failure in inbound_block_height_lookahead_limit, looks like another block out of sequence thing:
> 
> The same failure happened in mempool_advertise_transaction_ids, which passed in the previous run

Seems like the problem is in `setup()`, I tried to fix it in aef47858d4.

teor2345:
Looks like this all works, apart from the Rust bug fixed by PR #5251

upbqdn:
@mergifyio update

teor2345:
I need to call `clear_finalized_block_queue()` in the `Drop` impl, and instead of `self.queued_finalized_blocks.clear()` when we finish committing finalized blocks.

teor2345:
CI failed due to #5044

teor2345:
I'd also like to remove `Clone` from `NonFinalizedState`, because it doesn't work the same way as `FinalizedState.clone()`.

teor2345:
> I'd also like to remove `Clone` from `NonFinalizedState`, because it doesn't work the same way as `FinalizedState.clone()`.

That's not possible, so I just updated the docs.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -176,8 +176,13 @@ pub async fn run() -> Result<()> {
         assert_eq!(response, expected_response);
     }
 
-    tracing::info!("waiting for mempool to verify some transactions...");
-    zebrad.expect_stdout_line_matches("sending mempool transaction broadcast")?;
+    // The timing of verification logs are unrealiable, so we've disabled this check for now.
```
arya2, line 179: Not at all important.

```suggestion
    // The timing of verification logs are unreliable, so we've disabled this check for now.
```

arya2 (dismissed):
Excellent.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2, line 179: Not at all important so I'll update this in the following PR

```suggestion
    // The timing of verification logs are unreliable, so we've disabled this check for now.
```

zebra-state/src/service/finalized_state.rs
```diff
@@ -171,21 +163,23 @@ impl FinalizedState {
...
         };
 
-        let _ = rsp_tx.send(result.map_err(Into::into));
+        // Some io errors can't be cloned, so we format them instead.
+        let owned_result = result
+            .as_ref()
+            .map(|_hash| finalized)
+            .map_err(|error| format!("{:?}", error).into());
+
+        let _ = rsp_tx.send(result);
```
upbqdn, line 180: Could we swap these two statements, and return the adjusted result directly?

upbqdn (dismissed):
Looks great. I left one minor q.

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,138 @@
...
+        // So if there has been a block commit error,
+        // we need to drop all the descendants of that block,
+        // until we receive a block at the required next height.
+        let next_valid_height = finalized_state
+            .db
+            .finalized_tip_height()
+            .map(|height| (height + 1).expect("committed heights are valid"))
+            .unwrap_or(Height(0));
+
+        if ordered_block.0.height > next_valid_height {
```
upbqdn: Could it happen that we attempt to commit a block with a height lower than `next_valid_height`? In that case, we would need
```suggestion
        if ordered_block.0.height != next_valid_height {
```

zebra-state/src/service/finalized_state.rs
teor2345: `send()` takes ownership of the Result, and the error isn't cloneable, so we have to format the error before we send it to two different places.

zebra-state/src/service/write.rs
```diff
@@ -0,0 +1,138 @@
...
+            debug!(
+                ?next_valid_height,
+                invalid_height = ?ordered_block.0.height,
+                invalid_hash = ?ordered_block.0.hash,
+                "got a block that was too high. \
```
teor2345:
```suggestion
                "got a block that was the wrong height. \
```

zebra-state/src/service/write.rs
teor2345: You're right, we should use the same check as `commit_finalized()` does.

teor2345:
One last fix to the block height check in b150e3f, and I think we're done.

zebra-state/src/service/finalized_state.rs
upbqdn: This should have been obvious to me. Sorry.

upbqdn (approved):

zebra-state/src/service/finalized_state.rs
teor2345: > This should have been obvious to me. Sorry.

No worries, it's not obvious in the code, it trips me up all the time!

PR #4917, opened 2022-08-17
docs(release): add versioning and releases information

gustavovalverde:
## Motivation

Our release process and criteria needed to be updated.

## Review

This is a draft PR, but it needs review in this stage before continue adding more sections or details.

### Reviewer Checklist

  - [ ] Sections are the ones required in this document
  - [ ] The stated arguments can be complied by the Zebra team
  - [ ] None of the arguments goes against our Tier Policies

## Follow Up Work

Get feedback to make the needed changes

Review some Angular documents that might be helpful to adapt:
-  https://github.com/angular/angular/blob/main/docs/BRANCHES.md
- https://github.com/angular/angular/blob/main/docs/TRIAGE_AND_LABELS.md

Fixes #4778

PR #5138, opened 2022-09-12
feat(release): create Docker hub binaries when tagging

### LINKED ISSUE #1963: Work out how to release Zebra binaries
## Goals

Document how we release Zebra binaries.

## Tasks

- [ ] Get a Docker partner/verified account for the Zcash Foundation
- [ ] Work out how to release docker binaries securely
- [ ] Document how releases work
- [ ] Delete the outdated release RFC

## Out of Scope

- #4953 
- #1615

gustavovalverde:
## Motivation

We'd like to have binaries published in Docker Hub for easier pulling, as people is more used to use the `pull <org>/<image>:<tag>` approach when pulling Docker images, instead of long URLs which could also change.

### Designs

Publish binaries to Docker Hub when we're making a new release.

## Solution
- Create a new workflow for binary releases
- Just build this binaries when a release is created (not a pre-release)
- This binaries are meant to be push to Docker Hub

Fixes #1963

## Review
We might want to be more strict with which tags to release to Docker Hub.

### Reviewer Checklist

  - [ ] Tests for Expected Behavior

## Follow Up Work

Update https://github.com/ZcashFoundation/zebra/pull/4917 with this information (?)


.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow is meant to trigger a build of Docker binaries when a tag
+# is created, it uses the existing `build-docker-image.yml` workflow
+#
+# We use a separate action as we might want to trigger this with different under
+# different circumstances than a Continous Deployment, for example.
```
teor2345: Just fixing some typos:
```suggestion
# We use a separate action as we might want to trigger this under
# different circumstances than a Continuous Deployment, for example.
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
...
+      dockerfile_path: ./docker/Dockerfile
+      dockerfile_target: runtime
+      image_name: zebrad
+      network: Mainnet
+      # TODO: validate this are the values we want on a public image
+      checkpoint_sync: true
+      rust_backtrace: '1'
+      zebra_skip_ipv6_tests: '1'
+      rust_log: info
+    # This step needs access to Docker Hub secrets to run succesfully
```
teor2345:
```suggestion
    # This step needs access to Docker Hub secrets to run successfully
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
...
+  # The image will be commonly named `zebrad:<short-hash | github-ref | semver>`
+  build:
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    uses: ./.github/workflows/build-docker-image.yml
+    with:
+      dockerfile_path: ./docker/Dockerfile
+      dockerfile_target: runtime
+      image_name: zebrad
+      network: Mainnet
+      # TODO: validate this are the values we want on a public image
```
teor2345: These settings are fine for a release candidate.
We might want to make the logging less verbose later, if it is something that annoys or concerns users.

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow is meant to trigger a build of Docker binaries when a tag
```
teor2345: Can you add a short document describing our automated release builds to:
https://github.com/ZcashFoundation/zebra/tree/main/book/src/dev

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
...
+# is created, it uses the existing `build-docker-image.yml` workflow
+#
+# We use a separate action as we might want to trigger this with different under
+# different circumstances than a Continous Deployment, for example.
+name: Release binaries
+
+on:
+  create:
```
teor2345: I'm not sure if we want to use tag filters.

We can avoid them by building the image when a pre-release is "published".
(Pre-releases don't appear on the Zebra front page, and they have a pre-release tag in the release list.)

Then when the image is ready, we can change the pre-release to a release (the "released" event).

Here are the details:
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-41

```suggestion
on:
  release:
    types: [published]
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  # Each time this workflow is executed, a build will be triggered to create a new image
+  # with the corresponding tags using information from git
+  #
+  # The image will be commonly named `zebrad:<short-hash | github-ref | semver>`
+  build:
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```
teor2345: If we decide to remove the tag filter, we'll want to remove it here as well.

.github/workflows/build-docker-image.yml
```diff
@@ -107,6 +107,13 @@ jobs:
           password: ${{ steps.auth.outputs.access_token }}
           logout: false
 
+      - name: Login to DockerHub
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```
teor2345: If we decide to remove the tag filter, we'll want to remove it here as well.

teor2345 (changes requested):
This looks really good, thanks for pulling it together.

But I'm wondering if triggering releases using a tag filter is confusing or error-prone.

We've never used tags for development - we just seem to use branches.
But if we ever do, it would be a bit hard for developers to remember to start tags with letters other than `v`. (And if they forget, their tag gets published as an official release, and everyone using `zebrad:latest` automatically gets it.)

Can we use the release events instead?
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-41

.github/workflows/release-binaries.yml
gustavovalverde: I'll be adding it in this PR https://github.com/ZcashFoundation/zebra/pull/4917 as this should belong to the release process document.

.github/workflows/build-docker-image.yml
gustavovalverde: Fixed with [ci(release): just publish to DockerHub when a release is published](https://github.com/ZcashFoundation/zebra/pull/5138/commits/53191eaea9016ac8b0cd7633d77ed610fa9be62c)

.github/workflows/release-binaries.yml
gustavovalverde: Fixed with [ci(release): just publish to DockerHub when a release is published](https://github.com/ZcashFoundation/zebra/pull/5138/commits/53191eaea9016ac8b0cd7633d77ed610fa9be62c)

.github/workflows/release-binaries.yml
gustavovalverde: I decided on a slightly different approach (but similar to your proposal) with [ci(release): just publish to DockerHub when a release is published](https://github.com/ZcashFoundation/zebra/pull/5138/commits/53191eaea9016ac8b0cd7633d77ed610fa9be62c)

The event type `published` would also publish images to DockerHub when creating a pre-release, but we can keep those in GAR (Google Artifact Registry), if we'd like to publish it as an official release (even if that's a `beta`) then we would have to _un-check_ it as a `pre-release` in GitHub, and it would also be published to DockerHub.

This is mainly to keep DockerHub with just releases meant to be used in production, or after the images has been **tested** (as we should test those before deploying the binaries to DockerHub)

gustavovalverde: As expected.

When a release is published as `pre-release` it triggers our CD for internal testing: https://github.com/ZcashFoundation/zebra/actions/runs/3048068297

When the `pre-release` is _unchecked_ (converted to an official release) then it only runs the `Release binaries` workflow https://github.com/ZcashFoundation/zebra/actions/runs/3048093839

We should wait for both of these tests to finish

gustavovalverde:
Seems like this needs further testing as this condition is not working as expected . I'm pulling it back to `draft` in the meanwhile as I have to find a better way to make this work as expected https://github.com/ZcashFoundation/zebra/pull/5138/files#diff-5b40193e09025fd041de92d38c976d0b469e3ee85bef45a71d05d8b5078b1ee8R111

.github/workflows/build-docker-image.yml
```diff
@@ -105,7 +105,14 @@ jobs:
...
+
+      - name: Login to DockerHub
+        # We only publish images to DockerHub if a release is not a pre-release
+        if: github.event.release.action == 'released'
+        uses: docker/login-action@v2.0.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       # Build and push image to Google Artifact Registry
```
teor2345: I'm not sure if this is:
- always Google Artifact Registry and sometimes also DockerHub
- either Google Artifact Registry or DockerHub, but never both

```suggestion
      # Build and push image to Google Artifact Registry, and possibly DockerHub
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,35 @@
+# This workflow is meant to trigger a build of Docker binaries when a tag
```
teor2345, line 1:
```suggestion
# This workflow is meant to trigger a build of Docker binaries when a release
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,35 @@
...
+# is created, it uses the existing `build-docker-image.yml` workflow
```
teor2345:
```suggestion
# This workflow is meant to trigger a build of Docker binaries when a release
# is published, it uses the existing `build-docker-image.yml` workflow
```

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,35 @@
...
+    types:
+      - released
+
+jobs:
+  # Each time this workflow is executed, a build will be triggered to create a new image
+  # with the corresponding tags using information from git
+  #
+  # The image will be named `zebrad:<semver>`
+  build:
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```
teor2345: I think this can be removed now, as the release type makes sure that we're only publishing releases.

```suggestion
```

teor2345 (dismissed):
Thanks for these changes!

I just went through the comments and made them match what I think is happening. Feel free to edit them or reject them.

I'm happy for anyone to re-approve this PR.

gustavovalverde:
There were a bunch of edge cases, but this is finally ready!

.github/workflows/build-docker-image.yml
```diff
@@ -71,15 +71,16 @@ jobs:
           # list of Docker images to use as base name for tags
           images: |
             us-docker.pkg.dev/zealous-zebra/zebra/${{ inputs.image_name }}
+            zfnd/zebra,enable=${{ github.event_name == 'release' && !github.event.release.prerelease }}
           # generate Docker tags based on the following events/attributes
           tags: |
```
teor2345, line 76: Can you show me the full set of tags for each release?
And can you document them somewhere?

in particular, I am want to check these things:
- the version tag includes "-beta.15" or "-rc.0"
- the "sha" is long enough to be unique (git currently recommends 9 characters for our repository, and it will grow over time as we get more commits)
- we're using "v1" but "1.0", this seems inconsistent and a bit confusing?
- what if the ref looks like a version?

teor2345 (commented):
I just want to check the tags before we merge.

.github/workflows/build-docker-image.yml
gustavovalverde: > * the version tag includes "-beta.15" or "-rc.0"

These are tags for a `non-beta` or `non-rc` release:
https://github.com/ZcashFoundation/zebra/actions/runs/3053723143/jobs/4924709856#step:4:98

These are the tags for a `-beta.15` or `-rc.0`
https://github.com/ZcashFoundation/zebra/actions/runs/3053583083/jobs/4924390909#step:4:95

When a release specifies a `pre-release identifier` it just add the specific version tag and sha, it does not add `latest`, `v{{major}}`, `{{major}}.{{minor}}`, and neither of the other ones.


> * the "sha" is long enough to be unique (git currently recommends 9 characters for our repository, and it will grow over time as we get more commits)

This is automatically generated by https://github.com/docker/metadata-action#typesha and if we'd like to use the long version, this would need to be standardized on all the actions (this could be done later on)


> * we're using "v1" but "1.0", this seems inconsistent and a bit confusing?

If we have 1.0, 1.2, 1.3, and even another major version like 2.0, 2.2, 2,3, etc. We might want the latest version from version 1, this is `v1` as the `latest` would be 2.3. This is easier for the user to keep downloading the latest version from a specific major.

Rust has a good example in their [Docker Hub](https://hub.docker.com/_/rust), for example: `1-buster`, `1.63-buster`, `1.63.0-buster`, `buster`. If we remove the alias (`buster`) then we would have `1`, `1.63`, `1.63.0` 

In Rust case `buster` is their major alias, in our case `v1` would be our alias.


> * what if the ref looks like a version?

I'm not completely sure on the question, but the ref (from a `tag`) is just considered when publishing a release in GitHub as the workflow that publishes the official images is not triggered by other events. The only `ref` used is the one from the `tag` being published as a release.

Now...if the `tag` doesn't look like a version, it will deploy it as-is.

gustavovalverde:
Delete https://github.com/ZcashFoundation/zebra/releases/tag/v1.0.0 when this is approved

.github/workflows/build-docker-image.yml
```diff
@@ -71,15 +71,16 @@ jobs:
...
             type=schedule
-            type=sha
-            type=ref,event=branch
-            type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
+            type=semver,pattern=v{{major}}
```
teor2345: Is it going to confuse users if some version tags start with `v`, and others don't?
```
  zfnd/zebra:1.0.0
  zfnd/zebra:1.0
  zfnd/zebra:v1
```

```suggestion
            type=semver,pattern={{major}}
```

.github/workflows/build-docker-image.yml
teor2345: > > * the "sha" is long enough to be unique (git currently recommends 9 characters for our repository, and it will grow over time as we get more commits)
> 
> This is automatically generated by https://github.com/docker/metadata-action#typesha and if we'd like to use the long version, this would need to be standardized on all the actions (this could be done later on)

It's not a blocker, but it might be confusing for users if we ever tag two releases with the same short sha. We can fix this later if it becomes a significant problem.

> > * we're using "v1" but "1.0", this seems inconsistent and a bit confusing?
> 
> If we have 1.0, 1.2, 1.3, and even another major version like 2.0, 2.2, 2,3, etc. We might want the latest version from version 1, this is `v1` as the `latest` would be 2.3. This is easier for the user to keep downloading the latest version from a specific major.
> 
> Rust has a good example in their [Docker Hub](https://hub.docker.com/_/rust), for example: `1-buster`, `1.63-buster`, `1.63.0-buster`, `buster`. If we remove the alias (`buster`) then we would have `1`, `1.63`, `1.63.0`
> 
> In Rust case `buster` is their major alias, in our case `v1` would be our alias.

Rust does not have `v` in front of any of its versions. Can we remove the extra `v`, for consistency with the Rust ecosystem, and consistency between versions?
(See my suggestion below.)

teor2345 (changes requested):
We might want to format all the version numbers in our tags the same way.

teor2345 (approved):
Thanks for the changes, looks good!

PR #5163, opened 2022-09-14
chore(build): allow a custom `zebrad` config file path

gustavovalverde:
## Previous behavior:
The zebrad config file was hardcoded into the Dockerfile, allowing to change specific values, but not the config file as a whole

## Expected behavior:
Allow the user to specify the config file themselves, for example with the nginx image you can pass -v /host/path/nginx.conf:/etc/nginx/nginx.conf:ro

## Solution:
Use and ARG, with a default value, to allow the user to change the location of the zebrad configuration file. This also sets the location under /etc, honoring The Filesystem Hierarchy Standard (FHS) for linux

## Review

Anyone can review this, but @conradoplg proposed the change.

### Reviewer Checklist

  - [ ] CI passes
  - [ ] Tests for Expected Behavior

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -139,7 +139,7 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
-ARG ZEBRA_CONF_PATH=/etc/zebrad/zebrad.toml
+ARG ZEBRA_CONF_PATH=/etc/zebra/zebrad.toml
 ENV ZEBRA_CONF_PATH ${ZEBRA_CONF_PATH}
 
 # Build the `zebrad.toml` before starting the container, using the arguments from build
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -139,6 +139,9 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
+ARG ZEBRA_CONF_PATH=/etc/zebrad/zebrad.toml
+ENV ZEBRA_CONF_PATH ${ZEBRA_CONF_PATH}
+
 # Build the `zebrad.toml` before starting the container, using the arguments from build
 # time, or using the default values set just above.
 RUN set -ex; \
@@ -153,7 +156,7 @@ RUN set -ex; \
     echo "cache_dir = '/zebrad-cache'"; \
     echo "[tracing]"; \
     echo "endpoint_addr = '0.0.0.0:3000'"; \
-  } > "zebrad.toml"
+  } > "${ZEBRA_CONF_PATH}"
 
 EXPOSE 3000 8233 18233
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -139,6 +139,9 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
+ARG ZEBRA_CONF_PATH=/etc/zebrad/zebrad.toml
```
teor2345: If we want all the Zebra configs to be in the same directory, we might want to use:
```suggestion
ARG ZEBRA_CONF_PATH=/etc/zebra/zebrad.toml
```

For example, we might add configs for separate tools, like the checkpoints tool, or a future RPC or raw database access tool.

This also matches the name of the git repository.

docker/Dockerfile
```diff
@@ -153,7 +156,7 @@ RUN set -ex; \
     echo "cache_dir = '/zebrad-cache'"; \
     echo "[tracing]"; \
     echo "endpoint_addr = '0.0.0.0:3000'"; \
-  } > "zebrad.toml"
+  } > "${ZEBRA_CONF_PATH}"
```
teor2345, line 159: Do we need a test that checks that $ZEBRA_CONF_PATH works?
Do we already have a test that makes sure this default config works with the latest Zebra version, or do we need to add that test?

docker/Dockerfile
```diff
@@ -139,6 +139,9 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
+ARG ZEBRA_CONF_PATH=/etc/zebrad/zebrad.toml
+ENV ZEBRA_CONF_PATH ${ZEBRA_CONF_PATH}
```
teor2345, line 143: Can you add this environmental variable to the Docker user documentation?

teor2345 (commented):
I have a few suggestions, none of them are blocking, and only the path change needs to go in this PR. (If we decide to do it.)

But I'd like to make sure we have tickets any we need, so we don't forget.

docker/Dockerfile
teor2345: We added some config tests in #4684, so another option is adding this config to that directory.

gustavovalverde: I agree we might want to have a test to check `$ZEBRA_CONF_PATH` works, just as [oxarbitrage](https://github.com/oxarbitrage) added here [tests(config): Upgrade config file tests](https://github.com/ZcashFoundation/zebra/pull/5112#top)
#5112

docker/Dockerfile
gustavovalverde: Where's that documentation? 😅

gustavovalverde: Nevermind. Added this for tracking https://github.com/ZcashFoundation/zebra/issues/5167

docker/Dockerfile
gustavovalverde: I'm not the best one to create the description of this ticket, if this won't be done in this PR (or in the near future)

teor2345: I created ticket #5168

teor2345 (approved):

PR #5177, opened 2022-09-15
fix(build): a path must exist before creating a file in it

gustavovalverde:
## Motivation

Builds in the main branch are failing

## Solution

Other directories in the Dockerfile were automatically created with the `WORKDIR` and  `COPY` commands. In this case it has to be explicitly created 

## Review

Anyone can review this, and it's urgently needed

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -139,14 +139,21 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
-# Use a configurable path for the zebrad configuration file
-ARG ZEBRA_CONF_PATH=/etc/zebra/zebrad.toml
+# Use a configurable dir and file for the zebrad configuration file
+ARG ZEBRA_CONF_PATH=/etc/zebra
 ENV ZEBRA_CONF_PATH ${ZEBRA_CONF_PATH}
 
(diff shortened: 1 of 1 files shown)


dconnolly (approved):

docker/Dockerfile
```diff
@@ -139,11 +139,21 @@ RUN apt-get update && \
 ARG CHECKPOINT_SYNC=true
 ARG NETWORK=Mainnet
 
-ARG ZEBRA_CONF_PATH=/etc/zebra/zebrad.toml
+# Use a configurable dir and file for the zebrad configuration file
+ARG ZEBRA_CONF_PATH=/etc/zebra
```
upbqdn, line 143: We could change `PATH` to `DIR`
```suggestion
ARG ZEBRA_CONF_DIR=/etc/zebra
```

We could then have:
```
ZEBRA_CONF_PATH=${ZEBRA_CONF_DIR}/${ZEBRA_CONF_FILE}
```

upbqdn (approved):
Looks good. I left a minor suggestion.

docker/Dockerfile
gustavovalverde: I'll create a ticket to follow-up on this with a better implementation using the entrypoint.  But I do agree.

Issue #5187, opened 2022-09-19
Delete Google Cloud instances older than 3 days


teor2345:
## Motivation

Currently, we delete instance templates, disks, and images older than 2 days (if they are not being used).

But sometimes a test workflow can run and leave behind an instance. This costs money for CPU and RAM, as well as any related storage.

### Designs

Since our longest workflows run for 1-2 days, automatically delete any test instances older than 3 days.

### Example

These 2 instances have been running for an extra 1-2 weeks, even though their workflows have finished:

<img width="811" alt="Screen Shot 2022-09-19 at 11 21 24" src="https://user-images.githubusercontent.com/8951843/190936837-32793a04-778e-4913-84a8-4be178187192.png">

teor2345:
In PR #5191 I made new caches happen every 12 hours, double the current rate.

7 days after that PR merges, we should review costs for the last 3-4 days, and see if they are better or worse.

PR #5191, opened 2022-09-19
change(ci): Create a new cached state image every 12 hours

### LINKED ISSUE #5155: Decide if we want to create cached states more often
## Motivation

Currently we create updated cached states after a day (~1000 blocks).

But if we created them more often:
- CI would be faster
- we would use less CPU time, costing us less
- we would use more storage, costing us more

It seems like our storage costs are much lower than expected, so we could make it half a day if we want.
That would reduce our compute costs, because we would be syncing fewer blocks.
(We always sync 100 blocks, so it doesn't need to be much lower than 100-200.)

## Source Code

Here is the config for "how long to wait before creating another cached state image":
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/deploy-gcp-tests.yml#L97-L99

teor2345:
## Motivation

We want to try to reduce CPU cost (and increase CI speed) by creating a new cached state every 12 hours.
At the moment, we create them at most once per day.

Part of #5187.
Closes #5155.

## Review

This isn't urgent, but we'd like to get it merged to reduce costs.

### Reviewer Checklist

  - [ ] CI passes

## Follow Up Work

Check to see if it reduces costs.


teor2345:
@mergifyio update

gustavovalverde (approved):

gustavovalverde:
@Mergifyio update

Issue #4954, opened 2022-08-25
Update release checklist PR template for the release candidate series


teor2345:
## Motivation

When we've worked out how to release Zebra binaries, we need to update our release checklist.

### Tasks

Docker releases:
- [x] Update the release checklist for the new release process (PR #5193)
- [x] Add a "run zebrad" docker command to the README
- [x] Use the same docker command to test the released binaries
       (We already check that the released git tag will build, we also want to test the binaries)

`rc` status:
- [x] Change `zebrad` instructions from `beta` to `rc`
- [x] Keep the other crates on `beta` for now

PR #5193, opened 2022-09-19
doc(release): Update release template for Docker releases

teor2345:
## Motivation

1. We're releasing Docker Hub binaries now, so we need to add a "publish the release" step
2. The `Cargo.lock` description is confusing.
3. Semantic versioning is explained on their website, so we don't need a detailed description in the release template.

This is the Docker part of #4954, we still need to do the `rc` part.

## Solution

- Make the release a pre-release until the Docker binaries have been built and tested
- Explain that `cargo build` needs to be run
- Delete a long description of semantic versioning
- Fix a suggested branch name so it will work with Google Cloud

## Review

This is a routine update.

I'll need @gustavovalverde to check the new Docker release steps.

### Reviewer Checklist

  - [ ] Docs make sense


teor2345:
@mergifyio update

gustavovalverde (approved):

gustavovalverde:
@Mergifyio update

PR #5250, opened 2022-09-25
Revert "build(deps): bump codecov/codecov-action from 3.1.0 to 3.1.1"

teor2345:
## Motivation

After we merged this change, coverage builds started hanging during the download of the Zcash parameters.

We should try reverting it, and clearing the GitHub Actions caches. (The hang only seems to happen on the second run.)

## Solution

Reverts ZcashFoundation/zebra#5205


teor2345:
The bug wasn't caused by the codecov action version upgrade, because the parameter fetch is still timing out:
https://github.com/ZcashFoundation/zebra/actions/runs/3123961597/jobs/5066947753#step:9:18

PR #5257, opened 2022-09-26
change(state): Write non-finalized blocks to the state in a separate thread, to avoid network and RPC hangs

arya2:
## Motivation

This PR sends queued non-finalized blocks down a channel to the writer task, which makes state block writes concurrent

Part of [https://github.com/ZcashFoundation/zebra/issues/4937](https://github.com/ZcashFoundation/zebra/issues/4937).

### Designs

- [https://github.com/ZcashFoundation/zebra/issues/4937](https://github.com/ZcashFoundation/zebra/issues/4937)
- [https://docs.google.com/drawings/d/1FXpAUlenDAjl8nkftrypdAPsj0jr-Ut9gZlSP57nuyc](https://docs.google.com/drawings/d/1FXpAUlenDAjl8nkftrypdAPsj0jr-Ut9gZlSP57nuyc)

## Solution

Use block write task to:
 - Commit non-finalized blocks.
 - Propagate errors from queued parent blocks to queued children and skip validate_and_commit calls.
 - Finalize blocks on best chain past `MAX_REORG_HEIGHT`.

Related cleanups:
 - Removes unused `chain_tip_sender` and `non_finalized_state_sender` fields of StateService.

## Review

### Reviewer Checklist

## Follow Up Work

 - Avoid removing and re-adding known_utxos by nesting the SentHashes and QueuedBlocks inside another struct that holds that tracks the known_utxos
 - Avoid now-redundant checks against the chains in the Request::AwaitUtxo call by skipping to `db.utxo(&outpoint).map(|utxo| utxo.utxo)` instead of routing it to the ReadRequest after checking the queue and sent hashes on StateService.
 - Organize block write task into an object.


arya2:
I'm still looking into why `value_pool_is_updated` fails and `chain_tip_sender_is_updated` hangs

zebra-state/src/service.rs
```diff
@@ -718,98 +635,53 @@ impl StateService {
...
-    pub fn any_ancestor_blocks(&self, hash: block::Hash) -> block_iter::Iter<'_> {
-        block_iter::Iter {
-            service: self,
-            state: block_iter::IterState::NonFinalized(hash),
-        }
-    }
-
     /// Assert some assumptions about the prepared `block` before it is validated.
     fn assert_block_can_be_validated(&self, block: &PreparedBlock) {
         // required by validate_and_commit, moved here to make testing easier
```
teor2345: This comment needs to be updated, but maybe we want to move this method to the `write` module instead, so it is closer to the code it is related to?

```
        // required by validate_and_commit_non_finalized(), moved here to make testing easier
```

zebra-state/src/service/write.rs
```diff
@@ -108,31 +176,105 @@ pub fn write_blocks_from_channels(
...
+                let _ = rsp_tx.send(result.clone().map(|()| child_hash).map_err(BoxError::from));
+
+                if let Err(ref error) = result {
+                    parent_error_map.insert(parent_hash, error.clone());
+                }
+            }
+
+            NonFinalizedWriteCmd::FinishProcessingQueued => {
+                // Clear out any errors from processed queued blocks without children
+                parent_error_map.clear();
```
teor2345: I'm not sure if we want to clear the `parent_error_map` after each batch from the queue.
What if some of the children that need to be failed aren't in the queue yet?

Instead, we could use an `IndexMap`, and clear the oldest hashes when the map gets too long?
Let me know if you'd like me to do this rewrite in another PR.

zebra-state/src/service/write.rs
```diff
@@ -108,31 +176,105 @@ pub fn write_blocks_from_channels(
...
 
-    // We're finished receiving non-finalized blocks from the state.
-    //
-    // TODO:
-    // - make the task an object, and do this in the drop impl?
-    // - does the drop order matter here?
-    non_finalized_block_write_receiver.close();
-    std::mem::drop(non_finalized_block_write_receiver);
+                while non_finalized_state.best_chain_len()
+                    > crate::constants::MAX_BLOCK_REORG_HEIGHT
```
teor2345: What if the queue is very long?
If we delay finalizing thousands of blocks, we could run out of RAM, or have really slow write performance.
(We can add limited batching later, if we need to.)

Can we do the block writes and cached chain updates incrementally, every time we add a block to the non-finalized chain?
This might also fix some of the test failures, because it is more like the previous behaviour.

Let me know if you'd like me to do this rewrite in another PR.

zebra-state/src/service/write.rs
```diff
@@ -108,31 +176,105 @@ pub fn write_blocks_from_channels(
...
+                {
+                    tracing::trace!("finalizing block past the reorg limit");
+                    let finalized_with_trees = non_finalized_state.finalize();
+                    finalized_state
+                        .commit_finalized_direct(finalized_with_trees, "best non-finalized chain root")
+                        .expect(
+                            "expected that errors would not occur when writing to disk or updating note commitment and history trees",
```
teor2345: This failure message is outdated, we should have updated it as part of the NU5 work:
```suggestion
                            "unexpected finalized block commit error: note commitment and history trees were already checked by the non-finalized state",
```

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,106 @@
 //! Writing blocks to the finalized and non-finalized states.
+#![allow(clippy::too_many_arguments)]
```
teor2345: Nitpick: can you apply this clippy exception to the function that has the issue, rather than the entire module?

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,106 @@
...
-/// Reads blocks from the channels, writes them to the `finalized_state`,
+use std::collections::HashMap;
+
+pub enum NonFinalizedWriteCmd {
+    ProcessQueued {
+        parent_hash: block::Hash,
+        queued_child: QueuedNonFinalized,
+    },
+    FinishProcessingQueued,
+}
```
teor2345: We'll need extra tests for stateful updates like `NonFinalizedWriteCmd::FinishProcessingQueued`.
So if it's possible, I'd like to just send blocks on the channels.

See my suggestions below - let me know if you'd like me to do this rewrite in another PR.

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,106 @@
...
+
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
```
teor2345: Nitpick: we usually try to merge all the imports from the same crate together.

We also try to match Rust import order, which is:
- `std` library
- other crates
- `zebra*` crates
- this crate

Then the actual code and types in the module.

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,106 @@
...
+    non_finalized_state_sender: &watch::Sender<NonFinalizedState>,
+) -> Option<block::Height> {
+    let best_chain = non_finalized_state.best_chain();
+    let tip_block = best_chain
+        .and_then(|chain| chain.tip_block())
+        .cloned()
+        .map(ChainTipBlock::from);
+    let tip_block_height = tip_block.as_ref().map(|block| block.height);
+
+    // TODO: Update the non_finalized_state_receiver
```
teor2345: I'm not sure if this comment is outdated, or there is some other channel we should be updating?

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,106 @@
...
+    // If the final receiver was just dropped, ignore the error.
+    let _ = non_finalized_state_sender.send(non_finalized_state.clone());
+
+    chain_tip_sender.set_best_non_finalized_tip(tip_block);
+
+    tip_block_height
+}
+
+/// Reads blocks from the channels, write them to the `finalized_state`,
```
teor2345: Nitpick: Can you update this function comment?

```
/// Reads blocks from the channels, writes them to the `finalized_state` or `non_finalized_state`,
/// sends any errors on the `invalid_block_reset_sender` and (TODO????),
/// and updates the `chain_tip_sender` and `non_finalized_state_sender`.
```

zebra-state/src/service/write.rs
```diff
@@ -108,31 +176,105 @@ pub fn write_blocks_from_channels(
...
+                if let Some(parent_error) = parent_error {
+                    tracing::trace!(
+                        ?child_hash,
+                        ?parent_error,
+                        "rejecting queued child due to parent error"
+                    );
+                    result = Err(parent_error.clone());
+                } else {
+                    tracing::trace!(?child_hash, "validating queued child");
+                    result = validate_and_commit_non_finalized(
```
teor2345: Required Fix: `validate_and_commit_non_finalized()` will panic if the block is not ready to be committed:
https://github.com/ZcashFoundation/zebra/blob/bfdb29b757b3840cea01ffbb448f129eda669e01/zebra-state/src/service/non_finalized_state.rs#L475-L477

Since we want to clear old parent hashes out of `parent_error_map`, we need to check if all blocks are ready to be committed:
- check `parent_error_map`, return that error
- if the parent hash is not in any block in any non-finalized chain, or the finalized tip, return a generic "parent failed" error

Let me know if you'd like me to do this rewrite in another PR.

zebra-state/src/service.rs
```diff
@@ -333,20 +324,26 @@ impl StateService {
             tokio::sync::mpsc::unbounded_channel();
 
         let finalized_state_for_writing = finalized_state.clone();
-        let chain_tip_sender_for_writing = chain_tip_sender.clone();
+        let non_finalized_state_for_writing = non_finalized_state.clone();
```
teor2345: Required Fix: Cloning creates an independent clone of the non-finalized state, so `StateService.mem` never gets updated. Using an empty non-finalized state might cause some test failures.

Instead, remove `StateService.mem`, and use `StateService.read_service.non_finalized_state_receiver` to get a cached version of the latest non-finalized state.

`StateService.disk` is also the same as `StateService.read_service.db`, so let's remove `StateService.disk` at the same time.

zebra-state/src/service.rs
```diff
@@ -540,12 +535,11 @@ impl StateService {
 
     /// Send an error on a `QueuedFinalized` block's result channel, and drop the block
     fn send_finalized_block_error(queued: QueuedFinalized, error: impl Into<BoxError>) {
-        let (finalized, rsp_tx) = queued;
+        let (_finalized, rsp_tx) = queued;
 
         // The block sender might have already given up on this block,
         // so ignore any channel send errors.
         let _ = rsp_tx.send(Err(error.into()));
-        std::mem::drop(finalized);
```
teor2345: Nitpick: I'm not sure why this got changed, it seems more explicit the other way?

zebra-state/src/service.rs
```diff
@@ -718,98 +635,53 @@ impl StateService {
     /// Attempt to validate and commit all queued blocks whose parents have
     /// recently arrived starting from `new_parent`, in breadth-first ordering.
```
teor2345: Nitpick: Can you update this method documentation?

We might also want to rename the method so it's easy to understand what it does.
What about `send_ready_non_finalized_queued()`?

zebra-state/src/service.rs
```diff
@@ -718,98 +635,53 @@ impl StateService {
...
+                let queued_children = self
+                    .queued_non_finalized_blocks
+                    .dequeue_children(parent_hash);
+
+                for queued_child in queued_children {
+                    let child_hash = queued_child.0.hash;
+
+                    let send_result = non_finalized_block_write_sender.send(
+                        NonFinalizedWriteCmd::ProcessQueued {
+                            parent_hash,
```
teor2345: Nitpick: the `parent_hash` field is redundant, because it is the same as `queued_child.0.block.header.parent_hash` (or something like that).

zebra-state/src/service.rs
```diff
@@ -718,98 +635,53 @@ impl StateService {
...
-                .dequeue_children(parent_hash);
+                    if let Err(SendError(NonFinalizedWriteCmd::ProcessQueued {
+                        parent_hash: _,
+                        queued_child: (_, rsp_tx),
+                    })) = send_result
+                    {
+                        // If Zebra is shutting down, ignore dropped block result receivers
+                        let _ = rsp_tx.send(Err(
+                            "block commit task exited. Is Zebra shutting down?".into(),
+                        ));
```
teor2345: If the channel is closed, we also need to send errors on all the queued blocks, because they will never verify.
See `clear_finalized_block_queue()` for an example.

We also need to call this cleanup method in the `Drop` impl.
(I need to do a similar change in PR #5134.)

zebra-state/src/service.rs
```diff
@@ -618,98 +612,21 @@ impl StateService {
...
-            self.mem.commit_new_chain(prepared, &self.disk.db)?;
-        } else {
-            self.mem.commit_block(prepared, &self.disk.db)?;
-        }
-
-        Ok(())
-    }
-
     /// Returns `true` if `hash` is a valid previous block hash for new non-finalized blocks.
     fn can_fork_chain_at(&self, hash: &block::Hash) -> bool {
```
teor2345, line 688: Required Fix: `can_fork_chain_at()` is used to commit ready non-finalized blocks, but it does not currently work, because `self.mem` is always empty.

Instead:
- change `last_block_hash_sent` into a local set of the block hashes that have recently been sent, but haven't made it into the non-finalized or finalized state yet
- because it's a set, we don't need to reset `last_block_hash_sent` if there is a block commit error, but we do need to discard outdated hashes (below the finalized tip)
- use `StateService.read_service.non_finalized_state_receiver` to get a cached `NonFinalizedState`

Let me know if you'd like me to push a PR with these changes.

teor2345 (commented):
This looks great, it's a good design, and it seems like it will work.

I've marked a few reviews as a "Required Fix". They are probably the bugs causing the test failures. It might help to focus on them first.

I also made some comments about refactors that would simplify the code or improve cleanup. But they aren't as important.

Feel free to ignore the nitpicks until the sync and tests are working, some of them might get deleted or moved.

Let me know if you need help with any of these changes, or if you'd like me to push a PR for any change (or all of them).

teor2345 (changes requested):
(Just marking this so people know it has been reviewed.)

zebra-state/src/service/write.rs
arya2: That does feel like a wasted opportunity to avoid `validate_and_commit` calls when children that need to be failed aren't in the queue yet. 

I'll update it once the tests are passing since this is closer to the pre-existing logic (it's currently sending the FinishProcessingQueued message where it was previously dropping the errors of any parent blocks without queued children.)

zebra-state/src/service/write.rs
arya2: Oh certainly, this should not have been committed.

zebra-state/src/service.rs
arya2: I just noticed that it was already happening implicitly, but it's unrelated to the PR so I'll put it back.

zebra-state/src/service.rs
arya2: It looks like this is being used from the CommitBlock call before the block is queued. I think it makes more sense to update the comment to `/// Assert some assumptions about the prepared `block` before it is queued.` and leave it as a method on StateService?

arya2: Updated the comment in [`3703a2e`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/3703a2e1362c8ec59ec971023e0334de73643297), left it as a method in StateService since it's currently being used in the call fn and not the validate and commit

zebra-state/src/service/write.rs
arya2: Fixed in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service/write.rs
arya2: Fixed in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service/write.rs
arya2: It was outdated, fixed in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service/write.rs
arya2: Fixed in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service.rs
arya2: Updated the fn name and documentation in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service.rs
arya2: I probably should've noticed this, fixed in [`a27aa3f`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a27aa3fa449dd9d9bd482f0de627f349c738d3b7)

zebra-state/src/service/write.rs
arya2: Updated in [`a9d4e05`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/a9d4e0571959aa8aab067148840578445aff1147), let me know if that's right or what to change.

zebra-state/src/service/write.rs
arya2: Updated in [`76ceb6b`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/76ceb6b5f95a4b8af2d2d243327888812e3baa78) to finalize blocks past the re-org limit every time we add a block.

zebra-state/src/service/write.rs
arya2: Looks like it was possible and simpler!

zebra-state/src/service/write.rs
arya2: Updated in [`14460cf`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/14460cfc95d8f32ffda5bee8143b57b067381842) and the error message was changed again in [`6c4c7e8`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/6c4c7e82a4ed5092db43c5638f835840058816d6), please let me know if you have a better idea for the message.

zebra-state/src/service.rs
arya2: Oops. I removed `mem` and `disk` in [`64f1004`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/64f1004d470c9455a71680f0167097daddcd8949) and the three commits that followed that one.

zebra-state/src/service.rs
arya2: Good point. Fixed in [`3aad1b7`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/3aad1b7e80f413c12ff67efca8b601d858992036)

zebra-state/src/service/write.rs
```diff
@@ -1,33 +1,92 @@
...
+/// [`Chain`][1].
+///
+/// Returns the latest non-finalized chain tip height, or `None` if the
+/// non-finalized state is empty.
 ///
-/// TODO: pass the non-finalized state and associated update channel to this function
+/// [1]: non_finalized_state::Chain
+#[instrument(level = "debug", skip(chain_tip_sender))]
+fn update_latest_chain_channels(
+    finalized_state: &FinalizedState,
```
teor2345: Nitpick: This argument appears to be unused.

zebra-state/src/service/check/tests/nullifier.rs
```diff
@@ -144,15 +145,17 @@ proptest! {
...
+        finalized_state.populate_with_anchors(&block1);
 
-        let previous_mem = state.mem.clone();
+        let previous_mem = non_finalized_state.clone();
 
         let block1 = Arc::new(block1).prepare();
-        let commit_result = state.validate_and_commit(block1);
+        let commit_result = validate_and_commit_non_finalized(
+            &finalized_state,
+            &mut non_finalized_state,  block1);
```
teor2345, line 158: Replacing `state.validate_and_commit()` with `validate_and_commit_non_finalized()` reduces test coverage.

It's ok to do it to get the tests working. But once we're happy with the API and we've got everything passing, let's see if we can restore test coverage for `state.validate_and_commit()`.

zebra-state/src/service/check/tests/anchors.rs
```diff
@@ -42,7 +43,10 @@ fn check_sprout_anchors() {
 
     // Validate and commit [`block_1`]. This will add an anchor referencing the
     // empty note commitment tree to the state.
-    assert!(state.validate_and_commit(block_1).is_ok());
+    assert!(
+        validate_and_commit_non_finalized(&finalized_state, &mut non_finalized_state, block_1)
+            .is_ok()
+    );
```
teor2345, line 49: In these tests we also replace `state.validate_and_commit()` with `validate_and_commit_non_finalized()`, which reduces test coverage.

zebra-state/src/service/write.rs
teor2345: Looks good, I simplified the truncation, changed the max size, and added a constant in 0e37923 in PR #5299.

zebra-state/src/service.rs
teor2345: Thanks!

I wanted to get rid of the `mem` name, because it's a bit confusing, so I renamed it in 8f9aea5 in PR #5299.

zebra-state/src/service.rs
teor2345: I tweaked the error messages in 6e2c3b1 in PR #5299.

teor2345 (commented):
Thanks for these updates!

I pushed a PR #5299 with some suggested changes:
- Do error tasks on Err, and success tasks on Ok
- Rewrite best_tip() to use tip()
- Simplify latest_non_finalized_state() using a new WatchReceiver API
- Document why we send the result after updating the channels
- Rename a function to initial_contextual_validity()
- Fix doc link warnings

Feel free to merge that PR into this one, or ask for changes on it.

arya2:
Looks like there's a timing issue, seemingly related to the CommitFinalizedBlock call.

`components::mempool::tests::vector::mempool_cancel_mined` failed with:

>   left: `1`,
>  right: `0`
> Location: zebrad/src/components/mempool/tests/vector.rs:475

https://github.com/ZcashFoundation/zebra/actions/runs/3153382100/jobs/5129789960#step:10:3615

Same failure happened on main in the most recent commit [`75a6797`](https://github.com/ZcashFoundation/zebra/commit/75a679792bb787b95b4e9ce87aaefe205c48c97a)

https://github.com/ZcashFoundation/zebra/actions/runs/3153584855/jobs/5130226183#step:10:3674

zebra-state/src/service.rs
teor2345: Here's how we use `OrderedMap` in the address book:
https://github.com/ZcashFoundation/zebra/blob/97fb85dca9d00a717661ea0cbab7ce66b8d23977/zebra-network/src/address_book.rs#L130

https://github.com/ZcashFoundation/zebra/blob/97fb85dca9d00a717661ea0cbab7ce66b8d23977/zebra-network/src/address_book.rs#L331-L337

https://github.com/ZcashFoundation/zebra/blob/97fb85dca9d00a717661ea0cbab7ce66b8d23977/zebra-network/src/address_book.rs#L397-L403

zebra-state/src/service/check/tests/anchors.rs
teor2345: The extra test coverage from doing this is about 10 lines, so it's not worth it.

zebra-state/src/service/check/tests/nullifier.rs
teor2345: The extra test coverage from doing this is about 10 lines, so it's not worth it.

arya2:
There was a panic on the testnet sync


```
Message:  only called while blocks is populated
Location: zebra-state/src/service/non_finalized_state/chain.rs:561

Metadata:
version: 1.0.0-beta.15+121.gcbd8cdb
Zcash network: Testnet
state version: 25
branch: non-finalized-block-commit-channel
git commit: cbd8cdb
commit timestamp: 2022-09-28T23:55:45Z
build profile: debug

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: zebra_state::service::write::write_blocks_from_channels with finalized_state=FinalizedState { 
        network: Testnet, 
        debug_stop_at_height: None, 
        db: ZebraDb { 
            db: DiskDb { 
                ephemeral: false, db: RocksDB { 
                    path: "/home/ar/.cache/zebra/state/v25/testnet"
                } 
            } 
        } 
    } 
    non_finalized_state=NonFinalizedState { 
        chain_set: {}, 
        network: Testnet 
    } 
    non_finalized_state_sender=Sender { 
        shared: Shared { 
            value: RwLock(
                PhantomData, 
                RwLock { 
                    data: NonFinalizedState { 
                        chain_set: {}, 
                        network: Testnet 
                    } 
                }
            ), 
            state: AtomicState(0), 
            ref_count_rx: 2, 
            notify_rx: Notify { 
                state: 0, 
                waiters: Mutex(
                    PhantomData, 
                    Mutex { 
                        data: LinkedList { 
                            head: None, 
                            tail: None 
                        } 
                    }
                ) 
            }, 
            notify_tx: Notify { 
                state: 0, 
                waiters: Mutex(
                    PhantomData, 
                    Mutex { 
                        data: LinkedList { 
                            head: None, 
                            tail: None 
                        } 
                    }
                )
            } 
        } 
    }
at zebra-state/src/service/write.rs:76
```

teor2345:
> Looks like there's a timing issue, seemingly related to the CommitFinalizedBlock call.
> 
> `components::mempool::tests::vector::mempool_cancel_mined`

This is bug #5301, I'll increase its priority.

zebra-state/src/service.rs
arya2: Sent hashes added in [`bf125ac`](https://github.com/ZcashFoundation/zebra/pull/5257/commits/bf125ac575c4b320eb0fa521480686cb2551e584)

I used a VecDeque for pruning by height because OrderedMap doesn't have a `contain` or `get` method, and it seems low-cost to sometimes leave sent hashes that are below the finalized tip.

The check against the non-finalized chains in `can_fork_chain_at` was replaced with the check against sent hashes, since hashes aren't discarded until after they're below the finalized tip.

arya2:
Adding the non-finalized state receiver to `read_service` didn't seem to hurt performance, so I'm not too concerned about this, but I think if the watch channel is holding a reference to the chains, the Arc::make_mut calls in `NonFinalizedState.finalize()` may always find a strong count >1 and fall back on cloning all of the chain data in the non-finalized state.

That may be the case with the `Arc::try_unwrap` call on the `parent_chain` that falls back to a clone in `NonFinalizedState.validate_and_update_parallel` as well.

teor2345:
> Adding the non-finalized state receiver to `read_service` didn't seem to hurt performance, so I'm not too concerned about this, but I think if the watch channel is holding a reference to the chains, the Arc::make_mut calls in `NonFinalizedState.finalize()` may always find a strong count >1 and fall back on cloning all of the chain data in the non-finalized state.
> 
> That may be the case with the `Arc::try_unwrap` call on the `parent_chain` that falls back to a clone in `NonFinalizedState.validate_and_update_parallel` as well.

I'm not too concerned about it either, most of the large data in `Chain` is wrapped in `Arc`, inside the vectors or maps. If performance becomes a problem, we should try wrapping `Utxo` in `Arc`, it's the only potentially large data that isn't wrapped. (But it's usually small, so we'll need to benchmark if we make that change.)

zebra-state/src/service/queued_blocks.rs
```diff
@@ -209,3 +209,52 @@ impl QueuedBlocks {
...
+                self.buf.push_front((hash, height));
+                break;
+            } else {
+                self.set.remove(&hash);
+            }
+        }
+
+        // shrink to a reasonable cap in cases where there are many hashes that
+        // are not in order of their heights in the buf
+        self.buf.shrink_to(127);
```
teor2345: Nitpick: We prefer documented constants to integer literals, for code readability.

teor2345 (changes requested):
Required Fix: I think we need to check for duplicate blocks in `sent_non_finalized_block_hashes` at the top of `queue_and_commit_non_finalized`. This could be the cause of the panics in the tests.

For performance reasons, we might want to do all the memory-based duplicate checks first, then the ones that require database access.

Optional: We might want to turn the "duplicate chain tip" panic into an error. (But maybe not, because it helped us find a bug?)

Arya, it might help to set up your development environment or git commit hook to run `cargo fmt --all` when you save or commit. That will automatically fix the `rustfmt` errors in CI.

zebra-state/src/service.rs
teor2345: Optional: `sent_non_finalized_block_hashes.contains()` can return `true` for blocks that are:
- below the finalised tip
- at the finalised tip height, but not the same hash as the finalised tip block

This will cause an error when committing those blocks.

There are a few different ways we could fix this, here are some alternatives:
- make pruning reliable
- pass the finalised tip height to `contains()`, and return false for any height that is less than or equal to the finalised tip
- do nothing, and just accept that a few errors will happen

If we make these changes, there's still a race between the finalised tip and the channel, but it will happen less often. If the race does cause an error, we should make sure it's not a panic.

zebra-state/src/service/queued_blocks.rs
teor2345: Optional: There are a few alternatives here:
- use `shrink_to_fit()`
- don't shrink at all, and just keep using the extra memory - this is more efficient when we're syncing, because we don't reallocate, but less efficient when we've reached the tip
- use `shrink_to()` with a constant that's larger than the rollback limit (100) + maximum fork length (7) + channel capacity (1200 default lookahead limit)

The amount of extra memory is about 75 kB:
`(1200 look ahead limit - 100 non-finalised chain) * (32 byte hash * 2 + 4 byte height)`

Zebra typically uses 2-8 GB of memory.

teor2345: Nitpick: Whatever we do here, we should do it to the `HashSet` as well.

zebra-state/src/service.rs
arya2: I updated SentHashes to make pruning reliable in [`70beeb0`](https://github.com/ZcashFoundation/zebra/commit/70beeb07459d35940e4806342bba8daf48816d9c), it still may return true for blocks that are at the finalized tip height but not the same hash as the finalized tip block.

zebra-state/src/service/queued_blocks.rs
arya2: Oops. I updated it to call `shrink_to_fit` on the `HashSet` in [`70beeb0`](https://github.com/ZcashFoundation/zebra/commit/70beeb07459d35940e4806342bba8daf48816d9c).

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,80 @@ impl QueuedBlocks {
...
+
+#[derive(Debug, Default)]
+pub(crate) struct SentHashes {
+    /// A Vec of growable ring bufs to be used for pruning outdated hashes that are
+    /// inserted in the order of their height in a given batch.
+    sent_bufs: Vec<VecDeque<(block::Hash, block::Height)>>,
+
+    /// The growable ring buf to be used for pruning outdated hashes that are mostly
+    /// inserted in the order of their height for this batch of sent hashes
+    curr_buf: VecDeque<(block::Hash, block::Height)>,
```
teor2345: Nitpick: we try not to repeat details that are already in the code itself.

```suggestion
    /// A list of previously sent block batches, each batch is in increasing height order.
    /// We use this list to efficiently prune outdated hashes that are at or below the finalized tip.
    sent_bufs: Vec<VecDeque<(block::Hash, block::Height)>>,

    /// The list of blocks sent in the current batch, in increasing height order.
    curr_buf: VecDeque<(block::Hash, block::Height)>,
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,80 @@ impl QueuedBlocks {
...
+        }
+    }
+
+    /// Iterates over each buf in `sent_bufs`, removing hashes from SentHashes until reaching
+    /// the first hash with a height above the `height_bound`
+    ///
+    /// Assumes that hashes in each `sent_buf` will be added in order of their heights
+    /// in order to remove all hashes that are below `height_bound`.
+    pub fn prune_by_height(&mut self, height_bound: block::Height) {
+        self.sent_bufs.retain_mut(|buf| {
```
teor2345: We could accidentally misuse this API by calling `finish_batch()` after `prune_by_height()`.

```suggestion
    /// Iterates over each buf in `sent_bufs`, removing hashes from SentHashes until reaching
    /// the first hash with a height above the `height_bound`.
    ///
    /// Finishes the batch if `finish_batch()` hasn't been called already.
    ///
    /// Assumes that hashes in each `sent_buf` will be added in order of their heights
    /// in order to remove all hashes that are below `height_bound`.
    pub fn prune_by_height(&mut self, height_bound: block::Height) {
        self.finish_batch();
        
        self.sent_bufs.retain_mut(|buf| {
```

teor2345 (dismissed):
This looks good!

Let's run a full sync test before we merge, and a checkpoint sync as well?

All of my remaining suggestions are optional.

zebra-state/src/service.rs
```diff
@@ -565,9 +573,16 @@ impl StateService {
...
+        if self
+            .sent_non_finalized_block_hashes
+            .contains(&prepared.hash)
         {
+            let (rsp_tx, rsp_rx) = oneshot::channel();
+            let _ = rsp_tx.send(Err("block already sent to be committed to the state".into()));
+            return rsp_rx;
+        }
+
+        if self.read_service.db.hash(prepared.height).is_some() {
```
teor2345: I opened PR #5323 with a faster way of implementing this.

(RocksDB has a "key is present" and a "get the data" API.)

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,83 @@ impl QueuedBlocks {
...
+    /// Iterates over each buf in `sent_bufs`, removing hashes from SentHashes until reaching
+    /// the first hash with a height above the `height_bound`.
+    ///
+    /// Finishes the batch if `finish_batch()` hasn't been called already.
+    ///
+    /// Assumes that hashes in each `sent_buf` will be added in order of their heights
+    /// in order to remove all hashes that are below `height_bound`.
+    pub fn prune_by_height(&mut self, height_bound: block::Height) {
+        self.finish_batch();
+
```
teor2345: Oops, sorry I broke rustfmt

```suggestion

```

zebra-state/src/service.rs
arya2: that seems better.

teor2345:
I started a full and checkpoint sync by going to 'Actions/CI Docker', and running it manually:
Screen Shot 2022-10-04 at 09 08 37

Here is the running workflow:
https://github.com/ZcashFoundation/zebra/actions/runs/3178119035

Let's check back in about a day?

teor2345:
I just restarted my synced `zebrad` with this branch, and I got a block timeout:
```
2022-10-03T23:18:26.272444Z  INFO {zebrad="ef3e9f8" net="Main"}:init{config=Config { checkpoint_sync: true, debug_skip_parameter_preload: false } network=Mainnet debug_skip_parameter_preload=false}: zebra_consensus::chain: initializing chain verifier tip=Some((Height(1830536), block::Hash("000000000134d7c8df43537fd7faac1373d117e3e9c1faebb6d36a96d9610674"))) max_checkpoint_height=Height(1805214)
2022-10-03T23:18:26.272461Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::commands::start: initializing syncer
2022-10-03T23:18:26.272472Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::commands::start: initializing mempool
2022-10-03T23:18:26.272488Z  INFO {zebrad="ef3e9f8" net="Main"}: zebra_rpc::server: Trying to open RPC endpoint at 0.0.0.0:28232...
2022-10-03T23:18:26.272533Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::commands::start: spawned initial Zebra tasks
2022-10-03T23:18:26.272534Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::gossip: initializing block gossip task
2022-10-03T23:18:26.272541Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool::crawler: initializing mempool crawler task
2022-10-03T23:18:26.272540Z  INFO {zebrad="ef3e9f8" net="Main"}: zebra_state::config: finished old database version cleanup task
2022-10-03T23:18:26.272551Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool::gossip: initializing transaction gossip task
2022-10-03T23:18:26.272550Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool::queue_checker: initializing mempool queue checker task
2022-10-03T23:18:26.272632Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: starting sync, obtaining new tips state_tip=Some(Height(1830536))
2022-10-03T23:18:26.275303Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.995% current_height=Height(1830536) network_upgrade=Nu5 remaining_sync_blocks=95 time_since_last_state_block=0s
2022-10-03T23:18:26.305211Z  INFO {zebrad="ef3e9f8" net="Main"}: zebra_rpc::server: Opened RPC endpoint at 0.0.0.0:28232
2022-10-03T23:18:27.179475Z  INFO {zebrad="ef3e9f8" net="Main"}:init{config=Config { checkpoint_sync: true, debug_skip_parameter_preload: false } network=Mainnet debug_skip_parameter_preload=false}: zebra_consensus::chain: Groth16 pre-download and check task finished
2022-10-03T23:19:26.276929Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.997% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=62 time_since_last_state_block=0s
2022-10-03T23:20:26.278827Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.997% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=63 time_since_last_state_block=1m
2022-10-03T23:21:26.280399Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.997% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=64 time_since_last_state_block=2m
2022-10-03T23:22:26.281659Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.996% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=65 time_since_last_state_block=3m
2022-10-03T23:23:26.282543Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.996% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=65 time_since_last_state_block=4m
2022-10-03T23:24:26.283297Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.996% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=66 time_since_last_state_block=5m
2022-10-03T23:24:46.090050Z  WARN {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: error downloading and verifying block e=Invalid { error: Block(Transaction(InternalDowncastError("downcast to known transaction error type failed, original error: Elapsed(())"))), height: Height(1830583), hash: block::Hash("0000000000334454ba3daa548bf77b54735ece8e28ee9b73e480d4bb3b500a6a") }
2022-10-03T23:24:46.144306Z  INFO {zebrad="ef3e9f8" net="Main"}:sync: zebrad::components::sync: waiting to restart sync timeout=67s state_tip=Some(Height(1830578))
2022-10-03T23:25:26.284409Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.996% current_height=Height(1830578) network_upgrade=Nu5 remaining_sync_blocks=67 time_since_last_state_block=6m
2022-10-03T23:25:41.323297Z  INFO {zebrad="ef3e9f8" net="Main"}:crawl_and_dial{new_peer_interval=61s}: zebra_network::peer_set::candidate_set: timeout waiting for peer service readiness or peer responses
2022-10-03T23:25:53.145591Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: starting sync, obtaining new tips state_tip=Some(Height(1830578))
2022-10-03T23:26:00.914635Z  WARN {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: error downloading and verifying block e=Invalid { error: Block(Commit("block already sent to be committed to the state")), height: Height(1830588), hash: block::Hash("0000000000a82784cc38eb5155047eb4ef7218bba441361d3509cc54cb4ec1a7") }
2022-10-03T23:26:00.917005Z  INFO {zebrad="ef3e9f8" net="Main"}:sync: zebrad::components::sync: waiting to restart sync timeout=67s state_tip=Some(Height(1830584))
2022-10-03T23:26:26.285182Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.997% current_height=Height(1830594) network_upgrade=Nu5 remaining_sync_blocks=53 time_since_last_state_block=0s
2022-10-03T23:27:07.918808Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: starting sync, obtaining new tips state_tip=Some(Height(1830594))
2022-10-03T23:27:11.133649Z  WARN {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: error downloading and verifying block e=Invalid { error: Block(Commit("block already sent to be committed to the state")), height: Height(1830616), hash: block::Hash("00000000004da411d03906a2c60752d9b1ee23f7028b327de83083a48f565063") }
2022-10-03T23:27:11.133851Z  INFO {zebrad="ef3e9f8" net="Main"}:sync: zebrad::components::sync: waiting to restart sync timeout=67s state_tip=Some(Height(1830597))
2022-10-03T23:27:26.286034Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.998% current_height=Height(1830616) network_upgrade=Nu5 remaining_sync_blocks=37 time_since_last_state_block=0s
2022-10-03T23:28:18.135809Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: starting sync, obtaining new tips state_tip=Some(Height(1830616))
2022-10-03T23:28:26.287605Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=99.998% current_height=Height(1830616) network_upgrade=Nu5 remaining_sync_blocks=37 time_since_last_state_block=1m
2022-10-03T23:28:31.037042Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: extending tips tips.len=1 in_flight=19 extra_hashes=0 lookahead_limit=20 state_tip=Some(Height(1830623))
2022-10-03T23:28:31.340978Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: exhausted prospective tip set
2022-10-03T23:28:31.340990Z  INFO {zebrad="ef3e9f8" net="Main"}:sync: zebrad::components::sync: waiting to restart sync timeout=67s state_tip=Some(Height(1830623))
2022-10-03T23:28:31.412646Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool: activating mempool: Zebra is close to the tip tip_height=Some(Height(1830623))
2022-10-03T23:28:32.956587Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool: resetting mempool: switched best chain, skipped blocks, or activated network upgrade tip_height=Height(1830626)
2022-10-03T23:28:32.956614Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool: activating mempool: Zebra is close to the tip tip_height=Some(Height(1830626))
2022-10-03T23:28:36.413824Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool: resetting mempool: switched best chain, skipped blocks, or activated network upgrade tip_height=Height(1830636)
2022-10-03T23:28:36.413861Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool: activating mempool: Zebra is close to the tip tip_height=Some(Height(1830636))
2022-10-03T23:28:41.415867Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::mempool::gossip: sending mempool transaction broadcast request=AdvertiseTransactionIds({Witnessed(WtxId { id: transaction::Hash("d9c3da9f62a707d681b61a73cd8f24e2bddf21aca0c0e7ce203dd38d33cf14f6"), auth_digest: AuthDigest("ad3fe950acdf3fdcb7d7896e3c1ce4e72911267b779545281288f43ccaf675c6") }), Witnessed(WtxId { id: transaction::Hash("36a37d8562010d0cdedcc2c51d34510397c6a21b7190fadab1e891e235ff62f6"), auth_digest: AuthDigest("089841736a6084e2cf79fc075860c203b2460db79f7a60344c1fe03ee6e2676e") }), Witnessed(WtxId { id: transaction::Hash("578a8e17a2341861d24c2d8c8f1c1e631a80d6187697cdaa740de0faf8e60d3d"), auth_digest: AuthDigest("13f645d89d1299913e839fa21c5f2f85ad9fdb3a785ceeadaec151887f675090") })})
2022-10-03T23:29:26.288749Z  INFO {zebrad="ef3e9f8" net="Main"}: zebrad::components::sync::progress: finished initial sync to chain tip, using gossiped blocks sync_percent=99.999% current_height=Height(1830636) network_upgrade=Nu5 remaining_sync_blocks=12 time_since_last_state_block=0s
2022-10-03T23:29:38.342253Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: starting sync, obtaining new tips state_tip=Some(Height(1830636))
2022-10-03T23:29:38.664123Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: extending tips tips.len=1 in_flight=6 extra_hashes=0 lookahead_limit=20 state_tip=Some(Height(1830636))
2022-10-03T23:29:44.666403Z  INFO {zebrad="ef3e9f8" net="Main"}:sync:try_to_sync: zebrad::components::sync: exhausted prospective tip set
2022-10-03T23:29:44.666421Z  INFO {zebrad="ef3e9f8" net="Main"}:sync: zebrad::components::sync: waiting to restart sync timeout=67s state_tip=Some(Height(1830636))
```

It seems like this PR is not sending all the blocks that need to be committed, or they aren't being sent in the right order.

arya2:
> It seems like this PR is not sending all the blocks that need to be committed, or they aren't being sent in the right order.

I'm looking at [these lines](https://github.com/ZcashFoundation/zebra/pull/5257/files#diff-adfae901cefebad145d2cc719c260192358c4327859f2ab9a9f4bfa98e16ebd7R576-R583) with some suspicion, though it does panic in the testnet sync without that check.

teor2345:
> It seems like this PR is not sending all the blocks that need to be committed, or they aren't being sent in the right order.

I think I was wrong.

I added some metrics (PR #5327). It seems like we might have a race condition finding UTXOs.

Here is where we can find UTXOs:
- in the block queues (finalized or non-finalized)
- in the committed blocks (finalized or non-finalized)

Here is where we can't find UTXOs, but we need to be able to find them:
- in the block channel (non-finalized only)

If we get a UTXO request after a block has been removed from the queue, but before it has been committed to the non-finalized state, those UTXOs time out, because they are never checked again. (We only check for UTXOs when a new block or UTXO request arrives in the state.)

There are two ways to fix this:
1. re-check pending UTXOs every time a block is committed to the state
2. store UTXOs of sent blocks until they are committed

Storing more UTXOs might be the better option, because we are currently CPU-bound. And we can avoid most of the extra memory usage by wrapping the UTXO set for each block in an `Arc`.

teor2345:
@mergifyio update

teor2345:
(Let's update this PR to the latest `main` branch, just in case that fixes anything.)

teor2345:
> Storing more UTXOs might be the better option, because we are currently CPU-bound. And we can avoid most of the extra memory usage by wrapping the UTXO set for each block in an `Arc`.

I wonder if we just want to use two instances of `QueuedNonFinalizedBlocks`, rather than implementing and testing the same queue/UTXO/prune functionality twice. (I don't know if that would work, just a random idea.)

teor2345:
Coverage failed in `mempool_cancel_mined`, time for another `main` branch update.

teor2345:
@mergifyio update

arya2:
@teor2345  Thank you kindly for finding out why the timeout was happening!

I started a new Mainnet sync here: https://github.com/ZcashFoundation/zebra/actions/runs/3184782188

zebra-state/src/service.rs
```diff
@@ -1017,6 +918,16 @@ impl Service<Request> for StateService {
                     return response_fut;
                 }
 
+                // Check the sent non-finalized blocks
+                if let Some(utxo) = self.sent_non_finalized_block_hashes.utxo(&outpoint) {
+                    self.pending_utxos.respond(&outpoint, utxo);
+
+                    // We're finished, the returned future gets the UTXO from the respond() channel.
+                    timer.finish(module_path!(), line!(), "AwaitUtxo/queued-non-finalized");
```
```suggestion
                    timer.finish(module_path!(), line!(), "AwaitUtxo/sent-non-finalized");
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,104 @@ impl QueuedBlocks {
...
+    /// may not be in the finalized state yet.
+    sent: HashMap<block::Hash, Vec<transparent::OutPoint>>,
+
+    /// Known UTXOs.
+    known_utxos: HashMap<transparent::OutPoint, transparent::Utxo>,
+}
+
+impl SentHashes {
+    /// Inserts the block's utxos in `known_utxos` by outpoints, it's outpoints into `sent` by the block hash,
+    /// and pushes (block.hash, block.height) onto the current buf to be used for pruning outdated hashes.
```
teor2345: Nitpick: Comments usually focus on why the code exists, and a summary of what it does. We avoid exposing internal implementation details like private field names in public documentation.

```suggestion
    /// Stores the `block`'s hash, height, and UTXOs, so they can be used to check if a block or UTXO
    /// is available in the state.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,104 @@ impl QueuedBlocks {
...
+        self.sent.insert(block.hash, outpoints);
+    }
+
+    /// Try to look up this UTXO in any sent block.
+    #[instrument(skip(self))]
+    pub fn utxo(&self, outpoint: &transparent::OutPoint) -> Option<transparent::Utxo> {
+        self.known_utxos.get(outpoint).cloned()
+    }
+
+    /// Replaces `curr_buf` with an empty VecDeque and pushes the previous `curr_buf` to `sent_bufs`
```
teor2345: Nitpick: same as above.

```suggestion
    /// Finishes the current block batch, and stores it for efficient pruning.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,104 @@ impl QueuedBlocks {
...
+
+    /// Known UTXOs.
+    known_utxos: HashMap<transparent::OutPoint, transparent::Utxo>,
+}
+
+impl SentHashes {
+    /// Inserts the block's utxos in `known_utxos` by outpoints, it's outpoints into `sent` by the block hash,
+    /// and pushes (block.hash, block.height) onto the current buf to be used for pruning outdated hashes.
+    ///
+    /// Assumes that hashes are added in the order of their height between `finish_batch` calls
```
teor2345:
```suggestion
    /// Assumes that blocks are added in the order of their height between `finish_batch` calls
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,104 @@ impl QueuedBlocks {
...
+        }
+    }
+
+    /// Iterates over each buf in `sent_bufs`, removing hashes from SentHashes until reaching
+    /// the first hash with a height above the `height_bound`.
+    ///
+    /// Finishes the batch if `finish_batch()` hasn't been called already.
+    ///
+    /// Assumes that hashes in each `sent_buf` will be added in order of their heights
+    /// in order to remove all hashes that are below `height_bound`.
```
teor2345: Nitpick: same as above.

```suggestion
    /// Prunes sent blocks at or below `height_bound`.
    ///
    /// Finishes the batch if `finish_batch()` hasn't been called already.
    ///
    /// Assumes that blocks will be added in order of their heights between each `finish_batch()` call,
    /// so that blocks can be efficiently and reliably removed by height.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -196,4 +196,104 @@ impl QueuedBlocks {
...
+    pub fn prune_by_height(&mut self, height_bound: block::Height) {
+        self.finish_batch();
+        self.bufs.retain_mut(|buf| {
```
teor2345, line 300:
```suggestion

        // Iterates over each buf in `sent_bufs`, removing sent blocks until reaching
        // the first block with a height above the `height_bound`.
        self.bufs.retain_mut(|buf| {
```

teor2345 (commented):
I made some optional suggestions to improve the comments and metrics.

I'm happy for anyone to approve this PR as long as the full sync passes.

arya2:
The full Testnet sync succeeded locally.

Sharing the known_utxos field seems less important now that I'm looking at the end-product (https://github.com/ZcashFoundation/zebra/pull/5339), we can potentially close that PR. I replaced the panic on duplicate hashes in QueuedBlocks with an early return here too and am running another full Testnet sync locally.

teor2345:
> Sharing the known_utxos field seems less important now that I'm looking at the end-product (#5339), we can potentially close that PR

I agree it's a low priority (and it's not a release blocker). But we might still want to do it to simplify the state service implementation. That would also reduce the amount of code we have to test.

I have some more ideas for simplifications on top of PR #5339, I'll push another PR when I get time.

teor2345 (commented):
We're waiting for #5330 to merge, then we can run a full mainnet sync.
As long as that goes well, I'm happy for anyone to approve this PR.
(Please don't block on me being available!)

arya2:
Testnet sync succeeded locally again.

Updated workflow for the Mainnet sync now that https://github.com/ZcashFoundation/zebra/pull/5330 has merged: https://github.com/ZcashFoundation/zebra/actions/runs/3190092327

arya2:
@Mergifyio update

teor2345:
> I just restarted my synced `zebrad` with this branch, and I got a block timeout:

I re-did this local test, and it worked well.

teor2345:
@mergifyio update

arya2:
> Testnet sync succeeded locally again.
> 
> Updated workflow for the Mainnet sync now that #5330 has merged: https://github.com/ZcashFoundation/zebra/actions/runs/3190092327

[Zebra tip / Log full-sync-to-tip test (end)](https://github.com/ZcashFoundation/zebra/actions/runs/3190092327/jobs/5222573303) timed out.

I'm running a full Mainnet sync locally.

teor2345:
> > Testnet sync succeeded locally again.
> > Updated workflow for the Mainnet sync now that #5330 has merged: https://github.com/ZcashFoundation/zebra/actions/runs/3190092327
> 
> [Zebra tip / Log full-sync-to-tip test (end)](https://github.com/ZcashFoundation/zebra/actions/runs/3190092327/jobs/5222573303) timed out.

And we can't find out why, because PR #5330 stopped showing test output when running `ssh` commands.
I opened PR #5358 to see if we can fix that.

teor2345:
> [Zebra tip / Log full-sync-to-tip test (end)](https://github.com/ZcashFoundation/zebra/actions/runs/3190092327/jobs/5222573303) timed out.

There are a few possible reasons for this:
* bug #5125, fixed in my commits 29a8ada and 56cfa72 in this PR
* our checkpoints are out of date and need to be updated (I'll open another PR)
* the default checkpoint lookahead limit is too high, or the block timeout is too low, or the AboveLookaheadHeightLimit check is too low

teor2345:
I also merged PR #5358 into this PR, so I can run a full mainnet sync:
https://github.com/ZcashFoundation/zebra/actions/runs/3201662164

teor2345:
I made a few syncer tweaks, hopefully this full sync will pass:
https://github.com/ZcashFoundation/zebra/actions/runs/3201818199

teor2345:
I did a mainnet sync locally, and it still had a block time out right after we finished checkpointing. I think we might need to check a few more blocks for UTXOs, I just pushed commit 6551a74.

teor2345:
Let's try a full sync again, with the latest UTXO, syncer, and GitHub actions changes:
https://github.com/ZcashFoundation/zebra/actions/runs/3203613563

arya2:
> Let's try a full sync again, with the latest UTXO, syncer, and GitHub actions changes: https://github.com/ZcashFoundation/zebra/actions/runs/3203613563

There were SSH failures, I started another here: https://github.com/ZcashFoundation/zebra/actions/runs/3206336794

teor2345:
We're still seeing a hang after verifying the final checkpoint, we need to fix that bug before we merge this PR.

> 2022-10-08T18:10:12.430218Z  INFO {net="Main"}:sync:try_to_sync:extend_tips:checkpoint: zebra_consensus::checkpoint: verified final checkpoint: starting full validation final_checkpoint_height=Height(1805214)
...
> 2022-10-08T18:39:52.199233Z  WARN {net="Main"}: zebrad::components::sync::progress: chain updates have stalled, state height has not increased for 13 minutes. Hint: check your network connection, and your computer clock and time zone sync_percent=98.307% current_height=Height(1805214) network_upgrade=Nu5 time_since_last_state_block=13m target_block_spacing=PT75S max_block_spacing=None is_syncer_stopped=false
> 2022-10-08T18:40:04.782992Z  WARN {net="Main"}:sync:try_to_sync: zebrad::components::sync: error downloading and verifying block e=ValidationRequestError { error: Elapsed(()), height: Height(1805215), hash: block::Hash("00000000008b4c288a9bbbd87ed95e7dd3e5d65d445deea4b5eb4caa70ae2134") }

https://github.com/ZcashFoundation/zebra/actions/runs/3206336794/jobs/5250967839#step:6:2115

In this case, we needed to find out what block `1805215` was waiting for before it verified.

I suspect it was some UTXOs:
https://github.com/ZcashFoundation/zebra/blob/ee0edef8571d3ed82848303b142c7ae72e7135fc/zebra-consensus/src/transaction.rs#L485

But it could have been something in the state non-finalized block commit request:
https://github.com/ZcashFoundation/zebra/blob/ee0edef8571d3ed82848303b142c7ae72e7135fc/zebra-consensus/src/block.rs#L276

It's very unlikely to be this Depth call:
https://github.com/ZcashFoundation/zebra/blob/ee0edef8571d3ed82848303b142c7ae72e7135fc/zebra-consensus/src/block.rs#L140

I'll modify my metrics branch to show what is happening here.

teor2345:
> I'll modify my metrics branch to show what is happening here.

I didn't get to this today, feel free to have a go at it!

teor2345:
I compared the full sync before my sync tweaks:
https://github.com/ZcashFoundation/zebra/actions/runs/3190092327

With the full sync after the tweaks:
https://github.com/ZcashFoundation/zebra/actions/runs/3190092327

The sync after is about 25 minutes faster, and the jobs for recent blocks are faster than they used to be. So I think I'll keep most of the changes. (They should also reduce RAM usage.)

teor2345:
I added a workaround for #5125 in commit 74b29be.

That should be it for this PR, we just need to merge the checkpoint PR and get the full sync working.

teor2345:
I'm running a full sync with the new checkpoints here:
https://github.com/ZcashFoundation/zebra/actions/runs/3223410431

If any of the sync jobs time out, we can fix the workflows, but that doesn't need to block this PR merging.
(The timeouts could be from PR #5360 or this PR, but we want to wait to fix them until we have the final times from both.)

I'm also running a full mainnet and testnet sync locally.

zebrad/src/components/sync/downloads.rs
```diff
@@ -418,9 +436,17 @@ where
...
-                    .call(block);
+                    .call(block).boxed();
+
+                // Add a shorter timeout to workaround a known bug (#5125)
+                let short_timeout_max = (max_checkpoint_height + FINAL_CHECKPOINT_BLOCK_VERIFY_TIMEOUT_LIMIT).expect("checkpoint block height is in valid range");
+                if block_height >= max_checkpoint_height && block_height <= short_timeout_max {
+                    rsp = timeout(FINAL_CHECKPOINT_BLOCK_VERIFY_TIMEOUT, rsp)
+                        .map_err(|timeout| format!("initial fully verified block timed out: retrying: {:?}", timeout).into())
+                        .map(|nested_result| nested_result.and_then(convert::identity)).boxed();
+                }
```
teor2345, line 449: @arya2 or @upbqdn, once you're happy with the changes in commit 74b29be, you can resolve this comment, and the PR will start merging.

(I'm doing it this way because Arya can't approve PRs he has opened.)

teor2345 (approved):
I think the code is complete here, and we can do any other fixes in separate PRs.

arya2:
Failed to build `zebra-test` in the merge queue: https://github.com/ZcashFoundation/zebra/actions/runs/3228683588/jobs/5285115640

arya2:
@Mergifyio update

PR #5292, opened 2022-09-28
ci(deploy): retry `gcloud` ssh connections if it fails

gustavovalverde:
## Previous behavior
From time to time SSH connections to deployed VMs fails with the following error: `kex_exchange_identification: Connection closed by remote host`

## Expected behavior
If the connection fails, attempt to reconnect once again (or multiple times)

## Solution
Add the `ConnectionAttempts` and `ConnectTimeout` with 20 and 5 values respectively, which attempst to reconnect 19 more times every 5 seconds

## Review

Anyone can review this

### Reviewer Checklist

## Follow Up Work

If this issue keeps raising, even after the retries, we might want to apply some best SSH practices for Google Cloud (which we might want to apply anyways) as stated here https://cloud.google.com/compute/docs/instances/connecting-advanced

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -161,6 +161,8 @@ jobs:
           --zone ${{ env.ZONE }} \
           --quiet \
           --ssh-flag="-o ServerAliveInterval=5" \
+          --ssh-flag="-o ConnectionAttempts=20" \
+          --ssh-flag="-o ConnectTimeout=5" \
           --command \
           "\
           sudo mkfs.ext4 -v /dev/sdb \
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks, this seems like it might work, I guess we have to merge to find out.

Next time we need to change these parameters, let's put the settings in one place using a SSH config file or workflow environmental variables?
That way we don't have to modify the workflow in 10 places for the same change.

gustavovalverde:
> Next time we need to change these parameters, let's put the settings in one place using a SSH config file or workflow environmental variables?

Yeah, I think we should even see how to accomplish a DRY(er) approach on this workflows.

teor2345:
> > Next time we need to change these parameters, let's put the settings in one place using a SSH config file or workflow environmental variables?
> 
> Yeah, I think we should even see how to accomplish a DRY(er) approach on this workflows.

Another thing we can do is move workflow and Docker scripts into `.sh` files.

PR #5299, opened 2022-09-29
Suggested changes for "Write non-finalized blocks to the state in a separate thread"

teor2345:
## Motivation

Here are some suggestions from my review of PR #5257.
See that PR for details.

## Review

Arya can review and merge this PR into PR #5257.
(I don't think we want to keep it as a separate PR, that would cause merge conflicts.)


arya2 (approved):
Thank you Teor!

PR #5164, opened 2022-09-14
fix(test): Wait for zebrad and lightwalletd to reach the tip in tests, to improve test coverage

oxarbitrage:
## Motivation

We need to make sure lightwalletd and zebrad is at the tip for several tests. Fix https://github.com/ZcashFoundation/zebra/issues/4894 if merged.

Depends-On: #5307

## Solution

Add a new function and move common code to check if we are at the tip. Call it from tests. We are calling this function from all tests that call `lightwalletd_integration_test` and from `wallet_grpc_test` and `send_transaction_test`.

Also, the `lightwalletd_test_suite` now runs `wallet_grpc_test` before `send_transaction_test` so we know cache is up to date when running the tests using the suite.

Functional changes:
- use the same code to configure, launch, and wait for `zebrad` and `lightwalletd` across all 8 RPC integration tests,
  this makes all these tests check a bunch of things the same way 
- run the quick tests first in the test suite, then the slow tests

Refactors:
- make some test functions easier to call

## Review

This PR is ready for review.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?
    - [ ] Do the existing tests run, or are we accidentally skipping them? Check the test time and test logs.

PR #5309, opened 2022-10-02
fix(ci): Reduce false positives for the `C-trivial` label

upbqdn:
## Motivation

`release-drafter` erroneously labeled PR #5164 as `C-trivial`. It did so because the PR edited the file `zcash-rpc-diff` in `zebra-utils`.

## Solution

This PR deletes `zebra-utils` from the confing file for `release-drafter`.

## Review

We can't test this PR until we merge it to `main`.

### Reviewer Checklist

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -55,7 +55,6 @@ autolabeler:
       - 'CHANGELOG.md'
       - 'zebra-consensus/src/checkpoint/*-checkpoints.txt'
       # Developer-only changes
-      - 'only zebra-utils'
       - '.gitignore'
       # Test-only changes
       - 'zebra-test'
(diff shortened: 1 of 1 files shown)


.github/release-drafter.yml
```diff
@@ -55,7 +55,7 @@ autolabeler:
       - 'CHANGELOG.md'
       - 'zebra-consensus/src/checkpoint/*-checkpoints.txt'
       # Developer-only changes
-      - 'zebra-utils'
+      - 'only zebra-utils'
```
teor2345: I'm not sure if there is any way to say "only zebra-utils" in the regex syntax:
https://github.com/release-drafter/release-drafter#autolabeler

Maybe we'll just have to delete it.

```suggestion
```

upbqdn: Oh, I must've misunderstood one of your previous comments.

teor2345 (approved):

PR #5313, opened 2022-10-03
fix(docker): Make default command work in docker images, disable optional listener ports

teor2345:
## Motivation

1. The default Zebra docker command fails because the path variables are not substituted
2. The Zebra docker image opens world-readable metrics and logging ports by default, this is insecure

### Specifications

The variable substitution is undocumented:
https://docs.docker.com/engine/reference/run/#cmd-default-command-or-options

But there is some advice here:
https://stackoverflow.com/questions/39989045/dockerfile-cmd-not-accepting-variables-for-substitution/39989087#39989087

## Solution

- Use raw strings rather than a JSON array
- Disable optional ports

## Review

This is important because it is a security issue in a released Docker image.

We need to manually test the image created by this PR before merging:
```sh
docker run us-docker.pkg.dev/zealous-zebra/zebra:{commit-hash}
```

### Reviewer Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -148,27 +148,35 @@ ENV ZEBRA_CONF_FILE ${ZEBRA_CONF_FILE}
 
 # Build the `zebrad.toml` before starting the container, using the arguments from build
 # time, or using the default values set just above. And create the conf path and file if
-# it does not exist
+# it does not exist.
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
LGTM!

Thanks for this!

PR #5322, opened 2022-10-03
fix(test): resolve a timing issue in mempool test

arya2:
## Motivation

To fix a mempool test that's failing occasionally.

Closes #5301 

## Solution

- Await a `wait_for_tip_change` call before sending dummy calls to mempool for cancelling the transaction download

## Review

Part of regular scheduled work

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [x] How do you know it works? Does it have tests?

PR #5347, opened 2022-10-05
fix(test): resolve a timing issue in mempool test

### LINKED ISSUE #5301: Occasional failures in mempool_cancel_mined test
## Motivation

The mempool_cancel_mined test sometimes fails with:
> assert_eq!(mempool.tx_downloads().in_flight(), 0);
>
>  left: `1`,
> right: `0`
> Location: zebrad/src/components/mempool/tests/vector.rs:475

https://github.com/ZcashFoundation/zebra/actions/runs/3148267685/jobs/5118618631#step:10:3684

This is probably caused by the timing changes in PR #5134.

arya2:
## Motivation

To fix a mempool test that's failing occasionally.

The prior `mempool.dummy_call`s should also see the chain tip changes, and it's possible for the wait_for_tip_change call after the third block commit in this test to see the tip change from the first or second block commits.

See https://github.com/ZcashFoundation/zebra/pull/5322, closes https://github.com/ZcashFoundation/zebra/issues/5301

## Solution

Await `wait_for_tip_change` after every block commit.

## Review

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/mempool/tests/vector.rs
+++ b/zebrad/src/components/mempool/tests/vector.rs
@@ -417,6 +417,21 @@ async fn mempool_cancel_mined() -> Result<(), Report> {
         .await
         .unwrap();
 
+    // Wait for the chain tip update
+    if let Err(timeout_error) = timeout(
+        CHAIN_TIP_UPDATE_WAIT_LIMIT,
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks!

PR #5349, opened 2022-10-06
test(state): Test that Zebra resets correctly when committing blocks fails

### LINKED ISSUE #2654: Test if checkpoint verifier and state service are correctly reset on block commit errors
## Motivation

We want to test that errors in the state are handled correctly.

## Design

We want to do these tests here:
- when committing checkpointed blocks, fail a finalized state block commit
    - commit_finalized
- when committing fully verified blocks, fail a non-finalized state block commit
    - commit_block
    - commit_new_chain
- drop finalized blocks in the queue
- drop non-finalized blocks in the queue

If these tests take longer than a day to write, we're going to close this ticket as "too hard for now".

## Old Description

<details>

https://github.com/ZcashFoundation/zebra/pull/2633/ adds ZIP-221/244 block commitment validation in the checkpoint verifier. The check is actually done in the finalized state, and when that happens, the checkpoint verifier is reset to the tip of the finalized state in order to keep running correctly.

However, this hasn't been tested, since it is tricky to test.

Implement a test to verify that it's working correctly.

(Note that is only required if/when we have a checkpoint post-Nu5. Before that all block contents are committed to by the block hash, so they must be correct. Nu5-onward that no longer applies due to ZIP-244.)

### Previous attempt

I attempted to write this test in #2633 but gave up because it was too difficult and we had to keep working on other stuff. 
See [discussion](https://github.com/ZcashFoundation/zebra/pull/2633/files#r689198985). However, most of the problems were caused by trying to test it along with the "fake activation heights" feature in that PR, which would require generating fake blocks with correct difficulties, etc. in order to pass the checkpoint verifier. Another alternative would be to use real blocks similar to the `continuous_blockchain` - but the problem would then be how to ensure that one block fails to be comitted in order to test if that's handled correctly (the real blocks have small heights, and thus they don't use the block commitment in the header, so there is nothing to corrupt)

</details>

teor2345:
## Motivation

We want to test if the checkpoint verifier, state queues, and block write task are correctly reset on errors.

Closes #2654

## Designs

We might want to randomly fail some blocks in some tests, at low probability (1 in 1000 to 1 in 100,000).

We might want to cause errors the first time we reach a height, then validate the block properly the second time.

## Solution

- Pick some places where we can cause errors at random

## Review

This PR is incomplete.

### Reviewer Checklist


teor2345:
@upbqdn you might want to keep on working on this, I didn't get as far as I thought!

teor2345:
@mergifyio update

teor2345:
Automating these tests is out of scope, so I'm going to close this PR now.

PR #5365, opened 2022-10-09
fix(ci): Increase the Google Cloud instance sshd connection limit

teor2345:
## Motivation

SSH connections to Google Cloud test instances are failing with:
> ERROR: (gcloud.compute.start-iap-tunnel) Error while connecting [4010: 'destination read failed'].
> kex_exchange_identification: Connection closed by remote host

https://github.com/ZcashFoundation/zebra/actions/runs/3203311358/jobs/5234000346#step:6:106

Closes #5362

### Reference

> Weirdly, none actually try to authenticate to open a session.
>
> Some spiders and services like [Shodan](https://www.shodan.io/) scans public ipv4 addresses for open services, e.g. salt masters, ftp servers, RDPs, and also SSH services. These spiders usually only connect to the services without doing any valid authentication steps.

https://serverfault.com/questions/1015547/what-causes-ssh-error-kex-exchange-identification-connection-closed-by-remote/1015554#1015554

> I got this error when using docker command with remote host
> docker -H ssh://user@server compose up
> after some digging i found on my remote server in auth logs (/var/log/auth.log) this:
> Aug  8 14:51:46 user sshd[1341]: error: beginning MaxStartups throttling
> Aug  8 14:51:46 user sshd[1341]: drop connection #10 from [some_ip]:32992 on [some_ip]:22 past MaxStartups

https://stackoverflow.com/questions/67000681/kex-exchange-identification-connection-closed-by-remote-host/73292521#73292521

> MaxStartups
> Specifies the maximum number of concurrent unauthenticated connections to the SSH daemon.
> Additional connections will be dropped until authentication succeeds or the LoginGraceTime
> expires for a connection.

https://man7.org/linux/man-pages/man5/sshd_config.5.html

## Solution

Allow up to 500 unauthenticated connections to Google Cloud instances.
This tolerates a larger number of port scanning bots.

## Review

These errors are blocking most other PR merges.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

If this change works, we should:
- revert the change to the `ssh-compute` action in PRs #5330, #5358, and #5365, because they make test logging much worse
- modify the debian Docker image we use for Google Cloud instances, rather than editing the file after the instance has started using a SSH connection

PR #5358, opened 2022-10-06
fix(ci): Show logs after each Google Cloud test job

teor2345:
## Motivation

PR #5330 stopped showing any test logs in Google Cloud test runs that use the `ssh-compute` action.

But we need those logs to diagnose test failures.

## Solution

Add a step that shows the logs after each time we run `ssh-compute`

## Review

This PR is urgent because we need to be able to diagnose google cloud failures.

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

Issue #5085, opened 2022-09-06
Increase GCP disk size to 300 GB


teor2345:
## Motivation

Some GCP disk images are 160 GB, which means they could get to the current 200 GB size soon.

### Designs

We probably want to do this using `fastmod` or `sed`, because disk sizes are specified in a few places in a few different workflows.

teor2345:
We should do this soon to avoid CI failures.

They aren't happening yet, but they might happen soon on lightwalletd CI.

PR #5367, opened 2022-10-10
ci(ssh): revert using `ssh-compute` action & increase sshd connection limit

### LINKED ISSUE #639: [BUG] Can't get all changed files after pushing new commit
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does this issue exist in the latest version?

- [X] I'm using the latest release

### Describe the bug?

I want to get all changed files over entire PR after each push event (after adding new commit as well). But i always get changed files from the last commit only. I guess, that i am doing something wrong, so, could you please direct me in a right way

### To Reproduce

name: Echo Pull Request

on:
  push:
  pull_request:

jobs:
  echo:
    runs-on: ubuntu-latest
    name: Echo Changed Files
    steps:
      - uses: actions/checkout@v3
        with:
          fetch-depth: 0

      - name: Get changed files
        id: changed-files
        uses: tj-actions/changed-files@v29

      - name: List all changed files
        run: |
          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
            echo "$file was changed"
          done

### What OS are you seeing the problem on?

ubuntu-latest or ubuntu-20.04

### Expected behavior?

Expect to iterate over all changed files in a PR

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### LINKED ISSUE #5361: Show more than 1 MB of google cloud test logs
## Motivation

In PR #5358, we limited Google Cloud jobs to the last 1MB of test logs, to avoid a GitHub Actions output limit:
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-docker-container-and-javascript-actions

### Designs

We could use log artifacts:
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

Or modify the `ssh-compute` GitHub action so it shows the SSH output:
https://github.com/google-github-actions/ssh-compute/issues/29

### LINKED ISSUE #5362: Google Cloud ssh-compute action: "destination read failed"
## Motivation

> ERROR: (gcloud.compute.start-iap-tunnel) Error while connecting [4010: 'destination read failed'].
kex_exchange_identification: Connection closed by remote host

https://github.com/ZcashFoundation/zebra/actions/runs/3203311358/jobs/5234000346#step:6:106

## Alternatives

We could ask for a retry option on https://github.com/google-github-actions/ssh-compute

We could manually retry the ssh step ourselves.

If we can't fix these errors, we might want to revert all of PRs #5330 and #5358.

gustavovalverde:
## Motivation

We've been having multiple errors (more than before) after implementing [ssh-compute](https://github.com/google-github-actions/ssh-compute)

## Solution
- Revert the `ssh-compute` implementation
- Stop using IAP for TCP forwarding as there are [known limitations](https://cloud.google.com/iap/docs/using-tcp-forwarding) which might be counterproductive for our use cases
- Keep using `sudo` in some docker commands as we might be connecting with a non-root user to the VMs
- Set the `MaxStartups` setting in sshd to 500 to avoid a failing SSH connection causes GitHub Action to fail

**Note:** `tj-actions/changed-files` file comparison was failing in this PR and it has been failing in the `main` branch, so we're adding a fix here too as this would halt the PR from merging. Here's the actual explanation on this fix https://github.com/tj-actions/changed-files/issues/639#issuecomment-1256442729

Closes #5358
Closes #5365
Fixes #5362
Fixes #5361

## Review

If CI passes, anyone can review this PR

### Reviewer Checklist

    -  **Tested** I've tested this is working doing a manual SSH into the VM and confirming the file has changed with the actual configuration being used in the script file `gcp-vm-startup-script.sh`
    
## Follow Up Work

We might also want to wait up to 90 seconds after a VM has been created, just so we're sure all configurations are complete.


.github/workflows/scripts/gcp-vm-startup-script.sh
```diff
@@ -0,0 +1,22 @@
...
+echo 'MaxStartups 500' | sudo tee --append /etc/ssh/sshd_config \
+|| \
+(echo "updating instance sshd config failed: failing test"; exit 1)
+sudo grep MaxStartups /etc/ssh/sshd_config
+echo 'Modified config:'
+sudo cat /etc/ssh/sshd_config
+echo
+sudo systemctl reload sshd.service
+echo
+ps auxwww | grep sshd
```
teor2345: Optional: I think we might need a new line at the end of the script file, some tools complain when it isn't there

```suggestion
ps auxwww | grep sshd
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -150,6 +147,7 @@ jobs:
           --machine-type ${{ env.MACHINE_TYPE }} \
           --scopes cloud-platform \
           --metadata=google-monitoring-enabled=true,google-logging-enabled=true \
+          --metadata-from-file=startup-script=.github/workflows/scripts/gcp-vm-startup-script.sh \
```
teor2345, line 150: Optional: I tested this script on the Debian bullseye image. Since we're not seeing any of the output from the script, we can't be sure that it worked.

So I'd prefer to use the bullseye image for the instance, and I'll also add a check that it worked below.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -158,17 +156,21 @@ jobs:
...
+          gcloud compute ssh \
+          github-service-account@${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
+          --zone ${{ env.ZONE }} \
+          --quiet \
+          --ssh-flag="-o ServerAliveInterval=5" \
+          --ssh-flag="-o ConnectionAttempts=20" \
+          --ssh-flag="-o ConnectTimeout=5" \
+          --command \
+          "\
+          sudo mkfs.ext4 -v /dev/sdb \
```
teor2345, line 169: Let's check the sshd fix actually worked

```suggestion
          ps auxwww | grep sshd;
          sudo mkfs.ext4 -v /dev/sdb \
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -373,19 +381,23 @@ jobs:
...
+          gcloud compute ssh \
+          github-service-account@${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
+          --zone ${{ env.ZONE }} \
+          --quiet \
+          --ssh-flag="-o ServerAliveInterval=5" \
+          --ssh-flag="-o ConnectionAttempts=20" \
+          --ssh-flag="-o ConnectTimeout=5" \
+          --command \
+          "\
+          sudo e2fsck -v -f -p /dev/sdb \
```
teor2345, line 394: And also check for the cached state workflow

```suggestion
          ps auxwww | grep sshd;
          sudo e2fsck -v -f -p /dev/sdb \
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -373,19 +381,23 @@ jobs:
       # (The cached state and disk are usually the same size,
       # but the cached state can be smaller if we just increased the disk size.)
```
teor2345, line 382:
```suggestion
      # but the cached state can be smaller if we just increased the disk size.)
      #
      # Also checks that the sshd connection limit adjustment worked.
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -158,17 +156,21 @@ jobs:
       #
       # SSH into the just created VM, and create a docker volume with the newly created disk.
```
teor2345, line 157:
```suggestion
      # SSH into the just created VM, and create a docker volume with the newly created disk.
      #
      # Also checks that the sshd connection limit adjustment worked.
```

teor2345 (changes requested):
Looks great!

I'd like to check that the sshd connection limit adjustment actually worked before we merge this, since we're calling the script a different way now.

We might also want to use the bullseye image for the instances.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -362,6 +369,7 @@ jobs:
           --machine-type ${{ env.MACHINE_TYPE }} \
```
teor2345, line 369: Optional: We might also want to change to bullseye here.

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: I SSHed into the VM to confirm this change was applied, but using Debian should not hurt so I'm adding it here: 
[chore: update our VM image to bullseye](https://github.com/ZcashFoundation/zebra/pull/5367/commits/ce80c3af7e36bde0f4cb51ef12106c9cd65cb6be)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: I'm not completely sure if we should leave this troubleshooting commands in the pipelines, as this just to confirm it's working the first time. The `sshd` location is standard and it's very unusual for OpenSSH to make changes to this file.

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: https://github.com/ZcashFoundation/zebra/pull/5367#discussion_r991615060

gustavovalverde:
Sorry for adding [chore: fix tj-actions/changed-files file comparison](https://github.com/ZcashFoundation/zebra/pull/5367/commits/6ee879191a78189662a1e1b0835354e7b919466e) to this PR. But it's annoying having an ❌ in this PR and this will be pulled to `main`, fixing it here is minimal.

Edit: I had to fix it here anyways, as it wouldn't merge otherwise

gustavovalverde:
CI is failing with a new (unrelated?) error 🥲 

```
e2fsck: Cannot continue, aborting.
/dev/sdb is in use.
```

.github/workflows/deploy-gcp-tests.yml
teor2345: Thanks for letting me know, can you mention any manual checks you've done in the PR description next time?

.github/workflows/lint.yml
```diff
@@ -35,13 +35,12 @@ jobs:
       - uses: actions/checkout@v3.1.0
         with:
           persist-credentials: false
-          fetch-depth: 2
+          fetch-depth: 0
```
teor2345, line 38: Are you sure about this?
How does it work out the changed files with only one commit?

teor2345:
> CI is failing with a new (unrelated?) error 🥲
> 
> ```
> e2fsck: Cannot continue, aborting.
> /dev/sdb is in use.
> ```

We've changed two things related to the disk in this PR:
- bullseye image, which worked in PR #5365: https://github.com/ZcashFoundation/zebra/actions/runs/3216898611/jobs/5263285350#step:11:103
- moving the script to the image creation, which is new

Here are some things we could try:
- checking the server logs
- find out what is using the disk, by calling `lsof` when `e2fsck` or `resize2fs` fail 
- moving `e2fsck` and `resize2fs` to the startup script

Is there anything else we could try?

.github/workflows/lint.yml
gustavovalverde: Here's the explanation why we need to use `fetch-depth: 0` for our use case. Is not a great UX, but this solves our issue (it compares the whole PR against `main` instead of doing it with the previous `sha`...which we prefer anyways) https://github.com/tj-actions/changed-files/issues/639#issuecomment-1256442729

teor2345: Thanks, can you add a comment or some documentation explaining that, and linking to the ticket?

It doesn't have to be in this PR. But I think it's worth doing because these errors have been confusing for a while.

teor2345:
> Is there anything else we could try?

- Using `ssh` to adjust the `sshd` config, like PR #5365

That would only work if the instance startup script is causing the disk to be used.

teor2345:
@gustavovalverde I just realised that these disk resize commands are only needed when we change the disk size. 

Specifically, they are only needed between:
* when we merge the change to `main`, and
* when we stop using cached states with the smaller size (after the new larger full sync image is generated, or we've created new updated cached states with resized disks).

So let's ignore the failure for now, and fix it if it becomes a problem after #5085 ?

gustavovalverde:
> So let's ignore the failure for now, and fix it if it becomes a problem after #5085 ?

If a re-run fixes the issue, we can ignore it for now. I'm anyways testing a PR here to handle this repartitioning: https://github.com/ZcashFoundation/zebra/pull/5371

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -373,19 +381,23 @@ jobs:
...
+          --quiet \
+          --ssh-flag="-o ServerAliveInterval=5" \
+          --ssh-flag="-o ConnectionAttempts=20" \
+          --ssh-flag="-o ConnectTimeout=5" \
+          --command \
+          "\
+          sudo e2fsck -v -f -p /dev/sdb \
+          && \
+          sudo resize2fs -p /dev/sdb \
+          && \
```
teor2345, line 397: If resizing still fails, here is a fix:

```suggestion
          || \
          echo 'resizing disk failed: ignoring this error, because resizing is only needed when CI fails because the disk is full'
```

teor2345 (approved):
Thanks, looks good, let's get it fixed!

.github/workflows/lint.yml
gustavovalverde: I've created an issue to do it in a different PR so we don't reset the tests here, as we need it merge. https://github.com/ZcashFoundation/zebra/issues/5373

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: I'll follow-up with this in https://github.com/ZcashFoundation/zebra/pull/5371

PR #5370, opened 2022-10-10
ci(ssh): avoid ssh timeouts from the server

gustavovalverde:
## Motivation
We've been trying multiple solutions to our SSH connection issues, our last try solving this issues was PR https://github.com/ZcashFoundation/zebra/pull/5367/files

Depends-On: #5367

> This is an optional PR if #5367 fails to solve the issue

## Expected behavior
An SSH connection should not be terminated by the server, the connection must be kept alive indefinitely until it's killed by GitHub Actions

## Solution
Disable TCP keepalive messages from the server and set `ClientAliveCountMax` to 0, which disables connection termination

## Review
Anyone can review this

### Reviewer Checklist


.github/workflows/scripts/gcp-vm-startup-script.sh
```diff
@@ -11,6 +14,8 @@ sudo cat /etc/ssh/sshd_config
 echo
 echo 'Modifying config:'
 echo 'MaxStartups 500' | sudo tee --append /etc/ssh/sshd_config \
+echo 'TCPKeepAlive no' | sudo tee --append /etc/ssh/sshd_config \
```
teor2345, line 17: Disabling TCP keepalive packets will make some firewalls terminate the connection as inactive.
Is there some documentation supporting this change?

.github/workflows/scripts/gcp-vm-startup-script.sh
```diff
@@ -11,6 +14,8 @@ sudo cat /etc/ssh/sshd_config
...
+echo 'ClientAliveCountMax 0' | sudo tee --append /etc/ssh/sshd_config \
```
teor2345, line 18: We might want to set this to a high number instead (6 hours), to avoid stuck connections exhausting the connection limit.

Or we can also increase the authenticated connection limit so it's higher than the number of connections we expect from the longest workflow. (About 20.)

teor2345 (changes requested):
Just wondering if there is a specific log message or failure you're trying to fix here?

Edit: yes, this is a known issue.

teor2345:
I'm also going to ask the new CI change questions from the retro:
- how are we sure that this change is the best solution to the problem?
- how are we going to test this?
- what manual testing has been done?

teor2345:
> Just wondering if there is a specific log message or failure you're trying to fix here?

This is actually happening in CI:
> Timeout, server 34.123.232.71 not responding.

https://github.com/ZcashFoundation/zebra/pull/5391#issuecomment-1276724988

gustavovalverde:
We're no longer needing this. We might consider to open this once again if we see SSH issues reappear

PR #5371, opened 2022-10-10
ci(disk): use an official GCP image on CI VMs for disk auto-resizing, make CI & CD disks 300GB

### LINKED ISSUE #5085: Increase GCP disk size to 300 GB

gustavovalverde:
## Previous behavior
We've presented issues in the past with resizing as the device is busy, for example:

```
e2fsck: Cannot continue, aborting.
/dev/sdb is in use.
```

Depends-On: #5370
Fixes #5085

## Expected behavior
We've been manually resizing the disk as this task was not being done automatically, but having an official Public Image from GCP would make this easier (automatic) and it also integrates better with other GCP services

Configuration differences: https://cloud.google.com/compute/docs/images/os-details#notable-difference-debian

## Solution
- Use `debian-11` from the official public images https://cloud.google.com/compute/docs/images/os-details#debian
- Remove the manual disk resizing from the pipeline

### Why is this best solution to the problem?
- We've been using a custom image for VMs, which requires us to create tooling already implemented in official public images by GCP
- Resizing is one of the examples why we may want to change our custom image to an official GCP public image

### How are we going to test this?
- Resizing the disks and validating CI is working

### What manual testing has been done?
- The disk have been resized in this PR

## Review

Anyone from DevOps, but @teor2345 has interacted with this changes before

### Reviewer Checklist


teor2345:
I'm going to ask the new CI change questions from the retro:
- how are we sure that this change is the best solution to the problem?
- how are we going to test this?
- what manual testing has been done?

Let's see if they help us make these changes work.

gustavovalverde:
Personal note: Consider the changes done here while testing -> https://github.com/ZcashFoundation/zebra/pull/5367#discussion_r991701989

gustavovalverde:
> * how are we sure that this change is the best solution to the problem?
> * how are we going to test this?
> * what manual testing has been done?

Added to the PR description

teor2345:
> Configuration differences: https://cloud.google.com/compute/docs/images/os-details#notable-difference-debian

Some things to note, no blockers:
> IPv6 is enabled.

This might change sync speeds, it could be faster.

> The [Unattended-upgrades](https://wiki.debian.org/UnattendedUpgrades) package is installed and configured to download and install Debian security updates daily. This can be configured or disabled by changing the values in /etc/apt/apt.conf.d/50unattended-upgrades and /etc/apt/apt.conf.d/02periodic.

This is probably unnecessary, if it causes problems, it will mainly impact the full sync.

> The SSH server configuration is set up as follows:
>    Root login is disabled.

This change should be fine if CI passes.

teor2345 (approved):
This PR has about the same CI speed as before this change, both around 6 hours.

This PR:
https://github.com/ZcashFoundation/zebra/actions/runs/3227078141

`main` branch:
https://github.com/ZcashFoundation/zebra/actions/runs/3222101502

gustavovalverde:
@Mergifyio refresh

gustavovalverde:
Merging manually as this has been stalled for a while

PR #5377, opened 2022-10-11
feat(test): Check that the state service handles errors correctly when committing finalized blocks

upbqdn:
## Motivation

We recently refactored the process of committing blocks to the finalized and non-finalized state. This PR tests if Zebra handles errors correctly when committing blocks to the finalized state.

## Solution

In this PR, we trigger a failure when committing a block, and make sure the state service handles it as expected. However, the test dead-locks when Zebra is running on a single CPU. See comments below for details. For this reason, we closed this PR without merging.

The issue addressed in this PR is in #2654, but this PR doesn't test the non-finalized state, so it doesn't resolve the whole ticket.

### Reviewer Checklist

## Follow Up Work

We might need to do a similar test for the non-finalized state as well.


teor2345 (dismissed):
Looks good, thanks for working out how to do this!

I'm happy if you want to just delete the bits that didn't work, they're not release blockers.

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -421,6 +421,14 @@ impl DiskWriteBatch {
         note_commitment_trees: NoteCommitmentTrees,
         value_pool: ValueBalance<NonNegative>,
     ) -> Result<(), BoxError> {
+        // Return a fake error for some tests.
+        #[cfg(test)]
+        {
+            if finalized.height < Height(100) && finalized.hash == block::Hash([0; 32]) {
```
arya2, line 439: Just curious, why check the block's height?

arya2 (dismissed):
Looks good!

zebra-state/src/service/finalized_state/zebra_db/block.rs
upbqdn: The condition is checked for every test we run so just I wanted to restrict the set of blocks that could trigger the failure in case someone adds future tests that would be committing blocks with hashes equal to zero.

upbqdn: Let me add that as a comment.

upbqdn: Done in [b75f85e](https://github.com/ZcashFoundation/zebra/pull/5377/commits/b75f85ec73681837b2b6f7183209609a7d0f743a).

arya2: > The condition is checked for every test we run so just I wanted to restrict the set of blocks that could trigger the failure in case someone adds future tests that would be committing blocks with hashes equal to zero.

Ah. that makes sense. Thank you for adding the comment.

teor2345 (dismissed):
Thanks!

teor2345 (approved):
Thanks!

teor2345 (changes requested):
This change appears to make all the GitHub runner tests hang (Linux, macOS, and Linux Coverage).

I'm not sure exactly why, the logs seem to just stop?

Edit: the docker tests are also taking a long time, they might have hung too:
https://github.com/ZcashFoundation/zebra/actions/runs/3237624040/jobs/5305649035

upbqdn:
The new test works well locally.

teor2345:
> The new test works well locally.

The new test also works ok on my machine.

But I discovered that it hangs when there is only 1 CPU available. Which should be OK, some tests require multiple cores.

But maybe it also hangs in CI when there are only 2 virtual CPU cores available?
That might be a timing issue.

Here's how I tested it with just one core:
```sh
$ taskset 0x01 cargo test -- state_handles_error_correctly
```

I'm really not sure how to make progress on this one.
We could update the ticket with the progress we made, and close the PR?

(Sorry, I realise this has happened a bit to your work lately, I wish I could be more helpful!)

teor2345:
We could also disable the test by default in CI using `#[ignore]`, and merge it?

Then it will run locally when people ask for it, but it won't break CI or other weird test environments.

arya2:
> But maybe it also hangs in CI when there are only 2 virtual CPU cores available? That might be a timing issue.

I don't think it's a timing issue because it consistently it hangs in CI and passes locally. 

We could create an issue for when larger runners become available, they're currently in public beta for enterprise: https://github.blog/changelog/2022-09-01-github-actions-larger-runners-are-now-in-public-beta

teor2345:
We could add some logging to each line in each thread, and see there it hangs in CI?
But that could be a lot of work for a low-priority test.

@mpguerra what do you think? Should we work on this sync test more, or just close the PR and update the ticket with this draft branch?

mpguerra:
> @mpguerra what do you think? Should we work on this sync test more, or just close the PR and update the ticket with this draft branch?

Let's just close the PR and update with draft branch. Sorry @upbqdn  :(

upbqdn:
> I discovered that it hangs when there is only 1 CPU available.

You're right, running
```bash
taskset 0x1 cargo test -- state_handles_error_correctly
```
dead-locks since there' no CPU activity. However, running the test on two CPUs like this
```bash
taskset 0x3 cargo test -- state_handles_error_correctly
```
works fine. I was trying to find out where the dead-lock happens, but I didn't manage to figure it out.

upbqdn:
> Let's just close the PR and update with draft branch.

Sure, I also referenced this PR in ticket #2654 so that it's easy to find the description and the comments. I also updated the description of this PR.

PR #5393, opened 2022-10-12
ci(sync): only run the `lightwalletd` full sync on the `main` branch

### LINKED ISSUE #5316: Only run the `lightwalletd` full sync on the `main` branch
## Motivation

In PR #5164, we made `lightwalletd` sync all the way to the tip in its full sync test.

This increases that test's time from 1 hour to 4 hours, which makes the CI we run on each PR change increase from 3 hours to 6 hours.

This increases Google Cloud costs, and makes PR testing and merges slower for engineers.

### Designs

We're already running the full `zebrad` sync only on the `main` branch.
We can copy and paste the workflow code to the full `lightwalletd` sync, so this should be an easy change.

We might also need to update the patch workflows or the branch protection rules.

gustavovalverde:
## Previous behavior
In PR #5164, we made lightwalletd sync all the way to the tip in its full sync test.

This increases that test's time from 1 hour to 4 hours, which makes the CI we run on each PR change increase from 3 hours to 6 hours.

## Expected behavior
Run the lightwalletd full sync just on `main` or if a state disk for the actual version is not found.

## Solution
Add the `github.event_name == 'push' && github.ref_name == 'main'` condition to the `lightwalletd-full-sync` test.

Fixes #5316

## Review

Anyone from the DevOps team
**How to test:** This job should not run in this PR

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -287,6 +287,14 @@ jobs:
       saves_to_disk: true
       disk_suffix: checkpoint
       height_grep_text: 'flushing database to disk .*height.*=.*Height.*\('
+    # We want to prevent multiple checkpoint syncs running at the same time,
+    # but we don't want to cancel running syncs on `main` if a new PR gets merged,
+    # because we might never get a finished sync.
+    #
+    # See the concurrency comment on the zebrad test-full-sync job for details.
+    concurrency: 
+      group: ${{ github.workflow }}−${{ github.ref }}-regenerate-stateful-disks
+      cancel-in-progress: false
 
   # Test that Zebra syncs and fully validates a few thousand blocks from a cached mandatory checkpoint disk
   #
@@ -334,9 +342,11 @@ jobs:
       saves_to_disk: true
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
-    # We don't want to cancel running full syncs on `main` if a new PR gets merged,
-    # because we might never finish a full sync during busy weeks. Instead, we let the
-    # first sync complete, then queue the latest pending sync, cancelling any syncs in between.
+    # We want to prevent multiple full zebrad syncs running at the same time,
+    # but we don't want to cancel running syncs on `main` if a new PR gets merged,
+    # because we might never get a finished sync.
+    # 
+    # Instead, we let the first sync complete, then queue the latest pending sync, cancelling any syncs in between.
     # (As the general workflow concurrency group just gets matched in Pull Requests,
     # it has no impact on this job.)
     #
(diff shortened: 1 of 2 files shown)


arya2 (dismissed):
LGTM!

gustavovalverde:
Protection rules updated to allow this and other PRs to be merged

teor2345 (changes requested):
> Protection rules updated to allow this and other PRs to be merged

This PR disables the requirement for the `lwd` full sync job, but we added a patch job for the `zebrad` full sync job.

Screen Shot 2022-10-13 at 07 25 07

https://github.com/ZcashFoundation/zebra/blob/8646a03e03b9dda190c46f05fd64631a99c9da9d/.github/workflows/continous-integration-docker.patch-always.yml#L18-L22

Can we choose one way to do it, and make it consistent?

.github/workflows/continous-integration-docker.yml
```diff
@@ -385,7 +384,7 @@ jobs:
     # to also run on Mergify head branches,
     # add `|| (github.event_name == 'push' && startsWith(github.head_ref, 'mergify/merge-queue/'))`:
     # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request-1
-    if: ${{ (!cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true') || !fromJSON(needs.get-available-disks.outputs.lwd_tip_disk) }}
+    if: ${{ (!cancelled() && !failure() && (github.event_name == 'push' && github.ref_name == 'main') && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true') || !fromJSON(needs.get-available-disks.outputs.lwd_tip_disk) }}
```
teor2345: I'm wondering about the logic here:
- `!cancelled() && !failure()` - we must have success or skip for previous jobs, regardless of any other conditions, so I think these conditions need to be `&&` with all the rest of the conditions or triggers
- `github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true'` - these inputs will never be true on a push event, they are only set for manual workflow runs, so we can remove them
- I think we'll want to be able to manually trigger this job, like the zebrad checkpoint and full sync jobs (I'll open a PR for this)

So we might want to use the same conditions as the zebrad full sync, with an extra `!cancelled() && !failure()` at the start:
```suggestion
    if: ${{ !cancelled() && !failure() && ((github.event_name == 'push' && github.ref_name == 'main') || !fromJSON(needs.get-available-disks.outputs.lwd_tip_disk)) }}
```

Here are the zebrad full sync conditions, for comparison:
https://github.com/ZcashFoundation/zebra/blob/8646a03e03b9dda190c46f05fd64631a99c9da9d/.github/workflows/continous-integration-docker.yml#L321

teor2345 (changes requested):
I'm not sure about the conditions here, I think we might want to match the full sync job almost exactly?

I'll also push a PR with some more suggestions.

teor2345:
I opened PRs #5399 and #5400, feel free to merge them into this PR, or merge them into `main` separately.

PR #5400 will conflict with my job condition suggestion above, because it adds to that same line.

gustavovalverde:
This has been updated accordingly

teor2345:
> > Protection rules updated to allow this and other PRs to be merged
> 
> This PR disables the requirement for the `lwd` full sync job, but we added a patch job for the `zebrad` full sync job.

I added an "always" patch job, and re-added the branch protection rule.

teor2345 (approved):
Thanks, I think we're good to go here.

Issue #4649, opened 2022-06-20
Zebra falsely estimates that it's close to the tip when the network connection goes down


upbqdn:
## Motivation

Zebra currently estimates that it is close to the tip if the average number of synced blocks in the last four batches drops below 20. The decision is made in the function [`is_close_to_tip`](https://github.com/ZcashFoundation/zebra/blob/update-column-family-names/zebrad/src/components/sync/status.rs#L47) here https://github.com/ZcashFoundation/zebra/blob/update-column-family-names/zebrad/src/components/sync/status.rs#L67. Zebra uses the output of this function to activate the mempool.

However, when Zebra loses its peers during the syncing process, [`is_close_to_tip`](https://github.com/ZcashFoundation/zebra/blob/update-column-family-names/zebrad/src/components/sync/status.rs#L47) returns a false positive because the number of synced blocks in the last four batches naturally drops below 20. This happens, for example, when the internet connection goes down on the machine where Zebra is running.

A consequence of this bug is that the mempool is activated too early. Also, once the connection is back up, Zebra is likely to start crawling for tx IDs from other peers, and gossiping new block hashes to peers. This is unwanted load to other peers. However, this extra load won't last long because Zebra will figure out that it's not at the tip. It will also deactivate the mempool. 

### Solution

Make sure Zebra has enough connections before [`is_close_to_tip`](https://github.com/ZcashFoundation/zebra/blob/update-column-family-names/zebrad/src/components/sync/status.rs#L47) returns the final decision.

#### Steps

- Move `WatchReceiver` to `zebra-chain` so that you can use it from both `zebra-state` and `zebra-network`
- Create a new channel in https://github.com/zcashfoundation/zebra/blob/75a679792bb787b95b4e9ce87aaefe205c48c97a/zebra-network/src/peer_set/initialize.rs#L159, store the sender in `PeerSet`, and return the receiver.
- Use the sender in `update_metrics` in `PeerSet`.
- Store the receiver in `SyncStatus` as `WatchReceiver`, and use it in `is_close_to_tip`.
- Test that Zebra does not activate the mempool when it has no peers.
- Test that the `getblocktemplate` RPCs and other users of `SyncStatus` still work correctly.

teor2345:
> Make sure Zebra has enough connections before [`is_close_to_tip`](https://github.com/ZcashFoundation/zebra/blob/update-column-family-names/zebrad/src/components/sync/status.rs#L47) returns the final decision.

[...]

PR #5397, opened 2022-10-12
fix(sync): Make timeouts shorter to reduce full sync times

teor2345:
## Motivation

The full sync is timing out, so we want to reduce the amount of time we wait for stuck blocks.

## Solution

Reduce the default block timeout to 8 minutes, and the final checkpoint timeout to 2 minutes.

## Review

@arya2 is also working on fixing this.

### Reviewer Checklist

## Follow Up Work

Increase the Rust test or workflow job timeouts, if needed.


teor2345:
macOS tests failed due to a network error and bug #4649:
> assertion failed: output.stdout_line_contains(\"activating mempool\").is_err()

https://github.com/ZcashFoundation/zebra/actions/runs/3237875279/jobs/5305448541#step:15:520

oxarbitrage (approved):

PR #5357, opened 2022-10-06
feat(rpc): introduce `getblocktemplate-rpcs` feature

oxarbitrage:
## Motivation

We want to support mining rpc calls but as an isolated rust feature. Can close https://github.com/ZcashFoundation/zebra/issues/5305 with some more work. It also implement partially `get_block_count` (https://github.com/ZcashFoundation/zebra/issues/5303) so we can have something to test with.

Also closes #5303.

## Solution

PR attempts to do the [supertrait suggested implementation](https://github.com/ZcashFoundation/zebra/issues/5305#issuecomment-1269114778). It looks good and we can make tests but something is missing. When calling the rpc with curl:

```
$ curl --silent --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockcount", "params": [] }' -H 'Content-type: application/json' http://127.0.0.1:8232/ | jq
{
  "error": {
    "code": -32601,
    "message": "Method not found"
  },
  "id": "curltest"
}
$ 
```

## Review

I will like @teor2345 to take a look

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

PR #5418, opened 2022-10-18
suggestion for introduce `getblocktemplate-rpcs` feature

arya2:
## Motivation

This is a suggestion (or potential follow-up) for PR https://github.com/ZcashFoundation/zebra/pull/5357 - we could add a new RPC trait with just the field that's currently being used and easily add all of the rest of the fields as they become necessary.

## Review

@oxarbitrage can review & merge or close this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6506,7 +6506,6 @@ dependencies = [
 name = "zebra-rpc"
 version = "1.0.0-beta.15"
 dependencies = [
- "cfg-if 1.0.0",
  "chrono",
  "futures",
  "hex",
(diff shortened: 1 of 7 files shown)


oxarbitrage (commented):
I think this looks good.

I am concerned on how easy it will be to pass the rest of the fields we are going to need like state or mempool. Do you think it will be better to do that now ?

arya2:
> I think this looks good.
> 
> I am concerned on how easy it will be to pass the rest of the fields we are going to need like state or mempool. Do you think it will be better to do that now ?

I thought it might be difficult to add the mempool but then I tried adding everything except the `queue_sender` to check and it was very easy.

oxarbitrage (approved):
Yea, this is what i wanted initially. Thanks for making it happen. I think we should use it instead of what we have.

PR #4967, opened 2022-08-27
feat(rpc): add getblockhash rpc method

oxarbitrage:
## Motivation

After the release candidate Zebra will continue the development of more API calls. `getblockhash` is one of tme methods we will implement almost for sure. 

This is low priority however as it is pretty much done i want to push it to zebra reoo as a PR. 

This is a draft because is low priority however in other scenario it should be ready for review.

Close #5268.

## Solution

Implement `getblockhash`.

## Required Tests

- [x] Snapshot the output of the RPC
- [ ] Test calling the RPC from the mining pool software
- [x] Manually compare the `zcashd` and `zebrad` RPC output using https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/zcash-rpc-diff

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [x] How do you know it works? Does it have tests?

PR #5428, opened 2022-10-19
Suggestion for "add getblockhash rpc method" PR

teor2345:
## Motivation

This is a suggestion for PR #4967.

It makes two changes:
- explain or check that calculations don't overflow
- simplify error handling

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -1376,7 +1376,10 @@ fn get_height_from_int(index: i32, tip_height: Height) -> Result<Height> {
         }
         Ok(Height(height))
     } else {
-        let height = (tip_height.0 as i32).checked_add(index + 1);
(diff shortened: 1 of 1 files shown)


zebra-rpc/src/methods.rs
```diff
@@ -1376,10 +1376,13 @@ fn get_height_from_int(index: i32, tip_height: Height) -> Result<Height> {
         }
         Ok(Height(height))
     } else {
-        let height = (tip_height.0 as i32).checked_add(index + 1);
```
teor2345, line 1379: I try to avoid the `as` operator in production code, because it casts the bits without checking for overflow:
https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#numeric-cast

oxarbitrage: Yea, i generally avoid it as well but it seems i missed it. Thanks.

PR #5395, opened 2022-10-12
docs(audit): Add a list of Zebra dependencies for the audit

oxarbitrage:
## Motivation

This PR enumerates all Zebra dependencies into three sections for auditors to reference: 
- In Scope
- Partly In Scope
- Out Of Scope

Crate versions and links to the source code for those versions are provided as well but may be outdated for items that are out-of-scope.

Closes #5214
Close #3096

Issue #5214, opened 2022-09-21
Task: Make a list of Zebra dependencies


mpguerra:
## Motivation

When we engage a team to audit Zebra, we want to provide them with a number of lists outlining our dependencies and which of those we would like to be in scope for the audit.

## Task

We will need 3 lists:

- [ ] Direct ECC dependencies that are **in scope** for the audit (most will be out of scope as they have already been audited)
- [ ] Direct Zebra dependencies that are **in scope** as they are consensus or security critical 
- [ ] All other direct dependencies which are **out of scope**, including the ECC ones.

We can create the full list of all production Zebra dependencies with a command like:
> cargo tree --depth 1 --prefix none --locked --edges normal --workspace --exclude zebra-client --exclude zebra-test --features sentry,journald,filter-reload,prometheus

The options to `cargo tree` are listed here:
https://doc.rust-lang.org/cargo/commands/cargo-tree.html

ECC dependencies are usually from https://github.com/zcash
Consensus-critical dependencies are used to implement specific consensus rules, for example, `zcash_script`.

teor2345:
I've noticed that we're not automatically updating some dependencies using dependabot.

In some cases, this appears to be a bug (or incorrect documentation) for dependabot. Closing automated dependabot PRs seems to stop all dependency updates, not just the selected minor or major version.

If any of these dependencies are _in scope_ for the audit, we might want to upgrade them before tagging the release candidate.

I reopened dependabot PRs for these dependencies:
- color-eyre v0.6.1 -> v0.6.2
- metrics-exporter-prometheus v0.9.0 -> v0.11.0
- metrics v0.18.1 -> v0.20.1
- sha2 v0.9.9 -> v0.10.5 (just to check)
- tracing-subscriber v0.3.11 -> v0.3.15
- tracing v0.1.31 -> v0.1.36

These aren't all release blockers - we need to check the list created by this ticket.

These dependencies are deliberately kept in sync with the versions used by other dependencies:
- aes v0.7.5 -> v0.8.1
- gumdrop v0.7 -> v0.8.1 (abscissa)
- secp256k1 v0.21.3 -> v0.24.0
- sha2 v0.9.9 -> v0.10.5

These dependencies are too much work to upgrade everywhere:
- pin-project v0.4.29 -> v1.0.12
- proptest v0.10.1 -> v1.0.0

[...]

PR #5433, opened 2022-10-19
add(doc): Add draft audit process, and update some support docs

### LINKED ISSUE #5431: docs: Explain the process to be taken when a code audit is to take place
## Motivation

We've been defining our release process lately, and this also requires to define a process when auditors are going to start an audit on our code. Even though we can define the process on our own, it would be better to have feedback from our first audit before documenting a process that might change while the audit is happening.

## Related Work

https://github.com/ZcashFoundation/zebra/pull/5392

teor2345:
## Motivation

Let's document the audit decisions we've taken so far, we can update them later if they change.

This PR also makes some minor support documentation fixes.

Close #5431.

## Solution

The audit doc is based on meetings, slack conversations, ticket #5214, and PR #5395.

The release process changes fix an incorrect render on `main`, and make the document consistent with our other markdown docs.

The download workaround addresses a bug report in the #zebra discord channel.

The build doc changes replace an outdated build process with a link to the latest documentation in the README.

## Review

Anyone can review this PR, but @gustavovalverde opened the audit ticket.

### Reviewer Checklist

## Follow Up Work

Edit this document if the audit process changes significantly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,7 @@ So Zebra's state should always be valid, unless your OS or disk hardware is corr
 There are a few bugs in Zebra that we're still working on fixing:
 
 - Zebra requires Rust 1.63, due to [a compiler performance regression in Rust 1.64](https://github.com/ZcashFoundation/zebra/issues/5091)
+  - If Zebra fails downloading the Zcash parameters, use [the Zcash parameters download script](https://github.com/zcash/zcash/blob/master/zcutil/fetch-params.sh) instead. This script might be needed on macOS, even with Rust 1.63.
 - No Windows support [#3801](https://github.com/ZcashFoundation/zebra/issues/3801)
   - We used to test with Windows Server 2019, but not anymore; see issue for details
 
--- a/book/src/dev/release-process.md
+++ b/book/src/dev/release-process.md
@@ -8,7 +8,7 @@ Before reading, you should understand [Semantic Versioning](https://semver.org/s
 
 <a id="versioning"></a>
 
-## Zebra versioning {#zebra-versioning}
+## Zebra versioning
 
 Zebra version numbers show the impact of the changes in a release. They are composed of three parts: `major.minor.patch`.
 For example, version `3.1.11` indicates major version 3, minor version 1, and patch level 11.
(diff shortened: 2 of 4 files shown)


book/src/dev/release-process.md
```diff
@@ -8,7 +8,7 @@ Before reading, you should understand [Semantic Versioning](https://semver.org/s
 
 <a id="versioning"></a>
 
-## Zebra versioning {#zebra-versioning}
+## Zebra versioning
```
mpguerra, line 11: are these anchors that we are removing and if so, why?

mpguerra (commented):
Thank you for writing this!

README.md
```diff
@@ -227,6 +227,7 @@ So Zebra's state should always be valid, unless your OS or disk hardware is corr
 There are a few bugs in Zebra that we're still working on fixing:
 
 - Zebra requires Rust 1.63, due to [a compiler performance regression in Rust 1.64](https://github.com/ZcashFoundation/zebra/issues/5091)
+  - If Zebra fails downloading the Zcash parameters, use [the Zcash parameters download script](https://github.com/zcash/zcash/blob/master/zcutil/fetch-params.sh) instead. This script might be needed on macOS, even with Rust 1.63.
```
upbqdn, line 230: Is it worth adding this issue since there's a new version of `zcash_proofs`?
```suggestion
```

book/src/dev/release-process.md
upbqdn: Yes, these are anchors. The motivation to remove them is that they get rendered to the final `html` file: https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/release-process.md#zebra-versioning-zebra-versioning.

I think mdBook and GitHub add implicit anchors to each heading anyway. If that doesn't work, we can add the `<a></a>` `html` tag as you can see above.

upbqdn (approved):

README.md
teor2345: There's no PR for zcash_proofs yet, so I think it's ok to add this now, and remove it when we remove the whole issue.

PR #5434, opened 2022-10-20
change(mempool) reject transactions with spent outpoints or nullifiers

### LINKED ISSUE #2631: Send committed spent OutPoints & nullifiers to the mempool for rejection
## Motivation

We need to get a list of recently committed spent outpoints & nullifiers from the state service, so we can evict any mempool transactions that also spend/reveal them.

This fixes some subtle bugs:
1. Transaction `1` spends `B` and `C` - it gets in the mempool
2. Transaction `2` only spends `B` - it gets mined into the chain, but transaction `1` is not rejected from the mempool
    - this is an optional-priority bug, because we're accepting more transactions than `zcashd`
3. Transaction `3` only spends `C` - it is rejected from the mempool, but it should be accepted
    - this is a low-priority bug, because we're rejecting transactions that `zcashd` would accept

## Constraints

These rejections should persist until a chain reset, because:
* an outpoint can only be spent once in the chain, and
* a nullifier can only be revealed once in the chain.

arya2:
## Motivation

To reject transactions in the mempool when they're invalidated by recently committed blocks.

Closes #2631 

## Solution

- Add the transactions of committed blocks to the ChainTipBlock
- Reject and remove transactions in the mempool that are using spent outpoints or nullifiers from the latest chain tip
- Reject mined transactions

## Review

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -365,11 +365,39 @@ impl Service<Request> for Mempool {
             if let Some(TipAction::Grow { block }) = tip_action {
                 tracing::trace!(block_height = ?block.height, "handling blocks added to tip");
 
+                let spent_outpoints = block
+                    .transactions
+                    .iter()
+                    .flat_map(|tx| tx.spent_outpoints())
+                    .collect();
+
+                let sprout_nullifiers = block
+                    .transactions
+                    .iter()
+                    .flat_map(|transaction| transaction.sprout_nullifiers())
+                    .collect();
+                let sapling_nullifiers = block
+                    .transactions
+                    .iter()
+                    .flat_map(|transaction| transaction.sapling_nullifiers())
+                    .collect();
+                let orchard_nullifiers = block
+                    .transactions
+                    .iter()
+                    .flat_map(|transaction| transaction.orchard_nullifiers())
+                    .collect();
+
                 // Cancel downloads/verifications/storage of transactions
                 // with the same mined IDs as recently mined transactions.
                 let mined_ids = block.transaction_hashes.iter().cloned().collect();
                 tx_downloads.cancel(&mined_ids);
                 storage.remove_same_effects(&mined_ids);
+                storage.reject_invalidated_transactions(
+                    spent_outpoints,
+                    sprout_nullifiers,
+                    sapling_nullifiers,
+                    orchard_nullifiers,
+                );
                 storage.clear_tip_rejections();
             }
 
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -365,39 +365,11 @@ impl Service<Request> for Mempool {
             if let Some(TipAction::Grow { block }) = tip_action {
                 tracing::trace!(block_height = ?block.height, "handling blocks added to tip");
 
-                let spent_outpoints = block
-                    .transactions
-                    .iter()
-                    .flat_map(|tx| tx.spent_outpoints())
-                    .collect();
-
-                let sprout_nullifiers = block
-                    .transactions
-                    .iter()
-                    .flat_map(|transaction| transaction.sprout_nullifiers())
-                    .collect();
-                let sapling_nullifiers = block
-                    .transactions
-                    .iter()
-                    .flat_map(|transaction| transaction.sapling_nullifiers())
-                    .collect();
-                let orchard_nullifiers = block
-                    .transactions
-                    .iter()
-                    .flat_map(|transaction| transaction.orchard_nullifiers())
-                    .collect();
-
                 // Cancel downloads/verifications/storage of transactions
                 // with the same mined IDs as recently mined transactions.
                 let mined_ids = block.transaction_hashes.iter().cloned().collect();
                 tx_downloads.cancel(&mined_ids);
-                storage.remove_same_effects(&mined_ids);
-                storage.reject_invalidated_transactions(
-                    spent_outpoints,
-                    sprout_nullifiers,
-                    sapling_nullifiers,
-                    orchard_nullifiers,
-                );
+                storage.reject_and_remove_same_effects(&mined_ids, block.transactions);
                 storage.clear_tip_rejections();
             }
 
--- a/zebrad/src/components/mempool/storage.rs
+++ b/zebrad/src/components/mempool/storage.rs
@@ -81,7 +81,7 @@ pub enum SameEffectsChainRejectionError {
     #[error("best chain tip has reached transaction expiry height")]
     Expired,
 
-    #[error("transaction outpoints or nullifiers were committed to the best chain")]
+    #[error("transaction inputs were spent, or nullifiers were revealed, in the best chain")]
     DuplicateSpend,
 
     #[error("transaction was committed to the best chain")]
--- a/zebrad/src/components/mempool/storage.rs
+++ b/zebrad/src/components/mempool/storage.rs
@@ -10,15 +10,14 @@
 use std::{
     collections::{HashMap, HashSet},
     mem::size_of,
+    sync::Arc,
     time::Duration,
 };
 
 use thiserror::Error;
 
-use zebra_chain::{
-    orchard, sapling, sprout,
-    transaction::{self, Hash, UnminedTx, UnminedTxId, VerifiedUnminedTx},
-    transparent::OutPoint,
+use zebra_chain::transaction::{
+    self, Hash, Transaction, UnminedTx, UnminedTxId, VerifiedUnminedTx,
 };
 
 use self::{eviction_list::EvictionList, verified_set::VerifiedSet};
@@ -82,7 +81,13 @@ pub enum SameEffectsChainRejectionError {
     #[error("best chain tip has reached transaction expiry height")]
     Expired,
 
-    #[error("transaction outpoints or nullifiers were committed to the best chain")]
+    #[error(
+        "transaction rejected because another transaction in the chain or mempool \
+        has already spent some of its inputs"
+    )]
+    DuplicateSpend,
+
+    #[error("transaction was committed to the best chain")]
     Mined,
 
     /// Otherwise valid transaction removed from mempool due to [ZIP-401] random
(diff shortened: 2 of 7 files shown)


zebrad/src/components/mempool.rs
```diff
@@ -365,11 +365,39 @@ impl Service<Request> for Mempool {
...
+                let sapling_nullifiers = block
+                    .transactions
+                    .iter()
+                    .flat_map(|transaction| transaction.sapling_nullifiers())
+                    .collect();
+                let orchard_nullifiers = block
+                    .transactions
+                    .iter()
+                    .flat_map(|transaction| transaction.orchard_nullifiers())
+                    .collect();
```
upbqdn: It looks like the methods `remove_same_effects` and `reject_invalidated_transactions` could be merged, and this code could be moved to the unified method.

upbqdn (commented):
This looks good to me.

I don't know how often is `poll_ready` actually called, so my concern is that if there happen to be two or more blocks mined fast one after another (which can happen), Zebra would skip some blocks. This applies to other checks as well, for example, in `remove_same_effects`.

zebrad/src/components/mempool/storage.rs
```diff
@@ -78,6 +82,9 @@ pub enum SameEffectsChainRejectionError {
     #[error("best chain tip has reached transaction expiry height")]
     Expired,
 
+    #[error("transaction outpoints or nullifiers were committed to the best chain")]
+    Mined,
```
teor2345, line 88: Sometimes this happens because the transaction has been mined, but other times it happens because the output was spent by another transaction.

```suggestion
    DuplicateSpend,
```

zebrad/src/components/mempool/storage.rs
```diff
@@ -294,6 +301,56 @@ impl Storage {
...
+    ///
+    /// Returns the number of transactions that were removed and rejected.
+    pub fn reject_invalidated_transactions(
+        &mut self,
+        spent_outpoints: HashSet<OutPoint>,
+        sprout_nullifiers: HashSet<&sprout::Nullifier>,
+        sapling_nullifiers: HashSet<&sapling::Nullifier>,
+        orchard_nullifiers: HashSet<&orchard::Nullifier>,
+    ) -> usize {
+        let mined_ids: HashSet<_> = self
```
teor2345:
```suggestion
        let duplicate_spend_ids: HashSet<_> = self
```

teor2345 (commented):
Looks great!

(And it's simpler than I thought 🙂)

I have a minor suggestion about naming things.

teor2345:
> A possible alternative may be to pass the `non_finalized_state_receiver` to the mempool so that it can check the `spent_utxos` and nullifiers of all recently committed blocks on the best chain, rather than just the ChainTipBlock.

This isn't needed, because the mempool only updates by one block at a time. If there are multiple blocks, it assumes it has fallen out of sync, and resets itself.

zebrad/src/components/mempool.rs
teor2345: It would be good to move these collect statements into the new method, to keep poll_ready() short and readable.

zebrad/src/components/mempool.rs
```diff
@@ -365,11 +365,39 @@ impl Service<Request> for Mempool {
...
+
                 // Cancel downloads/verifications/storage of transactions
                 // with the same mined IDs as recently mined transactions.
                 let mined_ids = block.transaction_hashes.iter().cloned().collect();
                 tx_downloads.cancel(&mined_ids);
                 storage.remove_same_effects(&mined_ids);
```
teor2345: Let's cache the rejections of the mined transaction IDs using a new `Mined` error?
("transaction was committed to the best chain")

This is an existing bug, but fixing it will make the mempool faster, because the transactions won't be re-downloaded and re-validated.

teor2345:
This is a bug fix. We want to do it to improve the mempool, it's just a higher priority now we're doing mining RPCs.

So don't think we need to put it behind the `getblocktemplate-rpcs` feature, because it's a fix that we want in production. (And it's ok for the audit as well, because it's more like `zcashd`.)

teor2345:
> I don't know how often is `poll_ready` actually called, so my concern is that if there happen to be two or more blocks mined fast one after another (which can happen), Zebra would skip some blocks. This applies to other checks as well, for example, in `remove_same_effects`.

If Zebra gets multiple blocks between mempool updates, it drops all the mempool transactions, and starts fresh with an empty mempool.

This is implemented using a `Reset`, rather than a `Grow`:
https://github.com/ZcashFoundation/zebra/blob/3825caae0343d616b011dd666737934873954b8b/zebra-state/src/service/chain_tip.rs#L419-L433

We track this as part of `ChainTipChange.last_tip_change()`:
https://github.com/ZcashFoundation/zebra/blob/3825caae0343d616b011dd666737934873954b8b/zebra-state/src/service/chain_tip.rs#L543-L549

zebrad/src/components/mempool.rs
mpguerra: Do we have an issue number for this existing bug? If so, let's link it in to this PR and/or split the fix into a new PR.

zebrad/src/components/mempool.rs
arya2: Good points. Prevents repetition in tests too.

zebrad/src/components/mempool/storage.rs
arya2: Renamed to `duplicate_spend_ids` in [`b5cffbf`](https://github.com/ZcashFoundation/zebra/pull/5434/commits/b5cffbfe887e5701fb6b72020c5f390e05319c90) (just moved below the removal of `mined_ids` to avoid needless iteration over spent outpoints and nullifiers of mined transactions.)

zebrad/src/components/mempool/storage.rs
arya2: Renamed to DuplicateSpend in [`b5cffbf`](https://github.com/ZcashFoundation/zebra/pull/5434/commits/b5cffbfe887e5701fb6b72020c5f390e05319c90).

arya2:
> I don't know how often is `poll_ready` actually called, so my concern is that if there happen to be two or more blocks mined fast one after another (which can happen), Zebra would skip some blocks. This applies to other checks as well, for example, in `remove_same_effects`.

I was also concerned about this.

> If Zebra gets multiple blocks between mempool updates, it drops all the mempool transactions, and starts fresh with an empty mempool.

Good to know that's not an issue.

zebrad/src/components/mempool.rs
arya2: > Let's cache the rejections of the mined transaction IDs using a new `Mined` error? ("transaction was committed to the best chain")

I thought it might as well reject every mined id from blocks on the chain, let me know if it should only reject those that were removed from the verified transactions.

> Do we have an issue number for this existing bug? If so, let's link it in to this PR and/or split the fix into a new PR.

It's a small change and not worth creating an issue for in my opinion.

teor2345: > > Let's cache the rejections of the mined transaction IDs using a new `Mined` error? ("transaction was committed to the best chain")
> 
> I thought it might as well reject every mined id from blocks on the chain, let me know if it should only reject those that were removed from the verified transactions.

Yes, this is better, just in case our mempool didn't have those transactions in it yet.

zebrad/src/components/mempool/storage.rs
```diff
@@ -78,6 +81,12 @@ pub enum SameEffectsChainRejectionError {
     #[error("best chain tip has reached transaction expiry height")]
     Expired,
 
+    #[error("transaction outpoints or nullifiers were committed to the best chain")]
```
teor2345: Trivial clarification:

```suggestion
    #[error("transaction inputs were spent, or nullifiers were revealed, in the best chain")]
```

zebrad/src/components/mempool/storage.rs
```diff
@@ -276,22 +285,87 @@ impl Storage {
...
+            })
+            .collect();
+
+        let num_removed_duplicate_spend = self
+            .verified
+            .remove_all_that(|tx| duplicate_spend_ids.contains(&tx.transaction.id));
+
+        for &mined_id in mined_ids {
+            self.reject(
+                UnminedTxId::Legacy(mined_id),
```
teor2345, line 358: Does this reject v4 and v5 transactions with that `mined_id`?

In some cases, `Legacy` only applies to v4 transactions. But I think the code is correct, I just want to check (and maybe it needs a comment).

zebrad/src/components/mempool/storage/tests/prop.rs
```diff
@@ -367,7 +367,8 @@ proptest! {
         let inserted_transactions: HashSet<_> = input.transactions().filter_map(|transaction| {
             let id = transaction.transaction.id;
 
-            storage.insert(transaction.clone()).ok().map(|_| id)}).collect();
+            storage.insert(transaction.clone()).ok().map(|_| id)
+        }).collect();
```
teor2345: Nitpick: I'm surprised rustfmt accepts this, do we want to revert this whitespace change?

teor2345 (commented):
I think this looks good, just a few minor clarifications.

~Do you need help with the test failures?~
~I know we've done some fixes on the `main` branch, so I'll update this PR to the latest `main`.~

Edit: it was an unrelated test failure

teor2345:
@mergifyio update

teor2345:
Ah, the test failure was a ssh issue:
> Connection timed out during banner exchange

https://github.com/ZcashFoundation/zebra/actions/runs/3300645552/jobs/5445970948#step:6:112

teor2345 (dismissed):
Looks good, happy for this to go in without any of my suggestions.
(Or anyone can re-approve once they are done.)

zebrad/src/components/mempool/storage.rs
arya2: You're right, this needs a comment, added in [`c0e2f90`](https://github.com/ZcashFoundation/zebra/pull/5434/commits/c0e2f90c87856eb4a8f9a21d8e2aea98fc726bea). It only uses the mined_id so it should reject v5 transactions as well.

zebrad/src/components/mempool/storage/tests/prop.rs
arya2: It seems rustfmt doesn't check the prop tests at all, I moved the code elsewhere to be formatted and updated it in [`51b841b`](https://github.com/ZcashFoundation/zebra/pull/5434/commits/51b841b6d98289211ec1f259d67922e5ed20c6e8).

mpguerra (approved):

zebrad/src/components/mempool/storage/tests/prop.rs
upbqdn: Yea, the test is in the `proptest!` macro, and rustfmt doesn't check code passed to macros. I wonder if there's ever going to be any support at all since code passed to macros doesn't have to follow Rust's syntax.

PR #5448, opened 2022-10-21
fix(ci): Use correct lwd full sync patch job name

teor2345:
## Motivation

We got a workflow patch job name wrong, so Mergify isn't merging PRs.


teor2345:
This is a trivial typo, so I'm going to admin-merge it.

Issue #5436, opened 2022-10-20
Broken link in Zebra documentation Config section


mpguerra:
## Description

In the `zebrad` config section of the [documentation](https://doc.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html) the `ConsensusSection` link gives a `404 not found` error.

### Steps to Reproduce

Go to https://doc.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html and click on `ConsensusSection` 

![screenshot_from_2022-10-20_06-10-47](https://user-images.githubusercontent.com/1311133/196960488-1037bee9-e7ea-4602-8b2f-df7c77113f7d.png)

The link points to  https://doc.zebra.zfnd.org/zebra_consensus/zebra_consensus/config/struct.Config.html but maybe it should point to: https://doc.zebra.zfnd.org/zebra_consensus/config/struct.Config.html instead?

oxarbitrage:
In a quick look this is weird. In the internal docs the link is not broken https://doc-internal.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html

and i was not able to find anything in the code that can be causing it. I saw that zebra-consensus dependency is called with a trailing `/` in zebrad: https://github.com/ZcashFoundation/zebra/blob/main/zebrad/Cargo.toml#L72 while all the others do not trail. 

I don't think this is the cause here but we can try.

teor2345:
This seems like a rustdoc bug, rather than a Zebra bug.

We might be able to work around the bug by naming the types directly, rather than renaming them on import.

For example:
* ConsensusSection -> zebra_consensus::Config

oxarbitrage:
Unfortunately, the refactor we made in https://github.com/ZcashFoundation/zebra/pull/5460 do NOT fix the link issue: https://doc.zebra.zfnd.org/zebrad/config/struct.ZebradConfig.html

mpguerra:
This doesn't seem to be happening anymore...

PR #5460, opened 2022-10-21
refactor(config): Move configs to components

oxarbitrage:
## Motivation

This was inspired by https://github.com/ZcashFoundation/zebra/issues/5436 where there is a broken link and we don't know why. 

I am not sure if this will fix the link issue but in my opinion the docs about the config and also the code will be a bit more clear if each `Config` is where it needs to be.

config

## Solution

- Move the config code to each component (https://github.com/ZcashFoundation/zebra/commit/c21d0be05848ce1437f0c93a2a50506c135d3b04) and make needed fixes to make them work that way.
- Remove trailing slash from zebrad Cargo.toml when importing consensus (https://github.com/ZcashFoundation/zebra/commit/68a681ee904ee5db26b1f2cff9f7f09dc2ee92d2)
- Remove a doc warning that will be visible when running `cargo doc`. This is because the code the link is pointing to is behind a rust feature, so i prefer to just remove the link in those cases (https://github.com/ZcashFoundation/zebra/commit/5572f748efea36091438217984b2e1d4d73e2b1e)

## Review

I think anyone can review. In regards to the priority i think it can depend, we could want to add it to the new audit tag we are going to do. If so the priority can be medium, low otherwise.

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -7,6 +7,7 @@ use std::{cmp::max, collections::HashSet, pin::Pin, sync::Arc, task::Poll, time:
 use color_eyre::eyre::{eyre, Report};
 use futures::stream::{FuturesUnordered, StreamExt};
 use indexmap::IndexSet;
+use serde::{Deserialize, Serialize};
 use tokio::time::sleep;
 use tower::{
     builder::ServiceBuilder, hedge::Hedge, limit::ConcurrencyLimit, retry::Retry, timeout::Timeout,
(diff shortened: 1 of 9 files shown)


zebrad/src/components/sync.rs
```diff
@@ -204,6 +205,83 @@ const SYNC_RESTART_DELAY: Duration = Duration::from_secs(67);
 /// a denial of service on those peers.
 const GENESIS_TIMEOUT_RETRY: Duration = Duration::from_secs(5);
 
+/// Sync configuration section.
```
teor2345, line 208: Optional: This module is already over 1000 lines long, would you mind putting the config in a `sync/config.rs` submodule?

teor2345 (commented):
Looks good, just wondering if we can make a submodule to keep a file from getting too big?

What's the priority of this PR?

teor2345:
> I am not sure if this will fix the link issue but in my opinion the docs about the config and also the code will be a bit more clear if each `Config` is where it needs to be.

Can we build the docs and check if it fixes the links?

oxarbitrage:
> Can we build the docs and check if it fixes the links?

I don't think we should spend time doing this right now. Links work locally and making the CI build docs for this PR, push them to the website (https://doc.zebra.zfnd.org) could be hard (i honestly don't know).

The idea here is to make the refactor because i think it is the right thing to do, then when the CI build the docs we can figure out if the link was fixed and close the related issue or keep it open, but we will have the refactor done (which i think we should do anyway).

teor2345:
> > Can we build the docs and check if it fixes the links?
> 
> I don't think we should spend time doing this right now. Links work locally and making the CI build docs for this PR, push them to the website (https://doc.zebra.zfnd.org) could be hard (i honestly don't know).
> 
> The idea here is to make the refactor because i think it is the right thing to do, then when the CI build the docs we can figure out if the link was fixed and close the related issue or keep it open, but we will have the refactor done (which i think we should do anyway).

Sure, that makes sense, let's give it a shot!

teor2345 (approved):

Issue #5455, opened 2022-10-21
Populate `BlockTemplate` fields using a new state request


oxarbitrage:
## Motivation

Whatever is not fixed values(https://github.com/ZcashFoundation/zebra/issues/5452), coinbase transaction(https://github.com/ZcashFoundation/zebra/issues/5453) or mempool transaction(https://github.com/ZcashFoundation/zebra/issues/5454) will be state or Rust API values.

## Tasks

Return state values in the `getblocktemplate` RPC method response:
* `previousblockhash`
* `height`
* `defaultroots`
  * `chainhistoryroot`
* `target` (same as `bits`, but in a different format)
* `mintime`

These values should come from a new state request based on recent best blocks. We want to use a single state request for performance.

Return local node clock value in the `getblocktemplate` RPC method response:
* `curtime` - using the same standard library time function as the `time` field consensus rule check

## Related Tickets

Depend on https://github.com/ZcashFoundation/zebra/issues/5451
Can depend on https://github.com/ZcashFoundation/zebra/issues/5452 for identification

teor2345:
@oxarbitrage I think the fields we need here are:

New state request based on recent best blocks:
* `previousblockhash`
* `height`
* `defaultroots`
  * `chainhistoryroot`
* `target` (same as `bits`, but in a different format)
* `mintime`

Standard library time function:
* `curtime`

Can I edit the ticket, or did you want to?

oxarbitrage:
Please feel free to edit any of the getblocktemplate tickets. I did some some work already in the [first one](https://github.com/ZcashFoundation/zebra/issues/5451) but none in the rest of them.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5455?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage @teor2345 @upbqdn

teor2345:
PR #5659 populates all of these fields except `defaultroots.chainhistoryroot` (and the 5 fields that are copies of it).

@mpguerra @oxarbitrage did you want to split that out into a separate ticket?

mpguerra:
I think @oxarbitrage is working on them in a separate PR so I think it's ok to just keep it in this ticket.

Issue #5454, opened 2022-10-21
Populate transactions of `BlockTemplate`


oxarbitrage:
## Motivation

The `transactions` field of the `BlockTemplate`  response is an array of `BlockTemplateTx` structure with the transaction data and other fields.

## Tasks

- Create `BlockTemplateTx`
- bring transactions from the mempool and populate the `transactions.data` fields. https://github.com/ZcashFoundation/zebra/issues/5235#issuecomment-1285189453
  - the `depends` field is always an empty list, Zebra's mempool doesn't support dependent transactions
- Populate the additional fields of `BlockTemplateTx`

Depends on https://github.com/ZcashFoundation/zebra/issues/5451

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5454?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage @teor2345 @upbqdn

teor2345:
I split ZIP-317 transaction selection out of this ticket into #5473.

It seems like separate work, and it depends on some details that the `zcashd` developers haven't decided yet.

Issue #5453, opened 2022-10-21
Populate transparent coinbase transaction data for `BlockTemplate`


oxarbitrage:
## Motivation

The `coinbasetxn`  field of the `BlockTemplate` is a struct:(https://github.com/ZcashFoundation/zebra/issues/5235#issuecomment-1285174731) and can be complex enough to implement it in a separated ticket.

## Specifications

- [x] [NU5 onward] The transaction version number MUST be 4 or 5. If the transaction version number is 4 then the version group ID MUST be 0x892F2085. If the transaction version number is 5 then the version group ID MUST be 0x26A7270A.
- [x] [NU5 onward] If effectiveVersion ≥ 5, the nConsensusBranchId field MUST match the consensus branch ID used for SIGHASH transaction hashes, as specified in [ZIP-244].
- [ ] The total value in zatoshi of transparent outputs from a coinbase transaction, minus vbalanceSapling, minus vbalanceOrchard, MUST NOT be greater than the value in zatoshi of block subsidy plus the transaction fees paid by transactions in this block.
- [x] [NU5 onward] In a version 5 coinbase transaction, the enableSpendsOrchard flag MUST be 0.
- [x] [NU5 onward] In a version 5 transaction, the reserved bits 2 .. 7 of the flagsOrchard field MUST be zero.
- [x] A coinbase transaction for a block at block height greater than 0 MUST have a script that, as its first item, encodes the block height height as follows. For height in the range {1 .. 16}, the encoding is a single byte of value 0x50 + height. Otherwise, let heightBytes be the signed little-endian representation of height, using the minimum nonzero number of bytes such that the most significant byte is < 0x80. The length of heightBytes MUST be in the range {1 .. 5}. Then the encoding is the length of heightBytes encoded as one byte, followed by heightBytes itself. This matches the encoding used by Bitcoin in the implementation of [BIP-34] (but the description here is to be considered normative).
- [x] A coinbase transaction script MUST have length in {2 .. 100} bytes.
- [x] [NU5 onward] The nExpiryHeight field of a coinbase transaction MUST be equal to its block height .

https://zips.z.cash/protocol/protocol.pdf#txnconsensus

## Tasks

Shielded coinbase is out of scope, most miners don't use it.

- Create `CoinBaseTxn` struct.
- Populate on `getblocktemplate` response
  - The coinbase transaction has the following fixed values:
    - `required`: `true`
  - Here are the fields that are calculated from the transactions, including the coinbase transaction:
[...]

Issue #5452, opened 2022-10-21
Populate `Blocktemplate` responses with fixed values data


oxarbitrage:
## Motivation

Several `BlockTemplate`  fields introduced in https://github.com/ZcashFoundation/zebra/issues/5451 are fixed values. 

## Tasks

- [ ] Add fixed values to the response of the `getblocktemplate` RPC call. These values should use the relevant Zebra constants, or create a new constant.
  - `version`: `4` - existing block version number constant from block parser
  - `mutable`: `["time", "transactions", "prevblock"]` - new constant list of strings
  - `noncerange`: `00000000ffffffff` - new min and max nonce constants, set to `u32::MIN` and `u32::MAX`
  - `sigoplimit`: `20000` - existing sigops limit constant from consensus rule check
  - `sizelimit`: `2000000` - existing block byte limit constant from block parser
- [ ] Update the test created at 1- so it can also test for fixed values in the responses.

Depends on https://github.com/ZcashFoundation/zebra/issues/5451

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5452?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage @teor2345 @upbqdn

PR #5462, opened 2022-10-22
feat(rpc): introduce getblocktemplate rpc call with stub fields

### LINKED ISSUE #5451: Introduce `getblocktemplate` RPC method
Tasks:

- [ ] Add a `getblocktemplate` RPC method behind the `getblocktemplate-rpcs` rust feature.
- [ ] The method should have no arguments.
- [ ] Create the response structure `GetBlockTemplate` with fields: https://github.com/ZcashFoundation/zebra/issues/5235#issuecomment-1285174731
- [ ] Always return an empty `GetBlockTemplate`
- [ ] Test the rpc method is callable and returns a `GetBlockTemplate` type.

oxarbitrage:
## Motivation

This is pretty much what i have in mind for https://github.com/ZcashFoundation/zebra/issues/5451

It introduces the rpc call and all the fields using mostly strings as placeholders. All data will be filled in the other tickets, this is kind of the skeleton.

This is a draft, feel free to suggest changes. 

## Solution

Introduce `getblocktemplate` rpc methjod, return empty `GetblockTemplate` struct on call.

### Reviewer Checklist

## Follow Up Work

- https://github.com/ZcashFoundation/zebra/issues/5452
- https://github.com/ZcashFoundation/zebra/issues/5455
- https://github.com/ZcashFoundation/zebra/issues/5453
- https://github.com/ZcashFoundation/zebra/issues/5454

### DIFF
--- a/zebra-rpc/src/methods/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template.rs
@@ -171,7 +171,6 @@ where
                 },
                 transactions: vec![],
                 coinbasetxn: Coinbase {},
-                longpollid: empty_string.clone(),
                 target: empty_string.clone(),
                 mintime: 0,
                 mutable: vec![],
@@ -207,8 +206,6 @@ pub struct GetBlockTemplate {
     pub transactions: Vec<Transaction>,
     /// Add documentation.
     pub coinbasetxn: Coinbase,
-    /// Add documentation.
-    pub longpollid: String,
     /// Add documentation.
     pub target: String,
     /// Add documentation.
--- a/zebra-rpc/src/methods/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template.rs
@@ -35,8 +35,19 @@ pub trait GetBlockTemplateRpc {
     ///
     /// - If `index` is positive then index = block height.
     /// - If `index` is negative then -1 is the last known valid block.
+    /// - This rpc method is available only if zebra is built with `--features getblocktemplate-rpcs`.
     #[rpc(name = "getblockhash")]
     fn get_block_hash(&self, index: i32) -> BoxFuture<Result<GetBlockHash>>;
+
+    /// Documentation to be filled as we go.
+    ///
+    /// zcashd reference: [`getblocktemplate`](https://zcash-rpc.github.io/getblocktemplate.html)
+    ///
+    /// # Notes
+    ///
+    /// - This rpc method is available only if zebra is built with `--features getblocktemplate-rpcs`.
+    #[rpc(name = "getblocktemplate")]
+    fn get_block_template(&self) -> BoxFuture<Result<GetBlockTemplate>>;
 }
 
 /// RPC method implementations.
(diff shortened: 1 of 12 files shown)


zebra-rpc/src/methods/get_block_template.rs
```diff
@@ -139,8 +150,106 @@ where
...
+                finalsaplingroothash: empty_string.clone(),
+                defaultroots: DefaultRoots {
+                    merkleroot: empty_string.clone(),
+                    chainhistoryroot: empty_string.clone(),
+                    authdataroot: empty_string.clone(),
+                    blockcommitmentshash: empty_string.clone(),
+                },
+                transactions: vec![],
+                coinbasetxn: Coinbase {},
+                longpollid: empty_string.clone(),
```
teor2345: Zebra doesn't support the "long polling" feature, so we shouldn't include the `longpollid` field.
(It's an optional `getblocktemplate` feature.)

teor2345 (changes requested):
Looks good, but I think `longpollid` is out of scope for now.

zebra-rpc/src/methods/get_block_template.rs
teor2345: Here's the spec:

zebra-rpc/src/methods/get_block_template.rs
```diff
@@ -139,8 +150,106 @@ where
...
+    /// Add documentation.
+    pub mutable: Vec<String>,
+    /// Add documentation.
+    pub noncerange: String,
+    /// Add documentation.
+    pub sigoplimit: u32,
+    /// Add documentation.
+    pub sizelimit: u32,
+    /// Add documentation.
+    pub curtime: u32,
```
teor2345: Optional: For some other RPCs, we've used the Rust standard where words are separated by underscores, then renamed the RPC when serializing.

Did you want to do that here?

https://github.com/ZcashFoundation/zebra/blob/a2a75059d39dcd9aaabb9609a84e0dc59b966535/zebra-rpc/src/methods.rs#L1009

zebra-rpc/src/methods/get_block_template.rs
```diff
@@ -139,8 +150,106 @@ where
...
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct DefaultRoots {
+    /// Add documentation.
+    pub merkleroot: String,
+    /// Add documentation.
+    pub chainhistoryroot: String,
+    /// Add documentation.
+    pub authdataroot: String,
+    /// Add documentation.
+    pub blockcommitmentshash: String,
```
teor2345: Optional: same here, do we want to use underscores and renames?

teor2345 (commented):
I wonder if we want to use underscores to match the rest of the RPC implementations.

zebra-rpc/src/methods/get_block_template.rs
oxarbitrage: That is one less to worry about, thanks! I removed it at https://github.com/ZcashFoundation/zebra/pull/5462/commits/5756a77f89d8597ce146da071878d978c4706711

zebra-rpc/src/methods/get_block_template.rs
oxarbitrage: Yea, we definitely want to do this. https://github.com/ZcashFoundation/zebra/pull/5462/commits/5756a77f89d8597ce146da071878d978c4706711
Thanks.

zebra-rpc/src/methods/get_block_template.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/commit/5756a77f89d8597ce146da071878d978c4706711

zebra-rpc/src/methods/get_block_template.rs
```diff
@@ -139,8 +150,118 @@ where
...
+                bits: empty_string,
+                height: 0,
+            })
+        }
+        .boxed()
+    }
+}
+/// Documentation to be added after we document all the individual fields.
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct GetBlockTemplate {
```
arya2: I think it would be nice to separate the types for each method into their own files.

oxarbitrage: I think this is a good idea. We have already this file named `get_block_template.rs`. I am thinking on renaming it to `get_block_template_rpcs.rs` and then do `get_block_template.rs` to hold `GetBlockTemplate`. 

Then `default_roots.rs` for `DefaultRoots`, `transaction.rs` for `Transaction` and `coinbase.rs` for `Coinbase`.

Let me know if something like that is what you have in mind or something different.

oxarbitrage: Another option can be to create a `types` module/folder and drop the type files inside it.

arya2: I had a file for each method in mind, i.e. `methods/types/getblocktemplate.rs`, but on second thought that would be confusing, and I like your suggestion better.

teor2345 (approved):
Looks good!

I think this is a high priority, because it renames the module that all the other RPCs are in. If we don't get it merged soon, it will cause code conflicts.

PR #5435, opened 2022-10-20
ci(getblocktemplate): run getblocktemplate-rpcs feature tests in CI

oxarbitrage:
## Motivation

One option for https://github.com/ZcashFoundation/zebra/issues/5405 is to run a new test task with the getblocktemplate-rpcs feature enabled.

## Solution

Run a new task with `getblocktemplate-rpc` feature enabled. This will run the tests with and without the feature in a separated task for each case. I think is what we want to make sure everything pass when the feature is enabled/disabled in addition to run extra tests when enabled.

New tests can be seen in the logs at: 
https://github.com/ZcashFoundation/zebra/actions/runs/3291277916/jobs/5425229470#step:15:4896
https://github.com/ZcashFoundation/zebra/actions/runs/3291277916/jobs/5425229470#step:15:4912

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [x] How do you know it works? Does it have tests?

PR #5474, opened 2022-10-25
change(ci): Add patch jobs for lightwalletd build and getblocktemplate-rpcs tests

### LINKED ISSUE #4882: Add zcash-lightwalletd / Build images to required jobs
## Motivation

If this job fails because we broke it in a PR, we shouldn't merge the PR.

### Tasks

- [ ] add a patch job that runs when the real job is skipped
- [ ] enable the job in branch protection rules (needs Zebra admin permissions)

teor2345:
## Motivation

We want to add patch jobs so we can require the lightwalletd build and getblocktemplate-rpcs tests to succeed in CI, before merging PRs.

This is a follow-up to PR #5435.
It also closes ticket #4882.

## Solution

- Add the patch jobs
- Add a missing lightwalletd build concurrency rule

## Review

@oxarbitrage created PR #5435, but @gustavovalverde is more familiar with how patch jobs work.

### Reviewer Checklist

## Follow Up Work

- [ ] After this PR merges, add the two jobs to the branch protection rules (requires admin access)

### DIFF
--- a/.github/workflows/continous-integration-docker.patch.yml
+++ b/.github/workflows/continous-integration-docker.patch.yml
@@ -42,6 +42,12 @@ jobs:
     steps:
       - run: 'echo "No build required"'
 
+  test-all-getblocktemplate-rpcs:
+    name: Test all with getblocktemplate-rpcs feature
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


gustavovalverde (approved):
LGTM

Issue #4692, opened 2022-06-27
Check address and key lengths when parsing strings


teor2345:
## Scheduling

We don't need to fix these bugs until we start full wallet work.

## Motivation

Zebra doesn't check the length of address or key strings when parsing.

This might lead to loss of funds.

### Specifications

The Zcash consensus rules say that keys and addresses should be a specific length.

### Designs

In methods that handle untrusted data:
- `From`
- `FromStr`

Check that the key or address is the correct length, and return an error if there is any trailing data.

This issue applies to all keys and addresses in Zebra.

teor2345:
Also fixed by PR #5476 by removing unused address and key parsing code.

PR #5476, opened 2022-10-25
cleanup(cryptography): Remove unused shielded key and address code

### LINKED ISSUE #2041: Orchard: derive OutgoingCipherKey ( Sapling and Orchard)
Stub: 

```
//     From<(
//         OutgoingViewingKey,
//         ValueCommitment,
//         NoteCommitment,
//         EphemeralKey,
//     )> for OutgoingCipherKey
// {
//     type Error = &'static str;

//     fn from((): ()) -> Result<Self, Self::Error> {
//         Self(prf_ock())
//     }
// }
```

### LINKED ISSUE #2496: Handle a ivk == ⊥ (KA^Orchard private key) when generating an Orchard Spending Key
## Motivation

Currently, `SpendingKey::new` does not seem to handle the case where the derived ivk is ⊥, which should trigger retrying with a new key per the spec.

### Specifications

https://zips.z.cash/protocol/protocol.pdf#orchardkeycomponents

### Designs

Note that this is not urgent (AFAIK) since it's unlikely that ivk will be ⊥ per Theorem 5.4.4, but we should eventually match the spec.

## Related Work

N/A

### LINKED ISSUE #4691: Stop assuming testnet when parsing keys and addresses
## Scheduling

We don't need to fix these bugs until we start full wallet work.
(These bugs only affect Sprout, Sapling, and Orchard keys and addresses.)

## Motivation

When Zebra finds an unrecognised network in an address, it sometimes:
- panics, or
- assumes testnet.

This can lead to denial of service or loss of funds.

### Specifications

The Zcash consensus rules say that unrecognised networks should result in an error.

### Designs

In methods that handle untrusted data:
- `From`
- `FromStr`
- `ZcashDeserialize`

For addresses and keys in:
- Sprout
- Sapling
- Orchard

Stop panicking when an unrecognised human-readable part is found:
- return an error to the caller instead
- check all callers to make sure they don't panic

Stop assuming Mainnet or Testnet:
- Remove `Default for Network`, and replace all uses with traits like `ToAddressWithNetwork`
- Remove default matches (`_ =>`) and return an error instead
- Check for direct uses of `Mainnet` or `Testnet` as defaults, and remove them

### LINKED ISSUE #5446: Audit: remove zebra-chain cryptographic implementations that are unused
## Motivation

We want to remove unused full-node or full-wallet cryptographic code in zebra-chain, because it has known bugs. We don't want to spend audit time reviewing it or fixing it at the moment.

We can add it back in later if we need to.

Split out of #5214.

teor2345:
## Motivation

We want to remove unused shielded key and address code before the audit, because it contains known bugs that break consensus rules. 

We also want to remove all the code that is impacted by these known bugs:
- #4691
- the key parsing panics we found in PR #4732
- #2041
- #4692
- unimplemented cryptographic algorithms
- there might be 2-3 other bugs in our cryptographic parsing or conversion code. For example, panicking when point conversions fail, and panicking on malformed addresses.

Close #5446
Close #4691
Close #2496 
Close #2041

## Solution

This PR removes 3000 lines of unused cryptographic code:
- all unused shielded key code
- dependent unused shielded address, note, commitment, nullifier, and test code

As a consequence, it removes:
- unimplemented cryptographic algorithms and conversions
- buggy cryptographic algorithms and conversions
- known panics during conversions
- untested code

It also makes Zebra's code easier to review and audit, because some of the removed types had the same name as the types from the ECC crates.

## Review

Anyone can review this PR, it's a high priority because it's needed for the audit.

### Reviewer Checklist


teor2345:
I found this extra bug:
- #2041

zebra-chain/src/orchard/keys.rs
```diff
@@ -60,6 +59,9 @@ fn prp_d(K: [u8; 32], d: [u8; 11]) -> [u8; 11] {
 // TODO: This is basically a duplicate of the one in our sapling module, its
 // definition in the draft Nu5 spec is incomplete so I'm putting it here in case
 // it changes.
+//
+// Review question: @dconnolly is this code still incomplete?
+// It's used to generate new note commitments from notes.
```
teor2345: @dconnolly can you check if this code is still incomplete?

zebra-chain/src/orchard/keys.rs
```diff
@@ -278,6 +280,8 @@ impl Eq for SpendValidatingKey {}
 
 impl From<[u8; 32]> for SpendValidatingKey {
     fn from(bytes: [u8; 32]) -> Self {
+        // Review question: @dconnolly do we need to delete or fix this panic?
+        // It's used as part of the full viewing key, which is used to generate new spending keys.
```
teor2345: @dconnolly can you check if this code is a security risk, because it panics?

zebra-chain/src/orchard/keys.rs
```diff
@@ -579,7 +583,11 @@ impl PartialEq for FullViewingKey {
 #[derive(Copy, Clone)]
 pub struct IncomingViewingKey {
     dk: DiversifierKey,
+
     // TODO: refine type, so that IncomingViewingkey.ivk cannot be 0
+    //
+    // Review question: @dconnolly do we need to fix this bug before the audit?
+    // It's used to generate valid TransmissionKeys, is that code used in production?
```
teor2345: @dconnolly is this a bug we need to fix before the audit?

zebra-chain/src/orchard/keys.rs
```diff
@@ -675,11 +683,12 @@ impl TryFrom<FullViewingKey> for IncomingViewingKey {
     /// <https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents>
     /// <https://zips.z.cash/protocol/nu5.pdf#concreteprfs>
     #[allow(non_snake_case)]
-    #[allow(clippy::unwrap_in_result)]
     fn try_from(fvk: FullViewingKey) -> Result<Self, Self::Error> {
         let mut M: BitVec<u8, Lsb0> = BitVec::new();
 
         // I2LEBSP_l^Orchard_base(ak)︁
+        //
+        // Review question: @dconnolly do we need to fix this panic, or is it impossible?
```
teor2345: @dconnolly can you check if this code is a security risk, because it panics?

zebra-chain/src/orchard/keys.rs
```diff
@@ -986,6 +995,8 @@ impl From<[u8; 32]> for TransmissionKey {
     /// Attempts to interpret a byte representation of an affine point, failing
     /// if the element is not on the curve or non-canonical.
     fn from(bytes: [u8; 32]) -> Self {
+        // Review question: @dconnolly do we need to fix this panic, or is it impossible?
+        // It's used as part of the Orchard address.
```
teor2345: @dconnolly can you check if this code is a security risk, because it panics?

teor2345:
Actually, I think I might be able to remove some of the broken Orchard code here.

zebra-chain/src/orchard/keys.rs
dconnolly: This should be an `impl TryFrom` because the conversion from an arbitrary array of 32 bytes may not be a valid key; so instead of panic'ing it should return a `Result::Err`

teor2345:
> Actually, I think I might be able to remove some of the broken Orchard code here.

I worked out why it's tricky with Orchard - we need most of the key types to generate valid random Orchard addresses for tests.

So I'm going to move the keys that aren't used in production into a test module, and add a note about what needs to be done to the code to make it production-quality. (We're not getting test code audited, so that achieves the goals of the ticket.)

I might not get to finish it all today.

zebra-chain/src/orchard/keys.rs
teor2345: Thanks, I moved it to a test-only module.

zebra-chain/src/orchard/keys.rs
teor2345: I moved this to a test-only module.

zebra-chain/src/orchard/keys.rs
teor2345: This panic is impossible, I added a comment. But I'd still like a cryptographer to review this.

zebra-chain/src/orchard/keys.rs
teor2345: I moved this to a test-only module.

zebra-chain/src/orchard/keys.rs
teor2345: I'd still like a cryptographer to review this code, it looks like it might be used in production.

zebra-chain/src/orchard/arbitrary/keys/tests/generate.rs
```diff
@@ -0,0 +1,34 @@
+//! Key and address gewneration tests for Orchard.
+
+use rand_core::OsRng;
+
+use crate::{
+    orchard::{address::Address, arbitrary::keys::*, keys::*},
```
teor2345:
```suggestion
    orchard::{address::Address, arbitrary::keys::*},
```

zebra-chain/src/orchard/arbitrary/keys.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// <https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents>
+    /// <https://zips.z.cash/protocol/nu5.pdf#concreteprfs>
+    #[allow(non_snake_case)]
+    #[allow(clippy::unwrap_in_result)]
+    fn try_from(fvk: FullViewingKey) -> Result<Self, Self::Error> {
+        let mut M: BitVec<u8, Lsb0> = BitVec::new();
+
+        // I2LEBSP_l^Orchard_base(ak)︁
+        //
+        // Converting a Point to bytes and back is infalliable.
```
teor2345: I added this comment, is it correct?

zebra-chain/src/orchard/keys.rs
conradoplg: The code looks correct regarding the spec, but I don't think it's used anywhere. `NoteCommitment::new` is not used anywhere. Ultimately this `prf_expand` is used to implement `Arbitrary` for orchard `TransmissionKey` and `Address` which don't seem to be required currently.

zebra-chain/src/orchard/arbitrary/keys.rs
conradoplg: Yep, this is correct, there's just a typo. I think this piece of code can be ultimately removed, though

```suggestion
        // Converting a Point to bytes and back is infallible.
```

zebra-chain/src/orchard/keys.rs
teor2345: I think you're right!

Originally I thought we used it to generate arbitrary Orchard addresses, notes, note commitments, and transactions. But it turns out our EncryptedNote and WrappedNote key generation just uses random bytes.

Do you think it's worth removing it now? I don't mind either way.

zebra-chain/src/orchard/arbitrary/keys.rs
teor2345: Do you think it's worth removing this code now? I don't mind either way.

conradoplg: I think so, I don't see a reason to keep it if we want to make things easier for auditors.

teor2345: It seems like I can remove this easily, I'll push a commit.

teor2345:
Ok, I think I'm done here.

See the PR description for a summary of these changes, it's almost all removed code.

oxarbitrage:
This seems to be done. @conradoplg or @dconnolly  , can you please approve if that is the case ? Thank you.

zebra-chain/src/orchard/note/nullifiers.rs
```diff
@@ -1,46 +1,11 @@
-#![allow(clippy::unit_arg)]
+//! Orchard nullifier types and conversions.
```
dconnolly, line 1: 👍

zebra-chain/src/sprout/keys.rs
```diff
@@ -1,185 +1,18 @@
...
-    /// <https://zips.z.cash/protocol/protocol.pdf#concreteprfs>
-    fn from(spending_key: SpendingKey) -> ReceivingKey {
-        let derived_bytes = prf_addr(spending_key.bytes, 0);
-
-        ReceivingKey::from(derived_bytes)
-    }
-}
-
-/// Derived from a _SpendingKey_.
+/// Derived from a sprout Spending Key.
```
dconnolly:
```suggestion
/// Derived from a Sprout _spending key.
```

zebra-chain/src/sprout/keys.rs
```diff
@@ -1,185 +1,18 @@
...
-/// Keys are associated.
-mod sk_magics {
-    pub const MAINNET: [u8; 2] = [0xAB, 0x36];
-    pub const TESTNET: [u8; 2] = [0xAC, 0x08];
-}
-
-/// PRF^addr is used to derive a Sprout shielded payment address from
-/// a spending key, and instantiated using the SHA-256 compression
-/// function.
+/// A Sprout paying key.
```
dconnolly:
```suggestion
/// A Sprout _paying key_.
```

zebra-chain/src/sprout/keys.rs
```diff
@@ -1,185 +1,18 @@
-//! Sprout key types
+//! Sprout key types.
+//!
+//! Unused key types are not implemented, see PR #5476.
```
dconnolly, line 3: 👍

dconnolly (approved):
lgtm

If we do wallet work / scanning work in the future we can dig the sapling/orchard stuff out of the history and fix it up, if needed

Issue #5015, opened 2022-08-31
Get transactions from the non-finalized state in the send transactions test


teor2345:
## Motivation

Currently, the send transactions test is very slow, because it:
- copies the entire Zebra finalized state directory
- syncs to the tip
- gets transactions from the copied finalized state
- runs the test on the old finalized state

This is ok for now, but it might become a problem if the state gets much bigger, or we need to modify that test a lot.

### Designs

Instead, the test could:
- use the original Zebra cached state directory
- sync to the tip
- get transactions from at least 3 blocks in the non-finalized state via JSON-RPC
- run the test on the updated finalized state (which won't have all those non-finalized transactions)

teor2345:
@arya2 here's an example of an RPC call from the tests:
https://github.com/ZcashFoundation/zebra/blob/7b47aac3703afe2e5efb45cdeb537a238ef45d4e/zebrad/tests/acceptance.rs#L2016-L2029

There's no need to use `zebra-rpc` or anything complicated to get the transaction data, because we're not really testing that RPC here.

arya2:
@teor2345 Is there an easy way to use the original Zebra cached state directory?

Update: disregard, I was thinking only temp files could be accessed from the spawned process, but I found `spawn_zebrad_for_rpc` and realized we're already using the original cached state in some tests.

PR #5480, opened 2022-10-25
change(ci): only run the send transaction test on the main branch

### LINKED ISSUE #5426: Move the send transactions test to the main branch
## Motivation

The send transactions test is very slow and copies a lot of data.

We want to fix this test in #5015. But until then, we want to move it to the `main` branch.

If we move it to the `main` branch, we'll have a better chance of catching any changes that break this RPC.

### Designs

Move the workflow to `main` like we did in PR #5393.

## Rejected Alternatives

If we turn this test off, we need to either:
- test it manually before every release, or
- accept the risk that it might be broken in a release.

We're not planning on making any changes to this RPC, but we are working on some related RPC and mempool code that it calls.

Move the job definition into a separate workflow file, and turn off that workflow in GitHub Actions.

arya2:
## Motivation

The send transaction test is slow, so we want to move it to the main branch until we've made the changes in #5015 

Closes #5426 

## Review

### Reviewer Checklist

## Follow Up Work

Revert this change once #5015 has been addressed.

### DIFF
--- a/.github/workflows/continous-integration-docker.patch-always.yml
+++ b/.github/workflows/continous-integration-docker.patch-always.yml
@@ -26,7 +26,7 @@ jobs:
       - run: 'echo "No build required"'
 
   lightwalletd-transactions-test:
-    name: lightwalletd tip send
+    name: lightwalletd tip send / Run lwd-send-transactions test
     runs-on: ubuntu-latest
     steps:
       - run: 'echo "No build required"'
--- a/.github/workflows/continous-integration-docker.patch-always.yml
+++ b/.github/workflows/continous-integration-docker.patch-always.yml
@@ -24,3 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: 'echo "No build required"'
+
+  lightwalletd-transactions-test:
+    name: lightwalletd tip send
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -500,7 +500,6 @@ jobs:
   #
   # Runs:
   # - after every PR is merged to `main`
-  # - on every PR update
   #
   # If the state version has changed, waits for the new cached states to be created.
   # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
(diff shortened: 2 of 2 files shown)


arya2:
Should this PR allow the send transaction test to be triggered manually?

.github/workflows/continous-integration-docker.patch-always.yml
```diff
@@ -24,3 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: 'echo "No build required"'
+
+  lightwalletd-transactions-test:
+    name: lightwalletd tip send
```
teor2345: Patch job names must match the exact name of the original job, including the calling and called workflow's job names.

This is the name that GitHub checks in the existing branch protection rule:
```suggestion
    name: lightwalletd tip send / Run lwd-send-transactions test
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -509,7 +508,7 @@ jobs:
     name: lightwalletd tip send
     needs: lightwalletd-full-sync
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' }}
+    if: ${{ !cancelled() && !failure() && github.event_name == 'push' && github.ref_name == 'main' && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' }}
```
teor2345: I'm not sure if some of these conditions are needed.

The manual workflow inputs will always have their default values during a `push` event, because they are only used for `workflow_dispatch` events.

```suggestion
    if: ${{ !cancelled() && !failure() && github.event_name == 'push' && github.ref_name == 'main' }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -522,6 +521,14 @@ jobs:
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
       lwd_state_dir: 'lwd-cache'
+    # We want to prevent multiple lightwalletd send transaction tests running at the same time,
```
teor2345, line 529: Optional: I added the standard concurrency rule to the PR, like PR #5393.

We might need this rule because the test runs for a long time (2-3 hours, or longer if the cached state is old). If we don't have this rule, the test could run multiple times, because:
1. This test is waiting on the running full sync in an older commit
2. A newer commit skips the full sync due to the full sync's concurrency rule, and then runs this test (because it has no concurrency rule)

teor2345 (changes requested):
Just a few minor details to fix up so we satisfy the branch protection rules.

Do you think we will ever want to run this job manually?

I'm not sure, because we're not actually generating a cached state with this job. So the only reason to run it manually would be if we updated the code. (Which we might do in #5015, or we might just leave it.)

Here's what we did to add a manual full lightwalletd sync:
https://github.com/ZcashFoundation/zebra/pull/5393/files#diff-4dc8170bbb198e5ddb8c8d186aa2cf20ee5117960f71ad9197b2a800ad878bceR31-R35

.github/workflows/continous-integration-docker.yml
arya2: Good point, thank you.

arya2:
> Do you think we will ever want to run this job manually?

I don't see any harm in making it available for testing #5015.

.github/workflows/continous-integration-docker.yml
arya2: Thank you!

teor2345 (approved):
Thanks, looks good!

PR #5489, opened 2022-10-27
cleanup(clippy): Use inline format strings

teor2345:
## Motivation

Nightly clippy has a new fix that inlines format strings.

## Solution

- Apply the automated fixes

## Review

Anyone can review this PR.

It has a lot of changes, but they are all automatically generated. You can check them using the clippy commands in each commit description, with nightly clippy.

This is a low priority change, but the risk of merge conflicts is high. So I'm not sure what priority to put here.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/tests/prop.rs
+++ b/zebra-chain/src/block/tests/prop.rs
@@ -40,7 +40,7 @@ proptest! {
     fn block_hash_display_fromstr_roundtrip(hash in any::<Hash>()) {
         let _init_guard = zebra_test::init();
 
-        let display = format!("{}", hash);
+        let display = format!("{hash}");
         let parsed = display.parse::<Hash>().expect("hash should parse");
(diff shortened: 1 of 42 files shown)


oxarbitrage (approved):
Yes, i think we should try to push this sooner than later. The changes made all look good. 

In regard to the lint itself, i understand the inline way can add more clarity in strings with a lot of arguments, in the majority of our cases we just use one or two variables in the format string which is pretty much the same readability. 

I am pretty sure it will take me a while to get used to it but as the lint will warn, it should be ok.

teor2345:
> I am pretty sure it will take me a while to get used to it but as the lint will warn, it should be ok.

The lint is only on nightly Rust for now, it won't be on stable until the end of December.

But this PR will reduce the amount of changes we need to make in December when the lint hits stable.

PR #5505, opened 2022-10-30
deps(bump): zcash_primitives, zcash_note_encryption, zcash_encoding, orchard, and zcash_script

### LINKED ISSUE #3831: Remove duplicated zcash_proofs
## Motivation

We forked `zcash_proofs` because we need Sprout downloading support https://github.com/zcash/librustzcash/pull/459, which causes it to be duplicated as a dependency since other crates we use also import the upstream version.

### Designs

After the PR is merged upstream:

- Remove reference to the fork
- Remove the exceptions from `deny.toml`

oxarbitrage:
## Motivation

We want to bump this 3 dependencies at the same time. This PR will replace the other 3 if we go this way:

- https://github.com/ZcashFoundation/zebra/pull/5497
- https://github.com/ZcashFoundation/zebra/pull/5498
- https://github.com/ZcashFoundation/zebra/pull/5500

It also works around the fact that [address()](https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/legacy.rs#L46) is now private.

I am pushing this now as we might want to add it to the release candidate. 

Close #3831 (removing all duplicate ECC dependencies, not just `zcash_proofs`)

## Solution

- Upgrade the 4 ECC dependencies.
- Upgrade zcash_script.
- Fix address issue.

### Reviewer Checklist


zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -290,8 +290,15 @@ pub(crate) fn transparent_output_address(
     output: &transparent::Output,
     network: Network,
 ) -> Option<transparent::Address> {
-    let script = zcash_primitives::legacy::Script::from(&output.lock_script);
-    let alt_addr = script.address();
+    // Create a `TxOut` primitive from a zebra `Output`.
+    let tx_out = zcash_primitives::transaction::components::TxOut {
```
oxarbitrage: We could move this code to an `Output` method `recipient_address()` if we want.

upbqdn: I'd say that would be handy.

upbqdn (dismissed):
This PR looks great. I was working on a similar one, but my workaround for `address()` wasn't as smooth.

@oxarbitrage I have updates that make Zebra compatible with https://github.com/ZcashFoundation/zcash_script/pull/41, should I push them as a PR into this one, or would it be better to make a standalone PR?

oxarbitrage:
> This PR looks great. I was working on a similar one, but my workaround for `address()` wasn't as smooth.

Apologies for working on something you were already working on. I didn't saw https://github.com/ZcashFoundation/zebra/pull/5500#issuecomment-1295563898

> @oxarbitrage I have updates that make Zebra compatible with https://github.com/ZcashFoundation/zcash_script/pull/41, should I push them as a PR into this one, or would it be better to make a standalone PR?

That is great, i think we might want to have all the ECC or our own dependencies up to date in the release candidate but i am not totally sure if it is that important.

If your changes are not too big i think we can try to push to this one and merge all the dependencies alltogether.

upbqdn:
> @oxarbitrage I have updates that make Zebra compatible with https://github.com/ZcashFoundation/zcash_script/pull/41, should I push them as a PR into this one, or would it be better to make a standalone PR?

I just realized it would be best if we wait until we merge https://github.com/ZcashFoundation/zcash_script/pull/41, and then do the rest of the updates in Zebra, so a standalone PR will be a better solution.

upbqdn:
> Apologies for working on something you were already working on.

No worries at all, the only blocker was the `address()` function being private.

upbqdn:
> I just realized it would be best if we wait until we merge https://github.com/ZcashFoundation/zcash_script/pull/41, and then do the rest of the updates in Zebra, so a standalone PR will be a better solution.

I actually pushed a draft PR with the updates #5506 without https://github.com/ZcashFoundation/zcash_script/pull/41 being merged so that we can see what changes are needed.

zebra-chain/Cargo.toml
```diff
@@ -42,11 +42,11 @@ x25519-dalek = { version = "2.0.0-pre.1", features = ["serde"] }
 
 # ECC deps
 halo2 = { package = "halo2_proofs", version = "0.2.0" }
-orchard = "0.2.0"
+orchard = "0.3.0"
 zcash_encoding = "0.1.0"
 zcash_history = "0.3.0"
-zcash_note_encryption = "0.1"
-zcash_primitives = { version = "0.7.0", features = ["transparent-inputs"] }
+zcash_note_encryption = "0.2"
```
teor2345: Nitpick, for consistency with the other version numbers:

```suggestion
zcash_note_encryption = "0.2.0"
```

teor2345 (commented):
Just want to tweak a version number

zebra-chain/src/primitives/zcash_primitives.rs
teor2345: We're trying to keep everything that calls `zcash_primitives` directly in this module, so it's easy to see what we're using.

But we could add a convenience method to `Output`, like we did for `AuthDigest` and `Transaction`.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -290,8 +290,15 @@ pub(crate) fn transparent_output_address(
...
+        // We just set the value to zero as we are only interested in the
+        // address derived from the output lock script.
+        value: zcash_primitives::transaction::components::Amount::zero(),
+        script_pubkey: zcash_primitives::legacy::Script::from(&output.lock_script),
+    };
+    let alt_addr = tx_out.recipient_address();
```
teor2345: This code feels like a bit of a workaround, which could break if `librustzcash` changes.

If the address is only valid in the context of an output, then we should convert the whole output before getting the address. (It doesn't seem to make any difference now, but the `zcashd` developers suggested using the whole output.)

I made this change in PR #5507, what do you think?

teor2345 (dismissed):
I have a suggestion, but it's not a blocker.

teor2345:
> I am pushing this now as we might want to add it to the release candidate.

Yes, we do, using outdated or duplicate cryptographic dependencies is a security vulnerability. (And it's a pain to audit.)

teor2345:
> error: linking with `cc` failed: exit status: 1
https://github.com/ZcashFoundation/zebra/actions/runs/3357475081/jobs/5563308196#step:15:463

This is only happening in the GitHub Actions Linux build, so I cleared this PR's GitHub Actions caches, and the `main` branch GitHub Actions caches.

zebra-chain/src/primitives/zcash_primitives.rs
oxarbitrage: Yea, that looks a bit better, I merged suggestion to here.

conradoplg (approved):

PR #5513, opened 2022-10-31
release(1.0.0-rc.1): bump versions and changelog for release 

oxarbitrage:
Depends-On: #5512

---
name: Release Checklist Template
about: Checklist of versioning to create a taggable commit for Zebra
title: ''
labels:
assignees: ''

---

### How to Increment Versions

Zebra follows [semantic versioning](https://semver.org).
Semantic versions look like: MAJOR`.`MINOR`.`PATCH[`-`TAG`.`PRE-RELEASE]

The [draft `zebrad` changelog](https://github.com/ZcashFoundation/zebra/releases) will have an automatic version bump. This version is based on [the labels on the PRs in the release](https://github.com/ZcashFoundation/zebra/blob/main/.github/release-drafter.yml).

Check that the automatic `zebrad` version increment is correct:

If we're releasing a mainnet network upgrade, it is a `major` release:
1. Increment the `major` version of _*all*_ the Zebra crates.
2. Increment the `patch` version of the tower crates.

If we're not releasing a mainnet network upgrade, check for features, major changes, deprecations, and removals. If this release has any, it is a `minor` release:
1. Increment the `minor` version of `zebrad`.
2. Increment the `pre-release` version of the other crates.
3. Increment the `patch` version of the tower crates.

Otherwise, it is a `patch` release:
1. Increment the `patch` version of `zebrad`.
2. Increment the `pre-release` version of the other crates.
3. Increment the `patch` version of the tower crates.

Zebra's Rust API is not stable or supported, so we keep all the crates on the same beta `pre-release` version.

### Version Locations

Once you know which versions you want to increment, you can find them in the:

zebrad (rc):
- [x] zebrad `Cargo.toml`
- [x] `zebra-network` protocol user agent: https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/constants.rs
- [x] `README.md`
- [x] `book/src/user/install.md`

crates (pre-release):
- [x] zebra-* `Cargo.toml`s

tower (patch):
- [x] tower-* `Cargo.toml`s

auto-generated:
- [x] `Cargo.lock`: run `cargo build` after updating all the `Cargo.toml`s

#### Version Tooling

You can use `fastmod` to interactively find and replace versions.

For example, you can do something like:
```
fastmod --extensions rs,toml,md --fixed-strings '1.0.0-rc.0' '1.0.0-rc.1' zebrad README.md book/src/user/install.md zebra-network/src/constants.rs
[...]

PR #5523, opened 2022-11-01
suggestion: Only link to the audit version of each crate or file

teor2345:
## Motivation

In PR #5395, I think we should:
* only link to the files or crates we actually want audited
* link to the source code, not the release notes
* hide dependencies we don't want audited by default

Some of these links won't work until PR #5513 merges and the release gets tagged.

I also fixed up some formatting.


teor2345:
@arya2 would you mind reviewing this suggestion some time in the next day or so?

arya2:
This is perfect, thank you.

PR #5529, opened 2022-11-01
change(doc): Use HTTP to clone Zebra in Docker documentation

teor2345:
## Motivation

Git clones over HTTP are usually more reliable than SSH, because some users don't have SSH installed, or it doesn't work from their network.

## Review

Anyone can review this PR, but @gustavovalverde wrote the original docs.

### Reviewer Checklist


arya2:
I think this is good for usability, but it doesn't seem like a bug fix

arya2 (approved):

gustavovalverde (approved):

PR #5530, opened 2022-11-01
fix(batch): Return an error instead of panicking in the batch verifier on shutdown

teor2345:
## Motivation

Zebra can panic on shutdown in the batch verifier, but it should return an error to the transaction verifier instead. (This will cancel any pending transactions on shutdown.)

## Solution

- Handle `JoinError::Cancelled` correctly, the task can be cancelled by the tokio runtime on shutdown

## Review

Anyone can review this PR.

### Reviewer Checklist


arya2 (approved):
Looks good!

PR #5526, opened 2022-11-01
change(rpc): add submitblock RPC method

arya2:
## Motivation

Introduces basic support for the `submitblock` rpc method used by mining pools.

Closes #5236 as part of #5234

### Designs

[Data flow diagram](https://docs.google.com/drawings/d/1fNOY_9hjayzC3D31nKZvpHbONEgDy9CP2A0m9mwS-cs)

### Specifications

https://zcash.github.io/rpc/submitblock.html

## Solution

- Add field for the block verifier to the `GetBlockTemplateRpcImpl` struct
- Call the block verifier with the submitted block

## Review

This PR is part of regular scheduled work.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

- See #5527 
- See #5487
- Update acceptance test (similar to #5015)

Issue #5487, opened 2022-10-26
Return detailed errors from the `submitblock` RPC


teor2345:
## Motivation

If we find a mining pool that relies on detailed `submitblock` RPC errors, we should implement them.

Otherwise this ticket is optional.

This depends on #2908, because it changes the errors in the state.

### Mining Pool Usage

`s-nomp` calls `submitblock` from this code:
- https://github.com/s-nomp/node-stratum-pool/blob/d86ae73f8ff968d9355bb61aac05e0ebef36ccb5/lib/pool.js#L245

## Requirements

https://zcash.github.io/rpc/submitblock.html

https://en.bitcoin.it/wiki/BIP_0023

The goals of the `submitblock` RPC are to:
- get valid blocks onto the local and network chains as quickly as possible, so miners have the best chance of profiting from their work, and
- reject invalid blocks as quickly as possible, so miners can start on an updated block template.

## Possible Design

Return null if the block is accepted onto the best chain in the non-finalized state. Otherwise, return a detailed error containing the specified text.

See the TODO comments in the `submit_block()` RPC function for details.

#### Error handling

We need to check if miners rely on any specific error values.

Return
 - "duplicate": if Zebra has already committed the block to the non-finalized or finalized state
 - "duplicate-invalid": if Zebra already has the block, but it is invalid
    - Zebra will never return this status, because it does not store invalid blocks
 - "duplicate-inconclusive": if Zebra already has the block in the state queue or channel, but has not committed it to the non-finalized state yet. Checking this might not be supported by Zebra's current architecture.
 - "inconclusive": if Zebra has committed the block to the non-finalized state, but it is not on the best chain
 - "rejected": if Zebra rejected the block as invalid
 - Other possible errors listed here, if needed by mining pools: https://github.com/bitcoin/bips/blob/master/bip-0022.mediawiki#appendix-example-rejection-reasons

`zcashd` implementation:

https://github.com/zcash/zcash/blob/6c392dbb0b43d6988283e40ce423242833df2d30/src/rpc/mining.cpp#L869-L905

## Testing

- [ ] Snapshot the output of the RPC
- [ ] Test calling the RPC from the mining pool software
- [ ] Manually compare the `zcashd` and `zebrad` RPC output using https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/zcash-rpc-diff

We might also want to implement #5015 to make writing tests for this issue quicker.

arya2:
[...]

PR #5533, opened 2022-11-02
suggestion: Use the same ChainVerifier for downloaded and submitted blocks

teor2345:
## Motivation

Zebra isn't really designed to use multiple instances of the same verifiers. It might work, but it is easier to design, test, and maintain if we just have one running verifier of each kind.

This is a suggestion for PR #5526.

## Solution

See the explanations I'm going to post on each part of the PR.

## Review

@arya2 wrote the original PR.

I'm happy to revert any of these changes that aren't needed - let me know what you think!

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -12,7 +12,7 @@ use zebra_chain::{
     chain_tip::ChainTip,
     serialization::ZcashDeserializeInto,
 };
-use zebra_consensus::{BlockError, VerifyBlockError};
+use zebra_consensus::{BlockError, VerifyBlockError, VerifyChainError};
 use zebra_node_services::mempool;
 
 use crate::methods::{
@@ -97,7 +97,7 @@ where
         Response = zebra_state::ReadResponse,
         Error = zebra_state::BoxError,
     >,
-    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = VerifyBlockError>
+    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = zebra_consensus::BoxError>
         + Clone
         + Send
         + Sync
@@ -141,7 +141,7 @@ where
         + Sync
         + 'static,
     Tip: ChainTip + Clone + Send + Sync + 'static,
-    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = VerifyBlockError>
+    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = zebra_consensus::BoxError>
         + Clone
         + Send
         + Sync
@@ -182,7 +182,7 @@ where
         + 'static,
     <State as Service<zebra_state::ReadRequest>>::Future: Send,
     Tip: ChainTip + Send + Sync + 'static,
-    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = VerifyBlockError>
+    BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = zebra_consensus::BoxError>
         + Clone
         + Send
         + Sync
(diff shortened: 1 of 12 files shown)


zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -292,27 +292,56 @@ where
...
+                    code: ErrorCode::ServerError(0),
+                    message: error.to_string(),
+                    data: None,
+                })?
+                .call(Arc::new(block))
+                .await;
+
+            let chain_error = match block_verifier_response {
+                // Currently, this match arm returns `null` (Accepted) for blocks committed
+                // to any chain, but Accepted is only for blocks in the best chain.
```
teor2345, line 308: I checked each error to make sure it was being handled correctly, and added TODO comments for any differences. We can fix up any important differences when we do ticket #5487.

(There might not be any important differences.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -292,27 +292,56 @@ where
...
+                //   - "replaced by newer request"
+                // - keep the older request in the queue,
+                //   and return a duplicate error for the newer request immediately.
+                //   This improves the speed of the RPC response.
+                //
+                // Checking the download queues and ChainVerifier buffer for duplicates
+                // might require architectural changes to Zebra, so we should only do it
+                // if mining pools really need it.
+                Ok(_verify_chain_error) => Ok(submit_block::Response::Rejected),
+                Err(_boxed_string_error) => Ok(submit_block::Response::Rejected),
```
```suggestion

                // This match arm is currently unreachable, but if future changes add extra error types,
                // we want to turn them into `Rejected`.
                Err(_unknown_error_type) => Ok(submit_block::Response::Rejected),
```

zebra-rpc/src/methods/get_block_template_rpcs/types/submit_block.rs
```diff
@@ -1,12 +1,20 @@
...
+use crate::methods::get_block_template_rpcs::GetBlockTemplateRpc;
+
 /// Optional argument `jsonparametersobject` for `submitblock` RPC request
 ///
-/// See notes for [`Rpc::submit_block`] method
+/// See notes for [`GetBlockTemplateRpc::submit_block`] method
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct JsonParameters {
-    pub(crate) work_id: String,
+    pub(crate) work_id: Option<String>,
```
teor2345, line 12: The `work_id` field is also optional:
https://zcash.github.io/rpc/submitblock.html

zebra-rpc/src/server.rs
```diff
@@ -87,6 +72,12 @@ impl RpcServer {
             + 'static,
         State::Future: Send,
         Tip: ChainTip + Clone + Send + Sync + 'static,
+        BlockVerifier: Service<Arc<Block>, Response = block::Hash, Error = zebra_consensus::BoxError>
+            + Clone
+            + Send
+            + Sync
+            + 'static,
+        <BlockVerifier as Service<Arc<Block>>>::Future: Send,
```
teor2345, line 80: Using a generic here makes some tests easier to write, because we can use a `MockService`.

zebra-rpc/src/server.rs
```diff
@@ -47,35 +45,22 @@ pub struct RpcServer;
...
-                    >,
-                    zebra_state::Request,
-                >,
-                Buffer<
-                    BoxService<transaction::Request, transaction::Response, TransactionError>,
-                    transaction::Request,
-                >,
-            >,
-        >,
+        #[allow(unused_variables)] block_verifier: BlockVerifier,
```
teor2345, line 53: It seemed simpler to always accept this argument, and then handle it being optional within `zebra-rpc`.

zebrad/src/commands/start.rs
```diff
@@ -186,8 +181,7 @@ impl StartCmd {
             app_version(),
             mempool.clone(),
             read_only_state_service,
-            #[cfg(feature = "getblocktemplate-rpcs")]
-            Some(block_verifier),
+            chain_verifier.clone(),
```
teor2345, line 184: Zebra isn't really designed to have multiple instances of the same kind of verifier. It would probably work, but it seems trickier to test and maintain.

Instead, I just cloned the common `ChainVerifier`, and used it here. (And created a `ChainVerifier` or `MockService` in the tests.)

arya2 (approved):
These are good changes and decisions, thank you!

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: The best chain note hadn't occurred to me and will be good to keep in mind.

> There might not be any important differences

Spec conformance seems particularly important in lieu of substantial user feedback.

teor2345: > Spec conformance seems particularly important in lieu of substantial user feedback.

Just to be clear, RPC error values are not part of the Zcash specification.

Only consensus-critical rules get put in the specification. Everything else can be changed by nodes, as long it is compatible with other software in the ecosystem.

There is some RPC user documentation that is auto-generated from `zcashd`. But we have found some parts of it that are outdated already. So I'm not sure if any mining pools rely on these error values.

PR #5540, opened 2022-11-03
4. change(rpc): Refactor RPCs using a state tip height function

teor2345:
## Motivation

We call the same code 4 times to get the state tip height, so I made a function for it.

## Review

This is a tiny refactor, anyone can review it.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -360,7 +360,7 @@ where
         + Sync
         + 'static,
     State::Future: Send,
-    Tip: ChainTip + Send + Sync + 'static,
+    Tip: ChainTip + Clone + Send + Sync + 'static,
 {
     fn get_info(&self) -> Result<GetInfo> {
         let response = GetInfo {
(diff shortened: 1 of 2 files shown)


zebra-rpc/src/methods.rs
```diff
@@ -994,6 +992,19 @@ where
     }
 }
 
+/// Returns the best chain tip height of `latest_chain_tip`,
+/// or an RPC error if there are no blocks in the state.
+pub fn best_chain_tip_height<Tip>(latest_chain_tip: &Tip) -> Result<Height>
+where
+    Tip: ChainTip + Clone + Send + Sync + 'static,
+{
+    latest_chain_tip.best_tip_height().ok_or(Error {
```
arya2, line 1001: On this note it might be nice to have a `make_server_error` fn in the future.

arya2 (dismissed):
Looks good to me, thank you for the cleanup.

zebra-rpc/src/methods.rs
teor2345: Something like this, but maybe with an `error_code: Into<Option<i64>>` argument.

```rust
impl<T, E> MapServerError for Result<T, E> where E: Debug {
    fn map_server_error(result: Self) -> Result<T, jsonrpc_core::Error> {
        result.map_err(|error| jsonrpc_core::Error {
            code: ErrorCode::ServerError(0),
            message: format!("{error:?}"),
            data: None,
        })
    }
}

impl<T> OkOrServerError for Option<T> {
    fn ok_or_server_error<S: ToString>(option: Self, message: S) -> Result<T, jsonrpc_core::Error> {
        option.ok_or(jsonrpc_core::Error {
            code: ErrorCode::ServerError(0),
            message: message.to_string(),
            data: None,
        })
    }
}
```

dconnolly:
hm this may need a rebase
image

teor2345:
> hm this may need a rebase

GitHub does this when the workflows are different for PRs targeting `main`. Seems like a CI bug we should fix.

This PR will get rebased anyway, after we fix `main` branch failures in PR #5549.

teor2345:
> > hm this may need a rebase
> 
> GitHub does this when the workflows are different for PRs targeting `main`. Seems like a CI bug we should fix.

I opened PR #5550 which should fix this.

teor2345:
@mergifyio update

teor2345:
This PR conflicts with PR #5526, so I pushed a merge commit, and set it to merge after that PR merges.

arya2 (approved):

PR #5554, opened 2022-11-04
5. change(rpc): Implement coinbase conversion to RPC `TransactionTemplate` type

teor2345:
## Motivation

We want to calculate miner fees and transaction roots, then generate transaction templates.

This does everything except actually generating a coinbase transaction.

Part of #5453.

### Specifications

`coinbase_txn`, `transactions`, and some root fields in:
https://zcash.github.io/rpc/getblocktemplate.html

## Solution

- Add conversion from coinbase `Transaction`s into `TransactionTemplate`s
  - Calculate the sigops using `zebra-script`, because we don't want to call the transaction verifier
- Add a `Network` field to the getblocktemplate RPC handler
  - The network and block height are needed to generate the coinbase transaction

Refactors:
- Split out a select_mempool_transactions() function
- Cleanup redundant dependencies

## Review

Anyone can review this PR.

### Reviewer Checklist

## Follow Up Work

Actually generate the coinbase transaction

### DIFF
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -46,6 +46,7 @@ zebra-chain = { path = "../zebra-chain" }
 zebra-consensus = { path = "../zebra-consensus" }
 zebra-network = { path = "../zebra-network" }
 zebra-node-services = { path = "../zebra-node-services" }
+zebra-script = { path = "../zebra-script" }
 zebra-state = { path = "../zebra-state" }
 
 [dev-dependencies]
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -52,12 +52,10 @@ zebra-state = { path = "../zebra-state" }
 insta = { version = "1.21.0", features = ["redactions", "json"] }
 proptest = "0.10.1"
 proptest-derive = "0.3.0"
-serde_json = "1.0.87"
 thiserror = "1.0.37"
 
 tokio = { version = "1.21.2", features = ["full", "tracing", "test-util"] }
 
 zebra-chain = { path = "../zebra-chain", features = ["proptest-impl"] }
-zebra-consensus = { path = "../zebra-consensus" }
 zebra-state = { path = "../zebra-state", features = ["proptest-impl"] }
-zebra-test = { path = "../zebra-test/" }
+zebra-test = { path = "../zebra-test" }
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -253,10 +253,14 @@ where
             let _tip_height = best_chain_tip_height(&latest_chain_tip)?;
             let mempool_transactions = select_mempool_transactions(mempool).await?;
 
+            //let coinbase_transaction = Transaction::new_v5_coinbase(network, tip_height, miner_fee);
+
             let merkle_root;
             let auth_data_root;
 
-            // TODO: add the coinbase transaction to these lists, and delete the is_empty() check
+            // TODO:
+            // - add the coinbase transaction to these lists, and delete the is_empty() check
+            // - split this out into its own function
             if !mempool_transactions.is_empty() {
                 merkle_root = mempool_transactions.iter().cloned().collect();
                 auth_data_root = mempool_transactions.iter().cloned().collect();
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -266,9 +266,52 @@ where
             let miner_fee = miner_fee(&mempool_txs);
 
             /*
-            TODO: create a method Transaction::new_v5_coinbase(network, tip_height, miner_fee);
-            */
-            let coinbase_tx = mempool_txs[0].transaction.clone();
+            Fake a "coinbase" transaction by duplicating a mempool transaction,
+            or fake a response.
+
+            This is temporarily required for the tests to pass.
+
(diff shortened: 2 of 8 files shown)


zebra-rpc/Cargo.toml
```diff
@@ -46,18 +46,17 @@ zebra-chain = { path = "../zebra-chain" }
...
 insta = { version = "1.21.0", features = ["redactions", "json"] }
 proptest = "0.10.1"
 proptest-derive = "0.3.0"
-serde_json = "1.0.87"
 thiserror = "1.0.37"
 
 tokio = { version = "1.21.2", features = ["full", "tracing", "test-util"] }
 
 zebra-chain = { path = "../zebra-chain", features = ["proptest-impl"] }
-zebra-consensus = { path = "../zebra-consensus" }
```
teor2345, line 61: Already in `[dependencies]`

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -250,35 +261,20 @@ where
...
-                unreachable!("unmatched response to a mempool::FullTransactions request");
-            };
+            let miner_fee = miner_fee(&mempool_txs);
 
-            let merkle_root;
-            let auth_data_root;
+            /*
+            TODO: create a method Transaction::new_v5_coinbase(network, tip_height, miner_fee);
+            */
+            let coinbase_tx = mempool_txs[0].transaction.clone();
```
teor2345: Should we run the generated `coinbase_tx` through the `transaction::Verifier`?

We'll need a new request type which returns a coinbase result, because the `remaining_value_balance()` will be negative. (Maybe this can go in another ticket?)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -416,6 +393,70 @@ where
     }
 }
 
+// get_block_template support methods
+
+/// Returns selected transactions in the `mempool`, or an error if the mempool has failed.
+///
+/// TODO: select transactions according to ZIP-317 (#5473)
+pub async fn select_mempool_transactions<Mempool>(
```
teor2345, line 444: Same code, but refactored out of the `get_block_template()` method

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -416,6 +393,70 @@ where
...
+}
+
+/// Returns the total miner fee for `mempool_txs`.
+pub fn miner_fee(mempool_txs: &[VerifiedUnminedTx]) -> Amount<NonNegative> {
+    let miner_fee: amount::Result<Amount<NonNegative>> =
+        mempool_txs.iter().map(|tx| tx.miner_fee).sum();
+
+    miner_fee.expect(
+        "invalid selected transactions: \
+         must not spend more than MAX_MONEY",
```
```suggestion
         fees in a valid block can not be more than MAX_MONEY",
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -416,6 +393,70 @@ where
...
+    )
+}
+
+/// Returns the transaction effecting and authorizing roots
```
teor2345, line 483: Modified from the previous code in the `get_block_template()` method

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: I think it's optional since the generated coinbase is a function of verified transactions and it's being run through the `transaction::Verifier` by the `BlockVerifier` when the block is submitted, but it might be nice for testing and debugging.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -250,35 +261,20 @@ where
...
-                transactions
-            } else {
-                unreachable!("unmatched response to a mempool::FullTransactions request");
-            };
+            let miner_fee = miner_fee(&mempool_txs);
 
-            let merkle_root;
-            let auth_data_root;
+            /*
+            TODO: create a method Transaction::new_v5_coinbase(network, tip_height, miner_fee);
```
arya2: This might need to be part of the same call as the `FullTransactions` request to guarantee nothing else is added or evicted between these calls.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -416,6 +393,70 @@ where
...
+    miner_fee.expect(
+        "invalid selected transactions: \
+         fees in a valid block can not be more than MAX_MONEY",
+    )
+}
+
+/// Returns the transaction effecting and authorizing roots
+/// for `coinbase_tx` and `mempool_txs`.
+//
+// TODO: should this be spawned into a cryptographic operations pool?
```
arya2, line 486: We could create an issue for benchmarking this to decide if it's hard to say, but I think this is okay for now.

arya2 (commented):
This looks good! There's a small merge conflict with the `select_mempool_transactions` refactor.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: If that happened, what problems could it cause?

Either way, this seems like something we can deal with in the next PR.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: I'll see what I can do in the next PR.

teor2345:
@mergifyio update

teor2345:
@mergifyio update

teor2345:
@mergifyio update

arya2 (approved):

PR #5558, opened 2022-11-04
feature(rpc): add fixed values to getblocktemplate response

### LINKED ISSUE #5452: Populate `Blocktemplate` responses with fixed values data

oxarbitrage:
## Motivation

Close https://github.com/ZcashFoundation/zebra/issues/5452

## Solution

Add constants where needed, get constant from other crates, update tests.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/header.rs
+++ b/zebra-chain/src/block/header.rs
@@ -146,6 +146,12 @@ const BLOCK_HEADER_LENGTH: usize =
 /// A CountedHeader has BLOCK_HEADER_LENGTH bytes + 1 or more bytes for the transaction count
 pub(crate) const MIN_COUNTED_HEADER_LEN: usize = BLOCK_HEADER_LENGTH + 1;
 
+/// The Zcash accepted block version.
+///
+/// The consensus rules do not force the block version to be this value but just equal or greater than it.
+/// However, it is suggested that submitted block versions to be of this exact value.
+pub const ZCASH_BLOCK_VERSION: u32 = 4;
+
 impl TrustedPreallocate for CountedHeader {
     fn max_allocation() -> u64 {
         // Every vector type requires a length field of at least one byte for de/serialization.
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -1,6 +1,6 @@
 //! RPC methods related to mining only available with `getblocktemplate-rpcs` rust feature.
 
-use std::sync::Arc;
+use std::{iter, sync::Arc};
 
 use futures::{FutureExt, TryFutureExt};
 use jsonrpc_core::{self, BoxFuture, Error, ErrorCode, Result};
@@ -8,11 +8,17 @@ use jsonrpc_derive::rpc;
 use tower::{buffer::Buffer, Service, ServiceExt};
 
 use zebra_chain::{
-    amount::Amount,
-    block::{self, Block},
-    block::{Height, MAX_BLOCK_BYTES, ZCASH_BLOCK_VERSION},
+    amount::{self, Amount, NonNegative},
+    block::Height,
+    block::{
+        self,
+        merkle::{self, AuthDataRoot},
+        Block, MAX_BLOCK_BYTES, ZCASH_BLOCK_VERSION,
+    },
     chain_tip::ChainTip,
+    parameters::Network,
     serialization::ZcashDeserializeInto,
+    transaction::{UnminedTx, VerifiedUnminedTx},
 };
 use zebra_consensus::{
     BlockError, VerifyBlockError, VerifyChainError, VerifyCheckpointError, MAX_BLOCK_SIGOPS,
@@ -129,6 +135,9 @@ where
     // Configuration
     //
     // TODO: add mining config for getblocktemplate RPC miner address
+    //
+    /// The configured network for this RPC service.
+    _network: Network,
 
     // Services
     //
@@ -169,12 +178,14 @@ where
 {
     /// Create a new instance of the handler for getblocktemplate RPCs.
     pub fn new(
+        network: Network,
         mempool: Buffer<Mempool, mempool::Request>,
         state: State,
         latest_chain_tip: Tip,
         chain_verifier: ChainVerifier,
     ) -> Self {
         Self {
+            _network: network,
             mempool,
             state,
             latest_chain_tip,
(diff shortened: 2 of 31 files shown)


zebra-chain/src/block/header.rs
```diff
@@ -146,6 +146,12 @@ const BLOCK_HEADER_LENGTH: usize =
 /// A CountedHeader has BLOCK_HEADER_LENGTH bytes + 1 or more bytes for the transaction count
 pub(crate) const MIN_COUNTED_HEADER_LEN: usize = BLOCK_HEADER_LENGTH + 1;
 
+/// The Zcash accepted block version.
+///
+/// The consensus rules do not force the block version to be this value but just equal or greater than it.
+/// However, it is suggested that submitted block versions to be of this exact value.
```
teor2345, line 152: Optional: Do you think it makes more sense to move the consensus rule docs here?

zebra-rpc/src/methods/tests/snapshot/snapshots/get_block_template@mainnet_10.snap
```diff
@@ -1,10 +1,11 @@
 ---
 source: zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
+assertion_line: 128
```
teor2345, line 3: Optional: if you upgrade to `cargo insta` 1.13 or later, this line won't be added to snapshots any more. (Most of us have already upgraded.)

https://insta.rs/changelog/#1-13-0

zebra-rpc/src/methods/get_block_template_rpcs/constants.rs
```diff
@@ -0,0 +1,15 @@
...
+pub const GET_BLOCK_TEMPLATE_NONCE_RANGE_FIELD: &str = "00000000ffffffff";
+
+lazy_static! {
+    /// A hardcoded list of fields that the miner can change from the block.
+    pub static ref GET_BLOCK_TEMPLATE_MUTABLE_FIELD: Vec<String> = vec![
+        "time".to_string(),
+        "transactions".to_string(),
+        "prevblock".to_string()
+    ];
+}
```
teor2345: I'd like to avoid new uses of `lazy_static`, because its maintainer is slow, and it has known bugs:
- https://github.com/rust-lang-nursery/lazy-static.rs/issues/201
- https://github.com/rust-lang-nursery/lazy-static.rs/issues/204

Instead, we can use a constant value, see my PR #5559.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -324,12 +328,17 @@ where
...
+                nonce_range: constants::GET_BLOCK_TEMPLATE_NONCE_RANGE_FIELD.to_string(),
 
-                nonce_range: empty_string.clone(),
+                // Max legacy signature operations in the block.
+                sigop_limit: MAX_BLOCK_SIGOPS,
 
-                sigop_limit: 0,
-                size_limit: 0,
+                // Max block size in bytes
+                size_limit: MAX_BLOCK_BYTES,
```
teor2345: Can you add documentation to the `GetBlockTemplate` struct for these new fields?

teor2345 (commented):
Thanks, looks good!

I have a suggestion in PR #5559, and it will help me to review if we add documentation for the new fields.

teor2345:
@Mergifyio update

teor2345:
@mergifyio update

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/5558/commits/5590d0cab8339748ed97f88fcf40cf144916ec18

zebra-rpc/src/methods/get_block_template_rpcs/constants.rs
oxarbitrage: Thanks.

zebra-rpc/src/methods/tests/snapshot/snapshots/get_block_template@mainnet_10.snap
oxarbitrage: I am not sure how to do this. I do have cargo insta 1.21 on my system and that are the snapshot it is doing.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: My suggestion PR documented the roots fields that I missed in a previous PR.

But it's also important to document the mutable, nonce_range, sigop_limit, and size_limit fields on the struct. (The docs here are good, you can just copy them to the struct definition.)

zebra-rpc/src/methods/tests/snapshot/snapshots/get_block_template@mainnet_10.snap
teor2345: That's strange, maybe you have multiple versions installed?
You could try running `cargo install cargo-insta` to update it.

But it's not a big deal, it should keep on working until Zebra starts using a feature your version doesn't have.

teor2345:
This PR has conflicts with PR #5554, which just merged. So it will need to be rebased or it will need a merge commit.

oxarbitrage:
The tests will fail because they all get into this code section https://github.com/ZcashFoundation/zebra/blob/304048efe2a07eec86110d7c3cdeb08bb01f16cc/zebra-rpc/src/methods/get_block_template_rpcs.rs#L280

where everything is empty.

I am unsure if we should update the test and snapshots to be empty or if we should add the fixed values into that code as well.

teor2345:
> The tests will fail because they all get into this code section
> 
> https://github.com/ZcashFoundation/zebra/blob/304048efe2a07eec86110d7c3cdeb08bb01f16cc/zebra-rpc/src/methods/get_block_template_rpcs.rs#L280
> 
> where everything is empty.
> 
> I am unsure if we should update the test and snapshots to be empty or if we should add the fixed values into that code as well.

That code disappears in my next PR #5580, so you could base this PR on that one, or copy the fixed values into the code that is about to be deleted?

teor2345 (approved):
Thanks, looks good!

PR #5559, opened 2022-11-06
suggestion: Avoid new uses of lazy_static

teor2345:
## Motivation

This is a suggestion on PR #5558.

It removes a new use of `lazy_static`, and adds some missing documentation.


teor2345:
@Mergifyio update

teor2345:
@mergifyio update

oxarbitrage (approved):
Thanks, i tried to do the const array as you do here but i was not able to remember how to do it. I didn't thought it was a big deal (we have lazy statics already).

Anyway, thanks for the fix. I also missed those documentations so thank you for adding them too.

Issue #5494, opened 2022-10-28
gcloud ssh: Connection reset by peer


teor2345:
## Motivation

This error happens occasionally:
>  ERROR: (gcloud.compute.ssh) There was a problem refreshing your current auth tokens: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

https://github.com/ZcashFoundation/zebra/actions/runs/3341437225/jobs/5533308842#step:6:84

>  kex_exchange_identification: read: Connection reset by peer

https://github.com/ZcashFoundation/zebra/actions/runs/3352929291/jobs/5555664923#step:6:108

gustavovalverde:
I reopened https://github.com/ZcashFoundation/zebra/pull/5069, I'll let you decide which ticket to use.

> I'm not sure if reopening tickets is a thing in our workflow

teor2345:
> I reopened #5069, I'll let you decide which ticket to use.

I think the errors are different?
(They might have the same cause, I guess we'll find out when we fix one of them.)

> > I'm not sure if reopening tickets is a thing in our workflow

It works for me, but that's really a question for Pili.

teor2345:
> kex_exchange_identification: read: Connection reset by peer

On a launch step in PR #5521:
https://github.com/ZcashFoundation/zebra/actions/runs/3373347892/jobs/5597848235#step:7:108

teor2345:
From PR #5508:
> Writing 3 keys to /home/runner/.ssh/google_compute_known_hosts
> Updating project ssh metadata...
> ..................................................................Updated [https://www.googleapis.com/compute/v1/projects/zealous-zebra].
> .done.
> Waiting for SSH key to propagate.
> kex_exchange_identification: read: Connection reset by peer

https://github.com/ZcashFoundation/zebra/actions/runs/3381967679/jobs/5617006676

But this was an interim logging step that didn't run too long, so the test still passed.

teor2345:
> kex_exchange_identification: read: Connection reset by peer

On PR #5550 halfway through a test run:
https://github.com/ZcashFoundation/zebra/actions/runs/3391346405/jobs/5665340226#step:6:110

teor2345:
Same error on the main branch:
https://github.com/ZcashFoundation/zebra/actions/runs/3412654251/jobs/5679413392#step:6:106

mpguerra:
[...]

Issue #5366, opened 2022-10-10
Fix "the resource was not found" instance errors in CI


teor2345:
## Motivation

I'm seeing errors like this in CI:
>  - The resource 'projects/zealous-zebra/zones/us-central1-a/instances/update-to-tip-5360-merge-818a593' was not found

https://github.com/ZcashFoundation/zebra/actions/runs/3203311358/jobs/5239744583#step:6:101

### Command Reference

https://cloud.google.com/sdk/gcloud/reference/topic/filters

### Designs

I'm not sure what is causing these bugs, we only delete instances after 3 days:
https://github.com/ZcashFoundation/zebra/blob/ee0edef8571d3ed82848303b142c7ae72e7135fc/.github/workflows/delete-gcp-resources.yml#L56

And after the job fails:
https://github.com/ZcashFoundation/zebra/blob/ee0edef8571d3ed82848303b142c7ae72e7135fc/.github/workflows/deploy-gcp-tests.yml#L1329

Maybe we got a failed instance or outdated instance delete script wrong somewhere?
We could try using `name={{key}}` for failed job deletes, but it might not be causing this problem.

Let's check the Google Cloud logs to work out what is deleting these instances.

teor2345:
Failed in PR #5541, halfway through the test run:
https://github.com/ZcashFoundation/zebra/actions/runs/3390759314/jobs/5636067584#step:6:106

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5366?planning-poker) with Zenhub @arya2 @dconnolly @gustavovalverde @oxarbitrage @teor2345

teor2345:
@gustavovalverde did this get fixed by the recent changes?

teor2345:
We haven't seen this bug for a while.

teor2345:
I haven't seen these in ages?

PR #5560, opened 2022-11-07
fix(ci): Stop using multiple jobs for quick Google Cloud tests

teor2345:
## Motivation

This PR only uses multiple CI jobs for the full sync tests.
All the other test run in a single "Run" job.

If the SSH errors in these tickets are related to the number of `ssh` connections we make, this should reduce our CI error rates due to:
- #5494 
- #5069
- and possibly #5366

But it doesn't actually fix these bugs. (Unless the cause is an undocumented Google Cloud inbound SSH connection rate-limit.)

## Solution

- Add a new parameter for long tests, and only enable it for the full sync test
- Remove a job that's a duplicate of the "Run" job
- Remove unused steps and whitespace

## Review

@gustavovalverde might be the best person to review this PR.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -332,7 +332,7 @@ jobs:
     # because we might never get a finished sync.
     #
     # See the concurrency comment on the zebrad test-full-sync job for details.
-    concurrency: 
+    concurrency:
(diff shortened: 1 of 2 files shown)


teor2345:
@Mergifyio update

teor2345:
@mergifyio update

dconnolly (approved):

Issue #5579, opened 2022-11-08
Populate `BlockTemplate.blockcommitmentshash` from the transactions and state request


teor2345:
## Motivation

The `blockcommitmentshash` field is calculated from the transactions in #5453 and #5454, and the state request in #5455.

This field's value is repeated in:
    * `lightclientroothash`
    * `finalsaplingroothash`
    * `blockcommitmentshash`
    * `defaultroots.blockcommitmentshash`

### Specifications

https://zips.z.cash/zip-0244#block-header-changes

### Designs

We have code to calculate the block commitments hash in `zebra-chain` already, because we validate it in `zebra-state`.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5579?planning-poker) with Zenhub @arya2 @dconnolly @gustavovalverde @oxarbitrage @teor2345

Issue #5577, opened 2022-11-08
Document P2SH multisig address handling for funding stream receivers


teor2345:
## Motivation

Zebra's consensus rule documentation does not specifically mention P2SH multisig addresses, even though we handle them correctly.

### Specifications

> The standard redeem script hash is specified in [Bitcoin-Multisig] for P2SH multisig addresses...

https://zips.z.cash/protocol/protocol.pdf#fundingstreams
https://developer.bitcoin.org/devguide/transactions.html#multisig

### Designs

- Document multisig handling in `check_script_form()`
- We might want to assert we're getting a P2SH `Address` in `check_script_form()` and the mining config, because `Address` also has other variants

## Related Work

The `Address` variant bug was introduced as part of #5453

teor2345:
I documented why we can skip this consensus rule in PR #5580.

str4d:
> However, the current funding streams do not use multisig addresses

This is false. All of the current and past funding stream addresses are multisig P2SH addresses (the ones I spot-checked are 2-of-3 multisig).

teor2345:
> > However, the current funding streams do not use multisig addresses
> 
> This is false. All of the current and past funding stream addresses are multisig P2SH addresses (the ones I spot-checked are 2-of-3 multisig).

Thanks, I think I misunderstood what was happning here. Zebra's code is correct, but the consensus rule docs are incomplete.

PR #5580, opened 2022-11-08
change(rpc): generate coinbase transactions in the getblocktemplate RPC

### LINKED ISSUE #5453: Populate transparent coinbase transaction data for `BlockTemplate`

teor2345:
## Motivation

This PR generates coinbase transactions when the `getblocktemplate` RPC is called.

## Solution

Generate the coinbase transaction in the getblocktemplate RPC:
- Add coinbase creation to zebra-chain
- Add coinbase creation and miner subsidy to zebra-consensus

Related changes:
- Add the miner config to the GetBlockTemplateRpcImpl
- Add a getblocktemplate-rpcs feature to zebra-chain

Related fixes:
- Fix missing feature dependencies in zebra-consensus
- Provide fake valid block heights to getblocktemplate RPC tests
- Rewrite `check_script_form()` so it is easier to read

Testing:
- Add a getblocktemplate.coinbase_tx deserialized transaction snapshot
- Update the getblocktemplate RPC snapshots

Closes #5453.

## Review

This is a routine change, it might help getting two people to review it.

### Reviewer Checklist

## Follow Up Work

I split #5579 out of #5453, because the state request wasn't ready yet.


oxarbitrage (approved):
Looks really good.

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -124,20 +124,43 @@ pub fn funding_stream_address(
 ///
 /// [7.10]: https://zips.z.cash/protocol/protocol.pdf#fundingstreams
 pub fn check_script_form(lock_script: &Script, address: Address) -> bool {
-    let mut address_hash = address
+    // TODO: Verify P2SH multisig funding stream addresses (#5577).
+    //       As of NU5, the funding streams do not use multisig addresses,
```
arya2, line 128: This doesn't need to be addressed in this PR, but the funding streams might use multisig addresses: https://github.com/ZcashFoundation/zebra/issues/5577#issuecomment-1309292394

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -124,20 +124,43 @@ pub fn funding_stream_address(
...
+    //       so this is optional.
```
str4d, line 129: This is not the case. All of the (currently-revealed) funding stream addresses are P2SH ([as indicated by their `t3` prefixes](https://github.com/zcash/zcash/blob/fadbb794f16bee26513eb9a1fda51ac0cfa096e1/src/chainparams.cpp#L171-L225)), and of the addresses with publicly-revealed `scriptSig` values, all of the ones I spot-checked are multisig addresses. In particular, the ZF's funding stream address (the `scriptSig` for which is visible in [this transaction](https://blockchair.com/zcash/transaction/866de6abfee9ad02c267dccfc45710d8847b66a72e3b458f120c94ca0b4bf83c?i=999)) is a 2-of-3 multisig address (with an additional storage of 4 bytes of data that I can't identify):

```
// OP_0
00
// PUSHDATA(<SIGNATURE> || SIGHASH_ALL)
48
  3045022100a16d32df92e28624ae194fb4de9e8610aa1932e4e2b370fa7c0ca7dab62d34d102205cc4f9dac56d43e7fd9e3d70816e3aa7f72832774712193bfbc227fb5ffaa0bf
  01
// PUSHDATA(<SIGNATURE> || SIGHASH_ALL)
47
  304402202341788cde33e568cfc388acf36ae9b24c22883c7ad88822e54c8bbe07a4c93202207654bcaa7c8f14403c424c7d681b4c525f779aa314795066a825e921fd4525d4
  01
// PUSHDATA(multisig script)
4c 72
  52 // OP_2
  21 0352d50656796dd79e3e3385c29636d849b8705c9c6d9b86a1717adc38e4a567c7 // PUSHDATA(pubkey)
  21 02f39d0b69c3b8e06030ef9dc295cf8c848f51e398fe85414e33509afdc7f01fb3 // PUSHDATA(pubkey)
  21 027072ba2319ead2b0b387d69d36508f57b397da8d4ac76ea64676ee8a1356dae0 // PUSHDATA(pubkey)
  53 // OP_3
  af // OP_CHECKMULTISIGVERIFY
  04 d5f2a8b6 // PUSHDATA(0xb6a8f2d5)
  75 // OP_DROP
  74 // OP_DEPTH
  00 // OP_0
  87 // OP_EQUAL
```

zebra-consensus/src/block/subsidy/funding_streams.rs
teor2345: Thanks, I'll deal with this in ticket #5577.

zebra-consensus/src/block/subsidy/funding_streams.rs
teor2345: Thanks, I'll deal with this in ticket #5577.

PR #5561, opened 2022-11-07
fix(sync): Pause new downloads when Zebra reaches the lookahead limit

teor2345:
## Motivation

We're seeing a lot of AboveLookaheadLimit errors in full syncs. Instead of dropping those blocks, we should wait for them to verify.

Also we might need more time for a full sync.

### Designs

Add a watch channel to the downloader, which gets set when the downloader is a long way away from the state tip. Reset the channel when:
- a downloaded block is much closer to the tip, or
- there are no more blocks pending download or verify.

## Solution

Syncer fixes:
- Make the syncer pause new downloads when a downloaded block is 3000 blocks ahead of the state tip
    - If the block is 50,000 blocks ahead, still drop it anyway
- Make sure the default lookahead limit includes 2 full checkpoints (1000 blocks)
    - Add a new test config for this
- Allow 2000 extra blocks in the pre-download queue, or the state commit channel (was 4000 extra blocks)

Related fixes:
- Allow 36 hours for a full sync (was 32 hours)
- Only log lookahead issues once at info level for every pause, rather than for every block
    - Closes #5518

Related cleanups:
- Fix up a syncer test that didn't actually test what it said it did
- Add a `getblocktemplate-rpcs` alternative to the missing test config file message

## Review

Anyone can review this PR, it's a high priority because it fixes errors on `main`.

I'm doing a full sync locally to check that it works.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

PR #5586, opened 2022-11-08
fix(sync): Increase block download timeout and retries

teor2345:
## TODO

We might need to update some tests for the new values, or revert the block download timeout change.

These changes might not fix anything, they need a full sync test before they merge. It's possible we have some sync deadlocks, or an async task is not yielding often enough.

## Scheduling

We might need these changes if the full sync test fails. Otherwise, we can just close this PR.

## Motivation

After the fixes in PR #5561, there might still be some timeout warnings in Zebra's logs.

This a bug for a few reasons:
1. Users get alarmed at warnings
2. The timeouts could make Zebra's CI fail
3. It costs money to have full sync machines repeatedly doing nothing for an 8 minute timeout

## Solution

Try increasing the block download retries and timeout. This will help if peers are slow to send blocks.

I'm testing this locally.

(If that doesn't work, we might want to increase the full verification concurrency limit. That will help if we're getting blocks in the wrong order, then deadlock on both the concurrency limit, and getting a block we don't have.)

## Review

Anyone can review this PR.

### Reviewer Checklist


teor2345:
@mergifyio update

gustavovalverde:
@Mergifyio update

teor2345:
I'm running a full sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/3519147385

If it has fewer timeouts, I'll fix up the test failures and submit this PR for review.

teor2345:
This didn't help with the timeouts, so it's likely that we have some sync deadlocks, or an async task is not yielding often enough.

They are a low priority to fix, we can just extend the sync timeout for now.

Issue #5384, opened 2022-10-11
`fake_peer_set::mempool_requests_for_transactions` failed with Nil response


teor2345:
## Motivation

In PR #5383, `fake_peer_set::mempool_requests_for_transactions` failed with:
> internal error: entered unreachable code: `MempoolTransactionIds` requests should always respond `Ok(Vec<UnminedTxId>)`, got Ok(Nil)

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/5383#issuecomment-1275204946_
      

## Solution

Just allow a `Nil` response, the other tests cover this case as well.

teor2345:
Failed again in PR #5561
>  Message:  response to `MempoolTransactionIds` request should match added_transaction_ids [Legacy(transaction::Hash("c4eaa58879081de3c24a7b117ed2b28300e7ec4c4c1dff1d3f1268b7857a4ddb"))]

https://github.com/ZcashFoundation/zebra/actions/runs/3413929832/jobs/5681223874#step:14:8817

teor2345:
Failed again locally:
> Message:  response to `MempoolTransactionIds` request should match added_transaction_ids [Legacy(transaction::Hash("c4eaa58879081de3c24a7b117ed2b28300e7ec4c4c1dff1d3f1268b7857a4ddb"))]
> Location: zebrad/src/components/inbound/tests/fake_peer_set.rs:81
> failures:
>   components::inbound::tests::fake_peer_set::mempool_requests_for_transactions

teor2345:
Failed in CI:
https://github.com/ZcashFoundation/zebra/actions/runs/3423847000/jobs/5703529589#step:3:9323

@mpguerra this test seems to be failing a lot, maybe we should just fix it quickly by accepting a `Nil` response as well?

mpguerra:
How do we know that it failed with `Nil` response, as opposed to a different unexpected response? It's not clear to me from the logs...

teor2345:
> How do we know that it failed with `Nil` response, as opposed to a different unexpected response? It's not clear to me from the logs...

That specific log message is only used when `Nil` is returned:
https://github.com/ZcashFoundation/zebra/blob/71f5e63e6455a79c5a02f2dc5fc5261001fc7846/zebrad/src/components/inbound/tests/fake_peer_set.rs#L81-L85

The log should probably also say `Nil`, or log the actual response.

mpguerra:
[...]

PR #5589, opened 2022-11-08
fix(tests): add submitblock test to CI, and avoid copying the cached state directory in other tests

### LINKED ISSUE #5015: Get transactions from the non-finalized state in the send transactions test

### LINKED ISSUE #5527: Add submitblock test to CI
## Motivation

The acceptance test for submit block in #5526 uses the cached zebra state and is ignored in the regular coverage CI.

Add a CI job like the one for the send transaction test for the submit block RPC method.

arya2:
## Motivation

The send transactions and submitblock tests were needlessly copying the entire zebra cached state directory. 

The submitblock test also included duplicate code, was missing some details, and hadn't been added to our CI.

There are a few commits that are purely code movement and renaming.

- Closes #5015 
- Closes #5527 

## Solution

- Renames `LightwalletdTestType` to `TestType`, adds a `UpdateZebraCachedState` variant, and moves it to a new file
- Moves `random_known_rpc_port_config` fn to `config.rs` file
- Replaces duplicate logic with methods from `TestType` and adds a `needs_zebra_rpc_server` method for determining the config
- Adds `get_raw_future_blocks` and `get_future_blocks` fns for fully syncing Zebra and getting up to the first `max_num_blocks` in best chain of the non-finalized state
- Uses the new fns for getting blocks up to the first `max_num_blocks` past the cached state in the send transaction and submitblock tests
- Adds submitblock test to CI
 
## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Remove main branch constraint on send transactions test in CI


arya2:
Running a send transactions test here: https://github.com/ZcashFoundation/zebra/actions/runs/3424052393

.github/workflows/continous-integration-docker.patch.yml
```diff
@@ -90,6 +90,12 @@ jobs:
     steps:
       - run: 'echo "No build required"'
 
+  submit-block-test:
+    name: submit block test
```
teor2345: This test name needs to match the name used by the branch protection rules.
Can you change it to the name in the GitHub CI status list?

See the other jobs in this file for examples. (The first part of the name comes from the workflow name already.)

docker/entrypoint.sh
```diff
@@ -78,6 +78,10 @@ case "$1" in
             ls -lh "$ZEBRA_CACHED_STATE_DIR"/*/* || (echo "No $ZEBRA_CACHED_STATE_DIR/*/*"; ls -lhR  "$ZEBRA_CACHED_STATE_DIR" | head -50 || echo "No $ZEBRA_CACHED_STATE_DIR directory")
             ls -lhR "$LIGHTWALLETD_DATA_DIR/db" || (echo "No $LIGHTWALLETD_DATA_DIR/db"; ls -lhR "$LIGHTWALLETD_DATA_DIR" | head -50 || echo "No $LIGHTWALLETD_DATA_DIR directory")
             cargo test --locked --release --features lightwalletd-grpc-tests --package zebrad --test acceptance -- --nocapture --include-ignored sending_transactions_using_lightwalletd
+        elif [[ "$TEST_SUBMIT_BLOCK" -eq "1" ]]; then
+            # Starting with a cached Zebra and lightwalletd tip, test sending transactions gRPC call to lightwalletd, which calls Zebra.
```
teor2345:
```suggestion
            # Starting with a cached Zebra tip, test sending a block to Zebra's RPC port.
```

zebrad/tests/common/cached_state.rs
```diff
@@ -144,3 +158,149 @@ async fn copy_directory(
...
+    let tip_height: u32 = blockchain_info["result"]["blocks"]
+        .as_u64()
+        .expect("unexpected block height: doesn't fit in u64")
+        .try_into()
+        .expect("unexpected block height: doesn't fit in u32");
+
+    let estimated_finalized_tip_height = tip_height - MAX_BLOCK_REORG_HEIGHT;
+
+    tracing::info!(
+        ?estimated_finalized_tip_height,
```
teor2345, line 263: I think this might help with debugging:

```suggestion
        ?tip_height,
        ?estimated_finalized_tip_height,
```

zebrad/tests/common/cached_state.rs
```diff
@@ -144,3 +158,149 @@ async fn copy_directory(
...
+    let zebrad_state_path = test_type
+        .zebrad_state_path(test_name)
+        .expect("already checked that there is a cached state path");
+
+    let Height(finalized_tip_height) =
+        load_tip_height_from_state_directory(network, zebrad_state_path.as_ref()).await?;
+
+    tracing::info!(
+        ?finalized_tip_height,
+        "finalized tip height from state directory"
```
teor2345:
```suggestion
        ?finalized_tip_height,
        non_finalized_tip_height = ?tip_height,
        ?estimated_finalized_tip_height,
        "got finalized tip height from state directory"
```

zebrad/tests/common/test_type.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// If this test fails, the failure is in `zebrad` without RPCs or `lightwalletd`.
+    /// If it succeeds, but the RPC tests fail, the problem is caused by RPCs or `lightwalletd`.
+    ///
+    /// This test requires a cached Zebra state.
+    UpdateZebraCachedStateNoRpc,
+
+    /// Launch `zebrad` and sync it to the tip, but don't launch `lightwalletd`.
+    ///
+    /// This test requires a cached Zebra state.
+    UpdateZebraCachedState,
```
teor2345: Optional: The variant name seems a bit ambiguous, how about:

```suggestion
    /// Launch `zebrad`, open a RPC port, and sync it to the tip, but don't launch `lightwalletd`.
    ///
    /// This test requires a cached Zebra state.
    #[allow(dead_code)]
    UpdateZebraCachedStateWithRpc,
```

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -252,136 +245,36 @@ pub async fn run() -> Result<()> {
...
-
-    let transactions =
-        load_transactions_from_block_after(partial_sync_height, network, full_sync_path.as_ref())
-            .await?;
-
-    Ok(transactions)
-}
-
-/// Loads transactions from a block that's after the specified `height`.
+/// Loads transactions from 1-3 block(s) after the chain tip of the cached state.
```
teor2345: Nitpick: this depends on the exact value of a constant, so we could link to it, or just be vague.

```suggestion
/// Loads transactions from a few block(s) after the chain tip of the cached state.
```

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -252,136 +245,36 @@ pub async fn run() -> Result<()> {
...
-                 collected {} transactions from {} blocks before {target_height:?}",
-                transactions.len(),
-                target_height - height.0 - 1,
-            );
-            break;
-        }
-
-        target_height += 1;
-    }
+    let transactions = get_future_blocks(network, test_type, test_name, 3)
```
teor2345: Is 3 blocks always going to be enough to get 10 non-coinbase transactions?
Sometimes if the network is quiet, there will be a few blocks in a row without any non-coinbase transactions.

I'd suggest something like 20 or 50 blocks.

Optional: Can you make the number of blocks into a constant at the top of the module?

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -252,136 +245,36 @@ pub async fn run() -> Result<()> {
...
-    Ok(transactions)
-}
-
-/// Loads transactions from a block that's after the specified `height`.
+/// Loads transactions from 1-3 block(s) after the chain tip of the cached state.
 ///
-/// Starts at the block after the block at the specified `height`, and stops when it finds a block
-/// from where it can load at least one non-coinbase transaction.
+/// Returns a list of valid transactions that are not in any of the blocks finalized to disk
+/// in the original `zebrad_state_path`.
```
teor2345: This isn't quite what is happening in the updated function:

```suggestion
/// Returns a list of non-coinbase transactions from blocks that have not been finalized to disk
/// in the `ZEBRA_CACHED_STATE_DIR`.
```

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -252,136 +245,36 @@ pub async fn run() -> Result<()> {
...
-        target_height += 1;
-    }
+    let transactions = get_future_blocks(network, test_type, test_name, 3)
+        .await?
+        .into_iter()
+        .map(|block| block.transactions)
+        .reduce(|mut p, mut n| {
+            if (p.len() + n.len()) < max_sent_transactions() {
+                p.append(&mut n)
+            };
```
teor2345: This doesn't quite do what the previous test did:
1. Coinbase transactions are invalid in the mempool and must be skipped
2. If there are 10 transactions in the first block, the returned list will be empty. Instead, we want to get as many transactions as are available, then limit the list to 10 transactions.

```suggestion
        // Skip coinbase transactions
        .flat_map(|block| block.transactions[1..])
        .take(max_sent_transactions())
        .cloned()
        .collect();
```

zebrad/tests/common/get_block_template_rpcs/submit_block.rs
```diff
@@ -1,170 +1,88 @@
...
-//         After finishing the sync, it will get the first 20 blocks in the non-finalized state
-//         (past the MAX_BLOCK_REORG_HEIGHT) via getblock rpc calls, get the finalized tip height
-//         of the updated cached state, restart zebra without peers, and submit blocks above the
-//         finalized tip height.
-
-use std::path::PathBuf;
+//! This test requires a cached chain state that is partially synchronized close to the
+//! network chain tip height. It will finish the sync and update the cached chain state.
+//!
+//! After finishing the sync, it will get the first 20 blocks in the non-finalized state
```
teor2345: Nitpick: Let's be vague here, these docs don't match the current constant used in the code

```suggestion
//! After finishing the sync, it will get the first few blocks in the non-finalized state
```

zebrad/tests/common/get_block_template_rpcs/submit_block.rs
```diff
@@ -1,170 +1,88 @@
...
-        ?cached_sync_height,
-        ?future_block_height,
-        "got cached sync height, copying state dir to tempdir"
+        ?network,
+        ?test_type,
+        "running submitblock test using zebrad",
     );
 
-    let copied_state_path = copy_state_directory(network, &zebrad_state_path).await?;
+    let raw_blocks: Vec<String> = get_raw_future_blocks(network, test_type, test_name, 3).await?;
```
teor2345: Optional: let's use a constant so it is easy to change if the test fails

zebrad/tests/common/test_type.rs
```diff
@@ -0,0 +1,316 @@
+use std::{env, path::PathBuf, time::Duration};
```
teor2345, line 3: Please add a doc comment at the top of each new module.

.github/workflows/continous-integration-docker.yml
```diff
@@ -575,3 +575,28 @@ jobs:
...
+  # - after every PR is merged to `main`
+  # - on every PR update
+  #
+  # If the state version has changed, waits for the new cached states to be created.
+  # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
+  submit-block-test:
+    name: submit block test
+    needs: test-full-sync
+    uses: ./.github/workflows/deploy-gcp-tests.yml
+    if: ${{ !cancelled() && !failure() }}
```
teor2345: This job needs to be skipped when we're running manual test.
Can you change the job conditions to match similar jobs?

teor2345 (changes requested):
Thanks, looks good!

I like the way you have re-used a lot of the existing code.

I have a few minor suggestions, and I noticed some things from the old tests that are missing.

teor2345:
Note for other reviewers:

It was very tricky for me to review the changes to `zebrad/tests/common/test_type.rs`, because the code was modified and moved in the same PR. GitHub and `git` diffs don't really deal with that very well.

Here's how I got a diff:
```sh
git checkout origin/update-ci-and-long-running-tests zebrad/tests/common/test_type.rs
cd ..
git diff --unified=10 zebra/zebrad/tests/common/lightwalletd.rs zebra/zebrad/tests/common/test_type.rs
```

teor2345:
If this PR makes the send transactions test fast enough, we can open a ticket to run it on every PR again.

teor2345:
> error: none of the selected packages contains these features: getblocktemplate

https://github.com/ZcashFoundation/zebra/actions/runs/3424052393/jobs/5703959112#step:6:140

docker/entrypoint.sh
```diff
@@ -78,6 +78,10 @@ case "$1" in
...
+            ls -lh "$ZEBRA_CACHED_STATE_DIR"/*/* || (echo "No $ZEBRA_CACHED_STATE_DIR/*/*"; ls -lhR  "$ZEBRA_CACHED_STATE_DIR" | head -50 || echo "No $ZEBRA_CACHED_STATE_DIR directory")
+            cargo test --locked --release --features getblocktemplate --package zebrad --test acceptance -- --nocapture --include-ignored submit_block
```
teor2345: > error: none of the selected packages contains these features: getblocktemplate

https://github.com/ZcashFoundation/zebra/actions/runs/3424052393/jobs/5703959112#step:6:140

```suggestion
            cargo test --locked --release --features getblocktemplate-rpcs --package zebrad --test acceptance -- --nocapture --include-ignored submit_block
```

.github/workflows/continous-integration-docker.patch.yml
arya2: That explains why some of the patch job names don't appear to exactly match the original names.

Updated in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8)

zebrad/tests/common/test_type.rs
arya2: Renamed in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8)

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: Moved to a constant and updated to 50 blocks in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8)

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: I missed filtering out the coinbase transaction, I like this better, thank you. I thought transactions may have unordered dependencies on other transactions in that block, but looking at the transactions' verifier, that is not the case. 

Applied a slightly modified version in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8) and [`c34a955`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/c34a95506f00f3830e3a299c3ef35862ee90d169)

Note: while `fold` would return an empty list, `reduce` uses the first element as the accumulator and calls the provided fn starting from the second element.

zebrad/tests/common/get_block_template_rpcs/submit_block.rs
arya2: Good idea, updated in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8)

zebrad/tests/common/test_type.rs
arya2: 🤦 Updated in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8) and [`50bb03`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/50bb037d52f63bcfdcec9d67ddb8fcafddd1a364)

.github/workflows/continous-integration-docker.yml
arya2: Updated in [`d72ea40`](https://github.com/ZcashFoundation/zebra/pull/5589/commits/d72ea40650530d1dfa5e3bdd73d9d146eb28fba8). Looks like that's the case for every job in this file except `test-full-sync` and `regenerate-stateful-disks`. Noted.

docker/entrypoint.sh
arya2: Thank you!

arya2:
> If this PR makes the send transactions test fast enough, we can open a ticket to run it on every PR again.

I think ~30m is fast enough, I created issue #5600 for running it on every PR update.

arya2:
mempool_requests_for_transactions test failed in CI Docker / Test All:

> failures:
> 
> ---- components::inbound::tests::fake_peer_set::mempool_requests_for_transactions stdout ----
> 
> Message:  response to `MempoolTransactionIds` request should match added_transaction_ids [Legacy(transaction::Hash("c4eaa58879081de3c24a7b117ed2b28300e7ec4c4c1dff1d3f1268b7857a4ddb"))]
> Location: zebrad/src/components/inbound/tests/fake_peer_set.rs:81

https://github.com/ZcashFoundation/zebra/actions/runs/3431178507/jobs/5719616560#step:3:9361

Issue #5384

zebrad/tests/common/lightwalletd/send_transaction_test.rs
teor2345: > I thought transactions may have unordered dependencies on other transactions in that block, but looking at the transactions' verifier, that is not the case.

Later transactions in a block can spend the transparent outputs of earlier transactions. That is checked in the state, where we are no longer verifying transactions in parallel.

.github/workflows/continous-integration-docker.yml
teor2345: And `lightwalletd-transactions-test`, but we're going to revert that back to running on PRs.

teor2345 (approved):
Looks good, let's merge!

teor2345:
I added the new test to the branch protection rules for `main`.

PR #5607, opened 2022-11-10
change(tests): use OS-assigned unallocated ports for getting random known ports

arya2:
## Motivation

The `random_known_port` may return an allocated port, causing a test error.

This PR reduces the likelihood of those errors by having the OS find a port that is unallocated at the time that `random_known_port` is called, and by broadening the range of possible ports.

## Solution

- Create a TcpListener with port 0 to find an unallocated port.
- Read the port number of new socket and drop the listener to close it.

### Reviewer Checklist


zebra-test/src/net.rs
```diff
@@ -47,37 +48,39 @@ pub fn zebra_skip_ipv6_tests() -> bool {
...
-    //   - the MOM-Clear port 51515
-    //      - https://docs.microsoft.com/en-us/troubleshoot/windows-server/networking/service-overview-and-network-port-requirements
-    //   - the LDAP Kerberos byte-swapped reservation 53249
-    //      - https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/ldap-kerberos-server-reset-tcp-sessions
-    //   - macOS and Windows sequential ephemeral port allocations,
-    //     starting from 49152:
-    //      - https://dataplane.org/ephemeralports.html
-
-    rand::thread_rng().gen_range(53500..60999)
+    let host_ip = Ipv4Addr::new(127, 0, 0, 1);
```
teor2345, line 113: If we get port conflicts in tests, we might want to try 0.0.0.0 here, because some tests want to bind to all addresses.

But let's see how it goes first? Maybe 127.0.0.1 is better.

teor2345 (commented):
I'm happy to try this, and if it ends up not working on Windows, or in some tests, we can go back to using a numerically random port selection.

But every test it *does* work for is one less test that will randomly fail.

> It's impossible to do this check in a reliable, cross-platform way.

Here's some background for that comment:

Windows sometimes waits a few minutes before allowing port reuse:
https://serverfault.com/questions/541070/linux-tcp-source-port-reuse-and-delay/609829#609829

On Linux and macOS, it would be easier if we set the SO_REUSEADDR option on Zebra's ports, but that's impossible for tokio and std Rust ports, because they don't expose that option. (Maybe they've improved the API since?)

https://programmer.group/so_reuseaddr-and-so_reuseport-socket-options-for-tcp-ip-programming.html

arya2:
> Windows sometimes waits a few minutes before allowing port reuse

It's a 2-minute default `TcpTimedWaitDelay`. That makes sense.

Tokio seems to support `SO_REUSEADDR` (https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html#implementations) but that's not cross-platform or supported by Rust std.

We can close this now or put it under `#[cfg(any(target_os = "macos", target_os = "linux"))]` if the tests pass on MacOS.

teor2345:
> Tokio seems to support `SO_REUSEADDR` (https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html#implementations) but that's not cross-platform or supported by Rust std.

I think Rust std added it recently as well?

https://github.com/rust-lang/rust/blob/1b225414f325593f974c6b41e671a0a0dc5d7d5e/library/std/src/sys_common/net.rs#L402-L403

teor2345:
> We can close this now or put it under `#[cfg(any(target_os = "macos", target_os = "linux"))]` if the tests pass on MacOS.

It seems to work, so let's use the same `#[cfg(not(windows))]` logic as Rust std?

Then we can keep the old code and use it for Windows.

arya2:
> I think Rust std added it recently as well?

That explains why this worked.

teor2345 (approved):
Looks good, but let's merge it separately from #5589, in case we need to revert?

teor2345 (approved):

teor2345:
@mergifyio update

dconnolly (approved):

PR #5616, opened 2022-11-10
change(consensus) verify that mempool transaction UTXOs are in the best chain

### LINKED ISSUE #5386: Use BestChainUtxo rather than AwaitUtxo in the mempool transaction validator
## Motivation

We can't use `AwaitUtxo` for validating mempool transactions, because it gets UTXOs from any chain or any queue.

### Design

Use `BestChainUtxo` rather than `AwaitUtxo` to verify mempool transactions.
(Block transactions should still use `AwaitUtxo`.)

arya2:
## Motivation

The transaction verifier currently checks that spent UTXOs of mempool transactions are in the state queue or on any chain, but it needs to verify that those spent UTXOs are on the best chain so that mined blocks including those transactions will be contextually valid.

Closes #5386 

## Solution

- Call state service with `BestChainUtxo` request instead of `AwaitUtxo` when verifying mempool transactions.
- Add `InputNotFound` variant to `TransactionError` for when the spent UTXOs are not present on the best chain.

## Review

### Reviewer Checklist

## Follow Up Work

- Add a `retry_list` in the mempool storage for transactions that weren't verified with `InputNotFound`?

### DIFF
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -178,8 +178,8 @@ pub enum TransactionError {
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
 
-    #[error("could not find an input UTXO in best chain")]
-    InputNotFound,
+    #[error("could not find a mempool transaction input UTXO in the best chain")]
+    TransparentInputNotFound,
 }
 
 impl From<BoxError> for TransactionError {
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -177,6 +177,9 @@ pub enum TransactionError {
 
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
+
+    #[error("could not find an input UTXO in best chain")]
+    InputNotFound,
 }
 
 impl From<BoxError> for TransactionError {
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -488,7 +488,7 @@ where
                 } else if is_mempool {
                     let query = state.clone().oneshot(zs::Request::BestChainUtxo(*outpoint));
                     if let zebra_state::Response::BestChainUtxo(utxo) = query.await? {
-                        utxo.ok_or(TransactionError::InputNotFound)?
+                        utxo.ok_or(TransactionError::TransparentInputNotFound)?
                     } else {
                         unreachable!("BestChainUtxo always responds with Option<Utxo>")
                     }
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -486,11 +486,13 @@ where
                     tracing::trace!("UXTO in known_utxos, discarding query");
                     output.utxo.clone()
                 } else if is_mempool {
-                    let query = state.clone().oneshot(zs::Request::BestChainUtxo(*outpoint));
-                    if let zebra_state::Response::BestChainUtxo(utxo) = query.await? {
+                    let query = state
+                        .clone()
+                        .oneshot(zs::Request::UnspentBestChainUtxo(*outpoint));
+                    if let zebra_state::Response::UnspentBestChainUtxo(utxo) = query.await? {
                         utxo.ok_or(TransactionError::TransparentInputNotFound)?
                     } else {
-                        unreachable!("BestChainUtxo always responds with Option<Utxo>")
+                        unreachable!("UnspentBestChainUtxo always responds with Option<Utxo>")
                     }
                 } else {
                     let query = state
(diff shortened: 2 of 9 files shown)


zebra-consensus/src/error.rs
```diff
@@ -177,6 +177,9 @@ pub enum TransactionError {
 
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
+
+    #[error("could not find an input UTXO in best chain")]
+    InputNotFound,
```
teor2345: Nitpick: We already use `NoInputs` and `NoOutputs` to mean "shielded or transparent". So we might want to use a more specific name here.

```suggestion
    #[error("could not find a mempool transaction input UTXO in the best chain")]
    TransparentInputNotFound,
```

zebra-consensus/src/transaction.rs
```diff
@@ -481,6 +483,13 @@ where
                 let utxo = if let Some(output) = known_utxos.get(outpoint) {
```
teor2345, line 485:
```suggestion
                // Currently, Zebra only supports known UTXOs in block transactions.
                // But it might support them in the mempool in future.
                let utxo = if let Some(output) = known_utxos.get(outpoint) {
```

zebra-consensus/src/transaction/tests.rs
```diff
@@ -177,6 +180,38 @@ fn v5_transaction_with_no_inputs_fails_validation() {
     );
 }
 
+#[tokio::test]
+async fn mempool_request_with_missing_input_is_rejected() {
```
teor2345, line 184: Can we have a test for the success case as well?

zebra-state/src/request.rs
```diff
@@ -458,6 +458,15 @@ pub enum Request {
     /// * [`Response::Transaction(None)`](Response::Transaction) otherwise.
     Transaction(transaction::Hash),
 
+    /// Looks up a UTXO identified by the given [`OutPoint`](transparent::OutPoint),
+    /// returning `None` immediately if it is unknown.
+    ///
+    /// Checks verified blocks in the finalized chain and the _best_ non-finalized chain.
+    ///
+    /// This request is purely informational, there is no guarantee that
+    /// the UTXO remains unspent in the best chain.
```
teor2345: This isn't quite what we want - we should be checking for unspent `Utxo`s in the best chain. If the transaction output has been spent in the best chain, the mempool transaction is invalid and should be rejected.

It's ok to change the `BestChainUtxo` implementation to fix this, and rename it if you want.

We do a similar thing in `address_utxos()`, but the synchronisation required for multi-address, multi-utxo handling is much more complex:
https://github.com/ZcashFoundation/zebra/blob/cbc2e393caacaf82af3ac46603be344bd5cb2ea0/zebra-state/src/service/read/address/utxo.rs#L92-L102

Since an `OutPoint` uniquely identifies a UTXO in a chain, we will only ever find one create and one spend per chain. So we don't need to duplicate the chain overlap and multi-address synchronization logic in `address_utxos()`. Any create in either finalized or non-finalized state makes the UTXO available, and any non-finalized spend makes it unavailable again. (The finalized state automatically deletes spent UTXOs from its index.)

So we'll need an `OutPoint` argument to the new read function. Then we can call `db.utxo()`,  `chain.created_utxo()`, and a new `chain.spent_utxo()` method directly, combine the results, and we're done.

teor2345 (changes requested):
Thanks, this looks good!

But I think we need to check for unspent outputs, not just any available output. See my detailed note on the request comment.

We might also want to add a TODO to rename `AwaitUtxo` to `AwaitTransactionOutput` (the `U` means "Unspent"). There is already a TODO to combine the `Output`, `Utxo`, and `OrderedUtxo` types, maybe we can add a rename to it as well. But I don't want to do a mass rename during an audit freeze.

zebra-state/src/request.rs
arya2: Updated in [`5b8eadc`](https://github.com/ZcashFoundation/zebra/pull/5616/commits/5b8eadc66efbebfb6a1cfc2f6180a82fe40b0ed5).

I'm open to suggestions on naming.

zebra-consensus/src/transaction/tests.rs
arya2: Added a test for the success case in [`ece1be5`](https://github.com/ZcashFoundation/zebra/pull/5616/commits/ece1be55707a70f32b9a18ec121d640f132dbcf9).

zebra-state/src/request.rs
teor2345: Happy to deal with naming later if we need to, the current name is clear enough!

teor2345 (approved):
Thanks!

I particularly like how you re-used the utxo() function code.

PR #5602, opened 2022-11-09
feat(ssh): enable OS Login for GCP test instances

gustavovalverde:
## Motivation

We've been dealing with several different issues related to SSH connections. So far one of the solutions was to have a fixed SSH-key to connect from GitHub Actions, but this also reduces flexibility and maintainability, and goes against our security-wise decision of not having long-lived artifacts (JSON tokens, SSH keys) without expiration, which could be used in any point in time.

This PR implements OS Login as workaround (and possible root cause solution)

**Note:** _Not all changes to make this work are being done in GitHub Actions, as some tasks are tied to changes in the infrastructure_

Fixes: #5494
Fixes: #5069

### Specifications

https://cloud.google.com/compute/docs/oslogin#benefits_of_os_login

### Designs

- Activate OS Login on specific VMs
- Give the GitHub Actions service account the required permissions to connect as a user

## Solution

- Enable OS Login
- Expire OS Login SSH keys after 30 seconds (as we don't need those anymore)
- Delete any remaining keys (if any) at the end of the workflow run
- Use `busybox` instead of failing `debian-11` container image #5618 
- Add sudo to a missing `docker` command

## Review

Anyone from DevOps after running this PR successfully ~5+ times

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

Configure OS-login project-wide after this PR has been merged.

PR #5623, opened 2022-11-11
change(rpc): return an error from getblocktemplate method if Zebra is not synced to network tip

### LINKED ISSUE #5466: Return an error from getblocktemplate if Zebra isn't synced to the tip
## Motivation

`s-nomp` uses `getblocktemplate` to check if the node is synced to the tip:
https://github.com/s-nomp/node-stratum-pool/blob/d86ae73f8ff968d9355bb61aac05e0ebef36ccb5/lib/pool.js#L140

### Designs

- Use a `SyncStatus` to check that Zebra has finished syncing
- Allow the same clock skew as the Zcash network, which is 100 blocks, based on the standard rule:

> a full validator MUST NOT accept blocks with nTime more than two hours in the future according to its clock. This is not strictly a consensus rule because it is nondeterministic, and clock time varies between nodes.

Completed work:
<details>

Get the chain tip and the estimated height, and see if they are close enough.
Here is some existing RPC code that gets those fields:
https://github.com/ZcashFoundation/zebra/blob/a2dba8cc5dd482531883ea982398cf0e36c6fcdd/zebra-rpc/src/methods.rs#L370-L397

</details>

arya2:
## Motivation

We want to return an error if there's distance between Zebra's best chain tip height and the estimated network chain tip height because mining pools may use the `getblocktemplate` method to check if the node is synced.

Closes #5466

## Solution

- Return a server error if the estimated network height is higher than Zebra's chain tip height

## Review

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/chain_tip.rs
+++ b/zebra-chain/src/chain_tip.rs
@@ -58,6 +58,21 @@ pub trait ChainTip {
 
         Some(estimator.estimate_height_at(now))
     }
+
+    /// Return an estimate of how many blocks there are ahead of zebra's best chain tip
+    /// until the network chain tip.
+    ///
+    /// The estimate is calculated based on the current local time, the block time of the best tip
+    /// and the height of the best tip.
+    ///
+    /// This estimate may be negative if the current local time is behind the chain tip block's timestamp.
+    fn estimate_distance_to_network_chain_tip(&self, network: Network) -> Option<i32> {
+        self.estimate_network_chain_tip_height(network, Utc::now())
+            .and_then(|network_height| {
+                self.best_tip_height()
+                    .map(|zebra_height| network_height - zebra_height)
+            })
+    }
 }
 
 /// A chain tip that is always empty.
--- a/zebra-chain/src/chain_tip.rs
+++ b/zebra-chain/src/chain_tip.rs
@@ -59,19 +59,26 @@ pub trait ChainTip {
         Some(estimator.estimate_height_at(now))
     }
 
-    /// Return an estimate of how many blocks there are ahead of zebra's best chain tip
-    /// until the network chain tip.
+    /// Return an estimate of how many blocks there are ahead of Zebra's best chain tip
+    /// until the network chain tip, and Zebra's best chain tip height.
     ///
     /// The estimate is calculated based on the current local time, the block time of the best tip
     /// and the height of the best tip.
     ///
     /// This estimate may be negative if the current local time is behind the chain tip block's timestamp.
-    fn estimate_distance_to_network_chain_tip(&self, network: Network) -> Option<i32> {
-        self.estimate_network_chain_tip_height(network, Utc::now())
-            .and_then(|network_height| {
-                self.best_tip_height()
-                    .map(|zebra_height| network_height - zebra_height)
-            })
+    fn estimate_distance_to_network_chain_tip(
+        &self,
+        network: Network,
+    ) -> Option<(i32, block::Height)> {
+        let (current_height, current_block_time) = self.best_tip_height_and_block_time()?;
+
+        let estimator =
+            NetworkChainTipHeightEstimator::new(current_block_time, current_height, network);
+
+        Some((
+            estimator.estimate_height_at(Utc::now()) - current_height,
+            current_height,
+        ))
     }
 }
 
--- a/zebra-chain/src/chain_tip/mock.rs
+++ b/zebra-chain/src/chain_tip/mock.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use chrono::{DateTime, Utc};
 use tokio::sync::watch;
 
-use crate::{block, chain_tip::ChainTip, transaction};
+use crate::{block, chain_tip::ChainTip, parameters::Network, transaction};
 
 /// A sender to sets the values read by a [`MockChainTip`].
 pub struct MockChainTipSender {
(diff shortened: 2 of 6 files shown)


zebra-chain/src/chain_tip.rs
```diff
@@ -58,6 +58,21 @@ pub trait ChainTip {
...
         Some(estimator.estimate_height_at(now))
     }
+
+    /// Return an estimate of how many blocks there are ahead of zebra's best chain tip
+    /// until the network chain tip.
+    ///
+    /// The estimate is calculated based on the current local time, the block time of the best tip
+    /// and the height of the best tip.
+    ///
+    /// This estimate may be negative if the current local time is behind the chain tip block's timestamp.
```
dconnolly, line 68: Nice

arya2: Do we know why past times are handled this way?

https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/chain_tip/network_chain_tip_height_estimator.rs#L122-L137

dconnolly:
Restarting this, this failure may be related to flipping backend switches in gcloud related to OS Login:

https://github.com/ZcashFoundation/zebra/pull/5602

https://github.com/ZcashFoundation/zebra/actions/runs/3447560707/jobs/5754223166
image

gustavovalverde:
> Restarting this, this failure may be related to flipping backend switches in gcloud related to OS Login

@dconnolly this is a missing sudo command. I added it in #5602 but any of you can add it here if needed (not sure why this didn't fail before)

dconnolly (dismissed):
lgtm!

zebra-chain/src/chain_tip.rs
dconnolly: I don't recall, but I'd be willing to put money on "this is how zcashd handles it"

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1266,6 +1266,18 @@ jobs:
...
+      # could cause the following error:
+      # `Login profile size exceeds 32 KiB. Delete profile values to make additional space`
+      - name: Delete temporal SSH keys
+        continue-on-error: true
+        run: |
+          for i in $(gcloud compute os-login ssh-keys list --format="table[no-heading](value.fingerprint)"); do 
+            echo "$i";
+            gcloud compute os-login ssh-keys remove --key "$i" || true;
+          done
+
```
dconnolly: @gustavovalverde should we do this in #5602 and merge?

arya2: Seems to be needed for every PR to work so I reverted this for now and will wait for https://github.com/ZcashFoundation/zebra/pull/5602 to merge.

arya2:
@Mergifyio update

oxarbitrage:
This looks really good but i will lime to see a test with a chain synced up to tip and a call to `getblocktemplate` making sure we don't hit the error and the function is still usable.

We don't need to add this to CI but just a log here will be enough.

arya2:
> This looks really good but i will lime to see a test with a chain synced up to tip and a call to `getblocktemplate` making sure we don't hit the error and the function is still usable.
> 
> We don't need to add this to CI but just a log here will be enough.

This helped catch an issue!

I'm thinking I want to add a test that syncs to the tip for getblocktemplate in CI too for testing https://github.com/ZcashFoundation/zebra/pull/5630 with the non-finalized state, but for now:

```json
{
  "jsonrpc": "2.0",
  "result": {
    "capabilities": [],
    "version": 4,
    "previousblockhash": "0000000000000000000000000000000000000000000000000000000000000000",
    "blockcommitmentshash": "0000000000000000000000000000000000000000000000000000000000000000",
    "lightclientroothash": "0000000000000000000000000000000000000000000000000000000000000000",
    "finalsaplingroothash": "0000000000000000000000000000000000000000000000000000000000000000",
    "defaultroots": {
      "merkleroot": "f00a46d38fccf8b48554c1f4651eec6bbc294c79ac2e2bc91e63bfcdc574eaa9",
      "chainhistoryroot": "0000000000000000000000000000000000000000000000000000000000000000",
      "authdataroot": "8231fdfe4e63846249d0880168deb69bcf7d81cc2f762937677f203fa8d8fd2a",
      "blockcommitmentshash": "0000000000000000000000000000000000000000000000000000000000000000"
    },
    "transactions": [],
    "coinbasetxn": {
      "data": "050000800a27a726b4d0d6c20000000082b31c00010000000000000000000000000000000000000000000000000000000000000000ffffffff040382b31c0000000004286bee000000000017a914d45cb1adffb5215a42720532a076f02c7c778c908738c94d010000000017a914fd237ff37cc91b87b7cef8408ade1bc14a4e67158740787d010000000017a914931fec54c1fea86e574462cc32013f5400b891298780b2e60e0000000017a9147d46a730d31f97b1930d3368a967c309bd4d136a87000000",
      "hash": "f00a46d38fccf8b48554c1f4651eec6bbc294c79ac2e2bc91e63bfcdc574eaa9",
      "authdigest": "8231fdfe4e63846249d0880168deb69bcf7d81cc2f762937677f203fa8d8fd2a",
      "depends": [],
      "fee": 0,
      "sigops": 0,
      "required": true
    },
    "target": "",
    "mintime": 0,
    "mutable": ["time", "transactions", "prevblock"],
    "noncerange": "00000000ffffffff",
    "sigoplimit": 20000,
    "sizelimit": 2000000,
    "curtime": 0,
    "bits": "",
    "height": 0
  },
  "id": 123
}
```

arya2:
@Mergifyio update

dconnolly (approved):

PR #5630, opened 2022-11-15
change(rpc): check contextual validity of getblocktemplate response

### LINKED ISSUE #5376: Do contextual validation on mempool transactions using the state best tip
## Motivation

Edit: This means that each transaction in the mempool must be valid based on the best tip, not any tip.

We'll want a new state request for this.

### Design

Contextual validation:
https://github.com/ZcashFoundation/zebra/blob/afb8b3d4775d89b3f9223447341bf6ce152f5c3a/book/src/dev/rfcs/0002-parallel-verification.md

State design details:
https://github.com/ZcashFoundation/zebra/blob/afb8b3d4775d89b3f9223447341bf6ce152f5c3a/book/src/dev/rfcs/0005-state-updates.md

arya2:
## Motivation

The getblocktemplate response must be contextually valid based on the best tip in the state.

Needs other getblocktemplate fields for testing the success case.

Closes #5376

## Solution

- Add a new state request for checking the contextual validity of a prepared block without committing it.
- Make a prepared block with default `nonce` and `solution` fields to use the new state request.
- Update vectors and snapshot tests to use a MockService that returns a `Validated` response.

## Review

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -1522,13 +1522,6 @@ impl Service<ReadRequest> for ReadStateService {
             // Used by get_block_hash RPC.
             #[cfg(feature = "getblocktemplate-rpcs")]
             ReadRequest::BestChainBlockHash(height) => {
-                metrics::counter!(
-                    "state.requests",
-                    1,
-                    "service" => "read_state",
-                    "type" => "best_chain_block_hash",
-                );
-
                 let timer = CodeTimer::start();
 
                 let state = self.clone();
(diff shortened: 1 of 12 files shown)


zebra-state/src/service.rs
```diff
@@ -797,6 +797,84 @@ impl ReadStateService {
...
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    #[allow(clippy::unwrap_in_result)]
+    #[tracing::instrument(level = "debug", skip_all)]
+    fn check_best_chain_contextual_validity(
+        &self,
+        block: PreparedBlock,
+    ) -> Result<(), crate::ValidateContextError> {
+        if let Some(best_chain) = self.latest_non_finalized_state().best_chain() {
+            let best_chain_latest_non_finalized_state = NonFinalizedState::build(self.network)
+                .insert_chain(Arc::clone(best_chain))
```
arya2, line 811: qualifying the `Arc::clone` so it's obvious that nothing's being copied without looking at the type.

arya2:
@Mergifyio update

oxarbitrage:
This one will conflict with https://github.com/ZcashFoundation/zebra/pull/5659 in the tests as i added some stuff from this PR there. I will prefer https://github.com/ZcashFoundation/zebra/pull/5659 to go first just because i have another PR to push that is based on it and it will be easier for me to do it that way.

arya2:
> This one will conflict with #5659 in the tests as i added some stuff from this PR there. I will prefer #5659 to go first just because i have another PR to push that is based on it and it will be easier for me to do it that way.

Sounds good, I'd also prefer #5659 go first so this PR could use the populated `target` field for the `difficulty_threshold` of the prepared block.

teor2345 (changes requested):
I think the ticket for this PR was a bit confusing. We need to change how the mempool checks transactions, so it always uses the best chain.

If we wait until the transactions are in the `getblocktemplate` response, it is too late, because all we can do is return an error to the caller. We don't know which transactions to remove to make the block valid.

Edit: It is also going to be tricky to verify the block, because the solution in the header requires Proof of Work. (Correct equihash solutions are expensive to create, so we can't do that in tests or in the RPC.)

We might want to turn this PR into a ticket, because the code looks like it could be useful for testing.

I'll add some specific notes to the ticket about where we want to change things.

arya2:
> I think the ticket for this PR was a bit confusing. We need to change how the mempool checks transactions, so it always uses the best chain.
> 
> If we wait until the transactions are in the `getblocktemplate` response, it is too late, because all we can do is return an error to the caller. We don't know which transactions to remove to make the block valid.

Oops. 🤦 I thought I just hadn't come across the code that was checking the nullifiers, I understand now.

> It is also going to be tricky to verify the block, because the solution in the header requires Proof of Work. (Correct equihash solutions are expensive to create, so we can't do that in tests or in the RPC.)
>
> The code looks like it could be useful for testing.

I believe the solution in the header is only checked by the block verifier. I created issue #5681 for adding this as a test.

I'm closing this because it would needlessly hurt performance in production.

PR #5659, opened 2022-11-18
feat(rpc): populate some getblocktemplate RPC block header fields using the state best chain tip 

oxarbitrage:
## Motivation

[More to be added]

Populate the following fields:
- capabilities
- target
- min time
- cur time
- bits
- height
- previous block hash

https://github.com/ZcashFoundation/zebra/issues/5455

## Solution

[TBA]

### Reviewer Checklist

## Follow Up Work

Populate root/commitment fields


oxarbitrage:
`zebra_rpc_conflict` failing https://github.com/ZcashFoundation/zebra/actions/runs/3493963607/jobs/5849620107#step:3:9825

Retrying.

zebra-state/src/response.rs
```diff
@@ -115,6 +118,19 @@ pub enum ReadResponse {
...
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Response to [`ReadRequest::ChainInfo`](crate::ReadRequest::ChainInfo) with the state
+    /// information needed by the `getblocktemplate` RPC method.
+    ChainInfo(GetBlockTemplateChainInfo),
+}
+
+#[cfg(feature = "getblocktemplate-rpcs")]
+/// A structure with the information needed from the state to build a `getblocktemplate` RPC response.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct GetBlockTemplateChainInfo {
```
oxarbitrage, line 131: This structure exists just because i expect more fields to be added to populate the still missing getblocktemplate fields.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,62 @@
...
+    let relevant_data = relevant_chain.iter().map(|block| {
+        (
+            block.borrow().header.difficulty_threshold,
+            block.borrow().header.time,
+        )
+    });
+
+    // If we don't have enough context we just return 1.
+    if relevant_data.len() < MAX_CONTEXT_BLOCKS {
+        return CompactDifficulty::from(ExpandedDifficulty::from(U256::one()));
```
oxarbitrage: Zebra will panic here if there is not enough context. I made this to make tests pass but i am unsure if it will be the best way to handle.

arya2: I think we should return an error saying there's not enough context and use a `MockService` to return this in the test.

I used a `MockService` for the check contextual validity PR [here](https://github.com/ZcashFoundation/zebra/pull/5630/files#diff-78bcac22b320634c3859f8282d82427e7fa135c34607223ffdaad3c71c7d8379R116).

zebra-chain/src/work/difficulty.rs
```diff
@@ -112,6 +112,16 @@ impl fmt::Debug for ExpandedDifficulty {
     }
 }
 
+impl fmt::Display for ExpandedDifficulty {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut buf = [0; 32];
+        // Use the same byte order as block::Hash
```
arya2, line 119: 👍

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -286,6 +291,9 @@ where
                 data: None,
             })?;
 
+            let tip_hash = best_chain_tip_hash(&latest_chain_tip)?;
+            let tip_time = best_chain_tip_time(&latest_chain_tip)?;
+
             let (estimated_distance_to_chain_tip, tip_height) = latest_chain_tip
                 .estimate_distance_to_network_chain_tip(network)
```
arya2, line 295: Optional: replace this with a `get_tip_info` method to get the estimated distance, height, hash, and time all at once.

zebra-chain/src/work/difficulty.rs
```diff
@@ -257,6 +267,11 @@ impl CompactDifficulty {
         let expanded = self.to_expanded()?;
         Work::try_from(expanded).ok()
     }
+
+    /// Returns the raw inner value.
+    pub fn to_value(self) -> u32 {
```
arya2: Should this be feature flagged?

Optional: I think a copy op would be ok here

```suggestion
    #[cfg(feature = "getblocktemplate-rpcs")]
    pub fn to_value(&self) -> u32 {
```

zebra-chain/src/work/difficulty.rs
```diff
@@ -112,6 +112,16 @@ impl fmt::Debug for ExpandedDifficulty {
     }
 }
 
+impl fmt::Display for ExpandedDifficulty {
```
arya2, line 116: Should this be feature flagged?

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +337,38 @@ where
             // Convert into TransactionTemplates
             let mempool_txs = mempool_txs.iter().map(Into::into).collect();
 
-            let empty_string = String::from("");
+            // Calling state chainfo to additional data
```
```suggestion
            // Calling state with `ChainInfo` request for relevant chain data
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +337,38 @@ where
...
+            };
+
+            let expected_difficulty = match chain_info.expected_difficulty {
+                Some(difficulty) => difficulty,
+                _ => Err(Error {
+                    code: ErrorCode::ServerError(0),
+                    message: "Not enough state in the chain".to_string(),
+                    data: None,
+                })?,
+            };
```
```suggestion
            let expected_difficulty = chain_info.expected_difficulty.ok_or(Error {
                code: ErrorCode::ServerError(0),
                message: "Not enough blocks in the chain".to_string(),
                data: None,
            })?;
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +337,38 @@ where
...
+
             Ok(GetBlockTemplate {
-                capabilities: vec![],
+                capabilities: vec!["proposal".to_string()],
```
arya2: Optional: set default values in a constructor fn for `GetBlockTemplate`

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +337,38 @@ where
...
 
                 version: ZCASH_BLOCK_VERSION,
 
-                previous_block_hash: GetBlockHash([0; 32].into()),
+                previous_block_hash: GetBlockHash(tip_hash),
```
arya2, line 362: 👍

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -346,9 +383,14 @@ where
 
                 coinbase_txn: TransactionTemplate::from_coinbase(&coinbase_tx, miner_fee),
 
-                target: empty_string.clone(),
+                target: format!(
+                    "{}",
+                    expected_difficulty
+                        .to_expanded()
+                        .unwrap_or_else(|| ExpandedDifficulty::from(U256::zero()))
```
arya2: There's this assert that seems concerning: 

> `assert!(self.0 > 0.into(), "Zero difficulty values are invalid")`

https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/work/difficulty.rs#L345

I think we should try returning an error if it's `None` and reevaluate if that causes problems.

zebra-state/src/request.rs
```diff
@@ -744,6 +744,14 @@ pub enum ReadRequest {
...
     /// * [`ReadResponse::BlockHash(None)`](ReadResponse::BlockHash) otherwise.
     BestChainBlockHash(block::Height),
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Get state information from the best block chain.
+    ///
+    /// Returns [`ReadResponse::ChainInfo(info)`](ReadResponse::ChainInfo) where `info` is a
+    /// [`ReadResponse::GetBlockTemplateChainInfo`](ReadResponse::GetBlockTemplateChainInfo)` structure containing
+    /// best chain state information.
+    ChainInfo(),
```
arya2: Nitpick:

```suggestion
    ChainInfo,
```

zebra-state/src/service.rs
```diff
@@ -1554,6 +1557,60 @@ impl Service<ReadRequest> for ReadStateService {
                 .map(|join_result| join_result.expect("panic in ReadRequest::BestChainBlockHash"))
                 .boxed()
             }
+
+            // Used by get_block_template RPC.
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            ReadRequest::ChainInfo() => {
+                metrics::counter!(
```
arya2: A `req.count_metric()` call was recently added at the top of this fn that's already doing this.

zebra-state/src/service.rs
```diff
@@ -1554,6 +1557,60 @@ impl Service<ReadRequest> for ReadStateService {
...
+                    "state.requests",
+                    1,
+                    "service" => "read_state",
+                    "type" => "chain_info",
+                );
+
+                let timer = CodeTimer::start();
+
+                let state = self.clone();
+                let network = self.network;
```
arya2: Nitpick: we could use `state.network` below and prune this line.

zebra-state/src/service.rs
```diff
@@ -1554,6 +1557,60 @@ impl Service<ReadRequest> for ReadStateService {
...
+                        let tip = state.non_finalized_state_receiver.with_watch_data(
+                            |non_finalized_state| {
+                                read::tip(non_finalized_state.best_chain(), &state.db)
+                            },
+                        );
+
+                        // Difficulty
+                        let expected_difficulty = tip.map(|tip| {
+                            state.non_finalized_state_receiver.with_watch_data(
+                                |non_finalized_state| {
```
arya2: This could be part of the `with_watch_data` call above. There's also a `latest_non_finalized_state` method that could be used instead of `with_watch_data`.

zebra-state/src/service/check/difficulty.rs
```diff
@@ -99,16 +99,16 @@ impl AdjustedDifficulty {
         let previous_block_height = (candidate_block_height - 1)
             .expect("contextual validation is never run on the genesis block");
 
-        AdjustedDifficulty::new_from_header(
-            &candidate_block.header,
+        AdjustedDifficulty::new_from_header_time(
```
arya2, line 102: 👍

arya2 (commented):
This looks great! 🎉

zebra-state/src/service.rs
oxarbitrage: Thanks, i did this in https://github.com/ZcashFoundation/zebra/pull/5659/commits/e80514a675a1c278b8c01d3b072ca68164d1fa0d

zebra-state/src/service.rs
oxarbitrage: It seems i missed that change, thanks! I removed it at https://github.com/ZcashFoundation/zebra/pull/5659/commits/e80514a675a1c278b8c01d3b072ca68164d1fa0d . 

Also removed from `BestChainBlockHash`

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: Added at https://github.com/ZcashFoundation/zebra/pull/5659/commits/9e85ec38a582fde443413a0b9b3ff6e3cbb779ec

zebra-state/src/request.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/5659/commits/9e85ec38a582fde443413a0b9b3ff6e3cbb779ec

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/5659/commits/9e85ec38a582fde443413a0b9b3ff6e3cbb779ec

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: Yes, the idea here was to return one just as https://github.com/ZcashFoundation/zebra/pull/5659/files/6ba12c388493814f621b28681ad1e223a05001c7#diff-8c6198fdf18f8b2691f9c876d3a10f653827cee5a8cce4e8496451d03886c122 but i forgot to update.

Whatever is the case i will like to solve this 2 issues with the mocked service, however i am not totally sure how to do it yet.

zebra-chain/src/work/difficulty.rs
oxarbitrage: yea, added that in https://github.com/ZcashFoundation/zebra/pull/5659/commits/d30e5e4143063d685f140697a08c0f95cf429147

zebra-chain/src/work/difficulty.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/5659/commits/d30e5e4143063d685f140697a08c0f95cf429147

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: It might be the same as [here](https://github.com/ZcashFoundation/zebra/pull/5630/files#diff-39f9960d69871ed9095292c4272532f519193be4bb30b5e56c61c3cc4c706c16) or [here](https://github.com/ZcashFoundation/zebra/pull/5630/files#diff-78bcac22b320634c3859f8282d82427e7fa135c34607223ffdaad3c71c7d8379) but for `ChainInfo` with `ExpandedDifficulty::from(U256::one())` as the response.

zebra-state/src/service.rs
oxarbitrage: Thanks, this was done at https://github.com/ZcashFoundation/zebra/pull/5659/commits/9a12c17587376aab9220e331ea4accebfa17d7d0

zebra-chain/src/work/difficulty.rs
```diff
@@ -257,6 +268,12 @@ impl CompactDifficulty {
         let expanded = self.to_expanded()?;
         Work::try_from(expanded).ok()
     }
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Returns the raw inner value.
+    pub fn to_value(&self) -> u32 {
+        self.0
+    }
```
teor2345, line 276: Can we implement the formatting on `CompactDifficulty` instead of exposing the raw bytes?

`CompactDifficulty` isn't a `u32` integer, it is actually a floating-point number in a non-standard format. So exposing the value as a `u32` allows operations that we don't want. (For example, integer addition.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -286,6 +290,9 @@ where
                 data: None,
             })?;
 
+            let tip_hash = best_chain_tip_hash(&latest_chain_tip)?;
+            let tip_time = best_chain_tip_time(&latest_chain_tip)?;
```
teor2345: We need the tip height, hash, time, and `ChainInfo` to come from the same tip block.

Currently, this PR makes 4 different calls to the state, which means we could get info from 1-4 different tip blocks. (Blocks can be added between each call.)

It's ok to do the estimate separately, because it is just an estimate.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +336,35 @@ where
...
+
+            let expected_difficulty = chain_info.expected_difficulty.ok_or(Error {
+                code: ErrorCode::ServerError(0),
+                message: "Not enough blocks in the chain".to_string(),
+                data: None,
+            })?;
+
             Ok(GetBlockTemplate {
-                capabilities: vec![],
+                capabilities: vec!["proposal".to_string()],
```
teor2345: Zebra does not support the `proposal` feature, so the `capabilities` must be empty:
https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request

We can open an optional ticket for it - we don't know if any mining pools need it yet.

```suggestion
                capabilities: Vec::new(),
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -18,6 +18,8 @@ pub struct GetBlockTemplate {
     /// - `proposal`: <https://en.bitcoin.it/wiki/BIP_0023#Block_Proposal>
     /// - `longpoll`: <https://en.bitcoin.it/wiki/BIP_0022#Optional:_Long_Polling>
     /// - `serverlist`: <https://en.bitcoin.it/wiki/BIP_0023#Logical_Services>
+    ///
+    /// In zcashd, this field is hardcoded to ["proposal"]. We do the same in zebra.
```
teor2345: Zebra doesn't support the `proposal` feature, and we don't know if we will need to yet.

```suggestion
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -361,11 +403,13 @@ where
 
                 size_limit: MAX_BLOCK_BYTES,
 
-                cur_time: 0,
+                cur_time: chrono::Utc::now().timestamp(),
```
teor2345: I think the ticket was wrong here, the actual specification is:
> the current time as seen by the server (recommended for block time) - note this is not necessarily the system clock, and must fall within the mintime/maxtime rules 

https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request

The consensus rules are:
> For each block other than the genesis block , nTime MUST be strictly greater than the median-time-past of that block.
> For each block at block height 2 or greater on Mainnet , or block height 653606 or greater on Testnet , nTime MUST be less than or equal to the median-time-past of that block plus 90 · 60 seconds.

So it's ok to start with the current time, but it must be limited to the range specified by the consensus rules.

Since Zebra's mandatory checkpoint is greater than block 653606, and we return an error if Zebra isn't synced to the tip, we can apply these limits to every RPC response. We don't need to check the block height part of the rule.

This time must also be the same as the time used to calculate the difficulty, so it needs to be returned in the state request.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +336,35 @@ where
...
+            let chain_info = match response {
+                ReadResponse::ChainInfo(chain_info) => chain_info,
+                _ => unreachable!("chaininfo always respond even if all data inside is `None`"),
+            };
+
+            let expected_difficulty = chain_info.expected_difficulty.ok_or(Error {
+                code: ErrorCode::ServerError(0),
+                message: "Not enough blocks in the chain".to_string(),
+                data: None,
+            })?;
```
teor2345: It seems like a "Not enough blocks in the chain" error should apply to the whole response, rather than a specific field of the response?

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -346,9 +379,18 @@ where
...
-                target: empty_string.clone(),
+                target: format!(
+                    "{}",
+                    expected_difficulty
+                        .to_expanded()
+                        .ok_or(Error {
+                            code: ErrorCode::ServerError(0),
+                            message: "Not enough blocks in the chain".to_string(),
+                            data: None,
+                        })?
```
teor2345: The state should never return an invalid difficulty, because it validates blocks before committing them:
```suggestion
                        .expect("state always returns a valid difficulty value")
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -89,16 +91,16 @@ pub struct GetBlockTemplate {
     #[serde(rename = "sizelimit")]
     pub size_limit: u64,
 
-    /// Add documentation.
+    /// The current system time as seconds since epoch.
```
teor2345:
```suggestion
    /// > the current time as seen by the server (recommended for block time).
    /// > note this is not necessarily the system clock, and must fall within the mintime/maxtime rules
    ///
    /// <https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request>
```

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -150,7 +167,16 @@ fn snapshot_rpc_getblocktemplate(
     coinbase_tx: Transaction,
     settings: &insta::Settings,
 ) {
-    settings.bind(|| insta::assert_json_snapshot!("get_block_template", block_template));
+    settings.bind(|| {
+        insta::assert_json_snapshot!("get_block_template", block_template, {
+            ".curtime" => dynamic_redaction(|_value, _path| {
+                // TODO: assert that the value looks like a valid time.
```
teor2345: We might be able to remove this redaction, after we implement the min and max time consensus rules.

zebra-state/src/request.rs
```diff
@@ -744,6 +744,14 @@ pub enum ReadRequest {
     /// * [`ReadResponse::BlockHash(Some(hash))`](ReadResponse::BlockHash) if the block is in the best chain;
     /// * [`ReadResponse::BlockHash(None)`](ReadResponse::BlockHash) otherwise.
     BestChainBlockHash(block::Height),
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Get state information from the best block chain.
+    ///
+    /// Returns [`ReadResponse::ChainInfo(info)`](ReadResponse::ChainInfo) where `info` is a
+    /// [`ReadResponse::GetBlockTemplateChainInfo`](ReadResponse::GetBlockTemplateChainInfo)` structure containing
```
teor2345: Can you fix this doc warning?

zebra-state/src/service.rs
```diff
@@ -1522,13 +1525,6 @@ impl Service<ReadRequest> for ReadStateService {
             // Used by get_block_hash RPC.
             #[cfg(feature = "getblocktemplate-rpcs")]
             ReadRequest::BestChainBlockHash(height) => {
-                metrics::counter!(
```
teor2345, line 1525: Nice catch!

zebra-state/src/service.rs
```diff
@@ -1554,6 +1550,44 @@ impl Service<ReadRequest> for ReadStateService {
...
+                            read::tip(latest_non_finalized_state.best_chain(), &state.db)
+                                .map(|tip| {
+                                    read::difficulty::relevant_chain_difficulty(
+                                        &latest_non_finalized_state,
+                                        &state.db,
+                                        tip,
+                                        state.network,
+                                    )
+                                })
+                                .unwrap_or(None);
```
teor2345: This request reads `state.db` twice, which can return different values if a block is committed between the reads. If that happens, the RPC needs to correct the different values, or return an error.

It will be easier to do these corrections if `read::tip()` is called by `read::difficulty::relevant_chain_difficulty()`.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,64 @@
...
+    let relevant_data = relevant_chain.iter().map(|block| {
+        (
+            block.borrow().header.difficulty_threshold,
+            block.borrow().header.time,
+        )
+    });
+
+    if relevant_data.len() < MAX_CONTEXT_BLOCKS {
+        return None;
+    }
```
teor2345: Can we turn this into an `expect()` or returning an error?
Or do we need it for the tests?

```suggestion
    // The getblocktemplate RPC returns an error if Zebra is not synced to the tip.
    // So this will never happen in production code.
    if relevant_data.len() < MAX_CONTEXT_BLOCKS {
        return None;
    }
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -315,6 +322,10 @@ where
             let miner_fee = miner_fee(&mempool_txs);
 
             let block_height = (tip_height + 1).expect("tip is far below Height::MAX");
+            let block_time = tip_time
+                .checked_add_signed(Duration::seconds(POST_BLOSSOM_POW_TARGET_SPACING))
+                .expect("tip plus a small constant is far below i64::MAX")
+                .timestamp();
```
teor2345: The target spacing is a goal, not a consensus rule.

The actual `min_time` consensus rule is:
> For each block other than the genesis block , nTime MUST be strictly greater than the median-time-past of that block.

This time will need to come from the state request.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -65,14 +67,14 @@ pub struct GetBlockTemplate {
     #[serde(rename = "coinbasetxn")]
     pub coinbase_txn: TransactionTemplate<amount::NegativeOrZero>,
 
-    /// Add documentation.
+    /// The expected difficulty for the new block displayed in expanded form.
     // TODO: use ExpandedDifficulty type.
     pub target: String,
 
-    /// Add documentation.
+    /// Last block time plus block spacing.
```
teor2345: The `min_time` consensus rule is:

```suggestion
    /// > For each block other than the genesis block, nTime MUST be strictly greater than
    /// > the median-time-past of that block.
    ///
    /// <https://zips.z.cash/protocol/protocol.pdf#blockheader>
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let time = chrono::Utc::now();
```
teor2345: This time is used to check if the next block will be a testnet minimum difficulty block. So it needs to be exactly the same time as we use in the `getblocktemplate` RPC response.

We can't call `chrono::Utc::now()` twice, because we might get different times.

Here is one possible design:
1. If we're on testnet, and minimum difficulty blocks are allowed by any times in the valid range, set the minimum time to the lowest time that allows minimum difficulty blocks (this avoids the difficulty changing if the miner changes the time)
2. Clamp the time using the minimum and maximum time consensus rules
3. Return the minimum time, and the time we used to calculate the difficulty, as part of the state response

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: I think we might need to move all these fields into the state request instead, so the height, hash, time, and other fields are all from the same tip block.

(Tip blocks can change between calls to the state or `LatestChainTip`.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: This default value is incorrect, we don't support this feature.

teor2345 (changes requested):
This looks good, but I think we're missing the min and max time consensus rules.

There are also some issues with concurrent block commits that need to be fixed.

zebra-state/src/service.rs
teor2345: Actually, I think I was a bit confused by how the function was called.

This should work fine, but it would be clearer if we called `read::best_tip()` in `read::difficulty::relevant_chain_difficulty()`.

We'll also need a comment explaining why it is always correct to do this, even if blocks are added concurrently. The reasons are the same as the block locator hashes:
https://github.com/ZcashFoundation/zebra/blob/6446e0ec1bd42d46c115a1dc41fa78955f4aed30/zebra-state/src/service/read/find.rs#L162-L174

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
oxarbitrage: And what do we support exactly if this is the only thing zcash haves ?

oxarbitrage:
@teor2345 you number of comments you made in your review is too much, i am still trying to figure out which of them are important ... I think only 2 of them are and the rest is optional. I want to discuss this with you as i think is not helping.

zebra-chain/src/work/difficulty.rs
oxarbitrage: What is the worst thing that can happen if we leave this as is and open a ticket for this if needed ?

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: I don't understand this, zcash has this value hardcoded i dont see why we can't do the same. What is the worst thing that can happen if we just leave it as it is now ?

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: What is the big deal with this ? Can we leave it as it is ?

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: What is the worst it can happen if we just leave this error as is ?

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
oxarbitrage: What is the worst that can happen if we just leave this as it is ?

zebra-state/src/request.rs
oxarbitrage: Why the CI does not warns about this problem instead of having more comments inside the review ?

zebra-state/src/service.rs
oxarbitrage: What is the big deal if we just leave it as it is now that you understood whats going on ?

zebra-state/src/service/read/difficulty.rs
oxarbitrage: I think this is fine, we can have a panic here but what is the big deal with this as it is now ?

zebra-state/src/service/read/difficulty.rs
oxarbitrage: You added like 5 comment about time ? Can you just take the important information and place it in a single comment ? It is hard to follow.

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: I think this is important but is lost inside a bunch of other comments.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: > The block data MUST be validated and checked against the server's usual acceptance rules (excluding the check for a valid proof-of-work).

I suspect zcashd does support this (it was mentioned during an arborist call about zip-317 that miners could always include their own transactions and I've been wondering about how that could be reliable since then.)

We should consider supporting this too but a false positive might be jarring.

I created #5684 for adding support for the `proposal` capability and #5685 for using it to test the getblocktemplate's response in `template` mode.

zebra-chain/src/work/difficulty.rs
teor2345: This makes it easier for future code to break a consensus rule and not get picked up in reviews. But it is ok for now, the code in this PR uses it correctly.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: If mining pools get this capability, they might submit blocks using `getblocktemplate` rather than `submitblock`.

We decided that we don't need a ticket to support this feature, because we don't know if any mining pools use it.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: This makes the request slightly harder to use, but it is ok for now.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: This error should never happen, so it is ok to merge it.

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
teor2345: We might miss some test coverage of the time field. But if we're covering `mintime` that should be ok.

zebra-state/src/service.rs
teor2345: If we document why it works, then we don't need to make any code changes.

```suggestion
                            // # Correctness
                            //
                            // It is ok to do these lookups using multiple database calls. Finalized state updates
                            // can only add overlapping blocks, and block hashes are unique across all chain forks.
                            // 
                            // If there is a large overlap between the non-finalized and finalized states,
                            // where the finalized tip is above the non-finalized tip,
                            // Zebra is receiving a lot of blocks, or this request has been delayed for a long time.
                            //
                            // In that case, the `getblocktemplate` RPC will return an error because Zebra
                            // is not synced to the tip. That check happens before the RPC makes this request.
                            read::tip(latest_non_finalized_state.best_chain(), &state.db)
                                .map(|tip| {
                                    read::difficulty::relevant_chain_difficulty(
                                        &latest_non_finalized_state,
                                        &state.db,
                                        tip,
                                        state.network,
                                    )
                                })
                                .unwrap_or(None);
```

zebra-state/src/service/read/difficulty.rs
teor2345: It makes the request slightly harder to use, but it is ok to merge.

zebra-state/src/request.rs
teor2345: I opened a PR to turn doc warnings into errors:
https://github.com/ZcashFoundation/zebra/pull/5683

teor2345:
> @teor2345 you number of comments you made in your review is too much, i am still trying to figure out which of them are important ... I think only 2 of them are and the rest is optional. I want to discuss this with you as i think is not helping.

Sorry about that, and thanks for the catch-up. I've resolved all the comments that can wait for later, and I made a suggestion or PR for some other comments.

zebra-state/src/service.rs
```diff
@@ -1554,6 +1551,44 @@ impl Service<ReadRequest> for ReadStateService {
...
+                // Allow other async tasks to make progress while concurrently reading blocks from disk.
+                let span = Span::current();
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        let get_block_template_info = read::tip(
+                            latest_non_finalized_state.best_chain(),
+                            &state.db,
+                        )
+                        .map(|tip| GetBlockTemplateChainInfo {
+                            tip,
```
teor2345: I'm not sure if you're going to fix this later, but I'm letting you know just in case:

The time in this tip is the `nTime` for the chain tip block, but it is being used for `min_time` in the RPC.

The `min_time` consensus rule uses `median-time-past`, which is calculated from the `nTime` fields of 11 recent blocks:
> For each block other than the genesis block , nTime MUST be strictly greater than the median-time-past of that block.

So you'll want to use `difficulty_adjustment.median_time_past()` + 1 second for the `min_time`:
https://github.com/ZcashFoundation/zebra/blob/2e50ccc8f3d067c5aa3c8924c3ccedabc1af477b/zebra-state/src/service/check/difficulty.rs#L310-L323

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -314,7 +320,44 @@ where
...
+            let tip_hash = chain_info.tip.1;
+
             let block_height = (tip_height + 1).expect("tip is far below Height::MAX");
+            let min_time = chain_info.median_time_past.checked_add_signed(Duration::seconds(1))
+                .expect("median time plus a small constant is far below i64::MAX");
+
+            // > For each block at block height 2 or greater on Mainnet, or block height 653606 or greater on Testnet, nTime
+            // > MUST be less than or equal to the median-time-past of that block plus 90 * 60 seconds.
+            //
+            // We ignore the height as we are checkpointing on nu5 in Mainnet and Testnet.
```
teor2345: The mandatory checkpoint is Canopy, and the default checkpoint is past NU5. They are both a long way above block 653606, so this is just a documentation fix.

```suggestion
            // We ignore the height as we are checkpointing on Canopy or higher in Mainnet and Testnet.
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -314,7 +320,44 @@ where
...
+                ReadResponse::ChainInfo(Some(chain_info)) => chain_info,
+                _ => unreachable!("lets hope for always some until later"),
+            };
+
+            // Get the tip data from the state call
+            let tip_height = chain_info.tip.0;
+            let tip_hash = chain_info.tip.1;
+
             let block_height = (tip_height + 1).expect("tip is far below Height::MAX");
+            let min_time = chain_info.median_time_past.checked_add_signed(Duration::seconds(1))
```
teor2345: Let's write down the consensus rules here, so we are sure we have implemented them correctly.

```suggestion
            // > For each block other than the genesis block , nTime MUST be strictly greater than
            // > the median-time-past of that block.
            // https://zips.z.cash/protocol/protocol.pdf#blockheader
            let min_time = chain_info.median_time_past.checked_add_signed(Duration::seconds(1))
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -314,7 +320,44 @@ where
...
+            //
+            // We ignore the height as we are checkpointing on nu5 in Mainnet and Testnet.
+            let cur_time = chain_info.current_system_time;
+            if cur_time > min_time.checked_add_signed(Duration::seconds(5400)).expect("median time plus a small constant is far below i64::MAX") {
+                return Err(Error {
+                    code: ErrorCode::ServerError(0),
+                    message: format!("Current time {cur_time} is bigger than {min_time} + 5400 seconds."),
+                    data: None,
+                });
+            }
```
teor2345: These checks are tricky to get right, what do you think of these changes?

In the suggestion below:
1. The `max_time` calculation seems like it is off by one - the rule is:
> less than or equal to the median-time-past of that block plus 90 * 60 seconds

But this code calculates `median_time_past + 1 + 90*60 seconds`, so any user generating blocks with the largest possible time will create invalid blocks.

2. The `cur_time` must be greater than or equal to `min_time`, but this code does not check that.

3. If we return an error when `cur_time` is greater than `max_time`, users with bad clock skew won't be able to mine blocks. And if mainnet or testnet go down for 90 minutes, all Zebra instances won't be able to mine blocks. Instead, we can clamp the time so it is in the correct range.

Other changes:

After this, the time is correct for mainnet, but testnet has an extra consensus rule that can change the difficulty based on the time:
> if the block time of a block at height height ≥ 299188 is greater than 6 · PoWTargetSpacing(height) seconds after that of the preceding block, then the block is a minimum-difficulty block. In that case its nBits field MUST be set to ToCompact(PoWLimit)

4. This check can change the `cur_time`, so we need to move this whole check into the state request, before we call `AdjustedDifficulty::new_from_header_time()`. Otherwise, we could be using a one time to calculate the difficulty, and a different time in the block header.

5. The miner can also change the time to anything between `min_time` and `max_time`. But on testnet, changing the time can change the difficulty. So if `max_time` is a minimum difficulty block, we can either:
    - remove "time" from `getblocktemplate.mutable`
    - change `min_time` and `cur_time` to the first time that allows a minimum difficulty block, and call `AdjustedDifficulty::new_from_header_time()` with the new `cur_time`, so all times the miner can use are minimum difficulty blocks

What do you think we should do here?

I don't really mind either way, but changing `min_time` and `cur_time` might make testing faster, because some blocks will be easier. 

```suggestion
            const MAX_BLOCK_TIME_GAP: i64 = 90 * 60;
            let max_time = chain_info.median_time_past.checked_add_signed(Duration::seconds(MAX_BLOCK_TIME_GAP)).expect("median time plus a small constant is far below i64::MAX");
            let cur_time = chain_info.current_system_time.timestamp().clamp(min_time.timestamp(), max_time.timestamp());
       }
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -361,11 +412,13 @@ where
 
                 size_limit: MAX_BLOCK_BYTES,
 
-                cur_time: 0,
+                cur_time: cur_time.timestamp(),
```
teor2345: I changed the type of `cur_time` in my other suggestion.

```suggestion
                cur_time,
```

teor2345 (commented):
I think we're almost there, just need to deal with some tricky edge-cases with time and difficulty on testnet.

zebra-state/src/service.rs
```diff
@@ -1554,6 +1550,51 @@ impl Service<ReadRequest> for ReadStateService {
...
+                let state = self.clone();
+                let latest_non_finalized_state = self.latest_non_finalized_state();
+
+                // # Performance
+                //
+                // Allow other async tasks to make progress while concurrently reading blocks from disk.
+                let span = Span::current();
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        let get_block_template_info =
```
teor2345, line 1576: Can we add a comment here so we don't forget and try to fix it later?

```suggestion
                        // # Correctness
                        //
                        // It is ok to do these lookups using multiple database calls. Finalized state updates
                        // can only add overlapping blocks, and block hashes are unique across all chain forks.
                        // 
                        // If there is a large overlap between the non-finalized and finalized states,
                        // where the finalized tip is above the non-finalized tip,
                        // Zebra is receiving a lot of blocks, or this request has been delayed for a long time.
                        //
                        // In that case, the `getblocktemplate` RPC will return an error because Zebra
                        // is not synced to the tip. That check happens before the RPC makes this request.
                        let get_block_template_info =
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,105 @@
...
+        *current_system_time,
+        tip_height,
+        network,
+        relevant_data,
+    );
+
+    let compact_difficulty = match network {
+        Network::Testnet => {
+            // > if the block time of a block at height height ≥ 299188 is greater than 6 * PoWTargetSpacing(height) seconds after
+            // > that of the preceding block, then the block is a minimum-difficulty block.
```
oxarbitrage: The spec uses a bit of confusing words here for me:

`seconds after that of the preceding block`. I am assuming that for us this means seconds after our current tip time.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,105 @@
...
+        .map(|block| {
+            (
+                block.borrow().header.difficulty_threshold,
+                block.borrow().header.time,
+            )
+        })
+        .peekable();
+
+    // get the tip time from the relevant chain
+    let tip_time = relevant_data.peek().expect("tip should be here").1;
```
oxarbitrage: I am not totally sure that this is the time we need to do the check.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,105 @@
...
+            if current_system_time
+                > &tip_time
+                    .checked_add_signed(pow_target_spacing_mul)
+                    .expect("tip time plus a small constant is far below i64::MAX")
+            {
+                // We are in a minimum-difficulty block.
+                ExpandedDifficulty::target_difficulty_limit(Network::Testnet).to_compact()
+            } else {
+                difficulty_adjustment.expected_difficulty_threshold()
+            }
```
oxarbitrage: not totally sure if this is the correct logic, feel free to push changes or suggestions.

oxarbitrage (commented):
Some comments that i am not totally sure for the testnet consensus rule.

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: Lets follow up in new comments.

oxarbitrage: > The miner can also change the time to anything between min_time and max_time. But on testnet, changing the time can change the difficulty. So if max_time is a minimum difficulty block, we can either:
>
> - remove "time" from getblocktemplate.mutable
> - change min_time and cur_time to the first time that allows a minimum difficulty block, and call AdjustedDifficulty::new_from_header_time() with the new cur_time, so all times the miner can use are minimum difficulty blocks
>
> What do you think we should do here?
>
> I don't really mind either way, but changing min_time and cur_time might make testing faster, because some blocks will be easier.

This is still missing.

oxarbitrage:
By now, we don't have any tests for testnet specific code under minimum difficulty block mining.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -325,13 +368,21 @@ where
             // Convert into TransactionTemplates
             let mempool_txs = mempool_txs.iter().map(Into::into).collect();
 
-            let empty_string = String::from("");
+            let mut mutable: Vec<String> = constants::GET_BLOCK_TEMPLATE_MUTABLE_FIELD.iter().map(ToString::to_string).collect();
+
+            // If we are in the testnet and received expected difficulty equals to the target difficulty limit for the Testnet then we
+            // are mining a minimum difficulty block.
```
oxarbitrage: I am not sure if this is enough to apply the change.

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: > This is still missing.

I added some code for the remove time option in https://github.com/ZcashFoundation/zebra/pull/5659/commits/3a862aa942e24f0f28024f307358c7f32f6b128d

zebra-state/src/service/read/difficulty.rs
teor2345: That's correct, this is implemented by `NetworkUpgrade::is_testnet_min_difficulty_block()`, which is called by `AdjustedDifficulty::expected_difficulty_threshold()`.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+        Network::Mainnet => difficulty_adjustment.expected_difficulty_threshold(),
+    };
```
teor2345: This rule is already implemented by `AdjustedDifficulty::expected_difficulty_threshold()`:
```suggestion
    let compact_difficulty = difficulty_adjustment.expected_difficulty_threshold();
```

But it depends on the `cur_time`, which we must clamp before we send it to `AdjustedDifficulty::new_from_header_time()`.

zebra-state/src/service/read/difficulty.rs
teor2345: We need to use the `cur_time` we are going to put in the block header to calculate the difficulty we are going to put in the block header.

So all the time code needs to be moved from `ReadStateService::call()` and `get_block_template()` to here.

Something like this, but the types and imports are not quite right:
```rust
    // The getblocktemplate RPC returns an error if Zebra is not synced to the tip.
    // So this will never happen in production code.
    assert!(relevant_data.len() < MAX_CONTEXT_BLOCKS);

    let current_system_time = chrono::Utc::now();
                                        
    let difficulty_adjustment = AdjustedDifficulty::new_from_header_time(
        *current_system_time,
        tip_height,
        network,
        relevant_data,
    );

    // > For each block other than the genesis block , nTime MUST be strictly greater than
    // > the median-time-past of that block.
    // https://zips.z.cash/protocol/protocol.pdf#blockheader
    let min_time = difficulty_adjustment.median_time_past().checked_add_signed(Duration::seconds(1))
        .expect("median time plus a small constant is far below i64::MAX");

    // > For each block at block height 2 or greater on Mainnet, or block height 653606 or greater on Testnet, nTime
    // > MUST be less than or equal to the median-time-past of that block plus 90 * 60 seconds.
    //
    // We ignore the height as we are checkpointing on Canopy or higher in Mainnet and Testnet.
    let max_time = difficulty_adjustment.median_time_past().checked_add_signed(Duration::seconds(BLOCK_MAX_TIME_SINCE_MEDIAN))
        .expect("median time plus a small constant is far below i64::MAX");

    let current_system_time = current_system_time.timestamp().clamp(min_time.timestamp(), max_time.timestamp());

    let difficulty_adjustment = AdjustedDifficulty::new_from_header_time(
        *current_system_time,
        tip_height,
        network,
        relevant_data,
    );
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    (compact_difficulty, difficulty_adjustment.median_time_past())
```
teor2345: Now we have modified the time, we need to return that modified time in the state response:
```suggestion
    (compact_difficulty, current_system_time)
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: It's possible that `cur_time` is before the 15 minute limit, but `max_time` is after it.

So it's almost right, but we need to check if the difficulty for `max_time` is the minimum difficulty. (`chain_info.expected_difficulty` is the difficulty for `cur_time`.)

Since we're moving all the time stuff to the state anyway, we can do a difficulty check on `max_time` there?

teor2345 (commented):
Thanks for the changes, I think we're almost there.

I'm happy to help write these changes, but I need to finish the ZIP-317 work first. So feel free to keep making changes here, or ask me to submit a PR on top of this one.

teor2345:
> By now, we don't have any tests for testnet specific code under minimum difficulty block mining.

This existing test might help?
https://github.com/ZcashFoundation/zebra/blob/c812f880cf8e63d518c88908210764fd21a0cc62/zebra-chain/src/work/difficulty/tests/vectors.rs#L393-L396

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Did you want me to make this change?

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -75,17 +81,28 @@ pub async fn test_responses<State, ReadState>(
         miner_address: Some(transparent::Address::from_script_hash(network, [0xad; 20])),
     };
 
+    // nu5 block height
+    let fake_tip_height = NetworkUpgrade::Nu5.activation_height(network).unwrap();
+    // nu5 block hash
+    let fake_tip_hash =
+        Hash::from_hex("0000000000d723156d9b65ffcf4984da7a19675ed7e2f06d9e5d5188af087bf8").unwrap();
+    // nu5 block time
+    let fake_tip_time = Utc.timestamp_opt(1654008605, 0).unwrap();
```
teor2345: Optional: Can we make the test data obey the consensus rule that the `cur_time` must be equal or greater than the `min_time`?

It doesn't really matter for this test, but it makes reading the snapshot a bit confusing.

```rust
    // nu5 block time + 1
    let fake_tip_time = Utc.timestamp_opt(1654008606, 0).unwrap();
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -0,0 +1,115 @@
...
+    let current_system_time = DateTime::<Utc>::from_utc(
+        NaiveDateTime::from_timestamp_opt(
+            current_system_time
+                .timestamp()
+                .clamp(min_time.timestamp(), max_time.timestamp()),
+            0,
+        )
+        .expect("We should have an in range timestamp"),
+        Utc,
+    );
```
teor2345: Optional: I think this can be a bit simpler:

```suggestion
    let current_system_time = DateTime::timestamp_opt(
        current_system_time
            .timestamp()
            .clamp(min_time.timestamp(), max_time.timestamp()),
        0,
    )
    .expect("We should have an in range timestamp");
```

teor2345 (commented):
Thanks, I think we're almost there, let me know if you want me to do the minimum difficulty "time" change.

oxarbitrage:
Please apply any of the missing stuff and merge if you can. Thank you.

zebra-state/src/service/read/difficulty.rs
teor2345: Fixed by a22e3b5

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
teor2345: Fixed by eef133c

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Fixed by a22e3b5

teor2345 (approved):
Let's see if that works

PR #5671, opened 2022-11-19
fix(ssh): add a fixed SSH key to use with `gcloud`

### LINKED ISSUE #5645: ci: ssh error after implementing OS-Login: Permission denied (publickey)
## Motivation

After merging #5602 and #5617 we're still dealing with SSH issues

### Designs

- Use a more suitable SSH mechanism for automated scripts, like using a fixed SSH key (as this might be our best workaround so far)
- Validate if we can use plain SSH with OS-login with this action https://github.com/marketplace/actions/ssh-remote-commands

gustavovalverde:
## Motivation

After implementing #5602 most SSH errors were reduced to a single one, but this single error was being more constant than previous ones, as this error was raised on every other workflow job

Fixes: #5645

### Designs

- Disable OS Login
- Use a fixed SSH key
- Add a public key to the metadata of the project

## Solution

- As there's no easy way to convert a secret to a file, we're using https://github.com/shimataro/ssh-key-action to accomplish this
- `gcloud compute ssh` also expects the public key to be available locally, even though it has been added to the project metadata, so we're creating a public key on the fly (from the private key)

## Review

### Reviewer Checklist


.github/workflows/deploy-gcp-tests.yml
```diff
@@ -566,7 +605,6 @@ jobs:
       - name: Show logs for ${{ inputs.test_id }} test (sprout)
         run: |
           gcloud compute ssh ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
-          --ssh-key-expire-after=30s \
```
teor2345, line 569: This job does not install the key, but it looks like it should.

About half the jobs in this file are missing an install step.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1362,15 +1434,3 @@ jobs:
           else
             gcloud compute instances delete "${INSTANCE}" --zone "${{ env.ZONE }}" --delete-disks all --quiet
           fi
-
-      # Deletes SSH keys generated during this workflow run, as GCP has a limit of SSH keys
-      # that can exist at the same time in the OS Login metadata. Not deleting this keys
-      # could cause the following error:
-      # `Login profile size exceeds 32 KiB. Delete profile values to make additional space`
-      - name: Delete temporal SSH keys
```
teor2345, line 1370: Will outdated PRs continue to create these keys until they are all rebased on this PR? 
How will we do a final cleanup of these unused keys?

Do we need to keep on running this step until the last PR has been updated?

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: Solved with [fix: add missing SSH key installation steps](https://github.com/ZcashFoundation/zebra/pull/5671/commits/81c74a292f6f936f4ec574bf40688aca9dae4be0)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: This were specific keys created by the OS-Login function which was added project-wide on GCP. This is no longer happening (as it was disabled), and all unused keys have been manually deleted.

We anyways need to rebase all PRs to stop SSH errors from happening

arya2 (approved):
Let's try it!

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -125,6 +125,17 @@ jobs:
...
+      # Install our SSH secret
+      - name: Install private SSH key
+        uses: shimataro/ssh-key-action@v2.4.0
+        with:
+          key: ${{ secrets.GCP_SSH_PRIVATE_KEY }}
+          name: google_compute_engine
+          known_hosts: unnecessary
+
+      - name: Generate public SSH key
+        run: ssh-keygen -y -f ~/.ssh/google_compute_engine > ~/.ssh/google_compute_engine.pub
```
dconnolly, line 137: Oh, so this is generated each time and pushed? OK

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -162,7 +173,6 @@ jobs:
       - name: Create ${{ inputs.test_id }} Docker volume
         run: |
           gcloud compute ssh ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
-          --ssh-key-expire-after=30s \
```
dconnolly, line 165: 👍

dconnolly (commented):
lgtm, worth a shot, ephemeral ssh keys are even better

Issue #2908, opened 2021-10-19
Improve VerifyBlockError::Commit typing, so we don't accidentally break syncer error handling


conradoplg:
## Motivation

`VerifyBlockError::Commit` is a wrapper over a `BoxError`. However, in https://github.com/ZcashFoundation/zebra/pull/2890/ a check was added in the `should_restart_sync` function to ignore `Commit` errors corresponding to `"block is already committed to the state"` when deciding whether to reset the sync procedure. That check is fragile since it will break if the string changes.

Change `Commit` to wrap a specific error type enumeration, and create a specific item for that particular error to be used for matching and filtering it in `should_restart_sync`.

Currently, this check is implemented by the `is_duplicate_request()` method, which can't use `Commit` because it is a `BoxError`.

There's also a `BlockDownloadVerifyError::Invalid` error that comes from the chain verifier and a `BlockDownloadVerifyError::DownloadFailed` from the network service. 

We should downcast all `BoxError`s in the syncer's `BlockDownloadVerifyError` to concrete types and add them to the match statement in `should_restart_sync()`:

https://github.com/ZcashFoundation/zebra/blob/0152c8641bc41767a5ef7a67a8e01e84a26eb0a7/zebrad/src/components/sync.rs#L1088

https://github.com/ZcashFoundation/zebra/blob/0152c8641bc41767a5ef7a67a8e01e84a26eb0a7/zebrad/src/components/sync.rs#L1155-L1163

### Designs

- See #5682
- See #5757
    - we don't need to reorder fields manually to save space, because Rust 1.67 will do this automatically: https://github.com/rust-lang/rust/pull/102750
    - we can still turn unnamed tuple fields into named fields, because that makes debugging easier

## Related Work

- #1186

Follow up to https://github.com/ZcashFoundation/zebra/pull/2890/

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/2908?planning-poker) with ZenHub @conradoplg @dconnolly @jvff @oxarbitrage @teor2345 @upbqdn

teor2345:
We're de-prioritizing network changes, and network-related changes.

teor2345:
This is needed to do #5487

teor2345:
I added this as a TODO in the source code, so we can do it if we are rewriting that code for other reasons.

[...]

PR #5682, opened 2022-11-21
Improves block commit error typing (WIP)

arya2:
## Motivation

This is a rough draft and there are a lot of fixes to be made if it's to be pursued.

I had started on this when there wasn't much else in need of urgent attention. 

Briefly opening this up for comments on whether this issue is something that could be addressed in the near-term and if this PR is in the right direction.

This is likely very low priority (though I'd be happy to be convinced otherwise) and unlikely to be merged. 

Related to #2908.

## Review

This PR is a rough draft.

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -29,6 +29,7 @@ use zebra_chain::{
     work::equihash,
 };
 use zebra_state as zs;
+use zs::CommitBlockError;
 
 use crate::{error::*, transaction as tx, BoxError};
 
@@ -71,7 +72,7 @@ pub enum VerifyBlockError {
     Time(zebra_chain::block::BlockTimeError),
 
     #[error("unable to commit block after semantic verification")]
-    Commit(#[source] BoxError),
+    Commit(#[source] CommitBlockError),
 
     #[error("invalid transaction")]
     Transaction(#[from] TransactionError),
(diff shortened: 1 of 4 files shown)


zebra-consensus/src/block.rs
```diff
@@ -271,10 +272,10 @@ where
             match state_service
                 .ready()
                 .await
-                .map_err(VerifyBlockError::Commit)?
+                .expect("poll_ready method of state service and Ready::<StateService>::poll will never return Err")
```
teor2345, line 275: For your information - no changes needed:

This change locks us into always having the state service ready.

That might be ok, because our design requires the service is either ready: `Poll::Ready(Ok(_))`, or permanently failed: `Poll::Ready(Err(_))`. The state service can never be temporarily paused: `Poll::Pending`.

The state needs to commit blocks in order. So we want to pause higher blocks, but still let lower blocks commit. But Tower services can only pause all requests,  so we can't use `poll_ready()` to do that.

So as long as we're sure we'll never error here, even on shutdown, it would be ok to do this.

teor2345 (commented):
This looks like a good refactor. Just wondering if it is needed for the getblocktemplate work?

We are currently in an audit freeze for all other Rust changes, so we can make any audit changes quickly and easily.

We could add the branch to the ticket, and open a PR when we are out of the freeze?

zebra-consensus/src/block.rs
arya2: Got it! I was very uncertain about this part, that's good to know.

arya2:
> Just wondering if it is needed for the getblocktemplate work?

It's not needed (unless we find out #5487 is required), I was just curious.

Thank you for taking a look.

> We could add the branch to the ticket, and open a PR when we are out of the freeze?

This sounds good :)

zebra-consensus/src/block.rs
teor2345: I am also unsure about this!

PR #5670, opened 2022-11-18
fix(deploy): no dot name instance template disks to allow scaling

gustavovalverde:
## Motivation

We've been having the following errors when deploying:
- ```Instance template specifies a disk with a custom name. This will cause instance group not to scale beyond 1 instance per zone.```
- ```Error: Failed to start container: Volume pd-0: Could not determine if the GCE Persistent Disk zebrad-cache-b722e22 is attached read-only or read-write```

Fixes #4946
Fixes #5644

### Specifications

See: https://cloud.google.com/sdk/gcloud/reference/compute/instances/create-with-container#--create-disk:~:text=This%20option%20cannot%20be%20specified%20if%20more%20than%20one%20instance%20is%20being%20created

See: https://cloud.google.com/sdk/gcloud/reference/compute/instances/create-with-container#--create-disk:~:text=specifying%20disk%20name%20here%20is%20optional

## Solution

- Apply documentation recommendations to remove this errors and allow CD instances to increase as expected

## Review

Anyone can review this PR

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

PR #5697, opened 2022-11-22
fix(cd): allow deploying instance templates without disk errors

gustavovalverde:
## Motivation:

PR #5670 failed in `main` as it was tested with `gcloud compute instances create-with-container` and even the manual deployment uses `instances`, and it works.

But the one that failed uses `gcloud compute instance-templates create-with-container` using `instance-template` and it's complaining with: `When attaching or creating a disk that is also being mounted to a container, must specify the disk name`

Based on the documentation, the name is optional when using `create-with-container`, for both `instances` and `instance-templates`

Source: https://cloud.google.com/sdk/gcloud/reference/compute/instance-templates/create-with-container#--container-mount-disk

This partially reverts: #5670

## Solution:

Revert this specific job as how it was, and do not scale the instances above 1, as this would cause the following error:
`Instance template specifies a disk with a custom name. This will cause instance group not to scale beyond 1 instance per zone.`

## Review

@teor2345 reviewed the previous one

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-delivery.yml
+++ b/.github/workflows/continous-delivery.yml
@@ -116,6 +116,8 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v1.0.1
 
+      # TODO we should implement the fixes from https://github.com/ZcashFoundation/zebra/pull/5670 here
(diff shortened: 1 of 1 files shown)


dconnolly (approved):
"oy, google cloud docs!" 🫠

Issue #5473, opened 2022-10-25
Select `getblocktemplate` transactions according to ZIP-317


teor2345:
## Motivation

We want to select transactions based on ZIP-317.

### Specifications

The block production parts of:
- https://zips.z.cash/zip-0317#block-production
- https://hackmd.io/rgb38M_wSxa9kotyaON1fA (or any actual ZIP updates)

The outdated ZIP-313 might also have some relevant information:
https://zips.z.cash/zip-0313

### Design

When selecting transactions, we should do something similar to `zcashd`, based on ZIPs 317 and 313.

The ZIP editors have recommended including low-fee transactions in blocks with probability `fee / conventional_fee`. Zebra could implement this in the mempool, but that changes relaying as well, which we don't want to do yet. So using the RPC is best for now. The details are in https://hackmd.io/rgb38M_wSxa9kotyaON1fA

We should also check the `sigop_limit` before adding each new transaction to the block.

## Related Tickets

Other parts of transaction selection will be handled by the mempool changes in #5335.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5473?planning-poker) with Zenhub @arya2 @dconnolly @oxarbitrage @teor2345

Issue #5336, opened 2022-10-04
Relay and accept mempool transactions based on the ZIP-317 fee rules


teor2345:
### Motivation

We want to deploy the new relaying rules in ZIP-317, once enough wallets have updated:
> Node developers SHOULD coordinate on schedules for deploying restrictions to their policies for transaction mempool acceptance and peer-to-peer relaying. These policy changes SHOULD NOT be deployed before the changes to transaction inclusion policy by miners described in the preceding paragraph.

https://zips.z.cash/zip-0317#deployment

This should be released in Zebra around the time of the `zcashd` release that implements ZIP-317, which is release 5.5.0 in April 2023.

We're also adding a minimum fee rate to match `zcashd`.

#### ZIP 317

The relay parts of:
- https://zips.z.cash/zip-0317#transaction-relaying

#### Minimum Fee Rate

`zcashd` also enforces a standard rule:
- Transactions must pay a fee of at least 100 zatoshis per 1000 bytes of serialized `size`, with a maximum fee of 1000 zatoshis
- In `zcashd` this is DEFAULT_MIN_RELAY_TX_FEE

In Zebra, we can directly enforce this minimum fee rate during mempool transaction validation.

##### zcashd details

We don't need to implement exactly what `zcashd` does.

See Daira's summary here:
https://github.com/ZcashFoundation/zebra/issues/5336#issuecomment-1506748801

Their code is much more complicated:
https://github.com/zcash/zcash/blob/adfc7218435faa1c8985a727f997a795dcffa0c7/src/main.h#L71-L72
https://github.com/zcash/zcash/blob/6ebf01fa831c02c18d6c2cf76c15b0cd2a1fe978/src/main.cpp#L1967-L1971

### Design

Zebra currently relays all valid transactions.

We should try to match what `zcashd` does for their ZIP 317 implementation:
- transactions below the ZIP-317 fee per action are not relayed or stored in the mempool

The simplest way to do this in Zebra is to add the ZIP-317 check to mempool transaction validation as a standard rule. Then they won't be relayed, because they won't be in the mempool.

### Completed

<details>

#### ZIP 401 changes

As part of this ticket, we updated some of the ZIP-401 constants:
https://github.com/zcash/zips/issues/565#issuecomment-1489398544

This was done in
- #6521

</details>

### Outdated

Here is the previous behaviour:

<details>

Daira says that:
[...]

PR #5703, opened 2022-11-23
change(mempool): Evict transactions from the mempool using the ZIP-317 conventional fee

### LINKED ISSUE #5335: Evict transactions from the mempool based on the ZIP-317 fee rules
### Motivation

We can deploy the new mempool rules in ZIP-317 before wallets have updated to the new conventional fee, because dropping transactions is randomised.

This can be changed in Zebra at any time. It must not be behind the `getblocktemplate-rpcs` feature, because it is required as soon as possible to mitigate denial of service:
https://zips.z.cash/zip-0317#deployment

### Specification

The conventional fee part of:
https://zips.z.cash/zip-0317

teor2345:
## Motivation

We want to implement the ZIP-317 conventional fee calculation, and use it to evict transactions from the mempool.

This is a security fix independent of the getblocktemplate work.
(But we also use the conventional fee calculation when selecting transactions for new blocks.)

Close #5335.

### Specifications

https://zips.z.cash/zip-0317#fee-calculation
https://zips.z.cash/zip-0317#mempool-size-limiting

This change can be deployed immediately:
https://zips.z.cash/zip-0317#deployment

### Designs

Do the fee calculation in the specified order, using similar names for everything:
https://zips.z.cash/zip-0317#fee-calculation

Cache the result as a field on `UnminedTx`, because it only changes when the transaction data changes. This way, it is available during block construction and transaction relaying.

This is a standard rule, not a consensus rule, so variations in the implementation are acceptable.

## Solution

- Add a ZIP-317 conventional fee module
- Add ZIP-317 constants
- Calculate the ZIP-317 conventional fee
- Use the new conventional fee for mempool size limiting
-  Update existing tests

Related changes:
- Just return a usize from `zcash_serialized_size()`, removing the unused `Result<_, io::Error>`
  - This makes the new code a lot simpler

## Review

I'd like two people to review this code, including @dconnolly because she helped write the spec.

Here is the corresponding code in `zcash_primitives` (it doesn't have Orchard support yet):
https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/transaction/fees/zip317.rs#L117

Do we need any extra tests?

### Reviewer Checklist

## Follow Up Work

The changes to transaction relaying and block production depend on enough wallets updating to pay these fees:
https://zips.z.cash/zip-0317#deployment

So they have separate tickets:
- #5473 
- #5336

### DIFF
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -356,7 +356,7 @@ impl VerifiedUnminedTx {
     /// > a fee less than the conventional fee, otherwise 0.
     ///
     /// > zcashd and zebrad limit the size of the mempool as described in [ZIP-401].
-    /// > This specifies a that is added to the "eviction weight" if the transaction
+    /// > This specifies a low fee penalty that is added to the "eviction weight" if the transaction
     /// > pays a fee less than the conventional transaction fee. This threshold is
     /// > modified to use the new conventional fee formula.
     ///
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -52,6 +52,12 @@ mod zip317;
 /// [ZIP-401]: https://zips.z.cash/zip-0401
 const MEMPOOL_TRANSACTION_COST_THRESHOLD: u64 = 4000;
 
+/// When a transaction pays a fee less than the conventional fee,
+/// this low fee penatly is added to its cost for mempool eviction.
+///
+/// See [VerifiedUnminedTx::eviction_weight()] for details.
+const MEMPOOL_TRANSACTION_LOW_FEE_PENALTY: u64 = 16_000;
+
 /// A unique identifier for an unmined transaction, regardless of version.
 ///
 /// "The transaction ID of a version 4 or earlier transaction is the SHA-256d hash
@@ -203,6 +209,11 @@ pub struct UnminedTx {
 
     /// The size in bytes of the serialized transaction data
     pub size: usize,
+
+    /// The conventional fee for this transaction, as defined by [ZIP-317].
+    ///
+    /// [ZIP-317]: https://zips.z.cash/zip-0317#fee-calculation
+    pub conventional_fee: Amount<NonNegative>,
 }
 
 impl fmt::Display for UnminedTx {
@@ -210,6 +221,7 @@ impl fmt::Display for UnminedTx {
         f.debug_struct("UnminedTx")
             .field("transaction", &self.transaction)
             .field("serialized_size", &self.size)
+            .field("conventional_fee", &self.conventional_fee)
             .finish()
     }
 }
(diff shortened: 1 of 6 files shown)


zebra-chain/src/transaction/unmined.rs
```diff
@@ -50,6 +52,12 @@ use UnminedTxId::*;
 /// [ZIP-401]: https://zips.z.cash/zip-0401
 const MEMPOOL_TRANSACTION_COST_THRESHOLD: u64 = 4000;
 
+/// When a transaction pays a fee less than the conventional fee,
+/// this low fee penatly is added to its cost for mempool eviction.
```
daira:
```suggestion
/// this low fee penalty is added to its cost for mempool eviction.
```

zebra-chain/src/transaction/unmined.rs
```diff
@@ -327,31 +341,34 @@ impl VerifiedUnminedTx {
...
+    /// Standard rule:
     ///
     /// > Each transaction also has an eviction weight, which is cost +
     /// > low_fee_penalty, where low_fee_penalty is 16000 if the transaction pays
-    /// > a fee less than the conventional fee, otherwise 0. The conventional fee
-    /// > is currently defined as 1000 zatoshis
+    /// > a fee less than the conventional fee, otherwise 0.
     ///
+    /// > zcashd and zebrad limit the size of the mempool as described in [ZIP-401].
+    /// > This specifies a that is added to the "eviction weight" if the transaction
```
arya2:
```suggestion
    /// > This specifies a low fee penalty that is added to the "eviction weight" if the transaction
```

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -28,5 +31,59 @@ const P2PKH_STANDARD_OUTPUT_SIZE: usize = 34;
...
+
+    let n_join_split = transaction.joinsplit_count();
+    let n_spends_sapling = transaction.sapling_spends_per_anchor().count();
+    let n_outputs_sapling = transaction.sapling_outputs().count();
+    let n_actions_orchard = transaction.orchard_actions().count();
+
+    let tx_in_logical_actions = div_ceil(tx_in_total_size, P2PKH_STANDARD_INPUT_SIZE);
+    let tx_out_logical_actions = div_ceil(tx_out_total_size, P2PKH_STANDARD_OUTPUT_SIZE);
+
+    let logical_actions = max(tx_in_logical_actions, tx_out_logical_actions)
```
arya2, line 62: Optional: move logical actions calculation to an `impl Transaction { fn count_logical_actions(&self) }` next to this fn

arya2 (commented):
Code looks clean and correct, just needs minor updates to a couple doc comments.

I like that `zcash_serialized_size` was updated to return a `usize`.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: I'd prefer to just have `conventional_fee` as a field, there's no other uses for logical actions yet, and the serialization of transaction data can be slightly expensive due to cryptographic operations when serializing types.

zebra-rpc/src/methods/tests/vectors.rs
```diff
@@ -288,6 +289,7 @@ async fn rpc_getrawtransaction() {
                         id: UnminedTxId::Legacy(tx.hash()),
                         transaction: tx.clone(),
                         size: 0,
+                        conventional_fee: Amount::zero(),
```
arya2, line 292: I think existing tests are sufficient but I'll leaving this here to give @dconnolly a chance to disagree.

arya2 (approved):
Thank you!

PR #4941, opened 2022-08-24
ci(test): run build and test jobs on cargo and clippy config changes

gustavovalverde:
## Previous behavior:
If warnings or error are added in `.cargo/config.toml` or `clippy.toml`,
and those could generate CI failures, we wouldn't catch those new as the
pipelines are not run when this files are changed

## Expected behavior:
If warnings or error are added in `.cargo/config.toml` or `clippy.toml`,
run all the builds and test jobs which also track a `Cargo.toml`.

## Applied solution:
Add `.cargo/config.toml` and `clippy.toml` as paths to all the required
jobs which needs to be triggered when these files changes.

Fixes https://github.com/ZcashFoundation/zebra/issues/4652

## Review

Anyone can review this PR

PR #5705, opened 2022-11-23
fix(ci): run integration tests on `main` if configuration files are changed

mpguerra:
## Motivation

We are running the CI Docker workflow on PRs that change configuration files since PR #4941. However, we never added this to run on pushes to main. Should we have?

## Review

@gustavovalverde 

### Reviewer Checklist

## Follow Up Work

Are there other workflows that we missed?

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -73,6 +73,9 @@ on:
       # dependencies
       - '**/Cargo.toml'
       - '**/Cargo.lock'
+      # configuration files
+      - '.cargo/config.toml'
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
I'm approving this for consistency, as this is how we're doing it in other workflows. But as this tests are being run in the PR, and on Mergify (if the PR is outdated or being merged in a queue), this just ends up being redundant as you've mention before @mpguerra 

But we need to segment the workflow for this to happen. In the meanwhile this consistency is needed.

gustavovalverde:
I'm rerunning a failed job, and created a ticket for it 
- #5706

Issue #5425, opened 2022-10-19
Parallelise note commitment tree point conversions


teor2345:
## Motivation

We can speed up note commitment tree updates by running point conversions in parallel with tree updates.

### Designs

Run the following operations in parallel rayon iterators or threads:
- each point conversion, or batches of point conversions, in any order, in multiple threads (`cm_x.into()`)
- appending notes, in blockchain order, in a single thread (`self.inner.append(...)`)

We can start appending notes as soon as the first point conversion is completed.

This change can be made for Sprout, Sapling, and Orchard. Sprout is a lower priority because it's not currently being used much.

Orchard example:
https://github.com/ZcashFoundation/zebra/blob/267db7f956c741c5f4369722eb9aaa0c49c359ac/zebra-chain/src/orchard/tree.rs#L309-L310

`zcashd` example of out of order note insertion using `shardtree`:
https://github.com/zcash/incrementalmerkletree/pull/50

teor2345:
This is currently optional. We've talked about it a few times, so I wanted to write it down so we had a record.

But I think we can close it, unless we decide we need it to fix #5420, and that #5420 is an important enough fix to schedule soon.

Edit: or we decide we want to fix #5709, which we need to fix before this ticket will help with sync speed.

PR #5708, opened 2022-11-23
fix(ci): Allow full sync tests to take 42 hours

teor2345:
## Motivation

Full sync tests are taking slightly longer than expected.

We could fix bugs around block validation timeouts to speed this up, but they aren't a priority at the moment.

## Solution

Increase the full sync timeout to take 42 hours

## Review

This is urgent to fix full syncs on `main`.

### Reviewer Checklist

## Follow Up Work

- #5709

Make block commits more efficient:
- #5425

### DIFF
--- a/zebrad/tests/common/sync.rs
+++ b/zebrad/tests/common/sync.rs
@@ -74,7 +74,7 @@ pub const FINISH_PARTIAL_SYNC_TIMEOUT: Duration = Duration::from_secs(11 * 60 * 
 
 /// The maximum time to wait for Zebrad to synchronize up to the chain tip starting from the
 /// genesis block.
-pub const FINISH_FULL_SYNC_TIMEOUT: Duration = Duration::from_secs(36 * 60 * 60);
+pub const FINISH_FULL_SYNC_TIMEOUT: Duration = Duration::from_secs(42 * 60 * 60);
 
 /// The test sync height where we switch to using the default lookahead limit.
 ///


gustavovalverde (approved):

arya2 (approved):

zebrad/tests/common/sync.rs
```diff
@@ -74,7 +74,7 @@ pub const FINISH_PARTIAL_SYNC_TIMEOUT: Duration = Duration::from_secs(11 * 60 *
```
arya2, line 74: Do we want to increase `FINISH_PARTIAL_SYNC_TIMEOUT` too?

Edit: there's no reason to at the moment.

arya2:
The failure this is for:

> Error: stdout of command did not log any matches for the given regex,
> within the 1day 12h command timeout
> 
> Location:
>     /opt/zebrad/zebra-test/src/command.rs:856:22
> thread 'full_sync_mainnet' panicked at 'assertion failed: `(left == right)`
>   left: `1`,
>  right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/897e37553bba8b42751c67658967889d11ecd120/library/test/src/lib.rs:184:5

https://github.com/ZcashFoundation/zebra/actions/runs/3510523424/jobs/5908934707#step:6:12930

zebrad/tests/common/sync.rs
teor2345: The partial sync uses cached state, and it is not timing out or failing, so it should be fine.

Issue #4831, opened 2022-07-26
Revert deserializing state transactions in `rayon` threads


teor2345:
## Motivation

In PR #4828, we had to increase the `lightwalletd` full sync timeout to 2.5 hours.

This might be a Zebra bug, it looks like all the blocks are slow, not just the recent large blocks.
We could have introduced this bug in PR #4805.

### Things to try

Remove the `rayon` thread from transaction deserialization:
https://github.com/ZcashFoundation/zebra/blob/9b2185ad3d7417bf4bb5a570ad81c6a027aab3b7/zebra-state/src/service/finalized_state/disk_format/block.rs#L236-L246

Check if any of the `ReadRequest::Block` calls are slow enough to turn up in Zebra's logs in the lightwalletd full sync test.
If they are fast enough, we're done!

Otherwise, here are a few more things we could try...

<details>

If they are slow, try getting all the raw transaction data from the disk, then deserializing it all at once in a single `rayon` thread:
https://github.com/ZcashFoundation/zebra/blob/9b2185ad3d7417bf4bb5a570ad81c6a027aab3b7/zebra-state/src/service/finalized_state/zebra_db/block.rs#L127-L135

Check the logs for slow reads again.

Some other things we could try:
- just spawn threads to deserialize shielded transactions
    - how do we know before they are deserialized?
    - we know the serialized length, do we just want to do it for all large transactions?
- skip cryptographic validation when deserializing trusted transactions from the state
- try multiple RPC threads (see PR #4806 for details)
- just ignore the speed issues for now

</details>

teor2345:
I did the minimal version of this change, which was a simple revert of the slower code change.

PR #5716, opened 2022-11-24
change(mempool): Contextually validates mempool transactions in best chain

### LINKED ISSUE #5376: Do contextual validation on mempool transactions using the state best tip

arya2:
## Motivation

Mempool transactions collected into a block template need to be contextually valid in the best chain to produce a block template that's contextually valid on the best chain.

Closes #5376 

## Solution

- Add `check::nullifiers::no_duplicates_in_chain` that searches the non-finalized state as well as the finalized state for duplicate nullifiers
- Update `check::anchors` fns and errors accommodate transactions without a `PreparedBlock`
- Add new state request to check that the anchors and nullifiers of a transaction are valid on the best chain
- Use the new state request in the transactions' verifier for mempool transactions

## Review

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -5,8 +5,6 @@
 //! implement, and ensures that we don't reject blocks or transactions
 //! for a non-enumerated reason.
 
-use std::sync::Arc;
-
 use chrono::{DateTime, Utc};
 use thiserror::Error;
 
@@ -186,7 +184,7 @@ pub enum TransactionError {
 
     #[error("could not validate nullifiers and anchors on best chain")]
     #[cfg_attr(any(test, feature = "proptest-impl"), proptest(skip))]
-    ValidateNullifiersAndAnchorsError(#[from] Arc<ValidateContextError>),
+    ValidateNullifiersAndAnchorsError(#[from] ValidateContextError),
 }
 
 impl From<BoxError> for TransactionError {
@@ -198,7 +196,7 @@ impl From<BoxError> for TransactionError {
         }
 
         match err.downcast::<ValidateContextError>() {
-            Ok(e) => return Arc::new(*e).into(),
+            Ok(e) => return (*e).into(),
             Err(e) => err = e,
         }
 
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -9,6 +9,7 @@ use chrono::{DateTime, Utc};
 use thiserror::Error;
 
 use zebra_chain::{amount, block, orchard, sapling, sprout, transparent};
+use zebra_state::ValidateContextError;
 
 use crate::{block::MAX_BLOCK_SIGOPS, BoxError};
 
@@ -33,7 +34,7 @@ pub enum SubsidyError {
     SumOverflow,
 }
 
-#[derive(Error, Clone, Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 pub enum TransactionError {
     #[error("first transaction must be coinbase")]
@@ -180,6 +181,10 @@ pub enum TransactionError {
 
     #[error("could not find a mempool transaction input UTXO in the best chain")]
     TransparentInputNotFound,
+
+    #[error("could not validate nullifiers and anchors on best chain")]
+    #[cfg_attr(any(test, feature = "proptest-impl"), proptest(skip))]
+    ValidateNullifiersAndAnchorsError(#[from] ValidateContextError),
 }
 
 impl From<BoxError> for TransactionError {
@@ -209,7 +214,7 @@ impl From<SubsidyError> for BlockError {
     }
 }
 
-#[derive(Error, Clone, Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
 pub enum BlockError {
     #[error("block contains invalid transactions")]
     Transaction(#[from] TransactionError),
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -195,6 +195,11 @@ impl From<BoxError> for TransactionError {
             Err(e) => err = e,
         }
 
+        match err.downcast::<ValidateContextError>() {
+            Ok(e) => return (*e).into(),
+            Err(e) => err = e,
+        }
+
         // buffered transaction verifier service error
         match err.downcast::<TransactionError>() {
             Ok(e) => return *e,
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -219,7 +219,7 @@ impl From<SubsidyError> for BlockError {
     }
 }
 
-#[derive(Error, Debug, PartialEq, Eq)]
+#[derive(Error, Clone, Debug, PartialEq, Eq)]
 pub enum BlockError {
     #[error("block contains invalid transactions")]
     Transaction(#[from] TransactionError),
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -5,6 +5,8 @@
 //! implement, and ensures that we don't reject blocks or transactions
 //! for a non-enumerated reason.
 
+use std::sync::Arc;
+
 use chrono::{DateTime, Utc};
 use thiserror::Error;
 
@@ -34,7 +36,7 @@ pub enum SubsidyError {
     SumOverflow,
 }
 
-#[derive(Error, Debug, PartialEq, Eq)]
+#[derive(Error, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 pub enum TransactionError {
     #[error("first transaction must be coinbase")]
@@ -184,7 +186,7 @@ pub enum TransactionError {
 
     #[error("could not validate nullifiers and anchors on best chain")]
     #[cfg_attr(any(test, feature = "proptest-impl"), proptest(skip))]
-    ValidateNullifiersAndAnchorsError(#[from] ValidateContextError),
+    ValidateNullifiersAndAnchorsError(#[from] Arc<ValidateContextError>),
 }
 
 impl From<BoxError> for TransactionError {
@@ -196,7 +198,7 @@ impl From<BoxError> for TransactionError {
         }
 
         match err.downcast::<ValidateContextError>() {
-            Ok(e) => return (*e).into(),
+            Ok(e) => return Arc::new(*e).into(),
             Err(e) => err = e,
         }
 
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -347,7 +347,7 @@ where
             if req.is_mempool() {
                 let check_anchors_and_revealed_nullifiers_query = state
                     .clone()
-                    .oneshot(zs::Request::TransactionContextualValidity(
+                    .oneshot(zs::Request::CheckBestChainTipShieldedSpends(
                         req.transaction(),
                     ));
 
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -343,6 +343,15 @@ where
 
             check::spend_conflicts(&tx)?;
 
+
+            let check_anchors_and_revealed_nullifiers_query = state
+                .clone()
+                .oneshot(zs::Request::TransactionContextualValidity(
+                    req.transaction(),
+                ));
+
+            check_anchors_and_revealed_nullifiers_query.await?;
+
             tracing::trace!(?tx_id, "passed quick checks");
 
             // "The consensus rules applied to valueBalance, vShieldedOutput, and bindingSig
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -410,7 +410,7 @@ where
                                 .expect("CheckBestChainTipNullifiersAndAnchors error must be a ValidateContextError");
                             err.into()
                         });
-                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors);
+                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, "unexpected response to CheckBestChainTipNullifiersAndAnchors request");
                         Ok(())
                     }
                 );
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -351,7 +351,7 @@ where
                         req.transaction(),
                     ));
 
-                check_anchors_and_revealed_nullifiers_query.await?;
+                assert!(check_anchors_and_revealed_nullifiers_query.await? == zs::Response::ValidBestChainTipShieldedSpends);
             }
 
             tracing::trace!(?tx_id, "passed quick checks");
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -343,17 +343,6 @@ where
 
             check::spend_conflicts(&tx)?;
 
-
-            if req.is_mempool() {
-                let check_anchors_and_revealed_nullifiers_query = state
-                    .clone()
-                    .oneshot(zs::Request::CheckBestChainTipShieldedSpends(
-                        req.transaction(),
-                    ));
-
-                assert!(check_anchors_and_revealed_nullifiers_query.await? == zs::Response::ValidBestChainTipShieldedSpends);
-            }
-
             tracing::trace!(?tx_id, "passed quick checks");
 
             // "The consensus rules applied to valueBalance, vShieldedOutput, and bindingSig
@@ -368,15 +357,16 @@ where
 
             // Load spent UTXOs from state.
             // TODO: Make this a method of `Request` and replace `tx.clone()` with `self.transaction()`?
-            let (spent_utxos, spent_outputs) =
-                Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state).await?;
+            let check_spent_utxos =
+                Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state.clone());
+            let (spent_utxos, spent_outputs) = check_spent_utxos.await?;
 
             let cached_ffi_transaction =
                 Arc::new(CachedFfiTransaction::new(tx.clone(), spent_outputs));
 
             tracing::trace!(?tx_id, "got state UTXOs");
 
-            let async_checks = match tx.as_ref() {
+            let mut async_checks = match tx.as_ref() {
                 Transaction::V1 { .. } | Transaction::V2 { .. } | Transaction::V3 { .. } => {
                     tracing::debug!(?tx, "got transaction with wrong version");
                     return Err(TransactionError::WrongVersion);
@@ -406,6 +396,21 @@ where
                     orchard_shielded_data,
                 )?,
             };
+
+            if req.is_mempool() {
+                let check_anchors_and_revealed_nullifiers_query = state
+                    .clone()
+                    .oneshot(zs::Request::CheckBestChainTipShieldedSpends(
+                        req.transaction(),
+                    ))
+                    .map(|res| {
+                        assert!(res? == zs::Response::ValidBestChainTipShieldedSpends);
+                        Ok(())
+                    }
+                );
+
+                async_checks.push(check_anchors_and_revealed_nullifiers_query);
+            }
 
             tracing::trace!(?tx_id, "awaiting async checks...");
 
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -175,10 +175,10 @@ impl Request {
     }
 
     /// The unverified mempool transaction, if this is a mempool request.
-    pub fn into_mempool_transaction(self) -> Option<UnminedTx> {
+    pub fn mempool_transaction(&self) -> Option<UnminedTx> {
         match self {
             Request::Block { .. } => None,
-            Request::Mempool { transaction, .. } => Some(transaction),
+            Request::Mempool { transaction, .. } => Some(transaction.clone()),
         }
     }
 
@@ -397,11 +397,11 @@ where
                 )?,
             };
 
-            if req.is_mempool() {
+            if let Some(unmined_tx) = req.mempool_transaction() {
                 let check_anchors_and_revealed_nullifiers_query = state
                     .clone()
                     .oneshot(zs::Request::CheckBestChainTipShieldedSpends(
-                        req.transaction(),
+                        unmined_tx,
                     ))
                     .map(|res| {
                         assert!(res? == zs::Response::ValidBestChainTipShieldedSpends);
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -31,6 +31,7 @@ use zebra_chain::{
 
 use zebra_script::CachedFfiTransaction;
 use zebra_state as zs;
+use zs::ValidateContextError;
 
 use crate::{error::TransactionError, groth16::DescriptionWrapper, primitives, script, BoxError};
 
@@ -404,6 +405,11 @@ where
                         unmined_tx,
                     ))
                     .map(|res| {
+                        let res: Result<zs::Response, TransactionError> = res.map_err(|err| {
+                            let err: ValidateContextError = *err.downcast()
+                                .expect("CheckBestChainTipNullifiersAndAnchors error must be a ValidateContextError");
+                            err.into()
+                        });
                         assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors);
                         Ok(())
                     }
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -357,9 +357,9 @@ where
 
             // Load spent UTXOs from state.
             // TODO: Make this a method of `Request` and replace `tx.clone()` with `self.transaction()`?
-            let check_spent_utxos =
+            let load_spent_utxos_fut =
                 Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state.clone());
-            let (spent_utxos, spent_outputs) = check_spent_utxos.await?;
+            let (spent_utxos, spent_outputs) = load_spent_utxos_fut.await?;
 
             let cached_ffi_transaction =
                 Arc::new(CachedFfiTransaction::new(tx.clone(), spent_outputs));
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -31,7 +31,6 @@ use zebra_chain::{
 
 use zebra_script::CachedFfiTransaction;
 use zebra_state as zs;
-use zs::ValidateContextError;
 
 use crate::{error::TransactionError, groth16::DescriptionWrapper, primitives, script, BoxError};
 
@@ -405,11 +404,6 @@ where
                         unmined_tx,
                     ))
                     .map(|res| {
-                        let res: Result<zs::Response, TransactionError> = res.map_err(|err| {
-                            let err: ValidateContextError = *err.downcast()
-                                .expect("CheckBestChainTipNullifiersAndAnchors error must be a ValidateContextError");
-                            err.into()
-                        });
                         assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, "unexpected response to CheckBestChainTipNullifiersAndAnchors request");
                         Ok(())
                     }
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -400,11 +400,11 @@ where
             if let Some(unmined_tx) = req.mempool_transaction() {
                 let check_anchors_and_revealed_nullifiers_query = state
                     .clone()
-                    .oneshot(zs::Request::CheckBestChainTipShieldedSpends(
+                    .oneshot(zs::Request::CheckBestChainTipNullifiersAndAnchors(
                         unmined_tx,
                     ))
                     .map(|res| {
-                        assert!(res? == zs::Response::ValidBestChainTipShieldedSpends);
+                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors);
                         Ok(())
                     }
                 );
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -344,13 +344,15 @@ where
             check::spend_conflicts(&tx)?;
 
 
-            let check_anchors_and_revealed_nullifiers_query = state
-                .clone()
-                .oneshot(zs::Request::TransactionContextualValidity(
-                    req.transaction(),
-                ));
+            if req.is_mempool() {
+                let check_anchors_and_revealed_nullifiers_query = state
+                    .clone()
+                    .oneshot(zs::Request::TransactionContextualValidity(
+                        req.transaction(),
+                    ));
 
-            check_anchors_and_revealed_nullifiers_query.await?;
+                check_anchors_and_revealed_nullifiers_query.await?;
+            }
 
             tracing::trace!(?tx_id, "passed quick checks");
 
--- a/zebra-consensus/src/transaction/tests.rs
+++ b/zebra-consensus/src/transaction/tests.rs
@@ -227,41 +227,6 @@ async fn mempool_request_with_missing_input_is_rejected() {
 }
 
 #[tokio::test]
-async fn mempool_request_with_invalid_input_is_rejected() {
-    let mut state: MockService<_, _, _, _> = MockService::build().for_prop_tests();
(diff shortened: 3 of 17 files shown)


zebra-consensus/src/transaction/tests.rs
```diff
@@ -215,6 +223,38 @@ async fn mempool_request_with_missing_input_is_rejected() {
...
+    let verifier_response = verifier
+        .oneshot(Request::Mempool {
+            transaction: tx.into(),
+            height,
+        })
+        .await;
+
+    assert_eq!(
+        verifier_response,
+        Err(TransactionError::TransparentInputNotFound)
```
teor2345: In this test, the validator error doesn't seem to match the error from the state.
The state error is a shielded error, but the validator is a transparent error.

Is the test covering the intended code?
Does the error handling in the state or verifier need updating?

zebra-state/src/request.rs
```diff
@@ -539,6 +540,11 @@ pub enum Request {
         /// Optionally, the hash of the last header to request.
         stop: Option<block::Hash>,
     },
+
+    /// Contextually validates anchors and nullifiers of a transaction on the best chain
+    ///
+    /// Returns [`Response::Valid`].
+    TransactionContextualValidity(Arc<Transaction>),
```
teor2345: Nitpick: Can we make this name a bit more specific?

We're only checking shielded validity here, there are separate requests for transparent contextual validity. And knowing it is the best chain tip is important for maintaining this code. (It shows us we can only use this check for mempool transactions.)

In Rust, we usually name service actions as VerbAdjectiveNoun, but for simple data access we skip the Verb (which would be Get or Lookup).

So this request could be something like:
* CheckBestChainTipNullifiersAndAnchors
* CheckBestChainTipShieldedSpends
* CheckBestChainTipShieldedTransactionContextualValidity

zebra-consensus/src/transaction.rs
```diff
@@ -343,6 +343,17 @@ where
 
             check::spend_conflicts(&tx)?;
 
+
+            if req.is_mempool() {
+                let check_anchors_and_revealed_nullifiers_query = state
```
teor2345: Reading the database from disk can take a significant amount of time. Can you add this check to async_checks, so it is done in parallel with the other expensive checks?

zebra-consensus/src/transaction.rs
```diff
@@ -343,6 +343,17 @@ where
...
+                    .clone()
+                    .oneshot(zs::Request::TransactionContextualValidity(
+                        req.transaction(),
+                    ));
+
+                check_anchors_and_revealed_nullifiers_query.await?;
```
teor2345: Nitpick: We usually assert that service responses are the correct variant, to avoid bugs.

zebra-state/src/response.rs
```diff
@@ -51,6 +51,11 @@ pub enum Response {
 
     /// The response to a `FindBlockHeaders` request.
     BlockHeaders(Vec<block::CountedHeader>),
+
+    /// Response to [`Request::TransactionContextualValidity`].
+    ///
+    /// Does not check transparent UTXO inputs
+    ContextuallyValid,
```
teor2345: Nitpick: Can we make this name a noun form of the request name?

So this request could be something like:

* ValidBestChainTipNullifiersAndAnchors
* ValidBestChainTipShieldedSpends
* BestChainTipShieldedTransactionContextuallyValid
* BestChainTipValidShieldedTransaction

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,28 +3,29 @@
...
-/// `prepared` block.
+/// `transactions` iter.
 ///
 /// This method checks for anchors computed from the final treestate of each block in
 /// the `parent_chain` or `finalized_state`.
-#[tracing::instrument(skip(finalized_state, parent_chain, prepared))]
-pub(crate) fn sapling_orchard_anchors_refer_to_final_treestates(
+pub(crate) fn sapling_orchard_anchors_refer_to_final_treestates<
+    'a,
+    TransactionsIter: Iterator<Item = &'a Arc<Transaction>>,
```
teor2345: Optional: Cloning Arcs is cheap, if you want to simplify the lifetimes and references in this PR.

zebra-state/src/service.rs
```diff
@@ -1519,6 +1519,54 @@ impl Service<ReadRequest> for ReadStateService {
...
+                            None,
+                        )?;
+
+                        // Reads from disk
+                        let sprout_final_treestates = check::anchors::fetch_sprout_final_treestates(
+                            &state.db,
+                            latest_non_finalized_best_chain.as_ref(),
+                            std::iter::once(&transaction),
+                            None,
+                        );
```
teor2345: Thanks for checking the disk last!

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,28 +3,29 @@
...
+>(
     finalized_state: &ZebraDb,
-    parent_chain: &Chain,
-    prepared: &PreparedBlock,
+    parent_chain: Option<&Arc<Chain>>,
+    transactions: TransactionsIter,
+    height: Option<Height>,
```
teor2345: Should this function be internal to this module?

There are two valid ways to call this function:
- for a block: `Some(Chain)`, one or more mined transactions, `Some(Height)`
- for a mempool transaction: `Option<Chain>`, exactly one unmined transaction, `None`

But the compiler will let us call it with any combination of these arguments.

I'd suggest having two public wrapper functions, one for a prepared block, and one for an unmined transactions. Then we can make this function private, and it will be easier to check it is always called correctly.

Can you also add documentation for the new function arguments?

This comment applies to all the functions that have been modified in this module, because they all have a similar API.

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,28 +3,29 @@
...
 ) -> Result<(), ValidateContextError> {
-    for (tx_index_in_block, transaction) in prepared.block.transactions.iter().enumerate() {
+    for (tx_index_in_block, transaction) in transactions.enumerate() {
+        let tx_index_in_block = height.map(|_| tx_index_in_block);
```
teor2345: This code is hard to understand - it was hard for me to see that `tx_index_in_block` is `None` when height is `None`. And there is no comment explaining why we do this.

Instead, we could move the loop to the block wrapper function, leaving this function as just the loop body. Then the mempool transaction wrapper can just call it once.

This comment applies to all the modified functions in this module.

zebra-state/src/service/check/anchors.rs
```diff
@@ -47,26 +48,28 @@ pub(crate) fn sapling_orchard_anchors_refer_to_final_treestates(
...
+                .unwrap_or(false)
                 && !finalized_state.contains_sapling_anchor(&anchor)
             {
                 return Err(ValidateContextError::UnknownSaplingAnchor {
                     anchor,
-                    height: prepared.height,
+                    height,
                     tx_index_in_block,
-                    transaction_hash: prepared.transaction_hashes[tx_index_in_block],
+                    transaction_hash: transaction.hash(),
```
teor2345: Repeatedly hashing the serialized bytes of large transactions is expensive, so calling `transaction.hash()` here will hurt performance.

Instead, we could pass the `UnminedTx` to the state request, and use `transaction.id.mined_id()` here.

This change is needed to all the `hash()` calls in this PR.

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,28 +3,29 @@
...
+    ValidateContextError,
 };
 
 /// Checks the final Sapling and Orchard anchors specified by transactions in this
-/// `prepared` block.
+/// `transactions` iter.
 ///
 /// This method checks for anchors computed from the final treestate of each block in
 /// the `parent_chain` or `finalized_state`.
-#[tracing::instrument(skip(finalized_state, parent_chain, prepared))]
```
teor2345, line 21: Nitpick: can you please add instrumentation to each public function in this module? It helps a lot with debugging.

It is ok to skip large fields, or fields that can't be traced.

zebra-state/src/service/check/anchors.rs
```diff
@@ -287,8 +300,8 @@ pub(crate) fn sprout_anchors_refer_to_treestates(
                     return Err(ValidateContextError::UnknownSproutAnchor {
                         anchor: joinsplit.anchor,
                         height,
-                        tx_index_in_block,
-                        transaction_hash: transaction_hashes[tx_index_in_block],
+                        tx_index_in_block: Some(tx_index_in_block),
```
teor2345: This should be `None` for mempool transactions.

zebrad/src/components/mempool/downloads.rs
```diff
@@ -442,12 +442,11 @@ where
         self.pending.len()
     }
 
-    /// Check if transaction is already in the state.
+    /// Check if transaction is already in the best chain.
     async fn transaction_in_state(
```
teor2345: Nitpick: Could we rename this method as well?

zebra-state/src/service/check/anchors.rs
teor2345: When the API is stable, I'd like to see some unit tests for these new or modified functions.

We should have some existing tests, but I'm not sure where they are. And they won't be using mempool transactions.

zebra-state/src/service/check/nullifier.rs
```diff
@@ -1,12 +1,14 @@
 //! Checks for nullifier uniqueness.
```
teor2345, line 1: When the API of this module is stable, I'd like to see some unit tests for the new functions.

We should have some existing nullifier tests, but I'm not sure where they are. And they won't be using mempool transactions.

teor2345 (changes requested):
Looks great, seems like it should work!

Since the mempool currently has a lot of very large transactions, we need to be careful of performance. I've also made some suggestions for improving various APIs so they are easier to call correctly.

I'd like to see some unit tests for the new and modified consensus-critical checks.

zebra-consensus/src/transaction/tests.rs
arya2: I think I ran the tests before saving the file, this test definitely needs an update.

As did error handling in the verifier.

zebra-state/src/service/check/anchors.rs
arya2: I like this better than writing `clone`/`cloned` where the fns are used if it's optional.

Edit: we got both via the wrappers! 🎉

zebra-state/src/service/check/anchors.rs
arya2: I was a little concerned about that too but hadn't noticed that the mempool requests had the transaction hash already and was thinking errors happens occasionally enough that it wouldn't be too important for blocks.

Updated in [`55232d1`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/55232d14135a0e82ae17d1c150223dbf4ded989d).

teor2345: Yeah unfortunately this is a denial of service risk, if a remote node can send failing mempool transactions, and get us to hash a lot of transaction data.

Mempool transactions aren't protected by proof of work, so we have to be careful about performance.

zebra-state/src/response.rs
arya2: Great suggestions, `ContextuallyValid` didn't make much sense but I was having trouble coming up with something relatively succinct.

Updated in [`3f5efe3`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/3f5efe3c6f87af004ffc5e605a3f65141a2ccc0a).

zebra-consensus/src/transaction.rs
arya2: Good idea, updated in [`be723fe`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/be723fe2522ad5787c43bbe3972b9b737a43319e).

zebra-state/src/request.rs
arya2: Good suggestions, updated in [`3f5efe3`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/3f5efe3c6f87af004ffc5e605a3f65141a2ccc0a).

zebra-state/src/service/check/anchors.rs
arya2: Sorry about that, updated in [`00c2a4b`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/00c2a4bb1cb67b1ebaa00ec615ddb3257b6c75da) and [`55232d1`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/55232d14135a0e82ae17d1c150223dbf4ded989d).

zebra-state/src/service/check/anchors.rs
arya2: Yeah, that wasn't supposed to be removed, fixed in [`65f39ba`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/65f39ba7be9bd586d3a4762502ad1e56a864acdb). Good catch.

zebra-state/src/service/check/anchors.rs
arya2: Oops. Fixed [here](https://github.com/ZcashFoundation/zebra/pull/5716/commits/55232d14135a0e82ae17d1c150223dbf4ded989d#diff-6c9234d1fb8c8dacc8e48e5ae36978d86fdbb86ff43e71498d0413af6b730738R280) in [`55232d1`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/55232d14135a0e82ae17d1c150223dbf4ded989d).

zebrad/src/components/mempool/downloads.rs
arya2: Whoops, yeah, renamed in [`65f39ba`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/65f39ba7be9bd586d3a4762502ad1e56a864acdb).

zebra-consensus/src/transaction.rs
arya2: Ah, I forgot to do that here, thank you. Fixed in [`03eba19`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/03eba19c4c2b6cfd2c029c9a02a553a37c6a000c).

zebra-state/src/service/check/anchors.rs
arya2: > We should have some existing tests, but I'm not sure where they are.

The existing tests are going through `validate_and_commit` so the new fns for checking a block's anchors should be covered, it definitely needs new tests for the mempool transaction cases.

Updated the anchors tests to call the unmined_tx check in [`1b0d532`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/1b0d532e54539bc5ef36b0126b17316a68d436c6).

zebra-consensus/src/transaction.rs
```diff
@@ -396,6 +398,26 @@ where
...
 
+            if let Some(unmined_tx) = req.mempool_transaction() {
+                let check_anchors_and_revealed_nullifiers_query = state
+                    .clone()
+                    .oneshot(zs::Request::CheckBestChainTipNullifiersAndAnchors(
+                        unmined_tx,
+                    ))
+                    .map(|res| {
+                        let res: Result<zs::Response, TransactionError> = res.map_err(|err| {
+                            let err: ValidateContextError = *err.downcast()
```
teor2345: Can we do this downcast in `impl From<BoxError> for TransactionError`, which does all the other downcasts?

That way, they are all in the same place, and we avoid bugs like converting the same underlying type to two different error variants. It will also remove a panic due to a very strict error type requirement.

zebra-consensus/src/transaction.rs
```diff
@@ -396,6 +398,26 @@ where
...
+                                .expect("CheckBestChainTipNullifiersAndAnchors error must be a ValidateContextError");
+                            err.into()
+                        });
+                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors);
```
teor2345: If we get this panic, it helps to explain why it happened.

```suggestion
                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, "unexpected response to CheckBestChainTipNullifiersAndAnchors request");
```

zebra-consensus/src/transaction.rs
```diff
@@ -357,15 +358,16 @@ where
 
             // Load spent UTXOs from state.
             // TODO: Make this a method of `Request` and replace `tx.clone()` with `self.transaction()`?
-            let (spent_utxos, spent_outputs) =
-                Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state).await?;
+            let check_spent_utxos =
+                Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state.clone());
+            let (spent_utxos, spent_outputs) = check_spent_utxos.await?;
```
teor2345: Nitpick: This is technically an existence check, but it's really there just to get the data for verifying UTXO spends. (Rust skips `get_` when naming data fetch methods.)

The actual check is performed by `Self::verify_transparent_inputs_and_outputs()`, and in the state for block transactions.

So I don't really mind the name, just thought you might want to know.

zebra-state/src/request.rs
```diff
@@ -555,6 +561,7 @@ impl Request {
             Request::Block(_) => "block",
             Request::FindBlockHashes { .. } => "find_block_hashes",
             Request::FindBlockHeaders { .. } => "find_block_headers",
+            Request::CheckBestChainTipNullifiersAndAnchors(_) => "transaction_contextual_validity",
```
teor2345: Nitpick:
```suggestion
            Request::CheckBestChainTipNullifiersAndAnchors(_) => "best_chain_tip_nullifiers_anchors",
```

zebra-state/src/request.rs
```diff
@@ -764,6 +776,9 @@ impl ReadRequest {
             ReadRequest::AddressBalance { .. } => "address_balance",
             ReadRequest::TransactionIdsByAddresses { .. } => "transaction_ids_by_addesses",
             ReadRequest::UtxosByAddresses(_) => "utxos_by_addesses",
+            ReadRequest::CheckBestChainTipNullifiersAndAnchors(_) => {
+                "transaction_contextual_validity"
```
```suggestion
                "best_chain_tip_nullifiers_anchors"
```

zebra-state/src/service.rs
```diff
@@ -1519,6 +1519,41 @@ impl Service<ReadRequest> for ReadStateService {
                 .boxed()
             }
 
+            // Transaction Verifier expects errors from this request to be the ValidateContextError type
+            // and will panic if returned a different type of error.
```
teor2345: Since service readiness errors are combined with request errors by `oneshot()`, this restriction also applies to the `poll_ready()` methods of `StateService` and `ReadStateService`.

But that seems like an unrealistic restriction, so I suggest we fall back to a generic error variant, which is what the existing downcasts in `impl From<BoxError> for TransactionError` do.

This will also need a change to the request variant doc comment.

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,193 +3,390 @@
...
 
-/// Checks the final Sapling and Orchard anchors specified by transactions in this
-/// `prepared` block.
+/// Checks the final Sapling and Orchard anchors specified by `transaction`
 ///
 /// This method checks for anchors computed from the final treestate of each block in
 /// the `parent_chain` or `finalized_state`.
-#[tracing::instrument(skip(finalized_state, parent_chain, prepared))]
-pub(crate) fn sapling_orchard_anchors_refer_to_final_treestates(
+#[tracing::instrument(skip_all)]
```
teor2345: The hash, index, and height are useful for debugging

```suggestion
#[tracing::instrument(skip(finalized_state, parent_chain, transaction))]
```

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,193 +3,390 @@
...
+}
+
+/// This function fetches and returns the Sprout final treestates from the state,
+/// so [`sprout_anchors_refer_to_treestates()`] can check Sprout final and interstitial treestates,
+/// without accessing the disk.
+///
+/// Sprout anchors may also refer to the interstitial output treestate of any prior
+/// `JoinSplit` _within the same transaction_; these are created on the fly
+/// in [`sprout_anchors_refer_to_treestates()`].
+#[tracing::instrument(skip_all)]
```
teor2345:
```suggestion
#[tracing::instrument(skip(sprout_final_treestates, finalized_state, parent_chain, transaction))]
```

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,193 +3,390 @@
...
+///
+/// Sprout anchors may refer to some earlier block's final treestate (like
+/// Sapling and Orchard do exclusively) _or_ to the interstitial output
+/// treestate of any prior `JoinSplit` _within the same transaction_.
+///
+/// This method searches for anchors in the supplied `sprout_final_treestates`
+/// (which must be populated with all treestates pointed to in the `prepared` block;
+/// see [`fetch_sprout_final_treestates()`]); or in the interstitial
+/// treestates which are computed on the fly in this function.
+#[tracing::instrument(skip(sprout_final_treestates, transaction, transaction_hash))]
```
teor2345:
```suggestion
#[tracing::instrument(skip(sprout_final_treestates, transaction))]
```

zebra-state/src/service/check/anchors.rs
```diff
@@ -3,193 +3,390 @@
...
+#[tracing::instrument(skip_all)]
+pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
+    finalized_state: &ZebraDb,
+    parent_chain: &Arc<Chain>,
+    prepared: &PreparedBlock,
+) -> Result<(), ValidateContextError> {
+    prepared
+        .block
+        .transactions
+        .par_iter()
```
teor2345: I'm not sure if we need to improve the performance of this specific check. Have you seen this check impact Zebra's performance?

Each anchor check reads a 32 byte anchor from disk or memory, which can be very cheap if it is in the non-finalized state, cached by RocksDB, or cached by the OS.

Previously, we've had to revert parallelisation that we thought would improve performance, but actually slowed things down:
- #4831

In that case, reading a whole transaction is much more expensive than an anchor, and it can perform some cryptography. But trying to do it in parallel made `lightwalletd` RPC requests much slower.

zebra-state/src/service/check/anchors.rs
```diff
@@ -199,13 +396,13 @@ pub(crate) fn fetch_sprout_final_treestates(
 /// (which must be populated with all treestates pointed to in the `prepared` block;
 /// see [`fetch_sprout_final_treestates()`]); or in the interstitial
 /// treestates which are computed on the fly in this function.
-#[tracing::instrument(skip(sprout_final_treestates, block, transaction_hashes))]
-pub(crate) fn sprout_anchors_refer_to_treestates(
+#[tracing::instrument(skip(sprout_final_treestates, block))]
```
teor2345: There can be hundreds of transactions in a block.

```suggestion
#[tracing::instrument(skip(sprout_final_treestates, block, transaction_hashes))]
```

zebrad/src/components/mempool/downloads.rs
```diff
@@ -442,12 +442,11 @@ where
         self.pending.len()
     }
 
-    /// Check if transaction is already in the state.
-    async fn transaction_in_state(
+    /// Check if transaction is already in the best chain.
+    async fn transaction_in_chain(
```
teor2345: Nitpick:
```suggestion
    async fn transaction_in_best_chain(
```

zebra-consensus/src/error.rs
```diff
@@ -33,7 +34,7 @@ pub enum SubsidyError {
     SumOverflow,
 }
 
-#[derive(Error, Clone, Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
```
teor2345: Removing `Clone` from an error can make it and every error that wraps it tricky to work with. But it's ok if it just needs minor changes to existing code.

If those changes are too large, you can try using `CloneError` instead:
https://github.com/ZcashFoundation/zebra/blob/a28350e742532983e4244959fe50f10db1193393/zebra-state/src/error.rs#L19-L25

zebra-state/src/service/check/anchors.rs
```diff
@@ -214,119 +411,68 @@ pub(crate) fn sprout_anchors_refer_to_treestates(
...
-            tracing::debug!(
-                ?joinsplit.anchor,
-                ?joinsplit_index_in_tx,
-                ?tx_index_in_block,
-                ?height,
-                "validated sprout anchor",
-            );
+    block
+        .transactions
+        .par_iter()
```
teor2345: We know this kind of code has a performance impact, because note commitment tree updates are currently slowing down the full sync.

But can you make this change in another PR, just in case the extra performance cost of the threads is too much, and we need to revert it?

Please also skip running the parallel iterator if there are no sprout transactions in the block:
```suggestion
    // If there are no sprout transactions in the block, avoid running a parallel iterator for checks that will never fail.
    if sprout_final_treestates.is_empty() {
        return Ok(());
    }

    // This check is expensive, because it updates a note commitment tree for each sprout JoinSplit.
    block
        .transactions
        .par_iter()
```

zebra-state/src/service/check/anchors.rs
```diff
@@ -214,119 +411,68 @@ pub(crate) fn sprout_anchors_refer_to_treestates(
...
-            );
-        }
-    }
+    sprout_anchors_refer_to_treestates(
+        &sprout_final_treestates,
+        &unmined_tx.transaction,
+        unmined_tx.id.mined_id(),
+        None,
+        None,
+    )?;
```
teor2345, line 470: Like the block note commitment tree updates, I think we might want to run this verification on a CPU-bound thread pool.

It would go something like this, but maybe it will need some clones.
```suggestion
    // If there are no sprout transactions in the block, avoid running a rayon scope for checks that will never fail.
    if sprout_final_treestates.is_empty() {
        return Ok(());
    }
    
    // This check is expensive, because it updates a note commitment tree for each sprout JoinSplit.
    // Since we could be processing attacker-controlled mempool transactions, we need to run each one
    // in its own thread, separately from tokio's blocking I/O threads. And if we are under heavy load,
    // we want verification to finish in order, so that later transactions can't delay earlier ones.
    rayon::scope_fifo(move |s| {
        s.spawn_fifo(move |s| {
            sprout_anchors_refer_to_treestates(
                &sprout_final_treestates,
                &unmined_tx.transaction,
                unmined_tx.id.mined_id(),
                None,
                None,
            )
        })
    })?;
```

teor2345 (commented):
Looks good, I just have some questions and ideas around error handling.

Good catch with the expensive Sprout checks!

zebra-consensus/src/transaction.rs
arya2: Updated in [`42352a5`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/42352a5a1a09f6122c971d3c036538c63b3c9f5d) and created issue #5732 for considering other approaches, maybe by having conversion fns next to the struct definition?

zebra-consensus/src/error.rs
arya2: This was very helpful, rust-analyzer seemed to miss that `TransactionError` really needs to impl `Clone` because it's used for the rejection list in the mempool storage, but there's an `io::Error` variant in `HistoryTreeError` that was getting in the way of deriving Clone for `ValidateContextError`. 

We can just put the ValidateContextError in an Arc. Thank you.

Updated in [`5a064c0`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/5a064c014cf766f0fc4bea1ff2fb855f163cbc98).

teor2345: I would encourage you to put the `io::Error` in an `Arc`, so that everything that contains a `HistoryTreeError` can be `Clone`.

zebra-consensus/src/transaction.rs
arya2: I saw that they were being used for the verify_* calls below, just hadn't put much thought into the name, but `check_` didn't feel right so I changed it to `load_spent_utxos_fut` in [`5a064c0`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/5a064c014cf766f0fc4bea1ff2fb855f163cbc98).

zebra-state/src/service.rs
arya2: > Since service readiness errors are combined with request errors by oneshot()

Ah, I hadn't noticed that, fixed in [`5a064c0`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/5a064c014cf766f0fc4bea1ff2fb855f163cbc98).

zebra-state/src/service/check/nullifier.rs
arya2: Updates `check::nullifier` tests in [`513b811`](https://github.com/ZcashFoundation/zebra/pull/5716/commits/513b8110231debb65881e4f2c1e029ef658dd987).

zebra-state/src/service/check/anchors.rs
arya2: > Have you seen this check impact Zebra's performance?

Not at all, I just noticed that they *could* be done in parallel, and it looks like a case where that would be as fast or faster than doing it in sequence.

zebra-state/src/service/check/anchors.rs
arya2: I think if `sprout_final_treestates.is_empty()` it would return `UnknownSproutAnchor` in the first iteration of the loop in `sprout_anchors_refer_to_treestates`

zebra-consensus/src/error.rs
```diff
@@ -209,7 +219,7 @@ impl From<SubsidyError> for BlockError {
     }
 }
 
-#[derive(Error, Clone, Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
```
teor2345: Nitpick: I think we fixed it so we don't need to remove this clone?

```suggestion
#[derive(Error, Clone, Debug, PartialEq, Eq)]
```

zebra-state/src/service/check/anchors.rs
teor2345: I couldn't find your comment about it, but I think you're right: the function we need to use is `unmined_tx.transaction.has_sprout_joinsplit_data()`, not the looked up anchors.

teor2345 (commented):
Looking good, I don't mind if we do the joinsplit parallelisation in a different PR.

(But it's important to do soon, because it is a denial of service risk.)

arya2:
Any idea what "the trait bound `zcash_history::Error: Clone` is not satisfied" was about? 

Worked fine locally. 😢

zebra-state/src/service/check/anchors.rs
arya2: I realized the intent was to skip the rayon scope and that you must have meant something in this direction right after commenting. Good idea / catch / point.

Do you think this state request would often be the longest running check among the async checks? Might it be worth running the duplicate nullifier checks in parallel too? Why is it a DoS risk? 

I applied a slight modification of this suggestion in https://github.com/ZcashFoundation/zebra/pull/5742.

teor2345:
> Any idea what "the trait bound `zcash_history::Error: Clone` is not satisfied" was about?
> 
> Worked fine locally. 😢

As far as I can see, `zcash_history::Error` does not `impl Clone`:
https://docs.rs/zcash_history/latest/zcash_history/enum.Error.html
https://github.com/zcash/librustzcash/blob/580f87dc3ec5d12a449b437a0358228b0fe936f3/zcash_history/src/lib.rs#L20-L26

teor2345:
Feel free to wrap it in an `Arc`.

teor2345 (approved):
Thanks, looks good!

zebra-state/src/service/check/anchors.rs
teor2345: > Do you think this state request would often be the longest running check among the async checks?

During normal operation, no, because Sprout transactions are rare.

During an attack that uses lots of Sprout JoinSplits, yes. (But we haven't seen one of those yet.)

> Might it be worth running the duplicate nullifier checks in parallel too?

No, retrieving 32 bytes from disk is usually cheap enough to be run serially, because there are usually 0-2 nullifiers revealed in a transaction, and RocksDB stores a hash index in-memory for nullifiers.

And remember, this request is already running in parallel with the proofs and the other async checks run by the transaction verifier.

> Why is it a DoS risk?

Each Sprout interstitial anchor requires updating the Sprout note commitment tree, which is expensive, because it does `O(n*log(n))` cryptographic operations per `JoinSplit`.

This is a denial of service risk because the mempool verifies remotely-controlled transaction data, without any other proof of work.

If we put that much CPU-bound work into the tokio thread pool, it could slow down network and disk requests.

arya2:
> As far as I can see,` zcash_history::Error` does not impl `Clone`

Well, that would do it. I wonder how this happened:
image

Oh well.

Issue #5652, opened 2022-11-17
Add an acceptance test for the getblocktemplate method


arya2:
## Motivation

The vector and snapshot tests for the `getblocktemplate` RPC method won't be able to use the non-finalized state to test the contextual validity checks for https://github.com/ZcashFoundation/zebra/pull/5630.

## Designs

- Use `from_script_hash(Mainnet, [0x7e; 20])` as the `miner_address` in the config
- Sync Zebra to the network tip
- Call the getblocktemplate RPC method
- Assert that the call was successful

teor2345:
Re-opening this ticket, because the test PR was reverted due to intermittent CI failures.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5652?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage @teor2345

teor2345:
So I think this just needs a quick fix to remove an assertion, or fix a bug.

Did you want an estimate for the whole ticket, or just that fix?

teor2345:
Hey @arya2 I think the errors in your PR were from code I wrote, so I'm going to try to fix them.

PR #5724, opened 2022-11-25
change(rpc): Select getblocktemplate RPC transactions according to ZIP-317

### LINKED ISSUE #5473: Select `getblocktemplate` transactions according to ZIP-317

teor2345:
## Motivation

We want to select transactions for block production according to ZIP-317.

Close #5473.

Depends-On: #5761

### Specifications

https://zips.z.cash/zip-0317#recommended-algorithm-for-block-template-construction

### Designs

Follow the recommended algorithm in the specification.
Use `f32` for efficient calculations and weight storage.

Use a minimum fee weight of `1.0 / 20.0` to simplify the implementation. (This is not in the spec, but changes are allowed because it is not consensus-critical. If the ZIP editors disagree with this, I'll open another PR to fix it.)

Use the same random weight APIs that we use in mempool transaction eviction.

Some parts of this PR are outside the `getblocktemplate-rpcs` feature, because we will probably use them in ticket #5336.

## Solution

- Add a block_production_fee_weight field to VerifiedUnminedTx
- Add a custom Zebra minimum transaction weight for block production
- Randomly select transactions for block production according to ZIP-317

Related Refactors:
- Move getblocktemplate transaction selection into a new zip317 module
- Split the conventional fee check into its own method

This PR has some testing from the existing test vectors and snapshot tests.
The rest of its tests will be implemented as part of ticket #5652.

## Review

We should probably get two reviews on this.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/transaction/unmined/zip317.rs
+++ b/zebra-chain/src/transaction/unmined/zip317.rs
@@ -1,17 +1,14 @@
-//! The [ZIP-317 conventional fee calculation](https://zips.z.cash/zip-0317#fee-calculation)
-//! for [UnminedTx]s.
+//! An implementation of the [ZIP-317] fee calculations for [UnminedTx]s:
+//! - [conventional fee](https://zips.z.cash/zip-0317#fee-calculation)
+//! - [block production transaction weight](https://zips.z.cash/zip-0317#block-production)
 
 use std::cmp::max;
 
 use crate::{
     amount::{Amount, NonNegative},
     serialization::ZcashSerialize,
-    transaction::Transaction,
+    transaction::{Transaction, UnminedTx},
 };
-
-// For doc links
-#[allow(unused_imports)]
-use crate::transaction::UnminedTx;
 
 /// The marginal fee for the ZIP-317 fee calculation, in zatoshis per logical action.
 //
@@ -26,6 +23,9 @@ const P2PKH_STANDARD_INPUT_SIZE: usize = 150;
 
 /// The standard size of p2pkh outputs for the ZIP-317 fee calculation, in bytes.
 const P2PKH_STANDARD_OUTPUT_SIZE: usize = 34;
+
+/// The weight cap for ZIP-317 block production.
+const WEIGHT_CAP: f32 = 4.0;
 
 /// Returns the conventional fee for `transaction`, as defined by [ZIP-317].
 ///
@@ -70,6 +70,18 @@ pub fn conventional_fee(transaction: &Transaction) -> Amount<NonNegative> {
     let conventional_fee = marginal_fee * max(GRACE_ACTIONS, logical_actions);
 
     conventional_fee.expect("conventional fee is positive and limited by serialized size limit")
+}
+
+/// Returns the block production fee weight for `transaction`, as defined by [ZIP-317].
+///
+/// [ZIP-317]: https://zips.z.cash/zip-0317#block-production
+pub fn block_production_fee_weight(transaction: &UnminedTx, miner_fee: Amount<NonNegative>) -> f32 {
+    let miner_fee = i64::from(miner_fee) as f32;
+    let conventional_fee = i64::from(transaction.conventional_fee) as f32;
+
+    let uncapped_weight = miner_fee / conventional_fee;
+
+    uncapped_weight.min(WEIGHT_CAP)
 }
 
 /// Divide `quotient` by `divisor`, rounding the result up to the nearest integer.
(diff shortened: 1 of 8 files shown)


zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -0,0 +1,172 @@
...
+    let (conventional_fee_txs, low_fee_txs): (Vec<_>, Vec<_>) = mempool_transactions
+        .into_iter()
+        .partition(VerifiedUnminedTx::pays_conventional_fee);
+
+    // Set up limit tracking
+    let mut selected_txs = Vec::new();
+    let mut remaining_block_sigops = MAX_BLOCK_SIGOPS;
+    let mut remaining_block_bytes: usize = MAX_BLOCK_BYTES.try_into().expect("fits in memory");
+
+    // TODO: split these into separate functions?
```
teor2345: TODO: I might want to split this into separate functions

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -27,6 +24,26 @@ const P2PKH_STANDARD_INPUT_SIZE: usize = 150;
...
+///
+/// If blocks are full, this makes it likely that very low fee transactions
+/// will be mined:
+/// - after approximately 20 blocks delay,
+/// - but before they expire.
+///
+/// Note: Small transactions that pay the legacy ZIP-313 conventional fee have twice this weight.
+/// If blocks are full, they will be mined after approximately 10 blocks delay.
+///
+/// [ZIP-203]: https://zips.z.cash/zip-0203#changes-for-blossom>
```
oxarbitrage, line 44:
```suggestion
/// Note: Small transactions that pay the legacy [ZIP-313] conventional fee have twice this weight.
/// If blocks are full, they will be mined after approximately 10 blocks delay.
///
/// [ZIP-203]: https://zips.z.cash/zip-0203#changes-for-blossom
/// [ZIP-313]: https://zips.z.cash/zip-0313#specification
```

teor2345:
The code in this PR will have further changes after a ZIP update. I suggest we review and merge it based on the current ZIP-317, and then do the update in a separate PR based on ticket #5752.

That will make reviewing the update easier, because the diff will be smaller.

teor2345:
(Oops, pressed the wrong button.)

oxarbitrage (approved):
Looks good to me.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -0,0 +1,186 @@
...
+                choose_transaction_weighted_random(&low_fee_txs, tx_weights);
+            low_fee_tx_weights = tx_weights;
+
+            // > and add it to the block. If that transaction would exceed the `size_target`
+            // > or the block sigop limit, stop without adding it.
+            if candidate_tx.legacy_sigop_count > remaining_block_sigops
+                || candidate_tx.transaction.size > remaining_block_bytes
+            {
+                // We've exceeded the scaled quota size limit, or the absolute sigop limit
+                break;
```
arya2, line 104: ~Optional: move the while loop for conventional-fee transactions to its own function and re-use it here.~

Update: nevermind, we should wait for the ZIP update.

arya2 (approved):
This looks good!

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Thanks, I've added this to ticket #5752, we might end up rewriting this comment entirely.

teor2345:
@mergifyio refresh

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
teor2345: I thought about this, but the logic was different enough that the abstraction could be confusing.

Let's see what it's like after the update.

daira:
Note that the algorithm has changed, as discussed in ZIP sync. PR here: https://github.com/zcash/zips/pull/650

teor2345:
Please don't worry about restarting the tests in this PR, or merging in main, I'm going to update the code tomorrow anyway.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: This ended up getting deleted anyway.

PR #5734, opened 2022-11-28
fix(ci): Remove failing arm64 lightwalletd builds

teor2345:
## Motivation

The QEMU install step is currently failing in an unrelated PR:
> Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": received unexpected HTTP status: 503 Service Unavailable

https://github.com/ZcashFoundation/zebra/actions/runs/3569163606/jobs/5998829876#step:6:142

## Solution

We don't support arm64 or use it in our tests, so we can just remove it for now.

## Review

This is failing all CI, so it is critical.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/zcash-lightwalletd.yml
+++ b/.github/workflows/zcash-lightwalletd.yml
@@ -91,12 +91,14 @@ jobs:
             type=sha
             type=raw,value=latest,enable={{is_default_branch}}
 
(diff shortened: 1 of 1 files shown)


teor2345:
Actually it looks like all of Docker is down.

PR #5738, opened 2022-11-29
fix(rust): Make Zebra build with the latest nightly Rust

teor2345:
## Motivation

Zebra doesn't build on the latest nightly Rust due to a compiler bug:
https://github.com/rust-lang/rust/issues/105044

## Solution

It seems like it's going to be tricky to diagnose the bug, so I rewrote the code to avoid it.

I also dealt with some new clippy lints and compiler warnings.

## Review

Anyone can review this fix. It's not urgent.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer/client/tests.rs
+++ b/zebra-network/src/peer/client/tests.rs
@@ -90,10 +90,12 @@ impl ClientTestHarness {
 
     /// Drops the mocked heartbeat shutdown receiver endpoint.
     pub fn drop_heartbeat_shutdown_receiver(&mut self) {
-        let _ = self
+        let hearbeat_future = self
             .shutdown_receiver
             .take()
-            .expect("heartbeat shutdown receiver endpoint has already been dropped");
+            .expect("unexpected test failure: heartbeat shutdown receiver endpoint has already been dropped");
+
+        std::mem::drop(hearbeat_future);
     }
 
     /// Closes the receiver endpoint of [`ClientRequest`]s that are supposed to be sent to the
--- a/zebra-network/src/peer_set/set/tests/prop.rs
+++ b/zebra-network/src/peer_set/set/tests/prop.rs
@@ -145,7 +145,8 @@ proptest! {
             }
 
             // Send a request to all peers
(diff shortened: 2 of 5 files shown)


teor2345:
Failed due to #5384 and a clippy warning about the workaround.

zebra-network/src/peer/client/tests.rs
```diff
@@ -90,10 +90,12 @@ impl ClientTestHarness {
 
     /// Drops the mocked heartbeat shutdown receiver endpoint.
     pub fn drop_heartbeat_shutdown_receiver(&mut self) {
-        let _ = self
```
dconnolly, line 93: Did this trip a linter warning?

zebra-network/src/peer_set/set/tests/prop.rs
```diff
@@ -273,7 +275,8 @@ proptest! {
             }
 
             // this will panic as expected
-            let _ = peer_set.route_broadcast(Request::AdvertiseBlock(block_hash));
+            let response_future = peer_set.route_broadcast(Request::AdvertiseBlock(block_hash));
+            std::mem::drop(response_future);
```
dconnolly, line 279: Are these explicit `drop`s part of the bug or in response to new lints?

dconnolly (approved):

zebra-network/src/peer/client/tests.rs
teor2345: Yes, a future was implicitly dropped without awaiting it.

zebra-network/src/peer_set/set/tests/prop.rs
teor2345: In response to new lints, see the commit description of 98fee6cbc53c8ac0e5d72087035d29aba65d8755

teor2345:
Failed due to #5384

teor2345:
The underlying `rustc` bug was fixed in `nightly-2022-12-09` in https://github.com/rust-lang/rust/issues/105009

PR #4933, opened 2022-08-23
Revert: deserialize stored transactions in a rayon thread

teor2345:
## Motivation

This PR fixes a `lightwalletd` sync performance regression caused by PR #4801.

While that change was more correct, it also causes a 20% slowdown in `lightwalletd` sync performance:
- 34 minutes to get to block 1,700,000 after this PR
- 39-40 minutes to get to block 1,700,000 before this PR

Close #4831.

## Solution

Revert the rayon threads added in PR #4801.

## Review

This PR is important for some of our upcoming users, but it is not urgent.

### Reviewer Checklist

  - [x] CI passes

PR #5742, opened 2022-11-29
change(state): Check block and transaction Sprout anchors in parallel

arya2:
## Motivation

We want to run these anchor verifications on a CPU-bound thread pool.

## Solution

- Use a parallel iterator when checking `sprout_anchors_refer_to_treestates(..)` for a block if `sprout_final_treestates` is not empty.
- Use `in_place_scope_fifo` for running sprout anchors checks on an `UnminedTx` if the transaction has sprout `joinsplit_data`
 
## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/check/anchors.rs
+++ b/zebra-state/src/service/check/anchors.rs
@@ -333,12 +333,8 @@ pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
     parent_chain: &Arc<Chain>,
     prepared: &PreparedBlock,
 ) -> Result<(), ValidateContextError> {
-    prepared
-        .block
-        .transactions
-        .par_iter()
-        .enumerate()
-        .try_for_each(|(tx_index_in_block, transaction)| {
+    prepared.block.transactions.iter().enumerate().try_for_each(
+        |(tx_index_in_block, transaction)| {
             sapling_orchard_anchors_refer_to_final_treestates(
                 finalized_state,
                 Some(parent_chain),
@@ -347,7 +343,8 @@ pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
                 Some(tx_index_in_block),
                 Some(prepared.height),
             )
-        })
+        },
+    )
 }
 
 /// Accepts a [`ZebraDb`], [`Arc<Chain>`](Chain), and [`PreparedBlock`].
--- a/zebra-state/src/service/check/anchors.rs
+++ b/zebra-state/src/service/check/anchors.rs
@@ -3,6 +3,8 @@
 
 use std::{collections::HashMap, sync::Arc};
 
+use rayon::prelude::*;
+
 use zebra_chain::{
     block::{Block, Height},
     sprout,
@@ -331,8 +333,12 @@ pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
     parent_chain: &Arc<Chain>,
     prepared: &PreparedBlock,
 ) -> Result<(), ValidateContextError> {
-    prepared.block.transactions.iter().enumerate().try_for_each(
-        |(tx_index_in_block, transaction)| {
+    prepared
+        .block
+        .transactions
+        .par_iter()
+        .enumerate()
+        .try_for_each(|(tx_index_in_block, transaction)| {
             sapling_orchard_anchors_refer_to_final_treestates(
                 finalized_state,
                 Some(parent_chain),
@@ -341,8 +347,7 @@ pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
                 Some(tx_index_in_block),
                 Some(prepared.height),
             )
-        },
-    )
+        })
 }
 
 /// Accepts a [`ZebraDb`], [`Arc<Chain>`](Chain), and [`PreparedBlock`].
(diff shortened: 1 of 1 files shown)


zebra-state/src/service/check/anchors.rs
```diff
@@ -331,8 +333,12 @@ pub(crate) fn block_sapling_orchard_anchors_refer_to_final_treestates(
...
     prepared: &PreparedBlock,
 ) -> Result<(), ValidateContextError> {
-    prepared.block.transactions.iter().enumerate().try_for_each(
-        |(tx_index_in_block, transaction)| {
+    prepared
+        .block
+        .transactions
+        .par_iter()
+        .enumerate()
+        .try_for_each(|(tx_index_in_block, transaction)| {
```
teor2345: This is mainly a disk operation, so it should be on the tokio thread, not a rayon thread.

Checking if a nullifier exists is usually cheap enough to be run serially, because there are usually 0-2 nullifiers revealed in a transaction, and RocksDB stores a hash index in-memory for nullifiers.

And remember, this request is already running in parallel with the proofs and the other async checks run by the transaction verifier.

So I don't think we should make this change without any evidence that it is actually slowing Zebra down:
https://github.com/ZcashFoundation/zebra/pull/5716#discussion_r1033033771

zebra-state/src/service/check/anchors.rs
```diff
@@ -435,39 +454,67 @@ pub(crate) fn tx_anchors_refer_to_final_treestates(
...
+                let mut sprout_final_treestates = HashMap::new();
+
+                fetch_sprout_final_treestates(
+                    &mut sprout_final_treestates,
+                    finalized_state,
+                    parent_chain,
+                    &unmined_tx.transaction,
+                    None,
+                    None,
+                );
```
teor2345: This is mainly a disk operation, so it should be on the tokio thread, not a rayon thread.

zebra-state/src/service/check/anchors.rs
```diff
@@ -435,39 +454,67 @@ pub(crate) fn tx_anchors_refer_to_final_treestates(
...
+                sapling_orchard_anchors_result =
+                    Some(sapling_orchard_anchors_refer_to_final_treestates(
+                        finalized_state,
+                        parent_chain,
+                        &unmined_tx.transaction,
+                        unmined_tx.id.mined_id(),
+                        None,
+                        None,
+                    ));
+            });
```
teor2345: This is mainly a disk operation, so it should be on the tokio thread, not a rayon thread.

zebra-state/src/service/check/anchors.rs
```diff
@@ -435,39 +454,67 @@ pub(crate) fn tx_anchors_refer_to_final_treestates(
...
-        None,
-    );
+                sprout_anchors_result = Some(sprout_anchors_refer_to_treestates(
+                    &sprout_final_treestates,
+                    &unmined_tx.transaction,
+                    unmined_tx.id.mined_id(),
+                    None,
+                    None,
+                ));
+            });
```
teor2345, line 496: This is the only CPU-bound operation in this entire PR, because it does multiple cryptographic operations to update interstitial sprout note commitment trees.

zebra-state/src/service/check/anchors.rs
```diff
@@ -435,39 +454,67 @@ pub(crate) fn tx_anchors_refer_to_final_treestates(
...
+    } else {
+        // If there are no sprout transactions in the block, avoid running a rayon scope
+        sapling_orchard_anchors_refer_to_final_treestates(
+            finalized_state,
+            parent_chain,
+            &unmined_tx.transaction,
+            unmined_tx.id.mined_id(),
+            None,
+            None,
+        )?;
```
teor2345: I suggest doing this check first, to avoid the expensive Sprout check if it fails.

teor2345 (changes requested):
We try to only make performance-related changes where we have evidence of slow code or a denial of service risk.

Otherwise, we could spend a lot of time modifying code that's not performance critical, making it harder to maintain. And sometimes the setup costs exceed any performance gains.

teor2345:
Failed due to:
> Unable to find image 'us-docker.pkg.dev/zealous-zebra/zebra/zebrad-test:sha-e10d851' locally
> docker: Error response from daemon: Get "https://us-docker.pkg.dev/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers).

https://github.com/ZcashFoundation/zebra/actions/runs/3578145576/jobs/6018860424#step:8:80

@gustavovalverde any idea why this is happening?

zebra-state/src/service/check/anchors.rs
arya2: > this request is already running in parallel with the proofs and the other async checks run by the transaction verifier.

This one's for the block, so it would be splitting up the work for a lot of transactions and it's used in the `validate_and_commit` method of `NonFinalizedState`.

I think it's different from the changes [here](https://github.com/ZcashFoundation/zebra/pull/4933/files) in that the work can be split up cheaply.

> I don't think we should make this change without any evidence that it is actually slowing Zebra down

Fair enough, reverted in [`5ff585b`](https://github.com/ZcashFoundation/zebra/pull/5742/commits/5ff585beef8417051ff942ea38e792e19b73d8b5).

zebra-state/src/service/check/anchors.rs
arya2: Updated in [`b8d18d8`](https://github.com/ZcashFoundation/zebra/pull/5742/commits/b8d18d86450afd2d0ba15759017971a414f50553).

zebra-state/src/service/check/anchors.rs
arya2: Updated in [`b8d18d8`](https://github.com/ZcashFoundation/zebra/pull/5742/commits/b8d18d86450afd2d0ba15759017971a414f50553).

zebra-state/src/service/check/anchors.rs
teor2345: By the way, the average number of transactions per block is 3, and one of those is a coinbase transaction:
https://bitinfocharts.com/zcash/

So the available amount of parallelism in a block is very low. And since rayon automatically adjusts parallel iterator splits to reduce overhead, there are no guarantees that each transaction will actually run in parallel, or that using multiple threads will be more efficient.

(It's different for proofs and note commitment tree updates, which can take around 1 second of CPU, so we can be doing a lot of them at the same time.)

teor2345 (dismissed):
Thanks, looks good!

zebra-state/src/service/check/anchors.rs
arya2: Ah, I should've made this change in the PR that introduced the vulnerability so that the others could've been reverted independently. Do we want to revert the `par_iter` [here](https://github.com/ZcashFoundation/zebra/pull/5742/files#diff-6c9234d1fb8c8dacc8e48e5ae36978d86fdbb86ff43e71498d0413af6b730738R436) too for this PR?

zebra-state/src/service/check/anchors.rs
arya2: > By the way, the average number of transactions per block is 3

Oh. 🤦 Okay that makes sense. Thank you.

teor2345: It's not something I would expect anyone to know, I had to check it too 🙂

zebra-state/src/service/check/anchors.rs
teor2345: I think it's fine.

Since sprout transactions are rare, and we don't run this check for checkpointed blocks, it's unlikely we'll revert this PR.

And we review reverts!

teor2345:
Failed due to #5384

teor2345:
This will need a rebase, anyone can re-approve once that's done.

teor2345:
Failed due to a temporary network error:
> buildx failed with: ERROR: failed to solve: failed to compute cache key: failed to copy: read tcp 172.17.0.2:53916->142.250.114.82:443: read: connection timed out

https://github.com/ZcashFoundation/zebra/actions/runs/3586007091/jobs/6034687982#step:9:503

teor2345 (approved):

arya2:
Failed due to https://github.com/ZcashFoundation/zebra/issues/5384 *again*.

teor2345:
@mergifyio refresh

PR #5747, opened 2022-11-30
fix(rpc): Mine standard and minimum difficulty blocks on testnet

teor2345:
## Motivation

In PR #5659, I added a check for testnet minimum difficulty blocks. But it accidentally made all the mined blocks on testnet into minimum difficulty blocks.

Depends-On: #5761

## Solution

Check if the current time is close to the minimum difficulty time (2 target block intervals). If it is, use a minimum difficulty block. If there is plenty of time, use a standard difficulty block.

Split the testnet adjustment into its own function.

Cleanup some names and constants.

## Review

This is a routine bug fix, @oxarbitrage might want to review it.

### Reviewer Checklist


teor2345:
Failed due to #5384 😭

teor2345:
@mergifyio update

zebra-state/src/service/read/difficulty.rs
```diff
@@ -102,92 +98,102 @@ where
...
+    previous_block_height: Height,
+    relevant_data: Vec<(CompactDifficulty, DateTime<Utc>)>,
+) {
+    if network == Network::Mainnet {
+        return;
+    }
+
     // On testnet, changing the block time can also change the difficulty,
     // due to the minimum difficulty consensus rule:
     // > if the block time of a block at height height ≥ 299188
```
arya2:
```suggestion
    // > if the block time of a block at height ≥ 299188
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -102,92 +98,102 @@ where
...
     // > is greater than 6 * PoWTargetSpacing(height) seconds after that of the preceding block,
     // > then the block is a minimum-difficulty block.
     //
-    // In this case, we adjust the min_time and cur_time to the first minimum difficulty time.
+    // The max time is always a minimum difficulty block, because the minimum difficulty
+    // gap is 15 minutes, but the maximum gap is 90 minutes. This means that testnet blocks
+    // have two valid time ranges with different difficulties:
+    // * 1s - 15m: standard difficulty
```
arya2: > is greater than 6 * PoWTargetSpacing(height) seconds after that of the preceding block

Would the standard difficulty time range then be 1s - 9m?

zebra-state/src/service/read/difficulty.rs
teor2345: This is copied from the spec, so the correct syntax would be:

```suggestion
    // > if the block time of a block at height `height ≥ 299188`
```

zebra-state/src/service/read/difficulty.rs
teor2345: It's actually 7.5 minutes, the 15 minutes is from before the Blossom target spacing change:
https://github.com/ZcashFoundation/zebra/blob/5c4d3e1f5ca88032b4d81f8839997a642475aabd/zebra-chain/src/parameters/network_upgrade.rs#L200

I'll go fix that now.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -102,92 +98,102 @@ where
...
+    // * 15m1s - 90m: minimum difficulty
```
teor2345:
```suggestion
    // gap is 7.5 minutes, but the maximum gap is 90 minutes. This means that testnet blocks
    // have two valid time ranges with different difficulties:
    // * 1s - 7m30s: standard difficulty
    // * 7m31s - 90m: minimum difficulty
```

zebra-state/src/service/read/difficulty.rs
teor2345: Fixed in e26c1c4

arya2 (approved):
Looks good!

PR #5758, opened 2022-12-01
fix(ci): update fake_peer_set test to avoid spurious failures

arya2:
## Motivation

This test may still be unreliable due to timing issues if the transactions are being rejected because there is a second request in the `mempool_requests_for_transactions` for the same data that might also receive `Nil`.

Related to #5384.

## Solution

- Move `mempool.enable()` below `peer_set.expect_request().await` to wait for the second chain tip change

## Review

Anyone can review.

### Reviewer Checklist


teor2345:
@arya2 can you please add a priority to this PR? Thanks!

teor2345 (approved):
Thanks, let's see if it works!

Issue #5686, opened 2022-11-21
Manually test that each mining RPC works correctly


teor2345:
## Motivation

The fields in the getblocktemplate RPC are used by miners to create new blocks. We might have missed some fields, or made some mistakes.

So we need to test that the RPC implements the consensus rules correctly. These are manual tests.

## Test Ideas

- Call the RPC using `zcash-cli` once Zebra has synced to the tip, and inspect the output
    - [x] getblocktemplate
        - [x] getblocktemplate with long polling
    - [x] submitblock
    - [x] getblockcount
    - [x] getblockhash
    - [x] getmininginfo
    - [x] getnetworksolps
    - [x] getnetworkhashps
    - TODO: add new RPCs here
- Run `zcash_rpc_diff` on zcashd and Zebra, ideally with the same mempool transactions
    - [x] getblocktemplate
        - [x] Optional: getblocktemplate with long polling (might need custom code in `zcash_rpc_diff`)
    - [x] submitblock
    - [x] getblockcount
    - [x] getblockhash
    - [x] getmininginfo
    - [x] getnetworksolps
    - [x] getnetworkhashps
    - TODO: add new RPCs here

## Testing Hints

`zcashd` usually generates a new address for each template, but you can call `zcash-cli listaddresses` and then pin one of them in `zcash.conf` using `mineraddress=X`. This address must be a P2PKH address.

If there were any differences between the coinbase transactions, I used `zcash-cli decoderawtransaction (hex data)` to compare their parsed forms. We might want to make `zcash-rpc-diff` do this automatically.

## Final Testing

- Manually run some mining software using the RPC, if possible. This depends on the RPC being finished:
    - #5455
    - #5495
    - possibly #5719
    - possibly other RPCs depending on the exact mining software

## Related Work

We could automate this testing using #5304 

- #5686
- #5803
- #5934
- #5935

teor2345:
@mpguerra I opened this ticket for specific tests for the getblocktemplate RPC, because it has a lot of fields and tricky consensus rules.

mpguerra:
Let's figure out what the tests look like manually first before we try to automate anything

teor2345:
> Let's figure out what the tests look like manually first before we try to automate anything

Thanks, I've updated the ticket to make this clearer

mpguerra:
[...]

PR #5653, opened 2022-11-17
change(tests): add acceptance test for getblocktemplate method in CI

arya2:
## Motivation

Needed for CI to test the `getblocktemplate` method with the sync to tip and contextual validation checks.

Closes #5652

## Solution

- Add a `miner_address` to the config when the `getblocktemplate-rpcs` feature is enabled
- Sync to the tip
- Call getblocktemplate RPC method and assert that the request was successful
- Add new test to CI

## Review

Anyone can review.

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

PR #5761, opened 2022-12-01
change(ci): add acceptance test for getblocktemplate RPC in CI, and fix RPC bugs

### LINKED ISSUE #5652: Add an acceptance test for the getblocktemplate method

teor2345:
## Motivation

We want to test the getblocktemplate RPC with a synced Zebra instance.
We tried this in PR #5653, but the RPC had some bugs, so we had to revert it.

Closes #5652.

## Solution

- Re-apply PR #5653 (do a revert of the revert)
- Fix bugs in the RPC method
    - An assert that was doing the wrong check, causing a panic
    - A fee that wasn't actually negated, causing a panic
- Fix coverage bugs in the test
    - The original RPC query is sent right after Zebra activates the mempool, when it is usually empty
    - Wait for transactions to download and verify in the test, then call the RPC again

## Review

This is a high priority, because it stops us adding more bugs to the getblocktemplate RPC.

### Reviewer Checklist

## Follow Up Work

- #5686

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -593,6 +593,31 @@ jobs:
       lwd_state_dir: 'lwd-cache'
     secrets: inherit
 
+  # Test that Zebra can handle a getblocktemplate RPC call, using a cached Zebra tip state
+  #
+  # Runs:
+  # - after every PR is merged to `main`
+  # - on every PR update
+  #
+  # If the state version has changed, waits for the new cached states to be created.
+  # Otherwise, if the state rebuild was skipped, runs immediately after the build job.
+  get-block-template-test:
+    name: get block template
+    needs: test-full-sync
+    uses: ./.github/workflows/deploy-gcp-tests.yml
+    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' && github.event.inputs.run-lwd-send-tx != 'true' }}
+    with:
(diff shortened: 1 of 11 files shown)


teor2345:
I have run this RPC manually thousands of times, and it seems to work (and it's very fast)

.github/workflows/continous-integration-docker.yml
```diff
@@ -593,6 +593,31 @@ jobs:
...
+      app_name: zebrad
+      test_id: get-block-template
+      test_description: Test getblocktemplate RPC method via Zebra's rpc server
+      test_variables: '-e TEST_GET_BLOCK_TEMPLATE=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache'
+      needs_zebra_state: true
+      needs_lwd_state: false
+      saves_to_disk: false
+      disk_suffix: tip
+      root_state_path: '/var/cache'
+      zebra_state_dir: 'zebrad-cache'
```
teor2345, line 619:
```suggestion
      zebra_state_dir: 'zebrad-cache'
    secrets: inherit
```

arya2 (approved):
Looks good to me!

PR #5769, opened 2022-12-01
change(rpc): Return an error from getblocktemplate RPC if Zebra is still syncing lots of blocks

### LINKED ISSUE #5675: getblocktemplate RPC: Check if Zebra is still syncing lots of blocks, and return an error if it is
## Motivation

In order to be sure that Zebra is synced to the tip for the `getblocktemplate` RPC call we need to:

- check if Zebra has finished syncing blocks
- allow the same clock skew as the Zcash network, which is 100 blocks, based on the standard rule:

### Specifications

> a full validator MUST NOT accept blocks with nTime more than two hours in the future according to its clock. This is not strictly a consensus rule because it is nondeterministic, and clock time varies between nodes.

## Related Work

#5466

arya2:
## Motivation

Mining pools may use `getblocktemplate` RPC method to check if the node is synced to the tip so we want to check that Zebra is likely close to the network chain tip before responding with a block template.

Closes #5675.

## Solution

- Add a `ChainSyncStatus` trait to zebra-chain.
- Add `sync_status` field to `GetBlockTemplateRpcsImpl`.
- Return an error from `get_block_template` method if `!sync_status.is_close_to_tip()`.
- Add a `MockSyncStatus` for tests.

## Review

### Reviewer Checklist

## Follow Up Work

- Hide `MockSyncStatus` and other test-only code behind `#[cfg(test)]`?

### DIFF
--- a/zebra-chain/src/chain_sync_status.rs
+++ b/zebra-chain/src/chain_sync_status.rs
@@ -1,5 +1,10 @@
 //! Defines method signatures for checking if the synchronizer is likely close to the network chain tip.
 
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+
 /// An interface for checking if the synchronization is likely close to the network chain tip.
 pub trait ChainSyncStatus {
     /// Check if the synchronization is likely close to the network chain tip.
@@ -9,18 +14,19 @@ pub trait ChainSyncStatus {
 /// A mock [`ChainSyncStatus`] implementation that allows setting the status externally.
 #[derive(Clone, Default)]
 pub struct MockSyncStatus {
-    is_close_to_tip: bool,
+    is_close_to_tip: Arc<AtomicBool>,
 }
 
 impl MockSyncStatus {
     /// Sets mock sync status determining the return value of `is_close_to_tip()`
     pub fn set_is_close_to_tip(&mut self, is_close_to_tip: bool) {
-        self.is_close_to_tip = is_close_to_tip;
+        self.is_close_to_tip
+            .store(is_close_to_tip, Ordering::Release);
     }
 }
 
 impl ChainSyncStatus for MockSyncStatus {
     fn is_close_to_tip(&self) -> bool {
-        self.is_close_to_tip
+        self.is_close_to_tip.load(Ordering::Acquire)
     }
 }
--- a/zebra-chain/src/chain_sync_status.rs
+++ b/zebra-chain/src/chain_sync_status.rs
@@ -5,3 +5,22 @@ pub trait ChainSyncStatus {
     /// Check if the synchronization is likely close to the network chain tip.
     fn is_close_to_tip(&self) -> bool;
 }
+
+/// A mock [`ChainSyncStatus`] implementation that allows setting the status externally.
+#[derive(Clone, Default)]
+pub struct MockSyncStatus {
+    is_close_to_tip: bool,
+}
+
+impl MockSyncStatus {
+    /// Sets mock sync status determining the return value of `is_close_to_tip()`
+    pub fn set_is_close_to_tip(&mut self, is_close_to_tip: bool) {
+        self.is_close_to_tip = is_close_to_tip;
+    }
+}
+
+impl ChainSyncStatus for MockSyncStatus {
+    fn is_close_to_tip(&self) -> bool {
+        self.is_close_to_tip
+    }
+}
--- a/zebra-chain/src/chain_sync_status.rs
+++ b/zebra-chain/src/chain_sync_status.rs
@@ -1,0 +1,7 @@
+//! Defines method signatures for checking if the synchronizer is likely close to the network chain tip.
+
+/// An interface for checking if the synchronization is likely close to the network chain tip.
+pub trait ChainSyncStatus {
+    /// Check if the synchronization is likely close to the network chain tip.
+    fn is_close_to_tip(&self) -> bool;
+}
--- a/zebra-chain/src/chain_sync_status.rs
+++ b/zebra-chain/src/chain_sync_status.rs
@@ -21,12 +21,12 @@ impl MockSyncStatus {
     /// Sets mock sync status determining the return value of `is_close_to_tip()`
     pub fn set_is_close_to_tip(&mut self, is_close_to_tip: bool) {
         self.is_close_to_tip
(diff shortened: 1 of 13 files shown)


zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -278,6 +291,7 @@ where
         let mempool = self.mempool.clone();
         let latest_chain_tip = self.latest_chain_tip.clone();
         let mut state = self.state.clone();
+        let is_zebra_close_to_network_tip = self.sync_status.is_close_to_tip();
```
teor2345: This code runs when the future is created, which can be a long time from when the async body runs.

So if Zebra is under heavy load, the sync status here might be different from the sync status when the RPC call is executed.

So let's clone the sync_status instead, and check it near the start of the async block?

teor2345 (commented):
Thanks, looks good!

Do we need to add tests to make sure we cover the 4 possible combinations of estimate status and sync status?

I think some combinations will be covered already, but I wanted to make sure they are all covered.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Fixed in [`471330d`](https://github.com/ZcashFoundation/zebra/pull/5769/commits/471330dc4769eaa7e00969c4b44b7d36382efb89) (also added coverage for the other three combinations of estimate status and sync status).

zebra-chain/src/chain_sync_status.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// Sets mock sync status determining the return value of `is_close_to_tip()`
+    pub fn set_is_close_to_tip(&mut self, is_close_to_tip: bool) {
+        self.is_close_to_tip
+            .store(is_close_to_tip, Ordering::Release);
+    }
+}
+
+impl ChainSyncStatus for MockSyncStatus {
+    fn is_close_to_tip(&self) -> bool {
+        self.is_close_to_tip.load(Ordering::Acquire)
```
teor2345: Why did you choose these memory orderings?

Memory orderings are hard to reason about, and the behaviour can vary by architecture, so we usually use the strongest ordering.

If they don't matter because this is a test-only type, please mark the type as `#[cfg(any(test, feature = "proptest-impl"))]`, use `Relaxed`, and document that.

zebra-rpc/src/methods/tests/vectors.rs
```diff
@@ -919,6 +929,30 @@ async fn rpc_getblocktemplate() {
...
         ErrorCode::ServerError(-10)
     );
+
+    mock_sync_status.set_is_close_to_tip(false);
+
+    mock_chain_tip_sender.send_estimated_distance_to_network_chain_tip(Some(0));
+    let get_block_template_sync_error = get_block_template_rpc
+        .get_block_template()
+        .await
+        .expect_err("needs an error when syncer is not close to tip or estimated distance to network chain tip is far");
```
teor2345: Nitpick:

```suggestion
        .expect_err("needs an error when syncer is not close to tip");
```

teor2345 (commented):
I try to be careful with atomic memory orderings, because very few people understand them or can reason about them accurately.

(And in most cases they are not performance-critical, but they can cause subtle bugs.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -299,7 +313,7 @@ where
                     data: None,
                 })?;
 
-            if estimated_distance_to_chain_tip > MAX_ESTIMATED_DISTANCE_TO_NETWORK_CHAIN_TIP {
+            if !sync_status.is_close_to_tip() || estimated_distance_to_chain_tip > MAX_ESTIMATED_DISTANCE_TO_NETWORK_CHAIN_TIP {
                 tracing::info!(
                     estimated_distance_to_chain_tip,
                     ?tip_height,
```
teor2345, line 327: Also, should the error number below be a constant, like:
https://github.com/ZcashFoundation/zebra/blob/d778caebb8d032d62ddf4bf9c9bd3fe3add93960/zebra-rpc/src/methods.rs#L46-L50

This stops us accidentally changing it and breaking clients.

zebra-chain/src/chain_sync_status.rs
arya2: > Why did you choose these memory orderings?

I thought it would suffice to retain the order needed but now I feel slightly unsure, and I'd rather wait the extra nanosecond than risk a mistake there, so I replaced it with `SeqCst` in [`6d256c0`](https://github.com/ZcashFoundation/zebra/pull/5769/commits/6d256c01103098bd2388d57e830ed9fb481c25d5), which is what the watch channels are using. `Relaxed` doesn't offer the constraint needed here, though that might not matter for the architectures it's being tested on.

> If they don't matter because this is a test-only type, please mark the type as #[cfg(any(test, feature = "proptest-impl"))], use Relaxed, and document that.

I think this is a really good idea, but I'd like to mark it as `#[cfg(test)]` in another PR alongside other test-only code if possible. I just created issue #5775 for that.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Great point, yes it should, fixed in [`6d256c0`](https://github.com/ZcashFoundation/zebra/pull/5769/commits/6d256c01103098bd2388d57e830ed9fb481c25d5).

zebra-rpc/src/methods/tests/vectors.rs
arya2: Oops. Thank you.

teor2345:
@mergifyio update

zebra-chain/src/chain_sync_status.rs
teor2345: Most of the code in issue #5775 is already in modules or crates that are test-only, apart from `NoChainTip`, which I fixed in PR #5777.

arya2: Thank you!!

teor2345:
@mergifyio update

teor2345 (approved):
Looks good, I think I restarted everything it needs!

Issue #5720, opened 2022-11-25
Support long polling in the getblocktemplate RPC (excluding mempool changes)


teor2345:
## Motivation

Some mining pools need long polling support, where the RPC response is delayed until the block template changes.

### Specifications

https://en.bitcoin.it/wiki/BIP_0022#Optional:_Long_Polling

### Designs

RPC fields:
- [x] Declare `longpoll` support in the `getblocktemplate` RPC `capabilities`
- [x] Provide a `longpollid` field in the RPC response, which represents the contents of the tip and mempool used to create the response.
- [x] Accept a `longpollid` parameter to the RPC request,
    - [x] which activates long polling until the ID changes.

When long polling is requested by the client using a `longpollid` parameter, delay the RPC response until:
- [x] the tip block changes:
    - [x] add a `submitold: false` field, because the old template is invalid, and
    - [x] use a `ChainTipChange` watch channel that is updated by the state;
- [x] any transactions in the mempool change:
    - [x] add a `submitold: true` field, because the old template can still be a valid block, and
    - [ ] use a watch channel that is updated by the mempool, like `ChainTipChange`;
- [x] or the `max_time` is reached (or close?):
    - [x] add a `submitold: false` field, because the old template is invalid.

RPC server:
- [x] Increase the default number of RPC threads when the `getblocktemplate-rpcs` feature is on, because every long polling client could use one thread.
    - [ ] Test the `jsonrpc_core` implementation to see what happens when a RPC request future is held open. (We assume it will do whatever the underlying `tokio` executor does.)

## Testing

I have tested long polling with `zcashd`, and it works with this command:
```sh
 zcash-cli getblocktemplate '{ "longpollid": "0000000000ab159b36911c7860f637a8e5c2d99d17c2a922029ff0905655a30080497" }'
```

mpguerra:
We probably want to do this, however, this seems like a nice and should wait until we have the basic MVP functionality done (Scope section in https://github.com/zcashfoundation/zebra/issues/5234)

teor2345:
> We probably want to do this, however, this seems like a nice and should wait until we have the basic MVP functionality done (Scope section in #5234)

How will we know our MVP implementation actually works, if we don't implement enough functionality for a mining pool to run it?

mpguerra:
[...]

PR #5772, opened 2022-12-01
change(rpc): Adds ignored jsonrequestobject argument to the getblocktemplate RPC

### LINKED ISSUE #5495: Ignore the `jsonrequestobject` argument to the `getblocktemplate` RPC
## Motivation

Our stub `getblocktemplate` RPC will fail if it gets any arguments.

But we want to ignore the optional `jsonrequestobject` argument instead, because we're not going to support any optional features requested by clients.

### Specifications

https://zcash.github.io/rpc/getblocktemplate.html

### Designs

Accept an Option<String> argument and ignore it.

## Related Work

If we discover that clients need optional features, we can parse the argument and implement them later.

arya2:
## Motivation

We want to accept and ignore an optional jsonrequestobject in the getblocktemplate RPC.

Closes #5495.

## Solution

- Add an optional argument to the `getblocktemplate` method.

## Review

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
@@ -9,6 +9,40 @@ use crate::methods::{
     GetBlockHash,
 };
 
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum GetBlockTemplateRequestMode {
+    Template,
+    Proposal,
+}
+
+impl Default for GetBlockTemplateRequestMode {
+    fn default() -> Self {
+        Self::Template
+    }
+}
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum GetBlockTemplateCapability {
+    LongPoll,
+    CoinbaseTxn,
+    CoinbaseValue,
+    Proposal,
+    ServerList,
+    WorkId,
+}
+
+/// Optional `jsonparametersobject` argument
+#[derive(Debug, serde::Deserialize)]
+pub struct JsonParameters {
+    #[serde(default)]
+    pub mode: GetBlockTemplateRequestMode,
+    #[serde(default)]
+    pub capabilities: Vec<GetBlockTemplateCapability>,
+    pub longpollid: Option<String>,
+}
+
 /// Documentation to be added after we document all the individual fields.
 #[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct GetBlockTemplate {
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
@@ -9,40 +9,6 @@ use crate::methods::{
     GetBlockHash,
 };
 
-#[derive(Debug, serde::Deserialize)]
-#[serde(rename_all = "lowercase")]
-pub enum GetBlockTemplateRequestMode {
-    Template,
-    Proposal,
-}
-
-impl Default for GetBlockTemplateRequestMode {
-    fn default() -> Self {
-        Self::Template
-    }
-}
-
-#[derive(Debug, serde::Deserialize)]
-#[serde(rename_all = "lowercase")]
-pub enum GetBlockTemplateCapability {
-    LongPoll,
-    CoinbaseTxn,
-    CoinbaseValue,
-    Proposal,
-    ServerList,
-    WorkId,
-}
-
-/// Optional `jsonparametersobject` argument
-#[derive(Debug, serde::Deserialize)]
-pub struct JsonParameters {
-    #[serde(default)]
-    pub mode: GetBlockTemplateRequestMode,
-    #[serde(default)]
-    pub capabilities: Vec<GetBlockTemplateCapability>,
-    pub longpollid: Option<String>,
-}
-
 /// Documentation to be added after we document all the individual fields.
 #[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct GetBlockTemplate {
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
@@ -24,7 +24,6 @@ impl Default for GetBlockTemplateRequestMode {
 /// Valid `capabilities` values that indicate client-side support.
 #[derive(Clone, Debug, serde::Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
-#[allow(dead_code)]
 pub enum GetBlockTemplateCapability {
     /// Long Polling support.
     /// Currently ignored by zebra.
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
@@ -79,5 +79,7 @@ pub struct JsonParameters {
     /// An id to wait for, in zcashd this is the tip hash and an internal counter.
     ///
     /// If provided, the RPC response is delayed until the mempool or chain tip block changes.
+    ///
+    /// Currently unsupported and ignored by Zebra.
     pub longpollid: Option<String>,
 }
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
@@ -51,9 +51,10 @@ pub enum GetBlockTemplateCapability {
     /// Currently ignored by zcashd and zebra.
     WorkId,
 
-    /// Mutations
+    /// Unknown capability to fill in for mutations.
     // TODO: Fill out valid mutations capabilities.
-    UnknownCapability(String),
+    #[serde(other)]
+    UnknownCapability,
 }
 
 /// Optional argument `jsonrequestobject` for `getblocktemplate` RPC request.
@@ -80,7 +81,7 @@ pub struct JsonParameters {
     /// A list of client-side supported capability features
     // TODO: Fill out valid mutations capabilities.
     #[serde(default)]
-    pub capabilities: Vec<String>,
+    pub capabilities: Vec<GetBlockTemplateCapability>,
 
     /// An id to wait for, in zcashd this is the tip hash and an internal counter.
     ///
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
@@ -22,7 +22,7 @@ impl Default for GetBlockTemplateRequestMode {
 }
 
 /// Valid `capabilities` values that indicate client-side support.
-#[derive(Debug, serde::Deserialize)]
+#[derive(Clone, Debug, serde::Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 #[allow(dead_code)]
 pub enum GetBlockTemplateCapability {
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
@@ -21,39 +21,40 @@ impl Default for GetBlockTemplateRequestMode {
     }
 }
 
-// /// Valid `capabilities` values that indicate client-side support.
-// #[derive(Debug, serde::Deserialize)]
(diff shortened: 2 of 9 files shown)


zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -9,6 +9,40 @@ use crate::methods::{
...
+    Proposal,
+    ServerList,
+    WorkId,
+}
+
+/// Optional `jsonparametersobject` argument
+#[derive(Debug, serde::Deserialize)]
+pub struct JsonParameters {
+    #[serde(default)]
+    pub mode: GetBlockTemplateRequestMode,
```
teor2345: I think mode is optional according to the spec?
https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request

I understand that serde substitutes the default, but that also needs to be documented.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -9,6 +9,40 @@ use crate::methods::{
     GetBlockHash,
 };
 
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum GetBlockTemplateRequestMode {
+    Template,
+    Proposal,
+}
```
teor2345: Please document where the list of modes comes from, and what each mode means:
https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request

This will fail if a client specifies a mode we don't know about.
Would it make sense to have an `UnknownMode(String)` variant?
Or should we error, because we can't provide the correct response to an unknown mode?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -9,6 +9,40 @@ use crate::methods::{
...
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum GetBlockTemplateCapability {
+    LongPoll,
+    CoinbaseTxn,
+    CoinbaseValue,
+    Proposal,
+    ServerList,
+    WorkId,
```
teor2345: This list is missing the mutations:
https://en.bitcoin.it/wiki/BIP_0022#Block_Template_Request

This will fail if a client specifies a capability we don't know about.
Would it make sense to have an `UnknownCapability(String)` variant?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -9,6 +9,40 @@ use crate::methods::{
...
+}
+
+/// Optional `jsonparametersobject` argument
+#[derive(Debug, serde::Deserialize)]
+pub struct JsonParameters {
+    #[serde(default)]
+    pub mode: GetBlockTemplateRequestMode,
+    #[serde(default)]
+    pub capabilities: Vec<GetBlockTemplateCapability>,
+    pub longpollid: Option<String>,
```
teor2345: If we include the `longpollid` field but ignore it, then clients will expect us to handle it correctly. But we can't do that until we implement #5720.

If we don't list this field, then the RPC will return an error, and the client should stop trying to long poll:

teor2345 (changes requested):
Looks good, but I think we might want to return errors for some unsupported arguments.

~~We're missing documentation on all these new types. Please make the types modules and the other getblocktemplate RPC modules public, so documentation is required by CI.~~

What tests do we need for this?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: I think it should be an error.

Updated in [`575deba`](https://github.com/ZcashFoundation/zebra/pull/5772/commits/575debad65065ec49e332709dad682847f3c9492).

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: Yeah, updated docs to say "all other fields are optional" in [`575deba`](https://github.com/ZcashFoundation/zebra/pull/5772/commits/575debad65065ec49e332709dad682847f3c9492).

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: I commented it out and used a `Vec<String>` instead in [`423f546`](https://github.com/ZcashFoundation/zebra/pull/5772/commits/423f546839628d32b6fd93e609d6a45506bb847a). I'm not sure if `UnknownCapability(String)` would work, but it should definitely accept all valid options.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: Updated in [`9209e03`](https://github.com/ZcashFoundation/zebra/pull/5772/commits/9209e031c95a08328530fad2278cea0902abb03c) to return an error saying it's currently unsupported.

teor2345:
> We're missing documentation on all these new types. Please make the types modules and the other getblocktemplate RPC modules public, so documentation is required by CI.

I did this in PR #5776, because the missing docs impacted that PR as well.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Currently ignored by zcashd and zebra.
+    ServerList,
+
+    /// Currently ignored by zcashd and zebra.
+    WorkId,
+
+    /// Mutations
+    // TODO: Fill out valid mutations capabilities.
+    UnknownCapability(String),
+}
```
teor2345: I think this can work using `#[serde(other)]`, but I'm not sure of the exact syntax:
https://serde.rs/variant-attrs.html#other
https://serde.rs/enum-representations.html#internally-tagged

```suggestion
#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize)]
#[serde(tag = "type", rename_all = "lowercase")]
pub enum GetBlockTemplateCapability {
    /// Long Polling support.
    /// Currently ignored by zebra.
    LongPoll,

    /// Information for coinbase transaction, default template data with the `coinbasetxn` field.
    /// Currently ignored by zebra.
    CoinbaseTxn,

    /// Coinbase value, template response provides a `coinbasevalue` field and omits `coinbasetxn` field.
    /// Currently ignored by zebra.
    CoinbaseValue,

    /// Components of the coinbase transaction.
    /// Currently ignored by zebra.
    CoinbaseAux,

    /// Currently ignored by zcashd and zebra.
    Proposal,

    /// Currently ignored by zcashd and zebra.
    ServerList,

    /// Currently ignored by zcashd and zebra.
    WorkId,

    /// Mutations
    #[serde(other)]
    UnknownCapability(String),
}
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
```diff
@@ -0,0 +1,91 @@
+//! Parameter types for the `getblocktemplate` RPC.
+
+use super::hex_data::HexData;
+
+/// Defines whether the RPC method should generate a block template or attempt to validate a block proposal.
+/// `Proposal` mode is currently unsupported and will return an error.
+#[derive(Debug, serde::Deserialize, PartialEq, Eq)]
```
teor2345: Nitpick:
```suggestion
#[derive(Clone, Debug, serde::Deserialize, PartialEq, Eq)]
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Mutations
+    // TODO: Fill out valid mutations capabilities.
+    UnknownCapability(String),
+}
+
+/// Optional argument `jsonrequestobject` for `getblocktemplate` RPC request.
+///
+/// The `data` field must be provided in `proposal` mode, and must be omitted in `template` mode.
+/// All other fields are optional.
+#[derive(Debug, serde::Deserialize, Default)]
```
teor2345: Nitpick:
```suggestion
#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, Default)]
```

zebra-rpc/src/methods/get_block_template_rpcs/types/submit_block.rs
```diff
@@ -9,7 +9,21 @@ use crate::methods::get_block_template_rpcs::GetBlockTemplateRpc;
 /// See notes for [`GetBlockTemplateRpc::submit_block`] method
 #[derive(Debug, serde::Deserialize)]
```
teor2345: Nitpick:
```suggestion
#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize)]
```

teor2345 (commented):
Looks good for now, I have an optional suggestion to get `GetBlockTemplateCapability` to work.

teor2345 (dismissed):

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
arya2: Yes!!! 💯 It just has to a unit variant. I applied a slightly modified version of this.

arya2: Yes!!! 💯 It just has to be a unit variant. I applied a slightly modified version of this.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Must be omitted as "proposal" mode is currently unsupported.
+    ///
+    /// Hex-encoded block data to be validated and checked against the server's usual acceptance rules
+    /// (excluding the check for a valid proof-of-work) when `mode` is set to `proposal`.
+    pub data: Option<HexData>,
+
+    /// A list of client-side supported capability features
+    // TODO: Fill out valid mutations capabilities.
+    #[serde(default)]
+    pub capabilities: Vec<GetBlockTemplateCapability>,
```
teor2345, line 83: Can you please add a test for this field?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Currently ignored by zcashd and zebra.
+    ServerList,
+
+    /// Currently ignored by zcashd and zebra.
+    WorkId,
+
+    /// Unknown capability to fill in for mutations.
+    // TODO: Fill out valid mutations capabilities.
+    #[serde(other)]
+    UnknownCapability,
```
teor2345, line 56: Does this need to be `UnknownCapability(String)` to store mutations like "time"?

I guess we'll find out when we test it!

```suggestion
    /// Unknown capability to fill in for mutations.
    // TODO: Fill out valid mutations capabilities.
    #[serde(other)]
    UnknownCapability(String),
```

teor2345 (dismissed):
Thanks, looks good, but I'd like to make sure we cover the `GetBlockTemplateCapability` field in our tests.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
teor2345: Oh, I saw your other comment, I guess the exact value doesn't matter if we're ignoring them anyway.

teor2345:
Looks like this will also need a rebase before it merges.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template_opts.rs
arya2: I'm not sure what to test here since it's not used in the method. Serde handles the `other` attribute as documented: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=20cee774557e5aad8c4f56b1e94295af.

teor2345: I just want to check that the entire implementation works like we expect, because there are a lot of different interacting pieces.

You could run the RPC manually, and check that it accepts an unknown capability?

arya2: Got it. Updated the acceptance in [`b46be81`](https://github.com/ZcashFoundation/zebra/pull/5772/commits/b46be810522daf93c9c7c633df68014690ae651f) to test it with and without the optional parameter and included an unknown capability there.

teor2345 (approved):
Looks great, thanks for the extra test!

teor2345:
I tested this with:
```sh
zcash-cli getblocktemplate '{ "longpollid": "000190297833eb833116702948210000000001755ed052" }'
```

And it worked!

PR #5776, opened 2022-12-02
change(rpc): Update ZIP-317 transaction selection algorithm

### LINKED ISSUE #5752: Update block production algorithm once ZIP-317 updates
## Motivation

We're expecting a minor update to the ZIP-317 block production algorithm.

### Specifications

https://zips.z.cash/zip-0317

With the changes from the PR:
https://github.com/zcash/zips/pull/650

### Designs

1. Replace the dynamic size limit with a fixed action limit. This makes the design simpler.
    - Don't make the limit configurable for now, we'll see if we need to do that later.

2. Update comments to link to ZIP-313:
    - `[ZIP-313]: https://zips.z.cash/zip-0313#specification`
    - https://github.com/ZcashFoundation/zebra/pull/5724#discussion_r1036518490

3. Replace "transparent P2SH single signature address" with "transparent P2SH address", the single signature part is wrong

teor2345:
## Motivation

The previous transaction selection algorithm was easy to work around. So we want to implement a simpler one instead.

Close #5752.

### Specifications

https://github.com/zcash/zips/blob/a9ef04852749027e515b258dc60c1fff4f87e5e2/zip-0317.rst#block-production

### Designs

Similar to the previous design, but update the calculations.

## Solution

- Update transaction weight ratio calculation
- Add unpaid actions calculation
- Update transaction selection

Related changes:
- Rename fields
- Refactor and simplify code

## Review

Anyone can review this code, it's a routine change.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -307,7 +307,8 @@ pub struct VerifiedUnminedTx {
     /// transparent inputs and outputs.
     pub legacy_sigop_count: u64,
 
-    /// The number of unpaid actions `transaction`, as defined by [ZIP-317] for block production.
+    /// The number of unpaid actions for `transaction`,
+    /// as defined by [ZIP-317] for block production.
     ///
     /// The number of actions is limited by [`MAX_BLOCK_BYTES`], so it fits in a u32.
     ///
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -32,6 +32,10 @@ use UnminedTxId::*;
 #[cfg(any(test, feature = "proptest-impl"))]
 use proptest_derive::Arbitrary;
 
+// Documentation-only
+#[allow(unused_imports)]
+use crate::block::MAX_BLOCK_BYTES;
+
 mod zip317;
 
 /// The minimum cost value for a transaction in the mempool.
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -303,13 +303,20 @@ pub struct VerifiedUnminedTx {
     /// transparent inputs and outputs.
     pub legacy_sigop_count: u64,
 
-    /// The block production fee weight for `transaction`, as defined by [ZIP-317].
+    /// The number of unpaid actions `transaction`, as defined by [ZIP-317] for block production.
+    ///
+    /// The number of actions is limited by [`MAX_BLOCK_BYTES`], so it fits in a u32.
+    ///
+    /// [ZIP-317]: https://zips.z.cash/zip-0317#block-production
+    pub unpaid_actions: u32,
+
+    /// The fee weight ratio for `transaction`, as defined by [ZIP-317] for block production.
     ///
     /// This is not consensus-critical, so we use `f32` for efficient calculations
     /// when the mempool holds a large number of transactions.
     ///
     /// [ZIP-317]: https://zips.z.cash/zip-0317#block-production
-    pub block_production_fee_weight: f32,
+    pub fee_weight_ratio: f32,
 }
 
 impl fmt::Display for VerifiedUnminedTx {
@@ -318,6 +325,8 @@ impl fmt::Display for VerifiedUnminedTx {
             .field("transaction", &self.transaction)
             .field("miner_fee", &self.miner_fee)
             .field("legacy_sigop_count", &self.legacy_sigop_count)
+            .field("unpaid_actions", &self.unpaid_actions)
+            .field("fee_weight_ratio", &self.fee_weight_ratio)
             .finish()
     }
 }
@@ -330,14 +339,15 @@ impl VerifiedUnminedTx {
         miner_fee: Amount<NonNegative>,
         legacy_sigop_count: u64,
     ) -> Self {
-        let block_production_fee_weight =
-            zip317::block_production_fee_weight(&transaction, miner_fee);
+        let fee_weight_ratio = zip317::conventional_fee_weight_ratio(&transaction, miner_fee);
+        let unpaid_actions = zip317::unpaid_actions(&transaction, miner_fee);
 
         Self {
             transaction,
             miner_fee,
             legacy_sigop_count,
-            block_production_fee_weight,
+            fee_weight_ratio,
+            unpaid_actions,
         }
     }
 
--- a/zebra-chain/src/transaction/unmined/zip317.rs
+++ b/zebra-chain/src/transaction/unmined/zip317.rs
@@ -29,10 +29,10 @@ const P2PKH_STANDARD_OUTPUT_SIZE: usize = 34;
 /// `weight_ratio_cap` in ZIP-317.
 const BLOCK_PRODUCTION_WEIGHT_RATIO_CAP: f32 = 4.0;
 
-/// The minimum fee the block production weight ratio calculation, in zatoshis.
+/// The minimum fee for the block production weight ratio calculation, in zatoshis.
 /// If a transaction has a lower fee, this value is used instead.
 ///
-/// This avoids special handling transactions with zero weight.
+/// This avoids special handling for transactions with zero weight.
 const MIN_BLOCK_PRODUCTION_SUBSTITUTE_FEE: i64 = 1;
 
 /// Returns the conventional fee for `transaction`, as defined by [ZIP-317].
--- a/zebra-chain/src/transaction/unmined/zip317.rs
+++ b/zebra-chain/src/transaction/unmined/zip317.rs
@@ -6,6 +6,7 @@ use std::cmp::max;
 
 use crate::{
     amount::{Amount, NonNegative},
+    block::MAX_BLOCK_BYTES,
     serialization::ZcashSerialize,
     transaction::{Transaction, UnminedTx},
 };
@@ -13,10 +14,10 @@ use crate::{
 /// The marginal fee for the ZIP-317 fee calculation, in zatoshis per logical action.
 //
 // TODO: allow Amount<NonNegative> in constants
-const MARGINAL_FEE: i64 = 5_000;
+const MARGINAL_FEE: u64 = 5_000;
 
 /// The number of grace logical actions allowed by the ZIP-317 fee calculation.
-const GRACE_ACTIONS: u64 = 2;
+const GRACE_ACTIONS: u32 = 2;
 
 /// The standard size of p2pkh inputs for the ZIP-317 fee calculation, in bytes.
 const P2PKH_STANDARD_INPUT_SIZE: usize = 150;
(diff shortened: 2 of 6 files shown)


zebra-chain/src/transaction/unmined.rs
```diff
@@ -303,13 +303,20 @@ pub struct VerifiedUnminedTx {
     /// transparent inputs and outputs.
     pub legacy_sigop_count: u64,
 
-    /// The block production fee weight for `transaction`, as defined by [ZIP-317].
+    /// The number of unpaid actions `transaction`, as defined by [ZIP-317] for block production.
+    ///
+    /// The number of actions is limited by [`MAX_BLOCK_BYTES`], so it fits in a u32.
```
teor2345, line 313:
```suggestion
    /// The number of actions is limited by MAX_BLOCK_BYTES, so it will fit in a u32.
```

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -56,6 +47,66 @@ pub fn conventional_fee(transaction: &Transaction) -> Amount<NonNegative> {
...
+    miner_fee: Amount<NonNegative>,
+) -> f32 {
+    // Check that this function will always return a positive, non-zero value.
+    //
+    // The maximum number of logical actions in a block is actually
+    // MAX_BLOCK_BYTES / MIN_ACTION_BYTES. MIN_ACTION_BYTES is currently
+    // the minimum transparent output size, but future transaction versions could change this.
+    assert!(
+        MIN_BLOCK_PRODUCTION_SUBSTITUTE_FEE as f32 / MAX_BLOCK_BYTES as f32 > 0.0,
+        "invalid block production constants: the minumum fee ratio must not be zero"
```
teor2345, line 93:
```suggestion
        "invalid block production constants: the minimum fee ratio must not be zero"
```

teor2345:
Since Friday I added commit f8d781f, which subtracts the coinbase transaction from the block limits.

teor2345:
This change panics in debug mode, and returns wrong but irrelevant fake coinbase block subsidies in release mode.

teor2345:
> This change panics in debug mode, and returns wrong but irrelevant fake coinbase block subsidies in release mode.

Fixed in PR #5787.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -143,35 +126,74 @@ where
...
+        && candidate_tx.legacy_sigop_count <= *remaining_block_sigops
+        && candidate_tx.unpaid_actions <= *remaining_block_unpaid_actions
+    {
+        selected_txs.push(candidate_tx.clone());
+
+        *remaining_block_bytes -= candidate_tx.transaction.size;
+        *remaining_block_sigops -= candidate_tx.legacy_sigop_count;
+
+        // Unpaid actions are always zero for transactions that pay the conventional fee,
+        // so this limit always remains the same after they are added.
```
teor2345, line 180:
```suggestion
        // so this limit always remains the same after they are added.
        // It only changes for low fee transactions.
```

zebra-chain/src/transaction/unmined/zip317.rs
teor2345, line 93:
```suggestion
        "invalid block production constants: the minimum fee ratio must not be zero"
```

zebra-chain/src/transaction/unmined.rs
```diff
@@ -303,13 +307,20 @@ pub struct VerifiedUnminedTx {
     /// transparent inputs and outputs.
     pub legacy_sigop_count: u64,
 
-    /// The block production fee weight for `transaction`, as defined by [ZIP-317].
+    /// The number of unpaid actions `transaction`, as defined by [ZIP-317] for block production.
```
arya2:
```suggestion
    /// The number of unpaid actions for `transaction`, as defined by [ZIP-317] for block production.
```

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -6,43 +6,34 @@ use std::cmp::max;
...
 
-/// The recommended weight cap for ZIP-317 block production.
-const MAX_BLOCK_PRODUCTION_WEIGHT: f32 = 4.0;
+/// The recommended weight ratio cap for ZIP-317 block production.
+/// `weight_ratio_cap` in ZIP-317.
+const BLOCK_PRODUCTION_WEIGHT_RATIO_CAP: f32 = 4.0;
 
-/// Zebra's custom minimum weight for ZIP-317 block production,
-/// based on half the [ZIP-203] recommended transaction expiry height of 40 blocks.
+/// The minimum fee the block production weight ratio calculation, in zatoshis.
```
arya2:
```suggestion
/// The minimum fee for the block production weight ratio calculation, in zatoshis.
```

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -6,43 +6,34 @@ use std::cmp::max;
...
-/// will be mined:
-/// - after approximately 20 blocks delay,
-/// - but before they expire.
-///
-/// Note: Small transactions that pay the legacy ZIP-313 conventional fee have twice this weight.
-/// If blocks are full, they will be mined after approximately 10 blocks delay.
-///
-/// [ZIP-203]: https://zips.z.cash/zip-0203#changes-for-blossom>
-const MIN_BLOCK_PRODUCTION_WEIGHT: f32 = 1.0 / 20.0;
+/// This avoids special handling transactions with zero weight.
```
arya2:
```suggestion
/// This avoids special handling for transactions with zero weight.
```

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -33,82 +45,53 @@ where
...
-                remaining_block_bytes -= candidate_tx.transaction.size;
-            }
-        }
+    // Set up limit tracking
+    let mut remaining_block_bytes: usize = MAX_BLOCK_BYTES.try_into().expect("fits in memory");
+    let mut remaining_block_sigops = MAX_BLOCK_SIGOPS;
+    let mut remaining_block_unpaid_actions: u32 = BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT;
+
+    // Adjust the limits based on the coinbase transaction
+    remaining_block_bytes -= fake_coinbase_tx.data.as_ref().len();
```
arya2, line 63: 👍

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -549,6 +549,36 @@ pub fn standard_coinbase_outputs(
...
 }
 
+/// Returns a fake coinbase transaction that can be used during transaction selection.
+///
+/// This avoids a data dependency loop involving the selected transactions, the miner fee,
+/// and the coinbase transaction.
+///
+/// This transaction's serialized size and sigops must be at least as large as the real coinbase
+/// transaction with the correct height and fee.
+fn fake_coinbase_transaction(
```
arya2, line 592: Both optional: 
- set the returned value as a field on `GetBlockTemplateRpcImpl` so it's only computed once
- rename this to something like `coinbase_transaction_allocations` and only return the `sigops`/`data.len`

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -33,82 +45,53 @@ where
...
+    remaining_block_sigops -= fake_coinbase_tx.sigops;
+
+    // > Repeat while there is any candidate transaction
+    // > that pays at least the conventional fee:
+    let mut conventional_fee_tx_weights = setup_fee_weighted_index(&conventional_fee_txs);
+
+    while let Some(tx_weights) = conventional_fee_tx_weights {
+        conventional_fee_tx_weights = checked_add_transaction_weighted_random(
```
arya2, line 71: ❤️

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -33,82 +45,53 @@ where
...
-
-            selected_txs.push(candidate_tx.clone());
-
-            remaining_block_sigops -= candidate_tx.legacy_sigop_count;
-            remaining_block_bytes -= candidate_tx.transaction.size;
-        }
+    // > Repeat while there is any candidate transaction:
+    let mut low_fee_tx_weights = setup_fee_weighted_index(&low_fee_txs);
+
+    while let Some(tx_weights) = low_fee_tx_weights {
```
arya2, line 86: Optional:

```rust
for candidate_txs in [conventional_fee_txs, low_fee_txs] {
    let mut tx_weights = setup_fee_weighted_index(&candidate_txs);

    while let Some(weights) = tx_weights {
        tx_weights = checked_add_transaction_weighted_random(
            &candidate_txs,
            weights,
            &mut selected_txs,
            &mut remaining_block_bytes,
            &mut remaining_block_sigops,
            // The number of unpaid actions is always zero for transactions that pay the
            // conventional fee, so this check and limit is effectively ignored for conventional_fee_txs.
            &mut remaining_block_unpaid_actions,
        );
    }
}
```

arya2 (dismissed):
Looks great!

Nice catch subtracting the coinbase transaction from the block sigop and size limits.

zebra-chain/src/transaction/unmined.rs
teor2345: Thanks, I'll fix this in my next PR.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Thanks, I'll fix this in my next PR.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Thanks, I'll fix this in my next PR.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: >     * set the returned value as a field on `GetBlockTemplateRpcImpl` so it's only computed once

The size of the coinbase outputs depends on the block height, so this function is using the wrong height, and we can't store the result. (We can only store the result if all future blocks don't add funding streams, but there is a rule in ZIP-1017 that allows a large grant to be paid out via an extra funding stream.)

I'll fix this in my next PR.

>     * rename this to something like `coinbase_transaction_allocations` and only return the `sigops`/`data.len`

Thanks, I'll do this in my next PR.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
teor2345: Personally I find this harder to read. My first thought was that `[conventional_fee_txs, low_fee_txs]` is a flattened array of transactions, not an array of transaction lists.

zebra-chain/src/transaction/unmined.rs
teor2345: Fixed in 332a3ae

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Fixed in 332a3ae

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Fixed in 332a3ae

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: > > ```
> > * set the returned value as a field on `GetBlockTemplateRpcImpl` so it's only computed once
> > ```
> 
> The size of the coinbase outputs depends on the block height, so this function is using the wrong height, and we can't store the result. (We can only store the result if all future blocks don't add funding streams, but there is a rule in ZIP-1017 that allows a large grant to be paid out via an extra funding stream.)

I went and fixed this bug now, because it removes the dependency on PR #5787 getting reviewed and merged. See 53f53fe

I also used a 1 zat fee, just in case someone gets clever and starts skipping zero outputs. See 375b336

teor2345: > * rename this to something like `coinbase_transaction_allocations` and only return the `sigops`/`data.len`

I decided not to do this, because the fake coinbase transaction is typed, but the fields are `u64` and `usize`. So it would be easy to get them mixed up with the real fields for the template or other transactions.

teor2345:
Oops, too many commits in that last push 😶‍🌫️

arya2 (dismissed):

teor2345:
Network timeout:
> docker: error pulling image configuration: Get "https://us-docker.pkg.dev/artifacts-downloads/namespaces/zealous-zebra/repositories/zebra/downloads/AHi26s-xGU31nIfup2vjJB8eIGmyJtAUNrsSK9rHXtgvD5msfz9Sxl804rwnUBJ9t3t5-jOTHeGk8rBzY7oXXIXLB52DGXiGSIlAOBHU1jn-xxlmI9r4cljGDzvo4H8RINxkt_f1odo60aiO2k3HvZNMgauQFJQtWpWA6J4DMaaXrq5wqSnDiQZaIyE9G7OL7yNpb54PioeKSHib2DLOVmwaNFIrlNV5h2zk47LcdZfv4ATlN72oMLPjDanidvCvM5kXv2BYAS62b67vJHGRVw240yzIaAeXaSqwHR8usD_zJoY-DRxQcrest26_hVw8Gqr2ECsWNHB8HPqtb36mVDoQr_W04IVCY56PcKEu4SJlr6ZuIiZSMuNpahjGqJO8qeEZXEGah99XQxSPIMLcSJmxwDCzLcNNzLGmNdBwE8vbt9zBpzaq33UFEuzeyAR2gDflt3J4iwWKqXDKiNws8tpsec5EjGWY4RPQ3p_VpWA5-5rRPIGA9iGymZjVzE2FQt9Tjbgp-PfNeyB2t-DrzvSrMnTxnRYlMB833-8NiFmB426cvTimYbaa9_KnkW5XepkUZxt0-EdgXAWT6FU2Ee1S7QO9JciR-LYTuocKFzRgwDgbbIm3gHRgXVyHwEMDo2e87H0oF2pc_ILIUOTrFJQ=": dial tcp 173.194.198.82:443: i/o timeout.
> Error: Process completed with exit code 125.

https://github.com/ZcashFoundation/zebra/actions/runs/3642500278/jobs/6150421912#step:9:158

teor2345:
Just a rebase and resolving some trivial conflicts.

arya2 (approved):

PR #5783, opened 2022-12-05
fix(clippy): Put Rust format variables inline

teor2345:
## Motivation

Nightly clippy has more lints about inline variables in format strings.

## Solution

Run these commands and commit the result:
```sh
cargo clippy --fix --all-features --all-targets 
cargo fmt --all 
```

### Alternatives

We could disable the lint instead, and make this change after the audit. But most of these changes are in test or error code, which isn't being audited.

## Review

Normally this would be a low priority, but I've marked it as medium because it could cause merge conflicts.

### Reviewer Checklist

### DIFF
--- a/tower-batch/tests/worker.rs
+++ b/tower-batch/tests/worker.rs
@@ -39,8 +39,7 @@ async fn wakes_pending_waiters_on_close() {
     let err = assert_ready_err!(response.poll());
     assert!(
         err.is::<error::Closed>(),
-        "response should fail with a Closed, got: {:?}",
-        err,
+        "response should fail with a Closed, got: {err:?}",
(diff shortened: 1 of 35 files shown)


teor2345:
Actually this is pretty easy to regenerate using those commands, so we can fix conflicts that way.

arya2 (approved):
This looks good!

Interesting that clippy didn't modify `format!` calls where it could only put *some* of the arguments inline.

teor2345:
> Interesting that clippy didn't modify `format!` calls where it could only put _some_ of the arguments could be inline.

That might be the next stage of the lint, it seems like someone is actively working on it.

PR #5786, opened 2022-12-05
change(release): Exclude zebra-dependencies-for-audit.md from release version updates

teor2345:
## Motivation

We don't want to update the audited dependencies during a release. Instead, we'll check if there are any critical changes, and update it separately.

## Review

This is a routine change.

### Reviewer Checklist


arya2 (approved):
Makes sense.

PR #5787, opened 2022-12-05
fix(consensus): Refactor block subsidy to handle Height::MAX without panicking

teor2345:
## Motivation

Zebra panics when doing the block subsidy calculations for `Height::MAX`.

This is unlikely to ever happen on mainnet, because it only happens when block subsidies are all zero zats. But it's good to clean it up anyway, to avoid future changes causing actual panics in production.

### Specifications

Screen Shot 2022-12-05 at 18 25 28

https://zips.z.cash/protocol/protocol.pdf#subsidies

### Designs

Use `Option<u64>` internally whenever 2<sup>Halving(height)</sup> would overflow a `u64`.

## Solution

Refactor the halving divisor to avoid overflows and panics.

Add tests for `Height::MAX` and other large values.

## Review

This is a routine consensus rule fix, it would be good to get two people to review it.

This doesn't need to be included in the audit, because it only happens with very large heights.

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/src/block/check.rs
+++ b/zebra-consensus/src/block/check.rs
@@ -1,7 +1,8 @@
 //! Consensus check functions
 
-use chrono::{DateTime, Utc};
 use std::{collections::HashSet, sync::Arc};
+
+use chrono::{DateTime, Utc};
 
 use zebra_chain::{
     amount::{Amount, Error as AmountError, NonNegative},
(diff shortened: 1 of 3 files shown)


teor2345:
This isn't needed for anything else now, so it's a low priority.

oxarbitrage (approved):
I think this looks good so probably better to just merge now instead of leaving it hanging there.

PR #5796, opened 2022-12-06
change(rpc): Provide and parse a long poll ID, but don't use it yet

teor2345:
## Motivation

We want to implement long polling support in Zebra.

Depends-On: #5772

### Specifications

getblocktemplate request:
> capabilities: miners which support long polling SHOULD provide a list including the String "longpoll"
> longpollid: "longpollid" of job to monitor for expiration; required and valid only for long poll requests

getblocktemplate response:
> longpollid: identifier for long poll request; MUST be omitted if the server does not support long polling

### Designs

Add a `LongPollInput` type which contains all the data long polling can depend on.
Lossily convert it into a `LongPollId` type, which almost always changes when the `LongPollInput` changes.
Convert the `LongPollId` type to and from a string in the `getblocktemplate` RPC.

## Solution

RPC request:
- Declare longpoll support in the getblocktemplate RPC capabilities
- Provide a longpollid field in the RPC response, which represents the contents of the tip and mempool used to create the response.
- Accept a longpollid parameter to the RPC request, but don't implement long polling yet.

RPC server:
- Increase the default number of RPC threads when the getblocktemplate-rpcs feature is on, because every long polling client could use one thread.

This is part of #5720.

## Review

I'd like an initial review of the design and the data we're using, before I add long polling waits and tests in the next PR.

### Reviewer Checklist

## Follow Up Work

- Actually wait when long polling
- Add the `submit_old` field
- Check if `jsonrpc_core` blocks when a RPC waits
- Add tests

### DIFF
--- a/zebra-rpc/src/config.rs
+++ b/zebra-rpc/src/config.rs
@@ -55,6 +55,8 @@ pub struct Config {
     pub debug_force_finished_sync: bool,
 }
 
+// This impl isn't derivable because it depends on features.
+#[allow(clippy::derivable_impls)]
 impl Default for Config {
     fn default() -> Self {
         Self {
@@ -62,7 +64,12 @@ impl Default for Config {
             listen_addr: None,
 
             // Use a single thread, so we can detect RPC port conflicts.
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
             parallel_cpu_threads: 1,
+
+            // Use multiple threads, because we pause requests during getblocktemplate long polling
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            parallel_cpu_threads: 0,
 
             // Debug options are always off by default.
             debug_force_finished_sync: false,
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
@@ -194,6 +194,9 @@ impl From<LongPollInput> for LongPollId {
 fn update_checksum(checksum: &mut u32, item: [u8; 32]) {
     for chunk in item.chunks(4) {
         let chunk = chunk.try_into().expect("chunk is u32 size");
+
+        // The endianness of this conversion doesn't matter,
+        // so we make it efficient on the most common platforms.
         let chunk = u32::from_le_bytes(chunk);
 
         // It's ok to use xor here, because long polling checks are probabilistic,
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 
 use zebra_chain::{
     block::{self, Height},
-    transaction,
+    transaction::{self, UnminedTxId},
 };
 use zebra_node_services::BoxError;
 
@@ -30,11 +30,11 @@ pub struct LongPollInput {
     /// Old work is no longer valid.
     ///
     /// We use the height as well as the hash, to reduce the probability of a missed tip change.
-    pub tip_block_height: Height,
+    pub tip_height: Height,
 
     /// If the tip block hash changes, a new template must be provided.
     /// Old work is no longer valid.
-    pub tip_block_hash: block::Hash,
+    pub tip_hash: block::Hash,
 
     /// If the max time is reached, a new template must be provided.
     /// Old work is no longer valid.
@@ -52,6 +52,26 @@ pub struct LongPollInput {
     pub mempool_transaction_mined_ids: Arc<[transaction::Hash]>,
 }
 
+impl LongPollInput {
+    /// Returns a new [`LongPollInput`], based on the supplied fields.
+    pub fn new(
+        tip_height: Height,
+        tip_hash: block::Hash,
+        max_time: DateTime<Utc>,
+        mempool_tx_ids: impl IntoIterator<Item = UnminedTxId>,
+    ) -> Self {
+        let mempool_transaction_mined_ids =
+            mempool_tx_ids.into_iter().map(|id| id.mined_id()).collect();
+
+        LongPollInput {
+            tip_height,
+            tip_hash,
+            max_time,
+            mempool_transaction_mined_ids,
+        }
+    }
+}
+
 /// The encoded long poll ID, generated from the [`LongPollInput`].
 ///
 /// `zcashd` IDs are currently 69 hex/decimal digits long.
@@ -65,7 +85,7 @@ pub struct LongPollId {
     /// Old work is no longer valid.
     ///
     /// We use the height as well as the hash, to reduce the probability of a missed tip change.
-    pub tip_block_height: u32,
+    pub tip_height: u32,
 
     /// If the tip block changes, a new template must be provided.
     /// Old work is no longer valid.
@@ -73,7 +93,7 @@ pub struct LongPollId {
     ///
     /// It's ok to do a probabilistic check here,
     /// so we choose a 1 in 2^32 chance of missing a block change.
-    pub tip_block_checksum: u32,
+    pub tip_hash_checksum: u32,
 
     /// If the max time is reached, a new template must be provided.
     /// Old work is no longer valid.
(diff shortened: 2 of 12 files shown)


zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Zcash times are limited to 32 bits by the consensus rules.
+    pub max_timestamp: u32,
+
+    // Fields that allow old work:
+    //
+    /// If the number of mempool transactions changes, a new template might be provided.
+    /// Old work is still valid.
+    ///
+    /// The number of transactions is limited by the mempool DoS limit.
+    pub mempool_transaction_count: u32,
```
teor2345, line 138: This could be a `u16` if we want, because the number of transactions in the mempool is usually small. And it doesn't matter if we wrap when converting from `usize`, because we're only trying to detect changes. (The gap between identical wrapped `u16`s is 2<sup>16</sup>.)

But consistently using `u32`s is fine too.

zebra-rpc/src/config.rs
```diff
@@ -55,15 +55,22 @@ pub struct Config {
...
     fn default() -> Self {
         Self {
             // Disable RPCs by default.
             listen_addr: None,
 
             // Use a single thread, so we can detect RPC port conflicts.
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
             parallel_cpu_threads: 1,
 
+            // Use multiple threads, because we pause requests during getblocktemplate long polling
```
arya2, line 70: I doubt `jsonrpc_core` blocks when an RPC is waiting on a future.

arya2 (commented):
What's the advantage of this design over tip hash + mempool version?

arya2:
Alternative proposal: we could generate the template with a uuid in a different task when there's a tip change or mempool update then return the uuid as the longpollid.

zebra-rpc/src/config.rs
teor2345: It probably does whatever the tokio executor does, because that's what it uses internally to handle futures on its threads.

So it will block the thread if the future does a blocking task, or uses a lot of CPU, or doesn't yield. But otherwise it will be ok.

teor2345:
> What's the advantage of this design over tip hash + mempool version?

This is a partial implementation of a tip hash + mempool design, without the part where we check if the long poll id is the same, and wait to return the result.

`zcashd`'s long poll id implementation effectively places two restrictions on us:
- must use hex digits
- must be less than or equal to 69 hex digits

We might want to have exactly 69 digits, or we might want fewer, so our IDs can't be confused. I've gone for fewer here, we can always pad later if we find out that's needed during testing.

A block hash is 64 hex digits, but we also want to include the max time, so we can send a new template if there are no transaction changes in the mempool, and the template expires. We also need to fit some kind of mempool transaction content marker into the 69 digits.

There can be multiple transactions in the mempool, so we need some way of condensing their hashes into a smaller number of bytes. And once we have that, we might as well use it on the block hash.

The tip height and mempool transaction count are optional, but they are useful for avoiding duplicate checksums, and for debugging. We have space to write these fields and the time as decimal, which makes them more readable.

teor2345:
> Alternative proposal: we could generate the template with a uuid in a different task when there's a tip change or mempool update then return the uuid as the longpollid.

I think I see how this could work, but how would we avoid generating templates when there were no requests?

This also adds concurrency, which increases the complexity of this change. So if we can do it without concurrency, I'd like to try that first.

teor2345:
I need to handle parsing the long poll id correctly, it should be valid here:
```
---- methods::tests::vectors::rpc_getblocktemplate stdout ----

  left: `ServerError(-10)`,
 right: `InvalidParams`
Location: zebra-rpc/src/methods/tests/vectors.rs:1001
```

https://github.com/ZcashFoundation/zebra/actions/runs/3625573409/jobs/6114169640#step:3:3563

arya2:
> but how would we avoid generating templates when there were no requests?

Tower service driven by polling!

> This also adds concurrency, which increases the complexity of this change. So if we can do it without concurrency, I'd like to try that first.

I see your point. We could also do a uuid with a cached template and check the max_time there.

teor2345:
> > This also adds concurrency, which increases the complexity of this change. So if we can do it without concurrency, I'd like to try that first.
> 
> I see your point. We could also do a uuid with a cached template and check the max_time there.

I'd like to try for the simplest working implementation first, and then make it more efficient.

teor2345:
Just a rebase and some minor tidy-ups.

teor2345:
And finally a fix to the config file and its tests.

zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+/// Update `checksum` from `item`, so changes in `item` are likely to also change `checksum`.
+///
+/// This checksum is not cryptographically secure.
+fn update_checksum(checksum: &mut u32, item: [u8; 32]) {
+    for chunk in item.chunks(4) {
+        let chunk = chunk.try_into().expect("chunk is u32 size");
+        let chunk = u32::from_le_bytes(chunk);
```
arya2, line 200: Why call `from_le_bytes` here?

zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
```diff
@@ -0,0 +1,276 @@
...
+        )
+    }
+}
+
+impl FromStr for LongPollId {
+    type Err = BoxError;
+
+    /// Exact conversion from a string to LongPollId.
+    fn from_str(long_poll_id: &str) -> Result<Self, Self::Err> {
+        if long_poll_id.len() != LONG_POLL_ID_LENGTH {
```
arya2, line 246: 👍

zebra-rpc/src/methods/tests/vectors.rs
```diff
@@ -980,14 +991,36 @@ async fn rpc_getblocktemplate() {
     let get_block_template_sync_error = get_block_template_rpc
         .get_block_template(Some(
             get_block_template_rpcs::types::get_block_template_opts::JsonParameters {
-                longpollid: Some("".to_string()),
+                mode: GetBlockTemplateRequestMode::Proposal,
```
arya2: There's a call in proposal mode above, we can get rid of this.

zebrad/tests/acceptance.rs
```diff
@@ -654,6 +655,7 @@ fn valid_generated_config(command: &str, expect_stdout_line_contains: &str) -> R
 }
 
 /// Check if the config produced by current zebrad is stored.
+#[tracing::instrument]
```
arya2, line 658:
```suggestion
#[allow(clippy::print_stdout)]
#[tracing::instrument]
```

arya2 (commented):
Looks good, I like that you used a checksum of the tip hash instead of the full hash.

zebra-rpc/src/methods/get_block_template_rpcs/types/long_poll.rs
teor2345: We have to convert using some method, and it doesn't really matter which one.
I added a comment in 9cea1242ae.

zebra-rpc/src/methods/tests/vectors.rs
teor2345: Thanks, fixed in 7f0a612150.

zebrad/tests/acceptance.rs
teor2345: Thanks, fixed in 6c088c1930.

str4d:
> `zcashd`'s long poll id implementation effectively places two restrictions on us:
>
> - must use hex digits
> - must be less than or equal to 69 hex digits


Does it? The only restriction I see is that it must be encoded as a JSON string.

A long poll ID is effectively an identifier for the returned template, so the next call can know what you are currently using, and therefore what template state to measure "is updated" against. It therefore only makes sense in the context of the node instance that generated it, because only that instance can remember what the template was given the current API. Long poll IDs aren't even portable across different instances of the same implementation (`zcashd`'s `nTransactionsUpdatedLastLP` is derived from local state transitions, not global).

So unless there's something else in (Bitcoin Core's partial implementation of) BIP 22 that constrains it (which appears to not be the case; in fact "clients must not assume the long poll ID has any specific meaning" is stated) then you can do whatever you want here.

arya2 (approved):
> The only restriction I see is that it must be encoded as a JSON string.

That's my understanding as well, I don't think it could be used as anything besides an id.

This should work either way.

teor2345:
> So unless there's something else in (Bitcoin Core's partial implementation of) BIP 22 that constrains it (which appears to not be the case; in fact "clients must not assume the long poll ID has any specific meaning" is stated) then you can do whatever you want here.

While this is technically true according to the spec, there is nothing stopping client implementations:
- using a 69 byte array for the long poll id
- checking the length of the long poll id
- checking the format of the long poll id

So if we want to be maximally compatible with existing clients, we should avoid going over 69 bytes, and avoid using non-hex characters.

teor2345:
There's also no guarantee that any incompatibilities will reliably cause errors. In a memory-unsafe language, it's possible to write a few bytes past a 69 byte array without detecting it, depending on the allocator and memory pressure.

PR #5806, opened 2022-12-06
fix(ci): Fix cargo deny syntax

teor2345:
## Motivation

There's a copy-paste error in our cargo deny CI, but it still seems to work?

## Review

This is a low priority CI fix.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -270,4 +270,4 @@ jobs:
         uses: EmbarkStudios/cargo-deny-action@v1
         with:
           command: check ${{ matrix.checks }}
-          args: --workspace ${{ matrix.features }}
+          arguments: --workspace ${{ matrix.features }}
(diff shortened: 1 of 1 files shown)


.github/workflows/continous-integration-os.yml
```diff
@@ -270,4 +270,4 @@ jobs:
         uses: EmbarkStudios/cargo-deny-action@v1
         with:
           command: check ${{ matrix.checks }}
-          args: --workspace with features ${{ matrix.features }}
+          args: --workspace ${{ matrix.features }}
```
arya2:
```suggestion
          arguments: --workspace ${{ matrix.features }}
```

arya2 (commented):
Nice catch!

> it still seems to work?

It seems to ignore `args` altogether.

.github/workflows/continous-integration-os.yml
teor2345: Yep, that's the correct syntax:
https://github.com/EmbarkStudios/cargo-deny-action#use-specific-rust-version

arya2 (approved):
🎉

PR #5814, opened 2022-12-07
fix(ci): Run basic and ignored unit tests at the same time, speeding up CI by half an hour

### LINKED ISSUE #5812: Create a getblocktemplate-rpcs Docker image, and use it in the tests to speed up CI
## Motivation

Currently, the slowest Zebra test is "Test all with getblocktemplate-rpcs feature", which can take over two hours (including Docker image build time). All other tests take less than 1h 45m.

This test compiles Zebra twice, taking about half an hour each time.

If we sped this test up, our tests would finish quicker.

### Designs

- Build a Docker image with `--features getblocktemplate-rpcs`, ideally using a layer of the existing image (but the production image must not have `getblocktemplate-rpcs` active)
- Use that image in the "Test all with getblocktemplate-rpcs feature" and "get block template" tests

teor2345:
## Motivation

Currently we spent half an hour re-compiling Zebra for one ignored test.

See ticket #5812 for details.

## Solution

Run the ignored tests with the other tests, saving half an hour in CI

Closes #5812 by achieving the same goal, but in a simpler way.

## Review

Anyone can review this PR, I've marked it as critical so it goes first in the queue, and speeds up other batches.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -205,18 +205,13 @@ jobs:
         with:
           short-length: 7
 
-      # Run unit and basic acceptance tests, only showing command output if the test fails.
+      # Run unit, basic acceptance tests, and ignored tests, only showing command output if the test fails.
       #
       # If some tests hang, add "-- --nocapture" for just that test, or for all the tests.
-      - name: Run basic zebrad tests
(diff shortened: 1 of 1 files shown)


arya2 (approved):
Good idea.

PR #5808, opened 2022-12-07
change(rpc): Adds `getmininginfo`, `getnetworksolps` and `getnetworkhashps` methods

arya2:
## Motivation

We want to support the `getnetworksolps`, `getnetworkhashps`, and `getmininginfo` RPCs.

- Closes #5721
- Closes #5468

## Solution

- Adds a state request for getting the chain solution rate
  - Sums up total work for a range of blocks
  - Divides it by number of seconds elapsed between the first and last block in the range
- Adds RPC `getmininginfo`, `getnetworksolps` and `getnetworkhashps` RPCs

## Review

Part of regular scheduled work.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

- Populate other fields for the `getmininginfo` response

PR #5822, opened 2022-12-08
fix(dev): Make zcash-rpc-diff always check Zebra then zcashd

teor2345:
## Motivation

Sometimes they were in the opposite order in the script, which is confusing when reviewing diffs.

## Review

This is a minor usability fix for a developer testing tool.

I used it for the 3 diffs in:
https://github.com/ZcashFoundation/zebra/pull/5808#issuecomment-1342137431

### Reviewer Checklist

### DIFF
--- a/zebra-utils/zcash-rpc-diff
+++ b/zebra-utils/zcash-rpc-diff
@@ -69,15 +69,15 @@ echo
 
 if [ "$ZEBRAD_NET" != "$ZCASHD_NET" ]; then
     echo "WARNING: comparing RPC responses from different networks:"
-    echo "$ZCASHD is on: $ZCASHD_NET"
     echo "$ZEBRAD is on: $ZEBRAD_NET"
+    echo "$ZCASHD is on: $ZCASHD_NET"
     echo
 fi
 
 if [ "$ZEBRAD_HEIGHT" -ne "$ZCASHD_HEIGHT" ]; then
     echo "WARNING: comparing RPC responses from different heights:"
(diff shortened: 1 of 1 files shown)


arya2 (dismissed):

arya2:
@Mergifyio update

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -618,6 +618,56 @@ where
...
+                    message: "No blocks in state".to_string(),
+                    data: None,
+                })?,
+                _ => unreachable!("unmatched response to a solution rate request"),
+            };
+
+            Ok(solution_rate)
+        }
+        .boxed()
+    }
```
arya2: This needs a rebase.

```suggestion
```

teor2345:
There shouldn't be any Rust code in this PR, I'll fix it.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: I don't think this will work, it will delete that RPC method when merged to the main branch.

arya2: It's duplicate, the same code is right above this too, but a rebase would be better.

teor2345:
It should be fixed now, I probably caused this by basing my branch off the RPC I was testing.

arya2 (approved):

PR #5827, opened 2022-12-08
change(rpc): support transparent p2pkh miner addresses

### LINKED ISSUE #5719: Implement transparent P2PKH mining address support
## Motivation

Some mining pools have told us they use P2PKH transparent addresses for coinbase transactions, but Zebra only currently supports P2SH transparent addresses.

### Specification

```
In a P2PKH output, the pubkey script is:
OP_DUP OP_HASH160 <PubkeyHash> OP_EQUALVERIFY OP_CHECKSIG
```

https://developer.bitcoin.org/devguide/transactions.html#p2pkh-script-validation

In Zebra, `<PubkeyHash>` is implemented as:
```rust
bytes.push(OpCode::Push20Bytes);
bytes.extend(address.hash_bytes());
```

### Designs

- If the configured miner address is a P2PKH address, create the miner payout script using that P2PKH address (the funding streams must stay as P2SH scripts, this is a consensus rule)
    - Also support unified addresses that contain P2SH or P2PKH addresses (this can be a separate PR)
- Remove assertions that miner addresses are P2SH

#### Optional Changes

We might want to move the existing script opcodes to a new module, and add the others we need.
It will be easier to write code with the opcodes if they are `#[repr(u8)]`.

### Existing Code

We already define some of the required script opcodes here:
https://github.com/ZcashFoundation/zebra/blob/ce07f4d7af9eae3fd80c38ff3ecf5adb71b7b1b7/zebra-consensus/src/block/subsidy/funding_streams.rs#L175-L180

oxarbitrage:
## Motivation

We want to support P2PKH as miner address for the `getblocktemplate` rpc.

Closes https://github.com/ZcashFoundation/zebra/issues/5719

## Solution

- Move opcodes to its own module inside zebra-chain transparent.
- Add method to create script given any of the 2 possible transparent address.
- Use new methods in funding streams code.
- Use new method to validate config miners address and output addition.
- Expand `rpc_getblocktemplate` test to run with the 2 address types. 

## Review

I think anyone can review.

### Reviewer Checklist


teor2345:
Thanks, I'll have a look at this either this afternoon, or on Monday.

Would you mind rebasing or merging in `main`, so that CI runs?

zebra-chain/src/transparent/opcodes.rs
```diff
@@ -0,0 +1,13 @@
+//! Zebra script opcodes.
+
+/// Supported opcodes
+pub enum OpCode {
```
teor2345: This is part of the spec, so we need to quote where it comes from.

```suggestion
/// Supported opcodes
///
/// <https://github.com/zcash/zcash/blob/8b16094f6672d8268ff25b2d7bddd6a6207873f7/src/script/script.h#L39>
pub enum OpCode {
```

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -144,22 +143,11 @@ pub fn new_coinbase_script(address: Address) -> Script {
          Zebra only supports transparent 'pay to script hash' (P2SH) addresses",
```
teor2345:
```suggestion
         Funding streams only support transparent 'pay to script hash' (P2SH) addresses",
```

zebra-rpc/src/methods/get_block_template_rpcs/config.rs
```diff
@@ -9,7 +9,7 @@ use zebra_chain::transparent;
 #[serde(deny_unknown_fields, default)]
 pub struct Config {
     /// The address used for miner payouts.
-    /// Zebra currently only supports single-signature P2SH transparent addresses.
+    /// Zebra currently only supports single-signature P2SH and P2PKH transparent addresses.
```
teor2345: This is my mistake, I thought I'd got all of them

```suggestion
    /// Zebra currently only supports P2SH and P2PKH transparent addresses.
```

teor2345 (commented):
These are all comment or log fixes, so I'm going to commit them, and get someone else to approve.

teor2345 (dismissed):

zebra-chain/src/transparent/address.rs
```diff
@@ -215,6 +215,14 @@ impl Address {
         }
     }
 
+    /// Create an address for the given public key hash and network.
+    pub fn from_public_key_hash(network: Network, pub_key_hash: [u8; 20]) -> Self {
```
arya2: It looks like this method is identical to the `from_pub_key_hash` method above.

zebra-chain/src/transparent/address.rs
```diff
@@ -215,6 +215,14 @@ impl Address {
...
+        Self::PayToPublicKeyHash {
+            network,
+            pub_key_hash,
+        }
+    }
+
```
arya2:
```suggestion
```

arya2 (dismissed):
This looks good, there's a minor suggestion to remove the unused/duplicate `from_public_key_hash`.

zebra-chain/src/transparent/address.rs
teor2345: Good catch!

Since this is consensus-critical code, we can't have two implementations, even if they are identical.

arya2 (approved):

teor2345 (approved):

zebra-chain/src/transparent/address.rs
oxarbitrage: Thank you for that, i created the method for a test and later realized we already had it. It seems i used the correct one in the test but forgot to delete the created one. Thanks again.

PR #5829, opened 2022-12-08
change(rpc): return u64 from `get_network_sol_ps` and remove `arbitrary_precision` feature from serde

arya2:
## Motivation

Mining RPCs responding with the network solution rate added in https://github.com/ZcashFoundation/zebra/pull/5808 are returning a `u128`, requiring the `arbitrary_precision` feature from `serde`.

There are issues open in `serde` about the `arbitrary_precision` feature (see https://github.com/serde-rs/json/issues/505 and https://github.com/serde-rs/serde/issues/1183). We haven't run into these issues but we only need `u64` for any feasible solution rate that might be returned anyways, so we should use a `u64` instead of a `u128`.

## Solution

- Cast `solution_rate` to a `u64`
- Return a u64 from `getnetworksolps`, `getnetworkhashps`, and `getmininginfo`

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -154,7 +154,7 @@ pub trait GetBlockTemplateRpc {
         &self,
         num_blocks: Option<usize>,
         height: Option<i32>,
-    ) -> BoxFuture<Result<u128>>;
+    ) -> BoxFuture<Result<u64>>;
 
     /// Returns the estimated network solutions per second based on the last `num_blocks` before `height`.
     /// If `num_blocks` is not supplied, uses 120 blocks.
@@ -166,7 +166,7 @@ pub trait GetBlockTemplateRpc {
         &self,
         num_blocks: Option<usize>,
         height: Option<i32>,
-    ) -> BoxFuture<Result<u128>> {
+    ) -> BoxFuture<Result<u64>> {
         self.get_network_sol_ps(num_blocks, height)
     }
 }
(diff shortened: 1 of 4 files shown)


zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -614,7 +614,7 @@ where
                 _ => unreachable!("unmatched response to a solution rate request"),
             };
 
-            Ok(solution_rate)
+            Ok(solution_rate as u64)
```
teor2345: In general we avoid `as`, because it does silently wrapping integer casts, which can hide bugs.

Overflows are almost impossible in this situation, so we can do:
```suggestion
            Ok(solution_rate.try_into().expect("per-second solution rate always fits in u64"))
```

teor2345 (dismissed):
Thanks, this looks good.

`as` would be ok in this specific case, but since this module has consensus rules in it, we might want to avoid using it entirely here.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: (Otherwise we'd add a comment explaining why the cast is ok, but we try to implement those kinds of comments as `expect()`, so the code checks them.)

teor2345 (dismissed):
Thanks!

Does this also need a Cargo.lock update?
(CI will check that for us.)

teor2345:
Oh, sorry about the rustfmt in my suggestion.

teor2345 (approved):

arya2:
> Does this also need a Cargo.lock update? (CI will check that for us.)

Nope, it probably doesn't condition any dependencies on it.

> Oh, sorry about the rustfmt in my suggestion.

No worries, thank you for the note and for the quick review.

PR #5835, opened 2022-12-09
change(codec): Set relay to true if the last VersionMessage byte is omitted

### LINKED ISSUE #5713: relay (fRelayTxes) field when parse Message::Version from the remote node should be optional
## Required Changes

If the `Version` message bytes end at the `start_height` field, Zebra currently returns an `ErrorKind::UnexpectedEof`.
Instead of returning that error for that field, set the `relay` field to `true`.

If the message doesn't include enough bytes for any of the other fields, Zebra should continue to return `ErrorKind::UnexpectedEof`.

## Bug Report

The `relay` field read from the `Message::Version` should be optional as it implemented in `zcashd`:

https://github.com/zcash/zcash/blob/6284f3f8b08aa1f5fcaa27a8a4088fb204364401/src/main.cpp#L6505-L6510

As we are see here:
```cpp
        {
            LOCK(pfrom->cs_filter);
            if (!vRecv.empty())
                vRecv >> pfrom->fRelayTxes; // set to true after we get the first filter* message
            else
                pfrom->fRelayTxes = true;
        }
```
If `fRelayTxes` is not passed in version message we assume it `true`, but in the `zebrad` reading relay from the `Message::Version` is mandatory:

https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/protocol/external/codec.rs#L483

```rust
fn read_version<R: Read>(&self, mut reader: R) -> Result<Message, Error> {
        // Clippy 1.64 is wrong here, this lazy evaluation is necessary, constructors are functions. This is fixed in 1.66.
        #[allow(clippy::unnecessary_lazy_evaluations)]
        Ok(VersionMessage {
            version: Version(reader.read_u32::<LittleEndian>()?),
            // Use from_bits_truncate to discard unknown service bits.
            services: PeerServices::from_bits_truncate(reader.read_u64::<LittleEndian>()?),
            timestamp: Utc
                .timestamp_opt(reader.read_i64::<LittleEndian>()?, 0)
                .single()
                .ok_or_else(|| Error::Parse("version timestamp is out of range for DateTime"))?,
            address_recv: AddrInVersion::zcash_deserialize(&mut reader)?,
            address_from: AddrInVersion::zcash_deserialize(&mut reader)?,
            nonce: Nonce(reader.read_u64::<LittleEndian>()?),
            user_agent: String::zcash_deserialize(&mut reader)?,
            start_height: block::Height(reader.read_u32::<LittleEndian>()?),
            relay: match reader.read_u8()? {
                0 => false,
                1 => true,
                _ => return Err(Error::Parse("non-bool value supplied in relay field")),
            },
        }
        .into())
    }
```
If `relay` was unable to read here bcz it's not passed `read_version` will return with:
```
Err(error::const_io_error!(ErrorKind::UnexpectedEof, "failed to fill whole buffer"))
```
But `zcashd` default behavior is set it to `true`. 

To reproduce all of these we can start instance of `zebrad` and try to connect to it from `zcashd` qa `test_framework`. As the `class msg_version` haven't `relay` (`fRelayTxes`) it doesn't serialize it:

https://github.com/zcash/zcash/blob/master/qa/rpc-tests/test_framework/mininode.py#L1464

So, connection from the Python framework to `zebrad` will looks like this:
```
DEBUG:NodeConn(127.0.0.1:17770):Recv msg_version(nVersion=170011 nServices=1 nTime=Thu Nov 24 02:05:00 2022 addrTo=CAddress(nServices=1 ip=127.0.0.1 port=44424) addrFrom=CAddress(nServices=1 ip=0.0.0.0 port=17770) nNonce=0x3ACF450EA438557A strSubVer=b'/Zebra:1.0.0-beta.13/' nStartingHeight=0)
DEBUG:NodeConn(127.0.0.1:17770):Send msg_verack()
DEBUG:NodeConn(127.0.0.1:17770):MiniNode: Closing Connection to 127.0.0.1:17770...
```
I.e. closed with `HandshakeError::ConnectionClosed` on `zebrad` side. But the `zcashd` behaviour on the same connection with absent `relay` (`fRelayTxes`) in `version` p2p message will be different, connection will remains active, bcz in `zcashd` this param is `optional`.

arya2:
## Motivation

We want to make the relay field on `VersionMessage` optional.

Closes #5713.

## Solution

- Set `relay` field on `VersionMessage` to true when there's an `UnexpectedEof` error

Related changes: moves codec tests to their own module.
  
## Review

### Reviewer Checklist


teor2345:
> I'm having difficulty writing the test for this, ideally it might read `HEADER_LEN` bytes first, decrement `body_len` by 1 via a test impl method, and then continue reading the frame as usual, but I'm not sure how to do that yet. I'm open to alternative designs.

One possible test:
- parse a Version message with the relay field byte set to true/false, and check it is true/false
- parse a Version message with no relay field byte, and check it is true

The second version message should be one byte shorter than the other one.

(There's no need to check for unread bytes, because the true/false test makes sure that the relay byte was read.)

arya2:
> * parse a Version message with the relay field byte set to true/false, and check it is true/false

There's a test for this part here https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/isolated/tests/vectors.rs#L177

> * parse a Version message with no relay field byte, and check it is true

Ah, I was overcomplicating this part by trying to test it with a stream, I'll use `Codec` directly.

zebra-network/src/protocol/external/codec/tests.rs
```diff
@@ -0,0 +1,377 @@
...
+        fr.next()
+            .await
+            .expect("a next message should be available")
+            .expect("message should decode with the msg body size as max allowed value")
+    });
+}
+
+/// Check that the version test vector deserializes correctly without the relay byte
+#[test]
+fn version_message_omitted_relay() {
```
arya2: This is the only new code in this file, the rest was moved here from `codec.rs`.

zebra-network/src/protocol/external/codec.rs
```diff
@@ -24,6 +24,9 @@ use zebra_chain::{
 
 use crate::constants;
 
+#[cfg(test)]
+mod tests;
```
teor2345: Nitpick: We usually put the module declarations after the imports. (`use super` is an import.)

zebra-network/src/protocol/external/codec.rs
```diff
@@ -480,10 +483,11 @@ impl Codec {
...
             start_height: block::Height(reader.read_u32::<LittleEndian>()?),
-            relay: match reader.read_u8()? {
-                0 => false,
-                1 => true,
-                _ => return Err(Error::Parse("non-bool value supplied in relay field")),
+            relay: match reader.read_u8() {
+                Ok(val @ 0..=1) => val == 1,
+                Ok(_) => return Err(Error::Parse("non-bool value supplied in relay field")),
+                Err(err) if err.kind() == std::io::ErrorKind::UnexpectedEof => true,
+                Err(err) => Err(err)?,
```
teor2345, line 496: Please document that the `relay` field is optional, and link to the protocol reference:
https://developer.bitcoin.org/reference/p2p_networking.html#version

It might also be worth noting that `zcashd` only requires up to `address_recv`:
https://github.com/zcash/zcash/blob/11d563904933e889a11d9685c3b249f1536cfbe7/src/main.cpp#L6490-L6507

(But this is from pre-2010 Bitcoin, so we don't need to be compatible with it, and it is out of scope.)

zebra-network/src/protocol/external/codec/tests.rs
```diff
@@ -0,0 +1,366 @@
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
```
teor2345: Nitpicks:
- We usually put fixed test vectors in a `tests/vectors.rs` module
- Please add a short module doc comment

zebra-network/src/protocol/external/codec/tests.rs
```diff
@@ -0,0 +1,366 @@
...
+        .write_body(&version, &mut (&mut bytes).writer())
+        .expect("encoding should succeed");
+    bytes.truncate(bytes.len() - 1);
+
+    let relay = match codec.read_version(Cursor::new(&bytes)) {
+        Ok(Message::Version(VersionMessage { relay, .. })) => relay,
+        err => panic!("bytes should successfully decode to version message, got: {err:?}"),
+    };
+
+    assert!(relay, "relay should be true when omitted from message");
```
teor2345: Looks good!

Can you please add a test with a relay field/byte that is `false`, to catch bugs where we always set that field to `true`?

(I don't think this code has that bug, but we might refactor it in future.)

teor2345 (commented):
This looks good.

We could merge it as it is, but I'd like to improve test coverage, and document the specification we're implementing to.

zebra-network/src/protocol/external/codec.rs
arya2: Fixed in [`10c405c`](https://github.com/ZcashFoundation/zebra/pull/5835/commits/10c405ca251398f016e76909ecc970294bcb43d6).

zebra-network/src/protocol/external/codec.rs
arya2: Added in [`10c405c`](https://github.com/ZcashFoundation/zebra/pull/5835/commits/10c405ca251398f016e76909ecc970294bcb43d6) and [`b39b18c`](https://github.com/ZcashFoundation/zebra/pull/5835/commits/b39b18cf8a96c8df77d47e6686a03f033a040ffb).

zebra-network/src/protocol/external/codec/tests.rs
arya2: Fixed in [`10c405c`](https://github.com/ZcashFoundation/zebra/pull/5835/commits/10c405ca251398f016e76909ecc970294bcb43d6).

zebra-network/src/protocol/external/codec/tests.rs
arya2: Added in [`10c405c`](https://github.com/ZcashFoundation/zebra/pull/5835/commits/10c405ca251398f016e76909ecc970294bcb43d6).

teor2345 (approved):
Thanks for the extra test and docs!

PR #5840, opened 2022-12-11
fix(rpc): Wait for 3 minutes to check Zebra is synced to the tip, rather than 2

teor2345:
## Motivation

When I have been testing the `getblocktemplate` RPC, Zebra sometimes reports it is synced before it is actually at the tip. This will be bad for miners, but could also impact other RPCs and general Zebra usage.

This is a partial fix for #4649.

## Solution

Wait for 3 syncs to check that Zebra is synced, rather than 2.

## Review

Anyone can review this PR, it is a routine fix.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

We might want to fix the rest of #4649, or this could be enough for now.

PR #5841, opened 2022-12-12
change(mempool): Re-verify mempool transactions after a chain fork, rather than re-downloading them all

teor2345:
## Motivation

Zebra drops all its mempool transactions when there is a chain fork. But it would be faster and put less load on the network if we re-verified them instead.

I discovered this bug while testing the `getblocktemplate` RPC.

We need to make sure we clear the mempool, then re-verify all the stored or pending transactions.

Otherwise, we could break a consensus rule.

## Solution

- Re-verify mempool transactions after a fork, rather than re-downloading them all
- Update tests for mempool transaction re-verification after forks
- [Update README based on recent mitigations for some issues](https://github.com/ZcashFoundation/zebra/commit/e086e90c4c52b2d3f12dd389e85c8ac47fcd003a)
    - also includes some mitigations from PR #5840

Related bug fixes:
- [Move Drop from mempool::ActiveState to mempool::Downloads, to avoid bugs](https://github.com/ZcashFoundation/zebra/commit/a7f820d80a55247ac1842374254e446a60f4cde4)
    - In the previous design, we could drop the mempool Downloads without cancelling all the tasks

Related cleanups:
- Use a marker struct for cancelling mempool tasks
- Make mempool proptests easier to debug

This doesn't need to be behind a feature, because it's a bug fix for the production mempool as well.

## Review

This is a routine bug fix.

### Reviewer Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,19 +235,13 @@ So Zebra's state should always be valid, unless your OS or disk hardware is corr
 
 There are a few bugs in Zebra that we're still working on fixing:
 
-- Zebra falsely estimates that it's close to the tip when the network connection goes down [#4649](https://github.com/ZcashFoundation/zebra/issues/4649)
-
-  - One of the consequences of this issue is that Zebra might add unwanted load
-    to other peers when the connection goes back up. This load will last only
-    for a short period of time because Zebra will quickly find out that it's
-    still not close to the tip.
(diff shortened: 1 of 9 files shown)


oxarbitrage (approved):
Looks good, thanks for the fix.

PR #5843, opened 2022-12-12
feature(rpc): Add basic long polling support to the `getblocktemplate` RPC

teor2345:
## Motivation

This PR implements basic long polling support, checking for changes by re-fetching from the sync status, state and mempool.

Part of ticket #5720.

We need to hold the RPC request open until the response will change. This means running a loop and a `select!{}` in a future.

## Solution

Implement inefficient long polling by checking the state/mempool/syncer, waiting 5 seconds, then checking it again. Also wait for the max time.

Related changes:
- Change `impl From<LongPollInput> for LongPollId` into a named method
- Rename the `long_poll_id` field
- Cleanup some comments

## Review

This is ready for review, I'm doing some manual testing now. Automated testing will happen in the next PR.

Efficiency and refactors are out of scope for this PR, I'd like to focus on getting some kind of working implementation merged.

### Reviewer Checklist

## Follow Up Work

- medium priority: check for state changes efficiently
- low priority: check for sync status and mempool changes efficiently

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -1,8 +1,8 @@
 //! RPC methods related to mining only available with `getblocktemplate-rpcs` rust feature.
 
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
-use futures::{FutureExt, TryFutureExt};
+use futures::{future::OptionFuture, FutureExt, TryFutureExt};
 use jsonrpc_core::{self, BoxFuture, Error, ErrorCode, Result};
 use jsonrpc_derive::rpc;
 use tower::{buffer::Buffer, Service, ServiceExt};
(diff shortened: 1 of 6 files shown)


zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -333,40 +333,113 @@ where
...
+                    fetch_state_tip_and_local_time(state.clone()).await?;
 
-            // Fetch the mempool data for the block template:
-            // - if the mempool transactions change, we might return from long polling.
-            let mempool_txs = fetch_mempool_transactions(mempool).await?;
+                // Fetch the mempool data for the block template:
+                // - if the mempool transactions change, we might return from long polling.
+                //
+                // TODO:
+                // - add a `MempoolChange` type with an `async changed()` method.
```
arya2, line 374: We should use `mempool_transaction_receiver` to check if something is inserted into the mempool. https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/commands/start.rs#L163

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -333,40 +333,113 @@ where
             // These checks always have the same result during long polling.
             let miner_address = check_miner_address(miner_address)?;
 
+            let mut client_long_poll_id = None;
             if let Some(parameters) = parameters {
-                check_block_template_parameters(parameters)?;
+                check_block_template_parameters(&parameters)?;
+
+                client_long_poll_id = parameters.long_poll_id;
```
arya2, line 340: I think we should block this method here by awaiting `wait_for_max_time`, a changed `latest_chain_tip`, or a changed `mempool_transaction_receiver`, and leave the rest of this function as it is.

teor2345: I'm not sure if this would work the way it is meant to?

The long polling spec says that we need to return immediately if the client long poll ID is different from the current server long poll ID:
> This request SHOULD NOT be processed nor answered by the server until it wishes to replace the current block data as identified by the "longpollid".

Using watch channels is also out of scope for this PR, see the TODOs below.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Using watch channels is out of scope for this PR, see the TODOs.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: > I'm not sure if this would work the way it is meant to?

Ah, you're right, it needs to check that the long poll id is the same first.

teor2345:
This is ready for review, I'm doing some manual testing now. Automated testing will happen in the next PR.

teor2345:
A bunch of google cloud jobs seemed to have failed setup with:
> ssh: connect to host 104.198.199.90 port 22: Connection timed out

https://github.com/ZcashFoundation/zebra/actions/runs/3689513976/jobs/6246166936#step:11:78

teor2345:
@mergifyio update

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -333,40 +333,113 @@ where
...
+                // - add a `MempoolChange` type with an `async changed()` method.
+                // - if we are long polling, pause between state and mempool,
+                //   to allow transactions to re-verify (only works after PR #5841)
+                let mempool_txs = fetch_mempool_transactions(mempool.clone()).await?;
+
+                // - Long poll ID calculation
+                let server_long_poll_id = LongPollInput::new(
+                    chain_tip_and_local_time.tip_height,
+                    chain_tip_and_local_time.tip_hash,
+                    chain_tip_and_local_time.max_time,
```
arya2, line 383: Do we want to use the max time from the client's long poll id?

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -333,40 +333,113 @@ where
...
+                    mempool_txs.iter().map(|tx| tx.transaction.id),
+                )
+                .generate_id();
+
+                // The loop finishes if:
+                // - the client didn't pass a long poll ID,
+                // - the server long poll ID is different to the client long poll ID, or
+                // - the previous loop iteration waited until the max time.
+                if Some(&server_long_poll_id) != client_long_poll_id.as_ref() || max_time_reached {
```
arya2, line 392: I think this would return a template with a `max_time` in the past if `max_time_reached` is true but the long poll id is the same.

arya2 (commented):
This looks like it should work.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: I think that would make the next poll ignore any changes in the block header due to a new `max_time` from the state?

The `max_time` can change if the local clock is close to or past the minimum difficulty time on testnet. (It doesn't matter on mainnet.)

teor2345: All the gory details are here:
https://github.com/ZcashFoundation/zebra/blob/4fb2417adc70f57f42e9df4c8467279058b743d7/zebra-state/src/service/read/difficulty.rs#L263-L276

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Consensus rules can't depend on the local node time, because they need to be global. So it's definitely unusual, but I think it's still valid?

`max_time` is clamped to 90 minutes later than the `median-time-past` of recent blocks, based on this consensus rule:
https://github.com/ZcashFoundation/zebra/blob/4fb2417adc70f57f42e9df4c8467279058b743d7/zebra-state/src/service/read/difficulty.rs#L213-L219

(There is a standard 2 hour rule for the local node time, but standard rules can vary between implementations or configurations, so we don't use it to construct the block template.)

arya2: Ah, okay, so after the max time is reached there can't be any more valid changes to the template but the block can still be mined. This looks right then.

arya2 (approved):

teor2345:
@arya2 just checking why you marked this PR as "do-not-merge"?

It would be easier for me if this was part of the `main` branch and I didn't have to rebase my changes for PR #5862.

I'm also happy to sort out tests and any fixes in the next PR.

oxarbitrage:
I think is in case someone else wanted to review as @dconnolly is in the list as well.

oxarbitrage:
> I think is in case someone else wanted to review as @dconnolly is in the list as well.

But i think one approval is ok here so we can just remove the do not merge tag.

arya2:
> @arya2 just checking why you marked this PR as "do-not-merge"?

Just in case we wanted a second reviewer. I'll remove the label.

teor2345:
> > @arya2 just checking why you marked this PR as "do-not-merge"?
> 
> Just in case we wanted a second reviewer. I'll remove the label.

Always happy to get a second review, but I try not to block unless someone has committed to doing a review soon.

PR #5862, opened 2022-12-14
change(rpc): Return from long polling immediately when the chain tip changes

### LINKED ISSUE #5720: Support long polling in the getblocktemplate RPC (excluding mempool changes)

teor2345:
## Motivation

Miners want to start mining on the new tip immediately, to avoid wasted hash power and chain forks.

Closes #5720

Depends-On: #5843

This PR uses the block consensus rules - changing the previous block hash or block time makes previous work invalid.

It also changes the tokio watch channel API, and does some complex future implementations. (But mainly for testing.)

## Solution

- [Check for chain tip changes using a future, and return on error](https://github.com/ZcashFoundation/zebra/commit/29f78f1fcc65f96628f046f611c08a8ea5d2e111)
- [Ignore state changes before the most recent state fetch](https://github.com/ZcashFoundation/zebra/commit/a6062e625566f4c7671c30b1f4153d74dd6f0b60)
- [Add a submit old field to the getblocktemplate RPC during long polling](https://github.com/ZcashFoundation/zebra/commit/780a7531fb22eb515462a615129cd4b11a2b3a9e)

Related changes:
- [Add a best_tip_changed() method to trait ChainTip](https://github.com/ZcashFoundation/zebra/commit/2ad7964f1f51576e96f1435357f81e8b68c18d00)
- Add a mark as changed method to ChainTip

## Review

Anyone can review this PR, it's mostly changes to trait implementations and a `select!{}` loop.

### Reviewer Checklist

## Follow Up Work

I think we're done here, we're doing what miners have told us is their top priority.


teor2345:
I need to write tests before we merge this PR.

teor2345:
The test failures are in the snapshots:
>          38 │+  "submitold": null

https://github.com/ZcashFoundation/zebra/actions/runs/3691441568/jobs/6249438736#step:14:2473

`submitold` should be missing when it's `None`, rather than `null`. I'll fix this tomorrow.

I think I need to add `#[serde(skip_serializing_if = "Option::is_none()")]` and `#[serde(default = "Option::None")]`. And then check it doesn't serialize as `Some(true)`:
https://serde.rs/field-attrs.html#skip_serializing_if
https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html

mpguerra:
Can we evaluate this change in terms of how it will affect the code to be audited? 

We had talked about keeping the getblocktemplate work self contained as much as possible so it didn't affect the stable release candidate series. This change might be fine as it's just making additions and not changing existing functionality in zebra-state and zebra-chain but I would just like us to be mindful of this.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -390,14 +397,51 @@ where
...
                 if Some(&server_long_poll_id) != client_long_poll_id.as_ref() || max_time_reached {
-                    break (server_long_poll_id, chain_tip_and_local_time, mempool_txs);
+                    let mut submit_old = client_long_poll_id
+                        .as_ref()
+                        .map(|old_long_poll_id| server_long_poll_id.submit_old(old_long_poll_id));
+
+                    // On testnet, the max time changes the block difficulty,
+                    // so old shares are invalid.
+                    if max_time_reached {
+                        submit_old = Some(false);
```
arya2, line 409: Should this be `Some(if network == Network::Mainnet { true } else { false })` since mempool transactions haven't changed?

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -347,33 +349,38 @@ where
 
             // The loop returns the server long poll ID,
             // which should be different to the client long poll ID.
-            let (server_long_poll_id, chain_tip_and_local_time, mempool_txs) = loop {
+            let (server_long_poll_id, chain_tip_and_local_time, mempool_txs, submit_old) = loop {
```
arya2, line 352: You've probably already got this in mind but if we're using watch channels now, could we:
- add a `mempool_transaction_receiver` field on `GetBlockTemplateRpcs`
- move the contents of this loop until the check for if the `longpollid` has changed to a `fetch_template_data` fn
- call `mempool_transaction_receiver.borrow_and_update()`, `latest_chain_tip.mark_as_seen()`, and `fetch_template_data()`
- if the `longpollid` hasn't changed: wait for `mempool_transaction_receiver.changed()`, `latest_chain_tip.changed()`, or `wait_for_max_time` then call `fetch_template_data()` again 

Let me know if a suggestion PR would be helpful here.

zebra-state/src/service/chain_tip.rs
```diff
@@ -387,6 +391,38 @@ impl ChainTip for LatestChainTip {
         self.with_chain_tip_block(|block| block.transaction_hashes.clone())
             .unwrap_or_else(|| Arc::new([]))
     }
+
+    /// Returns when the state tip changes.
+    ///
+    /// Marks the state tip as seen.
+    #[instrument(skip(self))]
+    fn best_tip_changed(&mut self) -> BestTipChanged {
```
arya2, line 399: Nitpick: it seems simpler to use `std::pin::Pin<Box<dyn Future<Output = Result<(), watch::error::RecvError>> + Send + '_>>` (or add the async-trait crate and wait for the feature to hit rust stable https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#path-to-stabilization)

arya2 (commented):
Good call adding `submitold` and the `mark_as_seen` method on `ChainTip`.

arya2:
> Can we evaluate this change in terms of how it will affect the code to be audited?
> 
> We had talked about keeping the getblocktemplate work self contained as much as possible so it didn't affect the stable release candidate series. This change might be fine as it's just making additions and not changing existing functionality in zebra-state and zebra-chain but I would just like us to be mindful of this.

We should mark all of the new code here with `#[cfg(feature = "getblocktemplate-rpcs")]` too.

zebra-state/src/service/chain_tip.rs
```diff
@@ -387,6 +391,38 @@ impl ChainTip for LatestChainTip {
...
+        // so we don't need to use `with_chain_tip_block()` here.
+        //
+        // Clone the watch channel
+        let mut best_tip = self.receiver.clone();
+
+        // The cloned channel has the original change status,
+        // so we can mark the `self` channel as seen.
+        //
+        // Normally `changed()` does this, but we need the clone due to lifetimes.
+        self.mark_best_tip_seen();
```
arya2: I'm not sure this would work, because `changed()` might mark a message as seen after waiting to be notified of a new message if the version is initially the same.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Thanks, I already added optional TODOs for these (and for sync status, which we might also want to check using a watch channel.)

But this isn't a priority for miners, so I think it is optional and out of scope for now. If we want to do it, let's open a separate ticket, so we can schedule it separately.

I also think this might take significant changes to the rest of Zebra to get these channels to RPC initialisation in the start command. So we might want to delay it until after the audit. Or until a miner really wants it.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Max time is 90 minutes on mainnet, so it is extremely unlikely that there will be no new blocks and no new transactions in that time. (It's more likely the miner's network connection has gone down, so it doesn't matter what we do in that case.)

If it does happen, then I'm ok with an early return. Clients have to tolerate early returns anyway.

zebra-state/src/service/chain_tip.rs
teor2345: I don't really mind what we do here, but here are a few reasons I did it this way:

This is the way the standard library does it, see `FutureExt` for some similar future combinator types. So we know it will continue to work, even on nightly.

I didn't want to add an extra proc macro dependency just for one type. (It adds to compilation time, because it has to be compiled every time we compile Zebra.)

I tried spelling out the underlying type everywhere, and it didn't quite work, because it still carried over some of the input type generics in `MockChainTip::changed()`. So I had to use two levels of `BestTipChanged` to fully erase the type there.

I would also prefer to not repeat the exact type everywhere, so it's easy to change in the one place. And it's easier to read. If we didn't need to erase the extra generics, we could just use a type alias. But I don't think that would work here.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: That said, this is easy to fix or document, so I should do something here.

zebra-state/src/service/chain_tip.rs
teor2345: That seems like the expected behaviour to me, but I'm not sure if we're both thinking of the same situation.

Can you give me a specific scenario where that behaviour is a problem?

teor2345:
> > Can we evaluate this change in terms of how it will affect the code to be audited?
> > We had talked about keeping the getblocktemplate work self contained as much as possible so it didn't affect the stable release candidate series. This change might be fine as it's just making additions and not changing existing functionality in zebra-state and zebra-chain but I would just like us to be mindful of this.
> 
> We should mark all of the new code here with `#[cfg(feature = "getblocktemplate-rpcs")]` too.

There are only around 30 lines of added production code in this PR. The rest is comments, code that's already inside the feature, or tests.

I'm happy to put that code behind the feature after I write some tests. Just in case the API needs to change for the tests.

But I just wanted to let you know that it won't impact a significant amount of code. (A single hex serialisation implementation is larger, and we've been leaving them outside the feature.)

zebra-state/src/service/chain_tip.rs
teor2345: I'm happy to take a suggestion PR for this, but I'd like to avoid conflicts with the tests I need to write.

zebra-state/src/service/chain_tip.rs
arya2: > Can you give me a specific scenario where that behaviour is a problem?

If `best_tip_changed` is called when the newest value in the channel has been marked seen, it would mark the next value received as seen in `best_tip` but not in `self`.  Though I think it would work fine as it's used in `get_block_template` anyways.

Maybe just remove it and change the method name or update the doc comment?

teor2345: I think this is the correct behaviour, because the next value gets marked as seen in `self` just before we query the state for it.

teor2345: (`best_tip` is dropped when the `changed()` future completes, so it doesn't matter when it is marked as changed, because it is never used again.)

zebra-state/src/service/chain_tip.rs
arya2: > This is the way the standard library does it

> I didn't want to add an extra proc macro dependency just for one type. (It adds to compilation time, because it has to be compiled every time we compile Zebra.)

That makes sense, it would be simplifying marking the value as seen and I don't think we need to do that here anyways, so this is good.

zebra-state/src/service/chain_tip.rs
teor2345: I've checked the docs and comments here and in the RPC implementation, and I think they cover what we've discussed here.

But if you'd like me to make changes to make it clearer, I'm happy to do that - feel free to ask questions about specific comments, or make suggestions.

zebra-state/src/service/chain_tip.rs
```diff
@@ -387,6 +391,38 @@ impl ChainTip for LatestChainTip {
         self.with_chain_tip_block(|block| block.transaction_hashes.clone())
             .unwrap_or_else(|| Arc::new([]))
     }
+
+    /// Returns when the state tip changes.
+    ///
+    /// Marks the state tip as seen.
```
arya2: Maybe just this doc change? Though are you sure we need the clone due to lifetimes? It seems to compile okay when I change `best_tip.changed().err_into().await` to `self.receiver.changed().err_into().await`.

```suggestion
    /// Marks the state tip as seen when called (but not when BestTipChanged is polled)
```

teor2345:
> The test failures are in the snapshots:
> 
> > ```
> >      38 │+  "submitold": null
> > ```
> 
> https://github.com/ZcashFoundation/zebra/actions/runs/3691441568/jobs/6249438736#step:14:2473
> 
> `submitold` should be missing when it's `None`, rather than `null`. I'll fix this tomorrow.
> 
> I think I need to add `#[serde(skip_serializing_if = "Option::is_none()")]` and `#[serde(default = "Option::None")]`. And then check it doesn't serialize as `Some(true)`: https://serde.rs/field-attrs.html#skip_serializing_if https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html

In 9a34243ab2 I fixed the field serialization, made a separate snapshot that includes a submit old field, and changed the deserialized coinbase transaction format to RON. (Using JSON was a bit confusing due to the `null`s.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: I documented this in b312d5f52b.

zebra-state/src/service/chain_tip.rs
teor2345: You're right, I think I needed the old design when I required a `'static` lifetime in the `BestTipChanged` struct. But now it has a lifetime parameter, it's ok for it to depend on the lifetime of the receiver.

This is in e72d0ebac4, it simplifies the production and test code a lot.

teor2345:
> There are only around 30 lines of added production code in this PR.

Arya made a suggestion, and we're now down to around 20 lines of added production code. At this point, I don't think it matters if it is behind a feature. (But I'm happy to do that.)

arya2:
> At this point, I don't think it matters if it is behind a feature. (But I'm happy to do that.)

I don't think so either.

This looks good to me if it's ready for review.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -390,14 +397,53 @@ where
...
+                let wait_for_mempool_request = tokio::time::sleep(Duration::from_secs(
+                    GET_BLOCK_TEMPLATE_MEMPOOL_LONG_POLL_INTERVAL,
+                ));
+
+                // Return immediately if the chain tip has changed.
+                let wait_for_best_tip_change = latest_chain_tip.best_tip_changed();
+
+                // Wait for the maximum block time to elapse. This can change the block header
+                // on testnet. (On mainnet it can happen due to a network disconnection, or a
+                // rapid dropp in hash rate.)
```
```suggestion
                // rapid drop in hash rate.)
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -135,14 +135,30 @@ pub struct GetBlockTemplate {
...
+    /// > only relevant for long poll responses:
+    /// > indicates if work received prior to this response remains potentially valid (default)
+    /// > and should have its shares submitted;
+    /// > if false, the miner may wish to discard its share queue
+    ///
+    /// <https://en.bitcoin.it/wiki/BIP_0022#Optional:_Long_Polling>
+    ///
+    /// This field is not in `zcashd` or the Zcash RPC reference yet.
+    ///
+    /// In Zebra, `submit_old` is `false` when the tip block or max time is reached,
```
arya2:
```suggestion
    /// In Zebra, `submit_old` is `false` when the tip block changed or max time is reached,
```

teor2345:
> This looks good to me if it's ready for review.

I think it still needs more tests, even if they have to be manual tests, I want to do them.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Thanks, fixed!

teor2345:
I am happy with this PR, I did some tests as part of PR #5867. We can continue to test long polling in ticket #5686.

arya2 (approved):

PR #5866, opened 2022-12-15
Fix typo

rex4539:


teor2345 (approved):
Thanks!

teor2345:
Looks like we'll have to copy this branch into the ZF repository to get all the CI to run.

Maybe @gustavovalverde has a quick fix for getting Google Cloud to authenticate from a fork?

rex4539:
It's a simple typo fix. No need to run CI...

teor2345:
> It's a simple typo fix. No need to run CI...

Thanks, I understand and I agree, in this specific case.

But we automatically run CI and merge all our PRs. It takes admin access to override that, which is something we want to avoid. So I'm checking with the team if there's a quick way to get PRs from forks to work.

Another alternative is for us to add a separate merge workflow for fork PRs, which doesn't do the Google Cloud checks on the initial PR. (But does do them before we merge.)

oxarbitrage:
Replaced by https://github.com/ZcashFoundation/zebra/pull/5886 so we can marge because i don't think the issue with branches from outside zebra and CI will be fixed before at least mid january.

Thank you @rex4539

Issue #5685, opened 2022-11-21
Test the getblocktemplate RPC's response as a block template proposal in CI


arya2:
## Motivation

The acceptance test could confirm that the `getblocktemplate` response in 'template' mode will be accepted if it has a valid solution by submitting the same data to the `getblocktemplate` method in 'proposal' mode.

### Design

In the acceptance test that syncs until the tip and calls the `getblocktemplate` RPC method:
- Deserialize the response
- Convert the template into a block with empty `solution` and `nonce` fields
- Hex-encode the resulting block
- Call `getblocktemplate` in 'proposal' mode
- Assert a "null" response indicating successful validation (similar to the `submitblock` test)

teor2345:
> The acceptance test could confirm that the getblocktemplate response in 'template' mode will be accepted if it has a valid solution by submitting the same data to the getblocktemplate method in 'proposal' mode.

I'm not quite sure how this would work, do we need to mine a valid solution for the proposal to be valid?
Could we use `submitblock` in the test, rather than adding an extra unused `proposal` mode to `getblocktemplate`?

arya2:
> I'm not quite sure how this would work, do we need to mine a valid solution for the proposal to be valid? Could we use `submitblock` in the test, rather than adding an extra unused `proposal` mode to `getblocktemplate`?

It doesn't need a valid solution (it doesn't commit the block, just checks that it would be accepted if submitted with a valid solution), so a new block verifier request would skip the `check::equihash_solution_is_valid(&block.header)` [here](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/block.rs#L166) and use a `CheckContextualValidity` request instead of `CommitBlock`. Then if miners want to modify the template and then check that their modifications are still valid before mining, they could propose the new template.

> The block data MUST be validated and checked against the server's usual acceptance rules (excluding the check for a valid proof-of-work) - https://en.bitcoin.it/wiki/BIP_0023#Block_Proposal

teor2345:
Sounds good, can you update the ticket description with the spec details and your suggested design?

Sometimes they get lost in the comments!

mpguerra:
I don't think we should do this in CI until we can confirm it manually

mpguerra:
[...]

PR #5837, opened 2022-12-09
change(rpc): Refactor get block template RPC into stages

teor2345:
## Motivation

To implement long polling, we need to repeat some parts of the get block template RPC.

But currently the RPC function is very large, and in no specific order.

This refactor does not change any Zebra functionality, except that the long poll ID now tracks all mempool transactions.

## Solution

Functional Changes:
- [Split fetching mempool transactions out, include them in long poll id](https://github.com/ZcashFoundation/zebra/commit/c2d473bf3954cfebb5d8b9cb6b11bac007144a12)

Refactors:
- [Move and rename height_from_signed_int()](https://github.com/ZcashFoundation/zebra/commit/f08a54ff3685a3e038e69f1c89c7774a02ba3e27)
- [Move get_block_template() support functions to a submodule](https://github.com/ZcashFoundation/zebra/commit/38861d90545420bd37eee44fc5daa82fdce6b3de)
- [Split initial checks into their own functions](https://github.com/ZcashFoundation/zebra/commit/3de013d8b1d50a52bb9cd00ef8b7c47b46988edd)
- [Split state fetch into its own function](https://github.com/ZcashFoundation/zebra/commit/04d790cccedfb90120839250b24ff906d4eade3f)
- [Refactor coinbase generation](https://github.com/ZcashFoundation/zebra/commit/1a81275f0a86fd143e3f47700bef4ed1d77fcbd0)
- [Split default root calculation into a separate function](https://github.com/ZcashFoundation/zebra/commit/fe35d9b7439f5843962b66a503842ddde23085e0)
- [Split out a generate coinbase and roots function](https://github.com/ZcashFoundation/zebra/commit/bfd47ff2766a3f2d541bd937287cefb5caea1bde)
- [Use DateTime32 for getblocktemplate times](https://github.com/ZcashFoundation/zebra/commit/8d0f40e070ef5c1ef315b2706a2740f94507d23c)
- [Use typed difficulty fields rather than strings](https://github.com/ZcashFoundation/zebra/commit/a913c5adcb1ef92f19d99282c58c91aa61788bc6)
- [Move GetBlockTemplate construction into a method](https://github.com/ZcashFoundation/zebra/commit/fed5956024cbd6b25212a579bddc74dae68f66ae)

Cleanups:
- [Fix incorrect P2SH comments and logs](https://github.com/ZcashFoundation/zebra/commit/6e6daa1d81bc6efc676b75135c98409e98b6b097)
[...]

Issue #5803, opened 2022-12-06
Manually test the `getblocktemplate` RPC creates valid blocks


oxarbitrage:
Our final goal is to support mining pools to use zebra for their operations. This requires multiple rpc methods to be present.

There is a [discussion](https://github.com/ZcashFoundation/zebra/discussions/5804) open to exchange ideas on this.

## Details

From https://github.com/ZcashFoundation/zebra/discussions/5804#discussioncomment-4328539

I think we can do testing using a block proposal, which is a block without a proof of work.

Here's how we could do that:
- [x] Finish all the required getblocktemplate RPC fields
- [x] Create a block header from the RPC fields, with a zeroed-out equihash solution, and serialize it
- [x] Create the block transactions by combining all the coinbasetxn and transaction `data` fields
- [x] Send that block to `zcashd` using `getblocktemplate '{ "mode": "proposal", "data": "(hex data)" }'

`zcashd` has proposal support, but it's not documented. Here's the spec:
https://en.bitcoin.it/wiki/BIP_0023#Block_Proposal

## Related Work

- #5686
- #5803
- #5934
- #5935

teor2345:
@oxarbitrage please tag your tickets with A-rpc, so we can find them when we're looking for RPC work.

teor2345:
Please check your box when you have successfully run the `zcash-rpc-block-template-to-proposal` script from PR #5944, multiple times on mainnet and testnet:
- [x] @arya2 
- [ ] @oxarbitrage 
- [ ] @teor2345 

You'll need a synced `zcashd` instance and a synced Zebra instance. You can also test Zebra template to Zebra proposal if you just have Zebra.

We need to merge a few bug fixes to get the script to work all the time:
- [x] #5953 (fixes failures with multiple transactions)
- [x] #5925 (fixes failures on testnet using the min or max time test modes)
- [x] #5982 (fixes occasional failures due to duplicate transactions in the template)
- [ ] #5984 (fixes occasional failures due to invalid lock times, we haven't seen any of these yet)

teor2345:
I spent some time increasing the coverage of the test script in #5944, and making it easier to use.

I discovered two bugs during my testing:
- #5981 
- #5982 

Bug #5982 is a consensus rule bug that will block finishing off testing. (We can still check for other bugs until it is fixed.)

teor2345:
I am using these shell scripts to run continuous proposal tests until a test fails:
```sh
[...]

PR #5870, opened 2022-12-15
change(rpc): Add proposal capability to getblocktemplate

### LINKED ISSUE #5684: Add support for block proposals to `getblocktemplate` method
## Motivation

zcashd supports the `proposal` capability, so that if miners want to modify the template and then check that their modifications are still valid before mining, they could propose the new template.

This would also be useful for testing purposes, as we could propose the getblocktemplate's response and confirm that the block will be accepted if it has a valid solution.

This would require a `Request` enum in the block verifier with a variant for speculatively validating a block without checking its solution or committing it.

It also requires a new state request.

See #5630 for the state request.

### Specifications

> The block data MUST be validated and checked against the server's usual acceptance rules (excluding the check for a valid proof-of-work)

https://en.bitcoin.it/wiki/BIP_0022#appendix-example-rejection-reasons

### Designs

- Add a new block verifier request
  - Skip `check::equihash_solution_is_valid(&block.header)`
    (https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/block.rs#L166)
  - Use `CheckContextualValidity` request from #5630 instead of `CommitBlock`. 
- If `mode == GetBlockTemplateMode::Proposal` 
  - Deserialize `data` into the block
  - Call `BlockVerifier` with the new request variant instead of generating a template
  - Return null if successful or "rejected" if the block failed validation.

See zcashd impl: https://github.com/zcash/zcash/blob/master/src/rpc/mining.cpp#L538-L568
 
### Out Of Scope: 
- Detailed error handling

arya2:
## Motivation

We want to support using the getblocktemplate RPC in proposal mode to use in our acceptance test.

Closes #5684.

The new `CheckValidBlock` state request needs to check the block against all of the usual acceptance rules.

## Solution

- Add a state request that checks the contextual validity of a block by calling `validate_and_commit` with a clone of the non-finalized state
- Use a request enum in the chain/block verifiers with a `CheckProposal` variant that skips checking the solution and calls the state with `CheckValidBlock` instead of `CommitBlock`
- Check the `mode` and `data` fields of the `getblocktemplate` parameters
  - Deserialize the block bytes in the `data` field and call chain verifier with a `CheckProposal` request

Related changes:
- Fixes formatting in prop tests for checks in state service
- Accepts and ignores optional `workid` parameter

### Reviewer Checklist

## Follow Up Work

- #5685
- Detailed reject reasons (if needed)


teor2345:
@arya2 can you please rebase this PR on the latest `main` branch?

The diff currently shows changes from other PRs, which makes it hard to review.

zebra-consensus/src/block.rs
```diff
@@ -175,7 +177,10 @@ where
             // Do the difficulty checks first, to raise the threshold for
             // attacks that use any other fields.
             check::difficulty_is_valid(&block.header, network, &height, &hash)?;
-            check::equihash_solution_is_valid(&block.header)?;
+
+            if !request.is_proposal() {
+                check::equihash_solution_is_valid(&block.header)?;
```
teor2345, line 188: What is the specification for this check? Can you please quote it here?

Should block proposal block hashes:
1. satisfy the difficulty check without a valid equihash solution, or
2. not satisfy the difficulty check or the equihash solution

This code implements case 1, but case 2 allows more kinds of proposals.

In particular, case 1 requires the proposer to try multiple nonces until they find a block hash that satisfies the difficulty threshold. (The equihash solution doesn't need to be valid, but they still need to do a bit of hashing to make sure the block hash has enough leading zeroes.)

zebra-consensus/src/block.rs
```diff
@@ -279,19 +284,38 @@ where
...
-                .map_err(VerifyBlockError::Commit)?
-            {
-                zs::Response::Committed(committed_hash) => {
-                    assert_eq!(committed_hash, hash, "state must commit correct hash");
-                    Ok(hash)
-                }
-                _ => unreachable!("wrong response for CommitBlock"),
+
+            match request {
+                Request::Commit(_) => match state_service
```
teor2345: Optional: Since we're not using the outer `match` to deconstruct the type, it would simplify this code to use `Request::is_proposal()`.

zebra-consensus/src/block.rs
```diff
@@ -74,6 +70,10 @@ pub enum VerifyBlockError {
     // TODO: make this into a concrete type, and add it to is_duplicate_request() (#2908)
     Commit(#[source] BoxError),
 
+    #[error("unable to validate block after semantic verification")]
+    // TODO: make this into a concrete type
+    Validate(#[source] BoxError),
```
teor2345: All the rest of these errors apply to full validation, so it would be helpful to clearly distinguish them.

I think this description and name might help?


```rust
    #[error("unable to validate block proposal: failed semantic verification (proof of work is not checked for proposals)")]
    // TODO: make this into a concrete type
    ValidateProposal(#[source] BoxError),
```

zebra-consensus/src/block/request.rs
```diff
@@ -0,0 +1,39 @@
+//! Block verifier request type.
+
+use std::sync::Arc;
+
+use zebra_chain::block::Block;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// A request to the chain or block verifier
+pub enum Request {
+    /// Performs semantic validation then calls state with CommitBlock request
```
teor2345:
```suggestion
    /// Performs semantic validation, then asks the state to perform contextual validation and commit the block
```

zebra-consensus/src/block/request.rs
```diff
@@ -0,0 +1,39 @@
...
+    Commit(Arc<Block>),
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Performs semantic validation but skips checking the solution,
+    /// then calls the state with CheckBlockValid request
```
teor2345:
```suggestion
    /// Performs semantic validation but skips checking proof of work,
    /// then asks the state to perform contextual validation, excluding proof of work.
    /// Does not commit the block to the state.
```

zebra-consensus/src/block.rs
```diff
@@ -279,19 +284,38 @@ where
...
+                #[cfg(feature = "getblocktemplate-rpcs")]
+                Request::CheckProposal(_) => match state_service
+                    .ready()
+                    .await
+                    .map_err(VerifyBlockError::Validate)?
+                    .call(zs::Request::CheckBlockValidity(prepared_block))
+                    .await
+                    .map_err(VerifyBlockError::Validate)?
+                {
+                    zs::Response::Committed(committed_hash) => {
```
teor2345: This seems like the wrong response, because the block was not committed?

zebra-consensus/src/block/request.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub fn block(&self) -> Arc<Block> {
+        Arc::clone(match self {
+            Request::Commit(block) => block,
+
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            Request::CheckProposal(block) => block,
+        })
+    }
+
+    /// Checks if request is a proposal
```
teor2345:
```suggestion
    /// Returns `true` if the request is a proposal
```

zebra-consensus/src/chain.rs
```diff
@@ -167,14 +165,16 @@ where
         Poll::Ready(Ok(()))
     }
 
-    fn call(&mut self, block: Arc<Block>) -> Self::Future {
+    fn call(&mut self, request: Request) -> Self::Future {
+        let block = request.block();
+
         match block.coinbase_height() {
             Some(height) if height <= self.max_checkpoint_height => {
                 self.checkpoint.call(block).map_err(Into::into).boxed()
```
teor2345, line 186: What should happen if the checkpoint verifier is called with a proposal request?
Should we immediately return an error here?

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -316,7 +315,7 @@ where
     fn get_block_template(
         &self,
         parameters: Option<get_block_template::JsonParameters>,
-    ) -> BoxFuture<Result<GetBlockTemplate>> {
+    ) -> BoxFuture<Result<get_block_template::Response>> {
```
teor2345, line 318: Would you mind splitting this rename into a separate PR?

It will conflict with this change in PR #5867:
https://github.com/ZcashFoundation/zebra/pull/5867/files#diff-c0bec6265a8d68255f80da0e5f499035c5c0181cc218c3d68805094e2c7815f5R320-R324

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -326,256 +325,284 @@ where
...
+                    .call(zebra_consensus::Request::CheckProposal(Arc::new(block)))
+                    .await;
+
+                let response = chain_verifier_response
+                    .map(|_| get_block_template::ProposalResponse::Valid)
+                    .unwrap_or_else(|_| get_block_template::ProposalRejectReason::Rejected.into());
+
+                Ok(get_block_template::Response::ProposalMode(response))
+            } else {
+                // - One-off checks
```
teor2345: This change will conflict with PR #5867, can we avoid the `else` block in this PR?

I'm happy to do the code movement in a later PR. (We might also want to split out both template and proposal mode into functions as part of that refactor.)

As part of that change, we might also want to use separate futures for each mode, because that will make both futures smaller and more efficient.

```rust
                return Ok(get_block_template::Response::ProposalMode(response));
            }

            // - One-off checks
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -326,256 +325,284 @@ where
...
+                    server_long_poll_id,
+                    coinbase_txn,
+                    &mempool_txs,
+                    default_roots,
+                    submit_old,
+                );
+
+                Ok(get_block_template::Response::TemplateMode(Box::new(
+                    response,
+                )))
```
teor2345: Optional: to simplify this code, we could return the enum variant from `GetBlockTemplate::new()`, or add an `impl From<GetBlockTemplate> for get_block_template::Response`, and use `into()` here.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -326,256 +325,284 @@ where
...
-                // Fetch the state data and local time for the block template:
-                // - if the tip block hash changes, we must return from long polling,
-                // - if the local clock changes on testnet, we might return from long polling
+            let client_long_poll_id = parameters.as_ref().and_then(|params| params.long_poll_id.clone());
+
+            if let Some(HexData(block_proposal_bytes)) =
+                get_block_template::JsonParameters::block_proposal(parameters)?
+            {
+                let mut chain_verifier =
+                    chain_verifier.expect("chain_verifier is Some in proposal mode");
```
teor2345: Optional: Cloning is cheap, so I would just unconditionally do the clone. (In some cases, the panic machinery + `Option<T>` can be more expensive than always doing `T::clone()`.(

We can make it more efficient in a future PR, if we use different functions and futures for each kind of response.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
```diff
@@ -28,21 +28,70 @@ use crate::methods::get_block_template_rpcs::{
 
 pub use crate::methods::get_block_template_rpcs::types::get_block_template::*;
 
+use super::types::hex_data::HexData;
```
teor2345: Nitpick: items from this module are already imported above with `use crate::methods::get_block_template_rpcs::{ ... types::{ ... }, };`

zebra-consensus/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 
 [features]
 default = []
-getblocktemplate-rpcs = []
+getblocktemplate-rpcs = ["zebra-state/getblocktemplate-rpcs"]
```
teor2345: This is an existing dependency bug: every dependency's feature must be activated here.

For example, see:
https://github.com/ZcashFoundation/zebra/blob/80a6d3cdab2a9dc68e27ecfa729ba864394becb4/zebrad/Cargo.toml#L21-L28

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
```diff
@@ -28,21 +28,70 @@ use crate::methods::get_block_template_rpcs::{
...
-            message: "\"proposal\" mode is currently unsupported by Zebra".to_string(),
-            data: None,
-        });
+impl JsonParameters {
+    /// Checks that `data` is omitted in `Template` mode or provided in `Proposal` mode,
+    ///
+    /// Returns an error if there's a mismatch between the mode and whether `data` is provided.
+    /// Returns Ok(Some(data)) with the block proposal hexdata if in `Proposal` mode.
+    /// Returns Ok(None) if in `Template` mode.
+    pub fn block_proposal(parameters: Option<Self>) -> Result<Option<HexData>> {
```
teor2345: I'm finding this method hard to understand and review.

It seems to do two or three different things:
- In template mode, check that the proposal parameters are valid (that is, both not provided)
- In proposal mode, check that the proposal parameters are valid (that is, both provided)
- In proposal mode, return the proposal parameters

Looking at its name, it is not obvious that it needs to be called in both modes. It looks like a method that gets the data, regardless of any parameter checks.

Optional: We could also split it into separate check and get methods, if you want. And I wonder if taking `&self` would make this method easier to understand and use.

```rust
    /// Checks that `data` is omitted in `Template` mode or provided in `Proposal` mode,
    /// then returns the proposal data.
    ///
    /// Returns an error if there's a mismatch between the mode and whether `data` is provided.
    /// Returns `Ok(Some(hexdata))` with the block proposal `hexdata` if in `Proposal` mode.
    /// Returns `Ok(None)` if in `Template` mode.
    pub fn check_block_proposal(parameters: Option<Self>) -> Result<Option<HexData>> {
```

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
```diff
@@ -28,21 +28,70 @@ use crate::methods::get_block_template_rpcs::{
...
+                          omitted in \"template\" mode"
+                    .to_string(),
+                data: None,
+            }),
+        }
     }
 
-    Ok(())
+    /// Checks if `mode` parameters is `Proposal`.
+    pub fn is_proposal_mode(parameters: &Option<Self>) -> bool {
```
teor2345: Optional: I wonder if taking `&self` would make this method easier to understand and use.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -234,3 +234,60 @@ impl GetBlockTemplate {
...
+    /// Block was not successfully submitted, return error
+    ErrorResponse {
+        /// Reason the proposal was invalid as-is
+        reject_reason: ProposalRejectReason,
+
+        /// The getblocktemplate RPC capabilities supported by Zebra.
+        capabilities: Vec<String>,
+    },
+
+    /// Block successfully proposed, returns null
```
teor2345: This seems to serialize as `"valid"`, not `null`.

To automatically serialize as `null`, use an `Option<ErrorResponse>` instead of the `ProposalResponse` struct.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -234,3 +234,60 @@ impl GetBlockTemplate {
...
+#[serde(untagged, rename_all = "kebab-case")]
+pub enum ProposalResponse {
+    /// Block was not successfully submitted, return error
+    ErrorResponse {
+        /// Reason the proposal was invalid as-is
+        reject_reason: ProposalRejectReason,
+
+        /// The getblocktemplate RPC capabilities supported by Zebra.
+        capabilities: Vec<String>,
+    },
```
teor2345: Please link to the documentation that lists the fields in this JSON-RPC response.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/parameters.rs
```diff
@@ -11,7 +11,6 @@ pub enum GetBlockTemplateRequestMode {
     Template,
 
     /// Indicates a request to validate block data.
-    /// Currently unsupported and will return an error.
```
teor2345, line 14: Please also remove the comment on the enum that says this, and search for similar comments.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -234,3 +234,60 @@ impl GetBlockTemplate {
...
+    ProposalMode(ProposalResponse),
+}
+
+impl From<ProposalRejectReason> for ProposalResponse {
+    fn from(reject_reason: ProposalRejectReason) -> Self {
+        // Convert default values
+        let capabilities: Vec<String> = GET_BLOCK_TEMPLATE_CAPABILITIES_FIELD
+            .iter()
+            .map(ToString::to_string)
+            .collect();
```
teor2345: Optional: what happens if Zebra is configured in a way that changes the default capabilities?

zebra-state/src/error.rs
```diff
@@ -55,6 +55,11 @@ pub enum ValidateContextError {
     #[non_exhaustive]
     NotReadyToBeCommitted,
 
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    #[error("block parent not found in non-finalized best chain")]
```
teor2345: Do we need to mention that it is for a proposal here?

zebra-state/src/request.rs
```diff
@@ -790,6 +801,15 @@ pub enum ReadRequest {
         /// Optionally estimate the network speed at the time when a certain block was found
         height: Option<block::Height>,
     },
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Performs contextual validation of the given block.
+    ///
+    /// It is the caller's responsibility to perform semantic validation.
```
```suggestion
    /// Performs contextual validation of the given block, but does not commit it to the state.
    ///
    /// It is the caller's responsibility to perform semantic validation.
    /// (The caller does not need to check proof of work for block proposals.)
```

zebra-state/src/request.rs
```diff
@@ -558,6 +558,15 @@ pub enum Request {
...
     /// Returns [`Response::ValidBestChainTipNullifiersAndAnchors`]
     CheckBestChainTipNullifiersAndAnchors(UnminedTx),
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Performs contextual validation of the given block.
+    ///
+    /// It is the caller's responsibility to perform semantic validation.
+    ///
+    /// Returns [`Response::ValidBlock`] with the hash of the block when successful, or an error if
+    /// the block fails contextual validation.
```
```suggestion
    /// Performs contextual validation of the given block, but does not commit it to the state.
    ///
    /// See `[ReadRequest::CheckBlockValidity]` for details.
```

zebra-state/src/response.rs
```diff
@@ -59,6 +59,10 @@ pub enum Response {
     ///
     /// Does not check transparent UTXO inputs
     ValidBestChainTipNullifiersAndAnchors,
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Response to [`Request::CheckBlockValidity`](crate::Request::CheckBlockValidity)
+    ValidBlock(block::Hash),
```
teor2345: Is there a way to make it clear that this is not actually guaranteed to be a valid block?
We might also want to update the documentation on this response variant.

```rust
    ValidBlockProposal(block::Hash),
```

zebra-state/src/response.rs
```diff
@@ -137,6 +141,10 @@ pub enum ReadResponse {
     #[cfg(feature = "getblocktemplate-rpcs")]
     /// Response to [`ReadRequest::SolutionRate`](crate::ReadRequest::SolutionRate)
     SolutionRate(Option<u128>),
+
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Response to [`ReadRequest::CheckBlockValidity`](crate::ReadRequest::CheckBlockValidity)
+    ValidBlock(block::Hash),
```
teor2345: Same here, it would be good to be clear that these blocks aren't valid.

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
     fn latest_non_finalized_state(&self) -> NonFinalizedState {
         self.non_finalized_state_receiver.cloned_watch_data()
     }
+
+    /// Check the contextual validity of a block in the best chain
```
teor2345: Which consensus rules are checked here? Do we skip any?

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
...
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    #[allow(clippy::unwrap_in_result)]
+    #[tracing::instrument(level = "debug", skip_all)]
+    fn check_best_chain_contextual_validity(
+        &self,
+        prepared: &PreparedBlock,
+    ) -> Result<(), crate::ValidateContextError> {
```
teor2345: This seems like it is a duplicate of existing code that implements the consensus rules.

So it could easily go out of sync as we change the rules. (Or be called accidentally from somewhere we want to check the full set of block rules before we commit.)

Is there a way to re-use the existing consensus rule checks instead?

For example, we could:
1. Create a deep cloned non-finalized state, which only contains the best chain (or nothing, if there is no best chain)
2. Call `commit_block()` / `commit_new_chain()` on that non-finalized state (this automatically deep clones the best chain if its `Arc` is still shared)
3. Drop the modified finalized state clone, which drops the modified chain
4. Add an argument to `commit_block()` / `commit_new_chain()` that skips the metrics updates, because the block isn't actually committed

zebra-state/src/service.rs
```diff
@@ -1684,6 +1787,34 @@ impl Service<ReadRequest> for ReadStateService {
...
+
+                let state = self.clone();
+
+                // # Performance
+                //
+                // Allow other async tasks to make progress while concurrently reading blocks from disk.
+                let span = Span::current();
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        state.check_best_chain_contextual_validity(&prepared)?;
```
teor2345: Optional: Can we have this function explicitly take an `Arc<Chain>` and a `&ZebraDb`, like all the other `read_*` functions?

It's easier to review the concurrency if we can see exactly what parts of the state the function is using.

zebra-state/src/service.rs
```diff
@@ -1684,6 +1787,34 @@ impl Service<ReadRequest> for ReadStateService {
...
+                        timer.finish(
+                            module_path!(),
+                            line!(),
+                            "ReadRequest::CheckContextualValidity",
+                        );
+
+                        Ok(ReadResponse::ValidBlock(prepared.hash))
+                    })
+                })
+                .map(|join_result| join_result.expect("panic in ReadRequest::ChainInfo"))
```
```suggestion
                .map(|join_result| join_result.expect("panic in ReadRequest::CheckBlockValidity"))
```

zebra-state/src/service.rs
```diff
@@ -1684,6 +1787,34 @@ impl Service<ReadRequest> for ReadStateService {
...
+                // Allow other async tasks to make progress while concurrently reading blocks from disk.
+                let span = Span::current();
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        state.check_best_chain_contextual_validity(&prepared)?;
+                        // The work is done in the future.
+                        timer.finish(
+                            module_path!(),
+                            line!(),
+                            "ReadRequest::CheckContextualValidity",
```
```suggestion
                            "ReadRequest::CheckBlockValidity",
```

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
...
+        if best_chain.non_finalized_tip_hash() != prepared.block.header.previous_block_hash {
+            Err(crate::ValidateContextError::NotReadyToBeValidated)?
+        }
+
+        check::initial_contextual_validity(
+            self.network,
+            &self.db,
+            &latest_non_finalized_state,
+            prepared,
+        )?;
```
teor2345: Does this check the best chain?

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
...
+    ) -> Result<(), crate::ValidateContextError> {
+        let latest_non_finalized_state = self.latest_non_finalized_state();
+
+        let Some(best_chain) = latest_non_finalized_state.best_chain() else {
+            Err(crate::ValidateContextError::NotReadyToBeValidated)?
+        };
+
+        if best_chain.non_finalized_tip_hash() != prepared.block.header.previous_block_hash {
+            Err(crate::ValidateContextError::NotReadyToBeValidated)?
+        }
```
teor2345: Should this be a "not a child of the best chain tip" error?

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
...
+    #[tracing::instrument(level = "debug", skip_all)]
+    fn check_best_chain_contextual_validity(
+        &self,
+        prepared: &PreparedBlock,
+    ) -> Result<(), crate::ValidateContextError> {
+        let latest_non_finalized_state = self.latest_non_finalized_state();
+
+        let Some(best_chain) = latest_non_finalized_state.best_chain() else {
+            Err(crate::ValidateContextError::NotReadyToBeValidated)?
+        };
```
teor2345: We should be able to validate directly off the finalized tip, if we want to. All the functions that we call below support it.

If that's not what we want or need, please document that design choice in the function docs, and why we made it.

zebra-state/src/service.rs
```diff
@@ -803,6 +803,91 @@ impl ReadStateService {
...
+                    contextual.block,
+                    self.network,
+                    &best_chain.history_tree,
+                ));
+            });
+        });
+
+        block_commitment_result.expect("scope has finished")?;
+        sprout_anchor_result.expect("scope has finished")?;
+
```
teor2345: We're missing all the consensus checks that are done by this code:
https://github.com/ZcashFoundation/zebra/blob/80a6d3cdab2a9dc68e27ecfa729ba864394becb4/zebra-state/src/service/non_finalized_state.rs#L306-L308

For example, we need to apply the block chain value change to the chain value pools, and make sure none of the value pools are negative.

teor2345 (commented):
Looks good!

This is a large and risky change, but if we make some changes we can do it without breaking production code, or diverging from the consensus rules.

zebra-state/src/service.rs
arya2: Yeah, it checks `best_chain.non_finalized_tip_hash() != prepared.block.header.previous_block_hash` first and then only accesses the non-finalized state through `previous_block_hash`, I should document that.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -316,7 +315,7 @@ where
     fn get_block_template(
```
teor2345, line 315: Can you remove the part of the TODO about HexData, because this PR does that?

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Yeah, I didn't realize the chain verifier was in a Buffer and thought it was cloning CheckpointList. Fixed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: I think it serializes to null (see https://serde.rs/enum-representations.html#untagged), I'll double check.

Update: snapshot shows `null`.

zebra-state/src/service.rs
arya2: That's a brilliant idea. Yes, we should absolutely do this. Applied in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c). I think it should try committing it to any chain to align with the usual acceptance rules so I avoided using a non-finalized state with only the best chain.

Update: added a field to `NonFinalizedState` for configuring whether metrics should be counted in [`32e6ac9`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/32e6ac98a920c89a982cf57e78a5962441a47cc2).

zebra-state/src/service.rs
arya2: It should validate directly off of the finalized tip too, using `write::validate_and_commit_non_finalized` in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c) should achieve that.

zebra-state/src/service.rs
arya2: It probably should've been but we're just re-using the regular `validate_and_commit` fn now, which makes more sense given that that's what `submitblock` will call.

zebra-state/src/service.rs
arya2: Not anymore since it uses the regular `write::validate_and_commit_non_finalized` fn in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c), but I think it was a mistake to do that check in the first place since it's not part of the server's usual acceptance rules.

zebra-consensus/src/block.rs
arya2: It's "excluding the check for a valid proof-of-work" (https://en.bitcoin.it/wiki/BIP_0023#Block_Proposal), it should skip checking the hash as well, I thought `check::difficulty_is_valid` was just checking the difficulty threshold. 

Quoted the requirement in a comment and replaced `check::difficulty_is_valid` with `check::difficulty_threshold_is_valid` for `CheckProposal` requests in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-consensus/src/block.rs
arya2: Fixed in [`8978541`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/89785418947780226bb1e16d1d808008d13e5f7c).

zebra-consensus/src/block.rs
arya2: Fixed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-consensus/src/block.rs
arya2: Whoops. Fixed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Yeah, this else block was disrupting `rustfmt` too. Replaced with an early return in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c), moved to a separate future in [`8978541`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/89785418947780226bb1e16d1d808008d13e5f7c).

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: I'm not sure what to do about this one but it looks unchanged in https://github.com/ZcashFoundation/zebra/pull/5867.

Update: you were right, there was a conflict there, but it's resolved now.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Yeah, that's a good idea, added a `From` impl in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: Whoops, vscode probably auto-added the import, removed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-consensus/Cargo.toml
arya2: zebra-consensus isn't using zebra-rpc or zebra-node-services but I added zebra-chain/getblocktemplate-rpcs in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: I didn't want to repeat some of the code but it's probably better to split it into a different check if it's confusing. Updated in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: Removed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c). (It was only used for optionally cloning the chain verifier)

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: Added in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/parameters.rs
arya2: Fixed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c), I think all the comments to this effect included "unsupported" and searched for that.

zebra-state/src/error.rs
arya2: ~Yeah, I think that might be useful context for debugging, added  in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c)~ The error variant was removed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c) because it's using `write::validate_and_commit_non_finalized` instead. Added an info-level log in [`e3b6d28`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/e3b6d287b0d2b9a51489e16896e0f0232cc04b5e) to recover a bit of context.

zebra-state/src/response.rs
arya2: Renamed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-state/src/response.rs
arya2: Renamed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c).

zebra-state/src/service.rs
arya2: All of the usual ones that are checked in zebra-state for committing a block to the non-finalized state. 

> Do we skip any?

Certainly not, now that it's using `write::validate_and_commit_non_finalized`. (but we were skipping the ones from `chain.push` before.)

zebra-state/src/service.rs
arya2: Fixed in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c) (it now makes this call too since it mutates a cloned non-finalized state and drops it when it's finished).

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Yep, removed [here](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c#diff-c0bec6265a8d68255f80da0e5f499035c5c0181cc218c3d68805094e2c7815f5L314).

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: Moved to a fn so it stays consistent in [`31472a8`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/31472a86b1d0da1ff6f6a9ea577fb52b04017b81). I don't think it should be a problem but I'm open to better ideas.

zebra-consensus/src/chain.rs
arya2: We definitely shouldn't pass it along as a regular call to the checkpoint verifier, I think we should call the block verifier instead and skip `check::subsidy_is_valid` for block proposals with a height below `SLOW_START_INTERVAL`, but I'm not sure. 

Maybe it would be better to immediately return an error, or add a `CheckProposal` variant to the checkpoint verifier and handle `FinalizedBlock` in `CheckValidBlockProposal` too.

Update: I think it's fine to immediately return an error. Fixed in [`a2ae3fb`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/a2ae3fbd8f27ba6a55224c9b742b20ef35973bc4).

mpguerra:
Let's hold off on merging this one until 2023 Sprint 1...

zebra-state/src/service.rs
arya2: I re-used the same write fn but essentially did this in [`8420b75`](https://github.com/ZcashFoundation/zebra/pull/5870/commits/8420b75013c1a05ea11a9799d1f87bd338c3c27c) with your suggestion to attempt writing it to a cloned non-finalized state. 

A new block iterator would be needed to take an `Arc<Chain>` so it seems easier to rebuild a non-finalized state with only the best chain like [this](https://github.com/ZcashFoundation/zebra/pull/5630/files#diff-a2e205e7f47ebdd654a97d9a69c10fed917af8f6f6174e8cce994d5506eeded3R64-R79), but I think it should try committing it to any chain to align with the usual acceptance rules anyways.

teor2345:
Just letting you know that I've seen your changes to this PR, and I'm going to try to review them tomorrow.

Would you mind rebasing this PR to resolve conflicts?

zebra-state/src/service.rs
```diff
@@ -1684,6 +1702,51 @@ impl Service<ReadRequest> for ReadStateService {
                 .map(|join_result| join_result.expect("panic in ReadRequest::ChainInfo"))
                 .boxed()
             }
+
+            // Used by getmininginfo, getnetworksolps, and getnetworkhashps RPCs.
```
arya2:
```suggestion
```

zebra-consensus/src/block/check.rs
```diff
@@ -58,24 +58,22 @@ pub fn coinbase_is_first(block: &Block) -> Result<Arc<transaction::Transaction>,
     Ok(first.clone())
 }
 
-/// Returns `Ok(())` if `hash` passes:
-///   - the target difficulty limit for `network` (PoWLimit), and
-///   - the difficulty filter,
-/// based on the fields in `header`.
+/// Returns `Ok(ExpandedDifficulty)` if `difficulty_threshold` of `header` passes
```
teor2345: Nitpick: Now there's only one check, we can be a bit more specific here.

```suggestion
/// Returns `Ok(ExpandedDifficulty)` if the`difficulty_threshold` of `header` is at least as difficult as
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -332,21 +331,28 @@ where
...
+                .boxed();
+        }
+
         // To implement long polling correctly, we split this RPC into multiple phases.
         async move {
+            get_block_template::check_parameters(&parameters)?;
+
+            let client_long_poll_id = parameters
+                .as_mut()
+                .and_then(|params| params.long_poll_id.take());
```
teor2345: Using `take()` means that we are modifying `parameters`. So if we try to get the same data later in the function it will always be `None`:
https://doc.rust-lang.org/std/option/enum.Option.html#method.take

To review this code, I needed to check every access to `parameters` in the entire `get_block_template()` implementation. I'd like to avoid that if we can.

Can we use `Option::clone()` instead of `Option::take()`?

That way, we will always get the same results from `parameters`. And we won't need to take a mutable reference to `parameters`.

```suggestion
        if let Some(HexData(block_proposal_bytes)) = parameters
            .as_ref()
            .and_then(get_block_template::JsonParameters::block_proposal_data)
        {
            return validate_block_proposal(self.chain_verifier.clone(), block_proposal_bytes)
                .boxed();
        }

        // To implement long polling correctly, we split this RPC into multiple phases.
        async move {
            get_block_template::check_parameters(&parameters)?;

            let client_long_poll_id = parameters
                .as_ref()
                .and_then(|params| params.long_poll_id.clone());
```

(The other alternative is taking ownership of `parameters`. Then the compiler will make sure they are never used again. But I don't think that's needed here?)

zebra-state/src/service.rs
teor2345: This change looks really good, it simplifies the code a lot.

> I think it should try committing it to any chain to align with the usual acceptance rules so I avoided using a non-finalized state with only the best chain.

I'm not sure if accepting blocks on any chain will work for miners.

Miners want to make a profit, so they only want to mine on the best chain. I think this means that blocks that extend any other chain, or any parent of the best tip, should be invalid for the purposes of this RPC.

This rule is assumed in a lot of the Bitcoin and Zcash specs. It is also implied by these RPC errors:
> bad-prevblk | the previous-block is not the one the server intends to build on
> stale-prevblk | the previous-block is no longer the one the server intends to build on

https://en.bitcoin.it/wiki/BIP_0022#Appendix:_Example_Rejection_Reasons

Here are the underlying consensus rules in the Zcash specification:
> In order to choose the best valid block chain in its view of the overall block tree, a node sums the work, ... of all blocks in each valid block chain, and considers the valid block chain with greatest total work to be best. ...
> The consensus protocol is designed to ensure that for any given block height , the vast majority of well-connected nodes should eventually agree on their best valid block chain up to that height. A full validator SHOULD attempt to obtain candidate blocks from multiple sources in order to increase the likelihood that it will find a valid block chain that reflects a recent consensus state.

https://zips.z.cash/protocol/protocol.pdf#blockchain

But I haven't checked what `zcashd` does here. That's one of our other goals - matching the behaviour of its mining RPCs.

teor2345: We could make sure the block builds on the best tip by checking that the previous block hash matches the state tip hash?

For performance and concurrency reasons, we probably want to do this right before we call `validate_and_commit_non_finalized()`.

zebra-state/src/service.rs
teor2345: Thanks, this looks good!

zebra-state/src/service.rs
teor2345: Thanks for fixing this up, it's clear that we're doing all the required checks now.

zebra-consensus/src/block.rs
```diff
@@ -21,19 +21,15 @@ use thiserror::Error;
...
-};
+use zebra_chain::{amount::Amount, block, parameters::Network, transparent, work::equihash};
 use zebra_state as zs;
 
 use crate::{error::*, transaction as tx, BoxError};
 
 pub mod check;
+pub mod request;
 pub mod subsidy;
+pub use request::Request;
```
teor2345, line 33: Nitpick: I usually separate different types of imports from each other for readability.

```suggestion

pub use request::Request;
```

zebra-consensus/src/block/check.rs
```diff
@@ -58,24 +58,22 @@ pub fn coinbase_is_first(block: &Block) -> Result<Arc<transaction::Transaction>,
...
+/// the target difficulty limit for `network` (PoWLimit)
 ///
-/// If the block is invalid, returns an error containing `height` and `hash`.
-pub fn difficulty_is_valid(
+/// If the header is invalid, returns an error containing `height` and `hash`.
```
teor2345:
```suggestion
/// If the header difficulty threshold is invalid, returns an error containing `height` and `hash`.
```

zebra-consensus/src/block/request.rs
```diff
@@ -0,0 +1,40 @@
...
+        })
+    }
+
+    /// Returns `true` if the request is a proposal
+    pub fn is_proposal(&self) -> bool {
+        match self {
+            Request::Commit(_) => false,
+
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            Request::CheckProposal(_) => true,
```
teor2345, line 37: Thanks for doing it this way, it makes it easy to check that we're not accidentally calling proposal code in production.

zebra-rpc/src/methods/get_block_template_rpcs/constants.rs
```diff
@@ -26,8 +26,7 @@ pub const GET_BLOCK_TEMPLATE_MUTABLE_FIELD: &[&str] = &[
 ];
 
 /// A hardcoded list of Zebra's getblocktemplate RPC capabilities.
-/// Currently empty.
```
teor2345, line 29: Nitpick: Would you mind linking to the spec for this field, and the other constants before and after it?
(They also have missing spec links, this is an existing doc bug.)

https://en.bitcoin.it/wiki/BIP_0023#Mutations

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -262,7 +274,62 @@ pub async fn test_responses<State, ReadState>(
...
+            ..Default::default()
+        },
+    )));
+
+    let get_block_template = get_block_template
+        .await
+        .expect("unexpected panic in getblocktemplate RPC task")
+        .expect("unexpected error in getblocktemplate RPC call");
+
+    snapshot_rpc_getblocktemplate("invalid-proposal", get_block_template, None, &settings);
```
teor2345, line 299: Nitpick: consistent filenames

```suggestion
    snapshot_rpc_getblocktemplate("invalid_proposal", get_block_template, None, &settings);
```

zebra-state/src/service/check.rs
```diff
@@ -14,13 +14,13 @@ use zebra_chain::{
...
         block_iter::any_ancestor_blocks, check::difficulty::POW_ADJUSTMENT_BLOCK_SPAN,
-        finalized_state::FinalizedState, non_finalized_state::NonFinalizedState,
+        non_finalized_state::NonFinalizedState,
     },
     BoxError, PreparedBlock, ValidateContextError,
 };
 
 // use self as check
-use super::check;
+use super::{check, finalized_state::ZebraDb};
```
teor2345: Nitpick: I realise this is probably an auto-import thing, but this line is a special import that allows us to refer to functions in this module using `check::fn()`. They are more readable that way.

So I'm not sure we want any other imports in this list.

zebra-state/src/service/write.rs
```diff
@@ -27,6 +27,8 @@ use crate::service::{
     non_finalized_state::Chain,
 };
 
+use super::finalized_state::ZebraDb;
```
teor2345: Nitpick: please add this to the existing crate imports.

teor2345 (commented):
I'm really happy with the way you've made this change clearer and more consistent, and made it easier to see that we are following the consensus rules.

For testing, I'd like to see a diff between a Zebra and `zcashd` block proposal response. I'm also happy to do that test, if you don't have synced instances ready.

This PR took me a while to review, because it changes about 40 files. We might want to think about how to split up future large changes into stages, or how we can avoid unnecessary changes. (This is a general comment, there's no need to split this PR!)

For example, I think we could have done the `FinalizedState` to `ZebraDb` change in another PR, which would have split out about 10 files from this PR.

teor2345:
> For testing, I'd like to see a diff between a Zebra and `zcashd` block proposal response. I'm also happy to do that test, if you don't have synced instances ready.

This is now handled in ticket #5803, so it doesn't need to block this PR or the CI tests.

zebra-state/src/service.rs
```diff
@@ -1684,6 +1702,55 @@ impl Service<ReadRequest> for ReadStateService {
...
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        tracing::info!("attempting to validate and commit block proposal onto a cloned non-finalized state");
+                        let mut latest_non_finalized_state = state.latest_non_finalized_state();
+
+                        // Check that the previous block hash is the best chain tip if there are blocks in the 
+                        // non-finalized state. The previous block of a valid proposal must be on the best chain tip.
+                        if let Some(true) = latest_non_finalized_state.best_tip().map(|(_, tip_hash)| tip_hash != prepared.hash) {
+                            return Err("previous block hash is not the best chain tip".into())
+                        }
```
teor2345: I can see two potential issues here:
* we need to check the previous block hash in the proposal header, not the hash of the proposal header data
* it's a rare edge-case, but `validate_and_commit_non_finalized()` can also fork a non-finalized chain off the finalized tip

So I think what we might want here is something like:
```suggestion
                       // The previous block of a valid proposal must be on the best chain tip.
                       let Some((_best_tip_height, best_tip_hash) = read::best_tip(&latest_non_finalized_state, &state.db) {
                           if prepared.block.header.previous_block_hash !=  best_tip_hash {
                               return Err("proposal is not based on the current best chain tip: previous block hash must be the best chain tip".into());
                           }
                       } else {
                           return Err("state is empty: wait for Zebra to sync before submitting a proposal".into());
                       }
```

teor2345 (commented):
I'm happy for anyone to approve once we've fixed the previous block hash check.

zebra-state/src/service.rs
arya2: > we need to check the previous block hash in the proposal header, not the hash of the proposal header data

🤦 Thank you.

> it's a rare edge-case, but validate_and_commit_non_finalized() can also fork a non-finalized chain off the finalized tip

I was thinking when the non-finalized state is empty, either a call to `commit_block` will fail to find the previous block hash and return `ValidateContextError::NotReadyToBeCommitted`, or `commit_new_chain` will commit the best chain to the non-finalized state.

> We should be able to validate directly off the finalized tip, if we want to. All the functions that we call below support it.

I'm still not sure how the non-finalized state could be empty near the network tip but if calling `validate_and_commit_non_finalized` when there is no best chain in the non-finalized state works I think we should keep it.

arya2:
> This PR took me a while to review, because it changes about 40 files. We might want to think about how to split up future large changes into stages, or how we can avoid unnecessary changes. (This is a general comment, there's no need to split this PR!)

I agree with splitting changes into multiple PRs. I also liked how you organized the getblocktemplate refactor in PR https://github.com/ZcashFoundation/zebra/pull/5837 for when there are a lot of changes in one PR.

arya2:
An `expect_request` in the snapshot timed out:

> ---- methods::tests::snapshot::test_rpc_response_data stdout ----
> 
> Message:  timeout while waiting for a request
>  in zebra_test::mock_service::MockService<zebra_node_services::mempool::Request, zebra_node_services::mempool::Response, zebra_test::mock_service::PanicAssertion, alloc::boxed::Box<dyn core::error::Error+core::marker::Sync+core::marker::Send>>

https://github.com/ZcashFoundation/zebra/actions/runs/3886245054/jobs/6631117711#step:14:3001

Update: I made issue https://github.com/ZcashFoundation/zebra/issues/5940 for fixing the test.

zebra-state/src/service.rs
teor2345: > > it's a rare edge-case, but validate_and_commit_non_finalized() can also fork a non-finalized chain off the finalized tip
> 
> I was thinking when the non-finalized state is empty, either a call to `commit_block` will fail to find the previous block hash and return `ValidateContextError::NotReadyToBeCommitted`, or `commit_new_chain` will commit the best chain to the non-finalized state.
> 
> > We should be able to validate directly off the finalized tip, if we want to. All the functions that we call below support it.
> 
> I'm still not sure how the non-finalized state could be empty near the network tip but if calling `validate_and_commit_non_finalized` when there is no best chain in the non-finalized state works I think we should keep it.

I think it's a rare edge-case, it would only happen after a large chain fork, or if the node was disconnected from the network.

teor2345 (approved):
Thanks for fixing up my suggestions!

arya2:
> Thanks for fixing up my suggestions!

Thanks for the excellent suggestions!

teor2345:
Same network error as https://github.com/ZcashFoundation/zebra/pull/5938#issuecomment-1378201078

teor2345:
@mergifyio update

teor2345 (approved):

PR #5307, opened 2022-09-29
1. fix(rpc): Fix slow getblock RPC (verbose=1) using transaction ID index

teor2345:
## Motivation

Zebra's `lightwalletd` sync takes a lot longer than `zcashd`'s (7h vs 5h).

Fixes #5298.

## Solution

Use the transaction ID index for verbose `getblock` RPC requests, rather than deserializing the entire block, then hashing all the transactions in the block.

Also add timing output to `zcash-rpc-diff`.

## Review

This is a routine fix.

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [x] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
  - [x] How do you know it works? Does it have tests?

Issue #5298, opened 2022-09-28
Zebra full `lightwalletd` sync is significantly slower than `zcashd`


teor2345:
## Motivation

Currently, a full `lightwalletd` sync is 37% slower with Zebra than `zcashd` (7h vs 5h).

#### Full Sync Benchmarks

Here are some commands for full sync benchmarks:

Zebra:
```sh
time lightwalletd --zcash-conf-path /home/dev/.config/zebra/zebrad-mainnet-lightwalletd-tmp-zcash.conf --data-dir $(mktemp -d) --log-file /dev/stdout --grpc-bind-addr 0.0.0.0:0 --http-bind-addr 0.0.0.0:0 --no-tls-very-insecure 2>&1 | rg --passthru --max-count 1 'Adding block to cache 182[0-9]{4}'
```

`zcashd`:
```sh
time lightwalletd --zcash-conf-path /home/dev/.zcash/zcash.conf --data-dir $(mktemp -d) --log-file /dev/stdout --grpc-bind-addr 0.0.0.0:0 --http-bind-addr 0.0.0.0:0 --no-tls-very-insecure 2>&1 | rg --passthru --max-count 1 'Adding block to cache 182[0-9]{4}'
```

#### Quick Benchmarks

Here are some commands for quick benchmarks:

Zebra:
```sh
TODO
```

`zcashd`:
```sh
TODO
```

#### Benchmark Notes

You can change `182[0-9]{4}` to the current tip height minus 10,000. (For example, if the current tip height is 1,975,302 use `196[0-9]{4}`.)

`rg` (ripgrep) is a fast `grep` alternative written in Rust. You can use `tee /dev/stderr | grep --max-count=1` instead of `rg rg --passthru --max-count 1`.

`/home/dev/.config/zebra/zebrad-mainnet-lightwalletd-tmp-zcash.conf` is set up like it is in the `lightwalletd` tests:
```
rpcbind=127.0.0.1
rpcport=(your Zebra RPC port)
```

## TODO

- a benchmark that takes 1-5 minutes (or less!)
- check if older Zebra releases were faster
- use a flamegraph to find out what parts of Zebra are slow

teor2345:
Some changes that might impact RPC speed:
- #5257
- data compression in rocksdb
- #5253
- re-applying the revert in #4933

Some previous changes that might have regressed RPC speed:
- #5012 
- #4806 
- #5134 
- #5102

teor2345:
[...]

PR #5894, opened 2022-12-24
feat(zebra-checkpoints): make `zebra-checkpoints` work for zebrad backend

### LINKED ISSUE #5852: Make `zebra-checkpoints` work with Zebra RPCs, as well as `zcashd` RPCs
## Motivation

Currently, the `zebra-checkpoints` utility uses the `getblockcount` RPC, which Zebra doesn't have. But we do have other RPCs that provide the current block height.

If we use those RPCs, then `zebra-checkpoints` will work with both `zcashd` and Zebra.

It might also be useful to be able to compare checkpoints generated by `zcashd` and `zebrad`.

## Proposed Changes

Switch `zebra-checkpoints` from `getblockcount` to the `blocks` field of `getblockchaininfo`.

## Related Work

Another larger alternative change is:
- #795

oxarbitrage:
## Motivation

Zebra has an [utility tool to create checkpoints](https://github.com/ZcashFoundation/zebra/tree/main/zebra-utils#zebra-checkpoints). Until now this tool only work if zcashd is used as the backend. 

This PR make the changes to make it work with `zebrad` as the backend while `zcashd` is still supported.

Close https://github.com/ZcashFoundation/zebra/issues/5852

## Solution

When i started on this i was only expecting to make changes to the `zebra-checkpoints` binary, however i realized later we were going to need changes in the `getblock` rpc method.

The binary is expecting `height`, `hash` and `size` fields of a `getblock` call with verbosity = 1. Currently (and because ligthwalletd only make use of the `tx` field of the object response), we are not returning any additional data.

this PR updated the `getblock` rpc method to return the additional data needed to make the tool work.

Also, in the `zebra-checkpoints` we changed the use of `getblockcount` to the `blocks` field of `getblockchaininfo` call as suggested in the [proposed changes section of the ticket](https://github.com/ZcashFoundation/zebra/issues/5852).

## Review

This is optional and low priority. It can wait until people get back. Anyone can review.

### Reviewer Checklist

## Follow Up Work

- [x] Update some docs mentioning this in zebra book or checkpoint README ?
- [x] Do manual tests of differences between zcashd and zebrad generated checkpoints.
- [x] Compare the speed of using zebrad against zcashd as the backend.


oxarbitrage:
I did some manual test for comparison and performance.

In regards to time, zebra started a bit faster but they both ended up taking pretty much the same amount of time to get a full list of up to date checkpoints in the Mainnet. Here what i did:

zebra (i have a synchronized chain at port 6666):
```
$ time ./target/debug/zebra-checkpoints --cli /media/zcash/zcash/src/zcash-cli  -- -rpcport=6666 > mainnet_checkpoints_zebrad.txt

real	381m56,356s
user	60m57,259s
sys	36m19,635s
$ 
```

zcashd (synchronized and running in default rpc port):

```
$ time ./target/debug/zebra-checkpoints --cli /media/zcash/zcash/src/zcash-cli -- -rpccookiefile="/media/chain/zcash/.cookie" > mainnet_checkpoints_zcashd.txt
real	399m10,090s
user	67m53,213s
sys	37m6,481s
$
```

A comparassion of this 2 generated files show that they are exactly the same.

```
$ diff --color mainnet_checkpoints_zebrad.txt mainnet_checkpoints_zcashd.txt 
$
```

zebra-rpc/src/methods.rs
```diff
@@ -613,7 +615,14 @@ where
                 match response {
                     zebra_state::ReadResponse::TransactionIdsForBlock(Some(tx_ids)) => {
                         let tx_ids = tx_ids.iter().map(|tx_id| tx_id.encode_hex()).collect();
-                        Ok(GetBlock::Object { tx: tx_ids })
+                        Ok(GetBlock::Object {
+                            hash: block.hash(),
+                            height: block
+                                .coinbase_height()
+                                .expect("all blocks should have a coinbase height"),
+                            size: SerializedBlock::from(block).as_ref().len(),
```
arya2: Optional: could we use `zcash_serialized_size()` here (needs a trait import as well):

```suggestion
                            size: block.zcash_serialized_size(),
```

zebra-rpc/src/methods/tests/vectors.rs
```diff
@@ -113,7 +113,10 @@ async fn rpc_getblock() {
                     .transactions
                     .iter()
                     .map(|tx| tx.hash().encode_hex())
-                    .collect()
+                    .collect(),
+                height: block.coinbase_height().unwrap(),
```
arya2: Optional: could we use an `expect` call instead,
```suggestion
                height: block.coinbase_height().expect("test block should have coinbase height"),
```

arya2 (dismissed):
Looks good to me.

arya2 (dismissed):

arya2:
@Mergifyio refresh

zebra-rpc/src/methods.rs
```diff
@@ -613,7 +615,14 @@ where
                 match response {
                     zebra_state::ReadResponse::TransactionIdsForBlock(Some(tx_ids)) => {
                         let tx_ids = tx_ids.iter().map(|tx_id| tx_id.encode_hex()).collect();
-                        Ok(GetBlock::Object { tx: tx_ids })
+                        Ok(GetBlock::Object {
+                            hash: block.hash(),
+                            height: block
+                                .coinbase_height()
+                                .expect("all blocks should have a coinbase height"),
+                            size: block.zcash_serialized_size(),
```
teor2345: Reading entire block when `verbose = 1` will cause performance issues for `lightwalletd`.

`lightwalletd` calls the `getblock` RPC twice for every block during its initial sync, first with `verbose = 1` to get the transaction IDs, and again with `verbose = 0` to get the block data:
https://github.com/zcash/lightwalletd/blob/5d174f7feb702dc19aec5b09d8be8b3d5b17ce45/common/common.go#L246-L254

This caused a serious performance issue for Zebra, here's how we fixed it a few months ago:
- https://github.com/ZcashFoundation/zebra/issues/5298#issuecomment-1262775222
- https://github.com/ZcashFoundation/zebra/pull/5307

(I should have added some comments about this when I fixed it, sorry about that!)

When we were fetching the full block when `verbose = 1`, a full `lightwalletd` sync took 7 hours. (Compared with 3.4 hours after the fix in September.)

`zcashd` takes around 5 hours for a full `lightwalletd` sync. This change would make us much slower than `zcashd`.

This seems like a blocker, because one of our release goals is being faster than `zcashd`.

### Suggested Fix

We want Zebra to avoid reading the whole block for `lightwalletd`, which needs to efficiently get the `tx` field. But `zebra-checkpoints` needs the `hash`, `height`, and `size` fields.

Here are some ideas:
1. We could add a private `verbose` mode for `zebra-checkpoints`. Then `lightwalletd` can continue using `verbose=1`, and we can make any changes we want to the new mode.
2. Or we can add a separate private RPC with just the fields that `zebra-checkpoints` needs.

When `zebra-checkpoints` starts up, it can check if it's connected to Zebra, and use the private mode/RPC.

This is all temporary, until either:
- `lightwalletd` gets fixed so it calculates v5 transaction IDs correctly
- Zebra stores the block size in the database (ticket #5911)

What do you think?

### Other Alternatives

#### Wait for Fixes

That seems like a bit of work, and it's a bit complicated. So we could wait for the proper fixes in Zebra's state, or in `lightwalletd`. We could update the ticket with the progress we've made, and a list of blockers?

#### Get Fields Efficiently

I thought of some ways to get the RPC fields efficiently. But I couldn't get all of them without reading the block.

- hash: Call `ReadRequest::BestChainBlockHash(height)` or if the hash was supplied, use it
    - we'll need to change this request so it is compiled unconditionally, by removing the `getblocktemplate-rpcs` feature gate
- height: Add a new `ReadRequest::BestChainBlockHeight(hash)`, or if the height was supplied, use it
    - you can use `read::height_by_hash()` to do this efficiently
- size: RocksDB doesn't store the size of each value, it reads the disk instead. So we'd need to add the serialized block size to the database to do this efficiently (probably in `hash_by_height`).
    - this seems out of scope for this PR, so I opened another ticket #5911

#### Redesign zebra-checkpoints

We could use the number of transactions for checkpoint block spacing, rather than the number of bytes in each block. We already get this efficiently from `verbose=1`.

 I'm not sure what impact this would have on Zebra's sync performance, because transactions can be very different sizes.

zebra-consensus/src/checkpoint/README.md
```diff
@@ -22,7 +22,7 @@ Or jump straight to [the exact commands for updating the lists](https://github.c
 
 ### Use the `zebra-checkpoints` utility
 
-`zebra-checkpoints` is the program we use to collect checkpoints. Currently this program uses `zcash-cli` to get the hashes. `zcash-cli` must be available in your machine and it must be connected to a synchronized (Mainnet or Testnet) instance of `zcashd` to get the most recent hashes.
+`zebra-checkpoints` is the program we use to collect checkpoints. Currently this program uses `zcash-cli` to get the hashes. `zcash-cli` must be available in your machine and it must be connected to a synchronized (Mainnet or Testnet) instance of `zebrad` or `zcashd` to get the most recent hashes.
```
teor2345, line 25: Thanks for updating the docs!

teor2345 (dismissed):
This looks good, and I'd really like to merge it. But it seems to cause a performance issue with `lightwalletd`.

I spent some time trying to work out how to fix it, let me know what you think of my suggestions!

zebra-rpc/src/methods.rs
daira: > lightwalletd gets fixed so it calculates v5 transaction IDs correctly

Is there an issue for that? I don't see it at https://github.com/zcash/lightwalletd/issues .

teor2345: There was an issue:
https://github.com/zcash/lightwalletd/issues/392

But it was closed with this workaround:
https://github.com/zcash/lightwalletd/pull/393

```go
	// `block.ParseFromSlice` correctly parses blocks containing v5
	// transactions, but incorrectly computes the IDs of the v5 transactions.
	// We temporarily paper over this bug by fetching the correct txids via a
	// verbose getblock RPC call, which returns the txids.
	//
	// Unfortunately, this RPC doesn't return the raw hex for the block,
	// so a second getblock RPC (non-verbose) is needed (below).
```
https://github.com/zcash/lightwalletd/blob/07fd6a966a136a751b37b8c54a6679499def0caa/common/common.go#L246

I'm not sure how that team handles technical debt.

It is likely only a performance issue for `lightwalletd`'s initial sync, because the slowdown is much smaller than the target block interval.

arya2: We could also redesign zebra-checkpoints to call `getblock` with `verbose=0` and deserialize the response, or only return height/hash/size from `verbose=1` when compiled without the lightwalletd feature.

teor2345: > We could also redesign zebra-checkpoints to call `getblock` with `verbose=0` and deserialize the response,

That's a great idea! Let's do it.

It will be a bit slow to download all the block data, but we only update the checkpoints every few weeks, so that's ok.

As optional performance improvements, we could:
* keep the existing `getblock verbose=1` code, and use it with `zcashd`. Then `zcashd` will stay the same speed.
* only deserialize the block header in Zebra mode, then just count the rest of the bytes to get the block size.

> or only return height/hash/size from `verbose=1` when compiled without the lightwalletd feature.

Unfortunately, there is no lightwalletd feature in Zebra. (And Rust features are meant to be additive.) So I'm not sure if this alternative would work. But the other one seems promising.

mpguerra:
If we can't get this merged with minimal changes, e.g timeboxed to 1 day + 1 more round of reviews then we should park it for now

zebra-rpc/src/methods.rs
oxarbitrage: I don't see how deserializing the header only will give you the hash and height. Deserializing the entire block should work, with the known performance problem. I think i will try that.

oxarbitrage:
> If we can't get this merged with minimal changes, e.g timeboxed to 1 day + 1 more round of reviews then we should park it for now

I think this is easy enough for a fix in one day and another round of reviews. Working on it now.

zebra-utils/src/bin/zebra-checkpoints/args.rs
```diff
@@ -3,10 +3,40 @@
...
+use std::str::FromStr;
+
+/// The backend type the zebra-checkpoints utility will use to get data from.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Backend {
+    Zebrad,
+    Zcashd,
+}
+
+impl FromStr for Backend {
```
arya2, line 17: Optional/Nitpick: it would be nice to always use verbose=0 and deserialize the block to simplify the code.

arya2 (approved):
Thanks for the workaround to the performance issue!

zebra-utils/src/bin/zebra-checkpoints/args.rs
teor2345: ~~Downloading the entire block could significantly reduce performance, so I'd like to keep the faster mode available.~~

teor2345 (commented):
Looks good, just doing some testing to make sure all the backends produce the same results.

zebra-utils/src/bin/zebra-checkpoints/args.rs
teor2345: I just checked the performance, and it seems pretty similar. So I am not concerned about the difference, particularly for a tool we only run every few weeks.

So I don't mind what we do here, but we're trying to minimise changes so we can get this PR merged and move on with other work.

Here are the performance figures, it turns out the new mode is actually the fastest overall. It can take almost twice as long to query `zcashd` with `--backend zebrad` on mainnet, otherwise the times aren't that different.

Mainnet:
```sh
$ time zebra-checkpoints-release --backend zcashd --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/main-checkpoints.txt | cut -d" " -f1) | tee /tmp/main-zcashd-zcashd.txt
1941792 0000000000e0989263bef30dd453c6813dee07bdd1b61d3c846fae98d2c8929b
...
1954447 000000000086aa0f9e3375ef2c9211b524ab562aba10024c2d4b1e856bf223cc
real    6m5.974s
$ time zebra-checkpoints-release --backend zebrad --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/main-checkpoints.txt | cut -d" " -f1) | tee /tmp/main-zebrad-zcashd.txt
...
real    9m27.998s
$ time zebra-checkpoints-release --backend zebrad --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/main-checkpoints.txt | cut -d" " -f1) -- --rpcport=28232 | tee /tmp/main-zebrad-zebrad.txt
...
real    5m0.404s
```

Testnet:
```sh
$ time zebra-checkpoints-release --backend zcashd --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/test-checkpoints.txt | cut -d" " -f1) --cli zcash-cli-testnet | tee /tmp/test-zcashd-zcashd.txt
2183200 00388d6db62115418ba764acc6db4aa77e1dc40219d2d447e11605c75824549e
...
2196400 006756f092109fc63ebb76a9b5376d4a6a7e8a73efac32a682490546f647bf90
real    2m32.585s
$ time zebra-checkpoints-release --backend zebrad --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/test-checkpoints.txt | cut -d" " -f1) --cli zcash-cli-testnet | tee /tmp/test-zebrad-zcashd.txt
...
real    2m39.516s
$ time zebra-checkpoints-release --backend zebrad --last-checkpoint $(tail -1 zebra-consensus/src/checkpoint/test-checkpoints.txt | cut -d" " -f1) -- --rpcport=38232 | tee /tmp/test-zebrad-zebrad.txt
...
real    1m31.989s
```

teor2345:
> Looks good, just doing some testing to make sure all the backends produce the same results.

All 3 backend/node combinations produce the same results for new checkpoints, I'll run a full set of checkpoints overnight to be sure. I think it's worth doing this test because checkpoints are consensus-critical.

zebra-utils/src/bin/zebra-checkpoints/args.rs
teor2345: For the full sync, I used `zebrad` concurrently in both modes, so its times might be slower.

Mainnet:
```sh
$ time zebra-checkpoints-release --backend zcashd | tee /tmp/main-zcashd-zcashd.txt
1954681 00000000015a8986b4664632ecbcd9421b03af430991077b029dc2f6eb39709c
real    496m35.946s
$ time zebra-checkpoints-release --backend zebrad | tee /tmp/main-zebrad-zcashd.txt
real    846m48.880s
$ time zebra-checkpoints-release --backend zebrad -- --rpcport=28232 | tee /tmp/main-zebrad-zebrad.txt
real    604m2.113s
```

Testnet:
```sh
$ time zebra-checkpoints-release --backend zcashd --cli zcash-cli-testnet | tee /tmp/test-zcashd-zcashd.txt
2196800 0006a24636968703c7407a4444d34e313568352dc1a1179006b8ca84b5e004bd
real    438m13.955s
$ time zebra-checkpoints-release --backend zebrad --cli zcash-cli-testnet | tee /tmp/test-zebrad-zcashd.txt
real    437m48.286s
$ time zebra-checkpoints-release --backend zebrad -- --rpcport=38232 | tee /tmp/test-zebrad-zebrad.txt
real    257m32.831s
```

Zebra is almost twice as fast on testnet, and slower by about 20% on mainnet.

teor2345:
The full checkpoint files were exactly the same as each other, and they didn't change any of the existing checkpoints.

PR #5909, opened 2023-01-03
fix(test): Avoid spurious rpc_getblocktemplate test failure

arya2:
## Motivation

There are occasional test failures in CI from an `expect_request` call on the mock state or mempool panic on timeout. This failure can be observed locally as well.

<details>

> Message:  timeout while waiting for a request
>  in zebra_test::mock_service::MockService<zebra_node_services::mempool::Request, zebra_node_services::mempool::Response, zebra_test::mock_service::PanicAssertion, alloc::boxed::Box<dyn core::error::Error+core::marker::Sync+core::marker::Send>>
> Location: /opt/zebrad/zebra-test/src/mock_service.rs:449

https://github.com/ZcashFoundation/zebra/actions/runs/3825986911/jobs/6509836292#step:3:2254
</details>

## Solution

- Drive mock service request handling futures on main thread

## Review

Anyone can review.

I ran it 100x in a loop so it would usually fail before the change, this seems to have fixed it so it passes consistently.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/tests/vectors.rs
+++ b/zebra-rpc/src/methods/tests/vectors.rs
@@ -971,23 +971,19 @@ async fn rpc_getblocktemplate_mining_address(use_p2pkh: bool) {
     );
 
     // Fake the ChainInfo response
-    let mock_read_state_request_handler = {
-        async move {
-            read_state
-                .expect_request_that(|req| matches!(req, ReadRequest::ChainInfo))
-                .await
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks!

Usually I would just increase the mock service request timeout at:
https://github.com/ZcashFoundation/zebra/blob/53e836efdcb6f25319403d8e43daaff92408b009/zebra-test/src/mock_service.rs#L82

It was never really designed or tested for Tokio tasks under heavy load on multiple threads.

PR #5915, opened 2023-01-05
refactor(state): Make implementation of block consensus rules clearer

### LINKED ISSUE #5910: Remove duplicate `read::hash()` and `read::hash_by_height()` consensus-critical code
## Motivation

Zebra has two duplicate state functions for reading a block hash: `read::hash()` and `read::hash_by_height()`. We don't want to have duplicate code in Zebra, because it risks breaking consensus. It's also hard to maintain and audit.

This is a simple search-and-replace fix.

https://github.com/ZcashFoundation/zebra/blob/53e836efdcb6f25319403d8e43daaff92408b009/zebra-state/src/service/read/find.rs#L111-L112

https://github.com/ZcashFoundation/zebra/blob/53e836efdcb6f25319403d8e43daaff92408b009/zebra-state/src/service/read/block.rs#L187-L190

### Testing

This change will improve testing, because the existing tests will all check the same function, rather than some tests checking one function, and some checking the other function.

## Related Work

Discovered as part of PR #5894.

### LINKED ISSUE #5914: Unclear implementation of block "timestamp against prev" (time-too-old) consensus rule
It appears that the consensus rule of "checking the timestamp against the median time past" (`time-too-old`) which is present in ZCash (ZEC) is not implemented in Zebra:

https://github.com/zcash/zcash/blob/master/src/main.cpp#L4829-L4835
```cpp
    // Check timestamp against prev
    auto medianTimePast = pindexPrev->GetMedianTimePast();
    if (block.GetBlockTime() <= medianTimePast) {
        return state.Invalid(error("%s: block at height %d, timestamp %d is not later than median-time-past %d",
                                   __func__, nHeight, block.GetBlockTime(), medianTimePast),
                             REJECT_INVALID, "time-too-old");
    }
```

In Zebra sources I see only `time-too-new` rule implementation:
https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/block.rs#L193-L195
```rust
            let now = Utc::now();
            check::time_is_valid_at(&block.header, now, &height, &hash)
                .map_err(VerifyBlockError::Time)?;
```

It appears that there is no check for a `time-too-old` at all. I understand that Zebra uses "parallel block validation", which allows for blocks to come in an "unpredictable order" and could potentially not provide enough data to calculate the median time past at that moment. However, it looks like this check is not present in the contextual validation (nor in `initial_contextual_validity`, nor in `block_is_valid_for_recent_chain`) either. If I am mistaken, I apologize.

teor2345:
## Motivation

This PR changes two consensus-critical functions where the consensus rule implementation was unclear or duplicated.

This is not a security issue, but it might cause issues with future changes if we don't fix it.

## Solution

1. Rename a function to make it clear what consensus rules it implements
2. Delete a duplicate consensus-critical function

Closes #5914
Closes #5910

## Review

Anyone can review this PR.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -1573,7 +1573,11 @@ impl Service<ReadRequest> for ReadStateService {
                     span.in_scope(move || {
                         let hash = state.non_finalized_state_receiver.with_watch_data(
                             |non_finalized_state| {
-                                read::hash(non_finalized_state.best_chain(), &state.db, height)
+                                read::hash_by_height(
+                                    non_finalized_state.best_chain(),
+                                    &state.db,
+                                    height,
+                                )
                             },
                         );
 
--- a/zebra-state/src/service/check.rs
+++ b/zebra-state/src/service/check.rs
@@ -100,7 +100,7 @@ where
     });
     let difficulty_adjustment =
         AdjustedDifficulty::new_from_block(&prepared.block, network, relevant_data);
-    check::difficulty_threshold_is_valid(
+    check::difficulty_threshold_and_time_are_valid(
         prepared.block.header.difficulty_threshold,
         difficulty_adjustment,
     )?;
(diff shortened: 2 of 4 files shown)


dconnolly (approved):
lgtm!

Issue #5871, opened 2022-12-16
Fix getblocktemplate min and max times on testnet


teor2345:
## Motivation

On testnet only, there is a 1679 second difference between the `mintime` produced by Zebra and `zcashd` for the same height.

This amount doesn't seem to correspond to any particular consensus rule.

### Specifications

> Define the median-time-past of a block to be the median (as defined in § 7.7.3 ‘Difficulty adjustment’ on p. 131) of the nTime fields of the preceding PoWMedianBlockSpan blocks (or all preceding blocks if there are fewer than PoWMedianBlockSpan). The median-time-past of a genesis block is not defined.
> For each block other than the genesis block , nTime MUST be strictly greater than the median-time-past of that block.
> For each block at block height 2 or greater on Mainnet , or block height 653606 or greater on Testnet , nTime MUST be less than or equal to the median-time-past of that block plus 90 · 60 seconds.

https://zips.z.cash/protocol/protocol.pdf#blockheader

> This [difficulty] algorithm changed on Testnet, starting from block 299188, to allow "minimum-difficulty" blocks. If the block time of a block from this height onward is greater than 15 minutes after that of the preceding block, then the block is a minimum-difficulty block. In that case its nBits field MUST be set to ToCompact(PoWLimit)

https://zips.z.cash/zip-0205#change-to-difficulty-adjustment-on-testnet

### Complex Code or Requirements

Zebra adjusts the `mintime` or `maxtime` on testnet, so the block time range is either for a standard difficulty block, or a minimum difficulty block.

### Analysis

This is not a minimum difficulty block, so the `maxtime` should get adjusted to 15 minutes after the `mintime`. But instead it is 93.8 minutes later. But it should be 15m after the `mintime` for standard difficulty, or 75m (90m-15m) for minimum difficulty.

This seems like a bug in Zebra's testnet time adjustment. These comments are also wrong and don't match the spec:
https://github.com/ZcashFoundation/zebra/blob/80a6d3cdab2a9dc68e27ecfa729ba864394becb4/zebra-state/src/service/read/difficulty.rs#L263-L267

It is also possible that `zcashd` sets a higher minimum than the time permitted by the consensus rules.

## Related Work

This was discovered as part of:
- #5400 

## Logs

```sh
$ ZCASH_CLI="zcash-cli -testnet" zcash-rpc-diff 38232 getblocktemplate                                                                                               Checking first node release info...
Checking second node release info...
[...]

PR #5917, opened 2023-01-05
test(rpc): Add raw getblocktemplate RPC difficulty fields for testing only

teor2345:
## Motivation

As part of fixing bug #5871, I added some raw fields to the `getblocktemplate` RPC response. These fields have the difficulty values before the testnet difficulty adjustment.

### Specifications

> On Testnet from block height 299188 onward, the difficulty adjustment algorithm [6](https://zips.z.cash/zip-0208#protocol-diffadjustment) allows minimum-difficulty blocks, as described in [10](https://zips.z.cash/zip-0208#zip-0205), when the block time is greater than a given threshold. This specification changes this threshold to be proportional to the block target spacing.
>
> That is, if the block time of a block at height height ≥ 299188 is greater than 6 · PoWTargetSpacing(height) seconds after that of the preceding block, then the block is a minimum-difficulty block. In that case its nBits field MUST be set to ToCompact(PoWLimit), where PoWLimit is the value defined for Testnet in section [5](https://zips.z.cash/zip-0208#protocol-constants).3 of the Zcash Protocol Specification 5, and ToCompact is as defined in section [7](https://zips.z.cash/zip-0208#protocol-nbits).7.4 of that specification 7.
>
> ...
>
> This change does not affect Mainnet.

### Analysis

It looks like the `min_time` is being adjusted wrong, see https://github.com/ZcashFoundation/zebra/issues/5871#issuecomment-1371853411

## Solution

TODO

## Review

Not ready for review yet.

### Reviewer Checklist

## Follow Up Work

Actually fix the bug, either in this PR, or in another one.

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
@@ -159,6 +159,30 @@ pub struct GetBlockTemplate {
     #[serde(default)]
     #[serde(rename = "submitold")]
     pub submit_old: Option<bool>,
+
+    /// The expected difficulty for the new block displayed in expanded form,
+    /// with no testnet minimum difficulty adjutment.
+    #[serde(with = "hex")]
+    pub raw_target: ExpandedDifficulty,
+
+    /// The expected difficulty for the new block displayed in compact form,
+    /// with no testnet minimum difficulty adjutment.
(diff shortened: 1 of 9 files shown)


teor2345:
I fixed this bug in PR #5925.

Feel free to cherry-pick the raw fields commit 069e4ea if it's needed to test other PRs.

PR #5925, opened 2023-01-06
fix(rpc): Calculate getblocktemplate RPC testnet min and max times correctly

### LINKED ISSUE #5871: Fix getblocktemplate min and max times on testnet

teor2345:
## Motivation

The `min_time` field in the `getblocktemplate` RPC is not being calculated correctly on testnet.

Closes #5871

### Specifications

> On Testnet from block height 299188 onward, the difficulty adjustment algorithm [6](https://zips.z.cash/zip-0208#protocol-diffadjustment) allows minimum-difficulty blocks, as described in [10](https://zips.z.cash/zip-0208#zip-0205), when the block time is greater than a given threshold. This specification changes this threshold to be proportional to the block target spacing.
>
> That is, if the block time of a block at height height ≥ 299188 is greater than 6 · PoWTargetSpacing(height) seconds after that of the preceding block, then the block is a minimum-difficulty block. In that case its nBits field MUST be set to ToCompact(PoWLimit), where PoWLimit is the value defined for Testnet in section [5](https://zips.z.cash/zip-0208#protocol-constants).3 of the Zcash Protocol Specification 5, and ToCompact is as defined in section [7](https://zips.z.cash/zip-0208#protocol-nbits).7.4 of that specification 7.
>
> ...
>
> This change does not affect Mainnet.

This is a consensus rule, and the calculations are a bit complex. But it only applies to testnet, so it's less risky.

## Solution

- Use the tip block for the `min_time` calculation (it's actually the first block in reverse order)
- Only adjust times within the min and max range specified by the other consensus rules
- Improve documentation

## Review

Anyone can review this PR.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/read/difficulty.rs
+++ b/zebra-state/src/service/read/difficulty.rs
@@ -1,6 +1,9 @@
 //! Get context and calculate difficulty for the next block.
 
-use std::sync::Arc;
+use std::{
+    cmp::{max, min},
+    sync::Arc,
+};
 
 use chrono::{DateTime, Utc};
 
@@ -117,7 +120,9 @@ pub fn solution_rate(
 }
 
 /// Do a consistency check by checking the finalized tip before and after all other database queries.
-/// Returns and error if the tip obtained before and after is not the same.
+///
+/// Returns the state tip, recent blocks in reverse order from the tip, and the tip history tree.
+/// Returns an error if the tip obtained before and after is not the same.
 ///
 /// # Panics
 ///
@@ -172,6 +177,8 @@ fn relevant_chain_and_history_tree(
 }
 
 /// Returns the [`GetBlockTemplateChainInfo`] for the current best chain.
+///
+/// The `relevant_chain` has recent blocks in reverse order from the tip.
 ///
 /// See [`get_block_template_chain_info()`] for details.
 fn difficulty_time_and_history_tree(
@@ -245,6 +252,8 @@ fn difficulty_time_and_history_tree(
 }
 
 /// Adjust the difficulty and time for the testnet minimum difficulty rule.
+///
+/// The `relevant_data` has recent block difficulties and times in reverse order from the tip.
 fn adjust_difficulty_and_time_for_testnet(
     result: &mut GetBlockTemplateChainInfo,
     network: Network,
@@ -276,7 +285,8 @@ fn adjust_difficulty_and_time_for_testnet(
     // There is still a small chance that miners will produce an invalid block, if they are
     // just below the max time, and don't check it.
 
-    let previous_block_time = relevant_data.last().expect("has at least one block").1;
+    // The tip is the first relevant data block, because they are in reverse order.
+    let previous_block_time = relevant_data.first().expect("has at least one block").1;
     let previous_block_time: DateTime32 = previous_block_time
         .try_into()
         .expect("valid blocks have in-range times");
@@ -309,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
         .expect("a valid block time minus a small constant is in-range");
 
     if result.cur_time <= local_std_difficulty_limit {
-        // Standard difficulty: the max time needs to exclude min difficulty blocks
-        result.max_time = std_difficulty_max_time;
-    } else {
-        // Minimum difficulty: the min and cur time need to exclude min difficulty blocks
-        result.min_time = min_difficulty_min_time;
-        if result.cur_time < min_difficulty_min_time {
-            result.cur_time = min_difficulty_min_time;
-        }
+        // Standard difficulty: the cur and max time need to exclude min difficulty blocks
 
-        // And then the difficulty needs to be updated for cur_time
+        // The maximum time can only be decreased, and only as far as min_time.
+        // The old minimum is still required by other consensus rules.
+        result.max_time = min(std_difficulty_max_time, result.max_time);
+        result.max_time = max(result.min_time, result.max_time);
+
+        // The current time only needs to be decreased if the max_time decreased past it.
+        // Decreasing the current time can't change the difficulty.
+        result.cur_time = result.cur_time.clamp(result.min_time, result.max_time);
+    } else {
+        // Minimum difficulty: the min and cur time need to exclude std difficulty blocks
+
+        // The minimum time can only be increased, and only as far as max_time.
+        // The old maximum is still required by other consensus rules.
+        result.min_time = max(min_difficulty_min_time, result.min_time);
+        result.min_time = min(result.max_time, result.min_time);
+
+        // The current time only needs to be increased if the min_time increased past it.
+        result.cur_time = result.cur_time.clamp(result.min_time, result.max_time);
+
+        // And then the difficulty needs to be updated for cur_time.
         result.expected_difficulty = AdjustedDifficulty::new_from_header_time(
             result.cur_time.into(),
             previous_block_height,
--- a/zebra-state/src/service/read/difficulty.rs
+++ b/zebra-state/src/service/read/difficulty.rs
@@ -1,9 +1,6 @@
 //! Get context and calculate difficulty for the next block.
 
-use std::{
-    cmp::{max, min},
-    sync::Arc,
-};
+use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
 
--- a/zebra-state/src/service/read/difficulty.rs
+++ b/zebra-state/src/service/read/difficulty.rs
@@ -323,8 +323,7 @@ fn adjust_difficulty_and_time_for_testnet(
 
         // The maximum time can only be decreased, and only as far as min_time.
         // The old minimum is still required by other consensus rules.
-        result.max_time = min(std_difficulty_max_time, result.max_time);
-        result.max_time = max(result.min_time, result.max_time);
+        result.max_time = std_difficulty_max_time.clamp(result.min_time, result.max_time);
 
         // The current time only needs to be decreased if the max_time decreased past it.
         // Decreasing the current time can't change the difficulty.
@@ -334,8 +333,7 @@ fn adjust_difficulty_and_time_for_testnet(
 
         // The minimum time can only be increased, and only as far as max_time.
         // The old maximum is still required by other consensus rules.
-        result.min_time = max(min_difficulty_min_time, result.min_time);
-        result.min_time = min(result.max_time, result.min_time);
+        result.min_time = min_difficulty_min_time.clamp(result.min_time, result.max_time);
 
         // The current time only needs to be increased if the min_time increased past it.
         result.cur_time = result.cur_time.clamp(result.min_time, result.max_time);
--- a/zebra-state/src/service/read/difficulty.rs
+++ b/zebra-state/src/service/read/difficulty.rs
@@ -26,6 +26,11 @@ use crate::{
     BoxError, GetBlockTemplateChainInfo,
 };
 
+/// The amount of extra time we allow for a miner to mine a standard difficulty block on testnet.
+///
+/// This is a Zebra-specific standard rule.
+pub const EXTRA_TIME_TO_MINE_A_BLOCK: u32 = POST_BLOSSOM_POW_TARGET_SPACING * 2;
+
 /// Returns the [`GetBlockTemplateChainInfo`] for the current best chain.
 ///
 /// # Panics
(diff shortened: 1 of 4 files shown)


zebra-state/src/service/read/difficulty.rs
```diff
@@ -275,7 +285,8 @@ fn adjust_difficulty_and_time_for_testnet(
     // There is still a small chance that miners will produce an invalid block, if they are
     // just below the max time, and don't check it.
 
-    let previous_block_time = relevant_data.last().expect("has at least one block").1;
+    // The tip is the first relevant data block, because they are in reverse order.
+    let previous_block_time = relevant_data.first().expect("has at least one block").1;
```
teor2345, line 291: This is the biggest fix, we need the tip time here.

zebra-state/src/service/read/difficulty.rs
```diff
@@ -308,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
...
+        // Standard difficulty: the cur and max time need to exclude min difficulty blocks
+
+        // The maximum time can only be decreased, and only as far as min_time.
+        // The old minimum is still required by other consensus rules.
+        result.max_time = min(std_difficulty_max_time, result.max_time);
+        result.max_time = max(result.min_time, result.max_time);
+
+        // The current time only needs to be decreased if the max_time decreased past it.
+        // Decreasing the current time can't change the difficulty.
+        result.cur_time = result.cur_time.clamp(result.min_time, result.max_time);
```
teor2345: These fixes only matter if the last block's time is particularly large or small. (Which can happen, because the other consensus rules use the median time of recent blocks, not the tip block time.)

teor2345:
This fix gives us an exact match with `zcashd`:

```
$ ZCASH_CLI="zcash-cli-testnet" zcash-rpc-diff 38232 getblocktemplate
Connected to zebrad (port 38232) and zcashd (zcash-cli-testnet zcash.conf port).

Request:
getblocktemplate

Querying zebrad test chain at height >=2180007...

real    0m0.011s
user    0m0.005s
sys     0m0.006s

Querying zcashd test chain at height >=2180007...

real    0m0.023s
user    0m0.007s
sys     0m0.004s


--- /run/user/1000/tmp.0PycfC3DMm.rpc-diff/zebrad-test-2180007-getblocktemplate.json    2023-01-06 17:50:22.818172323 +1000
+++ /run/user/1000/tmp.0PycfC3DMm.rpc-diff/zcashd-test-2180007-getblocktemplate.json    2023-01-06 17:50:22.841172082 +1000
@@ -1,5 +1,6 @@
 {
   "capabilities": [
+    "proposal"
   ],
   "version": 4,
   "previousblockhash": "002d967dc24ef2ef34eb977fbfd67ef679f4fadf4c8caf84c3ba7d1dc9bf53b2",
@@ -24,7 +25,7 @@
     "sigops": 1,
     "required": true
   },
-  "longpollid": "00021800077623f6ef1672991854000000000000000000",
+  "longpollid": "002d967dc24ef2ef34eb977fbfd67ef679f4fadf4c8caf84c3ba7d1dc9bf53b226534",
   "target": "012bae0000000000000000000000000000000000000000000000000000000000",
   "mintime": 1672991128,
   "mutable": [
@@ -37,6 +38,5 @@
   "sizelimit": 2000000,
   "curtime": 1672991422,
   "bits": "20012bae",
-  "height": 2180008,
-  "maxtime": 1672991854
+  "height": 2180008
 }
```

Proposal is in PR #5870, long polling is allowed to be different, and maxtime is not implemented by `zcashd`.

teor2345:
Same network error as:
https://github.com/ZcashFoundation/zebra/pull/5938#issuecomment-1378201078

teor2345:
@mergifyio update

teor2345:
@arya2 would you mind giving this PR a review some time in the next few days?

zebra-state/src/service/read/difficulty.rs
```diff
@@ -308,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
...
 
     if result.cur_time <= local_std_difficulty_limit {
-        // Standard difficulty: the max time needs to exclude min difficulty blocks
-        result.max_time = std_difficulty_max_time;
+        // Standard difficulty: the cur and max time need to exclude min difficulty blocks
+
+        // The maximum time can only be decreased, and only as far as min_time.
+        // The old minimum is still required by other consensus rules.
+        result.max_time = min(std_difficulty_max_time, result.max_time);
+        result.max_time = max(result.min_time, result.max_time);
```
arya2: Optional:
```suggestion
        result.max_time = std_difficulty_max_time.clamp(result.min_time, result.max_time);
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -308,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
...
-        result.min_time = min_difficulty_min_time;
-        if result.cur_time < min_difficulty_min_time {
-            result.cur_time = min_difficulty_min_time;
-        }
+        // Minimum difficulty: the min and cur time need to exclude std difficulty blocks
+
+        // The minimum time can only be increased, and only as far as max_time.
+        // The old maximum is still required by other consensus rules.
+        result.min_time = max(min_difficulty_min_time, result.min_time);
+        result.min_time = min(result.max_time, result.min_time);
```
arya2: Optional:
```suggestion
        result.min_time = min_difficulty_min_time.clamp(result.min_time, result.max_time);
```

zebra-state/src/service/read/difficulty.rs
```diff
@@ -308,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
         .expect("a valid block time minus a small constant is in-range");
 
     if result.cur_time <= local_std_difficulty_limit {
-        // Standard difficulty: the max time needs to exclude min difficulty blocks
```
arya2, line 311: We may want to update this comment a few lines above:
> - if cur_time gets clamped to max_time, this is already a minimum difficulty block.

for when `max_time` from `median_time_past` is behind the `std_difficulty_max_time` from the previous block time.

This wasn't changed in this PR but I thought the condition was `result.cur_time <=  min_difficulty_min_time`, why subtract `POST_BLOSSOM_POW_TARGET_SPACING * 2` from the `std_difficulty_max_time`?

zebra-state/src/service/read/difficulty.rs
```diff
@@ -308,16 +319,28 @@ fn adjust_difficulty_and_time_for_testnet(
...
+        result.min_time = max(min_difficulty_min_time, result.min_time);
+        result.min_time = min(result.max_time, result.min_time);
+
+        // The current time only needs to be increased if the min_time increased past it.
+        result.cur_time = result.cur_time.clamp(result.min_time, result.max_time);
 
-        // And then the difficulty needs to be updated for cur_time
+        // And then the difficulty needs to be updated for cur_time.
         result.expected_difficulty = AdjustedDifficulty::new_from_header_time(
             result.cur_time.into(),
```
arya2, line 347: No change needed here but I wonder if we could set the `min_time` to `min_difficulty_min_time` as long as `max_time` is ahead of `std_difficulty_max_time` then use `min_difficulty_min_time` here instead of `cur_time`.

i.e. tell miners to use the minimum difficulty threshold then submit blocks with a timestamp ahead of their local clock.

https://github.com/ZcashFoundation/zebra/blob/691738226938a66c45527c5004b5710a294c6f0d/zebra-state/src/service/check.rs#L101-L106

arya2 (dismissed):
Looks good!

zebra-state/src/service/read/difficulty.rs
teor2345: I'm not sure if this is any easier to understand, maybe I'll need to fix the comments as well?

zebra-state/src/service/read/difficulty.rs
teor2345: The miner is going to take approximately ~75 seconds to mine the next block from this template, so there's no point starting a difficult block if we're close to doing an easy one.

I'll make a constant for this.

zebra-state/src/service/read/difficulty.rs
arya2: Maybe `result.max_time.clamp(result.min_time, std_difficulty_max_time);` would match the comment better than that suggestion, but I think the code and comments are clear in any case. `zcashd` uses min/max, either works, I just found it slightly verbose.

zebra-state/src/service/read/difficulty.rs
teor2345: > I wonder if we could set the `min_time` to `min_difficulty_min_time` as long as `max_time` is ahead of `std_difficulty_max_time`

You might have already worked this out, but this is almost always true, at least for the first few blocks, or if a node is a minority miner. The `max_time` is 90 minutes ahead of the median, and the `std_difficulty_max_time` is 7.5 minutes ahead of the previous block.

So if we want to test in both modes, we need to support mining both standard and minimum difficulty blocks.

zebra-state/src/service/read/difficulty.rs
teor2345: Fixed in 9842f52

arya2 (approved):

teor2345:
Failed in the merge queue due to a macOS timeout after downloading the Zcash parameters. Maybe we could increase that timeout:
https://github.com/ZcashFoundation/zebra/actions/runs/3932482431/jobs/6725103522

teor2345:
@mergifyio refresh

teor2345:
Failed in the merge queue due to #5958:
https://github.com/ZcashFoundation/zebra/actions/runs/3935154986/jobs/6730584282#step:9:2109

teor2345:
@mergifyio refresh

PR #5959, opened 2023-01-15
fix(clippy): Fix new lints in nightly clippy

teor2345:
## Motivation

We regularly fix new clippy lints to make new Rust releases easier to upgrade to, and to make Zebra easier to maintain.

## Solution

- derive Default in more places
- implement PartialEq manually when Hash is implemented manually
- document some functions to resolve missing docs warnings in Docker builds

## Review

Anyone can review this PR, it is a low priority.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -10,8 +10,6 @@
 //!
 //! A root of a note commitment tree is associated with each treestate.
 
-#![allow(clippy::derive_hash_xor_eq)]
-
 use std::{
     fmt,
     hash::{Hash, Hasher},
@@ -99,7 +97,7 @@ lazy_static! {
 /// The root hash in LEBS2OSP256(rt) encoding of the Orchard note commitment
 /// tree corresponding to the final Orchard treestate of this block. A root of a
 /// note commitment tree is associated with each treestate.
-#[derive(Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Default, Eq, Serialize, Deserialize)]
 pub struct Root(#[serde(with = "serde_helpers::Base")] pub(crate) pallas::Base);
 
 impl fmt::Debug for Root {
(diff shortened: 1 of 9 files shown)


zebra-chain/src/orchard/tree.rs
```diff
@@ -128,6 +126,13 @@ impl Hash for Root {
     }
 }
 
+impl PartialEq for Root {
+    fn eq(&self, other: &Self) -> bool {
+        // TODO: should we compare canonical forms here using `.to_repr()`?
```
arya2, line 131: Is there a benefit to comparing the canonical forms here? If not, I think it's preferable to avoid the compute cost.

zebra-chain/src/orchard/tree.rs
```diff
@@ -99,7 +97,7 @@ lazy_static! {
 /// The root hash in LEBS2OSP256(rt) encoding of the Orchard note commitment
 /// tree corresponding to the final Orchard treestate of this block. A root of a
 /// note commitment tree is associated with each treestate.
-#[derive(Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Default, Eq, Serialize, Deserialize)]
```
arya2, line 100: I wasn't able to reproduce the lint warning on nightly, are you sure we can't derive `PartialEq` here?

arya2 (approved):
Thank you for the proactive fixes!

zebra-chain/src/orchard/tree.rs
teor2345: If there is a non-canonical tree, then it could compare equal when hashed with `.to_repr()`, but unequal with `PartialEq`.

I think this is ok for now, because we always canonicalise our trees in Zebra. But I'm not sure enough, so I wanted to leave the question in a comment for later.

zebra-chain/src/orchard/tree.rs
teor2345: The problem is that the lint is renamed, so we have to use either the old or new name. But we also want to compile without warnings on stable and nightly.

The easiest solution is to fix the underlying problem the lint is identifying.

PR #5961, opened 2023-01-16
change(deps): Remove unused zebra-utils dependencies

teor2345:
## Motivation

`zebra-utils` only uses some constants from `zebra-state`, but it depends on the whole of `zebra-state`. (Including `rocksdb` and other large dependencies.)

It is much faster to compile the utilities if we minimise their dependencies. It also makes them easier to maintain and secure.

This is blocking RPC testing.

## Solution

Move some state constants into `zebra-node-services`, and update the corresponding dependency files and imports.

## Review

This is blocking RPC testing, it is a bit urgent.

### Reviewer Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5460,6 +5460,7 @@ dependencies = [
  "tracing-subscriber 0.3.16",
  "zcash_proofs",
  "zebra-chain",
+ "zebra-node-services",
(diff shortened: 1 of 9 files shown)


arya2 (approved):
Thank you for splitting this into its own PR!

PR #5963, opened 2023-01-16
change(ci): Run block proposal tests in CI

### LINKED ISSUE #5685: Test the getblocktemplate RPC's response as a block template proposal in CI

teor2345:
## Motivation

We want to use `getblocktemplate` in proposal mode in the acceptance test to check that it's producing valid templates.

Depends-On: #6027

Closes #5685

We want to manually test that block proposals work before merging this automated test.

We need to fix all known block proposal and mempool bugs before running this test in CI.

## Solution

- Test all possible valid times in block proposals (min, proposal current, max, and current clamped to [min, max])
- Run block proposal tests as part of the existing block template test in CI

## Review

Anyone can review this PR.

### Reviewer Checklist

### DIFF
--- a/zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
+++ b/zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
@@ -27,6 +27,10 @@ use crate::common::{
 /// We've seen it take anywhere from 1-45 seconds for the mempool to have some transactions in it.
 pub const EXPECTED_MEMPOOL_TRANSACTION_TIME: Duration = Duration::from_secs(45);
 
+/// Number of times we want to try submitting a block template as a block proposal at an interval
+/// that allows testing the varying mempool contents.
+const NUM_SUCCESSFUL_BLOCK_PROPOSALS_REQUIRED: usize = 3;
+
 /// Launch Zebra, wait for it to sync, and check the getblocktemplate RPC returns without errors.
 pub(crate) async fn run() -> Result<()> {
     let _init_guard = zebra_test::init();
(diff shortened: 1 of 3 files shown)


teor2345:
That was a bad merge, let's see how this one goes

arya2:
Since most of the rejections only happen some of the time, what do you think about running trying the proposal 5-10 times at an interval and asserting that it the block proposal was accepted every time to check that it works consistently for various mempool contents?

teor2345:
> Since most of the rejections only happen some of the time, what do you think about running trying the proposal 5-10 times at an interval and asserting that it the block proposal was accepted every time to check that it works consistently for various mempool contents?

I'm not sure if we can ensure that the mempool and chain tip remain the same for every repeated proposal check?

We might also want to think about how we handle "previous block is not the tip block" errors in CI, which are expected when a new block arrives. We might need to check the tip is the same before and after the proposal before we fail the test.

teor2345:
I'm going to put this in draft until we deal with expected errors.

teor2345:
@mergifyio update

teor2345:
I applied the Docker build fix, feel free to overwrite that commit.

teor2345:
@mergifyio update

teor2345:
@mergifyio update

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
```diff
@@ -27,6 +27,10 @@ use crate::common::{
 /// We've seen it take anywhere from 1-45 seconds for the mempool to have some transactions in it.
 pub const EXPECTED_MEMPOOL_TRANSACTION_TIME: Duration = Duration::from_secs(45);
 
+/// Number of times we want to try submitting a block template as a block proposal at an interval
+/// that allows testing the varying mempool contents.
+const NUM_SUCCESSFUL_BLOCK_PROPOSALS_REQUIRED: usize = 3;
```
arya2: I think this should be higher to consistently catch the reappearance of bugs that we've fixed or other somewhat infrequent failures.

```suggestion
const NUM_SUCCESSFUL_BLOCK_PROPOSALS_REQUIRED: usize = 10;
```

teor2345: This will test 40 time variants of 10 templates, and take around 8 minutes.

Since the whole test suite takes around 2 hours, I think that's ok.

teor2345:
@mergifyio update

arya2 (approved):
🔥

Thank you!

Update: we need another approver because I was the last pusher (it needed a rebase).

teor2345:
> 🔥
> 
> Thank you!
> 
> Update: we need another approver because I was the last pusher (it needed a rebase).

I can't approve because I opened the PR, so it will have to be @oxarbitrage

teor2345:
> > 🔥
> > Thank you!
> > Update: we need another approver because I was the last pusher (it needed a rebase).
> 
> I can't approve because I opened the PR, so it will have to be @oxarbitrage or @dconnolly 

If this is more annoying than it's worth, maybe we can disable this rule? Did you want to talk about it at the engineering sync?

mpguerra (approved):

arya2:
> If this is more annoying than it's worth, maybe we can disable this rule? Did you want to talk about it at the engineering sync?

It doesn't happen very often so I don't mind, and tagging Pili seems like a good solution for next time. I'll ask about review-latency in the next sync though because that could use some improvement.

teor2345:
@mergifyio refresh

teor2345:
Codespell failed due to a network error, but it's not required to merge.

PR #5968, opened 2023-01-16
ci: add a test to check that the Docker image config works

### LINKED ISSUE #5168: Add Docker tests for default config and $ZEBRA_CONF_PATH
## Motivation

We want to test that our Docker build scripts configure Zebra correctly:
- check that $ZEBRA_CONF_PATH works
- make sure the default Docker config works with the latest Zebra version

## Suggested Solution

- move the default Docker config into a file 
- add that file to the Zebra config test directory, all configs in this directory are automatically tested with Zebra:
    - https://github.com/ZcashFoundation/zebra/tree/main/zebrad/tests/common/configs
- add a Docker test for the default config
- add a Docker test for $ZEBRA_CONF_PATH

Here's one possible test script:
```sh
# make sure Zebra runs with the default docker config
docker run ... | grep --max-count=1 'estimated progress to chain tip.*BeforeOverwinter'
docker stop ...

# make sure Zebra runs with a custom docker config
TODO: copy the file into docker?
export ZEBRA_CONF_PATH=`zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml`
docker run ... | grep --max-count=1 'v1.0.0-rc.2.toml'
docker stop ...
```

## Related Work

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/5163#discussion_r971283944_

The test script depends on PR #5733 merging.

gustavovalverde:
## Motivation

We want to test that our Docker build scripts configure Zebra correctly:

- check that $ZEBRA_CONF_PATH works
- make sure the default Docker config works with the latest Zebra version

Fixes #5168

## Solution

- Add a Docker test for the default config
- Add a Docker test for $ZEBRA_CONF_PATH

## Review

Anyone can review this, validating the test is using the right configuration

### Reviewer Checklist

## Follow Up Work

Use a custom file instead of the existing one living inside `common/configs`

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -316,14 +316,14 @@ jobs:
       - name: Run tests using the default config
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
           docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
           docker stop default-conf-tests
 
       - name: Run tests using the $ZEBRA_CONF_PATH
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad -c $ZEBRA_CONF_PATH start
           docker logs --tail all --follow variable-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'v1.0.0-rc.2.toml'
           docker stop variable-conf-tests
         env:
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -320,6 +320,19 @@ jobs:
           docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
           docker stop default-conf-tests
 
+
+  # Test that Zebra works using the $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test Zebra configuration file
+    runs-on: ubuntu-latest
+    needs: build
+    if: ${{ github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' && github.event.inputs.run-lwd-send-tx != 'true' }}
+    steps:
+      - name: Inject slug/short variables
+        uses: rlespinasse/github-slug-action@v4
+        with:
+          short-length: 7
+
       - name: Run tests using the $ZEBRA_CONF_PATH
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -316,12 +316,12 @@ jobs:
       - name: Run tests using the default config
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --rm --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'estimated progress to chain tip.*BeforeOverwinter'
+          docker run --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'estimated progress to chain tip.*BeforeOverwinter'
 
       - name: Run tests using the $ZEBRA_CONF_PATH
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run -e ZEBRA_CONF_PATH --rm --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'v1.0.0-rc.2.toml'
+          docker run -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'v1.0.0-rc.2.toml'
         env:
           ZEBRA_CONF_PATH: 'zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml'
 
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -316,14 +316,14 @@ jobs:
       - name: Run tests using the default config
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} start
           docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
           docker stop default-conf-tests
 
       - name: Run tests using the $ZEBRA_CONF_PATH
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad -c $ZEBRA_CONF_PATH start
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} -c $ZEBRA_CONF_PATH start
           docker logs --tail all --follow variable-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'v1.0.0-rc.2.toml'
           docker stop variable-conf-tests
         env:
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -316,12 +316,12 @@ jobs:
       - name: Run tests using the default config
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'estimated progress to chain tip.*BeforeOverwinter'
+          docker run --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
 
       - name: Run tests using the $ZEBRA_CONF_PATH
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --max-count=1 'v1.0.0-rc.2.toml'
+          docker run -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --extended-regexp --max-count=1 -e 'v1.0.0-rc.2.toml'
         env:
           ZEBRA_CONF_PATH: 'zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml'
 
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -316,12 +316,16 @@ jobs:
       - name: Run tests using the default config
         run: |
           docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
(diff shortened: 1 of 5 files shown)


.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,36 @@ jobs:
...
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
+          docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
+          docker stop default-conf-tests
+
+      - name: Run tests using the $ZEBRA_CONF_PATH
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad -c $ZEBRA_CONF_PATH start
```
teor2345: I think we also might need a step that runs the Docker image without any arguments.

Thats what most users do, and I'm not sure we have a test for it yet.

teor2345: I think just passing the new conf in an environmental variable should work?
```suggestion
          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
```

gustavovalverde: I opened this discussion to decide before making more changes on the actual behavior https://github.com/ZcashFoundation/zebra/discussions/5972#discussion-4762947

teor2345 (approved):
Looks good, let's see how it goes!

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,36 @@ jobs:
...
+    steps:
+      - name: Inject slug/short variables
+        uses: rlespinasse/github-slug-action@v4
+        with:
+          short-length: 7
+
+      - name: Run tests using the default config
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} start
```
teor2345: Looks like the command is wrong here:
```
+ exec start
/entrypoint.sh: line 96: exec: start: not found
```

https://github.com/ZcashFoundation/zebra/actions/runs/3934277677/jobs/6730882649#step:3:174

```suggestion
          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,36 @@ jobs:
...
+          docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
+          docker stop default-conf-tests
+
+      - name: Run tests using the $ZEBRA_CONF_PATH
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} -c $ZEBRA_CONF_PATH start
```
teor2345, line 342:
```suggestion
          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad -c $ZEBRA_CONF_PATH start
```

teor2345 (dismissed):
I don't think the test commands are quite right yet.

We also need to add the new test to the branch protection rules, did you want to do that after this PR merges?

gustavovalverde:
In the `test` Docker stage we use `cargo test --locked --release --features lightwalletd-grpc-tests --workspace --no-run` which doesn't have `zebrad` as a binary

This is different from the `runtime` stage as it uses `cargo build --locked --release --features sentry --package zebrad --bin zebrad`

How should we run `zebrad` from the test stage @teor2345 ?

teor2345:
I'm not sure I fully understand the question, but I'll try to say some helpful things:

> In the `test` Docker stage we use `cargo test --locked --release --features lightwalletd-grpc-tests --workspace --no-run` which doesn't have `zebrad` as a binary

`--no-run` builds the binaries for the tests, without running any test binaries or the `zebrad` binary.

> This is different from the `runtime` stage as it uses `cargo build --locked --release --features sentry --package zebrad --bin zebrad`

This builds the `zebrad` release binary.

> How should we run `zebrad` from the test stage @teor2345 ?

The test stage should create a `zebrad` binary for the acceptance tests. It should be in a similar place to the release binary:
https://github.com/ZcashFoundation/zebra/blob/7e5253e27f77399962dd496551119b16a91dc7f3/docker/Dockerfile#L132

(It won't be in the debug directory, because we build tests with `--release`.)

I think one of the goals of the ticket is automating this manual check from the Zebra release process:
https://github.com/ZcashFoundation/zebra/blob/7e5253e27f77399962dd496551119b16a91dc7f3/.github/PULL_REQUEST_TEMPLATE/release-checklist.md?plain=1#L155

Is it possible to run this test on the release Docker image instead?
If it's not, or it's a lot of effort, that's ok. The test will still cover most of what we need.

gustavovalverde:
Actually, it's a lot easier to do it with the release image 😅

But if we're testing the actual PR or a stage before releasing, this will crash after we make the release, not before. But we can test it with the image from the main branch, which also uses the `runtime` stage

gustavovalverde:
@teor2345 so, `zebrad start` is working as expected as it's generating output and reaching the regex, but using `zebrad -c zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml start` is not working, it just hangs and get killed.

I'm not completely sure if this is an issue with the `zebrad` CLI, but this is reproducible even if I create a new configuration with the same configuration being used in the Dockerfile.

I went to our actual instances and can confirm those are also not generating output

docker/entrypoint.sh
```diff
@@ -15,7 +15,10 @@ echo "ZEBRA_CACHED_STATE_DIR=$ZEBRA_CACHED_STATE_DIR"
 echo "LIGHTWALLETD_DATA_DIR=$LIGHTWALLETD_DATA_DIR"
 
 case "$1" in
-    -- | cargo)
+    --* | -*)
```
teor2345, line 18: This entrypoint has 3 different behaviours:
- "zebrad -arguments" runs "zebrad -arguments"
- "-arguments" runs "zebrad -arguments"
- "" runs nothing

I wonder if that will be confusing, can we run zebrad with no arguments as well?

```suggestion
    --* | -* | "")
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,48 @@ jobs:
...
+      - name: Run tests using the default config
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
+          docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
+          docker stop default-conf-tests
+
+  # Test that Zebra works using the $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test Zebra configuration file
```
teor2345: Optional: Would it be easier to maintain if we make this new job two steps in one job, rather than two new jobs?

Two jobs means two new branch protection rules, and two statuses for developers to check.

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,48 @@ jobs:
...
+        uses: rlespinasse/github-slug-action@v4
+        with:
+          short-length: 7
+
+      - name: Run tests using the $ZEBRA_CONF_PATH
+        run: |
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} -c $ZEBRA_CONF_PATH start
+          docker logs --tail all --follow variable-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'v1.0.0-rc.2.toml'
+          docker stop variable-conf-tests
```
teor2345: I think there are two cases that this test needs to handle differently:
- if `zebrad` keeps running, the job needs to stop and fail 
- if `zebrad` stops and doesn't read the correct config file, the job needs to fail

teor2345:
> @teor2345 so, `zebrad start` is working as expected as it's generating output and reaching the regex, but using `zebrad -c zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml start` is not working, it just hangs and get killed.

How do you know that it is hanging? It looks to me like it finishes, but doesn't produce any output:
```
+ exec zebrad -c zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml start
tee: 'standard output': Broken pipe
variable-conf-tests
```

https://github.com/ZcashFoundation/zebra/actions/runs/3953227507/jobs/6769921730#step:3:168

gustavovalverde:
> How do you know that it is hanging? It looks to me like it finishes, but doesn't produce any output:

@teor2345 I built and ran the commands locally, even inside the container, and the container exits with an error after it hangs for around a minute.

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,50 @@ jobs:
...
+        with:
+          short-length: 7
+
+      - name: Run tests using the default config
+        run: |
+          set -x
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
+          docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
+          docker stop default-conf-tests
```
teor2345, line 321: Something else that might help with debugging:
```suggestion
          docker stop default-conf-tests
          docker logs default-conf-tests
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,50 @@ jobs:
...
+        with:
+          short-length: 7
+
+      - name: Run tests using the $ZEBRA_CONF_PATH
+        run: |
+          set -x
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach -e ZEBRA_CONF_PATH --name variable-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} -c $ZEBRA_CONF_PATH start
+          docker logs --tail all --follow variable-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'v1.0.0-rc.2.toml'
+          docker stop variable-conf-tests
```
teor2345, line 344: Something else that might help with debugging:
```suggestion
          docker stop variable-conf-tests
          docker logs variable-conf-tests
```

teor2345:
> > How do you know that it is hanging? It looks to me like it finishes, but doesn't produce any output:
> 
> @teor2345 I built and ran the commands locally, even inside the container, and the container exits with an error after it hangs for around a minute.

It looks like all the commands in the workflow and entrypoint are finishing:
```
+ exec zebrad -c zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml start
tee: 'standard output': Broken pipe
+ docker stop variable-conf-tests
variable-conf-tests
```
https://github.com/ZcashFoundation/zebra/actions/runs/3963283764/jobs/6791366909#step:3:154

But if `zebrad` was hanging, I think `docker logs --follow` would continue to run, and so `docker stop` would never run.
Can you show me the logs of `zebrad` hanging and getting killed?

gustavovalverde:
If you run `docker run zfnd/zebra` does it generate output or starts syncing? (I'm in a Mac with M1 and it doesn't, and the ones deployed on GCP are not syncing either)

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,52 @@ jobs:
         env:
           ZEBRA_TEST_LIGHTWALLETD: '1'
 
+  # Test that Zebra works using the default config with the latest Zebra version
+  #
+  # Also check that $ZEBRA_CONF_PATH works
+  test-configuration-file:
+    name: Test Zebra configuration file
```
teor2345: Nitpick: I had some trouble finding the right job in the summary view

```suggestion
    name: Test Zebra default Docker config file
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -299,6 +299,52 @@ jobs:
...
+          set -x
+          docker pull ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run --detach --name default-conf-tests -t ${{ env.GAR_BASE }}/${{ env.IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} zebrad start
+          docker logs --tail all --follow default-conf-tests | tee --output-error=exit /dev/stderr | grep --extended-regexp --max-count=1 -e 'estimated progress to chain tip.*BeforeOverwinter'
+          docker stop default-conf-tests
+          docker logs default-conf-tests
+
+  # Test that Zebra works using the $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test Zebra configuration file
```
```suggestion
    name: Test Zebra custom Docker config file
```

teor2345:
> If you run `docker run zfnd/zebra` does it generate output or starts syncing? (I'm in a Mac with M1 and it doesn't, and the ones deployed on GCP are not syncing either)

Since this PR changes the entry point script, I'm not sure if running the main branch image will help us here.

Here is the output from Run tests using the $ZEBRA_CONF_PATH: 
```
+ exec zebrad -c zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml start
error: Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file. You can generate a valid config by running "zebrad generate", and diff it against yours to examine any format inconsistencies.
error: zebrad fatal error: config error: path error: zebra/zebrad/tests/common/configs/v1.0.0-rc.2.toml
```
https://github.com/ZcashFoundation/zebra/actions/runs/3963766811/jobs/6792068479#step:3:174

It looks like the config file isn't at the expected path - where is the Zebra git checkout in the Docker image?

Also, tee isn't echoing the output before it exits - since tee works with longer "docker logs" output and with other tools, it could be a bug in "docker logs" where it doesn't send an end of file or flush its internal buffers before finishing. But printing all the logs at the end seems like enough for now. I don't think we need any further fixes for the log output.

.github/workflows/continous-integration-docker.yml
teor2345: > * if `zebrad` keeps running, the job needs to stop and fail

This is optional, these kinds of failures are rare, and the GitHub workflow timeout will catch it eventually.

> * if `zebrad` stops and doesn't read the correct config file, the job needs to fail

I think this might need to be fixed, we could do it by storing the result of the `grep` using:
```sh
... grep ...
EXIT_STATUS=$?
...
exit $EXIT_STATUS
```

docker/entrypoint.sh
gustavovalverde: > * "zebrad -arguments" runs "zebrad -arguments"
> * "-arguments" runs "zebrad -arguments"
> * "" runs nothing

We have to segregate this in two big scenarios:
1. The only image with an `entrypoint.sh` is the test image
2. When running `docker run zfnd/zebra` there's no `entrypoint.sh` and the command (`CMD`) at the end is established as `zebrad -c "${ZEBRA_CONF_PATH}" start` which will be overridden by any other command when running `docker run zfnd/zebra <command>`

From a `docker` perspective we have the following scenarios for the `tests` image:
- `docker run <test_image>`: this will _run nothing_ based on the entrypoint

https://github.com/ZcashFoundation/zebra/pull/5968/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7caR21

- `docker run <test_image> -<argument>` and docker run <test_image> --<argument> will run zebrad + the following arguments

gustavovalverde:
> Since this PR changes the entry point script, I'm not sure if running the main branch image will help us here.

The `main` branch does not build the Docker image with the entrypoint (is not included in the `runtime` stage), so this PR won't have an impact there, even after merged.

> https://github.com/ZcashFoundation/zebra/actions/runs/3963766811/jobs/6792068479#step:3:174
> 
> It looks like the config file isn't at the expected path - where is the Zebra git checkout in the Docker image?

Right here https://github.com/ZcashFoundation/zebra/blob/feat-conf-test/docker/Dockerfile#L105

This are the result of a `pwd` and `ls`
image

docker/entrypoint.sh
teor2345: Ok, I'm happy with whatever changes you want to make here, or if you just want to leave it as it is.

teor2345 (dismissed):
This looks like it should work.

Did you want to add the new jobs to the branch protection rules after it merges?

teor2345:
We'll also need to update the patch workflows with the new jobs 🙂

gustavovalverde:
@teor2345 sorry that I dismissed your review with this change
- https://github.com/ZcashFoundation/zebra/pull/5968/commits/1beaebd875da34abf968cea6d24c271b7bf1fb3a

Please have a look at it, as this will ensure we test the Dockerfile on both stages: `tests` and `runtime`, as the build process is different and this ensures it works in both cases.

gustavovalverde:
Manual deployment to test `runtime` stage https://github.com/ZcashFoundation/zebra/actions/runs/3982611624

teor2345 (approved):
Looks great, thanks for adding that extra test!

We'll have to manually check the test status in the CD workflow, but that's ok, that's what we're doing for everything else in that workflow anyway.

teor2345:
I added these new jobs to the `main` branch protection rules:
* Test Zebra default Docker config file
* Test Zebra custom Docker config file

Issue #5981, opened 2023-01-17
Use the same format as `zcashd` for `getblocktemplate` proposal mode errors 


teor2345:
## Motivation

When I use `zebra-utils/zcash-rpc-block-template-to-proposal` from PR #5944 to send a block proposal from Zebra to both `zcashd` and Zebra, I see these differences in the response format:
```
Proposal response diffs between ports  8232 6666 and time sources CurTime MinTime MaxTime ClampedNow:
--- /tmp/tmp.UWcAoSvcUZ.block-template-proposal/proposal-check-getblocktemplate-proposal.CurTime.8232.json	2023-01-26 15:15:37.881314992 -0300
+++ /tmp/tmp.UWcAoSvcUZ.block-template-proposal/proposal-check-getblocktemplate-proposal.MaxTime.6666.json	2023-01-26 15:15:38.329290891 -0300
@@ -0,0 +1,5 @@
+invalid proposal: Block {
+    source: ValidateProposal(
+        "proposal is not based on the current best chain tip: previous block hash must be the best chain tip",
+    ),
+}
```

Zebra should return a string to match the `zcashd` response format, but it is currently returning JSON. For example zcashd returns errors like `bad-txns-inputs-missingorspent`

### Specifications

We were following the specifications in the bitcoin improvement proposals here, but being compatible with the `zcashd` implementation is a higher priority.

It is enough to return any error string in `kebab-case` format - we don't need to match the exact error wording.

For the example above about best tip error zebra should return something like:

`invalid-proposal-proposal-is-not-based-on-the-current-best-chain-tip-previous-block-hash–must–be-the–best-chain-tip`

### Complex Code or Requirements

This should be a simple format change.

### Testing

We can use the `zebra-utils/zcash-rpc-block-template-to-proposal` script to manually re-test to make sure this bug is gone. We'll need to wait for an error.

(The previous block error happens if a new block has just arrived, or there is a chain fork.)

teor2345:
@mpguerra we might want to schedule this fix in this sprint, the different outputs could make testing harder for everyone, because we have to check the differences every time we run the script. But this is not a consensus rule bug, so it's not as important as #5982.

@arya2 what do you think?

arya2:
> @mpguerra we might want to schedule this fix in this sprint, the different outputs could make testing harder for everyone, because we have to check the differences every time we run the script. But this is not a consensus rule bug, so it's not as important as #5982.
> 
> @arya2 what do you think?

[...]

PR #5993, opened 2023-01-18
fix(rpc): Return detailed errors to the RPC client when a block proposal fails

teor2345:
## Motivation

We are currently testing Zebra's block templates in ticket #5803. But I can't work out if Zebra's errors are expected or the same as `zcashd`, because it doesn't return any error information at all.

This PR changes the error handling from bug #5981, it probably gets us closer to a fix.
(But to match `zcashd` we'd need to stop pretty printing, and change the format of the error string.)

### Specifications

We can do whatever we like here, because bug #5981 will fix up any mining pool compatibility issues.

## Solution

- Send the RPC client the kind of error, and the detailed error debug information

## Review

This seems like a high priority because testing is difficult without it.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
@@ -2,13 +2,14 @@
 //!
 //! `ProposalResponse` is the output of the `getblocktemplate` RPC method in 'proposal' mode.
 
-use std::{num::ParseIntError, str::FromStr, sync::Arc};
+use std::{error::Error, num::ParseIntError, str::FromStr, sync::Arc};
 
 use zebra_chain::{
     block::{self, Block, Height},
     serialization::{DateTime32, SerializationError, ZcashDeserializeInto},
     work::equihash::Solution,
 };
+use zebra_node_services::BoxError;
 
 use crate::methods::{
     get_block_template_rpcs::types::{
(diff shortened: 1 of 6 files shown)


teor2345:
I tested this PR with `zcash-rpc-block-template-to-proposal`.

I now see detailed errors like:
```
invalid proposal: Block {
    source: ValidateProposal(
        "proposal is not based on the current best chain tip: previous block hash must be the best chain tip",
    ),
}
```

Rather than the same error every time:
```
{
  "reject_reason": "rejected",
  "capabilities": [
    "proposal"
  ]
}
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
```diff
@@ -18,47 +19,46 @@ use crate::methods::{
...
-        reject_reason: ProposalRejectReason,
-
-        /// The getblocktemplate RPC capabilities supported by Zebra.
-        capabilities: Vec<String>,
-    },
+    /// Block proposal was rejected as invalid.
+    /// Contains the reason that the proposal was invalid.
+    ///
+    /// TODO: turn this into a typed error enum?
+    Rejected(String),
```
arya2, line 37: The renamed `ErrorResponse` variant needs an updated reference in the acceptance test.

arya2 (commented):
This looks great! 

Only change needed is an updated reference in the acceptance test for the renamed `ErrorResponse`.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
teor2345: Thanks, see fde6d52 for a fix!

arya2 (approved):

Issue #5935, opened 2023-01-09
Manually test that the Zebra and `zcashd` mempools are similar


teor2345:
## Motivation

Zebra and `zcashd` should have similar transactions in their mempools:
- Zebra must only put valid transactions into mined blocks
- Users expect Zebra to relay and mine their transactions if `zcashd` does

### Specifications

Mempools are governed by two different kinds of rules:
- Consensus rules must be the same in Zebra and `zcashd`:
    - Zebra's `getblocktemplate` RPC must create valid blocks
    - Zebra must not accept transactions into its mempool that `zcashd` says are invalid when mined into blocks
- Standard rules can be different between node implementations, but:
    - Users will report bugs if their transactions don't get mined by Zebra
    - If Zebra becomes a large part of the network, users will report bugs if their transactions don't get relayed by Zebra

### Complex Code or Requirements

If Zebra's mempool accepts a transaction that `zcashd` rejects in mined blocks, this is a high-priority bug that we must fix. These bugs are blockers for activating the mining RPCs in production.

If Zebra's mempool rejects a transaction that `zcashd`'s mempool accepts, this is a medium-priority bug that we need to document or fix.

### Testing

I think we can use `zcash-rpc-diff` with the `getrawmempool` RPC to test this. We might need to match `zcashd`'s transaction order to minimise the diff.

If it's the same order as the `getblocktemplate` RPC, we can do a similar change to PR #5867.

Please check your box when you have successfully run the `zcash-rpc-diff` script with `getrawmempool` (see https://github.com/ZcashFoundation/zebra/pull/5994 for [running the test continuously](https://github.com/ZcashFoundation/zebra/issues/5803#issuecomment-1386250961) until there's a diff to examine):
- [ ] @oxarbitrage 
- [ ] @teor2345 
- [x] @arya2 

## Related Work

- #5686
- #5803
- #5934
- #5935

teor2345:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5935?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage

arya2:
I ran zcash-rpc-diff about 10-15 times where the Zebra and zcashd mempools contained the same transactions in all instances:

```diff

[...]

PR #5992, opened 2023-01-18
fix(deps): Add test dependency from zebra-rpc to zebra-network with correct features

teor2345:
## Motivation

If I build the tests for the individual `zebra-rpc` crate, the build fails due to a missing dependency.

This is a bug from PR #5951.

## Solution

Add the correct dependencies for the `proptest-impl` feature.

## Review

Anyone can review this code. It blocks some kinds of testing, so I think it might be a high priority.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

I don't think we need any extra CI checks for this build bug, because it just happens in test code.

PR #5994, opened 2023-01-18
change(rpc): Sort transaction hashes like zcashd in getrawmempool RPC response

arya2:
## Motivation

It would be convenient for the `getrawmempool` output to match zcashd's exactly so we can run zcash-rpc-diff continuously until there's a difference to examine.

## Solution

- Gets full transactions from mempool
- Sorts transactions by `fee/size` then id before returning `getrawmempool` results

## Review

Anyone can review. This doesn't have to merge, we could just use this branch for https://github.com/ZcashFoundation/zebra/issues/5935.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/transaction/hash.rs
+++ b/zebra-chain/src/transaction/hash.rs
@@ -60,7 +60,7 @@ use super::{txid::TxIdBuilder, AuthDigest, Transaction};
 ///
 /// [ZIP-244]: https://zips.z.cash/zip-0244
 /// [Spec: Transaction Identifiers]: https://zips.z.cash/protocol/protocol.pdf#txnidentifiers
-#[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Hash)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 pub struct Hash(pub [u8; 32]);
 
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -648,66 +648,20 @@ where
     #[cfg(feature = "getblocktemplate-rpcs")]
     // TODO: use a generic error constructor (#5548)
     fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
-        use std::cmp::Ordering;
-        use zebra_chain::transaction::VerifiedUnminedTx;
-
-        /// Returns relative score for sorting [`VerifiedUnminedTx`]s (fee/size)
-        fn tx_sort_score(tx: &VerifiedUnminedTx, tx2: &VerifiedUnminedTx) -> Option<u64> {
-            u64::try_from(tx2.transaction.size)
-                .expect("transaction size should fit in u64")
-                .checked_mul(u64::from(tx.miner_fee))
-        }
+        /// Determines whether the output of this RPC is sorted like zcashd
+        const SHOULD_USE_ZCASHD_ORDER: bool = true;
 
         let mut mempool = self.mempool.clone();
 
         async move {
-            let request = mempool::Request::FullTransactions;
-
-            // `zcashd` doesn't check if it is synced to the tip here, so we don't either.
-            let response = mempool
-                .ready()
-                .and_then(|service| service.call(request))
-                .await
-                .map_err(|error| Error {
-                    code: ErrorCode::ServerError(0),
-                    message: error.to_string(),
-                    data: None,
-                })?;
-
-            if let mempool::Response::FullTransactions(mut transactions) = response {
-                transactions.sort_by(|tx, tx2| {
-                    match tx_sort_score(tx2, tx).cmp(&tx_sort_score(tx, tx2)) {
-                        Ordering::Equal => {
-                            tracing::info!("tx_sort_scores are equal");
-
-                            tx.transaction
-                                .id
-                                .mined_id()
-                                .cmp(&tx2.transaction.id.mined_id())
-                        }
-                        gt_or_lt => gt_or_lt,
-                    }
-                });
-
-                let tx_ids: Vec<String> = transactions
-                    .iter()
-                    .map(|unmined_tx| unmined_tx.transaction.id.mined_id().encode_hex())
-                    .collect();
-
-                Ok(tx_ids)
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            let request = if SHOULD_USE_ZCASHD_ORDER {
+                mempool::Request::FullTransactions
             } else {
-                unreachable!("unmatched response to a mempool::FullTransactions request")
-            }
-        }
-        .boxed()
-    }
+                mempool::Request::TransactionIds
+            };
 
-    #[cfg(not(feature = "getblocktemplate-rpcs"))]
-    // TODO: use a generic error constructor (#5548)
-    fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
-        let mut mempool = self.mempool.clone();
-
-        async move {
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
             let request = mempool::Request::TransactionIds;
 
             // `zcashd` doesn't check if it is synced to the tip here, so we don't either.
@@ -722,6 +676,23 @@ where
                 })?;
 
             match response {
+                #[cfg(feature = "getblocktemplate-rpcs")]
+                mempool::Response::FullTransactions(mut transactions) => {
+                    transactions.sort_by_cached_key(|tx| {
+                        std::cmp::Reverse((
+                            u64::from(tx.miner_fee) / tx.transaction.size as u64,
+                            tx.transaction.id.mined_id(),
+                        ))
+                    });
+
+                    let tx_ids: Vec<String> = transactions
+                        .iter()
+                        .map(|unmined_tx| unmined_tx.transaction.id.mined_id().encode_hex())
+                        .collect();
+
+                    Ok(tx_ids)
+                }
+
                 mempool::Response::TransactionIds(unmined_transaction_ids) => {
                     let mut tx_ids: Vec<String> = unmined_transaction_ids
                         .iter()
@@ -729,11 +700,11 @@ where
                         .collect();
 
                     // Sort returned transaction IDs in numeric/string order.
-                    // (zcashd's sort order appears arbitrary.)
                     tx_ids.sort();
 
                     Ok(tx_ids)
                 }
+
                 _ => unreachable!("unmatched response to a transactionids request"),
             }
         }
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -656,11 +656,15 @@ where
         let mut mempool = self.mempool.clone();
 
         async move {
-            let request = if SHOULD_USE_ZCASHD_ORDER && cfg!(feature = "getblocktemplate-rpcs") {
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            let request = if SHOULD_USE_ZCASHD_ORDER {
                 mempool::Request::FullTransactions
             } else {
                 mempool::Request::TransactionIds
             };
+
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
+            let request = mempool::Request::TransactionIds;
 
             // `zcashd` doesn't check if it is synced to the tip here, so we don't either.
             let response = mempool
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -657,7 +657,7 @@ where
 
         async move {
             #[cfg(feature = "getblocktemplate-rpcs")]
-            let request = if SHOULD_USE_ZCASHD_ORDER {
+            let request = if SHOULD_USE_ZCASHD_ORDER && cfg!(not(test)) {
                 mempool::Request::FullTransactions
             } else {
                 mempool::Request::TransactionIds
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -645,9 +645,10 @@ where
             })
     }
 
-    #[cfg(feature = "getblocktemplate-rpcs")]
     // TODO: use a generic error constructor (#5548)
     fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
+        use zebra_chain::block::MAX_BLOCK_BYTES;
+
         /// Determines whether the output of this RPC is sorted like zcashd
         const SHOULD_USE_ZCASHD_ORDER: bool = true;
 
@@ -679,8 +680,11 @@ where
                 #[cfg(feature = "getblocktemplate-rpcs")]
                 mempool::Response::FullTransactions(mut transactions) => {
                     transactions.sort_by_cached_key(|tx| {
+                        // zcashd uses modified fee here but Zebra doesn't currently
+                        // support prioritizing transactions
                         std::cmp::Reverse((
-                            u64::from(tx.miner_fee) / tx.transaction.size as u64,
+                            i64::from(tx.miner_fee) as u128 * MAX_BLOCK_BYTES as u128
+                                / tx.transaction.size as u128,
                             // transaction hashes are compared in their serialized byte-order.
                             tx.transaction.id.mined_id(),
                         ))
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -681,6 +681,7 @@ where
                     transactions.sort_by_cached_key(|tx| {
                         std::cmp::Reverse((
                             u64::from(tx.miner_fee) / tx.transaction.size as u64,
+                            // transaction hashes are compared in their serialized byte-order.
                             tx.transaction.id.mined_id(),
                         ))
                     });
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -645,6 +645,64 @@ where
             })
     }
 
+    #[cfg(feature = "getblocktemplate-rpcs")]
+    // TODO: use a generic error constructor (#5548)
+    fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
+        use std::cmp::Ordering;
+        use zebra_chain::transaction::VerifiedUnminedTx;
+
+        /// Returns relative score for sorting [`VerifiedUnminedTx`]s (fee/size)
+        fn tx_sort_score(tx: &VerifiedUnminedTx, tx2: &VerifiedUnminedTx) -> Option<u64> {
+            u64::try_from(tx2.transaction.size)
+                .expect("transaction size should fit in u64")
+                .checked_mul(u64::from(tx.miner_fee))
+        }
+
+        let mut mempool = self.mempool.clone();
+
+        async move {
+            let request = mempool::Request::FullTransactions;
+
+            // `zcashd` doesn't check if it is synced to the tip here, so we don't either.
+            let response = mempool
+                .ready()
+                .and_then(|service| service.call(request))
+                .await
+                .map_err(|error| Error {
+                    code: ErrorCode::ServerError(0),
+                    message: error.to_string(),
+                    data: None,
+                })?;
+
+            if let mempool::Response::FullTransactions(mut transactions) = response {
+                transactions.sort_by(|tx, tx2| {
+                    match tx_sort_score(tx2, tx).cmp(&tx_sort_score(tx, tx2)) {
+                        Ordering::Equal => {
+                            tracing::info!("tx_sort_scores are equal");
+
+                            tx.transaction
+                                .id
+                                .mined_id()
+                                .cmp(&tx2.transaction.id.mined_id())
+                        }
+                        gt_or_lt => gt_or_lt,
+                    }
+                });
+
+                let tx_ids: Vec<String> = transactions
+                    .iter()
+                    .map(|unmined_tx| unmined_tx.transaction.id.mined_id().encode_hex())
+                    .collect();
+
+                Ok(tx_ids)
+            } else {
+                unreachable!("unmatched response to a mempool::FullTransactions request")
+            }
+        }
+        .boxed()
+    }
+
+    #[cfg(not(feature = "getblocktemplate-rpcs"))]
     // TODO: use a generic error constructor (#5548)
     fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
         let mut mempool = self.mempool.clone();
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -647,6 +647,7 @@ where
 
     // TODO: use a generic error constructor (#5548)
     fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
+        #[cfg(feature = "getblocktemplate-rpcs")]
         use zebra_chain::block::MAX_BLOCK_BYTES;
 
         /// Determines whether the output of this RPC is sorted like zcashd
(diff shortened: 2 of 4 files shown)


zebra-rpc/src/methods.rs
```diff
@@ -645,6 +645,64 @@ where
...
+                .and_then(|service| service.call(request))
+                .await
+                .map_err(|error| Error {
+                    code: ErrorCode::ServerError(0),
+                    message: error.to_string(),
+                    data: None,
+                })?;
+
+            if let mempool::Response::FullTransactions(mut transactions) = response {
+                transactions.sort_by(|tx, tx2| {
```
arya2: zcashd does this here: https://github.com/zcash/zcash/blob/8bc5740003d54da824aa680e8914acbfff7b2ef7/src/txmempool.h#L220-L228

zebra-chain/src/transaction/hash.rs
```diff
@@ -60,7 +60,7 @@ use super::{txid::TxIdBuilder, AuthDigest, Transaction};
 ///
 /// [ZIP-244]: https://zips.z.cash/zip-0244
 /// [Spec: Transaction Identifiers]: https://zips.z.cash/protocol/protocol.pdf#txnidentifiers
-#[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Hash)]
```
teor2345, line 63: Is the order here based on the display order, or the serialised order? (The byte order of the hash is reversed for display order.)

Transaction hashes don't have a natural order, so we might want to use a method that is named after the order we are providing. That would make reviews easier.

What do you think?

zebra-rpc/src/methods.rs
```diff
@@ -645,6 +645,64 @@ where
             })
     }
 
+    #[cfg(feature = "getblocktemplate-rpcs")]
```
teor2345: I'd like to avoid making two copies of this code, because that makes it harder to maintain. (And review!)

We could define a USE_ZCASHD_ORDER constant that depends on the feature flag?
Anything that reduces the code duplication would also be ok here.

zebra-rpc/src/methods.rs
```diff
@@ -645,6 +645,64 @@ where
...
+    // TODO: use a generic error constructor (#5548)
+    fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
+        use std::cmp::Ordering;
+        use zebra_chain::transaction::VerifiedUnminedTx;
+
+        /// Returns relative score for sorting [`VerifiedUnminedTx`]s (fee/size)
+        fn tx_sort_score(tx: &VerifiedUnminedTx, tx2: &VerifiedUnminedTx) -> Option<u64> {
```
teor2345: This seems like a direct translation of the C code into Rust, but it's not clear to me exactly how it is achieving the desired order. (Or if the code is doing it correctly.)

There are also more efficient ways to do the same sort in Rust, so our code can be simpler than the C version.

In Rust, one idiomatic and efficient way to do this is:
https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_cached_key

So we might be able to use something like:
```rust
if USE_ZCASHD_ORDER {
    transactions.sort_by_cached_key(|tx| {
        Reverse((
            tx.miner_fee as f32 / tx.transaction.size as f32,
            tx.transaction.id.mined_id().bytes_in_display_order(),
        ))
    });
}
```

Or maybe just with the Reverse around fee/size, if the hash is sorted in ascending order.

teor2345 (dismissed):
I think this looks good, but I can't see if the order matches zcashd.

We might also want to make it easier to maintain - zcashd might change their order when they implement ZIP-317.

Can you post some zcash-rpc-diff results that demonstrate fee/size and transaction hash ordering?

zebra-rpc/src/methods.rs
teor2345: We could also use f64 if you wanted to match the zcashd code exactly. The performance difference would be minimal.

arya2:
> Can you post some zcash-rpc-diff results that demonstrate fee/size and transaction hash ordering?

I haven't seen it compare transaction hashes yet but for fee/size:

```diff
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952462...

real    0m0.004s
user    0m0.003s
sys     0m0.000s

Querying zcashd main chain at height >=1952462...

real    0m0.004s
user    0m0.004s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.aTB2QNAr5z.rpc-diff/zebrad-main-1952462-getrawmempool.json:
[
  "a6d4cd5bc965d581f2b125da30971993f034b370ffb54bcefaa8af0d452fc1bc",
  "74979ecc884fe5a4f01f78dfe5b8d252fee7b0a11c96492deb61d10024520ece",
  "ab0269e0caf72e467e246d9127affa58b1a28f65bca472d34ded8d1d816cd9e7",
  "36f5a844625e83ef9c01bc3c2a6ce368c22e313717f79784eed9db849a38e41c",
  "49cf002385c1503ddd1255fd5d2a6420f448d7f98dd17d826b5e10d081f917f9"
]
```
More:
<details>

```diff

➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952462...

real    0m0.003s
user    0m0.002s
sys     0m0.000s

Querying zcashd main chain at height >=1952462...

real    0m0.005s
user    0m0.000s
sys     0m0.004s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.zp5PHPZImD.rpc-diff/zebrad-main-1952462-getrawmempool.json:
[
  "a6d4cd5bc965d581f2b125da30971993f034b370ffb54bcefaa8af0d452fc1bc",
  "74979ecc884fe5a4f01f78dfe5b8d252fee7b0a11c96492deb61d10024520ece",
  "ab0269e0caf72e467e246d9127affa58b1a28f65bca472d34ded8d1d816cd9e7",
  "36f5a844625e83ef9c01bc3c2a6ce368c22e313717f79784eed9db849a38e41c",
  "49cf002385c1503ddd1255fd5d2a6420f448d7f98dd17d826b5e10d081f917f9"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952462...

real    0m0.003s
user    0m0.003s
sys     0m0.000s

Querying zcashd main chain at height >=1952462...

real    0m0.004s
user    0m0.003s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.ufdDnxDee3.rpc-diff/zebrad-main-1952462-getrawmempool.json:
[
  "a6d4cd5bc965d581f2b125da30971993f034b370ffb54bcefaa8af0d452fc1bc",
  "74979ecc884fe5a4f01f78dfe5b8d252fee7b0a11c96492deb61d10024520ece",
  "ab0269e0caf72e467e246d9127affa58b1a28f65bca472d34ded8d1d816cd9e7",
  "36f5a844625e83ef9c01bc3c2a6ce368c22e313717f79784eed9db849a38e41c",
  "49cf002385c1503ddd1255fd5d2a6420f448d7f98dd17d826b5e10d081f917f9"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.004s
user    0m0.003s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.003s
user    0m0.003s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.NVA9QwrtE9.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.002s
user    0m0.002s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.002s
user    0m0.002s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.wVSWLneKAN.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.004s
user    0m0.004s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.004s
user    0m0.004s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.qyAMIxOqEr.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.004s
user    0m0.003s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.003s
user    0m0.003s
sys     0m0.000s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.WU1AOkIgo4.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.004s
user    0m0.003s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.002s
user    0m0.000s
sys     0m0.002s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.bA4SJYm7xj.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]
➜  zebra git:(sort-getrawmempool) ✗ ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 3000 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zebrad (port 3000) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zebrad network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zebrad main chain at height >=1952464...

real    0m0.004s
user    0m0.004s
sys     0m0.000s

Querying zcashd main chain at height >=1952464...

real    0m0.004s
user    0m0.000s
sys     0m0.004s


Response diff between zebrad and zcashd:
RPC responses were identical

/tmp/tmp.lA4ztbtrh9.rpc-diff/zebrad-main-1952464-getrawmempool.json:
[
  "18bbe79487e157af34c1b8d00d51cedb3dd07625d8b76b17eb58e9f8c870d11e",
  "6077962d04cda6a99502cdfaf71c38079ec4f17186d4e53f94f18b1a39a24ddb",
  "bd8881165be076b6e1a342356bc7194f239908e0e8d72c4e75466e6bcefae2b7"
]

```

</details>

zebra-chain/src/transaction/hash.rs
arya2: > Is the order here based on the display order, or the serialised order? (The byte order of the hash is reversed for display order.)

Looks like it's based on ~display order~ serialized order in zcashd. (both reverse the order before hex-encoding)

> Transaction hashes don't have a natural order, so we might want to use a method that is named after the order we are providing. That would make reviews easier.
>
> What do you think?

I think that's a good idea and would be clearer.

zebra-rpc/src/methods.rs
arya2: Floats don't implement `Ord` which is a constraint for `sort_by_key` and `sort_by_cached_key`, it could use `sort_by` with `partial_cmp` or u64 division if sorted by key.

I went with integer division in [`cd11c81`](https://github.com/ZcashFoundation/zebra/pull/5994/commits/cd11c81c67c53a3fbc710f330fabbf9bc2688132) but it doesn't work so I might go with precomputing the f32s and using `sort_by` or revert to the direct translation of zcashd's logic.

zebra-rpc/src/methods.rs
arya2: Okay, I defined a constant in the same method in [`cd11c81`](https://github.com/ZcashFoundation/zebra/pull/5994/commits/cd11c81c67c53a3fbc710f330fabbf9bc2688132), let me know if it should be at the top of the file or in another module.

zebra-chain/src/transaction/hash.rs
arya2: On second thought, it would be returning the same value as `mined_id()`, so I added a comment clarifying that we want the serialized byte order in [`25be33d`](https://github.com/ZcashFoundation/zebra/pull/5994/commits/25be33d9094bc671fef33e57a354f58214153f1e).

zebra-rpc/src/methods.rs
arya2: Updated in [`2d7cf50`](https://github.com/ZcashFoundation/zebra/pull/5994/commits/2d7cf5004d277ed2cfbb16d6be58767d79b7e1cb) to use `fee * MAX_BLOCK_BYTES / tx.size` with a `u128`. It works consistently now:
<details>

```diff

ZCASH_CLI=../zcash/src/zcash-cli ./zebra-utils/zcash-rpc-diff 8232 getrawmempool
Checking first node release info...
Checking second node release info...
Connected to zcashd (port 8232) and zcashd (../zcash/src/zcash-cli zcash.conf port).

Checking zcashd network and tip height...
Checking zcashd network and tip height...

Request:
getrawmempool

Querying zcashd main chain at height >=1953445...

real    0m0.002s
user    0m0.002s
sys     0m0.000s

Querying zcashd main chain at height >=1953445...

real    0m0.003s
user    0m0.002s
sys     0m0.000s


Response diff between zcashd and zcashd:
RPC responses were identical

/tmp/tmp.wweYYik30y.rpc-diff/zcashd-main-1953445-getrawmempool.json:
[
  "19fffa587781a93d7d15085471d93db845f94b4dd4f11f03aac0a00f3fa8e706",
  "e689e9255e00f6953a5cf0441aeb770ebc099152d906fb90da0ef2eb6a8324dd",
  "7d3530f2dfe41e28d9371d45b1e26424c56c9ce96cb01a9fc34ce7be54bee4bd",
  "97b521dacd6d98d28ca8c073a857b339389925cf122c672d3297da68832ab041"
]

```

</details>

zebra-rpc/src/methods.rs
```diff
@@ -647,9 +647,22 @@ where
...
         async move {
+            #[cfg(feature = "getblocktemplate-rpcs")]
+            let request = if SHOULD_USE_ZCASHD_ORDER {
+                mempool::Request::FullTransactions
+            } else {
+                mempool::Request::TransactionIds
+            };
+
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
             let request = mempool::Request::TransactionIds;
```
teor2345, line 668: Optional: this might be a bit clearer with `cfg!()`:
https://doc.rust-lang.org/std/macro.cfg.html

```suggestion
            let request = if SHOULD_USE_ZCASHD_ORDER && cfg!(feature = "getblocktemplate-rpcs") {
                mempool::Request::FullTransactions
            } else {
                mempool::Request::TransactionIds
            };
```

zebra-rpc/src/methods.rs
```diff
@@ -647,9 +647,22 @@ where
 
     // TODO: use a generic error constructor (#5548)
     fn get_raw_mempool(&self) -> BoxFuture<Result<Vec<String>>> {
+        use zebra_chain::block::MAX_BLOCK_BYTES;
```
teor2345, line 651:
```suggestion
        #[cfg(feature = "getblocktemplate-rpcs")]
        use zebra_chain::block::MAX_BLOCK_BYTES;
```

zebra-rpc/src/methods.rs
```diff
@@ -664,18 +677,39 @@ where
                 })?;
 
             match response {
+                #[cfg(feature = "getblocktemplate-rpcs")]
+                mempool::Response::FullTransactions(mut transactions) => {
+                    transactions.sort_by_cached_key(|tx| {
```
teor2345, line 685: Can you please document the sort order here?

I think it's something like:
```suggestion
                    // Sort transactions in descending order by fee/size, using hash in serialized byte order as a tie-breaker
                    transactions.sort_by_cached_key(|tx| {
```

teor2345 (commented):
There are just a few compile warnings to clear up, then we're good to go.

teor2345:
Also I think there was a temporary network error which made a lot of PRs fail.

zebra-rpc/src/methods.rs
arya2: Whoops, thank you!

teor2345:
This might need the dependency fix in PR #5992 to build correctly, or some other changes to the dependencies, feature dependencies, or features on the code.

arya2:
@Mergifyio update

zebra-rpc/src/methods.rs
arya2: Skipping/reverting this one because the compiler didn't like it without the feature flag:

>    --> zebra-rpc/src/methods.rs:660:35
>     |
> 660 |                 mempool::Request::FullTransactions
>     |                                   ^^^^^^^^^^^^^^^^ variant or associated item not found in `zebra_node_services::mempool::Request`

https://github.com/ZcashFoundation/zebra/actions/runs/3955201675/jobs/6773274036#step:10:377

arya2:
I missed a couple test failures:

> ---- methods::tests::snapshot::test_rpc_response_data stdout ----
> 
> Message:  condition was false for request: FullTransactions,
>  in zebra_test::mock_service::MockService<zebra_node_services::mempool::Request, zebra_node_services::mempool::Response, zebra_test::mock_service::PanicAssertion, alloc::boxed::Box<dyn core::error::Error + core::marker::Send + core::marker::Sync>>

https://github.com/ZcashFoundation/zebra/actions/runs/3961606576/jobs/6787274003#step:14:2559

> ---- methods::tests::prop::mempool_transactions_are_sent_to_caller stdout ----
> proptest: Saving this and future failures in /home/runner/work/zebra/zebra/zebra-rpc/proptest-regressions/methods/tests/prop.txt
> proptest: If this test was run on a CI system, you may wish to add the following line to your copy of the file. (You may need to create it.)
> cc 92c0a529ff6f84a632eb97c92cc7e43a740c6c5bb5ad2338a28cd345cd73bb7b
>
> Message:  Test failed: assertion failed: `(left == right)` 
>   left: `FullTransactions`, 
>  right: `TransactionIds`: received an unexpected request
>  in zebra_test::mock_service::MockService<zebra_node_services::mempool::Request, zebra_node_services::mempool::Response, zebra_test::mock_service::PropTestAssertion, alloc::boxed::Box<dyn core::error::Error + core::marker::Send + core::marker::Sync>> at zebra-test/src/mock_service.rs:515; minimal failing input: transaction_ids = {}

https://github.com/ZcashFoundation/zebra/actions/runs/3961606576/jobs/6787274003#step:14:2473

zebra-rpc/src/methods/tests/prop.rs
```diff
@@ -334,16 +334,51 @@ proptest! {
...
-                .map(|id| id.mined_id().encode_hex())
-                .collect();
-            expected_response.sort();
 
-            mempool
-                .expect_request(mempool::Request::TransactionIds)
-                .await?
-                .respond(mempool::Response::TransactionIds(transaction_ids));
+
+            #[cfg(not(feature = "getblocktemplate-rpcs"))]
```
teor2345, line 339: Looks like there are some unused imports and incorrect variable names in this file?

I think the imports need to be inside ` #[cfg(not(feature = "getblocktemplate-rpcs"))]`.

arya2: > I think the imports need to be inside `#[cfg(not(feature = "getblocktemplate-rpcs"))]`.

That's strange, it's used here too: https://github.com/ZcashFoundation/zebra/pull/5994/files#diff-e5495906830cf674b84fb812a0cc57fa47692019f4aa7becd00c033a067ada9aR377

Added the missing `transaction_ids` in [`6114a96`](https://github.com/ZcashFoundation/zebra/pull/5994/commits/6114a969005a49a87c4dd414cd1a9e4e7a11d52f).

teor2345 (approved):
Thanks for fixing this up, let's go!

PR #6006, opened 2023-01-19
fix(rpc): Avoid selecting duplicate transactions in block templates

### LINKED ISSUE #5982: Zebra's block template can contain duplicate transactions
## Motivation

When I use `zebra-utils/zcash-rpc-block-template-to-proposal` from PR #5944 to send a block proposal from Zebra to both `zcashd` and Zebra, I see these errors in Zebra's block template:
```
Node said proposal was invalid, error response from /run/user/1000/tmp.Kr4uDEHYbz.block-template-proposal/proposal-check-getblocktemplate-proposal.CurTime.8232.json:
bad-txns-inputs-missingorspent
Node said proposal was invalid, error response from /run/user/1000/tmp.Kr4uDEHYbz.block-template-proposal/proposal-check-getblocktemplate-proposal.CurTime.28232.json:
{
  "reject_reason": "rejected",
  "capabilities": [
    "proposal"
  ]
}
```

The template in template-getblocktemplate-template.json contains two copies of transaction `f7b68748b40a49bf7cb0ce8b9565a29bfd65cd917c68d0ee18feaf2614055608`. This is incorrect and breaks a consensus rule.

Both Zebra and `zcashd` agree that this block proposal is invalid.

This bug is a blocker for mining pools using Zebra, because it produces invalid blocks.

### Specifications

Duplicate transactions are invalid because of the consensus rules that prevent:
- double-spending a transparent input
- revealing a shielded spend nullifier that has already been revealed

It is difficult to reproduce this bug, because it depends on the mempool's state at the time, and the chain tips of both nodes.

But the bug isn't happening due to a chain fork or a concurrent block commit, that would be a previous block hash error.

### Testing

We might need to re-test with the same script a few times to reproduce this error, and make sure it is gone.

### Tasks

- [ ] Fix the consensus bug in the block proposal or mempool code

arya2:
## Motivation

The block template should not contain duplicate transactions.

Closes #5982.

## Solution

- Logs hashes of mempool transactions returned by `FullTransactions` request.
- Logs selected mempool transaction hashes and unpaid actions.
- Removes candidate transactions from candidate transaction collections during selection.

## Review

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -568,6 +568,14 @@ where
             let next_block_height =
                 (chain_tip_and_local_time.tip_height + 1).expect("tip is far below Height::MAX");
 
+            tracing::info!(
+                mempool_tx_hashes = ?mempool_txs
+                    .iter()
+                    .map(|tx| tx.transaction.id.mined_id())
+                    .collect::<Vec<_>>(),
+                "Selecting transactions for the template from the mempool"
+            );
+
             // Randomly select some mempool transactions.
             //
             // TODO: sort these transactions to match zcashd's order, to make testing easier.
@@ -579,6 +587,14 @@ where
                 COINBASE_LIKE_ZCASHD,
             )
             .await;
+
+            tracing::info!(
+                mempool_tx_hashes = ?mempool_txs
+                    .iter()
+                    .map(|tx| tx.transaction.id.mined_id())
+                    .collect::<Vec<_>>(),
+                "Selected transactions for the template from the mempool"
+            );
 
             // - After this point, the template only depends on the previously fetched data.
 
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -568,12 +568,12 @@ where
             let next_block_height =
                 (chain_tip_and_local_time.tip_height + 1).expect("tip is far below Height::MAX");
 
-            tracing::info!(
+            tracing::debug!(
                 mempool_tx_hashes = ?mempool_txs
                     .iter()
                     .map(|tx| tx.transaction.id.mined_id())
                     .collect::<Vec<_>>(),
-                "Selecting transactions for the template from the mempool"
+                "selecting transactions for the template from the mempool"
             );
 
             // Randomly select some mempool transactions.
@@ -588,12 +588,12 @@ where
             )
             .await;
 
-            tracing::info!(
-                mempool_tx_hashes = ?mempool_txs
+            tracing::debug!(
+                selected_mempool_tx_hashes = ?mempool_txs
                     .iter()
                     .map(|tx| tx.transaction.id.mined_id())
                     .collect::<Vec<_>>(),
-                "Selected transactions for the template from the mempool"
+                "selected transactions for the template from the mempool"
             );
 
             // - After this point, the template only depends on the previously fetched data.
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
@@ -209,12 +209,11 @@ impl GetBlockTemplate {
         if like_zcashd {
             // Sort in serialized data order, excluding the length byte.
             // `zcashd` sometimes seems to do this, but other times the order is arbitrary.
-            mempool_txs_with_templates
-                .sort_by_cached_key(|(tx_template, _tx)| tx_template.data.clone());
+            mempool_txs_with_templates.sort_by_key(|(tx_template, _tx)| tx_template.data.clone());
         } else {
             // Sort by hash, this is faster.
             mempool_txs_with_templates
-                .sort_by_cached_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
+                .sort_by_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
         }
 
         let (mempool_tx_templates, mempool_txs): (Vec<_>, Vec<_>) =
@@ -245,12 +244,12 @@ impl GetBlockTemplate {
             .map(ToString::to_string)
             .collect();
 
-        tracing::info!(
+        tracing::debug!(
             selected_txs = ?mempool_txs
                 .iter()
                 .map(|tx| (tx.transaction.id.mined_id(), tx.unpaid_actions))
                 .collect::<Vec<_>>(),
-            "Creating template ... "
+            "creating template ... "
         );
 
         GetBlockTemplate {
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
@@ -209,11 +209,12 @@ impl GetBlockTemplate {
         if like_zcashd {
             // Sort in serialized data order, excluding the length byte.
             // `zcashd` sometimes seems to do this, but other times the order is arbitrary.
-            mempool_txs_with_templates.sort_by_key(|(tx_template, _tx)| tx_template.data.clone());
+            mempool_txs_with_templates
+                .sort_by_cached_key(|(tx_template, _tx)| tx_template.data.clone());
         } else {
             // Sort by hash, this is faster.
             mempool_txs_with_templates
-                .sort_by_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
+                .sort_by_cached_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
         }
 
         let (mempool_tx_templates, mempool_txs): (Vec<_>, Vec<_>) =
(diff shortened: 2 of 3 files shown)


zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -209,11 +209,12 @@ impl GetBlockTemplate {
...
             // Sort in serialized data order, excluding the length byte.
             // `zcashd` sometimes seems to do this, but other times the order is arbitrary.
-            mempool_txs_with_templates.sort_by_key(|(tx_template, _tx)| tx_template.data.clone());
+            mempool_txs_with_templates
+                .sort_by_cached_key(|(tx_template, _tx)| tx_template.data.clone());
         } else {
             // Sort by hash, this is faster.
             mempool_txs_with_templates
-                .sort_by_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
+                .sort_by_cached_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
```
teor2345: Nitpick: Are you sure that this change improves performance?

> For simple key functions (e.g., functions that are property accesses or basic operations), [sort_by_key](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_key) is likely to be faster.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by_cached_key

`sort_by_cached_key` calls more complex code, and it uses extra memory, so it's only worth doing if we think it will make a difference.

It's unlikely to be faster for the reversed hash, but it might be faster for large data. However, `sort_by_cached_key` also makes it impossible for the compiler to eliminate the clone, because it stores the cloned value.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -201,20 +201,13 @@ fn checked_add_transaction_weighted_random(
...
-    if weighted_index
-        .update_weights(&[(candidate_position, &0.0)])
-        .is_err()
-    {
-        // All weights are zero, so each transaction has either been selected or rejected
-        (None, candidate_tx)
-    } else {
-        (Some(weighted_index), candidate_tx)
-    }
+    // All weights are zero, so each transaction has either been selected or rejected
```
teor2345: These remaining comments are both incorrect now.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -568,6 +568,14 @@ where
             let next_block_height =
                 (chain_tip_and_local_time.tip_height + 1).expect("tip is far below Height::MAX");
 
+            tracing::info!(
```
teor2345: Can make all these logs debug or trace level?

They could be too verbose for users if there are a lot of transactions in the mempool or block.

Nitpick: Rust style does not start logs or error messages with a capital letter. I think that's because they are often added to an existing formatted line or sentence.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -201,20 +201,13 @@ fn checked_add_transaction_weighted_random(
...
+    (setup_fee_weighted_index(&candidate_txs), candidate_tx)
```
teor2345: Let's explain why we have to regenerate the index every time, so we don't try to optimise this code in future, and re-introduce the bug.

```suggestion
    // We have to regenerate this index each time we choose a transaction, due to floating-point sum inaccuracies.
    // If we don't, some chosen transactions can end up with a tiny non-zero weight, leading to duplicates.
    // <https://github.com/rust-random/rand/blob/4bde8a0adb517ec956fcec91665922f6360f974b/src/distributions/weighted_index.rs#L173-L183>
    (setup_fee_weighted_index(&candidate_txs), candidate_tx)
```

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -201,20 +201,13 @@ fn checked_add_transaction_weighted_random(
...
+    candidate_txs: &mut Vec<VerifiedUnminedTx>,
+    weighted_index: WeightedIndex<f32>,
 ) -> (Option<WeightedIndex<f32>>, VerifiedUnminedTx) {
     let candidate_position = weighted_index.sample(&mut thread_rng());
-    let candidate_tx = candidate_txs[candidate_position].clone();
+    let candidate_tx = candidate_txs.swap_remove(candidate_position);
 
     // Only pick each transaction once, by setting picked transaction weights to zero
-    if weighted_index
-        .update_weights(&[(candidate_position, &0.0)])
```
teor2345, line 212: Optional: Once we're sure this works, did you want to submit a PR to update the `update_weights` documentation?
https://github.com/rust-random/rand/blob/master/src/distributions/weighted_index.rs#L144

teor2345 (commented):
Looks good, just a few tweaks for usability and maintenance.

How do you know this fix works?
Can you find a second reviewer for this PR?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: I'm not sure, I don't think it'll make a big difference either way so I'll revert this for now and might write a quick benchmark to check later.

Reverted in [`b1cec4f`](https://github.com/ZcashFoundation/zebra/pull/6006/commits/b1cec4fb44d64521b16adca1532ead59a9104d51).

arya2:
> Looks good, just a few tweaks for usability and maintenance.
> 
> How do you know this fix works? 

It failed about 10 in ~30 tests before the change and 0 in ~20 tests after the change.

I checked the logic for the mempool and it shouldn't be returning any duplicate transactions, I also checked the hashes returned by the `FullTransactions` request several times to be sure that the duplicates were from the selection logic. 

The new selection logic doesn't clone candidate transactions anymore so I think that guarantees against duplication there. 

I tried to be thorough in confirming that the new logic works, I think `update_weights` wasn't behaving as expected due to subtracting floating point values from the total weight.

> Can you find a second reviewer for this PR?

Good idea, I added @oxarbitrage as a reviewer.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
arya2: Yeah that sounds good, I'll open a PR to ideally fix the bug (I think this is just undesirable behaviour) or at least update the documentation.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
arya2: Removed in [`b1cec4f`](https://github.com/ZcashFoundation/zebra/pull/6006/commits/b1cec4fb44d64521b16adca1532ead59a9104d51).

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: > Can make all these logs debug or trace level?

Yep, I agree these are too verbose for info-level logs, fixed in [`b1cec4f`](https://github.com/ZcashFoundation/zebra/pull/6006/commits/b1cec4fb44d64521b16adca1532ead59a9104d51).

> Nitpick: Rust style does not start logs or error messages with a capital letter. I think that's because they are often added to an existing formatted line or sentence.

Fixed in [`b1cec4f`](https://github.com/ZcashFoundation/zebra/pull/6006/commits/b1cec4fb44d64521b16adca1532ead59a9104d51).

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
arya2: Good idea, thank you. Applied in [`b1cec4f`](https://github.com/ZcashFoundation/zebra/pull/6006/commits/b1cec4fb44d64521b16adca1532ead59a9104d51).

daira: Btw I've never liked that zcashd uses floating point here, and that is something we might change. Let us know if you want to change it to integer or fixed point arithmetic.

oxarbitrage (approved):
I think it looks good, i made a few manual calls and got no duplicates. 

I was thinking on making a script that just runs `getblocktemplate` and check there are no duplicate transactions however i am not really sure if it worth it. 

If we think it does, we can think on the test for it in a separated issue with lower priority.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
teor2345: I don't think it's possible to fix the bug, because doing an efficient weights update requires re-using partial sums. That works fine for integers, but floating-point sums can't be re-ordered.

teor2345:
> > Looks good, just a few tweaks for usability and maintenance.
> > How do you know this fix works?
> 
> It failed about 10 in ~30 tests before the change and 0 in ~20 tests after the change.
> 
> I checked the logic for the mempool and it shouldn't be returning any duplicate transactions, I also checked the hashes returned by the `FullTransactions` request several times to be sure that the duplicates were from the selection logic.
> 
> The new selection logic doesn't clone candidate transactions anymore so I think that guarantees against duplication there.
> 
> I tried to be thorough in confirming that the new logic works, I think `update_weights` wasn't behaving as expected due to subtracting floating point values from the total weight.

We already have a CI test PR that will catch this bug, so I don't think we need any other tests.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
teor2345: We could use fixed point with u64, and scale the fee by 2^13 before dividing. That would get us close to the maximum number of actions. (Which I think is roughly 300K for transparent outputs.)

But we don't need to make that change any time soon, we're focused on finishing off all the RPCs that miners need.

PR #6013, opened 2023-01-20
use `reddsa` crate and remove duplicated RedPallas code

### LINKED ISSUE #40: Migrate/port `redpallas` to frost-core / FROST spec v05+

### LINKED ISSUE #2080: replace`redpallas` module with `reddsa` crate
## Motivation

There is a stub of `redpallas` in `zebra-chain::primitives` for now. There is a proposal to make `redjubjub` generic over traits to make it `reddsa`, that str4d made work over in the `orchard` crate. 

str4d published a minimal `reddsa` crate in #3070.

## Tasks

- Use the ECC's `redpallas` crate.
- Replace our stub in `zebra-chain` with `redpallas`.
- Deduplicate `redjubjub` and `redpallas`

### Scheduling

This is only required for the wallet work.

We might want to schedule this work around splitting FROST out of `redjubjub`.

conradoplg:
## Motivation

We have duplicated code in `zebra-chain/src/primitives/redpallas` not that it has been made generic and moved to the `reddsa` crate.

To avoid duplicated efforts, and missing porting improvements from one to another, we should remove the duplicated code and use `reddsa` instead.

Close #2080

Two consensus rules were being documented in the deleted code. However, they were also being documented elsewhere. Namely these ones:

https://github.com/ZcashFoundation/zebra/blob/81846d81e09e06252000660445919f369c8f27fb/zebra-chain/src/primitives/redpallas/batch.rs#L144-L151

https://github.com/ZcashFoundation/zebra/blob/81846d81e09e06252000660445919f369c8f27fb/zebra-chain/src/primitives/redpallas/batch.rs#L253-L260

are already documented in:

https://github.com/ZcashFoundation/zebra/blob/81846d81e09e06252000660445919f369c8f27fb/zebra-chain/src/orchard/action.rs#L62-L72

https://github.com/ZcashFoundation/zebra/blob/81846d81e09e06252000660445919f369c8f27fb/zebra-chain/src/orchard/action.rs#L80-L84

## Solution

Remove duplicated code and use `reddsa` instead.

This was fairly straightforward since the API is basically the same, the non-trivial part was creating newtypes to implement Arbitrary.

Closes https://github.com/ZcashFoundation/zebra/issues/2080
Closes https://github.com/ZcashFoundation/frost/issues/40

## Review

This is not a big priority, and I don't know how this is impacted by the audit, if we should wait for it or not.

To ensure it's correct you can compare the folders `zebra-chain/src/primitives/redpallas`  to `reddsa` to see they are very similar (I used `meld`)

Some changes we made after the crate was created were ported in https://github.com/ZcashFoundation/reddsa/pull/40 so me might want that to merge first.

### Reviewer Checklist


oxarbitrage:
I think merging this will impact the audit but in a positive way, as they will only have one source for this code instead of two.

teor2345:
> To avoid duplicated efforts, and missing porting improvements from one to another, we should remove the duplicated code and use `reddsa` instead.

> I think merging this will impact the audit but in a positive way, as they will only have one source for this code instead of two.

I agree, I don't want us to be maintaining duplicated code, it costs us time and is a source of bugs.

I put the audit scope on the agenda for the gardening meeting tomorrow.

zebra-consensus/src/transaction.rs
```diff
@@ -1029,11 +1031,13 @@ where
             // It calls [`pallas::Affine::from_bytes`] to parse R and
```
teor2345, line 1031: Nitpick: The comment above this line still references `redpallas`.

Can you search for any remaining references and update them?

zebra-chain/src/orchard/shielded_data.rs
```diff
@@ -8,15 +8,13 @@ use std::{
 
 use byteorder::{ReadBytesExt, WriteBytesExt};
 use halo2::pasta::pallas;
+use reddsa::{self, orchard::Binding, orchard::SpendAuth, Signature};
```
teor2345, line 11: We're currently doing manual tests that print these types, so I'd like to wait for the hex debug changes in `reddsa` to merge and be released:
https://github.com/ZcashFoundation/reddsa/pull/40

I guess that would be version 0.3.1 or 0.4.0:
https://github.com/ZcashFoundation/reddsa/tags

teor2345 (commented):
Looks good, I think we just need a few comment tweaks.

Thank you for this, I think it will simplify the code a lot!

zebra-consensus/src/transaction.rs
conradoplg: Good catch, fixed it and others in 5db57917c

teor2345:
```
#22 55.62 warning: spurious network error (1 tries remaining): failed to get successful HTTP response from `[https://crates.io/api/v1/crates/cfg-if/1.0.0/download`](https://crates.io/api/v1/crates/cfg-if/1.0.0/download%60), got 502
#22 55.62 body:
#22 55.62 <html>
#22 55.62 <head><title>502 Bad Gateway</title></head>
```
https://github.com/ZcashFoundation/zebra/actions/runs/3991429463/jobs/6846215229#step:9:1181

teor2345:
@mergifyio update

dconnolly:
https://github.com/ZcashFoundation/reddsa/pull/40 has merged

zebra-chain/benches/redpallas.rs
```diff
@@ -5,8 +5,11 @@
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 use rand::{thread_rng, Rng};
-
-use zebra_chain::primitives::redpallas::*;
+use reddsa::{
```
dconnolly, line 8: We should move this benchmark out of `zebra-chain` into `reddsa` now, the existing bench in `reddsa` is just the Sapling (RedJubjub) variant

dconnolly: (This is not a blocker to this PR merging but it's a thing we should track)

zebra-chain/src/orchard/arbitrary.rs
```diff
@@ -18,14 +16,14 @@ impl Arbitrary for Action {
...
-            any::<VerificationKeyBytes<SpendAuth>>(),
+            any::<SpendAuthVerificationKeyBytes>(),
             any::<note::EncryptedNote>(),
             any::<note::WrappedNoteKey>(),
         )
             .prop_map(|(nullifier, rk, enc_ciphertext, out_ciphertext)| Self {
                 cv: ValueCommitment(pallas::Affine::identity()),
                 nullifier,
-                rk,
+                rk: rk.0,
```
dconnolly, line 26: Hm i wonder if we can make this nicer, but not a blocker

dconnolly: Meh ignore

zebra-chain/src/orchard/shielded_data.rs
dconnolly: Now merged, need to version

dconnolly:
😊 

image

zebra-chain/benches/redpallas.rs
teor2345: Can you open a Zebra or `reddsa` ticket?
Whatever is easier for the `reddsa` developers and Pili to track.

zebra-chain/src/orchard/arbitrary.rs
teor2345: It has a deref impl, so `&rk` might also work, either as an argument or in the struct initialisation.

dconnolly:
https://github.com/ZcashFoundation/reddsa/releases/tag/0.4.0 has been published

zebra-chain/Cargo.toml
```diff
@@ -87,6 +87,7 @@ rayon = "1.6.1"
 # ZF deps
 ed25519-zebra = "3.1.0"
 redjubjub = "0.5.0"
+reddsa = "0.3.0"
```
dconnolly:
```suggestion
reddsa = "0.4.0"
```

zebra-chain/src/orchard/shielded_data.rs
dconnolly: Published, updated here in https://github.com/ZcashFoundation/zebra/pull/6013/commits/fe38ce718005b72971021db53b3587044ceeb855

zebra-chain/Cargo.toml
teor2345: This also requires an update to `Cargo.lock`, CI will fail without it, because we use it to check for duplicate dependencies:
https://github.com/ZcashFoundation/zebra/actions/runs/4050491155/jobs/6967928558

Docker also expects an up to date lock file:
https://github.com/ZcashFoundation/zebra/actions/runs/4050491179/jobs/6967928463#step:9:1696

dconnolly:
oop have to regenerate Cargo.lock

mpguerra:
@upbqdn Can you please give this one a final lookover and see if it's ready to go? :) Maybe after @conradoplg resolves the current unresolved conversations though...

zebra-chain/benches/redpallas.rs
conradoplg: Created https://github.com/ZcashFoundation/reddsa/issues/42

conradoplg:
[Failed](https://github.com/ZcashFoundation/zebra/actions/runs/4065946823/jobs/7002511587) due to https://github.com/ZcashFoundation/zebra/issues/6079 with

```
Message:  block proposal validation failed: 
   0: unsuccessful block proposal validation, reason: "invalid proposal: Block {\n    source: ValidateProposal(\n        \"proposal is not based on the current best chain tip: previous block hash must be the best chain tip\",\n    ),\n}"

Location:
   zebrad/tests/common/get_block_template_rpcs/get_block_template.rs:173
```

Retrying in case it's intermittent

zebra-chain/Cargo.toml
conradoplg: done in 3ea23326e

teor2345:
I retried all the jobs in that workflow, because only retrying the failed jobs caused this error:
> ERROR: (gcloud.compute.ssh) Could not fetch resource:
[86](https://github.com/ZcashFoundation/zebra/actions/runs/4065946823/jobs/7003747883#step:8:87)
> - The resource 'projects/zealous-zebra/zones/us-central1-a/instances/get-block-template-6013-merge-0a1eeb9' was not found

This happens because the instance is launched by one of the successful jobs, so it wasn't re-launched for the retry.

@gustavovalverde this is one of the reasons we might want to split the google cloud CI workflow into short and long jobs, so we don't have to retry all the slow jobs when a quick one fails. Or we might want to put the setup, launch and run steps in the same job, and then we could just retry failed jobs.

teor2345 (approved):
I'm not sure if Marek is working right now, and we're on a bit of a short deadline with these changes if we want to make changes to the audit.

PR #6027, opened 2023-01-24
fix(consensus): Verify the lock times of mempool transactions

### LINKED ISSUE #5984: Transaction nLockTIme check (CheckFinalTx) against MTP when adding to mempool possible missed
In `zcashd` implementation when we accepting transaction in memory pool (`AcceptToMemoryPool`) we checking is this transaction final or not - https://github.com/zcash/zcash/blob/master/src/main.cpp#L1816-L1820 :+1: 
```cpp

    // Only accept nLockTime-using transactions that can be mined in the next
    // block; we don't want our mempool filled up with transactions that can't
    // be mined yet.
    if (!CheckFinalTx(tx, STANDARD_LOCKTIME_VERIFY_FLAGS))
        return state.DoS(0, false, REJECT_NONSTANDARD, "non-final");
```
Here we called `CheckFinalTx` with `flags == 2` and inside it calls [`IsFinalTx(tx, nBlockHeight, nBlockTime)`](https://github.com/zcash/zcash/blob/master/src/main.cpp#L784) with `nBlockTime == chainActive.Tip()->GetMedianTimePast()`.

But in `zebrad` this check seems omitted. We have one [verifier](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/transaction.rs#L297) for block and mempool requests, and inside we have only one locktime related [check](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/transaction.rs#L309-L312):
```rust
            // Do basic checks first
            if let Some(block_time) = req.block_time() {
                check::lock_time_has_passed(&tx, req.height(), block_time)?;
            }
```
But this check only performed on `req` is `Request::Block` with block time, which corresponds to call `isFinalTx` in `ContextualCheckBlock` of `zcashd`, but on `Request::Mempool` - seems there is no check tx `nLockTIme` against MTP is implemented in `zebrad`. I apologize profusely if I am incorrect and once more overlooked something.

teor2345:
## Motivation

We need to check that mempool transaction lock times are valid, so the block templates created by Zebra are also valid.

Closes #5984

#### Mempool Transaction Verification

> Define the median-time-past of a block to be the median (as defined in § 7.7.3 ‘Difficulty adjustment’ on p. 131) of the nTime fields of the preceding PoWMedianBlockSpan blocks (or all preceding blocks if there are fewer than PoWMedianBlockSpan). The median-time-past of a genesis block is not defined.

> • As in Bitcoin, the nTime field MUST represent a time strictly greater than the median of the timestamps of the past PoWMedianBlockSpan blocks.

> If greater than or equal to 500 million, locktime is parsed using the [Unix epoch time](https://en.wikipedia.org/wiki/Unix_time) format (the number of seconds elapsed since 1970-01-01T00:00 UTC—currently over 1.395 billion). The transaction can be added to any block whose block time is greater than the locktime.

https://developer.bitcoin.org/devguide/transactions.html#locktime-and-sequence-number

If the transaction's lock time is less than the median-time-past, it will always be less than the next block's time, because the next block's time is strictly greater than the median-time-past.

This is the rule implemented by `zcashd`:
https://github.com/zcash/zcash/blob/9e1efad2d13dca5ee094a38e6aa25b0f2464da94/src/main.cpp#L776-L784
https://github.com/zcash/zcash/blob/9e1efad2d13dca5ee094a38e6aa25b0f2464da94/src/main.cpp#L735

#### Mempool Updates on State Tip Changes

When a new block is added to the state tip:
- existing transactions in the mempool will always have valid lock times, because the block time is strictly increasing
- previously rejected lock times might have become valid, but the mempool already clears all verifier errors on every new block

This PR modifies async code, and state code that runs concurrently with block writes.

## Solution

- Implement the BestChainNextMedianTimePast state request
- Verify the lock times of mempool transactions
- Document that the mempool already handles lock time rejections correctly

This is a mempool consensus rule bug, so we don't want these changes behind a feature flag.

### Testing

- Fix existing tests to have valid lock times
- Add new tests for each lock time consensus rule: unlocked, sequence numbers, past lock time, before lock time

## Review

I marked this as a high priority because it is blocking most of our testing work.
Because it is a high priority, I'd like to avoid doing refactors in this PR.

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -23,6 +23,7 @@ use zebra_chain::{
     parameters::{Network, NetworkUpgrade},
     primitives::Groth16Proof,
     sapling,
+    serialization::DateTime32,
     transaction::{
         self, HashType, SigHash, Transaction, UnminedTx, UnminedTxId, VerifiedUnminedTx,
     },
@@ -306,17 +307,15 @@ where
         async move {
             tracing::trace!(?tx_id, ?req, "got tx verify request");
 
-            // Do basic checks first
-            if let Some(block_time) = req.block_time() {
-                check::lock_time_has_passed(&tx, req.height(), block_time)?;
-            }
-
+            // Do quick checks first
             check::has_inputs_and_outputs(&tx)?;
             check::has_enough_orchard_flags(&tx)?;
 
+            // Validate the coinbase input consensus rules
             if req.is_mempool() && tx.is_coinbase() {
                 return Err(TransactionError::CoinbaseInMempool);
             }
+
             if tx.is_coinbase() {
                 check::coinbase_tx_no_prevout_joinsplit_spend(&tx)?;
             } else if !tx.is_valid_non_coinbase() {
@@ -345,6 +344,34 @@ where
 
             tracing::trace!(?tx_id, "passed quick checks");
 
+            if let Some(block_time) = req.block_time() {
+                check::lock_time_has_passed(&tx, req.height(), block_time)?;
+            } else {
+                // This state query is much faster than loading UTXOs from the database,
+                // so it doesn't need to be executed in parallel
+                let state = state.clone();
+                let next_median_time_past = Self::mempool_best_chain_next_median_time_past(state).await?;
+
+                // # Consensus
+                //
+                // > the nTime field MUST represent a time strictly greater than the median of the
+                // > timestamps of the past PoWMedianBlockSpan blocks.
+                // <https://zips.z.cash/protocol/protocol.pdf#blockheader>
+                //
+                // > The transaction can be added to any block whose block time is greater than the locktime.
+                // <https://developer.bitcoin.org/devguide/transactions.html#locktime-and-sequence-number>
+                //
+                // If the transaction's lock time is less than the median-time-past,
+                // it will always be less than the next block's time,
+                // because the next block's time is strictly greater than the median-time-past.
+                //
+                // This is the rule implemented by `zcashd`'s mempool:
+                // <https://github.com/zcash/zcash/blob/9e1efad2d13dca5ee094a38e6aa25b0f2464da94/src/main.cpp#L776-L784>
+                //
+                // This consensus check makes sure Zebra produces valid block templates.
+                check::lock_time_has_passed(&tx, req.height(), next_median_time_past.to_chrono())?;
+            }
+
             // "The consensus rules applied to valueBalance, vShieldedOutput, and bindingSig
             // in non-coinbase transactions MUST also be applied to coinbase transactions."
             //
@@ -356,7 +383,7 @@ where
             // https://zips.z.cash/zip-0213#specification
 
             // Load spent UTXOs from state.
-            // TODO: Make this a method of `Request` and replace `tx.clone()` with `self.transaction()`?
+            // The UTXOs are required for almost all the async checks.
             let load_spent_utxos_fut =
                 Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state.clone());
             let (spent_utxos, spent_outputs) = load_spent_utxos_fut.await?;
@@ -426,7 +453,7 @@ where
             // Calculate the fee only for non-coinbase transactions.
             let mut miner_fee = None;
             if !tx.is_coinbase() {
-                // TODO: deduplicate this code with remaining_transaction_value (#TODO: open ticket)
+                // TODO: deduplicate this code with remaining_transaction_value()?
                 miner_fee = match value_balance {
                     Ok(vb) => match vb.remaining_transaction_value() {
                         Ok(tx_rtv) => Some(tx_rtv),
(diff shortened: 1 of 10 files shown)


teor2345:
This PR needs 2 reviewers, so I've tagged everyone else who is working on Zebra for a review.

If you're the second reviewer, you can take the third person off!

zebra-state/src/service/read/find.rs
```diff
@@ -526,3 +535,102 @@ where
...
+    })?;
+
+    let state_tip_after_queries =
+        read::best_tip(non_finalized_state, db).expect("already checked for an empty tip");
+
+    if state_tip_before_queries != state_tip_after_queries {
+        return Err("Zebra is committing too many blocks to the state, \
+                    wait until it syncs to the chain tip"
+            .into());
+    }
```
arya2, line 598: Could we split this out of `relevant_chain_and_history_tree` and use the same fn for getting the `relevant_chain` in both places.

zebra-state/src/service/read/find.rs
```diff
@@ -526,3 +535,102 @@ where
...
+/// See [`next_median_time_past()`] for details.
+fn calculate_median_time_past(
+    relevant_chain: [Arc<Block>; POW_ADJUSTMENT_BLOCK_SPAN],
+) -> DateTime32 {
+    let relevant_data: Vec<(CompactDifficulty, DateTime<Utc>)> = relevant_chain
+        .iter()
+        .map(|block| (block.header.difficulty_threshold, block.header.time))
+        .collect();
+
+    // TODO: split out median-time-past into its own struct?
```
arya2, line 616: Optional: It would be nice to get rid of the ignored fields, perhaps by:
- using the `AdjustedDifficulty::median_time` associated function to collect 11 blocks instead of 28 blocks from the relevant chain
- making `AdjustedDifficulty::median_time_past` an associated function

I don't think it's worth splitting out median-time-past into its own struct but that would work too.

zebra-state/src/service/read/find.rs
```diff
@@ -526,3 +535,102 @@ where
...
+        relevant_data,
+    )
+    .median_time_past();
+
+    // > Define the median-time-past of a block to be the median of the nTime fields of the
+    // > preceding PoWMedianBlockSpan blocks (or all preceding blocks if there are fewer than
+    // > PoWMedianBlockSpan). The median-time-past of a genesis block is not defined.
+    // https://zips.z.cash/protocol/protocol.pdf#blockheader
+    DateTime32::try_from(median_time_past).expect("valid blocks have in-range times")
+}
```
arya2, line 636: Optional: use this function in `difficulty_time_and_history_tree` too. (either taking `time`, `height`, and `network` as arguments or getting the `median_time_past` without them)

zebra-consensus/src/transaction.rs
```diff
@@ -345,6 +344,34 @@ where
 
             tracing::trace!(?tx_id, "passed quick checks");
 
+            if let Some(block_time) = req.block_time() {
+                check::lock_time_has_passed(&tx, req.height(), block_time)?;
+            } else {
+                // This state query is much faster than loading UTXOs from the database,
+                // so it doesn't need to be executed in parallel
+                let state = state.clone();
+                let next_median_time_past = Self::mempool_best_chain_next_median_time_past(state).await?;
```
arya2, line 353: Optional: skip this request if the `lock_time` is a height.

zebra-consensus/src/transaction/tests.rs
```diff
@@ -261,6 +261,305 @@ async fn mempool_request_with_present_input_is_accepted() {
...
+            .expect("verifier should call mock state service with correct request")
+            .respond(zebra_state::Response::ValidBestChainTipNullifiersAndAnchors);
+    });
+
+    let verifier_response = verifier
+        .oneshot(Request::Mempool {
+            transaction: tx.into(),
+            height,
+        })
+        .await;
```
arya2, line 371: Optional/Nitpick: extract the common code from the new tests into a function.

arya2 (commented):
This looks good to me, it should work as-is but I left a few optional comments for clean up.

zebra-state/src/service/read/find.rs
teor2345: That's pretty tricky, because we don't need the history tree when we're just getting the median-time-past.

So I think the simplest way to do this is to refactor all the retry functions to use a generic "retry" closure. I've left TODOs, but I don't want to do that refactor in this PR. And it will change a bunch of existing production code, so we might want to wait until after the audit.

zebra-state/src/service/read/find.rs
teor2345: Thanks, I'll do that in PR #6031, so we're not blocking a high-priority fix and all the other work that depends on it.

zebra-state/src/service/read/find.rs
teor2345: This change is already in PR #6031.

zebra-consensus/src/transaction.rs
teor2345: I'll make this change in PR #6031.

zebra-consensus/src/transaction/tests.rs
teor2345: I don't think this is worth the effort it would take, but feel free to push a PR after I'm done with PR #6031.

arya2 (approved):
Looks good, thank you for the fix!

I left a `do-not-merge` label in case anyone else is currently reviewing this, feel free to remove it when you want this to embark on the merge train.

teor2345:
I would like two reviews, because this is consensus-critical concurrent code.

oxarbitrage (approved):
looks good to me.

zebra-state/src/service/read/find.rs
teor2345: Fixed in 6a748252d4 in PR #6031

zebra-consensus/src/transaction.rs
teor2345: Fixed in 8050557681 in PR #6031

PR #6031, opened 2023-01-24
refactor(state): Simplify difficulty and median-time-past state and mempool requests

teor2345:
## Motivation

This PR contains refactors and cleanups after PR #6027.

It shouldn't change any behaviour, or block any other work.

Depends-On: #6027

This PR changes a consensus rule implementation by replacing duplicate code.
It also changes another consensus rule implementation by making an argument optional if it is unused.

## Solution

Refactors:
- Only get the state tip median-time-past when needed in the transaction verifier
- Use the new `calculate_median_time_past()` function in the existing difficulty code, to remove duplicate code
- Simplify `calculate_median_time_past()` by removing dummy arguments and extra blocks
- Add a ReadState best chain clone method (currently unused)

Documentation and logging:
- Clarify difficulty state request docs, rename variables
- Fix state request log typos

## Review

This is a low priority.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/transaction.rs
+++ b/zebra-chain/src/transaction.rs
@@ -363,6 +363,17 @@ impl Transaction {
         }
     }
 
+    /// Returns `true` if this transaction's lock time is a time.
+    /// Returns `false` if it is a height, is unlocked,
+    /// or if the transparent input sequence numbers have disabled lock times.
+    pub fn lock_time_is_time(&self) -> bool {
+        if let Some(lock_time) = self.lock_time() {
+            return lock_time.is_time();
+        }
+
+        false
+    }
+
     /// Get this transaction's expiry height, if any.
     pub fn expiry_height(&self) -> Option<block::Height> {
         match self {
--- a/zebra-chain/src/transaction.rs
+++ b/zebra-chain/src/transaction.rs
@@ -363,8 +363,8 @@ impl Transaction {
         }
     }
 
-    /// Returns `true` if this transaction's lock time is a time.
-    /// Returns `false` if it is a height, is unlocked,
+    /// Returns `true` if this transaction's `lock_time` is a [`LockTime::Time`].
+    /// Returns `false` if it is a [`LockTime::Height`] (locked or unlocked), is unlocked,
     /// or if the transparent input sequence numbers have disabled lock times.
     pub fn lock_time_is_time(&self) -> bool {
         if let Some(lock_time) = self.lock_time() {
--- a/zebra-chain/src/transaction/lock_time.rs
+++ b/zebra-chain/src/transaction/lock_time.rs
@@ -93,6 +93,11 @@ impl LockTime {
                 .expect("in-range number of seconds and valid nanosecond"),
         )
     }
+
+    /// Returns `true` if this lock time is a time, or `false` if it is a height.
+    pub fn is_time(&self) -> bool {
+        matches!(self, LockTime::Time(_))
+    }
 }
 
 impl ZcashSerialize for LockTime {
--- a/zebra-chain/src/transaction/lock_time.rs
+++ b/zebra-chain/src/transaction/lock_time.rs
@@ -94,7 +94,7 @@ impl LockTime {
         )
     }
 
-    /// Returns `true` if this lock time is a time, or `false` if it is a height.
+    /// Returns `true` if this lock time is a [`LockTime::Time`], or `false` if it is a [`LockTime::Height`].
     pub fn is_time(&self) -> bool {
         matches!(self, LockTime::Time(_))
     }
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -347,29 +347,18 @@ where
             if let Some(block_time) = req.block_time() {
                 check::lock_time_has_passed(&tx, req.height(), block_time)?;
             } else {
-                // This state query is much faster than loading UTXOs from the database,
-                // so it doesn't need to be executed in parallel
-                let state = state.clone();
-                let next_median_time_past = Self::mempool_best_chain_next_median_time_past(state).await?;
+                let mut next_median_time_past = None;
 
-                // # Consensus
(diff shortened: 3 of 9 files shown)


teor2345:
I want to make some of the changes from a previous PR review in this PR:
https://github.com/ZcashFoundation/zebra/pull/6027#pullrequestreview-1270228944

zebra-chain/src/transaction.rs
```diff
@@ -363,6 +363,17 @@ impl Transaction {
         }
     }
 
+    /// Returns `true` if this transaction's lock time is a time.
+    /// Returns `false` if it is a height, is unlocked,
```
arya2: Nitpick:

```suggestion
    /// Returns `true` if this transaction's `lock_time` is a [`LockTime::Time`].
    /// Returns `false` if it is a [`LockTime::Height`] (locked or unlocked), is unlocked,
```

zebra-chain/src/transaction/lock_time.rs
```diff
@@ -93,6 +93,11 @@ impl LockTime {
                 .expect("in-range number of seconds and valid nanosecond"),
         )
     }
+
+    /// Returns `true` if this lock time is a time, or `false` if it is a height.
```
arya2: Nitpick:

```suggestion
    /// Returns `true` if this lock time is a [`LockTime::Time`], or `false` if it is a [`LockTime::Height`].
```

zebra-consensus/src/transaction.rs
```diff
@@ -347,29 +347,18 @@ where
...
                 check::lock_time_has_passed(&tx, req.height(), block_time)?;
             } else {
-                // This state query is much faster than loading UTXOs from the database,
-                // so it doesn't need to be executed in parallel
-                let state = state.clone();
-                let next_median_time_past = Self::mempool_best_chain_next_median_time_past(state).await?;
+                let mut next_median_time_past = None;
+
+                // Skip the state query if we don't need the time for this check.
+                if tx.lock_time_is_time() {
```
arya2: 🎉 

Nitpick/Optional/Subjective: assign `next_median_time_past` with:

```rust
// Skip the state query if we don't need the time for this check.
let next_median_time_past = if tx.lock_time_is_time() {
    // This state query is much faster than loading UTXOs from the database,
    // so it doesn't need to be executed in parallel
    let state = state.clone();
    Some(Self::mempool_best_chain_next_median_time_past(state).await?.to_chrono())
} else {
    None
};
```

instead of making it mutable.

zebra-consensus/src/transaction/check.rs
```diff
@@ -19,19 +19,47 @@ use crate::error::TransactionError;
...
+///
+/// > The transaction can be added to any block whose block time is greater than the locktime.
+/// <https://developer.bitcoin.org/devguide/transactions.html#locktime-and-sequence-number>
+///
+/// If the transaction's lock time is less than the median-time-past,
+/// it will always be less than the next block's time,
+/// because the next block's time is strictly greater than the median-time-past.
+/// (That is, `lock-time < median-time-past < block-header-time`.)
+///
+/// Using `median-time-past + 1` would also satisfy this consensus rule,
```
arya2: Nitpick:

```suggestion
/// Using `median-time-past + 1s` (the next block's mintime) would also satisfy this consensus rule,
```

(Just curious, why do we prefer this?)

zebra-state/src/service.rs
```diff
@@ -803,6 +803,15 @@ impl ReadStateService {
...
         self.non_finalized_state_receiver.cloned_watch_data()
     }
+
+    /// Gets a clone of the latest, best non-finalized chain from the `non_finalized_state_receiver`
+    #[allow(dead_code)]
+    fn latest_best_chain(&self) -> Option<Arc<Chain>> {
+        self.non_finalized_state_receiver
+            .cloned_watch_data()
+            .best_chain()
+            .cloned()
```
arya2: Nitpick:

```suggestion
        self.latest_non_finalized_state().best_chain().cloned()
```

zebra-state/src/service.rs
```diff
@@ -1152,7 +1161,7 @@ impl Service<ReadRequest> for ReadStateService {
                         Ok(ReadResponse::Depth(depth))
                     })
                 })
-                .map(|join_result| join_result.expect("panic in ReadRequest::Tip"))
+                .map(|join_result| join_result.expect("panic in ReadRequest::Depth"))
```
arya2, line 1161: Nice catch!

We should refactor the repetitive the timer/spawn_blocking/span/map into a function later.

arya2 (dismissed):
This is excellent, thank you for the cleanup.

All of my comments are optional.

zebra-chain/src/transaction.rs
teor2345: Thanks, I tweaked it a bit to match what the function does.

zebra-consensus/src/transaction.rs
teor2345: I don't mind either way, I'm happy if you want to push a commit to change it.

zebra-consensus/src/transaction/check.rs
teor2345: We prefer this rule because it matches `zcashd`, but I'd be happy to switch to the other one if we triple-checked my justification. (Or maybe asked the `zcashd` engineers.)

Or are you talking about `+ 1` vs `+ 1s`?

arya2: > We prefer this rule because it matches `zcashd`

Ah, okay.

> Or are you talking about `+ 1` vs `+ 1s`?

I meant the zcashd implementation of that rule in reference to the comment on the line right below.

`+ 1s` was just me taking a moment to wonder about the units.

arya2 (approved):
Thank you!

@mpguerra I was the last pusher (of a trivial change), could you please also approve this?

mpguerra (approved):

PR #6038, opened 2023-01-27
fix(docker): do not mix cache images with test images

### LINKED ISSUE #5933: GCP: Do not allow digests without tags in Artifact Registry
## Description
Cache images get their tag removed when updated, leaving "orphaned" images but with dependency on other digests 

image

### Expected behavior

When tagging the cache image use a specific tag so it doesn't get replaced by new ones and we don't get orphaned images which can't be automatically deleted by the enhancement in this PR https://github.com/ZcashFoundation/zebra/pull/5873

gustavovalverde:
## Motivation

Docker does not allow to use multiple tags for cache images in their `cache-from` and `cache-to` options. Which makes some images to lose previous tags, if the same.

This cause automated mechanims for deletion to fail, as those cache images were dependant on the main image, and deleting the cache alone was raising an error if the main image was not deleted first.

Fixes #5933 

- This required to create a new repository in GCP to hold all cache images: https://console.cloud.google.com/artifacts/docker/zealous-zebra/us/zebra-caching

## Solution

A workaround for this was making a separate repository which can hold this cache images. Which also reduces clutter from the main registries.

## Review

Anyone can review this, mainly testing a re-run is using the caching images

### Reviewer Checklist

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -84,7 +84,6 @@ jobs:
             type=ref,event=pr
             type=sha
             # edge is the latest commit on the default branch.
-            # TODO: We only want edge builds for `us-docker.pkg.dev`, because DockerHub doesn't keep up with every `main` branch commit.
             type=edge,enable={{is_default_branch}}
 
       # Setup Docker Buildx to allow use of docker cache layers from GH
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks!

PR #6040, opened 2023-01-27
ci(lwd): run the send transactions test on each PR update

### LINKED ISSUE #5600: Run the send transactions test on each PR update
## Motivation

The send transactions test was moved to the main branch in https://github.com/ZcashFoundation/zebra/pull/5480 because it was very slow.

It's much faster (~30m) with https://github.com/ZcashFoundation/zebra/issues/5015 and can be run for every PR update again.

### Designs

Revert the changes in https://github.com/ZcashFoundation/zebra/pull/5480 and remove the `github.event.inputs.run-lwd-send-tx != 'true'` anywhere else it may have been added.

gustavovalverde:
## Motivation

The send transactions test was moved to the main branch in #5480 because it was very slow.

It's much faster (~30m) with #5015 and now it can be run for every PR update again.

Fixes #5600

## Solution

Revert the changes in https://github.com/ZcashFoundation/zebra/pull/5480 and remove the github.event.inputs.run-lwd-send-tx != 'true' anywhere else it may have been added.

## Review

Anyone can review this

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -33,12 +33,6 @@ on:
         default: false
         description: 'Just run a lightwalletd full sync and update tip disks'
         required: true
-      run-lwd-send-tx:
-        type: boolean
-        default: false
-        description: 'Just run a lightwalletd send transactions test'
-        required: true
-
 
   pull_request:
     paths:
@@ -581,7 +575,7 @@ jobs:
     name: lightwalletd tip send
     needs: lightwalletd-full-sync
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' || github.event.inputs.run-lwd-send-tx == 'true' }}
+    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' }}
     with:
       app_name: lightwalletd
       test_id: lwd-send-transactions
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -198,7 +198,7 @@ jobs:
     name: Test all
     runs-on: ubuntu-latest
     needs: build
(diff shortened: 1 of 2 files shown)


.github/workflows/continous-integration-docker.yml
```diff
@@ -581,7 +581,7 @@ jobs:
     name: lightwalletd tip send
     needs: lightwalletd-full-sync
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    if: ${{ !cancelled() && !failure() && ((github.event_name == 'push' && github.ref_name == 'main') || github.event.inputs.run-lwd-send-tx == 'true') }}
+    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' || github.event.inputs.run-lwd-send-tx == 'true' }}
```
arya2: We should update this too if we remove the manual workflow dispatch:

```suggestion
    if: ${{ !cancelled() && !failure() && github.event.inputs.regenerate-disks != 'true' && github.event.inputs.run-full-sync != 'true' && github.event.inputs.run-lwd-sync != 'true' }}
```

arya2 (commented):
Looks good!

I don't think we'll need the [manual workflow dispatch] (https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/continous-integration-docker.yml#L36) anymore if this runs on PR updates.

I'm not sure if we need the [patch job](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/continous-integration-docker.patch-always.yml#L28) either.

.github/workflows/continous-integration-docker.yml
teor2345: Good catch!

teor2345:
> I'm not sure if we need the [patch job](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/continous-integration-docker.patch-always.yml#L28) either.

We don't need the "always" patch job. We still need the conditional patch job. (It would also be helpful to remove duplicate jobs across the patch files at the same time.)

.github/workflows/continous-integration-docker.yml
gustavovalverde: Fixed by: [fix(actions): remove references to the workflow_dispatch](https://github.com/ZcashFoundation/zebra/pull/6040/commits/fbba1d159393a0e45b3bda199a8ff3b76ff07e46)

teor2345 (approved):
Thanks for the fix, I think this is good to go now.

PR #6041, opened 2023-01-27
example: zcash_address parsing example

teor2345:
## Motivation

Arya asked for this example of how to use `zcash_address` to parse addresses and extract information for the RPCs.

It's not for review, and it shouldn't be merged.

### DIFF
--- a/zebra-chain/src/primitives/address.rs
+++ b/zebra-chain/src/primitives/address.rs
@@ -1,0 +1,51 @@
+//! `zcash_address` conversion to `zebra_chain` address types.
+//!
+//! Usage: <https://docs.rs/zcash_address/0.2.0/zcash_address/trait.TryFromAddress.html#examples>
+
+use crate::{parameters::Network, transparent, BoxError};
+
+impl TryFrom<zcash_address::Network> for Network {
+    // TODO: better error type
+    type Error = BoxError;
+
+    fn try_from(network: zcash_address::Network) -> Result<Self, Self::Error> {
+        match network {
+            zcash_address::Network::Main => Ok(Network::Mainnet),
+            zcash_address::Network::Test => Ok(Network::Testnet),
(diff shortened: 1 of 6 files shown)


zebra-chain/src/primitives/address.rs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+}
+
+// TODO:
+// addresses with embedded networks:
+//       implement TryFromRawAddress for sprout, sapling, unified/sapling, unified/orchard
+//       use convert_if_network(network.into()) to convert raw addresses
+// addresses which require a separate network:
+//       implement TryFromAddress for unified/transparent (p2pkh and p2sh)
```
teor2345, line 34: This might need some tweaks.

We don't have sapling or sprout address types, so we might need to re-export the `zcash_primitives` type. But we should get a `TryFromRawAddress` or `TryFromAddress` impl for free.

We do have an orchard type, so we can try to convert to that, if it contains enough information for the RPC. It's also ok to re-use the `zcash_primitives` orchard type.

(And if you'd like, you can use the `zcash_primitives` transparent type as well.)

teor2345:
Hopefully this is enough to get you started, here is what I get from this PR's RPC (it is not the right format):
```sh
$ zcash-cli parseaddress t1bDSccEgrRPu5J6EFsKGKbFPJqtDCXWdMJ
isvalid: true,
address: t1bDSccEgrRPu5J6EFsKGKbFPJqtDCXWdMJ,
scriptPubKey: TODO,
ismine: false,
isscript: false,
pubkey: TODO,
iscompressed: TODO,
account: ""

```

arya2:
> Hopefully this is enough to get you started

This is perfect, thank you!!

teor2345:
Arya has seen this, so I'm just going to close it

Issue #5823, opened 2022-12-08
transaction::tests::v4_with_modified_joinsplit_is_rejected fails with unexpected verifier error


teor2345:
## Motivation

> ---- transaction::tests::v4_with_modified_joinsplit_is_rejected stdout ----
> 
> The application panicked (crashed).
> Message:  assertion failed: `(left == right)`
>   left: `Err(RedJubjub(InvalidSignature))`,
>  right: `Err(Groth16("proof verification failed"))`
> Location: zebra-consensus/src/transaction/tests.rs:1578

https://github.com/ZcashFoundation/zebra/actions/runs/3645234772/jobs/6155200323#step:14:1505

### Specifications

TODO: quote the relevant spec sections from https://zips.z.cash/

### Complex Code or Requirements

This looks like a race condition between the proof and signature verifier. (Modifying the proof makes the signature invalid.)

We can fix this failure by accepting either error.

### Testing

It's going to be tricky to test that any fix works correctly.

conradoplg:
This is really baffling, I can't come up with a scenario where that happens. At first I thought the test could be changing the same transaction, but that's not it, since test_transactions deserializes blocks and thus returns fresh references each time.  The verifier is also freshly created each time.

teor2345:
Happened once on macOS:
> Error is for macos only in the v4_with_modified_joinsplit_is_rejected test: https://github.com/ZcashFoundation/zebra/actions/runs/4068002665/jobs/7006036670#step:14:1929

https://github.com/ZcashFoundation/zebra/pull/6044#issuecomment-1412710691

arya2:
> At first I thought the test could be changing the same transaction

The `async_checks` are unordered and the `verify_sapling_shielded_data` check may occasionally be faster.

I don't see it filtering out v5 transactions for the test so I wonder if it may be calling `verify_orchard_shielded_data` as well.

> Accept either error?

We could alternatively modify the signature when modifying the proof so only the proof verification check would fail if modifying the proof makes the signature invalid?

teor2345:
> > At first I thought the test could be changing the same transaction
> 
> The `async_checks` are unordered and the `verify_sapling_shielded_data` check may occasionally be faster.
> 
> I don't see it filtering out v5 transactions for the test so I wonder if it may be calling `verify_orchard_shielded_data` as well.
> 
> > Accept either error?
> 
[...]

PR #6044, opened 2023-01-27
fix(getblocktemplate): change error format in proposals

### LINKED ISSUE #5981: Use the same format as `zcashd` for `getblocktemplate` proposal mode errors

oxarbitrage:
## Motivation

We want to match zcashd error format in getblocktemplate proposal mode responses.

Close https://github.com/ZcashFoundation/zebra/issues/5981

## Solution

Create `kabeb-case` errors for `ProposalResponse` from zebra json.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -42,6 +42,7 @@ jsonrpc-core = "18.0.0"
 jsonrpc-derive = "18.0.0"
 jsonrpc-http-server = "18.0.0"
 num_cpus = "1.15.0"
+regex = "1.7.1"
 
 # zebra-rpc needs the preserve_order feature in serde_json, which is a dependency of jsonrpc-core
 serde_json = { version = "1.0.91", features = ["preserve_order"] }
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -56,7 +56,7 @@ tracing-futures = "0.2.5"
 hex = { version = "0.4.3", features = ["serde"] }
 serde = { version = "1.0.152", features = ["serde_derive"] }
 
-# Experimental features getblocktemplate-rpcs
+# Experimental feature getblocktemplate-rpcs
 rand = { version = "0.8.5", package = "rand", optional = true }
 
 # Test-only feature proptest-impl
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -15,7 +15,6 @@ default = []
 # Experimental mining RPC support
 getblocktemplate-rpcs = [
     "rand",
-    "regex",
     "zebra-consensus/getblocktemplate-rpcs",
     "zebra-state/getblocktemplate-rpcs",
     "zebra-node-services/getblocktemplate-rpcs",
@@ -59,7 +58,6 @@ serde = { version = "1.0.152", features = ["serde_derive"] }
 
 # Experimental features getblocktemplate-rpcs
 rand = { version = "0.8.5", package = "rand", optional = true }
-regex = { version = "1.7.1", package = "regex", optional = true }
 
 # Test-only feature proptest-impl
 proptest = { version = "0.10.1", optional = true }
--- a/zebra-rpc/Cargo.toml
+++ b/zebra-rpc/Cargo.toml
@@ -15,6 +15,7 @@ default = []
 # Experimental mining RPC support
 getblocktemplate-rpcs = [
     "rand",
+    "regex",
     "zebra-consensus/getblocktemplate-rpcs",
     "zebra-state/getblocktemplate-rpcs",
     "zebra-node-services/getblocktemplate-rpcs",
@@ -42,7 +43,6 @@ jsonrpc-core = "18.0.0"
 jsonrpc-derive = "18.0.0"
 jsonrpc-http-server = "18.0.0"
 num_cpus = "1.15.0"
-regex = "1.7.1"
 
 # zebra-rpc needs the preserve_order feature in serde_json, which is a dependency of jsonrpc-core
 serde_json = { version = "1.0.91", features = ["preserve_order"] }
@@ -57,8 +57,9 @@ tracing-futures = "0.2.5"
 hex = { version = "0.4.3", features = ["serde"] }
 serde = { version = "1.0.152", features = ["serde_derive"] }
 
-# Experimental feature getblocktemplate-rpcs
+# Experimental features getblocktemplate-rpcs
 rand = { version = "0.8.5", package = "rand", optional = true }
+regex = { version = "1.7.1", package = "regex", optional = true }
 
 # Test-only feature proptest-impl
 proptest = { version = "0.10.1", optional = true }
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
@@ -49,10 +49,10 @@ impl ProposalResponse {
         let error_kebab1 = format!("{error:?}")
             .replace(|c: char| !c.is_alphanumeric(), "-")
             .to_ascii_lowercase();
-
-        // Remove duplicated `-` characters.
-        let regex1 = regex::Regex::new(r"-+").expect("valid regex");
-        let error_kebab2 = regex1.replace_all(&error_kebab1, "-");
+        // Remove consecutive duplicated `-` characters.
+        let mut error_v: Vec<char> = error_kebab1.chars().collect();
+        error_v.dedup_by(|a, b| a == &'-' && b == &'-');
+        let error_kebab2: String = error_v.into_iter().collect();
         // Trim any leading or trailing `-` characters.
         let final_error = error_kebab2.trim_matches('-');
 
--- a/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
@@ -2,7 +2,7 @@
 //!
 //! `ProposalResponse` is the output of the `getblocktemplate` RPC method in 'proposal' mode.
 
-use std::{error::Error, num::ParseIntError, str::FromStr, sync::Arc};
+use std::{num::ParseIntError, str::FromStr, sync::Arc};
 
 use zebra_chain::{
     block::{self, Block, Height},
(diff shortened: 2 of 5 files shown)


zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
```diff
@@ -42,23 +42,22 @@ pub enum ProposalResponse {
...
+        let error_kebab1 = format!("{error:?}")
+            .replace(|c: char| !c.is_alphanumeric(), "-")
+            .to_ascii_lowercase();
+
+        // Remove consecutive `-` characters if any.
+        let regex1 = regex::Regex::new(r"-+").expect("valid regex");
+        let error_kebab2 = regex1.replace_all(&error_kebab1, "-");
+        // Remove trailing `-` character if any.
+        let regex2 = regex::Regex::new(r"-$").expect("valid regex");
+        let final_error = regex2.replace_all(&error_kebab2, "");
```
oxarbitrage: Maybe a list of regexes can be done here instead, i didn't tested it but i think this should be good enough as is too.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
```diff
@@ -42,23 +42,22 @@ pub enum ProposalResponse {
...
-        ProposalResponse::Rejected(format!("{error:#?}"))
-    }
-}
-
-impl<E: Error + Send + Sync + 'static> From<E> for ProposalResponse {
-    fn from(error: E) -> Self {
-        Self::error(error.into())
+    ///
+    /// Note: Error kind is currently ignored to match zcashd error format (`kebab-case` string).
+    pub fn rejected<S: ToString>(_kind: S, error: BoxError) -> Self {
```
arya2, line 47: Could we remove `_kind` altogether or add it to the message? (It's currently either "invalid proposal" or "invalid proposal format" so we could optionally make it an enum if we want to keep it).

Do you think we should also leave a TODO for explicitly downcasting and matching errors to constants?

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
```diff
@@ -42,23 +42,22 @@ pub enum ProposalResponse {
...
+        // Make error `kebab-case` to match zcashd format.
+        let error_kebab1 = format!("{error:?}")
+            .replace(|c: char| !c.is_alphanumeric(), "-")
+            .to_ascii_lowercase();
+
+        // Remove consecutive `-` characters if any.
+        let regex1 = regex::Regex::new(r"-+").expect("valid regex");
+        let error_kebab2 = regex1.replace_all(&error_kebab1, "-");
+        // Remove trailing `-` character if any.
+        let regex2 = regex::Regex::new(r"-$").expect("valid regex");
```
arya2: Optional/Nitpick: I don't think any of the errors that might be converted need to be checked for leading '-' characters, but in lieu of tests, could we check anyways?

```suggestion
        // Trim any leading or trailing `-` characters.
        let regex2 = regex::Regex::new(r"^-|-$").expect("valid regex");
```

zebra-rpc/Cargo.toml
```diff
@@ -42,6 +42,7 @@ jsonrpc-core = "18.0.0"
 jsonrpc-derive = "18.0.0"
 jsonrpc-http-server = "18.0.0"
 num_cpus = "1.15.0"
+regex = "1.7.1"
```
arya2: Could we make this an optional dependency that's required by the "getblocktemplate-rpcs" feature if we keep it.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
```diff
@@ -42,23 +42,22 @@ pub enum ProposalResponse {
...
+    /// Note: Error kind is currently ignored to match zcashd error format (`kebab-case` string).
+    pub fn rejected<S: ToString>(_kind: S, error: BoxError) -> Self {
+        // Make error `kebab-case` to match zcashd format.
+        let error_kebab1 = format!("{error:?}")
+            .replace(|c: char| !c.is_alphanumeric(), "-")
+            .to_ascii_lowercase();
+
+        // Remove consecutive `-` characters if any.
+        let regex1 = regex::Regex::new(r"-+").expect("valid regex");
+        let error_kebab2 = regex1.replace_all(&error_kebab1, "-");
```
arya2: I think adding the `regex` crate is okay, but if we want to remove it, we could use [`as_mut_vec`](https://doc.rust-lang.org/std/string/struct.String.html#method.as_mut_vec), [`dedup`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup), and [`trim_matches`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches) instead. 

`as_mut_vec` is unsafe because it might add invalid characters but in this case we would only be removing characters.

arya2 (commented):
This looks great, thank you!

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
teor2345: > Could we remove `_kind` altogether or add it to the message? (It's currently either "invalid proposal" or "invalid proposal format" so we could optionally make it an enum if we want to keep it).

I don't think it adds much to the error message, so I don't mind which way we fix this.

> Do you think we should also leave a TODO for explicitly downcasting and matching errors to constants?

This will only matter if miners care about the specific error text. I'll go ask some of the mining pools.

teor2345: (I just checked, the mining pools we're in contact with don't use block proposals currently.)

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
teor2345: We avoid using unsafe code when there is a safe alternative.

Here is a safe alternative to `as_mut_vec`:
```rust
let mut error: Vec<char> = error.chars().collect();
error.dedup();
let error: String = error.into_iter().collect();
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
teor2345: This will only trim a single leading or trailing character, using `trim_matches` would be more reliable here.

teor2345 (dismissed):
I think all the changes we've suggested are optional.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
oxarbitrage: Both suggestions are good to add, i made the `trim_matches` one at https://github.com/ZcashFoundation/zebra/pull/6044/commits/154386cc4ca94cfa9cff6b5fcce8eb6350556e33 so we have one regex less.

zebra-rpc/Cargo.toml
oxarbitrage: Yea, i had been lazy here, sorry. Added it at https://github.com/ZcashFoundation/zebra/pull/6044/commits/377025fada23ad4275262d7cb3aeb692e9341825

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
oxarbitrage: The only problem here and the reason we still have a regex left is with the `dedup()`. The method remove all the repeated characters so an error will end up looking as: 

`io-eror-kind-unexpectedeof-mesage-failed-to-fil-whole-bufer`

So i didn't applied it.

arya2: > Here is a safe alternative to as_mut_vec:

`as_mut_vec` wouldn't have worked either, I should've paid closer attention to the return type, it's a `Vec<u8>` so chars that are represented with two bytes would be two elements.

> The only problem here and the reason we still have a regex left is with the `dedup()`. The method remove all the repeated characters so an error will end up looking as:
> 
> `io-eror-kind-unexpectedeof-mesage-failed-to-fil-whole-bufer`
> 
> So i didn't applied it.

Oops. I think `dedup_by(|a, b| a == &'-' && b == &'-')` would work but it's optional so feel free to resolve this.

arya2 (dismissed):

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template/proposal.rs
oxarbitrage: That worked, thanks! https://github.com/ZcashFoundation/zebra/pull/6044/commits/a8b4d718bb8e5f1950c620de783c0f3f9d11d8d7

teor2345 (approved):
Thanks!

oxarbitrage:
There is an error in the CI but i think is unrelated to this PR so i am retrying.

Error is for macos only in the `v4_with_modified_joinsplit_is_rejected` test: https://github.com/ZcashFoundation/zebra/actions/runs/4068002665/jobs/7006036670#step:14:1929

teor2345:
> There is an error in the CI but i think is unrelated to this PR so i am retrying.
> 
> Error is for macos only in the `v4_with_modified_joinsplit_is_rejected` test: https://github.com/ZcashFoundation/zebra/actions/runs/4068002665/jobs/7006036670#step:14:1929

This seems to be bug #5823, which is unrelated.

PR #6064, opened 2023-01-31
fix(workflow): Name the GitHub projects workflow

teor2345:
## Motivation

If we add a name to this workflow, it is displayed in the actions list at:
https://github.com/ZcashFoundation/zebra/actions

This makes it easier to know what it does.

## Review

This is Pili's workflow.


mpguerra (approved):
Thank you!

PR #6067, opened 2023-02-01
Temporary RPC stubs for `get_difficulty` and `get_block` fields

teor2345:
## Motivation

This PR contains some RPC stubs or implementations that can be used until we implement these RPCs properly and merge them to `main`.


teor2345:
All the essential parts of this PR will be in the `main` branch after #6097 merges (it's currently in the queue).

PR #6066, opened 2023-02-01
fix(rpc): Make RPC "incorrect parameters" error code match `zcashd`

teor2345:
## Motivation

Some miners expect specific RPC error codes.

Closes #6054

## Solution

- return error code -1 when the parameters are incorrect

I checked the error strings and they already seem to match.

Related changes:
- log all RPC errors at info level (we can turn this off or skip some errors if it gets noisy)
- simplify RPC error logging implementation

## Review

This is on the critical path, so I guess it's a high priority?

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

The only other error codes I could find are in the payment code, which Zebra doesn't support because it doesn't have a wallet:
https://github.com/s-nomp/s-nomp/blob/8567a4c649ab6f5ea37ac3c3901f2d9c2597199a/libs/paymentProcessor.js#L750

Issue #5934, opened 2023-01-09
Manually test that Zcash mining pools can use Zebra to mine blocks on testnet


teor2345:
## Motivation

We should run Zebra with mining pool software before asking other mining pools to test it.

### Complex Code or Requirements

A bunch of us have done some work on this, it's tricky to get the right dependencies to deploy most mining pools.

## Related Work

- #5686
- #5803
- #5935

Here is our previous [discussion](https://github.com/ZcashFoundation/zebra/discussions/5804) about mining pool testing.

teor2345:
I think we should put a time limit of one day on this ticket. If we can't get it working quickly, we can contact an actual mining pool later to help with testing, when we've done more RPCs.

mpguerra:
> I think we should put a time limit of one day on this ticket. If we can't get it working quickly, we can contact an actual mining pool later to help with testing, when we've done more RPCs.

Agreed

@ZcashFoundation/zebra-team can whoever has done some research and/or testing add their notes to this issue.

Scheduling for Sprint 2 as I'd like to get all of the testing resolved in the next sprint

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/5934?planning-poker) with Zenhub @arya2 @conradoplg @dconnolly @oxarbitrage @teor2345 @upbqdn

mpguerra:
moving to Sprint 3 for now

teor2345:
I did some testing on this back in December, it was tricky because I couldn't get the node dependencies right. And some of the C libraries wouldn't compile on my machine.

But maybe using the node dependency lock files will help.

upbqdn:
I made it work by modifying the source code and installation configs. I could pack it all into some container.

teor2345:
Even just a diff from the original source would help!

mpguerra:
Let's document any successes we've had with this here and what we have done. This will help with #6028

teor2345:
> Let's document any successes we've had with this here and what we have done. This will help with #6028

These instructions are now at:
https://github.com/ZcashFoundation/s-nomp/blob/zebra-mining/zebra-instructions.md

Anyone can open PRs to improve them.

arya2:
[...]

PR #6069, opened 2023-02-01
change(log): Log `submitblock` RPC successes and failures at info level

teor2345:
## Motivation

I needed these logs to work out why my miner wasn't providing correct blocks. Other people might find them useful.

Related to #5934.

We might disable these or make them debug-level before this feature goes into production.

## Review

Anyone can review this PR.

### Reviewer Checklist

## Follow Up Work

This is similar to the error logging in PR #6066, we might not need both.


arya2 (approved):
Looks good! This should come in handy.

PR #6091, opened 2023-02-02
fix(test): avoid failing getblocktemplate acceptance test when the chain tip changes

### LINKED ISSUE #6079: Ignore expected "proposal is not based on the current chain tip" error in getblocktemplate acceptance test
## Motivation

We've seen the `getblocktemplate` test fail in CI with:

> "proposal is not based on the current best chain tip"

(see https://github.com/ZcashFoundation/zebra/pull/6013#issuecomment-1412381467)

## Possible Design

- Use getblocktemplate with long-polling in acceptance test
- Ignore/retry the getblocktemplate request in proposal mode if a new template is received from the long-polling request
- Wait a second for the long-polling to return after the block proposal result if it fails because it's not based on the current best chain tip?

## Alternative Design

- We already try the test multiple times. Make the test pass if there is at least one success, and all of the errors are "proposal is not based on the current best chain tip".

## Another Alternative

> We might also want to think about how we handle "previous block is not the tip block" errors in CI, which are expected when a new block arrives. We might need to check the tip is the same before and after the proposal before we fail the test.
https://github.com/ZcashFoundation/zebra/pull/5963#issuecomment-1396154807

arya2:
## Motivation

We need to ensure that block templates remain valid for repeated proposal checks.

See https://github.com/ZcashFoundation/zebra/pull/5963#issuecomment-1396154807.

Closes #6079.

I was hoping to add/retain test coverage but this grew more complicated than expected so let me know if one of the alternative designs on that issue would be preferable.

This PR adds some concurrency in `try_validate_block_template`:
- A spawned task for making long poll requests and sending new templates with `submit_old: false` to `long_poll_result_rx` until receiving a message on `done_rx`
- A `select!` for cancelling block proposals and making new requests when receiving messages from `long_poll_result_rx`

## Solution

- Spawns a thread that calls `getblocktemplate` with long polling
- Interrupts and retries block proposal check when new templates with a `submit_old: false` are received

Related changes:
- Increases required number of successful proposals and decreases the interval to test long polling 

## Review

Let me know if one of the alternative designs would be preferable.

### Reviewer Checklist

## Follow Up Work

- Address review comments


zebrad/tests/common/rpc_client.rs
```diff
@@ -8,6 +8,7 @@ use reqwest::Client;
 use color_eyre::{eyre::eyre, Result};
 
 /// An http client for making Json-RPC requests
+#[derive(Clone)]
```
teor2345: Optional nitpick:
```suggestion
#[derive(Clone, Debug)]
```

teor2345 (dismissed):
Looks good, let's run it a few times to make sure that it works?

arya2:
> Looks good, let's run it a few times to make sure that it works?

It ran successfully a few times but I changed the constants to make 300 block proposals at 5 second intervals to be sure. I'll try 3000 if that succeeds.

Update: It succeeded with 300.

Update: 3000 failed with "not based on the current best chain tip", possibly because it was dropping the send futures.

Update: 1000 succeeded. I'm going to try a shorter `EXTRA_LONGPOLL_WAIT_TIME` because I'd like to be sure that longpolling is working as expected too.

Update: a shorter extra longpoll wait time worked too (with 2500 iterations).

teor2345 (approved):

teor2345 (changes requested):
Looks like you're dropping some futures, but you might want to be awaiting them?
https://github.com/ZcashFoundation/zebra/actions/runs/4071294784/jobs/7012969586#step:8:970

Because send is cancel-safe, you might not be sending any messages on these channels at all:
https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send

Depending on the behaviour you want, you could await the futures or use blocking_send(). If there is any risk of deadlocks or hangs, then try_send() or send_timeout() might be better.

arya2:
> Looks like you're dropping some futures, but you might want to be awaiting them? https://github.com/ZcashFoundation/zebra/actions/runs/4071294784/jobs/7012969586#step:8:970
> 
> Because send is cancel-safe, you might not be sending any messages on these channels at all: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send
> 
> Depending on the behaviour you want, you could await the futures or use blocking_send(). If there is any risk of deadlocks or hangs, then try_send() or send_timeout() might be better.

🤦Thank you. I somehow missed the `async` on that fn. Updated my rust-analyzer settings to check clippy on save for next time.

arya2:
I updated the constants to run the proposal test 1000x at 300ms interval. The log for an interrupt by a template with `submit_old: false` is usually visible in CI now:

> 2023-02-02T22:12:38.388229Z  INFO got longpolling response with submitold of false before result of proposal tests

https://github.com/ZcashFoundation/zebra/actions/runs/4078725058/jobs/7030039421#step:8:1096

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
```diff
@@ -27,9 +30,18 @@ use crate::common::{
...
+/// Delay between attempts to validate a template as block proposals.
+///
+/// Running many iterations in short intervals tests that long poll requests correctly
+/// return `submit_old: false` when the old template becomes invalid.
+pub const BLOCK_PROPOSAL_INTERVAL: Duration = Duration::from_millis(300);
+
 /// Number of times we want to try submitting a block template as a block proposal at an interval
 /// that allows testing the varying mempool contents.
-const NUM_SUCCESSFUL_BLOCK_PROPOSALS_REQUIRED: usize = 10;
+const NUM_SUCCESSFUL_BLOCK_PROPOSALS_REQUIRED: usize = 1000;
```
teor2345, line 44: This seems like an unusually large number of test iterations to commit to CI, how long does it take?
We typically use 256 iterations for quick proptests, and 1-32 for slower tests.

Optional: If we need a larger number for some reason, please document it here. (Or refer to `BLOCK_PROPOSAL_INTERVAL` if the reason is covered by its docs - it seems like it might be, but it's easy to miss.)

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
```diff
@@ -27,9 +30,18 @@ use crate::common::{
...
 pub const EXPECTED_MEMPOOL_TRANSACTION_TIME: Duration = Duration::from_secs(45);
 
+/// Delay between getting block proposal results and cancelling long poll requests.
+pub const EXTRA_LONGPOLL_WAIT_TIME: Duration = Duration::from_millis(150);
+
+/// Delay between attempts to validate a template as block proposals.
+///
+/// Running many iterations in short intervals tests that long poll requests correctly
+/// return `submit_old: false` when the old template becomes invalid.
+pub const BLOCK_PROPOSAL_INTERVAL: Duration = Duration::from_millis(300);
```
teor2345, line 40: This is exactly `DEFAULT_MAX_REQUEST_DELAY`, should it be set to that constant, so it automatically adapts if CI gets more unreliable? 
```suggestion
pub const BLOCK_PROPOSAL_INTERVAL: Duration = zebra_test::mock_service::DEFAULT_MAX_REQUEST_DELAY;
```

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
```diff
@@ -27,9 +30,18 @@ use crate::common::{
 /// We've seen it take anywhere from 1-45 seconds for the mempool to have some transactions in it.
 pub const EXPECTED_MEMPOOL_TRANSACTION_TIME: Duration = Duration::from_secs(45);
 
+/// Delay between getting block proposal results and cancelling long poll requests.
+pub const EXTRA_LONGPOLL_WAIT_TIME: Duration = Duration::from_millis(150);
```
teor2345, line 34: This is shorter than the delays we've seen on busy CI runners:
https://github.com/ZcashFoundation/zebra/blob/dc43dca06f8eed275235110e57e643caef0daa26/zebra-test/src/mock_service.rs#L82

If our tests will fail if the test process is paused for more than this time, then we should set it to `DEFAULT_MAX_REQUEST_DELAY`. (We do this in other tests so they automatically adapt when we fix failures due to slow CI machines.)

If it needs to be lower than `BLOCK_PROPOSAL_INTERVAL`, then we might want to increase both.

Either way, can you document why this value is set to this particular amount, and what relationship it needs with `BLOCK_PROPOSAL_INTERVAL`? If it needs to be less, you could add an assert in the test that uses it to make sure.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
```diff
@@ -90,23 +102,29 @@ pub(crate) async fn run() -> Result<()> {
 
     assert!(is_response_success);
 
+    tokio::time::sleep(EXPECTED_MEMPOOL_TRANSACTION_TIME).await;
```
teor2345, line 105: Optional: Should we run our first iteration with an empty mempool, for better test coverage?

teor2345 (approved):
I have some concerns about the reliability of the timings used here, but I'm not sure if they are blockers.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
arya2: > This seems like an unusually large number of test iterations to commit to CI, how long does it take?

It takes ~15m locally for me, I think 6-30 minutes in general. 

It waits 45s for the mempool, and after that, `try_validate_block_template` seems to take about ~900ms, so the 300ms interval is now in parallel to those requests is effectively ignored right now. I think I should change that to 1 second (which would leave it at 18-30 minutes.)

I chose 1000 because `submit_old` is usually false ~1/1000 calls to `try_validate_block_template`.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
arya2: I think I need to increase this interval, it should be unrelated to CI reliability because it's only meant to be a minimum that ensures the test runs across various chain states regardless of how long `try_validate_block_template` takes.

But that's interesting!

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
arya2: > This is shorter than the delays we've seen on busy CI runners:

I'm not sure if the delay is necessary at all, validating a block proposal may take longer than creating a template, so this should perhaps be a minimum time rather than an added delay, but in any case this test uses `deploy-gcp-tests` in CI. I thought it was a Github runner and went with 10x the expected response time for a block template request originally (deserialization will take some time too), but I think 3x is okay for GCP?

> If it needs to be lower than BLOCK_PROPOSAL_INTERVAL, then we might want to increase both.

It does, I think we want to increase `BLOCK_PROPOSAL_INTERVAL` so that `try_validate_block_template` runs at consistent intervals and is useful, but that this is okay if it's running on a fast GCP compute instance.

> Either way, can you document why this value is set to this particular amount, and what relationship it needs with BLOCK_PROPOSAL_INTERVAL?

Yeah, these constants need definitely better explanations.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
arya2: Yes! I wasn't entirely sure why that was there before so I didn't want to change it, but I don't like it and will get rid of it if that's okay.

arya2:
> I have some concerns about the reliability of the timings used here, but I'm not sure if they are blockers.

These are excellent notes that have helped me notice some additional mistakes in this PR, but I'd like to handle these in a new PR so that the issue doesn't cause any more improper CI failures in the meantime.

Your concerns are perceptive, the timings - particularly `BLOCK_PROPOSAL_INTERVAL` - are wrong or flawed and in need of better explanations.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
teor2345: Ok, I think that's reasonable, but can you please check that this is not the longest CI job now?

If it is, we could reduce the number of iterations to 500.

zebrad/tests/common/get_block_template_rpcs/get_block_template.rs
teor2345: > I thought it was a Github runner and went with 10x the expected response time for a block template request originally (deserialization will take some time too), but I think 3x is okay for GCP?

At the moment these test runners have lots of CPUs, so we should be getting reasonably good latency. I think they are dedicated machines, but I am not sure, Gustavo would know.

teor2345:
@mergifyio rebase

teor2345:
@arya2 did you want to fix the merge conflict then I can approve?

teor2345 (approved):
Let's give this a shot!

teor2345:
In your next PR, would you mind reducing the verbose test output, if it isn't already fixed in this PR?

> 2023-02-03T03:52:28.131089Z  INFO got getblocktemplate response, might not have transactions response_text="{\"jsonrpc\":\"2.0\",\"result\":{\"capabilities\":[\"proposal\"],\"version\":4,\"previousblockhash\":\"00000000003a0c30f40de72e22e59a07e33adb56c9185b82b57a35ce415112d8\",\"blockcommitmentshash\":\"3eb2e1161099c57fc790abdddf84d13aadcb857873036a9b847f6e0006f55126\",\"lightclientroothash\":\"3eb2e1161099c57fc790abdddf84d13aadcb857873036a9b847f6e0006f55126\",\"finalsaplingroothash\":\"3eb2e1161099c57fc790abdddf84d13aadcb857873036a9b847f6e0006f55126\",\"defaultroots\":{\"merkleroot\":\"252af63d99a964648e66b8b0e2c4fb3ec4c9a78dacc5ba093d5a6079f151955b\",\"chainhistoryroot\":\"010537ba23f1a4932eae08cc122fa203afede82cbb939087b4ee8b9014c9e93f\",\"authdataroot\":\"dc40ac2b3a4ae92e4aa0d42abeea6934ef91e6ab488772c0466d7051180a4e83\",\"blockcommitmentshash\":\"3eb2e1161099c57fc790abdddf84d13aadcb857873036a9b847f6e0006f55126\"},\"transactions\":[{\"data\":\"040000808520...

https://github.com/ZcashFoundation/zebra/actions/runs/4080594267/jobs/7033830311#step:8:199

Some of these transactions, blocks, or templates can contain 4MB of hex data. And sometimes GitHub won't load the logs if they are too big.

PR #6098, opened 2023-02-03
change(log): Log a cute message for blocks that were mined by Zebra (off by default)

teor2345:
## Motivation

It's nice to see when Zebra has mined some blocks:
> 2023-02-03T05:52:06.475443Z  INFO zebra_state::service::write: looks like this block was mined by Zebra! text="z🦓" data="7af09fa693" height=2215022 hash=00531335a993d94aed848c76cceba5274e2913b14a286d6a485b44444eb3a9d6
> 2023-02-03T05:52:06.475598Z  INFO zebra_rpc::methods::get_block_template_rpcs: submit block accepted block_hash=block::Hash("00531335a993d94aed848c76cceba5274e2913b14a286d6a485b44444eb3a9d6") block_height="2215022"

This is part of #5934, testing v5 transactions, extra coinbase data, and other minor changes allowed by the consensus rules or RPCs.

### Specifications

Coinbase transactions can contain up to 100 bytes of data, but the consensus rules only specify the first 1-6 bytes.

## Solution

- Mark Zebra coinbase transactions with extra coinbase data if `like_zcashd` is false or if specifically configured
    - Make `debug_like_zcashd` into a config rather than a constant
    - Make a config for `extra_coinbase_data`
- Log when we commit a block mined by Zebra to our state
    - Don't log control characters
    - Rate-limit to 1 in 1000 blocks

Related changes:
- Reduce logging instrumentation during block writes
- Add network and commit to write task logs

## Review

This is a low-priority review.

### Reviewer Checklist


teor2345:
When I mine a block with this PR, here is what `zebrad` logs:
Screenshot 2023-02-03 at 16 08 25

And when it is loading blocks mined by other `zebrad`s:
Screenshot 2023-02-03 at 16 08 47

zebra-chain/src/transparent.rs
```diff
@@ -182,24 +194,54 @@ impl fmt::Display for Input {
...
+        // "No extra coinbase data" is the default.
+        let data = data.unwrap_or_default();
+        let height_size = height.coinbase_zcash_serialized_size();
+
+        assert!(
+            data.len() + height_size <= MAX_COINBASE_DATA_LEN,
+            "invalid coinbase data: extra data {} bytes + height {height_size} bytes \
+             must be {} or less",
+            data.len(),
+            MAX_COINBASE_DATA_LEN,
```
upbqdn, line 224: ###### Optional

The variable placeholders in the string feel non-consistent since two are empty, and one is not.

zebra-state/src/service/write.rs
```diff
@@ -328,13 +328,37 @@ fn log_if_mined_by_zebra(tip_block: &ChainTipBlock) {
...
+            );
+        } else {
+            // # Security
+            //
+            // Use the extra data as an allow-list, replacing unknown characters.
+            // This makes sure control characters and harmful messages don't get logged
+            // to the terminal.
+            let text = text.replace(
+                |c| {
+                    !EXTRA_ZEBRA_COINBASE_DATA
```
arya2, line 377: I think we should always omit the text, partial leading or trailing unicode char codes don't seem useful.

zebra-state/src/service/write.rs
```diff
@@ -328,13 +328,37 @@ fn log_if_mined_by_zebra(tip_block: &ChainTipBlock) {
...
-            %hash,
-            "looks like this block was mined by Zebra!"
-        );
+
+        if coinbase_data.as_ref() == EXTRA_ZEBRA_COINBASE_DATA.as_bytes() {
+            info!(
+                %text,
+                %height,
+                %hash,
+                "looks like this block was mined by Zebra!"
```
arya2, line 367: Nitpick: do we want to say "with Zebra" instead?

("looks like" feels like sufficient hedging for someone deciding to include the emoji in their coinbase data that they got from 'zcashd'.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -227,6 +229,14 @@ where
     /// Zebra currently only supports transparent addresses.
     miner_address: Option<transparent::Address>,
 
+    /// Extra data to include in coinbase transaction inputs.
+    /// Limited to around 95 bytes by the consensus rules.
```
upbqdn, line 233: ###### Optional

```suggestion
    /// Limited to around 95 bytes by the consensus rules,
    /// depending on the length of the serialized height.
```

zebra-rpc/src/methods/get_block_template_rpcs/config.rs
```diff
@@ -14,4 +14,28 @@ pub struct Config {
     /// Zebra sends mining fees and miner rewards to this address in the
     /// `getblocktemplate` RPC coinbase transaction.
     pub miner_address: Option<transparent::Address>,
+
+    /// Extra data to include in coinbase transaction inputs.
+    /// Limited to around 95 bytes by the consensus rules.
```
upbqdn, line 19: ###### Optional

```suggestion
    /// Limited to around 95 bytes by the consensus rules,
    /// depending on the length of the serialized height.
```

zebra-chain/src/transaction/builder.rs
```diff
@@ -112,20 +113,23 @@ impl Transaction {
...
         if like_zcashd {
-            // TODO: add a debug_like_zcashd config and use 0x00 for like_zcashd
-            //       add an arbitrary extra coinbase data config?
-            //extra_data = Some(vec![0x00]);
-            extra_data = Some(EXTRA_ZEBRA_COINBASE_DATA.as_bytes().to_vec());
+            extra_data = Some(vec![0x00]);
             sequence = Some(u32::MAX);
         }
 
+        // Override like_zcashd if extra_coinbase_data was supplied
```
arya2, line 128: 👍

zebra-state/src/service/write.rs
```diff
@@ -285,3 +323,72 @@ pub fn write_blocks_from_channels(
...
+    last_zebra_mined_log_height: &mut Option<Height>,
+) {
+    // This logs at most every 2-3 checkpoints, which seems fine.
+    const LOG_RATE_LIMIT: u32 = 1000;
+
+    let height = tip_block.height.0;
+
+    if let Some(last_height) = last_zebra_mined_log_height {
+        if height < last_height.0 + LOG_RATE_LIMIT {
+            // If we logged in the last 1000 blocks, don't log anything now.
```
arya2, line 342: Optional: a metrics counter would be nice here.

arya2 (approved):
This looks great! Thank you for adding this, it's nice to see Zebra committing blocks that were mined with Zebra.

zebra-state/src/service/write.rs
teor2345: Happy to do this in a follow-up PR to avoid conflicts.

zebra-state/src/service/write.rs
teor2345: Unfortunately, we can't guarantee that Zebra is the only node that is adding this specific coinbase data, or that every miner that uses Zebra keeps our coinbase data. So this describes the situation accurately.

zebra-state/src/service/write.rs
teor2345: `from_utf8_lossy()` deals with this already: partial unicode characters are illegal in Rust strings.
https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy

zebra-state/src/service/write.rs
```diff
@@ -285,3 +323,72 @@ pub fn write_blocks_from_channels(
...
+                %height,
+                hash = %tip_block.hash,
+                "looks like this block was mined by Zebra!"
+            );
+        } else {
+            // # Security
+            //
+            // Use the extra data as an allow-list, replacing unknown characters.
+            // This makes sure control characters and harmful messages don't get logged
+            // to the terminal.
```
upbqdn, line 374: Maybe we should change `logged to the terminal` to `written to the log`?

teor2345 (commented):
Thanks for the review, I'll do the metrics in a follow-up PR.

upbqdn (approved):
Looks great, I left a few minor suggestions.

I also noticed this comment https://github.com/zcashfoundation/zebra/blob/f85673c9aa8794efff68f34741afe49abebce5c0/zebra-chain/src/transparent/serialize.rs#L184 says the "The height may produce `0..=5` initial bytes of coinbase data.", which was confusing since the consensus rules say "The length of heightBytes MUST be in the range {1 .. 5}." Should we refactor that comment?

zebra-state/src/service/write.rs
teor2345: The specific security issue is about terminal control codes changing the appearance of the terminal, and causing the user to enter malicious commands. This is less of an issue for log files.

zebra-chain/src/transparent.rs
teor2345: Happy to fix this in a follow-up PR.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Happy to fix this in a follow-up PR.

teor2345:
Follow-ups:

Consensus rule comment mistake:
* should be `1..=6` initial bytes of coinbase data, small heights are 1, other heights are `heightBytes + 1`
* https://github.com/ZcashFoundation/zebra/pull/6098#pullrequestreview-1310017889

Optional:
* metrics for the number of likely Zebra blocks:
    * https://github.com/ZcashFoundation/zebra/pull/6098#discussion_r1114984962
* put a constant inline in a format string:
    * https://github.com/ZcashFoundation/zebra/pull/6098#discussion_r1114850773
* comment grammar:
    * https://github.com/ZcashFoundation/zebra/pull/6098#discussion_r1114955582
    * https://github.com/ZcashFoundation/zebra/pull/6098#discussion_r1114955890

zebra-state/src/service/write.rs
arya2: I think there could be an issue around comparing the character in `text` to the characters in the character code. Would filtering characters after the expected🦓 with [`is_control`](https://doc.rust-lang.org/stable/std/primitive.char.html#method.is_control) work if that's an issue?

teor2345: I'm not sure how that could happen, can you give an example?

Here is my analysis:

`replace()` is used as an allow-list, replacing every character except for `Z`, `z`, and the Zebra emoji with `?`.

These characters and their UTF-8 bytes are ok to write to a terminal, because they are all printable:
> [240, 159, 166, 147]

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7f5a7d8711a765c74414310e01a10b5f
https://www.ascii-code.com/

Here are the docs for replace():

replace:
> Replaces all matches of a pattern with another string.

https://doc.rust-lang.org/std/string/struct.String.html#method.replace

Pattern:
> F: FnMut(char) -> bool
> F returns true for a char in string

https://doc.rust-lang.org/std/str/pattern/trait.Pattern.html

arya2: > I'm not sure how that could happen, can you give an example?

It can't happen, your code is correct, I misunderstood `EXTRA_ZEBRA_COINBASE_DATA` to be a hex value in string format.

teor2345: Thanks for checking, it's important that we make sure our code is secure!

PR #6099, opened 2023-02-03
feat(rpc): implement `getdifficulty`

### LINKED ISSUE #6055: Implement the `getdifficulty` RPC
## Motivation

This is one of the initial RPCs called by `node-stratum-pool`:
https://github.com/s-nomp/node-stratum-pool/blob/d86ae73f8ff968d9355bb61aac05e0ebef36ccb5/lib/pool.js#L387-L392

It is required to detect proof of work or proof of stake coins:
https://github.com/s-nomp/node-stratum-pool/blob/d86ae73f8ff968d9355bb61aac05e0ebef36ccb5/lib/pool.js#L424-L427

The pool won't run if this RPC returns an error:
https://github.com/s-nomp/node-stratum-pool/blob/d86ae73f8ff968d9355bb61aac05e0ebef36ccb5/lib/pool.js#L413-L416

### Specifications

https://zcash.github.io/rpc/getdifficulty.html

We have an existing state request for the difficulty, so we could just use that for now?
(That state request also gets the history tree, which we might want to change in a later PR, but only if it impacts performance.)

### Testing

We can do a snapshot test and manually test with `s-nomp` or `equihash-solomining`.

## Related Work

I'm pretty sure this blocks #5934 with `s-nomp` and `equihash-solomining`.

oxarbitrage:
## Motivation

We need to implement `getdifficulty`. Close https://github.com/ZcashFoundation/zebra/issues/6055

## Solution

I ported the calculation from zcashd. Maybe it can be improved but it seems to work as it is.

## Review

Anyone can review.

### Reviewer Checklist


oxarbitrage:
I am using `zcash-rpc-diff` to compare results. Most of the time the output is the same in both implementations however in some cases there are some rounding differences. It seems rust `f64` has a bit more precision than c++. Here is an example:

```
$ ./zebra-utils/zcash-rpc-diff 6666 getdifficulty
Connected to zebrad (port 6666) and zcashd (/media/oxarbitrage/4eb53770-4e4f-4d80-a830-0914f3f5b89a/zcash/zcash/src/zcash-cli zcash.conf port).

Request:
getdifficulty

Querying zebrad main chain at height >=1971240...

real	0m0,002s
user	0m0,000s
sys	0m0,002s

Querying zcashd main chain at height >=1971240...

real	0m0,002s
user	0m0,002s
sys	0m0,000s


--- /tmp/tmp.KRvW7JPCAV.rpc-diff/zebrad-main-1971240-getdifficulty.json	2023-02-03 11:41:27.640150516 -0300
+++ /tmp/tmp.KRvW7JPCAV.rpc-diff/zcashd-main-1971240-getdifficulty.json	2023-02-03 11:41:27.644150290 -0300
@@ -1 +1 @@
-102712303.47499941
+102712303.4749994
$ 
```

teor2345:
> It seems rust `f64` has a bit more precision than c++.

I think this happens when printing `f64`, Rust guarantees that printing then reading a float will give the same value:
https://doc.rust-lang.org/std/primitive.f32.html

C++ does not give format guarantees for `double`, it depends on the platform and standard C++ or C library:
https://stackoverflow.com/questions/6950738/number-of-precision-digits-for-double-in-c-different-in-windows-and-linux-why

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -904,6 +910,58 @@ where
...
+
+            while n_shift < n_shift_amount {
+                d_diff *= 256.0;
+                n_shift += 1;
+            }
+
+            while n_shift > n_shift_amount {
+                d_diff /= 256.0;
+                n_shift -= 1;
+            }
```
teor2345, line 959: This looks like a re-implementation of `CompactDifficulty::to_expanded()`, but on `f64`, rather than `U256`:
https://github.com/ZcashFoundation/zebra/blob/d67b3b641d2a9bb5680c91cd3281bddf40670c42/zebra-chain/src/work/difficulty.rs#L148-L156

I think we can do this in a simpler way using our existing code, see PR #6105 for my suggestion. This helps us avoid duplicate implementations of the consensus rules.

This doesn't block merging this PR, but I'd like to get the simpler code and extra tests merged soon.

teor2345 (approved):
Thanks! This looks good, and the tests and differences seem ok.

I think we can make the code a bit simpler, see my suggestion in PR #6105. I also added some more tests in that PR for valid difficulties.

teor2345:
Can you add labels to this PR, and assign yourself?

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: I agree that your code is simpler but i don't see the consensus rules here, this is just for displaying (no `# Consensus` doc or similar was added here).

The problem i see with the new code is the diff which seems to be bigger than the one from this code. 

The extra tests are fine.

teor2345: > I agree that your code is simpler but i don't see the consensus rules here, this is just for displaying (no `# Consensus` doc or similar was added here).

The code isn't consensus-critical, but it uses the same consensus rules to display an accurate difficulty value to users.

So maybe there could be a doc in this PR that quotes the `ToTarget` consensus rules:
https://zips.z.cash/protocol/protocol.pdf#nbits

Or for PR #6105, we should turn these existing comments into consensus docs, because that's the consensus implementation we're using for the RPC and in other code that is actually consensus-critical:
https://github.com/ZcashFoundation/zebra/blob/7c36ff611fc9ad7fd5d826aa8380a0290ed5faab/zebra-chain/src/work/difficulty.rs#L96
https://github.com/ZcashFoundation/zebra/blob/7c36ff611fc9ad7fd5d826aa8380a0290ed5faab/zebra-chain/src/work/difficulty.rs#L318

teor2345: Just to be clear, this isn't a blocker, feel free to merge this PR and we can fix it up in the next PR.

PR #6103, opened 2023-02-05
fix(ci): Fix full sync timeouts

teor2345:
## Motivation

The full sync test is timing out, because the blockchain has grown, so we need to do the fixes here:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-ci-sync-timeouts

## Solution

- Add 2 full sync CI jobs to get us to 1,960,000 blocks
- Increase the full sync Rust test timeout to 58 hours (4 hours for each of the jobs, plus some extra)

## Review

This is urgent because CI is failing on the `main` branch

### Reviewer Checklist


arya2 (approved):
Looks good!

PR #6122, opened 2023-02-09
fix(state): Fix minute-long delays in block verification after a chain fork

### LINKED ISSUE #4794: Revert note commitment and history trees when forking non-finalized chains
## Motivation

When Zebra forks a non-finalized chain, it rebuilds the note commitment trees from the finalized tip.
But this can execute 100 blocks worth of note commitment tree updates, which uses a lot of CPU.

This change is a high priority for miners. They are unlikely to use Zebra unless it is fixed, because it prevents them mining 1-2 blocks after each chain fork.

We're seeing rebuild times of around 75 seconds, which is way too long:
> 2022-07-20T20:34:08.026902Z  INFO {zebrad="b52b10c" net="Main"}:{peer=Out("3.72.134.66:8233")}:msg_as_req{msg="inv"}:inbound:download_and_verify{hash=00000000017131670f50301f97d80699baaef991ffed0c4ec9b6a0b0c189e587}:state: zebra_state::service::non_finalized_state::chain: finished rebuilding note commitment trees after a non-finalized chain fork rebuild_time=75.654621465s fork_height=Height(1744525) fork_tip=block::Hash("0000000000f4dd1105f94c06f28189ebdd60c0895d8931486847cce3278590e5")

### Consensus Rules

The consensus rules require the trees to match the chain tip, including all the parent blocks back to genesis.

The revert needs to happen correctly, so the trees match the new fork tip block. (Forking is an efficiency thing, the consensus rules don't specify how to do forks.)

If we remove duplicate storage of tip trees by height and as their own field, we will only have to update them in one place. This makes keeping them consistent very easy.

### Designs

Instead of rebuilding the trees from the root, we should remove the trees and anchors that are above the fork.

Refactors:
- [x] replace tip sprout, sapling, orchard, history tree fields with methods that do a lookup using the tip height

Change:
- [x] during a fork, revert sprout, sapling, orchard, history trees by removing anchors and trees above the fork height in `revert_chain_with()` for sprout, sapling, orchard shielded data, and history trees
- [x] remove the code that rebuilds the history trees in `rebuild_trees_parallel()`

Cleanup:
- [x] remove `Chain::with_trees()` and use `derive(Clone)` instead
- [x] remove unused tree arguments from `Chain::fork()`
- [x] cleanup any methods in `NonFinalizedState` that call these methods

## Related Work

This might be needed by mining pools.

We added a log in PR #4795, which told us this can take around 75 seconds.

teor2345:
## Motivation

Zebra can take multiple minutes to rebuild note commitment trees after a chain fork in the non-finalized chain. This also delays mining template updates and verifying mined blocks.

It only impacts mainnet at the moment, it is caused by a large number of shielded transactions in blocks.

Closes #4794.

### Specifications

This is a data structure and processing refactor. The consensus rule checks don't change, but we need to correctly store data so that we implement the consensus rules correctly.

This PR deletes a lot of complex code.

It temporarily stores the finalized tip trees and anchors in the tree indexes at the finalized tip height. These temporary entries are deleted when the first root block in the chain is finalized.

## Solution

- Replace sprout, sapling, orchard, and history tree fields with lookup methods
- Remove unused tree rebuild code
- Replace a custom clone function with `derive(Clone)`
- Remove unused arguments

Related changes:
- Refactor tree and anchor addition and removal functions
- Print sprout and sapling tree Nodes as hex when debugging
- Show full debug info when tests fail because chains aren't equal

### Testing

We already have good test coverage for this code. I am also running a partial and full sync locally on both mainnet and testnet.

## Review

There is an initial refactor to get the height indexes working in 98dfa875d429a05c27bb83ea1f0b929153b48eb6.
Then this PR repeats similar changes for sprout, sapling, and orchard. The best example is probably commit 3b0abd49130b715d535811f105cceb45beb1b474.
The history tree change is slightly different, it's commit 09e514891820dcf1fb7df767dcf793f5a556f50f.

This is a blocker for mining pools on mainnet, so it is a routine bug fix.

### Reviewer Checklist

## Follow Up Work

We could do further refactors, but this seemed like enough for now.

### DIFF
--- a/zebra-chain/src/sapling/tree.rs
+++ b/zebra-chain/src/sapling/tree.rs
@@ -169,8 +169,14 @@ impl ZcashDeserialize for Root {
 ///
 /// Note that it's handled as a byte buffer and not a point coordinate (jubjub::Fq)
 /// because that's how the spec handles the MerkleCRH^Sapling function inputs and outputs.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 struct Node([u8; 32]);
+
+impl fmt::Debug for Node {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_tuple("Node").field(&hex::encode(self.0)).finish()
+    }
+}
 
 /// Required to convert [`NoteCommitmentTree`] into [`SerializedTree`].
 ///
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -10,7 +10,6 @@ use std::{
 
 use zebra_chain::{
     block::{self, Block},
-    history_tree::HistoryTree,
     parameters::Network,
     sprout, transparent,
 };
@@ -156,7 +155,7 @@ impl NonFinalizedState {
         let parent_hash = prepared.block.header.previous_block_hash;
         let (height, hash) = (prepared.height, prepared.hash);
 
-        let parent_chain = self.parent_chain(parent_hash, finalized_state.history_tree())?;
+        let parent_chain = self.parent_chain(parent_hash)?;
 
         // If the block is invalid, return the error,
         // and drop the cloned parent Arc, or newly created chain fork.
@@ -281,7 +280,7 @@ impl NonFinalizedState {
         // Clone function arguments for different threads
         let block = contextual.block.clone();
         let network = new_chain.network();
-        let history_tree = new_chain.history_tree.clone();
+        let history_tree = new_chain.history_tree_at_tip();
 
         let block2 = contextual.block.clone();
         let height = contextual.height;
@@ -470,7 +469,6 @@ impl NonFinalizedState {
     fn parent_chain(
         &mut self,
         parent_hash: block::Hash,
-        history_tree: Arc<HistoryTree>,
     ) -> Result<Arc<Chain>, ValidateContextError> {
         match self.find_chain(|chain| chain.non_finalized_tip_hash() == parent_hash) {
             // Clone the existing Arc<Chain> in the non-finalized state
@@ -482,7 +480,7 @@ impl NonFinalizedState {
                 let fork_chain = self
                     .chain_set
                     .iter()
-                    .find_map(|chain| chain.fork(parent_hash, history_tree.clone()).transpose())
+                    .find_map(|chain| chain.fork(parent_hash).transpose())
                     .transpose()?
                     .ok_or(ValidateContextError::NotReadyToBeCommitted)?;
 
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -11,7 +11,6 @@ use std::{
 use zebra_chain::{
     block::{self, Block},
     history_tree::HistoryTree,
-    orchard,
     parameters::Network,
     sprout, transparent,
 };
@@ -157,11 +156,7 @@ impl NonFinalizedState {
         let parent_hash = prepared.block.header.previous_block_hash;
         let (height, hash) = (prepared.height, prepared.hash);
 
-        let parent_chain = self.parent_chain(
-            parent_hash,
-            finalized_state.orchard_note_commitment_tree(),
-            finalized_state.history_tree(),
-        )?;
+        let parent_chain = self.parent_chain(parent_hash, finalized_state.history_tree())?;
 
         // If the block is invalid, return the error,
         // and drop the cloned parent Arc, or newly created chain fork.
@@ -453,7 +448,10 @@ impl NonFinalizedState {
 
     /// Returns `true` if the best chain contains `orchard_nullifier`.
     #[cfg(test)]
-    pub fn best_contains_orchard_nullifier(&self, orchard_nullifier: &orchard::Nullifier) -> bool {
+    pub fn best_contains_orchard_nullifier(
+        &self,
+        orchard_nullifier: &zebra_chain::orchard::Nullifier,
+    ) -> bool {
         self.best_chain()
             .map(|best_chain| best_chain.orchard_nullifiers.contains(orchard_nullifier))
             .unwrap_or(false)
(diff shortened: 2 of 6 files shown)


zebra-chain/src/sapling/tree.rs
```diff
@@ -169,9 +169,15 @@ impl ZcashDeserialize for Root {
 ///
 /// Note that it's handled as a byte buffer and not a point coordinate (jubjub::Fq)
 /// because that's how the spec handles the MerkleCRH^Sapling function inputs and outputs.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 struct Node([u8; 32]);
```
arya2, line 173: Optional: we could use `HexDebug` here.

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -278,151 +326,26 @@ impl Chain {
...
     #[allow(clippy::unwrap_in_result)]
-    pub fn fork(
-        &self,
-        fork_tip: block::Hash,
-        sprout_note_commitment_tree: Arc<sprout::tree::NoteCommitmentTree>,
-        sapling_note_commitment_tree: Arc<sapling::tree::NoteCommitmentTree>,
-        orchard_note_commitment_tree: Arc<orchard::tree::NoteCommitmentTree>,
-        history_tree: Arc<HistoryTree>,
-    ) -> Result<Option<Self>, ValidateContextError> {
+    pub fn fork(&self, fork_tip: block::Hash) -> Result<Option<Self>, ValidateContextError> {
```
arya2: This can return just the Option now.

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -480,17 +480,7 @@ impl NonFinalizedState {
...
-                                parent_hash,
-                                sprout_note_commitment_tree.clone(),
-                                sapling_note_commitment_tree.clone(),
-                                orchard_note_commitment_tree.clone(),
-                                history_tree.clone(),
-                            )
-                            .transpose()
-                    })
+                    .find_map(|chain| chain.fork(parent_hash).transpose())
                     .transpose()?
```
arya2: We can get rid of these two transpose calls if `fork` returns `Option<Self>`:

```suggestion
                    .find_map(|chain| chain.fork(parent_hash))
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -521,9 +460,111 @@ impl Chain {
...
+            self.sprout_trees_by_height.insert(height, tree.clone()),
+            None,
+            "incorrect overwrite of sprout tree: trees must be reverted then inserted",
+        );
+        #[cfg(not(test))]
+        assert_eq!(
+            self.sprout_anchors_by_height.insert(height, anchor),
+            None,
+            "incorrect overwrite of sprout anchor: anchors must be reverted then inserted",
+        );
```
arya2: Optional/Nitpick: group conditionally compiled statements in blocks?

```suggestion
        {
          assert_eq!(
              self.sprout_trees_by_height.insert(height, tree.clone()),
              None,
              "incorrect overwrite of sprout tree: trees must be reverted then inserted",
          );
          assert_eq!(
              self.sprout_anchors_by_height.insert(height, anchor),
              None,
              "incorrect overwrite of sprout anchor: anchors must be reverted then inserted",
          );
        }
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -547,6 +682,105 @@ impl Chain {
...
+    /// Returns the History tree of the tip of this [`Chain`],
+    /// including all finalized blocks, and the non-finalized blocks below the chain tip.
+    ///
+    /// If the chain is empty, instead returns the tree of the finalized tip,
+    /// which was supplied in [`Chain::new()`]
+    ///
+    /// # Panics
+    ///
+    /// If this chain has no history trees. (This should be impossible.)
+    pub fn history_tree_at_tip(&self) -> Arc<HistoryTree> {
```
arya2: Nitpick: rename this to `tip_history_tree` for consistency with the other method names.

arya2 (commented):
Looks good! I really like the simplified `fork()`/`parent_chain()` methods.

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: I think the most consistent name might be `history_block_commitment_tree`, to match `{sprout,sapling,orchard}_note_commitment_tree`?

We can do more renames of existing functions/fields later, I was hoping to keep this PR small.

zebra-chain/src/sapling/tree.rs
teor2345: Unfortunately that's a much larger change!

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: Thanks, fixed in 7ea4bf8db9

zebra-state/src/service/non_finalized_state.rs
teor2345: Thanks, fixed in 7ea4bf8db9

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: Changed in 6a26fe7ba7

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: Changed in 0e2335e836

arya2 (approved):
Thank you! 

(I left a do-not-merge label in case you'd like a second reviewer for the consensus rules)

teor2345:
> I am also running a partial and full sync locally on both mainnet and testnet.

The partial syncs on both networks seem normal, as does the full sync on testnet. I am still waiting for the full sync on mainnet to fully checkpoint, this is also normal with the latency to where I am.

Before the transaction spam we were getting about 0.3% of blocks with a chain fork on mainnet. So if there was anything seriously wrong with this PR, I'd expect my mainnet partial sync would have failed by now.

teor2345:
> (I left a do-not-merge label in case you'd like a second reviewer for the consensus rules)

I think @upbqdn or @conradoplg changed this code last, can one of you review this PR?

upbqdn:
I'm reviewing this PR.

upbqdn (approved):
This PR looks great. It was rewarding to see the gradual simplification in each commit.

PR #6131, opened 2023-02-09
Update copyright year in MIT License

sandakersmann:
Update copyright year in MIT License

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2019 Zcash Foundation
+Copyright (c) 2019-2023 Zcash Foundation
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated


LICENSE-MIT
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2019 Zcash Foundation
+Copyright (c) 2019-2023 Zcash Foundation
```
teor2345, line 1: This might be a legal change, so I'm going to check it internally before we approve it.

teor2345 (commented):
Thanks!

Can you please do the same thing for the Apache license?
https://github.com/ZcashFoundation/zebra/blob/main/LICENSE-APACHE

LICENSE-MIT
jackgavigan: This seems fine to me! 👍

oxarbitrage: I am not really sure how it works. Many projects seems to just use the year the license was added and assume the license is valid from that up to today ? The way this PR introduces seems to be a range which will force us to update it every year or we will be "unlicensed" while we don't?

daira: Specifying a date range for a copyright isn't legally necessary as far as I can tell. It's one of those things that people do because they copy what other people are doing.

(It used to be the case, before the US joined the Berne convention in 1989, that there was some benefit to explicitly claiming copyright in the US. Since then, copyright is established automatically in the US, just as was already the case in almost all other countries.  https://stackoverflow.com/a/28850974/393146 )

teor2345 (approved):
Thanks!

PR #6134, opened 2023-02-10
fix(rpc): Add `getblock` RPC fields to support the latest version of `zcash/lightwalletd`

### LINKED ISSUE #6085: Zebra returns "error: could not convert the input string to a hash or height" with zcash/lightwalletd
## Description

Using the normal LightwalletD branch https://github.com/zcash/lightwalletd.git returns an error on both zebra rc3 and rc4. The Zecwallet Lite version still functions.

### Analysis

This is likely a bug in the previous fix for this issue in PR #5861.

We might want to consider also adding a CI test or another integration or input snapshot test, because our manual testing didn't pick up this bug before we merged.

### Steps to Reproduce

I tried this:

Normal LightwalletD installation with the required 'zcash.conf' in the '/.config' directory. This is the error output

```sh
john@john-Latitude-5490:~/lightwalletd$ ./lightwalletd --no-tls-very-insecure --zcash-conf-path ~/.config/zcash.conf --data-dir . --log-file /dev/stdout
{"app":"lightwalletd","buildDate":"2023-02-01","buildUser":"john","gitCommit":"07fd6a966a136a751b37b8c54a6679499def0caa","level":"info","msg":"Starting gRPC server version v0.4.13-9-g07fd6a9 on 127.0.0.1:9067","time":"2023-02-01T14:00:09-07:00"}
{"app":"lightwalletd","level":"warning","msg":"Starting insecure no-TLS (plaintext) server","time":"2023-02-01T14:00:09-07:00"}
Starting insecure server
{"app":"lightwalletd","level":"info","msg":"Got sapling height 419200 block height 1969238 chain main branchID c2d6d0b4","time":"2023-02-01T14:00:09-07:00"}
{"app":"lightwalletd","level":"info","msg":"Found 0 blocks in cache","time":"2023-02-01T14:00:09-07:00"}
{"app":"lightwalletd","level":"info","msg":"getblock 419200 failed, will retry: error requesting block: 0: parse error: could not convert the input string to a hash or height","time":"2023-02-01T14:00:09-07:00"}
```

### Expected Behaviour

I expected to see this happen: LightwalletD server start and begin syncing.

### Actual Behaviour

Instead, this happened: retries 'getblock' and repeatedly fails

### Zebra Logs
The zebra terminal output is unaffected while LightwalletD repeats calls to getblock. The functional version does affect the zebra output mostly with mempool transaction calls but isn't relevant for this case. 

<details>

```

```

</details>

## Environment
Native, built from source

### Zebra Version
Both zebra 1.0.0-rc3 and rc4

### Operating System
Ubuntu 22 Jammy

Linux john-Latitude-5490 5.4.0-137-lowlatency #154-Ubuntu SMP PREEMPT Thu Jan 5 17:44:30 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

teor2345:
## Motivation

The latest version of `zcash/lightwalletd` expects the `getblock` response to always contain the block `hash` field:
https://github.com/zcash/lightwalletd/blob/5d174f7feb702dc19aec5b09d8be8b3d5b17ce45/common/common.go#L282

This PR gets the `hash` if needed, using a state index lookup.

Close #6085.

### Specifications

> "hash" : "hash",       (string) the block hash (same as provided hash)

https://zcash.github.io/rpc/getblock.html

The height to hash lookup isn't specified, but we can assume they want the hash at that height on the best chain.

For consistency even if the chain changes, we look up the block hash first, then use the hash to look up the transaction IDs and block confirmations.

## Solution

- Stabilise the BestChainBlockHash state request
    - previously, it was behind the `getblockchain-rpcs` feature, but we need it for the hash lookup in production RPCs
- Always include the block `hash` in the getblock RPC response
    - Since we always have the `hash`, it's easy to always get `confirmations` as well

Testing:
- Update existing test vectors to always require the `hash` and `confirmations` fields
- Update snapshots
- Update `lightwalletd` integration test to work with `zcash/lightwalletd`

I manually checked that `zcash/lightwalletd` works with Zebra's integration tests. (In CI we use `adityapk00/lightwalletd`.)

## Review

We might want to get this in the next release, but it's not a blocker.

### Reviewer Checklist

## Follow Up Work

We might want to test both `lightwalletd` forks in CI, to avoid issues like this in future.
@gustavovalverde can you put this in the CI review doc?

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -646,7 +646,7 @@ where
                             zebra_state::ReadResponse::BlockHash(Some(hash)) => hash,
                             zebra_state::ReadResponse::BlockHash(None) => {
                                 return Err(Error {
-                                    code: ErrorCode::ServerError(0),
+                                    code: MISSING_BLOCK_ERROR_CODE,
                                     message: "block height not in best chain".to_string(),
                                     data: None,
                                 })
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -613,24 +613,62 @@ where
                 // This RPC is used in `lightwalletd`'s initial sync of 2 million blocks,
                 // so it needs to load all its fields very efficiently.
                 //
-                // Currently, we get the transaction IDs from an index, which is much more
-                // efficient than loading all the block data and hashing all the transactions.
+                // Currently, we get the block hash and transaction IDs from indexes,
+                // which is much more efficient than loading all the block data,
+                // then hashing the block header and all the transactions.
 
-                // TODO: look up the hash if we only have a height,
-                //       and look up the height if we only have a hash
-                let hash = hash_or_height.hash().map(GetBlockHash);
-                let height = hash_or_height.height();
-
-                // TODO: do these state  queries in parallel?
-
-                // Get transaction IDs from the transaction index
+                // Get the block hash from the height -> hash index, if needed
                 //
                 // # Concurrency
+                //
+                // For consistency, this lookup must be performed first, then all the other
+                // lookups must be based on the hash.
+                //
+                // All possible responses are valid, even if the best chain changes. Clients
+                // must be able to handle chain forks, including a hash for a block that is
+                // later discovered to be on a side chain.
+
+                let hash = match hash_or_height {
+                    HashOrHeight::Hash(hash) => hash,
+                    HashOrHeight::Height(height) => {
+                        let request = zebra_state::ReadRequest::BestChainBlockHash(height);
+                        let response = state
+                            .ready()
+                            .and_then(|service| service.call(request))
+                            .await
+                            .map_err(|error| Error {
+                                code: ErrorCode::ServerError(0),
+                                message: error.to_string(),
+                                data: None,
+                            })?;
+
+                        match response {
+                            zebra_state::ReadResponse::BlockHash(Some(hash)) => hash,
+                            zebra_state::ReadResponse::BlockHash(None) => {
+                                return Err(Error {
+                                    code: ErrorCode::ServerError(0),
+                                    message: "block height not in best chain".to_string(),
+                                    data: None,
+                                })
+                            }
+                            _ => unreachable!("unmatched response to a block hash request"),
+                        }
+                    }
+                };
+
+                // TODO: do the txids and confirmations state queries in parallel?
+
+                // Get transaction IDs from the transaction index by block hash
+                //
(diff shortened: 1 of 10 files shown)


zebra-state/src/request.rs
```diff
@@ -53,6 +53,23 @@ impl HashOrHeight {
...
     }
 
+    /// Unwrap the inner hash or attempt to retrieve the hash for a given
+    /// height if one exists.
+    ///
+    /// # Consensus
+    ///
+    /// In the non-finalized state, a height can have multiple valid hashes.
+    /// We typically use the hash that is currently on the best chain.
+    pub fn hash_or_else<F>(self, op: F) -> Option<block::Hash>
```
arya2, line 63: Could this be removed if it's unused?

arya2 (approved):
Looks good to me! Thank you for fixing this.

Is the 'A-consensus' label correct? (is it for the `confirmations` field?)

zebra-rpc/src/methods.rs
```diff
@@ -613,24 +613,62 @@ where
...
+                // # Concurrency
+                //
+                // For consistency, this lookup must be performed first, then all the other
+                // lookups must be based on the hash.
+                //
+                // All possible responses are valid, even if the best chain changes. Clients
+                // must be able to handle chain forks, including a hash for a block that is
+                // later discovered to be on a side chain.
+
+                let hash = match hash_or_height {
```
arya2, line 631: Optional/Nitpick: use or remove the `HashOrHeight.hash()` method.

teor2345: We can't use that method here, because this match statement is async and uses the `?` return operator.

zebra-state/src/request.rs
teor2345: We already have `height_or_else()`, and I think we'll end up using `hash_or_else()` some time pretty soon.

I almost used it in this PR, but `Fn` generics can't accept async closures.

teor2345:
This blocks our next Zebra release because we want to be able to support this bug fix in `lightwalletd`:
https://github.com/adityapk00/lightwalletd/issues/13

For the Zebra team: this is an example of a race condition where using a hash rather than a height matters.

teor2345:
> Is the 'A-consensus' label correct? (is it for the `confirmations` field?)

I could go either way, it fixes a consensus bug in `lightwalletd`, but it's not really consensus-critical in Zebra. It's just re-using existing consensus rule implementations.

Feel free to take it off if you'd like?

PR #6139, opened 2023-02-10
cleanup(consensus): Remove unused `impl ZcashDeserialize for Height`

teor2345:
## Motivation

There are two different ways to serialize heights in Zcash, so we should use the existing named methods for that.

Having an unused `impl ZcashDeserialize for Height` risks us using it in the wrong context, or it could become outdated.

### Specifications

Heights can be serialized as a coinbase script, or as a signed or unsigned 32-bit integer.

## Review

This is a low priority consensus rule cleanup.

### Reviewer Checklist


arya2 (approved):
Thank you!

PR #6154, opened 2023-02-14
fix(mempool): Re-verify transactions that were verified at a different tip height

### LINKED ISSUE #6132: Fix "duplicate transparent spend" caused by mempool validation race condition
## Motivation

In an unrelated PR, block proposal validation failed with:
> Message:  block proposal validation failed: 
> 0: unsuccessful block proposal validation, reason: "block-source-validateproposal-commitblockerror-duplicatetransparentspend-outpoint-outpoint-hash-transaction-hash-20d2422d07a5b168e9a6d7d60f8d7eadc89f270ccce6ca984081d1f4c4258e07-index-111-location-the-non-finalized-chain"

https://github.com/ZcashFoundation/zebra/actions/runs/4131129255/jobs/7138887361#step:8:210

### Specifications

The consensus rules say that each transparent output must only be spent once in each chain. This should already be implemented in the mempool.

This bug needs further investigation: we won't know how complex it is until we know why.

### Testing

We might want to add a unit test for this rare scenario, because the existing unit and integration tests aren't catching it.

## Related Work

This likely blocks the next Zebra release, because we can't have miners test on `mainnet` with this bug.

arya2:
## Motivation

We want to ensure that transactions in the mempool will be valid in the next block at the end of `poll_ready()`.

Closes #6132 (it's hard to be entirely certain that this bug caused that failure but I don't see what else could've caused it.)

## Solution

- Returns the tip height at which a mempool transaction is verified
- Checks that the tip height hasn't changed before inserting the verified transaction into the mempool

## Review

Anyone can review, we may want 2 reviewers for the concurrency.

### Reviewer Checklist

## Follow Up Work

- See #6173

### DIFF
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -372,6 +372,8 @@ impl Service<Request> for Mempool {
             // Collect inserted transaction ids.
             let mut send_to_peers_ids = HashSet::<_>::new();
 
+            let best_tip_height = self.latest_chain_tip.best_tip_height();
+
             // Clean up completed download tasks and add to mempool if successful.
             while let Poll::Ready(Some(r)) = tx_downloads.as_mut().poll_next(cx) {
                 match r {
@@ -382,7 +384,7 @@ impl Service<Request> for Mempool {
                         // chain_tip_change.last_tip_change() returns a `TipAction::Reset` when
                         // the previous block hash doesn't match the `last_change_hash` and the
                         // mempool re-verifies all pending tx_downloads when there's a `TipAction::Reset`.
-                        if self.latest_chain_tip.best_tip_height() == Some(expected_tip_height) {
+                        if best_tip_height == Some(expected_tip_height) {
                             let insert_result = storage.insert(tx.clone());
 
                             tracing::trace!(
@@ -430,7 +432,7 @@ impl Service<Request> for Mempool {
             //
             // Lock times never expire, because block times are strictly increasing.
             // So we don't need to check them here.
-            if let Some(tip_height) = self.latest_chain_tip.best_tip_height() {
+            if let Some(tip_height) = best_tip_height {
                 let expired_transactions = storage.remove_expired_transactions(tip_height);
                 // Remove transactions that are expired from the peers list
                 send_to_peers_ids =
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -382,7 +382,7 @@ impl Service<Request> for Mempool {
                         //
                         // It's okay to use tip height here instead of the tip hash since
                         // chain_tip_change.last_tip_change() returns a `TipAction::Reset` when
-                        // the previous block hash doesn't match the `last_change_hash` and the
+                        // the best chain changes (which is the only way to stay at the same height), and the
                         // mempool re-verifies all pending tx_downloads when there's a `TipAction::Reset`.
                         if best_tip_height == Some(expected_tip_height) {
                             let insert_result = storage.insert(tx.clone());
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -376,6 +376,12 @@ impl Service<Request> for Mempool {
             while let Poll::Ready(Some(r)) = tx_downloads.as_mut().poll_next(cx) {
                 match r {
                     Ok((tx, expected_tip_height)) => {
+                        // # Correctness:
+                        //
+                        // It's okay to use tip height here instead of the tip hash since
+                        // chain_tip_change.last_tip_change() returns a `TipAction::Reset` when
+                        // the previous block hash doesn't match the `last_change_hash` and the
+                        // mempool re-verifies all pending tx_downloads when there's a `TipAction::Reset`.
                         if self.latest_chain_tip.best_tip_height() == Some(expected_tip_height) {
                             let insert_result = storage.insert(tx.clone());
 
@@ -447,9 +453,6 @@ impl Service<Request> for Mempool {
         }
 
         // TODO: Move the above into a loop, check that the tip action hasn't changed before returning
-        //       and leave a note about correctness for using tip_height instead of tip_hash
-        //       to check that pending verifications
-        //       are still valid .
 
         Poll::Ready(Ok(()))
     }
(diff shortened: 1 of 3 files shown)


arya2:
I'm not sure what happened here:

> ---- components::mempool::tests::vector::mempool_reverifies_after_tip_change stdout ----
>
> Message:  unexpected failure when checking for verified transactions: SendError({Legacy(transaction::Hash("4e3670eae21d0c7225716abacb3a95c203a75d7601594afddb07d175c3babe83"))})
> Location: zebrad/src/components/mempool/tests.rs:53

https://github.com/ZcashFoundation/zebra/actions/runs/4169566277/jobs/7217667829#step:14:3765

The tests passed locally, and that request should always returns `Ok(_)`; something must've gone wrong in the `poll_ready()` call.

zebrad/src/components/mempool/downloads.rs
```diff
@@ -282,12 +284,12 @@ where
 
             trace!(?txid, "transaction is not in best chain");
 
-            let next_height = match state.oneshot(zs::Request::Tip).await {
-                Ok(zs::Response::Tip(None)) => Ok(Height(0)),
+            let (tip_height, next_height) = match state.oneshot(zs::Request::Tip).await {
+                Ok(zs::Response::Tip(None)) => Ok((Height(0), Height(1))),
```
arya2: Should `tip_height` be an `Option` here? i.e.
```rust
Ok(zs::Response::Tip(None)) => Ok((None, Height(0))),
```
I'm open to suggestions here (though I don't think this state request will ever really return `None` anyways).

zebrad/src/components/mempool.rs
```diff
@@ -372,20 +372,36 @@ impl Service<Request> for Mempool {
...
-
-                        if let Ok(inserted_id) = insert_result {
-                            // Save transaction ids that we will send to peers
-                            send_to_peers_ids.insert(inserted_id);
+                    Ok((tx, expected_tip_height)) => {
+                        // # Correctness:
+                        //
+                        // It's okay to use tip height here instead of the tip hash since
+                        // chain_tip_change.last_tip_change() returns a `TipAction::Reset` when
+                        // the previous block hash doesn't match the `last_change_hash` and the
```
teor2345: Nitpick: it might help to explain the consensus rule, rather than the current implementation of it

```suggestion
                        // chain_tip_change.last_tip_change() returns a `TipAction::Reset` when
                        // the best chain changes (which is the only way to stay at the same height), and the
```

zebrad/src/components/mempool/downloads.rs
teor2345: Yes, the correct implementation has `tip_height: Option<Height>`, because:
- `None` is an empty state
- `Height(0)` is a state containing only the genesis block

I don't think `next_height == 1` happens in production, but I think it might be required for some tests. If it isn't required for any tests, it would also be ok to return an error for an empty state.

I don't think we can panic, because we have tests that activate the mempool on startup. But feel free to try it!

teor2345 (dismissed):
Looks good!

I think the test is good, and if it runs correctly a few times, we can merge this.

zebrad/src/components/mempool.rs
```diff
@@ -438,6 +454,8 @@ impl Service<Request> for Mempool {
             }
```
arya2: The test failure is from [this line above](https://github.com/ZcashFoundation/zebra/pull/6154/files#diff-ec634b649fdc36d20e3746cb9106eefafff1315ce871dd9b16fbbe5708d85784R453):
```rust
self.transaction_sender.send(send_to_peers_ids)?;
```

zebrad/src/components/mempool/downloads.rs
arya2: Fixed in [`99fa95e`](https://github.com/ZcashFoundation/zebra/pull/6154/commits/99fa95ebfdd2f468b984e0d9e166af0f4b148fb6).

zebrad/src/components/mempool.rs
arya2: Fixed in [`4b85ddc`](https://github.com/ZcashFoundation/zebra/pull/6154/commits/4b85ddc853d06559aa6f7a4438c8768ee249b94e).

zebrad/src/components/mempool/tests/vector.rs
```diff
@@ -794,6 +795,199 @@ async fn mempool_failed_download_is_not_rejected() -> Result<(), Report> {
...
+    )
+    .await
+    .map(|change_result| change_result.expect("unexpected chain tip update failure"))
+    {
+        info!(
+            timeout = ?humantime_seconds(CHAIN_TIP_UPDATE_WAIT_LIMIT),
+            ?timeout_error,
+            "timeout waiting for chain tip change after committing block"
+        );
+    }
```
arya2: Actually it matters here too.

```suggestion
    // Wait for the chain tip update without a timeout
    // (skipping the chain tip change here may cause the test to 
    // pass without reverifying transactions for  `TipAction::Grow`)
    chain_tip_change
        .wait_for_tip_change()
        .await
        .expect("unexpected chain tip update failure");
```

zebrad/src/components/mempool/tests/vector.rs
```diff
@@ -794,6 +795,192 @@ async fn mempool_failed_download_is_not_rejected() -> Result<(), Report> {
...
+            genesis_block.clone().into(),
+        ))
+        .await
+        .unwrap();
+
+    // Wait for the chain tip update without a timeout
+    // (skipping the chain tip change here may cause the test to 
+    // pass without reverifying transactions for  `TipAction::Grow`)
+    chain_tip_change
+        .wait_for_tip_change()
```
arya2, line 845: Should the `wait_for_tip_change()` calls in this test panic with a longer timeout instead?

teor2345 (approved):
These all look like reasonable changes, happy to merge if this passes PR CI and merge queue CI reliably.

arya2:
Failed to fetch zcash parameters in the merge queue:

> thread 'main' panicked at 'error downloading Sprout parameter files after 3 retries. IoError(Custom { kind: InvalidData, error: "sprout-groth16.params failed validation:\nexpected: e9b238411bd6c0ec4791e9d04245ec350c9c5744f5610dfcce4365d5ca49dfefd5054e371842b3f88fa1b9d7e8e075249b3ebabd167fa8b0f3161292d36c180a hashing 725523612 bytes,\nactual:   

https://github.com/ZcashFoundation/zebra/actions/runs/4192757597/jobs/7268756930#step:9:18

arya2:
@Mergifyio refresh

PR #6155, opened 2023-02-14
fix(mempool): Process chain tip actions that occur during poll_ready() execution before returning

arya2:
## Motivation

We want to ensure that transactions in the mempool will be valid in the next block at the end of poll_ready().

## Solution

- Processes new tip actions if the best chain tip changes during `poll_ready` execution .

## Review

This PR was split out of https://github.com/ZcashFoundation/zebra/pull/6154 for readability.

The diff is messy but it's a minor change.

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -317,7 +317,7 @@ impl Service<Request> for Mempool {
             return Poll::Ready(Ok(()));
         }
 
-        let tip_action = self.chain_tip_change.last_tip_change();
+        let mut tip_action = self.chain_tip_change.last_tip_change();
 
         // If the mempool was just freshly enabled,
         // skip resetting and removing mined transactions for this tip.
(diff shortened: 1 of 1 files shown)


zebrad/src/components/mempool.rs
```diff
@@ -317,144 +317,152 @@ impl Service<Request> for Mempool {
...
 
-            // Send transactions that were not rejected nor expired to peers
-            if !send_to_peers_ids.is_empty() {
-                tracing::trace!(?send_to_peers_ids, "sending new transactions to peers");
+            tip_action = self.chain_tip_change.last_tip_change();
 
-                self.transaction_sender.send(send_to_peers_ids)?;
+            if tip_action.is_none() {
+                break;
             }
```
teor2345, line 457: Nitpick: this loop might be easier to understand as:

```rust
while let Some(tip_action) = self.chain_tip_change.last_tip_change() {
    ...
}
```

teor2345 (commented):
This looks good, and I think it might avoid a whole class of bugs where mempool requests are processed while the mempool is out of sync with the state. But I'd like to be clear about what those bugs actually are.

zebrad/src/components/mempool.rs
```diff
@@ -317,144 +317,152 @@ impl Service<Request> for Mempool {
...
+        // Collect inserted transaction ids.
+        let mut send_to_peers_ids = HashSet::<_>::new();
 
-            // Re-initialise an empty state.
-            self.update_state();
-
-            // Re-verify the transactions that were pending or valid at the previous tip.
-            // This saves us the time and data needed to re-download them.
-            if let ActiveState::Enabled { tx_downloads, .. } = &mut self.active_state {
+        loop {
```
teor2345, line 331: I think this is a good change that will keep the mempool in sync with the state.

But I'm not sure if desync bugs can actually happen in practice. `poll_ready()` happens before every `call()`, so I don't know if it is possible to observe the mempool being out of sync with the state?

I also wonder what happens if blocks arrive in the state at just the right rate to make the loop infinite? (If they are too slow, the loop exits, and if they are too fast, they are combined and create a Reset.)

Could you add a comment explaining:
- why this loop is needed, ideally with a short description of a specific example
- how we avoid livelock if blocks are arriving in the state at a rate that matches the loop
- that `self.chain_tip_change` only ever contains a single action (it's not a queue, but the loop makes it look like one)

arya2: > I don't know if it is possible to observe the mempool being out of sync with the state?

The example I had in mind was:
- `poll_ready()` is called shortly after a chain tip change
- chain tip changes again at the same time as the `poll_ready()` call

It looks like it's called every ~5 seconds here: https://github.com/ZcashFoundation/zebra/blob/recheck-last-tip-change/zebrad/src/components/mempool/queue_checker.rs#L52-L63

I think that another call that requests transactions *could* coincide with the *next* chain tip change, albeit very very rarely.

> I also wonder what happens if blocks arrive in the state at just the right rate to make the loop infinite? (If they are too slow, the loop exits, and if they are too fast, they are combined and create a Reset.)

I think block commits should be too slow for a third iteration. We could limit it to a maximum number of iterations and add a field for hiding the mempool storage in requests until it's back in sync if `poll_ready()` is a lot slower than I'm imagining, i.e. return no transactions if the chain tip changes >2 times during a call to `poll_ready()` so that it doesn't block template generation.

A comment would be good, I'll add that.

teor2345: > > I don't know if it is possible to observe the mempool being out of sync with the state?
> 
> The example I had in mind was:
>     * `poll_ready()` is called shortly after a chain tip change
>     * chain tip changes again at the same time as the `poll_ready()` call

Sorry, I'm having a bit of trouble understanding the concurrency issues here, and what bugs they cause. How can mempool callers see this happening? What's the impact?

I guess I'm looking for significant differences from existing concurrency behaviour. For example, the chain tip can change after the response has been created, but before it is received by the caller. This is an unavoidable issue in concurrent async code, so callers have to be able to handle it.

It also feels a bit like we are repeating similar comments in this thread. Did you want to jump on a call some time to talk it over?

> I think block commits should be too slow for a third iteration.

This is currently true, but it isn't guaranteed to always happen. For example, small transparent-only blocks commit very quickly.

> We could limit it to a maximum number of iterations and add a field for hiding the mempool storage in requests until it's back in sync if `poll_ready()` is a lot slower than I'm imagining, i.e. return no transactions if the chain tip changes >2 times during a call to `poll_ready()` so that it doesn't block template generation.

This loop and its edge cases seem a bit complicated. That makes them tricky to reason about, and harder to test. Instead, can we reset the mempool if there are multiple chain tip changes?

(That's what we'd do if the blocks arrived a bit closer together, they'd be counted as a multi-block Reset. It's also a well-tested and easy to reason about operation.)

arya2: > How can mempool callers see this happening? What's the impact?

The `FullTransactions` request could return transactions that were already mined in the `get_block_template` method when called without a `longpollid`, but then it might not match the `GetBlockTemplateChainInfo` response.

> the chain tip can change after the response has been created, but before it is received by the caller.

That's a good point, I think we can close this PR and check that the tip hasn't changed while waiting for `fetch_mempool_transactions()` in the `get_block_template` method (occasionally advertising transaction ids that will be queued for re-verification in the next `poll_ready()` call shouldn't be an issue).

> Did you want to jump on a call some time to talk it over?

Sure!

arya2:
Checking for tip changes during the `poll_ready()` execution won't prevent seeing transactions that are invalidated as of the latest chain tip and wouldn't match the `GetBlockTemplateChainInfo` response if it did. 

We could check that the latest tip height seen by `poll_ready()` matches the `GetBlockTemplateChainInfo` response to avoid excluding transactions when the best tip has changed but there are transactions in the mempool that match the `GetBlockTemplateChainInfo`. I'll check with Teor to see if that's worth doing (it would only be relevant for submitting a valid block that forks the chain).

zebrad/src/components/mempool.rs
teor2345: > Instead, can we reset the mempool if there are multiple chain tip changes?

Just a reminder to myself that this would be a small alternative fix.

PR #6176, opened 2023-02-16
fix(ci): Increase macOS CI timeout

teor2345:
## Motivation

When macOS refreshes the Zcash parameter cache, its CI sometimes takes longer than 165 minutes:
https://github.com/ZcashFoundation/zebra/actions/runs/4188656916/jobs/7260060480

(This one timed out in the last step, so it appears successful, but fails the PR.)

## Review

This might be urgent if the cache was not actually written, because the next PR will do the same thing.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -61,9 +61,9 @@ jobs:
   test:
     name: Test ${{ matrix.rust }} on ${{ matrix.os }}${{ matrix.features }}
     # The large timeout is to accommodate:
-    # - Windows builds (75 minutes, typically 30-50 minutes), and
-    # - parameter downloads (an extra 90 minutes, but only when the cache expires)
-    timeout-minutes: 165
+    # - macOS and Windows builds (90 minutes, typically 30-70 minutes), and
+    # - parameter downloads (an extra 100 minutes, but only when the cache expires)
+    timeout-minutes: 190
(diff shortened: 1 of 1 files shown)


.github/workflows/continous-integration-os.yml
```diff
@@ -61,9 +61,9 @@ jobs:
   test:
     name: Test ${{ matrix.rust }} on ${{ matrix.os }}${{ matrix.features }}
     # The large timeout is to accommodate:
-    # - Windows builds (75 minutes, typically 30-50 minutes), and
-    # - parameter downloads (an extra 90 minutes, but only when the cache expires)
-    timeout-minutes: 165
+    # - macOS and Windows builds (90 minutes, typically 30-70 minutes), and
+    # - parameter downloads (an extra 100 minutes, but only when the cache expires)
+    timeout-minutes: 190
```
arya2, line 66: Why not the default 6hrs?

arya2 (approved):
Looks good!

.github/workflows/continous-integration-os.yml
teor2345: Because if our builds suddenly jump from 3 to 6 hours, we likely have a compile or production performance bug, and we shouldn't merge that PR.

We don't do any other performance monitoring, so I don't want to increase these timeouts too much.

PR #6171, opened 2023-02-15
feat(rpc): Implement `z_listunifiedreceivers`

oxarbitrage:
## Motivation

We want to have `z_listunifiedreceivers` as part of the mining pool efforts.

Close https://github.com/ZcashFoundation/zebra/issues/6030

## Solution

Add support for `z_listunifiedreceivers`

## Review

Anyone can review, there are some parts of the code that can be improved or refactored. Diff tests seems to work with several test vectors taken from zcash.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #6185, opened 2023-02-17
feat(rpc): Implement the `z_validateaddress` RPC

### LINKED ISSUE #6083: Implement z_validateaddress RPC
## Motivation

Some mining pools use this RPC to check addresses.

### Specifications

https://zcash.github.io/rpc/z_validateaddress.html

### Designs

We can call the [`zcash_address`](https://docs.rs/zcash_address/latest/zcash_address/) crate to validate shielded and unified addresses. (Zebra also has its own transparent address parsing, but we'd be better to just be consistent and use `zcash_address` for everything.)

We just need to implement these fields:
- isvalid
- address
- type
- address_type (same as type)
- ismine - always false, because Zebra doesn't have a wallet

The `Address` enum added in [#6086](https://github.com/ZcashFoundation/zebra/pull/6086/files#diff-7f5d342981d90d652271da72fb36b58d1af35b1f92431d3dcc79a1113d0ad957) can be extended to convert shielded addresses as well for implementing this RPC.

upbqdn:
## Motivation

Close #6083.

### Specifications

https://zcash.github.io/rpc/z_validateaddress.html

## Solution

This PR implements the following fields:
- `isvalid`,
- `address`,
- `address_type`,
- `is_mine`.

To implement UAs, I used the `zcash_address` crate.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/primitives/address.rs
+++ b/zebra-chain/src/primitives/address.rs
@@ -7,7 +7,6 @@ use zcash_primitives::sapling;
 use crate::{parameters::Network, transparent, BoxError};
 
 /// Zcash address variants
-// TODO: Add Sprout addresses
 pub enum Address {
     /// Transparent address
     Transparent(transparent::Address),
--- a/zebra-chain/src/primitives/address.rs
+++ b/zebra-chain/src/primitives/address.rs
@@ -4,7 +4,7 @@
 
 use zcash_primitives::sapling;
 
-use crate::{orchard, parameters::Network, transparent, BoxError};
+use crate::{parameters::Network, transparent, BoxError};
 
 /// Zcash address variants
 // TODO: Add Sprout addresses
@@ -26,14 +26,8 @@ pub enum Address {
         /// Address' network
         network: Network,
 
-        /// Transparent address
-        transparent_address: transparent::Address,
-
-        /// Sapling address
-        sapling_address: sapling::PaymentAddress,
-
-        /// Orchard address
-        orchard_address: orchard::Address,
+        /// Unified address
+        address: zcash_address::unified::Address,
     },
 }
 
@@ -93,7 +87,13 @@ impl zcash_address::TryFromAddress for Address {
             .ok_or_else(|| BoxError::from("not a valid sapling address").into())
     }
 
-    // TODO: Add sprout and unified/orchard converters
+    fn try_from_unified(
+        network: zcash_address::Network,
+        address: zcash_address::unified::Address,
+    ) -> Result<Self, zcash_address::ConversionError<Self::Error>> {
+        let network = network.try_into()?;
+        Ok(Self::Unified { network, address })
+    }
 }
 
 impl Address {
--- a/zebra-chain/src/primitives/address.rs
+++ b/zebra-chain/src/primitives/address.rs
@@ -2,6 +2,7 @@
 //!
 //! Usage: <https://docs.rs/zcash_address/0.2.0/zcash_address/trait.TryFromAddress.html#examples>
 
+use zcash_address::unified::{self, Container};
 use zcash_primitives::sapling;
 
 use crate::{parameters::Network, transparent, BoxError};
@@ -27,6 +28,15 @@ pub enum Address {
 
         /// Unified address
         address: zcash_address::unified::Address,
+
+        /// Orchard address
+        orchard: Option<orchard::Address>,
+
+        /// Sapling address
+        sapling: Option<sapling::PaymentAddress>,
+
+        /// Transparent address
+        transparent: Option<transparent::Address>,
     },
 }
 
@@ -91,7 +101,55 @@ impl zcash_address::TryFromAddress for Address {
         address: zcash_address::unified::Address,
     ) -> Result<Self, zcash_address::ConversionError<Self::Error>> {
         let network = network.try_into()?;
-        Ok(Self::Unified { network, address })
+        let mut orchard = None;
+        let mut sapling = None;
+        let mut transparent = None;
+
+        for receiver in address.items().into_iter() {
+            match receiver {
+                unified::Receiver::Orchard(data) => {
+                    orchard = orchard::Address::from_raw_address_bytes(&data).into();
+                    // ZIP 316: Consumers MUST reject Unified Addresses/Viewing Keys in
+                    // which any constituent Item does not meet the validation
+                    // requirements of its encoding.
+                    if orchard.is_none() {
+                        return Err(BoxError::from(
+                            "Unified Address contains an invalid Orchard receiver.",
+                        )
+                        .into());
+                    }
+                }
+                unified::Receiver::Sapling(data) => {
+                    sapling = sapling::PaymentAddress::from_bytes(&data);
+                    // ZIP 316: Consumers MUST reject Unified Addresses/Viewing Keys in
+                    // which any constituent Item does not meet the validation
+                    // requirements of its encoding.
+                    if sapling.is_none() {
+                        return Err(BoxError::from(
+                            "Unified Address contains an invalid Sapling receiver",
+                        )
+                        .into());
+                    }
+                }
+                unified::Receiver::P2pkh(data) => {
+                    transparent = Some(transparent::Address::from_pub_key_hash(network, data));
+                }
+                unified::Receiver::P2sh(data) => {
+                    transparent = Some(transparent::Address::from_script_hash(network, data));
+                }
+                unified::Receiver::Unknown { .. } => {
+                    return Err(BoxError::from("Unsupported receiver in a Unified Address.").into());
+                }
+            }
+        }
+
+        Ok(Self::Unified {
+            network,
+            address,
+            orchard,
+            sapling,
+            transparent,
+        })
     }
 }
 
--- a/zebra-chain/src/primitives/address.rs
+++ b/zebra-chain/src/primitives/address.rs
@@ -176,4 +176,19 @@ impl Address {
     pub fn is_transparent(&self) -> bool {
         matches!(self, Self::Transparent(_))
     }
+
+    /// Returns the payment address for transparent or sapling addresses.
+    pub fn payment_address(&self) -> Option<String> {
+        use zcash_address::{ToAddress, ZcashAddress};
+
+        match &self {
+            Self::Transparent(address) => Some(address.to_string()),
+            Self::Sapling { address, network } => {
+                let data = address.to_bytes();
+                let address = ZcashAddress::from_sapling((*network).into(), data);
+                Some(address.encode())
+            }
+            Self::Unified { .. } => None,
+        }
+    }
 }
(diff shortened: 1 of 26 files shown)


upbqdn:
<details>
<summary>An initial basic test shows that the output is identical to zcashd.</summary>

```
Request:
z_validateaddress t2UNzUUx8mWBCRYPRezvA363EYXyEpHokyi

Querying zebrad test chain at height >=2232411...

real    0m0.002s
user    0m0.002s
sys     0m0.000s

Querying zcashd test chain at height >=2232411...

real    0m0.002s
user    0m0.002s
sys     0m0.000s


RPC responses were identical

/tmp/tmp.ESy1lXAX0v.rpc-diff/zebrad-test-2232411-z_validateaddress.json:
{
  "isvalid": true,
  "address": "t2UNzUUx8mWBCRYPRezvA363EYXyEpHokyi",
  "address_type": "p2sh",
  "ismine": false
}
```

</details>

zebra-chain/src/primitives/address.rs
```diff
@@ -26,14 +25,8 @@ pub enum Address {
...
-        /// Transparent address
-        transparent_address: transparent::Address,
-
-        /// Sapling address
-        sapling_address: sapling::PaymentAddress,
-
-        /// Orchard address
-        orchard_address: orchard::Address,
+        /// Unified address
+        address: zcash_address::unified::Address,
```
teor2345: This seems ok for now, but I think PR #6171 will want to expand it back out again. (The conversion from an address list to fields is already in that PR, you don't need to include it here.)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -53,6 +53,8 @@ use crate::methods::{
     height_from_signed_int, GetBlockHash, MISSING_BLOCK_ERROR_CODE,
 };
 
+use self::types::z_validate_address::{self, AddressType};
```
teor2345: Optional Nitpick: please move this to `crate::methods::types::` above.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -851,6 +863,49 @@ where
...
+                    is_mine: Some(false),
+                })
+            } else {
+                tracing::info!(
+                    ?network,
+                    address_network = ?address.network(),
+                    "invalid address network in z_validateaddress RPC: address is for {:?} but Zebra is on {:?}",
+                    address.network(),
+                    network
+                );
```
teor2345, line 908: Optional: does `zcashd` also do this check?
I think it does, but I'd like to double-check, and maybe add a comment to this RPC and `validateaddress`.

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -499,6 +517,17 @@ fn snapshot_rpc_validateaddress(
...
+/// Snapshot `z_validateaddress` response, using `cargo insta` and JSON serialization.
+fn snapshot_rpc_z_validateaddress(
+    variant: &'static str,
+    z_validate_address: z_validate_address::Response,
+    settings: &insta::Settings,
+) {
+    settings.bind(|| {
+        insta::assert_json_snapshot!(format!("z_validate_address_{variant}"), z_validate_address)
+    });
+}
```
teor2345, line 528: To pass CI, this snapshot test will need new snapshot files:
https://github.com/ZcashFoundation/zebra/blob/0ee67d5d7cbcf1107a0ae94e212ccfac8f27b50d/zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs#L3-L6

teor2345 (dismissed):
This looks good, thanks for the address API changes.

We're just missing the snapshot files, so this PR will fail CI until they are added.

zebra-chain/src/primitives/address.rs
teor2345: Actually, it might be easier doing it this way.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: > does zcashd also do this check?

Yeah, it's here for unified addresses: https://github.com/zcash/zcash/blob/master/src/rust/src/address_ffi.rs#L71.

It checks the network in the prefix for transparent addresses via the keyConstants instance here: https://github.com/zcash/zcash/blob/master/src/key_io.cpp#L273 and does the same for Sapling/Sprout.

zebra-chain/src/primitives/address.rs
arya2: I think we'll need to convert the addresses to match zcashd's behaviour if we want to support unified addresses later, it should return `isvalid: false` if any conversions fail (see https://github.com/zcash/zcash/blob/master/src/rpc/misc.cpp#L324 https://github.com/zcash/zcash/blob/master/src/zcash/Address.cpp#L80-L93 https://github.com/zcash/zcash/blob/master/src/rust/src/address_ffi.rs#L148-L152 and https://github.com/zcash/zcash/blob/master/src/rust/src/address_ffi.rs#L89).

upbqdn: I've been trying to find suitable UAs that would fail the check, but I couldn't find any. I have only valid UAs. :/

arya2: I'm not sure if it's necessary to do it for this PR but I think we could use [`try_from_items`](https://docs.rs/zcash_address/latest/zcash_address/unified/trait.Encoding.html#method.try_from_items) for any random "address" values/items with the correct length and encoding.

zebra-rpc/src/methods/tests/snapshot/snapshots/z_validate_address_basic@mainnet_10.snap.new
```diff
@@ -0,0 +1,11 @@
+---
```
teor2345: These snapshot files need a different name, there's two ways to fix this:
1. Take the ".new" off the end manually
2. Run the full `cargo insta` command again, and accept the changes interactively at the end (it will do the renames for you)

zebra-chain/src/primitives/address.rs
teor2345: This PR is on the critical path for miner testing, so I'd like to get it merged soon.
Can we open a separate ticket for the unified address change?

Here's a previous conversation where Pili suggests we split out unified address support entirely:
https://github.com/ZcashFoundation/zebra/issues/6083#issuecomment-1432260991

zebra-rpc/src/methods/tests/snapshot/snapshots/z_validate_address_basic@mainnet_10.snap.new
upbqdn: Hm, I ran `cargo insta test --features getblocktemplate-rpcs`, but the filenames didn't change.

Edit: nvm, I needed to run `cargo insta review`.

zebra-rpc/src/methods/get_block_template_rpcs.rs
upbqdn: Done in [1cec031](https://github.com/ZcashFoundation/zebra/pull/6185/commits/1cec0319dcbc61b264278e74757fe2d137b19a52).

zebra-rpc/src/methods/tests/snapshot/snapshots/z_validate_address_basic@mainnet_10.snap.new
upbqdn: Done in [c984d4d](https://github.com/ZcashFoundation/zebra/pull/6185/commits/c984d4d4c506558665b7e40a14d33b309d573bf3).

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
upbqdn: Done in [1c14599](https://github.com/ZcashFoundation/zebra/pull/6185/commits/1c145991e7354f25129fa32617c691dba118ddcd).

zebra-chain/src/primitives/address.rs
upbqdn: A suitable UA support is now implemented in [d47abd1](https://github.com/ZcashFoundation/zebra/pull/6185/commits/d47abd133fcda832b16999ca929951934e4192e9).

upbqdn:
I implemented all suggestions, semantic UA checks, and I updated the PR description to reflect the new changes.

zebra-chain/src/primitives/address.rs
```diff
@@ -93,7 +96,61 @@ impl zcash_address::TryFromAddress for Address {
...
+                    }
+                }
+                unified::Receiver::P2pkh(data) => {
+                    transparent = Some(transparent::Address::from_pub_key_hash(network, data));
+                }
+                unified::Receiver::P2sh(data) => {
+                    transparent = Some(transparent::Address::from_script_hash(network, data));
+                }
+                unified::Receiver::Unknown { .. } => {
+                    return Err(BoxError::from("Unsupported receiver in a Unified Address.").into());
```
teor2345, line 141: I'm not sure if unknown receivers are invalid?

Unified addresses are designed to be forwards-compatible, so new or experimental receiver types can be added at any time. In this case, we can ignore them because the RPC doesn't have fields for them.

zebra-chain/src/primitives/address.rs
```diff
@@ -2,12 +2,12 @@
...
 //! Usage: <https://docs.rs/zcash_address/0.2.0/zcash_address/trait.TryFromAddress.html#examples>
 
+use zcash_address::unified::{self, Container};
 use zcash_primitives::sapling;
 
-use crate::{orchard, parameters::Network, transparent, BoxError};
+use crate::{parameters::Network, transparent, BoxError};
 
 /// Zcash address variants
-// TODO: Add Sprout addresses
```
teor2345, line 10: Optional Nitpick: There's no sprout address implemented, so I'd like to leave this comment in.

sprout itself is out of scope because it's not used for coinbase transactions.

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -391,6 +391,24 @@ pub async fn test_responses<State, ReadState>(
         .expect("We should have a validate_address::Response");
     snapshot_rpc_validateaddress("invalid", validate_address, &settings);
 
+    // `z_validateaddress`
+    let founder_address = match network {
```
teor2345, line 395: Optional: Can you please add a p2pkh and sapling address, and the unified test vectors from z_listunifiedreceivers to the snapshot tests?
Then we'll have full coverage of all the address types.

It's ok to use the same p2pkh and sapling addresses as the ones inside those unified addresses. We don't need to cover both networks for all of these.

Sorry I didn't see this before.

teor2345 (approved):
I think we're missing some snapshot cases, but luckily we already have test vectors in z_listunifiedreceivers.

There might also be a minor future compatibility issue.

I don't think these are blockers to going ahead and merging this, so I'd suggest doing them in a follow-up PR.

We can also repeat the zcash-rpc-diff test on that PR.

PR #6210, opened 2023-02-23
Manual test: Run unit tests on nightly

teor2345:
I'm seeing some issues with halo2 proof verification on my machine, but they aren't appearing in CI.

One of the possibilities is that I'm using nightly. (Others are a corrupt disk, or a different Rust configuration.)

If nightly also fails in CI, then we should ask ECC to open a Rust miscompilation ticket.
If CI passes, then my machine has the issue.

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -70,7 +70,7 @@ jobs:
       matrix:
         # TODO: Windows was removed for now, see https://github.com/ZcashFoundation/zebra/issues/3801
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


mpguerra:
Seems we need to ask ECC to open a Rust miscopilation issue:

https://github.com/ZcashFoundation/zebra/actions/runs/4251106918/jobs/7393038041 
https://github.com/ZcashFoundation/zebra/actions/runs/4251106918/jobs/7393038170

teor2345:
This PR got us the diagnostics we need, I opened a `halo2` issue:
https://github.com/zcash/halo2/issues/737

PR #6228, opened 2023-02-24
feat(release): create Docker hub binary with mining enabled on release

### LINKED ISSUE #6200: Publish Docker image with mining enabled
## Motivation

We'd like to provide an easy way for miners to test their own mining software with Zebra.

The required changes in the Dockerfile are:
- Compile with `--features getblocktemplate-rpcs`
- Enable the testnet RPC port in the default config, and change the network to testnet
- Expose the port in the Docker image

And in the workflows:
- Publish this Docker image to DockerHub with every release, named something like `zebra-mining-rpcs-testnet-v1.0.0-rc.5.experimental`

### Testing

We can run this image using the Docker commands in the readme, and connect to its RPC port using `zcash-cli`.

We'll be asking mining pools to test this.

## Related Work

This is a much smaller task than:
- #6118

arya2:
## Motivation

We want to make it easy for miners to test Zebra's getblocktemplate RPC methods.

Closes #6200.

## Solution

- Add a Dockerfile with:
  - `getblocktemplate-rpcs` feature enabled
  - Network=Testnet by default
  - An exposed RPC listener port
- Add a workflow for building and publishing images on 'release'

## Review

Anyone can review.

### Reviewer Checklist


docker/mining.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+
+# In this stage we build tests (without running then)
+#
+# We also download needed dependencies for tests to work, from other images.
+# An entrypoint.sh is only available in this step for easier test handling with variables.
+FROM deps AS tests
+# TODO: do not hardcode the user /root/ even though is a safe assumption
+# Pre-download Zcash Sprout, Sapling parameters and Lightwalletd binary
+COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/zcash-params /root/.zcash-params /root/.zcash-params
+COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/lightwalletd /opt/lightwalletd /usr/local/bin
```
teor2345: `lightwalletd` is not required for mining, please make it conditional on "not mining" in the original Dockerfile.

docker/mining.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+# Re-hydrate the minimum project skeleton identified by `cargo chef prepare` in the planner stage,
+# and build it to cache all possible sentry and test dependencies.
+#
+# This is the caching Docker layer for Rust!
+#
+# TODO: is it faster to use --tests here?
+RUN cargo chef cook --release --features sentry,lightwalletd-grpc-tests,getblocktemplate-rpcs --workspace --recipe-path recipe.json
+
+COPY . .
+RUN cargo test --locked --release --features getblocktemplate-rpcs lightwalletd-grpc-tests --workspace --no-run
```
teor2345: Make `lightwalletd` conditional here as well.

docker/mining.Dockerfile
```diff
@@ -0,0 +1,196 @@
+# We are using five stages:
```
teor2345: Having two different docker files to build Zebra means that we will need to remember to make all our changes in both. But in the past, we have forgotten to make these kinds of duplicate changes.

Can you please implement the mining-specific changes in the original `Dockerfile` using a Docker `ARG` instead?
https://github.com/ZcashFoundation/zebra/issues/6200#issuecomment-1439483686

Gustavo can help with how Docker works, and I can help with the shell scripting.

You might want something like this:
```
ARG ENABLE_MINING_RPCS
...
RUN cargo chef cook --release --features sentry${ENABLE_MINING_RPCS:+,getblocktemplate-rpcs} --recipe-path recipe.json 
```

teor2345 (dismissed):
I'm concerned that we are creating a duplicate Dockerfile, it will be much harder to maintain than the single test/production Dockerfile we have at the moment.

Can we make the mining RPCs depend on a Docker argument instead?

.github/workflows/release-binaries.yml
```diff
@@ -33,3 +33,19 @@ jobs:
...
     secrets: inherit
+
+  # The image will be named `zebrad-mining-rpcs-testnet:<semver>`
+  build-mining-testnet:
+    name: Build Release Testnet Mining Docker
+    uses: ./.github/workflows/build-docker-image.yml
+    with:
+      dockerfile_path: ./docker/mining.Dockerfile
+      dockerfile_target: runtime
+      image_name: zebrad-mining-rpcs-testnet
```
teor2345, line 44: This is experimental code, but we're publishing it alongside production images. So I'd like to make it clear that it's not for production.

It might be better to do this by adding `.experimental` to the version tag, so that we don't have to maintain image names with and without `-experimental` after it gets audited.

```
      image_name: zebrad-mining-rpcs-testnet-experimental
```

teor2345:
This PR will need a rebase on the latest `main` to pick up some new CI jobs.

docker/mining.Dockerfile
gustavovalverde: I wasn't able to see this PR previously, but I do agree we need to make it on a single Dockerfile. @arya2 let me know if you'd like some help with this.

gustavovalverde (dismissed):
I wasn't able to see this PR previously, but I do agree we need to make this happen on our existing Dockerfile.

@arya2 let me know if you'd like some help with this changes.

.github/workflows/release-binaries.yml
arya2: I think adding `flavour` in [`ced6cbe`](https://github.com/ZcashFoundation/zebra/pull/6228/commits/ced6cbeb8317592d236fb890ab4de74868143aec#diff-5b40193e09025fd041de92d38c976d0b469e3ee85bef45a71d05d8b5078b1ee8R86) should fix this.

docker/mining.Dockerfile
arya2: The new `mining.Dockerfile` was removed and ARGs were added in [`ced6cbe`](https://github.com/ZcashFoundation/zebra/pull/6228/commits/ced6cbeb8317592d236fb890ab4de74868143aec).

docker/mining.Dockerfile
arya2: I had some trouble handling booleans so I added `FEATURES` and `TEST_FEATURES` ARGS in [`63158a4`](https://github.com/ZcashFoundation/zebra/pull/6228/commits/63158a4b1daa80b7c437afc4c0d89dfbad1a542b) for configuring features. `TEST_FEATURES` is 'lightwalletd-grpc-tests` by default, so it's built with FEATURES="getblocktemplate-rpcs" and TEST_FEATURES="" in [the docker build workflow](https://github.com/ZcashFoundation/zebra/pull/6228/commits/63158a4b1daa80b7c437afc4c0d89dfbad1a542b#diff-5b40193e09025fd041de92d38c976d0b469e3ee85bef45a71d05d8b5078b1ee8R164) to skip the tests.

.github/workflows/release-binaries.yml
teor2345: I'm not sure if adding a tag helps, because after the stable release, the version will be `1.0.0`, which makes it look like supported production code. Instead, I'd like the version to be `1.0.0-experimental` or something similar, to show it is still pre-release quality.

Currently, it's not release-candidate quality code, so I guess we should be using `1.0.0-rc.6.experimental` or `1.0.0-experimental`. But the exact format is less important.

Here's the semantic versioning reference for pre-release code:
https://semver.org/#spec-item-9

What I'm trying to avoid is users installing a docker image containing pre-release code, but thinking it's release quality.

docker/Dockerfile
```diff
@@ -169,20 +174,22 @@ RUN set -ex; \
   { \
     echo "[network]"; \
     echo "network = '${NETWORK}'"; \
+    [ $NETWORK = "Testnet" ] && echo "listen_addr = '127.0.0.1:18233'"; \
```
teor2345: This is not required, the Rust deserialization code already uses the correct port for the network:
https://github.com/ZcashFoundation/zebra/blob/f1bedafbbb1bab97b2536fee50062db47bc4bbea/zebra-network/src/config.rs#L332

So this should work:

```suggestion
    echo "listen_addr = '127.0.0.1'"; \
```

teor2345: Actually, I think there might be a bug in the deserializer code:
https://github.com/ZcashFoundation/zebra/blob/f1bedafbbb1bab97b2536fee50062db47bc4bbea/zebra-network/src/config.rs#L312-L324

I think the default deserialized listen address needs to be "0.0.0.0", so that we get the correct default port when the address is missing. This should probably be a separate PR.

teor2345: Using an address of "127.0.0.1" in the config should still work though!

.github/workflows/release-binaries.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+
+  # The image will be named `zebrad-mining-rpcs-testnet:<semver>`
+  build-mining-testnet:
+    name: Build Release Testnet Mining Docker
+    uses: ./.github/workflows/build-docker-image.yml
+    with:
+      dockerfile_path: ./docker/Dockerfile
+      dockerfile_target: runtime
+      image_name: zebrad-mining-rpcs-testnet
+      tag_suffix: .experimental
```
teor2345, line 46: If we're adding this to the semver, we'll need to tweak it slightly when Zebra is no longer in pre-release.

```suggestion
      # TODO: change this to `-experimental` when we release Zebra `1.0.0`
      tag_suffix: .experimental
```

.github/workflows/release-binaries.yml
```diff
@@ -33,3 +33,23 @@ jobs:
       rust_log: info
     # This step needs access to Docker Hub secrets to run successfully
     secrets: inherit
+
+  # The image will be named `zebrad-mining-rpcs-testnet:<semver>`
```
teor2345: Is this correct? I'm a bit confused about what is happening here.

```suggestion
  # The image will be named `zebrad-mining-rpcs-testnet:<semver>.experimental`
```

docker/mining.Dockerfile
teor2345: Thanks, that seems like a good solution!

docker/Dockerfile
teor2345: Let's open another issue for the config fix, and just use an address in this one?

That way we can schedule the config fix based on our priorities, and get this PR merged without needing any Zebra fixes.

teor2345: I opened this ticket for the Rust default testnet port config fix:
- #6386

.github/workflows/build-docker-image.yml
```diff
@@ -72,6 +86,9 @@ jobs:
           images: |
             us-docker.pkg.dev/zealous-zebra/zebra/${{ inputs.image_name }}
             zfnd/zebra,enable=${{ github.event_name == 'release' && !github.event.release.prerelease }}
+          # appends inputs.tag_suffix to tags
```
arya2:
```suggestion
          # appends inputs.tag_suffix to image tags/names
```

teor2345 (approved):
This looks good, but I think we need to test it by making a fake `.test` release. `workflow_dispatch` won't work on this workflow, so we'll need to do the test after merging to `main`.

Here's the kinds of things I want to check:
1. Is the test or production image the first image that users see?
2. If users are pulling down the `latest` image, do they always get the production image?
3. Is the test image clearly marked as experimental?

If it could be confusing, our other option is to:
1. Use a different project name like `zfnd/zebra-experimental`
2. Remove any images we've pushed to the `zfnd/zebra` DockerHub

teor2345:
@mergifyio refresh

teor2345:
> This looks good, but I think we need to test it by making a fake `.test` release. `workflow_dispatch` won't work on this workflow, so we'll need to do the test after merging to `main`.
> 
> Here's the kinds of things I want to check:
> 
>     1. Is the test or production image the first image that users see?
> 
>     2. If users are pulling down the `latest` image, do they always get the production image?
> 
>     3. Is the test image clearly marked as experimental?
> 
> 
> If it could be confusing, our other option is to:
> 
>     1. Use a different project name like `zfnd/zebra-experimental`
> 
>     2. Remove any images we've pushed to the `zfnd/zebra` DockerHub

@mpguerra @arya2 when do you think we should do this testing? Is it ok to wait until the next release and just see what happens?

PR #6230, opened 2023-02-26
fix(mempool): Stop ignoring some transaction broadcasts

teor2345:
## Motivation

Sometimes Zebra will overwrite transactions in the mempool gossip watch channel, and never broadcast them to the network.

Fixing this is important for reliably sending newly created `lightwalletd` transactions. (For all other transactions, nodes can get them from elsewhere on the network.)

### Specifications

Nodes are not required to broadcast new mempool transactions, but they should:
https://developer.bitcoin.org/devguide/p2p_network.html#transaction-broadcasting

Zebra does not rely on peers broadcasting transactions, it also actively crawls memory pools.

This PR fixes a concurrency bug by repeatedly checking for new transactions from the mempool before broadcasting any.

It has a limit on the number of times it will wait for new changes, which ensures we eventually broadcast and don't infinitely loop.

## Solution

- Combine some transaction IDs that arrive close together rather than overwriting and ignoring them
- Limit the number of times we will wait for new transaction IDs

Related changes:
- Reduce verbosity of transaction broadcast logging

## Review

This is a low-priority bug fix.

### Reviewer Checklist

## Follow Up Work

If we're still dropping some transaction IDs, we might want to change from a watch channel to a broadcast channel. But based on my logs that's unlikely, I'm only seeing a new transaction every few seconds.

### DIFF
--- a/zebrad/src/components/mempool/gossip.rs
+++ b/zebrad/src/components/mempool/gossip.rs
@@ -3,18 +3,15 @@
 //! This module is just a function [`gossip_mempool_transaction_id`] that waits for mempool
 //! insertion events received in a channel and broadcasts the transactions to peers.
 
-use tower::{timeout::Timeout, Service, ServiceExt};
-
-use zebra_network as zn;
-
-use tokio::sync::watch;
-use zebra_chain::transaction::UnminedTxId;
-
 use std::collections::HashSet;
 
-use crate::BoxError;
+use tokio::sync::watch;
+use tower::{timeout::Timeout, Service, ServiceExt};
 
-use crate::components::sync::TIPS_RESPONSE_TIMEOUT;
+use zebra_chain::transaction::UnminedTxId;
+use zebra_network as zn;
+
+use crate::{components::sync::TIPS_RESPONSE_TIMEOUT, BoxError};
 
 /// Runs continuously, gossiping new [`UnminedTxId`] to peers.
 ///
@@ -37,10 +34,20 @@ where
     let mut broadcast_network = Timeout::new(broadcast_network, TIPS_RESPONSE_TIMEOUT);
 
     loop {
-        // once we get new data in the channel, broadcast to peers
+        // once we get new data in the channel, broadcast to peers,
+        // the mempool automatically combines some transactions that arrive close together
         receiver.changed().await?;
+        let mut txs = receiver.borrow().clone();
+        tokio::task::yield_now().await;
 
-        let txs = receiver.borrow().clone();
+        // also combine transactions that arrived shortly after this one
+        while receiver.has_changed()? {
+            // Correctness: reset has_changed(), don't hold the lock
+            let extra_txs = receiver.borrow_and_update().clone();
+            txs.extend(extra_txs.iter());
+            tokio::task::yield_now().await;
+        }
+
         let txs_len = txs.len();
         let request = zn::Request::AdvertiseTransactionIds(txs);
 
--- a/zebrad/src/components/mempool/gossip.rs
+++ b/zebrad/src/components/mempool/gossip.rs
@@ -60,10 +60,10 @@ where
         let request = zn::Request::AdvertiseTransactionIds(txs);
 
         // TODO: rate-limit this info level log?
-        info!(%request, %combined_changes, "sending mempool transaction broadcast");
+        info!(%request, changes = %combined_changes, "sending mempool transaction broadcast");
         debug!(
             ?request,
-            ?combined_changes,
+            changes = ?combined_changes,
             "full list of mempool transactions in broadcast"
         );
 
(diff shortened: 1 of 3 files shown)


teor2345:
The Docker jobs are likely to fail with a known Google PGP key error, feel free to restart all the Docker jobs.

teor2345:
Ah yeah there it is:
https://github.com/ZcashFoundation/zebra/actions/runs/4277535424/jobs/7446412852#step:9:622

teor2345:
@mergifyio update

teor2345:
This PR looks like it has a GitHub status reporting issue, the "Setup update-to-tip test" job has actually finished, but it's not showing up as finished.

That job isn't required, so mergify should still work fine.

zebrad/src/components/mempool/gossip.rs
```diff
@@ -37,14 +37,35 @@ where
...
+        let mut combined_changes = 1;
+
+        // once we get new data in the channel, broadcast to peers,
+        // the mempool automatically combines some transactions that arrive close together
         receiver.changed().await?;
+        let mut txs = receiver.borrow().clone();
+        tokio::task::yield_now().await;
+
+        // also combine transactions that arrived shortly after this one
+        while receiver.has_changed()? && combined_changes < MAX_CHANGES_BEFORE_SEND {
```
oxarbitrage, line 49:
```suggestion
        while receiver.has_changed()? && combined_changes <= MAX_CHANGES_BEFORE_SEND {
```

Do we want to loop 10 times or 9 ?

zebrad/src/components/mempool/gossip.rs
```diff
@@ -37,14 +37,35 @@ where
...
+            // Correctness: reset has_changed(), don't hold the lock
```
oxarbitrage: This comment is not totally clear to me. When we are resetting and what ?

oxarbitrage (dismissed):
Looks pretty good but how do we know this is doing what we need ? Any chance to do a test ?

zebrad/src/components/mempool/gossip.rs
teor2345: I don't think it matters either way. We have already checked once before the loop, so it's 10 combined changes at the end of the loop.

teor2345:
> Looks pretty good but how do we know this is doing what we need ? Any chance to do a test ?

I have looked at the log, and I haven't seen any combined changes yet. So I think it will only happen under load.

I don't think we can write a test for this without using something like `loom` to get an unusual execution order:
https://docs.rs/loom/latest/loom/

(We need to run both tasks at the same time, and have an update happen in-between two of the times that the gossip task checks for changes. We can't do that in test code easily.)

zebrad/src/components/mempool/gossip.rs
teor2345:
```suggestion
            // Correctness
            // - set the has_changed() flag to false using borrow_and_update()
            // - clone() so we don't hold the watch channel lock while modifying txs 
```

teor2345: I expanded it in commit 09912976983923ec59dcecb366c0979a98ffa981. The relevant API docs are:
https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.borrow_and_update
https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.has_changed

zebrad/src/components/mempool/gossip.rs
```diff
@@ -37,14 +37,37 @@ where
...
+        // the mempool automatically combines some transactions that arrive close together
         receiver.changed().await?;
+        let mut txs = receiver.borrow().clone();
+        tokio::task::yield_now().await;
+
+        // also combine transactions that arrived shortly after this one
+        while receiver.has_changed()? && combined_changes < MAX_CHANGES_BEFORE_SEND {
+            // Correctness
+            // - set the has_changed() flag to false using borrow_and_update()
+            // - clone() so we don't hold the watch channel lock while modifying txs
```
teor2345:
```suggestion
            // - clone() so we don't hold the watch channel lock while modifying txs
```

oxarbitrage (approved):

teor2345:
@mergifyio refresh

teor2345:
Failed in the merge queue with:
> - Constraint constraints/compute.vmExternalIpAccess violated for project 143793276228. Add instance projects/zealous-zebra/zones/us-central1-a/instances/sync-past-checkpoint-6260-merge-9ce0bb0 to the constraint to use external IP with it.

https://github.com/ZcashFoundation/zebra/actions/runs/4309018317/jobs/7516676191#step:10:84

Gustavo is working on an infrastructure fix.

To re-queue the PR, comment `@mergify refresh` once he's fixed it.

teor2345:
@mergifyio refresh

Issue #4529, opened 2022-05-29
Make PRs from external user repositories pass or skip CI jobs


teor2345:
## Motivation

Currently, Zebra's CI fails on PRs from external contributors. This can be a really confusing and negative experience for a first-time contributor.

Example: (recent examples first)
https://github.com/ZcashFoundation/zebra/pull/7516
https://github.com/ZcashFoundation/zebra/pull/4527

### Jobs that need to be skipped

We don't want external PRs running on our Google Cloud, so we need to skip all those jobs.
We want to skip release and deployment jobs as well, because they use secret keys.

See the full list here:
https://github.com/ZcashFoundation/zebra/issues/4529#issuecomment-1396055038

### Jobs that need to be fixed

All other jobs should be fixed if they don't work for external PRs.

### Designs

We can skip some tests that would otherwise fail, because Mergify will catch them when it does a full test run. But we need to disable in-place merges to make sure Mergify does a separate run.

This is more secure, because we review code, approve it, then run Mergify on it.

## Related Work

Our previous attempts to fix this issue didn't work:
- #3419

teor2345:
We're getting closer to the release candidate series, so this is a medium priority now.

teor2345:
Attracting external developers is not one of the goals of the release candidate, so this is a low priority.

mpguerra:
Is it just as simple as skipping the CI Docker workflow or are there others that would need to be skipped?

How will this change affect the complexity of our CI rules and/or setup?

teor2345:
> Is it just as simple as skipping the CI Docker workflow or are there others that would need to be skipped?

We'll also need to skip other workflows that write to our GitHub or Docker:
- release drafter (PR auto-labels)
- zcash-lightwalletd (google cloud Docker image push)

And avoid sending our GitHub access token to the protoc installer, by doing one of these things:
- just skip those entire workflows (simplest option, and a tiny amount of work)
- skip the protoc and --all-features parts of those workflows for external users (slightly more work)
- commit the generated files to our git repository, and have a workflow that checks they are up to date (this would take slightly more work, but simplify a lot of other workflows, and decrease our CI failure

If we decide we really want one of these workflows to run, we can do the more complicated thing for it later.

> How will this change affect the complexity of our CI rules and/or setup?

If we choose the simplest option, it is:
- change the workflow triggers for about 5 workflows by adding 2 lines of conditions
- change the patch workflows for those workflows to run instead

[...]

PR #6236, opened 2023-02-27
change: Improve CommitBlockError message by including underlying error message

### LINKED ISSUE #6218: Implement custom Debug trait for CommitBlockError
## Motivation
For CommitBlockError tuple the default implementation for Debug is used now so with statements like `println!("{:?}, err);` there is no formatted error description is printed for the base ValidateContextError but only the error parameters.
Suggestion is to implement the std::fmt::Debug trait for CommitBlockError which would also print expanded ValidateContextError, like:
```
impl std::fmt::Debug for CommitBlockError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "CommitBlockError(\"{}\": ValidateContextError(\"{}\"))", self.to_string(), self.0.to_string())
    }
}
``` 
Or maybe we could use debug_tuple with the formatted ValidateContextError field as well.

dimxy:
## Motivation

Improve logging of CommitBlockError logged with the default Display method by extending with the underlying ValidateContextError message.

For example, before this fix a CommitBlockError with a ValidateContextError::Orphaned error would be printed as:
```
block is not contextually valid
```
with this fix:
```
block is not contextually valid: block height 12 is lower than the current finalized height 127
```
fixes #6218

### DIFF
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -43,7 +43,7 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
 /// An error describing the reason a block could not be committed to the state.
 #[derive(Debug, Error, PartialEq, Eq)]
-#[error("block is not contextually valid")]
+#[error("block is not contextually valid: {}", .0)]
 pub struct CommitBlockError(#[from] ValidateContextError);
 
 /// An error describing why a block failed contextual validation.


upbqdn (approved):
This PR looks good, thank you.

Edit: I moved my question into the conversation below.

zebra-state/src/error.rs
```diff
@@ -43,7 +43,7 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
 /// An error describing the reason a block could not be committed to the state.
 #[derive(Debug, Error, PartialEq, Eq)]
-#[error("block is not contextually valid")]
+#[error("block is not contextually valid: {}", .0)]
```
upbqdn, line 46: Would you be able to provide a reference to the ` .0` part?. I haven't seen that syntax yet, and I didn't find any docs mentioning it.

oxarbitrage:
Hey, thank you for the contribution. Can you show us how were you able to see that `block is not contextually valid: block height 12 is lower than the current finalized height 127` is now the error message ? I was wondering if we could add this to a test but maybe it will not worth it.

zebra-state/src/error.rs
conradoplg: I think it's documented in the third bullet point here: https://docs.rs/thiserror/latest/thiserror/#details

In this case I think it could be just `{0}`

teor2345:
@mergifyio update

teor2345:
> Hey, thank you for the contribution. Can you show us how were you able to see that `block is not contextually valid: block height 12 is lower than the current finalized height 127` is now the error message ? I was wondering if we could add this to a test but maybe it will not worth it.

We don't usually test log messages, because they aren't part of Zebra's supported functionality. (We only use them in acceptance tests to see that the node has done the thing we actually want to test.)

dimxy:
> 

here is how this fix is used in my branch: https://github.com/dimxy/zebra/blob/0a5eaa4e9d59cdaa3a937d116ed0e6aa4b5a8b76/zebra-state/src/service/check/tests/komodo_nn.rs#L193:
it prints the ValidateContextError message in addition to CommitBlockError default message ("block is not contextually valid"):
```
commit_result1=block is not contextually valid: immature transparent coinbase spend: attempt to spend OutPoint { hash: transaction::Hash("cd5b8e29b0c893c57cc3c30d170f99a82e56157e3992374faad6d242ba0863ae"), index: 0 } at Height(139), but spends are invalid before Height(232), which is 100 blocks after it was created at Height(132)
```
Command to run: `cargo test -p zebra-state komodo_nn::komodo_valid_fork_from_below_last_notarised_height`

teor2345:
> > 
> 
> here is how this fix is used in my branch: https://github.com/dimxy/zebra/blob/0a5eaa4e9d59cdaa3a937d116ed0e6aa4b5a8b76/zebra-state/src/service/check/tests/komodo_nn.rs#L193: it prints the ValidateContextError message in addition to CommitBlockError default message ("block is not contextually valid"):
> 
> ```
> commit_result1=block is not contextually valid: immature transparent coinbase spend: attempt to spend OutPoint { hash: transaction::Hash("cd5b8e29b0c893c57cc3c30d170f99a82e56157e3992374faad6d242ba0863ae"), index: 0 } at Height(139), but spends are invalid before Height(232), which is 100 blocks after it was created at Height(132)
> ```
> 
> Command to run: `cargo test -p zebra-state komodo_nn::komodo_valid_fork_from_below_last_notarised_height`

Are you trying to track down this bug?
- #6195

teor2345:
Note for the Zebra developers: we'll need to re-push this PR to our repository to get the Google Cloud tests to run, because the secrets are only available to developers with write access.

dimxy:
> > > 
> > 
> > 
> > here is how this fix is used in my branch: https://github.com/dimxy/zebra/blob/0a5eaa4e9d59cdaa3a937d116ed0e6aa4b5a8b76/zebra-state/src/service/check/tests/komodo_nn.rs#L193: it prints the ValidateContextError message in addition to CommitBlockError default message ("block is not contextually valid"):
> > ```
> > commit_result1=block is not contextually valid: immature transparent coinbase spend: attempt to spend OutPoint { hash: transaction::Hash("cd5b8e29b0c893c57cc3c30d170f99a82e56157e3992374faad6d242ba0863ae"), index: 0 } at Height(139), but spends are invalid before Height(232), which is 100 blocks after it was created at Height(132)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Command to run: `cargo test -p zebra-state komodo_nn::komodo_valid_fork_from_below_last_notarised_height`
> 
> Are you trying to track down this bug?
> 
> * [Fix invalid transparent spends in `getblocktemplate` RPC responses #6195](https://github.com/ZcashFoundation/zebra/issues/6195)

No, we are just playing with our project's params (coinbase maturity for testnet)

upbqdn:
> we'll need to re-push this PR to our repository

I re-pushed the branch from this PR to the Zebra repo here https://github.com/ZcashFoundation/zebra/tree/fix-commit-block-error-msg, but I don't know what to do next since I can't change the source branch of this PR.

teor2345:
> > we'll need to re-push this PR to our repository
> 
> I re-pushed the branch from this PR to the Zebra repo here https://github.com/ZcashFoundation/zebra/tree/fix-commit-block-error-msg, but I don't know what to do next since I can't change the source branch of this PR.

I opened a PR at #6251, anyone can approve it.

Because the PR is from a branch in our repository, it has access to our GitHub secrets for CI. We want to fix this eventually in #4529.

teor2345:
I'll need to close this PR to get Mergify to check the other PR, but we are going to merge it!

PR #6249, opened 2023-02-28
Show `Cargo.lock` files in the github diff view

mpguerra:
## Motivation

During this week's sprint retrospective we discussed how we sometimes miss unintended changes to the `Cargo.lock` file as this is not shown by default on the diff page. 

## Solution

This PR adds a `.gitattributes` file as per: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github to enable `Cargo.lock` files to be shown in the github diff. 

### Reviewer Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+**/Cargo.lock linguist-generated=false
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,17 +38,17 @@ checksum = "74f5722bc48763cb9d81d8427ca05b6aa2842f6632cf8e4c0a29eef9baececcc"
 dependencies = [
  "darling 0.10.2",
  "ident_case",
- "proc-macro2 1.0.51",
- "quote 1.0.23",
- "syn 1.0.109",
+ "proc-macro2 1.0.47",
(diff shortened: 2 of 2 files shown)


mpguerra:
Seems like `Cargo.lock` it wouldn't be shown anyway because it's too large...

Screenshot 2023-02-28 at 10 06 56

mpguerra:
This is how it looked before: 

Screenshot 2023-02-28 at 10 09 40

mpguerra:
We need to decide whether we want to make this change

teor2345 (approved):
Diffs above 400 lines are hidden on GitHub, regardless of the type of file:
https://docs.github.com/en/repositories/creating-and-managing-repositories/about-repositories#diff-limits

This is not configurable for performance reasons.

But we'll be able to see smaller Cargo.lock diffs, and hopefully we'll notice when they are hidden.

If that's not enough, we can put a warning in CI if the diff is too large. But let's see how this goes first?

teor2345:
For example, here's a large diff warning action:
https://github.com/cornell-dti/big-diff-warning

PR #6251, opened 2023-02-28
change(format): improve CommitBlockError message: include underlying ValidateContextError message

### LINKED ISSUE #6218: Implement custom Debug trait for CommitBlockError

teor2345:
This is a copy of PR #6236 in the Zebra repository, so all our CI can run on it.

### DIFF
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -43,7 +43,7 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
 /// An error describing the reason a block could not be committed to the state.
 #[derive(Debug, Error, PartialEq, Eq)]
-#[error("block is not contextually valid")]
+#[error("block is not contextually valid: {}", .0)]
 pub struct CommitBlockError(#[from] ValidateContextError);
 
 /// An error describing why a block failed contextual validation.


teor2345:
@mergifyio refresh

upbqdn (dismissed):

upbqdn:
I can't approve this PR since I pushed it, even though I didn't make the commit.

teor2345:
> I can't approve this PR since I pushed it, even though I didn't make the commit.

If that rule gets annoying an admin can turn it off. But every time I've seen it come up, it's a situation where we actually want another review!

teor2345:
@mergifyio update

mpguerra (approved):

teor2345:
Failed due to an infrastructure configuration issue, Gustavo is working on a fix.

To re-queue the PR, comment `@mergify refresh` once he's fixed it.

PR #6273, opened 2023-03-07
change(docs): Update Zebra's build instructions

upbqdn:
-------

## Motivation

Building Zebra requires the [pkg-config](http://pkgconf.org/) tool, which wasn't mentioned in the build instructions. The tool is required by the [pkg-config crate](https://crates.io/crates/pkg-config) [here](https://github.com/zcashfoundation/zebra/blob/25e656b26dc55fa689d34fe18c3e524417065e12/Cargo.lock#L2944). As part of the solution, I tried to refine the build instructions. 

## Solution

This PR:

- adds `pkg-config` to the build dependencies;
- adds a specific example of installing dependencies for Arch;
- renames the _Build Instructions_ section to _Building Zebra_ for consistency with other sections
- moves the _Optional Features_ subsection into the _Building Zebra_ section, and consolidates all Zebra's build features;
- explicitly states that users can combine the features;
- contains a bunch of additional small edits.

### Reviewer Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@
   - [Using Zebra](#using-zebra)
 - [Release Candidates](#release-candidates)
 - [Getting Started](#getting-started)
-  - [Build Instructions](#build-instructions)
+  - [Building Zebra](#building-zebra)
+    - [Optional Features](#optional-features)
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
Changes looks good to me, thanks!

PR #6274, opened 2023-03-07
feat(elasticsearch): Export block data to elasticsearch database

### LINKED ISSUE #6270: Save blockchain data to elasticsearch database
We want to export block data (the whole block history) into an [elasticsearch](https://www.elastic.co/) database so we can use [kibana](https://www.elastic.co/kibana/) for data exploration.

## TODO

- [x] Use [Rust elasticsearch](https://docs.rs/elasticsearch/8.5.0-alpha.1/elasticsearch/) crate in order to save blocks into elasticsearch.
  - [x] Use a bulk request when wee are far away from the tip.
  - [x] Reduce the bulk size when we get closer to the tip so we can have real time data in kibana as the chain is in sync.
- [x] Hide all elasticsearch code behind a rust feature.
- [x] Add a config section for elastic host, username and password.
- [x] Manual test the full thing in Testnet.
- [x] Create an entry in the book on how to setup and run.

oxarbitrage:
## Motivation

We want to export block data from the Zcash blockchain using Zebra into an elasticsearch database

https://github.com/ZcashFoundation/zebra/issues/6270

### Specifications

All finalized blocks pass through the `commit_finalized_direct` method of the `FinalizedState` structure.

Here, if the `elasticsearch` rust feature is enabled we call a new method `elasticsearch()` that will take the additional work of sending the blocks to the elasticsearch database.

Url, username and elasticsearch database password are configurable. 

Indexing blocks one by one is a performance pain for the entire blockchain. The idea is to use the bulk elasticsearch api to send a batch of blocks from time to time when we are far away from the tip and then starting indexing by one when we get closer so we will end up with a real time thing.

## Solution

TBA

### Reviewer Checklist

## Follow Up Work

Possible TODOs:

- Test with the mainnet.
- Add to CI?
- Improve performance.

### DIFF
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -357,14 +357,10 @@ impl FinalizedState {
         finalized: &FinalizedBlock,
         committed_tip_height: &Option<block::Height>,
     ) {
-        use chrono::Utc;
-        use elasticsearch::BulkParts;
-        use serde_json::Value;
-
         if let Some(client) = self.elastic_db.clone() {
             let block = &finalized.block;
             let block_time = block.header.time.timestamp();
-            let local_time = Utc::now().timestamp();
+            let local_time = chrono::Utc::now().timestamp();
 
             const AWAY_FROM_TIP_BULK_SIZE: usize = 800;
             const CLOSE_TO_TIP_BULK_SIZE: usize = 2;
@@ -393,13 +389,14 @@ impl FinalizedState {
 
             // We are in bulk time, insert to ES all we have.
             if self.elastic_blocks.len() >= blocks_size_to_dump {
-                let rt = tokio::runtime::Runtime::new().unwrap();
+                let rt = tokio::runtime::Runtime::new()
+                    .expect("wuntime creation for elasticsearch should not fail.");
                 let blocks = self.elastic_blocks.clone();
                 let network = self.network;
 
                 rt.block_on(async move {
                     let response = client
-                        .bulk(BulkParts::Index(
+                        .bulk(elasticsearch::BulkParts::Index(
                             format!("zcash_{}", network.to_string().to_lowercase()).as_str(),
                         ))
                         .body(blocks)
@@ -409,7 +406,7 @@ impl FinalizedState {
 
                     // Make sure no errors ever.
                     let response_body = response
-                        .json::<Value>()
+                        .json::<serde_json::Value>()
                         .await
                         .expect("ES response parsing to a json_body should never fail");
                     let errors = response_body["errors"].as_bool().unwrap_or(true);
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -230,7 +230,7 @@ impl FinalizedState {
 
         // Save blocks to elasticsearch if the feature is enabled.
         #[cfg(feature = "elasticsearch")]
-        self.elasticsearch(&finalized, &committed_tip_height);
+        self.elasticsearch(&finalized.block, &committed_tip_height);
 
         // Assert that callers (including unit tests) get the chain order correct
         if self.db.is_empty() {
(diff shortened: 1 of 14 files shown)


oxarbitrage:
The current design is to panic on any indexing or any other error. I am sycronizing the testnet and it is looking good by now, i am at about half of the sync, but i need to reach the phase where blocks are indexed one by one to see if that is working. In the performance side, it is of course, slower than a sync without elastic enabled.

testnet_sync

I will let this going overnight and see how it ends.

oxarbitrage:
The testnet got in sync overnight however the closer to tip interval was too small so had to change it here: https://github.com/ZcashFoundation/zebra/pull/6274/commits/014a108750e11502945aae4f44571ff04eefaeef

The performance is not very good, but it was expected in the current state, it took around 12 hours to sync the testnet.

elastic2

zebra-state/src/service/finalized_state.rs
```diff
@@ -53,7 +53,7 @@ pub use disk_db::{DiskWriteBatch, WriteDisk};
 ///
 /// This is different from `NonFinalizedState::clone()`,
 /// which returns an independent copy of the chains.
-#[derive(Clone, Debug, Eq, PartialEq)]
```
oxarbitrage, line 56: It is important here to remove this or we can't add the type `elasticsearch::Elasticsearch` as a field otherwise.

Fortunately, this two derivations were not used at all in current Zebra.

zebra-state/src/service/finalized_state.rs
```diff
@@ -206,6 +228,10 @@ impl FinalizedState {
         let committed_tip_hash = self.db.finalized_tip_hash();
         let committed_tip_height = self.db.finalized_tip_height();
 
+        // Save blocks to elasticsearch if the feature is enabled.
+        #[cfg(feature = "elasticsearch")]
+        self.elasticsearch(&finalized.block, &committed_tip_height);
```
arya2: Could this call be moved to the `if result.is_ok()` block at the bottom of the method.

zebra-state/src/service/finalized_state.rs
```diff
@@ -321,6 +347,77 @@ impl FinalizedState {
...
+            if local_time - block_time < CLOSE_TO_TIP_SECONDS {
+                blocks_size_to_dump = CLOSE_TO_TIP_BULK_SIZE;
+            }
+
+            // Insert the operation line.
+            let height_number = committed_tip_height.unwrap_or(block::Height(0)).0;
+            self.elastic_blocks.push(
+                serde_json::json!({
+                    "index": {
+                        "_id": height_number.to_string().as_str()
```
arya2, line 380: Should the index be different every time? This is currently the finalized tip height before the block is committed, is it supposed to be the block's height?

zebra-state/src/service/finalized_state.rs
```diff
@@ -321,6 +347,77 @@ impl FinalizedState {
...
+                    }
+                })
+                .to_string(),
+            );
+
+            // Insert the block itself.
+            self.elastic_blocks
+                .push(serde_json::json!(block).to_string());
```
arya2, line 388: Is there any block data that should be excluded here?

arya2 (commented):
This looks fantastic!

zebra-state/src/service/finalized_state.rs
oxarbitrage: Moved at https://github.com/ZcashFoundation/zebra/pull/6274/commits/bf7b0fb07c957adc626a434f49bc34d79e221b8a

zebra-state/src/service/finalized_state.rs
oxarbitrage: Yea, `_id` needs to be different for each block or it will overwrite the data. Height is what makes more sense but you are right, it should be the block height from the block. I fixed this at https://github.com/ZcashFoundation/zebra/pull/6274/commits/8b2c474aa4432c19e374269c310573d897cbb8a9

zebra-state/src/service/finalized_state.rs
oxarbitrage: I am not sure, probably not everything is useful or makes sense but for now i am just dumping the entire block into the database, users can filter stuff they don't want to see from the front end.

arya2 (approved):

PR #6287, opened 2023-03-10
change(rpc): Add confirmations to getrawtransaction method response

### LINKED ISSUE #6286: Add 'confirmations' field to `getrawtransaction` RPC
## Motivation

Mining pools may need a way to see the confirmation progress for a particular transaction.

## Specifications

Zebra needs to return the `confirmations` field.

## Design

- Return `confirmations: 1 + latest_chain_tip_height - transaction_block_height` if the transaction is in the state

arya2:
## Motivation

Mining pools may need a way to see the confirmation progress for a particular transaction.

Closes #6286.

## Solution

- Add `confirmations` field to the `getrawtransaction` response if `verbose != 0` and the transaction is in the best chain
- Update tests

## Review

Anyone can review.

### Reviewer Checklist


zebra-rpc/src/methods.rs
```diff
@@ -902,15 +901,21 @@ where
...
-                        },
-                    )?,
-                ),
+                zebra_state::ReadResponse::Transaction(Some((tx, height))) => {
+                    // # Correctness:
+                    //
+                    // The transaction request only checks the best chain, so it's okay
+                    // to get the transaction depth by subtracting the transaction height
+                    // from the best chain tip height just before the transaction request.
+                    let depth = best_tip_height.map(|tip_height| (tip_height.0 - height.0));
```

We could merge #6330 first, and use its checked subtraction.

upbqdn (dismissed):
Looks great! I left one optional suggestion.

zebra-rpc/src/methods.rs
teor2345: What happens if the best chain tip changes between the time we store the tip height, and the transaction request?

I think the transactions returned by the request could be higher than the previous tip if we add a block to the tip, or if we switch to another chain with more blocks.

zebra-rpc/src/methods.rs
```diff
@@ -1439,6 +1444,10 @@ pub enum GetRawTransaction {
         /// The height of the block that contains the transaction, or -1 if
         /// not applicable.
         height: i32,
+        /// The confirmations of the block that contains the transaction,
+        /// or None if the transaction is in the mempool.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        confirmations: Option<u32>,
```
teor2345: `zcashd` always provides this field, but uses  `0` for mempool transactions:
https://github.com/zcash/zcash/blob/c0fbeb809bf2303e30acef0d2b74db11e9177427/src/rpc/rawtransaction.cpp#L332

zebra-rpc/src/methods.rs
arya2: > What happens if the best chain tip changes between the time we store the tip height, and the transaction request?

It'll return slightly outdated information - it gets `best_tip_height` before making the transaction request, so if the tip grows, confirmations will be lower than the blocks current depth in the state, and if it switches to a fork between requests, the height from the previous best chain should be the same or lower.

> I think the transactions returned by the request could be higher than the previous tip if we add a block to the tip, or if we switch to another chain with more blocks.

Good point, the transactions are always on the best chain so confirmations should be greater than or equal to 1.

arya2: Added a default `confirmations` of 1 in [`bb50243`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/bb50243b03a4c768fb1aa43e46aad452a35b9a48).

zebra-rpc/src/methods.rs
arya2: Whoops, I missed this, thank you. Fixed in [`bb50243`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/0b5c7bfc4c1a75ce3953f947738a3d830e073eb8).

zebra-rpc/src/methods.rs
```diff
@@ -902,15 +901,23 @@ where
...
-                    )?,
-                ),
+                zebra_state::ReadResponse::Transaction(Some((tx, height))) => {
+                    // # Correctness:
+                    //
+                    // The transaction request only checks the best chain, so it's okay
+                    // to get the transaction depth by subtracting the transaction height
+                    // from the best chain tip height just before the transaction request.
+                    let depth = best_tip_height
+                        .map(|tip_height| (tip_height.0 - height.0))
```
teor2345: If the best chain changes, and the old `tip_height` is below the transaction height in the latest best chain, this subtraction will:
* panic in debug mode
* produce a large invalid value in release mode

Instead, we could:
* return the depth in the state request, so it is always valid and consistent
* replace the result with `0` (but that might cause problems for clients, because that's incorrect data - we know the transaction is now in the best chain)

I'd like to avoid re-checking the best tip, because the same chain-switch issue can happen again between the state request and the re-check.

zebra-rpc/src/methods.rs
teor2345: > if it switches to a fork between requests, the height from the previous best chain should be the same or lower.

This is not guaranteed by the protocol rules: better chains can have fewer blocks if those blocks have done more work.

teor2345: I'm not sure this potential panic has been fixed, I added a follow-up comment here:
https://github.com/ZcashFoundation/zebra/pull/6287/files#r1144127025

zebra-rpc/src/methods.rs
teor2345: We should probably also add a test with this edge-case, so we don't make this mistake in future RPC changes. (We almost missed it and merged the code.)

zebra-rpc/src/methods.rs
arya2: > better chains can have fewer blocks if those blocks have done more work

Ah, then I think it has to be returned by the state service. Fixed in [`2610124`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/26101243da0af4652dc281448084d51249a23a31).

zebra-rpc/src/methods.rs
arya2: Added a test in [`e12437d`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/e12437d7f37913bd2159c414013fb261af10d320)/[`07fe752`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/07fe7521079f0482c1200daadaadf9a4d0063aa7), and updated the method to get confirmations from the state service in [`d6791a5`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/d6791a5e50fe0ff94742bdb59cee1c742b92b374).

zebra-rpc/src/methods.rs
```diff
@@ -878,19 +879,17 @@ where
...
                     }
                 }
                 _ => unreachable!("unmatched response to a transactionids request"),
             };
 
             // Now check the state
-            let request = zebra_state::ReadRequest::Transaction(txid);
+            let request = zebra_state::ReadRequest::Transaction {
+                hash: txid,
+                should_return_confirmations: true,
```
teor2345: Can we just always return the confirmations? Is the extra complexity of an additional argument and optional return value required?

zebra-rpc/src/methods.rs
```diff
@@ -902,16 +901,19 @@ where
...
-                            code: ErrorCode::ServerError(0),
-                            message: error.to_string(),
-                            data: None,
-                        },
-                    )?,
-                ),
-                zebra_state::ReadResponse::Transaction(None) => Err(Error {
+                zebra_state::ReadResponse::Transaction {
+                    transaction_and_height: Some((tx, height)),
+                    confirmations: Some(confirmations),
```
teor2345: This might be easier to handle (and review) as an optional struct with 3 non-optional fields. Otherwise, there are combinations of optional fields that can never occur, but callers still have to handle them.

Edit: switched to confirmations optional, then switched back

zebra-rpc/src/methods.rs
```diff
@@ -1439,6 +1441,9 @@ pub enum GetRawTransaction {
         /// The height of the block that contains the transaction, or -1 if
         /// not applicable.
```
teor2345:
```suggestion
        /// The height of the block in the best chain that contains the transaction, or -1 if
        /// the transaction is in the mempool.
```

teor2345: Edit: we don't check side-chains

zebra-rpc/src/methods.rs
```diff
@@ -1439,6 +1441,9 @@ pub enum GetRawTransaction {
...
         height: i32,
+        /// The confirmations of the block that contains the transaction,
+        /// or 0 if the transaction is in the mempool.
```
teor2345:
```suggestion
        /// The confirmations of the block in the best chain that contains the transaction,
        /// or 0 if the transaction is in the mempool.
```

zebra-state/src/service.rs
```diff
@@ -1247,21 +1247,39 @@ impl Service<ReadRequest> for ReadStateService {
...
+                            read::transaction(latest_best_chain.as_ref(), &state.db, hash);
+
+                        let confirmations = should_return_confirmations
+                            .then(|| {
+                                transaction_and_height.as_ref().and_then(|(_tx, height)| {
+                                    latest_best_chain
+                                        .map(|latest_best_chain| {
+                                            latest_best_chain.non_finalized_tip_height()
+                                        })
+                                        .or_else(|| state.db.finalized_tip_height())
```
teor2345: What happens if a block is pushed to the finalized state after we clone the `latest_best_chain`?
(I think that both the lower and higher values are valid, but we need a correctness comment to explain why, including a justification that the subtraction won't panic.)

Optional: this code could also get moved to the `read::block` module, along with the rest of the complex queries with correctness comments.

arya2: > What happens if a block is pushed to the finalized state after we clone the `latest_best_chain`? (I think that both the lower and higher values are valid, but we need a correctness comment to explain why, including a justification that the subtraction won't panic.)

It would read the more current finalized tip height if the non-finalized chain is empty, or a potentially outdated non-finalized tip height that's current at the time that it's cloned. I left the same correctness comment as the `read::depth()` function in [`f883072`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/f8830721e7a6e150b646163d24bc5977e3304db1).
 
> Optional: this code could also get moved to the `read::block` module, along with the rest of the complex queries with correctness comments.

Moved to `read::block::mined_transaction()` function in [`f883072`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/f8830721e7a6e150b646163d24bc5977e3304db1).

zebra-rpc/src/methods.rs
arya2: Agreed, updated in [`f883072`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/f8830721e7a6e150b646163d24bc5977e3304db1).

zebra-rpc/src/methods.rs
arya2: Yes, I wanted to avoid checking the tip height, but I don't think we should be adding boolean flags for what fields to include in state service responses, removed so that it always returns the confirmations in [`f883072`](https://github.com/ZcashFoundation/zebra/pull/6287/commits/f8830721e7a6e150b646163d24bc5977e3304db1).

teor2345 (approved):
Thanks, this looks great, and thanks for the tests!

PR #6320, opened 2023-03-14
fix(network): validate addr length before reading

### LINKED ISSUE #6280: [NCC-E005955-HV6] `zebra-network`: Buffer length validation after memory allocation
## Motivation

We want to track all of the findings from the zebra audit.

## Details

When [deserializing a version 2 address](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/protocol/external/addr/v2.rs#L285-L296), the length of the (variable-length) address is validated after the address is read into a buffer. An unexpectedly large address can temporarily lead to a large memory allocation on the heap. It is recommended to peek at the length of the address and validate it before reading the `addr`:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/protocol/external/addr/v2.rs#L285-L296

Note that the deserialization logic ensures that the length of the `addr` vector is smaller than [`MAX_U8_ALLOCATION`](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/serialization/zcash_deserialize.rs#L187) (currently 2097147 bytes) which is significantly higher than `MAX_ADDR_V2_ADDR_SIZE` (currently set to 512).

oxarbitrage:
## Motivation

We want to validate addr length before reading and deserializing the addr itself.

Close https://github.com/ZcashFoundation/zebra/issues/6280

## Solution

There is a `sizeAddr` field in the [spec](https://zips.z.cash/zip-0155#specification) that we can use first to validate. Then we use `zcash_deserialize_bytes_external_count` to read the address with the length we already have.

## Review

Not sure if there is a better alternative. Anyone can review.

We already have tests for this, for example `parses_msg_addr_v2_ip`.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/protocol/external/addr/v2.rs
+++ b/zebra-network/src/protocol/external/addr/v2.rs
@@ -15,8 +15,8 @@ use byteorder::{BigEndian, ReadBytesExt};
 use thiserror::Error;
 
 use zebra_chain::serialization::{
-    CompactSize64, DateTime32, SerializationError, TrustedPreallocate, ZcashDeserialize,
-    ZcashDeserializeInto,
+    zcash_deserialize_bytes_external_count, CompactSize64, DateTime32, SerializationError,
+    TrustedPreallocate, ZcashDeserialize, ZcashDeserializeInto,
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


upbqdn (approved):
All looks great to me.

PR #6330, opened 2023-03-15
change(chain): Refactor the handling of height differences

### LINKED ISSUE #6279: [NCC-E005955-XVE] `zebra-chain`: Inconsistent error management in `Add` and `Sub` for `Height`
## Motivation

We want to track all of the findings from the zebra audit.

## Details

Some arithmetic operations for `Height` are implemented in [zebra-chain/src/block/height.rs](https://github.com/ZcashFoundation/zebra/blob/audit-v1.0.0-rc.0/zebra-chain/src/block/height.rs#L68-L100) for both `Height` and `i32`

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/block/height.rs#L68-L100

The `Add` function handles overflow with an `Option` , but `Sub` will panic. The `Sub` function later in the file for `i32` subtraction returns an `Option` as well, making the panic behavior an outlier. Panics should be used as a last resort, when there is no possibility of recovery. Otherwise, an attacker may attempt to intentionally trigger a panic as part of a denial-of-service attack. Additionally, the `Sub` function **does not** enforce the same constraint checks shown on Line 73 above, unlike the other functions in the same file.

upbqdn:
## Motivation

Close #6279. 

## Solution

This PR handles height differences similarly to `chrono::DateTime` and `chrono::Duration`.

### Reviewer Checklist


upbqdn:
Tests for GBT RPCs are failing. I forgot to run them locally. I'm working on that.

upbqdn:
> Tests for GBT RPCs are failing. I forgot to run them locally. I'm working on that.

Fixed.

zebra-chain/src/chain_tip.rs
```diff
@@ -95,26 +95,29 @@ pub trait ChainTip {
...
         let estimator =
             NetworkChainTipHeightEstimator::new(current_block_time, current_height, network);
 
-        Some((
-            estimator.estimate_height_at(Utc::now()) - current_height,
-            current_height,
-        ))
+        let distance_to_tip = estimator.estimate_height_at(Utc::now()) - current_height;
+
+        distance_to_tip.map(|distance_to_tip| (distance_to_tip, current_height))
```
arya2: Optional: I think the minimum estimated network height should be `current_height` so `distance_to_tip` would be a minimum of zero.

```suggestion
        Some((distance_to_tip.unwrap_or(block::Height(0)), current_height))
```

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,37 +25,35 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
         .activation_height(network)
         .expect("blossom activation height should be available");
 
-    if height < SLOW_START_SHIFT {
+    let slow_start_shift_diff = height - SLOW_START_SHIFT;
+    let slow_start_shift_diff = slow_start_shift_diff.unwrap_or_else(|| {
```
arya2: Optional: use `let .. else` statement here, i.e.

```suggestion
    let Some(slow_start_shift_diff) = height - SLOW_START_SHIFT else {
```

arya2 (dismissed):
This looks excellent!

Feel free to ignore the couple trivial suggestions.

zebra-chain/src/chain_tip.rs
upbqdn: The estimated network height can be lower than `current_height` when Zebra's local time is behind the chain tip block's timestamp. There's a comment describing this scenario above the function.

arya2 (dismissed):

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,37 +25,34 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
...
         unreachable!(
             "unsupported block height: checkpoints should handle blocks below {:?}",
             SLOW_START_SHIFT
         )
-    } else if height < blossom_height {
-        let pre_blossom_height: u32 = (height - SLOW_START_SHIFT)
-            .try_into()
-            .expect("height is above slow start, and the difference fits in u32");
-        let halving_shift = pre_blossom_height / PRE_BLOSSOM_HALVING_INTERVAL.0;
+    });
```
arya2:
```suggestion
    };
```

arya2 (dismissed):

zebra-chain/src/chain_tip.rs
teor2345: In some cases, the callers of this function use the tip height, even if the estimate is below the current tip. It's not clear to me if this PR breaks that functionality.

For example:
* the `getblocktemplate` RPC could return "No Chain tip available yet" if the estimate is below the current tip, but it should instead return a template
* the `getblockchaininfo.estimatedheight` field is [specified as the current best height if we've finished syncing]( https://zcash.github.io/rpc/getblockchaininfo.html). But instead we might return a "No Chain tip available yet" error for the whole RPC.

Are there existing unit tests for these two edge cases, or should we add them to this PR?

Can we return a struct with two named `Option<Height>` fields?
Now that the two tuple fields are the same type, it's easy for the caller to mix them up. And since the fields aren't named, reviewers will find it hard to pick up a mistake like that.

teor2345 (commented):
I'm concerned that this change might have unintended consequences, particularly for the RPCs that use the estimated height.

But I like the way that it simplifies the consensus code!

Do we need to do a full sync test or a manual mining test before we approve this PR? Or are the unit tests good enough?

teor2345:
Some background that might be helpful here:

> Panics should be used as a last resort, when there is no possibility of recovery. Otherwise, an attacker may attempt to intentionally trigger a panic as part of a denial-of-service attack.

The original `Sub` implementation can't actually panic with valid heights, because `Height::MIN - Height::MAX` is `i32::MIN + 1`. Zebra doesn't allow invalid heights to be generated, they are checked when they are deserialized.

> Additionally, the `Sub` function **does not** enforce the same constraint checks shown on Line 73 above, unlike the other functions in the same file.

`i32` is not a `Height`, so these checks are not required. When it's used in consensus-critical code as a `Height`, those checks are done in the consensus code.

But I think it's better to do those checks in the `Sub` impl, like this PR does, so the checks always get done correctly.

zebra-chain/src/block/height.rs
```diff
@@ -83,18 +83,19 @@ impl Add<Height> for Height {
...
-    fn sub(self, rhs: Height) -> i32 {
-        // We construct heights from integers without any checks,
-        // so the inputs or result could be out of range.
-        let lhs = i32::try_from(self.0)
-            .expect("out of range input `self`: inputs should be valid Heights");
-        let rhs =
-            i32::try_from(rhs.0).expect("out of range input `rhs`: inputs should be valid Heights");
-        lhs.checked_sub(rhs)
-            .expect("out of range result: valid input heights should yield a valid result")
+    type Output = Option<Height>;
```
teor2345: Should this be `Option<u32>` instead?

`Height`s in Zebra represent the height of a block in the blockchain, but this operation returns the difference between two heights. (It's like `DateTime` and `Duration`, where a duration is the difference between two times.)

The returned height might be a numerically correct height, but it doesn't actually represent a block. So using it instead of the height of a block could lead to consensus rule mistakes in future changes.

zebra-chain/src/chain_tip.rs
teor2345: I thought about this a bit more, and I think we're actually trying to represent four different outcomes here:
- Empty: the blockchain is empty
- Syncing: the `estimated_height` is above the `state_height`
- SyncedEqual: the estimated and `state_height` are equal
- SyncedAbove: the estimated height is above the `state_height`

Would it help to have an enum, or is that too complicated?

If we don't have any code that cares about the difference between the last two outcomes, we could represent those two outcomes in the same way, with just one enum variant and height field.

arya2: > * the `getblocktemplate` RPC could return "No Chain tip available yet" if the estimate is below the current tip, but it should instead return a template

This method would've panicked before this PR if it tried to subtract a bigger height from the estimated network height.

I still think a minimum of zero (i.e. `Some((distance_to_tip.unwrap_or(block::Height(0)), current_height))`) would work nicely here.

upbqdn: > Do we need to do a full sync test or a manual mining test before we approve this PR? Or are the unit tests good enough?

I can manually test both mining and a full sync before a merge.

> Are there existing unit tests for these two edge cases, or should we add them to this PR?

I don't think we have them. I'll try to add them.

> Would it help to have an enum, or is that too complicated?

The enum you outlined sounds good.

zebra-chain/src/block/height.rs
upbqdn: > It's like DateTime and Duration, where a duration is the difference between two times.

> Should this be Option<u32> instead?

I was also thinking about this, and considering wrapping `i32` into a new type, say, `HeightDiff`, so that we can keep negative diffs. I'd then return `Option<HeightDiff>` from the `Sub` impl.

I guess we can go with `Option<u32>` to express it's not height we're returning, but we'd still lose negative diffs.

My rationale was that returning `None` would "cover" all negative results as well as overflows and that `Height` would suffice, just as in https://github.com/zcashfoundation/zebra/blob/a97a9752d74c8505652de666cca3c67fa4ea7c81/zebra-chain/src/block/height.rs#L119-L120.

I don't really know what the ideal way to model this is. I'd go with `struct HeightDiff(i32)` for both `impl Sub<i32> for Height` and `impl Sub<Height> for Height`, but that sounds like overkill now.

zebra-chain/src/chain_tip.rs
teor2345: > > * the `getblocktemplate` RPC could return "No Chain tip available yet" if the estimate is below the current tip, but it should instead return a template
> 
> This method would've panicked before this PR if it tried to subtract a bigger height from the estimated network height.

I'm not sure where the panic is, can you explain it to me?

I think I've seen the estimate below the tip before, but I'm not sure.

> I still think a minimum of zero (i.e. `Some((distance_to_tip.unwrap_or(block::Height(0)), current_height))`) would work nicely here.

This code is not consensus-critical, so it doesn't really matter what we do here. But is there any RPC specification, RPC error, or any other code that cares about the difference between a negative and a zero distance to the tip?

zebra-chain/src/block/height.rs
teor2345: I think it depends on what the requirements are. If we don't have any code that uses negative `HeightDiff`s, then we could just return `None` for them. What do the existing callers of this code need?

We can always change it later if we need something more complicated.

I also think that some RPCs might manually construct heights that are above `Height::MAX`. We should check they aren't creating invalid heights.

We could also open a ticket to make the `u32` inside `Height` a private field, because until we do that, we can write code that constructs heights above `Height::MAX`, which can then cause panics later. But that's a huge change, which seems larger than what we need to fix this issue.

Here's one possible minor change to the implementation:
```rust
pub type HeightDiff = i32;

// Delete impl Add<Height> for Height
// because it doesn't actually make sense to add two heights

// Do we want to return `None` rather than panicking, because other code can construct invalid heights?
// Do we want to return `None` rather than negative values, because they aren't needed?
// If we have two different invalid outcomes, do we want to return a `Result` instead?
impl Sub<Height> for Height {
    type Output = Option<HeightDiff>;
    ...
}

// This is the existing impl, but using the type alias
impl Add<HeightDiff> for Height {
    type Output = Option<Height>;
    ...
}

// This is the existing impl, but using the type alias
impl Sub<HeightDiff> for Height {
    type Output = Option<Height>;
    ...
}
```

We might also want to test anything that gets externally provided heights with inputs near `Height::MAX`:
* syncer
* inbound block downloader
* mempool
* transaction verifier
* RPCs

Edit: I think we might already have tests for these services, or they don't actually use this code. But let's double-check?

zebra-chain/src/chain_tip.rs
arya2: > I'm not sure where the panic is, can you explain it to me?

It *was* [here](https://github.com/ZcashFoundation/zebra/pull/6330/files#diff-929cb1ee6ff0e87f011f57a1a5f68132642202776e071648e4138ca28ce032d2L96-L97), where a `Height` was being subtracted from a possibly lower `Height` in this `estimate_distance_to_network_chain_tip()` method.

teor2345: `i32::checked_sub()` does not panic or return `None` on negative results, it only returns `None` on overflow/underflow.

The current range of `Height` makes sure there is no overflow/underflow, except for manually constructed invalid heights. (And in that case, the bug is in the code that constructs the height.)

mpguerra:
What's the status with this PR? Is it still relevant/valid given https://github.com/ZcashFoundation/zebra/issues/6279#issuecomment-1478669249 ?

upbqdn:
I have a refactor that addresses the comments almost ready to be pushed.

upbqdn:
I pushed the new refactor. I also changed the PR name and description.

zebrad/src/components/sync/progress.rs
```diff
@@ -129,7 +129,7 @@ pub async fn show_block_chain_progress(
             );
 
             let remaining_sync_blocks =
-                (estimated_height - current_height).map_or(0, |height| height.0);
+                (estimated_height - current_height).map_or(0, |height| height);
```
arya2:
```suggestion
                (estimated_height - current_height).unwrap_or(0);
```

zebra-chain/src/block/height.rs
```diff
@@ -64,69 +64,60 @@ impl Height {
...
-            _ => None,
+        let res = lhs.checked_add(rhs)?;
+        let res = u32::try_from(res).ok()?;
+        let height = Height(res);
+
+        // Check the bounds.
+        if Height::MIN <= height && height <= Height::MAX {
+            Some(height)
+        } else {
+            None
```
arya2: Optional: `TryFrom<u32 | i32>` impls for `Height` that check these bounds would be nice.

zebra-chain/src/block/height.rs
```diff
@@ -191,13 +182,12 @@ fn operator_tests() {
...
-    assert_eq!(Height::MAX_AS_U32 as i32, Height::MAX - Height(0));
-    assert_eq!(1, Height::MAX - Height(Height::MAX_AS_U32 - 1));
-    assert_eq!(-1, Height(Height::MAX_AS_U32 - 1) - Height::MAX);
-    assert_eq!(-(Height::MAX_AS_U32 as i32), Height(0) - Height::MAX);
+    assert_eq!(1, (Height(2) - Height(1)).unwrap());
+    assert_eq!(0, (Height(1) - Height(1)).unwrap());
+    assert_eq!(Some(-1), Height(0) - Height(1));
+    assert_eq!(Some(-5), Height(2) - Height(7));
+    assert_eq!(Height::MAX, (Height::MAX - 0).unwrap());
+    assert_eq!(1, (Height::MAX - Height(Height::MAX_AS_U32 - 1)).unwrap());
```
arya2: Nitpick/Optional:

```suggestion
    assert_eq!(Some(1), (Height(2) - Height(1)));
    assert_eq!(Some(0), (Height(1) - Height(1)));
    assert_eq!(Some(-1), Height(0) - Height(1));
    assert_eq!(Some(-5), Height(2) - Height(7));
    assert_eq!(Some(Height::MAX), (Height::MAX - 0));
    assert_eq!(Some(1), (Height::MAX - Height(Height::MAX_AS_U32 - 1)));
```

zebra-chain/src/chain_tip.rs
arya2: > Are there existing unit tests for these two edge cases, or should we add them to this PR?

There aren't any existing unit tests that I'm aware of, but I think we should add those in a different PR.

zebra-chain/src/chain_tip/network_chain_tip_height_estimator.rs
```diff
@@ -87,15 +87,17 @@ impl NetworkChainTipHeightEstimator {
...
+            remaining_blocks.is_positive().then(|| {
+                let remaining_blocks = i64::from(remaining_blocks);
+                let target_spacing_seconds = self.current_target_spacing.num_seconds();
+                let time_to_activation =
+                    Duration::seconds(remaining_blocks * target_spacing_seconds);
+
+                self.current_block_time += time_to_activation;
+                self.current_height = max_height;
+            })
+        });
```
arya2: Nitpick/Optional:

```suggestion
        if let Some(remaining_blocks) =
            (max_height - self.current_height).filter(|r| r.is_positive())
        {
            let remaining_blocks = i64::from(remaining_blocks);
            let target_spacing_seconds = self.current_target_spacing.num_seconds();
            let time_to_activation = Duration::seconds(remaining_blocks * target_spacing_seconds);

            self.current_block_time += time_to_activation;
            self.current_height = max_height;
        }
```

zebra-chain/src/chain_tip/tests/prop.rs
```diff
@@ -72,7 +72,7 @@ fn estimate_time_difference(
 ) -> Duration {
     let spacing_seconds = active_network_upgrade.target_spacing().num_seconds();
 
-    let height_difference = i64::from(end_height - start_height);
+    let height_difference = (end_height - start_height).map_or(0, |height| height.into());
```
arya2: Nitpick/Optional:

```suggestion
    let height_difference = (end_height - start_height).map_or(0, i64::from);
```

zebra-state/src/service/check/utxo.rs
```diff
@@ -200,7 +200,7 @@ pub fn transparent_coinbase_spend(
     match spend_restriction {
         OnlyShieldedOutputs { spend_height } => {
             let min_spend_height =
-                utxo.utxo.height + block::Height(MIN_TRANSPARENT_COINBASE_MATURITY);
+                utxo.utxo.height + MIN_TRANSPARENT_COINBASE_MATURITY.try_into().unwrap();
```
arya2, line 203: Optional: It would be nice for `MIN_TRANSPARENT_COINBASE_MATURITY` to be a `HeightDiff` too.

arya2 (dismissed):
This looks great, thank you!

I left some suggestions, all are optional.

zebra-state/src/service/check/utxo.rs
teor2345: I think this might cause issues with the type in other code, so that seems like something for a later cleanup after the audit.

zebra-chain/src/block/height.rs
teor2345: We can probably only implement one of these, because if we implement both, it makes type inference impossible on integer constants.

zebra-chain/src/block/height.rs
```diff
@@ -64,69 +64,60 @@ impl Height {
...
+        // [`None`] for negative results. We must check the conversion since it's possible to
+        // construct heights outside the valid range for [`i32`].
+        let lhs = i32::try_from(self.0).ok()?;
+        let rhs = i32::try_from(rhs.0).ok()?;
         lhs.checked_sub(rhs)
-            .expect("out of range result: valid input heights should yield a valid result")
     }
 }
 
-// We don't implement Add<u32> or Sub<u32>, because they cause type inference issues for integer constants.
```
teor2345, line 101: This note is important and I'd like to keep it

zebra-chain/src/chain_tip.rs
```diff
@@ -95,26 +95,29 @@ pub trait ChainTip {
...
+    ///
+    /// The first element in the returned tuple is the estimate.
+    /// The second element in the returned tuple is the current best chain tip.
     ///
     /// The estimate is calculated based on the current local time, the block time of the best tip
     /// and the height of the best tip.
     ///
-    /// This estimate may be negative if the current local time is behind the chain tip block's timestamp.
+    /// This estimate may be negative if the current local time is behind the chain tip block's
+    /// timestamp.
```
teor2345, line 108: Can you please document when this method returns None?

zebra-consensus/src/block/check.rs
```diff
@@ -154,7 +154,7 @@ pub fn subsidy_is_valid(block: &Block, network: Network) -> Result<(), BlockErro
         .activation_height(network)
         .expect("Canopy activation height is known");
 
-    if height < SLOW_START_INTERVAL {
+    if (height - SLOW_START_INTERVAL).is_none() {
```
teor2345: The original comparison seems clearer here, is there a reason to also reject heights over i32::MAX?

Edit: Feel free to `impl PartialOrd` and `impl Ord` if needed. (And if it makes sense for the types.)

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -74,7 +74,7 @@ fn funding_stream_address_period(height: Height, network: Network) -> u32 {
     // - In Rust, "integer division rounds towards zero":
     //   https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#arithmetic-and-logical-binary-operators
     //   This is the same as `floor()`, because these numbers are all positive.
-    (height.0 + (POST_BLOSSOM_HALVING_INTERVAL.0) - (height_for_first_halving(network).0))
+    (height.0 + (POST_BLOSSOM_HALVING_INTERVAL as u32) - (height_for_first_halving(network).0))
```
teor2345: Optional: I'd prefer we use try_into() here, because `as` can do some unexpected conversions and wrapping.

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,38 +25,36 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
...
-        // Some far-future shifts can be more than 63 bits
-        1u64.checked_shl(halving_shift)
-    }
+    let post_blossom_height_diff = (height - blossom_height)?;
+
+    post_blossom_height_diff
+        .is_negative()
+        .then(|| {
+            let pre_blossom_height = height_after_slow_start_shift.0;
+            let halving_shift = pre_blossom_height / (PRE_BLOSSOM_HALVING_INTERVAL as u32);
```
teor2345: Optional: I'd prefer we avoid `as`, see above.

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,38 +25,36 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
...
+                .expect("pre-blossom heights produce small shifts")
+        })
+        .or_else(|| {
+            let pre_blossom_height = (blossom_height - SLOW_START_SHIFT)
+                .expect("Blossom height must be above SlowStartShift.")
+                .0;
+            let scaled_pre_blossom_height = pre_blossom_height * BLOSSOM_POW_TARGET_SPACING_RATIO;
+
+            let halving_shift = (scaled_pre_blossom_height + (post_blossom_height_diff as u32))
+                / (POST_BLOSSOM_HALVING_INTERVAL as u32);
```
teor2345: Optional: I'd prefer we avoid `as`, see above.

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,38 +25,36 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
...
+            1u64.checked_shl(halving_shift)
+                .expect("pre-blossom heights produce small shifts")
+        })
+        .or_else(|| {
+            let pre_blossom_height = (blossom_height - SLOW_START_SHIFT)
+                .expect("Blossom height must be above SlowStartShift.")
+                .0;
+            let scaled_pre_blossom_height = pre_blossom_height * BLOSSOM_POW_TARGET_SPACING_RATIO;
+
+            let halving_shift = (scaled_pre_blossom_height + (post_blossom_height_diff as u32))
```
teor2345: This `as` is particularly risky because it operates on a number that's not a constant.

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -25,38 +25,36 @@ pub fn halving_divisor(height: Height, network: Network) -> Option<u64> {
...
+        .is_negative()
+        .then(|| {
+            let pre_blossom_height = height_after_slow_start_shift.0;
+            let halving_shift = pre_blossom_height / (PRE_BLOSSOM_HALVING_INTERVAL as u32);
+
+            1u64.checked_shl(halving_shift)
+                .expect("pre-blossom heights produce small shifts")
+        })
+        .or_else(|| {
+            let pre_blossom_height = (blossom_height - SLOW_START_SHIFT)
```
teor2345: Isn't this just `height_after_slow_start_shift.0` now?

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -75,7 +73,7 @@ pub fn block_subsidy(height: Height, network: Network) -> Result<Amount<NonNegat
         return Ok(Amount::zero());
     };
 
-    if height < SLOW_START_INTERVAL {
+    if (height - SLOW_START_INTERVAL).is_none() {
```
teor2345: I think the original comparison is clearer here.

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -4,19 +4,23 @@ use std::collections::HashMap;
...
+    amount::COIN,
+    block::{Height, HeightDiff},
+    parameters::Network,
+};
 
 /// An initial period from Genesis to this Height where the block subsidy is gradually incremented. [What is slow-start mining][slow-mining]
 ///
 /// [slow-mining]: https://z.cash/support/faq/#what-is-slow-start-mining
-pub const SLOW_START_INTERVAL: Height = Height(20_000);
+pub const SLOW_START_INTERVAL: HeightDiff = 20_000;
```
teor2345: This seems more like a Height, in practice it's the maximum height that the slow start was active until.

it's never used as a difference or a repeated interval.

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -4,19 +4,23 @@ use std::collections::HashMap;
...
 
 /// `SlowStartShift()` as described in [protocol specification §7.8][7.8]
 ///
 /// [7.8]: https://zips.z.cash/protocol/protocol.pdf#subsidies
 ///
 /// This calculation is exact, because `SLOW_START_INTERVAL` is divisible by 2.
-pub const SLOW_START_SHIFT: Height = Height(SLOW_START_INTERVAL.0 / 2);
+pub const SLOW_START_SHIFT: HeightDiff = SLOW_START_INTERVAL / 2;
```
teor2345: This also seems like a Height, for similar reasons to SLOW_START_INTERVAL.

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -33,11 +37,11 @@ pub const BLOSSOM_POW_TARGET_SPACING_RATIO: u32 = 2;
...
 ///
 /// `(60 * 60 * 24 * 365 * 4) / 150 = 840960`
-pub const PRE_BLOSSOM_HALVING_INTERVAL: Height = Height(840_000);
+pub const PRE_BLOSSOM_HALVING_INTERVAL: HeightDiff = 840_000;
 
 /// After Blossom the block time is reduced to 75 seconds but halving period should remain around 4 years.
-pub const POST_BLOSSOM_HALVING_INTERVAL: Height =
-    Height(PRE_BLOSSOM_HALVING_INTERVAL.0 * BLOSSOM_POW_TARGET_SPACING_RATIO);
+pub const POST_BLOSSOM_HALVING_INTERVAL: HeightDiff =
+    PRE_BLOSSOM_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as i32);
```
teor2345: Optional: this would be clearer.

```suggestion
    PRE_BLOSSOM_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff);
```

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -134,7 +138,7 @@ lazy_static! {
 ///
 /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
 pub const FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL: Height =
-    Height(POST_BLOSSOM_HALVING_INTERVAL.0 / 48);
+    Height(POST_BLOSSOM_HALVING_INTERVAL as u32 / 48);
```
teor2345: Optional: This seems like it's a HeightDiff, but we can make that change in another PR if you want.

teor2345 (commented):
Looks great!

My comments are mainly nitpicks, I've tried to call out the one or two cases where the new code is particularly unclear or risky.

mpguerra:
@upbqdn is going to be out for this week. Do we want to do anything with this PR in the meantime or should it wait until he gets back?

teor2345:
> @upbqdn is going to be out for this week. Do we want to do anything with this PR in the meantime or should it wait until he gets back?

@arya2 and I have enough context to finish it off, the remaining comments are mainly tweaks. I think it would be good to finish it to avoid merge conflicts with other work.

zebra-consensus/src/block/subsidy/general.rs
teor2345: Ah no, I'm wrong.

zebra-chain/src/block/height.rs
teor2345: 53115ab5a3 does the u32 version

zebra-chain/src/block/height.rs
teor2345: 06d7914507

zebra-chain/src/chain_tip/network_chain_tip_height_estimator.rs
teor2345: 976316ae90

zebra-chain/src/chain_tip.rs
teor2345: f1d0921956

zebra-chain/src/block/height.rs
teor2345: f834ae1c6c

zebra-consensus/src/block/check.rs
teor2345: 155ce35d1d

zebra-consensus/src/block/subsidy/funding_streams.rs
teor2345: 155ce35d1d

zebra-consensus/src/block/subsidy/general.rs
teor2345: 155ce35d1d

zebra-consensus/src/block/subsidy/general.rs
teor2345: 155ce35d1d

zebra-consensus/src/block/subsidy/general.rs
teor2345: 155ce35d1d

zebra-state/src/service/check/utxo.rs
teor2345: I added a TODO for this.

zebra-chain/src/chain_tip/network_chain_tip_height_estimator.rs
teor2345: And also extra changes in 4ad832e2e4

zebra-chain/src/chain_tip/tests/prop.rs
teor2345: Turns out this is unnecessary after 4ad832e2e4

zebra-consensus/src/block/subsidy/general.rs
teor2345: 155ce35d1d

zebra-consensus/src/parameters/subsidy.rs
teor2345: 155ce35d1d

zebra-consensus/src/parameters/subsidy.rs
teor2345: 155ce35d1d

zebra-consensus/src/parameters/subsidy.rs
teor2345: a765be43bf

zebra-consensus/src/block/subsidy/general.rs
teor2345: Turns out this is unnecessary with a765be43bf

zebra-consensus/src/parameters/subsidy.rs
teor2345: 155ce35d1d + 3238412dca

teor2345:
It turns out if we make `HeightDiff` an `i64`, a whole lot of `expect()`, `as`, and other tricky interim checks go away. See a765be43bf & 7ac8b89d13.

(We still have to check at the end of each calculation.)

teor2345:
@arya2 this is ready for a re-review, the changes were a bit more substantial than I thought, but they helped remove a lot of code that's tricky to review.

I can do a video review if that would help?

zebra-chain/src/block/height.rs
```diff
@@ -65,60 +65,70 @@ impl Height {
...
+/// even if they are outside the valid height range (for example, in buggy RPC code).
+pub type HeightDiff = i64;
+
+impl TryFrom<u32> for Height {
+    type Error = &'static str;
+
+    /// Checks that the `height` is within the valid [`Height`] range.
+    fn try_from(height: u32) -> Result<Self, Self::Error> {
+        // Check the bounds.
+        if Height::MIN.0 <= height && height <= Height::MAX.0 {
```
arya2, line 79: Nitpick/Optional: the first part of this condition will always be true for a u32

```suggestion
        if height <= Height::MAX.0 {
```

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -69,13 +69,23 @@ pub fn height_for_first_halving(network: Network) -> Height {
...
-        / (FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL.0)
+
+    let height_after_first_halving = height - height_for_first_halving(network);
+
+    let address_period = (height_after_first_halving + POST_BLOSSOM_HALVING_INTERVAL)
+        / FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL;
+
+    address_period
+        .try_into()
+        .expect("all values are positive and smaller than the input height")
```
arya2, line 88: Nitpick/Optional: Could we make this look more like the spec

```suggestion
    ((height - (height_for_first_halving(network) - POST_BLOSSOM_HALVING_INTERVAL))
        / FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL)
        .try_into()
        .expect("all values are positive and smaller than the input height")
```

arya2 (approved):
This looks great!

I left a couple optional suggestions, feel free to ignore either or both of them.

zebra-chain/src/block/height.rs
teor2345: Yep, but we want it in there in case we change the underlying type.

zebra-consensus/src/block/subsidy/funding_streams.rs
teor2345: I'm not sure if the types work out this way, and I'm not too concerned about matching the form of the spec equations.

PR #6335, opened 2023-03-15
change(state): Stop re-downloading blocks that are in non-finalized side chains

### LINKED ISSUE #862: Make sync, inbound, and block verifier check if a block hash is in any chain or any queue
## Motivation

In the sync service (#853), inbound downloader, and block verifier, we are using the GetDepth request to check if a block is present in the state:
https://github.com/ZcashFoundation/zebra/pull/853/files#r467684617

But the sync, inbound, and block verifier actually need to know if a block hash is present in any chain. (This is a bug.)

They also don't need a depth, or any other extra data.

### Scheduling

This risk is acceptable for the stable release, but we need to fix it before we support lightwalletd.

We should also fix this bug if Zebra continues to hang, after we fix known hang bugs.

## Solution

Each section should be implemented in a separate PR.

A. Add a new "contains" block hash request:
- [ ] add an additional state service request, which checks if a block is in:
  - [ ] any chain
  - [ ] the pending non-finalized state queue
  - [ ] the pending finalized state queue
- [ ] use it in the sync service to ignore block downloads (or check for genesis)
- [ ] use it in the inbound downloader to ignore block downloads
- [ ] make it a `HashSet` of block hashes, to reduce the number of state requests

B. We want to do these checks:
- [ ] before downloading the block, and
- [ ] after the download, but before verifying it (#2327)
  - See PR #2327 for a draft of this change

For example, see this code:
https://github.com/ZcashFoundation/zebra/blob/ebe1c9f88ecc44570ee335c3e8a25bcb53672acd/zebra-state/src/service.rs#L99-L104

D. Check if the blocks are already waiting in a checkpoint verifier queue:
- [ ] change the request type for the checkpoint verifier into an enum
- [ ] check if blocks are already waiting in the checkpoint verifier queue

E. De-duplicate the `sync` and `inbound` block downloaders:
- [ ] Use an `enum` to distinguish `sync` and `inbound` requests
- [ ] Don't apply the `inbound` limits to `sync` blocks
- [ ] Use a buffer or channels to get responses back to the right service?

## Alternatives

We could implement the new request as a wrapper function around a more specific query, which discards any extra data. But there aren't any requests that find blocks in any chain.

We might also want to return success on duplicate blocks, rather than an error. But sync restarts should be even rarer once we fix this bug.

## Related

The state request for this could support [more detailed errors for the submitblock method (#5487)](#5487) if it specifies where the block is in the state (i.e. best chain, side chain, or queued).

arya2:
## Motivation

The sync task, inbound downloader, and the block verifier need to check if a block hash is present in the state to avoid redundant work and unnecessary errors. They currently call the state service with `Depth` requests, but this only checks if a block hash has been committed to the best chain, and returns `None` when the block is in a side chain or queued for validation.

This PR adds a `Contains` request to the state for checking if a block hash is present in any chain and uses it from sync, inbound and the block verifier instead of `Depth`.

Closes #862.

## Solution

- Add a `Contains` request to the state service that responds with an `Option<BlockLocation>`
  - Checks if the block hash is in:
    - non-finalized best chain
    - non-finalized side chains
    - finalized best chain
- Call state with `Contains` request from sync, inbound, and block verifier to check if the block hash is in any chain

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -159,7 +159,7 @@ where
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
                 zs::Response::KnownBlock(Some(location)) => {
-                    return Err(BlockError::AlreadyInState(hash, location).into())
+                    return Err(BlockError::AlreadyInChain(hash, location).into())
                 }
                 zs::Response::KnownBlock(None) => {}
                 _ => unreachable!("wrong response to Request::KnownBlock"),
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -154,7 +154,7 @@ where
                 .ready()
                 .await
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
-                .call(zs::Request::Contains(hash))
+                .call(zs::Request::KnownBlock(hash))
                 .await
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -159,7 +159,7 @@ where
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
                 zs::Response::KnownBlock(Some(location)) => {
-                    return Err(BlockError::AlreadyInChain(hash, location).into())
+                    return Err(BlockError::AlreadyInState(hash, location).into())
                 }
                 zs::Response::KnownBlock(None) => {}
                 _ => unreachable!("wrong response to Request::Depth"),
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -154,14 +154,14 @@ where
                 .ready()
                 .await
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
-                .call(zs::Request::Depth(hash))
+                .call(zs::Request::Contains(hash))
                 .await
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
-                zs::Response::Depth(Some(depth)) => {
-                    return Err(BlockError::AlreadyInChain(hash, depth).into())
+                zs::Response::BlockLocation(Some(location)) => {
+                    return Err(BlockError::AlreadyInChain(hash, location).into())
                 }
-                zs::Response::Depth(None) => {}
+                zs::Response::BlockLocation(None) => {}
                 _ => unreachable!("wrong response to Request::Depth"),
             }
 
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -158,10 +158,10 @@ where
                 .await
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
-                zs::Response::BlockLocation(Some(location)) => {
+                zs::Response::KnownBlock(Some(location)) => {
                     return Err(BlockError::AlreadyInChain(hash, location).into())
                 }
-                zs::Response::BlockLocation(None) => {}
+                zs::Response::KnownBlock(None) => {}
                 _ => unreachable!("wrong response to Request::Depth"),
             }
 
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -162,7 +162,7 @@ where
                     return Err(BlockError::AlreadyInState(hash, location).into())
                 }
                 zs::Response::KnownBlock(None) => {}
-                _ => unreachable!("wrong response to Request::Depth"),
+                _ => unreachable!("wrong response to Request::KnownBlock"),
             }
 
             tracing::trace!("performing block checks");
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -246,7 +246,7 @@ pub enum BlockError {
     DuplicateTransaction,
 
     #[error("block {0:?} is already in present in the state {1:?}")]
-    AlreadyInState(zebra_chain::block::Hash, zebra_state::KnownBlock),
+    AlreadyInChain(zebra_chain::block::Hash, zebra_state::KnownBlock),
 
     #[error("invalid block {0:?}: missing block height")]
     MissingHeight(zebra_chain::block::Hash),
@@ -311,6 +311,6 @@ impl BlockError {
     /// Returns `true` if this is definitely a duplicate request.
     /// Some duplicate requests might not be detected, and therefore return `false`.
     pub fn is_duplicate_request(&self) -> bool {
-        matches!(self, BlockError::AlreadyInState(..))
+        matches!(self, BlockError::AlreadyInChain(..))
     }
 }
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -246,7 +246,7 @@ pub enum BlockError {
     DuplicateTransaction,
 
     #[error("block {0:?} is already in the chain at depth {1:?}")]
-    AlreadyInChain(zebra_chain::block::Hash, u32),
+    AlreadyInChain(zebra_chain::block::Hash, zebra_state::BlockLocation),
 
     #[error("invalid block {0:?}: missing block height")]
     MissingHeight(zebra_chain::block::Hash),
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -246,7 +246,7 @@ pub enum BlockError {
     DuplicateTransaction,
 
     #[error("block {0:?} is already in the chain at depth {1:?}")]
-    AlreadyInChain(zebra_chain::block::Hash, zebra_state::BlockLocation),
+    AlreadyInChain(zebra_chain::block::Hash, zebra_state::KnownBlock),
 
     #[error("invalid block {0:?}: missing block height")]
     MissingHeight(zebra_chain::block::Hash),
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -245,8 +245,8 @@ pub enum BlockError {
     #[error("block contains duplicate transactions")]
     DuplicateTransaction,
 
-    #[error("block {0:?} is already in the chain at depth {1:?}")]
-    AlreadyInChain(zebra_chain::block::Hash, zebra_state::KnownBlock),
+    #[error("block {0:?} is already in present in the state {1:?}")]
+    AlreadyInState(zebra_chain::block::Hash, zebra_state::KnownBlock),
 
     #[error("invalid block {0:?}: missing block height")]
     MissingHeight(zebra_chain::block::Hash),
@@ -311,6 +311,6 @@ impl BlockError {
     /// Returns `true` if this is definitely a duplicate request.
     /// Some duplicate requests might not be detected, and therefore return `false`.
     pub fn is_duplicate_request(&self) -> bool {
-        matches!(self, BlockError::AlreadyInChain(..))
+        matches!(self, BlockError::AlreadyInState(..))
     }
 }
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -641,7 +641,7 @@ impl Request {
             }
             Request::BestChainNextMedianTimePast => "best_chain_next_median_time_past",
             Request::BestChainBlockHash(_) => "best_chain_block_hash",
-            Request::Contains(_) => "contains",
+            Request::KnownBlock(_) => "known_block",
             #[cfg(feature = "getblocktemplate-rpcs")]
             Request::CheckBlockProposalValidity(_) => "check_block_proposal_validity",
         }
@@ -955,7 +955,7 @@ impl TryFrom<Request> for ReadRequest {
                      Manually convert the request to ReadRequest::AnyChainUtxo, \
                      and handle pending UTXOs"),
 
-            Request::Contains(_) => Err("ReadService does not track queued blocks"),
+            Request::KnownBlock(_) => Err("ReadService does not track queued blocks"),
 
             #[cfg(feature = "getblocktemplate-rpcs")]
             Request::CheckBlockProposalValidity(prepared) => {
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -607,11 +607,12 @@ pub enum Request {
     /// * [`Response::BlockHash(None)`](Response::BlockHash) otherwise.
     BestChainBlockHash(block::Height),
 
-    /// Checks if a block is present anywhere in the state service; Looks up hash in block stores.
+    /// Checks if a block is present anywhere in the state service.
+    /// Looks up `hash` in block queues as well as the finalized chain and all non-finalized chains.
     ///
     /// Returns [`Response::BlockLocation(Some(Location))`](Response::BlockLocation) if the block is in the best state service.
     /// Returns [`Response::BlockLocation(None)`](Response::BlockLocation) otherwise.
-    Contains(block::Hash),
+    KnownBlock(block::Hash),
 
     #[cfg(feature = "getblocktemplate-rpcs")]
     /// Performs contextual validation of the given block, but does not commit it to the state.
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -607,6 +607,12 @@ pub enum Request {
     /// * [`Response::BlockHash(None)`](Response::BlockHash) otherwise.
     BestChainBlockHash(block::Height),
 
+    /// Checks if a block is present anywhere in the state service; Looks up hash in block stores.
+    ///
+    /// Returns [`Response::BlockLocation(Some(Location))`](Response::BlockLocation) if the block is in the best state service.
+    /// Returns [`Response::BlockLocation(None)`](Response::BlockLocation) otherwise.
+    Contains(block::Hash),
+
     #[cfg(feature = "getblocktemplate-rpcs")]
     /// Performs contextual validation of the given block, but does not commit it to the state.
     ///
@@ -634,6 +640,7 @@ impl Request {
             }
             Request::BestChainNextMedianTimePast => "best_chain_next_median_time_past",
             Request::BestChainBlockHash(_) => "best_chain_block_hash",
+            Request::Contains(_) => "contains",
             #[cfg(feature = "getblocktemplate-rpcs")]
             Request::CheckBlockProposalValidity(_) => "check_block_proposal_validity",
         }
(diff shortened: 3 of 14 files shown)


zebra-consensus/src/error.rs
```diff
@@ -246,7 +246,7 @@ pub enum BlockError {
     DuplicateTransaction,
 
     #[error("block {0:?} is already in the chain at depth {1:?}")]
```
upbqdn: We should change this error message since we're no longer using depth.

zebra-state/src/service/queued_blocks.rs
```diff
@@ -286,6 +292,17 @@ impl SentHashes {
         self.update_metrics_for_block(block.height);
     }
 
+    /// Stores the finalized `block`'s hash, so it can be used to check if a
+    /// block is in the state queue.
+    ///
+    /// For more details see `add()`.
+    pub fn add_finalized_hash(&mut self, hash: block::Hash, height: block::Height) {
```
upbqdn: The docs for `curr_buf` say the blocks are in increasing height order, but none of the functions that push to it enforce the order. I'm probably missing something here.

zebra-state/src/service.rs
```diff
@@ -491,6 +485,10 @@ impl StateService {
             }
 
             self.drain_queue_and_commit_finalized();
+
+            if let Some(finalized_tip_height) = self.read_service.db.finalized_tip_height() {
+                self.sent_blocks.prune_by_height(finalized_tip_height);
```
upbqdn: `prune_by_height` calls `finish_batch`. Doesn't that imply that batches are effectively single blocks?

upbqdn (commented):
Some sections mentioned in #862 are not in this PR or are outdated. I think we should update #862 before we merge this PR.

zebra-state/src/service.rs
upbqdn: It doesn't. Sorry for the confusion.

zebra-state/src/request.rs
```diff
@@ -607,6 +607,12 @@ pub enum Request {
     /// * [`Response::BlockHash(None)`](Response::BlockHash) otherwise.
     BestChainBlockHash(block::Height),
 
+    /// Checks if a block is present anywhere in the state service; Looks up hash in block stores.
```
```suggestion
    /// Checks if a block is present anywhere in the state service.
    /// Looks up `hash` in block queues as well as the finalized chain and all non-finalized chains.
```

zebra-state/src/request.rs
```diff
@@ -607,6 +607,12 @@ pub enum Request {
...
+    ///
+    /// Returns [`Response::BlockLocation(Some(Location))`](Response::BlockLocation) if the block is in the best state service.
+    /// Returns [`Response::BlockLocation(None)`](Response::BlockLocation) otherwise.
+    Contains(block::Hash),
```
teor2345: Can we rename this request so it's clear that the block is known, but might not be valid?
(Also it would be helpful to have "block" in the name!)

```suggestion
    KnownBlock(block::Hash),
```

zebra-state/src/response.rs
```diff
@@ -69,11 +69,27 @@ pub enum Response {
     /// specified block hash.
     BlockHash(Option<block::Hash>),
 
+    /// Response to [`Request::Contains`].
+    BlockLocation(Option<BlockLocation>),
```
teor2345: Can we rename this response and enum? It is very similar to "block locator", which is part of the network protocol and state service.

We could use `KnownBlock` instead?

zebra-state/src/response.rs
```diff
@@ -69,11 +69,27 @@ pub enum Response {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+/// An enum of block stores in the state where a block hash could be found.
+pub enum BlockLocation {
+    /// Block is in the best chain.
+    BestChain,
+
+    /// Block is in a side chain.
+    SideChain,
+
+    /// Block is queued to be validated and committed.
```
teor2345:
```suggestion
    /// Block is queued to be validated and committed, or rejected and dropped.
```

zebra-state/src/service/queued_blocks.rs
teor2345: I think this is implemented by `drain_queue_and_commit_finalized()`, because parent blocks are added before their child blocks. (And the checkpoint verifier only sends one block per height.)

For the non-finalized state, `send_ready_non_finalized_queued()` implements the same rule, but there can be multiple child blocks for the one parent block. (This forks the chain if the blocks are valid.)

arya2: > I think this is implemented by drain_queue_and_commit_finalized(), because parent blocks are added before their child blocks.

Yes.

Updated the doc comments in [`de9730f`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/de9730f08db982e1c952d1b3bf78e2c012657f9e).

zebra-consensus/src/error.rs
arya2: Updated in [`348fadc`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/348fadc9c46c34e773cce9ce12e72590b8df88a6).

zebra-state/src/response.rs
arya2: Yeah, renamed in [`7bf56da`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/7bf56da36c09487803f890fb0a9a399085ac3b8f).

zebra-consensus/src/block.rs
```diff
@@ -154,14 +154,14 @@ where
...
                 .map_err(|source| VerifyBlockError::Depth { source, hash })?
             {
-                zs::Response::Depth(Some(depth)) => {
-                    return Err(BlockError::AlreadyInChain(hash, depth).into())
+                zs::Response::KnownBlock(Some(location)) => {
+                    return Err(BlockError::AlreadyInState(hash, location).into())
                 }
-                zs::Response::Depth(None) => {}
+                zs::Response::KnownBlock(None) => {}
                 _ => unreachable!("wrong response to Request::Depth"),
```
teor2345:
```suggestion
                _ => unreachable!("wrong response to Request::KnownBlock"),
```

zebra-state/src/service.rs
```diff
@@ -162,7 +163,7 @@ pub(crate) struct StateService {
 
     /// A set of non-finalized block hashes that have been sent to the block write task.
     /// Hashes of blocks below the finalized tip height are periodically pruned.
-    sent_non_finalized_block_hashes: SentHashes,
+    sent_blocks: SentHashes,
```
teor2345: I'm wondering if this field rename also means we need to update its comment?

arya2: Whoops, fixed in [`b325a4e`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/b325a4eb05335f38420ba34a9e5f132cedd9d2e0).

zebra-state/src/response.rs
```diff
@@ -69,11 +69,27 @@ pub enum Response {
...
+/// An enum of block stores in the state where a block hash could be found.
+pub enum KnownBlock {
+    /// Block is in the best chain.
+    BestChain,
+
+    /// Block is in a side chain.
+    SideChain,
+
+    /// Block is queued to be validated and committed, or rejected and dropped.
+    Queue,
```
teor2345, line 90: I think that we might accidentally label some non-finalized blocks as `Queued`?
This isn't important for this PR, but we should document it to avoid surprises later.

```suggestion
    /// Block is queued to be validated and committed, or rejected and dropped.
    /// Some non-finalized blocks might also appear as `Queued` until the sent blocks have been pruned.
    Queue,
```

teor2345: Oh hang on let me check the rest of the code.

teor2345: I think we can fix this in the code.

zebra-state/src/service.rs
```diff
@@ -1063,6 +1072,32 @@ impl Service<Request> for StateService {
...
+                let read_service = self.read_service.clone();
+
+                async move {
+                    let response = is_block_queued.then_some(KnownBlock::Queue).or_else(|| {
+                        read::contains(
+                            &read_service.latest_non_finalized_state(),
+                            &read_service.db,
+                            hash,
+                        )
+                    });
```
teor2345: I think that we might accidentally label some non-finalized blocks as `Queued`, because we don't prune `sent_blocks` above the finalized tip.

So I suggest we check for blocks in this order:
* non-finalized state
* queued
* finalized state

This can become a TODO and a doc update to `KnownBlocks` if you don't want to do it now. (The current code doesn't care where the block is.)

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -420,6 +420,12 @@ impl Chain {
         self.height_by_hash.get(&hash).cloned()
     }
 
+    /// Returns true is the chain contains the given block hash.
+    /// Returns false otherwise.
+    pub fn contains(&self, hash: &block::Hash) -> bool {
```
teor2345: For consistency

```suggestion
    pub fn contains_block_hash(&self, hash: &block::Hash) -> bool {
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -217,6 +217,12 @@ impl QueuedBlocks {
 
         self.blocks.drain()
     }
+
+    /// Returns true if QueuedBlocks contains a block with the given hash.
+    /// Returns false otherwise.
+    pub fn contains(&self, hash: &block::Hash) -> bool {
```
teor2345: To avoid confusion with UTXO transaction hashes.

```suggestion
    pub fn contains_block_hash(&self, hash: &block::Hash) -> bool {
```

zebra-state/src/service/read/find.rs
```diff
@@ -101,6 +101,27 @@ where
     Some(tip.0 - height.0)
 }
 
+/// Returns the location of the block in the state service if present.
+/// Returns None if the block hash is not found in the state.
+pub fn contains(
```
teor2345: For consistency

```suggestion
pub fn contains_block_hash(
```

zebrad/src/components/sync.rs
```diff
@@ -1057,21 +1057,17 @@ where
...
             .map_err(|e| eyre!(e))?
-            .call(zebra_state::Request::Depth(hash))
+            .call(zebra_state::Request::KnownBlock(hash))
             .await
             .map_err(|e| eyre!(e))?
         {
-            zs::Response::Depth(Some(_)) => Ok(true),
-            zs::Response::Depth(None) => Ok(false),
+            zs::Response::KnownBlock(loc) => Ok(loc.is_some()),
             _ => unreachable!("wrong response to depth request"),
```
teor2345:
```suggestion
            _ => unreachable!("wrong response to known block request"),
```

zebra-state/src/service/read/find.rs
arya2: Applied in [`7f7d070`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/7f7d07094ab731d18ce3c3a799ec46beda848add).

zebra-state/src/service.rs
arya2: Thank you!! That's a great idea. Applied in [`7f7d070`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/7f7d07094ab731d18ce3c3a799ec46beda848add).

teor2345 (dismissed):
Thanks for this change, I've been wanting to fix it for ages!

I think it's ok to let it merge without a full sync test, but if the next full sync test stalls or it's very slow, we should consider reverting this PR.

zebra-state/src/service.rs
```diff
@@ -468,13 +469,6 @@ impl StateService {
         let queued_prev_hash = finalized.block.header.previous_block_hash;
```
teor2345: This PR appears to have a bug that causes syncing to hang under some network conditions:
> 2023-03-22T04:55:46.396678Z  WARN {net="Main"}: zebrad::components::sync::progress: chain updates have stalled, state height has not increased for 160 minutes. Hint: check your network connection, and your computer clock and time zone sync_percent=99.993% current_height=Height(2024520) network_upgrade=Nu5 time_since_last_state_block=2h 40m target_block_spacing=PT75S max_block_spacing=None is_syncer_stopped=false

https://github.com/ZcashFoundation/zebra/actions/runs/4485179712/jobs/7887130253?pr=6335

Since this is supposed to be an optimisation-only PR, I'm concerned that we're making more changes that are actually needed. (It would be ok to re-download some redundant blocks, but it's not ok to skip the re-download of a block that's actually needed.)

 I'd suggest reverting any unnecessary changes, or splitting the PR into different stages.

zebra-state/src/service.rs
```diff
@@ -468,13 +469,6 @@ impl StateService {
...
         let queued_height = finalized.height;
 
-        // If we're close to the final checkpoint, make the block's UTXOs available for
-        // full verification of non-finalized blocks, even when it is in the channel.
-        if self.is_close_to_final_checkpoint(queued_height) {
```
teor2345: I'm not sure why this code was moved, is this change needed?

zebra-state/src/service.rs
```diff
@@ -491,6 +485,10 @@ impl StateService {
             }
 
             self.drain_queue_and_commit_finalized();
+
+            if let Some(finalized_tip_height) = self.read_service.db.finalized_tip_height() {
```
teor2345: Is this code move or extra code needed?

teor2345: Do we also need to clear the queue in the "finished finalized blocks" case?

zebra-state/src/service.rs
```diff
@@ -724,17 +727,15 @@ impl StateService {
...
         }
 
         rsp_rx
     }
 
     /// Returns `true` if `hash` is a valid previous block hash for new non-finalized blocks.
     fn can_fork_chain_at(&self, hash: &block::Hash) -> bool {
-        self.sent_non_finalized_block_hashes.contains(hash)
-            || &self.read_service.db.finalized_tip_hash() == hash
+        self.sent_blocks.contains(hash) || &self.read_service.db.finalized_tip_hash() == hash
```
teor2345: We can't fork the chain at a finalized block height below the finalized tip height, but this data structure now contains those heights.

We might need to store the queued finalized block list separately, rather than re-using an existing data structure that other code depends on.

teor2345: I think this might be an existing bug, but only near the final checkpoint. We can't fork from those blocks, but we store their UTXOs in this data structure.

zebrad/src/components/sync.rs
```diff
@@ -1160,7 +1156,7 @@ where
                 // https://github.com/ZcashFoundation/zebra/issues/2909
                 let err_str = format!("{e:?}");
                 if err_str.contains("AlreadyVerified")
-                    || err_str.contains("AlreadyInChain")
+                    || err_str.contains("AlreadyInState")
```
teor2345: Is this change necessary? Did you do a search and replace for all instances?

zebra-state/src/service.rs
teor2345: I'm wondering what happens when:
1. A block is sent to the channel and put in the queued list
2. That block is rejected by the block write task due to a validation error
3. But it stays in the queued list because the rejection doesn't change what's in the queued list

As an alternative, we could just accept that we might sometimes re-download queued blocks across the syncer and inbound service. (Each download service already checks the block hashes of its own pending tasks, and doesn't download duplicate blocks. Those tasks only return once the block has been committed successfully.)

That would mean removing the queued code, and only checking for blocks in side-chains.

teor2345 (commented):
It feels like this change has grown a lot in scope, but the original issue wasn't actually that much of a problem in practice.

Would it help to simplify or split this PR?
We could split the queue part into another PR and fix this bug there?
Or not do the queue check at all?

I feel like there might be a simpler solution - did you want to catch up and talk it through?

teor2345:
Oh and sorry for all the comments, I'm having a bit of trouble keeping up with the changes in this PR. (And it's a complicated part of the code, that really does need a better design eventually!)

arya2:
> did you want to catch up and talk it through?

Sure!

zebra-state/src/service.rs
arya2: > Is this code move or extra code needed?

It's needed to prune the finalized block hashes that are added every time a block is sent to the write task, it was only being pruned when non-finalized block commits were queued.

> Do we also need to clear the queue in the "finished finalized blocks" case?

I think it's okay to leave them there until the finalized tip grows past their heights, and the state needs UTXOs from the finalized blocks near the last checkpoint to verify the first non-finalized blocks.

zebra-state/src/service.rs
arya2: > But it stays in the queued list because the rejection doesn't change what's in the queued list

That could cause syncing to hang, because the `KnownBlock` request will respond with `Some(_)` when a block fails validation so Zebra won't be able to re-download or verify the block.

Fixed in [`c508a08`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/c508a0835276126ae47b4d352e761cdcb55db5da), where it clears sent blocks if there's a reset message from the write task.

zebra-state/src/service.rs
arya2: It was to help keep the `sent_blocks` in order of height, but not strictly necessary and reverted in [`c508a08`](https://github.com/ZcashFoundation/zebra/pull/6335/commits/c508a0835276126ae47b4d352e761cdcb55db5da).

zebrad/src/components/sync.rs
arya2: > Is this change necessary?

Not strictly necessary, I'll split this rename out into a PR for checking the queues.

> Did you do a search and replace for all instances?

That's how I caught it 😎

mpguerra:
> > did you want to catch up and talk it through?
> 
> Sure!

Can we have an update here once you both have a catch up? :)

zebra-state/src/service.rs
arya2: I'll split this change into a follow-up PR and check that the finalized block sender has closed in `can_fork_at()`.

arya2:
@teor2345 it only checks the chains now, I'll open a follow-up PR for checking the queues.

zebra-state/src/service.rs
```diff
@@ -713,13 +713,13 @@ impl StateService {
...
-        // Wait until block commit task is ready to write non-finalized blocks before dequeuing them
         if self.finalized_block_write_sender.is_none() {
+            // Wait until block commit task is ready to write non-finalized blocks before dequeuing them
             self.send_ready_non_finalized_queued(parent_hash);
 
             let finalized_tip_height = self.read_service.db.finalized_tip_height().expect(
-                "Finalized state must have at least one block before committing non-finalized state",
-            );
+            "Finalized state must have at least one block before committing non-finalized state",
+        );
```
teor2345, line 722: Rustfmt?

teor2345 (approved):
Thanks, this looks like a worthwhile change, and it's low risk.

arya2:
> Can we have an update here once you both have a catch up? :)

We'll try merging this PR (which is low risk because it only checks for blocks that have been validated and committed), the follow-up PR that checks the queues is riskier and less important, so we'll try that too if it works well without any further changes but will close it otherwise.

PR #6337, opened 2023-03-16
change(state): Condense CodeTimer::start() and Span::current() calls in state service

arya2:
## Motivation

There is some redundant code in the state services that starts timers and gets handles to the current span for all request variants.

This PR starts timers and gets the current span before the match statements in the state service `call()` methods to condense the code.

## Solution

- Start timer and define `span` at top of state service `call()` methods

## Review

Anyone can review.

### Reviewer Checklist


oxarbitrage (commented):
Yea, this is good. Is this or remove it completely as i think we used them in the past but not anymore since a while.

I will be ok to go with this approach at least for now.

oxarbitrage (approved):

PR #6351, opened 2023-03-16
feat(zebrad): Refuse to run zebrad when release is too old

### LINKED ISSUE #1870: Make old Zebra versions eventually refuse to run
## Motivation

1. Nodes stuck on outdated network upgrades can be attacked using a small amount of hash power. For example, this is a security issue for users if that Zebra instance backs a lightwalletd server. 

2. Include the latest checkpoints to prevent long-range attacks under both Proof of Work (this is even more important under Proof of Stake).

3. Having an end of support halt puts an upper bound on how long security issues can persist on the network.

4. If an old Zebra version launches after a network upgrade, it will keep trying each peer it gets from the DNS seeders, even if they have been upgraded, and reject its connections. This is a distributed denial of service risk, and it places extra load on the network.

Issue 4 is mitigated by:
* #1848 Limit reconnection rate to each individual peer address
* #1865 Running Zebra nodes should eventually stop trying to contact peers that always fail
  * this fix doesn't work when the node is restarted frequently, so we need to eventually make Zebra refuse to run

## Solution

Zebra should refuse to run on launch, if a lot of blocks have been generated since the last release. The exact number of blocks isn't important, but it should be some fraction of the block/times between network upgrades.

Zebra should exit when it reaches the end of support block height.

Design Tasks:
- [ ] work out the average time between recent network upgrades
- [ ] decide how long Zebra should run before it requires an upgrade
- [ ] decide which time we want to use as the "last change" time, this time must be stored somewhere in Zebra's source code
  - an explicit support date constant
  - a date calculated from: a checkpoint date stored in the checkpoint files each time they are updated, and the target block time interval (this requires an update to `zebra-checkpoints` and the checkpoint parsing code)
- [ ] get this design reviewed by other Zebra developers

Implementation Tasks:
- [ ] make Zebra exit on launch if it's a long time since the "last change" time chosen during the design process

Testing:
- [ ] Make a test that fails if the end of support height is in the next few weeks, so we don't forget to update it

Documentation:
- [ ] Add "update the end of support halt" to the release checklist

## Alternatives

This is a serious security issue, so we must do something. But it won't have a major impact until a lot of Zebra nodes are deployed and run for a long time.

## Context

`zcashd` has a support interval of around 16 weeks between required upgrades, with new versions coming out every 6 weeks. It's important that the end of support halt is not a multiple of the release cycle.

This is similar to Zebra's 4 month peer address limit in #1865.

Here is the initial `zcashd` issue:
- https://github.com/zcash/zcash/issues/2274

oxarbitrage:
## Motivation

We want to halt Zebra when a hardcoded release height plus an interval of days is reached.

Close https://github.com/ZcashFoundation/zebra/issues/1870

## Solution

- Added new hardcoded constants so zebra can make checks during sync and refuse to keep if needed. 
- Panic if version is older than 120 days. (to be decided).
- Warn if verson is older than 180 - 14 days.
- Avoid asking users to report the panic.
- Update checklist template to upgrade some of the new constants on each release.

## Review

Anyone can review.

### Reviewer Checklist

# Follow up work

- Add a zebra book entry similar to https://zcash.github.io/zcash/user/release-support.html#end-of-support-heights
- Add rpc call with deprecation info: https://zcash.github.io/rpc/getdeprecationinfo.html


zebrad/src/constants.rs
```diff
@@ -0,0 +1,20 @@
+//! Zebrad constants
+
+/// The name of the current Zebra release.
+pub const RELEASE_NAME: &str = "Zebra 1.0.0-rc.5";
```
arya2: Could we re-use [`zebra_network::constants::USER_AGENT`](https://doc-internal.zebra.zfnd.org/zebra_network/constants/constant.USER_AGENT.html) instead of adding a `RELEASE_NAME` in the constants here.

zebrad/src/constants.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// The date of the current Zebra release.
+///
+/// This must be a valid `chrono::DateTime` with timezone string.
+pub const RELEASE_DATE: &str = "2023-02-23 00:00:00 +00:00";
```
arya2: Could we move this to `zebra_network::constants` to group Rust constants that need to be updated for releases.

zebrad/src/application.rs
```diff
@@ -394,6 +403,21 @@ impl Application for ZebradApp {
...
         if is_server {
             tracing::info!("Diagnostic {}", metadata_section);
+
+            // Check if Zebra is too old by using the last commit timestamp and the `RELEASE_DURATION` constant.
+            let release_date: chrono::DateTime<chrono::Utc> =
+                chrono::DateTime::from_str(RELEASE_DATE).expect("Release date must be valid");
+            let max_time = release_date
+                .checked_add_days(chrono::Days::new(RELEASE_DURATION_DAYS))
+                .expect("`RELEASE_DURATION` should never be that big to overflow");
+            if chrono::Utc::now() > max_time {
```
arya2: I think we might need this logic in the start command as well to wait on a timeout for when the release expires.

Optional: 
It could be nice if it also logs a reminder to update Zebra at ~`max_time - now < RELEASE_DURATION_DAYS / 4`.

arya2 (commented):
This looks great!

teor2345 (commented):
`zcashd` checks the chain tip height before shutting down, rather than the node's local clock. This prevents an attack using NTP servers to set node clocks to a future time, causing them to shut down early.

Docs:
https://github.com/zcash/zcash/blob/b177eeb89e1e2bfeb06b02869a3eb5c668812574/doc/book/src/user/release-support.md#end-of-support-halt

Code:
https://github.com/zcash/zcash/blob/38f7ed3ce4b9f7b678ab8ed93bbd911d8889d8e0/src/deprecation.h#L21

Do we also want to use the tip height for Zebra's shutdown check?

Currently, there is only one standard block validation rule that depends on the local clock:
> In addition, a full validator MUST NOT accept blocks with nTime more than two hours in the future according to its clock. This is not strictly a consensus rule because it is nondeterministic, and clock time varies between nodes. Also note that a block that is rejected by this rule at a given point in time may later be accepted.

This rule stops attackers setting node clocks to a time in the past, and keeping them active forever. But it allows node clocks to be set to any future time.

teor2345:
Here's the similar reasoning for `zcashd`:
https://github.com/zcash/zcash/issues/2274#issuecomment-295297796

mpguerra:
What's happening with this PR? Are we going to use this PR for the approach using block heights or will it be done in a different PR?

oxarbitrage:
I will work in this PR today, force push if needed.

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -123,6 +123,14 @@ From "Keep a Changelog":
 
 </details>
 
+## Release hardcoded constants
+
+Zebra needs release information constants to be updated so it can  know if a release is too old. Please update the top constants [in this file](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/constants.rs):
+
+- [ ] `RELEASE_NAME` (required) - Replace with the new release name here, for example `Zebra 1.0.0-rc.6` if `fastmod` didn't made the change already.
+- [ ] `RELEASE_DATE` (required) - Replace with the estimated date where the release will be created, for example `2023-03-23 00:00:00 +00:00`
```
teor2345: Nitpick:

```suggestion
- [ ] `RELEASE_DATE` (required) - Replace with the estimated date when the release will be created, for example `2023-03-23 00:00:00 +00:00`
```

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -123,6 +123,14 @@ From "Keep a Changelog":
...
+- [ ] `RELEASE_DURATION_DAYS` (optional) - Replace if you want the release to be valid for a different numbers of days into the future.
```
teor2345: Optional: It might help to list the default duration here:

> - [ ] `RELEASE_DURATION_DAYS` (optional) - Replace if you want the release to be valid for a different numbers of days into the future. The default release duration is TODO weeks (in constant TODO).

zebra-network/src/constants.rs
```diff
@@ -21,6 +21,29 @@ use zebra_chain::{
...
+/// Notes:
+///
+/// - Zebra will exit with a panic if the current date is bigger than the `RELEASE_DATE` date plus this number of days.
+pub const EOS_PANIC_AFTER: u64 = 180;
+
+/// The number of days before the end of support where Zebra will display warnings.
+pub const EOS_WARN_AFTER: u64 = EOS_PANIC_AFTER - 5;
+
+/// A string which is part of the panic that will be displayed if Zebra release is too old.
+pub const EOS_PANIC_MESSAGE_HEADER: &str = "Zebra refuses to run";
```
teor2345: Optional: It seems like these belong in the `zebrad` application?

Other applications that use `zebra-network` might find them confusing.

zebra-network/src/constants.rs
```diff
@@ -21,6 +21,29 @@ use zebra_chain::{
...
+
+/// The maximum number of days after `RELEASE_DATE` where a Zebra server runs.
+///
+/// Notes:
+///
+/// - Zebra will exit with a panic if the current date is bigger than the `RELEASE_DATE` date plus this number of days.
+pub const EOS_PANIC_AFTER: u64 = 180;
+
+/// The number of days before the end of support where Zebra will display warnings.
+pub const EOS_WARN_AFTER: u64 = EOS_PANIC_AFTER - 5;
```
teor2345: `zcashd` uses 14 days here, some people need more than a week to plan downtime and an upgrade:
https://github.com/zcash/zcash/blob/38f7ed3ce4b9f7b678ab8ed93bbd911d8889d8e0/src/deprecation.h#L15

```suggestion
pub const EOS_WARN_AFTER: u64 = EOS_PANIC_AFTER - 14;
```

zebrad/src/components/sync.rs
```diff
@@ -1174,3 +1179,35 @@ where
...
+            "{EOS_PANIC_MESSAGE_HEADER} if the release date is older than {EOS_PANIC_AFTER} days. \
+            \nRelease name: {RELEASE_NAME}, Release date: {RELEASE_DATE} \
+            \nHint: Download and install the latest Zebra release from: https://github.com/ZcashFoundation/zebra/releases/latest"
+        );
+    } else if block_tip_time > max_time_warning {
+        warn!(
+            "Your Zebra release is too old and it will stop running in {max_time_panic}. \
+            \nRelease name: {RELEASE_NAME}, Release date: {RELEASE_DATE} \
+            \nHint: Download and install the latest Zebra release from: https://github.com/ZcashFoundation/zebra/releases/latest"
+        );
```
teor2345: Can you add this warning and panic to the list of integration test `ZEBRA_FAILURE_MESSAGES`?
That will give us extra test coverage.

zebrad/src/components/sync.rs
```diff
@@ -1174,3 +1179,35 @@ where
...
+    let release_date: chrono::DateTime<chrono::Utc> =
+        chrono::DateTime::from_str(RELEASE_DATE).expect("Release date must be valid");
+    let max_time_panic = release_date
+        .checked_add_days(chrono::Days::new(EOS_PANIC_AFTER))
+        .expect("`EOS_DURATION_PANIC` should never be that big to overflow");
+    let max_time_warning = release_date
+        .checked_add_days(chrono::Days::new(EOS_WARN_AFTER))
+        .expect("`EOS_DURATION_WARNING` should never be that big to overflow");
+
+    if block_tip_time > max_time_panic {
```
teor2345: Using the block time is a good fix, and it helps avoid an attack that just modifies the local clock.

But I think there still might be some ways to attack `zebrad` using the time. If an attacker advances the local clock of a Zebra node to the end of support time time, and then mines a single block with that time, Zebra will accept the block, and exit.

There are also similar attacks which advance the times of the majority of the mining pools on the network.

These attacks significantly harder than the original attack, but they are still much less difficult than attacking the `zcashd` end of support height. To attack `zcashd`, an attacker has to mine and sync all the blocks up to that height, which is tens of thousands of times harder:
https://github.com/zcash/zcash/blob/38f7ed3ce4b9f7b678ab8ed93bbd911d8889d8e0/src/deprecation.h#L15

Do you think checking the block time is enough of a change, or should we just be checking the block height?

zebrad/src/components/sync.rs
```diff
@@ -604,6 +604,11 @@ where
                 })?;
             }
             self.update_metrics();
+
+            // check for the end of support of this release.
+            if let Some(block_tip_time) = self.latest_chain_tip.best_tip_block_time() {
+                end_of_support(block_tip_time);
+            }
```
teor2345: We've had bugs that have hung the syncer task, but the rest of `zebrad` keeps running fine. (Including the inbound block downloader, which syncs gossiped blocks after we've reached the tip.)

So I'm wondering if this check should be in a different task?

For example, it could go in the progress task, which doesn't block on anything:
https://github.com/ZcashFoundation/zebra/blob/5db2243c25cdde03d9cdccd72a137ccc6f2370a5/zebrad/src/components/sync/progress.rs#L109-L111

zebrad/src/components/sync/tests/vectors.rs
```diff
@@ -981,3 +987,61 @@ fn setup() -> (
         mock_chain_tip_sender,
     )
 }
+
+/// Test that the `end_of_support` function is working as expected.
+#[test]
+#[tracing_test::traced_test]
+#[should_panic(expected = "Zebra refuses to run if the release date is older than")]
+fn end_of_support_function() {
```
teor2345: Can we also add an integration test that runs the EOS-checking task with a fake chain tip?
That makes sure the task is actually calling the `end_of_support()` check regularly.

(It's tricky to do a full integration test with the whole of `zebrad`, so I don't think that's worth the effort.)

zebrad/src/components/sync/tests/vectors.rs
```diff
@@ -981,3 +987,61 @@ fn setup() -> (
...
+
+    // We are in panic
+    let panic = release_date
+        .checked_add_days(chrono::Days::new(
+            zebra_network::constants::EOS_PANIC_AFTER + 1,
+        ))
+        .unwrap();
+
+    sync::end_of_support(panic);
+    assert!(logs_contain("Checking if Zebra release is too old"));
```
teor2345: Optional: it might help to split the checks that should panic from the checks that don't panic. That will make test failures easier to diagnose.

zebrad/src/components/sync/tests/vectors.rs
```diff
@@ -981,3 +987,61 @@ fn setup() -> (
...
+}
+
+/// Test that we are never in end of support warning or panic.
+#[test]
+#[tracing_test::traced_test]
+fn end_of_support_date() {
+    // We check this with local clock.
+    let now = chrono::Utc::now();
+
+    sync::end_of_support(now);
```
teor2345: This is a clever way to test!

teor2345 (commented):
Thanks for these fixes, and thanks for working through the release checklist changes!

I think this is pretty good, it seems like it should work and we could merge it.

I haven't had time to think through the security, how it's different from `zcashd`, and if we need to make any changes. But I wanted to share my thoughts and some questions, so you could think about it too.

zebrad/src/components/sync/tests.rs
```diff
@@ -1,4 +1,5 @@
 //! Syncer tests
 
+mod eos;
```
oxarbitrage: This tests only work when I run them separated from others. When running the whole suite they fail with :

```
Could not set global tracing subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

There is a conflict with `tracing_test` and already subscribed code. I tried different alternatives today here, none of them solved it. I might go in a different way than capturing the logs in the test.

zebrad/tests/acceptance.rs
```diff
@@ -2206,3 +2206,34 @@ async fn get_block_template() -> Result<()> {
 async fn submit_block() -> Result<()> {
     common::get_block_template_rpcs::submit_block::run().await
 }
+
+/// Check that the the end oif support code is called.
+#[test]
+fn end_of_support_is_checked() -> Result<()> {
```
oxarbitrage: This test do not work as is, the idea was a simple test to see if the eos code is called. I didn't had enough time to fix it but i suppose it should be something simple missing.

zebrad/src/components/sync.rs
oxarbitrage: I think we can change to height with some effort but i didn't had enough time to do it.

teor2345:
Bumping this down to low priority because the audit is a higher priority, and the author of the PR is away.

zebra-network/src/constants.rs
```diff
@@ -21,6 +21,29 @@ use zebra_chain::{
...
+///
+/// This must be a valid `chrono::DateTime` with timezone string.
+pub const RELEASE_DATE: &str = "2023-03-23 00:00:00 +00:00";
+
+/// The maximum number of days after `RELEASE_DATE` where a Zebra server runs.
+///
+/// Notes:
+///
+/// - Zebra will exit with a panic if the current date is bigger than the `RELEASE_DATE` date plus this number of days.
+pub const EOS_PANIC_AFTER: u64 = 180;
```
teor2345: Let's ask the team how long we should support Zebra?
Currently `zcashd` uses 16 weeks.

mpguerra: Unless anyone sees any strong reason why we should do it differently I would match what `zcashd` does. I also wouldn't block on this necessarily, as long as the length of time can be easily changed as needs evolve. I feel like the mechanism is more important than the actual choice of support time length.

zebrad/src/components/sync.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/6351/commits/89345b72694cba307e6347abf8a1a285d4e5d5de

zebra-network/src/constants.rs
arya2: > Unless anyone sees any strong reason why we should do it differently I would match what zcashd does.

I don't have a strong opinion here but I think Str4d mentioned zcashd's 16 weeks feeling too short in hindsight.

Update: I think 16 weeks sounds better if we implement https://github.com/ZcashFoundation/zebra/issues/6564.

teor2345: Since we might be fixing some serious network security issues as part of the audit, I'd like to make the support lifetime short for now. That way, all vulnerable versions of Zebra will be removed from the network sooner.

For example, if we merge this PR with a 6 month limit, we'll have those Zebra versions on the network until November 2023. But if we use a 16 week limit, they'll be around until August 2023.

We can extend it later once we're confident Zebra is secure and working well.

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
...
+use color_eyre::Report;
+
+use zebra_chain::{
+    block::Height,
+    chain_tip::ChainTip,
+    parameters::{Network, NetworkUpgrade},
+};
+
+/// The name of the current Zebra release.
+pub const RELEASE_NAME: &str = "Zebra:1.0.0-rc.6";
```
teor2345: Optional: Can we re-use the `zebra-network` user agent constant here?

This would simplify the release process. So if we make this change, we'll also need to delete those parts of the release checklist.

In `zebra-network`:
```rust
pub const RELEASE_VERSION: &str = "1.0.0-rc.7";
lazy_static! {
    pub const USER_AGENT: &str = format!("/Zebra:{RELEASE_VERSION}/");
}
```

```rust
lazy_static! {
    pub const RELEASE_NAME: &str = format!("Zebra {}", zebra_network::constants::RELEASE_VERSION);
}
```

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -123,6 +124,20 @@ From "Keep a Changelog":
...
 </details>
 
+## Release support constants
+
+Needed for the end of support feature. Please update the top constants [in this file](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/components/sync/end_of_support.rs):
+
+- [ ] `RELEASE_NAME` (required) - Replace with the new release name here, for example `Zebra 1.0.0-rc.6` if `fastmod` didn't made the change already.
+- [ ] `ESTIMATED_RELEASE_HEIGHT` (required) - Replace with the estimated height you estimate the release will be tagged.
+      <details>
+      - Find where the Zcash blockchain tip is now by using a Zcash explorer or other tool.
```
teor2345: Nitpick: Can you link to a reliable block explorer here?

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
...
+            \nHint: Download and install the latest Zebra release from: https://github.com/ZcashFoundation/zebra/releases/latest"
+        );
+    } else if tip_height > warn_height {
+        warn!(
+            "{EOS_WARN_MESSAGE_HEADER} at block {}. \
+            \nRelease name: {RELEASE_NAME}, Estimated release height: {ESTIMATED_RELEASE_HEIGHT} \
+            \nHint: Download and install the latest Zebra release from: https://github.com/ZcashFoundation/zebra/releases/latest", panic_height.0
+        );
+    } else {
+        info!("Zebra release is under support");
```
teor2345: Let's tell users the height that support ends at.

Optional: would it also help to give an estimated support end date here? Or will users be annoyed if the estimate is wrong? What does `zcashd` do?

```suggestion
        info!("Zebra release is supported until block {}, please report bugs at https://github.com/ZcashFoundation/zebra/issues", panic_height.0);
```

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
...
+pub const EOS_WARN_AFTER: u32 = EOS_PANIC_AFTER - 14;
+
+/// A string which is part of the panic that will be displayed if Zebra halts.
+pub const EOS_PANIC_MESSAGE_HEADER: &str = "Zebra refuses to run";
+
+/// A string which is part of the warning that will be displayed if Zebra release is close to halting.
+pub const EOS_WARN_MESSAGE_HEADER: &str = "Your Zebra release is too old and it will stop running";
+
+/// The amount of time between end of support checks.
+const CHECK_INTERVAL: Duration = Duration::from_secs(60 * 60);
```
teor2345, line 43: How often does `zcashd` do this check?

Optional: I think we should check every minute or so. A node can load a lot of blocks in an hour, and I'd like to avoid outdated nodes running that long. And the check is really cheap.

If we make this change, we should only log a check at info level every hour. (It's ok to log every check at debug level.)

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
...
+        Height(ESTIMATED_RELEASE_HEIGHT + (EOS_PANIC_AFTER * estimated_blocks_per_day));
+    let warn_height =
+        Height(ESTIMATED_RELEASE_HEIGHT + (EOS_WARN_AFTER * estimated_blocks_per_day));
+
+    if tip_height > panic_height {
+        panic!(
+            "{EOS_PANIC_MESSAGE_HEADER} if the release date is older than {EOS_PANIC_AFTER} days. \
+            \nRelease name: {RELEASE_NAME}, Estimated release height: {ESTIMATED_RELEASE_HEIGHT} \
+            \nHint: Download and install the latest Zebra release from: https://github.com/ZcashFoundation/zebra/releases/latest"
+        );
```
teor2345: Optional: Can we return from the `loop {}` in `check()` instead of panicking?

That gives a much better error message to the user.

If we make this change, we should also:
- remove the panic exception in `application.rs`
- update the docs that mention panicking

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
+//! End of support checking task.
```
teor2345, line 1: > Follow up work
> * Add a zebra book entry similar to https://zcash.github.io/zcash/user/release-support.html#end-of-support-heights

Ending support for old Zebra versions seems like something we really want to tell users about. So for the first release where we enable this feature, we might want:
- a Zebra book entry
- ~~a summary in the README~~
- a summary in the CHANGELOG
- ~~a blog post~~

These can all use similar wording.

@oxarbitrage @mpguerra which of these things should we do as part of this PR? And which can be another ticket?

Edit: ~~strike through~~ the things Pili said we're not doing

zebrad/src/components/sync/end_of_support.rs
```diff
@@ -0,0 +1,92 @@
...
+    block::Height,
+    chain_tip::ChainTip,
+    parameters::{Network, NetworkUpgrade},
+};
+
+/// The name of the current Zebra release.
+pub const RELEASE_NAME: &str = "Zebra:1.0.0-rc.6";
+
+/// The estimated height that this release started to run.
+pub const ESTIMATED_RELEASE_HEIGHT: u32 = 2_026_000;
```
teor2345, line 22: `zcashd` only does this check on mainnet, do we want to end support on testnet?

It would be nice to test this feature on testnet, and get rid of old nodes. But it also adds another step to the release process. 

If we do, we'll need a different `ESTIMATED_RELEASE_HEIGHT` on testnet. Because this height is below the current testnet tip.

Optional: Do we want a different support lifetime (`EOS_PANIC_AFTER`) on testnet?

teor2345 (commented):
Looks great, thanks for all these changes!

I think these might be blockers for merging this PR:
- It looks like Zebra will exit immediately when run on testnet
- Change the support period to match `zcashd` for now (16 weeks)

We'll also want to tell users that support for old Zebra versions will end. But those CHANGELOG and README updates are only needed for the next release after this PR merges.

daira:
For information, zcashd has no end-of-support when run on testnet. (This isn't a problem for network upgrades, because testnet users are typically paying more attention to announcements about what version they should be running to test the upgrade, and also there's no loss of real funds potential if they don't.)

zebrad/src/components/sync/end_of_support.rs
teor2345: This might be optional.

zebrad/src/components/sync/end_of_support.rs
mpguerra: We should always document new features in the relevant places and in particular where the behaviour significantly difers from `zcashd`. 
So we should add a Zebra book entry, probably under "[Running Zebra](https://zebra.zfnd.org/user/run.html)" and we should definitely highlight it in the `CHANGELOG` and release notes when we tag the next release.
However, given that we don't have many users yet, I don't think we need a blogpost. I also don't see it as something that needs to go in the `README`, I seem to remember people find it too long and overwhelming as it is and I'm not sure that this is some detail that people would go looking for in there.

mpguerra: After looking at where this is mentioned in the `zcashd` book again, I should add that the location in the zebra book (under "Running Zebra") can be temporary. 

From the stable release onwards, we might decide we want to dedicate a whole section in the zebra book to Zebra releases to include info such as release cadence and end of support details for each zebra release version. 

I don't think we're there yet though.

zebrad/src/components/sync/end_of_support.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/6351/commits/460fcfd6056ae9be7c4029be122d0017984e054e

zebrad/src/components/sync/end_of_support.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/6351/commits/ff72c1938e93e0b9ca65d515df38b57fcd37d92f

zebrad/src/components/sync/end_of_support.rs
oxarbitrage: Will not do, i think 60 minutes is ok for this.

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/6351/commits/dc7daba4c936188a046536f29a8b2a433111c0e6

zebrad/src/components/sync/end_of_support.rs
oxarbitrage: We are now skipping the checks in the testnet. https://github.com/ZcashFoundation/zebra/pull/6351/commits/3a9a3ee180cb09f68b0ae6d11cedb4facfa5e93d

zebra-network/src/constants.rs
oxarbitrage: I changed to the exact 16 weeks(112 days) at https://github.com/ZcashFoundation/zebra/pull/6351/commits/11f34721b73b73b919545c5a45e80c7004fa1d58

zebrad/src/components/sync/end_of_support.rs
teor2345: > From the stable release onwards, we might decide we want to dedicate a whole section in the zebra book to Zebra releases to include info such as release cadence and end of support details for each zebra release version.

Part of this already exists in:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/release-process.md#supported-releases

So we could just update that book section in this PR, rather than adding a new section with conflicting info.

Edit: and link to it from "Running Zebra"

teor2345 (dismissed):
This looks good, someone just needs to decide when we're doing the doc updates to supported releases and the CHANGELOG.

zebrad/src/components/sync/end_of_support.rs
mpguerra: ah right, this section seems to be more of a section for zebra developers rather than zebra users though. 

Anyway, if we just want to inform users of this particular feature we should probably put it somewhere in the user documentation section of the book.

book/src/dev/release-process.md
```diff
@@ -33,10 +33,11 @@ The pre-release version is denoted by appending a hyphen and a series of dot sep
 
 ### Supported Releases
 
-Older Zebra versions are always supported until the next Zebra major, minor or patch release. Initially, we can only guarantee support for the latest Zebra release.
-We might be able to provide support for earlier releases, or we might ask you to upgrade to the latest release.
+Every Zebra version released by the Zcash Foundation is supported up to a specific height. Currently we support each version for about **16 weeks** but this can change from release to release.
 
-Our support periods will be extended as we gain experience with supporting Zebra releases.
+When The Zcash chain reaches this end of support height, `zebrad` will shut down and the binary will refuse to start.
```
teor2345: Typo?

```suggestion
When the Zcash chain reaches this end of support height, `zebrad` will shut down and the binary will refuse to start.
```

book/src/dev/release-process.md
```diff
@@ -33,10 +33,11 @@ The pre-release version is denoted by appending a hyphen and a series of dot sep
...
+
+Our process in this subject is very similar to `zcashd`: https://zcash.github.io/zcash/user/release-support.html
```
teor2345: Nitpick: Simplify sentence

```suggestion
Our process is similar to `zcashd`: https://zcash.github.io/zcash/user/release-support.html
```

book/src/user/run.md
```diff
@@ -11,6 +11,10 @@ structure, and documentation for all of the config options can be found
 
 * `zebrad start` starts a full node.
 
+## Supported versions
+
+Always make sure your version is under support and that you upgrade regularly to avoid halting. [More information](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/release-process.md#supported-releases).
```
teor2345: "Under support" sounds a lot like "unsupported", but I don't think that's what you meant here.

```suggestion
Always run a supported version of Zebra, and upgrade it regularly, so it doesn't become unsupported and halt. [More information](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/release-process.md#supported-releases).
```

teor2345 (dismissed):
Looks good, I just want to fix some confusing wording

zebrad/src/components/sync/end_of_support.rs
teor2345: I think we might want to move some of `dev/release-process.md` into the `user` side of the book before the stable release. But that can happen when we've worked out what we want there.

teor2345 (approved):
These are just docs tweaks, so I'm going to apply them and merge this PR.

oxarbitrage:
Failed in the build docker stage with https://github.com/ZcashFoundation/zebra/actions/runs/4826536076/jobs/8598460445#step:42:4

oxarbitrage:
@mergifyio update

PR #6357, opened 2023-03-20
refactor(ci): use GitHub variables for non-sensitive info

### LINKED ISSUE #6312: refactor(ci): define configuration variables for multiple workflows
## Motivation

We stopped using secrets for non-sensitive information that could vary in some point in time, as there were some limitations with secrets while using reusable workflows, but GitHub recently added configuration variables to be used in multiple workflows, so we don't have to repeat ourselves: https://github.blog/2023-01-10-introducing-required-workflows-and-configuration-variables-to-github-actions/

### Specifications

Documentation: https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows

Some of the code/variables that needs to be added or changed:
- Our GCP Project name
- Our registry URLs
- GCP machine types
- Other variables that have an impact on all or most workflows

### Testing

We must confirm a PR run all test successfully after the change 

## Related Work

https://github.com/ZcashFoundation/infra/issues/8

gustavovalverde:
## Motivation

We've been using values that are variable across multiple workflows,
and those can only be changed if modifying the workflows, but we should
be able to change the values without committing new changes in the code
for this purpose we're now using GitHub Variables, and even moving
non-sensitive information into variables instead of secrets. Allowing
more flexibility and other scenarios that should be easier to manage,
like deploying to Mainnet or Testnet.

Closes: #6312

This will also make it easier to fix our actual `main` branch failures

### Specifications

https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows

## Solution

Use GitHub variables for:
- Our GCP Project name
- Our registry URLs
- GCP machine types
- Other variables that have an impact on all or most workflows

## Review

Anyone from the DevOps team, after test are passing (or before for desing feedback)

**Check the actual values here**: https://github.com/ZcashFoundation/zebra/settings/variables/actions

### Reviewer Checklist


gustavovalverde:
This should also make it easier to fix `main` branch CI failures, when related to CPUs not being available on a specific region.

.github/workflows/build-docker-image.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         with:
           # list of Docker images to use as base name for tags
           images: |
-            us-docker.pkg.dev/zealous-zebra/zebra/${{ inputs.image_name }}
+            us-docker.pkg.dev/${{ vars.GCP_PROJECT }}/zebra/${{ inputs.image_name }}
```
teor2345, line 87: How can we make sure we have replaced all the hard-coded strings with variables?
Was this PR created using a tool like `fastmod` or `grep`?

.github/workflows/build-crates-individually.yml
```diff
@@ -37,11 +37,11 @@ on:
       - '.github/workflows/build-crates-individually.yml'
```
teor2345, line 37: This seems like a large change. Half the team goes to a conference at the end of the week, so do we need to make this change now, or can it wait until we have most people back?

teor2345 (commented):
I like this change, I think it will make things a lot easier!

But I'm not sure how to review or test it, since we only run some workflows on `main` or on specific events (like releases).

.github/workflows/build-docker-image.yml
gustavovalverde: I used my VSCode replacement tool, but I had to anyways validate most of the mass replacement before applying it as not all of them applied _as-is_ like the `IMAGE_NAME` (which I'll improve in the next refactor)

.github/workflows/build-crates-individually.yml
gustavovalverde: In the worst case scenario we at least have fewer people making changes and the snowball is way smaller if something bad happens, and we can revert the commit or apply a fix. Theoretically, we're not changing the behavior on the workflows, but the source of the values (while keeping the actual values).

I'm taking into consideration being able to run all the tests on this kind of changes, on an isolated environment, before reaching main...sadly, we're not there yet (but almost)

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -89,10 +89,6 @@ on:
 env:
   # where we get the Docker image from
   IMAGE_NAME: zebrad-test
-  GAR_BASE: us-docker.pkg.dev/zealous-zebra/zebra
-  # what kind of Google Cloud instance we want to launch
-  ZONE: us-central1-f
-  MACHINE_TYPE: c2d-standard-16
```
teor2345, line 95: This machine type is larger than the machine type in other files, so it needs a different variable

.github/workflows/continous-delivery.yml
```diff
@@ -25,12 +25,6 @@ on:
     types:
       - published
 
-env:
-  GAR_BASE: us-docker.pkg.dev/zealous-zebra/zebra
-  REGION: us-central1
-  ZONE: us-central1-a
-  MACHINE_TYPE: c2-standard-4
```
teor2345, line 32: This is the standard machine type, so it needs a different variable name

.github/workflows/deploy-gcp-tests.yml
teor2345: It also possibly needs a different zone, in the rare case some machine types are available in different zones, and we can't find a common zone that has both.

.github/workflows/zcashd-manual-deploy.yml
```diff
@@ -12,12 +12,6 @@ on:
         description: 'GCP Managed Instance Group size'
         required: true
 
-env:
-  PROJECT_ID: zealous-zebra
-  REGION: us-central1
-  ZONE: us-central1-a
-  MACHINE_TYPE: c2-standard-4
```
teor2345, line 19: This is the standard machine type, it needs a different variable name

teor2345:
This is probably something for the next refactor, but can you change some of the secrets that don't need to be confidential into variables?

it's much easier to edit the value if we know what the previous value was.

mpguerra:
> This is probably something for the next refactor, but can you change some of the secrets that don't need to be confidential into variables?
> 
> it's much easier to edit the value if we know what the previous value was.

Can this be merged "as is" with follow ups in other PRs? Otherwise this will have to be parked until Sprint 8

teor2345:
> > This is probably something for the next refactor, but can you change some of the secrets that don't need to be confidential into variables?
> > it's much easier to edit the value if we know what the previous value was.
> 
> Can this be merged "as is" with follow ups in other PRs? Otherwise this will have to be parked until Sprint 8

Unfortunately, this PR uses a single variable for the machine type, but our current CI uses two different machine types. Changing all workflows to a single machine type could break CI or be really expensive, so we need to fix that before we merge:
https://github.com/ZcashFoundation/zebra/pull/6357#discussion_r1145388175

teor2345:
All the other suggestions could be done as follow-ups.

teor2345:
> This should also make it easier to fix `main` branch CI failures, when related to CPUs not being available on a specific region.

We might want to try switching the large machine workflows from Zone 1f to 1b, if that would help with "service unavailable" errors:
https://github.com/ZcashFoundation/zebra/pull/6426#issuecomment-1487749393

teor2345:
I'm bumping this down to a medium priority because I think we fixed the urgent bug another way in PR #6365

gustavovalverde:
Aside from `secrets.DOCKERHUB_USERNAME`, all other _non-sensitive_ information was replaced with a variable.

teor2345 (dismissed):
Thanks for these fixes, it looks good!

.github/workflows/continous-integration-docker.yml
```diff
@@ -106,19 +99,19 @@ jobs:
...
 
       # Disk images in GCP are required to be in lowercase, but the blockchain network
-      # uses sentence case, so we need to downcase ${{ env.NETWORK or github.event.inputs.network }}
+      # uses sentence case, so we need to downcase ${{ vars.ZCASH_NETWORK or github.event.inputs.network }}
       #
       # Passes a lowercase Network name to subsequent steps using $NETWORK env variable
       - name: Downcase network name for disks
         run: |
-          NETWORK_CAPS=${{ env.NETWORK || github.event.inputs.network }}
+          NETWORK_CAPS=${{ vars.ZCASH_NETWORK || github.event.inputs.network }}
```
teor2345: This is an existing bug, the supplied network should override the default network variable:
```suggestion
          NETWORK_CAPS=${{ github.event.inputs.network || vars.ZCASH_NETWORK }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -169,9 +162,9 @@ jobs:
     with:
       dockerfile_path: ./docker/Dockerfile
       dockerfile_target: tests
-      image_name: zebrad-test
+      image_name: ${{ vars.CI_IMAGE_NAME }}
       # TODO: validate how to use variable/conditional values for Testnet
```
```suggestion
      # The default network in the image is based on the default network variable in GitHub
```

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1725,13 +1719,13 @@ jobs:
           MINIMUM_UPDATE_HEIGHT=$((ORIGINAL_HEIGHT+CACHED_STATE_UPDATE_LIMIT))
           if [[ -z "$UPDATE_SUFFIX" ]] || [[ "$SYNC_HEIGHT" -gt "$MINIMUM_UPDATE_HEIGHT" ]]; then
              gcloud compute images create \
-              "${{ inputs.disk_prefix }}-${SHORT_GITHUB_REF}-${{ env.GITHUB_SHA_SHORT }}-v${{ env.STATE_VERSION }}-${{ env.NETWORK }}-${{ inputs.disk_suffix }}${UPDATE_SUFFIX}-${TIME_SUFFIX}" \
+              "${{ inputs.disk_prefix }}-${SHORT_GITHUB_REF}-${{ env.GITHUB_SHA_SHORT }}-v${{ env.STATE_VERSION }}-${{ vars.ZCASH_NETWORK }}-${{ inputs.disk_suffix }}${UPDATE_SUFFIX}-${TIME_SUFFIX}" \
```
teor2345: We need the lowercase selected network name, because the search won't work with the titlecase default network name:
```suggestion
              "${{ inputs.disk_prefix }}-${SHORT_GITHUB_REF}-${{ env.GITHUB_SHA_SHORT }}-v${{ env.STATE_VERSION }}-${NETWORK}-${{ inputs.disk_suffix }}${UPDATE_SUFFIX}-${TIME_SUFFIX}" \
```

In general, it would help if we put more details in variable names, so that it's clearer what they should be used for:
- `vars.ZCASH_NETWORK` -> `vars.DEFAULT_ZCASH_NETWORK`
- `${NETWORK}` -> `${NETWORK_LOWERCASE}`

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1725,13 +1719,13 @@ jobs:
...
               --force \
               --source-disk=${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }} \
-              --source-disk-zone=${{ env.ZONE }} \
+              --source-disk-zone=${{ vars.GCP_ZONE }} \
               --storage-location=us \
               --description="Created from commit ${{ env.GITHUB_SHA_SHORT }} with height ${{ env.SYNC_HEIGHT }}" \
-              --labels="height=${{ env.SYNC_HEIGHT }},purpose=${{ inputs.disk_prefix }},commit=${{ env.GITHUB_SHA_SHORT }},state-version=${{ env.STATE_VERSION }},network=${{ env.NETWORK }},target-height-kind=${{ inputs.disk_suffix }},update-flag=${UPDATE_SUFFIX},updated-from-height=${ORIGINAL_HEIGHT},test-id=${{ inputs.test_id }},app-name=${{ inputs.app_name }}"
+              --labels="height=${{ env.SYNC_HEIGHT }},purpose=${{ inputs.disk_prefix }},commit=${{ env.GITHUB_SHA_SHORT }},state-version=${{ env.STATE_VERSION }},network=${{ vars.ZCASH_NETWORK }},target-height-kind=${{ inputs.disk_suffix }},update-flag=${UPDATE_SUFFIX},updated-from-height=${ORIGINAL_HEIGHT},test-id=${{ inputs.test_id }},app-name=${{ inputs.app_name }}"
```
teor2345: Labels must be lowercase. (And the default network is incorrect here, it needs to be the input network.)

```suggestion
              --labels="height=${{ env.SYNC_HEIGHT }},purpose=${{ inputs.disk_prefix }},commit=${{ env.GITHUB_SHA_SHORT }},state-version=${{ env.STATE_VERSION }},network=${NETWORK},target-height-kind=${{ inputs.disk_suffix }},update-flag=${UPDATE_SUFFIX},updated-from-height=${ORIGINAL_HEIGHT},test-id=${{ inputs.test_id }},app-name=${{ inputs.app_name }}"
```

teor2345 (approved):
This PR failed in the merge queue, because it tried to create a disk using the incorrect network name:
> ERROR: (gcloud.compute.images.create) argument --labels: Bad value [Mainnet]: Only hyphens (-), underscores (_), lowercase characters, and numbers are allowed. International characters are allowed.

https://github.com/ZcashFoundation/zebra/actions/runs/4683572250/jobs/8299559492?pr=6493#step:13:81

I have applied a quick fix to use the correct network variable, and a fix for another incorrect network default variable.

In general, it would help if we put more details in variable names, so that it's clearer what they should be used for:
- `vars.ZCASH_NETWORK` -> `vars.DEFAULT_ZCASH_NETWORK`
- `${NETWORK}` -> `${NETWORK_LOWERCASE}`

(I didn't make these changes in this PR, happy to leave them to a later refactor.)

gustavovalverde:
```
I have applied a quick fix to use the correct network variable, and a fix for another incorrect network default variable.

In general, it would help if we put more details in variable names, so that it's clearer what they should be used for:

vars.ZCASH_NETWORK -> vars.DEFAULT_ZCASH_NETWORK
${NETWORK} -> ${NETWORK_LOWERCASE}

```

Thanks for this, I'll take this into account for further refactors I have pending.

teor2345:
Failed in the merge queue due to #6498

teor2345:
@mergifyio refresh

PR #6385, opened 2023-03-22
fix(errors): Handle randomness generation and invalid random values as errors in cryptographic code

### LINKED ISSUE #6338: [NCC-E005955-DBV] `zebra-chain`: Unbounded Rejection Sampling with Possibility of Panics
## Impact

The `zebra-chain` module implements random number generation for some types which are primarily used for testing, without guarding against possible panics.

## Description

The use of `fill_bytes()` in the code snippet below (from `zebra-chain/src/sapling/keys.rs`) [can potentially cause panics](https://docs.rs/rand/latest/rand/trait.RngCore.html#tymethod.fill_bytes). It is recommended to use `try_fill_bytes()` to catch the random number generator’s failures<sup>1</sup>.

In addition, the number of sampling retries by the loop has to be bounded.

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/sapling/keys.rs#L176-L196

There is one instance where this can potentially be problematic for a full node with mining support. Generating [a random](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/orchard/note.rs#L65) `Rho` for an Orchard note uses the same unchecked RNG API.

A node with mining capabilities might use this API to generate [a dummy Orchard input note](https://zips.z.cash/protocol/nu5.pdf#orcharddummynotes) as a nullifier for the output note when creating a miner’s reward note:

https://github.com/ZcashFoundation/zebra/blob/161bb80ce8fc131bd7fea3a95723b34f1795716e/zebra-chain/src/orchard/note.rs#L59-L69

---
1. For instance `OsRng` can, in rare occasions, fail on some platforms, and `thread_rng` might return a
warning if it fails to reseed itself. See https://rust-random.github.io/book/guide-err.html for more
details.
## Recommendation

Consider replacing the `fill_bytes()` usage with `try_fill_bytes()` and properly handle its possible failure.

## Location

- [zebra-chain/src/orchard/keys.rs, line 110](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/orchard/keys.rs#L110)
- [zebra-chain/src/sapling/keys.rs, line 189](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/sapling/keys.rs#L189)
- [zebra-chain/src/orchard/commitment.rs, line 30](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/orchard/commitment.rs#L30)
- [zebra-chain/src/sapling/commitment.rs, line 42](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/sapling/commitment.rs#L42)
- [zebra-chain/src/orchard/note.rs, line 30](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/orchard/note.rs#L30)
- [zebra-chain/src/orchard/note.rs, line 65](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/orchard/note.rs#L65)

oxarbitrage:
## Motivation

We want to avoid some panics in some cryptography code. There are some specific problems that were pointed by the audit.

This should close https://github.com/ZcashFoundation/zebra/issues/6338

The code modified here is currently not used but we expect to use it in the future and we will want to avoid panics.

## Solution

Most related code use a string as an error with [this pattern](https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/keys.rs#L143).
However, this is not always possible, for example if you want to do that [here](https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/keys.rs#L176) you will get the following compiler error:

```
inherent associated types are unstable
see issue #8995 <https://github.com/rust-lang/rust/issues/8995> for more information
```

So i think that limitation might be the cause of not using `try_fill_bytes()` in the first place.

What is needed is our own errors for handling this and that is what this PR tries to do.

I think all the string errors should be eventually refactored to this but i didnt wanted to do that here as it will be too intrusive so i just fixed what was pointed by the audit ticket. 

## Review

I think anyone can review, the names chosen could be improved maybe.

### Reviewer Checklist

## Follow Up Work

Add errors for all string errors used inside zebra-chain ?


zebra-chain/src/error.rs
```diff
@@ -0,0 +1,38 @@
+//! Errors that can occur inside any `zebra-chain` submodule.
+use thiserror::Error;
+
+/// Number of times a `diversify_hash` will try to obtain a diversified base point.
+pub const DIVERSIFY_HASH_TRIES: u8 = 5;
```
oxarbitrage: I honestly don't have idea if this number is a good choice.

teor2345: What does `zcashd` do?

zebra-chain/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Errors related to random bytes generation.
```
teor2345, line 4: I'd like a cryptographer to review these error names and descriptions to make sure they match the terminology in the spec (or standard cryptographic terminology).

zebra-chain/src/sapling/commitment.rs
```diff
@@ -120,11 +123,9 @@ impl NoteCommitment {
         // The `TryFrom<Diversifier>` impls for the `jubjub::*Point`s handles
         // calling `DiversifyHash` implicitly.
 
-        let g_d_bytes: [u8; 32] = if let Ok(g_d) = jubjub::AffinePoint::try_from(diversifier) {
-            g_d.to_bytes()
-        } else {
-            return None;
-        };
+        let g_d_bytes = jubjub::AffinePoint::try_from(diversifier)
```
teor2345, line 126: This might also need multiple retries, but I'd like to check with a cryptographer to be sure.

teor2345 (commented):
Looks great, thanks for making minimal changes!

I'd just like to check a few cryptographic details.

mpguerra:
> Looks great, thanks for making minimal changes!
> 
> I'd just like to check a few cryptographic details.

I'm not sure we're going to be able to get a cryptography review until after RWC and/or Easter so this may have to move to Sprint 8. 

EDIT: Unless @upbqdn is around today or tomorrow?

upbqdn:
I'll review this.

zebra-chain/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+    #[error(
+        "failed to generate a diversified base point after {} tries",
+        DIVERSIFY_HASH_TRIES
+    )]
+    DiversifyHash,
+}
+
+/// Errors related to `AffinePoint`.
+#[derive(Error, Copy, Clone, Debug, PartialEq, Eq)]
+pub enum AffineError {
```
upbqdn: ###### Optional

I'd go with `AffinePointError` here. 


```suggestion
pub enum AffinePointError {
```

upbqdn (commented):
The error names make sense to me, but I'd rather let @dconnolly  give it the final approval.

teor2345:
This is waiting for a final review from Deirdre when she gets back.

zebra-chain/src/error.rs
upbqdn: Resolved in [421c9b8](https://github.com/ZcashFoundation/zebra/pull/6385/commits/421c9b86d6cacd1de6c4a73332ab29a175998d07).

zebra-chain/src/error.rs
oxarbitrage: It seems they do infinite loops to go over this issue. https://github.com/zcash/librustzcash/blob/3ff5fa52881d615e207a7dbe25591d36c8f2113b/zcash_primitives/src/transaction/components/sapling/builder.rs#L452

zebra-chain/src/sapling/commitment.rs
oxarbitrage: It seems it will not be needed here.

The `NoteCommitment::new` takes a `Diversifier` as an argument. Creating a `Diviersifier` should be done by `Diversifier::new` where we already have the loop.

`NoteCommitment::new` is currently not used in Zebra, i guess that there is a small risk of passing an invalid `Diversifier` here but in that case i think is ok to just error.

zebra-chain/src/error.rs
dconnolly: They also loop infinitely in related Orchard code

teor2345: It seems like we are talking about things that are extremely low probability here. An infinite loop will never happen unless the RNG is broken, at which point a hang is (arguably) the safest behaviour for users.

This is also the case for any number of loop iterations more than 2. (Assuming the ratio between the number of valid and invalid values is around 2^100 or greater.)

Memory bit errors are much more likely to cause problems in cryptographic software:
https://www.zdnet.com/article/dram-error-rates-nightmare-on-dimm-street/

So I would be comfortable going back to the auditors and saying "an infinite loop will never happen with a working RNG, we're going to follow what the rest of the ecosystem does here".

dconnolly: I'm asking some cryptographic peers their opinions on a hard number between 1 and infinity, as this does not seem to have a solid consensus

dconnolly: Let's go with 2 total tries:

```suggestion
pub const DIVERSIFY_HASH_TRIES: u8 = 2;
```

dconnolly: https://github.com/ZcashFoundation/zebra/pull/6385/files#r1170302073

dconnolly: I would actually move this const to the sapling/keys module as that's where it's used, including the number of tries in the error message is not something a general user can change or take an action on, so we don't strictly need it in the message

zebra-chain/src/sapling/commitment.rs
```diff
@@ -120,11 +123,9 @@ impl NoteCommitment {
...
+            .map_err(|_| CryptoError::Affine(AffinePointError::FromDiversifier))?
```
dconnolly: `AffinePointError` as a top-level `Error` is a bit ambiguous as we have several curve point types that can be represented in that manner; I would suggest having a `CryptoError::InvalidPoint` type that an `InvalidPointError` variant can be converted from

dconnolly:
@oxarbitrage @upbqdn @teor2345 I made some suggestions in https://github.com/ZcashFoundation/zebra/pull/6385/commits/ed6596a1b8c591872ec8d731d343978d04a87a53, lmk what you think?

teor2345 (dismissed):
Looks good to me, ~~just needs to resolve a merge conflict?~~

teor2345 (dismissed):

zebra-chain/src/orchard/commitment.rs
```diff
@@ -26,14 +27,16 @@ use super::sinsemilla::*;
...
 {
     let mut bytes = [0u8; 64];
-    csprng.fill_bytes(&mut bytes);
-    // pallas::Scalar::from_bytes_wide() reduces the input modulo q_P under the hood.
-    pallas::Scalar::from_uniform_bytes(&bytes)
+    csprng
+        .try_fill_bytes(&mut bytes)
+        .map_err(|_| RandError::FillBytes)?;
+    // pallas::Scalar::from_uniform_bytes() reduces the input modulo q_P under the hood.
+    ok(pallas::Scalar::from_uniform_bytes(&bytes))
```
teor2345: Not sure how this happened, maybe autocorrect?

```suggestion
    Ok(pallas::Scalar::from_uniform_bytes(&bytes))
```

teor2345 (dismissed):

daira:
I question the idea that we want to avoid a panic when a cryptographic RNG fails. https://github.com/ZcashFoundation/zebra/issues/6338#issuecomment-1514918706

zebra-chain/src/orchard/note.rs
```diff
@@ -57,14 +62,16 @@ impl From<Nullifier> for Rho {
...
         T: RngCore + CryptoRng,
     {
         let mut bytes = [0u8; 32];
-        csprng.fill_bytes(&mut bytes);
+        csprng
+            .try_fill_bytes(&mut bytes)
+            .map_err(|_| RandError::FillBytes)?;
 
-        Self(extract_p(pallas::Point::from_bytes(&bytes).unwrap()))
+        Ok(Self(extract_p(pallas::Point::from_bytes(&bytes).unwrap())))
```
teor2345: I don't think we should panic here, instead, we should handle the identity point by returning an error:
https://docs.rs/pasta_curves/latest/pasta_curves/struct.Ep.html#method.from_bytes

zebra-chain/src/orchard/commitment.rs
```diff
@@ -26,14 +27,16 @@ use super::sinsemilla::*;
...
 where
     T: RngCore + CryptoRng,
 {
     let mut bytes = [0u8; 64];
-    csprng.fill_bytes(&mut bytes);
-    // pallas::Scalar::from_bytes_wide() reduces the input modulo q_P under the hood.
-    pallas::Scalar::from_uniform_bytes(&bytes)
+    csprng
+        .try_fill_bytes(&mut bytes)
+        .map_err(|_| RandError::FillBytes)?;
```
teor2345, line 37: Daira suggests that the correct behaviour if a CSPRNG panics is to fail, to protect user funds:
https://github.com/ZcashFoundation/zebra/issues/6338#issuecomment-1514918706

This seems reasonable to me. @dconnolly can you make a decision on this?

If the correct behaviour is panicking, then we can revert the change to `try_fill_bytes`, and remove `RandError::FillBytes`. We can keep the error handling for invalid curve points.

dconnolly: Replied here: https://github.com/ZcashFoundation/zebra/issues/6338#issuecomment-1515552707

I do not think we should panic in `zebra-chain`, as it is a library. The consumers of the library (probably `zebra-consensus` or another dependent) should handle the error, and probably panic at _that_ layer. 

We may want to document this in the rustdoc or on the shielded coinbase tickets

teor2345: Thanks, I'll leave it to you or @mpguerra to update those tickets or open a PR with updated docs.

zebra-chain/src/orchard/note.rs
dconnolly: Done here: https://github.com/ZcashFoundation/zebra/pull/6385/commits/47a3adb948b751f51fd459e9317028ee7f25512b

teor2345 (approved):

PR #6387, opened 2023-03-22
change(state): Stop dropping blocks in state

arya2:
## Motivation

In PR https://github.com/ZcashFoundation/zebra/pull/6335, the state adds the hashes of queued finalized blocks to `sent_blocks` to check when responding to a `KnownBlock` request, and clears `sent_blocks` when there's a reset signal from the write task so blocks that could not be validated can be re-downloaded/verified.

This PR keeps queued finalized blocks in the state until they fall below the finalized tip height so Zebra won't need to re-download or re-verify them.

## Solution

- Puts `oneshot::Sender` into a `Arc<Mutex<Option<T>>>` 
- Sends a clone of `QueuedFinalized` to the block write task without removing it from `queued_finalized_blocks`
- Omits finalized hashes from `sent_blocks` (I'll revert the rename if we go with this approach)

## Review

This is meant to merge into https://github.com/ZcashFoundation/zebra/pull/6335 and needs some polish before it replaces adding finalized block hashes to `sent_blocks`.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -19,7 +19,7 @@ use std::{
     convert,
     future::Future,
     pin::Pin,
-    sync::Arc,
+    sync::{Arc, Mutex},
     task::{Context, Poll},
     time::{Duration, Instant},
 };
@@ -476,7 +476,7 @@ impl StateService {
         }
 
         let (rsp_tx, rsp_rx) = oneshot::channel();
-        let queued = (finalized, rsp_tx);
+        let queued = (finalized, Arc::new(Mutex::new(Some(rsp_tx))));
 
         if self.finalized_block_write_sender.is_some() {
             // We're still committing finalized blocks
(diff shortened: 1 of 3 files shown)


teor2345:
> In PR #6335, the state adds the hashes of queued finalized blocks to `sent_blocks` to check when responding to a `KnownBlock` request, and clears `sent_blocks` when there's a reset signal from the write task so blocks that could not be validated can be re-downloaded/verified.
> 
> This PR keeps queued finalized blocks in the state until they fall below the finalized tip height so Zebra won't need to re-download or re-verify them.

I'm not sure if this will work with the syncer, because it resets its entire queue on error, so it must re-download all the blocks above the error:
https://github.com/ZcashFoundation/zebra/blob/7eb08208eaf5a9a8ce3e3145d4e3a162d38c2fdc/zebrad/src/components/sync.rs#L508-L510

This seems like the correct behaviour in the presence of malicious or misbehaving peers.

PR #6410, opened 2023-03-27
fix(net): Avoid some self-connection nonce removal attacks

### LINKED ISSUE #6339: [NCC-E005955-MMC] `zebra-network`: Uncaught Nonce Reuse and Fragile Nonce Cache Eviction
## Impact

Failure to check if the nonce cache already contains the new nonce may prevent the self-connection detection from working as intended. Incorrectly evicting an entry from the nonce cache will have the same effect.

## Description

The function `negotiate_version()` in [zebra-network/src/peer/handshake.rs](https://github.com/ZcashFoundation/zebra/blob/audit-v1.0.0-rc.0/zebra-network/src/peer/handshake.rs#L571-L782) is used to negotiate the network version used when connecting to a new peer. In order to match outgoing messages with incoming responses, a nonce is included, and cached locally to help identify self-connection attempts:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/peer/handshake.rs#L584-L591

It was observed that the return value of the insert function is not checked. This function returns `true` if the value was successfully inserted, and `false` if the value was already contained in the set. Therefore, nonce reuse could be detected at this step by checking the return value of this function. The above freshly generates each nonce via `Nonce::default()`, so the probability of a collision is negligible, but it is nevertheless recommended to check the result of the insert operation as a precaution.

Later, in the same function, the following code handles received nonces:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/peer/handshake.rs#L677-L693

If the incoming message contains a nonce currently in the cache, that means it corresponds to a negotiation version message originating from itself, indicating a self-connection attempt. When this occurs, an error is returned and the nonce associated with the request is evicted from the cache.

Earlier, this finding highlighted a potential case where nonces could be re-used without correct detection. If such a nonce is evicted after the first connection attempt then a second connection attempt might succeed. Building on this observation, a malicious party may attempt to replay messages or craft messages containing an observed nonce, thereby
causing the nonce to be evicted incorrectly. Subsequently, a self-connection attempt would not be correctly identified.

Based on the above, it is recommended that the approach to nonce handling in this use case be re-evaluated to ensure the intended security goals are met. It may be safer to cache all nonces for a set duration to ensure that malicious or incorrect behavior cannot force nonce eviction from the cache.

## Recommendation

- Consider adding an explicit check for nonce reuse when adding values to the cache and returning `HandshakeError::NonceReuse` as necessary.
- Consider caching nonce entries for a longer period of time to ensure that nonces are not prematurely evicted from the cache.

## Location

[zebra-network/src/peer/handshake.rs](https://github.com/ZcashFoundation/zebra/blob/audit-v1.0.0-rc.0/zebra-network/src/peer/handshake.rs#L571-L782)

teor2345:
## Motivation

We want Zebra to continue to reject self-connections, even if a malicious peer replays one of our nonces.

Fixes #6339

This code changes concurrent futures with futures-based mutex locking.
The amount of work done in the critical sections is small.

## Solution

- Check for duplicate nonces from our internal random number generator
- Stop removing remote nonces after the first attempt that uses them: all attempts should be rejected
- Add a test that checks that self-connections fail

### Manual Testing

I ran:
```sh
cd zebra-network
RUST_LOG=info cargo test -- --nocapture self_connections_should_fail
```

And got these logs:
```
...
2023-04-19T02:36:56.524554Z  INFO crawl_and_dial{new_peer_interval=1s}: starting the peer address crawler crawl_new_peer_interval=1s outbound_connections=0
2023-04-19T02:36:56.525989Z  INFO accept_inbound_connections{min_inbound_peer_connection_interval=1s listener_addr=Ok(127.0.0.1:39689)}:listen_accept{peer=In("127.0.0.1:39760")}: rejecting self-connection attempt connected_addr=In("127.0.0.1:39760")
2023-04-19T02:36:56.526043Z  INFO accept_inbound_connections{min_inbound_peer_connection_interval=1s listener_addr=Ok(127.0.0.1:39689)}:listen_accept{peer=In("127.0.0.1:39760")}: rejecting self-connection attempt connected_addr=Out("127.0.0.1:39689")
...
```

This shows that both sides of a self-connection reject that connection with a nonce reuse error.

## Review

This is a routine audit fix.

### Reviewer Checklist

## Follow Up

Add more unit tests:
- self-connections fail specifically because of a  nonce rejection error
- multiple self-connections with the same nonce fail

These tests are complicated, so I don't think they are worth the effort to do now.

### DIFF
--- a/zebra-network/src/peer/handshake.rs
+++ b/zebra-network/src/peer/handshake.rs
@@ -595,13 +595,15 @@ where
     // It is ok to wait for the lock here, because handshakes have a short
     // timeout, and the async mutex will be released when the task times
     // out.
-    //
-    // Duplicate nonces don't matter here, because 64-bit random collisions are very rare.
-    // If they happen, we're probably replacing a leftover nonce from a failed connection,
-    // which wasn't cleaned up when it closed.
     {
         let mut locked_nonces = nonces.lock().await;
-        locked_nonces.insert(local_nonce);
+
+        // Duplicate nonces are very rare, because they require a 64-bit random number collision,
+        // and the nonce set is limited to a few hundred entries.
+        let is_unique_nonce = locked_nonces.insert(local_nonce);
+        if !is_unique_nonce {
+            return Err(HandshakeError::LocalDuplicateNonce);
+        }
 
         // # Security
         //
@@ -615,7 +617,7 @@ where
         // - avoiding memory denial of service attacks which make large numbers of connections,
         //   for example, 100 failed inbound connections takes 1 second.
         // - memory usage: 16 bytes per `Nonce`, 3.2 kB for 200 nonces
-        // - collision probability: 2^32 has ~50% collision probability, so we use a lower limit
+        // - collision probability: two hundred 64-bit nonces have a very low collision probability
         //   <https://en.wikipedia.org/wiki/Birthday_problem#Probability_of_a_shared_birthday_(collision)>
         while locked_nonces.len() > config.peerset_total_connection_limit() {
             locked_nonces.shift_remove_index(0);
@@ -731,7 +733,7 @@ where
         nonce_reuse
     };
     if nonce_reuse {
-        Err(HandshakeError::NonceReuse)?;
+        Err(HandshakeError::RemoteNonceReuse)?;
     }
 
     // SECURITY: Reject connections to peers on old versions, because they might not know about all
--- a/zebra-network/src/peer/handshake.rs
+++ b/zebra-network/src/peer/handshake.rs
@@ -595,6 +595,8 @@ where
     // It is ok to wait for the lock here, because handshakes have a short
     // timeout, and the async mutex will be released when the task times
     // out.
+    //
+    // START CRITICAL SECTION
     {
         let mut locked_nonces = nonces.lock().await;
 
@@ -624,7 +626,8 @@ where
         }
 
         std::mem::drop(locked_nonces);
-    };
+    }
+    // END CRITICAL SECTION
 
     // Don't leak our exact clock skew to our peers. On the other hand,
     // we can't deviate too much, or zcashd will get confused.
@@ -723,20 +726,16 @@ where
     //
     // # Security
     //
-    // Connections that get a `Version` message will remove their nonces here,
-    // but connections that fail before this point can leave their nonces in the nonce set.
-    let nonce_reuse = {
-        let mut locked_nonces = nonces.lock().await;
-        let nonce_reuse = locked_nonces.contains(&remote.nonce);
-        // Regardless of whether we observed nonce reuse, remove our own nonce from the nonce set.
-        locked_nonces.remove(&local_nonce);
-        nonce_reuse
-    };
+    // We don't remove the nonce here, because peers that observe our network traffic could
+    // maliciously remove nonces, and force us to make self-connections.
+    let nonce_reuse = nonces.lock().await.contains(&remote.nonce);
     if nonce_reuse {
         Err(HandshakeError::RemoteNonceReuse)?;
     }
 
-    // SECURITY: Reject connections to peers on old versions, because they might not know about all
+    // # Security
+    //
+    // Reject connections to peers on old versions, because they might not know about all
     // network upgrades and could lead to chain forks or slower block propagation.
     let min_version = minimum_peer_version.current();
     if remote.version < min_version {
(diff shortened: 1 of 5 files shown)


zebra-network/src/peer/handshake.rs
```diff
@@ -596,12 +596,16 @@ where
     // timeout, and the async mutex will be released when the task times
     // out.
     //
-    // Duplicate nonces don't matter here, because 64-bit random collisions are very rare.
-    // If they happen, we're probably replacing a leftover nonce from a failed connection,
-    // which wasn't cleaned up when it closed.
+    // START CRITICAL SECTION
```
oxarbitrage: Can you explain why this is considered critical section and if the plan is to use this "tag" for other critical zones of our codebase ?

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1273,14 +1400,19 @@ async fn local_listener_port_with(listen_addr: SocketAddr, network: Network) {
     );
 }
 
-/// Initialize a peer set with `peerset_initial_target_size` and `inbound_service` on `network`.
-/// Returns the newly created [`AddressBook`] for testing.
+/// Initialize a peer set with`peerset_initial_target_size`, `inbound_service`, and `network`.
```
oxarbitrage:
```suggestion
/// Initialize a peer set with `peerset_initial_target_size`, `inbound_service`, and `network`.
```

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1095,6 +1097,128 @@ async fn add_initial_peers_is_rate_limited() {
...
+        TEST_PEERSET_INITIAL_TARGET_SIZE,
+        unreachable_inbound_service,
+        Mainnet,
+        force_listen_addr,
+        no_initial_peers_config,
+    )
+    .await;
+
+    // Insert our own address into the address book, and make sure it works
+    // BEGIN CRITICAL SECTION
```
oxarbitrage:
```suggestion
```

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1095,6 +1097,128 @@ async fn add_initial_peers_is_rate_limited() {
...
+        unlocked_address_book.set_local_listener("192.168.0.0:1".parse().unwrap());
+
+        let updated_addr = unlocked_address_book.update(
+            real_self_listener
+                .new_gossiped_change()
+                .expect("change is valid"),
+        );
+
+        std::mem::drop(unlocked_address_book);
+        // END CRITICAL SECTION
```
oxarbitrage:
```suggestion
```

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1095,6 +1097,128 @@ async fn add_initial_peers_is_rate_limited() {
...
+        );
+
+        real_self_listener
+    };
+
+    // Wait until the crawler has tried at least one self-connection
+    tokio::time::sleep(TEST_CRAWL_NEW_PEER_INTERVAL * 3).await;
+
+    // Check that the self-connection failed
+    // BEGIN CRITICAL SECTION
```
oxarbitrage:
```suggestion
```

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1095,6 +1097,128 @@ async fn add_initial_peers_is_rate_limited() {
...
+        let mut unlocked_address_book = address_book
+            .lock()
+            .expect("unexpected panic in address book");
+
+        let self_connection_status = unlocked_address_book
+            .get(&real_self_listener.addr())
+            .expect("unexpected dropped listener address in address book");
+
+        std::mem::drop(unlocked_address_book);
+        // END CRITICAL SECTION
```
oxarbitrage:
```suggestion
```

oxarbitrage (dismissed):
Looks good to me, i made a few optional documentation suggestions.

zebra-network/src/peer/handshake.rs
teor2345: https://en.wikipedia.org/wiki/Critical_section

Usually we just use a block with a comment above it, I can switch to that if you'd like.

oxarbitrage (approved):

PR #6444, opened 2023-04-01
fix(script): updates search-issue-refs to group references by issue and output markdown

arya2:
## Motivation

It's more convenient to view all references to a closed issue at once.

## Solution

- Group references by issue
- Only output issue ids and permalinks to where they were found in the repo

## Review

This is optional priority, anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-utils/src/bin/search-issue-refs/main.rs
+++ b/zebra-utils/src/bin/search-issue-refs/main.rs
@@ -33,6 +33,7 @@
 //! > Found 3 references to closed issues.
 
 use std::{
+    collections::HashMap,
     env,
     ffi::OsStr,
     fs::{self, File},
@@ -89,8 +90,18 @@ fn github_issue_url(issue_id: &str) -> String {
 }
 
 fn github_remote_file_ref(file_path: &str, line: usize) -> String {
+    let file_path = &crate_mod_path(file_path, line);
+    format!("https://github.com/ZcashFoundation/zebra/blob/main/{file_path}")
+}
+
+fn github_permalink(file_path: &str, line: usize) -> String {
+    let file_path = &crate_mod_path(file_path, line);
+    format!("https://github.com/ZcashFoundation/zebra/blob/5db2243c25cdde03d9cdccd72a137ccc6f2370a5/{file_path}")
+}
+
+fn crate_mod_path(file_path: &str, line: usize) -> String {
     let file_path = &file_path[2..];
-    format!("https://github.com/ZcashFoundation/zebra/blob/main/{file_path}#L{line}")
+    format!("{file_path}#L{line}")
 }
 
 fn github_issue_api_url(issue_id: &str) -> String {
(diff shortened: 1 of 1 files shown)


zebra-utils/src/bin/search-issue-refs/main.rs
```diff
@@ -89,8 +90,18 @@ fn github_issue_url(issue_id: &str) -> String {
 }
 
 fn github_remote_file_ref(file_path: &str, line: usize) -> String {
+    let file_path = &crate_mod_path(file_path, line);
+    format!("https://github.com/ZcashFoundation/zebra/blob/main/{file_path}")
+}
+
+fn github_permalink(file_path: &str, line: usize) -> String {
+    let file_path = &crate_mod_path(file_path, line);
+    format!("https://github.com/ZcashFoundation/zebra/blob/5db2243c25cdde03d9cdccd72a137ccc6f2370a5/{file_path}")
```
oxarbitrage, line 99: I don't think this should be hard coded as we might want to run this script in the future where new `TODO`s might appear.  We should just link to main and assume the links are temporal or use the current last commit.

oxarbitrage (approved):
I think this is better, thanks for adjusting it. Just an issue with the permalink that i think we should change but it maybe has a reason i am not seeing. Feel free to clarify and merge.

zebra-utils/src/bin/search-issue-refs/main.rs
arya2: I'm not sure how to get a blob id for the latest commit on main, it seems easy enough to update the hardcoded value but let me know if there's a better way to get the blob.

(The reason for using the permalink is for generating previews when posted on Github.)

teor2345: If you use the JSON REST API:
https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28

A request like:
```sh
curl -L \
-H "Accept: application/vnd.github+json" \
-H "X-GitHub-Api-Version: 2022-11-28" \
https://api.github.com/repos/ZcashFoundation/zebra/git/ref/heads/main
```

will get you a response like:
```json
{
  "ref": "refs/heads/main",
  "node_id": "MDM6UmVmMjA1MjU1NjgzOnJlZnMvaGVhZHMvbWFpbg==",
  "url": "https://api.github.com/repos/ZcashFoundation/zebra/git/refs/heads/main",
  "object": {
    "sha": "d1990ca98fbd0a1d35a3244fc2d67dc6f8c5b4e1",
    "type": "commit",
    "url": "https://api.github.com/repos/ZcashFoundation/zebra/git/commits/d1990ca98fbd0a1d35a3244fc2d67dc6f8c5b4e1"
  }
}
```

And then you can parse the json or run it through another tool to get the commit hash:
```sh
$ curl ... | jq -r .object.sha
d1990ca98fbd0a1d35a3244fc2d67dc6f8c5b4e1
```

arya2: Thank you!! Fixed/applied in https://github.com/ZcashFoundation/zebra/pull/6494.

PR #6447, opened 2023-04-02
security(state): Limit the number of non-finalized chains tracked by Zebra

teor2345:
## Motivation

There is no limit on the number of non-finalized chains that Zebra can track.

Theoretically, this allows unbounded use of memory. But this bug is difficult to exploit on mainnet, because it would require mining hundreds of side-chain blocks at mainnet difficulty.

### Specifications

The number of chains tracked by a node implementation is a standard rule, not a consensus rule. For `zcashd`, it is currently 1 in memory, and some number on disk.

For Zebra, after this PR it will be 10 in memory. (Zebra doesn't store any unverified or side-chain blocks on disk.)

This is concurrent code, the non-finalized state is cloned before reading it concurrently. We ensure that the limit of 10 is kept by making the list of chains into a private field, and only accessing them via a method.

## Solution

- Limit the number of non-finalized chains to 10
  - If the limit is reached, drop the chain with the least work
  - Make the chain list private, and provide iterator, count, and insert chain methods

## Review

This is a low priority security fix.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -87,9 +87,9 @@ impl NonFinalizedState {
             && self.network == other.network
     }
 
-    /// Returns an iterator over the non-finalized chains.
-    #[cfg(any(test, feature = "proptest-impl"))]
-    #[allow(dead_code)]
+    /// Returns an iterator over the non-finalized chains, with the best chain first.
+    //
+    // TODO: replace chain_set.iter().rev() with this method
     pub fn chain_iter(&self) -> impl Iterator<Item = &Arc<Chain>> {
         self.chain_set.iter()
     }
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -39,7 +39,8 @@ pub(crate) use chain::Chain;
 pub struct NonFinalizedState {
     /// Verified, non-finalized chains, in ascending order.
     ///
-    /// The best chain is `chain_set.last()` or `chain_set.iter().next_back()`.
+    /// The best chain is `chain_iter().next()`.
+    /// Using `chain_set.last()` or `chain_set.iter().next_back()` is deprecated, and should migrate to `chain_iter().next()`.
     chain_set: BTreeSet<Arc<Chain>>,
 
     /// The configured Zcash network.
@@ -91,7 +92,7 @@ impl NonFinalizedState {
     //
     // TODO: replace chain_set.iter().rev() with this method
     pub fn chain_iter(&self) -> impl Iterator<Item = &Arc<Chain>> {
-        self.chain_set.iter()
+        self.chain_set.iter().rev()
     }
 
     /// Insert `chain` into `self.chain_set`, apply `chain_filter` to the chains,
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -15,6 +15,7 @@ use zebra_chain::{
 };
 
 use crate::{
+    constants::MAX_NON_FINALIZED_CHAIN_FORKS,
     request::{ContextuallyValidBlock, FinalizedWithTrees},
     service::{check, finalized_state::ZebraDb},
     PreparedBlock, ValidateContextError,
@@ -39,7 +40,7 @@ pub struct NonFinalizedState {
     /// Verified, non-finalized chains, in ascending order.
     ///
     /// The best chain is `chain_set.last()` or `chain_set.iter().next_back()`.
-    pub chain_set: BTreeSet<Arc<Chain>>,
+    chain_set: BTreeSet<Arc<Chain>>,
 
     /// The configured Zcash network.
     pub network: Network,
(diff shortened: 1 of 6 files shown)


zebra-state/src/service/read/find.rs
```diff
@@ -107,7 +107,7 @@ pub fn non_finalized_state_contains_block_hash(
     non_finalized_state: &NonFinalizedState,
     hash: block::Hash,
 ) -> Option<KnownBlock> {
-    let mut chains_iter = non_finalized_state.chain_set.iter().rev();
+    let mut chains_iter = non_finalized_state.chain_iter();
```
arya2, line 110: This needs to be reversed here to correctly return `BestChain` or `SideChain`, I think it's optional everywhere else it's been updated (all tests) so we could also return `self.chain_set.iter().rev()` from `chain_iter()`.

```suggestion
    let mut chains_iter = non_finalized_state.chain_iter().rev();
```

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -86,6 +87,34 @@ impl NonFinalizedState {
...
+    /// Returns an iterator over the non-finalized chains, with the best chain first.
+    //
+    // TODO: replace chain_set.iter().rev() with this method
+    pub fn chain_iter(&self) -> impl Iterator<Item = &Arc<Chain>> {
+        self.chain_set.iter()
+    }
+
+    /// Insert `chain` into `self.chain_set`, apply `chain_filter` to the chains,
+    /// then limit the number of tracked chains.
+    fn insert_with<F>(&mut self, chain: Arc<Chain>, chain_filter: F)
```
arya2, line 100: Optional: Would an insert method that returns a drop guard for pruning the chain set be nice here?

```rust
struct PruneGaurd<'a>(&'a mut BTreeSet<Arc<Chain>>);

impl<'a> PruneGaurd<'a> {
    fn apply_and_prune<F: FnOnce(&mut BTreeSet<Arc<Chain>>)>(self, chain_filter: F) {
        chain_filter(self.0);
    }
}

impl<'a> Drop for PruneGaurd<'a> {
    fn drop(&mut self) {
        while self.0.len() > MAX_NON_FINALIZED_CHAIN_FORKS {
            // The first chain is the chain with the lowest work.
            self.0.pop_first();
        }
    }
}

/// Insert `chain` into `self.chain_set`, then limit the number of tracked chains.
fn insert(&mut self, chain: Arc<Chain>) -> PruneGaurd {
    self.chain_set.insert(chain);

    PruneGaurd(&mut self.chain_set)
}

pub fn commit_block(..) -> Result<(), ValidateContextError> {
    ..
    self.insert(modified_chain).apply_and_prune(|chain_set| {
        chain_set.retain(|chain| chain.non_finalized_tip_hash() != parent_hash)
    });
    ..
}
```

arya2 (commented):
This looks excellent!

zebra-state/src/service/non_finalized_state.rs
teor2345: Drop guards aren't guaranteed to run, so it would be a less reliable implementation. What you're looking for is called a "linear type", which isn't implemented in Rust yet:
https://blog.yoshuawuyts.com/linearity-and-control/#what-are-linear-types

This particular implementation would also hold a mutable reference to the chains for the entire calling method, which would likely cause problems with the borrow-checker.

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -39,7 +40,7 @@ pub struct NonFinalizedState {
     /// Verified, non-finalized chains, in ascending order.
     ///
     /// The best chain is `chain_set.last()` or `chain_set.iter().next_back()`.
```
teor2345:
```suggestion
    /// The best chain is `chain_iter().next()`.
    /// Using `chain_set.last()` or `chain_set.iter().next_back()` is deprecated, and should migrate to `chain_iter().next()`.
```

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -86,6 +87,34 @@ impl NonFinalizedState {
             && self.network == other.network
     }
 
+    /// Returns an iterator over the non-finalized chains, with the best chain first.
+    //
+    // TODO: replace chain_set.iter().rev() with this method
+    pub fn chain_iter(&self) -> impl Iterator<Item = &Arc<Chain>> {
+        self.chain_set.iter()
```
teor2345:
```suggestion
        self.chain_set.iter().rev()
```

zebra-state/src/service/read/find.rs
teor2345: Thanks, that was the original implementation, but I didn't catch it in the cherry-pick.

Fixed in de1269f59f6edb1a032b39a4ba40b70c87424b6c.

arya2 (approved):

teor2345:
Temporary network errors:
> warning: spurious network error (2 tries remaining): early EOF; class=Net (12); code=Eof (-20)
> Warning: warning: spurious network error (1 tries remaining): early EOF; class=Net (12); code=Eof (-20)
> error: Unable to update registry `crates-io`
> Error: Unable to update registry `crates-io`
> Caused by:
>   failed to fetch `[https://github.com/rust-lang/crates.io-index`](https://github.com/rust-lang/crates.io-index%60)

https://github.com/ZcashFoundation/zebra/actions/runs/4601125572/jobs/8128642182#step:5:15

> Error: -03 20:40:13 [ERROR] `cargo metadata` exited with an error:     Updating crates.io index
> warning: spurious network error (2 tries remaining): early EOF; class=Net (12); code=Eof (-20)
> warning: spurious network error (1 tries remaining): early EOF; class=Net (12); code=Eof (-20)
> error: Unable to update registry `crates-io`
> 
> Caused by:
>   failed to fetch `[https://github.com/rust-lang/crates.io-index`](https://github.com/rust-lang/crates.io-index%60)

https://github.com/ZcashFoundation/zebra/actions/runs/4601125562/jobs/8128592665?pr=6447#step:4:14

zebra-state/src/service/non_finalized_state.rs
arya2: > Drop guards aren't guaranteed to run

I think they would be in this case?

The apply fn is effectively the same as [`pub fn drop<T>(_x: T) {}`](https://doc.rust-lang.org/stable/src/core/mem/mod.rs.html#980), and it should be dropped immediately if it's unassigned?

teor2345: > > Drop guards aren't guaranteed to run
> 
> I think they would be in this case?
> 
> The apply fn is effectively the same as [`pub fn drop<T>(_x: T) {}`](https://doc.rust-lang.org/stable/src/core/mem/mod.rs.html#980), and it should be dropped immediately if it's unassigned?

Yes, in this implementation it would be dropped eventually. But with an API like this, we have to make sure we drop the guard at runtime, rather than storing it or forgetting it. And there's no way to tell the compiler to verify we've done that.

The guard can also be dropped at different times depending on how we assign, match, or ignore the return value. That probably doesn't matter here, but it requires extra thought to make sure.

So I went for the simple check that's obviously correct.

PR #6448, opened 2023-04-02
fix(mempool): Stop ignoring new transactions after the mempool is newly activated

teor2345:
## Motivation

Currently, when Zebra's mempool is newly activated, it skips adding newly verified transactions until the next request.

Instead, the first request to a newly activated mempool should show all available transactions.

This could be related to #6195, but I'm not sure if it is.

### Specifications

Mempool transaction selection is a standard rule, so either implementation is valid here.

## Solution

Only skip the reset when the mempool is newly activated, rather than skipping all of `poll_ready()`.

## Review

This is a routine mempool bug.

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/mempool.rs
+++ b/zebrad/src/components/mempool.rs
@@ -319,14 +319,11 @@ impl Service<Request> for Mempool {
 
         let tip_action = self.chain_tip_change.last_tip_change();
 
-        // If the mempool was just freshly enabled,
-        // skip resetting and removing mined transactions for this tip.
-        if is_state_changed {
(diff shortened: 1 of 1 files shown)


teor2345:
Temporary error:
> - Internal error. Please try again or contact Google Support. (Code: '5F864782189D7.6B5868C.50A111B')

https://github.com/ZcashFoundation/zebra/actions/runs/4591441080/jobs/8108460458?pr=6448#step:10:83

arya2 (approved):
Looks good!

PR #6450, opened 2023-04-03
fix(state): Remove an outdated anchor assertion in a comment

### LINKED ISSUE #6389: Re-enable or delete the assertion in `fetch_sprout_final_treestates()`
## Motivation

The auditors have suggested that this commented-out assertion should either be deleted, or enabled but disabled in tests:
https://github.com/ZcashFoundation/zebra/blob/c838383fd53610da01bfee4f7555f219f84853d0/zebra-state/src/service/check/anchors.rs#L158-L172

We can do this using `#[cfg(not(test))]`.

### Testing

Our existing integration tests should cover this.

teor2345:
## Motivation

There is an outdated assertion in the state Sprout anchor checking code, which has already been commented out.

This assertion could be expensive, so we can only do it during tests.

Close #6389

### Specifications

This assertion is not consensus-critical, the existing production code already constructs the anchor map correctly.

## Solution

- Remove the commented-out assertion
- Update the TODO to only apply to tests

## Review

This is an audit fix, it should be reviewed first.

### Reviewer Checklist

## Follow Up Work

We could fix the tests, but it's not important at all.

### DIFF
--- a/zebra-state/src/service/check/anchors.rs
+++ b/zebra-state/src/service/check/anchors.rs
@@ -155,23 +155,14 @@ fn fetch_sprout_final_treestates(
             .or_else(|| finalized_state.sprout_note_commitment_tree_by_anchor(&joinsplit.anchor));
 
         if let Some(input_tree) = input_tree {
-            /* TODO:
-                 - fix tests that generate incorrect root data
-                 - assert that roots match the fetched tree during tests
-                 - move this CPU-intensive check to sprout_anchors_refer_to_treestates()
-
-            assert_eq!(
-                input_tree.root(),
-                joinsplit.anchor,
-                "anchor and fetched input tree root did not match:\n\
-                 anchor: {anchor:?},\n\
-                 input tree root: {input_tree_root:?},\n\
-                 input_tree: {input_tree:?}",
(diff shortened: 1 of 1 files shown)


zebra-state/src/service/check/anchors.rs
```diff
@@ -155,24 +155,15 @@ fn fetch_sprout_final_treestates(
...
-            );
-             */
-
             sprout_final_treestates.insert(joinsplit.anchor, input_tree);
 
+            /* TODO:
+               - fix tests that generate incorrect root data
+               - assert that joinsplit.anchor matches input_tree.root() during tests,
+                 but don't assert in production, because the check is CPU-intensive,
+                 and sprout_anchors_refer_to_treestates() constructs the map correctly
```
arya2, line 164: Do you want to uncomment the `assert_eq!()` as a `debug_assert_eq!()` here instead of removing the comment?

arya2 (approved):
This looks good!

zebra-state/src/service/check/anchors.rs
teor2345: No, because the assert fails for some tests, so the tests need to be fixed first.

PR #6460, opened 2023-04-03
fix(net): Fix off-by-one error in DNS seed peer retries, and clarify logs

### LINKED ISSUE #6393: [NCC-E005955-VM7] `zebra-network`: Off-by-One Error in `zebra-network` Retry Parameter
## Impact
The parameter `MAX_SINGLE_PEER_RETRIES` may not behave as expected due to an off-by-one error.

## Description

Finding [Off-by-One Errors and Inconsistent Usage of `PARAMETER_DOWNLOAD_MAX_RETRIES`](https://github.com/ZcashFoundation/zebra/issues/6340) documented an off-by-one issue with retry behavior in `zebra-consensus`. This finding documents a similar issue in `zebra-network`. Note that this does not represent a security issue or vulnerability, outside of potential user error during configuration.

The following code is part of peer DNS resolution, where each peer is resolved individually, and the complete process is repeated if no resolution is successful. The parameter `MAX_SINGLE_PEER_RETRIES` implies that it dictates the number of retry attempts for a given peer.

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/config.rs#L25-L27

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/config.rs#L161-L170

The `retry` function loops using `for retry_count in 1..=max_retries`, meaning that the value `MAX_SINGLE_PEER_RETRIES` will result in 1 single attempt and not 1 retry. Therefore, this parameter may be better named as `MAX_SINGLE_PEER_ATTEMPTS` or the behavior could be updated to loop one additional time.

## Recommendation

Ensure the naming of the parameter accurately reflects its usage. Either rename the parameter to reflect the maximum number of connection attempts or revise the behavior to reflect its current name.

## Location

[zebra-network/src/config.rs](https://github.com/ZcashFoundation/zebra/blob/audit-v1.0.0-rc.0/zebra-network/src/config.rs)

teor2345:
## Motivation

Zebra's DNS seed peer retries aren't documented or named correctly: the number of DNS attempts implied by the name doesn't match the implementation.

Closes #6393

This is concurrent async code, but this PR doesn't change the number of retries.

## Solution

- Change the value of the constant and the implementation of the loop to actually implement retries, not attempts
- Don't wait for the next retry if we're not actually going to retry
- Rename the constant to clarify what it does
- Split the log messages to clarify what is happening

## Review

This is a routine audit fix.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/config.rs
+++ b/zebra-network/src/config.rs
@@ -1,3 +1,5 @@
+//! Configuration for Zebra's network communication.
+
 use std::{
     collections::HashSet,
     net::{IpAddr, SocketAddr},
@@ -24,7 +26,14 @@ mod tests;
 
 /// The number of times Zebra will retry each initial peer's DNS resolution,
 /// before checking if any other initial peers have returned addresses.
-const MAX_SINGLE_PEER_RETRIES: usize = 1;
+///
+/// After doing this number of retries of a failed single peer, Zebra will
+/// check if it has enough peer addresses from other seed peers. If it has
+/// enough addresses, it won't retry this peer again.
+///
+/// If the number of tries is `0`, other peers are checked after every successful
+/// or failed DNS attempt.
+const MAX_SINGLE_SEED_PEER_DNS_RETRIES: usize = 0;
 
 /// Configuration for networking code.
 #[derive(Clone, Debug, Serialize)]
@@ -164,7 +173,7 @@ impl Config {
             // address peers. Individual retries avoid this issue.
             let peer_addresses = peers
                 .iter()
-                .map(|s| Config::resolve_host(s, MAX_SINGLE_PEER_RETRIES))
+                .map(|s| Config::resolve_host(s, MAX_SINGLE_SEED_PEER_DNS_RETRIES))
                 .collect::<futures::stream::FuturesUnordered<_>>()
                 .concat()
                 .await;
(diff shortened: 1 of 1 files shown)


zebra-network/src/config.rs
```diff
@@ -24,7 +26,14 @@ mod tests;
 
 /// The number of times Zebra will retry each initial peer's DNS resolution,
 /// before checking if any other initial peers have returned addresses.
-const MAX_SINGLE_PEER_RETRIES: usize = 1;
+///
+/// After doing this number of retries of a failed single peer, Zebra will
+/// check if it has enough peer addresses from other seed peers. If it has
+/// enough addresses, it won't retry this peer again.
+///
+/// If the number of tries is `0`, other peers are checked after every successful
```
teor2345:
```suggestion
/// If the number of retries is `0`, other peers are checked after every successful
```

zebra-network/src/config.rs
```diff
@@ -188,12 +197,25 @@ impl Config {
...
+                );
+                tokio::time::sleep(DNS_LOOKUP_TIMEOUT).await;
+            } else {
+                tracing::info!(
+                    ?host,
+                    attempts = ?(retries + 1),
+                    "Seed peer DNS resolution failed, checking for addresses from other seed peers",
+                );
+            }
         }
```
arya2, line 219: Nitpick/Optional: refactor this to remove the if statement

```suggestion
        if let Ok(addresses) = Config::resolve_host_once(host).await {
            return addresses;
        };

        for retries in 1..=max_retries {
            tracing::info!(
                ?host,
                previous_attempts = ?(retries + 1),
                "Waiting {DNS_LOOKUP_TIMEOUT:?} to retry seed peer DNS resolution",
            );
            tokio::time::sleep(DNS_LOOKUP_TIMEOUT).await;

            if let Ok(addresses) = Config::resolve_host_once(host).await {
                return addresses;
            };
        }

        tracing::info!(
            ?host,
            attempts = ?(max_retries + 1),
            "Seed peer DNS resolution failed, checking for addresses from other seed peers",
        );
```

arya2 (approved):
Looks good!

zebra-network/src/config.rs
teor2345: I usually prefer to duplicate logging rather than duplicating the action outside the loop, because then it's possible for the duplicate actions to get out of sync.

(If the logging gets out of sync it doesn't matter as much.)

PR #6475, opened 2023-04-08
fix(chain): Validate header versions when serializing blocks

### LINKED ISSUE #497: [NCC-E005955-M2F] `zebra-chain`: Check that header block version field is valid when serializing blocks
## Impact

Zebra assumes all serialization/deserialization functions cannot produce invalid data. Unenforced constraints violate this assumption and may affect consensus or interoperability.

## Description

Zebra implements the `ZCashSerialize` trait for consensus-critical data.

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/serialization/zcash_serialize.rs#L19-L30

Note the highlighted requirement that illegal states must be unrepresentable.

The function `zcash_deserialize()` for `Header` contains historical information regarding the version field of the header. In particular, it specifies that the only valid version number is 4, but that incorrect implementations in the past resulted in several blocks with an incorrect value. The function has some special case handling for this, as well as an explicit check that the parsed version is greater than or equal to 4:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/block/serialize.rs#L75-L82

This is similarly captured in other annotations for the version field, where it is specified that:

> The only constraint is that it must be at least `4` when interpreted as an `i32` .

In comparison, the corresponding `zcash_deserialize()` function does not perform validation on the version field:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-chain/src/block/serialize.rs#L28-L43

No instances within the codebase will currently set this value incorrectly, but it appears to be a violation of the specified constraint. It is possible to construct a `Header` such that the output of `zcash_serialize` will not correctly deserialize within Zebra.

## Recommendation

Align the constraint checks for the version field such that any serialized Header will correctly deserialize without an error.

## Location

[zebra-chain/src/block/serialize.rs](https://github.com/ZcashFoundation/zebra/blob/audit-v1.0.0-rc.0/zebra-chain/src/block/serialize.rs)

## Original issue description

Once Zebra starts generating its own blocks, we'll need to do the version check in a few more places:
- [ ] make sure serialized blocks contain a version that's 4 or greater, based on the consensus rules
- [x] impose a stricter version check on generated blocks, to ensure the version is always 4

This task is open to contributors, but some parts of it depend on future Zebra features.

arya2:
## Motivation

This PR updates the `ZcashSerialize` implementation for block headers to apply to same constraints as `ZcashDeserialize`.

Closes #497.

## Solution

- Splits version constraint checks in `zcash_deserialize` to a `check_version` function
- Calls `check_version()` from `zcash_serialize`

Related changes:
- Removes `LOW_31_BITS` constant and masking of high bit when deserializing (a high bit causes an error anyways)

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/serialize.rs
+++ b/zebra-chain/src/block/serialize.rs
@@ -22,7 +22,16 @@ use crate::{
 /// transaction in the chain is approximately 1.5 kB smaller.)
 pub const MAX_BLOCK_BYTES: u64 = 2_000_000;
 
-/// Checks if a block header version is valid
+/// Checks if a block header version is valid.
+///
+/// Zebra could encounter a [`Header`] with an invalid version when serializing a block header constructed
+/// in memory with the wrong version in tests or the getblocktemplate RPC.
+///
+/// The getblocktemplate RPC generates a template with version 4. The miner generates the actual block,
+/// and then we deserialize it and do this check.
+///
+/// All other blocks are deserialized when we receive them, and never modified,
+/// so the deserialisation would pick up any errors.
 fn check_version(version: u32) -> Result<(), &'static str> {
     match version {
         // The Zcash specification says that:
--- a/zebra-chain/src/block/serialize.rs
+++ b/zebra-chain/src/block/serialize.rs
@@ -22,9 +22,40 @@ use crate::{
 /// transaction in the chain is approximately 1.5 kB smaller.)
 pub const MAX_BLOCK_BYTES: u64 = 2_000_000;
 
+/// Checks if a block header version is valid
+fn check_version(version: u32) -> Result<(), &'static str> {
+    match version {
+        // The Zcash specification says that:
+        // "The current and only defined block version number for Zcash is 4."
+        // but this is not actually part of the consensus rules, and in fact
+        // broken mining software created blocks that do not have version 4.
+        // There are approximately 4,000 blocks with version 536870912; this
+        // is the bit-reversal of the value 4, indicating that that mining pool
+        // reversed bit-ordering of the version field. Because the version field
+        // was not properly validated, these blocks were added to the chain.
+        //
+        // The only possible way to work around this is to do a similar hack
+        // as the overwintered field in transaction parsing, which we do here:
+        // treat the high bit (which zcashd interprets as a sign bit) as an
+        // indicator that the version field is meaningful.
+        version if version >> 31 != 0 => Err("high bit was set in version field"),
+
+        // # Consensus
+        //
+        // > The block version number MUST be greater than or equal to 4.
+        //
+        // https://zips.z.cash/protocol/protocol.pdf#blockheader
+        version if version < ZCASH_BLOCK_VERSION => Err("version must be at least 4"),
+
+        _ => Ok(()),
+    }
+}
+
 impl ZcashSerialize for Header {
     #[allow(clippy::unwrap_in_result)]
     fn zcash_serialize<W: io::Write>(&self, mut writer: W) -> Result<(), io::Error> {
+        check_version(self.version).map_err(|msg| io::Error::new(io::ErrorKind::Other, msg))?;
+
         writer.write_u32::<LittleEndian>(self.version)?;
         self.previous_block_hash.zcash_serialize(&mut writer)?;
         writer.write_all(&self.merkle_root.0[..])?;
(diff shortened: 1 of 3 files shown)


oxarbitrage (commented):
It looks good, thank you for the fix.

`wrong_checkpoint_hash_fail_test` is failing. This seems to be because Zebra now panic in the serialization instead of an error in the verifier.

Can be fixed by changed the version to `5` at https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/checkpoint/tests.rs#L507

zebra-chain/src/block/serialize.rs
```diff
@@ -22,9 +22,40 @@ use crate::{
...
 pub const MAX_BLOCK_BYTES: u64 = 2_000_000;
 
+/// Checks if a block header version is valid
+fn check_version(version: u32) -> Result<(), &'static str> {
+    match version {
+        // The Zcash specification says that:
+        // "The current and only defined block version number for Zcash is 4."
+        // but this is not actually part of the consensus rules, and in fact
+        // broken mining software created blocks that do not have version 4.
+        // There are approximately 4,000 blocks with version 536870912; this
```
teor2345, line 41: Can you add documentation to explain when Zebra could encounter invalid versions when serializing a block header?

I think these errors could only happen if we constructed a generated block in memory with the wrong version. All other blocks are deserialized when we receive them, and never modified. So the deserialisation would pick up any errors.

And in the getblocktemplate RPC, we generate a template with version 4, then the miner generates the actual block, and then we deserialize it and do this check.

Can we also add one of the high version blocks to our test vectors, and add a test that serializes it? That will make sure we don't panic anywhere in that code path.

teor2345 (commented):
Looks good, I just want to check we handle the serialization error correctly.

In some cases, we panic because we assume serialization errors are not possible. But we only do this when we serialize to a vector or memory. (If we're using the disk or network, we can encounter an IO error like a full disk or a broken connection.)

zebra-chain/src/block/serialize.rs
arya2: > Can you add documentation to explain when Zebra could encounter invalid versions when serializing a block header?
> 
> I think these errors could only happen if we constructed a generated block in memory with the wrong version. All other blocks are deserialized when we receive them, and never modified. So the deserialisation would pick up any errors.
> 
> And in the getblocktemplate RPC, we generate a template with version 4, then the miner generates the actual block, and then we deserialize it and do this check.

Yes, added in https://github.com/ZcashFoundation/zebra/pull/6475/commits/7d7594f86b9ae0dab79a2e41792b7f41c89a6a33.

arya2: Added testing for block header (de)serialization errors in https://github.com/ZcashFoundation/zebra/pull/6475/commits/6b4ec1b79804e0bd9c7b3f61f3c060aa44a49b8c.

arya2:
> Can be fixed by changed the version to `5` at https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/checkpoint/tests.rs#L507

Thank you!! Applied in https://github.com/ZcashFoundation/zebra/pull/6475/commits/86ec19b4a39bcf2001818e098c9e2da6a82b5588.

> In some cases, we panic because we assume serialization errors are not possible. But we only do this when we serialize to a vector or memory. (If we're using the disk or network, we can encounter an IO error like a full disk or a broken connection.)

It looks like `Header`s are only created in production via `ZcashDeserialize` and never modified, so the `check_version()` call in `ZcashSerialize` should only ever return an error in tests?

teor2345:
> > In some cases, we panic because we assume serialization errors are not possible. But we only do this when we serialize to a vector or memory. (If we're using the disk or network, we can encounter an IO error like a full disk or a broken connection.)
> 
> It looks like `Header`s are only created in production via `ZcashDeserialize` and never modified, so the `check_version()` call in `ZcashSerialize` should only ever return an error in tests?

That makes sense to me!

teor2345 (approved):
Thanks for the extra docs and tests!

PR #6488, opened 2023-04-12
doc(net): Explain how we prioritise peer messages, and why it is secure

### LINKED ISSUE #6344: Resolve open Peer Starving `TODO`
## Details

The following open `TODO` item from [src/peer/connection.rs](https://github.com/ZcashFoundation/zebra/blob/a4cb835e129b9203cc84ce8cd3495689a04f37a2/zebra-network/src/peer/connection.rs) is noted:

https://github.com/ZcashFoundation/zebra/blob/a4cb835e129b9203cc84ce8cd3495689a04f37a2/zebra-network/src/peer/connection.rs#L598-L613

This would seem to be an important item to prioritize for resolution. While NCC Group concurs that it is unlikely in practice, it nevertheless could compound with other factors to strengthen network-level attacks. It is further noted that in the case of attackers with  extremely influential positions on the network, mitigating factors such as network latency
may have less of a mitigating influence than originally expected. 

It is noted that if the great majority of outstanding messages originate from attackers, then randomly chosen messages will likely be attacker-sent, and will still be prioritized over honest ones; it may be preferable to respond to requests in the order they are received, in order to ensure all requests are (eventually) answered. This may result in worse behavior in
the case where the incoming message rate exceeds the message processing rate and a backlog develops (since in this case response times would be monotonically increasing); however, the solution to this is likely not random choice but rather rate-limiting.

## Resolution

- [ ] Update existing comment to specify that the current existing behaviour is what is desired

teor2345:
## Motivation

The auditors asked about a TODO about peer starvation.

Closes #6344

This PR documents concurrent code. If it's unclear, please ask questions or suggest changes in the PR.

## Solution

- Explain why we want to prefer inbound messages when waiting for a request
- Explain why we prefer cancellations and timeouts when waiting for a response

## Review

This is a routine audit fix.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -603,7 +603,7 @@ where
                     // - incoming messages from the remote peer, then
                     // - outgoing messages to the remote peer.
                     //
-                    // This improves the performance of peer responses to Zebra requests,and new
+                    // This improves the performance of peer responses to Zebra requests, and new
                     // peer requests to Zebra's inbound service.
                     //
                     // `futures::StreamExt::next()` is cancel-safe:
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -595,22 +595,45 @@ where
             match self.state {
                 State::AwaitingRequest => {
                     trace!("awaiting client request or peer message");
-                    // CORRECTNESS
+                    // # Correctness
                     //
-                    // Currently, select prefers the first future if multiple
-                    // futures are ready.
+                    // Currently, select prefers the first future if multiple futures are ready.
+                    // We use this behaviour to prioritise messages on each individual peer
+                    // connection in this order:
+                    // - incoming messages from the remote peer, then
+                    // - outgoing messages to the remote peer.
                     //
-                    // The peer can starve client requests if it sends an
-                    // uninterrupted series of messages. But this is unlikely in
-                    // practice, due to network delays.
+                    // This improves the performance of peer responses to Zebra requests,and new
+                    // peer requests to Zebra's inbound service.
                     //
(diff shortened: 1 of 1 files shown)


zebra-network/src/peer/connection.rs
```diff
@@ -595,22 +595,45 @@ where
...
+                    // Currently, select prefers the first future if multiple futures are ready.
+                    // We use this behaviour to prioritise messages on each individual peer
+                    // connection in this order:
+                    // - incoming messages from the remote peer, then
+                    // - outgoing messages to the remote peer.
                     //
-                    // The peer can starve client requests if it sends an
-                    // uninterrupted series of messages. But this is unlikely in
-                    // practice, due to network delays.
+                    // This improves the performance of peer responses to Zebra requests,and new
```
teor2345:
```suggestion
                    // This improves the performance of peer responses to Zebra requests, and new
```

zebra-network/src/peer/connection.rs
```diff
@@ -595,22 +595,45 @@ where
...
-                    // practice, due to network delays.
+                    // This improves the performance of peer responses to Zebra requests, and new
+                    // peer requests to Zebra's inbound service.
                     //
-                    // If both futures are ready, there's no particular reason
-                    // to prefer one over the other.
+                    // `futures::StreamExt::next()` is cancel-safe:
+                    // <https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety>
+                    // This means that messages from the future that isn't selected will be selected
+                    // the next time that future is ready.
```
teor2345:
```suggestion
                    // This means that messages from the future that isn't selected stay in the stream,
                    // and they will be returned next time the future is checked.
```

dconnolly (approved):
🎉

Issue #2336, opened 2021-06-17
Add hashAuthDataRoot to Block network messages, and semantically verify it


teor2345:
## Motivation

In #2134, Zebra validates the `hashAuthDataRoot` in the finalized state. But this is very late in the validation pipeline.

We'd prefer to validate it as the first step in block and checkpoint verification. That way, the transaction authorizing data is bound to the block header before we do any other verification. That improves the security of Zebra's verification pipeline.

### Specifications

Transaction Identifier Non-Malleability: https://zips.z.cash/zip-0244

### Designs

Zebra's parallel verification pipeline: https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/rfcs/0002-parallel-verification.md#definitions

## Solution

- [ ] Write a ZIP that adds the commitments used to generate `hashBlockCommitments` to the end of `Block` messages in the network protocol
- [ ] Add these commitments to Zebra's `block::Header`, parsing them on the way in, and sending them on the way out
- [ ] Validate the `hashAuthDataRoot` as the first step in block and checkpoint verification

### Alternatives

This change improves security, but it's not required for full validation. We'll probably do it during the network protocol upgrades after NU5.

## Related Work

#1874 and other ZIP-244 work.

daira:
I assigned ZIP number ~~240~~ 260 for this. (The title is just a placeholder.)

teor2345:
This protocol change would be useful, but other things are higher priority.

Issue #6281, opened 2023-03-09
Tracking: TODOs with closed tasks


mpguerra:
## Motivation

We want to track all of the findings from the zebra audit.

## Details

There are a number of `TODO`s in the codebase that use task numbers that are already closed. For instance, issue #862 states that the Sync process’s `state_contains()` API only checks the best chain for a given block hash ([zebrad/src/components/sync.rs](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebrad/src/components/sync.rs#L961)), rather than checking all the available chains in the mempool. The audited version of the library still queries the best chain in state and the task is closed.

Another example is issue #2214 which suggests that the fanout handling should be done by the `PeerSet` so that all the fanouts wouldn’t use the same peer and block themselves. This seems to only be an issue on testnet where nodes aren’t as well connected as mainnet and the task is closed. This issue is referenced in a `TODO` in [zebrad/src/components/
sync.rs](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebrad/src/components/sync.rs#L547).

This issue will be used to track any further issues created to address `TODO`s which refer to issues which have already been closed.

## Scope

- [x] #6278
- [x] #862
- [ ] Evaluate list of `TODO`s with closed issues and identify:
    - which `TODO`s can be removed because the issue is already closed as done, otherwise resolved, or no longer relevant
    - which, if any, issues need to be reopened because they are still relevant and were closed because they were no longer a priority
    - which `TODO`s need to be updated with relevant context
    - which `TODO`s should be updated to `WONTFIX` in the case the issue was closed as not planned or something that we will not fix.
- [ ] Update `TODO` comment referring to #2214

mpguerra:
Copied from #6347 

```console
➜ GITHUB_TOKEN=XXX cargo run --package zebra-utils --bin search-issue-refs --all-features --features search-issue-refs

Searching files in this repo with a ".rs", ".yml", ".yaml", or ".toml" file extension for issue references..

Found 209 possible references to 104 issues, checking statuses on Github..

{results - see below}

Confirmed 179 references to 91 closed issues.
```

### References to Closed Issues

--------------------------------------
- #65

https://github.com/ZcashFoundation/zebra/blob/5db2243c25cdde03d9cdccd72a137ccc6f2370a5/zebra-network/src/protocol/internal/response_status.rs#L52

[...]

PR #6510, opened 2023-04-13
fix(rpc): Omit transactions with transparent coinbase spends that are immature at the next block height from block templates

### LINKED ISSUE #6482: Fix Immature transparent coinbase spends in `getblocktemplate` RPC responses
## Motivation

If we send invalid templates to miners, they will make invalid blocks. This wastes mining power and loses the ZEC they could have mined.

These invalid templates are happening in PRs and on the main branch.

#### Immature transparent coinbase spend

This might be a missing mempool contextual validation query to the state.

Click for details...

<details>

```
2023-02-20T22:21:48.939092Z  INFO calling getblocktemplate RPC method at 127.0.0.1:40585, with a mempool that is likely empty...
2023-02-20T22:21:48.939771Z  INFO got getblocktemplate response, might not have transactions response_text="{\"jsonrpc\":\"2.0\",\"result\":{\"capabilities\":[\"proposal\"],\"version\":4,\"previousblockhash\":\"0000000000621ae12cf1f90e8c1f6821d94529f511fd8b3df595dd596d835d5a\",\"blockcommitmentshash\":\"ccc674e0e3b7a72613d0a5cecf5fa50dbd68e126217bcdf5b5ab08314c1c4b43\",\"lightclientroothash\":\"ccc674e0e3b7a72613d0a5cecf5fa50dbd68e126217bcdf5b5ab08314c1c4b43\",\"finalsaplingroothash\":\"ccc674e0e3b7a72613d0a5cecf5fa50dbd68e126217bcdf5b5ab08314c1c4b43\",\"defaultroots\":{\"merkleroot\":\"25b3c88b456b8ffdce9f8fa0730eff194f2c19d695010e84ed53ad266f040439\",\"chainhistoryroot\":\"c65e1ee71afddd3729cffec49b190b02541cbc3819ab7e6562a6e02dcb3d17d8\",\"authdataroot\":\"ad52724e37f52c500d815c47e0abd56a5caa0e0910a8feb4dfc8ed9deba4040e\",\"blockcommitmentshash\":\"ccc674e0e3b7a72613d0a5cecf5fa50dbd68e126217bcdf5b5ab08314c1c4b43\"},\"transactions\":[{\"data\":\"0400008085202f8906f7b4faf96eaf5bb356458ca72a1efe0810a0c7add08283ea23b76dd1bdf91adc190000006b483045022100fca6418f7991ca54b92c5b88ac8ea8b4ac187eee6ae02396a213f991c3d515a702203960c8c1fdc7fe5581657a1a1365139084130f8dce2f6071fcb640de922c5d4a012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff1601cd5fb49903e44cbc17613295e77a20a3b6e89ea8a15e99d1c08fc9839233160000006a47304402201bb95aee6860765a81c51d5ea3a1c9930ed568b413ef73dfa04301f868c10d1002207b88e3e0d8162fb51a6dfb06b9c08d16b1c538703c57332eb812f78db140a43e012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff888f86fb84652235a0da5ce4669461cb1fbd50247c1ead59fb545431d77b4029000000006a473044022023b55776a48595cfec49255dacf520cffa2872788dd1baa1b6293c176261ae190220272911fbf65ec94a505a022bb91d4ac6064f9264defbe577b934e577b5118a1f012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff65e923890ce78ffff639da9a4b4a0f52117e8ca383e9b361c33ee5cd89a91d12000000006a47304402207a82d8b2dedfdabfb9cd0c115aa3a32c99da39315a87fc9d40c7230bdcb7a08102201774f63dd8a57e0184fc2b0ad2542cca89cbdc7e8915b68203a91005d27eafa6012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff3b4c20ee11662439215d8f97c4f6eb5a0868cc8a13a485e97c60bf2b0632da43000000006a47304402205d78418bc7b0b6b6cc45a1d0f23236b1d1be66beabd9748de0d3de8de585640e0220528350193c55a9f983ecf38507aaf9fe06d7820b12794aa9fff560893049cacc012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff1f3f2044c6d871762b440f8b58b3117c669b788323eda70c67f18e0204a027a0150000006a47304402207810cc82a15fcc292c0f22bcfe762f6cf439fc15199826a00e74b928389791aa02205b987d38f9fea62cedbff0ba4f20fb623cc1ea0c0d845ab38ba905284b11f3b8012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff0102190c00000000001976a91494c1e92e3544ec2d9124e77298ec9ee57adbdffc88ac00000000000000000000000000000000000000\",\"hash\":\"5c0da8ae26573fd170049489e031c17bb982af09d4cc0df1fb990af42f432436\",\"authdigest\":\"ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff\",\"depends\":[],\"fee\":9320,\"sigops\":1,\"required\":false}],\"coinbasetxn\":{\"data\":\"0400008085202f89010000000000000000000000000000000000000000000000000000000000000000ffffffff0503b5611e00ffffffff04e8d6e60e0000000017a9147e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e8740787d010000000017a914931fec54c1fea86e574462cc32013f5400b891298738c94d010000000017a914c2d76726a137b2228cbf2d4229061c42156054b487286bee000000000017a914d45cb1adffb5215a42720532a076f02c7c778c908700000000b5611e000000000000000000000000\",\"hash\":\"ef46e300dbd50e250f101c13af0d477d84fa4e95ef1abf2f22c1454fd266a14d\",\"authdigest\":\"ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff\",\"depends\":[],\"fee\":-9320,\"sigops\":0,\"required\":true},\"longpollid\":\"0001991092f022a90516769365320000000002c5c2eeef\",\"target\":\"0000000001543200000000000000000000000000000000000000000000000000\",\"mintime\":1676931133,\"mutable\":[\"time\",\"transactions\",\"prevblock\"],\"noncerange\":\"00000000ffffffff\",\"sigoplimit\":20000,\"sizelimit\":2000000,\"curtime\":1676931708,\"bits\":\"1c015432\",\"height\":1991093,\"maxtime\":1676936532},\"id\":123}\n"
2023-02-20T22:22:03.228297Z  INFO failed to validate template as a block proposal reject_reason="block-source-validateproposal-commitblockerror-immaturetransparentcoinbasespend-outpoint-outpoint-hash-transaction-hash-c7cfb238596f9509dec27c4a1826f789fc4ddd9b5410d278085ff91becc8d1ed-index-0-spend-height-height-1991093-min-spend-height-height-1991094-created-height-height-1990994" template=GetBlockTemplate { capabilities: ["proposal"], version: 4, previous_block_hash: GetBlockHash(block::Hash("0000000000621ae12cf1f90e8c1f6821d94529f511fd8b3df595dd596d835d5a")), block_commitments_hash: ChainHistoryBlockTxAuthCommitmentHash("095321b49d42c6ff7fcd0138698344e90d1700998c94f2f39239cfc83bca8780"), light_client_root_hash: ChainHistoryBlockTxAuthCommitmentHash("095321b49d42c6ff7fcd0138698344e90d1700998c94f2f39239cfc83bca8780"), final_sapling_root_hash: ChainHistoryBlockTxAuthCommitmentHash("095321b49d42c6ff7fcd0138698344e90d1700998c94f2f39239cfc83bca8780"), default_roots: DefaultRoots { merkle_root: Root("b8931c5a09b6080c4c97ad86b7c4fcdd23d1258d9119162c7ed67715c3fe521e"), chain_history_root: ChainHistoryMmrRootHash("c65e1ee71afddd3729cffec49b190b02541cbc3819ab7e6562a6e02dcb3d17d8"), auth_data_root: AuthRoot("ff4fc133b81ec603eaa391a31c2035812f5ebb55e1b28bd186a8b28ab615acd4"), block_commitments_hash: ChainHistoryBlockTxAuthCommitmentHash("095321b49d42c6ff7fcd0138698344e90d1700998c94f2f39239cfc83bca8780") }, transactions: [TransactionTemplate { data: SerializedTransaction("0400008085202f8906f7b4faf96eaf5bb356458ca72a1efe0810a0c7add08283ea23b76dd1bdf91adc190000006b483045022100fca6418f7991ca54b92c5b88ac8ea8b4ac187eee6ae02396a213f991c3d515a702203960c8c1fdc7fe5581657a1a1365139084130f8dce2f6071fcb640de922c5d4a012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff1601cd5fb49903e44cbc17613295e77a20a3b6e89ea8a15e99d1c08fc9839233160000006a47304402201bb95aee6860765a81c51d5ea3a1c9930ed568b413ef73dfa04301f868c10d1002207b88e3e0d8162fb51a6dfb06b9c08d16b1c538703c57332eb812f78db140a43e012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff888f86fb84652235a0da5ce4669461cb1fbd50247c1ead59fb545431d77b4029000000006a473044022023b55776a48595cfec49255dacf520cffa2872788dd1baa1b6293c176261ae190220272911fbf65ec94a505a022bb91d4ac6064f9264defbe577b934e577b5118a1f012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff65e923890ce78ffff639da9a4b4a0f52117e8ca383e9b361c33ee5cd89a91d12000000006a47304402207a82d8b2dedfdabfb9cd0c115aa3a32c99da39315a87fc9d40c7230bdcb7a08102201774f63dd8a57e0184fc2b0ad2542cca89cbdc7e8915b68203a91005d27eafa6012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff3b4c20ee11662439215d8f97c4f6eb5a0868cc8a13a485e97c60bf2b0632da43000000006a47304402205d78418bc7b0b6b6cc45a1d0f23236b1d1be66beabd9748de0d3de8de585640e0220528350193c55a9f983ecf38507aaf9fe06d7820b12794aa9fff560893049cacc012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff1f3f2044c6d871762b440f8b58b3117c669b788323eda70c67f18e0204a027a0150000006a47304402207810cc82a15fcc292c0f22bcfe762f6cf439fc15199826a00e74b928389791aa02205b987d38f9fea62cedbff0ba4f20fb623cc1ea0c0d845ab38ba905284b11f3b8012103df9038a4cd7d003e8193c2a38c3f3873dfd2c1cec6bd967f03b8c838caf9c8e5feffffff0102190c00000000001976a91494c1e92e3544ec2d9124e77298ec9ee57adbdffc88ac00000000000000000000000000000000000000"), hash: transaction::Hash("5c0da8ae26573fd170049489e031c17bb982af09d4cc0df1fb990af42f432436"), auth_digest: AuthDigest("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"), depends: [], fee: Amount<zebra_chain::amount::NonNegative>(9320), sigops: 1, required: false }, TransactionTemplate { data: SerializedTransaction("0400008085202f892607ed2a826986d53cc25346d8a83c524d53a18eecb50d2a169ffe3c0a868842de000000006b483045022100faf3132bc08dca256de1014f96c4760a6cfb5a45123ee9157dedebec0dc6193c02205fcab0646c3acfe6f1d6d39a2370395b9d2b5780f59faacfcdc69697f3bde23a0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff0d9689aa00dd8e0541ee8e92c258406761ea6ab6cf4bff61e22d4844137b839e000000006a47304402205338ea8d55e3066752adbeeb945b627011ad2ead43ba5230a3865422213dbb9a02202af994227d5e42df17dd06fd03ee36fe66ab499062a72b505975675b8c7e592c0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff16e5c23b753f68c2d58c386408cf49a797f60da3110af4a368d331f0f56e79c1000000006b483045022100f24825dc8cd1cb73897a50f891c8ba0d634f3dcbe2e52ab26f06adea2e0648ac02202b7fbe732d2e2c7ad1b737955760d2f64bb8495b1f55e1c89e00079e2d8cf0a30121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff375fdf384491e54ed8a50b1f4972255b50a4b15a5e0ef70dc11478a1de49433a000000006a47304402205adacf9cd35b077f339c2e31f2ef5f97dde9d7519f70d3240ee66a04b7c4ed99022020d09acebb114ef2094c64958547132d6c61ff4cf1f8cfca14a572bfafdda9090121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff3b595e82b94c3104994b4458e79f45a8862f03a3f939498eb40401cb20d19fd2000000006a473044022051b5cc5223357b0346704bb0b4ddbfdb2d70aa970c733c85a1de217296bf9d7e02200f8fd3aff7304b6cddef8b42b82257c73ef300148c80fbecc47f5c4bd1d7f3b10121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff45b3186017523e13309719a692aad6897cc1eb380afeabd2552edaf44bc6e674000000006a473044022025f8489f6b60d13a4535d68ddd7709189dcc92fbad78d1c9b4f8375ebd961c91022006731571664ec79d7e75b4102c5a4abc28f6a525c4d0464da065120783bfde1a0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff4af14eb6325dff8cd59c830b9a2bbaae7e9047bd665b942b1180abc6d11ec532000000006a4730440220784e33603780a04dafb3d23fb0d20e81b9d51eb33c3806026ad67d4cb126bda902205a3ff5487c63ae8e54eeb19ddb4d44999fd0cbc516a9ae5ad1ed8bb08faf9b890121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff5226e212b38dd57557332ea1f46273a208185efc580a684550c6ecb5e9acab35000000006a4730440220646ef83a203fd03909d879495cb3348be54fa92af0356dad3b06af5ac2c77b5802205d30c0626db3f727df76c72ecde85666e7ff7d76894a15c2f1c1b637ca6f82c30121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff658aaf62ab91b51858c216fb42b3337e48fa37dd7222053ba4f229fab0344efb000000006a473044022068ea4cc66ad85752ecefc98214c88a19e53f8e4d6492f763d6ff692340dc2aca02206dea617524119e2e66515fec63313f939487ea4d0ffb54b6aeec3452c72b31b70121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff683d73118c26a5dbeb08e25e17420b16c87020b80ef741a9cdd704663d501fec000000006b4830450221008280ca3424548f33b0d78973d93e14e5d39ebf7094451d6c05be17682b37387a02206544dea4c13c03183e2bf4ea743daa07167a1874d796f1e17d32f7742ad5d88d0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff759bb0b3a2a17e2366b2a6765628dd30890641b7dfb4fc1a7c0ab102cb83a363000000006b483045022100de1817e0d3a0465cab4f7c893c0c5209c5aff650c69fd3317568f891cfe86830022029c4b9e3094cdb3fc6845330562fed65c4d726bebbad5edf4d9ce858ba1f38460121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff78a89b07f08e97e6c64ebd29b111ab341a8636162692cc550588c1b1a6bd44e0000000006b483045022100d9eade0d9fa2ebd831fab3254ba7fe5b645b7f58322a0bb1b52b399f902bba9802205843c25fb4f5121d0827614f2358f89a5ea5c2e75e6c7382a118d9bc5a0bcc9e0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff7b7cc8f50bdaf598ec7903a8c7bb9de01d2d3032169b4dfbd36231d969d340a2000000006a473044022008a494deda8770a806f7ebb64a9cf49b907d7ed0ddd4de01dd0dc660bd3b2caf0220159451a8bd65304f6ca401859b52eb66b53b4579f7cdd505592e16c2d380f28c0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff7e058e6ec9664b405027d0708f50f59100f29a85326b0d6a1b9f853505735fc0000000006a4730440220459876f074e5705829556b60dc4765ab4e57a41d2a8eb4fc7d68a7b35c1456a7022061044ecca78c28020f3c0cafb72c5553a945a2baef0decffcccae4681c9544030121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff896a15286922cde36bf15fb47bee55fa741a1a72170a34e2812cdbfdc85c6b24000000006b4830450221008b9600ec0b3c192ec9f74650e121c1a4bd3b1bbe811abd04119b7b755aed30b302204f791146816cf8e620ad076a11b5f31b747d3639ae974499382107d67114c9f70121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff8a3850a631e2eebcc102bc819087784ebefbd392bf41d14529c8684841de34fd000000006a4730440220060371ab7f43ac3c6694e3c402d421dab6a79c58db85b0041b565d7826b5da24022021cd4d093215f3b7dcc221eaddff29bb905bb9cddbe5b93ac7ea00964a70bdb60121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff8ab2138d32fdb520fd8a7032848c6c4b3fd8776f4927ee9e408c415159b698ac000000006a473044022065c6a5fc818ff8f407a0a5fdbfddf180eb91bcf6c2eabb7c3a0ba3068abb054602203477e0060f201e242fcfb5a81d11103e1be7e0f2bd93135391570da3a7881a1d0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff8db13ffb130688ae6760d98f013ffe7723b305824b579a2048fc3f30382a0463000000006a4730440220029f4b0a986418be30649966b87de98a31e9f29b247d6ff3ac46cee2ea058b950220075348e6c4609af1ecc6101f91436deb45afe928390122e5d9538ec6d7e6962e0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff9012814fa451868a93f6b929c9d414d212df23c22673765ff1d775664265fab8000000006b483045022100fdde2e254e1d0b668913795b414c7fb238de036c59f301d56ee8421a3978d113022036af8290945fb71ed673b151668f3e17b8dcff682038ebc7288c51d2261c6e560121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff904228e523e79a5b96c6f01c4fa71eb1e1d32c32482df97121b7bed543cd406c000000006b483045022100d1b79946254f5b911e2ad32604cff4ce9bfb931ae9b17d01f6d705437aad677b02202bfce90d5466370152d4cddc59a6cb21bb1d22139de36f5f31806589604da7620121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff90eb312f8db224bc8d9afd4db18b1b7f963d0629efc1d362161725d5f21f6c92000000006a4730440220256e859516bcf8e123aadd133daf395daac5a4867db60cf93ab03f3e7ade426d02207e693961fcb31cd640ee9d5f4705c474da860aae63894c434f529a468692da9e0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff95f35c812835df9d9b7d302852016af98eccc8061ffa7edfccd9056e4fe98a20000000006b4830450221008df8cd0829ab4758c34a0ca8155a542deb4e7c5107c66f2eb916e782dcd07ad1022070747ad612f2184e0d4e768a7a6b70a7bdd322899fc0121dd728dedd6201ac7a0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff9a18b65a0e06b8e74df84fc74cf910619cf48addbd37b78db04838a78b64186f000000006a473044022074f64a65b3a5de29c83261caa4b82e2c959e9ca68118fa6b106c23fd240a1c3a02202c665362b7557e4997ff34bed7d9f1c46bbe2d9f7a1be2e0469d64438fba2f670121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff9cbd8b3b94fba871cd86cbd5713f4678ec8d44a821c59a77f9911d74460a8139000000006a473044022029304b35bd82e6b968a85558cc276f5d3a0a2ee6cd025437b29e7069e48110da02201bcebbd8b4985c799d6a478e6dbe63a8f4ef071b42ae48136cba74d902ef25560121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffa6664c01ac8cecabb27fa62c66b81b1cb4bd03b2f4ada139d3af332d9deaa2c3000000006b48304502210081a3d02bb06a942f6e8bc400ba5ebd68ac9581a93aa5ead8965b719a1c1fc78c022034f273bf2821d9ebfd680dfcea9fb5bb3ba5f9e6d277d24dccd3e0abd99ffd060121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffa75498cd4e209048069f4f029981c87a1a3c06cbb7455156944e665d0402924c000000006a47304402201a2732cd427597d9658d3e314cdd4942ae9780771771bd1ed55e61902fef7355022042f816bfc135483d37de308e43725faf387902d829ddeff6265c0402f9e824b80121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffadd79280dcbbffb6fedb9796c86ffaa1299ecc207c5b666c6b4210f03abe8b06000000006a47304402202d3207fa9f93141a20513c1c18800a40928faa1888b986b0bbb7b17aea7391a6022014bc333ad010405852cc225f8edb5b017f19b7f62fb85fcc3bb089e94305de5c0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffaed7d93fad8fc35af90e7767391ee122d158b514373bae4629b72aaf32c83126000000006a4730440220266f26b6ff640a693ee2ced544451a971744d7f6504cc8ecde8d7be7677a7af00220495d789ab0c0ff7eec2866c42bc4a70fb8e88ee13b817618f4751ddacf34eca40121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffcaac51f2349e210dd799dd20fd0ed6a34dfd713d8ac649e45d536ab9bed3e52e000000006a47304402206dc9fb79040b12bf6234e2dde42f3c758152fd60132f6785a0f7f007ee38185402206a8354c36413705966f1e0211f8e539815a7aceb7a362f926d11ebf5b96c5c470121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffdbfea10c00e23257fe12d201fc52fe79ffee4494bd871724e2dd3775c186b59b000000006a47304402202b8253b6fea0c27088a517af6b8ab0edfe68d0ba4af647cf4a997a71cd667edb022026f56c8bf4548d2ae014bed0556cc959870086d6f70c505943bf98f12e7212fa0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffdd6d76ed75d0b2666a2a4975201a7c531ce4c415a14604d36d4d0c5b20c7e3b8000000006a473044022032e01aa9356dcbeb38298a92495abcaf94699f0307cf2fbad9127a6690c3e8f7022010280cdf54fcb822700ad78bae58f9b328ab71eec0a85ba55b7989f3caf018700121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffde9b719342b13aa5507dd572ed63b84ccdd73aa6d603c553babc45e48e10865b000000006a473044022050cc84e813e7e17d689049748c4e2c3bae1439d4c7c94be812097b34b958199002206e16232963055f1f0f19819b800ddc56604e38e19fe04e10471440428b2becd20121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffdf47d505be18583cdcb43b56165ad4ce9f27357a8318b216a3455e436a92b51e000000006b483045022100b9079aa733738a55ce7f1c368180987a869706e2b67a1cc2b57a204884e3766002206ec0f9eaa91f9c21ec5d76570050ceb32aeed8d9afd3503de782e303238302820121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffe1ca75795e6154afdd7df43cb1c9e3a62997e10d26626f80699bed904c49e510000000006b483045022100ce435b8210838a7840b4f6e26a6a90690cc6d659eb884eca562b01529f9ea5da02204d25f3a2541c828a9d5632297acb258609d43e00de85a167606c7639740c246e0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffffedd1c8ec1bf95f0878d210549bdd4dfc89f726184a7cc2de09956f5938b2cfc7000000006b483045022100e09036270845696124b0c49aa10f7a99ba8e4555fb773499dab205fcc7c074aa02203d5978157578b5f5dfd4a1b86ce0cf64355df36f0183d0bd0b93e3101f0d197b0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005fffffffff37c8b502095b1d3dd22dfa62659cdb39ec12f1ca8291fff19c8089b580e313a000000006b48304502210088498eff709b18e4f7fed668d5f3a68d15134f5e17fd392000bed8f5bd92a2d402200a84afe27a6eaabddfd4215e957b306556e2e5d841ffb35575ff15bd5cd4df7d0121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005fffffffff6b8513460799ff1f154225e95a1b4dd6d828983363ca4f8c60bdaf92fd2740f000000006a473044022027df1c98a10e78739b27493113219c0d1957f03d0f8975ee7c89d56510956881022010f887f71339a32b7c9f788410baf5bbc735e4dd174835f86f3869df79bce5300121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005fffffffffa28134f39776dada8517678ffd83cc9d216cffa2d2c01673c1dae4fd81951c1000000006a47304402207adb86a40da02d72c0de40d8d24413a2c369bf8b4f2d229087f4aeff2888ea3402207e8a60a360639b4c0fbf2d3fc9a90f978271a7ea0e9aade193a51f0ff0d249290121037da8d6e0518d0968e70fce58c0d55313f634b5a02b4998d53288d70989665005ffffffff0000000000e0611e0012eabac9fdffffff00013afc684e277eb287d9a5907b71e59e3db6bc65b6ffa2cdcb15ceb3909ca59cb91399f506c7f4e7ac3ba0628083bacef769951e8726c69fe84a6b5da535546e5283fbc741fbe2fc96688a430dd6c489d515692e7a40743dd7f538d1f97a2e922d017a593a926e616f224cb7c6136719f00d0337368f6dd470e109902ff8360214aa0cb483e242365654d8a8cd4b7beece3a39f4ff2e4bcfc4d9361225215107ca1170e30b2b4b5e021c08fd2fcf770284f5c5713cac5eb692a4146e30dfc08882d27a5eb8c5c8392d0064c8bfd8c0e46626810d7fb1580ffb051b749862ba0cf37e13c1e113510daee32a397b6c692f4337b3ef3d4419f20de516db5b137d504d19baabe3cebde3eed9aa554ec9da2d4793cd297f042705ab091305b472223a9ee9ca04b6c95581fb0d9379a08bb33a5fb0ecc76a961a1701282a7fffec5fc3a5922813496ca2cc07543500df8e135cf9d4ee2fedca9602f2b545c0dfc7266118f72fc6adc437cab9513107fa01c210a251df851ef391f1fdf0da02b6f17e46336e3433d171bd72831455a593ec1dddbcc1c9b8d7d3a1bde5b8a71d1660022c8e7173f2046ceb4068b0c27d7aca5a9026bbc5fe9bb2c9319bf92c3904e5bfd5f42a1399c0cbb8b8043490fa07aa20f33b9c7b7d611cebaad7112bc3a1a66d95130145fc2639df1a39d08ce8d162b804e7ed663cd22c13adda41024e6865c0f461ef168e4bd14a49ba8ea27ce199ef62af755d04dfffcee7a2f07a94bb154ef1817606c9ae0715b44ca937d5ccb570aa5d64f2c950fd47630337d29db96f386102c1c22023fc2965c46bd41b8f3cc2aa70bfa87e365c2fdc0da8d90e42105595ce2f1874eca43514a1068250e0d6724484b895eb0bf714ed9dee5cf77e42008894be1d46f875aed6fff79f6377d7b982da7202b75068a16ffc7fc40d21e09f419076b66c09bfe86b1b633fbc4dfff4905b0820a6c37e9ded32bf4a681e4c694fb03a43acea8a9a42020b61f226c4bfcf1fb220f2974db1d374ef758c8a28841d8aaaab518a170366df1c19bc155f6e6fbf73e68ab690a0e4c9b047227bbe089c33bae4d5dc4ba3d6ef5dcbf16f3d18e455cfe4f9fd64cc7e5455d38139528ac584d6ce4b4080e527e1be09ba0617697d07aef0dabd283168552b554af526ad301355964d15074d2ca3ef6911a2226ed9c264a73040071c5752c9dd6943ad1573e33e8bba7214c9cc9d6bac874454b60b004b94b7d99b06fff241aad6e568fdea90ef42a5789446a96c946b1d211b9e3d5a5650fbbf10e0c067856ee297fee54157e06b8ae0dfeb18da056da5ff03a894e280fd763a0088b66fd240c5f05a04fdba14d2dd95931cab53439d561533f86b1a9bb7b9fc48e8355dd453911784d8bbff5d87c9f314b0a3ef481c210c42ec3be2943418fe03"), hash: transaction::Hash("a5a240e1a6c7f6537e25dd1391763ec454d0aaf7c289fd355a7203ab067a6379"), auth_digest: AuthDigest("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"), depends: [], fee: Amount<zebra_chain::amount::NonNegative>(1000), sigops: 0, required: false }], coinbase_txn: TransactionTemplate { data: SerializedTransaction("0400008085202f89010000000000000000000000000000000000000000000000000000000000000000ffffffff0503b5611e00ffffffff04d0dae60e0000000017a9147e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e7e8740787d010000000017a914931fec54c1fea86e574462cc32013f5400b891298738c94d010000000017a914c2d76726a137b2228cbf2d4229061c42156054b487286bee000000000017a914d45cb1adffb5215a42720532a076f02c7c778c908700000000b5611e000000000000000000000000"), hash: transaction::Hash("1fa2e68e7329fee2a6906a9748d50ce2eeedad3597acb10f7da940b6ffcabb12"), auth_digest: AuthDigest("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"), depends: [], fee: Amount<zebra_chain::amount::NegativeOrZero>(-10320), sigops: 0, required: true }, long_poll_id: LongPollId { tip_height: 1991092, tip_hash_checksum: 4028803333, max_timestamp: 1676936532, mempool_transaction_count: 3, mempool_transaction_content_checksum: 3819277466 }, target: ExpandedDifficulty("0000000001543200000000000000000000000000000000000000000000000000"), min_time: DateTime32 { timestamp: 1676931133, calendar: 2023-02-20T22:12:13Z }, mutable: ["time", "transactions", "prevblock"], nonce_range: "00000000ffffffff", sigop_limit: 20000, size_limit: 2000000, cur_time: DateTime32 { timestamp: 1676931722, calendar: 2023-02-20T22:22:02Z }, bits: CompactDifficulty(0x1c015432, Some(ExpandedDifficulty("0000000001543200000000000000000000000000000000000000000000000000"))), height: 1991093, max_time: DateTime32 { timestamp: 1676936532, calendar: 2023-02-20T23:42:12Z }, submit_old: None } time_source=CurTime

Message:  block proposal validation failed: 
   0: unsuccessful block proposal validation, reason: "block-source-validateproposal-commitblockerror-immaturetransparentcoinbasespend-outpoint-outpoint-hash-transaction-hash-c7cfb238596f9509dec27c4a1826f789fc4ddd9b5410d278085ff91becc8d1ed-index-0-spend-height-height-1991093-min-spend-height-height-1991094-created-height-height-1990994"

Location:
   zebrad/tests/common/get_block_template_rpcs/get_block_template.rs:255
```

</details>

https://github.com/ZcashFoundation/zebra/actions/runs/4227058491/jobs/7341912371#step:8:203

### Specifications

This rule is already checked for blocks in the state: 
*transparent coinbase outputs can only be spent after 100 blocks.*

### Possible Design

- Move the check from the state to the transaction verifier

- Check latest transparent coinbase spends in transaction verifier
  - Return an error for block requests
  - Add a field `maturity_height: Option<Height>` to `VerifiedUnminedTx` for mempool requests

- Filter out transactions that will be immature at the height of the next block on the best chain from `FullTransactions` mempool response.

### Testing

Testing is going to be a bit tricky here, any ideas?

## Related Work

- #6195

arya2:
## Motivation

The getblocktemplate method can currently construct invalid block templates if there are mempool transactions with transparent coinbase spends that are immature at the next block height.

This PR adds a `maturity_height` field to the transaction verifier's mempool response and uses it in `get_block_template()` to filter out transactions that would be invalid at the next block height.

Closes #6482.

## Solution

- Adds `maturity_height: Option<Height>` to `VerifiedUnminedTx` in the transaction verifier
- Filters out transactions that would be invalid at the next block height in `fetch_mempool_transactions()`

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

Once authorizing data is bound to the block data during semantic validation, return errors for block requests to replace this check in the state service:
- #2336

### DIFF
--- a/zebra-chain/src/transparent/utxo.rs
+++ b/zebra-chain/src/transparent/utxo.rs
@@ -54,18 +54,6 @@ pub struct OrderedUtxo {
     pub tx_index_in_block: usize,
 }
 
-impl<'a> From<&'a OrderedUtxo> for &'a Utxo {
-    fn from(ordered_utxo: &'a OrderedUtxo) -> Self {
-        &ordered_utxo.utxo
-    }
-}
-
-impl From<OrderedUtxo> for Utxo {
-    fn from(ordered_utxo: OrderedUtxo) -> Self {
-        ordered_utxo.utxo
-    }
-}
-
 impl AsRef<Utxo> for OrderedUtxo {
     fn as_ref(&self) -> &Utxo {
         &self.utxo
--- a/zebra-chain/src/transparent/utxo.rs
+++ b/zebra-chain/src/transparent/utxo.rs
@@ -54,6 +54,24 @@ pub struct OrderedUtxo {
     pub tx_index_in_block: usize,
 }
 
+impl<'a> From<&'a OrderedUtxo> for &'a Utxo {
+    fn from(ordered_utxo: &'a OrderedUtxo) -> Self {
+        &ordered_utxo.utxo
+    }
+}
+
+impl From<OrderedUtxo> for Utxo {
+    fn from(ordered_utxo: OrderedUtxo) -> Self {
+        ordered_utxo.utxo
+    }
+}
+
+impl AsRef<Utxo> for OrderedUtxo {
+    fn as_ref(&self) -> &Utxo {
+        &self.utxo
+    }
+}
+
 impl Utxo {
     /// Create a new UTXO from its fields.
     pub fn new(output: transparent::Output, height: block::Height, from_coinbase: bool) -> Utxo {
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -226,20 +226,6 @@ impl Request {
     pub fn is_mempool(&self) -> bool {
         matches!(self, Request::Mempool { .. })
     }
-
-    /// Returns lowest height at which all transparent coinbase spends will be valid,
-    /// or None if this transaction has no transparent coinbase spends.
-    pub fn check_maturity_height(
-        &self,
-        spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
-    ) -> Result<(), TransactionError> {
-        check::tx_transparent_coinbase_spends_maturity(
-            self.transaction(),
-            self.height(),
-            self.known_utxos(),
-            spent_utxos,
-        )
-    }
 }
 
 impl Response {
@@ -388,6 +374,12 @@ where
                 Self::spent_utxos(tx.clone(), req.known_utxos(), req.is_mempool(), state.clone());
             let (spent_utxos, spent_outputs) = load_spent_utxos_fut.await?;
 
+            // WONTFIX: Return an error for Request::Block as well to replace this check in
+            //       the state once #2336 has been implemented?
+            if req.is_mempool() {
+                Self::check_maturity_height(&req, &spent_utxos)?;
+            }
+
             let cached_ffi_transaction =
                 Arc::new(CachedFfiTransaction::new(tx.clone(), spent_outputs));
 
@@ -467,12 +459,6 @@ where
             }
 
             let legacy_sigop_count = cached_ffi_transaction.legacy_sigop_count()?;
-
-            // TODO: Return an error for Request::Block as well to replace this check in
-            //       the state once #2336 has been implemented?
-            if req.is_mempool() {
-                req.check_maturity_height(&spent_utxos)?;
-            }
 
             let rsp = match req {
                 Request::Block { .. } => Response::Block {
@@ -585,6 +571,28 @@ where
             }
         }
         Ok((spent_utxos, spent_outputs))
+    }
+
+    /// Accepts `request`, a transaction verifier [`&Request`](Request),
+    /// and `spent_utxos`, a HashMap of UTXOs in the chain that are spent by this transaction.
+    ///
+    /// Gets the `transaction`, `height`, and `known_utxos` for the request and checks calls
+    /// [`check::tx_transparent_coinbase_spends_maturity`] to verify that every transparent
+    /// coinbase output spent by the transaction will have matured by `height`.
+    ///
+    /// Returns `Ok(())` if every transparent coinbase output spent by the transaction is
+    /// mature and valid for the request height, or a [`TransactionError`] if the transaction
+    /// spends transparent coinbase outputs that are immature and invalid for the request height.
+    pub fn check_maturity_height(
+        request: &Request,
+        spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
+    ) -> Result<(), TransactionError> {
+        check::tx_transparent_coinbase_spends_maturity(
+            request.transaction(),
+            request.height(),
+            request.known_utxos(),
+            spent_utxos,
+        )
     }
 
     /// Verify a V4 transaction.
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -229,16 +229,22 @@ impl Request {
 
     /// Returns lowest height at which all transparent coinbase spends will be valid,
     /// or None if this transaction has no transparent coinbase spends.
-    pub fn maturity_height(&self, spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>) -> Option<block::Height> {
+    pub fn maturity_height(
+        &self,
+        spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
+    ) -> Option<block::Height> {
         match self {
             // TODO: return an error for Request::Block to replace this check in the state (#2336)
             Request::Block { .. } => None,
 
-            Request::Mempool { transaction, height } => check::tx_transparent_coinbase_spends_maturity(
+            Request::Mempool {
+                transaction,
+                height,
+            } => check::tx_transparent_coinbase_spends_maturity(
                 transaction.transaction.clone(),
                 *height,
                 Default::default(),
-                spent_utxos
+                spent_utxos,
             ),
         }
     }
@@ -434,7 +440,7 @@ where
                     ))
                     .map(|res| {
                         assert!(
-                            res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, 
+                            res? == zs::Response::ValidBestChainTipNullifiersAndAnchors,
                             "unexpected response to CheckBestChainTipNullifiersAndAnchors request"
                         );
                         Ok(())
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -229,24 +229,16 @@ impl Request {
 
     /// Returns lowest height at which all transparent coinbase spends will be valid,
     /// or None if this transaction has no transparent coinbase spends.
-    pub fn maturity_height(
+    pub fn check_maturity_height(
         &self,
         spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
-    ) -> Option<block::Height> {
-        match self {
-            // TODO: return an error for Request::Block to replace this check in the state (#2336)
-            Request::Block { .. } => None,
-
-            Request::Mempool {
-                transaction,
-                height,
-            } => check::tx_transparent_coinbase_spends_maturity(
-                transaction.transaction.clone(),
-                *height,
-                Default::default(),
-                spent_utxos,
-            ),
-        }
+    ) -> Result<(), TransactionError> {
+        check::tx_transparent_coinbase_spends_maturity(
+            self.transaction(),
+            self.height(),
+            self.known_utxos(),
+            spent_utxos,
+        )
     }
 }
 
@@ -475,7 +467,12 @@ where
             }
 
             let legacy_sigop_count = cached_ffi_transaction.legacy_sigop_count()?;
-            let maturity_height = req.maturity_height(&spent_utxos);
+
+            // TODO: Return an error for Request::Block as well to replace this check in
+            //       the state once #2336 has been implemented?
+            if req.is_mempool() {
+                req.check_maturity_height(&spent_utxos)?;
+            }
 
             let rsp = match req {
                 Request::Block { .. } => Response::Block {
@@ -490,7 +487,6 @@ where
                             "unexpected mempool coinbase transaction: should have already rejected",
                         ),
                         legacy_sigop_count,
-                        maturity_height,
                     ),
                 },
             };
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -224,9 +224,22 @@ impl Request {
 
     /// Returns true if the request is a mempool request.
     pub fn is_mempool(&self) -> bool {
+        matches!(self, Request::Mempool { .. })
+    }
+
+    /// Returns lowest height at which all transparent coinbase spends will be valid,
+    /// or None if this transaction has no transparent coinbase spends.
+    pub fn maturity_height(&self, spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>) -> Option<block::Height> {
         match self {
-            Request::Block { .. } => false,
-            Request::Mempool { .. } => true,
+            // TODO: return an error for Request::Block to replace this check in the state (#2336)
+            Request::Block { .. } => None,
+
+            Request::Mempool { transaction, height } => check::tx_transparent_coinbase_spends_maturity(
+                transaction.transaction.clone(),
+                *height,
+                Default::default(),
+                spent_utxos
+            ),
         }
     }
 }
@@ -420,7 +433,10 @@ where
                         unmined_tx,
                     ))
                     .map(|res| {
-                        assert!(res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, "unexpected response to CheckBestChainTipNullifiersAndAnchors request");
+                        assert!(
+                            res? == zs::Response::ValidBestChainTipNullifiersAndAnchors, 
+                            "unexpected response to CheckBestChainTipNullifiersAndAnchors request"
+                        );
                         Ok(())
                     }
                 );
@@ -453,6 +469,7 @@ where
             }
 
             let legacy_sigop_count = cached_ffi_transaction.legacy_sigop_count()?;
+            let maturity_height = req.maturity_height(&spent_utxos);
 
             let rsp = match req {
                 Request::Block { .. } => Response::Block {
@@ -467,6 +484,7 @@ where
                             "unexpected mempool coinbase transaction: should have already rejected",
                         ),
                         legacy_sigop_count,
+                        maturity_height,
                     ),
                 },
             };
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -19,11 +19,7 @@ use zebra_chain::{
     parameters::{Network, NetworkUpgrade},
     primitives::zcash_note_encryption,
     transaction::{LockTime, Transaction},
-    transparent::{self, MIN_TRANSPARENT_COINBASE_MATURITY},
-};
-
-use zebra_chain::transparent::CoinbaseSpendRestriction::{
-    OnlyShieldedOutputs, SomeTransparentOutputs,
+    transparent,
 };
 
 use crate::error::TransactionError;
@@ -474,56 +470,6 @@ fn validate_expiry_height_mined(
     Ok(())
 }
 
-/// Check that `utxo` is spendable, based on the coinbase `spend_restriction`.
-///
-/// # Consensus
-///
-/// > A transaction with one or more transparent inputs from coinbase transactions
-/// > MUST have no transparent outputs (i.e. tx_out_count MUST be 0).
-/// > Inputs from coinbase transactions include Founders’ Reward outputs and
-/// > funding stream outputs.
-///
-/// > A transaction MUST NOT spend a transparent output of a coinbase transaction
-/// > from a block less than 100 blocks prior to the spend.
-/// > Note that transparent outputs of coinbase transactions include
-/// > Founders’ Reward outputs and transparent funding stream outputs.
-///
-/// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
-// TODO: Move tests for `zebra_state::service::check::utxo::transparent_coinbase_spend()` to this crate
-//       once the transaction verifier returns an error for block transactions that fail this check.
-fn transparent_coinbase_spend_maturity(
-    outpoint: transparent::OutPoint,
-    spend_restriction: transparent::CoinbaseSpendRestriction,
-    utxo: transparent::Utxo,
-) -> Result<(), TransactionError> {
-    if !utxo.from_coinbase {
-        return Ok(());
-    }
-
-    let min_spend_height = utxo.height + MIN_TRANSPARENT_COINBASE_MATURITY.try_into().unwrap();
-    let min_spend_height =
-        min_spend_height.expect("valid UTXOs have coinbase heights far below Height::MAX");
-
-    match spend_restriction {
-        OnlyShieldedOutputs { spend_height } => {
-            if spend_height >= min_spend_height {
-                Ok(())
-            } else {
-                Err(TransactionError::ImmatureTransparentCoinbaseSpend {
-                    outpoint,
-                    spend_height,
-                    min_spend_height,
-                    created_height: utxo.height,
-                })
-            }
-        }
-        SomeTransparentOutputs => Err(TransactionError::UnshieldedTransparentCoinbaseSpend {
-            outpoint,
-            min_spend_height,
-        }),
-    }
-}
-
 /// Accepts a transaction, UTXOs in the same block, and spent UTXOs from the chain.
 ///
 /// Returns the lowest/earlier height at which every transparent coinbase spend
@@ -545,7 +491,7 @@ pub fn tx_transparent_coinbase_spends_maturity(
 
         let spend_restriction = tx.coinbase_spend_restriction(height);
 
-        transparent_coinbase_spend_maturity(spend, spend_restriction, utxo)?;
+        zebra_state::transparent_coinbase_spend(spend, spend_restriction, &utxo)?;
     }
 
     Ok(())
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -2,7 +2,13 @@
 //!
 //! Code in this file can freely assume that no pre-V4 transactions are present.
 
-use std::{borrow::Cow, collections::HashSet, convert::TryFrom, hash::Hash};
+use std::{
+    borrow::Cow,
+    collections::{HashMap, HashSet},
+    convert::TryFrom,
+    hash::Hash,
+    sync::Arc,
+};
 
 use chrono::{DateTime, Utc};
 
@@ -13,6 +19,11 @@ use zebra_chain::{
     parameters::{Network, NetworkUpgrade},
     primitives::zcash_note_encryption,
     transaction::{LockTime, Transaction},
+    transparent::{self, MIN_TRANSPARENT_COINBASE_MATURITY},
+};
+
+use zebra_chain::transparent::CoinbaseSpendRestriction::{
+    OnlyShieldedOutputs, SomeTransparentOutputs,
 };
 
 use crate::error::TransactionError;
@@ -461,4 +472,102 @@ fn validate_expiry_height_mined(
     }
 
     Ok(())
+}
+
+/// Check that `utxo` is spendable, based on the coinbase `spend_restriction`.
+///
+/// # Consensus
+///
+/// > A transaction with one or more transparent inputs from coinbase transactions
+/// > MUST have no transparent outputs (i.e. tx_out_count MUST be 0).
+/// > Inputs from coinbase transactions include Founders’ Reward outputs and
+/// > funding stream outputs.
+///
+/// > A transaction MUST NOT spend a transparent output of a coinbase transaction
+/// > from a block less than 100 blocks prior to the spend.
+/// > Note that transparent outputs of coinbase transactions include
+/// > Founders’ Reward outputs and transparent funding stream outputs.
+///
+/// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+fn transparent_coinbase_spend_maturity(
+    outpoint: transparent::OutPoint,
+    spend_restriction: transparent::CoinbaseSpendRestriction,
+    utxo: transparent::Utxo,
+) -> Result<(), TransactionError> {
+    if !utxo.from_coinbase {
+        return Ok(());
+    }
+
+    let min_spend_height = utxo.height + MIN_TRANSPARENT_COINBASE_MATURITY.try_into().unwrap();
+    let min_spend_height =
+        min_spend_height.expect("valid UTXOs have coinbase heights far below Height::MAX");
+
+    match spend_restriction {
+        OnlyShieldedOutputs { spend_height } => {
+            if spend_height >= min_spend_height {
+                Ok(())
+            } else {
+                Err(TransactionError::ImmatureTransparentCoinbaseSpend {
+                    outpoint,
+                    spend_height,
+                    min_spend_height,
+                    created_height: utxo.height,
+                })
+            }
+        }
+        SomeTransparentOutputs => Err(TransactionError::UnshieldedTransparentCoinbaseSpend {
+            outpoint,
+            min_spend_height,
+        }),
+    }
+}
+
+/// Accepts a transaction, UTXOs in the same block, and spent UTXOs from the chain.
+///
+/// Returns the lowest/earlier height at which every transparent coinbase spend
+/// in the provided [`Transaction`] will be mature,
+///
+/// Returns None if the transaction has no transparent coinbase spends.
+pub fn tx_transparent_coinbase_spends_maturity(
+    tx: Arc<Transaction>,
+    height: Height,
+    block_new_outputs: Arc<HashMap<transparent::OutPoint, transparent::OrderedUtxo>>,
+    spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
+) -> Option<Height> {
+    let mut maturity_height = None;
+
+    for spend in tx.spent_outpoints() {
+        let utxo = block_new_outputs
+            .get(&spend)
+            .map(|ordered_utxo| ordered_utxo.utxo.clone())
+            .or_else(|| spent_utxos.get(&spend).cloned())
+            .expect("load_spent_utxos_fut.await should return an error if a utxo is missing");
+
+        let spend_restriction = tx.coinbase_spend_restriction(height);
+
+        let check_coinbase_spend_result =
+            transparent_coinbase_spend_maturity(spend, spend_restriction, utxo);
+
+        match check_coinbase_spend_result {
+            Err(TransactionError::ImmatureTransparentCoinbaseSpend {
+                min_spend_height, ..
+            })
+            | Err(TransactionError::UnshieldedTransparentCoinbaseSpend {
+                min_spend_height, ..
+            }) if maturity_height
+                .map(|h| h < min_spend_height)
+                .unwrap_or(true) =>
+            {
+                maturity_height = Some(min_spend_height);
+            }
+
+            Err(TransactionError::ImmatureTransparentCoinbaseSpend { .. })
+            | Err(TransactionError::UnshieldedTransparentCoinbaseSpend { .. })
+            | Ok(()) => {}
+
+            Err(_other) => panic!("unexpected error from transparent_coinbase_spend_maturity"),
+        };
+    }
+
+    maturity_height
 }
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -535,9 +535,7 @@ pub fn tx_transparent_coinbase_spends_maturity(
     height: Height,
     block_new_outputs: Arc<HashMap<transparent::OutPoint, transparent::OrderedUtxo>>,
     spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
-) -> Option<Height> {
-    let mut maturity_height = None;
-
+) -> Result<(), TransactionError> {
     for spend in tx.spent_outpoints() {
         let utxo = block_new_outputs
             .get(&spend)
@@ -547,29 +545,8 @@ pub fn tx_transparent_coinbase_spends_maturity(
 
         let spend_restriction = tx.coinbase_spend_restriction(height);
 
-        let check_coinbase_spend_result =
-            transparent_coinbase_spend_maturity(spend, spend_restriction, utxo);
-
-        match check_coinbase_spend_result {
-            Err(TransactionError::ImmatureTransparentCoinbaseSpend {
-                min_spend_height, ..
-            })
-            | Err(TransactionError::UnshieldedTransparentCoinbaseSpend {
-                min_spend_height, ..
-            }) if maturity_height
-                .map(|h| h < min_spend_height)
-                .unwrap_or(true) =>
-            {
-                maturity_height = Some(min_spend_height);
-            }
-
-            Err(TransactionError::ImmatureTransparentCoinbaseSpend { .. })
-            | Err(TransactionError::UnshieldedTransparentCoinbaseSpend { .. })
-            | Ok(()) => {}
-
-            Err(_other) => panic!("unexpected error from transparent_coinbase_spend_maturity"),
-        };
+        transparent_coinbase_spend_maturity(spend, spend_restriction, utxo)?;
     }
 
-    maturity_height
+    Ok(())
 }
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -489,6 +489,8 @@ fn validate_expiry_height_mined(
 /// > Founders’ Reward outputs and transparent funding stream outputs.
 ///
 /// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+// TODO: Move tests for `zebra_state::service::check::utxo::transparent_coinbase_spend()` to this crate
+//       once the transaction verifier returns an error for block transactions that fail this check.
 fn transparent_coinbase_spend_maturity(
     outpoint: transparent::OutPoint,
     spend_restriction: transparent::CoinbaseSpendRestriction,
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -5,7 +5,6 @@
 use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
-    convert::TryFrom,
     hash::Hash,
     sync::Arc,
 };
@@ -470,12 +469,11 @@ fn validate_expiry_height_mined(
     Ok(())
 }
 
-/// Accepts a transaction, UTXOs in the same block, and spent UTXOs from the chain.
+/// Accepts a transaction, block height, block UTXOs, and
+/// the transaction's spent UTXOs from the chain.
 ///
-/// Returns the lowest/earlier height at which every transparent coinbase spend
-/// in the provided [`Transaction`] will be mature,
-///
-/// Returns None if the transaction has no transparent coinbase spends.
+/// Returns `Ok(())` if spent transparent coinbase outputs are
+/// valid for the block height, or a [`Err(TransactionError)`](TransactionError)
 pub fn tx_transparent_coinbase_spends_maturity(
     tx: Arc<Transaction>,
     height: Height,
@@ -491,7 +489,7 @@ pub fn tx_transparent_coinbase_spends_maturity(
 
         let spend_restriction = tx.coinbase_spend_restriction(height);
 
-        zebra_state::transparent_coinbase_spend(spend, spend_restriction, &utxo)?;
+        zebra_state::check::transparent_coinbase_spend(spend, spend_restriction, &utxo)?;
     }
 
     Ok(())
--- a/zebra-consensus/src/transaction/tests.rs
+++ b/zebra-consensus/src/transaction/tests.rs
@@ -25,7 +25,7 @@ use zebra_chain::{
         },
         Hash, HashType, JoinSplitData, LockTime, Transaction,
     },
-    transparent::{self, CoinbaseData, MIN_TRANSPARENT_COINBASE_MATURITY},
+    transparent::{self, CoinbaseData},
 };
 
 use zebra_test::mock_service::MockService;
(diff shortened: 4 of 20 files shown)


teor2345 (dismissed):
> The getblocktemplate method can currently construct invalid block templates if there are mempool transactions with transparent coinbase spends that are immature at the next block height.

So this is kind of a big question, but should these invalid transactions be in the mempool at all?

If we make the changes in this PR, we'll store invalid transactions in the mempool, and relay them to other nodes. And those invalid transactions could cause valid transactions to be rejected from the mempool due to conflicts, or due to mempool capacity limits. (They also waste CPU and bandwidth across the Zcash network.)

Looking at our design principles, we'd be breaking one of the core properties of Zebra's implementation: we don't store invalid data. It would be particularly confusing to store an invalid transaction in a type called `VerifiedUnminedTx`:
https://github.com/ZcashFoundation/zebra/blob/166526a088ae127d88714d93b791b95a35c42447/zebra-chain/src/transaction/unmined.rs#L294-L299

As an alternative, if the mempool transaction verifier checks transactions against the best tip height, and returns a `StorageEffectsTip` error if they can't be spent yet, they won't be stored in our mempool. This means:
- they can't be mined into blocks
- they won't be relayed to other nodes

And when the tip updates, Zebra will re-download them and try to verify them again. Which is what we want.

(If we want to store unmined transactions that could potentially become valid, like `zcashd`, then that's a bigger decision. I think it should probably go back to the whole team.)

I'm sorry I have to give feedback like this, because I think this is a neat idea. (And it's quite a clever workaround!) But I'm concerned it could cause maintenance issues like subtle bugs and developer confusion.

zebra-consensus/src/transaction.rs
```diff
@@ -224,9 +224,28 @@ impl Request {
...
+    /// or None if this transaction has no transparent coinbase spends.
+    pub fn maturity_height(
+        &self,
+        spent_utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
+    ) -> Option<block::Height> {
         match self {
-            Request::Block { .. } => false,
-            Request::Mempool { .. } => true,
+            // TODO: return an error for Request::Block to replace this check in the state (#2336)
+            Request::Block { .. } => None,
```
teor2345: In the non-finalized state, mined transactions are verified in the context of their specific chain fork. So this block transaction check actually belongs in the state.

(If we try to do it here against the best chain tip, we'll be rejecting some valid blocks on side chains, which stops those side chains later becoming the best chain. This could cause a consensus split.)

Sorry if I was the one to put that note in the ticket, I can see it's wrong now.

arya2: > (If we try to do it here against the best chain tip, we'll be rejecting some valid blocks on side chains, which stops those side chains later becoming the best chain. This could cause a consensus split.)

It's comparing block (coinbase) heights to spent utxo heights (which should be reliable after #2336), and the transaction verifier calls the state with `AwaitUtxo`  requests to for block transactions so it should find spent UTXOs in any chain.

> Sorry if I was the one to put that note in the ticket, I can see it's wrong now.

I added the possible design, hoping the transaction verifier already had the information to check [`utxo::transparent_spend()`](https://github.com/ZcashFoundation/zebra/blob/2a48d4cf259f82b050e6d08a943844f7d78ef662/zebra-state/src/service/check/utxo.rs) in parallel. I still think that should work after #2336?

arya2:
> So this is kind of a big question, but should these invalid transactions be in the mempool at all?

I don't think so, they're not in the zcashd mempool.

>If we make the changes in this PR, we'll store invalid transactions in the mempool, and relay them to other nodes. And those invalid transactions could cause valid transactions to be rejected from the mempool due to conflicts, or due to mempool capacity limits. (They also waste CPU and bandwidth across the Zcash network.)

Great point, it looks like zcashd always immediately rejects transactions with immature transparent coinbase spends at the next block height from the mempool  (at the end of [`AcceptToMemoryPool()`](https://github.com/zcash/zcash/blob/0b9126a846723e447f65cc080e4ce857f241dd77/src/main.cpp#L2025-L2028)), Zebra should do the same (or at least avoid relaying these transactions).

> (If we want to store unmined transactions that could potentially become valid, like zcashd, then that's a bigger decision. I think it should probably go back to the whole team.)

I thought zcashd stashes these (it stashes other invalid transactions that may become valid but not these, whoops).

> I'm sorry I have to give feedback like this, because I think this is a neat idea. (And it's quite a clever workaround!) But I'm concerned it could cause maintenance issues like subtle bugs and developer confusion.

Hey, that's what it takes to make good code, and I took the risk by not looking closely enough at what zcashd is doing. Thank you for helping catch the mistake.

zebra-consensus/src/transaction.rs
teor2345: > > (If we try to do it here against the best chain tip, we'll be rejecting some valid blocks on side chains, which stops those side chains later becoming the best chain. This could cause a consensus split.)
> 
> It's comparing block (coinbase) heights to spent utxo heights (which should be reliable after #2336), and the transaction verifier calls the state with `AwaitUtxo` requests to for block transactions so it should find spent UTXOs in any chain.

I'm not sure, I'd have to think about it a bit more.

It seems to go against the general design principle of doing contextual verification in the state. But that's not a blocker if the change is required to fix a serious bug.

Just making sure we're both on the same page, #2336 wasn't implemented, and it isn't planned any time soon.

arya2: > It seems to go against the general design principle of doing contextual verification in the state. But that's not a blocker if the change is required to fix a serious bug.

It's not required to fix anything but the transaction verifier seems to already have all of the information needed to do the check in parallel. It also avoids having the check in both the transaction verifier and the state.

> Just making sure we're both on the same page, https://github.com/ZcashFoundation/zebra/issues/2336 wasn't implemented, and it isn't planned any time soon.

It's currently only a TODO for when #2336 is implemented, it skips the check for block requests.

arya2:
Updated in https://github.com/ZcashFoundation/zebra/pull/6510/commits/d8c8160d24b9765f6c45246655723b8bae1ef8da so it rejects transactions with spends that aren't valid at the next block height.

Unified `transparent_coinbase_spend()` with `transparent_coinbase_spend_maturity()` in https://github.com/ZcashFoundation/zebra/pull/6510/commits/a45a6f78c849069622c7f0bdba4a01e2d1940219.

I left the TODO for moving the check out of the state, I still think it would work, parallelize the check, and possibly avoid duplicate lookups.

zebra-chain/src/transparent/utxo.rs
```diff
@@ -54,6 +54,24 @@ pub struct OrderedUtxo {
...
+        &ordered_utxo.utxo
+    }
+}
+
+impl From<OrderedUtxo> for Utxo {
+    fn from(ordered_utxo: OrderedUtxo) -> Self {
+        ordered_utxo.utxo
+    }
+}
+
```
arya2: These aren't currently used anywhere.

```suggestion
```

teor2345: I think it's ok to delete them, `ordered_utxo.utxo` is clearer and shorter.

zebra-consensus/src/transaction.rs
```diff
@@ -454,6 +468,12 @@ where
 
             let legacy_sigop_count = cached_ffi_transaction.legacy_sigop_count()?;
 
+            // TODO: Return an error for Request::Block as well to replace this check in
```
teor2345: This needs to be a WONTFIX because ticket #2336 is closed, see our new process in:
- #6281

```suggestion
            // WONTFIX: Return an error for Request::Block as well to replace this check in
```

zebra-consensus/src/transaction/check.rs
```diff
@@ -2,7 +2,13 @@
 //!
 //! Code in this file can freely assume that no pre-V4 transactions are present.
 
-use std::{borrow::Cow, collections::HashSet, convert::TryFrom, hash::Hash};
+use std::{
+    borrow::Cow,
+    collections::{HashMap, HashSet},
+    convert::TryFrom,
```
teor2345: Nitpick: In Rust 2021 edition, TryFrom and TryInto are redundant and can be deleted. (They are in the prelude already.)

```suggestion
```

zebra-consensus/src/transaction.rs
```diff
@@ -224,10 +224,21 @@ impl Request {
 
     /// Returns true if the request is a mempool request.
     pub fn is_mempool(&self) -> bool {
-        match self {
-            Request::Block { .. } => false,
-            Request::Mempool { .. } => true,
-        }
+        matches!(self, Request::Mempool { .. })
```
teor2345, line 227: Note: We've been writing out all the variants explicitly so we are forced to update them all if the enum changes.

But this particular enum probably won't change, and this implementation of this method will likely remain valid if it does.

zebra-consensus/src/transaction.rs
```diff
@@ -224,10 +224,21 @@ impl Request {
...
+    }
+
+    /// Returns lowest height at which all transparent coinbase spends will be valid,
+    /// or None if this transaction has no transparent coinbase spends.
+    pub fn check_maturity_height(
```
teor2345: Nitpick: Currently, this module follows a pattern where inexpensive field lookups are on the Request and Response types, and verification is on the Verifier type.

Can you move this method to the verifier type?

Can you also document what this function actually does, including each argument and the return value?

zebra-consensus/src/transaction.rs
```diff
@@ -454,6 +468,12 @@ where
 
             let legacy_sigop_count = cached_ffi_transaction.legacy_sigop_count()?;
 
+            // TODO: Return an error for Request::Block as well to replace this check in
+            //       the state once #2336 has been implemented?
+            if req.is_mempool() {
+                req.check_maturity_height(&spent_utxos)?;
```
teor2345: This check is much faster than the proof or signature verifiers, and we know that it sometimes fails. (And that we'll be retrying it every time the tip changes.)

So this check should be done before the async checks, to avoid running the expensive shielded proof or signature verifiers for transactions with immature coinbase spends.

zebra-consensus/src/transaction/check.rs
```diff
@@ -462,3 +469,30 @@ fn validate_expiry_height_mined(
...
     Ok(())
 }
+
+/// Accepts a transaction, UTXOs in the same block, and spent UTXOs from the chain.
+///
+/// Returns the lowest/earlier height at which every transparent coinbase spend
+/// in the provided [`Transaction`] will be mature,
+///
+/// Returns None if the transaction has no transparent coinbase spends.
+pub fn tx_transparent_coinbase_spends_maturity(
```
teor2345, line 477: Please update this function's documentation to match the arguments and return values.

zebra-consensus/src/transaction/tests.rs
```diff
@@ -564,6 +564,83 @@ async fn mempool_request_with_past_lock_time_is_accepted() {
     );
 }
 
+#[tokio::test]
+async fn mempool_response_maturity_height_correct() {
```
teor2345: Can you add a test for the success case as well?

Nitpick: this test's name makes it seem like it is testing the success case, and it has no documentation.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
```diff
@@ -271,7 +271,6 @@ where
         unreachable!("unmatched response to a mempool::FullTransactions request")
     };
 
-    // Check that the mempool and state were in sync when we made the requests
```
teor2345: This deleted comment seems important, and it's unrelated to this PR

zebra-state/src/lib.rs
```diff
@@ -36,7 +36,7 @@ pub use request::{FinalizedBlock, HashOrHeight, PreparedBlock, ReadRequest, Requ
 pub use response::{KnownBlock, MinedTx, ReadResponse, Response};
 pub use service::{
     chain_tip::{ChainTipChange, LatestChainTip, TipAction},
-    init, spawn_init,
+    init, spawn_init, transparent_coinbase_spend,
```
teor2345: Nitpick: for readability, this method is meant to be called as `check::transparent_coinbase_spend`, but if we don't want to expose the module, we can make it an underscore instead.

```suggestion
    init, spawn_init, transparent_coinbase_spend as check_ transparent_coinbase_spend,
```

zebra-consensus/src/transaction/tests.rs
```diff
@@ -564,6 +564,83 @@ async fn mempool_request_with_past_lock_time_is_accepted() {
...
+
+    let verifier_response = verifier
+        .oneshot(Request::Mempool {
+            transaction: tx.into(),
+            height,
+        })
+        .await;
+
+    assert!(
+        verifier_response.is_err(),
```
teor2345: Can we `assert_eq!` the specific failure we're expecting?

Otherwise, a refactor could make the test fail for an unrelated reason, and it would still pass.

zebra-state/src/service/check/tests/utxo.rs
```diff
@@ -53,8 +53,12 @@ fn accept_shielded_mature_coinbase_utxo_spend() {
     };
 
     let result =
-        check::utxo::transparent_coinbase_spend(outpoint, spend_restriction, ordered_utxo.clone());
-    assert_eq!(result, Ok(ordered_utxo));
+        check::utxo::transparent_coinbase_spend(outpoint, spend_restriction, ordered_utxo.as_ref());
+
+    assert!(
+        result.is_ok(),
```
teor2345: Using `assert_eq!` provides better diagnostics on test failure, because it shows the specific error that was returned.

```rust
assert_eq!(result, Ok(()), "mature transparent coinbase spend check should return Ok");
```

teor2345 (commented):
This seems like something that we could merge as-is, but I'd like to improve the test coverage and method documentation.

zebra-consensus/src/transaction.rs
arya2: Noted.

I think we should also try to use `matches!` in these cases so we _aren't_ forced to update methods that remain correct when the enum changes.

zebra-consensus/src/transaction.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-consensus/src/transaction/check.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-consensus/src/transaction.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-consensus/src/transaction.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-consensus/src/transaction/check.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-consensus/src/transaction/tests.rs
arya2: It actually was a test for the success case when it was returning `maturity_height: Some(_)` instead of an error, now the success case is covered by the test right above it.

Renamed/updated for the failure case and documented in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: It was very important and mistakenly deleted when reverting the changes in this file. Fixed in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290, thank you for catching that.

zebra-state/src/lib.rs
arya2: Exported the `check` module instead in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290, `transparent_coinbase_spend` is the only member that's public beyond the crate.

zebra-consensus/src/transaction/tests.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

zebra-state/src/service/check/tests/utxo.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/6510/commits/e6c6c0f33593095568ffef531379d445f2baa290.

teor2345 (approved):
Thanks, let's get this merged to fix CI!

PR #6521, opened 2023-04-16
Increase ZIP-401 mempool cost thresholds for Orchard transactions

teor2345:
## Motivation

This PR changes Zebra's mempool cost thresholds to match `zcashd` 5.5.0, as specified in the latest ZIP-401 update:
https://github.com/zcash/zips/pull/685/files#diff-e8ce48f91a8494325daeea2a8bfb2f5a4f1d6d261d0c3a985da463b03a24e14cR82

This is a standard rule, so it is not consensus-critical.

## Solution

Update the mempool cost threshold constants to match ZIP-401 and `zcashd`

## Review

This is a routine change.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -43,15 +43,23 @@ mod zip317;
 /// Contributes to the randomized, weighted eviction of transactions from the
 /// mempool when it reaches a max size, also based on the total cost.
 ///
+/// # Standard Rule
+///
 /// > Each transaction has a cost, which is an integer defined as:
 /// >
-/// > max(serialized transaction size in bytes, 4000)
+/// > max(memory size in bytes, 10000)
 /// >
-/// > The threshold 4000 for the cost function is chosen so that the size in bytes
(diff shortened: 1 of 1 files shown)


zebra-chain/src/transaction/unmined.rs
```diff
@@ -54,13 +54,13 @@ mod zip317;
 /// > transparent transactions because of their size.
 ///
 /// [ZIP-401]: https://zips.z.cash/zip-0401
-pub const MEMPOOL_TRANSACTION_COST_THRESHOLD: u64 = 4000;
+pub const MEMPOOL_TRANSACTION_COST_THRESHOLD: u64 = 10_000;
```
daira, line 65: Also update the comment with the new spec and rationale.

arya2 (dismissed):
This looks good, thank you!

(Please add updating the comment to follow-up work if this merges as-is)

zebra-chain/src/transaction/unmined.rs
teor2345: Thanks, fixed in 0da6b5f0dc

arya2 (approved):

PR #6556, opened 2023-04-22
feat(mempool): add ZIP-317 rules to mempool

### LINKED ISSUE #5336: Relay and accept mempool transactions based on the ZIP-317 fee rules

oxarbitrage:
## Motivation

We want to add ZIP-317 relay rules to the mempool verifier, so Zebra can drop transactions before they get into the mempool storage.

Close https://github.com/ZcashFoundation/zebra/issues/5336

### Specifications

- https://zips.z.cash/zip-0317#transaction-relaying

## Solution

This pull request is checking 2 rules currently:

- Check if transaction unpaid actions is under a threshold.
- Check if the miner fee is greater than the transaction conventional fee.

If there are more rules they need to be added to `zebra_chain::transactions::unmined::mempool_checks` function.

### Reviewer Checklist


zebra-chain/src/transaction/unmined.rs
```diff
@@ -335,6 +335,9 @@ impl fmt::Display for VerifiedUnminedTx {
             .finish()
     }
 }
+/// The ZIP-317 recommended limit on the number of unpaid actions per block.
+/// `block_unpaid_action_limit` in ZIP-317.
+pub const BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT: u32 = 50;
```
teor2345: Optional: this is a ZIP-317 constant, so it can go in the zip317 module.

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -153,3 +158,34 @@ fn div_ceil(quotient: usize, divisor: usize) -> usize {
...
 }
+
+/// Make ZIP-317 checks before inserting a transaction into the mempool.
+pub fn mempool_checks(
+    unpaid_actions: u32,
+    unpaid_action_limit: u32,
+    miner_fee: Amount<NonNegative>,
+    conventional_fee: Amount<NonNegative>,
+) -> Result<(), Error> {
+    // Check unpaid actions is below the thershold.
```
teor2345: Typo

```suggestion
    // Check unpaid actions is below the threshold.
```

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -153,3 +158,34 @@ fn div_ceil(quotient: usize, divisor: usize) -> usize {
...
+    if unpaid_actions > unpaid_action_limit {
```
teor2345: Can you quote the part of the ZIP that is being implemented here?

I've just been using `# Standard Rule` as the heading, because ZIP-317 is only a recommended set of standard rules.

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -153,3 +158,34 @@ fn div_ceil(quotient: usize, divisor: usize) -> usize {
...
+        return Err(Error::UnpaidActions);
+    }
+
+    // Check the fee is not below the calculated conventional fee for the transaction.
```
teor2345: I'm not sure where this rule comes from, can you quote it here?

zebra-consensus/src/transaction/tests.rs
```diff
@@ -2726,3 +2817,145 @@ fn shielded_outputs_are_not_decryptable_for_fake_v5_blocks() {
         );
     }
 }
+
+#[tokio::test]
+async fn mempool_zip317_error() {
```
teor2345, line 2822: Thanks for all these tests!

zebra-chain/src/transaction/unmined.rs
```diff
@@ -353,17 +356,24 @@ impl VerifiedUnminedTx {
...
         let unpaid_actions = zip317::unpaid_actions(&transaction, miner_fee);
 
-        Self {
+        zip317::mempool_checks(
+            unpaid_actions,
+            BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT,
+            miner_fee,
+            transaction.conventional_fee,
+        )?;
+
```
teor2345, line 361: As part of this ticket, we also wanted to enforce a minimum mempool fee, like `zcashd` does. This is not from ZIP-317, it's from Bitcoin.

Here's Daira's summary of what `zcashd` does:
https://github.com/ZcashFoundation/zebra/issues/5336#issuecomment-1506748801

I think the ticket was really complicated, there were a lot of updates over a few months. I've just edited it to remove the outdated and completed work.

teor2345 (commented):
Looks good, thanks for doing this!
I hope it wasn't too complicated, it seems like updating the tests was a bit of work.

I just want to check where these rules are coming from, to make sure we got them all.

zebra-chain/src/transaction/unmined/zip317.rs
oxarbitrage: Done.

zebra-chain/src/transaction/unmined.rs
oxarbitrage: Done.

zebra-chain/src/transaction/unmined/zip317.rs
oxarbitrage: Done in https://github.com/ZcashFoundation/zebra/pull/6556/commits/2140ce270cd0c29c7d707b1bb36ba4c197e6a240

zebra-chain/src/transaction/unmined/zip317.rs
oxarbitrage: Quoted https://github.com/ZcashFoundation/zebra/pull/6556/commits/2140ce270cd0c29c7d707b1bb36ba4c197e6a240

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -139,3 +147,64 @@ fn conventional_actions(transaction: &Transaction) -> u32 {
...
+    }
+
+    // # Standard rule
+    //
+    // > Minimum Fee Rate
+    // >
+    // > Transactions must pay a fee of at least 100 zatoshis per 1000 bytes of serialized size,
+    // > with a maximum fee of 1000 zatoshis. In zcashd this is `DEFAULT_MIN_RELAY_TX_FEE`.
+    //
+    // <https://github.com/ZcashFoundation/zebra/issues/5336#issuecomment-1506748801>
```
oxarbitrage: This probably needs to be changed, i was unsure from where to quote this rule.

zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -139,3 +147,64 @@ fn conventional_actions(transaction: &Transaction) -> u32 {
...
+    let limit = std::cmp::min(
+        1000,
+        (f32::floor(100.0 * transaction_size as f32 / 1000.0)) as u64,
+    );
+    if miner_fee < Amount::<NonNegative>::try_from(limit).expect("limit value is invalid") {
```
oxarbitrage: This can probably be improved, i just added [Daira's formula](https://github.com/ZcashFoundation/zebra/issues/5336#issuecomment-1506748801) as it is posted but it needs more work. Any hint is appreciated.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: Ah, I see where this first rule came from! This rule says that we can refuse to relay all of these transactions, but it is our choice which ones we relay. It doesn't force us to stop relaying any transactions.

In the ZIP editors meeting, we decided to be similar to `zcashd`, which is actually the next check you've already written.

So we can use these docs to justify the next check, but delete the first check, because it is too strict:
```suggestion
    // # Standard Rule
    //
    // > Nodes that normally relay transactions are expected to do so for transactions that pay at least the
    // > conventional fee as specified in this ZIP.
    //
    // <https://zips.z.cash/zip-0317#transaction-relaying>
    //
    // In Zebra, we use a similar minimum fee rate to `zcashd` v5.5.0 and later.
    // Transactions must pay a fee of at least 100 zatoshis per 1000 bytes of serialized size,
    // with a maximum fee of 1000 zatoshis. 
    //
    // <https://github.com/zcash/zcash/blob/9e856cfc5b81aa2607a16a23ff5584ea10014de6/src/amount.cpp#L24-L37>
    //
    // In zcashd this is `DEFAULT_MIN_RELAY_TX_FEE` and `LEGACY_DEFAULT_FEE`:
    // <https://github.com/zcash/zcash/blob/f512291ff20098291442e83713de89bcddc07546/src/main.h#L71-L72>
    // <https://github.com/zcash/zcash/blob/9e856cfc5b81aa2607a16a23ff5584ea10014de6/src/amount.h#L35-L36>
```

teor2345: It's tricky because the `zcashd` code is the only reference, and it has been changing while this PR has been open. I think I found the most recent code, I linked to it below.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: I'm going to need the compiler to help me with this one, so I'll open another PR on top of this PR.

teor2345 (commented):
Thanks for these changes, I am going to go and see what I can do with the fee rate code using the Rust compiler.

zebra-chain/src/transaction/unmined/zip317.rs
teor2345: https://github.com/ZcashFoundation/zebra/pull/6585/files

teor2345 (approved):
Thanks for this, looks great!

PR #6565, opened 2023-04-25
change(release): Add arborist and comms summaries to the release checklist

teor2345:
## Motivation

We want to make it easy to share Zebra releases with users. Currently we do that in Arborist calls, and on Twitter.

## Solution

Add a checklist item for posting a release summary on Slack.

## Review

This is a low priority change.

### Reviewer Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -166,8 +166,9 @@ and the updated changelog:
 - [ ] Un-freeze the [`batched` queue](https://dashboard.mergify.com/github/ZcashFoundation/repo/zebra/queues) using Mergify.
 
 
-## Blog Post
+## Telling Zebra Users
 
+- [ ] Post a summary of the important changes in the release in the `#arborist` and `#communications` Slack channels
 If the release contains new features (`major` or `minor`), or high-priority bug fixes:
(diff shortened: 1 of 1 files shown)


dconnolly (approved):

mpguerra:
Could we automate this somehow in future in discussion with comms team? We're adding more things to this checklist when we have a pending review of it to see if there are any steps that can be removed

teor2345:
> Could we automate this somehow in future in discussion with comms team? We're adding more things to this checklist when we have a pending review of it to see if there are any steps that can be removed

I don't think we should block changes to the release checklist on that discussion, since it has been pending for about a month now. Instead, we should justify any changes in terms of complexity. (We already have a section for complexity in the PR template.)

Here's a justification for this PR:

Forgetting to do the tasks added in this PR could have worse consequences than adding another step to the release checklist:
- whoever is giving the Arborist update needs this information, and we've missed important changes in the past
- we have an open PR that will require users to upgrade Zebra regularly #6351

I think automation is a good idea, but we shouldn't block changes on it either. Because it helps to have a full list of things we want to automate, so we can prioritise them, and choose a solution that covers the most important ones. In this case, we could automate posting a link to the release notes to those Slack channels.

We could split parts of the release checklist between different people, if that helps. Some teams have an entire dedicated release manager role.

Issue #1471, opened 2020-12-07
Panic: ClientRequest oneshot sender must not be dropped


rex4539:
```
Dec 07 08:56:27.460  INFO {zebrad="5b687f87"}:sync:obtain_tips: zebra_network::peer_set::set: network request with no ready peers: finding more peers, waiting for 18 peers to answer requests
Dec 07 10:42:24.595  INFO {zebrad="5b687f87"}:peer{addr=185.153.231.82:8233}:msg_as_req{msg=getheaders}:inbound:state: zebra_state::service: responding to peer GetBlocks or GetHeaders final_height=Height(1067845) response_len=127 chain_tip_height=Height(1067845) stop_height=None intersection_height=Some(Height(1067718))
Dec 07 10:42:24.775  INFO {zebrad="5b687f87"}:peer{addr=47.253.0.14:8233}:msg_as_req{msg=getheaders}:inbound:state: zebra_state::service: responding to peer GetBlocks or GetHeaders final_height=Height(1067845) response_len=127 chain_tip_height=Height(1067845) stop_height=None intersection_height=Some(Height(1067718))
Dec 07 10:42:25.279  INFO {zebrad="5b687f87"}:peer{addr=47.94.92.186:8233}:msg_as_req{msg=getheaders}:inbound:state: zebra_state::service: responding to peer GetBlocks or GetHeaders final_height=Height(1067845) response_len=127 chain_tip_height=Height(1067845) stop_height=None intersection_height=Some(Height(1067718))
```

## Error
```
ClientRequest oneshot sender must not be dropped before send: Canceled
```

## Metadata
|key|value|
|--|--|
|**version**|3.0.0-alpha.0|
|**git commit**|5b687f87|
|**location**|zebra-network/src/peer/client.rs:84:26|

## SpanTrace

<details>

```
SpanTrace:
   0: zebrad::components::sync::obtain_tips
             at zebrad/src/components/sync.rs:255
   1: zebrad::components::sync::sync
             at zebrad/src/components/sync.rs:150
   2: zebrad::commands::
           with zebrad="5b687f87"
             at zebrad/src/commands.rs:56
```
</details>

## Backtrace

<details>

```
Backtrace:
   0: backtrace::backtrace::libunwind::trace
             at .cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/backtrace/libunwind.rs:90:5
      backtrace::backtrace::trace_unsynchronized
[...]

PR #6566, opened 2023-04-25
fix(net): Avoid a rare panic when a connection is dropped

teor2345:
## Motivation

Zebra sometimes panics when a `Connection` is dropped and there is no error response:
> Message:  ClientRequest oneshot sender must not be dropped before send: Canceled
> Location: zebra-network/src/peer/client.rs:625
> Metadata:
> version: 1.0.0-rc.6

https://github.com/ZcashFoundation/zebra/actions/runs/4764031851/jobs/8495315588#step:8:543

This fixes a regression, the original ticket was #1471.

This is a rare panic due to concurrency.

## Solution

Replace the panic with returning a generic error.

## Review

This is a low-priority security fix. It seems like it could be remotely triggerable, but I'm not sure.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -621,8 +621,13 @@ impl Service<Request> for Client {
             Ok(()) => {
                 // The receiver end of the oneshot is itself a future.
                 rx.map(|oneshot_recv_result| {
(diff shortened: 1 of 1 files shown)


dconnolly (approved):
lgtm

PR #6567, opened 2023-04-25
fix(net): Reject nodes using ZClassic ports, and warn if configured with those ports

teor2345:
## Motivation

We want to reject ZClassic nodes by port, because some block explorers are picking them up and sharing their addresses:
https://github.com/zcash/zcash/issues/6505#issuecomment-1520649629

Zebra currently rejects ZClassic nodes because their protocol versions are too low, but we need to make a connection to do that. ZClassic could also increase their protocol versions in future.

We also want to warn the user if they configure their node with those ports.

## Solution

- Add the ZClassic ports to our existing node port check
- Do the regtest and coin checks even if we don't have a network
- Replace our Mainnet/Tesnet wrong port check with the check we do on remote node addresses, including ports used by other coins

## Review

This is a low priority peer handling and config issue.

### Reviewer Checklist


zebra-network/src/peer/priority.rs
```diff
@@ -123,9 +123,11 @@ fn address_is_valid_for_outbound_connections(
...
                 "invalid peer port: port is for Regtest, but Zebra does not support that network",
             );
-        } else if [16125, 26125].contains(&peer_addr.port()) {
-            // 16125/26125 is used by Flux/ZelCash, which uses the same network magic numbers as Zcash,
-            // so we have to reject it by port
+        } else if [8033, 18033, 16125, 26125].contains(&peer_addr.port()) {
+            // These coins use the same network magic numbers as Zcash, so we have to reject them by port:
+            // - ZClassic: 8033/18033
+            //   https://github.com/ZclassicCommunity/zclassic/blob/504362bbf72400f51acdba519e12707da44138c2/src/chainparams.cpp#L130
+            // - Flux/ZelCash: 16125/26125
```
daira: Does Zebra also error/warn if it is configured to listen on these ports?

teor2345: Thanks for the reminder, I updated this PR to make that happen:
https://github.com/ZcashFoundation/zebra/pull/6567/files#diff-890f028f155b45da59a5072f406a55b97d98ea27b27b40bad3d716627997e318R471-R476

I think we always meant to re-use that code for the listener, we just hadn't got around to it.

zebra-network/src/peer/priority.rs
```diff
@@ -105,30 +105,47 @@ fn address_is_valid_for_outbound_connections(
...
-            return Err("invalid peer port: port is for a non-Zcash network");
         }
     }
 
+    // Ignore ports used by potentially compatible nodes: other coins and unsupported Zcash regtest.
+    if listener_addr.port() == 18344 {
+        return Err(
+            "invalid peer port: port is for Regtest, but Zebra does not support that network",
+        );
+    } else if [8033, 18033, 16125, 26125].contains(&listener_addr.port()) {
```
oxarbitrage, line 142: Question: Our acceptance test use random ports. If one of these ports is chosen then the test will fail probably. Is this correct and accepted ?

teor2345: This is a good question!

In most tests, we ask the OS for an unallocated port in our tests, and it should choose from an ephemeral port range:
https://github.com/ZcashFoundation/zebra/blob/beb45fc514d78f91fde18a602d475ff55102c4fe/zebra-test/src/net.rs#L110-L132

The ephemeral port range is usually above 49,000.

If it's not, the probability of test failure would be approximately 6/65535 or 1 in 10,000 test functions. I don't think we've ever seen a failure like this happen, so it seems like the OS is choosing from the upper range.

oxarbitrage: Ok, thats good to know, thanks.

oxarbitrage (approved):

teor2345:
@mergifyio update

PR #6516, opened 2023-04-14
change(utils): Add a direct connection mode to zebra-checkpoints

teor2345:
## Instructions

Test the new direct JSON-RPC connections using:
```sh
cargo run --release --features=zebra-checkpoints --bin zebra-checkpoints -- --transport direct --addr 127.0.0.1:<zebrad port>
```

And the refactored existing functionality using:
```sh
cargo run --release --features=zebra-checkpoints --bin zebra-checkpoints -- --transport cli --addr 127.0.0.1:<zebrad port>
# automatically uses zcashd
cargo run --release --features=zebra-checkpoints --bin zebra-checkpoints -- --transport cli --backend zcashd
# actually uses zcashd, but in zebrad RPC parameter mode
cargo run --release --features=zebra-checkpoints --bin zebra-checkpoints -- --transport cli --backend zebrad
```

## Motivation

In #6367 we want to generate checkpoints in CI. This is the Rust tool part of that ticket.

### Complex Code or Requirements

There are 4 modes in this tool now:
- zebrad direct
- zcashd CLI
- zebrad CLI
- zcashd CLI in zebrad mode (not supported but it should work)

This means we need to test all of them. (Which I've done.)

Technically this code is async, but it doesn't actually use much concurrency.

## Solution

- [Move all the CLI-specific code into a single function](https://github.com/ZcashFoundation/zebra/commit/86b7cb291078d57ffda48259ee737eeecf7bd6d7)
- [Add transport and addr arguments to zebra-checkpoints](https://github.com/ZcashFoundation/zebra/commit/e14aab69b21b89dc158c115b8229544d86ce7c6e)
- [Make zebra-checkpoints support both CLI and direct JSON-RPC connections](https://github.com/ZcashFoundation/zebra/commit/f10a7889dc4d395d6207ed5918892519fb1a533e)
- [Document required features](https://github.com/ZcashFoundation/zebra/commit/9daf7d87654455c69c3f84a77a00b668f06e4318)

Related Changes:
- simplify the `zebra-checkpoints` implementation
- [Add a json-conversion feature for converting JSON to valid Heights](https://github.com/ZcashFoundation/zebra/commit/b9ec68bfd845d26f7fc5ff8d1465a97875d934bb)
    - Add a `TryIntoHeight` convenience trait for converting integers, strings, and JSON to valid heights (we'll use this in other fixes like #6279)
[...]

PR #6579, opened 2023-04-26
fix(log): Fix confusing comments and error messages in zebra-checkpoints

teor2345:
## Motivation

Minor fixes from @upbqdn's review of PR #6516.

## Solution

Fix the issues listed in this PR review:
https://github.com/ZcashFoundation/zebra/pull/6516#pullrequestreview-1402212516

## Review

Marek can review this, it's not urgent at all.

### Reviewer Checklist


teor2345:
@mergifyio update

zebra-utils/src/bin/zebra-checkpoints/main.rs
```diff
@@ -227,15 +234,19 @@ async fn main() -> Result<()> {
 
                 let block_bytes: Vec<u8> = hex::decode(block_bytes)?;
 
-                // TODO: is it faster to call both `getblock height 0` and `getblock height 1`,
-                //       rather than deserializing the block and calculating its hash?
+                // TODO: is it faster to call both `getblock height version=0`
+                //       and `getblock height version=1`, rather than deserializing the block
```
upbqdn: I spent a while trying to figure out what's so special about heights 0 and 1 that we mention them here. Then I finally noticed you added the versions!

upbqdn (dismissed):
Thank you very much for all the fixes and clarifications.

zebra-utils/src/bin/zebra-checkpoints/main.rs
teor2345: Yeah it was a confusing way of saying it, because "height" was meant to stand for an actual height, then the second parameter is the verbosity of the RPC. (And it's optional.)

zebra-utils/src/bin/zebra-checkpoints/main.rs
```diff
@@ -232,15 +239,19 @@ async fn main() -> Result<()> {
...
```
teor2345:
```suggestion
                // TODO: is it faster to call both `getblock height verbosity=0`
                //       and `getblock height verbosity=1`, rather than deserializing the block
```

upbqdn: Oh, it's actually verbosity. Sorry, I missed it as well.

upbqdn (approved):

PR #6585, opened 2023-04-27
change(mempool): Refactor the ZIP-317 minimum fee rate calculation to use usize

teor2345:
## Motivation

@oxarbitrage asked for some help refactoring this.

### Specifications

See the links in the code.

## Solution

- Follow the `zcashd` code, but make it simpler
- Use `usize` for all the calculations

This might break the tests, I'll let you update them all at once in whatever way you'd like, to avoid merge issues.

## Review

Alfredo asked for this PR.

Feel free to merge or copy-paste.

### DIFF
--- a/zebra-chain/src/transaction/unmined/zip317.rs
+++ b/zebra-chain/src/transaction/unmined/zip317.rs
@@ -45,6 +45,23 @@ const MIN_BLOCK_PRODUCTION_SUBSTITUTE_FEE: i64 = 1;
 /// `block_unpaid_action_limit` in ZIP-317.
 pub const BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT: u32 = 50;
 
+/// The minimum fee per kilobyte for Zebra mempool transactions.
+/// Also used as the minimum fee for a mempool transaction.
+///
+/// Based on `DEFAULT_MIN_RELAY_TX_FEE` in `zcashd`:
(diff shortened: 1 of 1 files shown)


zebra-chain/src/transaction/unmined/zip317.rs
```diff
@@ -184,11 +201,24 @@ pub fn mempool_checks(
...
+    // and usize is at least 4 GB.
+    assert!(
+        MIN_MEMPOOL_TX_FEE_RATE
+            < usize::MAX / usize::try_from(MAX_BLOCK_BYTES).expect("constant fits in usize"),
+        "the fee rate multiplication must never overflow",
     );
-    if miner_fee < Amount::<NonNegative>::try_from(limit).expect("limit value is invalid") {
+
+    let min_fee = (MIN_MEMPOOL_TX_FEE_RATE * transaction_size / KILOBYTE)
+        .clamp(MIN_MEMPOOL_TX_FEE_RATE, MEMPOOL_TX_FEE_REQUIREMENT_CAP);
```
teor2345, line 215: It seemed easier to do this calculation in `usize` then convert to `Amount`, which is what `zcashd` does.

Otherwise every calculation has to be checked for overflow, even though we clamp to `[100, 1000]` at the end anyway.

oxarbitrage (approved):
Thanks!

PR #6601, opened 2023-05-03
feat(zebra-network): add user agent argument

### LINKED ISSUE #2375: Allow different `zebra-network` apps to use different user agents
## Motivation

1. We want `zebrad` to have a different user agent from the ECC's DNS seeder, and other apps that use `zebra-network`

2. We'd like to simplify Zebra's release process, by automatically using a `zebrad` version in the zebra-network user agent.

### Specifications

https://github.com/bitcoin/bips/blob/master/bip-0014.mediawiki

## Solution

- [ ] Add a `user_agent` argument to `zebra_network::init`, and pass it to the handshaker
- [ ] In `zebrad::commands::start`, pass `application::app_version()` to `zebra_network::init`
  - Use the BIP-14 format: "/Zebra:1.0.0-alpha.11/"
- [ ] Add the supplied user agent to the front of the `zebra-network` user agent string, using the BIP-14 syntax

https://github.com/ZcashFoundation/zebra/blob/c40cbee42f0897932b03ba9fbb53bf6219e4c5f8/zebra-network/src/isolated.rs#L43-L59

## Related Work

Make sure custom user agents conform to BIP-14 #794

oxarbitrage:
## Motivation

We want to give `zebra-network` the possibility of changing the user agent it will use and we also want Zebra to automatically build the user agent by using git data.

Close https://github.com/ZcashFoundation/zebra/issues/2375

## Solution

- From `zebrad`, build and pass valid user agent to network.
- Build the user agent too in `zebra-rpc` as it is needed in the `getinfo` rpc method. We have `app_version` data already here, was needed for the same `getinfo` function, so we have everything to build user agent there too.

## Review

I think anyone can review, this reduce the number of steps from the release checklist. 

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/initialize/tests/vectors.rs
+++ b/zebra-network/src/peer_set/initialize/tests/vectors.rs
@@ -1245,7 +1245,7 @@ async fn remnant_nonces_from_outbound_connections_are_limited() {
     let hs = peer::Handshake::builder()
         .with_config(config.clone())
         .with_inbound_service(nil_inbound_service)
-        .with_user_agent(crate::constants::USER_AGENT.to_string())
+        .with_user_agent(("Test user agent").to_string())
         .with_latest_chain_tip(NoChainTip)
         .want_transactions(true)
         .finish()
@@ -1351,7 +1351,12 @@ async fn add_initial_peers_deadlock() {
 
     let nil_inbound_service = service_fn(|_| async { Ok(Response::Nil) });
 
-    let init_future = init(config, nil_inbound_service, NoChainTip);
+    let init_future = init(
+        config,
+        nil_inbound_service,
+        NoChainTip,
+        "Test user agent".to_string(),
+    );
 
     assert!(tokio::time::timeout(TIME_LIMIT, init_future).await.is_ok());
 }
@@ -1372,7 +1377,13 @@ async fn local_listener_port_with(listen_addr: SocketAddr, network: Network) {
     let inbound_service =
         service_fn(|_| async { unreachable!("inbound service should never be called") });
 
-    let (_peer_service, address_book) = init(config, inbound_service, NoChainTip).await;
+    let (_peer_service, address_book) = init(
+        config,
+        inbound_service,
+        NoChainTip,
+        "Test user agent".to_string(),
+    )
+    .await;
     let local_listener = address_book.lock().unwrap().local_listener_meta_addr();
 
     if listen_addr.port() == 0 {
@@ -1429,7 +1440,13 @@ where
         ..default_config
     };
 
-    let (_peer_service, address_book) = init(config, inbound_service, NoChainTip).await;
+    let (_peer_service, address_book) = init(
+        config,
+        inbound_service,
+        NoChainTip,
+        "Test user agent".to_string(),
+    )
+    .await;
 
     address_book
 }
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -28,7 +28,6 @@ use zebra_chain::{
     transaction::{self, SerializedTransaction, Transaction, UnminedTx},
     transparent::{self, Address},
 };
-use zebra_network::constants::USER_AGENT;
 use zebra_node_services::mempool;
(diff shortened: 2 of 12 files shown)


teor2345:
I just want to say that I love deleting stuff from the release checklist 🎉

teor2345:
Just letting you know that PR #6606 changes the format of the `git` version string in some cases. I don't know if that will impact this PR.

If you'd like, you can add:
```
Depends-On: #6601
```
to the PR #6606 description, which will make my PR wait for this one to merge. Then if anything breaks I can fix it.

(Just so you know, I'm not planning on reviewing this PR, because Marek is assigned to it.)

zebra-rpc/src/methods.rs
```diff
@@ -365,10 +364,25 @@ where
     State::Future: Send,
     Tip: ChainTip + Clone + Send + Sync + 'static,
 {
+    #[allow(clippy::unwrap_in_result)]
     fn get_info(&self) -> Result<GetInfo> {
+        // Build a [BIP 14] valie user agent with release info.
```

```suggestion
        // Build a [BIP 14] valid user agent with release info.
```

zebra-rpc/src/methods.rs
```diff
@@ -365,10 +364,25 @@ where
...
+        //
+        // [BIP 14]: https://github.com/bitcoin/bips/blob/master/bip-0014.mediawiki
+        let release_version = self
+            .app_version
+            // remove everything after the `+` character if any
+            .split('+')
+            .next()
+            .expect("always at least 1 slice")
+            // remove the previously added `v` character at start, this do not form part of the user agent.
```

```suggestion
            // remove the previously added `v` character at the start since it's not a part of the user agent.
```

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1245,7 +1245,7 @@ async fn remnant_nonces_from_outbound_connections_are_limited() {
     let hs = peer::Handshake::builder()
         .with_config(config.clone())
         .with_inbound_service(nil_inbound_service)
-        .with_user_agent(crate::constants::USER_AGENT.to_string())
+        .with_user_agent(("Test user agent").to_string())
```
upbqdn, line 1248: Why don't we go with the real user agent?

zebra-network/src/peer_set/initialize/tests/vectors.rs
```diff
@@ -1351,7 +1351,12 @@ async fn add_initial_peers_deadlock() {
 
     let nil_inbound_service = service_fn(|_| async { Ok(Response::Nil) });
 
-    let init_future = init(config, nil_inbound_service, NoChainTip);
+    let init_future = init(
+        config,
+        nil_inbound_service,
+        NoChainTip,
+        "Test user agent".to_string(),
```
upbqdn, line 1358: Same question as above: why not the actual user agent?

upbqdn (dismissed):
Looks great. I left two optional suggestions and one question.

zebra-network/src/peer_set/initialize/tests/vectors.rs
oxarbitrage: We don't have a "real user agent" here if that means a Zebra user agent. Zebra user agent is built in the `zebrad` crate and passed to `zebra-network`. Then this user agent is used https://github.com/ZcashFoundation/zebra/blob/8615dd750f28cb2b6d1956e022f26e83067ec011/zebra-network/src/peer_set/initialize.rs#L135
Here the test is starting a new handshake so it does not have any user agent being pass, so that string was added.

oxarbitrage:
Thanks! I added your doc suggestions and i tried to reply the question.

zebra-network/src/peer_set/initialize/tests/vectors.rs
upbqdn: Ah, of course. I should have thought of that. Thanks for the clarification.

upbqdn (approved):

zebra-network/src/peer_set/initialize/tests/vectors.rs
upbqdn: Explained in https://github.com/ZcashFoundation/zebra/pull/6601#discussion_r1185440905.

PR #6604, opened 2023-05-03
fix: Avoid wrong test results for modified JoinSplits

### LINKED ISSUE #5823: transaction::tests::v4_with_modified_joinsplit_is_rejected fails with unexpected verifier error

upbqdn:
## Motivation

Issue #5823, reports that we occasionaly get the following error when running the test `transaction::tests::v4_with_modified_joinsplit_is_rejected`:

> ---- transaction::tests::v4_with_modified_joinsplit_is_rejected stdout ----
> 
>   left: `Err(RedJubjub(InvalidSignature))`,
>  right: `Err(Groth16("proof verification failed"))`
> Location: zebra-consensus/src/transaction/tests.rs:1578

The problem is that modifying the JoinSplit data invalidates the tx signatures. The signatures are checked concurrently with the ZK proofs, and when a signature check finishes before the proof check, the verifier reports an invalid signature instead of invalid proof.

Note that the invalid sig is for Sapling since the sig is on RedJubjub even though we modify a JoinSplit data in the test. The reason is that Sapling binding sigs use a sighash that covers JoinSplits in the tx as well. 

## Solution

Mocking the signatures so that the verifier accepts them would be one way of solving the issue. However, since we're using transactions to which we don't have the signing keys, recomputing the sigs is infeasible. [§4.13](https://zips.z.cash/protocol/protocol.pdf#saplingbalance) in the spec contains the explanation.

The race condition described above happens only occasionally, so the solution I chose is to run the verifier in a loop with a small iteration threshold until it returns the correct error. I tried running the verifier ten times in a row (without exiting the loop any earlier), and the test took less than two seconds. The test will, therefore, run for two seconds in the worst case and fail afterwards, which should happen with low probability.

This PR also does a few minor refactors along the way.

Closes #5823.

### Reviewer Checklist


zebra-consensus/src/transaction/tests.rs
```diff
@@ -2016,30 +2016,52 @@ async fn v4_with_joinsplit_is_rejected_for_modification(
...
-        .await;
+    // Test the transaction verifier.
+    //
+    // Note that modifying the JoinSplit data invalidates the tx signatures. The signatures are
+    // checked concurrently with the ZK proofs, and when a signature check finishes before the proof
+    // check, the verifier reports an invalid signature instead of invalid proof. This race
+    // condition happens only occasionaly, so we run the verifier in a loop with a small iteration
+    // threshold until it returns the correct error.
+    let mut i = 1;
+    let result = loop {
```
upbqdn, line 2046: I use a `loop` here to avoid manually initializing the `result` since I didn't really know how to initialize it. I didn't find any better way.

teor2345 (approved):
Looks great, thanks for this fix!

oxarbitrage:
@mergifyio update

Issue #3320, opened 2022-01-04
Stop logging IP addresses and mempool transaction IDs by default, to protect user privacy


teor2345:
## Motivation

Zebra collects, logs, and has metrics on two kinds of private information:
* IP addresses of Zcash nodes, which can be linked to organisations or individuals, and
* transaction IDs of mempool transactions, because they can reveal the origin of a transaction.

Zebra's logs also attach detailed timestamps to this information.

Zebra should avoid logging and metrics containing this private information by default.

### Suggested Solution

Stop logging and including in metrics by default:
- `SocketAddr`s
- `UnminedTxId`s

## Suggested Design

Modify the `Debug` and `Display` impls of `SocketAddr` and  `UnminedTxId` to just log the type name by default. This would need a `PrivateSocketAddr` wrapper with a `Deref` impl.

This should work for logging and metrics.

## Follow Up Work

Add an IP address logging mode. If we add this config, then users can be compelled to collect this information. So we might want to put it behind a feature that's off by default. Or we might not want to expose this information at all.

ftm1000:
redistributing issues that can be separately worked from Epic #2311

ftm1000:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/3320?planning-poker) with ZenHub @gustavovalverde @dconnolly @conradoplg @jvff @oxarbitrage @teor2345 @upbqdn

teor2345:
This isn't somethign we need to fix any time soon.

teor2345:
@mpguerra Dodger suggested that we consider this privacy risk for the `zebra-network` part of the DNS seeder. We might also want to consider it for the Zebra stable release.

PR #6616, opened 2023-05-05
fix(log): Stop logging potentially sensitive user information from unmined transactions

teor2345:
## Motivation

We don't want to log sensitive user information like transaction IDs and details.

This is the transaction part of #3320.

## Solution

Change the following Debug and Display impls so they don't log any transaction info:
- UnminedTxId
- UnminedTx
- VerifiedUnminedTx

## Review

Anyone can review this, it's not a release blocker.

### Reviewer Checklist

## Follow Up Work

Check for logging or metrics of individual fields from mempool/RPC transactions:
* transaction IDs
* sensitive transaction info

Consider modifying the `WtxId` type in the same way.

### DIFF
--- a/zebra-chain/src/transaction/unmined.rs
+++ b/zebra-chain/src/transaction/unmined.rs
@@ -86,7 +86,7 @@ const MEMPOOL_TRANSACTION_LOW_FEE_PENALTY: u64 = 40_000;
 /// [ZIP-239]: https://zips.z.cash/zip-0239
 /// [ZIP-244]: https://zips.z.cash/zip-0244
 /// [Spec: Transaction Identifiers]: https://zips.z.cash/protocol/protocol.pdf#txnidentifiers
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 pub enum UnminedTxId {
     /// A legacy unmined transaction identifier.
@@ -104,6 +104,24 @@ pub enum UnminedTxId {
     ///
     /// For more details, see [`WtxId`].
     Witnessed(WtxId),
+}
+
+impl fmt::Debug for UnminedTxId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Legacy(hash) => f.debug_tuple("Legacy").field(hash).finish(),
+            Self::Witnessed(id) => f.debug_tuple("Witnessed").field(id).finish(),
+        }
+    }
+}
+
+impl fmt::Display for UnminedTxId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Legacy(hash) => hash.fmt(f),
+            Witnessed(id) => id.fmt(f),
+        }
+    }
 }
 
 impl From<Transaction> for UnminedTxId {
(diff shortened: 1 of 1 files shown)


zebra-chain/src/transaction/unmined.rs
```diff
@@ -106,6 +106,29 @@ pub enum UnminedTxId {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Legacy(_hash) => f
+                .debug_tuple("transaction::Hash")
+                .field(&"private")
+                .finish(),
+            Witnessed(_id) => f.debug_tuple("WtxId").field(&"private").finish(),
+        }
+    }
+}
```
dconnolly, line 130: 👍

zebra-chain/src/transaction/unmined.rs
```diff
@@ -106,6 +106,29 @@ pub enum UnminedTxId {
...
+
+impl fmt::Display for UnminedTxId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Legacy(_hash) => f
+                .debug_tuple("transaction::Hash")
+                .field(&"private")
+                .finish(),
+            Witnessed(_id) => f.debug_tuple("WtxId").field(&"private").finish(),
```
dconnolly, line 127: Debug writes "Legacy", "Witnessed"; Display writes "transaction::Hash" and "WtxId"; should we align these?

zebra-chain/src/transaction/unmined.rs
```diff
@@ -334,14 +354,20 @@ pub struct VerifiedUnminedTx {
     pub fee_weight_ratio: f32,
 }
 
+impl fmt::Debug for VerifiedUnminedTx {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Logging unmined transactions can leak sensitive user information,
+        // particularly when Zebra is being used as a `lightwalletd` backend.
+        f.debug_tuple("VerifiedUnminedTx")
+            .field(&"private")
```
dconnolly, line 362: Should this `Debug` be private too?

zebra-chain/src/transaction/unmined.rs
```diff
@@ -230,13 +244,17 @@ pub struct UnminedTx {
     pub conventional_fee: Amount<NonNegative>,
 }
 
+impl fmt::Debug for UnminedTx {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Logging unmined transactions can leak sensitive user information,
+        // particularly when Zebra is being used as a `lightwalletd` backend.
+        f.debug_tuple("UnminedTx").field(&"private").finish()
```
dconnolly, line 251: Should this call `self.to_string()` like `Unmined` does above? Seems like a nice pattern, but duplicating 'private' everywhere seems easy to maintain

dconnolly (commented):
some q's

zebra-chain/src/transaction/unmined.rs
teor2345: I'm not too concerned about the difference here, happy if you want to make another PR to fix it up.

zebra-chain/src/transaction/unmined.rs
teor2345: This debug is private, did you mean to tag a different part of the code?

zebra-chain/src/transaction/unmined.rs
teor2345: I'm not too concerned about the difference here, happy if you want to make another PR to fix it up.

zebra-chain/src/transaction/unmined.rs
teor2345: @dconnolly ?

dconnolly: Huh, this may have been erroneous; nevermind!

dconnolly (approved):

PR #6624, opened 2023-05-07
change(deps): Upgrade zcash_script to 0.1.12 to match zcashd 5.5.0

### LINKED ISSUE #6535: Upgrade `zcash_script` after the `zcashd` 5.5.0 release
## Motivation

After `zcashd` 5.5.0 is released:
https://github.com/zcash/zcash/tags

We want to update our script engine to match its implementation. There haven't been any significant changes to script behaviour, but the surrounding code has been refactored.

## Scheduling

This doesn't need to be in the same release as #6232, but builds will be slower until we update our entire set of dependencies. There's also a small but acceptable risk of consensus divergence.

### Specifications

Zcash transparent scripts are mostly unspecified, so we must match `zcashd`'s behaviour.

This is consensus-critical code.

### Testing

Our existing unit and integration tests have good coverage of this code.

Manually use `cargo tree` to check that we don't depend on any vulnerable `orchard` versions: we must only depend on orchard 0.4 or later.

## Related Work

Follow up to:
- #6232

teor2345:
## Motivation

We need to upgrade to the `zcash_script` dependency to match `zcashd` 5.5.0.

Closes #6535

### Specifications

Zcash script handling is not specified, so we just run `zcashd`'s code.

This is unlikely to be a consensus-critical change, because `zcashd` has not changed its script handling. However, it has refactored some of its transaction parsing code.

## Solution

- Run [cargo upgrade --workspace zcash_script](https://github.com/ZcashFoundation/zebra/commit/491ef973646950338c80768aa6ea66473ccdf2ca)
- Automatically [update Cargo.lock](https://github.com/ZcashFoundation/zebra/commit/c0cb0541f351cc38c7cefc7c530ce77b839ed4aa)
- Manually [update deny.toml](https://github.com/ZcashFoundation/zebra/commit/de714a9df9ec33bc52831a53b38f11c8beb500ce)

## Review

This is a high priority because it's blocking the missed version updates, which is needed to tag the next release.

### Reviewer Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -48,9 +48,8 @@ skip-tree = [
 
     # wait for zcashd and zcash_script to upgrade
     # https://github.com/ZcashFoundation/zcash_script/pulls
-    { name = "zcash_primitives", version = "=0.9.1" },
+    { name = "metrics", version = "=0.20.1" },
     { name = "sha2", version = "=0.9.9" },
-    { name = "reddsa", version = "=0.3.0" },
 
     # wait for ed25519-zebra, indexmap, metrics-util, and metrics to upgrade
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
@@ -58,15 +57,12 @@ skip-tree = [
 
     # ECC crates
 
-    # Wait until `orchard` updates `aes`, which depends on `cipher`
-    { name = "cipher", version = "=0.3.0" },
-
     # wait for zcash_primitives to remove duplicated dependencies
     { name = "block-buffer", version = "=0.9.0" },
 
     # zebra-utils dependencies
 
-    # wait for structopt upgrade (or upgrade to clap 3)
+    # wait for structopt upgrade (or upgrade to clap 4)
     { name = "clap", version = "=2.34.0" },
     { name = "heck", version = "=0.3.3" },
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,24 +71,12 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.7.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
-dependencies = [
- "cfg-if 1.0.0",
- "cipher 0.3.0",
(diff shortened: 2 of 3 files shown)


deny.toml
```diff
@@ -48,25 +48,21 @@ skip-tree = [
 
     # wait for zcashd and zcash_script to upgrade
     # https://github.com/ZcashFoundation/zcash_script/pulls
```
dconnolly, line 50: Should we remove these lines now?

dconnolly (approved):

deny.toml
teor2345: Nope, we still need `zcashd` to upgrade, unfortunately!

daira: https://github.com/zcash/zcash/issues/6648

deny.toml
```diff
@@ -48,25 +48,21 @@ skip-tree = [
...
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
     # ECC crates
 
-    # Wait until `orchard` updates `aes`, which depends on `cipher`
-    { name = "cipher", version = "=0.3.0" },
-
     # wait for zcash_primitives to remove duplicated dependencies
     { name = "block-buffer", version = "=0.9.0" },
```
daira, line 61: This dependency is via `sha2 0.9.9` so it should go away with https://github.com/zcash/zcash/issues/6648 .

daira (approved):
utACK

deny.toml
mpguerra: @teor2345 should we wait for this one before merging this PR?

mpguerra: actually I guess not since this will be in zcash 5.6.0 and I don't think we want to wait that long

teor2345: Duplicate dependencies usually aren't a blocker, unless they're consensus-critical and we're actually calling both.

teor2345:
@mergify refresh

teor2345:
@mergifyio refresh

teor2345:
@mergifyio refresh

teor2345:
> Manually use cargo tree to check that we don't depend on any vulnerable orchard versions: we must only depend on orchard 0.4 or later.

I forgot this check when making this PR. `orchard` was upgraded in that PR, and some other missed updates are handled by PR #6656.

Issue #6168, opened 2023-02-15
actions: keep scripts out of the workflow scope


gustavovalverde:
## Motivation

Some workflows are script intensive, making the workflows unreadable.

### Specifications

A good approach to handle this would be moving all the scripts to multiple files outside the workflow, or create a GitHub Action to execute them in a reusable manner.

### Complex Code or Requirements

- Define a place to hold all the script files
- Keep linting working on these files
- Use a name convention for script files

teor2345:
The other advantage of this is that it's easier to test the scripts locally outside the workflows.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/6168?planning-poker) with Zenhub @arya2 @gustavovalverde @oxarbitrage @teor2345 @upbqdn

teor2345:
This is potentially a very large task, depending on how much work we put into de-duplicating or parameterising our scripts.

What's the intended scope of this ticket? Is it just moving the existing scripts into files? Or will there be a lot of refactoring?

(CI refactoring is risky and often breaks things, because it's difficult to test before merging.)

gustavovalverde:
> Is it just moving the existing scripts into files? Or will there be a lot of refactoring?

It's basically moving scripts into `.sh` files, but just doing that can be considered _a lot_ of refactoring.

teor2345:
> > Is it just moving the existing scripts into files? Or will there be a lot of refactoring?
> 
> It's basically moving scripts into `.sh` files, but just doing that can be considered _a lot_ of refactoring.

I guess what I meant was: are we just moving the script text into files and calling those files? Or are we modifying the functionality of the scripts at the same time?

I think it would reduce the risk and scope of this ticket to restrict it to copy-pasting scripts, and also do it in stages. Maybe we could try it with a small workflow or a single script first, to check if there are any issues?

gustavovalverde:
Yeah, it should basically be copy-pasting, and we can surely test with the less criticals first. I've anyways remove this from the Sprint backlog as we have these dependencies:
- #5572
- #6167

teor2345:
In PR #8132 I reverted this PR because it has been causing internal GitHub errors. Unfortunately no-one picked up on this during the review process or after it merged.

teor2345:
[...]

PR #6628, opened 2023-05-08
refactor(ci): migrate from abandoned `actions-rs/*`

### LINKED ISSUE #6048: ci: migrate actions using the unmaintained `actions-rs/*` repository
## Motivation

We might have to stop using `actions-rs/toolchain` and `actions-rs/cargo` as those actions are no longer maintained, and might break sooner or later based on existing warnings:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1, actions-rs/cargo@v1.0.3
```

## Related Work

https://github.com/ZcashFoundation/zebra/pull/6039

gustavovalverde:
## Motivation

All Rust related repositories at `actions-rs/*` are no longer being maintained

Fixes: #6048 

## Solution

- Do not depend on external actions if we have a way to do it manually with little verbosity
- Setup Rust and Cargo using their official documentation https://doc.rust-lang.org/cargo/getting-started/installation.html
- Install Toolchains, Components and Profiles with manual instructions: https://rust-lang.github.io/rustup/concepts/toolchains.html

## Review

### Reviewer Checklist

### DIFF
--- a/.github/workflows/build-crates-individually.patch.yml
+++ b/.github/workflows/build-crates-individually.patch.yml
@@ -25,13 +25,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3.5.2
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
+      # Setup Rust with stable toolchain and minimal profile
+      - name: Setup Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=stable --profile=minimal
 
-      - uses: actions-rs/cargo@v1.0.3
       # This step is meant to dynamically create a JSON containing the values of each crate
       # available in this repo in the root directory. We use `cargo tree` to accomplish this task.
       #
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -52,13 +52,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3.5.2
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
+      # Setup Rust with stable toolchain and minimal profile
+      - name: Setup Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=stable --profile=minimal
 
-      - uses: actions-rs/cargo@v1.0.3
       # This step is meant to dynamically create a JSON containing the values of each crate
       # available in this repo in the root directory. We use `cargo tree` to accomplish this task.
       #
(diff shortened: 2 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -81,27 +81,17 @@ jobs:
...
-          sharedKey: "clippy-cargo-lock"
-
-      - name: Run clippy action to produce annotations
-        uses: actions-rs/clippy-check@v1.0.7
-        if: ${{ steps.check_permissions.outputs.has-permission }}
-        with:
-          # GitHub displays the clippy job and its results as separate entries
-          name: Clippy (stable) Results
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --all-targets -- -D warnings
```
teor2345: Please don't delete these clippy annotations, they are really useful for developers and reviewers to see where the clippy problems are in the GitHub diff.

.github/workflows/lint.yml
```diff
@@ -124,20 +114,18 @@ jobs:
...
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: actions-rs/toolchain@v1.0.6
-        with:
-          toolchain: stable
-          components: rustfmt
-          override: true
+      # Setup Rust with stable toolchain and default profile
+      - name: Setup Rust
+        if: ${{ !steps.check_permissions.outputs.has-permission }}
```
teor2345: This job now needs to depend on the check permissions job.

.github/workflows/lint.yml
```diff
@@ -165,16 +153,14 @@ jobs:
...
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: actions-rs/toolchain@v1.0.6
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
+      # Setup Rust with stable toolchain and default profile
+      - name: Setup Rust
+        if: ${{ !steps.check_permissions.outputs.has-permission }}
```
teor2345: This job now needs to depend on the check permissions job.

.github/workflows/build-crates-individually.patch.yml
```diff
@@ -25,13 +25,11 @@ jobs:
...
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
+      # Setup Rust with stable toolchain and minimal profile
+      - name: Setup Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=stable --profile=minimal
```
teor2345, line 31: ~~What is the security impact of this change to using a bare URL? Did the action used to verify the downloaded binaries?~~

teor2345: ~~What's the CI performance impact of this change?~~

teor2345: > What is the security impact of this change to using a bare URL? Did the action used to verify the downloaded binaries?

It looks like this is exactly what the maintained action does:
https://github.com/dtolnay/rust-toolchain/blob/b44cb146d03e8d870c57ab64b80f04586349ca5d/action.yml#L65

That action also applies some additional environmental variables, I've added them to the Zebra repository settings:
- CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
- CARGO_TERM_COLOR=always

This should speed up builds and show output nicely.

.github/workflows/build-crates-individually.yml
```diff
@@ -52,13 +52,11 @@ jobs:
...
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
+      # Setup Rust with stable toolchain and minimal profile
+      - name: Setup Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=stable --profile=minimal
```
teor2345, line 58: Optional: We repeat this same script a lot, should it be a shell script? Or would that be more complicated?

I'm concerned that these duplicate scripts might diverge over time, because we might miss some when we do fixes or changes. The action was only a name and a version, so it was easier to keep in sync.

.github/workflows/lint.yml
teor2345: It seems like the top two forks here are maintained:
https://github.com/actions-rs/clippy-check/forks

I think they're getting node updates, but please double-check.

teor2345 (changes requested):
I think removing the clippy annotations is a blocker. Can we please keep them?

teor2345:
What's the deadline for this? Have GitHub announced a final date when support for these actions will end?

.github/workflows/lint.yml
gustavovalverde: I'll leave the previous one as it's secure. Those are being updated with dependency bots, and some have changes that I can't deep dive into them right now to understand if the changes are "secure".

To be on the safe side, let's keep `actions-rs/clippy-check` and I'll keep this task as we have time (until September 27th) to make a better decision for this specific action

.github/workflows/build-crates-individually.yml
gustavovalverde: Yes, I'll do this globally with: 
- https://github.com/ZcashFoundation/zebra/issues/6168

gustavovalverde:
```
I think removing the clippy annotations is a blocker. Can we please keep them?
```

I kept it, as this is a blocker and we have time to find a better replacement.


```
What's the deadline for this? Have GitHub announced a final date when support for these actions will end?
```
September 27th

teor2345 (approved):
Looks good.

I reduced the priority to "medium", because the deadline is still 4 months away.

teor2345:
@gustavovalverde is this PR causing internal errors in GitHub actions:
> GitHub Actions has encountered an internal error when running your job.

https://github.com/ZcashFoundation/zebra/actions/runs/4940550408?pr=6657

Or is that happening because of the current GitHub infrastructure failures:
https://www.githubstatus.com/

PR #6634, opened 2023-05-08
fix(build): sync LWD builder with our `build-docker-image.yml` workflow

gustavovalverde:
## Motivation

After migrating to our new GCP project, some references were not being applied correctly, as this workflow was not referencing new resources the right way. It was even outdated on some specific parts.
## Solution

Sync this build steps with the ones we're using at https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml

## Review

Anyone from the DevOps team

### Reviewer Checklist


teor2345 (approved):
Thank you for the quick fix!

PR #6643, opened 2023-05-09
fix(deployment): explicitly use a service account and region

gustavovalverde:
## Motivation

Some deployments were using default networks, which we no longer allow on our deployments.

## Solution

- Adding the network zone specified at our `${{ vars.GCP_ZONE }}` variable
- Also adding `${{ vars.GCP_DEPLOYMENTS_SA }}` to avoid future issues

## Review

Anyone from the DevOps team

### Reviewer Checklist


teor2345 (approved):
Thanks!

Issue #3665, opened 2022-02-28
GitHub Actions cache fails, causing "Build images" job to fail


teor2345:
## Motivation

I keep seeing CI failures in "Build images" when running full sync tests, but they don't happen on every run.

> Error: buildx failed with: error: failed to solve: Get "https://artifactcache.actions.githubusercontent.com/iLo8c1TT59QRZKeUa4bO5aBdMPTRIlKlqwBwOSus1iy4D4L9OS/_apis/artifactcache/cache?keys=buildkit-blob-1-sha256%3A0c6b8ff8c37e92eb1ca65ed8917e818927d5bf318b6f18896049b5d9afc28343&version=693bb7016429d80366022f036f84856888c9f13e00145f5f6f4dce303a38d6f2": dial tcp 13.107.42.16:443: i/o timeout

https://github.com/ZcashFoundation/zebra/runs/5353953645?check_suite_focus=true#step:7:2528

ftm1000:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/3665?planning-poker) with ZenHub @conradoplg @dconnolly @gustavovalverde @jvff @oxarbitrage @teor2345 @upbqdn

teor2345:
Also happened in PR #3668:
> Error: buildx failed with: error: failed to solve: error writing layer blob: Patch "https://artifactcache.actions.githubusercontent.com/iLo8c1TT59QRZKeUa4bO5aBdMPTRIlKlqwBwOSus1iy4D4L9OS/_apis/artifactcache/caches/19616": read tcp 10.1.0.82:52776->13.107.43.16:443: read: connection reset by peer


https://github.com/ZcashFoundation/zebra/runs/5364814787?check_suite_focus=true#step:7:1797

teor2345:
Also happened in PR #3641, the error is slightly different:
> Error: buildx failed with: error: failed to solve: failed commit on ref "manifest-sha256:446e770f6ae99a4d1cd9de5e5916e8057ecca7d11ad6f4bc69499a0f4963bda9": cannot reuse body, request must be retried

teor2345:
Probably also failed on #3647:
https://github.com/ZcashFoundation/zebra/runs/5383367400?check_suite_focus=true

teor2345:
Failed in PR #3691:
[...]

PR #6645, opened 2023-05-09
ci: unpin specific `buildkit` version and fix deployment subnets

gustavovalverde:
## Motivation

We previously had an issue with the following error: `cannot reuse body, request must be retried` and it's happening once again, but it might be for a different reason.

This commonly was a wrong error, caused by a containerd issue which has being tracked and solved here: https://github.com/docker/build-push-action/issues/761#issuecomment-1406261692

We had this previously reported in: https://github.com/ZcashFoundation/zebra/issues/3665

This PR also fixes the following error:

```
Invalid value for field 'resource.networkInterfaces[0].network': 'https://compute.googleapis.com/compute/v1/projects/zfnd-dev-zebra/global/networks/default'. The referenced network resource cannot be found.
```

### Specifications

We're having errors when building, and this might be caused by an underlying error which containerd is not showing us correctly.

## Solution

- Unpin our old `buildkit` version and use the latest with included fixes
- GCP VMs not only need a specific region set, but also a subnetwork, as we don't have default subnetworks available (those have to be explicitly created)

## Review

Anyone from DevOps team

### Reviewer Checklist

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -110,12 +110,6 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v2
-        with:
-          # TODO: change after new buildkit version gets fixed
-          # https://github.com/moby/buildkit/issues/3347
-          # https://github.com/docker/build-push-action/issues/761#issuecomment-1383822381
-          driver-opts: |
-            image=moby/buildkit:v0.10.6
 
       - name: Authenticate to Google Cloud
         id: auth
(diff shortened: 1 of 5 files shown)


teor2345 (dismissed):
Thank you for this bug fix, it's been a bit annoying!

gustavovalverde:
@teor2345 I had to fix another issue which was not ultimately solved by https://github.com/ZcashFoundation/zebra/commit/b3691c514a7a7ef8d13928436029f4b62d70facd 

The VMs which did not start by the error are now running.

teor2345 (approved):

PR #6659, opened 2023-05-10
fix(ci): Allow testnet syncs to take longer than 6 hours

teor2345:
## Motivation

Testnet syncs can be as short as 2 hours, but also take over 6 hours, depending on which testnet nodes are running at the time.

## Solution

Allow this job to take longer than 6 hours.

## Review

This is a fix for a CI failure on `main`, so it's urgent. It doesn't need to be documented in the CHANGELOG.

### Reviewer Checklist


gustavovalverde (approved):
LGTM

PR #6660, opened 2023-05-10
fix(ci): Run all the duplicate dependency checks, even if one fails

teor2345:
## Motivation

Sometimes we need to see the output of a specific duplicate dependency job to fix a dependency update PR.

### Specifications

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

## Solution

Run all the duplicate dependency jobs to completion, even if one fails.

These jobs are quick, so this won't change CI run time.

## Review

This is a fix that is needed to fix other open PRs. It doesn't need to go in the changelog.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -250,6 +250,8 @@ jobs:
           - bans
           - sources
         features: ['', '--all-features', '--no-default-features']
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
LGTM!

dconnolly (approved):

PR #6632, opened 2023-05-08
v1.0.0-rc.8 release

oxarbitrage:
---
name: Release Checklist Template
about: Checklist of versioning to create a taggable commit for Zebra
title: ''
labels:
assignees: ''

---

### How to Increment Versions

Zebra follows [semantic versioning](https://semver.org). Semantic versions look like: MAJOR`.`MINOR`.`PATCH[`-`TAG`.`PRE-RELEASE]

The [draft `zebrad` changelog](https://github.com/ZcashFoundation/zebra/releases) will have an automatic version bump. This version is based on [the labels on the PRs in the release](https://github.com/ZcashFoundation/zebra/blob/main/.github/release-drafter.yml).

Check that the automatic `zebrad` version increment matches the changes in the release:

<details>

If we're releasing a mainnet network upgrade, it is a `major` release:
1. Increment the `major` version of _*all*_ the Zebra crates.
2. Increment the `patch` version of the tower crates.

If we're not releasing a mainnet network upgrade, check for features, major changes, deprecations, and removals. If this release has any, it is a `minor` release:
1. Increment the `minor` version of `zebrad`.
2. Increment the `pre-release` version of the other crates.
3. Increment the `patch` version of the tower crates.

Otherwise, it is a `patch` release:
1. Increment the `patch` version of `zebrad`.
2. Increment the `pre-release` version of the other crates.
3. Increment the `patch` version of the tower crates.

Zebra's Rust API is not stable or supported, so we keep all the crates on the same beta `pre-release` version.

</details>

### Version Locations

Once you know which versions you want to increment, you can find them in the:

zebrad (rc):
- [x] zebrad `Cargo.toml`
- [x] `README.md`
- [x] `book/src/user/docker.md`

crates (beta):
- [x] zebra-* `Cargo.toml`s

tower (patch):
- [x] tower-* `Cargo.toml`s

auto-generated:
- [x] `Cargo.lock`: run `cargo build` after updating all the `Cargo.toml`s

#### Version Tooling

You can use `fastmod` to interactively find and replace versions.

For example, you can do something like:
```
fastmod --extensions rs,toml,md --fixed-strings '1.0.0-rc.0' '1.0.0-rc.1' zebrad README.md zebra-network/src/constants.rs book/src/user/docker.md
fastmod --extensions rs,toml,md --fixed-strings '1.0.0-beta.15' '1.0.0-beta.16' zebra-*
[...]

PR #6664, opened 2023-05-11
fix(ci): Stop running multiple full syncs on different branches

teor2345:
## Motivation

We are accidentally running about 8 simultaneous full syncs on testnet, which is slowing down testnet.

### Specifications

https://docs.github.com/en/actions/using-jobs/using-concurrency

https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

## Solution

Fix long job concurrency:
- Only run one checkpoint/full/lwd sync at a time, by removing the branch from the concurrency group label
- Allow a manual sync to run at the same time as automatic syncs, by adding the manual sync setting to the label

Fix cache-using job conditions:
- If cache generation jobs are skipped due to concurrency, and the cache is missing, also skip their dependent jobs, rather than failing them when the disk isn't found
- This is implemented by requiring that the disk was found in the cache, or the disk was previously generated in this workflow run by a successful job

Related fixes:
- Remove outdated concurrency and fix docs on the lightwalletd send transactions test, which now runs on every PR

## Review

This is urgent because we're putting a lot of load on testnet.

See my comment about manual testing of this fix below.

### Reviewer Checklist

  - [ ] Will the PR name make sense to developers?

## Follow Up Work

I'm going to cancel all the extra jobs now, and just leave the one in this PR, the release PR #6632, and the one that's about to finish.

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -314,7 +314,7 @@ jobs:
     #
     # See the concurrency comment on the zebrad test-full-sync job for details.
     concurrency:
-      group: ${{ github.workflow }}−${{ github.ref }}-regenerate-stateful-disks
+      group: ${{ github.workflow }}−manual-${{ fromJSON(github.event.inputs.regenerate-disks) }}-regenerate-stateful-disks
       cancel-in-progress: false
 
   # Test that Zebra syncs and fully validates a few thousand blocks from a cached mandatory checkpoint disk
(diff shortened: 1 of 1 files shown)


teor2345:
I did a manual test of these changes. In this workflow, the full testnet sync was skipped due to concurrency, and so the testnet checkpoint generation was also skipped:
https://github.com/ZcashFoundation/zebra/actions/runs/4944180938/jobs/8839453807

This is the correct behaviour. (Previously the checkpoint generation job would fail.)

gustavovalverde (approved):
I agree with this as a bugfix, but I have a "feeling" that we should move the testnet sync to the `main` branch, and just trigger it on PRs on-demand (using a label or comment).

teor2345:
> I agree with this as a bugfix, but I have a "feeling" that we should move the testnet sync to the `main` branch,

it is already just running on the main branch

> and just trigger it on PRs on-demand (using a label or comment).

it only runs on PRs when there is no cached state disk, or when manually triggered using workflow dispatch.

PR #6679, opened 2023-05-15
fix(security): Limit blocks and transactions sent in response to a single request 

teor2345:
## Motivation

Zebra doesn't limit how many blocks or transactions it will send in response to a `getdata` request.

Since `getdata` requests are small hashes, this can become a network amplification attack.

### Specifications

Network response limits are not specified, nodes are free to implement any behaviour.

This PR replaces complex concurrent iterator code with simpler code that is not concurrent.

These changes should be compatible with both Zebra's and `zcashd`'s block syncing and mempool behaviour.

## Solution

- Stop queuing additional block and transaction responses after 1 MB of data has been queued
- Only answer the first 16 requests in a block request (`zcashd` sends up to 16, Zebra always sends 1)

### Testing

There are existing tests for responses. We don't test the limits, but they seem pretty trivial in the code.

I'm happy to add tests if anyone wants.

## Review

This PR is a routine security fix.

### Reviewer Checklist

### DIFF
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -46,6 +46,15 @@ mod tests;
 
 use downloads::Downloads as BlockDownloads;
 
+/// The number of bytes the [`Inbound`] service will queue in response to a single block or
+/// transaction request, before ignoring any additional block or transaction IDs in that request.
+///
+/// This is the same as `zcashd`'s default send buffer limit:
+/// <https://github.com/zcash/zcash/blob/829dd94f9d253bb705f9e194f13cb8ca8e545e1e/src/net.h#L84>
+/// as used in `ProcessGetData()`:
+/// <https://github.com/zcash/zcash/blob/829dd94f9d253bb705f9e194f13cb8ca8e545e1e/src/main.cpp#L6410-L6412>
+pub const GETDATA_SENT_BYTES_LIMIT: usize = 1_000_000;
+
 type BlockDownloadPeerSet =
     Buffer<BoxService<zn::Request, zn::Response, zn::BoxError>, zn::Request>;
 type State = Buffer<BoxService<zs::Request, zs::Response, zs::BoxError>, zs::Request>;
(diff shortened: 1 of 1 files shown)


zebrad/src/components/inbound.rs
```diff
@@ -369,66 +392,97 @@ impl Service<zn::Request> for Inbound {
...
+                        //
+                        // If there was a database error, return the error,
+                        // and stop processing further chunks.
+                        match response {
+                            zs::Response::Block(Some(block)) => {
+                                // If checking the serialized size of the block performs badly,
+                                // return the size from the state using a wrapper type.
+                                total_size += block.zcash_serialized_size();
+
+                                blocks.push(Available(block))
```
arya2, line 427: Could we check that `blocks.is_empty()` before adding blocks past the limit here:

```suggestion
                                let new_total_size = total_size + block.zcash_serialized_size();

                                if new_total_size <= GETDATA_SENT_BYTES_LIMIT || blocks.is_empty() {
                                    total_size = new_total_size;
                                    blocks.push(Available(block));
                                };
```

zebrad/src/components/inbound.rs
```diff
@@ -46,6 +50,25 @@ mod tests;
...
 use downloads::Downloads as BlockDownloads;
 
+/// The number of bytes the [`Inbound`] service will queue in response to a single block or
+/// transaction request, before ignoring any additional block or transaction IDs in that request.
+///
+/// This is the same as `zcashd`'s default send buffer limit:
+/// <https://github.com/zcash/zcash/blob/829dd94f9d253bb705f9e194f13cb8ca8e545e1e/src/net.h#L84>
+/// as used in `ProcessGetData()`:
+/// <https://github.com/zcash/zcash/blob/829dd94f9d253bb705f9e194f13cb8ca8e545e1e/src/main.cpp#L6410-L6412>
+pub const GETDATA_SENT_BYTES_LIMIT: usize = 1_000_000;
```
arya2, line 60: 👍

It looked like the constant is 1000 in zcashd but it says elsewhere that that's in units of 1000 bytes.

arya2 (approved):
This is looking great, I left a suggestion for enforcing the limit once there's at least 1 block for the response.

zebrad/src/components/inbound.rs
teor2345: We've already done the work to retrieve this block from the database, so we might as well send it on the network.

The limit isn't a network protocol limit, it's just a performance/denial of service limit.

zebrad/src/components/inbound.rs
teor2345: Yep, it's multiplied in the code before use.

PR #6683, opened 2023-05-15
docs(book): Update Zebra book summary to include supported platforms, platform tier policy, and versioning

mpguerra:
## Motivation

In preparation for the stable release we want to make sure that information about supported platforms is readily available to zebra users.

This change includes:

- Adding some previously "hidden" user documentation to the zebra book menu
- Fixing some broken links
- Fixing links to point to the book rather than github
### Reviewer Checklist

## Follow Up Work

Review whether there is any other documentation (both user and developer targeted) that is not currently exposed in the zebra book and publish if relevant.


arya2 (approved):
Looks good!

PR #6686, opened 2023-05-15
fix(ci): Use valid names for concurrency groups

teor2345:
## Motivation

The concurrency group names aren't parsing correctly in CI, instead, they are silently failing.

### Specifications

https://docs.github.com/en/actions/learn-github-actions/expressions#format

## Solution

Compare the string to `true` rather than trying to convert a potentially empty string from JSON

## Review

This is an urgent fix.

### Reviewer Checklist


dconnolly (approved):

teor2345:
This worked, the full sync jobs are running now.

PR #6690, opened 2023-05-16
change(ci): Turn CI errors into PR annotations for most CI jobs

teor2345:
## Motivation

At today's engineering sync, we talked about making PR failures easier to diagnose. One way to do that is to put the error next to the failing code in the PR.

### Specifications

> Problem Matchers are a way to scan the output of actions for a specified regex pattern and surface that information prominently in the UI. Both [GitHub Annotations](https://developer.github.com/v3/checks/runs/#annotations-object-1) and log file decorations are created when a match is detected.

https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

This new action:
https://github.com/r7kamura/rust-problem-matchers#usage

## Solution

- Add the rust-problem-matchers action to most workflows

## Review

This is tricky to review unless we deliberately cause failures in this PR.

### Reviewer Checklist

## Follow Up Work

Fork the action and add our own regexes to catch Zebra-specific failures.

### DIFF
--- a/.github/workflows/build-crates-individually.yml
+++ b/.github/workflows/build-crates-individually.yml
@@ -51,6 +51,7 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - uses: actions/checkout@v3.5.2
+      - uses: r7kamura/rust-problem-matchers@v1.3.0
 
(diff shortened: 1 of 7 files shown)


dconnolly (approved):
Hell yes, I didn't know this finally exists as a first-party github feature 😍

Issue #6506, opened 2023-04-13
Multiple tests fail in CI with "inbound service overloaded" warnings


arya2:
## Motivation

This test is failing in CI on main: https://github.com/ZcashFoundation/zebra/actions/runs/4687263326/jobs/8306335685#step:15:598

<details>
 
> Error: 
>    0: stdout of command did not log any matches for the given regex,
>       within the 3m command timeout
> 
> Location:
>    /Users/runner/work/zebra/zebra/zebra-test/src/command.rs:854
> 
> Match Regex:
>    [
>        "stopping at configured height",
>    ]
> 
> Command:
>    cd "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebrad_testsKJDaJD" && "/Users/runner/work/zebra/zebra/target/release/zebrad" "-c" "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebrad_testsKJDaJD/zebrad.toml" "start"
> 
>   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
>    0: zebra_test::command::expect_line_matching_regexes with success_regexes=RegexSet(["stopping at configured height"]) stream_name="stdout"
>       at zebra-test/src/command.rs:810
>    1: zebra_test::command::expect_stdout_line_matches with success_regex="stopping at configured height"
>       at zebra-test/src/command.rs:701
>    2: acceptance::common::sync::check_sync_logs_until with network=Mainnet stop_regex="stopping at configured height" mempool_behavior=ForceActivationAt(Height(0)) check_legacy_chain=true
>       at zebrad/tests/common/sync.rs:303
> 

> Error: 
>    0: stdout of command did not log any matches for the given regex,
>       within the 3m command timeout
> 
> Location:
>    /opt/zebrad/zebra-test/src/command.rs:854
> 
> Match Regex:
>    [
>        "stopping at configured height",
>    ]
> 
> Command:
>    cd "/tmp/zebrad_testsnttLId" && "/opt/zebrad/target/release/zebrad" "-c" "/tmp/zebrad_testsnttLId/zebrad.toml" "start"
> 
[...]

PR #6691, opened 2023-05-16
fix(ci): Give more time to the inbound service when queueing requests

teor2345:
## Motivation

This is a potential fix for #6506.

### Specifications

https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html

This doesn't change concurrency behaviour much - it just yields time to other tasks.

## Solution

Wait for the inbound service to process other requests before and after queueing a request

#### Unrelated changes

Remove redundant deny.toml entry

## Review

This is a routine fix.

### Reviewer Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -41,9 +41,6 @@ skip-tree = [
     # wait for primitive-types to upgrade
     { name = "proc-macro-crate", version = "=0.1.5" },
 
-    # wait for rocksdb to upgrade
-    { name = "bindgen", version = "=0.64.0" },
-
     # wait for prost-build to upgrade
     { name = "prettyplease", version = "=0.1.25" },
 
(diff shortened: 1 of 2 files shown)


dconnolly (approved):
lgtm but is it expected to yield back to the executor multiple times in one async worker? just curious

teor2345:
> lgtm but is it expected to yield back to the executor multiple times in one async worker? just curious

Yep, because some of the requests have multiple await points in them, so they might need to run multiple times to stop using the services that are heavily loaded.

teor2345:
This PR failed with an overloaded error:
https://github.com/ZcashFoundation/zebra/actions/runs/4988361494/jobs/8931748876?pr=6691#step:3:3911

It is possible but unlikely that these changes make these errors happen more often. So if we see them increase after merging this PR, we should revert it.

Issue #6553, opened 2023-04-20
Make the latest production release sort first on DockerHub


teor2345:
## Motivation

Currently, the docker image for the latest production release is the 5th entry on DockerHub:
https://hub.docker.com/r/zfnd/zebra/tags

But we want people to use that image, so it should be listed first to avoid confusion. This is a usability issue.

## Potential Solution

This might be fixed by adding a `latest` tag, which should sort first in alphabetical order.

### Testing

Do a release, and check what appears here:
https://hub.docker.com/r/zfnd/zebra/tags

teor2345:
@mpguerra this is something we probably want to start doing in a release candidate at least two releases before the stable release.

gustavovalverde:
This should be done automatically when we release v1.0.0 without `-rc.0`. For Docker, `latest` is only the last `stable` release. https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml#L99

This can be closed, or we can wait until we release to confirm.

teor2345:
It would be nice to check this happens how we expect after the stable release.

gustavovalverde:
Confirmed https://hub.docker.com/r/zfnd/zebra/tags

PR #6710, opened 2023-05-17
fix(release): do not hardcode the name of our images

gustavovalverde:
## Motivation

Our main image in DockerHub is called `zebra` not `zebrad`.

This hardcoded value is also causing the mining image to also be called `zebra` instead of `zebrad-mining-rpcs-testnet`

We should also decide between having a different image `name`, `tag` or both. Basically deciding between:
- `zebra:v1.0.0-rc.8.experimental`
- `zebra:v1.0.0-rc.8.mining`
- `zebrad-mining-rpcs-testnet:v1.0.0-rc.8.experimental`

## Solution

- Use `zfnd/${{ inputs.image_name }}` instead of `zfnd/zebra`
- Do not name experimental tags with a different image name for Docker Hub

## Review

Anyone from the DevOps team can review this.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -88,7 +88,7 @@ jobs:
           # list of Docker images to use as base name for tags
           images: |
             us-docker.pkg.dev/${{ vars.GCP_PROJECT }}/zebra/${{ inputs.image_name }}
-            zfnd/zebra,enable=${{ github.event_name == 'release' && !github.event.release.prerelease }}
+            zfnd/${{ inputs.image_name }},enable=${{ github.event_name == 'release' && !github.event.release.prerelease }}
           # appends inputs.tag_suffix to image tags/names
           flavor: |
             suffix=${{ inputs.tag_suffix }}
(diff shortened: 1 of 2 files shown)


teor2345:
I think we should use "zebra" because:
* it's the better known name
* we plan to add other binaries to the image in future

But it's slightly harder to search for:
* people searching for zebrad might not find zebra, but people searching for zebra will find zebrad

teor2345:
I also think it would be better to have a different name for experimental images, so their tags don't appear in the same list as production images. This avoids users accidentally using experimental images when they actually wanted a production image.

That would also close:
- #6553

mpguerra:
My vote goes to `zebra:v1.0.0-rc.8.experimental`

gustavovalverde:
> I also think it would be better to have a different name for experimental images, so their tags don't appear in the same list as production images. This avoids users accidentally using experimental images when they actually wanted a production image.

I've taken some time to evaluate multiple organizations in DockerHub, and it's commonly unexpected to have a different `repository` for test, unstable, experimental and/or nightly images.  A good example is https://hub.docker.com/_/debian which has multiple of these tags in their repo.

Nonetheless, having `latest` on top would help for people who use the DockerHub UI, instead of using the CLI. But this should be clear in the README anyways, and we'll have the `latest` tag when we make our stable release, as commented here: https://github.com/ZcashFoundation/zebra/issues/6553#issuecomment-1557956352

arya2 (approved):
Looks good, thank you for the fix!

PR #6713, opened 2023-05-17
cleanup(tests): Move end_of_support.rs into acceptance.rs so tests are built faster

teor2345:
## Motivation

We currently have a separate test file for end of support tests, which makes `cargo` link `zebrad` into two different test binaries. This increases build time and test size on disk.

### Specifications

> Each file in the tests directory is a separate crate, so we need to bring our library into each test crate’s scope. For that reason we add use adder at the top of the code, which we didn’t need in the unit tests.

https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests

## Solution

Move those tests into `acceptance.rs`

## Review

### Reviewer Checklist


teor2345:
This doesn't work due to a dependency on `tracing_test`

PR #6715, opened 2023-05-17
fix(test): Check for early exits in the end of support task test

teor2345:
## Motivation

Our integration tests could ignore early exits in the end of support task.

### Specifications

`Result::ok()` turns a `Result` into an `Option`:
https://doc.rust-lang.org/std/result/enum.Result.html#method.ok

## Solution

Expect a timeout error instead of ignoring the result of the task

## Review

This is a low priority test fix.

### Reviewer Checklist


oxarbitrage (approved):
Looks good, thanks.

PR #6716, opened 2023-05-18
DRAFT: radicle contrib init

patitieso:
Introduction and basic instructions for Radicle peer-to-peer network for code collaboration, and seednode.link


mpguerra:
Thank you for the PR. 

Our current priority is to release a stable version of our software and we're not currently looking into these sorts of collaborations.

PR #6717, opened 2023-05-18
fix(network): Ignore out of order Address Book changes, unless they are concurrent

### LINKED ISSUE #6672: [NCC-E005955-7DU] `zebra-network`: Fragile State Transition During Address Book Update
## Impact

Failure to reject out of order address change requests corrupts the Address Book’s state and opens the Zebra node to state manipulation attacks.

## Description
The `zebra_network`’s `AddressBook` update implementation uses `MetaAddrChange`’s `apply_to_meta_addr()` to update the entry’s previous state to the received updated state.
The `apply_to_meta_addr()` function validates the change against the previous state and optionally returns the new `MetaAddr`. If the received state is not the never-attempted state (the else condition on line 831) the current state is one of { `AttemptPending`, `Responded`, `Failed`}. In order to tolerate an address change request that is received out of order, the implementation picks the maximum of { `last_response`, `last_attempt`, `last_failure`} timestamps. Thus these timestamps will never revert to their previous values. However, independent of what the previous state was, on line 853, the new address state is returned. The `last_connection_state` records the outcome of local node’s most recent
communication attempt with this peer:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/meta_addr.rs#L785-L860

## Recommendation

Update `apply_to_meta_addr()` to return `None` when the state transition is invalid, e.g., the request is received out-of-order.

## Location

[zebra-network/src/meta_addr.rs, line 853](https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/meta_addr.rs#L853)

teor2345:
## Motivation

We want to ignore earlier address book changes which are delayed, so we don't accidentally reset the peer state to an earlier value. But some changes can be concurrent, so the change application order doesn't always match the network order.

Close #6672 

### Specifications

There is no formal specification for peer network behaviour, but in general, the peer's internal address book state should reflect the actual behaviour of its peers.

This code runs concurrently, it is accessed from both the address book updater task and the candidate set peer dialler task. 

Changes can arrive earlier or later than their network order because of:
- waiting on the address book mutex
- async runtime or OS thread scheduling
- monotonic clock resolution, or OS or hardware time bugs
- wall clock time updates

The application can also independently initiate a change to the connection state, which can be simultaneous (or before) a change from the peer. In this case, the application-provided change needs to take priority. (Because it is typically failing the connection.)

## Solution

Fix the code that applies changes to the address book, so that:
- Concurrent changes are applied if they progress the connection state, but ignored if they revert it
- Much earlier changes are ignored, even if they progress the connection state

Refactor the code that applies changes so it is easier to read.

Add TODOs for further changes that are not required to fix this security issue.

### Testing

- Add tests for the failure and success cases for the new checks, the extra success cases are:
    - Much later changes are applied, even if they revert the connection state
- Allow tests to provide synthetic times for address book updates
    - This resolves some technical debt caused by a partial refactor
    - Calling time APIs less will also improve performance on some systems

There are existing tests for the other kinds of changes.

## Review

This is one of the more complex audit changes that we have left to do. I'm happy to do a video review if that helps!

### Reviewer Checklist

## Follow Up Work

Make it much harder for remote peers to delay changes by flooding the change channel:
- #6738

### DIFF
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -90,6 +90,13 @@ pub const REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
 /// nodes, and on testnet.
 pub const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(3);
 
+/// The maximum time difference for two address book changes to be considered concurrent.
+///
+/// This timeout should be less than the reconnection and keepalive/heartbeat timeouts,
+/// but more than the amount of time between connection events and address book updates,
+/// even under heavy load. (In tests, we have observed delays up to 500ms.)
+pub const CONCURRENT_ADDRESS_CHANGE_PERIOD: Duration = HANDSHAKE_TIMEOUT;
+
 /// We expect to receive a message from a live peer at least once in this time duration.
 ///
 /// This is the sum of:
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -92,10 +92,21 @@ pub const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(3);
 
 /// The maximum time difference for two address book changes to be considered concurrent.
 ///
-/// This timeout should be less than the reconnection and keepalive/heartbeat timeouts,
-/// but more than the amount of time between connection events and address book updates,
-/// even under heavy load. (In tests, we have observed delays up to 500ms.)
-pub const CONCURRENT_ADDRESS_CHANGE_PERIOD: Duration = HANDSHAKE_TIMEOUT;
+/// This prevents simultaneous or nearby important changes or connection progress
+/// being overridden by less important changes.
+///
+/// This timeout should be less than:
+/// - the [peer reconnection delay](MIN_PEER_RECONNECTION_DELAY), and
+/// - the [peer keepalive/heartbeat interval](HEARTBEAT_INTERVAL).
+///
+/// But more than:
+/// - the amount of time between connection events and address book updates,
+///   even under heavy load (in tests, we have observed delays up to 500ms),
+/// - the delay between an outbound connection failing,
+///   and the [CandidateSet](crate::peer_set::CandidateSet) registering the failure, and
+/// - the delay between the application closing a connection,
+///   and any remaining positive changes from the peer.
+pub const CONCURRENT_ADDRESS_CHANGE_PERIOD: Duration = Duration::from_secs(5);
 
 /// We expect to receive a message from a live peer at least once in this time duration.
 ///
--- a/zebra-network/src/meta_addr.rs
+++ b/zebra-network/src/meta_addr.rs
@@ -87,11 +87,7 @@ impl PeerAddrState {
     fn connection_state_order(&self, other: &Self) -> Ordering {
         use Ordering::*;
         match (self, other) {
-            (NeverAttemptedAlternate, NeverAttemptedAlternate)
-            | (NeverAttemptedGossiped, NeverAttemptedGossiped)
-            | (AttemptPending, AttemptPending)
-            | (Responded, Responded)
-            | (Failed, Failed) => Equal,
+            _ if self == other => Equal,
             // Peers start in one of the "never attempted" states,
             // then typically progress towards a "responded" or "failed" state.
             //
@@ -109,7 +105,10 @@ impl PeerAddrState {
             (_, AttemptPending) => Greater,
             (Responded, _) => Less,
             (_, Responded) => Greater,
-            // Failed is covered by the other cases
+            // These patterns are redundant, but Rust doesn't assume that `==` is reflexive,
+            // so the first is still required (but unreachable).
+            (Failed, _) => Less,
+            //(_, Failed) => Greater,
         }
     }
 }
@@ -133,11 +132,7 @@ impl Ord for PeerAddrState {
     fn cmp(&self, other: &Self) -> Ordering {
         use Ordering::*;
         match (self, other) {
-            (Responded, Responded)
-            | (NeverAttemptedGossiped, NeverAttemptedGossiped)
-            | (NeverAttemptedAlternate, NeverAttemptedAlternate)
-            | (Failed, Failed)
-            | (AttemptPending, AttemptPending) => Equal,
+            _ if self == other => Equal,
             // We reconnect to `Responded` peers that have stopped sending messages,
             // then `NeverAttempted` peers, then `Failed` peers
             (Responded, _) => Less,
@@ -148,7 +143,10 @@ impl Ord for PeerAddrState {
             (_, NeverAttemptedAlternate) => Greater,
             (Failed, _) => Less,
             (_, Failed) => Greater,
-            // AttemptPending is covered by the other cases
+            // These patterns are redundant, but Rust doesn't assume that `==` is reflexive,
+            // so the first is still required (but unreachable).
+            (AttemptPending, _) => Less,
+            //(_, AttemptPending) => Greater,
         }
     }
 }
--- a/zebra-network/src/meta_addr.rs
+++ b/zebra-network/src/meta_addr.rs
@@ -730,7 +730,7 @@ impl MetaAddrChange {
     }
 
     /// Return the untrusted last seen time for this change, if available.
-    pub fn untrusted_last_seen(&self) -> Option<DateTime32> {
+    pub fn untrusted_last_seen(&self, now: DateTime32) -> Option<DateTime32> {
         match self {
             NewInitial { .. } => None,
             NewGossiped {
@@ -739,7 +739,7 @@ impl MetaAddrChange {
             } => Some(*untrusted_last_seen),
             NewAlternate { .. } => None,
             // We know that our local listener is available
-            NewLocal { .. } => Some(DateTime32::now()),
+            NewLocal { .. } => Some(now),
             UpdateAttempt { .. } => None,
             UpdateResponded { .. } => None,
             UpdateFailed { .. } => None,
@@ -766,7 +766,7 @@ impl MetaAddrChange {
     //     so they are called in the connection tasks
     //
     /// Return the last attempt for this change, if available.
-    pub fn last_attempt(&self) -> Option<Instant> {
+    pub fn last_attempt(&self, now: Instant) -> Option<Instant> {
         match self {
             NewInitial { .. } => None,
             NewGossiped { .. } => None,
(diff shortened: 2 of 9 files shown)


zebra-network/src/constants.rs
```diff
@@ -90,6 +90,13 @@ pub const REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
 /// nodes, and on testnet.
 pub const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(3);
 
+/// The maximum time difference for two address book changes to be considered concurrent.
+///
+/// This timeout should be less than the reconnection and keepalive/heartbeat timeouts,
+/// but more than the amount of time between connection events and address book updates,
+/// even under heavy load. (In tests, we have observed delays up to 500ms.)
+pub const CONCURRENT_ADDRESS_CHANGE_PERIOD: Duration = HANDSHAKE_TIMEOUT;
```
teor2345: Any opinions on how long the concurrency period should be?
It's currently 3 seconds.

zebra-network/src/meta_addr.rs
```diff
@@ -79,6 +79,39 @@ impl PeerAddrState {
...
+            // but we respect the same order here as a precaution.)
+            (NeverAttemptedAlternate, _) => Less,
+            (_, NeverAttemptedAlternate) => Greater,
+            (NeverAttemptedGossiped, _) => Less,
+            (_, NeverAttemptedGossiped) => Greater,
+            (AttemptPending, _) => Less,
+            (_, AttemptPending) => Greater,
+            (Responded, _) => Less,
+            (_, Responded) => Greater,
+            // Failed is covered by the other cases
```
teor2345: This method returns the connection state machine order, to help us choose between concurrent changes.

zebra-network/src/meta_addr.rs
```diff
@@ -703,15 +739,34 @@ impl MetaAddrChange {
...
             UpdateFailed { .. } => None,
         }
     }
 
+    // # Concurrency
+    //
+    // We assign a time to each change when it is applied to the address book by either the
+    // address book updater or candidate set tasks. This is the time that the change was received
+    // from the updater channel, rather than the time that the message was read from the peer
+    // connection.
```
teor2345, line 752: This was surprising to me: address book updates get given a time when they are received from the channel, rather than when they actually happen.

But we don't need to fix this design issue to fix the audit issue.

zebra-network/src/meta_addr.rs
```diff
@@ -776,93 +831,197 @@ impl MetaAddrChange {
...
+            .unwrap_or_default()
+            || local_previous
+                .map(|local_previous| local_previous > local_now)
+                .unwrap_or_default();
+
+        // Is this change typically from a connection state that has more progress?
+        let connection_has_more_progress = self
+            .peer_addr_state()
+            .connection_state_order(&previous.last_connection_state)
+            == Ordering::Greater;
```
teor2345, line 921: These are the new checks we do to apply addresses in order.

zebra-network/src/meta_addr.rs
```diff
@@ -776,93 +831,197 @@ impl MetaAddrChange {
...
+            //
+            // The connection state machine order is used so that state transitions which are
+            // typically close together are preserved. These transitions are:
+            // - NeverAttempted*->AttemptPending->(Responded|Failed)
+            // - Responded->Failed
+            //
+            // State transitions like (Responded|Failed)->AttemptPending only happen after the
+            // reconnection timeout, so they will never be considered concurrent.
+            return None;
+        }
```
teor2345: These are the new ways we can ignore an address change.

zebra-network/src/meta_addr.rs
```diff
@@ -776,93 +831,197 @@ impl MetaAddrChange {
...
+                untrusted_last_seen: previous.untrusted_last_seen,
+                // This is a wall clock time, but we already checked that responses are in order.
+                // Even if the wall clock time has jumped, we want to use the latest time.
+                last_response: self.last_response(local_now).or(previous.last_response),
+                // These are monotonic times, we already checked the responses are in order.
+                last_attempt: self.last_attempt(instant_now).or(previous.last_attempt),
+                last_failure: self.last_failure(instant_now).or(previous.last_failure),
+                // Replace the state with the updated state.
+                last_connection_state: self.peer_addr_state(),
+            })
```
teor2345, line 1037: This is the new way to update an address book entry, if all the checks pass.

zebra-network/src/meta_addr/tests/vectors.rs
```diff
@@ -227,3 +238,210 @@ fn responded_long_ago_peer_is_not_gossipable() {
 
     assert!(!peer.is_active_for_gossip(chrono_now));
 }
+
+/// Test that a change that is delayed for a long time is not applied to the address state.
+#[test]
+fn long_delayed_change_is_not_applied() {
```
teor2345, line 244: The new tests start here, there are 4 of them: `2 new checks * success/failure`

zebra-network/src/constants.rs
arya2: I think anything between 2-5 seconds would be appropriate.

Is there any way for the timestamps to reflect the precise order that changes should be applied? Could we do that in the follow-up work for the rest of the issue?

Are there any risks with this approach while it's shorter than the keepalive/heartbeat timeouts? (Could you add references to those constants too.)

zebra-network/src/meta_addr.rs
```diff
@@ -79,6 +79,39 @@ impl PeerAddrState {
...
+    ///
+    /// See [`MetaAddrChange::apply_to_meta_addr()`] for more details.
+    fn connection_state_order(&self, other: &Self) -> Ordering {
+        use Ordering::*;
+        match (self, other) {
+            (NeverAttemptedAlternate, NeverAttemptedAlternate)
+            | (NeverAttemptedGossiped, NeverAttemptedGossiped)
+            | (AttemptPending, AttemptPending)
+            | (Responded, Responded)
+            | (Failed, Failed) => Equal,
```
arya2: Optional/Nitpick:

```suggestion
            _ if self == other => Equal,
            (Failed, Failed) => Equal,
```

arya2 (commented):
This is looking great, thank you for fixing this.

I like the workaround for the imprecise timestamps (though I'd like to review it more thoroughly).

zebra-network/src/constants.rs
teor2345: > Is there any way for the timestamps to reflect the precise order that changes should be applied? Could we do that in the follow-up work for the rest of the issue?

I don't think there is, because changes are sent by different tasks that must run concurrently: the candidate set dialer, and multiple connection message receiver closures. So we can't ensure that each received message is tagged with a time in order without a significant redesign to the entire network stack.

Even if we did, some machines have limited time resolution, so we can't guarantee that two messages on the same connection will always have different times. Sometimes the times would be equal, so we'd have to have state ordering logic anyway. (There are also known bugs impacting monotonic times which make them go backwards or stay equal.)

teor2345: > So we can't ensure that each received message is tagged with a time in order without a significant redesign to the entire network stack.

Also, some state change events are triggered by the application, so they can happen independently and simultaneously with peer messages. In that case, if we fail a connection, we don't want a peer to be able to overwrite it with "responded".

teor2345: > Are there any risks with this approach while it's shorter than the keepalive/heartbeat timeouts?

I can't think of any right now. The important state transition to preserve is `Failed/Responded -> AttemptPending`, which only happens after the reconnection timeout.

zebra-network/src/meta_addr.rs
teor2345: Applied in [cf7f618](https://github.com/ZcashFoundation/zebra/pull/6717/commits/cf7f618463462c8a30935e9e170574b8805c7d78)

zebra-network/src/constants.rs
teor2345: > (Could you add references to those constants too.)

Done in [3e3e32c](https://github.com/ZcashFoundation/zebra/pull/6717/commits/3e3e32cd6613322b5b64ec3c57d64c34b7e974a6), along with docs for more edge cases and an increase to the concurrency period.

arya2: > So we can't ensure that each received message is tagged with a time in order without a significant redesign to the entire network stack.

> Also, some state change events are triggered by the application, so they can happen independently and simultaneously with peer messages. In that case, if we fail a connection, we don't want a peer to be able to overwrite it with "responded".

Got it, thank you.

> Even if we did, some machines have limited time resolution, so we can't guarantee that two messages on the same connection will always have different times.

I think sequential response ids could fix that if there weren't other issues.

zebra-network/src/meta_addr.rs
```diff
@@ -703,15 +737,34 @@ impl MetaAddrChange {
...
+    // # Concurrency
+    //
+    // We assign a time to each change when it is applied to the address book by either the
+    // address book updater or candidate set tasks. This is the time that the change was received
+    // from the updater channel, rather than the time that the message was read from the peer
+    // connection.
+    //
+    // Since the connection tasks run concurrently in an unspecified order, and the address book
+    // updater runs in a separate thread, these times are almost always very similar. If Zebra's
+    // address book is under load, we should increase the rate-limits for new inbound or outbound
```
arya2:
```suggestion
    // address book is under load, we should reduce the rate-limits for new inbound or outbound
```

zebra-network/src/meta_addr.rs
```diff
@@ -776,93 +829,212 @@ impl MetaAddrChange {
...
+            // - Responded->Failed
+            //
+            // State transitions like (Responded|Failed)->AttemptPending only happen after the
+            // reconnection timeout, so they will never be considered concurrent.
+            return None;
+        }
+
+        // Valid changes
+
+        if change_to_never_attempted && !previous_has_been_attempted {
```
arya2, line 995: Nitpick/Optional: the second part of the condition isn't strictly necessary (but may help readability).

```suggestion
        if change_to_never_attempted {
```

zebra-network/src/meta_addr/tests/vectors.rs
```diff
@@ -227,3 +238,210 @@ fn responded_long_ago_peer_is_not_gossipable() {
...
+        outcome, None,
+        "\n\
+         unexpected application of a much earlier change to a peer:\n\
+         change: {change:?}\n\
+         times: {instant_early:?} {chrono_early}\n\
+         peer: {peer:?}"
+    );
+}
+
+/// Test that a change that happens long time after the previous change
```
arya2:
```suggestion
/// Test that a change that happens a long time after the previous change
```

arya2 (dismissed):
Thank you!

zebra-network/src/meta_addr.rs
teor2345: Thanks for picking that up, it's there for readability and maintainability.

zebra-network/src/meta_addr.rs
teor2345:
```suggestion
    // address book is under load, we should use lower rate-limits for new inbound or outbound
```

teor2345:
@oxarbitrage I'm going to go ahead and merge this PR now to avoid merge conflicts.

Feel free to review it or skip the review, whatever you work out with Pili.

arya2 (approved):

PR #6724, opened 2023-05-18
fix(test): Refactor a network test function to be more robust

### LINKED ISSUE #6674: Fragile Address Limit Implementation in Tests
In the following code snippet from `zebra-network`’s `AddressBook` implementation, the helper function adds a list of peer addresses to the address book. However, it first takes `addr_limit` elements from the list and then deduplicates it, as such if there are repeated elements in the truncated list, the resulting address book will be smaller than the required limit. It is more robust to deduplicate the list and then take `addr_limit` elements from it. As this function is only used in tests, it is left as a note:

https://github.com/ZcashFoundation/zebra/blob/5a88fe783c4c219beb94362b78974555fa1a398b/zebra-network/src/address_book.rs#L143-L188

oxarbitrage:
## Motivation

Fix https://github.com/ZcashFoundation/zebra/issues/6674. This is a note from the audit where a test function creates a list taking some elements and then deduplicating as they get inserted with the same key to another collection.
 
## Solution

Do not truncate in creating the iterator and insert to the final collection until a limit is reached.

`addr_limit` of `0` will not work anymore so i added an assert, i think requiring a max value to be at least 1 makes sense.

## Review

I think anyone can review, the changes are small and in test code only.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -180,6 +180,9 @@ impl AddressBook {
         let instant_now = Instant::now();
         let chrono_now = Utc::now();
 
+        // The maximum number of addresses should be always greater than 0
+        assert!(addr_limit > 0);
+
         let mut new_book = AddressBook::new(local_listener, network, span);
         new_book.addr_limit = addr_limit;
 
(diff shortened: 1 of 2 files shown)


arya2 (approved):
Looks good!

teor2345:
Looks like this PR had a temporary CI conflict with the crate changes in #6726. The missing required job was:
> crates-matrix

I tried re-running this workflow to fix it:
https://github.com/ZcashFoundation/zebra/actions/runs/5018012172

If that doesn't work, then we might want to delete and re-add the branch protection rule with the correct name. (I didn't change the CI workflow definition in that PR, so it might be a cache issue.)

teor2345:
@mergifyio refresh

teor2345:
> If that doesn't work, then we might want to delete and re-add the branch protection rule with the correct name. 

This was the problem: the branch protection rule used the patch workflow name, but the real workflow used a different name.

I fixed it in the branch protection rules for this PR, and for non-Rust PRs it's fixed in PR #6729.

PR #6732, opened 2023-05-19
Delete outdated `TODOs` refering to closed issues

mpguerra:
## Motivation

We want to clean up any code comments relating to `TODO` tasks that have already been addressed.

## Solution

Deletes comments related to issues that have been closed as fixed.

Depends-On: #6737 
## Review

For each deleted comment the reviewer should double check that the PR really addresses the issue which is mentioned in the code comment. 

Informational comments that refer to closed issues should not be removed unless they are no longer relevant. 

### Reviewer Checklist

## Follow Up Work

This is just a first pass. There are still a number of issues that have been closed as `wont-fix` referenced in the code and we should decide what we do about those.

Depends-

### DIFF
--- a/zebra-chain/src/work/difficulty/tests/vectors.rs
+++ b/zebra-chain/src/work/difficulty/tests/vectors.rs
@@ -457,7 +457,9 @@ fn check_testnet_minimum_difficulty_block(height: block::Height) -> Result<(), R
             .signed_duration_since(previous_block.header.time);
 
         // zcashd requires a gap that's strictly greater than 6 times the target
-        // threshold.
+        // threshold, as documented in ZIP-205 and ZIP-208:
+        // https://zips.z.cash/zip-0205#change-to-difficulty-adjustment-on-testnet
+        // https://zips.z.cash/zip-0208#minimum-difficulty-blocks-on-testnet
         match NetworkUpgrade::minimum_difficulty_spacing_for_height(Network::Testnet, height) {
             None => Err(eyre!("the minimum difficulty rule is not active"))?,
             Some(spacing) if (time_gap <= spacing) => Err(eyre!(
--- a/zebra-chain/src/work/difficulty/tests/vectors.rs
+++ b/zebra-chain/src/work/difficulty/tests/vectors.rs
@@ -457,7 +457,7 @@ fn check_testnet_minimum_difficulty_block(height: block::Height) -> Result<(), R
             .signed_duration_since(previous_block.header.time);
 
         // zcashd requires a gap that's strictly greater than 6 times the target
-        // threshold, but ZIP-205 and ZIP-208 are ambiguous. See bug #1276.
+        // threshold.
         match NetworkUpgrade::minimum_difficulty_spacing_for_height(Network::Testnet, height) {
             None => Err(eyre!("the minimum difficulty rule is not active"))?,
             Some(spacing) if (time_gap <= spacing) => Err(eyre!(
--- a/zebra-network/src/config.rs
+++ b/zebra-network/src/config.rs
@@ -347,7 +347,6 @@ impl<'de> Deserialize<'de> for Config {
 
         let config = DConfig::deserialize(deserializer)?;
 
-        // TODO: perform listener DNS lookups asynchronously with a timeout (#1631)
         let listen_addr = match config.listen_addr.parse::<SocketAddr>() {
             Ok(socket) => Ok(socket),
             Err(_) => match config.listen_addr.parse::<IpAddr>() {
--- a/zebra-network/src/peer_set/limit.rs
+++ b/zebra-network/src/peer_set/limit.rs
@@ -178,6 +178,9 @@ impl Drop for ConnectionTracker {
 
         // We ignore disconnected errors, because the receiver can be dropped
         // before some connections are dropped.
+        // # Security
+        //
+        // This channel is actually bounded by the inbound and outbound connection limit.
         let _ = self.close_notification_tx.send(ConnectionClosed);
     }
 }
--- a/zebra-network/src/peer_set/limit.rs
+++ b/zebra-network/src/peer_set/limit.rs
@@ -178,8 +178,6 @@ impl Drop for ConnectionTracker {
 
         // We ignore disconnected errors, because the receiver can be dropped
         // before some connections are dropped.
-        //
-        // TODO: This channel will be bounded by the connection limit (#1850, #1851, #2902).
         let _ = self.close_notification_tx.send(ConnectionClosed);
     }
 }
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -418,8 +418,6 @@ where
         for guard in self.guards.iter() {
             guard.abort();
         }
-
-        // TODO: implement graceful shutdown for InventoryRegistry (#1678)
     }
 
     /// Check busy peer services for request completion or errors.
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -504,7 +504,6 @@ where
     /// Checks if the minimum peer version has changed, and disconnects from outdated peers.
     fn disconnect_from_outdated_peers(&mut self) {
         if let Some(minimum_version) = self.minimum_peer_version.changed() {
-            // TODO: Remove when the code base migrates to Rust 2021 edition (#2709).
             let preselected_p2c_peer = &mut self.preselected_p2c_peer;
 
             self.ready_services.retain(|address, peer| {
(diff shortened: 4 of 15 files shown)


zebra-network/src/peer_set/set.rs
```diff
@@ -504,7 +504,6 @@ where
     /// Checks if the minimum peer version has changed, and disconnects from outdated peers.
     fn disconnect_from_outdated_peers(&mut self) {
         if let Some(minimum_version) = self.minimum_peer_version.changed() {
-            // TODO: Remove when the code base migrates to Rust 2021 edition (#2709).
```
mpguerra: Should the below line be removed also?

zebra-network/src/config.rs
```diff
@@ -347,7 +347,6 @@ impl<'de> Deserialize<'de> for Config {
 
         let config = DConfig::deserialize(deserializer)?;
 
-        // TODO: perform listener DNS lookups asynchronously with a timeout (#1631)
```
mpguerra, line 350: Do we instead want to update this comment to `WONTFIX` instead of `TODO`?

zebra-chain/src/work/difficulty/tests/vectors.rs
```diff
@@ -457,7 +457,7 @@ fn check_testnet_minimum_difficulty_block(height: block::Height) -> Result<(), R
             .signed_duration_since(previous_block.header.time);
 
         // zcashd requires a gap that's strictly greater than 6 times the target
-        // threshold, but ZIP-205 and ZIP-208 are ambiguous. See bug #1276.
+        // threshold.
```
teor2345: These ZIPs got fixed in response to our questions:
```suggestion
        // threshold, as documented in ZIP-205 and ZIP-208:
        // https://zips.z.cash/zip-0205#change-to-difficulty-adjustment-on-testnet
        // https://zips.z.cash/zip-0208#minimum-difficulty-blocks-on-testnet
```

zebra-network/src/peer_set/limit.rs
```diff
@@ -178,8 +178,6 @@ impl Drop for ConnectionTracker {
 
         // We ignore disconnected errors, because the receiver can be dropped
         // before some connections are dropped.
-        //
-        // TODO: This channel will be bounded by the connection limit (#1850, #1851, #2902).
         let _ = self.close_notification_tx.send(ConnectionClosed);
```
teor2345, line 184: This security constraint was implemented:

```suggestion
        // # Security
        //
        // This channel is actually bounded by the inbound and outbound connection limit.
        let _ = self.close_notification_tx.send(ConnectionClosed);
```

zebra-network/src/policies.rs
```diff
@@ -34,9 +34,6 @@ impl<Req: Clone + std::fmt::Debug, Res, E: std::fmt::Debug> Policy<Req, Res, E>
                 Some(
                     // Let other tasks run, so we're more likely to choose a different peer,
                     // and so that any notfound inv entries win the race to the PeerSet.
-                    //
-                    // TODO: move syncer retries into the PeerSet,
-                    //       so we always choose different peers (#3235)
                     Box::pin(tokio::task::yield_now().map(move |()| retry_outcome)),
```
teor2345, line 43: This security fix was actually implemented, but in a different way:

```suggestion
                    // # Security
                    //
                    // We want to choose different peers for retries, so we have a better chance of getting each block.
                    // This is implemented by the connection state machine sending synthetic `notfound`s to the 
                    // `InventoryRegistry`, as well as forwarding actual `notfound`s from peers.
                    Box::pin(tokio::task::yield_now().map(move |()| retry_outcome)),
```

zebra-network/src/protocol/external/codec.rs
```diff
@@ -184,10 +184,6 @@ impl Codec {
     /// Obtain the size of the body of a given message. This will match the
     /// number of bytes written to the writer provided to `write_body` for the
     /// same message.
-    ///
-    /// TODO: Replace with a size estimate, to avoid multiple serializations
-    /// for large data structures like lists, blocks, and transactions.
-    /// See #1774.
     fn body_length(&self, msg: &Message) -> usize {
```
teor2345, line 191: This is still a potential problem, but we don't want to fix it unless it's clear it's needed:

```suggestion
    // # Performance TODO
    //
    // If this code shows up in profiles, replace with a size estimate or cached size,
    // to avoid multiple serializations for large data structures like lists, blocks, and transactions.
    fn body_length(&self, msg: &Message) -> usize {
```

zebra-network/src/protocol/external/message.rs
```diff
@@ -401,8 +401,6 @@ impl TryFrom<Message> for VersionMessage {
     }
 }
 
-// TODO: add tests for Error conversion and Reject message serialization (#4633)
-// (Zebra does not currently send reject messages, and it ignores received reject messages.)
 impl<E> From<E> for Message
```
teor2345, line 406: This test coverage is still missing, if we ever modify this code we should add these tests:

```suggestion
// TODO: add tests for Error conversion and Reject message serialization
 // (Zebra does not currently send reject messages, and it ignores received reject messages.)
impl<E> From<E> for Message
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -110,8 +110,7 @@ pub trait GetBlockTemplateRpc {
     /// - the parent block is a valid block that Zebra already has, or will receive soon.
     ///
     /// Zebra verifies blocks in parallel, and keeps recent chains in parallel,
-    /// so moving between chains is very cheap. (But forking a new chain may take some time,
-    /// until bug #4794 is fixed.)
+    /// so moving between chains is very cheap.
```
```suggestion
    /// so moving between chains and forking chains is very cheap.
```

zebra-network/src/peer_set/set.rs
teor2345: Yes, but it needs a separate PR because it changes multiple lines, see PR #6737

teor2345 (dismissed):
Looks good, but I think we can update some of these comments based on how their tickets were resolved.

zebra-network/src/peer_set/set.rs
teor2345: Yes, but it needs a separate PR because it changes multiple lines, see PR #6737

zebra-network/src/config.rs
teor2345: It turns out we actually implemented this change, it's in `Config::resolve_peers()`. So there is nothing left to do (or to WONTFIX!).

zebra-network/src/peer_set/set.rs
mpguerra: Added mergify dependency on #6737

gustavovalverde:
Rustfmt is failing after the changes: https://github.com/ZcashFoundation/zebra/actions/runs/5043379478/jobs/9045106343?pr=6732

zebra-network/src/policies.rs
```diff
@@ -34,9 +34,11 @@ impl<Req: Clone + std::fmt::Debug, Res, E: std::fmt::Debug> Policy<Req, Res, E>
                 Some(
                     // Let other tasks run, so we're more likely to choose a different peer,
                     // and so that any notfound inv entries win the race to the PeerSet.
+                    // # Security
                     //
-                    // TODO: move syncer retries into the PeerSet,
-                    //       so we always choose different peers (#3235)
+                    // We want to choose different peers for retries, so we have a better chance of getting each block.
+                    // This is implemented by the connection state machine sending synthetic `notfound`s to the
```
teor2345:
```suggestion
                    // This is implemented by the connection state machine sending synthetic `notfound`s to the
```

zebra-network/src/policies.rs
```diff
@@ -34,9 +34,11 @@ impl<Req: Clone + std::fmt::Debug, Res, E: std::fmt::Debug> Policy<Req, Res, E>
                 Some(
                     // Let other tasks run, so we're more likely to choose a different peer,
                     // and so that any notfound inv entries win the race to the PeerSet.
+                    // # Security
```
teor2345, line 38:
```suggestion
                    //
                    // # Security
```

zebra-network/src/protocol/external/message.rs
```diff
@@ -401,8 +401,8 @@ impl TryFrom<Message> for VersionMessage {
     }
 }
 
-// TODO: add tests for Error conversion and Reject message serialization (#4633)
-// (Zebra does not currently send reject messages, and it ignores received reject messages.)
+// TODO: add tests for Error conversion and Reject message serialization
+ // (Zebra does not currently send reject messages, and it ignores received reject messages.)
```
teor2345:
```suggestion
// (Zebra does not currently send reject messages, and it ignores received reject messages.)
```

teor2345 (approved):
Thanks for reviewing my fixes, I fixed the rustfmt and the merge conflict, both caused by my PRs and suggestions.

PR #6738, opened 2023-05-21
fix(net): Rate-limit MetaAddrChange::Responded from peers

teor2345:
## Motivation

There is no limit to the amount of `MetaAddrChange::Responded` updates a peer can send to the shared address book updater channel.

This is a remotely-triggerable security issue, likely resulting in a hang or very slow network performance.

This code runs concurrently for every peer, then writes to a channel shared between all peers. Writing on the channel blocks if it is full.

## Solution

- Stop allowing peers to send a `Ping` or `Pong` message to generate a `MetaAddrChange::Responded`
- Instead, generate a `MetaAddrChange::Responded` after every successful heartbeat, which is already rate-limited by Zebra

## Review

This is a security fix that we might want to block the stable release for.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer/handshake.rs
+++ b/zebra-network/src/peer/handshake.rs
@@ -914,6 +914,10 @@ where
             // addresses. Otherwise, malicious peers could interfere with the
             // address book state of other peers by providing their addresses in
             // `Version` messages.
+            //
+            // New alternate peer address and peer responded updates are rate-limited because:
+            // - opening connections is rate-limited
+            // - we only send these messages once per handshake
             let alternate_addrs = connected_addr.get_alternate_addrs(remote_canonical_addr);
             for alt_addr in alternate_addrs {
                 let alt_addr = MetaAddr::new_alternate(alt_addr, &remote_services);
@@ -1411,6 +1415,12 @@ where
         Err(err) => {
             tracing::debug!(?err, "heartbeat error, shutting down");
 
+            // # Security
+            //
+            // Peer errors and shutdowns are rate-limited because:
+            // - opening connections is rate-limited
+            // - the number of connections is limited
+            // - after the first error or shutdown, the peer is disconnected
             if let Some(book_addr) = connected_addr.get_address_book_addr() {
                 let _ = address_book_updater
                     .send(MetaAddr::new_errored(book_addr, *remote_services))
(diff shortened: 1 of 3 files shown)


teor2345:
@mpguerra this seems serious enough to block the stable release, but it's a small change so it should be easy to review.

teor2345:
Temporary network failure:
> Error: buildx failed with: ERROR: failed to solve: failed to push us-docker.pkg.dev/zfnd-dev-zebra/zebra/zebrad-test:pr-6738: failed to copy: failed to do request: Put "https://us-docker.pkg.dev/artifacts-uploads/namespaces/zfnd-dev-zebra/repositories/zebra/uploads/...": write tcp 172.17.0.2:47038->142.251.116.82:443: write: connection reset by peer

https://github.com/ZcashFoundation/zebra/actions/runs/5039819774/jobs/9038203151?pr=6738#step:10:926

mpguerra:
> @mpguerra this seems serious enough to block the stable release, but it's a small change so it should be easy to review.

Is there a corresponding issue for it? It's ok if not, I just want to make sure we track it somehow.

teor2345:
> > @mpguerra this seems serious enough to block the stable release, but it's a small change so it should be easy to review.
> 
> Is there a corresponding issue for it? It's ok if not, I just want to make sure we track it somehow.

I didn't open a separate ticket.

This is related to #6672, because it was one of the ways that remote peers could easily manipulate their address state inside Zebra. (If we'd fixed this bug earlier, the audit finding might have been low severity, because peer state changes would all have been rate-limited or triggered by the local Zebra node.)

Also happy to call it a separate bug if you want.

mpguerra:
> > > @mpguerra this seems serious enough to block the stable release, but it's a small change so it should be easy to review.
> > 
> > 
> > Is there a corresponding issue for it? It's ok if not, I just want to make sure we track it somehow.
> 
> I didn't open a separate ticket.
> 
> This is related to #6672, because it was one of the ways that remote peers could easily manipulate their address state inside Zebra. (If we'd fixed this bug earlier, the audit finding might have been low severity, because peer state changes would all have been rate-limited or triggered by the local Zebra node.)
> 
> Also happy to call it a separate bug if you want.

Nope, I see I already linked it to the relevant audit issue, it's all I need :)

arya2 (approved):
Looks good!

PR #6739, opened 2023-05-22
feat(net): Cache a list of useful peers on disk

### LINKED ISSUE #1956: Zebra should store a persistent list of recent successful peers
## Motivation

Zebra currently fetches a new list of peers from DNS each time it launches.

This design has some benefits:
- Zebra behaviour is consistent and has minimal state

And some drawbacks:
- Zebra will fail if DNS is down or slow when it launches
  - see also #1892 where Zebra regularly queries DNS for peers
- Zebra relies on peer addresses from DNS, which is not encrypted or authenticated by default
- Zebra leaks the exact time that it launches by making a DNS query

## Solution

Zebra should keep a cached list of peer addresses on disk, and update it regularly. This list can be separate from the state cache.

To avoid privacy leaks:
- the list should only store peer addresses, and no other data
- it should be sorted to destroy any order in the peer list

### Alternative 1: Only use DNS if cached peers don't work

`zcashd` tries both the cached list of peer addresses and new addresses from DNS. This design has the following tradeoffs:
- Zebra doesn't rely on DNS
- Every time Zebra starts, it can have bad peers added by its current DNS
- Launching is as fast as possible, regardless of the quality of the cached list or DNS
- Zebra leaks some information by making a DNS query every time it starts up
- Zebra can easily recover from a compromised or poorly performing peer list, but some bad peers will stay in the list

### Alternative 2: Only use DNS if cached peers don't work

`zcashd` first tries the cached list of peer addresses, then makes a DNS query if none connect. This design has the following tradeoffs:
- Zebra doesn't rely on DNS unless it really needs to
- Launching might be slower if it takes a while to work out that all peers have failed
- Zebra leaks some information about the quality of its cached peer address list
- Zebra doesn't have a peer reputation system, so it's hard to work out if all peers are "bad"
- It's harder for Zebra to recover from a compromised or poorly performing peer list, because it might never use DNS again

teor2345:
## Motivation

Zebra currently uses DNS seeders as its only source of peers. To make it more secure and reliable, we store a list of recent useful peers on disk, and read it at startup.

Close #1956

### Specifications

Peer behaviour is not specified, but `zcashd` does something like this. (In fact it tries to rely on DNS even less.)

This PR introduces a new async task, and new filesystem code that must be run in an async context.

## Solution

- Add a new `zebra-network` config for the peer cache directory
- Regularly write recent useful peers to a file
    - Do it atomically so the file isn't corrupted if Zebra shuts down or panics while it is being written
    - If the peer cache would become empty, keep the previous cache file
    - Limit the number of peers in the peer cache
    - Add a new `zebra-network` peer set task to run this code 
- At startup, load the peers from a file, and use them in addition to DNS
- Add metrics and logging for the peer cache

Related changes:
- Update state cache docs to fix outdated info
- Add some zebra-chain and zebra-network convenience functions
- Panic if the configured peer addresses in `zebrad.toml` are invalid, rather than continuing

### Testing

Config:
- Add 3 new zebrad.toml config test files containing the variants of the new config field

Unit:
- make sure the peer cache is written by zebra_network::initialize and can be read back manually
- make sure the peer cache is written by zebra_network::initialize and is automatically loaded by the next instance

Acceptance:
- make sure the peer cache file path is written by `zebrad` in the existing persistent cache test
- make sure that the peer cache is not written when the cache is disabled (existing tests)
- make sure that a missing cache_dir config gets the default path, but a disabled config stays disabled when written to disk (existing tests)

Test Fixes:
- Update some existing tests to disable or configure the peer cache as needed
- Update the config test failure message for the new toml format using standard quotes

## Review

This is a routine audit fix.

Design questions:
- should cached peers replace DNS peers, or be used in combination with DNS peers?
- the peer cache is shared by all Zebra instances run by a single user, is this ok?

### Reviewer Checklist

## Follow Up Work

We might also want to add hard-coded IP addresses to the default list of peers.


zebra-network/src/config.rs
```diff
@@ -71,9 +76,54 @@ pub struct Config {
...
+    ///
+    /// # Implementation Details
+    ///
+    /// Each network has a separate peer list, which is updated regularly from the current
+    /// address book. These lists are stored in `network/mainnet.peers` and
+    /// `network/testnet.peers` files, underneath the `cache_dir` path.
+    ///
+    /// Previous peer lists are automatically loaded at startup, and used to populate the
+    /// initial peer set and address book.
+    pub cache_dir: Option<PathBuf>,
```
teor2345: This is wrong because Option::None is not written to disk, and then when it is read back it is replaced by Option::Some(default_path).

teor2345: Use a custom untagged enum with `bool` and `PathBuf` types.
https://serde.rs/enum-representations.html#untagged

arya2: This should work:

```rust

#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
#[serde(untagged)]
pub enum ShouldCachePeers {
    Path(PathBuf),
    Bool(bool),
}

impl ShouldCachePeers {
    pub fn cache_dir(&self) -> Option<PathBuf> {
        match self {
            Self::Path(cache_dir) => Some(cache_dir.to_owned()),
            Self::Bool(should_cache) => should_cache.then(|| {
                dirs::cache_dir()
                    .unwrap_or_else(|| std::env::current_dir().unwrap().join("cache"))
                    .join("zebra")
            }),
        }
    }
}

pub fn peer_cache_file_path(&self, network: Network) -> Option<PathBuf> {
    Some(
        self.should_cache_peers
            .cache_dir()?
            .join("network")
            .join(format!("{}.peers", network.lowercase_name())),
    )
}

```

teor2345: Thanks, I switched some names around, but the overall implementation is the same.

teor2345:
@arya2 anything I can do to help get this reviewed and merged?

zebra-network/src/config.rs
```diff
@@ -260,6 +347,13 @@ impl Config {
 
                 Ok(ip_addrs.into_iter().collect())
             }
+            Ok(Err(e)) if e.kind() == ErrorKind::InvalidInput => {
+                // TODO: add testnet/mainnet ports, like we do with the listener address
+                panic!(
+                    "Invalid peer IP address in Zebra config: addresses must have ports:\n\
```
arya2, line 353: ~Nitpick: include the path to the cached peer addresses here?~

Update: invalid peer cache files should be covered by `.filter_map(|peer| { peer.parse() })`

arya2 (dismissed):
This looks perfect, thank you!

teor2345:
@mergifyio refresh

teor2345:
@arya2 this is ready for re-approval. (Or anyone can do it, it's just a dependency merge.)

arya2 (approved):

PR #6755, opened 2023-05-24
fix(release): Allow inbound connections to Zebra running in Docker

upbqdn:
## Motivation & Solution

Binding `127.0.0.1` means that Zebra will accept inbound connections coming only from the loopback network interface. This is desirable as long as Zebra runs on a native machine.

When Zebra runs inside a Docker container, incoming connections coming from the host machine don't come from the container's loopback interface. In order to be able to connect to Zebra from the host machine, we can listen on `0.0.0.0` so Zebra accepts inbound connections coming from any interface. Users then can limit inbound connection to the loopback of their host by

```bash
docker run -p 127.0.0.1:8232:8232 zfnd/zebra:1.0.0-rc.8
```

We should maybe document this, although Docker operators should be familiar with it.

### Reviewer Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -180,18 +180,18 @@ RUN set -ex; \
   { \
     echo "[network]"; \
     echo "network = '${NETWORK}'"; \
-    echo "listen_addr = '127.0.0.1'"; \
+    echo "listen_addr = '0.0.0.0'"; \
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
Thank you for taking the time testing and fixing this!

PR #6790, opened 2023-05-30
fix(security): Randomly drop connections when inbound service is overloaded

### LINKED ISSUE #6596: Security Issue #38: Stop disconnecting all peers when the inbound service is overloaded
## Motivation

When Zebra's inbound service is overloaded, it disconnects every peer that makes a request to Zebra, until the inbound service processes at least one request.

This allows a single peer to disconnect all other peers by repeatedly making inbound connections, and sending lots of inbound requests on those connections.

### Goals

- [ ] Make sure a single peer can't make Zebra drop connections from other peers
- [ ] Replace `Overloaded` error during shutdown with a new error type, so that connections shut down properly after this fix

https://github.com/ZcashFoundation/zebra/blob/185d1384fbf4e7a31c794b17d7fc69dd4aeb5aa7/zebra-network/src/peer/connection.rs#L1256-L1261

All requests concurrently received from peers go into a single queue, where they are processed one by one.

The load shedding is configured here, in response to the buffer reaching its limit:
https://github.com/ZcashFoundation/zebra/blob/166526a088ae127d88714d93b791b95a35c42447/zebrad/src/commands/start.rs#L133-L139

#### Rate-limit connections

- Define `INBOUND_USAGE_LIMIT` and `INBOUND_USAGE_LIMIT_REFRESH_INTERVAL` constants
- Keep a count of requests made by a connection during an interval with an `AtomicUsize` that is reset to zero at the interval
- Delay requests until the next usage interval if a connection has already reached its limit
- Start the next interval when the inbound service buffer is emptied

---

#### Randomly keep some overloaded connections

When the inbound service is overloaded, randomly choose to keep or drop connections that send inbound requests. If a short time has elapsed since the last overload, increase this probability. Otherwise, reset this probability.

This can be done without locking by storing the last overload time in an `AtomicU64`. But a `Mutex` would probably be fine, too.

---

#### Add a PendingRequests load measurer to the service stack, above the buffer:
https://docs.rs/tower/latest/tower/load/pending_requests/struct.PendingRequests.html

If the service isn't fully overloaded, but the load is past a lower limit, randomly return an Overloaded error for some requests:
https://github.com/ZcashFoundation/zebra/blob/166526a088ae127d88714d93b791b95a35c42447/zebra-network/src/peer/connection.rs#L1242-L1261

Make the overloaded errors more likely as the load gets closer to the load shed limit.

--- 
#### Apply a layer to the inbound service in zebra-network to buffer requests and reserve service capacity for each peer connection.

Call the underlying service with one queued request from each peer before calling it with a second queued request from any given peer.

### Testing

Deliberately cause a partial overload, and make sure only some connections are dropped.
Make sure all connections that make an inbound request are dropped during a full overload.

## Related Work

We've seen all connections dropped in some of our tests, see:
- https://github.com/ZcashFoundation/zebra/issues/6506

teor2345:
## Motivation

Zebra should avoid dropping all peer connections when the inbound service is overloaded by a subset of peers.

This PR will randomly drop connections that get an overloaded error, with a high probability 

Closes #6596.

## Solution

- Randomly drop connections that receive overloaded errors, with greater likelihood when a second overloaded error is seen soon after the first.
- Add a `ServiceShutdown` variant to `PeerError` for when the inbound service has failed or shutdown.

Testing:
- more overloads mean a higher probability of being dropped
- less overloads mean a lower probability
- frequent overloads significantly increase the probability
- when we make multiple connections, some are dropped and some aren't

Related fixes:
- add some test utility methods and fix docs

## Review

This is a routine fix.

The production code has already been reviewed privately, the tests are new.

### Reviewer Checklist

## Follow Up Work

- Use a skiplist to keep an ordered buffer queue

### DIFF
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -316,20 +316,16 @@ pub const EWMA_DECAY_TIME_NANOS: f64 = 200.0 * NANOS_PER_SECOND;
 /// The number of nanoseconds in one second.
 const NANOS_PER_SECOND: f64 = 1_000_000_000.0;
 
-/// The minimum interval that must elapse for the drop probability of an overloaded connection to decrease.
+/// The duration it takes for the drop probability of an overloaded connection to
+/// reach [`MIN_OVERLOAD_DROP_PROBABILITY`].
 ///
 /// Peer connections that receive multiple overloads have a higher probability of being dropped.
-/// Each time this interval elapses after the last overload, the probability of a connection being
-/// dropped gradually decreases, until it reaches the default drop probability.
+///
+/// The probability of a connection being dropped gradually decreases during this interval
+/// until it reaches the default drop probability ([`MIN_OVERLOAD_DROP_PROBABILITY`]).
 ///
 /// Increasing this number increases the rate at which connections are dropped.
-pub const OVERLOAD_DROP_PROBABILITY_INTERVAL: Duration = Duration::from_millis(50);
-
-/// The number of [`OVERLOAD_DROP_PROBABILITY_INTERVAL`]s that must elapse to reach the minimum
-/// drop probability for an overloaded connection.
-///
-/// Increasing this number increases the rate at which connections are dropped.
-pub const NUM_OVERLOAD_DROP_PROBABILITY_INTERVALS: f32 = 10.0;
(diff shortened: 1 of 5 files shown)


teor2345:
Sorry for the doctest compile errors, I didn't catch them locally.

(Next time I'll fix the test API separately.)

arya2 (approved):

PR #6024, opened 2023-01-24
fix(clippy): Silence future-incompat warnings until we upgrade Abscissa

teor2345:
## Motivation

Nightly Rust has started warning that abscissa 0.5.2 might not be supported by the 2024 Rust edition.

Let's turn that warning off until we are ready to upgrade Abscissa:
https://github.com/ZcashFoundation/zebra/issues/5502

## Solution

Disable all future incompatibility warnings for now.

There is a small risk we will miss another incompatibility warning, but it will also likely be in the 2024 Rust edition, which is optional to upgrade to.

## Review

This is a low priority.

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [x] How do you know it works? Does it have tests?

## Follow Up Work

We might want to upgrade abscissa some time in the next year or so:
https://github.com/ZcashFoundation/zebra/issues/5502

PR #6801, opened 2023-05-31
fix(zebrad): accept default subcommand arguments and print consistent usage information for top-level 'help' subcommand

### LINKED ISSUE #5502: Migrate to the latest version of Abscissa (0.7 or later)
## Motivation

We need to update our command-line argument parsing to bump abscissa_core from 0.5.2 to 0.7.0 or later.

This is also needed to make the output of the `--help` flag and `help` command consistent with that of invalid options or commands, by replacing `gumdrop` with `clap`.

It is also needed to avoid a compilation error that is likely to be added to the Rust 2024 edition:
https://github.com/rust-lang/rust/issues/79813

This error is currently a warning on `nightly`.

### Security

This will fix a low-severity security issue - we don't actually call the vulnerable abscissa 0.5.2 or tracing-subscriber 0.1.6 code, so it's only an issue at startup.

It will also make our dependency tree smaller, which improves compile speed, binary size, and security.

Vulnerability:
https://github.com/advisories/GHSA-wcg3-cvx6-7396
https://github.com/time-rs/time/issues/293

Fix:
https://github.com/chronotope/chrono/issues/602#issuecomment-1242149249

## Steps

[Migration guide](https://github.com/iqlusioninc/abscissa/blob/main/CHANGELOG.md#migration-guide)

1. Upgrade command-line argument parsing
2. Upgrade abscissa
    a. Consider replacing Zebra's custom `tokio` and `metrics` components with the current Abscissa implementations
3. Turn future incompatibility warnings back on, by reverting https://github.com/ZcashFoundation/zebra/pull/6024

## Related Work

- #3372

### LINKED ISSUE #5624: zebrad: command-line help info is hard to access and understand
## Description

Initially, I wanted to check the description of the `--verbose` option since one of our scripts in the Ziggurat project uses that option.

I had to find the zebrad's help output, but to find out what the help subcommand looks like, I just used an invalid subcommand:

```sh
➜  codecs git:(main) ✗ zebrad random-things
error: unrecognized command `random-things`

zebrad 1.0.0-rc.0
Zcash Foundation <zebra@zfnd.org>

FLAGS:
    -c, --config CONFIG       path to configuration file
    -h, --help                print help message
    -v, --verbose             be verbose
```

After seeing `    -h, --help                print help message`,  I used `--help` with no effect. Not only did I not get the help text, it even started the process.
```sh
➜  codecs git:(main) ✗ zebrad --help
2022-11-13T12:20:45.661166Z  INFO zebrad::components::tracing::component: started tracing component filter="info" TRACING_STATIC_MAX_LEVEL=LevelFilter::INFO LOG_STATIC_MAX_LEVEL=Info
2022-11-13T12:20:45.661203Z  INFO zebrad::application: Diagnostic Metadata:
version: 1.0.0-rc.0
Zcash network: Mainnet
state version: 25
branch: master
git commit: a4cb835
commit timestamp: 2022-10-12T01:17:55Z
target triple: aarch64-apple-darwin
build profile: release
[...]
```
Then I realized the command `zebrad help` would give me what I needed. And there I also saw that the `verbose` flag doesn't exist anymore.

### Steps to Reproduce

Run:
```sh
zebrad invalid-input
```
or
```sh
zebrad --helpppp
```
but not (could be considered as a separate issue):
```sh
zebrad --help # this one actually starts the zebrad process - it's probably the same root cause that causes this behaviour
```

### Expected Behaviour

I expected that the invalid subcommand input would result in the `zebrad help` help text output.

### Actual Behaviour

Instead, this happened:
I got the obsolete help text with the wrong information:
```
zebrad 1.0.0-rc.0
Zcash Foundation <zebra@zfnd.org>

FLAGS:
    -c, --config CONFIG       path to configuration file
    -h, --help                print help message
    -v, --verbose             be verbose
```

### Zebra Logs

N/A

### Zebra Version
```sh
➜  ~ zebrad version
zebrad 1.0.0-rc.0
```
### Operating System
```sh
➜  ~ uname -a
Darwin EQML220902-02 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:20:07 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T8110 arm64
```

## Designs

- Print the same information for invalid commands or options as is printed by the `help` command/flag

### LINKED ISSUE #5707: Improve consistency of usage information printed by zebrad `help` command
## Motivation

Print the long-form usage information for the help to match that which is printed with the --help flag

### Designs

- Copy Abscissa's `Help` Cmd into a new command module
- Move the code used to print out usage information for the `--help` flag into a fn of the new module (https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/application/entry_point.rs#L46)
- Call that fn instead of `print_usage_and_exit` from `HelpCmd::run` if no subcommand is provided

arya2:
## Motivation

This PR is needed to avoid a compilation error that is likely to be added to the Rust 2024 edition, to make our dependency tree smaller.

Related: This makes the output of `zebrad --help` consistent with that of `zebrad help`.

Closes #5502
Closes #5707
Closes #5624

### Major

- The `version` subcommand has been replaced with a `--version`/`-V` flag
- Zebra now accepts filters for the start command when no subcommand is provided

### Minor

- Zebra now ignores unrecognized arguments passed after a `--version` flag

## Solution

- Bumps `abscissa_core` from 0.5 to 0.7.0 in `zebrad/Cargo.toml`
- Replaces gumdrop with the latest version of clap
- Updates `ZebradApp`'s impl of `Application` to match new method signatures
- Updates commands to use `clap::Parser` instead of `gumdrop::Options`
- Uses `ColorChoice::Never` when initializing the updated `Terminal` component to avoid panicking when `color_spantrace::set_theme()` is called after it's been set
- Adds a `EntryPoint::process_cli_args()` method for inserting the default subcommand if one is not provided
- Moves `EntryPoint` from the application module to commands now that it's been excluded from `abscissa_core`
- Reverts https://github.com/ZcashFoundation/zebra/pull/6024

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- #6809


arya2:
```console
➜  zebrad help
zebrad 1.0.0-rc.8

Zcash Foundation <zebra@zfnd.org>

Usage: zebrad [OPTIONS] [FILTERS]... [COMMAND]

Commands:
  copy-state  The `copy-state` subcommand, used to debug cached chain state
  download    Pre-download required parameter files
  generate    Generate a skeleton configuration
  start       Start the application
  tip-height  Get the block height of Zebra's persisted chain state
  help        Print this message or the help of the given subcommand(s)

Arguments:
  [FILTERS]...
          tracing filters which override the zebrad.toml config

Options:
  -c, --config <CONFIG>
          path to configuration file

  -v, --verbose
          be verbose

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

zebrad/src/application.rs
```diff
@@ -503,25 +452,34 @@ impl Application for ZebradApp {
...
             Shutdown::Forced => process::exit(1),
             Shutdown::Crash => process::exit(2),
         }
     }
+}
 
-    fn version(&self) -> Version {
-        app_version()
-    }
+/// Boot the given application, parsing subcommand and options from
```
arya2, line 457:
```suggestion
// (See https://docs.rs/abscissa_core/latest/src/abscissa_core/application.rs.html#174-178)
/// Boot the given application, parsing subcommand and options from
```

.cargo/config.toml
```diff
@@ -57,13 +51,13 @@ rustflags = [
     # Panics
     "-Wclippy::fallible_impl_from",
     "-Wclippy::unwrap_in_result",
-
+
```
teor2345: Nitpick: please do a manual revert to avoid adding trailing whitespace, some editors delete it automatically.

teor2345:
Can you add the tickets that are closed by this PR to the PR description?

> * Replaces gumdrop with the latest version of clap
> * Updates commands to use `clap::Parser` instead of `gumdrop::Options`

Can you open a ticket to upgrade uses of structopt in other Zebra tools to the latest clap?

.github/workflows/continous-integration-os.yml
```diff
@@ -19,14 +19,11 @@ on:
       # production code and test code
       - '**/*.rs'
       # hard-coded checkpoints
-      # TODO: skip proptest regressions?
+      # TODO: skip proptest regressions
```
teor2345: This file should not be reverted, this was my mistake when writing the ticket. We need to keep commit efeab382bda32b8e22c3a92faa82fecd12607b9c and only revert parts of f6c74f1bd7d69a236476a941cc4ba23785f57d6f

.github/workflows/coverage.yml
```diff
@@ -35,8 +35,6 @@ on:
       - '**/*.snap'
       - '**/Cargo.toml'
       - '**/Cargo.lock'
-      - '.cargo/config.toml'
-      - '**/clippy.toml'
```
teor2345: This file should not be reverted, see above for details

.github/workflows/docs.yml
```diff
@@ -21,9 +21,6 @@ on:
       - '**/*.rs'
       - '**/Cargo.toml'
       - '**/Cargo.lock'
-      # configuration files
-      - '.cargo/config.toml'
-      - '**/clippy.toml'
```
teor2345: This file should not be reverted, see above for details

zebrad/src/application.rs
```diff
@@ -173,43 +136,49 @@ impl Application for ZebradApp {
...
-        command: &Self::Cmd,
+        _command: &Self::Cmd,
     ) -> Result<Vec<Box<dyn Component<Self>>>, FrameworkError> {
+        // TODO: Open a PR in abscissa to add a TerminalBuilder for opting out
+        //       of the `color_eyre::install` part of `Terminal::new` without
+        //       ColorChoice::Never?
+        //
         // Automatically use color if we're outputting to a terminal
         //
         // The `abcissa` docs claim that abscissa implements `Auto`, but it
```
teor2345: Is this still true for the 0.7 docs?

zebrad/src/application.rs
```diff
@@ -173,43 +136,49 @@ impl Application for ZebradApp {
...
-            //
-            // We'd also like to check `config.tracing.use_color` here, but the
-            // config has not been loaded yet.
-            if !Self::outputs_are_ttys() {
-                term_colors = ColorChoice::Never;
-            }
-        }
-        let terminal = Terminal::new(term_colors);
+        // does not.
+        // let mut term_colors = self.term_colors(command);
```
teor2345: The auditors gave us feedback that commented-out code was hard to review and maintain.

Instead, we could:
- open a ticket and give this code as an example
- have this code compiled but programmatically disabled by an `if false` (compiled but ignored at runtime) or `#[cfg(abscissa_terminal_fix)]` (parsed but not compiled)
- delete this code and let whoever fixes the TODO work it out

zebrad/src/application.rs
```diff
@@ -503,25 +452,34 @@ impl Application for ZebradApp {
...
+
             fatal_error(app_name, &e);
         }
-
-        // Swap out a fake app so we can trigger the destructor on the original
-        let _ = std::mem::take(self);
+        self.state()
+            .components_mut()
+            .get_downcast_mut::<Tracing>()
+            .map(Tracing::shutdown);
```
teor2345, line 442: Nitpick: If this code is moved above the `if`, it won't need to be duplicated.

zebrad/src/application.rs
teor2345: Nitpicks:
- In general, Rust documentation puts the purpose of the function at the top of the docs, then any references near the bottom
- Links are enclosed in angle brackets to avoid quoting issues: `<link>`

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
 use std::path::PathBuf;
 
 /// Zebrad Configuration Filename
 pub const CONFIG_FILE: &str = "zebrad.toml";
 
 /// Zebrad Subcommands
-#[derive(Command, Debug, Options)]
+#[derive(Command, Debug, clap::Subcommand)]
 pub enum ZebradCmd {
     /// The `copy-state` subcommand, used to debug cached chain state
```
teor2345:
```suggestion
    /// The `copy-state` subcommand, used to debug cached chain state (expert users only)
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
 
-    /// The `download` subcommand
-    #[options(help = "pre-download required parameter files")]
+    // The `download` subcommand
+    /// Pre-download required parameter files
     Download(DownloadCmd),
 
-    /// The `generate` subcommand
-    #[options(help = "generate a skeleton configuration")]
+    // The `generate` subcommand
```
teor2345: Nitpick: this comment effectively repeats what the code below says

```suggestion
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
 
-    /// The `help` subcommand
-    #[options(help = "get usage information, \
-        use help <subcommand> for subcommand usage information, \
-        or --help flag to see top-level options")]
-    Help(Help<Self>),
-
-    /// The `start` subcommand
-    #[options(help = "start the application")]
+    // The `start` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
-    /// The `download` subcommand
-    #[options(help = "pre-download required parameter files")]
+    // The `download` subcommand
+    /// Pre-download required parameter files
     Download(DownloadCmd),
 
-    /// The `generate` subcommand
-    #[options(help = "generate a skeleton configuration")]
+    // The `generate` subcommand
+    /// Generate a skeleton configuration
```
teor2345: "Skeleton" is an obscure term for this

```suggestion
    /// Generate a default `zebrad.toml` configuration
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
-    /// The `help` subcommand
-    #[options(help = "get usage information, \
-        use help <subcommand> for subcommand usage information, \
-        or --help flag to see top-level options")]
-    Help(Help<Self>),
-
-    /// The `start` subcommand
-    #[options(help = "start the application")]
+    // The `start` subcommand
+    /// Start the application
```
teor2345:
```suggestion
    /// Start the application (default command)
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
     Start(StartCmd),
 
-    /// The `tip-height` subcommand
-    #[options(help = "get the block height of Zebra's persisted chain state")]
+    // The `tip-height` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
+    /// Get the block height of Zebra's persisted chain state
```
teor2345:
```suggestion
    /// Print the tip block height of Zebra's chain state on disk
```

zebrad/src/commands.rs
```diff
@@ -73,27 +61,26 @@ impl ZebradCmd {
...
             // - is used by automated tools, or
             // - needs to be read easily.
-            Generate(_) | TipHeight(_) | Help(_) | Version(_) => true,
+            Generate(_) | TipHeight(_) => true,
 
             // Commands that generate informative logging output by default.
             CopyState(_) | Download(_) | Start(_) => false,
         };
 
-        // set to warn so that usage info is printed without info-level logs from component registration
```
teor2345, line 95: This comment seems useful, does it need to be updated for Zebra's new component initialisation behaviour?

zebrad/src/commands.rs
```diff
@@ -143,3 +128,132 @@ impl Configurable<ZebradConfig> for ZebradCmd {
...
+            .as_ref()
+            .expect("should default to start if not provided")
+    }
+
+    /// Returns a string that parses to the default subcommand
+    pub fn default_cmd_as_str() -> &'static str {
+        "start"
+    }
+
+    /// Process arguments from [`env::args_os()`] and insert the default subcommand
```
arya2:
```suggestion
    /// Process command arguments and insert the default subcommand
```

zebrad/src/commands.rs
```diff
@@ -143,3 +128,132 @@ impl Configurable<ZebradConfig> for ZebradCmd {
         }
     }
 }
+
+/// Toplevel entrypoint command.
```
teor2345: Moving this code from entry_point.rs and updating it at the same time makes it harder to review. 

Do you think it's worth moving it back and putting the fields and functions in the original order, to minimise the diff?

zebrad/src/commands/copy_state.rs
```diff
@@ -54,38 +55,43 @@ use crate::{
 /// How often we log info-level progress messages
 const PROGRESS_HEIGHT_INTERVAL: u32 = 5_000;
 
-/// `copy-state` subcommand
-#[derive(Command, Debug, Options)]
+// `copy-state` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands/copy_state.rs
```diff
@@ -54,38 +55,43 @@ use crate::{
...
+/// copy cached chain state (debug only)
```
teor2345:
```suggestion
/// copy cached chain state (expert users only)
```

zebrad/src/commands/download.rs
```diff
@@ -5,10 +5,11 @@
 //! This command should be used if you're launching lots of `zebrad start` instances for testing,
 //! or you want to include the parameter files in a distribution package.
 
-use abscissa_core::{Command, Options, Runnable};
+use abscissa_core::{Command, Runnable};
 
-/// `download` subcommand
-#[derive(Command, Debug, Default, Options)]
+// `download` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands/download.rs
```diff
@@ -5,10 +5,11 @@
...
+/// pre-download required parameter files
```
teor2345:
```suggestion
/// Pre-download required Zcash Sprout and Sapling parameter files
```

zebrad/src/commands.rs
```diff
@@ -5,59 +5,47 @@ mod download;
...
 pub enum ZebradCmd {
     /// The `copy-state` subcommand, used to debug cached chain state
     // TODO: hide this command from users in release builds (#3279)
-    #[options(help = "copy cached chain state (debug only)")]
     CopyState(CopyStateCmd),
 
-    /// The `download` subcommand
-    #[options(help = "pre-download required parameter files")]
+    // The `download` subcommand
+    /// Pre-download required parameter files
```
teor2345:
```suggestion
    /// Pre-download required Zcash Sprout and Sapling parameter files
```

zebrad/src/commands/generate.rs
```diff
@@ -1,13 +1,19 @@
 //! `generate` subcommand - generates a skeleton config.
```
teor2345:
```suggestion
//! `generate` subcommand - generates a default `zebrad.toml` config.
```

zebrad/src/commands/generate.rs
```diff
@@ -1,13 +1,19 @@
...
 
 use crate::config::ZebradConfig;
-use abscissa_core::{Command, Options, Runnable};
+use abscissa_core::{Command, Runnable};
+use clap::Parser;
 
-/// `generate` subcommand
-#[derive(Command, Debug, Options)]
+// `generate` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands/generate.rs
```diff
@@ -1,13 +1,19 @@
...
+/// generate a skeleton configuration
```
teor2345:
```suggestion
/// Generate a default `zebrad.toml` configuration
```

zebrad/src/commands/generate.rs
```diff
@@ -1,13 +1,19 @@
...
+#[derive(Command, Debug, Default, Parser)]
 pub struct GenerateCmd {
     /// The file to write the generated config to.
```
teor2345, line 10:
```suggestion
    /// The file to write the generated config to.
    //
    // TODO: use PathBuf here instead, to support non-UTF-8 paths
```

zebrad/src/commands/start.rs
```diff
@@ -91,17 +91,18 @@ use crate::{
     prelude::*,
 };
 
-/// `start` subcommand
-#[derive(Command, Debug, Options, Default)]
+// `start` subcommand
```
teor2345:
```suggestion
```

zebrad/src/commands/start.rs
```diff
@@ -91,17 +91,18 @@ use crate::{
...
+/// start the application
```
teor2345:
```suggestion
/// Start the application (default command)
```

zebrad/src/commands/tests.rs
```diff
@@ -0,0 +1,41 @@
+use clap::Parser;
```
teor2345, line 3: Please add a module comment at the top of this file

zebrad/src/commands/tests.rs
```diff
@@ -0,0 +1,41 @@
...
+use super::EntryPoint;
+
+#[test]
+fn args_with_subcommand_pass_through() {
+    let test_cases = [
+        (true, false, vec!["zebrad"]),
+        (true, true, vec!["zebrad", "-v"]),
+        (true, true, vec!["zebrad", "--verbose"]),
+        (true, false, vec!["zebrad", "-h"]),
+        (true, false, vec!["zebrad", "--help"]),
```
teor2345: Should these help cases use the start command? I'm pretty sure they used to just show the help.

zebrad/src/commands/tests.rs
```diff
@@ -0,0 +1,41 @@
...
+        assert!(args.cmd.is_some(), "args.cmd should not be none");
+        assert_eq!(
+            args.verbose, should_be_verbose,
+            "process_cli_args should preserve top-level args"
+        );
+
+        if should_be_start {
+            assert!(
+                matches!(args.cmd(), ZebradCmd::Start(_)),
+                "args.cmd should be start"
```
teor2345: Can you also test that the start command is not used when this is false?

zebrad/src/commands/tip_height.rs
```diff
@@ -15,17 +16,18 @@ use zebra_chain::{
 };
 use zebra_state::LatestChainTip;
 
-use crate::prelude::app_config;
+use crate::prelude::APPLICATION;
 
-/// `zebra-tip-height` subcommand
-#[derive(Command, Debug, Options)]
+// `zebra-tip-height` subcommand
```
teor2345: This is incorrect as well as being redundant
```suggestion
```

zebrad/src/commands/tip_height.rs
```diff
@@ -15,17 +16,18 @@ use zebra_chain::{
...
+/// get the block height of Zebra's persisted chain state
```
teor2345:
```suggestion
/// Print the tip block height of Zebra's chain state on disk
```

zebrad/src/components/tracing/component.rs
```diff
@@ -275,10 +275,16 @@ impl Tracing {
             initial_filter: filter,
             #[cfg(feature = "flamegraph")]
             flamegrapher,
-            _guard,
+            _guard: Some(worker_guard),
         })
     }
 
+    /// Drops guard for worker thread of non-blocking logger,
```
teor2345, line 282: I'm not sure we want to shutdown the log writer until the very last moment. See the note about the drop order of this field in the struct definition.

Shutting down here would stop other tasks logging shutdown messages, which could make debugging shutdowns harder.

zebrad/tests/acceptance.rs
```diff
@@ -264,11 +264,11 @@ fn help_no_args() -> Result<()> {
         is_zebrad_version,
         &output.output.stdout,
         "stdout",
-        "a valid zebrad semantic version",
+        "with a valid zebrad semantic version",
     )?;
 
     // Make sure we are in help by looking usage string
```
teor2345: Nitpick 

```suggestion
    // Make sure we are in help by looking for the usage string
```

zebrad/tests/acceptance.rs
```diff
@@ -264,11 +264,11 @@ fn help_no_args() -> Result<()> {
         is_zebrad_version,
         &output.output.stdout,
         "stdout",
-        "a valid zebrad semantic version",
+        "with a valid zebrad semantic version",
```
teor2345: Nitpick: This change isn't quite grammatically correct:
https://github.com/ZcashFoundation/zebra/blob/618d3fcca0f27a38ef0d7ce1c1d8e0436cc0117b/zebra-test/src/command.rs#L1129

```suggestion
        "are valid zebrad semantic versions",
```

zebrad/tests/acceptance.rs
```diff
@@ -522,7 +522,7 @@ fn version_no_args() -> Result<()> {
 
     let testdir = testdir()?.with_config(&mut default_test_config()?)?;
 
-    let child = testdir.spawn_child(args!["version"])?;
+    let child = testdir.spawn_child(args!["--version"])?;
```
teor2345, line 539: Removing the version command is a breaking change so there should be a changelog note for it.

As an alternative, add a version command that calls --version.

zebrad/tests/acceptance.rs
```diff
@@ -544,15 +544,18 @@ fn version_args() -> Result<()> {
     let testdir = testdir()?.with_config(&mut default_test_config()?)?;
     let testdir = &testdir;
 
-    // unexpected free argument `argument`
```
teor2345, line 561: You could use the tip-height command here instead of deleting this test. That command doesn't expect any free arguments.

The default config and an empty test directory should be fine, because this test is about argument parsing.

zebrad/tests/acceptance.rs
```diff
@@ -544,15 +544,18 @@ fn version_args() -> Result<()> {
...
-    let child = testdir.spawn_child(args!["version", "argument"])?;
-    let output = child.wait_with_output()?;
-    output.assert_failure()?;
-
     // unrecognized option `-f`
```
teor2345, line 561: This comment is out of date, it doesn't match the outcome of the test any more. Could tip-height be used here instead?

Allowing unrecognised arguments after version is potentially a breaking change, but it's minor.

teor2345 (commented):
Thanks, this looks like it should work, but I haven't tested it yet.

I'm having trouble reviewing the new entrypoint code, because it was moved and updated in this PR. Do you think it's worth minimising that diff?

Can you add a CHANGELOG entry with the list of breaking changes in this PR?

Feel free to tell me to submit a PR on top of this PR for cleanups. I'm happy for you just to ignore all those comments, or batch merge them in a single commit.

Just a note for revisions: please don't mark my comments as resolved. That makes it harder for me to work out which ones were actually resolved and which ones I still need to check.

## Follow up in other tickets

Would it help to add snapshot tests for quick commands and help/version output, so we don't accidentally change them in future? This should be a separate ticket.

teor2345:
> Related: This makes the output of `zebrad --help` consistent with that of `zebrad help`.

This is another minor breaking change that needs to be listed somewhere. Maybe the PR can have the full list, and we can just list the major breaking changes in the CHANGELOG with a link to the PR?

teor2345:
> * Will the PR name make sense to users?

Can you change the PR name to describe the user-visible changes, and put the implementation details in the PR?

zebrad/src/application.rs
arya2: I don't think so, it appears to implement `Auto` via termcolor, https://docs.rs/termcolor/1.2.0/termcolor/struct.StandardStream.html#method.stdout

zebrad/src/application.rs
```diff
@@ -503,25 +452,34 @@ impl Application for ZebradApp {
...
         }
     }
+}
 
-    fn version(&self) -> Version {
-        app_version()
-    }
+/// Boot the given application, parsing subcommand and options from
+/// command-line arguments, and terminating when complete.
+pub fn boot(app_cell: &'static AppCell<ZebradApp>) -> ! {
```
arya2, line 460:
```suggestion
// <https://docs.rs/abscissa_core/latest/src/abscissa_core/application.rs.html#174-178>
pub fn boot(app_cell: &'static AppCell<ZebradApp>) -> ! {
```

teor2345: Nitpick: these line numbers could go out of date as the latest abscissa version changes. We could link to the specific version we copied the code from, or name the specific function that people should be looking for in the comment.

zebrad/src/commands.rs
arya2: It's not necessary with the default subcommand being added outside the parsing logic.

This is maybe why clap doesn't support default subcommands.

zebrad/src/commands/tests.rs
arya2: > Should these help cases use the start command?

No, and they don't, [this test was wrong](https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b#diff-cacfd20083244388ead431a9d6c4997edae553b05a4d7d80901ee73bcd1eb85bR10).

Fixed in https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b.

.cargo/config.toml
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6801/commits/c01d33031d94060ab52308ad7bf7c6053b036ed8.

.github/workflows/continous-integration-os.yml
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6801/commits/c01d33031d94060ab52308ad7bf7c6053b036ed8.

zebrad/src/commands.rs
arya2: > Moving this code from entry_point.rs and updating it at the same time makes it harder to review.

Moved into its own file in https://github.com/ZcashFoundation/zebra/pull/6801/commits/82fb931bfd5d76c90a5dcd639b5804a72301793b.

Part of it is because of updates to the `Command` trait and part of it is from the [migration guide](https://github.com/iqlusioninc/abscissa/blob/main/CHANGELOG.md#migration-guide):

```rust
/// Entry point for the application. It needs to be a struct to allow using subcommands!
#[derive(Command, Debug, Clap)]
#[clap(author, about, version)]
pub struct EntryPoint {
    #[clap(subcommand)]
    // Replace this with you app name
    cmd: AppCmd,

    /// Enable verbose logging
    #[clap(short, long)]
    pub verbose: bool,

    /// Use the specified config file
    #[clap(short, long)]
    pub config: Option<String>,
}

impl Runnable for EntryPoint {
    fn run(&self) {
        self.cmd.run()
    }
}
```

zebrad/src/commands/tests.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b.

zebrad/src/commands/tests.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b.

zebrad/src/components/tracing/component.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b. Thank you.

It calls `components().shutdown()` - which calls `before_shutdown()` for each component -, `Tracing::shutdown`, then immediately exits the process.

`TokioComponent` handles interrupts/shutdown independently, TracingEndpoint and MetricsEndpoint don't have custom Drop impls, and `Tracing` should be okay with `before_shutdown()` + `shutdown()`.

zebrad/tests/acceptance.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/6801/commits/6b553d290759d64255389c43c1dfb6698313b28b.

zebrad/src/application.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/commit/6b553d290759d64255389c43c1dfb6698313b28b.

zebrad/src/application.rs
arya2: Deleted in https://github.com/ZcashFoundation/zebra/commit/6b553d290759d64255389c43c1dfb6698313b28b.

I doubt it'll be necessary for the TODO now that abscissa implements `Auto` via termcolor (albeit with different logic than we had).

zebrad/src/application.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/commit/6b553d290759d64255389c43c1dfb6698313b28b.

zebrad/tests/acceptance.rs
arya2: > This comment is out of date, it doesn't match the outcome of the test any more. Could tip-height be used here instead?

I used `tip-height` instead and added "is ignored" to the comment for this check in https://github.com/ZcashFoundation/zebra/commit/6b553d290759d64255389c43c1dfb6698313b28b.

> Allowing unrecognised arguments after version is potentially a breaking change, but it's minor.

I'll leave that in the PR description.

zebrad/tests/acceptance.rs
arya2: > Removing the version command is a breaking change so there should be a changelog note for it.

I'll add something to the changelog for now.

> As an alternative, add a version command that calls --version.

A suggestion PR for this would be very welcome.

teor2345: Fixing a breaking change that most users won't notice isn't a priority for me right now, I still have some stable release blockers to implement.

Feel free to open a ticket, but if it doesn't get done by the first stable release we should just close that ticket.

zebrad/src/commands.rs
teor2345: This PR changes the path of entry_point.rs, so I'll need to do a manual diff.

teor2345 (commented):
This PR changes the path of entry_point.rs, so I'll need to do a manual diff. I haven't had a chance to do that yet, I'll post an update when I've finished reviewing.

zebrad/src/application.rs
```diff
@@ -1,27 +1,24 @@
 //! Zebrad Abscissa Application
```
teor2345, line 1: This PR seems to have memory safety issues, or possibly this is another instance of bug #6812.

>   process didn't exit successfully: `/home/runner/work/zebra/zebra/target/release/deps/zebrad-ec601d4e69bdf498` (signal: 11, SIGSEGV: invalid memory reference)

https://github.com/ZcashFoundation/zebra/actions/runs/5151819470/jobs/9277360620?pr=6801#step:15:3194

zebrad/tests/acceptance.rs
arya2: Opened issue https://github.com/ZcashFoundation/zebra/issues/6823 to optionally follow up on this.

zebrad/src/application.rs
teor2345: We've confirmed that bug #6812 only happens on macOS x86_64 with Rust 1.70, and it's in `zebra-state`.

So a Linux bug on Rust 1.71-beta in `zebrad` seems like a different issue.

How did you want to move forward with this?

zebrad/src/commands/entry_point.rs
```diff
@@ -0,0 +1,138 @@
...
+        // Add the default subcommand to args after the top-level args if cmd is None
+        if should_add_default_subcommand {
+            // try_parse_from currently produces an error if the first argument is not the binary name,
+            let mut num_top_level_args = 1;
+
+            // update last_top_level_arg_idx to the number of top-level args
+            for (idx, arg) in args.iter().enumerate() {
+                num_top_level_args = match arg.to_str() {
+                    Some("--verbose" | "-v" | "--version" | "-V" | "--help") => idx + 1,
+                    Some("--config" | "-c") => idx + 2,
```
teor2345, line 81: This hard-coded list of special arguments seems fragile. Where does it come from?

Is there a way to look up this list, rather than listing it out in full?
Can we add tests to make sure each one of these arguments (and their alternative names) is handled correctly?

teor2345:
I opened this PR in case other reviewers wanted a smaller diff for `entry_point.rs`:
https://github.com/ZcashFoundation/zebra/pull/6826

zebrad/src/commands/entry_point.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            let mut updated_args: Vec<_> = args.iter().take(num_top_level_args).cloned().collect();
+            updated_args.push(EntryPoint::default_cmd_as_str().into());
+            updated_args.extend(args.into_iter().skip(num_top_level_args));
+
+            updated_args
+        } else {
+            args
+        }
```
teor2345: Nitpick: this seems like a complicated way to insert an item into a vector

```suggestion
            args.insert(num_top_level_args, EntryPoint::default_cmd_as_str().into());
        }

        args
```

zebrad/src/commands/entry_point.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /// Returns a string that parses to the default subcommand
+    pub fn default_cmd_as_str() -> &'static str {
+        "start"
+    }
+
+    /// Process command arguments and insert the default subcommand
+    /// if no subcommand is provided.
+    pub fn process_cli_args(args: Vec<OsString>) -> Vec<OsString> {
```
teor2345: Required for the simplification below

```suggestion
    pub fn process_cli_args(mut args: Vec<OsString>) -> Vec<OsString> {
```

zebrad/src/commands/entry_point.rs
arya2: > Where does it come from?

It's from the `help` output for top-level options.

> Is there a way to look up this list, rather than listing it out in full?

That may require a macro, we could delete this altogether if we open a PR in abscissa to take default subcommands.

> Can we add tests to make sure each one of these arguments (and their alternative names) is handled correctly?

Yeah, I'll add some test cases to `args_with_subcommand_pass_through()`.

zebrad/src/application.rs
arya2: > How did you want to move forward with this?

I think it happened in `args_with_subcommand_pass_through()`, I'm hoping it'll happen again for the most recent commit as a starting point.

zebrad/src/commands/entry_point.rs
teor2345: Does `try_parse_from()` guarantee that each `-c` is followed by another argument?
Please add a test case where it's not.

zebrad/src/commands/entry_point.rs
```diff
@@ -0,0 +1,134 @@
...
+            // update last_top_level_arg_idx to the number of top-level args
+            for (idx, arg) in args.iter().enumerate() {
+                num_top_level_args = match arg.to_str() {
+                    Some("--verbose" | "-v" | "--version" | "-V" | "--help") => idx + 1,
+                    Some("--config" | "-c") => idx + 2,
+                    _ => num_top_level_args,
+                }
+            }
+
+            args.insert(num_top_level_args, EntryPoint::default_cmd_as_str().into());
```
teor2345, line 87: Here's another possible fix: if the argument calculation is wrong, just add `start` as the final argument.

```suggestion
            num_top_level_args = min(num_top_level_args, args.len());
            args.insert(num_top_level_args, EntryPoint::default_cmd_as_str().into());
```

zebrad/src/commands/entry_point.rs
```diff
@@ -0,0 +1,135 @@
+//! Zebrad EntryPoint
+
```
teor2345, line 2:
```suggestion

use std::cmp::min;

```

zebrad/src/application.rs
teor2345: This has been fixed by making sure that the index is always within the vector.

zebrad/src/commands/entry_point.rs
teor2345: I'd like to get these changes in the final release candidate, so I'm going to approve this PR.

Feel free to add extra tests as another PR or a separate ticket.

teor2345 (dismissed):
This seems to be working well enough for now.

arya2:
In the merge queue:
>   process didn't exit successfully: `/home/runner/work/zebra/zebra/target/release/deps/zebrad-4d9e827350fddfd8` (signal: 11, SIGSEGV: invalid memory reference)
> Error: Process completed with exit code 101.
>
> https://github.com/ZcashFoundation/zebra/actions/runs/5194379481/jobs/9365975834#step:15:3531

zebrad/src/commands/tests.rs
```diff
@@ -0,0 +1,38 @@
+//! Tests for parsing zebrad commands
+
+use clap::Parser;
+
+use crate::commands::ZebradCmd;
+
+use super::EntryPoint;
+
+#[test]
+fn args_with_subcommand_pass_through() {
```
teor2345, line 10: It seems like this test is occasionally failing with:
> process didn't exit successfully: `/home/runner/work/zebra/zebra/target/release/deps/zebrad-4d9e827350fddfd8` (signal: 11, SIGSEGV: invalid memory reference)

https://github.com/ZcashFoundation/zebra/actions/runs/5194379481/jobs/9365975834#step:15:3532

arya2: I think it was because of the `err.exit()` in `process_cli_args()`.

Possibly fixed by https://github.com/ZcashFoundation/zebra/pull/6801/commits/85b26149ef46cd02b4c3c637f49ca11b337b9a47.

teor2345: It looks like the output of `zebrad help` has been corrupted:

```
Zcash Foundation <zebra@zfnd.org>
brad [OPTIONS] [FILTERS]... [COMMAND]
Commands:
```
https://github.com/ZcashFoundation/zebra/actions/runs/5194379481/jobs/9365975834#step:15:3509

But when I run the test on my x86_64 Linux and M1 macOS I see:
```
Zcash Foundation <zebra@zfnd.org>
Usage: zebrad [OPTIONS] [FILTERS]... [COMMAND]
Commands:
```

I have done thousands of test runs that succeed, with stable Rust 1.70 and nightly Rust 1.72.

teor2345: The previous failure had slightly different memory corruption:

```
Zcash Foundation <zebra@zfnd.org>
sage: zebrad [OPTIONS] [FILTERS]... [COMMAND]
Commands:
```

https://github.com/ZcashFoundation/zebra/actions/runs/5194379481/jobs/9365975834#step:15:3509

teor2345: I don't know if the memory corruption is diagnostic, I think it's also happening in successful test runs:
https://github.com/ZcashFoundation/zebra/actions/runs/5192882552/jobs/9362667847#step:15:3167

Maybe we should add a snapshot test for the usage output to catch bugs like this. (But in another ticket/PR.)

arya2: > I have done thousands of test runs that succeed, with stable Rust 1.70 and nightly Rust 1.72.

I haven't seen it locally either but I think it happens pretty often in CI, the passing tests were skipped.

> Maybe we should add a snapshot test for the usage output to catch bugs like this

This should be part of https://github.com/ZcashFoundation/zebra/issues/6809

teor2345: It seems like exiting the test process without warning could cause memory issues and output corruption, so I'm happy to approve this and see if it passes.

If we see any memory errors in the zebrad lib tests after it merges, let's revert this PR.

teor2345 (approved):
Conditional approval: if we see any memory errors in the zebrad lib tests, we should revert this PR.

Issue #6568, opened 2023-04-25
fix(ci): Add another job to the full sync test


teor2345:
## Motivation

The full sync test is failing on the main branch:
> 2023-05-08T04:04:59.428138Z  INFO {net="Main"}:sync:try_to_sync:extend_tips:checkpoint: zebra_consensus::checkpoint: verified checkpoint range block_count=169 current_range=(Excluded(Height(2053210)), Included(Height(2053379)))
> 2023-05-08T04:05:20.967223Z  INFO {net="Main"}: zebrad::components::sync::progress: estimated progress to chain tip sync_percent=98.747% current_height=Height(2052566) network_upgrade=Nu5 remaining_sync_blocks=26050 time_since_last_state_block=0s
> Error: The operation was canceled.

https://github.com/ZcashFoundation/zebra/actions/runs/4888643906/jobs/8765973268#step:8:1845

Here is the latest full sync, it is also timing out, even in the new environment:
https://github.com/ZcashFoundation/zebra/actions/runs/4931890966/jobs/8832449145

### How to do this fix

Follow the "Fixing CI Sync Timeouts" instructions, starting at step 4:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-ci-sync-timeouts

The job has been failing for at least a week, and the checkpoints were recently updated. So we have already done steps 1-3.

### Testing

Wait for the next full sync to finish successfully after this PR is merged.

teor2345:
I talked with Gustavo, we're going to switch to `us-east` later this week, so we'll see how it goes after that.

teor2345:
Now we're in the new environment, we need to check that this works.

Issue #5709, opened 2022-11-23
Fix repeated block timeouts during initial sync


teor2345:
## Motivation

During the initial sync, Zebra's block validation repeatedly times out, and the syncer resets:
> 2022-11-23T09:22:15.197732Z  WARN {net="Main"}:sync:try_to_sync: zebrad::components::sync: error downloading and verifying block e=ValidationRequestError { error: Elapsed(()), height: Height(1856228), hash: block::Hash("0000000000d433e30155445ddc72854eb7395827c1c241b249a255585ea2716a") }

> https://github.com/ZcashFoundation/zebra/actions/runs/3519147385/jobs/5923785413#step:8:12498

This only happens in some parts of the chain, and it seems to happen at slightly different block heights each time. When it does happen, it can happen repeatedly every 10-15 minutes:
https://github.com/ZcashFoundation/zebra/actions/runs/3519147385/jobs/5923785413#step:8:12465

This is not a priority because Zebra is still much faster than `zcashd`.

### Designs

This could be happening because:
- blocks are committed out of order
- we're downloading blocks in the wrong order (not strict height order) then hitting a concurrency limit so we can't download the rest before the timeout
- the block timeout is too short for the number of blocks allowed in the validate and commit pipeline
  - decreasing the pipeline capacity also reduces RAM use, but makes runs of small blocks download and verify slower
  - increasing the timeout increases the amount of time we wait for blocks that will never verify, increasing RAM usage
  - we tried increasing the timeout and it didn't work, but a slight increase might be ok
- our state isn't responding to all the UTXOs for the blocks it has (I think this is a known issue)
- the verifier is blocked in some other way

## Related Work

If we speed up block commits, it might solve this issue:
- #5425

teor2345:
Due to this block verification timeout bug, full syncs are currently taking 40-48 hours. That's still acceptable performance, I think?

mpguerra:
Is this still happening? if so, when was the last time we saw it?

teor2345:
> Is this still happening? if so, when was the last time we saw it?

Yes, it happens multiple times in every full sync, this sync is running right now:
https://github.com/ZcashFoundation/zebra/actions/runs/4931890966/jobs/8824296617#step:8:8745

In the last completed full sync, it happened at least 200 times, which causes a delay of 3-5 hours in  50 hour sync:
https://github.com/ZcashFoundation/zebra/actions/runs/4888643906/jobs/8769200765#step:8:17512

[...]

Issue #6793, opened 2023-05-30
Do the state renames that were not done in #6762


oxarbitrage:
## Motivation

In #6762 we renamed a lot of stuff from the tasklist in https://github.com/ZcashFoundation/zebra/issues/6680

However there are more things we want to rename, for example some methods, but we don't want #6762 to keep growing. So here is a new task list to do after #6762 gets merged.

The verification parts of the request, state service, and consensus rule checks use incorrect terms that are very confusing for developers and auditors. The storage parts use those terms correctly and should be left alone.

These renames can be implemented and reviewed easily using a `sed` script:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/mass-renames.md#universal-renames-with-sed

### Possible Renames

Checkpoint Blocks in zebra-consensus:
- [x] Rename `VerifyCheckpointError::CommitFinalized` variant to `CommitCheckpointVerified`
- [x] Rename `commit_finalized_block` variable to `commit_checkpoint_verified`

Finalized Queues:
- [x] Rename `drain_queue_and_commit_finalized` method to `drain_finalized_queue_and_commit`
- [x] Rename `queue_and_commit_finalized` method to `queue_and_commit_to_finalized_state`
- [x] Rename `queued_finalized_blocks` to `finalized_state_queued_blocks`
- [x] Replace "queued finalized block" with "finalized state queue block" in comments and logs
- [x] Rename `max_queued_finalized_height` to `max_finalized_queue_height`
- [x] Rename `send_finalized_block_error` method to `send_checkpoint_verified_block_error`
- [x] Rename `QueuedFinalized` to `QueuedCheckpointVerified`

Non-Finalized Queues:
- [x] Rename `queue_and_commit_non_finalized` to `queue_and_commit_to_non_finalized_state`
- [x] Rename `queued_non_finalized_blocks` to `non_finalized_state_queued_blocks`
- [x] Replace "queued non-finalized block" with "non-finalized state queue block" in comments and logs
- [x] Rename `send_non_finalized_block_error` to `send_semantically_verified_block_error`
- [x] Rename `QueuedNonFinalized` to `QueuedSemanticallyVerified`

Channels:
- [x] Rename `last_sent_finalized_block_hash` to `finalized_block_write_last_sent_hash`
- [x] Rename `sent_non_finalized_block_hashes` to `non_finalized_block_write_sent_hashes`
[...]

Issue #6681, opened 2023-05-15
Update documentation to clarify terminology around block verification


mpguerra:
## Motivation

The verification parts of the request, state service, and state consensus rule checks use incorrect terms that are very confusing for developers and auditors.

The storage parts use those terms correctly and should be left alone.

See the comments of this ticket for potential changes.

### Background Notes

Here is some of the confusion in this auditor note that we should resolve in our documentation:

#### Checkpoint vs Finalized

> Note that a checkpointed/finalized block can be a settled network upgrade or a block beyond the rollback/reorg limit. The state service processes the finalized block and queues it to be committed to state

This is the core of the confusion:
- a settled network upgrade is specified as a network upgrade that has "social consensus": it has enough confirmations on enough nodes that it can't be rolled back
- checkpointed blocks are a Zebra implementation detail: every release, we generate checkpoints for all finalized blocks in the chain, because those blocks can't be rolled back by Zebra, and are very unlikely to be be rolled back by `zcashd`. This includes settled network upgrades, but it also goes beyond them to include almost all settled blocks.
- finalized blocks can come from checkpoints, or from a contextually verified block which is beyond the rollback limit
- the state service processes:
    - checkpoint verified blocks direct to the finalized state
    - contextually verified blocks into the non-finalized state, and then into the finalized state once they reach the rollback limit (or if they are on a side chain that will never be committed to the finalized state, they get dropped)

#### Verified Block Cache Behaviour

> (finalized blocks can be optionally saved to disk.)

All valid finalized blocks are saved to disk, that's what finalization means in Zebra. Saving to disk is not under user control, and it's not a choice the application makes ~~automatically~~ dynamically. The only thing that can stop a finalized block being saved to disk is invalid data, which causes a verification failure in the block commitment, authorization, or chain history roots.

Non-finalized blocks are saved to disk if they are part of the best chain, and beyond the rollback limit. If they are not part of the best chain and beyond the rollback limit, they are dropped.

##### User Confusion about Caching

There's also some potential user confusion here between:
* thin clients: programs that don't download or store the whole chain, and skip some validation
* full nodes: programs that download the whole chain, fully validate it, and store it

We've had multiple requests for Zebra to become a thin client, so it's important for us to avoid confusion in our user documentation.

[...]

PR #6810, opened 2023-06-02
fix(state): Avoid temporary failures verifying the first non-finalized block or attempting to fork the chain before the final checkpoint

### LINKED ISSUE #5125: Avoid temporary failures verifying the first non-finalized block

### LINKED ISSUE #6388: can_fork_chain_at() should ignore blocks below the finalized tip
## Motivation

We can't fork the chain at a finalized block height below the finalized tip height, but the `sent_non_finalized_blocks` data structure contains the heights of some finalized blocks close to the final checkpoint.

We might need to store the queued finalized block list separately, rather than re-using an existing data structure that other code depends on.

(This is an existing bug, but the code in the linked audit PR added more finalized blocks to the data structure.)

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/6335#discussion_r1144234829_

## Scheduling

We reviewed this code, and checked that even if it was called incorrectly, the write block task would return a similar error. So this is just a technical debt cleanup.

arya2:
## Motivation

This PR updates the `can_fork_chain_at()` method in zebra-state to ignore blocks below the finalized tip and fixes a bug where the first block past the final checkpoint will fail verification if it arrives in zebra-state before the final checkpoint block is written to disk.

Close #6388
Close #5125

Part of #5709

## Solution

- Add a `can_fork_chain_at_hashes` flag to `SentHashes`
- Add and use a `can_fork_chain_at(hash)` method to `SentHashes` that checks the new flag
- When closing finalized block write channel:
  - Replace `block_write_sent_hashes` with an empty `SentHashes` and mark `can_fork_chain_at_hashes` as true 
  - Start committing queued non-finalized child blocks of finalized tip

Related:

- Adds `shrink_to_fit()` calls in `prune_by_height()`

## Review

This is low-priority.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -707,7 +707,8 @@ impl StateService {
             // Tell the block write task to stop committing checkpoint verified blocks to the finalized state,
             // and move on to committing semantically verified blocks to the non-finalized state.
             std::mem::drop(self.finalized_block_write_sender.take());
-            self.prune_by_height();
+            // Remove any checkpoint-verified block hashes from `block_write_sent_hashes`.
+            self.block_write_sent_hashes = SentHashes::default();
             // Mark `SentHashes` as usable by the `can_fork_chain_at()` method.
             self.block_write_sent_hashes.can_fork_chain_at_hashes = true;
 
@@ -729,27 +730,19 @@ impl StateService {
         if self.finalized_block_write_sender.is_none() {
             // Wait until block commit task is ready to write non-finalized blocks before dequeuing them
             self.send_ready_non_finalized_queued(parent_hash);
-            self.prune_by_height();
+
+            let finalized_tip_height = self.read_service.db.finalized_tip_height().expect(
+                "Finalized state must have at least one block before committing non-finalized state",
+            );
+
+            self.non_finalized_state_queued_blocks
+                .prune_by_height(finalized_tip_height);
+
+            self.block_write_sent_hashes
+                .prune_by_height(finalized_tip_height);
         }
 
         rsp_rx
-    }
-
-    /// Prunes queued blocks and sent hashes up to the finalized tip height.
-    ///
-    /// # Panics
-    ///
-    /// If called when the finalized state has no blocks.
-    fn prune_by_height(&mut self) {
-        let finalized_tip_height = self.read_service.db.finalized_tip_height().expect(
-            "Finalized state must have at least one block before committing non-finalized state",
-        );
-
-        self.non_finalized_state_queued_blocks
-            .prune_by_height(finalized_tip_height);
-
-        self.block_write_sent_hashes
-            .prune_by_height(finalized_tip_height);
     }
 
     /// Returns `true` if `hash` is a valid previous block hash for new non-finalized blocks.
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -158,7 +158,7 @@ pub(crate) struct StateService {
 
     /// A set of block hashes that have been sent to the block write task.
     /// Hashes of blocks below the finalized tip height are periodically pruned.
-    block_write_sent_hashes: SentHashes,
+    non_finalized_block_write_sent_hashes: SentHashes,
 
     /// If an invalid block is sent on `finalized_block_write_sender`
     /// or `non_finalized_block_write_sender`,
@@ -405,7 +405,7 @@ impl StateService {
             non_finalized_block_write_sender: Some(non_finalized_block_write_sender),
             finalized_block_write_sender: Some(finalized_block_write_sender),
             finalized_block_write_last_sent_hash,
-            block_write_sent_hashes: SentHashes::default(),
+            non_finalized_block_write_sent_hashes: SentHashes::default(),
             invalid_block_write_reset_receiver,
             pending_utxos,
             last_prune: Instant::now(),
@@ -468,7 +468,7 @@ impl StateService {
         // If we're close to the final checkpoint, make the block's UTXOs available for
         // semantic block verification, even when it is in the channel.
         if self.is_close_to_final_checkpoint(queued_height) {
-            self.block_write_sent_hashes
+            self.non_finalized_block_write_sent_hashes
                 .add_finalized(&checkpoint_verified)
         }
 
@@ -647,7 +647,7 @@ impl StateService {
         let parent_hash = semantically_verrified.block.header.previous_block_hash;
 
         if self
-            .block_write_sent_hashes
+            .non_finalized_block_write_sent_hashes
             .contains(&semantically_verrified.hash)
         {
             let (rsp_tx, rsp_rx) = oneshot::channel();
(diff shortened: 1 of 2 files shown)


zebra-state/src/service/queued_blocks.rs
```diff
@@ -219,18 +219,49 @@ impl QueuedBlocks {
     }
 }
 
+#[derive(Debug)]
```
teor2345: Please derive the standard set of impls on new types, so they are easier to use in future code changes.

teor2345 (dismissed):
"non-finalized" and "finalized" are outdated terms that we're trying to move away from, because they were very confusing during the audit. When we're talking about the verification status of blocks, we use "semantically verified", "checkpoint verified" or "contextually verified".

But it's ok to say "finalized queue" when talking about block storage, or which state blocks are queued for.

There's a list of name changes in ticket #6793 which might be helpful.

Edit: name changes have finished

zebra-state/src/service/queued_blocks.rs
```diff
@@ -219,18 +219,49 @@ impl QueuedBlocks {
...
+
+    /// Stores a set of hashes that may have been sent to the block write task via
+    /// the finalized block write channel but may not be in the finalized state yet.
+    sent_finalized: HashMap<block::Hash, Vec<transparent::OutPoint>>,
 
-    /// Stores a set of hashes that have been sent to the block write task but
-    /// may not be in the finalized state yet.
-    sent: HashMap<block::Hash, Vec<transparent::OutPoint>>,
+    /// Stores a set of hashes that have been sent to the block write task via the
+    /// non-finalized block write channel but may not be in the finalized state yet.
```
teor2345: Which state do you mean here?

zebra-state/src/service/queued_blocks.rs
```diff
@@ -314,30 +347,41 @@ impl SentHashes {
...
-        self.sent.shrink_to_fit();
+        self.sent_finalized.shrink_to_fit();
+        self.sent_non_finalized.shrink_to_fit();
+        self.known_utxos.shrink_to_fit();
 
         self.update_metrics_for_cache();
     }
 
     /// Returns true if SentHashes contains the `hash`
     pub fn contains(&self, hash: &block::Hash) -> bool {
```
teor2345: These docs and method name are now outdated

zebra-state/src/service.rs
```diff
@@ -720,28 +720,31 @@ impl StateService {
...
-        rsp_rx
-    }
+        self.sent_non_finalized_block_hashes
+            .prune_by_height(finalized_tip_height);
 
-    /// Returns `true` if `hash` is a valid previous block hash for new non-finalized blocks.
-    fn can_fork_chain_at(&self, hash: &block::Hash) -> bool {
-        self.sent_non_finalized_block_hashes.contains(hash)
-            || &self.read_service.db.finalized_tip_hash() == hash
+        self.sent_non_finalized_block_hashes.contains(hash) || &finalized_tip_hash == hash
```
teor2345: This field name is now outdated for two reasons:
1. It uses old verification status terminology
2. The type contains both checkpointed and semantically verified blocks

zebra-state/src/service.rs
```diff
@@ -720,28 +720,31 @@ impl StateService {
...
-            // Wait until block commit task is ready to write non-finalized blocks before dequeuing them
-            self.send_ready_non_finalized_queued(parent_hash);
+        self.send_ready_non_finalized_queued(parent_hash);
 
-            let finalized_tip_height = self.read_service.db.finalized_tip_height().expect(
+        rsp_rx
+    }
+
+    /// Returns `true` if `hash` is a valid previous block hash for new non-finalized blocks.
+    fn can_fork_chain_at(&mut self, hash: &block::Hash) -> bool {
```
teor2345: This PR changes syncing behaviour, can you run a full sync test before we merge it?
You'll have to do it manually, or get Gustavo to fix #6568 first.

zebra-state/src/service/queued_blocks.rs
arya2: The finalized state, I'll add some context on how it's used in `can_fork_chain_at()`.

I think I could just add a flag to this struct and simplify this PR.

teor2345:
```suggestion
    /// non-finalized block write channel, but haven't been committed to the finalized state yet.
    /// (They are either in the channel, or in the non-finalized state.)
```

teor2345: I wasn't asking for a code change here, I just wanted to check that "finalized state" wasn't a copy-paste error.

zebra-state/src/service.rs
teor2345: > 1. It uses old verification status terminology

~~Let's do these name changes automatically in ticket #6793?
Any complicated documentation changes can be done in this PR, or added to the list in #6681.~~

Edit: we've finished name changes

zebra-state/src/service.rs
teor2345: I fixed #6568, I hope!

teor2345: Since there's a sync risk, this PR should go at the end of the PR backlog.

teor2345 (dismissed):
Just marking this as needing changes.

teor2345:
I updated my review because the name changes in ticket #6793 have merged.

(There's no need to work on this PR any time soon, please check with Pili first.)

zebra-state/src/service/queued_blocks.rs
arya2: > I wasn't asking for a code change here

I know - but I think I initially overcomplicated this, I updated it to prune out all the finalized sent hashes and mark a `can_fork_chain_at_hashes` flag as true. It should actually just drop it and replace it with a new one since it'll always be emptied anyways. ~I'll do this in a follow-up PR.~

zebra-state/src/service.rs
arya2: Mainnet sync started here: https://github.com/ZcashFoundation/zebra/actions/runs/5661476853

Update: Finished in 2d 4h, last full sync time I'd seen was 3d 20h.

zebra-state/src/service/queued_blocks.rs
arya2: I started over and removed this struct, and noted.

zebra-state/src/service.rs
arya2: Should this PR still update this field name?

arya2:
@Mergifyio update

arya2:
@Mergifyio update

oxarbitrage (commented):
I think this looks good, is short, clear and it seems to work.

What i am not totally sure is if it closes all the stuff that are inside the related tickets.

I will like @teor2345 to make the final review as they have more context in the tickets this is closing. I think the next few days could be a good time to merge this in.

zebra-state/src/service.rs
teor2345: Ok great that seems fine.

The scheduled full syncs usually complete in just over 2 days.

teor2345 (approved):
Thanks for this fix!

> What i am not totally sure is if it closes all the stuff that are inside the related tickets.

That's a good question.

#5125 is partially resolved, I opened #7352 with the remaining work. (Which I don't think we should fix unless it actually happens.)

I checked the logs and #5709 is not resolved, it seems to be just as frequent as before this fix. (The warnings are still there, and syncing takes about the same amount of time.)

Issue #6642, opened 2023-05-09
Tracking: Update `lightwalletd` RPCs for "spend before sync" algorithm


teor2345:
## Motivation

The `zcashd` and `lightwalletd` teams have been working on a "spend before sync" algorithm, targeted for `zcashd` 5.6.0 in June 2023.

As part of that change, the JSON-RPC interface used by `lightwalletd` will need extra methods or fields.

Preparation:

<details>
<summary>completed preparation</summary>

- [x] #6746
- [x] #6955
- [x] #6859 (Update `incrementalmerkletree` crate to 0.4.0)
    - [x] #7143

- [x] #7292 
    - [x] #7390
    - [x] #7349

</details>

Zebra will need to implement some of those changes to remain compatible with the latest `lightwalletd` versions:
- [x] #6952
- [x] #6953
    - [x] #7532
    - [x] #7407
    - [x] #7570
- [x] #6954 
    - [x] #7449
        - [x] #7446
        - [x] #7450 
        - [x] #7451 
        - [x] #7513

Extra Dependencies & Tests:

<details>

<summary>completed</summary>

- [x] #7201
- [x] #6638
- [x] #7335

</details>

### Lower Priority

Minor bug or test fixes:
- [x] #7443
- [x] #7447 
- [x] #7461
- [x] #7444
- [x] #7351
- [x] #7463

<details>
<summary>completed performance & documentation</summary>

To get reasonable lookup performance:
- [x] #4784

Documentation:
- [x] #7215
- [x] #7319
    - [x] #7370

</details>

Optional Cleanups & Fixes:

<details>
<summary>these tickets are optional, ask Pili before starting them</summary>

Bug fixes:
- [ ] #7392
- [x] #7263 

Technical debt:
- [x] #7383
- [ ] #7377

Diagnostics:
- [x] #7369 
- [x] #7416

Needs decision/triage:
- [x] #7441

</details>

### Timing

`zcashd` aims to finalize the RPC design by the end of May, and then release it in the first two weeks of June.

`lightwalletd` will start calling the new RPCs around the same time, but they aim to be compatible with old RPC versions using an off-by-default flag.

### Implementation Notes

Zebra already stores the data we need for these lookups. We might need extra indexes for good performance, but that's not a breaking state change. (So we don't need to upgrade the state version, we can just automatically add/update a new column family with the new index at startup.)

We will need to implement #4784 to have acceptable note commitment tree scanning performance.

[...]

Issue #4784, opened 2022-07-17
Store only the first tree state in each series of identical tree states


upbqdn:
Storing only the first tree state in each identical series of tree states saves memory, and makes `Chain` cloning cheaper.

### Possible Implementation

We can access the deduped trees using:

Non-finalized state:
```rust
Chain.{sapling,orchard}_note_commitment_tree_by_height.range(..=target_height).next_back()
```

Finalized state (reading is implemented, writing new blocks and updating cached blocks is not):
```rust
db.zs_prev_key_value_back_from({sapling,orchard}_note_commitment_tree, target_height)
```
### Performance

This needs a full sync performance test:
> The performance of reversed iteration is usually much worse than forward iteration
https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ

teor2345:
This is important to do eventually because large blocks take up more space on disk. But we don't need to do it any time soon.

teor2345:
Performance and disk space usage are not a priority for Zebra right now.

teor2345:
This is required for the note commitment tree scanning in #6642 to have reasonable performance.

teor2345:
Note from the engineering sync: This seems risky to do between the last release candidate and the first stable release. There is no specific deadline for this change.

teor2345:
> Finalized state:
> 
> ```rust
> db.iterator_cf({sapling,orchard}_note_commitment_tree, IteratorMode::From(target_height, Direction::Reverse)).next()
> ```

The database read part of this ticket is implemented in PR #6813.

Everything else still needs to be implemented:
- non-finalized state data structure deduplication and lookups
- finalized state deduplication for newly verified blocks
- finalized state deduplication for existing blocks on disk

PR #6813 introduces state minor versions and the idea of a state format update task. But I didn't create the actual task in that PR. I'm happy to help out with that change, currently it's just a vague idea of a design, and a bunch of TODOs.

teor2345:
@upbqdn lets catch up some time and work out how to make these database changes without breaking compatibility?

mpguerra:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/4784?planning-poker) with Zenhub @arya2

teor2345:
Here's a draft checklist for state format changes, let me know if it works for you!

[...]

Issue #6763, opened 2023-05-24
Intermittent long delays to inbound peer connection handshakes, Credit: Ziggurat Team


teor2345:
## Motivation

The Ziggurat team have reported some long delays (25-40 seconds) in Zebra's inbound connection handler.

The most likely cause is running blocking code inside async futures on the same thread or task as the connection handler.

The amount of code between accepting an inbound connection and sending a version message is quite small. It's possible the async mutex in the nonce set or reading the state tip from the watch channel are causing delays, but it's unlikely.

### Next Diagnostic Steps

Run Zebra under `strace` or `dtrace` to work out what it's waiting on during those 40 seconds.
Run Zebra with `tokio-console` to see what the blocking task is.

### Complex Code or Requirements

This seems like a concurrency bug.

### Testing

Manually start up Zebra and check how long it takes to accept inbound connections.

## Related Work

This issue is possibly related to:
- #6911
- #6506 
- #5709
- #3288
- #3234

### Discussions of this bug

- https://forum.zcashcommunity.com/t/ziggurat-3-0/43350/31
- https://discord.com/channels/809218587167293450/809251029673312267/1111072914624757770
- private communication to avoid leaking user IP addresses

teor2345:
The Ziggurat team have done some more testing:

> So one more small related finding - we've set up a synth node which sends `GetAddr` message requests periodically every few seconds.
> 
> What we have also observed in our active connection - once the zebrad enters this stuck state (where it can't handle new inbound connections for a while) - it also stops replying to our `GetAddr` messages during the same period on this other active connection, we have established.

This suggests that either the inbound service is blocked, or the entire async thread or executor is blocked.

arya2:
I think it's the codec in some cases.

Tasks spawned with `task::block_in_place()` are queued for execution if there are already `max_blocking_threads` on the runtime even if every spawned blocking thread is idle.

teor2345:
> I think it's the codec in some cases.
> 
> Tasks spawned with `task::block_in_place()` are queued for execution if there are already `max_blocking_threads` on the runtime even if every spawned blocking thread is idle.

Do you mean "blocked", rather than "idle"?

Blocking threads are designed to be blocked on network, filesystem, or other blocking operations, so they should never use much CPU. (That's why we use rayon.)

Have you checked the queue depth? About how long is it?
https://docs.rs/tokio/latest/tokio/runtime/struct.RuntimeMetrics.html#method.blocking_queue_depth

[...]

PR #6699, opened 2023-05-16
fix(security): Stop panicking on state RPC or block requests with very large heights

teor2345:
## Motivation

We have a few panics that can be triggered remotely by rpc methods or blocks with very large heights.

Close #6595.

## Solution

This fixes the panics in `getblock`, `z_gettreestate`, `getblockhash`, and committing blocks to the state.

The following calls do not panic anymore.

- ` curl --silent --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblock", "params": ["2147483647", 1] }' -H 'Content-type: application/json' http://127.0.0.1:6666/ | jq`
- `curl --silent --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblock", "params": ["2147483647", 0] }' -H 'Content-type: application/json' http://127.0.0.1:6666/ | jq`
- `curl --silent --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "z_gettreestate", "params": ["2147483647"] }' -H 'Content-type: application/json' http://127.0.0.1:6666/ | jq`

### Not Impacted

`getaddresstxids` is not affected by this issue because it returns another error first, but it was fixed anyway.
`getblocksubsidy` is not affected because it doesn't use the state.

`get_network_sol_ps` and `get_network_hash_ps` have a different bug that will be fixed in #6688.

## Review

@oxarbitrage (or anyone else) can review the new tests in this PR.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

- #6688

PR #6813, opened 2023-06-05
change(db): Make the first stable release forward-compatible with planned state changes

### LINKED ISSUE #6746: Make the first stable release forward-compatible with planned state changes
## Motivation

We're going to make some state changes that are backward-compatible as part of the "fast spendability" RPCs in #6642. Later Zebra versions can already open and read earlier states and update them, without changing the major version of the state.

But to make them forward-compatible (earlier versions can read later states), we need to make some minor changes.

### Scheduling

Ideally, these changes should be made before the stable release, so we can guarantee that any stable Zebra version can open any stable state.

## Compatibility Issues

RocksDB requires all column families to be listed when we open the state read-write:
- [x] use [list_cf()](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.list_cf) so earlier Zebra versions can open column families added in later state versions

In #4784, we will delete some duplicate sapling and orchard tree states:
- [x] use `zs_prev_key_value_back_from()` to look up sapling and orchard note commitment trees
- [x] sprout already deletes its old note commitment trees when it adds a new one, because historical roots are looked up by their hash

Version tracking:
- [x] add a minor version `25.1` to the state, and call the previous version `25.0`

## Testing

I think we should support these changes on a best-effort basis, and just use our existing tests.

teor2345:
## Motivation

We're going to change how the state database is implemented in ticket #6642, after the first stable release.

It would be nice if all stable Zebra versions could read those states, even if they have some format changes.

Close #6746

### Specifications

https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.list_cf
https://docs.rs/rocksdb/latest/rocksdb/struct.DBIteratorWithThreadMode.html

This is all fairly straightforward, the tricky part will be tickets #6642 and #4784, where we actually change the state format.

## Solution

- Add a minor and patch version to the state disk database format, and log compatibility at startup
- Open all column families in the database, including ones added by future Zebra versions (compatibility with #6642)
- If there isn't a new note commitment tree for a height, look for the previous tree (compatibility with #4784)

### Testing

These changes are covered by existing state and RPC tests.

I don't think we should have specific forward compatibility tests, we can just support it on a best-effort basis.

## Review

This is a routine change that we'd like to get in the first stable release.

### Reviewer Checklist

## Follow Up Work

Actually change the state format.

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -1,15 +1,25 @@
 //! Cached state configuration for Zebra.
 
 use std::{
-    fs::{canonicalize, remove_dir_all, DirEntry, ReadDir},
+    fs::{self, canonicalize, remove_dir_all, DirEntry, ReadDir},
+    io::ErrorKind,
     path::{Path, PathBuf},
 };
 
+use semver::Version;
 use serde::{Deserialize, Serialize};
 use tokio::task::{spawn_blocking, JoinHandle};
 use tracing::Span;
 
 use zebra_chain::parameters::Network;
+
+use crate::{
+    constants::{
+        DATABASE_FORMAT_MINOR_VERSION, DATABASE_FORMAT_PATCH_VERSION, DATABASE_FORMAT_VERSION,
+        DATABASE_FORMAT_VERSION_FILE_NAME,
+    },
+    BoxError,
+};
 
 /// Configuration for the state service.
 #[derive(Clone, Debug, Deserialize, Serialize)]
@@ -117,6 +127,15 @@ impl Config {
                 .join(format!("v{}", crate::constants::DATABASE_FORMAT_VERSION))
                 .join(net_dir)
         }
+    }
+
+    /// Returns the path of the database format version file.
+    pub fn version_file_path(&self, network: Network) -> PathBuf {
+        let mut version_path = self.db_path(network);
+
+        version_path.push(DATABASE_FORMAT_VERSION_FILE_NAME);
+
+        version_path
     }
 
     /// Construct a config for an ephemeral database
@@ -261,8 +280,70 @@ fn parse_dir_name(entry: &DirEntry) -> Option<String> {
 /// Parse the state version number from `dir_name`.
 ///
 /// Returns `None` if parsing fails, or the directory name is not in the expected format.
-fn parse_version_number(dir_name: &str) -> Option<u32> {
+fn parse_version_number(dir_name: &str) -> Option<u64> {
     dir_name
         .strip_prefix('v')
         .and_then(|version| version.parse().ok())
+}
+
+/// Returns the full semantic version of the currently running database format code.
+///
+/// This is the version implemented by the Zebra code that's currently running,
+/// the minor and patch versions on disk can be different.
+pub fn database_format_version_in_code() -> Version {
+    Version::new(
+        DATABASE_FORMAT_VERSION,
+        DATABASE_FORMAT_MINOR_VERSION,
+        DATABASE_FORMAT_PATCH_VERSION,
+    )
+}
+
+/// Returns the full semantic version of the on-disk database.
+///
+/// This is the format of the data on disk, the minor and patch versions
+/// implemented by the running Zebra code can be different.
+pub fn database_format_version_on_disk(
+    config: &Config,
+    network: Network,
+) -> Result<Version, BoxError> {
+    let version_path = config.version_file_path(network);
+
+    let version = match fs::read_to_string(version_path) {
+        Ok(version) => version,
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            // If the version file doesn't exist, assume both minor and patch are zero.
+            // (The major versions must be the same: we create a new database if they are not.)
+            return Ok(Version::new(DATABASE_FORMAT_VERSION, 0, 0));
+        }
+        Err(e) => Err(e)?,
+    };
+
+    let (minor, patch) = version
+        .split_once('.')
+        .ok_or("invalid database format version file")?;
+
+    Ok(Version::new(
+        DATABASE_FORMAT_VERSION,
+        minor.parse()?,
+        patch.parse()?,
+    ))
+}
+
+/// Writes the currently running semantic database version to the on-disk database.
+/// This should only be called after all running format upgrades are complete.
+pub fn write_database_format_version_to_disk(
+    config: &Config,
+    network: Network,
+) -> Result<(), BoxError> {
+    let version_path = config.version_file_path(network);
+
+    // The major version is already in the directory path.
+    let version = format!(
+        "{}.{}",
+        DATABASE_FORMAT_MINOR_VERSION, DATABASE_FORMAT_PATCH_VERSION
+    );
+
+    fs::write(version_path, version.as_bytes())?;
+
+    Ok(())
 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -299,21 +299,22 @@ pub fn database_format_version_in_code() -> Version {
 }
 
 /// Returns the full semantic version of the on-disk database.
+/// If there is no existing on-disk database, returns `Ok(None)`.
 ///
 /// This is the format of the data on disk, the minor and patch versions
 /// implemented by the running Zebra code can be different.
 pub fn database_format_version_on_disk(
     config: &Config,
     network: Network,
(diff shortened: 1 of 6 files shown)


zebra-state/src/constants.rs
```diff
@@ -19,13 +26,29 @@ pub use zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY;
...
 pub const MAX_BLOCK_REORG_HEIGHT: u32 = MIN_TRANSPARENT_COINBASE_MATURITY - 1;
 
-/// The database format version, incremented each time the database format changes.
-pub const DATABASE_FORMAT_VERSION: u32 = 25;
+/// The database format major version, incremented each time the on-disk database format has a
+/// breaking change.
+///
+/// Use [`database_format_version_in_code()`] or [`database_format_version_on_disk()`]
+/// to get the full semantic format version.
+pub const DATABASE_FORMAT_VERSION: u64 = 25;
```
oxarbitrage, line 42: Do you think this and the new constants are something we should add to the release checkliist ? Something like "check if any change was done to the database and increase  `DATABASE_FORMAT_VERSION`, `DATABASE_FORMAT_MINOR_VERSION` or `DATABASE_FORMAT_PATCH_VERSION` accordingly".

oxarbitrage (dismissed):
This is very clear, thank you for making it that way. Looks great.

zebra-state/src/constants.rs
teor2345: Good question!

I haven't thought through the design for format upgrades yet, I was waiting until we actually started on ticket #6642 or #4784 after the first stable release.

I think our tests will fail if we accidentally change the database format on disk, because they won't be able to read cached states.

We could add a comment to the database format compatibility task, that says to update the database format version when updating the format itself. That task doesn't exist yet, but I can add a TODO with that comment. (The task won't exist until we make database format changes.)

We can also add a test that a format upgrade only runs once. This test is important anyway, but if we forget to update the format version the test will fail. (I'll add a TODO for this too.)

#### More Details

The last time we changed the major database format version was PR #4639 in July 2022. We've done 12 releases since then, and none of them changed the database format on disk.

(PR #6699 changed how we read that format to avoid a panic, but that wasn't a data format compatibility issue. So I wouldn't have even given it a patch release.)

We will change the database format once for these new RPCs, and then again for NU6. So that's maybe 2 changes planned in the next year. Hopefully they will both be minor version changes (extra data stored), and we won't need any patch version changes (compatibility fixes).

zebra-state/src/config.rs
```diff
@@ -261,8 +280,70 @@ fn parse_dir_name(entry: &DirEntry) -> Option<String> {
...
+) -> Result<(), BoxError> {
+    let version_path = config.version_file_path(network);
+
+    // The major version is already in the directory path.
+    let version = format!(
+        "{}.{}",
+        DATABASE_FORMAT_MINOR_VERSION, DATABASE_FORMAT_PATCH_VERSION
+    );
+
+    fs::write(version_path, version.as_bytes())?;
```
teor2345, line 359: Comment needed: we don't need to worry about concurrency here, because this method is only called after RocksDB exclusively locks the state directory.

zebra-state/src/constants.rs
```diff
@@ -19,13 +26,29 @@ pub use zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY;
...
+
+/// The database format minor version, incremented each time the on-disk database format has a
+/// significant functional change.
+pub const DATABASE_FORMAT_MINOR_VERSION: u64 = 0;
+
+/// The database format patch version, incremented each time the on-disk database format has a
+/// significant bug fix.
```
teor2345:
```suggestion
/// significant format compatibility fix.
```

zebra-state/src/constants.rs
```diff
@@ -19,13 +26,29 @@ pub use zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY;
...
-pub const DATABASE_FORMAT_VERSION: u32 = 25;
+/// The database format major version, incremented each time the on-disk database format has a
+/// breaking change.
+///
+/// Use [`database_format_version_in_code()`] or [`database_format_version_on_disk()`]
+/// to get the full semantic format version.
+pub const DATABASE_FORMAT_VERSION: u64 = 25;
+
+/// The database format minor version, incremented each time the on-disk database format has a
+/// significant functional change.
```
teor2345:
```suggestion
/// significant format change. This includes adding new column families, or changing the format of a column family in a compatible way.
```

zebra-state/src/constants.rs
```diff
@@ -19,13 +26,29 @@ pub use zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY;
 // TODO: change to HeightDiff
 pub const MAX_BLOCK_REORG_HEIGHT: u32 = MIN_TRANSPARENT_COINBASE_MATURITY - 1;
 
-/// The database format version, incremented each time the database format changes.
-pub const DATABASE_FORMAT_VERSION: u32 = 25;
+/// The database format major version, incremented each time the on-disk database format has a
+/// breaking change.
```
teor2345:
```suggestion
/// breaking change. This includes deleting a column family, or changing a column family's data format in an incompatible way.
/// (Unless we previously added compatibility code, and it's available in all supported Zebra versions.)
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -453,6 +480,13 @@ impl DiskDb {
 
                 db.assert_default_cf_is_empty();
 
+                // Now we've checked that the database format is up-to-date,
```
teor2345, line 493: Concurrency note

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -386,61 +391,83 @@ impl DiskDb {
...
-            rocksdb::ColumnFamilyDescriptor::new("sapling_nullifiers", db_options.clone()),
-            rocksdb::ColumnFamilyDescriptor::new("sapling_anchors", db_options.clone()),
-            rocksdb::ColumnFamilyDescriptor::new(
-                "sapling_note_commitment_tree",
-                db_options.clone(),
+        let running_version = database_format_version_in_code();
+        let disk_version = database_format_version_on_disk(config, network)
+            .expect("unable to read database format version file");
+
+        match running_version.cmp(&disk_version) {
```
teor2345: There should be a "creating new database" case here, I could do it by returning Ok(None) when there's no on disk version. Currently we're logging new databases as an older version, which is incorrect.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -453,6 +480,13 @@ impl DiskDb {
 
                 db.assert_default_cf_is_empty();
 
+                // Now we've checked that the database format is up-to-date,
+                // mark it as updated on disk.
+                //
+                // TODO: only update the version at the end of the format upgrade task (#6642)
```
teor2345:
```suggestion
                // TODO:
                // - only update the version at the end of the format upgrade task (#6642)
                // - add a note to that code to update the version constants
                // - add a test that the format upgrade runs exactly once if an older cached state format is opened, then if Zebra is launched again the format is not upgraded
```

zebra-state/src/constants.rs
teor2345: Fixed in commit c02f8f4ac0

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: Fixed in e90b718d79

teor2345:
This PR is ready for re-review

teor2345:
Unrelated failure, possibly #6763 but in the RPC code:
> {"app":"lightwalletd","error":"Post \"[http://127.0.0.1:41291\](http://127.0.0.1:41291/)": read tcp 127.0.0.1:45786-\u003e127.0.0.1:41291: read: connection reset by peer","level":"fatal","msg":"error zcashd getbestblockhash rpc","time":"2023-06-06T01:10:56Z"}

https://github.com/ZcashFoundation/zebra/actions/runs/5182998267/jobs/9341002724?pr=6813#step:8:426

teor2345:
@mergifyio update

oxarbitrage (approved):

Issue #6824, opened 2023-06-05
Restore macOS tests


oxarbitrage:
We have an issue with macOS tests and the last rust version upgrade (1.70). We don't know exactly what test is failing nor why so we [disabled all macOS tests](https://github.com/ZcashFoundation/zebra/pull/6825) for now as we are not supporting it in the first stable release.

This ticket is for researching whats going on and potentially restore the tests if possible.

teor2345:
> We don't know exactly what test is failing nor why

One of the state tests is failing, likely because the RocksDB database shutdown code is not running at the correct time during `zebra-state` shutdown. (We've seen this issue before in the state.)

teor2345:
I added this to the list of tasks in the original bug #6812

Issue #6812, opened 2023-06-04
Process abort in state tests on some macOS versions with Rust 1.70


teor2345:
## Tasks

- [x] #7834 
- [x] #7842
  - [x] Admin: Add a macOS test branch protection rule

## Motivation

Zebra's state tests are crashing on macOS across multiple PRs:

<details>

> Running `/Users/runner/work/zebra/zebra/target/release/deps/zebra_state-52d6a4509c1a5e9a`
> running 97 tests
> test service::chain_tip::tests::vectors::chain_tip_change_is_initially_not_ready ... ok
> test service::chain_tip::tests::vectors::current_best_tip_is_initially_empty ... ok
> test service::chain_tip::tests::vectors::empty_latest_chain_tip_is_empty ... ok
> test service::chain_tip::tests::prop::best_tip_is_latest_non_finalized_then_latest_finalized ... ok
> 2023-06-01T20:21:10.039357Z  INFO Opened Zebra state cache at /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebra-state-v25-mainnet-T3MWk1
> 2023-06-01T20:21:10.039660Z  INFO loaded Zebra state cache tip=None
> 2023-06-01T20:21:10.040026Z  INFO Opened Zebra state cache at /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebra-state-v25-mainnet-n6VK4c
> 2023-06-01T20:21:10.040049Z  INFO loaded Zebra state cache tip=None
> 2023-06-01T20:21:10.130630Z  INFO Opened Zebra state cache at /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/zebra-state-v25-mainnet-OSJIVt
> 2023-06-01T20:21:10.130704Z  INFO loaded Zebra state cache tip=None
> error: test failed, to rerun pass `-p zebra-state --lib`
> Caused by:

</details>

>   process didn't exit successfully: `/Users/runner/work/zebra/zebra/target/release/deps/zebra_state-52d6a4509c1a5e9a` (signal: 6, SIGABRT: process abort signal)
> Error: Process completed with exit code 101.

https://github.com/ZcashFoundation/zebra/actions/runs/5148228655/jobs/9269653348#step:15:3285
https://github.com/ZcashFoundation/zebra/actions/runs/5151593461/jobs/9276906631?pr=6810#step:15:3274

### Investigation

[...]

PR #6820, opened 2023-06-05
fix(state): Stop aborting on shutdown in state tests on macOS

teor2345:
## Motivation

We're seeing process aborts in the state on some macOS versions starting in Rust 1.70. This is the code that last caused that problem.

Close #6812

### Specifications

https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.cancel_all_background_work

### Complex Code or Requirements

This is an ongoing concurrency bug: the database must shut down before its destructors run, but running the shutdown code also sometimes causes crashes.

## Solution

- Only run the shutdown code on macOS
- Ignore "shutdown in progress" errors in the state shutdown code, and just log other errors

### Testing

I haven't tested this yet because I don't have an x86_64 macOS machine. CI will test it.

## Review

This fixes a CI failure that's stopping all Rust PRs merging. We've disabled that job requirement, so it is urgent but not critical.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

- [ ] Admin: re-enable the macOS requirement after this PR has merged, and all failing PRs have been merged/updated

PR #6825, opened 2023-06-05
fix(ci): remove macOS tests from CI

oxarbitrage:
## Motivation

We have crashes in every PR after rust upgraded to 1.70 in macOS tests. We don't know what tests are causing it so we are disabling all tests for this platform as they are not supported by Zebra in tier 1.

Fixes the CI error in https://github.com/ZcashFoundation/zebra/issues/6812, but not the underlying bug.

## Solution

Remove all macOS tests from CI.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

Investigate whats going on and re enable if possible. https://github.com/ZcashFoundation/zebra/issues/6824

### DIFF
--- a/.github/workflows/continous-integration-os.patch.yml
+++ b/.github/workflows/continous-integration-os.patch.yml
@@ -22,7 +22,8 @@ jobs:
     strategy:
       matrix:
         # TODO: Windows was removed for now, see https://github.com/ZcashFoundation/zebra/issues/3801
-        os: [ubuntu-latest, macos-latest]
+        # TODO: macOS tests were removed for now, see https://github.com/ZcashFoundation/zebra/issues/6824
+        os: [ubuntu-latest]
         rust: [stable, beta]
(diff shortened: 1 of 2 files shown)


teor2345:
I am a bit surprised, but my PR actually seems to fix the underlying bug:
https://github.com/ZcashFoundation/zebra/pull/6820

The test failures I saw were easy to fix.

So let's leave this one in draft until we see if that one passes CI?

oxarbitrage:
> I am a bit surprised, but my PR actually seems to fix the underlying bug: #6820
> 
> The test failures I saw were easy to fix.
> 
> So let's leave this one in draft until we see if that one passes CI?

That is even better, so yea, lets do that.

teor2345:
Looks like my PR didn't work, let's try this one.

teor2345 (approved):
Thanks for this.

I left the issue open because the bug isn't actually fixed, but I changed its priority to low.

PR #6828, opened 2023-06-06
fix(state): Avoid panicking on state errors during shutdown

teor2345:
## Motivation

Zebra can panic if there's a disk error during shutdown, but there's nothing we can do about that error anyway.

### Specifications

https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.flush
https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.flush_wal

This is concurrent code, but this PR just updates error handling and comments.

## Solution

- Ignore errors during state shutdown and just log them
- Document the different ways we've tried to handle database shutdown
- Add the macOS shutdown crash to the README known issues

### Testing

This is covered by existing state tests.

## Review

### Reviewer Checklist

## Follow Up Work

Fix the actual shutdown bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ There are a few bugs in Zebra that we're still working on fixing:
 
 - Block download and verification sometimes times out during Zebra's initial sync [#5709](https://github.com/ZcashFoundation/zebra/issues/5709). The full sync still finishes reasonably quickly.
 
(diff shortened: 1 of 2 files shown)


teor2345:
@mergifyio update

oxarbitrage (approved):
looks good.

PR #6832, opened 2023-06-06
fix(net): Try harder to drop connections when they shut down, Credit: Ziggurat Team

### LINKED ISSUE #6805: Zebra's peer set should drop multiple outbound connections to the same IP and port, Credit: Ziggurat Team
## Motivation

It looks like some Zebra instances are making multiple outbound connections to the same node:
https://forum.zcashcommunity.com/t/ziggurat-3-0/43350/40

This is a potential remote denial of service issue.

This is concurrent code. The `PeerSet` should be dropping all but one outbound connection to the same IP address and port.

We should check that behaviour is actually happening, and that the connections are closed when they are dropped. 

### Testing

We have similar `PeerSet` tests with different IPs and ports, we should copy them and test with identical IPs and ports instead.

## Related Work

- Part of this issue might be fixed by PR #6672 in Zebra v1.0.0-rc.9

teor2345:
## Changelog

The Ziggurat team needs to be credited in this PR's changelog entry. This should happen automatically.

## Motivation

Some users report that Zebra v1.0.0-rc.8 opens multiple connections to the same node.

Close #6805

This is concurrent code, but this PR mainly modifies the shutdown sequence. The task shutdowns could have timing issues, but the other shutdown code is not concurrent.

## Solution

- Force Client tasks to shut down when it is dropped
- Try to close the Connection peer sender sink on drop
- Reliably shut down the peer sender when the Connection is shut down
- Add PeerSet logging for duplicate peer connections, checking both IP:port and just IP 

### Testing

I couldn't reproduce the errors on the latest `main` branch, and I didn't find any bugs that would be easy to test.

I added logging so these kinds of issues are easier to detect in future.

## Review

### Reviewer Checklist

## Follow Up Work

Ask the Ziggurat team to re-test.

### DIFF
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -98,6 +98,7 @@ use std::{
     fmt::Debug,
     future::Future,
     marker::PhantomData,
+    net::IpAddr,
     pin::Pin,
     task::{Context, Poll},
     time::Instant,
@@ -109,6 +110,7 @@ use futures::{
     prelude::*,
     stream::FuturesUnordered,
 };
+use itertools::Itertools;
 use tokio::{
     sync::{broadcast, oneshot::error::TryRecvError, watch},
     task::JoinHandle,
@@ -123,6 +125,7 @@ use zebra_chain::chain_tip::ChainTip;
 
 use crate::{
     address_book::AddressMetrics,
+    constants::MIN_PEER_SET_LOG_INTERVAL,
     peer::{LoadTrackedClient, MinimumPeerVersion},
     peer_set::{
         unready_service::{Error as UnreadyError, UnreadyService},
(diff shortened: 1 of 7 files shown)


zebra-network/src/peer_set/set.rs
```diff
@@ -810,33 +813,84 @@ where
...
+        self.last_peer_log = Some(now);
+
+        // Log potential duplicate connections.
+        let peers = self.peer_set_addresses();
+
+        // Check for duplicates by address and port: these are unexpected and represent a bug.
+        let duplicates: Vec<PeerSocketAddr> = peers.iter().duplicates().cloned().collect();
+
+        let mut peer_counts = peers.iter().counts();
+        peer_counts.retain(|peer, _count| duplicates.contains(peer));
```
arya2, line 858: Nitpick:

```suggestion
        let mut peer_counts = peers.iter().counts();
        peer_counts.retain(|_peer, count| *count > 1);
```

zebra-network/src/peer_set/set.rs
```diff
@@ -810,33 +813,84 @@ where
...
+            );
+        }
+
+        // Check for duplicates by address: these can happen if there are multiple nodes
+        // behind a NAT or on a single server.
+        let peers: Vec<IpAddr> = peers.iter().map(|addr| addr.ip()).collect();
+        let duplicates: Vec<IpAddr> = peers.iter().duplicates().cloned().collect();
+
+        let mut peer_counts = peers.iter().counts();
+        peer_counts.retain(|peer, _count| duplicates.contains(peer));
```
arya2, line 877:
```suggestion
        let mut peer_counts = peers.iter().counts();
        peer_counts.retain(|_peer, count| *count > 1);
```

arya2 (approved):
Thank you!

zebra-network/src/peer_set/set.rs
teor2345: Thanks for this cleanup. This fix really needs to go in the release candidate, so I'm going to let it merge now.

Feel free to submit this as a separate PR.

PR #6865, opened 2023-06-08
add(doc): Add instructions for doing mass renames easily

teor2345:
## Motivation

Here are some instructions for doing mass renames, we can use them in ticket #6793.

## Review

This just needs a quick check to see if they make sense.

We can update these instructions after we do ticket #6793.

### Reviewer Checklist


mpguerra (approved):
thank you!

PR #6869, opened 2023-06-08
fix(net): Avoid potential concurrency bugs in outbound handshakes

teor2345:
## Motivation

This PR fixes a bunch of potential concurrency bugs in Zebra's outbound handshake code.

Part of #6763.

This is not a blocker for the first stable release.

This is concurrent code, but the final code should be a lot easier to reason about than the previous code, because it's all running concurrently in independent tasks.

## Solution

- [Stop sending peer errors on the PeerSet channel](https://github.com/ZcashFoundation/zebra/commit/626bf47a52accfe19304ea8978371ecd2bdde789)
    - this fixes a potential deadlock or panic when the channel gets full
- [Move locking out of the cralwer select!](https://github.com/ZcashFoundation/zebra/commit/c5a514bf47eb3dbcffaca53af58867e903f84420)
    - this avoids using the address book lock in a select! statement, which is very hard to reason about
- [Move report_failed() out of the CandidateSet](https://github.com/ZcashFoundation/zebra/commit/99592273acc0634d0c5d031900ee5727ec25a830)
    - another move that makes it easier to reason about concurrency, because the code is now in the same file with no other locking
- [Make CandidateSet Send](https://github.com/ZcashFoundation/zebra/commit/74ed89d7d7d6b6133551ce7b880995ce3f6b2022)
    - just changes to generic bounds, it seems the compiler has got smarter about this since I last tried (or I have!)
- [Make all CandidateSet operations concurrent](https://github.com/ZcashFoundation/zebra/commit/5effe36b5716070eef8b6926e0d613e49bbce4ba)
    - this code has caused hang and deadlock bugs in the past, now it is fully concurrent, so there are no weird dependencies
- Reduce the gap between handshakes and peer set updates
    - this is another concurrency risk - we used to delay updating the peer set or address book until the next time the crawler task looped around, now we just do it concurrently in the handshake task
- Exit the crawler task on shutdown
    - when channels are disconnected, exit the task, rather than panicking or hanging

### Testing

Our existing unit and integration tests have good coverage of this code, and "is this task concurrent" is hard to test.

## Review

@arya2 and I discussed part of this design about a week ago.

I can split this into 2-3 PRs if it would be easier, or do a video review.

### Reviewer Checklist

## Follow Up Work

Try diagnosing any remaining bugs in #6763.

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -669,11 +669,10 @@ where
 enum CrawlerAction {
     /// Drop the demand signal because there are too many pending handshakes.
     DemandDrop,
-    /// Initiate a handshake to `candidate` in response to demand.
-    DemandHandshake { candidate: MetaAddr },
-    /// Crawl existing peers for more peers in response to demand, because there
-    /// are no available candidates.
-    DemandCrawl,
+    /// Initiate a handshake to the next candidate peer in response to demand.
+    ///
+    /// If there are no available candidates, crawl existing peers.
+    DemandHandshakeOrCrawl,
     /// Crawl existing peers for more peers in response to a timer `tick`.
     TimerCrawl { tick: Instant },
     /// Handle a successfully connected handshake `peer_set_change`.
@@ -778,21 +777,25 @@ where
         );
 
         let crawler_action = tokio::select! {
+            biased;
+            // Check for completed handshakes first, because the rest of the app needs them.
+            // Pending handshakes are limited by the connection limit.
             next_handshake_res = handshakes.next() => next_handshake_res.expect(
                 "handshakes never terminates, because it contains a future that never resolves"
             ),
+            // The timer is rate-limited
             next_timer = crawl_timer.next() => next_timer.expect("timers never terminate"),
-            // turn the demand into an action, based on the crawler's current state
+            // Turn any new demand into an action, based on the crawler's current state.
+            //
+            // # Concurrency
+            //
+            // Demand is potentially unlimited, so it must go last in a biased select!.
             _ = demand_rx.next() => {
                 if active_outbound_connections.update_count() >= config.peerset_outbound_connection_limit() {
                     // Too many open outbound connections or pending handshakes already
                     DemandDrop
-                } else if let Some(candidate) = candidates.next().await {
-                    // candidates.next has a short delay, and briefly holds the address
-                    // book lock, so it shouldn't hang
-                    DemandHandshake { candidate }
                 } else {
-                    DemandCrawl
+                    DemandHandshakeOrCrawl
                 }
             }
         };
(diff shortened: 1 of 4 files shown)


zebra-network/src/peer_set/initialize.rs
```diff
@@ -755,12 +751,17 @@ where
 
     let address_book = candidates.address_book().await;
 
+    // # Concurrency
+    //
+    // Allow tasks using the candidate set to be spawned, so they can run concurrently.
+    // Previously, Zebra has had deadlocks and long hangs caused by running dependent
+    // candidate set futures in the same async task.
+    let candidates = Arc::new(futures::lock::Mutex::new(candidates));
```
arya2, line 754: 🎉

zebra-network/src/peer_set/initialize.rs
```diff
@@ -804,97 +809,112 @@ where
...
                     //
-                    // We only try peers if there was actually an update.
-                    // So if all peers have had a recent attempt,
-                    // and there was recent update with no peers,
-                    // the channel will drain.
-                    // This prevents useless update attempt loops.
-                    if let Some(more_peers) = more_peers {
-                        let _ = demand_tx.try_send(more_peers);
+                    // Hold the lock for as short a time as possible.
+                    let candidate = { candidates.lock().await.next().await };
```
arya2, line 838: How come this is a block?

zebra-network/src/peer_set/initialize.rs
```diff
@@ -804,97 +809,112 @@ where
...
+                    if let Some(candidate) = candidate {
+                        // we don't need to spawn here, because there's nothing running concurrently
+                        dial(candidate, outbound_connector, outbound_connection_tracker).await
+                    } else {
+                        // There weren't any peers, so try to get more peers.
+                        debug!("demand for peers but no available candidates");
+
+                        crawl(candidates, demand_tx).await;
+
+                        DemandCrawlFinished
```
arya2: 👍

zebra-network/src/peer_set/initialize.rs
```diff
@@ -904,20 +924,48 @@ where
...
+) where
+    S: Service<Request, Response = Response, Error = BoxError> + Send + Sync + 'static,
+    S::Future: Send + 'static,
+{
+    // update() has timeouts, and briefly holds the address book
+    // lock, so it shouldn't hang.
+    // Try to get new peers, holding the lock for as short a time as possible.
+    let result = {
+        let result = candidates.lock().await.update().await;
+        std::mem::drop(candidates);
```
arya2, line 940: Nitpick: it's not necessary to drop `candidates` here, it drops the lock above.

zebra-network/src/peer_set/initialize.rs
```diff
@@ -904,20 +924,48 @@ where
     }
 }
 
-/// Mark `addr` as a failed peer.
-async fn report_failed(address_book: Arc<std::sync::Mutex<AddressBook>>, addr: MetaAddr) {
-    let addr = MetaAddr::new_errored(addr.addr, addr.services);
+/// Try to get more peers using `candidates`, then queue a connection attempt using `demand_tx`.
+/// If there were no new peers, the connection attempt is skipped.
+#[instrument(skip(candidates, demand_tx))]
+async fn crawl<S>(
```
arya2, line 927: Would it make sense to define `crawl()` and `dial()` in `crawl_and_dial()`?

zebra-network/src/peer_set/initialize.rs
```diff
@@ -881,17 +921,79 @@ where
...
+        if let Err(send_error) = demand_tx.try_send(more_peers) {
+            if send_error.is_disconnected() {
+                // Zebra is shutting down
+                return Err(send_error.into());
+            }
+        }
+    }
+
+    Ok(())
+}
```
arya2, line 974:
```suggestion
    Ok(try_send_demand(more_peers, &mut demand_tx)?)
}

fn try_send_demand(
    more_peers: Option<MorePeers>,
    demand_tx: &mut futures::channel::mpsc::Sender<MorePeers>,
) -> Result<(), futures::channel::mpsc::TrySendError<MorePeers>> {
    more_peers
        .and_then(|more_peers: MorePeers| demand_tx.try_send(more_peers).err())
        .filter(futures::channel::mpsc::TrySendError::is_disconnected) // Zebra is shutting down?
        .map_or(Ok(()), Err)
}
```

zebra-network/src/peer_set/initialize.rs
```diff
@@ -911,35 +1013,59 @@ where
...
+            //
+            // # Security
+            //
+            // Handshake failures are rate-limited by peer attempt timeouts.
+            if let Err(send_error) = demand_tx.try_send(MorePeers) {
+                if send_error.is_disconnected() {
+                    // Zebra is shutting down
+                    return Err(send_error.into());
                 }
             }
```
arya2, line 1049:
```suggestion
            try_send_demand(Some(MorePeers), &mut demand_tx)?;
```

arya2 (approved):
This is marvelous, thank you.

zebra-network/src/peer_set/initialize.rs
teor2345: To hold the lock for as short a time as possible, by dropping the lock guard temporary within the block.

zebra-network/src/peer_set/initialize.rs
teor2345: I agree, but it makes it earlier to reason about concurrency, because the lock guard and any other locking can't be done past that point.

zebra-network/src/peer_set/initialize.rs
teor2345: That seems like it would be harder to read, because the code would be indented, and the closures or functions could look like they execute where they're defined.

zebra-network/src/peer_set/initialize.rs
teor2345: Thanks, please submit this as another PR so this PR makes the merge freeze.

Issue #6859, opened 2023-06-07
Tracking: Upgrade shared ECC dependencies and zcash_script for zcashd 5.6.0


teor2345:
## Motivation

We need to do a routine update to our shared ECC dependencies in Zebra:
- [x] #6836 
- [x] https://github.com/ZcashFoundation/zebra/pull/6855
- [x] #7107 
- [x] #6857
- [x] https://github.com/ZcashFoundation/zebra/pull/6854
- [x] https://github.com/ZcashFoundation/zebra/pull/6851
- [x] #7151
- [x] #6638

These should be done together to avoid duplicate dependencies.
We might want to do incrementalmerkletree in a separate PR because it has breaking API changes.

### Testing

Our existing tests cover these crates, except for the note commitment tree serialisation format in RPCs. That will need a test, ideally a snapshot test that is written based on the format before the upgrade.

## Related Work

This probably needs to happen before the RPC changes, but after the first stable release.

mpguerra:
~~Part of this ticket is blocked by the zcashd 5.6.0 release~~

I see it's already been released 😅

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/6859?planning-poker) with Zenhub @arya2 @dconnolly @oxarbitrage @teor2345 @upbqdn

teor2345:
Estimation note: there are some API changes in `incrementalmerkletree`, see PR #6836 and maybe ticket #6642 for details.

dconnolly:
@oxarbitrage for upgrading `incrementalmerkletree`, here's a summary of the changes I spotted as we upgrade to 0.4.0

https://github.com/ZcashFoundation/zebra/pull/6836#issuecomment-1579444904

teor2345:
`zcash_proofs` updated to 0.12.1 (#7107)

mpguerra:
@oxarbitrage do you want to update this estimate? :D

oxarbitrage:
> @oxarbitrage do you want to update this estimate? :D

Changed to 8.

PR #6881, opened 2023-06-08
build(deps): Upgrade to ed25519-zebra 4.0.0

dconnolly:
## Motivation

Aligns with other dependencies that also rely on `curve25519-dalek 4.0.0+`

## Solution

- Upgrade `zebra-chain`
- Upgrade `tower-batch` dev dependencies
- Remove `curve25519-dalek` denial in `deny.toml`
- Update Cargo.lock via `cargo update`

Confirmed via `cargo tree --duplicates` that we have no `curve25519-dalek` duplicate versions.

### Reviewer Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -55,9 +55,6 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
-    # wait for ed25519-zebra to upgrade
-    { name = "curve25519-dalek", version = "=3.2.0" },
-
     # ECC crates
 
     # wait for zcash_primitives to remove duplicated dependencies
--- a/deny.toml
+++ b/deny.toml
@@ -57,6 +57,8 @@ skip-tree = [
 
     # wait for indexmap, toml_edit, serde_json, tower to upgrade
     { name = "hashbrown", version = "=0.12.3" },
+    # wait for metrics-exporter-prometheus to upgrade
+    { name = "hashbrown", version = "=0.13.2" },
 
     # ECC crates
 
--- a/deny.toml
+++ b/deny.toml
@@ -55,6 +55,9 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
+    # wait for metrics-exporter-prometheus to upgrade
+    { name = "hashbrown", version = "=0.13.2" },
+
     # ECC crates
 
     # wait for zcash_primitives to remove duplicated dependencies
--- a/deny.toml
+++ b/deny.toml
@@ -55,8 +55,8 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
-    # wait for metrics-exporter-prometheus to upgrade
-    { name = "hashbrown", version = "=0.13.2" },
+    # wait for indexmap, toml_edit, serde_json, tower to upgrade
+    { name = "hashbrown", version = "=0.12.3" },
 
     # ECC crates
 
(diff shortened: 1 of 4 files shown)


teor2345 (changes requested):
This is failing due to a duplicate hashbrown dependency.

The command-line for checking for duplicate dependencies is:
```sh
cargo deny check bans
```

We can add this to #6859 if you'd like instead.

dconnolly:
> This is failing due to a duplicate hashbrown dependency.
> 
> The command-line for checking for duplicate dependencies is:
> ```sh
> cargo deny check bans
> ```
> 
> We can add this to #6859 if you'd like instead.

Ah I think I updated the `deny.toml` just as you replied about that 😁

deny.toml
```diff
@@ -55,8 +55,8 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
-    # wait for ed25519-zebra to upgrade
-    { name = "curve25519-dalek", version = "=3.2.0" },
+    # wait for metrics-exporter-prometheus to upgrade
+    { name = "hashbrown", version = "=0.13.2" },
```
teor2345, line 61:
```suggestion
    { name = "hashbrown", version = "=0.12.3" },
```

deny.toml
```diff
@@ -55,8 +55,8 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
-    # wait for ed25519-zebra to upgrade
-    { name = "curve25519-dalek", version = "=3.2.0" },
+    # wait for metrics-exporter-prometheus to upgrade
```
teor2345, line 60:
```suggestion
    # wait for indexmap, toml_edit, serde_json, tower to upgrade
```

teor2345 (dismissed):

teor2345:
> Ah I think I updated the `deny.toml` just as you replied about that

It might be nice to have a "bless" script, because it would always get the crates and versions right. The tricky part would be when it creates over-broad exceptions. I guess we could just add exceptions for each duplicate crate directly.

deny.toml
```diff
@@ -55,8 +55,8 @@ skip-tree = [
     # ed25519-zebra/hashbrown: https://github.com/ZcashFoundation/ed25519-zebra/pull/65
     { name = "ahash", version = "=0.7.6" },
 
-    # wait for ed25519-zebra to upgrade
-    { name = "curve25519-dalek", version = "=3.2.0" },
+    # wait for indexmap, toml_edit, serde_json, tower to upgrade
+    { name = "hashbrown", version = "=0.12.3" },
```
dconnolly: @teor2345 This is interesting, I ran the command from CI which uses `--all-features` and it pinned on `0.13.2` and only because of `ed25519-zebra` and `metrics-exporter-prometheus`; is the one in CI incorrect?

deny.toml
```diff
@@ -55,8 +55,8 @@ skip-tree = [
...
```
teor2345, line 59:
```suggestion
    { name = "hashbrown", version = "=0.12.3" },
    # wait for metrics-exporter-prometheus to upgrade
    { name = "hashbrown", version = "=0.13.2" },
```

teor2345 (approved):

dconnolly:
oop github is being weird for me so i'll put this at the top level, 

@teor2345 This is interesting, I ran the command from CI which uses --all-features and it pinned on 0.13.2 and only because of ed25519-zebra and metrics-exporter-prometheus; I do see the other crates and other `hashbrown` version with just `cargo deny check bans`; is the one in CI incorrect?


image

teor2345:
> I ran the command from CI which uses --all-features and it pinned on 0.13.2 and only because of ed25519-zebra and metrics-exporter-prometheus; I do see the other crates and other `hashbrown` version with just `cargo deny check bans`; is the one in CI incorrect?

CI runs both commands in a matrix, and both of them must pass. (With and without `--all-features`.)

dconnolly:
> > I ran the command from CI which uses --all-features and it pinned on 0.13.2 and only because of ed25519-zebra and metrics-exporter-prometheus; I do see the other crates and other `hashbrown` version with just `cargo deny check bans`; is the one in CI incorrect?
> 
> CI runs both commands in a matrix, and both of them must pass. (With and without `--all-features`.)

Ah I missed that, thanks!

PR #6883, opened 2023-06-08
fix(release): Fix incorrect commands in release-checklist.md

teor2345:
## Motivation

There are some mistakes in the testing steps of the release checklist.

I also deleted some optional steps, and made the checklist apply to stable-to-stable releases.


gustavovalverde (approved):
LGTM

We not necessarily need to specify zfnd/zebra:`v1.0.0`, but this will be replaced anyways on each new release.

This is considering the latest stable will have `latest` as a tag, which is an alias to `zfnd/zebra` without tags.

PR #6884, opened 2023-06-08
change(checkpoints): Update checkpoints for 1.0.0

dconnolly:
## Motivation

For performance and security, we want to update the Zebra checkpoints in every release.

## Solution

Get the latest generated checkpoints from #main: https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/README.md#automatic-checkpoint-generation-in-ci

### Reviewer Checklist


teor2345 (changes requested):
I think we should update checkpoints the day before we tag the release.

Otherwise, there's not much point, because there's only 4 extra checkpoints right now.

teor2345:
(This means the release tag will have to include the checkpoints PR as well as the change log PR.)

teor2345:
We could also make this optional if we're happy with last week's checkpoints.

teor2345:
This needs to be rebased on the latest `main` to pick up the CI job changes in PR #6907.

teor2345 (approved):

PR #6885, opened 2023-06-08
fix(compatibility): Replace or add RPC content type header when applicable

### LINKED ISSUE #6363: Accept text/plain RPC encodings automatically
## Motivation

A user describes a Zebra incompatibility with `zcashd` RPC examples in this blog post:

> The curl command originally came from [the RPC docs](https://zcash.github.io/rpc/getblockchaininfo.html) as a sample, but it didn’t work out of the box. I had to change the Content-Type header from text/plain to application/json.

https://blog.nerdbank.net/2022/12/hosting-zcash-litewalletd-via-zebra

### Specifications

The `curl` example in:
https://zcash.github.io/rpc/getblockchaininfo.html

This should be a trivial change, we already adjust missing `Content-Type` headers here:
https://github.com/ZcashFoundation/zebra/blob/a97a9752d74c8505652de666cca3c67fa4ea7c81/zebra-rpc/src/server/http_request_compatibility.rs#L106-112

### Testing

- Make sure the `curl` example works after this change
    - Add a test in CI, like the existing raw RPC tests
- Make sure `zcash-cli` still works

oxarbitrage:
## Motivation

We want Zebra to support the curl examples from the zcashd rpc documentation like https://zcash.github.io/rpc/getblockchaininfo.html where the content type header is `text/plain`.

Close https://github.com/ZcashFoundation/zebra/issues/6363

## Solution

The solution here actually ignores any content type sent by the client and always use `application/json` which is the only type we support. I could be wrong here.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-node-services/src/rpc_client.rs
+++ b/zebra-node-services/src/rpc_client.rs
@@ -43,7 +43,7 @@ impl RpcRequestClient {
             .await
     }
 
-    /// Builds rpc request with a variable `content-type`. Used only for testing.
+    /// Builds rpc request with a variable `content-type`.
     pub async fn call_with_content_type(
         &self,
         method: impl AsRef<str>,
@@ -59,6 +59,24 @@ impl RpcRequestClient {
                 r#"{{"jsonrpc": "2.0", "method": "{method}", "params": {params}, "id":123 }}"#
             ))
             .header("Content-Type", content_type)
+            .send()
+            .await
+    }
+
+    /// Builds rpc request with no content type.
+    pub async fn call_with_no_content_type(
+        &self,
+        method: impl AsRef<str>,
+        params: impl AsRef<str>,
+    ) -> reqwest::Result<reqwest::Response> {
+        let method = method.as_ref();
+        let params = params.as_ref();
+
+        self.client
+            .post(format!("http://{}", &self.rpc_address))
+            .body(format!(
+                r#"{{"jsonrpc": "2.0", "method": "{method}", "params": {params}, "id":123 }}"#
+            ))
             .send()
             .await
     }
--- a/zebra-node-services/src/rpc_client.rs
+++ b/zebra-node-services/src/rpc_client.rs
@@ -43,6 +43,26 @@ impl RpcRequestClient {
             .await
     }
 
+    /// Builds rpc request with a variable `content-type`. Used only for testing.
+    pub async fn call_with_content_type(
+        &self,
+        method: impl AsRef<str>,
+        params: impl AsRef<str>,
+        content_type: String,
+    ) -> reqwest::Result<reqwest::Response> {
+        let method = method.as_ref();
+        let params = params.as_ref();
+
+        self.client
+            .post(format!("http://{}", &self.rpc_address))
+            .body(format!(
+                r#"{{"jsonrpc": "2.0", "method": "{method}", "params": {params}, "id":123 }}"#
+            ))
+            .header("Content-Type", content_type)
+            .send()
+            .await
+    }
+
     /// Builds rpc request and gets text from response
     pub async fn text_from_call(
         &self,
--- a/zebra-rpc/src/server/http_request_compatibility.rs
+++ b/zebra-rpc/src/server/http_request_compatibility.rs
@@ -111,6 +111,16 @@ impl FixHttpRequestMiddleware {
     /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint:
     ///
     /// <https://github.com/paritytech/jsonrpc/blob/38af3c9439aa75481805edf6c05c6622a5ab1e70/http/src/handler.rs#L582-L584>
+    ///
+    /// # Security
+    ///
+    /// - `content-type` headers exist so that applications know they are speaking the correct protocol with the correct format.
+    /// We can be a bit flexible, but there are some types (such as binary) we shouldn't allow.
+    /// In particular, the "application/x-www-form-urlencoded" header should be rejected, so browser forms can't be used to attack
+    /// a local RPC port. See "The Role of Routers in the CSRF Attack" in
+    /// <https://www.invicti.com/blog/web-security/importance-content-type-header-http-requests/>
+    /// - Checking all the headers is secure, but only because hyper has custom code that just reads the first content-type header.
+    /// <https://github.com/hyperium/headers/blob/f01cc90cf8d601a716856bc9d29f47df92b779e4/src/common/content_type.rs#L102-L108>
     pub fn insert_or_replace_content_type_header(headers: &mut hyper::header::HeaderMap) {
         if !headers.contains_key(hyper::header::CONTENT_TYPE)
             || headers
--- a/zebra-rpc/src/server/http_request_compatibility.rs
+++ b/zebra-rpc/src/server/http_request_compatibility.rs
@@ -44,7 +44,7 @@ impl RequestMiddleware for FixHttpRequestMiddleware {
         tracing::trace!(?request, "original HTTP request");
 
         // Fix the request headers
-        FixHttpRequestMiddleware::add_missing_content_type_header(request.headers_mut());
+        FixHttpRequestMiddleware::add_content_type_header(request.headers_mut());
 
         // Fix the request body
         let request = request.map(|body| {
@@ -106,7 +106,7 @@ impl FixHttpRequestMiddleware {
     /// Ignore client supplied `content-type` HTTP header if any and always send requests with `application/json`.
     ///
     /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint.
-    pub fn add_missing_content_type_header(headers: &mut hyper::header::HeaderMap) {
+    pub fn add_content_type_header(headers: &mut hyper::header::HeaderMap) {
         headers.insert(
             hyper::header::CONTENT_TYPE,
             hyper::header::HeaderValue::from_static("application/json"),
--- a/zebra-rpc/src/server/http_request_compatibility.rs
+++ b/zebra-rpc/src/server/http_request_compatibility.rs
@@ -103,11 +103,13 @@ impl FixHttpRequestMiddleware {
             .replace(", \"jsonrpc\": \"1.0\"", "")
     }
 
-    /// If the `content-type` HTTP header is not present,
-    /// add an `application/json` content type header.
+    /// Ignore client supplied `content-type` HTTP header if any and always send requests with `application/json`.
+    ///
+    /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint.
     pub fn add_missing_content_type_header(headers: &mut hyper::header::HeaderMap) {
-        headers
-            .entry(hyper::header::CONTENT_TYPE)
-            .or_insert(hyper::header::HeaderValue::from_static("application/json"));
+        headers.insert(
+            hyper::header::CONTENT_TYPE,
+            hyper::header::HeaderValue::from_static("application/json"),
+        );
     }
 }
--- a/zebra-rpc/src/server/http_request_compatibility.rs
+++ b/zebra-rpc/src/server/http_request_compatibility.rs
@@ -106,7 +106,10 @@ impl FixHttpRequestMiddleware {
     /// Insert or replace client supplied `content-type` HTTP header to `application/json` in the following cases:
     ///
     /// - no `content-type` supplied.
-    /// - supplied `content-type` is `text/plain`.
+    /// - supplied `content-type` start with `text/plain`, for example:
+    ///   - `text/plain`
+    ///   - `text/plain;`
+    ///   - `text/plain; charset=utf-8`
     ///
     /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint:
     ///
@@ -125,7 +128,13 @@ impl FixHttpRequestMiddleware {
         if !headers.contains_key(hyper::header::CONTENT_TYPE)
             || headers
                 .get(hyper::header::CONTENT_TYPE)
-                .filter(|value| value.to_str().ok() == Some("text/plain"))
+                .filter(|value| {
+                    value
+                        .to_str()
+                        .ok()
+                        .unwrap_or_default()
+                        .starts_with("text/plain")
+                })
                 .is_some()
         {
             headers.insert(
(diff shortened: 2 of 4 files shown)


oxarbitrage:
Before the change:

```
$ curl --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockchaininfo", "params": [] }' -H 'content-type: text/plain;' http://127.0.0.1:6666/
Supplied content type is not allowed. Content-Type: application/json is required
$ 
```

After:

```
$ curl --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockchaininfo", "params": [] }' -H 'content-type: text/plain;' http://127.0.0.1:6666/
{"result":{"chain":"main","blocks":11793,"bestblockhash":"00000001036cd5b8c7730a662a4a7533c4cacc0b0a8989f629ea4e6a754acbf2","estimatedheight":2128414,"upgrades":{"5ba81b19":{"name":"Overwinter","activationheight":347500,"status":"pending"},"76b809bb":{"name":"Sapling","activationheight":419200,"status":"pending"},"2bb40e60":{"name":"Blossom","activationheight":653600,"status":"pending"},"f5b9230b":{"name":"Heartwood","activationheight":903000,"status":"pending"},"e9ff75a6":{"name":"Canopy","activationheight":1046400,"status":"pending"},"c2d6d0b4":{"name":"Nu5","activationheight":1687104,"status":"pending"}},"consensus":{"chaintip":"00000000","nextblock":"00000000"}},"id":"curltest"}
$ curl --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockchaininfo", "params": [] }' http://127.0.0.1:6666/
{"result":{"chain":"main","blocks":11843,"bestblockhash":"00000000d3dccd08c7b4a1d56cb7d34462cd686ccd5f866490168b6ed36cefe6","estimatedheight":2128419,"upgrades":{"5ba81b19":{"name":"Overwinter","activationheight":347500,"status":"pending"},"76b809bb":{"name":"Sapling","activationheight":419200,"status":"pending"},"2bb40e60":{"name":"Blossom","activationheight":653600,"status":"pending"},"f5b9230b":{"name":"Heartwood","activationheight":903000,"status":"pending"},"e9ff75a6":{"name":"Canopy","activationheight":1046400,"status":"pending"},"c2d6d0b4":{"name":"Nu5","activationheight":1687104,"status":"pending"}},"consensus":{"chaintip":"00000000","nextblock":"00000000"}},"id":"curltest"}
$
```

Zcash-cli after the change:

```
$ ./zcash-cli -rpcport=6666 getblockchaininfo
{
  "chain": "main",
  "blocks": 20952,
  "bestblockhash": "00000000511a302da9b0de8ec02e744b39a7ffe3e0deda88bae493d574ecf9df",
  "estimatedheight": 2128470,
  "upgrades": {
    "5ba81b19": {
      "name": "Overwinter",
      "activationheight": 347500,
      "status": "pending"
    },
    "76b809bb": {
      "name": "Sapling",
      "activationheight": 419200,
      "status": "pending"
    },
    "2bb40e60": {
      "name": "Blossom",
      "activationheight": 653600,
      "status": "pending"
    },
    "f5b9230b": {
      "name": "Heartwood",
      "activationheight": 903000,
      "status": "pending"
    },
    "e9ff75a6": {
      "name": "Canopy",
      "activationheight": 1046400,
      "status": "pending"
    },
    "c2d6d0b4": {
      "name": "Nu5",
      "activationheight": 1687104,
      "status": "pending"
    }
  },
  "consensus": {
    "chaintip": "00000000",
    "nextblock": "00000000"
  }
}
$
```

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -103,11 +103,13 @@ impl FixHttpRequestMiddleware {
...
 
-    /// If the `content-type` HTTP header is not present,
-    /// add an `application/json` content type header.
-    pub fn add_missing_content_type_header(headers: &mut hyper::header::HeaderMap) {
-        headers
-            .entry(hyper::header::CONTENT_TYPE)
-            .or_insert(hyper::header::HeaderValue::from_static("application/json"));
+    /// Ignore client supplied `content-type` HTTP header if any and always send requests with `application/json`.
+    ///
+    /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint.
```
teor2345: This is correct, I confirmed it by searching the `jsonrpc_http_server` GitHub for "application/json", and finding the code that validates headers.

```suggestion
    /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint:
    /// <https://github.com/paritytech/jsonrpc/blob/38af3c9439aa75481805edf6c05c6622a5ab1e70/http/src/handler.rs#L582-L584>
```

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -103,11 +103,13 @@ impl FixHttpRequestMiddleware {
...
+    pub fn add_content_type_header(headers: &mut hyper::header::HeaderMap) {
```
teor2345: Nitpick: this is actually replacing the header, not adding an extra header with a different value
```suggestion
    pub fn replace_content_type_header(headers: &mut hyper::header::HeaderMap) {
```

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -44,7 +44,7 @@ impl RequestMiddleware for FixHttpRequestMiddleware {
         tracing::trace!(?request, "original HTTP request");
 
         // Fix the request headers
-        FixHttpRequestMiddleware::add_missing_content_type_header(request.headers_mut());
+        FixHttpRequestMiddleware::add_content_type_header(request.headers_mut());
```
teor2345:
```suggestion
        FixHttpRequestMiddleware::replace_content_type_header(request.headers_mut());
```

zebra-node-services/src/rpc_client.rs
```diff
@@ -43,6 +43,26 @@ impl RpcRequestClient {
             .await
     }
 
+    /// Builds rpc request with a variable `content-type`. Used only for testing.
+    pub async fn call_with_content_type(
```
teor2345, line 47: Test-only methods should only be compiled in tests. This also requires an update to zebrad/Cargo.toml, I'll push that to this branch now.

```suggestion
    #[cfg(any(test, feature = "proptest-impl"))]
    pub async fn call_with_content_type(
```

zebrad/tests/acceptance.rs
```diff
@@ -1467,6 +1467,52 @@ async fn rpc_endpoint(parallel_cpu_threads: bool) -> Result<()> {
...
+    let res = client
+        .call_with_content_type("getinfo", "[]".to_string(), "text/plain".to_string())
+        .await?;
+
+    // Test if rpc endpoint respond
+    assert!(res.status().is_success());
+
+    // Make the call to the `getinfo` RPC method with a random string as content type.
+    let res = client
+        .call_with_content_type("getinfo", "[]".to_string(), "whatever".to_string())
```
teor2345: I don't think we want to allow any content type value, because that would allow some kinds of cross-application attacks.

This is a potential security issue.

Can we only allow:
- no content-type
- text/plain
- anything else we can find in actual clients or documentation examples?

teor2345 (changes requested):
I've had a look at this PR, and it solves the compatibility problem in a really neat way.

But it also potentially introduces other security issues.

content-type headers exist so that applications know they are speaking the correct protocol with the correct format. We can be a bit flexible, but there are some types (such as binary) we shouldn't allow.

In particular, the "application/x-www-form-urlencoded" header should be rejected, so browser forms can't be used to attack a local RPC port. See "The Role of Routers in the CSRF Attack" in:
https://www.invicti.com/blog/web-security/importance-content-type-header-http-requests/

zebra-rpc/src/server/http_request_compatibility.rs
teor2345: This is an outdated comment.

zebra-rpc/src/server/http_request_compatibility.rs
teor2345: This is an outdated comment.

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: Was added as part of https://github.com/ZcashFoundation/zebra/pull/6885/commits/6ef6969397f44c560c7c772461f7ccbc0fb6e1c9

zebra-node-services/src/rpc_client.rs
oxarbitrage: I don't think this change is needed so i reverted https://github.com/ZcashFoundation/zebra/pull/6885/commits/752be9486aab236ba0d3b2401d11a7ffc676f034 in https://github.com/ZcashFoundation/zebra/pull/6885/commits/d23175c56d9d572da3e5399343085724fc1513b5

The new methods are already under the rpc-client feature https://github.com/ZcashFoundation/zebra/blob/752be9486aab236ba0d3b2401d11a7ffc676f034/zebra-node-services/src/lib.rs#L6 and this is only called from tests: https://github.com/ZcashFoundation/zebra/blob/752be9486aab236ba0d3b2401d11a7ffc676f034/zebrad/Cargo.toml#L227

zebrad/tests/acceptance.rs
oxarbitrage: This was changed at https://github.com/ZcashFoundation/zebra/pull/6885/commits/6ef6969397f44c560c7c772461f7ccbc0fb6e1c9

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -103,11 +103,36 @@ impl FixHttpRequestMiddleware {
...
+                    .filter_map(|x| x.to_str().ok())
+                    .map(|s| s == "text/plain")
+                    .next()
+                    .unwrap_or(false)
+                {
+                    headers.insert(
+                        hyper::header::CONTENT_TYPE,
+                        hyper::header::HeaderValue::from_static("application/json"),
+                    );
+                }
```
oxarbitrage: There is probably a simpler way to do this but i was not able to find it.

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -103,11 +103,36 @@ impl FixHttpRequestMiddleware {
...
+    pub fn insert_or_replace_content_type_header(headers: &mut hyper::header::HeaderMap) {
+        match headers.entry(hyper::header::CONTENT_TYPE) {
+            hyper::header::Entry::Vacant(_) => {
+                headers.insert(
+                    hyper::header::CONTENT_TYPE,
+                    hyper::header::HeaderValue::from_static("application/json"),
+                );
+            }
+            hyper::header::Entry::Occupied(header_data) => {
+                if header_data
```
teor2345: This code checks all the headers, and replaces any that are text/plain.

Checking all the headers is secure, but only because `hyper` has custom code that just reads the first content-type header. Can you add a security comment?

The relevant hyper code is:
`<https://github.com/hyperium/headers/blob/f01cc90cf8d601a716856bc9d29f47df92b779e4/src/common/content_type.rs#L102-L108>`

zebra-rpc/src/server/http_request_compatibility.rs
teor2345: Optional: You could do something like:
```rust
if !headers.contains_key(CONTENT_TYPE) ||
    headers.get(CONTENT_TYPE).is_some_and(|value| value == "text/plain") {
    // insert
}
```

`is_some_and()` is a new feature in Rust 1.70 (just released last week). If we want to build on older Rust versions we can use `filter(|value| value == "text/plain").is_some()`.

`get()`` is secure because hyper only allows one content-type header.
https://docs.rs/hyper/latest/hyper/header/struct.HeaderMap.html#method.get

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -103,11 +103,36 @@ impl FixHttpRequestMiddleware {
...
-            .entry(hyper::header::CONTENT_TYPE)
-            .or_insert(hyper::header::HeaderValue::from_static("application/json"));
+    /// Insert or replace client supplied `content-type` HTTP header to `application/json` in the following cases:
+    ///
+    /// - no `content-type` supplied.
+    /// - supplied `content-type` is `text/plain`.
+    ///
+    /// `application/json` is the only `content-type` accepted by the Zebra rpc endpoint:
+    ///
+    /// <https://github.com/paritytech/jsonrpc/blob/38af3c9439aa75481805edf6c05c6622a5ab1e70/http/src/handler.rs#L582-L584>
```
teor2345, line 116: This needs a security note about headers and blocking browser-based attacks, feel free to copy and paste what I wrote here:
https://github.com/ZcashFoundation/zebra/pull/6885#pullrequestreview-1470849861

teor2345 (dismissed):
Looks good!

The security comments are needed so we know why we only replaced the text/plain header. The code cleanup is optional.

I've labelled this PR so it doesn't merge yet, because we started a stable release freeze on Friday, and this isn't a release blocker.

teor2345:
> Zcash-cli after the change:

Can you do these manual tests again when you've finished changing the code?

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: Thanks, that simplifies the method code a lot. Applied at https://github.com/ZcashFoundation/zebra/pull/6885/commits/5d6a55703b16357f598966dfdc87cc729b5fb3a7

teor2345 (commented):
Thanks, this is good to go as soon as we've re-done the manual tests. (And tagged the stable release.)

oxarbitrage:
it seems we have a problem, the [zcashd documentation](https://zcash.github.io/rpc/getblockchaininfo.html) use `text/plain;` in their curl examples instead of `text/plain` which is the string are looking for. The easier thing to do will be to replace it but i was wondering if we should support both ?

teor2345:
> it seems we have a problem, the [zcashd documentation](https://zcash.github.io/rpc/getblockchaininfo.html) use `text/plain;` in their curl examples instead of `text/plain` which is the string are looking for. The easier thing to do will be to replace it but i was wondering if we should support both ?

Good question!

We could replace anything that starts with `text/plain`, and add these common test cases to the tests:
- `text/plain`
- `text/plain;`
- `text/plain; charset=utf-8`

https://stackoverflow.com/questions/30262241/specifying-a-character-set-in-http-headers/30266220#30266220

teor2345:
You could use:
https://doc.rust-lang.org/std/primitive.str.html#method.starts_with

teor2345 (approved):
Thanks for the fix and the extra tests!

PR #6887, opened 2023-06-09
cleanup(consensus): Avoid blocking threads by awaiting proof verification results from rayon in async context

arya2:
## Motivation

This PR updates `verify_single_spawning()` and `flush_spawning()` methods to delegate CPU-bound work using rayon instead of blocking tokio threads.

`rayon::iter::once` admits no parallelism on its own and will execute operations on the current thread if collected without being chained to another parallel iterator.

Depends-On: #7158

## Solution

- Replace `tokio::spawn_blocking()` calls with oneshot channels and `rayon::spawn_fifo()`

## Review

Anyone can review.

### Reviewer Checklist


zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -120,36 +119,41 @@ impl Verifier {
 
     /// Flush the batch using a thread pool, and return the result via the channel.
     /// This function returns a future that becomes ready when the batch is completed.
-    fn flush_spawning(batch: BatchVerifier, tx: Sender) -> impl Future<Output = ()> {
+    async fn flush_spawning(batch: BatchVerifier, tx: Sender) {
+        let (rsp_tx, rsp_rx) = tokio::sync::oneshot::channel();
+
         // Correctness: Do CPU-intensive work on a dedicated thread, to avoid blocking other futures.
-        tokio::task::spawn_blocking(|| {
```
arya2, line 125: Using `spawn_fifo` directly to avoid idle blocking threads counting towards `max_blocking_threads` (Zebra uses the default of 512).

teor2345: This seems like something important to put in the docs?

zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -120,36 +119,41 @@ impl Verifier {
...
+        rayon::spawn_fifo(|| {
```
teor2345: This changes panic behaviour:
https://docs.rs/rayon/latest/rayon/fn.scope_fifo.html#panics
https://docs.rs/rayon/latest/rayon/fn.spawn_fifo.html#panic-handling

Are we setting the rayon panic handler to our standard panic handler?

zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -120,36 +119,41 @@ impl Verifier {
...
             //   then run it using scope_fifo() in the worker task,
             //   limiting the number of concurrent batches to the number of rayon threads
-            rayon::scope_fifo(|s| s.spawn_fifo(|_s| Self::verify(batch, tx)))
-        })
-        .map(|join_result| join_result.expect("panic in ed25519 batch verifier"))
+            Self::verify(batch, tx);
+            let _ = rsp_tx.send(());
+        });
+
+        rsp_rx.await.expect("sender should not be dropped")
```
teor2345: This could panic during shutdown depending on the order of sender/receiver drops.
(This applies to all repetitions of this code in this PR.)

teor2345 (dismissed):
This seems like a useful change, do you think it should be scheduled before or after the first stable release?

Do you have any evidence that this specific issue is causing the hangs?
What testing do we need to do on this change to make sure it works as expected?

Also, can you please update the PR title to describe the user-visible behaviour change? It doesn't have to be detailed, just a quick summary of the most significant change is fine.

zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -11,7 +11,6 @@ use futures::{future::BoxFuture, FutureExt};
 use once_cell::sync::Lazy;
```
teor2345, line 11: I'm making this comment to block merging this PR there's a test plan.

(We could test after merging and revert if there are any issues, but in that case, this PR should go at the end of the PR backlog.)

Edit: updated to remove the stable release

arya2:
> Should it be scheduled before or after the first stable release?
> ..
> Any evidence that this specific issue is causing the hangs?

This should go after the stable release, I checked the queue depth and this is definitely *not* causing the hangs.

zebra-consensus/src/primitives/ed25519.rs
arya2: > Are we setting the rayon panic handler to our standard panic handler?

I think we should, it could help with debugging later.

teor2345: I have a preference for panicking at the site of the problem, but it's not a blocker.

Also in this specific case, the code we're calling is low risk.

teor2345 (changes requested):
Just marking this as needing changes, so I can see the status easily in the PR list.

zebra-consensus/src/primitives/ed25519.rs
teor2345: Since we've made a few stable releases now, I think it's ok to merge this PR without a test plan.

teor2345:
Let's try to merge this PR a while before a release, so it gets time on the `main` branch for testing unexpected behaviour or panics?

zebra-consensus/src/primitives/ed25519.rs
arya2: > Are we setting the rayon panic handler to our standard panic handler?

Actually, it appears it is using the std panic hook.

zebra-consensus/src/primitives/ed25519.rs
arya2: It turns out this was irrelevant, Zebra doesn't seem to approach anywhere near the limit except briefly at startup.

zebra-consensus/src/primitives/ed25519.rs
teor2345: I'm not sure if you meant to resolve this?

Users have opened bug reports for panics on shutdown, so I'd like to fix this.

I think it would be better to either:
* ignore any errors, optionally logging them with a note "Is Zebra shutting down?"
* if we need to return something, return the closest Result::Err variant

We can't return `Ok(())`, because technically that could make verification of an invalid proof or signature pass, and get committed to the finalized state.

arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/6887/commits/e4fcc93ed2b85adfcf6eb9fcdc912afd5bfd2273, updated to use `BoxError` in https://github.com/ZcashFoundation/zebra/pull/6887/commits/8c6fb80e8af4445bc096255299deba35dc62768a, simplified in https://github.com/ZcashFoundation/zebra/pull/6887/commits/08649438e91f6dce7bfa258add4b8911280f7e22 and https://github.com/ZcashFoundation/zebra/pull/6887/commits/08649438e91f6dce7bfa258add4b8911280f7e22.

> I'm not sure if you meant to resolve this?

No, probably a mis-click.

> if we need to return something, return the closest Result::Err variant

I went with `BoxError` to avoid repetitive or confusing logs (they're all used through `Fallback` which converts to `BoxError` anyway).

zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -120,43 +126,32 @@ impl Verifier {
...
-            // - spawn fallback individual verifications so rayon executes them in FIFO order,
-            //   if possible
-            rayon::iter::once(item)
-                .map(|item| item.verify_single())
-                .collect()
-        })
-        .map(|join_result| join_result.expect("panic in ed25519 fallback verifier"))
+        // TODO:
+        // - when a batch fails, spawn all its individual items into rayon using Vec::par_iter()
+        // - spawn fallback individual verifications so rayon executes them in FIFO order,
```
teor2345: Optional: I think this TODO item has been achieved in this PR, and the other TODO is now redundant?
This applies to all of the verify_single_spawning() methods.

zebra-consensus/src/primitives/ed25519.rs
```diff
@@ -120,43 +126,32 @@ impl Verifier {
 
     /// Flush the batch using a thread pool, and return the result via the channel.
     /// This function returns a future that becomes ready when the batch is completed.
```
teor2345, line 128: Nitpick: Can you please update the docs of all the methods you've modified?
The return values are now different.

teor2345 (dismissed):
This looks good to me, thanks for this cleanup.

This is concurrent, consensus critical, cryptographic code, so changes are higher risk. I'd like to make sure it doesn't impact consensus compatibility or syncing behaviour.

Can you please run a full sync on this branch?
It might help to start with testnet syncs, and then run one final mainnet sync before this PR merges.

I'd also like to get a cryptographer to review before it merges. (I've marked it as needing two reviews, and depending on PR #7158 which will enforce two reviews.)

teor2345:
> `rayon::iter::once` admits no parallelism on its own and will execute operations on the current thread if collected without being chained to another parallel iterator.

Because this is a bug that could block other futures in a similar way to PR #7103, should this PR be marked as low or medium priority?

zebra-consensus/src/primitives/ed25519.rs
teor2345: Setting the rayon panic hook seems like something for another ticket or PR.

arya2:
> Can you please run a full sync on this branch?

Initial testnet sync is running here: https://github.com/ZcashFoundation/zebra/actions/runs/5482290801
Mainnet sync: https://github.com/ZcashFoundation/zebra/actions/runs/5490276483

zebra-consensus/src/primitives/ed25519.rs
arya2: > Setting the rayon panic hook seems like something for another ticket or PR.

I meant that it seems to be printing ideal diagnostic information already (to my surprise):

image

teor2345: That's good, thanks for checking!

zebra-consensus/src/primitives/ed25519.rs
arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/6887/commits/02684ca1349a0c0513feecdba27a0dd5afb02a18.

zebra-consensus/src/primitives/ed25519.rs
arya2: Thank you, removed in https://github.com/ZcashFoundation/zebra/pull/6887/commits/02684ca1349a0c0513feecdba27a0dd5afb02a18 and https://github.com/ZcashFoundation/zebra/pull/6887/commits/126a6e57b02737b036784bec48b5276ed2d37502.

arya2:
Build for coverage tests failed:

> error: linking with `cc` failed: exit status: 1
> ..
>   = note: /usr/bin/ld: final link failed: No space left on device
>
> https://github.com/ZcashFoundation/zebra/actions/runs/5489130816/jobs/10002885276#step:10:415
> https://github.com/ZcashFoundation/zebra/actions/runs/5489130816/jobs/10004921424#step:10:415

teor2345:
> > Can you please run a full sync on this branch?
> 
> Initial testnet sync is running here: [ZcashFoundation/zebra/actions/runs/5482290801](https://github.com/ZcashFoundation/zebra/actions/runs/5482290801)

It looks like this workflow run didn't run the testnet sync, did you check the box for full sync and change the network to testnet?

> Mainnet sync: [ZcashFoundation/zebra/actions/runs/5490276483](https://github.com/ZcashFoundation/zebra/actions/runs/5490276483)

Thanks, looks like it will finish around Wednesday or Thursday, are you ok to wait until then?

zebra-consensus/src/primitives.rs
```diff
@@ -1,5 +1,9 @@
 //! Asynchronous verification of cryptographic primitives.
```
teor2345, line 1: This PR is waiting for a full sync before it merges.

teor2345 (approved):
The code looks good, thanks!

teor2345:
> Build for coverage tests failed:
> 
> > error: linking with `cc` failed: exit status: 1
> > ..
> > = note: /usr/bin/ld: final link failed: No space left on device
> > [ZcashFoundation/zebra/actions/runs/5489130816/jobs/10002885276#step:10:415](https://github.com/ZcashFoundation/zebra/actions/runs/5489130816/jobs/10002885276#step:10:415)
> > [ZcashFoundation/zebra/actions/runs/5489130816/jobs/10004921424#step:10:415](https://github.com/ZcashFoundation/zebra/actions/runs/5489130816/jobs/10004921424#step:10:415)

This will be fixed in ticket #7180, I've disabled those jobs and their branch protection rule for now.

teor2345:
> > > Can you please run a full sync on this branch?
> > 
> > 
> > Initial testnet sync is running here: [ZcashFoundation/zebra/actions/runs/5482290801](https://github.com/ZcashFoundation/zebra/actions/runs/5482290801)
> 
> It looks like this workflow run didn't run the testnet sync, did you check the box for full sync and change the network to testnet?

Ah, I see what happened here. There is a bug in the workflow where `Testnet` isn't being passed to all the test jobs. I opened PR #7186 to fix it.

Until then, we have to select the full sync job when using testnet:
Screenshot 2023-07-10 at 12 42 49

If we don't select anything, or select the checkpoint job, it will regenerate the missing checkpoint cached state on testnet, but write it to a disk named mainnet.

And if we select the lightwalletd job, it might write a testnet lightwalletd cached state to a disk labelled mainnet. (Or fail because we haven't configured lightwalletd for testnet, or it doesn't work on testnet.)

conradoplg (approved):
Looks good!

zebra-consensus/src/primitives.rs
teor2345: This mainnet sync:
https://github.com/ZcashFoundation/zebra/actions/runs/5490276483

~~And I just started a testnet sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/5511409905~~

teor2345:
@mergifyio update

zebra-consensus/src/primitives.rs
teor2345: I accidentally cancelled the testnet sync when I was trying to cancel just the checkpoint and lightwalletd jobs. I'm really sorry about that!

I started another testnet sync here:
https://github.com/ZcashFoundation/zebra/actions/runs/5514327024

The checkpoint and lightwalletd jobs will run because they don't have cached states for testnet. If they fail, we can ignore them and open a PR to skip them in manual testnet runs.

arya2: > This mainnet sync:
> https://github.com/ZcashFoundation/zebra/actions/runs/5490276483

This one passed.

teor2345: The mainnet full sync took 182232s (50.6 hours):
https://github.com/ZcashFoundation/zebra/actions/runs/5490276483/jobs/10071236412#step:8:19629

Compared to 179831s (50.0h) and 180705s (50.2h) for the last two scheduled syncs (Friday's CI failed due to a temporary network error, but the test actually finished):
https://github.com/ZcashFoundation/zebra/actions/runs/5422688804/jobs/9886836808#step:8:17919
https://github.com/ZcashFoundation/zebra/actions/runs/5486195320/jobs/10023005486#step:8:18014

That's a 1% performance decrease, which is ok, it's about what we'd expect over 3 weeks, and is partly due to the full verification since the last checkpoint update.

But it also produces 1600 more log lines (30 per hour). Would you mind checking that out? It doesn't really matter much, but I thought it was worth pointing out.

teor2345 (approved):
This is good to go!

zebra-consensus/src/primitives.rs
teor2345: The extra log lines might have been a log we deleted in the last 135 commits. The branch we were running the manual sync from hadn't been rebased in a while.

arya2: > The extra log lines might have been a log we deleted in the last 135 commits.

Yeah, ~it may have been missing a change from `info` level logging to `debug` in `route_inv()`~, I don't think there would be 1600 fewer handshake errors if `reconnection_peers()` checks by IP but perhaps a couple extra logs came from there too.

Update: it's wasn't the log in `route_inv()`

teor2345: It's likely either overload logs:
> 2023-07-10T22:49:59.623623Z  INFO {net="Main"}:{peer=Out("v4redacted:8233")}:msg_as_req{msg="inv"}: zebra_network::peer::connection: inbound service is overloaded, closing connection drop_connection_probability=0.05000000074505806 remote_user_agent="/MagicBean:5.6.1/" negotiated_version=Version(170100) peer="v4redacted:8233" last_peer_state=Some("AwaitingRequest::In::Req::AdvertiseTransactionIds") remote_height=Height(2151066) cached_addrs=2 connection_state=AwaitingRequest

(There are at least a few hundred of these.)

The duplicate IP address info messages:
> 2023-07-11T13:10:47.276913Z  INFO {net="Main"}:crawl_and_dial{new_peer_interval=30s}: zebra_network::peer_set::set: duplicate IP addresses in peer set duplicate_connections=2 duplicated_peers=1 peers=75

Or various logs or warnings that don't happen since we fixed the 100% CPU future.

teor2345:
Looks like mergify is waiting for the `lightwalletd tip / Run lwd-full-sync test`:
https://github.com/ZcashFoundation/zebra/pull/6887/checks?check_run_id=14929147913

Screenshot 2023-07-12 at 13 50 44

But that patch job actually ran:
https://github.com/ZcashFoundation/zebra/actions/runs/5514322056/jobs/10053440684?pr=6887

I can't work out what went wrong here, so I'm going to re-try running the patch job.

teor2345:
@mergifyio refresh

teor2345:
@mergifyio refresh

PR #6890, opened 2023-06-09
fix(deployment): Fix log file path and log colour

teor2345:
## Motivation

I can't say "no log file" for manually deployed instances.

Google Cloud logs have colour escapes in them.

## Solution

Only use the GItHub repository variable for auto-deployments.
Allow colours to be forced on or off.

## Review

This is a routine fix.

Please check I spelt color the North American way in the PR code.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #6893, opened 2023-06-09
fix(deploy): allow the container to raise in MIGs

### LINKED ISSUE #5644: infra: Some disks are not being recognized as RO/RW and containers are failing to get raised
## Motivation

We've been reducing the amount of errors happening in CI, and in GCP logs, to reduce the spectrum of possibilities causing SSH errors from CI.

One of the latest errors is:

```
Failed to start container: Volume pd-0: Could not determine if the GCE Persistent Disk zebrad-cache-7b73aa0 is attached read-only or read-write
```

### Designs

- Verify if there's documentation on how to properly mount disks on Compute Engine
- Validate if we're mounting disks in a standard way for all instances in CI and CD.

### LINKED ISSUE #6861: Enable logging on Google Cloud continuous delivery instances
## Motivation

Currently logging doesn't work on these instances: they just show a few instance security logs at startup.

This is urgent because it impacts final release candidate testing.

### Reference

The service account might not have enough permissions:
https://stackoverflow.com/questions/59188515/gcloud-compute-instances-create-with-container-not-logging-to-stackdriver/66466152#66466152

Or the rest of this checklist:
https://cloud.google.com/logging/docs/agent/logging/troubleshooting

This requires Google Cloud admin permissions.

### Testing

1. Change the CD_LOG_FILE repository variable to an empty string
2. Manually run this workflow: https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/continous-delivery.yml
3. Check that the instance logs some `zebrad` logs

## Related Work

Discovered as part of:
- #6842

gustavovalverde:
## Motivation

The Managed Instance Groups in GCP were not raising the container as those were expecting the device name and the correct RW/RO mode to match between the created disk and the disk we're mounting in the container.

This also fixes monitoring, and uses the same caching directory as in CI, as the Container Optimized Image does not allow mounting in non-standard linux paths for security reasons.

Fixes: #5644
Fixes: #6861  

### Specifications

See: https://cloud.google.com/sdk/gcloud/reference/compute/instance-templates/create-with-container#--container-mount-disk

## Solution

- Add the missing `--boot-disk-size 300GB` to the MIGs template, just as we have on CI and Single instance deploy
- Allow user output and container `stdin` and `tty` to standardize with other approaches
- Enable Google Monitoring for single and long-live instances (MIGs)
- Use `ZEBRA_CACHED_STATE_DIR` with `/var/cache/zebrad-cache` as we do on CI, as the `cos-cloud` doesn't behave correctly with non-standard paths.

## Review

Anyone from DevOps

### Reviewer Checklist

## Follow Up Work

We should make further changes in a following PR:

- For releases, it should sync from scratch (1). Currently it's 2-4 weeks between releases, so a 3 day mainnet sync is ok.

- For `main` branch pushes, it should have a cached state (2 or 3, whatever is easier or more reliable). We merge new PRs multiple times a day, so a cached state is needed.

### DIFF
--- a/.github/workflows/continous-delivery.yml
+++ b/.github/workflows/continous-delivery.yml
@@ -211,15 +211,18 @@ jobs:
       - name: Create instance template for ${{ matrix.network }}
         run: |
           gcloud compute instance-templates create-with-container zebrad-${{ needs.versioning.outputs.major_version || env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}-${NETWORK} \
+          --boot-disk-size 300GB \
           --boot-disk-type=pd-ssd \
           --image-project=cos-cloud \
           --image-family=cos-stable \
           --user-output-enabled \
-          --metadata google-logging-enabled=true,google-logging-use-fluentbit=true \
+          --metadata google-logging-enabled=true,google-logging-use-fluentbit=true,google-monitoring-enabled=true \
+          --container-stdin \
+          --container-tty \
           --container-image ${{ vars.GAR_BASE }}/zebrad@${{ needs.build.outputs.image_digest }} \
-          --container-env "NETWORK=${{ matrix.network }},LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }},SHORT_SHA=${{ env.GITHUB_SHA_SHORT }}" \
-          --create-disk=name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK},auto-delete=yes,size=300GB,type=pd-ssd \
-          --container-mount-disk=mount-path="/zebrad-cache",name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK} \
+          --container-env "NETWORK=${{ matrix.network }},ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache,LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }},SHORT_SHA=${{ env.GITHUB_SHA_SHORT }}" \
+          --create-disk=name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK},device-name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK},auto-delete=yes,size=300GB,type=pd-ssd,mode=rw \
+          --container-mount-disk=mount-path='/var/cache/zebrad-cache',name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK},mode=rw \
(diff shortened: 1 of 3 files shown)


.github/workflows/continous-delivery.yml
```diff
@@ -211,15 +211,18 @@ jobs:
...
           --user-output-enabled \
-          --metadata google-logging-enabled=true,google-logging-use-fluentbit=true \
+          --metadata google-logging-enabled=true,google-logging-use-fluentbit=true,google-monitoring-enabled=true \
+          --container-stdin \
+          --container-tty \
           --container-image ${{ vars.GAR_BASE }}/zebrad@${{ needs.build.outputs.image_digest }} \
-          --container-env "NETWORK=${{ matrix.network }},LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }},SHORT_SHA=${{ env.GITHUB_SHA_SHORT }}" \
-          --create-disk=name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK},auto-delete=yes,size=300GB,type=pd-ssd \
-          --container-mount-disk=mount-path="/zebrad-cache",name=zebrad-cache-${{ env.GITHUB_SHA_SHORT }}-${NETWORK} \
+          --container-env "NETWORK=${{ matrix.network }},ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache,LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }},SHORT_SHA=${{ env.GITHUB_SHA_SHORT }}" \
```
teor2345: Using a different state path will need changes in `runtime-entrypoint.sh`. You should just be able to substitute the variable in directly, like this:
https://github.com/ZcashFoundation/zebra/blob/286ebdbf8c9e40891164821852567bc86de30012/docker/runtime-entrypoint.sh#L39

It would also be nice to set this path as the default in the `Dockerfile`, to avoid confusion.

If you're looking for an example of how runtime config creation works, see PR #6890. (But terminal colour is a bit more complicated because the variable doesn't match the config setting.)

teor2345:
> * For releases, it should sync from scratch (1). Currently it's 2-4 weeks between releases, so a 3 day mainnet sync is ok.

These further changes could be optional. For testing, it's enough to just run a scheduled full sync in our test environment. To support the network, using a cached state will be better because the instances are live faster.

.github/workflows/continous-delivery.yml
gustavovalverde: Solved with: [fix(docker): add the ZEBRA_CACHED_STATE_DIR as a default ENV](https://github.com/ZcashFoundation/zebra/pull/6893/commits/10c3fc9cc6e6a813f20a8f9074b108b53b9d5936)

teor2345 (dismissed):

dconnolly (approved):
Approving; interesting if mode=rw matters, docs say it is the default when unspecified for both `--create-disk` and `--container-mount-disk`

PR #6920, opened 2023-06-12
ci(tests): run network syncs on a weekly schedule

### LINKED ISSUE #6631: Execute Zebra full sync weekly on the `main` branch
### Discussed in https://github.com/ZcashFoundation/zebra/discussions/6603

<div type='discussions-op-text'>

<sup>Originally posted by **gustavovalverde** May  3, 2023</sup>
GitHub changed the behavior of the concurrency, showing cancellation statuses as failures, rather than ignoring them and showing it as successes. Having all those :x: in the main branch is confusing, and we should decide how to handle this situation.

We have several options (including others not listed here):
- Not running the full sync on each merge, as we're anyways cancelling those until a previous one finishes
- Have it on its own workflow and have a more specific decision tree on when to trigger it
- Schedule it to run twice a week or weekly (as it anyways takes 3+ days to complete)
- Having a trigger on the PR, as we might want to validate if a critical PR runs a Full sync safely before merging it (required check for specific labels [?])</div>

gustavovalverde:
## Motivation

GitHub changed the behavior of the concurrency, showing cancellation statuses as failures, rather than ignoring them and showing it as successes. Having all those ❌ in the main branch is confusing and, as the full sync takes multiple days to complete, we don't need to run it on each merge to `main`, and doing it weekly should be enough.

This change is being applied to the following tests:
- Zebra full Mainnet sync
- Zebra full Testnet sync
- Lightwalletd full Mainnet sync (which depends on the zebra sync)

Fixes #6631 

### Specifications

We're using the `schedule` event of GitHub Actions: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

We have multiple tests that depend on each other, and the actual state version. This implementation can be improved, segmenting the jobs, but it requires a big refactor which we're not willing to do before our stable release.

## Solution

- Run the workflow on a `schedule`, weekly on Fridays at midday UTC
- Change the condition on the sync jobs which are being triggered on the push to main, to be on the `schedule` event
- Add some missing comments, and remove extra one.

## Review

Anyone can review this

### Reviewer Checklist

## Follow Up Work

The `schedule` will run all the jobs. We should ideally move this jobs to a different Workflow file, but this requires a more complex and bigger PR.


.github/workflows/continous-integration-docker.yml
```diff
@@ -8,6 +8,12 @@ concurrency:
   cancel-in-progress: true
 
 on:
+  schedule:
+    # Run this job every Friday at mid-day UTC
+    # This is limited to the Zebra Full Sync job
```
teor2345: Can you add the full list of jobs here?

```suggestion
    # This is limited to the Zebra and lightwalletd Full Sync jobs
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -552,18 +555,17 @@ jobs:
   # Test full sync of lightwalletd with a Zebra tip state
   #
   # Runs:
-  # - after every PR is merged to `main`
+  # - on schedule, as defined at the top of the workflow
+  # - on every PR update, but only if the state version in constants.rs has no cached disk
+  # - in manual workflow runs, when run-lwd-sync is 'true' and network is 'Mainnet'
```
teor2345: Optional

```suggestion
  # - in manual workflow runs, when run-lwd-sync is 'true' and network is 'Mainnet' (the network is required by the test-full-sync job)
```

teor2345 (commented):
Thanks! Let's check with the rest of the Zebra engineers before we merge, because we're in a release freeze.

None of my suggestions are blockers.

teor2345:
We discussed this in the Engineering sync and decided it was better to merge it before the stable release, so it's easier to see actual failures in CI.

teor2345 (approved):
Thanks, this looks great!

gustavovalverde:
@Mergifyio requeue

teor2345:
GitHub says they fixed an infrastructure issue 3 minutes ago:
https://www.githubstatus.com/history

teor2345:
@mergifyio refresh

PR #6923, opened 2023-06-12
fix(logs): Avoid grouping logs for separate tasks

arya2:
## Motivation

We've seen some malformed logs from the `peer_cache_updater` task:

> 2023-06-11T22:18:21.656595Z INFO {net="Test"}:accept_inbound_connections{min_inbound_peer_connection_interval=1s listener_addr=Ok(0.0.0.0:18233)}:listen_accept{peer=In("v4redacted:41812")}:crawl: zebra_network::peer_set::candidate_set: timeout waiting for peer service readiness or peer responses
> accept_inbound_connections{min_inbound_peer_connection_interval=1s listener_addr=Ok(0.0.0.0:18233)}:listen_accept{peer=In("v4redacted:47028")}: zebra_network::config: updated cached peer IP addresses cached_ip_count=11 peer_cache_file="/root/.cache/zebra/network/testnet.peers

## Solution

- Drop `Entered` guard in `accept_inbound_connections()` before yielding

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -603,8 +603,7 @@ where
             );
 
             let connected_addr = peer::ConnectedAddr::new_inbound_direct(addr);
-            let accept_span = info_span!("listen_accept", peer = ?connected_addr);
-            let _guard = accept_span.enter();
+            info!(peer = ?connected_addr, "listen_accept");
 
             debug!("got incoming connection");
             handshaker.ready().await?;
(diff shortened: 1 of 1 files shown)


zebra-network/src/peer_set/initialize.rs
```diff
@@ -607,6 +607,13 @@ where
             let _guard = accept_span.enter();
 
             debug!("got incoming connection");
+
+            // # Correctness
+            //
+            // Holding the drop guard returned by Span::enter across .await points will
+            // result in incorrect traces if the executor yields.
```
arya2:
```suggestion
            // result in incorrect traces if it yields.
```

teor2345 (approved):
Thanks!

Looks like manual span handling is an anti-pattern, we should use the instrument macros on split-out functions instead.

PR #6929, opened 2023-06-13
fix(cleanup): redundant configuration feature

oxarbitrage:
## Motivation

Minor mistake a i found which is not causing any issue but it is definitely redundant code.

This has no ticket associated with as it is too small.

## Solution

Fix the redudancy.

Check for other similars: There is no other find for `any(any` in the zebra codebase. 

## Review

Anyone can review.

### Reviewer Checklist


oxarbitrage:
Marked as `do not merge` by the release freeze.

teor2345 (approved):

PR #6930, opened 2023-06-13
fix(ci): Disable scheduled testnet full sync by default

teor2345:
## Motivation

The testnet full sync is currently extremely slow, we should stop running it until the network recovers:
https://github.com/ZcashFoundation/zebra/actions/runs/5240305921/jobs/9476158691#step:8:2240

## Solution

Disable the scheduled testnet full sync job unless the `SCHEDULE_TESTNET_FULL_SYNC` actions variable is set to `true`.

## Review

This needs to be merged before the scheduled job runs on Friday midday UTC.

### Reviewer Checklist

## Follow Up Work

If the full sync job needs fixes, we should fix it in another ticket.
If Zebra needs fixes on testnet, that should also be another ticket.


gustavovalverde (approved):
LGTM

PR #6933, opened 2023-06-13
fix(docker): Improve Dockerfile cache use by making build commands match exactly

teor2345:
## Motivation

I think Docker builds are slow because the cook and build/test commands are using different settings.

## Solution

Use exactly the same settings for cook and build/test
Add a TODO for --locked once it is supported 

## Review

This is important but not a release blocker.

We might want to merge it separately to PR #6934, so we can see the cache speed improvement. The full speed improvement won't be visible until this PR merges to `main`.

### Reviewer Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -107,10 +107,11 @@ COPY --from=us-docker.pkg.dev/zealous-zebra/zebra/lightwalletd /opt/lightwalletd
 #
 # This is the caching Docker layer for Rust!
 #
-# TODO: is it faster to use --tests here?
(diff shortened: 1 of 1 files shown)


gustavovalverde (changes requested):
I do agree with merging this, I'm just not approving yet as I'd like to see how #6934 behaves without this changes.

arya2 (approved):

gustavovalverde (approved):
I removed the dependency on PR #6934 as it will be more complex than initially expected

PR #6940, opened 2023-06-13
fix: prevent progress bar from panicking using workaround

### LINKED ISSUE #6926: Panic running 1.0.0-rc.9+23.gc058f77 with progress-bar feature
I got a crash while running Zebra with progress bars, I had it previously synced a month ago.

Log: 
[log.txt](https://github.com/ZcashFoundation/zebra/files/11728972/log.txt)

## Error
```
overflow when adding durations
```

## Metadata
|key|value|
|--|--|
|**version**|1.0.0-rc.9+23.gc058f77|
|**Zcash network**|Mainnet|
|**state version**|25|
|**features**|howudoin,indicatif,metrics_exporter_prometheus,progress_bar,prometheus|
|**branch**|main|
|**git commit**|c058f77|
|**commit timestamp**|2023-06-12T22:10:09.000000000Z|
|**rust compiler**|1.70.0|
|**rust release date**|2023-05-31|
|**optimization level**|3|
|**debug checks**|false|
|**location**|library/core/src/time.rs:920:31|

conradoplg:
## Motivation

Running Zebra with progress bar makes it panic after a while.

This seems to be caused by a bug in `howudoin`/`indicatif` where if a progress bar gets stuck, the ETA keeps growing and growing until it overflows. I'll eventually try to get a simple reproducer and report it, but in the meantime we can do a workaround.

## Solution

This removes the limit on the progress bars that are not guaranteed to complete, which removes the ETA and avoids the panic. The other progress bars (e.g. Blocks) could still panic if the node gets offline but this seems enough for now.

We could copy&paste and modify the `TermLine` implementation from `howudoin` to not display the ETA, but seems more work and would duplicate code.

They look like this with this PR:

```
 | Known Peers: 70                                                                                                    
 | Failed Peers: 3                                                                                                    
 | Inbound Connections: 0                                                                                             
 | Outbound Connections: 39                                                                                           
 / Queued Checkpoint Blocks (6y)
 [=============================>]  99% (2092738/2112016)                                                              
 / Verified Checkpoints (39w)
 [=============================>]  99% (10795/10875)                                                                  
 / Blocks (2h)
 [=============================>]  99% (2091696/2120401) Nu5                                                          
 | Never Attempted Peers: 27                                                             
 ```

Closes https://github.com/ZcashFoundation/zebra/issues/6926

## Review

This is not urgent since progress bars are optional. But I think it would be nice to get this in the stable release so that we can show the progress bars working during demos.

Anyone can review.

### Reviewer Checklist

## Follow Up Work

Need to report the bug in `indicatif` (and maybe propose a PR to fix), will do that later

### DIFF
--- a/zebra-consensus/src/checkpoint.rs
+++ b/zebra-consensus/src/checkpoint.rs
@@ -311,9 +311,7 @@ where
         if matches!(howudoin::cancelled(), Some(true)) {
             self.finish_diagnostics();
         } else {
-            self.queued_blocks_bar
-                .set_pos(max_queued_height.0)
-                .set_len(u64::from(self.checkpoint_list.max_height().0));
+            self.queued_blocks_bar.set_pos(max_queued_height.0);
         }
     }
 
--- a/zebra-consensus/src/checkpoint.rs
+++ b/zebra-consensus/src/checkpoint.rs
@@ -311,7 +311,9 @@ where
         if matches!(howudoin::cancelled(), Some(true)) {
             self.finish_diagnostics();
         } else {
-            self.queued_blocks_bar.set_pos(max_queued_height.0);
+            self.queued_blocks_bar
+                .set_pos(max_queued_height.0)
+                .set_len(u64::from(self.checkpoint_list.max_height().0));
         }
     }
 
--- a/zebra-state/src/service/non_finalized_state.rs
+++ b/zebra-state/src/service/non_finalized_state.rs
@@ -662,9 +662,7 @@ impl NonFinalizedState {
                 .best_chain()
                 .map(|chain| chain.non_finalized_root_height().0 - 1);
 
-            chain_count_bar
-                .set_pos(u64::try_from(self.chain_count()).expect("fits in u64"))
-                .set_len(u64::try_from(MAX_NON_FINALIZED_CHAIN_FORKS).expect("fits in u64"));
+            chain_count_bar.set_pos(u64::try_from(self.chain_count()).expect("fits in u64"));
 
             if let Some(finalized_tip_height) = finalized_tip_height {
                 chain_count_bar.desc(format!("Finalized Root {finalized_tip_height}"));
@@ -701,10 +699,7 @@ impl NonFinalizedState {
                 // - the chain this bar was previously assigned to might have changed position.
                 chain_length_bar
                     .label(format!("Fork {fork_height}"))
-                    .set_pos(u64::try_from(chain.len()).expect("fits in u64"))
-                    .set_len(u64::from(
-                        zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY,
-                    ));
+                    .set_pos(u64::try_from(chain.len()).expect("fits in u64"));
 
                 // display work as bits
                 let mut desc = format!(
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: zebra-consensus/src/checkpoint.rs [hunk 1]
```diff
@@ -311,9 +311,7 @@ where
         if matches!(howudoin::cancelled(), Some(true)) {
             self.finish_diagnostics();
         } else {
-            self.queued_blocks_bar
-                .set_pos(max_queued_height.0)
-                .set_len(u64::from(self.checkpoint_list.max_height().0));
+            self.queued_blocks_bar.set_pos(max_queued_height.0);
```
teor2345: The queued blocks progress bar is guaranteed to "complete". As long as Zebra is connected to the network, it will queue the final checkpoint, verify it, then shut down this progress bar.

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -701,10 +699,7 @@ impl NonFinalizedState {
                 // - the chain this bar was previously assigned to might have changed position.
                 chain_length_bar
                     .label(format!("Fork {fork_height}"))
-                    .set_pos(u64::try_from(chain.len()).expect("fits in u64"))
-                    .set_len(u64::from(
-                        zebra_chain::transparent::MIN_TRANSPARENT_COINBASE_MATURITY,
-                    ));
+                    .set_pos(u64::try_from(chain.len()).expect("fits in u64"));
```
teor2345, line 703: The chain length bar should either complete or get pruned within 100 blocks. If it doesn't, we should check if we're updating it before the block commit, rather than after. And we should check we're using the right constant, we might want MAX_BLOCK_REORG_HEIGHT instead.

teor2345 (commented):
Thanks for this fix!

I think we should restart the Zebra instance before the demo as a workaround.

The long-term fix would be disabling the ETA on those progress bars (an API change), and fixing the overflow bug in those libraries. But that's not a priority right now.

Would you mind moving these limits into comments? It's important that we document them anyway, and we might as well refer to the existing constants.

zebra-consensus/src/checkpoint.rs
conradoplg: Restored in 74365c42a2ac646690d6d7c0e1fdc840d5aca043

zebra-state/src/service/non_finalized_state.rs
conradoplg: It panicked here after a while, I think it stops at 99? For precaution I kept it without length.

conradoplg:
> Thanks for this fix!
> 
> I think we should restart the Zebra instance before the demo as a workaround.

It takes 4m30s to crash on my machine, so that may lead to a very stressful demo :sweat_smile: 
This seems a very low risk change, but it's up to you. We can always cherry pick this when doing any demos.

> Would you mind moving these limits into comments? It's important that we document them anyway, and we might as well refer to the existing constants.

Done in 74365c42a2ac646690d6d7c0e1fdc840d5aca043

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -701,10 +700,10 @@ impl NonFinalizedState {
...
+                // .set_len(u64::from(
```
teor2345, line 705:
```suggestion
                // TODO: should this be MAX_BLOCK_REORG_HEIGHT?
                // .set_len(u64::from(
```

teor2345:
> > Thanks for this fix!
> > I think we should restart the Zebra instance before the demo as a workaround.
> 
> It takes 4m30s to crash on my machine, so that may lead to a very stressful demo 😅 This seems a very low risk change, but it's up to you. We can always cherry pick this when doing any demos.

Oh wow, that's not what was happening when I wrote the feature. This seems like a new bug in the libraries we're using.

teor2345 (approved):
Thanks for the fix!

PR #6944, opened 2023-06-13
fix(build): Suppress warnings about doc links pointing to private items

upbqdn:
## Motivation

I was checking if things work before the upcoming release, and I noticed compiling the docs with 

``` bash
``` 

produces lots of warnings because our documentation for some public items contains links to private items. We thought we had these warnings suppressed in `Cargo.toml`, but it wasn't correct. Note that these warnings show up only locally because CI uses its own config.

## Solution

This PR (hopefully) correctly suppresses the warnings and updates CI.

### Reviewer Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -55,12 +55,6 @@ rustflags = [
     # Documentation
     "-Wmissing_docs",
 
-    # These rustdoc -A and -W settings must be the same as the RUSTDOCFLAGS in:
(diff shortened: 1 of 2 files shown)


upbqdn:
I don't know whether it's worth merging this PR before the stable release. I preemptively labeled it as `do-not-merge`.

arya2 (approved):

arya2:
Good catch!

teor2345:
CI failed due to bug #6948, it seems unrelated.

PR #6945, opened 2023-06-13
print a Zebra logo and some text if stderr is terminal

conradoplg:
## Motivation

The logo zcashd prints is fun, and I wanted something similar.

## Solution

This just prints a big static string before the progress bar, which just works. I used the same technique with `img2txt` that zcashd used.

This is a WIP because we need to decide:

- Whether to include in the stable release;
- What logos to print (Just zebra, zebra and heart, zebra and zcash?)
- What text to print (I just copied from zcashd, and removed the link to the security overview)
- Whether to mention if it's mainnet or testnet like zcashd does (we don't have this information easily available where the printing happens)

## Review

Not urgent unless we plan to include in stable release.

Depends on https://github.com/ZcashFoundation/zebra/pull/6940

### Reviewer Checklist


arya2:
Do we want to try [viuer](https://crates.io/crates/viuer) for full resolution images in supported environments? 

It supports transparent backgrounds too.

https://github.com/atanunq/viu

mpguerra:
I think this will be great for the next release :)

zebrad/src/components/tracing/component.rs
```diff
@@ -21,6 +21,9 @@ use crate::{application::app_version, components::tracing::Config};
 #[cfg(feature = "flamegraph")]
```
teor2345, line 22: Note for reviewers: the actual changes in this PR start here, the rest is from a merge conflict with another PR.

zebrad/src/components/tracing/component.rs
```diff
@@ -266,7 +269,19 @@ impl Tracing {
...
+                // Clear screen
+                eprint!("\x1B[2J");
+                {
+                    eprintln!(
+                        "{}",
+                        std::str::from_utf8(&ZEBRA_ART)
+                            .expect("should always work on a UTF-8 encoded constant")
+                    );
+                }
+                eprintln!("Thank you for running a zebrad {} node!", app_version());
```
teor2345: > * Whether to mention if it's mainnet or testnet like zcashd does (we don't have this information easily available where the printing happens)

Let's pass `cfg_ref.network.network` to the `Tracing::new()` method, then print `network.lowercase_name()` as part of this message?

zebrad/src/components/tracing/component.rs
conradoplg: rebased and fixed this

zebrad/src/components/tracing/component.rs
conradoplg: done in cacac5ac

conradoplg:
Now with a heart 
image

conradoplg:
> Do we want to try [viuer](https://crates.io/crates/viuer) for full resolution images in supported environments?
> 
> It supports transparent backgrounds too.
> 
> https://github.com/atanunq/viu

I like it but it seems overkill :rofl:

zebrad/src/components/tracing/component.rs
teor2345: Thanks!

teor2345:
> > Do we want to try [viuer](https://crates.io/crates/viuer) for full resolution images in supported environments?
> > It supports transparent backgrounds too.
> > [atanunq/viu](https://github.com/atanunq/viu)
> 
> I like it but it seems overkill 🤣

We can make this decision in a later ticket, it's not a breaking change.

zebrad/src/components/tracing/component.rs
```diff
@@ -266,7 +270,23 @@ impl Tracing {
             info!("activated progress bar");
 
             if config.log_file.is_some() {
-                eprintln!("waiting for initial progress reports...");
+                // Clear screen
```
teor2345: I don't think this logo needs to depend on the progress bar feature. We could activate it whenever the log file is enabled. This would also fix the "unused variable network" warnings.

zebrad/src/components/tracing/component.rs
```diff
@@ -266,7 +270,23 @@ impl Tracing {
...
+                eprint!("\x1B[2J");
```
teor2345: These escape codes should only be sent if output is going to a terminal.

Edit: the current code isn't correct

zebrad/src/components/tracing/component.rs
```diff
@@ -266,7 +270,23 @@ impl Tracing {
...
+                {
+                    eprintln!(
+                        "{}",
+                        std::str::from_utf8(&ZEBRA_ART)
+                            .expect("should always work on a UTF-8 encoded constant")
+                    );
+                }
+                eprintln!(
```
teor2345: These plain-text logs can be sent regardless of whether the output is a terminal.

zebrad/src/components/tracing/component.rs
teor2345: The `use_color` variable in this method is `true` if you can use terminal colour escapes.

It's refactored in PR #6997, but the behaviour of tracing doesn't change in that PR.

zebrad/src/components/tracing/component.rs
conradoplg: Good point. Thinking about it, I think it makes sense to print the logo if it's a terminal. It seems wasteful sending it to a log. And that's what zcashd does too. I did that on da446250

zebrad/src/components/tracing/component.rs
conradoplg: I did that on da446250, with the difference that I test stderr, because the logo is printed there (and that's because the progress bars are also printed there. Which makes sense overall)

zebrad/src/components/tracing/component.rs
conradoplg: Good point. Did that on da446250;

conradoplg (commented):
I've addressed comments, but with some changes. The intended behaviour is:

- stderr to terminal: print logo and message.
  If you redirect stdout to a file, then just the logo and message will be printed on the terminal which seem nice (in comparison to nothing being printed).
- stderr to file: only message is printed to stderr.
  Seems wasteful outputting the logo to anywhere not a terminal.
  
 If progress-bar is enabled, since it uses stderr and also checks for terminal, it will be printed after the logo as an "extension" of it, so the same reasoning applies.
 
 (I think I found a harmless bug in the progress bars: if you redirect stdout to a file, they lose color. That seems a bug in the libraries, I'll poke around later and report it)

teor2345:
> (I think I found a harmless bug in the progress bars: if you redirect stdout to a file, they lose color. That seems a bug in the libraries, I'll poke around later and report it)

I reported the same bug a while ago:
https://github.com/kdr-aus/how-u-doin/issues/2

I think the panic bug seems more critical to fix though!

teor2345 (approved):
This is looking good on my end.

Release manager note:

This is a user-visible feature. But in the past we haven't considered logging changes to be significant features. So I'm still happy to make the next release 1.0.1. (But I don't mind either way, open to suggestions.)

Default config:
Screenshot 2023-06-27 at 12 44 07

Default config with a log file:
Screenshot 2023-06-27 at 12 39 05

With `progress-bar` feature, initial sync:
Screenshot 2023-06-27 at 13 07 04



With `progress-bar` feature, end of initial sync:

Screenshot 2023-06-27 at 12 59 39

With `progress-bar` feature, finished sync, mempool active:

Screenshot 2023-06-27 at 13 01 18

PR #6969, opened 2023-06-15
fix(handshake): Add extra timeout logging to peer TCP connections

teor2345:
## Motivation

This PR adds logs for inbound and outbound timeouts, and fixes up some logging inconsistencies.

### Specifications

How TCP connect works in async Rust:
https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.connect

Tokio can yield in any library function:
https://docs.rs/tokio/latest/tokio/task/index.html#cooperative-scheduling

This is async code, but it's just adding (hopefully) redundant timeouts and splitting it into functions.

## Solution

Outbound connections:
- Add an extra 3 second timeout to TCP connect, to see if that's what's timing out

Inbound connections:
- Add an additional 3.5 second timeout to the spawned handshake task, and log if it fails. This will help locate inbound connection hangs.
- Replace the manual span enter with an instrumented macro.

Related cleanups:
- Delete a useless manual span in zebra_network::Config

I checked all the manual spans in zebra_network::AddressBook, they are all in non-async code, and there aren't any sync locks in those functions.

## Review

### Reviewer Checklist

## Follow Up Work

Use these logs to diagnose #6763.
Add extra logs to work out exactly where it's hanging.

### DIFF
--- a/zebra-network/src/config.rs
+++ b/zebra-network/src/config.rs
@@ -320,11 +320,6 @@ impl Config {
             Ok(Ok(ip_addrs)) => {
                 let ip_addrs: Vec<PeerSocketAddr> = ip_addrs.map(canonical_peer_addr).collect();
 
-                // if we're logging at debug level,
-                // the full list of IP addresses will be shown in the log message
-                let debug_span = debug_span!("", remote_ip_addrs = ?ip_addrs);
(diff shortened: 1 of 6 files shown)


upbqdn (approved):
Looks great!

Issue #6912, opened 2023-06-11
Change `ContextuallyVerifiedBlockWithTrees` into an enum to correctly represent verification statuses


teor2345:
## Motivation

`ContextuallyVerifiedBlockWithTrees` actually represents two different types with two different verification statuses. This is a maintainability issue.

### Design Background

Blocks from the checkpoint verifier are checkpoint-verified: they are semantically verified, but their transaction authorizing data isn't bound to their header.

Blocks from the non-finalized state are contextually verified: their transaction authorizing data is bound to their header. (And their trees have been calculated.)

### Complex Code or Requirements

The consensus code is correct, but the names and types we're using could cause bugs during future changes.

### Design

The current code is:
https://github.com/ZcashFoundation/zebra/blob/c058f77a342d5513420f50542593c11e76731a87/zebra-state/src/request.rs#L277-L282

The different verification statuses can be represented using an enum:
```rust
pub enum FinalizableBlock {
    Checkpoint {
        checkpoint_verified: CheckpointVerifiedBlock
    },
    Contextual {
        contextually_verified: ContextuallyVerifiedBlock,
        treestate: Treestate,
    }
}
```

Here are some example code changes that we could do:

In the match statement in `commit_finalized_direct()`:
- `FinalizableBlock::Checkpoint { checkpoint_verified }` replaces `ContextuallyVerifiedBlockWithTrees { checkpoint_verified, treestate: None }`
- `FinalizableBlock::Contextual { contextually_verified,  treestate }` replaces `ContextuallyVerifiedBlockWithTrees { checkpoint_verified, treestate: Some(treestate) }`

The `Checkpoint` case can construct a `ContextuallyVerifiedBlockWithTrees` from the checkpoint block and the trees in `commit_finalized_direct()`. Then the rest of the finalized state can use `ContextuallyVerifiedBlockWithTrees` instead of passing separate `finalized, history_tree, note_commitment_trees` arguments.

### Testing

Our existing tests should cover this change.

## Related Work

This is part of:
- #6680

mpguerra:
@upbqdn can you please add an estimate here?

PR #6971, opened 2023-06-15
change(state): Use `OrderedUtxo` in `CheckpointVerifiedBlock`

upbqdn:
## Motivation

After looking into #6912, I realized `CheckpointVerifiedBlock` and `SemanticallyVerifiedBlock` are the same, except that the first contains `Utxo`, and the second `OrderedUtxo`. This PR makes `CheckpointVerifiedBlock` use `OrderedUtxo` as well. This makes the block types the same, which will make the refactor of `ContextuallyVerifiedBlockWithTrees` simpler since `ContextuallyVerifiedBlock` uses `OrderedUtxo` as well.

## Solution

Use `OrderedUtxo` instead of `Utxo` in `CheckpointVerifiedBlock`.

### Reviewer Checklist

## Follow Up Work

- #7025

### DIFF
--- a/zebra-state/src/arbitrary.rs
+++ b/zebra-state/src/arbitrary.rs
@@ -183,7 +183,8 @@ impl CheckpointVerifiedBlock {
         height: block::Height,
     ) -> Self {
         let transaction_hashes: Arc<[_]> = block.transactions.iter().map(|tx| tx.hash()).collect();
-        let new_outputs = transparent::new_outputs_with_height(&block, height, &transaction_hashes);
+        let new_outputs =
+            transparent::new_ordered_outputs_with_height(&block, height, &transaction_hashes);
 
         Self {
             block,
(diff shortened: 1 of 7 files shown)


upbqdn:
The CI says "Some checks were not successful", but I didn't manage to find what's wrong.

teor2345:
> The CI says "Some checks were not successful", but I didn't manage to find what's wrong.

It's the testnet checkpoint job, due to ongoing testing on testnet. We can ignore these temporary failures. (I think they might be fixed by now, so feel free to restart that job if you want.)

I use Refined GitHub to make the failed jobs sort at the top, it works in every browser I've tried except Firefox:
https://github.com/refined-github/refined-github#-refined-github

teor2345:
Testnet jobs aren't required, so the PR will merge as soon as Alfredo approves it.

oxarbitrage (approved):
looks good to me, thanks.

PR #6980, opened 2023-06-15
fix(network): Limit number of peer connections per IP address, Ignore new peer connections from the same IP and port

### LINKED ISSUE #6936: Zebra node establishes multiple inbound or outbound connections to the same peer
### What happened?

I set up a small testnet with two nodes, zcash and zebra.
When I ask a zcash node to add a connection to a zebra node with an rpc `addnode 'ip:port' onetry` I expect that the zebra node creates only a single inbound connection to this peer. 

Instead, this happened:
After that inbound conn is set up, the zebra node establishes a second outbound connection to the same zcash peer node, by learning, I believe, its address from the version message.
I think, this is a sort of a problem as the second connection wastes a slot in the limited connection list.

When komodod or zcashd node gets an inbound connection they won't create an outbound connection to the same peer, I believe, thankfully to this code:
https://github.com/zcash/zcash/blob/fa3827656df8f97fb4700b22022cc581fbda61fa/src/net.cpp#L1663 (checking that there is no peer with this ip address already).
Maybe this check was not implemented deliberately in Zebra but I have not found any explanations for that so I'm creating this issue just for sure.

`zcash-cli addnode 'ip:port' onetry` to connect a zcash node to a zebra node

### Zebra logs

_No response_

### Zebra Version

Zebra:1.0.0-rc.9 (build from the main branch on 12 June 2023)

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

not os dependent

### Additional information

Sample getpeerinfo output from a zcash node from a small testnet, with both inbound and outbound conns to a zebra node (7999 port is used for listening) :
```
./zcash-cli -testnet getpeerinfo
[
  {
    "id": 1,
    "addr": "XX.XX.87.135:7999",
    "addrlocal": "XX.XX.238.59:40006",
    "services": "0000000000000001",
    ...
    "version": 170100,
    "subver": "/Zebra:1.0.0-rc.9/",
     "inbound": false,
     ...
  },
  {
    "id": 2,
    "addr": "XX.XX.87.135:48982",
    "addrlocal": "XX.XX.238.59:7999",
    "services": "0000000000000001",
    ...
    "version": 170100,
    "subver": "/Zebra:1.0.0-rc.9/",
    "inbound": true,
     ...
  }
]
```

arya2:
## Motivation

This PR drops additional peer connections from a given IP after there are already `MAX_CONNS_PER_IP` in the peer set.

Closes #6936.

## Solution

- Define `MAX_CONNS_PER_IP`
- Count the number of keys in `PeerSet.ready_services` and `PeerSet.cancel_handles` before inserting new connections into the peer set
- Drop new connections when there are already `MAX_CONNS_PER_IP`

## Review

### Reviewer Checklist

## Follow Up Work

- #6981 
- #6982

### DIFF
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -67,7 +67,7 @@ pub const INBOUND_PEER_LIMIT_MULTIPLIER: usize = 5;
 /// See [`INBOUND_PEER_LIMIT_MULTIPLIER`] for details.
 pub const OUTBOUND_PEER_LIMIT_MULTIPLIER: usize = 3;
 
-/// The maximum number of peer connections Zebra will for a given IP address
+/// The maximum number of peer connections Zebra will keep for a given IP address
 /// before it drops any additional peer connections with that IP.
 pub const MAX_CONNS_PER_IP: usize = 3;
 
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -67,6 +67,10 @@ pub const INBOUND_PEER_LIMIT_MULTIPLIER: usize = 5;
 /// See [`INBOUND_PEER_LIMIT_MULTIPLIER`] for details.
 pub const OUTBOUND_PEER_LIMIT_MULTIPLIER: usize = 3;
 
+/// The maximum number of peer connections Zebra will for a given IP address
+/// before it drops any additional peer connections with that IP.
+pub const MAX_CONNS_PER_IP: usize = 3;
+
 /// The buffer size for the peer set.
 ///
 /// This should be greater than 1 to avoid sender contention, but also reasonably
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -541,7 +541,7 @@ where
                     //
                     // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
                     // the same IP address in the peer set.
-                    if self.num_peers_with_ip(key.ip()) > self.max_conns_per_ip {
+                    if self.num_peers_with_ip(key.ip()) >= self.max_conns_per_ip {
                         std::mem::drop(svc);
                         continue;
                     }
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -476,6 +476,17 @@ where
         }
     }
 
+    /// Returns the number of peer connections Zebra already has with
+    /// the provided IP address
+    #[cfg(not(test))]
+    fn num_peers_with_ip(&self, ip: IpAddr) -> usize {
+        self.ready_services
+            .keys()
+            .chain(self.cancel_handles.keys())
+            .filter(|addr| addr.ip() == ip)
+            .count()
+    }
+
     /// Checks for newly inserted or removed services.
     ///
     /// Puts inserted services in the unready list.
@@ -497,6 +508,14 @@ where
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
                     self.remove(&key);
+
+                    // drop the new peer if its IP is in cancel_handles or ready_services,
+                    // this is skipped for tests so we can mock peer connections
+                    #[cfg(not(test))]
+                    if self.num_peers_with_ip(key.ip()) > crate::constants::MAX_CONNS_PER_IP {
+                        continue;
+                    }
+
                     self.push_unready(key, svc);
                 }
             }
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -507,13 +507,15 @@ where
                     // - always do the same checks on every ready peer, and
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
-                    self.remove(&key);
 
+                    // # Security
+                    //
                     // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
                     // the same IP address in the peer set
                     // this is skipped for tests so we can mock peer connections
                     #[cfg(not(test))]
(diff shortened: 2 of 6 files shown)


zebra-network/src/peer_set/set.rs
```diff
@@ -497,6 +508,14 @@ where
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
                     self.remove(&key);
+
+                    // drop the new peer if its IP is in cancel_handles or ready_services,
```
arya2:
```suggestion
                    // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
                    // the same IP address in the peer set
```

zebra-network/src/constants.rs
```diff
@@ -67,6 +67,10 @@ pub const INBOUND_PEER_LIMIT_MULTIPLIER: usize = 5;
 /// See [`INBOUND_PEER_LIMIT_MULTIPLIER`] for details.
 pub const OUTBOUND_PEER_LIMIT_MULTIPLIER: usize = 3;
 
+/// The maximum number of peer connections Zebra will for a given IP address
```
arya2:
```suggestion
/// The maximum number of peer connections Zebra will keep for a given IP address
```

zebra-network/src/peer_set/set.rs
```diff
@@ -497,6 +508,15 @@ where
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
                     self.remove(&key);
```
teor2345: This `remove()` call is now incorrect.

For security/reliability reasons, I think old connections with exactly the same IP and port should stay, rather than replacing them with new connections. This matches the behaviour of the IP-based security fix in this PR.

Otherwise, two nodes could get in a loop where they initiate new connections, and they continually replace the other node's connection.

```suggestion
```

zebra-network/src/peer_set/set.rs
```diff
@@ -497,6 +508,15 @@ where
...
+
+                    // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
+                    // the same IP address in the peer set
+                    // this is skipped for tests so we can mock peer connections
+                    #[cfg(not(test))]
+                    if self.num_peers_with_ip(key.ip()) > crate::constants::MAX_CONNS_PER_IP {
+                        continue;
```
teor2345, line 546: Nitpick: I prefer explicit drops, particularly in security fixes.

```suggestion
                        std::mem::drop(svc);
                        continue;
```

zebra-network/src/peer_set/set.rs
```diff
@@ -497,6 +508,15 @@ where
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
                     self.remove(&key);
+
+                    // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
```
teor2345, line 542:
```suggestion
                    // # Security
                    //
                    // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
```

teor2345 (changes requested):
I think this change needs some tests, what do you think?

zebra-network/src/peer_set/set.rs
```diff
@@ -496,7 +507,18 @@ where
...
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
-                    self.remove(&key);
+
+                    // # Security
+                    //
+                    // drop the new peer if there are already `MAX_CONNS_PER_IP` peers with
+                    // the same IP address in the peer set
+                    // this is skipped for tests so we can mock peer connections
+                    #[cfg(not(test))]
```
teor2345: I just realised that disabling this check in tests will also disable it in integration and acceptance tests. So we'll still get the testnet failures we're seeing on other PRs in CI.

zebra-network/src/peer_set/set.rs
```diff
@@ -496,7 +507,18 @@ where
                     // - always do the same checks on every ready peer, and
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
-                    self.remove(&key);
```
teor2345, line 499: I just realised my suggestion here is wrong, we still have to do something about duplicate connections to the same IP and port.

So I opened #6983.

arya2: > for outbound connections, this can cause connection thrashing, if two peers disagree on which of their connections is "latest"

Good thinking keeping the older connection.

zebra-network/src/peer_set/set.rs
arya2: Replaced with a configurable `max_conns_per_ip` in https://github.com/ZcashFoundation/zebra/pull/6980/commits/cbd9041affac75f684b41058379e17caa060743b.

teor2345: Thanks!

zebra-network/src/peer_set/set.rs
```diff
@@ -282,6 +287,9 @@ where
     /// - `inv_stream`: receives inventory changes from peers,
     ///                 allowing the peer set to direct inventory requests;
     /// - `address_book`: when peer set is busy, it logs address book diagnostics.
+    /// - `minimum_peer_version`: endpoint to see the minimum peer protocol version in real time.
```
teor2345, line 290: Technically minimum peer version updates are delayed until blocks are verified, even if the network has a higher minimum

```suggestion
    /// - `minimum_peer_version`: endpoint to see the minimum peer protocol version based on the latest state tip height.
```

teor2345 (approved):
Thanks for the tests!

zebra-network/src/peer_set/set.rs
teor2345: This isn't a blocker so I'm just going to merge the PR

PR #6983, opened 2023-06-16
fix(network): Ignore new peer connections from the same IP and port

### LINKED ISSUE #6936: Zebra node establishes multiple inbound or outbound connections to the same peer

teor2345:
## Motivation

Currently, Zebra replaces old peer connections with new peer connections from the same address in the peer set.

But this has two issues:
- for inbound connections, the node will use a different ephemeral port, so it still allows multiple connections (fixed by PR #6980) 
- for outbound connections, this can cause connection thrashing, if two peers disagree on which of their connections is "latest"

So it's better to keep the older connection, and rely on OS errors, the keepalive task, or connection timeouts to close it when it fails.

This changes Zebra's behaviour to match `zcashd`.

## Solution

- Close new connections to an existing peer on the same outbound IP and port, rather than dropping the old connection

## Review

This is part of #6980.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -478,6 +478,11 @@ where
 
     /// Returns the number of peer connections Zebra already has with
     /// the provided IP address
+    ///
+    /// # Performance
+    ///
+    /// This method is `O(connected peers)`, so it should not be called from a loop
+    /// that is already iterating through the peer set.
     #[cfg(not(test))]
     fn num_peers_with_ip(&self, ip: IpAddr) -> usize {
         self.ready_services
@@ -485,6 +490,11 @@ where
             .chain(self.cancel_handles.keys())
             .filter(|addr| addr.ip() == ip)
             .count()
+    }
+
+    /// Returns `true` if Zebra is already connected to the IP and port in `addr`.
+    fn has_peer_with_addr(&self, addr: PeerSocketAddr) -> bool {
+        self.ready_services.contains_key(&addr) || self.cancel_handles.contains_key(&addr)
     }
 
     /// Checks for newly inserted or removed services.
@@ -507,6 +517,15 @@ where
                     // - always do the same checks on every ready peer, and
                     // - check for any errors that happened right after the handshake
                     trace!(?key, "got Change::Insert from Discover");
+
+                    // # Security
+                    //
+                    // Drop the new peer if we are already connected to it.
+                    // Preferring old connections avoids connection thrashing.
+                    if self.has_peer_with_addr(key) {
+                        std::mem::drop(svc);
+                        continue;
+                    }
 
                     // # Security
                     //


teor2345:
Temporary google cloud error:
> There was a problem refreshing your current auth tokens: HTTPSConnectionPool(host='iamcredentials.googleapis.com', port=443): Max retries exceeded with url: /v1/projects/-/serviceAccounts/instance-deployer@zfnd-dev-zebra.iam.gserviceaccount.com:generateAccessToken (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f091000beb0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))

https://github.com/ZcashFoundation/zebra/actions/runs/5285113156/jobs/9564056694#step:10:77

arya2 (approved):

PR #6998, opened 2023-06-19
refactor(consensus): Rename router_verifier to block_verifier_router

teor2345:
## Motivation

`router_verifier` sounds like something that verifies routes or routers. But what it actually does it route blocks to other verifiers.

## Solution

Do a mass rename to `block_verifier_router` using these commands:

```sh
git ls-tree --full-tree -r --name-only HEAD | xargs sed -i -e 's/router_verifier/block_verifier_router/g'
cargo fmt --all
```

## Review

You can automatically review this code using the review instructions at:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/mass-renames.md#universal-renames-with-sed

### Reviewer Checklist


oxarbitrage (approved):

teor2345:
@mergifyio refresh

PR #7013, opened 2023-06-19
change(network): Configurable maximum connections per IP

arya2:
## Motivation

This could be helpful for integration testing.

## Solution

- Adds an optional `maximum_connections_per_ip` field to zebra-network config
- Adds the new generated configs to `zebrad/tests/common/configs`

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/config.rs
+++ b/zebra-network/src/config.rs
@@ -19,8 +19,8 @@ use zebra_chain::parameters::Network;
 
 use crate::{
     constants::{
-        DEFAULT_CRAWL_NEW_PEER_INTERVAL, DNS_LOOKUP_TIMEOUT, INBOUND_PEER_LIMIT_MULTIPLIER,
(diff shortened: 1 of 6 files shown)


teor2345 (dismissed):
Thanks, looks great!

We might want to think about how to name the config files when we don't know if the next release will be 1.0.1 or 1.1.0. But that's not an issue for this PR to solve!

teor2345 (dismissed):

teor2345 (approved):

teor2345:
This new config could be a new user-visible feature. So if it merges, that would make the next release a minor version (1.1.0).

This config is likely to be used in test environments. But we've also had production users say that they wanted node replication on internal networks or on the same machine, which would benefit from this config as well.

Just checking to see if anyone wants to call it a debugging feature instead?
If we want to do that, we could add "debug_" to the start of its name, and document that it is developer-focused.

(I don't mind either way, I'm just putting the question out there, so it's decided before the release.)

PR #7014, opened 2023-06-19
change(commands): Ignore error from loading config if running the 'generate' or 'download' commands

arya2:
## Motivation

Zebra should ignore errors from loading the config when generating a default config.

## Solution

- Ignore error from `load_config()` when the command is `Generate` and use default config

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebrad/src/application.rs
+++ b/zebrad/src/application.rs
@@ -14,7 +14,7 @@ use zebra_network::constants::PORT_IN_USE_ERROR;
 use zebra_state::constants::{DATABASE_FORMAT_VERSION, LOCK_FILE_ERROR};
 
 use crate::{
-    commands::{EntryPoint, ZebradCmd},
+    commands::EntryPoint,
     components::{sync::end_of_support::EOS_PANIC_MESSAGE_HEADER, tracing::Tracing},
     config::ZebradConfig,
 };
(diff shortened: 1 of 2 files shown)


zebrad/src/application.rs
```diff
@@ -235,11 +235,13 @@ impl Application for ZebradApp {
...
-        // report an error to the terminal if it occurs.
+        // report an error to the terminal if it occurs (unless used with the 'generate' cmd).
         let config = match command.config_path() {
-            Some(path) => match self.load_config(&path) {
-                Ok(config) => config,
-                Err(e) => {
+            Some(path) => match (&command.cmd, self.load_config(&path)) {
+                (_, Ok(config)) => config,
+                // Ignore errors loading the config when generating a default config.
+                (&Some(ZebradCmd::Generate(_)), Err(_e)) => Default::default(),
```
teor2345: `Download` doesn't use the config either, should we also ignore errors for that command?
Do we need to ignore errors for `help` and `--version`, or are they handled before loading the config?

teor2345: Optional: The list of commands could be implemented as a method on `ZebradCmd` instead, so that we are more likely to update it when we add new commands.

teor2345 (dismissed):
I'm going to approve this just in case all the commands are handled correctly already. But feel free to add more commands and ask for a re-review.

teor2345:
Oh and nice catch, by the way!

zebrad/src/application.rs
arya2: > Download doesn't use the config either, should we also ignore errors for that command?

Yeah.

> Do we need to ignore errors for help and --version, or are they handled before loading the config?

No, those are handled before it gets here (https://github.com/ZcashFoundation/zebra/blob/skip-loading-config/zebrad/src/application.rs#L519).

arya2: Good idea, applied in https://github.com/ZcashFoundation/zebra/pull/7014/commits/35f253f53a9872e99beab0ae72cfee6bcf1b8323.

zebrad/src/application.rs
```diff
@@ -235,10 +235,12 @@ impl Application for ZebradApp {
         let mut components = self.framework_components(command)?;
 
         // Load config *after* framework components so that we can
-        // report an error to the terminal if it occurs.
+        // report an error to the terminal if it occurs (unless used with the 'generate' cmd).
```
teor2345: Nitpick, I'll just merge and approve.

```suggestion
        // report an error to the terminal if it occurs (unless used with a command that doesn't need the config).
```

teor2345 (approved):

PR #6972, opened 2023-06-15
add(metrics): Track mempool actions and size bucketed by weight

str4d:
## Motivation

It is useful to have insight into the composition of the mempool along the axes used for the default block construction algorithm. The algorithm from the now-deployed [ZIP 317](https://zips.z.cash/zip-0317) takes into account both a transaction's weight, and the number of "unpaid actions" it contains.

## Solution

This PR adds the following new metrics:

- (gauge) `zcash.mempool.actions.unpaid { "bk" = ["< 0.2", "< 0.4", "< 0.6", "< 0.8", "< 1"] }`
- (gauge) `zcash.mempool.actions.paid`
- (gauge) `zcash.mempool.size.weighted { "bk" = ["< 1", "1", "> 1", "> 2", "> 3"] }`

These are proposed as common Zcash metrics (with a `zcash.` prefix). The corresponding `zcashd` PR is https://github.com/zcash/zcash/pull/6632.

### Reviewer Checklist

  - [x] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [x] Are the PR labels correct?
  - [x] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #7019, opened 2023-06-20
add(metrics): Track mempool actions and size bucketed by weight (copy of #6972, credit @str4d)

mpguerra:
## Motivation

This is a copy of #6972 so that tests can be run

### Reviewer Checklist


teor2345:
@mergifyio refresh

teor2345:
This seems like a real bug, but I don't know if it is in the generated proptest data, or in the code.

Multiple tests fail on the same Rust code in this PR:
> ---- components::mempool::storage::tests::prop::reject_lists_are_limited_insert_eviction stdout ----
> Message:  attempt to subtract with overflow
> Location: zebrad/src/components/mempool/storage/verified_set.rs:312
> 312 >             paid_actions += entry.conventional_actions - entry.unpaid_actions;

https://github.com/ZcashFoundation/zebra/actions/runs/5321638950/jobs/9636992835?pr=7019#step:10:3627

teor2345 (changes requested):
Panic on "attempt to subtract with overflow"

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -286,10 +286,110 @@ impl VerifiedSet {
...
+        // Track the sum of transaction sizes (the metric by which they are mainly limited) across
+        // several buckets.
+        let mut size_with_weight_lt1 = 0;
+        let mut size_with_weight_eq1 = 0;
+        let mut size_with_weight_gt1 = 0;
+        let mut size_with_weight_gt2 = 0;
+        let mut size_with_weight_gt3 = 0;
+
+        for entry in self.full_transactions() {
+            paid_actions += entry.conventional_actions - entry.unpaid_actions;
```
arya2, line 312:
```suggestion
            paid_actions = paid_actions
                .checked_add_unsigned(entry.conventional_actions)
                .and_then(|v| v.checked_sub_unsigned(entry.unpaid_actions))
                .unwrap_or_else(|| {
                    warn!(?paid_actions, ?entry, "attempt to add or subtract with overflow");
                    0
                });
```

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -286,10 +286,110 @@ impl VerifiedSet {
...
+        let mut unpaid_actions_with_weight_lt20pct = 0;
+        let mut unpaid_actions_with_weight_lt40pct = 0;
+        let mut unpaid_actions_with_weight_lt60pct = 0;
+        let mut unpaid_actions_with_weight_lt80pct = 0;
+        let mut unpaid_actions_with_weight_lt1 = 0;
+
+        // Track the total number of paid actions across all transactions in the mempool. This
+        // added to the bucketed unpaid actions above is equal to the total number of conventional
+        // actions in the mempool.
+        let mut paid_actions = 0;
```
arya2, line 301:
```suggestion
        let mut paid_actions: u32 = 0;
```

zebrad/src/components/mempool/storage/verified_set.rs
teor2345: This shouldn't be necessary, the number of conventional actions should be greater than or equal to the number of unpaid conventional actions. This is enforced by the consensus rules.

So it seems like we need to fix the test data generation instead.

zebrad/src/components/mempool/storage/verified_set.rs
teor2345: This should be the same type as conventional_actions and unpaid_actions. I think that's u32. (It definitely shouldn't be signed, negative values aren't valid.)

zebra-chain/src/transaction/unmined.rs
```diff
@@ -443,4 +452,20 @@ impl VerifiedUnminedTx {
 
         cost
     }
+
+    ///
+    #[cfg(any(test, feature = "proptest-impl"))]
+    pub fn fix_arbitrary_generated_action_overflows(&mut self) -> Self {
```
teor2345: Optional: This fix might be easier and more reliable as a manual Arbitrary implementation?

Using a random number generator inside a proptest will make it slow. Also, test seeds won't be able to reproduce failures because of the extra randomness.

But if we implement `Arbitrary for VerifiedUnminedTx`, it will be faster and we can reproduce failures.

It's been a while since I've done one of these, but here's a way to do it efficiently by turning everything over the total into a value between 0 and total:
```rust
any::<(u32, u32)>().prop_map(|(unpaid, total)| (unpaid % (total + 1), total))
```

But total could be `u32::MAX`, so we also need to limit `total + 1` to `u32::MAX`:
```rust
any::<(u32, u32)>().prop_map(|(unpaid, total)| (unpaid % total.saturating_add(1), total))
```

Here's an example of a manual `Arbitrary` implementation:
https://github.com/ZcashFoundation/zebra/blob/1e47fe5b2bb1b9579fedffbcf1d830194f753666/zebra-chain/src/sprout/arbitrary.rs#L10-L54

teor2345:
@oxarbitrage CI won't run on this branch until you fix the merge conflicts

zebrad/src/components/mempool/storage/verified_set.rs
teor2345: This is probably fine as it is.

zebra-chain/src/transaction/unmined.rs
```diff
@@ -443,4 +455,28 @@ impl VerifiedUnminedTx {
...
+    /// new conventional and unpaid actions where rules are considered.
+    #[cfg(any(test, feature = "proptest-impl"))]
+    pub fn fix_arbitrary_generated_action_overflows(&mut self) -> Self {
+        let mut rng = rand::thread_rng();
+
+        // The number of actions can't be greater than 2^16.
+        let conventional_actions = rng.gen_range(0..2 ^ 16);
+
+        // The number of unpaid actions can't be greater than `BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT`.
+        let mut unpaid_actions = rng.gen_range(0..zip317::BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT);
```
teor2345: `..value` means "less than `value`", but the verifier accepts "less than or equal to":
https://github.com/ZcashFoundation/zebra/blob/1e47fe5b2bb1b9579fedffbcf1d830194f753666/zebra-chain/src/transaction/unmined/zip317.rs#L181-L183

```suggestion
        let mut unpaid_actions = rng.gen_range(0..=zip317::BLOCK_PRODUCTION_UNPAID_ACTION_LIMIT);
```

zebra-chain/src/transaction/unmined.rs
```diff
@@ -443,4 +455,28 @@ impl VerifiedUnminedTx {
...
+
+    /// When deriving `Arbitrary` for `VerifiedUnminedTx` action fields can be generated
+    /// that are against consensus rules. This function prevent those issues by generating
+    /// new conventional and unpaid actions where rules are considered.
+    #[cfg(any(test, feature = "proptest-impl"))]
+    pub fn fix_arbitrary_generated_action_overflows(&mut self) -> Self {
+        let mut rng = rand::thread_rng();
+
+        // The number of actions can't be greater than 2^16.
+        let conventional_actions = rng.gen_range(0..2 ^ 16);
```
teor2345: In Rust, `^` means "bitwise exclusive or":
https://doc.rust-lang.org/std/ops/trait.BitXor.html

But the power of two operator is actually "bitwise shift left":
https://doc.rust-lang.org/std/ops/trait.Shl.html

Also the limit in the spec is `< 2^16`, rather than `<= 2^16`:

```suggestion
        // The number of actions must be less than 2^16.
        let conventional_actions = rng.gen_range(0..(1 << 16));
```

Or you can use `u16::MAX`:

```suggestion
        // The number of actions must be less than 2^16.
        let conventional_actions = rng.gen_range(0..(u16::MAX as u32));
```

teor2345 (dismissed):
Thanks for this!

It looks like all the tests are passing, but we're not quite testing the correct ranges.

Also we could make this implementation simpler and easier to use by replacing the derived `Arbitrary` implementation with a manual one. Did you want me to show you how to do that?

teor2345:
CI hit a GitHub HTTP request rate-limit:
>   /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=2 origin +1782753784d560a037dd6786bfae6780ba3693a6:refs/remotes/pull/7019/merge
>   Error: fatal: unable to access 'https://github.com/ZcashFoundation/zebra/': The requested URL returned error: 429

https://github.com/ZcashFoundation/zebra/actions/runs/5590449347/jobs/10221580034?pr=7019#step:2:54

The solution to this is to stop checking out Zebra in jobs and steps that don't need it, or to combine jobs and steps so that we check out Zebra fewer times.

teor2345 (approved):
Thanks for fixing this up!

PR #7029, opened 2023-06-21
fix(network): Avoid initiating outbound handshakes with IPs for which Zebra already has an active peer.

### LINKED ISSUE #6982: Don't initiate handshakes to outbound IPs that were recently connected
## Motivation

This prevents some kinds of attacks, by making Zebra skip the handshake for these outbound connections.

Edit: some possible attacks are:
- rapidly repeated outbound reconnections and handshakes to the same IP

Skip addresses in `CandidateSet::next()` if their IPs are recently connected, according to the AddressBook.

The iteration order should be carefully written so this new check is O(n) rather than O(n^2):
> In [computer science](https://en.wikipedia.org/wiki/Computer_science), big O notation is used to [classify algorithms](https://en.wikipedia.org/wiki/Computational_complexity_theory) according to how their run time or space requirements grow as the input size grows.

https://en.wikipedia.org/wiki/Computational_complexity_theory

### Testing

Try to make a lot of outbound connections:
- existing tests do this, their asserts can be changed
- we can copy the failing tests and adjust their test data to use different IPs to test this new code

arya2:
## Motivation

This prevents some kinds of attacks.

Closes #6982.

## Solution

- Adds a `most_recent_by_ip` field to `AddressBook`.
- Updates the new field when there's an update with a later `last_response` time, or when addresses are removed.
- Adds `&& !self.has_active_peer_with_ip()` to the filter in `reconnection_peers()`. 
- Tests that the iterator from `reconnection_peers()` skips over address when there's already a live address with the same IP.
 
## Review

### Reviewer Checklist

## Follow Up Work

- #7040

### DIFF
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -78,7 +78,7 @@ pub struct AddressBook {
     ///
     /// This is used to avoid initiating outbound connections past [`Config::max_connections_per_ip`](crate::config::Config), and
     /// currently only supports a `max_connections_per_ip` of 1, and must be `None` when used with a greater `max_connections_per_ip`.
-    // TODO: Replace with `by_ip` to support configured `max_connections_per_ip`
+    // TODO: Replace with `by_ip: HashMap<IpAddr, BTreeMap<DateTime32, MetaAddr>>` to support configured `max_connections_per_ip` greater than 1
     most_recent_by_ip: Option<HashMap<IpAddr, MetaAddr>>,
 
     /// The local listener address.
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -533,7 +533,7 @@ impl AddressBook {
     }
 
     /// Is this IP ready for a new outbound connection attempt?
-    /// Checks if the most recently updated address with this IP `was_recently_updated()`.
+    /// Checks if the most recently updated outbound address with this IP `was_recently_updated()`.
     ///
     /// Note: last_response times may remain live for a long time if the local clock is changed to an earlier time.
     fn is_ready_for_connection_attempt_with_ip(
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -203,6 +203,12 @@ impl AddressBook {
         for (socket_addr, meta_addr) in addrs {
             // overwrite any duplicate addresses
             new_book.by_addr.insert(socket_addr, meta_addr);
+            // Add the address to `most_recent_by_ip` if it has responded
+            if new_book.should_update_most_recent_by_ip(meta_addr) {
+                new_book
+                    .most_recent_by_ip
+                    .insert(socket_addr.ip(), meta_addr);
+            }
             // exit as soon as we get enough addresses
             if new_book.by_addr.len() >= addr_limit {
                 break;
@@ -435,7 +441,7 @@ impl AddressBook {
                 self.by_addr.remove(&surplus_peer.addr);
 
                 // Check if this surplus peer's addr matches that in `most_recent_by_ip`
-                // for this the surplus peer's ip.
+                // for this the surplus peer's ip to remove it there as well.
                 if self.is_addr_most_recent_by_ip(surplus_peer.addr) {
                     self.most_recent_by_ip.remove(&surplus_peer.addr.ip());
                 }
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -358,9 +358,10 @@ impl AddressBook {
         };
 
         if let Some(previous) = most_recent_by_ip.get(&updated.addr.ip()) {
-            updated.last_response() > previous.last_response()
+            updated.last_connection_state == PeerAddrState::Responded
+                && updated.last_response() > previous.last_response()
         } else {
-            true
+            updated.last_connection_state == PeerAddrState::Responded
         }
     }
 
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -3,8 +3,9 @@
 
 use std::{
     cmp::Reverse,
+    collections::HashMap,
     iter::Extend,
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     sync::{Arc, Mutex},
     time::Instant,
 };
@@ -71,6 +72,9 @@ pub struct AddressBook {
     /// ([`BTreeMap`](std::collections::BTreeMap)) sorts in ascending order, but
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
+
+    /// The most recently updated addresses by ip.
+    most_recent_by_ip: HashMap<IpAddr, MetaAddr>,
 
     /// The local listener address.
     local_listener: SocketAddr,
@@ -149,6 +153,7 @@ impl AddressBook {
             span,
             address_metrics_tx,
             last_address_log: None,
+            most_recent_by_ip: HashMap::new(),
         };
 
         new_book.update_metrics(instant_now, chrono_now);
@@ -314,6 +319,33 @@ impl AddressBook {
         meta_addr
     }
 
+    /// Returns true if `updated` has a last_connection_state of `Responded`
+    /// and there is either:
+    /// - no previous addr with the updated.addr IP, or
+    /// - updated has a more recent last_response time than the existing [`MetaAddr`]
+    ///   for that IP in `most_recent_by_ip`.
+    fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
+        updated
+            .last_response()
+            .is_some_and(|updated_last_response| {
+                self.most_recent_by_ip
+                    .get(&updated.addr.ip())
+                    .map_or(true, |prev| {
+                        prev.last_response() < Some(updated_last_response)
+                    })
+            })
+    }
+
+    /// Returns true if the provided `addr` matches that in `most_recent_by_ip`
+    /// for the IP of the address.
+    fn is_addr_most_recent_by_ip(&self, addr: PeerSocketAddr) -> bool {
+        self.most_recent_by_ip
+            .get(&addr.ip())
+            .map_or(false, |most_recent_peer_at_ip_addr| {
+                most_recent_peer_at_ip_addr.addr == addr
+            })
+    }
+
     /// Apply `change` to the address book, returning the updated `MetaAddr`,
     /// if the change was valid.
     ///
@@ -373,6 +405,11 @@ impl AddressBook {
 
             self.by_addr.insert(updated.addr, updated);
 
+            // Add the address to `most_recent_by_ip` if it has responded
+            if self.should_update_most_recent_by_ip(updated) {
+                self.most_recent_by_ip.insert(updated.addr.ip(), updated);
+            }
+
             debug!(
                 ?change,
                 ?updated,
@@ -396,6 +433,12 @@ impl AddressBook {
                     .expect("just checked there is at least one peer");
 
                 self.by_addr.remove(&surplus_peer.addr);
+
+                // Check if this surplus peer's addr matches that in `most_recent_by_ip`
+                // for this the surplus peer's ip.
+                if self.is_addr_most_recent_by_ip(surplus_peer.addr) {
+                    self.most_recent_by_ip.remove(&surplus_peer.addr.ip());
+                }
 
                 debug!(
                     surplus = ?surplus_peer,
@@ -463,6 +506,13 @@ impl AddressBook {
         self.by_addr.descending_values().cloned()
     }
 
+    /// Returns true if there is a recently live peer connection with the provided IP address.
+    fn has_active_peer_with_ip(&self, ip: &IpAddr, chrono_now: chrono::DateTime<Utc>) -> bool {
+        self.most_recent_by_ip
+            .get(ip)
+            .map_or(false, |peer| peer.was_recently_live(chrono_now))
+    }
+
     /// Return an iterator over peers that are due for a reconnection attempt,
     /// in reconnection attempt order.
     pub fn reconnection_peers(
@@ -478,6 +528,7 @@ impl AddressBook {
             .descending_values()
             .filter(move |peer| {
                 peer.is_ready_for_connection_attempt(instant_now, chrono_now, self.network)
+                    && !self.has_active_peer_with_ip(&peer.addr.ip(), chrono_now)
             })
             .cloned()
     }
@@ -699,6 +750,7 @@ impl Clone for AddressBook {
             span: self.span.clone(),
             address_metrics_tx,
             last_address_log: None,
+            most_recent_by_ip: self.most_recent_by_ip.clone(),
         }
     }
 }
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -73,7 +73,7 @@ pub struct AddressBook {
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
 
-    /// The most recently updated addresses by ip.
+    /// The address with a last_connection_state of [`Responded`] and the most recent `last_response` time by IP.
     most_recent_by_ip: HashMap<IpAddr, MetaAddr>,
 
     /// The local listener address.
@@ -331,15 +331,16 @@ impl AddressBook {
     /// - updated has a more recent last_response time than the existing [`MetaAddr`]
     ///   for that IP in `most_recent_by_ip`.
     fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
-        updated
-            .last_response()
-            .is_some_and(|updated_last_response| {
-                self.most_recent_by_ip
-                    .get(&updated.addr.ip())
-                    .map_or(true, |prev| {
-                        prev.last_response() < Some(updated_last_response)
-                    })
-            })
+        updated.last_connection_state == PeerAddrState::Responded
+            && updated
+                .last_response()
+                .is_some_and(|updated_last_response| {
+                    self.most_recent_by_ip
+                        .get(&updated.addr.ip())
+                        .map_or(true, |prev| {
+                            prev.last_response() < Some(updated_last_response)
+                        })
+                })
     }
 
     /// Returns true if the provided `addr` matches that in `most_recent_by_ip`
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -336,7 +336,12 @@ impl AddressBook {
     /// or if `updated` has a more recent update requiring the outbound connector to wait
     /// longer before initiating handshakes with peers at this IP.
     ///
-    /// See [`MetaAddr::is_ready_for_connection_attempt`] for more details.
+    /// This code only needs to check a single cache entry, rather than the entire address book,
+    /// because other code maintains these invariants:
+    /// - `last_attempt`, `last_failure`, and `last_response` times for an entry can only increase.
+    /// - these are the only fields checked by `was_recently_updated()`
+    ///
+    /// See [`AddressBook::is_ready_for_connection_attempt_with_ip`] for more details.
     fn should_update_most_recent_by_ip(
         &self,
         updated: MetaAddr,
@@ -350,8 +355,8 @@ impl AddressBook {
         }
     }
 
-    /// Returns true if the provided `addr` matches that in `most_recent_by_ip`
-    /// for the IP of the address.
+    /// Returns true if `addr` is the latest entry for its IP, which is stored in `most_recent_by_ip`.
+    /// The entry is checked for an exact match to the IP and port of `addr`.
     fn is_addr_most_recent_by_ip(&self, addr: PeerSocketAddr) -> bool {
         self.most_recent_by_ip
             .get(&addr.ip())
@@ -529,6 +534,8 @@ impl AddressBook {
 
     /// Is this IP ready for a new outbound connection attempt?
     /// Checks if the most recently updated address with this IP `was_recently_updated()`.
+    ///
+    /// Note: last_response times may remain live for a long time if the local clock is changed to an earlier time.
     fn is_ready_for_connection_attempt_with_ip(
         &self,
         ip: &IpAddr,
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -73,7 +73,8 @@ pub struct AddressBook {
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
 
-    /// The address with a last_connection_state of [`Responded`] and the most recent `last_response` time by IP.
+    /// The address with a last_connection_state of [`PeerAddrState::Responded`] and
+    /// the most recent `last_response` time by IP.
     most_recent_by_ip: HashMap<IpAddr, MetaAddr>,
 
     /// The local listener address.
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -330,7 +330,9 @@ impl AddressBook {
         meta_addr
     }
 
-    /// Returns true if there are no existing entries in the address book with this IP,
+    /// Returns true if `updated` needs to be applied to the recent outbound peer connection IP cache.
+    ///
+    /// Checks if there are no existing entries in the address book with this IP,
     /// or if `updated` has a more recent update requiring the outbound connector to wait
     /// longer before initiating handshakes with peers at this IP.
     ///
@@ -526,13 +528,14 @@ impl AddressBook {
     }
 
     /// Is this IP ready for a new outbound connection attempt?
+    /// Checks if the most recently updated address with this IP `was_recently_updated()`.
     fn is_ready_for_connection_attempt_with_ip(
         &self,
         ip: &IpAddr,
         instant_now: Instant,
         chrono_now: chrono::DateTime<Utc>,
     ) -> bool {
-        self.most_recent_by_ip.get(ip).map_or(false, |peer| {
+        self.most_recent_by_ip.get(ip).map_or(true, |peer| {
             !peer.was_recently_updated(instant_now, chrono_now)
         })
     }
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -217,7 +217,7 @@ impl AddressBook {
             // overwrite any duplicate addresses
             new_book.by_addr.insert(socket_addr, meta_addr);
             // Add the address to `most_recent_by_ip` if it has responded
-            if new_book.should_update_most_recent_by_ip(meta_addr, instant_now, chrono_now) {
+            if new_book.should_update_most_recent_by_ip(meta_addr) {
                 new_book
                     .most_recent_by_ip
                     .as_mut()
@@ -343,27 +343,22 @@ impl AddressBook {
     /// Returns true if `updated` needs to be applied to the recent outbound peer connection IP cache.
     ///
     /// Checks if there are no existing entries in the address book with this IP,
-    /// or if `updated` has a more recent update requiring the outbound connector to wait
+    /// or if `updated` has a more recent `last_response` requiring the outbound connector to wait
     /// longer before initiating handshakes with peers at this IP.
     ///
     /// This code only needs to check a single cache entry, rather than the entire address book,
     /// because other code maintains these invariants:
-    /// - `last_attempt`, `last_failure`, and `last_response` times for an entry can only increase.
-    /// - these are the only fields checked by `was_recently_updated()`
+    /// - `last_response` times for an entry can only increase.
+    /// - this is the only field checked by `has_connection_recently_responded()`
     ///
     /// See [`AddressBook::is_ready_for_connection_attempt_with_ip`] for more details.
-    fn should_update_most_recent_by_ip(
-        &self,
-        updated: MetaAddr,
-        now: Instant,
-        chrono_now: chrono::DateTime<Utc>,
-    ) -> bool {
+    fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
         let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
             return false
         };
 
         if let Some(previous) = most_recent_by_ip.get(&updated.addr.ip()) {
-            updated.gt_most_recent_update(previous, now, chrono_now)
+            updated.last_response() > previous.last_response()
         } else {
             true
         }
@@ -444,7 +439,7 @@ impl AddressBook {
 
             // Add the address to `most_recent_by_ip` if it sent the most recent
             // response Zebra has received from this IP.
-            if self.should_update_most_recent_by_ip(updated, instant_now, chrono_now) {
+            if self.should_update_most_recent_by_ip(updated) {
                 self.most_recent_by_ip
                     .as_mut()
                     .expect("should be some when should_update_most_recent_by_ip is true")
@@ -557,19 +552,18 @@ impl AddressBook {
     }
 
     /// Is this IP ready for a new outbound connection attempt?
-    /// Checks if the most recently updated outbound address with this IP `was_recently_updated()`.
+    /// Checks if the outbound connection with the most recent response at this IP has recently responded.
     ///
     /// Note: last_response times may remain live for a long time if the local clock is changed to an earlier time.
     fn is_ready_for_connection_attempt_with_ip(
         &self,
         ip: &IpAddr,
-        instant_now: Instant,
         chrono_now: chrono::DateTime<Utc>,
     ) -> bool {
         self.most_recent_by_ip
             .as_ref()
             .and_then(|most_recent_by_ip| most_recent_by_ip.get(ip))
-            .filter(|peer| peer.was_recently_updated(instant_now, chrono_now))
+            .filter(|peer| peer.has_connection_recently_responded(chrono_now))
             .is_none()
     }
 
@@ -588,11 +582,7 @@ impl AddressBook {
             .descending_values()
             .filter(move |peer| {
                 peer.is_ready_for_connection_attempt(instant_now, chrono_now, self.network)
-                    && self.is_ready_for_connection_attempt_with_ip(
-                        &peer.addr.ip(),
-                        instant_now,
-                        chrono_now,
-                    )
+                    && self.is_ready_for_connection_attempt_with_ip(&peer.addr.ip(), chrono_now)
             })
             .cloned()
     }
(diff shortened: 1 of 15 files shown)


zebra-network/src/address_book.rs
```diff
@@ -72,6 +73,9 @@ pub struct AddressBook {
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
 
+    /// The most recently updated addresses by ip.
```
arya2:
```suggestion
    /// The address with a last_connection_state of [`Responded`] and the most recent `last_response` time by IP.
```

zebra-network/src/address_book.rs
```diff
@@ -314,6 +325,33 @@ impl AddressBook {
         meta_addr
     }
 
+    /// Returns true if `updated` has a last_connection_state of `Responded`
+    /// and there is either:
+    /// - no previous addr with the updated.addr IP, or
+    /// - updated has a more recent last_response time than the existing [`MetaAddr`]
+    ///   for that IP in `most_recent_by_ip`.
+    fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
+        updated
```
arya2, line 334: `last_response()` will currently only return `Some(_)` when `last_connection_state` is `Responded`, but in case that changes:

```suggestion
        updated.last_connection_state == PeerAddrState::Responded
            && updated
```

zebra-network/src/address_book.rs
```diff
@@ -314,6 +325,33 @@ impl AddressBook {
...
+            .last_response()
+            .is_some_and(|updated_last_response| {
+                self.most_recent_by_ip
+                    .get(&updated.addr.ip())
+                    .map_or(true, |prev| {
+                        prev.last_response() < Some(updated_last_response)
+                    })
+            })
```
arya2: `last_response()` will currently only return `Some(_)` when `last_connection_state` is `Responded`, but in case that changes:

```suggestion
        updated.last_connection_state == PeerAddrState::Responded
            && updated
                .last_response()
                .is_some_and(|updated_last_response| {
                    self.most_recent_by_ip
                        .get(&updated.addr.ip())
                        .map_or(true, |prev| {
                            prev.last_response() < Some(updated_last_response)
                        })
                })
```

zebra-network/src/address_book/tests/vectors.rs
```diff
@@ -110,3 +110,38 @@ fn address_book_peer_order() {
...
+            Instant::now(),
+            Utc::now().try_into().expect("will succeed until 2038"),
+        );
+    let meta_addr2 = MetaAddr::new_gossiped_meta_addr(
+        addr2,
+        PeerServices::NODE_NETWORK,
+        DateTime32::MIN.saturating_add(Duration32::from_seconds(1)),
+    );
+
+    // Regardless of the order of insertion, the most recent address should be chosen first
```
arya2:
```suggestion
    // The second address should be skipped because the first address has a
    // recent `last_response` time and the two addresses have the same IP.
```

zebra-network/src/address_book.rs
```diff
@@ -72,6 +73,10 @@ pub struct AddressBook {
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
 
+    /// The address with a last_connection_state of [`PeerAddrState::Responded`] and
+    /// the most recent `last_response` time by IP.
```
teor2345, line 77: `HashMap<IpAddr, MetaAddr>` only supports `max_conns_per_ip = 1`. That's ok for now, but needs to be documented here, on the constant, and in that user config field's docs.

Or we could mark that config as unsupported (developer/tester-only).

Supporting larger values is a different ticket with a different priority, that's not a security issue. And selfishly, I don't want to do that review right now, there's a lot of other things going on!

zebra-network/src/address_book.rs
```diff
@@ -397,6 +442,12 @@ impl AddressBook {
 
                 self.by_addr.remove(&surplus_peer.addr);
 
+                // Check if this surplus peer's addr matches that in `most_recent_by_ip`
+                // for this the surplus peer's ip to remove it there as well.
+                if self.is_addr_most_recent_by_ip(surplus_peer.addr) {
```
teor2345: `AddressBook::take()` also needs to remove the IP from the cache. Are there any other methods that remove addresses from the address book?

zebra-network/src/address_book.rs
```diff
@@ -314,6 +326,34 @@ impl AddressBook {
         meta_addr
     }
 
+    /// Returns true if `updated` has a last_connection_state of `Responded`
+    /// and there is either:
+    /// - no previous addr with the updated.addr IP, or
+    /// - updated has a more recent last_response time than the existing [`MetaAddr`]
+    ///   for that IP in `most_recent_by_ip`.
+    fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
+        updated.last_connection_state == PeerAddrState::Responded
```
teor2345: This condition implies an invariant, but I'm not sure exactly what it is.

Is it something like:
* the the only updates which change `last_response` are updates to the `Responded` state
* the only field checked by `was_recently_live()` is `last_response`

Are these invariants already documented, implemented, and tested somewhere else?

teor2345: An IP that is currently in the `AttemptPending` state should also prevent other connections to that IP from being initiated. See my note below about how to implement this check.

What's the equivalent invariant for the `AttemptPending` state?

This will need a method docs update (but it doesn't need to be detailed, that's what the code and code comments are for).

zebra-network/src/address_book.rs
```diff
@@ -314,6 +326,34 @@ impl AddressBook {
...
+                    self.most_recent_by_ip
+                        .get(&updated.addr.ip())
+                        .map_or(true, |prev| {
+                            prev.last_response() < Some(updated_last_response)
+                        })
+                })
+    }
+
+    /// Returns true if the provided `addr` matches that in `most_recent_by_ip`
+    /// for the IP of the address.
```
teor2345: Nitpick: this doc is a bit unclear to me. I had to read the code to find out it was an exact IP and port match, rather than just an IP address match. ("Address" can mean either.)

It's tricky to explain, here's my attempt at it:
```suggestion
    /// Returns true if `addr` is the latest entry for its IP, which is stored in `most_recent_by_ip`.
    /// The entry is checked for an exact match to the IP and port of `addr`.
```

zebra-network/src/address_book.rs
```diff
@@ -314,6 +326,34 @@ impl AddressBook {
         meta_addr
     }
 
+    /// Returns true if `updated` has a last_connection_state of `Responded`
```
teor2345: Nitpick: Can we add a summary line to explain what the method does?

```suggestion
    /// Returns true if `updated` needs to be applied to the recent outbound peer connection IP cache.
    ///
    /// Cache entries are updated if they have a last_connection_state of `Responded`
```

zebra-network/src/address_book.rs
```diff
@@ -463,6 +514,13 @@ impl AddressBook {
         self.by_addr.descending_values().cloned()
     }
 
+    /// Returns true if there is a recently live peer connection with the provided IP address.
+    fn has_active_peer_with_ip(&self, ip: &IpAddr, chrono_now: chrono::DateTime<Utc>) -> bool {
+        self.most_recent_by_ip
+            .get(ip)
+            .map_or(false, |peer| peer.was_recently_live(chrono_now))
```
teor2345: We also need to check `peer.was_connection_recently_attempted()` here, and update the method docs.

zebra-network/src/address_book/tests/vectors.rs
```diff
@@ -110,3 +110,39 @@ fn address_book_peer_order() {
         Some(meta_addr2),
     );
 }
+
+/// Check that `reconnection_peers` skips addresses with IPs for which
+/// Zebra already has recently connected peers.
+#[test]
+fn reconnection_peers_skips_live_ip() {
```
teor2345: Can we have a similar test with AttemptPending, and one where the cache entry was recently Attempted/Responded, but has now Failed?

zebra-network/src/address_book.rs
teor2345: Should moving that `MetaAddr.addr` to a `Failed` state invalidate the cached entry, and replace it with the next oldest entry for that IP?

Or is there some invariant that makes it ok to keep the latest entry, even if its state has changed since?

I think the invariant is something like:
* `last_response` and `last_attempt` times only increase

But I can't remember if that holds after our latest fixes.

zebra-network/src/address_book.rs
teor2345: If we're going to break support for initiating more than 1 outbound connection to an IP, we might want to delay merging this PR until after the next release. It's hard to say though, because it's only really important for testing.

What do you think users would want? ("They won't care" is an ok answer.)

zebra-network/src/address_book.rs
teor2345: Note: `last_response` times can remain live for a long time if the local clock is changed to a much earlier time. There's not much we can do about that, but it might be worth documenting, since checking by IP makes the problem slightly worse.

(It impacts all nodes with that IP, not just the node with that exact IP and port.)

teor2345 (changes requested):
This is a nice minimal design!

I think checking `AttemptPending` peers is a blocking change. If we decide to put this PR in a later release, that could also be blocking.

The rest of my comments provide context, ask for more context so we can check the design works as expected, or ask for docs so future developers and auditors can do those checks.

Feel free to ask me to find some of these details myself.

zebra-network/src/address_book.rs
```diff
@@ -463,6 +514,13 @@ impl AddressBook {
         self.by_addr.descending_values().cloned()
     }
 
+    /// Returns true if there is a recently live peer connection with the provided IP address.
```
teor2345:
```suggestion
    /// Returns true if there is a recently live outbound peer connection with the provided IP address.
```

zebra-network/src/address_book/tests/vectors.rs
```diff
@@ -110,3 +110,39 @@ fn address_book_peer_order() {
         Some(meta_addr2),
     );
 }
+
+/// Check that `reconnection_peers` skips addresses with IPs for which
+/// Zebra already has recently connected peers.
```
teor2345:
```suggestion
/// Zebra already has recently connected outbound peers.
```

zebra-network/src/address_book.rs
```diff
@@ -3,8 +3,9 @@
```
teor2345, line 3: The title of this PR needs a tweak, because the `AddressBook` only stores the addresses of outbound peer connections.

So this PR only prevents rapidly repeated outbound reconnections and handshakes to the same IP. (Which is the attack I'm trying to address, and what I should have written in the ticket.)

It's ok to leave any duplicate outbound/inbound IPs for the peer set to drop. That would be a different attack, which seems much less plausible to me. But if you can think of a high-impact attack across inbound and outbound connections, please open a ticket!

zebra-network/src/address_book.rs
arya2: > It's hard to say though, because it's only really important for testing.

I could open a suggestion PR replacing `has_active_peer_with_ip()` with `num_active_peers_with_ip()` to make the review a little easier if you think it may be worth doing for the tests?

> What do you think users would want?

I think users would expect `max_conns_per_ip` to at least configure the number of connections Zebra will accept, but that it would be better for it to define how many connections Zebra will try to initiate as well. It's not worth doing immediately unless we need it for integration tests.

zebra-network/src/address_book.rs
arya2: > Are these invariants already documented, implemented, and tested somewhere else?

It's the `MetaAddr` that's most likely to fail the `!MetaAddr.has_connection_recently_responded()` check in `MetaAddr.is_ready_for_connection_attempt()`. I'll add that to the method's docs.

The redundant `self.last_connection_state == PeerAddrState::Responded` is from `was_recently_live()` (though I'm not sure if it's there for clarity or to short-circuit).

zebra-network/src/address_book.rs
arya2: > AddressBook::take() also needs to remove the IP from the cache.

Thank you.

> Are there any other methods that remove addresses from the address book?

No, the only other remove is in `get()` where it's immediately re-inserted.

zebra-network/src/address_book.rs
teor2345: > > It's hard to say though, because it's only really important for testing.
> 
> I could open a suggestion PR replacing `has_active_peer_with_ip()` with `num_active_peers_with_ip()` to make the review a little easier if you think it may be worth doing for the tests?

That's a different change with a lower priority, because it's not a security fix. Also, I'd prefer we all shift to focusing on the RPC work soon, rather than adding more network changes. (I'd like to get our current set of changes tested in a release first.)

But ultimately this is @mpguerra's decision, and she uses tickets to make those decisions. 

> > What do you think users would want?
> 
> I think users would expect `max_conns_per_ip` to at least configure the number of connections Zebra will accept, but that it would be better for it to define how many connections Zebra will try to initiate as well. It's not worth doing immediately unless we need it for integration tests.

I think you're right, and it's ok as long as we document it.

zebra-network/src/address_book.rs
teor2345: > > Are there any other methods that remove addresses from the address book?
> 
> No, the only other remove is in `get()` where it's immediately re-inserted.

Yes, as you can see I dislike that particular collection API 😂

zebra-network/src/address_book.rs
teor2345: Hmm, I'm not sure if were talking about the same things here.

What I'm thinking of is documentation like:

> This code only needs to check a single cache entry, rather than the entire address book, because other code maintains these invariants:
> * the the only updates which change last_response are updates to the Responded state
> * the only field checked by was_recently_live() is last_response

arya2: What about `Failed`?

teor2345: An ~~IP~~ peer address and port that's currently in the `Failed` state can't have an outbound connection:

https://github.com/ZcashFoundation/zebra/blob/31bc46bb2a736110d215de2fac5cdfd79bcbd058/zebra-network/src/meta_addr.rs#L66-L68

Edit: it's based on address and port, not just IP

arya2: > Should moving that MetaAddr.addr to a Failed state invalidate the cached entry, and replace it with the next oldest entry for that IP?

I don't think so, it should always prefer the `meta_addr` that will fail `is_ready_for_connection_attempt()` the longest, including `last_failure` times, and it should check all of the times that `is_ready_for_connection_attempt()` checks in a renamed `has_active_peer_with_ip()`.

> last_response and last_attempt times only increase

That's still the case.

zebra-network/src/address_book.rs
```diff
@@ -463,6 +521,18 @@ impl AddressBook {
...
     }
 
+    /// Is this IP ready for a new outbound connection attempt?
+    fn is_ready_for_connection_attempt_with_ip(
+        &self,
+        ip: &IpAddr,
+        instant_now: Instant,
+        chrono_now: chrono::DateTime<Utc>,
+    ) -> bool {
+        self.most_recent_by_ip.get(ip).map_or(false, |peer| {
```
arya2:
```suggestion
        self.most_recent_by_ip.get(ip).map_or(true, |peer| {
```

zebra-network/src/address_book.rs
arya2: Documented in https://github.com/ZcashFoundation/zebra/pull/7029/commits/d57f2b1c54883ae56ecec8f7fafba21a43663bc9.

teor2345:
I have about 5 things on my TODO list before I get to this review, so I'm not sure if it's going to make it into this release. You could ask Pili to give the review to someone else, or re-prioritise this ticket?

zebra-network/src/address_book.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7029/commits/489e63d14a9c9db1796c275c60234f832b7f497a.

zebra-network/src/address_book.rs
arya2: Documented in https://github.com/ZcashFoundation/zebra/pull/7029/commits/0c769b657fbe1f66dd20b7ed874ce4b529395395.

zebra-network/src/address_book.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7029/commits/0c769b657fbe1f66dd20b7ed874ce4b529395395.

zebra-network/src/address_book.rs
arya2: Updated to check last_attempt, last_response, and last_failure times in https://github.com/ZcashFoundation/zebra/pull/7029/commits/489e63d14a9c9db1796c275c60234f832b7f497a.

zebra-network/src/address_book/tests/vectors.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/pull/7029/commits/77833d9ab6eac162f5e0e39b68d35187ab33cc63.

zebra-network/src/address_book.rs
arya2: Documented in https://github.com/ZcashFoundation/zebra/pull/7029/commits/0c769b657fbe1f66dd20b7ed874ce4b529395395.

zebra-network/src/address_book.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7029/commits/77833d9ab6eac162f5e0e39b68d35187ab33cc63.

zebra-network/src/address_book/tests/vectors.rs
arya2: It looks like the address book should be tracking inbound connections as well: https://github.com/ZcashFoundation/zebra/blob/issue6681/zebra-network/src/peer/handshake.rs#L933

zebra-network/src/address_book.rs
teor2345: I was thinking about this a bit more yesterday. Some users will need Zebra to accept multiple connections from the same IP, either for testing, or for internal caching setups.

So could we disable the checks in this PR when `max_connections_per_ip > 1`?
(And document the security implications on that config.)

That way, the default config is secure, but users who need it can get Zebra to make multiple outbound connections.

zebra-network/src/address_book.rs
teor2345: Why does last_failure matter?

zebra-network/src/address_book/tests/vectors.rs
teor2345: I'm having trouble working out what the tests do, and what the closure argument means. Would you mind documenting what each test case covers?

zebra-network/src/address_book/tests/vectors.rs
teor2345: This method returns `None` for inbound connections, so the address book never finds out about them:
https://github.com/ZcashFoundation/zebra/blob/b6ba3f70319f8283feb8e7033b25f453b90ad843/zebra-network/src/peer/handshake.rs#L245-L254
https://github.com/ZcashFoundation/zebra/blob/b6ba3f70319f8283feb8e7033b25f453b90ad843/zebra-network/src/peer/handshake.rs#L263-L270

zebra-network/src/address_book.rs
arya2: So that `reconnection_peers()` skips addresses with IPs where there's a recently failed connection as it would if the connection had recently failed (in `is_ready_for_connection_attempt()`).

> Why does last_failure matter?

I don't think it actually does matter, it's just what I would expect it to be doing if I were debugging later.

arya2: Should it be updated to make `last_response` an `Instant`, or a tuple with both?

zebra-network/src/address_book.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/pull/7029/commits/0c769b657fbe1f66dd20b7ed874ce4b529395395.

zebra-network/src/address_book/tests/vectors.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7029/commits/0c769b657fbe1f66dd20b7ed874ce4b529395395.

zebra-network/src/address_book/tests/vectors.rs
arya2: Good point, added comments in https://github.com/ZcashFoundation/zebra/pull/7029/commits/9897b37cc98faa8ff8c0241515d4f0150a6f59fe.

zebra-network/src/address_book.rs
arya2: Great idea, thank you, applied in https://github.com/ZcashFoundation/zebra/pull/7029/commits/6d17a9cbd0e37f2a5196d046e2c94dfc0eacd043.

zebra-network/src/address_book.rs
arya2: > it's based on address and port, not just IP

Should it check for recent failures by IP as well?

zebra-network/src/address_book.rs
```diff
@@ -141,6 +155,8 @@ impl AddressBook {
         // and it gets replaced by `update_metrics` anyway.
         let (address_metrics_tx, _address_metrics_rx) = watch::channel(AddressMetrics::default());
 
+        // Avoid initiating outbound handshakes when max_connections_per_ip is 1.
+        let should_limit_outbound_conns_per_ip = max_connections_per_ip == 1;
```
teor2345, line 159: Currently we assert that `initial_peer_set_size` is greater than zero, should we do the same with `max_connections_per_ip`?

Or should we remove the assertions on both and just warn?

```suggestion
        let should_limit_outbound_conns_per_ip = max_connections_per_ip <= 1;
```

zebra-network/src/address_book.rs
teor2345: I think that checking the failure time enables an attack which pollutes our address book with fake peers on the same IP as a real peer. If the real peer is on a non-default port, the fake peers will all be tried before it, delaying a connection to the real peer by 2 minutes per fake peer. (Which is effectively never, because each IP address has 65000 TCP ports, so the limit is the 4000 address book addresses.)

So I think we should treat recently successful connections differently to connections that were never successful and immediately failed.

Specifically, an outbound connection to the same IP should not be attempted when:
* the peer recently had a successful connection and responded (~ 2 minute timeout)

But an outbound connection to the same IP should be attempted even if:
* we recently attempted a handshake to that peer
* we recently failed a connection to that peer without completing the handshake (`last_response` was not updated)

If we want to do anything about avoiding attempting multiple connections to the same IP at the same time, let's do that in a separate ticket with a lower priority?

teor2345: > Should it be updated to make `last_response` an `Instant`, or a tuple with both?

I think that's beyond the scope of a small security fix?

(We need to have a clock time because responded times are sent as part of the Zcash network protocol.)

zebra-network/src/address_book.rs
```diff
@@ -72,6 +73,14 @@ pub struct AddressBook {
     /// [`OrderedMap`] sorts in descending order.
     by_addr: OrderedMap<PeerSocketAddr, MetaAddr, Reverse<MetaAddr>>,
 
+    /// The address with a last_connection_state of [`PeerAddrState::Responded`] and
+    /// the most recent `last_response` time by IP.
+    ///
+    /// This is used to avoid initiating outbound connections past [`Config::max_connections_per_ip`](crate::config::Config), and
+    /// currently only supports a `max_connections_per_ip` of 1, and must be `None` when used with a greater `max_connections_per_ip`.
+    // TODO: Replace with `by_ip` to support configured `max_connections_per_ip`
```
teor2345: Optional:

This is a potential design for another ticket, which would keep the computational complexity to `O(1)`, except for removal which would `O(addresses for that IP with the same last responded time)`. (Which is usually 1.)

```suggestion
    // TODO: Replace with `by_ip: HashMap<IpAddr, BTreeMap<DateTime32, MetaAddr>>` to support configured `max_connections_per_ip` greater than 1
```

zebra-network/src/address_book.rs
```diff
@@ -314,6 +340,49 @@ impl AddressBook {
...
+    /// Returns true if `updated` needs to be applied to the recent outbound peer connection IP cache.
+    ///
+    /// Checks if there are no existing entries in the address book with this IP,
+    /// or if `updated` has a more recent update requiring the outbound connector to wait
+    /// longer before initiating handshakes with peers at this IP.
+    ///
+    /// This code only needs to check a single cache entry, rather than the entire address book,
+    /// because other code maintains these invariants:
+    /// - `last_attempt`, `last_failure`, and `last_response` times for an entry can only increase.
+    /// - these are the only fields checked by `was_recently_updated()`
```
teor2345: If we just look at last response, this function (and its docs) needs to be updated.

```suggestion
    /// - `last_response` time for an entry can only increase
    /// - this is the only field checked by `has_connection_recently_responded()`
```

zebra-network/src/address_book.rs
```diff
@@ -463,6 +556,23 @@ impl AddressBook {
         self.by_addr.descending_values().cloned()
     }
 
+    /// Is this IP ready for a new outbound connection attempt?
+    /// Checks if the most recently updated outbound address with this IP `was_recently_updated()`.
```
teor2345: If we just look at last response, this function (and its docs) needs to be updated.



```suggestion
    /// Checks if the most recently updated outbound address with this IP `has_connection_recently_responded()`.
```

zebra-network/src/constants.rs
```diff
@@ -71,6 +71,12 @@ pub const OUTBOUND_PEER_LIMIT_MULTIPLIER: usize = 3;
 /// before it drops any additional peer connections with that IP.
 ///
 /// This will be used as Config.max_connections_per_ip if no value is provided.
+///
+/// Note: Zebra will currently avoid initiating outbound connections where it
+///       has already recently sent or received a message to or from an address
+///       with that IP. Zebra will not initiate more than 1 outbound connection
+///       to an IP based on this configuration, but it will accept more inbound
+///       connections to an IP.
```
teor2345: If we just look at last response, these docs need to be updated.

```suggestion
/// Note: Zebra will currently avoid initiating outbound connections where it
///       has recently had a successful handshake with any address
///       on that IP. Zebra will not initiate more than 1 outbound connection
///       to an IP based on the default configuration, but it will accept more inbound
///       connections to an IP.
```

zebra-network/src/config.rs
```diff
@@ -158,6 +158,12 @@ pub struct Config {
     /// before it drops any additional peer connections with that IP.
     ///
     /// The default and minimum value are 1.
+    ///
+    /// Note: Zebra currently requires this to be set to 1 in order to avoid initiating
+    ///       additional outbound handshakes. It will otherwise still initiate outbound handshakes
+    ///       beyond this limit. Zebra will also currently still accept more inbound connections
+    ///       per IP. In cases where additional handshakes are permitted beyond this limit,
+    ///       connections will be dropped when attempting to add them to the peer set.
```
teor2345: For security-sensitive configs, it's important to warn the user in the first sentence.

Can I suggest a format for user-focused documentation?
- Short summary of the most important user impact
- Summary of how to avoid that impact
- Technical details

```suggestion
    /// # Security
    ///
    /// Increasing this config above 1 reduces Zebra's network security.
    ///
    /// If this config is greater than 1, Zebra can initiate multiple outbound handshakes to the same
    /// IP address.
    ///
    /// This config does not currently limit the number of inbound connections that Zebra will accept
    /// from the same IP address.
    ///
    /// If Zebra makes multiple inbound or outbound connections to the same IP, they will be dropped
    /// after the handshake, but before adding them to the peer set. The total numbers of inbound and
    /// outbound connections are also limited to a multiple of `peerset_initial_target_size`.
```

teor2345 (changes requested):
It seems like the scope of this PR has grown a bit. I'm partly responsible for that, I think my suggestion about pending handshakes was out of scope. (And it's difficult to do securely.)

So I'd like to make sure all the code we add in this PR has a strong security motivation. That way, it's less likely to have other negative security impacts.

zebra-network/src/address_book.rs
teor2345: For security reasons, I think failures should be ignored entirely in this security fix:
https://github.com/ZcashFoundation/zebra/pull/7029#discussion_r1252379827

zebra-network/src/address_book.rs
arya2: I think we should remove the assertions on both, use the default values, and log a warning.

Update: added warnings in https://github.com/ZcashFoundation/zebra/pull/7029/commits/3ac16d57cafd380f5678f0a1efc373bdfccb7b1a.

zebra-network/src/address_book.rs
arya2: Applied something similar in https://github.com/ZcashFoundation/zebra/pull/7029/commits/788bce3b9ea89d3d5b3d535d09720b9b04c169a0.

zebra-network/src/address_book.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7029/commits/788bce3b9ea89d3d5b3d535d09720b9b04c169a0.

zebra-network/src/address_book.rs
arya2: > But an outbound connection to the same IP should be attempted even if:
> 
> - we recently attempted a handshake to that peer
> - we recently failed a connection to that peer without completing the handshake (last_response was not updated)

Fixed in https://github.com/ZcashFoundation/zebra/pull/7029/commits/788bce3b9ea89d3d5b3d535d09720b9b04c169a0 and https://github.com/ZcashFoundation/zebra/pull/7029/commits/f8338f1871f829851582fd3d7015effa7f0e04be.

zebra-network/src/address_book.rs
```diff
@@ -435,6 +518,12 @@ impl AddressBook {
         );
 
         if let Some(entry) = self.by_addr.remove(&removed_addr) {
+            // Check if this surplus peer's addr matches that in `most_recent_by_ip`
+            // for this the surplus peer's ip to remove it there as well.
+            if self.should_remove_most_recent_by_ip(entry.addr) {
+                self.most_recent_by_ip.as_mut()?.remove(&entry.addr.ip());
```
teor2345: I think this will return `None` when `most_recent_by_ip` is disabled, even if a peer was actually removed. But this method should return the removed peer.

```suggestion
                if let Some(most_recent_by_ip) = self.most_recent_by_ip.as_mut() {
                    most_recent_by_ip.remove(&entry.addr.ip());
                }
```

zebra-network/src/address_book.rs
```diff
@@ -463,6 +552,22 @@ impl AddressBook {
...
+    fn is_ready_for_connection_attempt_with_ip(
+        &self,
+        ip: &IpAddr,
+        chrono_now: chrono::DateTime<Utc>,
+    ) -> bool {
+        self.most_recent_by_ip
+            .as_ref()
+            .and_then(|most_recent_by_ip| most_recent_by_ip.get(ip))
+            .filter(|peer| peer.has_connection_recently_responded(chrono_now))
+            .is_none()
```
teor2345: I think the logic here might be inverted?

But I'm not sure, because the tests look like they are correct, and they are passing. Maybe we're missing a test for this specific case.

Here's a minimal fix:

```rust
            .filter(|peer| !peer.has_connection_recently_responded(chrono_now))
```

Here's a refactor where that is easier to see:
```suggestion
        let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
            // if we're not checking IPs, any connection is allowed
            return true;
        }
        
        let Some(same_ip_peer) = most_recent_by_ip.get(ip) else {
            // If there's no entry for this IP, any connection is allowed
            return true;
        }
        
        !same_ip_peer.has_connection_recently_responded(chrono_now)
```

teor2345 (dismissed):
Looks great!

I just want to double-check the logic in two places where we might be returning an incorrect value.

zebra-network/src/address_book.rs
arya2: > I think the logic here might be inverted?
> 
> But I'm not sure

The logic was correct, but your version is much clearer.

zebra-network/src/address_book.rs
arya2: > I think this will return `None`

It wasn't clear but it's guaranteed to be `Some` if `should_remove_most_recent_by_ip()` is true.

zebra-network/src/address_book.rs
```diff
@@ -463,6 +554,28 @@ impl AddressBook {
...
+    /// Note: last_response times may remain live for a long time if the local clock is changed to an earlier time.
+    fn is_ready_for_connection_attempt_with_ip(
+        &self,
+        ip: &IpAddr,
+        chrono_now: chrono::DateTime<Utc>,
+    ) -> bool {
+        let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
+            // if we're not checking IPs, any connection is allowed
+            return true;
+        }
```
teor2345: oops

```suggestion
        };
```

zebra-network/src/address_book.rs
```diff
@@ -463,6 +554,28 @@ impl AddressBook {
...
+        
+        let Some(same_ip_peer) = most_recent_by_ip.get(ip) else {
+            // If there's no entry for this IP, any connection is allowed
+            return true;
+        }
```
teor2345:
```suggestion
        };
```

teor2345 (dismissed):
Sorry about the Rust syntax, I didn't check it before suggesting

zebra-network/src/address_book.rs
```diff
@@ -463,6 +554,28 @@ impl AddressBook {
...
+    fn is_ready_for_connection_attempt_with_ip(
+        &self,
+        ip: &IpAddr,
+        chrono_now: chrono::DateTime<Utc>,
+    ) -> bool {
+        let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
+            // if we're not checking IPs, any connection is allowed
+            return true;
+        };
+
```
teor2345:
```suggestion

```

zebra-network/src/address_book.rs
```diff
@@ -463,6 +554,28 @@ impl AddressBook {
...
+        let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
+            // if we're not checking IPs, any connection is allowed
+            return true;
+        };
+        
+        let Some(same_ip_peer) = most_recent_by_ip.get(ip) else {
+            // If there's no entry for this IP, any connection is allowed
+            return true;
+        };
+
```
teor2345:
```suggestion

```

teor2345 (approved):

arya2:
@Mergifyio refresh

PR #7031, opened 2023-06-21
change(state): Prepare for in-place database format upgrades, but don't make any format changes yet

### LINKED ISSUE #6955: Create an empty database format update task and format update tests
## Motivation

We want to do in-place format upgrades to the Zebra on-disk database.

### Design

We can make state database format changes by:
- [x] add a "minor" version to the state
- [x] launch a format update task when we open the database
- [x] regularly check if the format upgrade task has finished, and check for panics after it exits
- [x] mark the update as finished when the task finishes
- [x] restart the task if the previous zebrad instance didn't finish updating the format (or if it downgraded the format)

After initial PR review:
- [x] increment the database format patch version
- [x] log the database version as part of the startup/panic logs
- [x] add full disk and running state versions to cached state images in CI

### Tests

- [x] newly created database is not updated / version is not updated on restart
- [x] standard update / state version is incremented
- [x] repeated update only runs once / current version is not updated on restart
- [x] downgrade (best effort compatibility/testing)
- [ ] partial update and restart (TODO in another PR)
- [x] CI workflow changes

## Documentation Follow-Up

In ticket #7215.

## Refactor Follow-Ups

- refactor the `block_write_task` and `format_change_handle` into a generic `std::thread` wrapper.
- move config code to the new format upgrade module

## Debugging Follow-Ups

- add a small log file for debugging database format changes in the state directory, and limit its length

teor2345:
## Motivation

We need a way to upgrade the on-disk database format, without re-syncing the whole chain.

Close #6955

This is concurrent code, we need to be careful to shut down correctly, and regularly check for panics in the inner thread.

A lot of this code is modified from the block write task. We could refactor that task to use a generic `ThreadHandle` type in a future PR.

## Solution

Implementation:
- [Move format upgrades to their own module and enum](https://github.com/ZcashFoundation/zebra/pull/7031/commits/451f97cc1e7ac2a26cdeff72f7d8f799b3f21d11)
- [Launch a format change thread if needed](https://github.com/ZcashFoundation/zebra/pull/7031/commits/79f0fbbe86215e0e7df47dc2c5dd27b98c080d01)
- Shut down the format change thread when Zebra is shutting down
- [Regularly check for panics in the state upgrade task](https://github.com/ZcashFoundation/zebra/pull/7031/commits/c82aa581331f7031150163f6f1617ea6080ddc51)
- [Increment database format to 25.0.2](https://github.com/ZcashFoundation/zebra/pull/7031/commits/2dbc6cee00536d5f40d10f95d35a51f67c429d65)

Logging:
- [Log the running and initial disk database format versions on startup](https://github.com/ZcashFoundation/zebra/pull/7031/commits/d7572cb304fa116ba958506cc961942b80101888)

### Testing

`zebrad` tests:
- Test that a newly created state is the current version, and re-opening it doesn't upgrade or downgrade the version
- Test that a fake older state gets upgraded to the current version
- Test that a fake newer state gets downgraded to the current version
- Add extra test support functions and rename some existing functions

CI:
- [Add initial disk and running state versions to cached state images in CI](https://github.com/ZcashFoundation/zebra/pull/7031/commits/de2453596ad7cb25174488727c9be1476e6e9a6f)
    - This also tests `zebrad` logging the state versions
- [Add a force_save_to_disk argument to the CI workflow](https://github.com/ZcashFoundation/zebra/pull/7031/commits/5a50516773dd3738b727171b6f413bf77202329d)

Related changes:
- [Add some TODOs and remove a redundant timer](https://github.com/ZcashFoundation/zebra/pull/7031/commits/6633f2f68d1890189c790f4823b1644b21a24c86)

## Review

This is a large PR, so I'm happy to do a video review, or split the test harness refactor and the CI changes into separate PRs.

### Reviewer Checklist

## Follow Up Work

See #6955:
- documentation for format upgrades
- optional format upgrade debugging log file
- optional refactors

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -437,7 +437,7 @@ jobs:
       needs_zebra_state: true
       # update the disk on every PR, to increase CI speed
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -510,7 +510,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
     secrets: inherit
@@ -554,7 +554,7 @@ jobs:
       # update the disk on every PR, to increase CI speed
       # we don't have a test-update-sync-testnet job, so we need to update the disk here
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -587,7 +587,7 @@ jobs:
       needs_zebra_state: true
       needs_lwd_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
       disk_prefix: lwd-cache
       disk_suffix: tip
       root_state_path: '/var/cache'
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -35,6 +35,11 @@ on:
         default: false
         description: 'Just run a lightwalletd full sync and update tip disks'
         required: true
+      force_save_to_disk:
+        required: false
+        type: boolean
+        default: false
+        description: 'Force tests to always create a cached state disk, if they already create disks'
       no_cache:
         description: 'Disable the Docker cache for this build'
         required: false
@@ -330,6 +335,7 @@ jobs:
       test_variables: '-e TEST_DISK_REBUILD=1 -e ZEBRA_FORCE_USE_COLOR=1'
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_suffix: checkpoint
       height_grep_text: 'flushing database to disk .*height.*=.*Height.*\('
     secrets: inherit
@@ -391,6 +397,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
     secrets: inherit
@@ -430,6 +437,7 @@ jobs:
       needs_zebra_state: true
       # update the disk on every PR, to increase CI speed
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -502,6 +510,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
     secrets: inherit
@@ -545,6 +554,7 @@ jobs:
       # update the disk on every PR, to increase CI speed
       # we don't have a test-update-sync-testnet job, so we need to update the disk here
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -577,6 +587,7 @@ jobs:
       needs_zebra_state: true
       needs_lwd_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
       disk_prefix: lwd-cache
       disk_suffix: tip
       root_state_path: '/var/cache'
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -335,7 +335,7 @@ jobs:
       test_variables: '-e TEST_DISK_REBUILD=1 -e ZEBRA_FORCE_USE_COLOR=1'
       needs_zebra_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_suffix: checkpoint
       height_grep_text: 'flushing database to disk .*height.*=.*Height.*\('
     secrets: inherit
@@ -397,7 +397,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
     secrets: inherit
@@ -437,7 +437,7 @@ jobs:
       needs_zebra_state: true
       # update the disk on every PR, to increase CI speed
       saves_to_disk: true
-      force_save_to_disk: ${{ inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -510,7 +510,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_suffix: tip
       height_grep_text: 'current_height.*=.*Height.*\('
     secrets: inherit
@@ -554,7 +554,7 @@ jobs:
       # update the disk on every PR, to increase CI speed
       # we don't have a test-update-sync-testnet job, so we need to update the disk here
       saves_to_disk: true
-      force_save_to_disk: ${{ inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_suffix: tip
       root_state_path: '/var/cache'
       zebra_state_dir: 'zebrad-cache'
@@ -587,7 +587,7 @@ jobs:
       needs_zebra_state: true
       needs_lwd_state: false
       saves_to_disk: true
-      force_save_to_disk: ${{ inputs.force_save_to_disk }}
+      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       disk_prefix: lwd-cache
       disk_suffix: tip
       root_state_path: '/var/cache'
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -1719,6 +1719,96 @@ jobs:
           echo "UPDATE_SUFFIX=$UPDATE_SUFFIX" >> "$GITHUB_ENV"
           echo "TIME_SUFFIX=$TIME_SUFFIX" >> "$GITHUB_ENV"
 
+      # Get the full initial and running database versions from the test logs.
+      # These versions are used as part of the disk description and labels.
+      #
+      # If these versions are missing from the logs, the job fails.
+      #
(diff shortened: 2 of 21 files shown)


teor2345:
I'm running a manual test of the workflow changes here:
~~https://github.com/ZcashFoundation/zebra/actions/runs/5470498900~~
~~https://github.com/ZcashFoundation/zebra/actions/runs/5479531847~~
~~https://github.com/ZcashFoundation/zebra/actions/runs/5482043126~~
~~https://github.com/ZcashFoundation/zebra/actions/runs/5482393626~~
~~https://github.com/ZcashFoundation/zebra/actions/runs/5501815291~~

https://github.com/ZcashFoundation/zebra/actions/runs/5502224802

It needs the new "force_save_to_disk" argument in this PR, to make sure the disk job gets tested.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -0,0 +1,511 @@
...
+            //
+            // Keep upgrading until the initial database has been upgraded,
+            // or this task is cancelled by a shutdown.
+            while upgrade_height <= initial_tip_height
+                && matches!(cancel_receiver.try_recv(), Err(mpsc::TryRecvError::Empty))
+            {
+                // TODO: Do one format upgrade step here
+
+                upgrade_height = (upgrade_height + 1).expect("task exits before maximum height");
+            }
```
teor2345: There's a lot of task code, versioning code, and test code in this PR. But the main part of it is here - this is where we can add the code to do the format upgrades.

We can call whatever database methods we like here, and they can take as long as is needed, because this code runs in a separate thread.

~~I just noticed I need to pass the database into this code, then we can call it! But that's an easy change, the database can be cloned here:
https://github.com/ZcashFoundation/zebra/pull/7031/files#diff-17dd8a7e992ef6adf7d3c5e78546718069747564f3b60ab3662747a0ceaddc68R84~~

teor2345: The database is available in this code now.

teor2345:
> I'm running a manual test of the workflow changes here: 
> [ZcashFoundation/zebra/actions/runs/5502224802](https://github.com/ZcashFoundation/zebra/actions/runs/5502224802)

This worked and created a disk with the correct version description and labels:


Screenshot 2023-07-10 at 13 01 34

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -0,0 +1,533 @@
...
+                && matches!(cancel_receiver.try_recv(), Err(mpsc::TryRecvError::Empty))
+            {
+                // TODO: Do one format upgrade step here
+                //
+                // This fake step just shows how to access the database.
+                let _replace_me_ = upgrade_db.tip();
+
+                upgrade_height = (upgrade_height + 1).expect("task exits before maximum height");
+            }
+        }
```
teor2345, line 286: There's a lot of task code, versioning code, and test code in this PR. But the main part of it is here - this is where we can add the code to do the format upgrades.

We can call whatever database methods we like here, and they can take as long as is needed, because this code runs in a separate thread.

.github/workflows/continous-integration-docker.yml
```diff
@@ -330,6 +335,7 @@ jobs:
       test_variables: '-e TEST_DISK_REBUILD=1 -e ZEBRA_FORCE_USE_COLOR=1'
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -391,6 +397,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ github.event.inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -430,6 +437,7 @@ jobs:
       needs_zebra_state: true
       # update the disk on every PR, to increase CI speed
       saves_to_disk: true
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -502,6 +510,7 @@ jobs:
       is_long_test: true
       needs_zebra_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -545,6 +554,7 @@ jobs:
       # update the disk on every PR, to increase CI speed
       # we don't have a test-update-sync-testnet job, so we need to update the disk here
       saves_to_disk: true
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

.github/workflows/continous-integration-docker.yml
```diff
@@ -577,6 +587,7 @@ jobs:
       needs_zebra_state: true
       needs_lwd_state: false
       saves_to_disk: true
+      force_save_to_disk: ${{ inputs.force_save_to_disk }}
```
```suggestion
      force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
```

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: @upbqdn here is the place to start working on the finalized state change. Let me know if it doesn't work, or if anything is confusing.

I haven't written the instructions yet, I'm waiting to see if we need to change the process as part of the first two format changes.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -108,19 +175,42 @@ impl DbFormatChange {
...
+
+        // Example format change.
+        //
+        // TODO: link to format upgrade instructions doc here
+        let mut upgrade_height = Height(0);
+
+        // Keep upgrading until the initial database has been upgraded,
+        // or this task is cancelled by a shutdown.
+        while upgrade_height <= initial_tip_height
+            && matches!(cancel_receiver.try_recv(), Err(mpsc::TryRecvError::Empty))
```
arya2: Optional: could we re-use `is_shutting_down()` from zebra-chain if `cancel_receiver` is only used for shutdowns?

zebra-state/src/constants.rs
```diff
@@ -52,7 +52,7 @@ pub const DATABASE_FORMAT_MINOR_VERSION: u64 = 0;
 
 /// The database format patch version, incremented each time the on-disk database format has a
 /// significant format compatibility fix.
-pub const DATABASE_FORMAT_PATCH_VERSION: u64 = 1;
+pub const DATABASE_FORMAT_PATCH_VERSION: u64 = 2;
```
arya2: ~Could we pass in `crate_version!()`/`CARGO_PKG_VERSION` from zebrad instead?~

Update: Nevermind.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -0,0 +1,533 @@
...
+        let running_version = database_format_version_in_code();
+
+        assert_eq!(
+            disk_version, None,
+            "can't overwrite the format version in an existing database:\n\
+             disk: {disk_version:?}\n\
+             running: {running_version}"
+        );
+
+        write_database_format_version_to_disk(&running_version, config, network)
```
arya2, line 341: Do we want to add the format version to RocksDB and use transactions in case it's interrupted during a migration? 

We could update ~100 blocks at a time and update the disk version to something like `v1(1131..2210), v2(0..1131)` leaving the height range in parens during a migration when there are multiple data formats.

arya2 (dismissed):
Looks good to me!

Thank you for the well-organized tests and related cleanups.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: I'd always thought of the shutdown flag as an advisory flag that wasn't required for code correctness. I think that's how it's currently used, but it's not documented that way. So we could change it if we want.

But in this PR, the task is shut down, then the state shuts itself down. Shutting down the state in a specific order is required to avoid memory safety bugs with C++ threads and destructors. I want to avoid having to diagnose and fix those kinds of bugs again. (It's very tricky because we get almost no log output, and tracing is shutting down at the same time.)

The timing of shutdown is also less reliable, and the task wouldn't be stopped if the state is dropped. So the database could remain open and the upgrade task would keep running, even though the state was forcing itself to shut down.  

So the global shutdown flag might work if the state is being used by the `zebrad start` application, as long as the only time we want to stop the task is at shutdown. But it would make it harder to open multiple states in the same app (like `zebrad copy-state`), temporarily (`zebrad tip-height`), or re-use zebra-state in another app entirely.

Here's some background to why making shutdown local is a more reliable design:
https://blog.yoshuawuyts.com/tree-structured-concurrency/#what-s-the-worst-that-can-happen

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: That's a good question! I'm not sure if we have enough information to answer it yet. (That's why I haven't written up the state upgrade documentation yet.)

Here are my current thoughts:

We could handle partial migrations by expecting upgrade authors to make their upgrade code work for full or partial migrations. Then that code could also handle downgrades to an earlier state version as well. (A downgrade adds new blocks after the tip in an earlier state format, a partial migration leaves some higher or lower blocks in an earlier state format, depending on the height order it does the upgrade in.)

If upgrades need to track how many blocks have been upgraded in order to be correct, we can add that code later. But earlier versions won't track that info, so they'll have to assume any un-marked blocks are not upgraded. (That's a disadvantage of not building it into the upgrade framework from the start.)

I wanted to have the version outside the database for 2 reasons:
- we already have the major version in the directory path outside the database
- if we need to do an upgrade or fix an issue before opening the database, then we might need to know the version to open it correctly (for example, if we need to change the open options or column families)

Transactions can lock the entire database, blocking other reads and writes. So I'd prefer to keep them as small as possible. We could do each height (or other consistent update) in its own transaction, if needed. But none of our planned upgrades need transactions yet.

Here's how the 3 upgrades we have planned for RPCs could work:
- delete redundant tree states: correct if applied multiple times (don't panic if the deleted item isn't there), correct but lowers performance if some duplicate tree states aren't deleted, no transactions needed (single column family), can be done in any order
- ~~add tree sizes: correct if applied multiple times (skip if the size has already been added), no transactions needed (single column family), might need to be added in reverse height order for lightwalletd to do fast sync correctly~~
- create subtree index: correct if applied multiple times (just writes the same data), no transactions needed (single column family), needs to be added in reverse height order for lightwalletd to do fast sync correctly

Using reverse height order might be the way to go for all upgrades, because the latest blocks are usually the ones being accessed most frequently.

teor2345: Edit: "add tree sizes" isn't a database format change, it's a new API that uses the existing tree data

teor2345: > We could update ~100 blocks at a time and update the disk version to something like `v1(1131..2210), v2(0..1131)` leaving the height range in parens during a migration when there are multiple data formats.

I was thinking of logging progress in a separate text file or the Zebra logs, but I cut it from scope because the rest of the PR was too big!

arya2: > I'm not sure if we have enough information to answer it yet.
> ..
> If upgrades need to track how many blocks have been upgraded in order to be correct, we can add that code later.

Okay, I was thinking of a disk format change for note commitment trees (not planned), but on second thought it's better to let upgrade authors make their upgrade code work for full or partial migrations.

> Here's how the 3 upgrades we have planned for RPCs could work

It seems easy to figure out where it left off by checking the data in these cases, that may be ideal.

teor2345: > > Here's how the 3 upgrades we have planned for RPCs could work
> 
> It seems easy to figure out where it left off by checking the data in these cases, that may be ideal.

I think the subtree indexes will be `index -> (start_height, end_height)`, so we could look up the earliest index key, find the earliest height, and go from there. The others have height keys.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -1719,6 +1724,96 @@ jobs:
...
+          RUNNING_DB_VERSION="v${RUNNING_DB_VERSION//./-}"
+          echo "Found running database version in logs: $RUNNING_DB_VERSION"
+          echo "RUNNING_DB_VERSION=$RUNNING_DB_VERSION" >> "$GITHUB_ENV"
+
+          if [[ "$INITIAL_DISK_DB_VERSION" = "$RUNNING_DB_VERSION" ]]; then
+              DB_VERSION_SUMMARY="$RUNNING_DB_VERSION"
+          elif [[ "$INITIAL_DISK_DB_VERSION" = "new" ]]; then
+              DB_VERSION_SUMMARY="$RUNNING_DB_VERSION in new database"
+          else
+              DB_VERSION_SUMMARY="$INITIAL_DISK_DB_VERSION upgrading to $RUNNING_DB_VERSION"
```
```suggestion
              DB_VERSION_SUMMARY="$INITIAL_DISK_DB_VERSION changing to $RUNNING_DB_VERSION"
```

teor2345:
@mergifyio update

arya2 (approved):

PR #7032, opened 2023-06-21
fix(concurrency): Use Arc::into_inner() to avoid potential concurrency issues, needs Rust 1.70

teor2345:
## Motivation

Some of our code uses an older `Arc` method, which isn't guaranteed to return a value. This can cause hangs or panics in our code if two threads call it at the same time.

### Specifications

The replacement method guarantees that exactly one thread will get the inner value:
https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_inner

This is a concurrency bug fix, it requires Rust 1.70.

## Solution

Replace `Arc::try_unwrap()` with `Arc::into_inner()`
Update Zebra's Rust version requirement to 1.70

Replacements aren't needed if the code falls back to another method that is guaranteed to be correct.

## Review

This is a routine concurrency fix.

### Reviewer Checklist


arya2 (approved):
Nice catch!

PR #7025, opened 2023-06-20
change(state): Refactor the structure of verified blocks

upbqdn:
## Motivation & Solution

In preparation for issue #6912, this PR:

- Wraps `SemanticallyVerifiedBlock` as `CheckpointVerifiedBlock` instead of having a standalone `CheckpointVerifiedBlock`. This was straightforward due to PR #6971.
- Uses `SemanticallyVerifiedBlock` in `ContextuallyVerifiedBlockWithTrees`.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests? 

## Follow-Up Work

- #7035

PR #7035, opened 2023-06-21
change(state): Refactor the structure of finalizable blocks

### LINKED ISSUE #6912: Change `ContextuallyVerifiedBlockWithTrees` into an enum to correctly represent verification statuses

upbqdn:
## Motivation

Close #6912.

## Solution

- Create `FinalizableBlock` and use it instead of `ContextuallyVerifiedBlockWithTrees` in `commit_finalized_direct()`.
- Pass `ContextuallyVerifiedBlockWithTrees` instead of passing separate `finalized`, `history_tree`, and `note_commitment_trees` when storing blocks in the finalized state. This also simplifies the internal API when storing blocks since the parameter list is shorter.

This PR is based on PR #7025.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -273,7 +273,7 @@ pub enum FinalizableBlock {
 }
 
 impl FinalizableBlock {
-    /// Create a new `FinalizedBlock` given a `ContextuallyVerifiedBlock`.
+    /// Create a new [`FinalizableBlock`] given a [`ContextuallyVerifiedBlock`].
     pub fn new(contextually_verified: ContextuallyVerifiedBlock, treestate: Treestate) -> Self {
         Self::Contextual {
             contextually_verified,
@@ -282,7 +282,7 @@ impl FinalizableBlock {
     }
 
     #[cfg(test)]
-    /// Extract a `Block` from a `FinalizedBlock` variant.
+    /// Extract a [`Block`] from a [`FinalizableBlock`] variant.
     pub fn inner_block(&self) -> Arc<Block> {
         match self {
             FinalizableBlock::Checkpoint {
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -273,6 +273,7 @@ pub enum FinalizableBlock {
 }
 
 impl FinalizableBlock {
+    /// Create a new `FinalizedBlock` given a `ContextuallyVerifiedBlock`.
     pub fn new(contextually_verified: ContextuallyVerifiedBlock, treestate: Treestate) -> Self {
         Self::Contextual {
             contextually_verified,
@@ -281,6 +282,7 @@ impl FinalizableBlock {
     }
 
     #[cfg(test)]
+    /// Extract a `Block` from a `FinalizedBlock` variant.
     pub fn inner_block(&self) -> Arc<Block> {
         match self {
             FinalizableBlock::Checkpoint {
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -262,6 +262,10 @@ pub struct SemanticallyVerifiedBlockWithTrees {
     pub treestate: Treestate,
 }
 
+/// Contains a block ready to be committed.
+///
+/// Zebra's state service passes this `enum` over to the finalized state
+ /// when committing a block.
 pub enum FinalizableBlock {
     Checkpoint {
         checkpoint_verified: CheckpointVerifiedBlock,
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -257,41 +257,54 @@ impl Treestate {
 /// database and recompute a new one.
 pub struct ContextuallyVerifiedBlockWithTrees {
     /// A block ready to be committed.
-    pub block: SemanticallyVerifiedBlock,
+    pub verified: SemanticallyVerifiedBlock,
     /// The tresstate associated with the block.
-    pub treestate: Option<Treestate>,
+    pub treestate: Treestate,
 }
 
-impl ContextuallyVerifiedBlockWithTrees {
+pub enum FinalizableBlock {
+    Checkpoint {
+        checkpoint_verified: CheckpointVerifiedBlock,
+    },
+    Contextual {
+        contextually_verified: ContextuallyVerifiedBlock,
+        treestate: Treestate,
+    },
+}
+
+impl FinalizableBlock {
     pub fn new(contextually_verified: ContextuallyVerifiedBlock, treestate: Treestate) -> Self {
-        Self {
-            block: SemanticallyVerifiedBlock::from(contextually_verified),
-            treestate: Some(treestate),
+        Self::Contextual {
+            contextually_verified,
+            treestate,
+        }
+    }
+
+    #[cfg(test)]
+    pub fn inner_block(&self) -> Arc<Block> {
+        match self {
+            FinalizableBlock::Checkpoint {
+                checkpoint_verified,
+            } => checkpoint_verified.block.clone(),
+            FinalizableBlock::Contextual {
+                contextually_verified,
+                ..
+            } => contextually_verified.block.clone(),
         }
     }
 }
 
-impl From<Arc<Block>> for ContextuallyVerifiedBlockWithTrees {
-    fn from(block: Arc<Block>) -> Self {
-        Self::from(SemanticallyVerifiedBlock::from(block))
-    }
-}
-
-impl From<SemanticallyVerifiedBlock> for ContextuallyVerifiedBlockWithTrees {
-    fn from(semantically_verified: SemanticallyVerifiedBlock) -> Self {
-        Self {
-            block: semantically_verified,
-            treestate: None,
-        }
-    }
-}
-
-impl From<CheckpointVerifiedBlock> for ContextuallyVerifiedBlockWithTrees {
+impl From<CheckpointVerifiedBlock> for FinalizableBlock {
     fn from(checkpoint_verified: CheckpointVerifiedBlock) -> Self {
-        Self {
-            block: checkpoint_verified.0,
-            treestate: None,
+        Self::Checkpoint {
+            checkpoint_verified,
         }
+    }
+}
+
+impl From<Arc<Block>> for FinalizableBlock {
+    fn from(block: Arc<Block>) -> Self {
+        Self::from(CheckpointVerifiedBlock::from(block))
     }
 }
 
@@ -410,6 +423,12 @@ impl From<ContextuallyVerifiedBlock> for SemanticallyVerifiedBlock {
             new_outputs,
             transaction_hashes,
         }
+    }
+}
+
+impl From<CheckpointVerifiedBlock> for SemanticallyVerifiedBlock {
+    fn from(checkpoint_verified: CheckpointVerifiedBlock) -> Self {
+        checkpoint_verified.0
     }
 }
 
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -255,7 +255,7 @@ impl Treestate {
 /// when committing a block. The associated treestate is passed so that the
 /// finalized state does not have to retrieve the previous treestate from the
 /// database and recompute a new one.
-pub struct ContextuallyVerifiedBlockWithTrees {
+pub struct SemanticallyVerifiedBlockWithTrees {
     /// A block ready to be committed.
     pub verified: SemanticallyVerifiedBlock,
     /// The tresstate associated with the block.
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -265,7 +265,7 @@ pub struct SemanticallyVerifiedBlockWithTrees {
 /// Contains a block ready to be committed.
 ///
 /// Zebra's state service passes this `enum` over to the finalized state
- /// when committing a block.
+/// when committing a block.
 pub enum FinalizableBlock {
     Checkpoint {
         checkpoint_verified: CheckpointVerifiedBlock,
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -236,7 +236,7 @@ impl FinalizedState {
                 // treestate of the finalized tip from the database and update it for the block
                 // being committed, assuming the retrieved treestate is the parent block's
                 // treestate. Later on, this function proves this assumption by asserting that the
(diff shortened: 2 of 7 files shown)


upbqdn:
The initial base of this PR is #7025, so it would be suitable if we merge #7025 first, and then rebase this PR onto `main`. I, therefore, marked this PR as a draft, even though it's complete.

upbqdn:
Rebased onto `main`.

zebra-state/src/request.rs
```diff
@@ -257,41 +257,54 @@ impl Treestate {
 /// database and recompute a new one.
 pub struct ContextuallyVerifiedBlockWithTrees {
     /// A block ready to be committed.
-    pub block: SemanticallyVerifiedBlock,
+    pub verified: SemanticallyVerifiedBlock,
```
teor2345, line 260: Should this type or the outer type name be different? Currently they don't match, which was part of your original feedback on the rename PR.

```suggestion
    pub verified: ContextuallyVerifiedBlock,
```

zebra-state/src/request.rs
```diff
@@ -257,41 +257,54 @@ impl Treestate {
...
+        checkpoint_verified: CheckpointVerifiedBlock,
+    },
+    Contextual {
+        contextually_verified: ContextuallyVerifiedBlock,
+        treestate: Treestate,
+    },
+}
+
+impl FinalizableBlock {
     pub fn new(contextually_verified: ContextuallyVerifiedBlock, treestate: Treestate) -> Self {
```
teor2345, line 281: Can you add documentation to these methods? It can just be short, but it helps to explain what they do.

zebra-state/src/service/finalized_state.rs
```diff
@@ -225,53 +225,22 @@ impl FinalizedState {
...
-            ),
-            // If the treestate was not supplied, retrieve a previous treestate
-            // from the database, and update it for the block being committed.
-            None => {
+        let (height, hash, finalized) = match finalizable_block {
+            FinalizableBlock::Checkpoint {
+                checkpoint_verified,
+            } => {
+                // Checkpoint-verified blocks don't have an associated treestate, so we get the most
+                // recent one from the database and update it for the block being committed.
```
teor2345: This isn't valid unless finalizable_block is a child of the finalised tip block. So it would be more accurate to say "parent block's treestate".

There's an assert below that checks this, maybe we could say "we check that it's the parent block in an assert later in this function" or something.

(This was a problem in the existing docs, but we might as well fix it now.)

zebra-state/src/service/finalized_state.rs
```diff
@@ -303,41 +272,75 @@ impl FinalizedState {
...
             #[cfg(feature = "elasticsearch")]
             self.elasticsearch(&finalized_block);
 
             // TODO: move the stop height check to the syncer (#3442)
-            if self.is_at_stop_height(finalized_height) {
+            if self.is_at_stop_height(height) {
                 tracing::info!(
-                    height = ?finalized_height,
-                    hash = ?finalized_hash,
+                    height = ?height,
```
teor2345: Nitpick:
```suggestion
                    ?height,
```

zebra-state/src/service/finalized_state.rs
```diff
@@ -303,41 +272,75 @@ impl FinalizedState {
...
+                    hash = ?hash,
```
teor2345:
```suggestion
                    ?hash,
```

teor2345 (dismissed):
Looks great, thanks!

All my comments are optional, but it would be nice to fix the docs and the type name.

zebra-state/src/request.rs
upbqdn: Oh, thanks for catching this. You're right.

However, after looking into it again, I realized `ContextualyVerifiedBlock`s have two extra fields compared to `SemanticallyVerifiedBlock`s which aren't used anywhere in `ContextuallyVerifiedBlockWithTrees`. Why do we even call the struct `ContextuallyVerifiedBlockWithTrees`?

We construct `ContextuallyVerifiedBlockWithTrees` only in one place and either directly from a `ContextuallyVerifiedBlock` or `CheckpointVerifiedBlock`. I could simply construct a `ContextuallyVerifiedBlock` from a `CheckpointVerifiedBlock` with the two aforementioned fields being empty, but couldn't we rename `ContextuallyVerifiedBlockWithTrees` to `SemanticallyVerifiedBlockWithTrees` instead? It would generalize both the `CheckpointVerifiedBlock` and `ContextuallyVerifiedBlock`, while having all the fields necessary for committing the block.

teor2345: Because of the way we've defined different verification stages, a checkpoint verified block needs to have its header bound to the authorising data in its transactions. We also bind the header to the chain history of the previous block, but that's redundant verification. (We do it so light clients can use the history tree to quickly check we haven't left out any transactions in RPC responses.)

However, a block from the non-finalised state doesn't need either of these verification steps, because the non-finalised state has already contextually verified it, and generated its trees.

So I suggest we use names and types that reflect the verification status of the block, then adjust the data to match those names. If we have to add new names, types, or data, that's ok.

Are we constructing contextually verified blocks from checkpoint verified blocks after doing the required verification? If we're not, that misrepresents their verification status.

Edit:
Also, I don't think we should be downgrading verification status from contextual or checkpoint to semantic, unless it's a quick downgrade using into(), and then calling a common verification function. Most of the time we've already done that verification on those types though!

zebra-state/src/request.rs
oxarbitrage: Added in https://github.com/ZcashFoundation/zebra/pull/7035/commits/d61add359e14cf2251b3c76053a7653e09607314

zebra-state/src/service/finalized_state.rs
oxarbitrage: Fixed in https://github.com/ZcashFoundation/zebra/pull/7035/commits/d5e0e2b323973fe7c1a25f1fdab17616dc70623c

zebra-state/src/request.rs
oxarbitrage: It seems that with the current state of the renames the best thing to do here is to change the outer type from `ContextuallyVerifiedBlockWithTrees` to `SemanticallyVerifiedBlockWithTrees`.

This rename will work for the checkpointing because `SemanticallyVerifiedBlock` is enough to commit and for write a finalized block into the database because the block is already contextually verified : https://github.com/ZcashFoundation/zebra/blob/add-FinalizableBlock/zebra-state/src/service/write.rs#L293

I made this change at https://github.com/ZcashFoundation/zebra/pull/7035/commits/47a6ab80fa9da6440a3d83e0f675be8029ec07b5

zebra-state/src/request.rs
upbqdn: Refactored in [55def20](https://github.com/ZcashFoundation/zebra/pull/7035/commits/55def2032ae6b070251dff7ab8cce8ab9404228d).

zebra-state/src/service/finalized_state.rs
upbqdn: Clarified in [67d880d](https://github.com/ZcashFoundation/zebra/pull/7035/commits/67d880d7f03a5758727de11d342faf5aedb671d5).

zebra-state/src/request.rs
upbqdn: > I made this change at https://github.com/ZcashFoundation/zebra/commit/47a6ab80fa9da6440a3d83e0f675be8029ec07b5

@oxarbitrage I thought of doing the same, but given Teor's comment, I'm not sure about that solution.

> Are we constructing contextually verified blocks from checkpoint verified blocks after doing the required verification? If we're not, that misrepresents their verification status.

The only two places where we construct `SemanticallyVerifiedBlockWithTrees` (called `ContextuallyVerifiedBlockWithTrees` before Alfredo's change) is 
1. https://github.com/zcashfoundation/zebra/blob/67d880d7f03a5758727de11d342faf5aedb671d5/zebra-state/src/service/finalized_state.rs#L283, and
2. https://github.com/zcashfoundation/zebra/blob/67d880d7f03a5758727de11d342faf5aedb671d5/zebra-state/src/service/finalized_state.rs#L298.

Both are from the new `FinalizableBlock`, which carries either a `CheckpointVerifiedBlock`, or `ContextuallyVerifiedBlock`. So what about calling it `FinalizableWithTrees`?

oxarbitrage: My apologies, i touched this PR only because i had the information that you were going to be out the whole week. 

Feel free to reverse any of changes i did.

upbqdn: The structure would look like this 
image 

I don't like that `FinalizableWithTrees` would contain `SemanticallyVerified` again, though.

upbqdn: @oxarbitrage, no bother. I want to finish it and don't leave it dangling.

upbqdn: > I don't like that FinalizableWithTrees would contain SemanticallyVerified again, though.

We could create another wrapper of `SemanticallyVerified`, call it just `Verified`, and comment in the docs that it's either a former `CheckpointVerified`, or `ContextuallyVerified` block?

zebra-state/src/request.rs
```diff
@@ -255,43 +255,58 @@ impl Treestate {
...
     /// A block ready to be committed.
-    pub block: SemanticallyVerifiedBlock,
+    pub verified: SemanticallyVerifiedBlock,
     /// The tresstate associated with the block.
-    pub treestate: Option<Treestate>,
+    pub treestate: Treestate,
 }
 
-impl ContextuallyVerifiedBlockWithTrees {
+pub enum FinalizableBlock {
```
teor2345, line 269:
```suggestion
/// Contains a block ready to be committed.
///
/// Zebra's state service passes this `enum` over to the finalized state
 /// when committing a block.
pub enum FinalizableBlock {
```

zebra-state/src/request.rs
teor2345: What about `FinalizableBlockWithTrees`, and containing a `FinalizableBlock`?

I think we can merge what we have for now, and then open a new ticket and wait for Pili to schedule it?

zebra-state/src/service/finalized_state.rs
```diff
@@ -225,53 +225,25 @@ impl FinalizedState {
...
-            None => {
+        let (height, hash, finalized) = match finalizable_block {
+            FinalizableBlock::Checkpoint {
+                checkpoint_verified,
+            } => {
+                // Checkpoint-verified blocks don't have an associated treestate, so we retrieve the
+                // treestate of the finalized tip from the database and update it for the block
+                // being committed, assuming the retrieved treestate is the parent block's
+                // treestate. Later on, this function proves this assumption by asserting that the
+                // finalized tip is the parent block of the block being commited.
```
teor2345:
```suggestion
                // finalized tip is the parent block of the block being committed.
```

teor2345 (dismissed):
Let's merge this as it is, then open tickets for any further changes.

zebra-state/src/request.rs
```diff
@@ -255,43 +255,62 @@ impl Treestate {
...
     /// The tresstate associated with the block.
-    pub treestate: Option<Treestate>,
+    pub treestate: Treestate,
 }
 
-impl ContextuallyVerifiedBlockWithTrees {
+/// Contains a block ready to be committed.
+///
+/// Zebra's state service passes this `enum` over to the finalized state
+ /// when committing a block.
```
```suggestion
/// when committing a block.
```

teor2345 (approved):

teor2345:
Here's the link to the long discussion about what to call "block with trees":
https://github.com/ZcashFoundation/zebra/pull/7035#discussion_r1242876101

teor2345:
@mergifyio update

teor2345:
Failed due to bug #7077

zebra-state/src/request.rs
upbqdn: > What about `FinalizableBlockWithTrees`, and containing a `FinalizableBlock`?

The issue is that [`FinalizableBlock`](https://github.com/zcashfoundation/zebra/blob/7e7ce2ba712e07971d96ef4c44b8af9092041ae0/zebra-state/src/request.rs#L269) contains a `Treestate` in its `Contextual` variant, so `FinalizableBlockWithTrees` would contain the `Treestate` twice if `FinalizableBlock` was of that variant.

I'll think of a suitable solution, and if I can't come up with anything, I'll open a ticket.

If you have any further ideas, please share.

teor2345: I think I assumed that we'd remove the duplicate tree field from either `FinalizableBlockWithTrees` or `FinalizableBlock`.

But I can't keep those structs in my head, so listing them in a ticket might help.

PR #7041, opened 2023-06-21
fix(network): Rate-limit inbound connections per IP.

### LINKED ISSUE #6981: Ignore repeated inbound connections from the same IP
## Motivation

This prevents some kinds of attacks, by making Zebra skip the handshake for these inbound connections.

Edit: some possible attacks are:
* A single IP can have multiple inbound connections in flight between the inbound connection task and the peer set. (About 3-4 if it uses both timeouts fully.) These connections must be successful to be sent to the peer set.
* A single IP can also deny service to other inbound peers by continually making connections, and hitting the 1 second rate-limit. These connections can be successful, failed, or time out. (So there's no guarantee they will reach the peer set.)

To avoid complex code, we could keep a list in a vector and reset every N seconds.

We could re-use the outbound reconnection interval.

### Testing

Try to make a lot of inbound connections:
- existing tests do this, their asserts can be changed
- we can copy the failing tests and adjust their test data to use different IPs to test this new code

arya2:
## Motivation

This prevents some kinds of attacks by making Zebra skip the handshake for these inbound connections.

Closes #6981.

## Solution

- Count recent inbound connection attempts by IP
- Drop connections if there were already `max_connections_per_ip` recent inbound connections from their IP
 
## Review

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -615,6 +615,9 @@ where
                 // Too many open inbound connections or pending handshakes already.
                 // Close the connection.
                 std::mem::drop(tcp_stream);
+                // Allow invalid connections to be cleared quickly,
+                // but still put a limit on our CPU and network usage from failed connections.
+                tokio::time::sleep(constants::MIN_INBOUND_PEER_FAILED_CONNECTION_INTERVAL).await;
                 continue;
             }
 
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -51,6 +51,8 @@ use crate::{
 #[cfg(test)]
 mod tests;
 
+mod recent_by_ip;
+
 /// A successful outbound peer connection attempt or inbound connection handshake.
 ///
 /// The [`Handshake`](peer::Handshake) service returns a [`Result`]. Only successful connections
@@ -576,6 +578,9 @@ where
         + Clone,
     S::Future: Send + 'static,
 {
+    #[cfg(not(any(test, feature = "proptest-impl")))]
+    let mut recent_inbound_connections = recent_by_ip::RecentByIp::default();
+
     let mut active_inbound_connections = ActiveConnectionCounter::new_counter_with(
         config.peerset_inbound_connection_limit(),
         "Inbound Connections",
@@ -603,9 +608,13 @@ where
         if let Ok((tcp_stream, addr)) = inbound_result {
             let addr: PeerSocketAddr = addr.into();
 
-            if active_inbound_connections.update_count()
-                >= config.peerset_inbound_connection_limit()
-            {
+            let should_drop = active_inbound_connections.update_count()
+                >= config.peerset_inbound_connection_limit();
+
+            #[cfg(not(any(test, feature = "proptest-impl")))]
+            let should_drop = should_drop || recent_inbound_connections.has_or_add(addr.ip());
+
+            if should_drop {
                 // Too many open inbound connections or pending handshakes already.
                 // Close the connection.
                 std::mem::drop(tcp_stream);
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -578,8 +578,8 @@ where
         + Clone,
     S::Future: Send + 'static,
 {
-    #[cfg(not(any(test, feature = "proptest-impl")))]
-    let mut recent_inbound_connections = recent_by_ip::RecentByIp::default();
+    let mut recent_inbound_connections =
+        recent_by_ip::RecentByIp::new(None, Some(config.max_connections_per_ip));
 
     let mut active_inbound_connections = ActiveConnectionCounter::new_counter_with(
         config.peerset_inbound_connection_limit(),
@@ -608,13 +608,10 @@ where
         if let Ok((tcp_stream, addr)) = inbound_result {
             let addr: PeerSocketAddr = addr.into();
 
-            let should_drop = active_inbound_connections.update_count()
-                >= config.peerset_inbound_connection_limit();
-
-            #[cfg(not(any(test, feature = "proptest-impl")))]
-            let should_drop = should_drop || recent_inbound_connections.has_or_add(addr.ip());
-
-            if should_drop {
+            if active_inbound_connections.update_count()
+                >= config.peerset_inbound_connection_limit()
+                || recent_inbound_connections.is_past_limit_or_add(addr.ip())
+            {
                 // Too many open inbound connections or pending handshakes already.
                 // Close the connection.
                 std::mem::drop(tcp_stream);
--- a/zebra-network/src/peer_set/initialize/recent_by_ip.rs
+++ b/zebra-network/src/peer_set/initialize/recent_by_ip.rs
@@ -32,6 +32,7 @@ impl Default for RecentByIp {
 }
 
 impl RecentByIp {
+    /// Creates a new [`RecentByIp`]
     pub fn new(time_limit: Option<Duration>, max_connections_per_ip: Option<usize>) -> Self {
         let (by_time, by_ip) = Default::default();
         Self {
@@ -61,10 +62,10 @@ impl RecentByIp {
         }
     }
 
-    /// Prunes entries older than [`constants::MIN_PEER_RECONNECTION_DELAY`].
+    /// Prunes entries older than `time_limit`, decrementing or removing their counts in `by_ip`.
     fn prune_by_time(&mut self, now: Instant) {
         while let Some((ip, time)) = self.by_time.pop_front() {
-            if now.saturating_duration_since(time) <= constants::MIN_PEER_RECONNECTION_DELAY {
+            if now.saturating_duration_since(time) <= self.time_limit {
                 return self.by_time.push_front((ip, time));
             }
 
@@ -80,8 +81,11 @@ impl RecentByIp {
 
 #[test]
 fn old_connection_attempts_are_pruned() {
+    const TEST_TIME_LIMIT: Duration = Duration::from_secs(5);
+
     let _init_guard = zebra_test::init();
-    let mut recent_connections = RecentByIp::default();
+
+    let mut recent_connections = RecentByIp::new(Some(TEST_TIME_LIMIT), None);
     let ip = "127.0.0.1".parse().expect("should parse");
 
     assert!(
@@ -93,22 +97,21 @@ fn old_connection_attempts_are_pruned() {
         "should be past max_connections_per_ip limit"
     );
 
-    std::thread::sleep(constants::MIN_PEER_RECONNECTION_DELAY / 3);
+    std::thread::sleep(TEST_TIME_LIMIT / 3);
 
     assert!(
         recent_connections.is_past_limit_or_add(ip),
-        "should still contain entry after a third of the reconnection delay"
+        "should still contain entry after a third of the time limit"
     );
 
-    std::thread::sleep(3 * constants::MIN_PEER_RECONNECTION_DELAY / 4);
+    std::thread::sleep(3 * TEST_TIME_LIMIT / 4);
 
     assert!(
         !recent_connections.is_past_limit_or_add(ip),
-        "should prune entry after 13/12 * MIN_PEER_RECONNECTION_DELAY"
+        "should prune entry after 13/12 * time_limit"
     );
 
     const TEST_MAX_CONNS_PER_IP: usize = 3;
-    const TEST_TIME_LIMIT: Duration = Duration::from_secs(5);
 
     let mut recent_connections =
         RecentByIp::new(Some(TEST_TIME_LIMIT), Some(TEST_MAX_CONNS_PER_IP));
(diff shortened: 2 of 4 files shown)


teor2345 (changes requested):
Some overall design advice:

This ticket and PR are about preventing repeated inbound handshakes from the same IP. (Or a small set of IPs.)

We already rate-limit these handshakes, and drop duplicates when they reach the peer set.

***Here's what I think the remaining security issues are:***

A single IP can have multiple inbound connections in flight between the inbound connection task and the peer set. (About 3-4 if it uses both timeouts fully.) These connections must be successful to be sent to the peer set.

A single IP can also deny service to other inbound peers by continually making connections, and hitting the 1 second rate-limit. These connections can be successful, failed, or time out. (So there's no guarantee they will reach the peer set.)

I should have written this in the ticket, but I hadn't thought it through, because I didn't realise it was going to be scheduled so soon.

So I'd suggest keeping a list of recent inbound connection IPs in the inbound connection task, and expiring them after a timeout.

***Outbound peers and the PeerSet are not involved:***

I don't think these security issues can be fixed by checking the connections tracked by the peer set or address book.

The `AddressBook` only tracks addresses that are valid for outbound connections. The ports of inbound connections are ephemeral, so they aren't stored in the `AddressBook`, because we can't make outbound connections to those ports.

The `PeerSet` tracks both inbound and outbound connections, but only after they successfully complete a handshake.

It's also ok to leave any duplicate outbound/inbound IPs for the peer set to drop.

***Related bugs:***

~~We're currently fixing a concurrency bug in the inbound connection handler, so I'd like to avoid adding more locking to that function. Particularly when it's hidden in trait methods.~~

Edit: it turns out the bug was #7103, a future using 100% CPU in a busy-loop. So it's not related to locking.

teor2345:
I updated my review to remove the note about a locking bug, because it turned out to be incorrect.

Edit: there's no need to work on this either just because I commented 😅

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,144 @@
...
+            time_limit: time_limit.unwrap_or(constants::MIN_PEER_RECONNECTION_DELAY),
+            max_connections_per_ip: max_connections_per_ip
+                .unwrap_or(constants::DEFAULT_MAX_CONNS_PER_IP),
+        }
+    }
+
+    /// Prunes outdated entries, checks if there's a recently attempted inbound connection with
+    /// this IP, and adds the entry to `by_time` if it wasn't already in `by_ip`.
+    ///
+    /// Returns true if there was a recently attempted inbound connection.
```
arya2:
```suggestion
    /// Returns true if the recently attempted inbound connection count is past the configured limit.
```

teor2345:
Looks like testnet Docker failed due to some issues with the testnet DNS seeders:
https://github.com/ZcashFoundation/zebra/actions/runs/5525812340/jobs/10080142157?pr=7041#step:5:161

If it continues, maybe we should restart our seeder.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,144 @@
...
+                if *count == 0 {
+                    self.by_ip.remove(&ip);
+                }
+            }
+        }
+    }
+}
+
+#[test]
+fn old_connection_attempts_are_pruned() {
```
teor2345: Optional Nitpick: can you put tests in a separate test module?

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,144 @@
...
+            self.by_time.push_back((ip, now));
+            false
+        }
+    }
+
+    /// Prunes entries older than `time_limit`, decrementing or removing their counts in `by_ip`.
+    fn prune_by_time(&mut self, now: Instant) {
+        while let Some((ip, time)) = self.by_time.pop_front() {
+            if now.saturating_duration_since(time) <= self.time_limit {
+                return self.by_time.push_front((ip, time));
```
teor2345: Optional Nitpick: Usually when a method returns nothing, we don't return its return value.

```suggestion
                self.by_time.push_front((ip, time));
                return;
```

zebra-network/src/peer_set/initialize.rs
```diff
@@ -605,6 +610,7 @@ where
 
             if active_inbound_connections.update_count()
                 >= config.peerset_inbound_connection_limit()
+                || recent_inbound_connections.is_past_limit_or_add(addr.ip())
             {
                 // Too many open inbound connections or pending handshakes already.
```
teor2345, line 615: If an attacker continuously makes connections to Zebra, they could make this task use 100% CPU. This attack becomes easier the more code we add to the failed connection loop.

Can we add a short delay *after* dropping the connection?
```rust
                tokio::time::sleep(constants::MIN_INBOUND_PEER_FAILED_CONNECTION_INTERVAL).await;
```

If they use different IP addresses (easy with IPv6), they can also add an unlimited number of entries to `recent_inbound_connections`.

The change above limits the number of entries to `MIN_PEER_RECONNECTION_DELAY / MIN_INBOUND_PEER_FAILED_CONNECTION_INTERVAL = 11,900`. This takes about 1 MB.

Do you think it would take a lot of CPU to create and prune twelve thousand entries?
Should we put a lower limit on the number of entries?

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,144 @@
...
+            by_time,
+            by_ip,
+            time_limit: time_limit.unwrap_or(constants::MIN_PEER_RECONNECTION_DELAY),
+            max_connections_per_ip: max_connections_per_ip
+                .unwrap_or(constants::DEFAULT_MAX_CONNS_PER_IP),
+        }
+    }
+
+    /// Prunes outdated entries, checks if there's a recently attempted inbound connection with
+    /// this IP, and adds the entry to `by_time` if it wasn't already in `by_ip`.
```
teor2345:
```suggestion
    /// this IP, and adds the entry to `by_time`, and `by_ip` if needed.
```

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,144 @@
...
+    time::{Duration, Instant},
+};
+
+use crate::constants;
+
+#[derive(Debug)]
+/// Stores IPs of recently attempted inbound connections.
+pub struct RecentByIp {
+    /// The list of IPs in increasing connection time order.
+    pub by_time: VecDeque<(IpAddr, Instant)>,
```
teor2345, line 18: Optional: If we use `BTreeMap`, it will automatically sort times in increasing order. We can also use `self.by_time.split_off(now + self.time_limit)` to find the expired entries, which is slightly more CPU efficient than popping and comparing each entry.

https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.split_off

```suggestion
    pub by_time: BTreeMap<Instant, IpAddr>,
```

teor2345 (dismissed):
This looks great, thanks so much for rewriting it! I really appreciate the simpler code.

I have some questions about how this code will handle an attack where Zebra is flooded with inbound connections from different IP addresses. It would be good to fix them here or in a quick follow-up.

I was also wondering about the PR title. I think what this code does now is rate-limit inbound connections to `max_connections_per_ip` per `MIN_PEER_RECONNECTION_DELAY`, rather than ignoring them completely?
(Which still fixes the original bug where the handshaker could get overloaded.)

Otherwise all my suggestions are optional.

As a follow-up, I think we might want to hide these `IpAddr` and the ones in the `AddressBook` from the logs, in the same way as `PeerSocketAddr`. But I think that's a low priority, because we don't actually log those types or instrument their functions. So the `IpAddr`s we've added won't appear in any logs yet.

zebra-network/src/peer_set/initialize.rs
arya2: Great catch.

> Can we add a short delay after dropping the connection?

Yes.

> The change above limits the number of entries to MIN_PEER_RECONNECTION_DELAY / MIN_INBOUND_PEER_FAILED_CONNECTION_INTERVAL = 11,900

~It may not limit the number of entries at all, it would currently drop the connection and continue with the loop if an attacker used the same IP repeatedly.~

Update: it wouldn't add the entry if it's continuing without sleeping, but still a mistake.

> Do you think it would take a lot of CPU to create and prune twelve thousand entries?

Yeah, far more than warranted.

> Should we put a lower limit on the number of entries?

Good idea, yeah, there should be a constant limit.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7041/commits/397e70391a5b80a0097b58f740fc1a9eb8c0e7a8 and noted.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
arya2: Moved in https://github.com/ZcashFoundation/zebra/pull/7041/commits/397e70391a5b80a0097b58f740fc1a9eb8c0e7a8.

zebra-network/src/peer_set/initialize.rs
arya2: > Can we add a short delay after dropping the connection?

Applied in https://github.com/ZcashFoundation/zebra/pull/7041/commits/397e70391a5b80a0097b58f740fc1a9eb8c0e7a8.

> Do you think it would take a lot of CPU to create and prune twelve thousand entries?

It's a bit better now that it's using `split_off()`, and I don't think it would take 10ms to update `by_ip`, but I'm not sure.

> Do you think it would take a lot of CPU to create and prune twelve thousand entries?
> Should we put a lower limit on the number of entries?

I don't think it would get there, because if it adds the entry, then it should sleep for `MIN_INBOUND_PEER_CONNECTION_INTERVAL` (1 second), I think the default limit right now is 30.

It doesn't add entries when `active_inbound_connections.update_count() >= config.peerset_inbound_connection_limit()` because Zebra is dropping the connection without initiating a handshake.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
arya2: Good idea.

I kept the `VecDeque` because I wasn't sure how to make a `BTreeMap` work faster with a configurable `max_connections_per_ip`, but changed it to use `partition_point()` (binary search) and `split_off()` https://github.com/ZcashFoundation/zebra/pull/7041/commits/397e70391a5b80a0097b58f740fc1a9eb8c0e7a8.

zebra-network/src/peer_set/initialize.rs
teor2345: Great, let's merge it and see how it goes?

zebra-network/src/peer_set/initialize/recent_by_ip/tests.rs
```diff
@@ -0,0 +1,67 @@
+use std::time::Duration;
```
teor2345, line 3: Optional Nitpick: please add a module comment

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use crate::constants;
+
+#[cfg(test)]
+mod tests;
+
+#[derive(Debug)]
+/// Stores IPs of recently attempted inbound connections.
+pub struct RecentByIp {
+    /// The list of IPs in increasing connection time order.
```
teor2345: New IPs are added to the back, and old IPs are split from the front.

I originally read "time" as "how long ago the connection was made" (connection age or connected time). But I think you meant "the instant the connection was made" (connection start or initiation time).

```suggestion
    /// The list of IPs in decreasing connection age order.
```

```suggestion
    /// The list of IPs in increasing connection initiation time order.
```

teor2345: I added this to PR #7213, we could also say both if you like, see PR #7214 as an example.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
teor2345: I opened PR #7214 as an example of this change, please don't feel like you have to merge it. PR #7213 also contains a necessary simplification for this change: moving the time calculation out of the partition closure.

zebra-network/src/peer_set/initialize/recent_by_ip/tests.rs
teor2345: I fixed this in PR #7213.

teor2345 (dismissed):
I think this is good to go, feel free to accept or reject my suggestions in PRs #7213 and #7214.

teor2345:
> I was also wondering about the PR title. I think what this code does now is rate-limit inbound connections to `max_connections_per_ip` per `MIN_PEER_RECONNECTION_DELAY`, rather than ignoring them completely?
> (Which still fixes the original bug where the handshaker could get overloaded.)

Oh this is also an optional nitpick, I doubt most users will notice the difference between "Ignore for 2 minutes" and "rate-limit for 2 minutes".

zebra-network/src/peer_set/initialize/recent_by_ip/tests.rs
arya2: Whoops, thank you, merged #7213.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
arya2: > But I think you meant "the instant the connection was made"

Yep. I often struggle with these wordings, thank you for the assist.

Merged #7213.

teor2345 (approved):

PR #7045, opened 2023-06-22
refactor(docker): allow more flexible zebra configuration

### LINKED ISSUE #6752: Pass mining config to the mining Docker container at runtime
## Motivation

In order to use Zebra for mining, a user needs to set their Zcash address in Zebra's config. Zebra uses this address in coinbase transactions to pay the mining reward.

We have a Docker image that packages Zebra with support for mining. However, since Zebra lives inside this image, the user needs to be able to pass the address to the container at runtime.

Users of the image will also need to be able to choose if they want to mine on Testnet or Mainnet. They will need to pass this information to the container at runtime as well.

## Solution

We need to enable passing of `network.network` and `mining.miner_address` to Zebra's config at the runtime of the Docker container that has mining enabled.

We could set a default `mining.miner_address`, but if a miner accidentally forgets to set their own, and they happen to mine a block on Mainnet, they might be upset. I would therefore lean towards forcing users to set their `mining.miner_address` before starting the Docker container. This is also consistent with non-containerized Zebra. I'd do the same for Testnet so that there's no difference.

### LINKED ISSUE #6759: Inconsistent management of ports in the Dockerfile
## Motivation

- I noticed we set the ports for `metrics` and `tracing` for Zebra in the Dockerfile (https://github.com/zcashfoundation/zebra/blob/3706ff6e67e3214a7acc8beb98dacd878771f0ab/docker/Dockerfile#L191-L194), but we don't expose them (https://github.com/zcashfoundation/zebra/blob/3706ff6e67e3214a7acc8beb98dacd878771f0ab/docker/Dockerfile#L198). Note that neither `metrics` nor `tracing` are enabled in our published images, but users could enable them when building the image manually.
- We allow setting an arbitrary RPC port (https://github.com/zcashfoundation/zebra/blob/3706ff6e67e3214a7acc8beb98dacd878771f0ab/docker/Dockerfile#L189), which seems like a redundant overhead because users can pick any port when they are publishing the ports at runtime.

gustavovalverde:
## Motivation

Changing the configuration was a bit confusing for Docker users and it invalidated a specific use-case for users using mounting their configuration files from their host machine, which is very common when using `docker-compose`

This also fixes some inconsistencies between ports, variable configurations and expected outcome.

Fixes #6759
It creates a path to fix ~~extra text so this ticket doesn't auto close~~ #6752

Depends-On: #7055

## Solution

Set default values in the `entrypoint`, and allow users to:
- Mount their own configuration file
- Expose Zebra outside the container or not
- Turn off sync, or keep it enabled
- enable `metrics` and `tracing`, exposing them or not

Having the `-x` option prints variable expansions, so we don't have to echo each value.

## Review

Review by commit, as the file renaming makes the whole review harder.

### Reviewer Checklist

## Follow Up Work

After this is merge we can continue with #6752 and documentation


teor2345 (commented):
This is going to be tricky to review, because the files are renamed and edited in the same PR. So the diffs are very complicated.

Would you mind splitting this PR into a bug fix PR, and a file rename PR?

gustavovalverde:
> This is going to be tricky to review, because the files are renamed and edited in the same PR. So the diffs are very complicated.
> 
> Would you mind splitting this PR into a bug fix PR, and a file rename PR?

@teor2345 yes, reviewing as a whole is tricky, but it should be considerably easier reviewing the 3 commits one by one (I made each commit as it would be a different PR).

Nonetheless, if a separate PR is prefered I can do that also.

gustavovalverde:
Let's keep this simple. I renamed the files back to their original name. I'll do that in a different PR.

cc: @upbqdn you might want to review this also.

docker/Dockerfile
```diff
@@ -161,45 +162,9 @@ ENV NETWORK ${NETWORK:-Mainnet}
...
-
-# Set this to change the default cached state directory
-ARG ZEBRA_CACHED_STATE_DIR
-ENV ZEBRA_CACHED_STATE_DIR ${ZEBRA_CACHED_STATE_DIR:-/var/cache/zebrad-cache}
-
-# Zebra automatically detects if it is attached to a terminal, and uses colored output.
-# Set this to 'true' to force using color even if the output is not a terminal.
-# Set this to 'false' to disable using color even if the output is a terminal.
-ARG LOG_COLOR
-ENV LOG_COLOR ${LOG_COLOR}
```
teor2345, line 198: This variable makes sure that deployed Google Cloud instances have readable logs, by disabling colour escape codes. Can you check that still works?

https://github.com/ZcashFoundation/zebra/actions/workflows/continous-delivery.yml

docker/Dockerfile
```diff
@@ -161,45 +162,9 @@ ENV NETWORK ${NETWORK:-Mainnet}
...
-ENV ZEBRA_CONF_PATH ${ZEBRA_CONF_PATH:-$ZEBRA_CONF_DIR/$ZEBRA_CONF_FILE}
-
-# Other settings
-
-ARG SHORT_SHA
-ENV SHORT_SHA ${SHORT_SHA}
-
-# Set this to send sentry reports when Zebra crashes
-ARG SENTRY_DSN
-ENV SENTRY_DSN ${SENTRY_DSN}
```
teor2345, line 223: These variables control logging and panic reports:
- SHORT_SHA tells us the git commit in the logs and the sentry reports
- SENTRY_DSN sends crash reports to sentry, which sends a notification to Slack

The deployed Google Cloud instances should log both values when they start up. If that's not working, we'll see messages like "sentry is disabled", and log lines without any git commit hashes in them.

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+: "${LOG_COLOR:='false'}"
+# [rpc]
+: "${RPC_LISTEN_ADDR:='127.0.0.1'}"
 
 # Create the conf path and file if it does not exist.
-mkdir -p "$ZEBRA_CONF_DIR"
-touch "$ZEBRA_CONF_PATH"
+if [[ -n "$ZEBRA_CONF_PATH" ]]; then
+    mkdir -p "$ZEBRA_CONF_DIR"
+    touch "$ZEBRA_CONF_FILE"
```
teor2345: The file is just a file name, not a path, so this command creates an empty config file in the current directory. I don't think that actually does anything useful.

Looking at the script below, it seems like we might not want to create an empty file here. (Which is what `touch` does: it creates a file if it doesn't exist, and updates its timestamp if it does.)

```suggestion
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
-echo "SHORT_SHA=$SHORT_SHA"
-echo "SENTRY_DSN=$SENTRY_DSN"
+# [network]
+: "${NETWORK:='Mainnet'}"
+: "${ZEBRA_LISTEN_ADDR:='127.0.0.1'}"
+# [consensus]
+: "${ZEBRA_CHECKPOINT_SYNC:='true'}"
+# [state]
+: "${ZEBRA_CACHED_STATE_DIR:='/var/cache/zebrad-cache'}"
+: "${LOG_COLOR:='false'}"
```
teor2345: Nitpick: The logging config is in the tracing section, not state.

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
-mkdir -p "$ZEBRA_CONF_DIR"
-touch "$ZEBRA_CONF_PATH"
+if [[ -n "$ZEBRA_CONF_PATH" ]]; then
+    mkdir -p "$ZEBRA_CONF_DIR"
+    touch "$ZEBRA_CONF_FILE"
+fi
 
 # Populate `zebrad.toml` before starting zebrad, using the environmental
-# variables set by the Dockerfile.
+# variables set by the Dockerfile or the user.
```
teor2345:
```suggestion
# variables set by the Dockerfile or the user. If the user has already created a config, don't replace it.
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+if [[ -n "$ZEBRA_CONF_DIR" ]] && [[ -n "$ZEBRA_CONF_FILE" ]]; then
+    ZEBRA_CONF_PATH="$ZEBRA_CONF_DIR/$ZEBRA_CONF_FILE"
+fi
 
-echo "Other variables:"
-echo "SHORT_SHA=$SHORT_SHA"
-echo "SENTRY_DSN=$SENTRY_DSN"
+# [network]
+: "${NETWORK:='Mainnet'}"
+: "${ZEBRA_LISTEN_ADDR:='127.0.0.1'}"
```
teor2345: This changes the default from 0.0.0.0, which is a breaking change. The Zcash listener is expected to be available outside the container.

```suggestion
: "${ZEBRA_LISTEN_ADDR:='0.0.0.0'}"
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
-#
-# TODO:
-#  - make `cache_dir`, `metrics.endpoint_addr`, and `tracing.endpoint_addr` into Docker arguments
-#  - add an $EXTRA_CONFIG or $REPLACEMENT_CONFIG environmental variable
+if [[ ! -f "$ZEBRA_CONF_PATH" ]]; then
 cat <<EOF > "$ZEBRA_CONF_PATH"
 [network]
 network = "$NETWORK"
-listen_addr = "0.0.0.0"
+listen_addr = $ZEBRA_LISTEN_ADDR:8233
```
teor2345: 8233 is the default port for mainnet. If it is not supplied, Zebra automatically chooses the default port for the configured network.

Changing this port would be a breaking change for testnet.

https://doc.zebra.zfnd.org/zebra_network/struct.Config.html#structfield.listen_addr

```suggestion
listen_addr = $ZEBRA_LISTEN_ADDR
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+
+[consensus]
+checkpoint_sync = $ZEBRA_CHECKPOINT_SYNC
```
teor2345: `checkpoint_sync = false` is a deprecated insecure config, I don't think we should support it in new work:
https://doc.zebra.zfnd.org/zebra_consensus/struct.Config.html#security-1

```suggestion
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
 [state]
-cache_dir = "$ZEBRA_CACHED_STATE_DIR"
+cache_dir = $ZEBRA_CACHED_STATE_DIR
+EOF
 
+if [[ -n "$METRICS_ENDPOINT_ADDR" ]]; then
+cat <<EOF >> "$ZEBRA_CONF_PATH"
 [metrics]
-#endpoint_addr = "0.0.0.0:9999"
+endpoint_addr = ${METRICS_ENDPOINT_ADDR}:9999
```
teor2345: Let's allow the user to choose their own port, rather than forcing them to use 9999?

```suggestion
endpoint_addr = ${METRICS_ENDPOINT_ADDR}
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+# [network]
+: "${NETWORK:='Mainnet'}"
+: "${ZEBRA_LISTEN_ADDR:='127.0.0.1'}"
+# [consensus]
+: "${ZEBRA_CHECKPOINT_SYNC:='true'}"
+# [state]
+: "${ZEBRA_CACHED_STATE_DIR:='/var/cache/zebrad-cache'}"
+: "${LOG_COLOR:='false'}"
+# [rpc]
+: "${RPC_LISTEN_ADDR:='127.0.0.1'}"
```
teor2345: Changing the default RPC bind address is a breaking change. RPC ports are (sometimes) expected to be available to other containers on the same host, but not exposed to the internet.

```suggestion
: "${RPC_LISTEN_ADDR:='0.0.0.0'}"
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+listen_addr = ${RPC_LISTEN_ADDR}:${RPC_PORT}
 EOF
 fi
 
-if [[ -n "$LOG_FILE" ]] || [[ -n "$LOG_COLOR" ]]; then
+if [[ -n "$TRACING_ENDPOINT_ADDR" ]]; then
 cat <<EOF >> "$ZEBRA_CONF_PATH"
 [tracing]
-#endpoint_addr = "0.0.0.0:3000"
+endpoint_addr = "${TRACING_ENDPOINT_ADDR}:3000"
```
teor2345: Allow the user to choose their own port.

```suggestion
endpoint_addr = "${TRACING_ENDPOINT_ADDR}"
```

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
 [rpc]
-listen_addr = "0.0.0.0:${RPC_PORT}"
+listen_addr = ${RPC_LISTEN_ADDR}:${RPC_PORT}
 EOF
 fi
 
-if [[ -n "$LOG_FILE" ]] || [[ -n "$LOG_COLOR" ]]; then
+if [[ -n "$TRACING_ENDPOINT_ADDR" ]]; then
 cat <<EOF >> "$ZEBRA_CONF_PATH"
 [tracing]
```
teor2345, line 73: The tracing section must go before the endpoint address, log colour, and log path, if any of them are enabled. Currently, this script only adds it if the endpoint address is enabled.

I suggest always putting it in the config, empty sections are valid.

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+# Set this to 'false' to disable using color even if the output is a terminal.
+if [[ "$LOG_COLOR" = "true" ]] || [[ "$LOG_COLOR" = "false" ]]; then
 cat <<EOF >> "$ZEBRA_CONF_PATH"
-force_use_color = true
-EOF
-elif [[ "$LOG_COLOR" = "false" ]]; then
-cat <<EOF >> "$ZEBRA_CONF_PATH"
-use_color = false
+force_use_color = $LOG_COLOR
 EOF
```
teor2345: The previous script set `force_use_color` if the env var was true, and `use_color` if it was false. This is the correct behaviour, changing it is a breaking change.

Currently in this PR there's no way to force-disable colours.

docker/Dockerfile
```diff
@@ -57,6 +57,7 @@ RUN if [ "$(uname -m)" != "aarch64" ]; then \
     && \
     rm -rf /var/lib/apt/lists/* /tmp/*
 
+# TODO: this ENV variables are not required at build time, so they're not needed as ARGs either
```
teor2345: The backtrace, colorbt, rust_log, and cargo_home variables are required at build time.

docker/runtime-entrypoint.sh
teor2345: The listener address change will need a manual check, like the one we did for the Ziggurat testnet instance using: `nc -v -z IP port`

I guess each new or changed port might need a test like this?

teor2345 (dismissed):
This looks like a good rewrite, thanks for cleaning things up.

It seems to be more than a refactor. Do you think we should rename the PR, so users can see it contains new functionality?

Some of these changes break compatibility with older docker scripts and configs. Is this deliberate?

What's the plan for testing this PR? Did you do a manual deploy?
(The manual deploy also does some tests of the mainnet and testnet configs.)

I've added comments about some changes that will need manual checks of the Google Cloud logs, or using network commands.

teor2345:
> Let's keep this simple. I renamed the files back to their original name. I'll do that in a different PR.

Thanks!

GitHub doesn't have a good interface for reviewing by commit, comments stay on the original commit, and I can't make suggestions.

docker/Dockerfile
```diff
@@ -161,45 +162,9 @@ ENV NETWORK ${NETWORK:-Mainnet}
...
-
-# Zebra automatically detects if it is attached to a terminal, and uses colored output.
-# Set this to 'true' to force using color even if the output is not a terminal.
-# Set this to 'false' to disable using color even if the output is a terminal.
-ARG LOG_COLOR
-ENV LOG_COLOR ${LOG_COLOR}
-
 # Expose configured ports
-
 EXPOSE 8233 18233 $RPC_PORT
```
upbqdn: @teor2345 suggests we let users set their own ports in `runtime-entrypoint.sh`, which is inconsistent with this instruction. Plus, some ports are missing, and at least one will always be redundant.

I guess we could just get rid of it completely, since

> The `EXPOSE` instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.

The `EXPOSE` instruction is useful for automated builds using the `-P` flag, but I don't think it makes much sense for Zebra, since

> The `-P` option publishes all the ports to the host interfaces. Docker binds each exposed port to a random port on the host.

upbqdn (dismissed):
utACK, except for the `EXPOSE` instruction.

docker/runtime-entrypoint.sh
```diff
@@ -1,82 +1,95 @@
...
+endpoint_addr = ${METRICS_ENDPOINT_ADDR}:9999
 EOF
+fi
 
+# Set this to enable the RPC port
 if [[ -n "$RPC_PORT" ]]; then
 cat <<EOF >> "$ZEBRA_CONF_PATH"
 [rpc]
-listen_addr = "0.0.0.0:${RPC_PORT}"
+listen_addr = ${RPC_LISTEN_ADDR}:${RPC_PORT}
```
upbqdn: I could be useful if we detect the `getblocktemplate-rpcs` feature in `$FEATURES` and bind the default port if it is present.

docker/runtime-entrypoint.sh
upbqdn: I could be useful if we detect the `prometheus` feature in `$FEATURES` and bind the default port if it is present.

docker/runtime-entrypoint.sh
upbqdn: I could be useful if we detect the `filter-reload` feature in `$FEATURES` and bind the default port if it is present.

upbqdn (commented):
I noticed we could automatically bind some ports if the image was built with their corresponding feature.

upbqdn:
AFAIK, the effects of `runtime-entrypoint.sh` persist in the container, and `runtime-entrypoint.sh` is run on each restart. I guess it's good we're writing to the config file only if it doesn't exist, but then users can't alter the config by using different values of env vars when re-running the container. We should document this since I don't think this would be the expected runtime behavior. 

Another solution would be to overwrite the file each time (that's what we've been doing before this PR), but I guess we don't want that.

Also, we should never arbitrarily append to the config because Zebra doesn't start when there are redefined tables.

docker/Dockerfile
upbqdn: Also, the exposed `$RPC_PORT` can differ from the one used in `runtime-entrypoint.sh`.

upbqdn:
Or we could overwrite the config file if there are any relevant env vars present. That would suggest that the user intends to use the env vars instead of an external config file.

docker/Dockerfile
teor2345: Thanks, I wasn't expecting it to work that way!

But I was going to ask a question about updating EXPOSE.

teor2345:
> Or we could overwrite the config file if there are any relevant env vars present. That would suggest that the user intends to use the env vars instead of an external config file.

Another alternative is having a specific env var to force updating the config, or to force no updates, in case our automatic config update logic doesn't work for some use cases.

teor2345:
> What's the plan for testing this PR? Did you do a manual deploy?
> (The manual deploy also does some tests of the mainnet and testnet configs.)

Since we're going to be updating this config generator a few times over the next few weeks, I'll activate the release Docker image tests on each PR.

Then we can automate some of the tests for these environmental variables. Let me know if you like help with that.

(Previously the tests ran on every main branch merge, before we deployed machines. But I turned off deployments on main branch merges until we had cached state, to reduce network load. I didn't really mean to turn off the tests, I don't want to discover they're failing only after I've tagged the next release.)

docker/runtime-entrypoint.sh
teor2345: That feature should also require a mining address, but we can do that in another ticket.

docker/Dockerfile
upbqdn: Clarification: I took the cited parts from the official docs.

teor2345:
> > What's the plan for testing this PR? Did you do a manual deploy?
> > (The manual deploy also does some tests of the mainnet and testnet configs.)
> 
> Since we're going to be updating this config generator a few times over the next few weeks, I'll activate the release Docker image tests on each PR.
>
> Then we can automate some of the tests for these environmental variables. Let me know if you like help with that.

This is done in PR #7055, which should make testing this PR a lot easier. I've marked this PR so it will merge after #7055.

docker/runtime-entrypoint.sh
gustavovalverde: Fixed with [fix(docker): revert some breaking changes](https://github.com/ZcashFoundation/zebra/pull/7045/commits/b7a6fde8ddea961ea2cc67f7cf7f5bd7a1f4701e)

docker/runtime-entrypoint.sh
gustavovalverde: Fixed with [imp(docker): allow more flexibility with FEATURES config](https://github.com/ZcashFoundation/zebra/pull/7045/commits/6039c5a89b6f6c08d0132c7499471f5be9e97f42)

docker/runtime-entrypoint.sh
gustavovalverde: Fixed with [imp(docker): allow more flexibility with FEATURES config](https://github.com/ZcashFoundation/zebra/pull/7045/commits/6039c5a89b6f6c08d0132c7499471f5be9e97f42)

docker/runtime-entrypoint.sh
gustavovalverde: Change reverted

docker/Dockerfile
gustavovalverde: I've removed $RPC_PORT and left `8233 18233` as

> It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.

gustavovalverde:
> but then users can't alter the config by using different values of env vars when re-running the container

What would be `re-running` in this context? As this Dockerfile would create a _stateless_ container, unless a volume is mounted on a directory, like we do with the cached states, or how we recently saw a user mounting the config file.

If, for example, the container crashes and gets restarted, the state is lost and you start with a fresh image.

P.S.: We should also document that if the state directory is not mounted somehow, it will get lost on a restart.

upbqdn:
> What would be re-running in this context?

I thought of `docker stop` and `docker start`, assuming `docker start` could take env vars, which was a wrong assumption. It turns out Docker doesn't properly support changing env vars in containers: https://github.com/moby/moby/issues/8838, which seems to be a good thing. We can therefore ignore this use case. Sorry for the confusion.

docker/Dockerfile
```diff
@@ -158,12 +158,8 @@ RUN apt-get update && \
 ARG NETWORK
 ENV NETWORK ${NETWORK:-Mainnet}
 
-# Set this to enable the RPC port
-ARG RPC_PORT
-ENV RPC_PORT ${RPC_PORT}
-
 # Expose configured ports
-EXPOSE 8233 18233 $RPC_PORT
+EXPOSE 8233 18233
```
upbqdn, line 184: Zebra uses only one of these ports, depending on whether it runs on Mainnet or Testnet. Is it OK to have an extra port here?

Also, what about the rest of the ports?

gustavovalverde: > Is it OK to have an extra port here?

Considering that at least one of these ports is expected to _always_ be exposed, yes.

> Also, what about the rest of the ports?

If there's any other port that's expected to be opened by default, it should be added here.

gustavovalverde:
BTW, this was supposed to be a Draft PR, but it's too late for that.

I started this deploy so we can confirm we have the expected values on the GCP instances https://github.com/ZcashFoundation/zebra/actions/runs/5371770342

gustavovalverde:
Started a new deploy after some fixes:

https://github.com/ZcashFoundation/zebra/actions/runs/5372451179

docker/Dockerfile
gustavovalverde: Confirmed with deployed instance: https://console.cloud.google.com/compute/instancesDetail/zones/us-east1-c/instances/zebrad-fix-entrypoint-ports-ad5a3a4-mainnet?authuser=1&project=zfnd-dev-zebra

docker/Dockerfile
gustavovalverde: Confirmed with deployed instance: https://console.cloud.google.com/compute/instancesDetail/zones/us-east1-c/instances/zebrad-fix-entrypoint-ports-ad5a3a4-mainnet?authuser=1&project=zfnd-dev-zebra

gustavovalverde:
This should be good to go now

teor2345:
This needs to be rebased to pick up PR #7063, I'll do that now.

teor2345:
@mergifyio update

teor2345 (approved):
Looks good, thanks for making these changes.

Optional: `cargo` accepts space-separated and comma-separated features, do you want to match them as well? Comma-separated seems to be the preferred format.

https://doc.rust-lang.org/cargo/commands/cargo-build.html#feature-selection

I'd suggest making a separate ticket for that change.

teor2345:
Temporary network error:
> Error: buildx failed with: ERROR: failed to solve: failed to push us-docker.pkg.dev/zfnd-dev-zebra/zebra/zebrad-test:pr-7045: failed to copy: failed to do request: Put "https://us-docker.pkg.dev/artifacts-uploads/namespaces/zfnd-dev-zebra/repositories/zebra/uploads/...?digest=sha256%3A036d5ebcdf9f2c294bda9859f6cb453423042dce221cefb9507edde556914b66": write tcp 172.17.0.2:56886->142.251.2.82:443: write: connection reset by peer

https://github.com/ZcashFoundation/zebra/actions/runs/5385043749/jobs/9773642536?pr=7045#step:10:1532

gustavovalverde:
@Mergifyio update

teor2345:
@mergifyio refresh

PR #7047, opened 2023-06-22
fix(deps): Replace openssl with rustls in tests and experimental features

teor2345:
## Motivation

OpenSSL continues to have memory safety vulnerabilities:
https://github.com/advisories/GHSA-xcf7-rvmh-g6q4

`zebrad` isn't impacted by this one in production, because openssl is only a dependency of our tests and experimental features. But it could accidentally become a production dependency when we stabilise features or add new dependencies.

### Reference

GitHub Actions ternary expressions:
https://docs.github.com/en/actions/learn-github-actions/expressions#example

`--all-features` activates the openssl dependency, because it's an optional dependency of some of Zebra's dependencies. So I changed the CI check to ignore banned crates for `--all-features`. (And ignore missing dependencies for default features.)

## Solution

- Ban `openssl` as a Zebra dependency
- Make CI check that openssl is not used in default production builds and production Docker images

Related fixes:
- Add a `default-release-binaries` feature to `zebrad` and use it in Docker files and production/prod test workflows
- Ignore expected warnings for default features and `default-release-binaries`
- Ban wildcard ("any version") dependencies, except in tests

## Review

This is a low priority security fix.

### Reviewer Checklist

## Follow Up Work

It would be good to automate keeping the duplicate dependency exception list up to date.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -38,7 +38,7 @@ on:
       # https://github.com/ZcashFoundation/zebra/blob/main/docker/Dockerfile#L83
       features:
         required: false
-        default: "default-docker"
+        default: "default-release-binaries"
         type: string
       test_features:
         required: false
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -38,7 +38,7 @@ on:
       # https://github.com/ZcashFoundation/zebra/blob/main/docker/Dockerfile#L83
       features:
         required: false
-        default: "sentry"
+        default: "default-docker"
         type: string
       test_features:
         required: false
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -261,7 +261,7 @@ jobs:
         # We don't need to check `--no-default-features` here, because (except in very rare cases):
         # - disabling features isn't going to add duplicate dependencies
         # - disabling features isn't going to add more crate sources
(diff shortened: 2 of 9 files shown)


gustavovalverde:
I don't think I'm following up why we would require a new `feature` name (`default-docker`) shouldn't `default sentry` have the same behavior? It might be confusing as why Docker needs a "special" flag.

gustavovalverde (commented):
I don't think I'm following up why we would require a new feature name (default-docker) shouldn't default sentry have the same behavior? It might be confusing as why Docker needs a "special" flag.

teor2345:
> I don't think I'm following up why we would require a new feature name (default-docker) shouldn't default sentry have the same behavior? It might be confusing as why Docker needs a "special" flag.

At the moment we're repeating the `sentry` feature in about 4 different places in our test and production release CI workflows. And this PR adds a 5th location that's security-sensitive. (Banning OpenSSL in released binaries.)

So if we change our set of default features in released binaries, we'll need to remember to update all those locations. But we'll probably forget one!

Would it be clearer if I renamed it to something like:
* `default-release-binaries`
* `zf-published-binaries`
* `zf-binary-releases`

We should be using the same features for all our released binaries. Even if we're only releasing Docker binaries at the moment.

gustavovalverde:
Understood!

I think both of this options should be good:
- default-release-binaries
- zf-binary-releases

I'd vote for the former: `default-release-binaries`

teor2345:
Done, and merge conflicts resolved.

gustavovalverde (approved):
LGTM!

Issue #6638, opened 2023-05-09
Remove duplicate dependencies from `deny.toml` once `ed25519-zebra` and `zcashd` are released


mpguerra:
## Motivation

We want to remove as many duplicate dependencies as possible from the zebra codebase. This should be possible after https://github.com/zcash/zcash/issues/6648 currently planned for the the `zcashd` 5.6.0 release.

### Specifications

With the `zcashd` 5.6.0 release, we we can upgrade and delete these dependencies in `deny.toml`

https://github.com/ZcashFoundation/zebra/blob/39c599713abce08654de30a4f99c2bdcb69a21b1/deny.toml#L71-L74

https://github.com/ZcashFoundation/zebra/blob/39c599713abce08654de30a4f99c2bdcb69a21b1/deny.toml#L85-L89

https://github.com/ZcashFoundation/zebra/blob/39c599713abce08654de30a4f99c2bdcb69a21b1/deny.toml#L96-L97

https://github.com/ZcashFoundation/zebra/blob/39c599713abce08654de30a4f99c2bdcb69a21b1/deny.toml#L76-L78

https://github.com/ZcashFoundation/zebra/blob/4b6e2fe2236956d32102bddf74a979cb97fbf095/deny.toml#L64-L65

## Related Work

- #6624 
- https://github.com/zcash/zcash/issues/6648 
- https://github.com/ZcashFoundation/ed25519-zebra/issues/79

teor2345:
@mpguerra the next step here is to publish an ed25519-zebra release. The updated dependencies are in its main branch, but they haven't been released yet.

Once ed25519-zebra is done, it should be automatically picked up by the zcashd and zcash_script release processes. Then Zebra will pick it up automatically using dependabot.

Just a note: the deny.toml entries linked in the ticket will be outdated by the time we do this ticket, because PR #6624 will modify them.

(This ticket is also related to https://github.com/ZcashFoundation/ed25519-zebra/issues/77 , because the release instructions for all our crates should include updating dependencies.)

daira:
Note that we are cutting an rc of zcashd, so this is at risk of not getting into zcashd 5.6.0 if it is not done very soon.

mpguerra:
tentatively scheduling for next sprint

teor2345:
This needs to be done as part of #6859, can we close this ticket and just make it part of the PR for #6959?

[...]

PR #7053, opened 2023-06-23
build(deps): update ecc dependencies for `zcashd` 5.6.0, and create legacy state format compatibility layer

### LINKED ISSUE #6859: Tracking: Upgrade shared ECC dependencies and zcash_script for zcashd 5.6.0

oxarbitrage:
## Motivation

Update ecc dependencies.

Close https://github.com/ZcashFoundation/zebra/issues/6859

Depends-On: #7158

## Solution

Create legacy code ion order to keep serialization the same as it was.

## Review

@teor2345 and @conradoplg . There was already a lot of discussion about the options, solutions, etc.

### Reviewer Checklist


teor2345:
@dconnolly can you help out with this PR since Marek isn't available?

teor2345:
I added some sprout snapshot tests in PR #7057, and marked this PR so that those tests will merge first.

Otherwise, I'm not sure how much I can help with this PR. (Deirdre and Marek did the reviews on the tree changes.)

oxarbitrage:
I was able to make some progress by using some of the read/write functions from `librustzcash::zcash_primitives::merke_tree` @conradoplg pointed me too.

There are several things pending, some of them are:

- There are no functions in the library to deal with sprout.
- The library has `read_frontier_v0` and `read_frontier_v1` but only `write_frontier_v1`. It is not totally clear what to use n each case but it seems `read_frontier_v0` should be used in sapling.
- Checking the output of the following serialization tests:

Sapling:
```
$ cargo test sapling_note_commitment_tree_serialization -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 3.17s
     Running unittests src/lib.rs (target/debug/deps/tower_batch_control-c2bd02070ca729ce)

...

running 1 test

  left: `"0102000000000000007c3ea01a6e3a3d90cf59cd789e467044b5cd78eb2c84cc6816f960746d0e036c000162324ff2c329e99193a74d28a585a3c167a93bf41a255135529c913bd9b1e666ddaa1ab86de5c153993414f34ba97e9674c459dfadde112b89eeeafa0e5a204c"`,
 right: `"0102007c3ea01a6e3a3d90cf59cd789e467044b5cd78eb2c84cc6816f960746d0e036c0162324ff2c329e99193a74d28a585a3c167a93bf41a255135529c913bd9b1e66601ddaa1ab86de5c153993414f34ba97e9674c459dfadde112b89eeeafa0e5a204c"`
Location: zebra-state/src/service/finalized_state/tests/vectors.rs:41

...
```

Orchard:

```
$ cargo test orchard_note_commitment_tree_serialization -- --nocapture
...

     Running unittests src/lib.rs (target/debug/deps/zebra_state-a7a295af75b2f260)

running 1 test

  left: `"010200000000000000ee9488053a30c596b43014105d3477e6f578c89240d1d1ee1743b77bb6adc40a0001a34b69a4e4d9ccf954d46e5da1004d361a5497f511aeb4d481d23c0be1778133a0be6dab19bc2c65d8299258c16e14d48ec4d4959568c6412aa85763c222a702"`,
 right: `"010200ee9488053a30c596b43014105d3477e6f578c89240d1d1ee1743b77bb6adc40a01a34b69a4e4d9ccf954d46e5da1004d361a5497f511aeb4d481d23c0be177813301a0be6dab19bc2c65d8299258c16e14d48ec4d4959568c6412aa85763c222a702"`
Location: zebra-state/src/service/finalized_state/tests/vectors.rs:90
...
```

Shows that the serialization is a bit longer. This seems to be because the `Position` is now a `u64` field.

Will post more details as i discover them.

zebra-chain/src/orchard/tree.rs
```diff
@@ -302,6 +289,51 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+impl ZcashSerialize for Frontier<Node, MERKLE_DEPTH> {
+    fn zcash_serialize<W: io::Write>(&self, mut writer: W) -> Result<(), io::Error> {
+        //
+        merkle_tree::write_frontier_v1(&mut writer, self)
```
teor2345: You can use:
```rust
let mut data = Vec::new();
merkle_tree::write_frontier_v1(&mut data, tree);
```

Because `Vec` implements `Write`:
https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Write-for-Vec%3Cu8,+A%3E

zebra-chain/src/orchard/tree.rs
```diff
@@ -265,7 +252,7 @@ pub enum NoteCommitmentTreeError {
 }
 
 /// Orchard Incremental Note Commitment Tree
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug)]
 pub struct NoteCommitmentTree {
```
teor2345: Two options here:
- make a `SerializedNoteCommitmentTree` struct that has the same field order and similar types to the previous `bridgetree` crate, then derive `Serialize, Deserialize` on it ourselves
- manually implement the same format as bincode serialization

Here's the bincode format we need to match:
https://docs.rs/bincode/latest/bincode/index.html

mpguerra:
Do we need 2 reviewers here?

teor2345:
> Do we need 2 reviewers here?

Yep, we decided in the meeting that Conrado and I are going to review.

Do you want to add a label for that, so Mergify checks for us?
Otherwise it will merge after the first approval.

teor2345:
PR #7158 will make sure we do multiple reviews before merging.

zebra-chain/src/orchard/tree.rs
teor2345: This comment is outdated now.

zebra-chain/src/sapling/tree.rs
```diff
@@ -498,10 +487,11 @@ impl From<&NoteCommitmentTree> for SerializedTree {
         //
         // [1]: <https://github.com/zcash/librustzcash/blob/a63a37a/zcash_primitives/src/merkle_tree.rs#L125>
         if let Some(frontier) = tree.inner.value() {
-            let (left_leaf, right_leaf) = match frontier.leaf() {
-                Leaf::Left(left_value) => (Some(left_value), None),
-                Leaf::Right(left_value, right_value) => (Some(left_value), Some(right_value)),
-            };
```
oxarbitrage: We can't do any of these anymore. I think we will just have to go with `write_frontier_v1` just as orchard do and delete the rest of the code. I am unsure if the doc comments above make sense here or in orchard so we might want to remove them too.

We could also remove `SerializedTree` fully but i don't think we should do that in this PR.

teor2345: The JSON-RPC format used by `zcashd` is `zcash_primitives::merkle_tree::write_commitment_tree()`, so I think we have to use that:
https://docs.rs/zcash_primitives/0.12.0/zcash_primitives/merkle_tree/fn.write_commitment_tree.html

I don't know why we implemented the serialization ourselves, maybe that method didn't exist, or we didn't know about it.

Here's how I found that:

Orchard tree RPC:
https://github.com/zcash/zcash/blob/4c32e3b3e3e0f576cf0c37aa260a5a9d1619ec3b/src/rpc/blockchain.cpp#L1432

OrchardMerkleFrontierLegacySer type:
https://github.com/zcash/zcash/blob/4c32e3b3e3e0f576cf0c37aa260a5a9d1619ec3b/src/zcash/IncrementalMerkleTree.hpp#L423C29-L423C45

serialize_legacy method:
https://github.com/zcash/zcash/blob/4c32e3b3e3e0f576cf0c37aa260a5a9d1619ec3b/src/rust/src/merkle_frontier.rs#L47

write_commitment_tree_method:
https://github.com/zcash/librustzcash/blob/4d5dc28ab18a020f46a87c1b76d0bf7883ae55c7/zcash_primitives/src/merkle_tree.rs#L182

teor2345: In zcash_primitives 0.11 `write_commitment_tree()` is `CommitmentTree::write()`:
https://docs.rs/zcash_primitives/0.11.0/zcash_primitives/merkle_tree/struct.CommitmentTree.html#method.write

~~I can make this change in a separate PR if it's easier?~~

teor2345: I think we have to do the upgrade first, then use `CommitmentTree::from_parts()` to replace our custom Sapling code:
https://docs.rs/incrementalmerkletree/0.4.0/src/incrementalmerkletree/frontier.rs.html#321

zebra-chain/src/orchard/tree/legacy.rs
```diff
@@ -0,0 +1,102 @@
...
+    Left(A),
+    Right(A, A),
+}
+
+impl<A> LegacyLeaf<A> {
+    ///
+    pub fn value(&self) -> &A {
+        match self {
+            LegacyLeaf::Left(a) => a,
+            LegacyLeaf::Right(_, a) => a,
```
teor2345: Based on @conradoplg's comments on Slack, I think we should be using both parts of the rightmost root?

zebra-state/src/service/finalized_state/tests/vectors.rs
```diff
@@ -73,21 +80,8 @@ fn sprout_note_commitment_tree_serialization() {
     // The purpose of this test is to make sure the serialization format does
```
teor2345, line 60: Just turning this into a checklist, so I can make sure the tests cover both serialization formats:
(Or we can work out which tests aren't actually needed.)

Currently, these tests check these structs are equal:
- [x] commitments -> tree struct
- [x] commitments -> tree struct -> serialize -> deserialize -> tree struct

And these serialized formats are equal:
- [x] fixed serialized test vector
- [x] commitments -> tree struct -> serialize
- [x] commitments -> tree struct -> serialize -> deserialize -> tree struct -> serialize

TODO:
- [x] apply these tests to the new tree structs, and update the serialization format
- [ ] do we need to test the old format as well, or do the tests below cover it?

TODO: test that old and new serializations produce the same format:
Tree roots built from the same commitments should match:
- [ ] commitments -> old tree struct -> new tree struct -> un-cached root
- [x] commitments -> new tree struct -> un-cached root

Even when serialized and deserialized:
- [ ] commitments -> old tree struct -> old serialize -> old deserialize -> old tree struct ->  new tree struct -> un-cached root
- [x] commitments -> new tree struct -> new serialize -> new deserialize -> new tree struct -> un-cached root
- [x] commitments -> new tree struct -> un-cached root

teor2345: I think we're missing these tests:
- using the old tree struct with the same commitments produces the same root
- using the old serialization format to create a new tree struct produces the same root

We can test this two different ways:
Keep the old serialization code, and do these tests:
- commitments -> old tree struct -> new tree struct -> un-cached root
- commitments -> old tree struct -> old serialize -> old deserialize -> old tree struct -> new tree struct -> un-cached root
- (existing new root)

Keep the old serialzation format and root as fixed test vectors, and do these tests:
- old serialization format -> new tree struct -> un-cached root
- old root fixed test vector
- (existing new root)

@oxarbitrage which way did you want to go?

zebra-chain/src/orchard/tree/legacy.rs
oxarbitrage: That implementation is a copy of https://github.com/zcash/incrementalmerkletree/blob/incrementalmerkletree-v0.3.1/src/bridgetree.rs#L21

conradoplg: This can be correct depending on how it's used, if it matches `Right` then additional logic is required in the caller.

zebra-chain/src/orchard/tree/legacy.rs
```diff
@@ -0,0 +1,102 @@
...
+    frontier: Option<LegacyNonEmptyFrontier<H>>,
+}
+
+impl From<LegacyFrontier<Node, MERKLE_DEPTH>> for Frontier<Node, MERKLE_DEPTH> {
+    fn from(legacy_frontier: LegacyFrontier<Node, MERKLE_DEPTH>) -> Self {
+        if let Some(legacy_frontier_data) = legacy_frontier.frontier {
+            Frontier::from_parts(
+                Position::from(legacy_frontier_data.position.0 as u64),
+                *legacy_frontier_data.leaf.value(),
+                legacy_frontier_data.ommers,
```
conradoplg: You will need to also extend `ommers` with `insert(0, left)` if `leaf` is `Right(left,right)`

zebra-chain/src/orchard/tree/legacy.rs
```diff
@@ -0,0 +1,102 @@
...
+            Frontier::empty()
+        }
+    }
+}
+
+impl From<Frontier<Node, MERKLE_DEPTH>> for LegacyFrontier<Node, MERKLE_DEPTH> {
+    fn from(frontier: Frontier<Node, MERKLE_DEPTH>) -> Self {
+        if let Some(frontier_data) = frontier.value() {
+            // We don't have information to build the `Right` side of the `LegacyLeaf` so we just put what we have into the left.
+            let leaf = LegacyLeaf::Left(*frontier_data.leaf());
```
conradoplg: Here, if `frontier.position().is_odd()`, then you need to remove the first element from `ommers()` and use that as the left side, and use the leaf as the right side.

oxarbitrage: Got it, thanks! Trying ...

oxarbitrage: It seems there is something else also needed related to the position that need to change. I am not totally sure how it is yet.

conradoplg: I think you're passing the original `ommers` (with `frontier_data.ommers().to_vec()`) when instantiating `LegacyNonEmptyFrontier` below, instead of the possibly-changed `ommers`

oxarbitrage: D'oh. That's right. I have all this working now. Thanks for the help!

zebra-chain/src/sapling/tree.rs
oxarbitrage: This code was converted to what it was by converting the `NoteCommitmentTree` into a `LegacyNoteCommitmentTree` in https://github.com/ZcashFoundation/zebra/pull/7053/commits/f07c5716e52c7e47e21651971a9791512a76802f

Also for orchard the old code was added back (always with some minor mods to make it work) in https://github.com/ZcashFoundation/zebra/pull/7053/commits/7962f315163fea30d7fef26c62e91454ce44b62d

I don't think we should try to use new functions or way of doing it better in this PR and just stick with what we know it was working.

zebra-state/src/service/finalized_state/tests/vectors.rs
oxarbitrage: I don't think we will need so many tests now that we are keeping the old serialization.

teor2345: Yep, if we're only using one format, then it's enough to check that the new struct roots match the old struct roots.

Something like `assert_eq!(tree.recalculate_root(), tree.cached_root().expect("cached root was serialized"));`

oxarbitrage: Added those at https://github.com/ZcashFoundation/zebra/pull/7053/commits/0b494db159eacd7b388f268110404c7997031999

deny.toml
```diff
@@ -82,6 +82,33 @@ skip-tree = [
...
+    { name = "zcash_address", version = "=0.2.1" },
+
+    # wait for zebra-script to upgrade
+    { name = "zcash_note_encryption", version = "=0.3.0" },
+
+    # wait for zebra-script to upgrade
+    { name = "zcash_primitives", version = "=0.11.0" },
+
+    # wait for zebra-script to upgrade
+    { name = "zcash_proofs", version = "=0.11.0" },
```
teor2345: Upgrading to `zcash_script` 1.1.13 is on the list of upgrades in the ticket, it should fix these duplicate dependencies.

It should also let us fix #6638 from the ticket, which is about removing some duplicate dependencies from this file. Let me know if you'd like help with that.

(After upgrading `zcash_script`, the warnings in the `CI OSes / Check deny.toml bans --all-features`  job will give you a list you can delete from this file.)

zebra-chain/src/orchard/tree.rs
```diff
@@ -498,8 +492,9 @@ impl From<&NoteCommitmentTree> for SerializedTree {
         // Convert the note commitment tree from
         // [`Frontier`](bridgetree::Frontier) to
         // [`CommitmentTree`](merkle_tree::CommitmentTree).
-        let tree = CommitmentTree::from_frontier(&tree.inner);
-        tree.write(&mut serialized_tree)
+        let tree = incrementalmerkletree::frontier::CommitmentTree::from_frontier(&tree.inner);
+
+        write_commitment_tree(tree, &mut serialized_tree)
```
teor2345: This looks like we could just call the function from `zcash_primitives`?

```suggestion
         zcash_primitives::merkle_tree::write_commitment_tree(tree, &mut serialized_tree)
```

zebra-chain/src/orchard/tree/legacy.rs
```diff
@@ -0,0 +1,107 @@
+//! Orchard serialization legacy code.
+
+use incrementalmerkletree::{frontier::Frontier, Position};
+
+use super::{Node, NoteCommitmentTree, Root, MERKLE_DEPTH};
+
+///
```
teor2345: Nitpick: Missing doc comment - please add doc comments on all 3 shielded protocols in the PR, they can be very short.

Did you want me to help write the comments?

zebra-chain/src/orchard/tree/legacy.rs
```diff
@@ -0,0 +1,107 @@
+//! Orchard serialization legacy code.
```
teor2345, line 1: It looks like some of the code in this file has been copied from `incrementalmerkletree` or `zcash_primitives`, can you link to where the code was copied from?

This is part of the legal requirements for the source code license. It needs to be done for the 3 copies in this PR. Did you want me to help write these?

Here's an example:
https://github.com/ZcashFoundation/zebra/blob/be5cfad07f0e401de84f19ec9b418684a93afac2/zebra-network/src/peer_set/initialize.rs#L3-L5

zebra-state/src/service/finalized_state/tests/vectors.rs
```diff
@@ -496,19 +379,156 @@ fn orchard_note_commitment_tree_serialization_pow2() {
...
+    );
+    assert_eq!(
+        incremental_tree.recalculate_root(),
+        deserialized_legacy_tree_as_new.recalculate_root()
+    );
+
+    // Check reclaculated and cached roots are the same
+    assert_eq!(
+        incremental_tree.recalculate_root(),
+        incremental_tree
```
teor2345: We actually need to check the cached root from the serialized format, the one that we construct in the test is guaranteed to be the same.
```suggestion
        deserialised_tree
```

zebra-state/src/service/finalized_state/tests/vectors.rs
```diff
@@ -496,19 +379,156 @@ fn orchard_note_commitment_tree_serialization_pow2() {
...
```
teor2345:
```suggestion
        deserialised_tree
```

zebra-state/src/service/finalized_state/tests/vectors.rs
```diff
@@ -496,19 +379,156 @@ fn orchard_note_commitment_tree_serialization_pow2() {
...
```
teor2345:
```suggestion
        deserialised_tree
```

teor2345 (dismissed):
Looks great, thanks for sticking with this!

I think the `zcash_script` upgrade accidentally got skipped, but it will be easier to do it at the same time. We also need to say where we've copied code from to satisfy the code license.

It would be nice to test the serialized root rather than the one we've just generated, but we might have test coverage for that already.

zebra-state/src/service/finalized_state/tests/vectors.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/7053/commits/4d05ead48c2505551405d354c05c85ad18868f71

zebra-chain/src/orchard/tree.rs
oxarbitrage: You are right, thanks. Done at https://github.com/ZcashFoundation/zebra/pull/7053/commits/ed57b42254dc6b2547cb6ccada2e6c332a0e575a

deny.toml
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/7053/commits/953f32f6461a8c9ba45503a38157f2b689c8609e and https://github.com/ZcashFoundation/zebra/pull/7053/commits/8b6649605f173a594d2637d9807e7215956d57a0

conradoplg (dismissed):
Looks great, thanks!

deny.toml
```diff
@@ -82,6 +82,12 @@ skip-tree = [
     # wait for metrics-exporter-prometheus to upgrade
     { name = "hashbrown", version = "=0.13.2" },
 
+    # wait for zebra-chain to upgrade
```
teor2345, line 80: zebra-chain is in this repository, so we can upgrade it in this PR. We want to keep this dependency in sync with the ECC version so it stays consensus-compatible:
https://github.com/ZcashFoundation/zebra/blob/f1ee502ddf2bf8ffcdb60dbabdc91885f3d723c9/zebra-chain/Cargo.toml#L67-L68

deny.toml
```diff
@@ -70,26 +70,24 @@ skip-tree = [
 
     # wait for zcashd and zcash_script to upgrade
     # https://github.com/ZcashFoundation/zcash_script/pulls
-    { name = "metrics", version = "=0.20.1" },
     { name = "sha2", version = "=0.9.9" },
```
teor2345, line 73: The only remaining dependency on this older `sha2` version is in `zebra-chain`, so we can upgrade it in this PR and remove this exception. This is part of #6638.

deny.toml
```diff
@@ -70,26 +70,24 @@ skip-tree = [
...
+    # wait for zebra-chain to upgrade
+    { name = "secp256k1", version = "=0.21.3" },
+
+    # wait for zebra-chain to upgrade `secp256k1`
+    { name = "secp256k1-sys", version = "=0.4.2" },
+
     # ECC crates
 
     # wait for zcash_primitives to remove duplicated dependencies
     { name = "block-buffer", version = "=0.9.0" },
```
teor2345, line 89: `block-buffer` can also be removed once sha2 is upgraded

mpguerra:
Let's try to wrap up this PR so we can unblock the following work. If there is any other work that would be nice to have, let's create separate PRs for it

deny.toml
teor2345: Moved to ticket #6638

deny.toml
teor2345: Moved to ticket #6638

deny.toml
teor2345: Moved to ticket #6638

teor2345 (approved):
I moved the remaining dependency updates to ticket #6638, which is now a release blocker

PR #7061, opened 2023-06-25
docs(state): Use different terms for block verification and state queues

### LINKED ISSUE #6681: Update documentation to clarify terminology around block verification

oxarbitrage:
## Motivation

Close https://github.com/ZcashFoundation/zebra/issues/6681

## Solution

Do suggested renames.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -494,8 +494,8 @@ pub enum Request {
     ///
     /// # Note
     ///
-    /// `SemanticallyVerifiedBlock`, `ContextuallyVerifiedBlock` and
-    /// `CheckpointVerifiedBlock` are an internal Zebra implementation detail.
+    /// [`SemanticallyVerifiedBlock`], [`ContextuallyVerifiedBlock`] and
+    /// [`CheckpointVerifiedBlock`] are an internal Zebra implementation detail.
     /// There is no difference between these blocks on the Zcash network, or in Zebra's
     /// network or syncer implementations.
     ///
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -163,7 +163,7 @@ pub struct SemanticallyVerifiedBlock {
 }
 
 /// A block ready to be committed directly to the finalized state with
-/// no checks.
+/// a small number of checks if compared with a `ContextuallyVerifiedBlock`.
 ///
 /// This is exposed for use in checkpointing.
 ///
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -455,12 +455,11 @@ impl DerefMut for CheckpointVerifiedBlock {
 /// A query about or modification to the chain state, via the
 /// [`StateService`](crate::service::StateService).
 pub enum Request {
-    /// Performs contextual validation of the given block, committing it to the
-    /// state if successful.
+    /// Performs contextual validation of the given semantically verified block,
+    /// committing it to the state if successful.
     ///
-    /// It is the caller's responsibility to perform semantic validation. This
-    /// request can be made out-of-order; the state service will queue it until
-    /// its parent is ready.
+    /// This request can be made out-of-order; the state service will queue it
+    /// until its parent is ready.
     ///
     /// Returns [`Response::Committed`] with the hash of the block when it is
     /// committed to the state, or an error if the block fails contextual
@@ -478,12 +477,12 @@ pub enum Request {
     /// documentation for details.
     CommitSemanticallyVerifiedBlock(SemanticallyVerifiedBlock),
 
-    /// Commit a checkpointed block to the state, skipping most block validation.
+    /// Commit a checkpointed block to the state, skipping most but not all
+    /// contextual validation.
     ///
-    /// This is exposed for use in checkpointing, which produces finalized
-    /// blocks. It is the caller's responsibility to ensure that the block is
-    /// semantically valid and final. This request can be made out-of-order;
-    /// the state service will queue it until its parent is ready.
+    /// This is exposed for use in checkpointing, which produces checkpoint vefified
+    /// blocks. This request can be made out-of-order; the state service will queue
+    /// it until its parent is ready.
     ///
     /// Returns [`Response::Committed`] with the hash of the newly committed
     /// block, or an error.
@@ -495,8 +494,9 @@ pub enum Request {
     ///
     /// # Note
     ///
-    /// Finalized and non-finalized blocks are an internal Zebra implementation detail.
-    /// There is no difference between these blocks on the network, or in Zebra's
+    /// `SemanticallyVerifiedBlock`, `ContextuallyVerifiedBlock` and
+    /// `CheckpointVerifiedBlock` are an internal Zebra implementation detail.
+    /// There is no difference between these blocks on the Zcash network, or in Zebra's
     /// network or syncer implementations.
     ///
     /// # Consensus
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -970,11 +970,16 @@ impl Service<Request> for StateService {
             Request::CommitCheckpointVerifiedBlock(finalized) => {
                 // # Consensus
                 //
-                // A checkpoint block verification could have called AwaitUtxo
-                // before this finalized block arrived in the state.
-                // So we need to check for pending UTXOs here for non-finalized blocks,
-                // even though it is redundant for most finalized blocks.
-                // (Finalized blocks are verified using block hash checkpoints
+                // A semantic block verification could have called AwaitUtxo
+                // before this checkpoint verified block arrived in the state.
+                // So we need to check for pending UTXO requests sent by running
+                // semantic block verifications.
+                //
+                // This check is redundant for most checkpoint verified blocks,
+                // because semantic verification can only succeed near the final 
+                // checkpoint, when all the UTXOs are available for the verifying block.
+                //
+                // (Checkpoint block UTXOs are verified using block hash checkpoints
                 // and transaction merkle tree block header commitments.)
                 self.pending_utxos
                     .check_against_ordered(&finalized.new_outputs);
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -455,7 +455,7 @@ impl StateService {
         (state, read_service, latest_chain_tip, chain_tip_change)
     }
 
-    /// Queue a finalized block for verification and storage in the finalized state.
+    /// Queue a checkpoint verified block for verification and storage in the finalized state.
     ///
     /// Returns a channel receiver that provides the result of the block commit.
     fn queue_and_commit_to_finalized_state(
@@ -481,32 +481,32 @@ impl StateService {
         let queued = (checkpoint_verified, rsp_tx);
 
         if self.finalized_block_write_sender.is_some() {
-            // We're still committing finalized blocks
+            // We're still committing checkpoint verified blocks
             if let Some(duplicate_queued) = self
                 .finalized_state_queued_blocks
                 .insert(queued_prev_hash, queued)
             {
                 Self::send_checkpoint_verified_block_error(
                     duplicate_queued,
-                    "dropping older finalized block: got newer duplicate block",
+                    "dropping older checkpoint verified block: got newer duplicate block",
                 );
             }
 
             self.drain_finalized_queue_and_commit();
         } else {
-            // We've finished committing finalized blocks, so drop any repeated queued blocks,
-            // and return an error.
+            // We've finished committing checkpoint verified blocks to finalized queue,
+            // so drop any repeated queued blocks, and return an error.
             //
             // TODO: track the latest sent height, and drop any blocks under that height
             //       every time we send some blocks (like QueuedSemanticallyVerifiedBlocks)
             Self::send_checkpoint_verified_block_error(
                 queued,
-                "already finished committing finalized blocks: dropped duplicate block, \
+                "already finished committing checkpoint verified blocks to finalized queue: dropped duplicate block, \
                  block is already committed to the state",
             );
 
             self.clear_finalized_block_queue(
-                "already finished committing finalized blocks: dropped duplicate block, \
+                "already finished committing checkpont verified blocks to finalized queue: dropped duplicate block, \
                  block is already committed to the state",
             );
         }
@@ -694,8 +694,8 @@ impl StateService {
             rsp_rx
         };
 
-        // We've finished sending finalized blocks when:
-        // - we've sent the finalized block for the last checkpoint, and
+        // We've finished sending semantically verrified blocks when:
+        // - we've sent the semantically verrified block for the last checkpoint, and
         // - it has been successfully written to disk.
         //
         // We detect the last checkpoint by looking for non-finalized blocks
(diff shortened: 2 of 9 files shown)


oxarbitrage:
This PR should be rebased to main after https://github.com/ZcashFoundation/zebra/pull/7035 is merged. 

I forgot how to set up an automatic rebase. I will add that when i find the feature.

zebra-state/src/request.rs
```diff
@@ -495,8 +494,9 @@ pub enum Request {
     ///
     /// # Note
     ///
-    /// Finalized and non-finalized blocks are an internal Zebra implementation detail.
-    /// There is no difference between these blocks on the network, or in Zebra's
+    /// `SemanticallyVerifiedBlock`, `ContextuallyVerifiedBlock` and
+    /// `CheckpointVerifiedBlock` are an internal Zebra implementation detail.
```
teor2345: If we use doc links here, `rustdoc` will tell us if we have typos or missed renames:
```suggestion
    /// [`SemanticallyVerifiedBlock`], [`ContextuallyVerifiedBlock`] and
    /// [`CheckpointVerifiedBlock`] are an internal Zebra implementation detail.
```

zebra-state/src/service.rs
```diff
@@ -156,7 +156,7 @@ pub(crate) struct StateService {
     /// - the hash of the last valid committed block (the parent of the invalid block).
     //
     // TODO:
```
teor2345: I think this TODO was implemented as `SentHashes`, which is just below this field.

```suggestion
```

zebra-state/src/service.rs
```diff
@@ -156,7 +156,7 @@ pub(crate) struct StateService {
...
-    // - turn this into an IndexMap containing recent non-finalized block hashes and heights
+    // - turn this into an IndexMap containing recent semantically verified block hashes and heights
     //   (they are all potential tips)
     // - remove block hashes once their heights are strictly less than the finalized tip
```
```suggestion
```

zebra-state/src/service.rs
```diff
@@ -471,7 +471,7 @@ impl StateService {
         let queued_height = checkpoint_verified.height;
 
         // If we're close to the final checkpoint, make the block's UTXOs available for
-        // full verification of non-finalized blocks, even when it is in the channel.
+        // full verification of checkpoint blocks, even when it is in the channel.
```
teor2345: This is about taking the UTXOs of checkpoint verified blocks, and making them available for semantic verification:

```suggestion
        // semantic block verification, even when it is in the channel.
```

zebra-state/src/service.rs
```diff
@@ -481,32 +481,32 @@ impl StateService {
...
+                    "dropping older checkpoint verified block: got newer duplicate block",
                 );
             }
 
             self.drain_finalized_queue_and_commit();
         } else {
-            // We've finished committing finalized blocks, so drop any repeated queued blocks,
-            // and return an error.
+            // We've finished committing checkpoint verified blocks to finalized queue,
+            // so drop any repeated queued blocks, and return an error.
```
teor2345: When we've finished with checkpoint verified blocks, we're also finished with the queue. (Blocks get queued in the queue then stored or committed to the state.)

```suggestion
            // We've finished committing checkpoint verified blocks to the finalized state,
            // so drop any repeated queued blocks, and return an error.
```

zebra-state/src/service.rs
```diff
@@ -481,32 +481,32 @@ impl StateService {
...
             //
             // TODO: track the latest sent height, and drop any blocks under that height
             //       every time we send some blocks (like QueuedSemanticallyVerifiedBlocks)
             Self::send_checkpoint_verified_block_error(
                 queued,
-                "already finished committing finalized blocks: dropped duplicate block, \
+                "already finished committing checkpoint verified blocks to finalized queue: dropped duplicate block, \
                  block is already committed to the state",
```
```suggestion
                "already finished committing checkpoint verified blocks: dropped duplicate block, \
                 block is already committed to the state",
```

zebra-state/src/service.rs
```diff
@@ -481,32 +481,32 @@ impl StateService {
...
             );
 
             self.clear_finalized_block_queue(
-                "already finished committing finalized blocks: dropped duplicate block, \
+                "already finished committing checkpont verified blocks to finalized queue: dropped duplicate block, \
                  block is already committed to the state",
```
```suggestion
                "already finished committing checkpoint verified blocks: dropped duplicate block, \
                 block is already committed to the state",
```

zebra-state/src/service.rs
```diff
@@ -694,8 +694,8 @@ impl StateService {
             rsp_rx
         };
 
-        // We've finished sending finalized blocks when:
-        // - we've sent the finalized block for the last checkpoint, and
+        // We've finished sending semantically verrified blocks when:
+        // - we've sent the semantically verrified block for the last checkpoint, and
```
teor2345: This comment is about ending checkpoint verification and starting semantic verification, it describes the checkpoint end condition.

```suggestion
        // We've finished sending checkpoint verified blocks when:
        // - we've sent the verified block for the last checkpoint, and
```

zebra-state/src/service.rs
```diff
@@ -709,13 +709,13 @@ impl StateService {
             && self.read_service.db.finalized_tip_hash()
                 == self.finalized_block_write_last_sent_hash
         {
-            // Tell the block write task to stop committing finalized blocks,
-            // and move on to committing non-finalized blocks.
+            // Tell the block write task to stop committing semantically verified blocks to the finalized queue,
+            // and move on to committing to non-finalized blocks.
```
```suggestion
            // Tell the block write task to stop committing checkpoint verified blocks to the finalized state,
            // and move on to committing semantically verified blocks to the non-finalized state.
```

zebra-state/src/service.rs
```diff
@@ -709,13 +709,13 @@ impl StateService {
...
             std::mem::drop(self.finalized_block_write_sender.take());
 
-            // We've finished committing finalized blocks, so drop any repeated queued blocks.
+            // We've finished committing semantically verified blocks to finalized queue, so drop any repeated queued blocks.
```
```suggestion
            // We've finished committing checkpoint verified blocks to finalized state, so drop any repeated queued blocks.
```

zebra-state/src/service.rs
```diff
@@ -709,13 +709,13 @@ impl StateService {
...
             self.clear_finalized_block_queue(
-                "already finished committing finalized blocks: dropped duplicate block, \
+                "already finished committing semantically verified blocks to finalized queue: dropped duplicate block, \
                  block is already committed to the state",
```
```suggestion
                "already finished committing checkpoint verified blocks: dropped duplicate block, \
                 block is already committed to the state",
```

zebra-state/src/service.rs
```diff
@@ -754,7 +754,7 @@ impl StateService {
 
     /// Returns `true` if `queued_height` is near the final checkpoint.
     ///
-    /// The non-finalized block verifier needs access to UTXOs from finalized blocks
+    /// The checkpoint block verifier needs access to UTXOs from finalized blocks
     /// near the final checkpoint, so that it can verify blocks that spend those UTXOs.
```
```suggestion
    /// The semantic block verifier needs access to UTXOs from checkpoint verified blocks
    /// near the final checkpoint, so that it can verify blocks that spend those UTXOs.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -15,13 +15,13 @@ use crate::{BoxError, CheckpointVerifiedBlock, SemanticallyVerifiedBlock};
 #[cfg(test)]
 mod tests;
 
-/// A finalized state queue block, and its corresponding [`Result`] channel.
+/// A checkpoint verified block queue, and its corresponding [`Result`] channel.
```
teor2345: This is just a single block

```suggestion
/// A queued checkpoint verified block, and its corresponding [`Result`] channel.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -15,13 +15,13 @@ use crate::{BoxError, CheckpointVerifiedBlock, SemanticallyVerifiedBlock};
...
 pub type QueuedCheckpointVerified = (
     CheckpointVerifiedBlock,
     oneshot::Sender<Result<block::Hash, BoxError>>,
 );
 
-/// A non-finalized state queue block, and its corresponding [`Result`] channel.
+/// A semantically verified block queue, and its corresponding [`Result`] channel.
```
teor2345:
```suggestion
/// A queued semantically verified block, and its corresponding [`Result`] channel.
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -264,10 +264,10 @@ impl SentHashes {
         self.update_metrics_for_block(block.height);
     }
 
-    /// Stores the finalized `block`'s hash, height, and UTXOs, so they can be used to check if a
+    /// Stores the checkpoint verified `block`'s hash, height, and UTXOs into finalized, so they can be used to check if a
```
teor2345: There's no finalized state here yet, that happens in the block write task

```suggestion
    /// Stores the checkpoint verified `block`'s hash, height, and UTXOs, so they can be used to check if a
```

zebra-state/src/service/queued_blocks.rs
```diff
@@ -264,10 +264,10 @@ impl SentHashes {
...
     /// block or UTXO is available in the state.
     ///
-    /// Used for finalized blocks close to the final checkpoint, so non-finalized blocks can look up
+    /// Used for checkpoint verified blocks close to the final checkpoint, so state non-finalized blocks can look up
```
teor2345:
```suggestion
    /// Used for checkpoint verified blocks close to the final checkpoint, so the semantic block verifier can look up
```

zebra-state/src/service.rs
```diff
@@ -970,7 +970,7 @@ impl Service<Request> for StateService {
             Request::CommitCheckpointVerifiedBlock(finalized) => {
                 // # Consensus
```
teor2345, line 966: This comment needs a rewrite to remove all instances of "finalized block" and "non-finalised block", so I'm just going to push a commit

teor2345 (dismissed):
Looks good, thanks for doing this work! I really appreciate it.

The old comments are very confusing, so I made some suggestions where we could make them clearer. I'm also going to ask @arya2 to review as well, because he's more familiar with some parts of this code.

zebra-state/src/service/queued_blocks.rs
teor2345: The UTXOs are being cached in this data structure when blocks are queued, before we do contextual validation on them.

zebra-state/src/service.rs
teor2345: See my commit 4a9b841f2d14cdd12362822da679e26b771edd56

zebra-state/src/service.rs
teor2345: (This is the comment that the auditors needed to have clarified!)

zebra-state/src/service.rs
```diff
@@ -970,11 +970,16 @@ impl Service<Request> for StateService {
...
-                // So we need to check for pending UTXOs here for non-finalized blocks,
-                // even though it is redundant for most finalized blocks.
-                // (Finalized blocks are verified using block hash checkpoints
+                // A semantic block verification could have called AwaitUtxo
+                // before this checkpoint verified block arrived in the state.
+                // So we need to check for pending UTXO requests sent by running
+                // semantic block verifications.
+                //
+                // This check is redundant for most checkpoint verified blocks,
+                // because semantic verification can only succeed near the final
```
```suggestion
                // because semantic verification can only succeed near the final
```

zebra-state/src/service.rs
arya2: We should be able to get rid of this field as part of https://github.com/ZcashFoundation/zebra/issues/6388.

zebra-state/src/service.rs
arya2: We may need to update these checks in `should_restart_sync()`:

https://github.com/ZcashFoundation/zebra/blob/issue6681/zebrad/src/components/sync.rs#L1137
https://github.com/ZcashFoundation/zebra/blob/issue6681/zebrad/src/components/sync.rs#L1171

teor2345: Good catch, I've removed the change to that line, this is meant to be a documentation-only PR.

zebra-state/src/service/check.rs
```diff
@@ -375,23 +375,23 @@ where
 pub(crate) fn initial_contextual_validity(
```
arya2, line 375: Optional: we may also want to rename `prepared` in `validate_and_commit_non_finalized()` to match this, or it may be enough that it has `non_finalized` in the method name like the methods in `NonFinalizedState`.

Update: This isn't worth blocking the PR, I think it's okay for now.

arya2 (approved):
This looks great, much clearer, thank you!

PR #7063, opened 2023-06-26
fix(cd): Rename a CD job with the same name as a CI job

teor2345:
## Motivation

One of our CD test jobs has the same name as a CI test job, so we could merge a PR if either one of them passes. (We want both to pass.)

## Solution

Rename the CD job

- [x] Admin: add a branch protection rule for the new name after merging this PR

## Review

This should be merged before merging PRT #7045, because it tests that PR.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-delivery.patch.yml
+++ b/.github/workflows/continous-delivery.patch.yml
@@ -30,7 +30,7 @@ jobs:
       - run: 'echo "No build required"'
 
   test-configuration-file:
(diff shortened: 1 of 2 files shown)


gustavovalverde (approved):

teor2345:
> * Admin: add a branch protection rule for the new name after merging this PR

I added a branch protection rule that requires this job to pass on all PRs:
`Test Zebra CD Docker config file`

PR #7072, opened 2023-06-27
fix(state): Replace a chain length assertion with a NotReadyToBeCommitted error

teor2345:
## Motivation

The auditors were concerned about the possibility of a panic in a state verification check.

## Solution

Replace the assertion with a block verification error.

## Review

This is a low-priority audit cleanup.

### Reviewer Checklist

## Follow Up Work

We should change the tests that currently check for success to check for this error instead, and remove the test workaround.


dconnolly (approved):
lgtm!

PR #7096, opened 2023-06-28
fix(docs): Add a few more details to metrics book page

oxarbitrage:
## Motivation

I found that even though the tutorial says in the top this needs the `prometheus` rust feature, this is not part of the steps so can be easily bypassed. Metrics endpoint will not open, no errors are shown.

I think we might want to panic if endpoint is set but no `prometheous` feature is present but that is not part of this PR. 

Here i just update the tutorial a bit to make sure the feature is an step of it. 

Added an image on how it look as well at the end of the tutorial. 

## Solution

Add some more info to the tutorial.

## Review

Anyone can review.

### Reviewer Checklist


arya2 (approved):
This looks good, nice catch!

> I think we might want to panic if endpoint is set but no prometheous feature is present but that is not part of this PR.

I don't think it's worth panicking for unused config fields, but I suspect the warning would be more visible if logged immediately after loading the config (I added this to #7118).

Issue #6922, opened 2023-06-12
Use correct tx modifications in transaction::tests::v4_with_modified_joinsplit_is_rejected


arya2:
This is still quite fragile, and there's no reason why correct modifications (e.g. to make the signature verification faster) wouldn't start making it fail all the time.

It's up to yous of course, but I would suggest creating transactions for which you know the keys so that you can recreate the signature. An advantage of this is that you can recreate the signature only for v4, and that would effectively test that v5 signatures don't depend on proofs.

Accepting either a proof or signature failure is not the right thing to do: suppose that you do that, then improve signature verification time, and then introduce a regression in the proof checking. The test could silently succeed in that situation. (Just a proof checking regression might be enough to do the same thing if it increased proof checking time.)

_Originally posted by @daira in https://github.com/ZcashFoundation/zebra/issues/6921#issuecomment-1587744961_

conradoplg:
Reopening per https://github.com/ZcashFoundation/zebra/pull/11296#pullrequestreview-4985422664

Issue #6867, opened 2023-06-08
RPC connection closed by Zebra during sending_transactions_using_lightwalletd test


teor2345:
## Motivation

The `sending_transactions_using_lightwalletd` test occasionally fails with a "connection closed" or "connection refused" error:
> {"app":"lightwalletd","error":"Post \"[http://127.0.0.1:39183\](http://127.0.0.1:39183/)": read tcp 127.0.0.1:37352-\u003e127.0.0.1:39183: read: connection reset by peer","level":"fatal","msg":"error zcashd getbestblockhash rpc","time":"2023-06-08T01:52:28Z"}

https://github.com/ZcashFoundation/zebra/actions/runs/5205948564/jobs/9392912256?pr=6842#step:8:384

Zebra just seems to keep on running, without logging any errors.

We could try adding extra logs or extra waits when the `zebrad` or `lightwalletd` processes in these tests fail?
This might be tricky because we're seeing a test panic, rather than a test error.

We could also remove this error from the lightwalletd error list, and see if the test passes. (It probably won't, because `lightwalletd` says it's a fatal error. But it might give us more time to see Zebra's logs.)

### Complex Code or Requirements

This is probably a `zebrad` RPC concurrency bug or a test harness concurrency bug.

### Testing

This is going to be very tricky to reproduce and test.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/6867?planning-poker) with Zenhub @arya2 @dconnolly @oxarbitrage @teor2345 @upbqdn

teor2345:
@mpguerra we've only seen this happen once every few weeks, maybe it should be a low priority? (Feel free to switch it back.)

mpguerra:
that's fine, we should use the opportunity that we'll be working on lightwalletd RPC methods to try to fix this if it comes up. Otherwise I will un-schedule if it's not finished by the time we're done with our RPC changes.

teor2345:
This failed again in the merge queue in PR #6992:
https://github.com/ZcashFoundation/zebra/actions/runs/5305905415/jobs/9603454438?pr=6992#step:8:547

teor2345:
We haven't seen this happen since PR #7103 was merged, so I'm going to assume it was fixed by that PR.

teor2345:
[...]

PR #6235, opened 2023-02-27
feat(ui): Add a terminal-based progress bar to Zebra

teor2345:
## Motivation

Some users just want to see how Zebra's sync is progressing, without detailed logs.

You can see the progress bar by compiling Zebra with the `progress-bar` feature:
```sh
cargo run --features progress-bar --bin zebrad
```

This automatically redirects your logs to a file in the standard log path.

Closes #6245

### Complex Code or Requirements

There are some shutdown delays, but they seem to be about the same as without the progress bar.

## Solution

Terminal Display:
- Add a progress-bar feature to zebrad
- Add the progress bar writer to the tracing component
    - Redirect the logs to a file when the progress bar is on
    - Create missing directories for the log file
- Shut down the progress bar when there is an interrupt like Ctrl-C

Block Progress:
- Add a block progress bar transmitter
- Add checkpoint progress bar transmitters

Network Connections:
- Add inbound and outbound connection count transmitter
- Add address book peer status transmitters

Mempool:
- Add progress bars for transaction queue, transaction count, transaction memory, and rejected transactions

Related changes:
- Add a Display/to_string() implementation for NetworkUpgrade
- Add an optional limit and label to connection counters
- Improve CheckpointList API and CheckpointVerifier debugging
- [Fix progress log formatting](https://github.com/ZcashFoundation/zebra/pull/6235/commits/04fe2e7b4744e2c454b5b71cd7830b30a3639f06)
- [Add file security docs for running Zebra with elevated permissions](https://github.com/ZcashFoundation/zebra/pull/6235/commits/1eecebecf2ff9438e7f699ac7a4b9576462f3969)

## Review

This is a low priority change.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

Write a blog post.
Tweak progress bars display.

PR #7103, opened 2023-06-29
fix(hang): Stop blocking some Zebra futures for up to a minute using a CPU busy-loop, Credit: Ziggurat Team (#6763), james_katz (#7000)

### LINKED ISSUE #6763: Intermittent long delays to inbound peer connection handshakes, Credit: Ziggurat Team

### LINKED ISSUE #7000: Reduce Zebra CPU and memory usage during initial sync
## Motivation

A user has noted that Zebra 1.0.0's CPU usage is high during the initial sync: 
https://forum.zcashcommunity.com/t/zebra-1-0-0-stable-release/44867/17

Another user asked us to track memory usage.

We need to find out what is causing the CPU and memory usage before we can make progress on this ticket.

Ideally we should focus on things that make the sync faster.

### Testing

Run any benchmarks or manual tests after making each change.

Check the full mainnet zebrad + lightwalletd sync time at:
https://github.com/ZcashFoundation/zebra/actions/workflows/continous-integration-docker.yml?query=branch%3Amain+event%3Aschedule

## Related Work

Possible causes:
- #4784 
- #5425
- #4841
- #4672

teor2345:
## Motivation

Since PR #6235 was merged in March 2023, Zebra's progress logging future has been busy-waiting in a loop for at least 55 seconds out of every minute.

Since this loop doesn't call into the `tokio` executor, it never yields, so it blocks all other futures running on that thread.

This is a likely cause for these bugs:
Close #6763
Close #7000 

It might also possibly be causing these bugs, or making them worse:
* #6867
* #5709
* #6922
* #6812

### Specifications

https://docs.rs/tokio/latest/tokio/#cpu-bound-tasks-and-blocking-code

`tokio` uses cooperative multitasking with internal yields, but only `tokio` APIs can yield to `tokio`:
https://docs.rs/tokio/latest/tokio/task/index.html#cooperative-scheduling

## Solution

Sleep before continuing the loop.
Use monotonic `Instant` times, so that changing the computer's clock doesn't cause progress updates to stop.

## Review

This seems like it might be a release blocker, given the number of bugs it impacts.

### Reviewer Checklist

## Follow Up Work

We should ask users to re-test and see if any of these bugs are still happening.

### DIFF
--- a/zebrad/src/components/sync/progress.rs
+++ b/zebrad/src/components/sync/progress.rs
@@ -154,8 +154,8 @@ pub async fn show_block_chain_progress(
                     .desc(network_upgrade.to_string());
             }
 
-            // Skip logging if it isn't time for it yet
-            let elapsed_since_log = last_log_time.saturating_duration_since(instant_now);
+            // Skip logging and status updates if it isn't time for them yet.
+            let elapsed_since_log = instant_now.saturating_duration_since(last_log_time);
             if elapsed_since_log < LOG_INTERVAL {
                 tokio::time::sleep(PROGRESS_BAR_INTERVAL).await;
                 continue;
--- a/zebrad/src/components/sync/progress.rs
+++ b/zebrad/src/components/sync/progress.rs
@@ -144,6 +144,8 @@ pub async fn show_block_chain_progress(
             let network_upgrade = NetworkUpgrade::current(network, current_height);
 
             // Send progress reports for block height
+            //
+            // TODO: split the progress bar height update into its own function.
             #[cfg(feature = "progress-bar")]
             if matches!(howudoin::cancelled(), Some(true)) {
                 block_bar.close();
@@ -162,6 +164,8 @@ pub async fn show_block_chain_progress(
             } else {
                 last_log_time = instant_now;
             }
+
+            // TODO: split logging / status updates into their own function.
 
             // Work out the sync progress towards the estimated tip.
             let sync_progress = f64::from(current_height.0) / f64::from(estimated_height.0);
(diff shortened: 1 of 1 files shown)


teor2345:
@mpguerra is it ok if we get this merged before the release?

It would finish off all our current network bugs, and get Zebra ready for re-testing.

conradoplg (dismissed):
Good catch! I monitored CPU usage before & after and I think it shows the difference, before there was a core at 100% most of the time

Screenshot from 2023-06-29 11-08-35

Screenshot from 2023-06-29 11-06-50

teor2345:
> Good catch! I monitored CPU usage before & after and I think it shows the difference, before there was a core at 100% most of the time

Thanks, I was going to check this!

teor2345:
I got the elapsed time calculation wrong, the tests should be fixed now.

teor2345:
I can confirm, the only CPU heavy threads left on my machine are my `zcashd` testnet miner. All my Zebra instances are at 50% CPU or less.

Screenshot 2023-06-30 at 10 15 13

conradoplg (approved):

PR #7104, opened 2023-06-29
fix(panic): Stop panicking on shutdown in the syncer and network init

teor2345:
## Motivation

I saw another panic on shutdown in the syncer while preparing for the Zebra showcase.

This is the same kind of bug as the last one: a panic on JoinError::Cancelled, which should just be a log.

## Solution

Log shutdown errors rather than panicking in the syncer and when initialising zebra-network.

## Review

This should be fixed soon, but it's not a release blocker.

### Reviewer Checklist


arya2 (approved):
Looks good.

PR #7155, opened 2023-07-05
change(Docker): Use standard syntax for the `ENV` instruction in Dockerfile

upbqdn:
## Motivation

The documented syntax for the `ENV` instruction is as follows:

```Dockerfile
ENV <key>=<value> ...
```

We've been using:

```Dockerfile
ENV <key> <value> ...
```

The Docker docs say:

> This syntax does not allow for multiple environment-variables to be set in a single ENV instruction, and can be confusing. For example, the following sets a single environment variable (ONE) with value "TWO= THREE=world":

```Dockerfile
ENV ONE TWO= THREE=world
```

> This alternative syntax is supported for backward compatibility, but discouraged for the reasons outlined above, and may be removed in a future release.

## Solution

Use the standard syntax.

### Reviewer Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -64,20 +64,20 @@ RUN if [ "$(uname -m)" != "aarch64" ]; then \
 #
 # We set defaults to all variables.
 ARG RUST_BACKTRACE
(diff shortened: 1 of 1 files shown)


arya2 (approved):
Looks good!

PR #7162, opened 2023-07-06
fix(Docker): Set default `RPC_PORT`

upbqdn:
## Motivation

I noticed we were using an unset `RPC_PORT` in the runtime entry point.

## Solution

The `runtime-entrypoint.sh` uses the `RPC_PORT` env var when the user
specifies the `getblocktemplate-rpc` feature, but this env var is unset
unless the user sets it. This commit sets the default values for
`RPC_PORT` depending on `NETWORK`.

### Reviewer Checklist

### DIFF
--- a/docker/runtime-entrypoint.sh
+++ b/docker/runtime-entrypoint.sh
@@ -31,6 +31,11 @@ fi
 : "${TRACING_ENDPOINT_PORT:=3000}"
 # [rpc]
 : "${RPC_LISTEN_ADDR:=0.0.0.0}"
+if [[ "${NETWORK}" = "Mainnet" ]]; then
+: "${RPC_PORT:=8232}"
+elif [[ "${NETWORK}" = "Testnet" ]]; then
+: "${RPC_PORT:=18232}"
+fi
 
 
 # Populate `zebrad.toml` before starting zebrad, using the environmental


docker/Dockerfile
```diff
@@ -175,11 +175,6 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     ca-certificates
 
-# Config settings
-
-ARG NETWORK
-ENV NETWORK=${NETWORK:-Mainnet}
```
arya2: I think this is needed for building with `--target runtime`.

docker/runtime-entrypoint.sh
```diff
@@ -31,6 +31,11 @@ fi
 : "${TRACING_ENDPOINT_PORT:=3000}"
 # [rpc]
 : "${RPC_LISTEN_ADDR:=0.0.0.0}"
+if [[ "${NETWORK}" = "Mainnet" ]]; then
+: "${RPC_PORT:=8232}"
+elif [[ "${NETWORK}" = "Testnet" ]]; then
+: "${RPC_PORT:=18232}"
+fi
```
arya2, line 38: Optional: it's fairly easy to expose a port when running the image, but it'd be nice to expose `RPC_PORT` in the Dockerfile too.

arya2 (dismissed):
Good catch!

docker/Dockerfile
upbqdn: I think so too. I just force-pushed a branch that omits this change before I read your note. Thanks for spotting it.

docker/runtime-entrypoint.sh
upbqdn: I assume you mean the `EXPOSE` instruction. It doesn't actually do anything unless you use the `-P` (`--publish-all`) flag with `docker run`. I would personally get rid of it completely because when used with the flag, it'll map all the listed ports to random ports on the host, even though Zebra doesn't use them. Also, if we let users pick their port at build time and a potentially different port at run time, the port in `EXPOSE` might differ from the one used by Zebra.

arya2: > I assume you mean the EXPOSE instruction.

Yep.

> It doesn't actually do anything unless you use the -P (--publish-all) flag

Well that explains a lot. Okay.

arya2 (approved):
Thank you!

teor2345 (approved):
> The `runtime-entrypoint.sh` uses the `RPC_PORT` env var when the user
> specifies the `getblocktemplate-rpc` feature

Some users might want to use the Docker image with `lightwalletd`, but without configuring `getblocktemplate-rpcs`. So let's work out how to do that in another ticket.

Previously, the Dockerfile only activated the RPC port when it was set, but I agree that doesn't make sense with `getblocktemplate-rpcs`:
https://github.com/ZcashFoundation/zebra/commit/8c90f653915b1587a2fcc3f227a791a421f832ad#diff-5b838857fc3410c2438196effdbdcffae6548bdba431a4a938e76a0be62d008eL49

So we might want logic like:
* If `getblocktemplate-rpcs` is activated, activate the default RPC port if it is unset
* If there aren't any active RPC features, activate the RPC port only if it is set
* Otherwise, don't activate the RPC port

This covers the 3 different use cases of mining, light wallet, and no RPCs.

PR #7169, opened 2023-07-07
fix(ci): Increase peer cache startup wait time and test time

teor2345:
## Motivation

The peer cache tests sometimes fail because the task doesn't wait long enough for the seed peer DNS queries:
https://github.com/ZcashFoundation/zebra/actions/runs/5481072579/jobs/9984955195?pr=7152#step:15:4516

I've seen this happen two or three times in the past week. 

## Solution

Wait longer for the DNS timeout and initial peer connections, before writing peers to disk
Make the test timeout longer as well

## Review

This is causing CI failures so it would be good to get it fixed.

### Reviewer Checklist

## Follow Up Work

We could wait for the peers to actually be put in the address book instead. But in production it doesn't matter because they just get written to disk the next time the task runs.

### DIFF
--- a/zebra-network/src/peer_cache_updater.rs
+++ b/zebra-network/src/peer_cache_updater.rs
@@ -25,7 +25,7 @@ pub async fn peer_cache_updater(
     //
     // TODO: turn the initial sleep time into a parameter of this function,
     //       and allow it to be set in tests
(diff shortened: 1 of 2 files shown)


arya2 (approved):
Thank you!

PR #7176, opened 2023-07-08
fix(CI): Resize GCP disks from 300 GB to 400 GB

upbqdn:
## Motivation

In PRs #7174 and #7173, there is the following error in the CI logs:

```
  = note: /usr/bin/ld: final link failed: No space left on device
```

## Solution

I assumed the error was due to insufficient disk size in GCP, so I increased it from 300 GB to 400 GB. If the solution is wrong, let's close and ignore this PR.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -159,11 +159,11 @@ jobs:
         id: create-instance
         run: |
           gcloud compute instances create-with-container "${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}" \
-          --boot-disk-size 300GB \
+          --boot-disk-size 400GB \
           --boot-disk-type pd-ssd \
           --image-project=cos-cloud \
           --image-family=cos-stable \
-          --create-disk=name="${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }}",device-name="${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }}",size=300GB,type=pd-ssd \
+          --create-disk=name="${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }}",device-name="${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }}",size=400GB,type=pd-ssd \
           --container-image=gcr.io/google-containers/busybox \
           --machine-type ${{ vars.GCP_LARGE_MACHINE }} \
           --network-interface=subnet=${{ vars.GCP_SUBNETWORK }} \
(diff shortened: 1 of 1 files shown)


teor2345:
Coverage is a GitHub actions job, it doesn't have any GCP steps:
https://github.com/ZcashFoundation/zebra/actions/runs/5488261146/jobs/10013394589?pr=7174#step:3:1

So I think we might need to clear the compiler caches instead:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-disk-full-errors-and-zcash-parameter-errors

I'll just go do that now, and see if it works.

teor2345:
> Coverage is a GitHub actions job, it doesn't have any GCP steps: [ZcashFoundation/zebra/actions/runs/5488261146/jobs/10013394589?pr=7174#step:3:1](https://github.com/ZcashFoundation/zebra/actions/runs/5488261146/jobs/10013394589?pr=7174#step:3:1)
> 
> So I think we might need to clear the compiler caches instead: [`main`/book/src/dev/continuous-integration.md#fixing-disk-full-errors-and-zcash-parameter-errors](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md?rgh-link-date=2023-07-09T19%3A22%3A15Z#fixing-disk-full-errors-and-zcash-parameter-errors)

That job also doesn't have any explicit Rust cache steps, but it might have implicit caching, I'm not sure.

I tried clearing all the test caches using `gh api` to access GitHub Actions, and I restarted one of the jobs:
https://github.com/ZcashFoundation/zebra/actions/runs/5488261146

If that doesn't work, we might need to turn coverage off, then open a ticket for larger runners. Standard runners have 14 GB of disk space:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

But the smallest "larger runner" has 150 GB:
https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners#machine-sizes-for-larger-runners

Larger runners also have a cost. (The default runners are free for open-source projects.) So we'd need to ask ops to set a cost cap for GitHub actions before we use them. Or @gustavovalverde might want to wait to switch to self-hosted runners.

teor2345:
This is now ticket #7180.

teor2345:
I manually deleted some 400 GB disks created by this PR, because I was getting errors in other PRs:
>  - Invalid value for field 'resource.disks[1].initializeParams.diskSizeGb': '300'. Requested disk size cannot be smaller than the image size (400 GB)

https://github.com/ZcashFoundation/zebra/actions/runs/5501815291/jobs/10026074730#step:11:79

I didn't realise this would happen, but it's an easy fix.

PR #7177, opened 2023-07-09
fix(Docker): Enable RPC port in general scenarios

### LINKED ISSUE #7166: Activate lightwalletd RPCs in Docker image when RPC port is set
## Motivation

> The `runtime-entrypoint.sh` uses the `RPC_PORT` env var when the user
> specifies the `getblocktemplate-rpc` feature

Some users might want to use the Docker image with `lightwalletd`, but without configuring `getblocktemplate-rpcs`. So let's work out how to do that!

Previously, the Dockerfile only activated the RPC port when it was set, but I agree that doesn't make sense with `getblocktemplate-rpcs`:
https://github.com/ZcashFoundation/zebra/commit/8c90f653915b1587a2fcc3f227a791a421f832ad#diff-5b838857fc3410c2438196effdbdcffae6548bdba431a4a938e76a0be62d008eL49

So we might want logic like:
* If `getblocktemplate-rpcs` is activated, activate the default RPC port if it is unset
* If there aren't any active RPC features, activate the RPC port only if it is set
* Otherwise, don't activate the RPC port

This covers the 3 different use cases of mining, light wallet, and no RPCs.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/7162#pullrequestreview-1517161441_

upbqdn:
## Motivation

Close #7166.

## Solution

Enable the RPC port if `$RPC_PORT` is set. If `$RPC_PORT` is not set, enable the default RPC port if `$FEATURES` contains `getblocktemplate-rpcs`.

### Reviewer Checklist

### DIFF
--- a/docker/runtime-entrypoint.sh
+++ b/docker/runtime-entrypoint.sh
@@ -31,12 +31,15 @@ fi
 : "${TRACING_ENDPOINT_PORT:=3000}"
 # [rpc]
 : "${RPC_LISTEN_ADDR:=0.0.0.0}"
+if [[ -z "${RPC_PORT}" ]]; then
+if [[ " ${FEATURES} " =~ " getblocktemplate-rpcs " ]]; then
 if [[ "${NETWORK}" = "Mainnet" ]]; then
 : "${RPC_PORT:=8232}"
 elif [[ "${NETWORK}" = "Testnet" ]]; then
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks!

PR #7178, opened 2023-07-09
add(Docker): Enable miners to set their address for mining rewards.

### LINKED ISSUE #6752: Pass mining config to the mining Docker container at runtime

upbqdn:
## Motivation

Close #6752.

## Solution

This PR became trivial, given all the preceding updates we did.

### Reviewer Checklist

## Follow Up Work

#7179

### DIFF
--- a/docker/runtime-entrypoint.sh
+++ b/docker/runtime-entrypoint.sh
@@ -105,6 +105,13 @@ use_color = false
 EOF
 fi
 fi
+
+if [[ -n "$MINER_ADDRESS" ]]; then
+cat <<EOF >> "$ZEBRA_CONF_PATH"
+[mining]
+miner_address = "${MINER_ADDRESS}"
+EOF
+fi
 fi
 
 echo "Using zebrad.toml:"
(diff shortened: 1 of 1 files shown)


teor2345 (approved):

gustavovalverde (commented):
LGTM

gustavovalverde (approved):

Issue #7008, opened 2023-06-19
Document Docker build-time & run-time config params


upbqdn:
## Motivation

We have recently refactored the Dockerfile so that it allows users to specify various parameters when launching their Docker containers. There are also a bunch of parameters that can be specified when building images.

These aren't documented anywhere.

## Suggested Solution

Document the parameters users can set both at build-time and when starting the container.

## Related Work

Issue #6759 might change how we handle Zebra's ports in Docker.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7008?planning-poker) with Zenhub @arya2 @gustavovalverde @oxarbitrage @teor2345 @upbqdn

Issue #7136, opened 2023-07-04
Publish next Zebra release: 1.1.0


mpguerra:
# Prepare for the Release

These release steps can be done a week before the release, in separate PRs.
They can be skipped for urgent releases.

## Checkpoints

For performance and security, we want to update the Zebra checkpoints in every release.
- [ ] You can copy the latest checkpoints from CI by following [the zebra-checkpoints README](https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/README.md#zebra-checkpoints).

## Missed Dependency Updates

Sometimes `dependabot` misses some dependency updates, or we accidentally turned them off.

This step can be skipped if there is a large pending dependency upgrade. (For example, the shared ECC crates in PR #7053)

Here's how we make sure we got everything:
- [x] Run `cargo update` on the latest `main` branch, and keep the output
- [x] If needed, [add duplicate dependency exceptions to deny.toml](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-duplicate-dependencies-in-check-denytoml-bans)
- [x] If needed, remove resolved duplicate dependencies from `deny.toml`
- [x] Open a separate PR with the changes
- [x] Add the output of `cargo update` to that PR as a comment

# Prepare and Publish the Release

Follow the steps in the [release checklist](https://github.com/ZcashFoundation/zebra/blob/main/.github/PULL_REQUEST_TEMPLATE/release-checklist.md) to prepare the release:

Release PR:
- [x] Update Changelog
- [x] Update README
- [x] Update Zebra Versions
- [x] Update End of Support Height

Publish Release:
- [x] Create & Test GitHub Pre-Release
- [x] Publish GitHub Release
- [x] Publish Rust Crates
- [x] Publish Docker Images

PR #7179, opened 2023-07-09
add(Docker):  Docs for mining with Docker

### LINKED ISSUE #6611: Document how to switch from Testnet to Mainnet in Docker images
## Motivation

We recently published a Docker image with mining enabled on Testnet. We would like to invite miners to try mining with Zebra on Mainnet as well. We should document how they can do the switch so we can point them to the documentation.

upbqdn:
## Motivation

Close #6611.

Depends-On: #7200 and #7231.

### Reviewer Checklist

## Follow Up Work

#7008.

### DIFF
--- a/book/src/user/mining-docker.md
+++ b/book/src/user/mining-docker.md
@@ -1,0 +1,33 @@
+# Mining with Zebra in Docker
+
+All [published images](https://hub.docker.com/r/zfnd/zebra/tags) containing the
+`.experimental` suffix are compiled with the `getblocktemplate-rpcs` feature, so
+you can use them for your mining operations. The images are pre-configured for
+Testnet, and you must pass your rewards address as an environment variable to
+the container when starting it. For example, executing
+
+```bash
+docker run -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/zebra:v1.0.1.experimental
+```
+
+will start a container on Testnet and bind port 18232 on your Docker host,
+which is the standard Testnet RPC port. If you want to use a different port, you
+can specify it in the `RPC_PORT` environment variable, as shown in the next
+example. Also, notice that the address we used starts with the prefix `t2`,
+indicating that it is a Testnet address. A Mainnet address would not work.
+
+If you want to mine on Mainnet, you can execute the following command.
+
+```bash
+docker run -e NETWORK="Mainnet" -e RPC_PORT="8232" -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:v1.0.1.experimental
+```
+
+Note that besides specifying the network, we also changed the port to 8232,
+which is the standard RPC port for Mainnet. Also, the rewards address now begins
+with the prefix `t1`, meaning it is a Mainnet P2PKH address. Please remember to
+use your own rewards address.
+
+Instead of listing the environment variables on the command line, you can use
+Docker's `--env-file` flag to specify a file containing the variables. You can
+find more info here
+https://docs.docker.com/engine/reference/commandline/run/#env.
--- a/book/src/user/mining-docker.md
+++ b/book/src/user/mining-docker.md
@@ -9,7 +9,8 @@ operations. For example, executing
 docker run -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:v1.1.0.experimental
 ```
 
-will start a container on Mainnet and bind port 8232 on your Docker host.
+will start a container on Mainnet and bind port 8232 on your Docker host. If you
+want to start generating blocks, you need to let Zebra sync first.
 
 Note that you must pass the address for your mining rewards via the
 `MINER_ADDRESS` environment variable when you are starting the container, as we
--- a/book/src/user/mining-docker.md
+++ b/book/src/user/mining-docker.md
@@ -1,33 +1,44 @@
 # Mining with Zebra in Docker
 
-All [published images](https://hub.docker.com/r/zfnd/zebra/tags) containing the
-`.experimental` suffix are compiled with the `getblocktemplate-rpcs` feature, so
-you can use them for your mining operations. The images are pre-configured for
-Testnet, and you must pass your rewards address as an environment variable to
-the container when starting it. For example, executing
+Some of our published [Docker images](https://hub.docker.com/r/zfnd/zebra/tags)
+have the `.experimental` suffix in their name. We compile these images with the
+`getblocktemplate-rpcs` feature, and you can use them for your mining
+operations. For example, executing
 
 ```bash
-docker run -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/zebra:v1.0.1.experimental
+docker run -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:v1.1.0.experimental
 ```
 
-will start a container on Testnet and bind port 18232 on your Docker host,
-which is the standard Testnet RPC port. If you want to use a different port, you
-can specify it in the `RPC_PORT` environment variable, as shown in the next
-example. Also, notice that the address we used starts with the prefix `t2`,
-indicating that it is a Testnet address. A Mainnet address would not work.
+will start a container on Mainnet and bind port 8232 on your Docker host.
 
-If you want to mine on Mainnet, you can execute the following command.
+Note that you must pass the address for your mining rewards via the
+`MINER_ADDRESS` environment variable when you are starting the container, as we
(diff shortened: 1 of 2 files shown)


upbqdn:
We'll need to test the commands in the docs when we do the next release. It would be good if we could test them before the release, but I don't know how.

teor2345:
> We'll need to test the commands in the docs when we do the next release. It would be good if we could test them before the release, but I don't know how.

You can test the commands on the release image in the CD builds, they use the `runtime` target:
https://github.com/ZcashFoundation/zebra/blob/f6afec2be81d8cc7af76c59b453ac5add9f7dccb/.github/workflows/continous-delivery.yml#L114

It seems like you'll want to test against `us-docker.pkg.dev/zfnd-dev-zebra/zebra/zebrad:pr-7178`, which is from:
https://github.com/ZcashFoundation/zebra/actions/runs/5501189414/jobs/10024596480?pr=7178#step:10:1985
(the CD build job on PR #7178).

But it doesn't use the `getblocktemplate-rpcs` feature, so I guess we could add a `features` list to the manual deployment and build arguments?
https://github.com/ZcashFoundation/zebra/blob/f6afec2be81d8cc7af76c59b453ac5add9f7dccb/.github/workflows/continous-delivery.yml#L16-L22
https://github.com/ZcashFoundation/zebra/blob/f6afec2be81d8cc7af76c59b453ac5add9f7dccb/.github/workflows/continous-delivery.yml#L109-L115

Or we could add a manual release trigger and tag it with the branch name. But that seems riskier, we could accidentally publish something that looks like a real release, and users would use it. (It might also update the `latest` tag and get used automatically by Docker.)

gustavovalverde (approved):
LGTM!

teor2345 (dismissed):
Have these commands been tested against the Docker image changes yet?

You can build Docker images locally if that helps?

gustavovalverde:
I tested this locally, as a result of executing the commands in this PR I got these configurations:

Running `docker run -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/<local_rpc_image>`

Creates this `zebrad.toml`:
```ini
[network]
network = "Testnet"
listen_addr = "0.0.0.0"
[state]
cache_dir = "/var/cache/zebrad-cache"
[tracing]
use_color = false
[mining]
miner_address = "t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v"
```

Running `docker run -e NETWORK="Mainnet" -e RPC_PORT="8232" -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:<local_rpc_image>`

Creates this `zebrad.toml`:
```ini
[network]
network = "Mainnet"
listen_addr = "0.0.0.0"
[state]
cache_dir = "/var/cache/zebrad-cache"
[rpc]
listen_addr = "0.0.0.0:8232"
[tracing]
use_color = false
[mining]
miner_address = "t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc"
```

teor2345:
> I tested this locally, as a result of executing the commands in this PR I got these configurations:

It looks like a RPC address is missing from this config, maybe it needs some parts of PR #7175? Or the `getblocktemplate-rpcs` feature?

mpguerra (commented):
Content looks good to me. I appreciate the clear examples. 

I have not run any of the commands so I don't know whether they work or not which is why I'm not explicitly approving this PR.

upbqdn:
> I tested this locally, ...

I think testing the commands locally doesn't suffice. We need to test them against the images we publish.

> It looks like a RPC address is missing from this config, maybe it needs some parts of PR https://github.com/ZcashFoundation/zebra/pull/7175? Or the getblocktemplate-rpcs feature?

Yes, this is also true for the currently published images. #7175 attempts to fix it. I tried describing what the issue is and why the port doesn't propagate to Zebra's config in the PR description.

I don't see how we could leverage the `getblocktemplate-rpcs` for published images.

upbqdn:
I set #7175 as a blocker for this PR. Let's figure out a proper config for the images, and then test this PR?

mpguerra:
After speaking with Marek, these instructions won't work until we publish the next release so I'm blocking this on #7136

book/src/user/mining-docker.md
```diff
@@ -0,0 +1,33 @@
+# Mining with Zebra in Docker
+
+All [published images](https://hub.docker.com/r/zfnd/zebra/tags) containing the
+`.experimental` suffix are compiled with the `getblocktemplate-rpcs` feature, so
+you can use them for your mining operations. The images are pre-configured for
+Testnet, and you must pass your rewards address as an environment variable to
+the container when starting it. For example, executing
+
+```bash
+docker run -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/zebra:v1.0.1.experimental
```
gustavovalverde:
```suggestion
docker run -e NETWORK="Testnet" -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/zebra:v1.1.0.experimental
```

book/src/user/mining-docker.md
```diff
@@ -0,0 +1,33 @@
...
+will start a container on Testnet and bind port 18232 on your Docker host,
+which is the standard Testnet RPC port. If you want to use a different port, you
+can specify it in the `RPC_PORT` environment variable, as shown in the next
+example. Also, notice that the address we used starts with the prefix `t2`,
+indicating that it is a Testnet address. A Mainnet address would not work.
+
+If you want to mine on Mainnet, you can execute the following command.
+
+```bash
+docker run -e NETWORK="Mainnet" -e RPC_PORT="8232" -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:v1.0.1.experimental
```
gustavovalverde: We already set the port depending on the `NETWORK` env variable

```suggestion
docker run -e NETWORK="Mainnet" -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd/zebra:v1.1.0.experimental
```

gustavovalverde (changes requested):
Minimum changes to get it in

mpguerra:
Did we do the required tests to ensure that these instructions will work?

gustavovalverde:
> Did we do the required tests to ensure that these instructions will work?

@mpguerra With the required changes from my review, it's working.

upbqdn:
> Did we do the required tests to ensure that these instructions will work?

I didn't, and we ended up with images with Mainnet enabled by default which isn't what we intended. However, after thinking about it, I think this change is actually suitable because there will be no API change when we start officially supporting mining, and I don't really see how having Testnet enabled by default protects from wasting the miner's resources, which is the risk we wanted to avoid. The reason is that to waste resources, the miner has to use a significant mining power for an extended period of time, and by that time, they must know whether they're on Mainnet or Testnet.

gustavovalverde:
> The reason is that to waste resources, the miner has to use a significant mining power for an extended period of time, and by that time, they must know whether they're on Mainnet or Testnet.

@upbqdn when I tested this the first time I used this command, which defaults to `Mainnet`:
```shell
docker run -e MINER_ADDRESS="t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v" -p 18232:18232 zfnd/zebra:v1.1.0.experimental
```

And the application panicked with the following error:
```shell
  left: `Testnet`,
 right: `Mainnet`: incorrect miner address config: t27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v network.network Mainnet and miner address network Testnet must match
Location: /opt/zebrad/zebra-rpc/src/server.rs:158
```

From my perspective, I doubt the user will be able to use the wrong Network, but I might not be considering other use-cases.

upbqdn:
> From my perspective, I doubt the user will be able to use the wrong Network, but I might not be considering other use-cases.

The issue isn't related to using an incompatible address; Zebra will always catch that. Our concern was that since we didn't have the means to test some of the mining functionality on Mainnet, users might run into bugs when running the image on Mainnet out of the box. However, as you pointed out, the image won't even start out of the box. Users have to first provide an address, which is either for Mainnet or Testnet, so they have to implicitly pick the network anyway. It's just a bit less of a hassle to start on Mainnet now, which I think is good.

book/src/user/mining-docker.md
upbqdn: Addressed in [f4ec740](https://github.com/ZcashFoundation/zebra/pull/7179/commits/f4ec740f302ba0a72752c64a0939538d9b533bf9).

book/src/user/mining-docker.md
upbqdn: Addressed in [f4ec740](https://github.com/ZcashFoundation/zebra/pull/7179/commits/f4ec740f302ba0a72752c64a0939538d9b533bf9).

gustavovalverde (dismissed):
Tested and LGTM!

upbqdn:
I refactored almost all of the docs.

upbqdn:
The examples work for me as well.

gustavovalverde (approved):

PR #7186, opened 2023-07-10
fix(ci): Pass the network input to all cached state tests

teor2345:
## Motivation

Some manual workflow runs can write testnet cache data to mainnet cached state disks. This makes any future jobs that use those cached states run for a long time or fail.

## Solution

- Pass the input network to all the cached state jobs
- Check that the input network matches the network that `zebrad` is using right after it launches

## Review

This is urgent because it's causing CI failures.

### Reviewer Checklist

## Follow Up Work

We should also apply the network to non-cached-state tests, but that's not urgent.

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -135,6 +135,9 @@ jobs:
 
   # zebrad tests without cached state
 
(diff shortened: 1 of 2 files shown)


gustavovalverde (approved):
LGTM

PR #7198, opened 2023-07-11
fix(ci): When testnet is manually selected, disable tests that don't work on it

teor2345:
## Motivation

Some tests try to run on mainnet when given a testnet variable and cached state:
https://github.com/ZcashFoundation/zebra/actions/runs/5514327024/jobs/10054409296#step:8:105

## Solution

Disable all the lightwalletd cached state jobs on testnet, because the lightwalletd full sync job doesn't work.

## Review

This should be fixed eventually, but we don't do testnet manual runs that often.

### Reviewer Checklist

## Follow Up Work

Some other jobs might not work, we can disable them later if we ever use testnet manual runs with them.

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -386,7 +386,7 @@ jobs:
     name: Zebra tip
     needs: [ build, get-available-disks ]
     uses: ./.github/workflows/deploy-gcp-tests.yml
-    if: ${{ github.event_name == 'schedule' || !fromJSON(needs.get-available-disks.outputs.zebra_tip_disk) || (github.event.inputs.run-full-sync == 'true' && github.event.inputs.network == 'Mainnet') }}
+    if: ${{ github.event_name == 'schedule' || !fromJSON(needs.get-available-disks.outputs.zebra_tip_disk) || (github.event.inputs.run-full-sync == 'true' && (inputs.network || vars.ZCASH_NETWORK) == 'Mainnet') }}
     with:
       app_name: zebrad
       test_id: full-sync-to-tip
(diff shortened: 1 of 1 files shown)


teor2345:
Testnet checkpoints failed due to a temporary testnet DNS seeder DNS error:
https://github.com/ZcashFoundation/zebra/actions/runs/5523991607/jobs/10075820704?pr=7198#step:8:158

mpguerra:
> Testnet checkpoints failed due to a temporary testnet DNS seeder DNS error: https://github.com/ZcashFoundation/zebra/actions/runs/5523991607/jobs/10075820704?pr=7198#step:8:158

@gustavovalverde ☝️

teor2345:
@mergifyio update

teor2345:
@mpguerra just wondering if we can get this reviewed this week? I won't be available for revisions next week.

dconnolly (approved):
i think this is fine, i'm not 100% sure if `(inputs.network || vars.ZCASH_NETWORK) == 'Mainnet'` returns a string, which string if both, when returning from the `()` expression, but it might be fine?

teor2345:
> i think this is fine, i'm not 100% sure if `(inputs.network || vars.ZCASH_NETWORK) == 'Mainnet'` returns a string, which string if both, when returning from the `()` expression, but it might be fine?

It seems to work, what it means is: "use the string from the inputs or env, then compare it with 'Mainnet' and return a boolean".

teor2345:
@mergifyio refresh

teor2345:
@mergifyio refresh

teor2345:
Temporary Google Cloud "service is not available" error, possibly a rate limit.

Issue #7170, opened 2023-07-07
Zebra sometimes panics in the state and network tasks on shutdown


teor2345:
### What happened?

I expected to see this happen:

A shutdown with no panics.

Instead, this happened:

> 2023-07-06T02:30:10.476975Z  INFO {net="Test"}: zebrad::components::tokio: waiting for async tokio tasks to shut down TOKIO_SHUTDOWN_TIMEOUT=20s
> The application panicked (crashed).
> Message:  panic in ReadRequest::Tip: JoinError::Cancelled(Id(23437))
> Location: zebra-state/src/service.rs:1208

https://github.com/ZcashFoundation/zebra/actions/runs/5470776835/jobs/9961459769#step:5:347

> 2023-07-10T04:02:08.269874Z  INFO {net="Test"}: zebrad::components::tokio: waiting for async tokio tasks to shut down TOKIO_SHUTDOWN_TIMEOUT=20s
> The application panicked (crashed).
> Message:  panic in next peer address book task: JoinError::Cancelled(Id(9554))
> Location: /opt/zebrad/zebra-network/src/peer_set/candidate_set.rs:407

https://github.com/ZcashFoundation/zebra/actions/runs/5503766375/jobs/10029501421#step:5:326

### What were you doing when the issue happened?

Running Zebra in CI and terminating the process.

### Zebra logs

You can see the full logs at:
https://github.com/ZcashFoundation/zebra/actions/runs/5470776835/jobs/9961459769#step:5:347
https://github.com/ZcashFoundation/zebra/actions/runs/5503766375/jobs/10029501421#step:5:326

### Zebra Version

1.0.1 with PR #7029
1.0.1 with PRs #7177 + #7183 in merge queue PR #7190

### Which operating systems does the issue happen on?

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

Ubuntu 4   22.04.2 5   LTS GitHub Runner

### Additional information

This is a low priority because the cache is not corrupted, and the user was shutting down anyway. It also causes rare CI failures. (Any tokio update could make them more common.)

Every JoinError::Cancelled in every state and network request will need to be ignored, something like this:
https://github.com/ZcashFoundation/zebra/blob/1014e3c7855df1c53d0bf0ef571d2cecd9b795f8/zebra-network/src/peer_set/initialize.rs#L633-L640

[...]

PR #7200, opened 2023-07-11
ref(docker): organize `ENV` and `ARG` values based on their usage

### LINKED ISSUE #7197: docker: Use of `ENV` and `ARG` it's misleading across the repo
## Motivation

While @upbqdn was working with #7179 I realized the actual panorama is very confusing as we've been using variables and arguments in different ways and in places where some of those shouldn't be.

### Testing

We have to build and make manual deployments to GCP to confirm the behavior we're expecting is being kept

## Related Work
- #7179

gustavovalverde:
## Motivation

Docker environment variables (ENV) and build arguments (ARG) were being used in a confusing way across the Dockerfile and entrypoint. This PR is scoping each variable and build argument under the stage they were meant to be used.

As a consequence this PR: 
Overrides PR #7175
Allows PR  #7179 to work as expected
Fixes #7197

### Specifications

Summary: https://vsupalov.com/docker-arg-vs-env/
Deep-dive: https://vsupalov.com/docker-arg-env-variable-guide/

## Solution

- Add global ARGs which are required across stages
- Add `SHORT_SHA` in the `deps` stage for build scope
- Use a single `TEST_FEATURES` argument and variable for tests purposes
- Add the `FEATURES` ARG to the `release` stage to change build time features
- Add the `FEATURES` and `NETWORK` ARG and ENV to the `runtime` stage for consistency with global/previous ARGs 

## Review

@teor2345 and @upbqdn might want to have a look at this

### Reviewer Checklist

## Follow Up Work

Based on this changes, also remove inconsistency from workflows.
Test this changes
I might also remove `NETWORK` as an ARG, but I need to fully evaluate the whole pipeline (but this should only be an ENV variable, as it has no impact at build time).


docker/Dockerfile
```diff
@@ -175,10 +182,12 @@ RUN apt-get update && \
...
 
-# Config settings
-
+# Config settings for zebrad
 ARG NETWORK
-ENV NETWORK=${NETWORK:-Mainnet}
+ENV NETWORK=${NETWORK}
+
+ARG FEATURES
+ENV FEATURES=${FEATURES}
```
upbqdn, line 185:
```suggestion
ENV FEATURES=${FEATURES}

ARG RPC_PORT
ENV RPC_PORT=${RPC_PORT}
```

I think we need this to pass the port from here https://github.com/zcashfoundation/zebra/blob/3944c3f199e31ef2e992cd31bc67d7d08d9c124b/.github/workflows/release-binaries.yml#L46 to `runtime-entrypoint.sh`

docker/Dockerfile
```diff
@@ -8,6 +8,16 @@
...
 #
+# We first set default values for build arguments used across the stages.
+# Each stage must define the build arguments (ARGs) it uses.
+#
+# Build zebrad with these features
+# Keep these in sync with:
+# https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml#L42
+ARG FEATURES="default-release-binaries"
+ARG TEST_FEATURES="${FEATURES} lightwalletd-grpc-tests zebra-checkpoints"
+ARG NETWORK="Mainnet"
```
upbqdn: ###### Optional

Wouldn't it be clearer to have the default values set in each stage that uses the build arg? We need to redeclare these args later on anyway, and if there are declared new ones in a given stage, it might be hard to tell if they have a default value without looking up here.

docker/Dockerfile
```diff
@@ -116,7 +115,7 @@ COPY . .
 # It creates fake empty test binaries so dependencies are built, but Zebra is not fully built.
 #
 # TODO: add --locked when cargo-chef supports it
-RUN cargo chef cook --tests --release --features "${TEST_FEATURES} ${FEATURES}" --workspace --recipe-path recipe.json
+RUN cargo chef cook --tests --release --features "${TEST_FEATURES}" --workspace --recipe-path recipe.json
```
upbqdn: Will we not miss some features if we only use `$TEST_FEATURES`?

docker/entrypoint.sh
```diff
@@ -25,43 +25,43 @@ case "$1" in
         if [[ "$RUN_ALL_TESTS" -eq "1" ]]; then
             # Run all the available tests for the current environment.
             # If the lightwalletd environmental variables are set, we will also run those tests.
-            cargo test --locked --release --features "$ENTRYPOINT_FEATURES" --workspace -- --nocapture --include-ignored
+            cargo test --locked --release --features "$TEST_FEATURES" --workspace -- --nocapture --include-ignored
```
upbqdn: `$ENTRYPOINT_FEATURES` used to combine `$FEATURES` and `$TEST_FEATURES`. Similarly to the previous comment, will we not miss `$FEATURES`?

docker/Dockerfile
```diff
@@ -58,39 +68,20 @@ RUN if [ "$(uname -m)" != "aarch64" ]; then \
...
     rm -rf /var/lib/apt/lists/* /tmp/*
 
-# TODO: just keep the backtrace, colorbt, rust_log, and cargo_home variables as those are the only needed at build time.
-# Build arguments and variables set to change how tests are run, tracelog levels,
-# and Network to be used (Mainnet or Testnet)
+# Build arguments and variables set for tracelog levels and debug information
 #
 # We set defaults to all variables.
 ARG RUST_BACKTRACE
 ENV RUST_BACKTRACE=${RUST_BACKTRACE:-0}
```
teor2345: I'd like to turn on backtraces, so we can to diagnose bugs in Zebra tests. This requires some environmental variables to be set to `1`. They don't need to be user-configurable.

For example, I can't make sure I've fixed the panic bugs in #7170 without a backtrace, because this is all the output I get:
> Message:  panic in ReadRequest::Tip: JoinError::Cancelled(Id(23437))
> Location: zebra-state/src/service.rs:1208
> Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.

https://github.com/ZcashFoundation/zebra/actions/runs/5470776835/jobs/9961459769#step:5:347

```suggestion
ENV RUST_BACKTRACE=${RUST_BACKTRACE:-1}

ARG RUST_LIB_BACKTRACE
ENV RUST_LIB_BACKTRACE=${RUST_LIB_BACKTRACE:-1}
```

docker/Dockerfile
```diff
@@ -58,39 +68,20 @@ RUN if [ "$(uname -m)" != "aarch64" ]; then \
...
 
-ARG RUST_LIB_BACKTRACE
-ENV RUST_LIB_BACKTRACE=${RUST_LIB_BACKTRACE:-0}
-
 ARG COLORBT_SHOW_HIDDEN
 ENV COLORBT_SHOW_HIDDEN=${COLORBT_SHOW_HIDDEN:-0}
```
teor2345:
```suggestion
ENV COLORBT_SHOW_HIDDEN=${COLORBT_SHOW_HIDDEN:-1}
```

docker/Dockerfile
```diff
@@ -175,10 +182,12 @@ RUN apt-get update && \
...
 
-# Config settings
-
+# Config settings for zebrad
 ARG NETWORK
-ENV NETWORK=${NETWORK:-Mainnet}
+ENV NETWORK=${NETWORK}
+
+ARG FEATURES
+ENV FEATURES=${FEATURES}
```
teor2345, line 186: We need these environmental variables to diagnose bugs in production images as well.
They don't need to be user-configurable.

```suggestion

ARG RUST_BACKTRACE
ENV RUST_BACKTRACE=${RUST_BACKTRACE:-1}

ARG RUST_LIB_BACKTRACE
ENV RUST_LIB_BACKTRACE=${RUST_LIB_BACKTRACE:-1}

ARG COLORBT_SHOW_HIDDEN
ENV COLORBT_SHOW_HIDDEN=${COLORBT_SHOW_HIDDEN:-1}

```

teor2345 (commented):
I'd like backtraces in tests and production, please.

docker/Dockerfile
gustavovalverde: If this is meant to be set at runtime (tests and/or production), we should handle those in the pipelines with `vars.RUST_BACKTRACE` and the others, as we've done here: https://github.com/search?q=repo%3AZcashFoundation%2Fzebra%20vars.RUST_BACKTRACE&type=code

docker/Dockerfile
gustavovalverde: `$TEST_FEATURES` includes `$FEATURES` from here: https://github.com/ZcashFoundation/zebra/pull/7200/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR18

docker/Dockerfile
gustavovalverde: The `$RPC_PORT` is just used as an env variable at runtime, we'd only need an `ARG` here if this required to be passed at build time, but it's not.

docker/Dockerfile
gustavovalverde: To have these environment variables in production, is not required to have those in the Dockerfile. We can set these defaults (if needed) in the `entrypoint` script

docker/Dockerfile
upbqdn: I don't think that's right for the mining image. The currently released `.experimental` image doesn't work because it has no RPC port specified. You can see it yourself if you run the container and look at the config.

upbqdn: Here's the config of `zfnd/zebra:v1.0.1.experimental`:

```toml
[network]
network = "Testnet"
listen_addr = "0.0.0.0"
[state]
cache_dir = "/var/cache/zebrad-cache"
[tracing]
use_color = false
```

The RPC port specified at https://github.com/zcashfoundation/zebra/blob/3944c3f199e31ef2e992cd31bc67d7d08d9c124b/.github/workflows/release-binaries.yml#L46 is missing.

docker/Dockerfile
upbqdn: Oh, I assumed that if I used `--build-arg TEST_FEATURES="foo"`, I would override the default value, and `TEST_FEATURES` would contain only `foo`, excluding `FEATURES`. Was that wrong?

upbqdn: In other words, `TEST_FEATURES` will contain `FEATURES` only if we don't set `TEST_FEATURES`. However, here https://github.com/zcashfoundation/zebra/blob/7b9a0aed04efaf61d35988283fea5d2c841da0f3/.github/workflows/build-docker-image.yml#L165 we set `TEST_FEATURES`.

gustavovalverde:
This can be tested locally (for mining) with the following commands:

Build with:
```shell
docker build -f ./docker/Dockerfile --target runtime --build-arg FEATURES='default-release-binaries getblocktemplate-rpcs'  --tag zfnd-local/zebra.mining:latest .
```

Optionally, add these `build-args`: 
```shell
--build-arg RUST_BACKTRACE=full --build-arg RUST_LOG=debug --build-arg COLORBT_SHOW_HIDDEN=1
```

Run with:
```shell
docker run -e NETWORK="Mainnet" -e RPC_PORT="8232" -e MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 zfnd-local/zebra.mining
```

This outputs the following `zebrad.toml`:

```ini
[network]
network = "Mainnet"
listen_addr = "0.0.0.0"
[state]
cache_dir = "/var/cache/zebrad-cache"
[rpc]
listen_addr = "0.0.0.0:8232"
[tracing]
use_color = false
[mining]
miner_address = "t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc"
```

This config file tackles the bug identified by @teor2345 here: https://github.com/ZcashFoundation/zebra/pull/7179#issuecomment-1629783599

docker/Dockerfile
gustavovalverde: Oh, now I got it! I oversighted that use case. Let me think if the previous approach is the only way to handle this.

docker/Dockerfile
teor2345: That works for our environment, but we also want these variables set to get better bug reports from users.

Do our CI environment variables get propagated into the production Docker images as defaults?

docker/Dockerfile
teor2345: Great! However you want to do it is fine with me.

docker/Dockerfile
gustavovalverde: Yes, that was an issue which has been fixed in this commit: https://github.com/ZcashFoundation/zebra/pull/7200/commits/55ea8019b8274a4c3540e64e3bc9798c68ab3fcc

docker/Dockerfile
gustavovalverde: **Summary**
- Setting variables–_without_ defaults–at `runtime` is the easiest, those doesn't have to be defined in the `Dockerfile` nor the `entrypoint`
- Setting _default_ variables at `runtime` just requires to set them in the `entrypoint`
- Setting _default_ variables at `build time` requires them to be `ARG` (`--build-arg`) and `ENV`, as that's the only way to set them.

Our CI environment variables _can_ get propagated to the **deployments** if those are set in the workflow (globally or on the specific job/task) and added, for example, here:
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/continous-delivery.yml#L279

Or here, using `-e`:
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/deploy-gcp-tests.yml#L256

If we want defaults (at runtime) for our production images, we would add them in the entrypoint as we're doing here:
- https://github.com/ZcashFoundation/zebra/blob/main/docker/runtime-entrypoint.sh#L20

If we want defaults (at build time) for our production and/or tests images, we should add them in Dockerfile as `ARG` and `ENV` here:
- https://github.com/ZcashFoundation/zebra/blob/main/docker/Dockerfile#L66
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml#L159

**Note**: `NETWORK` is not required as a variable at build time, so we can easily remove those `ARG`, they still exist as our pipeline was built expecting this variable as an `ARG` (that was a wrong understanding on my side as when this variable was actually required).

cc: @upbqdn

docker/Dockerfile
upbqdn: The docs say that `-n string` is analogous to just `string`: 
image

so adding `-n` shouldn't do anything?

I must be missing something. How will the port propagate from https://github.com/ZcashFoundation/zebra/blob/7b0dedd3a28b3bcb24214ecae7d00408e5349749/.github/workflows/release-binaries.yml#L46 to `runtime-entrypoint.sh`?

Another solution would be to remove https://github.com/ZcashFoundation/zebra/blob/7b0dedd3a28b3bcb24214ecae7d00408e5349749/.github/workflows/release-binaries.yml#L46 and https://github.com/ZcashFoundation/zebra/blob/7b0dedd3a28b3bcb24214ecae7d00408e5349749/.github/workflows/build-docker-image.yml#L166 and let `runtime-entrypoint.sh` handle the port based only on `getblocktemplate-rpcs` in `$FEATURES`.

gustavovalverde: > The docs say that `-n string` is analogous to just `string`

That's correct, maybe it's because we're using double quotes when passing the argument
- https://github.com/ZcashFoundation/zebra/pull/7200#issuecomment-1634653085
- https://github.com/ZcashFoundation/zebra/pull/7179/files#diff-573d08e7251621753237189e973a77b6c83e29e9b4b5f16d3f8fc8f26ed053b3R22

The following `ARGs` should not exist in the `release-binaries.yml` workflow:
- network
- rpc_port
- test_features
- zebra_skip_ipv6_tests

I'd even remove others. Those changes will follow on a PR depending on this one. I explained why those are not needed here: https://github.com/ZcashFoundation/zebra/pull/7200#discussion_r1263572502 but I'm not sure if that's clear.

upbqdn: I see now. Shouldn't we then remove https://github.com/ZcashFoundation/zebra/blob/7b0dedd3a28b3bcb24214ecae7d00408e5349749/.github/workflows/build-docker-image.yml#L166 as well?

gustavovalverde: Yes, we should also remove that.

upbqdn: All clear now. Thanks, Gustavo.

I think we can even remove the `-n`.

docker/Dockerfile
gustavovalverde: Reverted: [revert(docker): revert to using extra ENTRYPOINT_FEATURES](https://github.com/ZcashFoundation/zebra/pull/7200/commits/285c67468978bb29528658afaf049e9100ff0d89)

docker/Dockerfile
gustavovalverde: Fixed here: 
[fix(docker): enable backtraces for errors and panics](https://github.com/ZcashFoundation/zebra/pull/7200/commits/37b591441125d53824746ea4a2abc4fa60cdb51b)

docker/Dockerfile
teor2345: Thanks for the explanation! What we need to debug panics and other bugs is for the 3 backtrace environmental variables to be set in ZF tests, ZF deployments, and user-run images.

Since they only change the output during panics, it's unlikely that users will want to turn them off, so making them user-configurable is optional.

I'm focused on other work right now, so I'm going to leave the details to you both.

gustavovalverde: As making this change will require further changes in our pipeline (aside from the Dockerfile), I'll make a different PR for that

docker/Dockerfile
```diff
@@ -8,6 +8,15 @@
 # - release: builds release binary
 # - runtime: is our runtime environment
 #
+# We first set default values for build arguments used across the stages.
+# Each stage must define the build arguments (ARGs) it uses.
+#
+# Build zebrad with these features
+# Keep these in sync with:
+# https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml#L42
```
upbqdn:
```suggestion
# https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/build-docker-image.yml#L37
```

docker/entrypoint.sh
```diff
@@ -8,7 +8,9 @@ set -e
 set -o pipefail
 
 # TODO: expand this section if needed (#4363)
```
upbqdn:
```suggestion
```

upbqdn (dismissed):
utACK

upbqdn (approved):

PR #7213, opened 2023-07-13
suggestion: Make recent IP splitting more efficient by doing the time calculation once

teor2345:
This is a suggestion for @arya2 for PR #7041:
- make recent IP splitting more efficient by doing the time calculation once
- tweak some comments


arya2 (approved):
Thank you!

PR #7214, opened 2023-07-13
suggestion: Make recent IP splitting more efficient using a BTreeSet to compare once

teor2345:
This is a suggestion for @arya2 for PR #7041:
- Use a BTreeSet to compare times once, rather than a binary search

Even during an attack, a binary search would only be comparing at most `log2(12_000) = 14` times per connection, which is a trivial amount compared to all the other connection operation. So I'm not sure if this change is worth it.

It has an increased risk of bugs due to duplicate `Instant`s from the OS. (Which can happen due to low resolution timers or OS bugs.)

It's totally up to you, please close it if you are unsure about the duplicate time risk.

### DIFF
--- a/zebra-network/src/peer_set/initialize/recent_by_ip.rs
+++ b/zebra-network/src/peer_set/initialize/recent_by_ip.rs
@@ -1,8 +1,8 @@
 //! A set of IPs from recent connection attempts.
 
 use std::{
-    collections::{HashMap, VecDeque},
-    net::IpAddr,
+    collections::{BTreeSet, HashMap},
+    net::{IpAddr, Ipv4Addr},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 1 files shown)


teor2345:
@arya2 I thought you might want to see how this could work, even if we don't merge it.

zebra-network/src/peer_set/initialize/recent_by_ip.rs
```diff
@@ -74,13 +79,12 @@ impl RecentByIp {
...
         // After the handshake, the peer set will remove any duplicate connections over the limit.
         let age_limit = now - self.time_limit;
+        // This is the lowest IP address among all IpAddr.
+        let dummy_ip = Ipv4Addr::UNSPECIFIED.into();
 
-        // `by_time` must be sorted for this to work.
-        let split_off_idx = self.by_time.partition_point(|&(_, time)| time <= age_limit);
-
-        let updated_by_time = self.by_time.split_off(split_off_idx);
+        let updated_by_time = self.by_time.split_off(&(age_limit, dummy_ip));
```
arya2, line 85: This seems to check every element until it finds something less than or equal to the key:

image

teor2345: You're right, that's weird:
> Currently, our implementation simply performs naive linear search. 

https://doc.rust-lang.org/std/collections/struct.BTreeMap.html

However, due to caching, performance is usually ok:
> This provides excellent performance on small nodes of elements which are cheap to compare. However in the future we would like to further explore choosing the optimal search strategy based on the choice of B, and possibly other factors. Using linear search, searching for a random element is expected to take B * log(n) comparisons, which is generally worse than a BST. In practice, however, performance is excellent.

But this is unlikely to hold for large numbers of elements.

And a `VecDeque` deals with the cache-miss part of the issues with binary trees:
> In theory, a binary search tree (BST) is the optimal choice for a sorted map, as a perfectly balanced BST performs the theoretical minimum amount of comparisons necessary to find an element (log2n). However, in practice the way this is done is very inefficient for modern computer architectures. In particular, every element is stored in its own individually heap-allocated node. This means that every single insertion triggers a heap-allocation, and every single comparison should be a cache-miss. Since these are both notably expensive things to do in practice, we are forced to, at the very least, reconsider the BST strategy.

teor2345: So I don't think we should merge this, thanks for checking!

arya2:
I learned that `BTreeMap::split_off()` will return the index of the first element that is `Ordering::Less` than the key, thank you!

PR #7217, opened 2023-07-14
fix(tests): Place coinbase tx as the first tx in fake child blocks

upbqdn:
## Motivation

I noticed we weren't making fake child blocks correctly. The issue was that we weren't placing the updated coinbase tx as the first one in the block. All tests using the `make_fake_child` function were working fine because they all use blocks with only one tx. 

## Solution

Place the updated coinbase tx as the first one in the block.

### Reviewer Checklist


oxarbitrage (approved):
Yea, i think is more correct to do `insert` here even if the effect is the same. Thanks!

PR #7218, opened 2023-07-15
fix(state): Remove workarounds for storing trees

upbqdn:
## Motivation

We were using some workarounds for committing trees into the non-finalized state. Namely, we turned off asserting that we don't commit more than a single tree at a particular height in tests. It would be great to have this resolved before we do the tree deduplication.

## Solution

This PR removes the workarounds and fixes tests so they don't push more than a single block at a particular height.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/non_finalized_state/tests/vectors.rs
+++ b/zebra-state/src/service/non_finalized_state/tests/vectors.rs
@@ -67,7 +67,7 @@ fn construct_many() -> Result<()> {
         zebra_test::vectors::BLOCK_MAINNET_434873_BYTES.zcash_deserialize_into()?;
     let initial_height = block
         .coinbase_height()
-        .expect("Block 434873 should have its height in its coibase tx.");
+        .expect("Block 434873 should have its height in its coinbase tx.");
     let mut blocks = vec![];
 
     while blocks.len() < 100 {
--- a/zebra-state/src/service/non_finalized_state/tests/vectors.rs
+++ b/zebra-state/src/service/non_finalized_state/tests/vectors.rs
@@ -65,6 +65,9 @@ fn construct_many() -> Result<()> {
 
     let mut block: Arc<Block> =
         zebra_test::vectors::BLOCK_MAINNET_434873_BYTES.zcash_deserialize_into()?;
+    let initial_height = block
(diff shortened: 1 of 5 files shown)


zebra-state/src/service/non_finalized_state/tests/vectors.rs
```diff
@@ -65,6 +65,9 @@ fn construct_many() -> Result<()> {
 
     let mut block: Arc<Block> =
         zebra_test::vectors::BLOCK_MAINNET_434873_BYTES.zcash_deserialize_into()?;
+    let initial_height = block
+        .coinbase_height()
+        .expect("Block 434873 should have its height in its coibase tx.");
```
teor2345:
```suggestion
        .expect("Block 434873 should have its height in its coinbase tx.");
```

teor2345 (approved):
Thanks for these fixes!

I just committed one typo fix.

teor2345 (changes requested):
Hmm, looks like 3 proptests failed for 3 different reasons?

> ---- service::non_finalized_state::tests::prop::forked_equals_pushed_genesis stdout ----
> Message:  hash is present
> Location: zebra-state/src/service/non_finalized_state/tests/prop.rs:206

https://github.com/ZcashFoundation/zebra/actions/runs/5570108032/jobs/10174308429?pr=7218#step:4:4059

We didn't get a proptest seed out of these tests, so we might need to run a few tests to make sure we've fixed this completely.

upbqdn:
It looks like it's two proptests, but the log contains three entries because one of the tests panicked in an `expect`, which got reported, and that made the test fail, which got reported again.

I ran both protests five times locally, and none of them failed once. :/ I think I found the issue for both failures, though. I pushed a fix.

teor2345:
> I ran both protests five times locally, and none of them failed once. :/ I think I found the issue for both failures, though. I pushed a fix.

I usually run all the related proptests in a loop for a few thousand iterations, overnight or while I'm doing other work.

teor2345 (approved):
Let's try it and see

teor2345:
This was automatically marked with `extra-reviews` because it contains the word "remove", but it doesn't actually remove any user-visible features.

PR #7231, opened 2023-07-17
ref(docker): remove all unrequired docker arguments from CI/CD pipelines

gustavovalverde:
## Motivation

This is a follow-up work based on:
- #7200

Considering we refactor all the `ARGs` and `ENVs` from the Dockerfile, we had a pending task of syncing those changes with the pipelines.

Some inputs were being used for build arguments which are no longer required, in most scenarios those were using the same defaults we have on our Dockerfile, and thus were removed to reduce confusion for other engineers which might see those as a requirement.

The `NETWORK` has been removed from most inputs, and instead moved to the place where it should be: Docker environment variables (ENV) at runtime. Based on our `entrypoint` approach, this is safer.

## Solution

- Remove build arguments which are not required from the calling action and the image build workflow
- Remove environment variables if those are build arguments, like `SHORT_SHA` (which is still part of the VM name)
- Prefer using `-e` (Docker `ENV`) instead of `inputs` or `build arguments` for `$NETWORK`

## Review

Anyone from DevOps, but it might be easier for @upbqdn or Teor, as they participated in the previous review. 

### Reviewer Checklist

## Follow Up Work

None

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -131,11 +131,8 @@ jobs:
       dockerfile_target: tests
       image_name: ${{ vars.CI_IMAGE_NAME }}
       no_cache: ${{ inputs.no_cache || false }}
-      network: ${{ inputs.network || vars.ZCASH_NETWORK }}
       rust_backtrace: full
       rust_lib_backtrace: full
-      colorbt_show_hidden: '1'
-      zebra_skip_ipv6_tests: '1'
       rust_log: info
 
   # zebrad tests without cached state
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -189,7 +189,7 @@ jobs:
       - name: Run zebrad tests
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker -e NETWORK run --name zebrad-tests --tty -e ${{ inputs.network || vars.ZCASH_NETWORK }} ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features "lightwalletd-grpc-tests getblocktemplate-rpcs" --workspace -- --include-ignored
+          docker run -e NETWORK run --name zebrad-tests --tty -e ${{ inputs.network || vars.ZCASH_NETWORK }} ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features "lightwalletd-grpc-tests getblocktemplate-rpcs" --workspace -- --include-ignored
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -42,12 +39,7 @@ jobs:
       dockerfile_target: runtime
       image_name: zebra
       tag_suffix: .experimental
-      network: Testnet
```
teor2345, line 45: This changes the default mining image target from testnet to mainnet. Is that intentional?

upbqdn: I'd stick with Testnet until we move mining out of `experimental`.

gustavovalverde: Oh yeah, that was unintended. I'll fix that.

.github/workflows/continous-integration-docker.yml
```diff
@@ -190,7 +189,9 @@ jobs:
       - name: Run zebrad tests
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
-          docker run --name zebrad-tests --tty ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features "lightwalletd-grpc-tests getblocktemplate-rpcs" --workspace -- --include-ignored
+          docker -e NETWORK run --name zebrad-tests --tty -e ${{ inputs.network || vars.ZCASH_NETWORK }} ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features "lightwalletd-grpc-tests getblocktemplate-rpcs" --workspace -- --include-ignored
```
upbqdn:
```suggestion
          docker run -e NETWORK run --name zebrad-tests --tty -e ${{ inputs.network || vars.ZCASH_NETWORK }} ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features "lightwalletd-grpc-tests getblocktemplate-rpcs" --workspace -- --include-ignored
```

upbqdn (dismissed):
utACK

.github/workflows/release-binaries.yml
gustavovalverde: Fixed by: [fix(release): do not change default $NETWORK for experimental image](https://github.com/ZcashFoundation/zebra/pull/7231/commits/bd5b6c831b7807f8038d0cf33dcb19e09fd44ca7)

gustavovalverde:
This will need re-approval @upbqdn

.github/workflows/release-binaries.yml
gustavovalverde: Nevermind, we should fix this in the `entrypoint` as the argument is not being passed to the `ENV` anyways. Based on our latest Dockerfile

upbqdn (dismissed):

gustavovalverde:
@upbqdn sorry to ping you again for re-approval. I had to fix a typo as a tests was failing.

upbqdn (approved):

PR #7239, opened 2023-07-17
change(state): Deduplicate note commitment trees in non-finalized state

### LINKED ISSUE #4784: Store only the first tree state in each series of identical tree states

upbqdn:
## Motivation

Address a part of #4784.

## Solution

This PR stores only the first tree in a series of identical trees in the non-finalized state.

## Review

This PR doesn't contain any additional tests. If reviewers have suggestions for suitable tests, I'd be glad to hear them.

### Reviewer Checklist


teor2345:
It looks like this contains a copy of PR #7218. Do you mind if I change the PR base to make the diff smaller to review?

(I'm a big fan of this PR and I'm glad we're getting it done.)

upbqdn:
This PR is built atop #7218, so we should merge that first. I forgot to target that branch instead of main. I don't think I can change it now.

teor2345:
> This PR is built atop #7218, so we should merge that first. I forgot to target that branch instead of main. I don't think I can change it now.

The base branch can be changed using the PR name edit button. (It's a bit weird but it works.)

IMG_3953

upbqdn:
Excellent, I thought you couldn't change the base branch. It turns out you can.

upbqdn:
(I  wrote the two comments above at the same time as you posted yours, so they might seem a bit off.)

teor2345 (dismissed):
This all looks pretty good!

I want to check we handle chain forks correctly, is there a test we can do to make sure it works?

I think we have a lot of unit and proptest coverage already. But we could manually query the last 100 or so heights using the RPC, and do it in a loop?

I don't know if we need to check the RPC responses against zcashd or an unmodified Zebra. Do you think it's more important to test for panics or wrong data?

upbqdn:
> Do you think it's more important to test for panics or wrong data?

Both would have a severe impact, so they seem equally important to me.

I'll try the RPC on the last 100 blocks. I'll compare it to an unmodified Zebra.

teor2345:
Thanks!

If you have a script I'm happy to run it as well, it would be good to do a test during a fork.

zebra-chain/src/block/height.rs
```diff
@@ -65,6 +65,29 @@ impl Height {
...
+    }
+
+    /// Returns the previous [`Height`].
+    ///
+    /// # Panics
+    ///
+    /// - If the current height is at its minimum.
+    pub fn previous(self) -> Self {
+        (self - 1).expect("Height should not be at its minimum.")
+    }
```
dconnolly, line 85: I don't disagree with the panics in principle, but most of the functionality implemented for `Height` returns errors vs panics (except the `TryFrom` impl which panics _and_ returns an Error and I'm not sure if that's what we want?) I mostly raise this as we fixed several panics that we turned into errors in the course of the audit so I figured I'd ask

// @teor2345 @arya2 curious on your thoughts too

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -500,14 +500,25 @@ impl Chain {
...
-    /// Add the Sprout `tree` to the tree and anchor indexes at `height`.
+    /// Adds the Sprout `tree` to the tree and anchor indexes at `height`.
     ///
     /// `height` can be either:
+    ///
     /// - the height of a new block that has just been added to the chain tip, or
-    /// - the finalized tip height: the height of the parent of the first block of a new chain.
+    /// - the finalized tip height—the height of the parent of the first block of a new chain.
+    ///
+    /// Stores only the first tree in each series of identical trees.
```
dconnolly, line 516: 👍

zebra-chain/src/block/height.rs
teor2345: I agree that as a general principle we should handle errors in the caller, rather than panicking for conditions that can happen during normal operation.

But I don't think it's a blocker for this specific PR, so I'd like to deal with the error handling on these new methods and `TryFrom` in a separate ticket or PR.

dconnolly (commented):
generally looks good, just one q about panics vs returning errors around Height

zebra-chain/src/block/height.rs
upbqdn: Thanks for bringing this up. I'll keep in mind that returning errors is more suitable. I'll fix it in this, or a subsequent PR.

upbqdn:
I synced the last ~ 1% of the Mainnet chain with this PR with no issues. Then, I got the tip with
```bash
zcash-cli getblockcount
```
which was 2160512. Then, I got the Sapling and Orchard treestates of the last 200 blocks with
```bash
for height in $(seq 2160312 2160512); do
    zcash-cli z_gettreestate "${height}"
done
```
Then, I switched to the main branch and collected the treestates of the same 200 blocks again. Zebra was still at the same tip after I collected the second batch. I `diff`ed them, and there was no difference.

upbqdn:
> I want to check we handle chain forks correctly, is there a test we can do to make sure it works?

This test is passing https://github.com/zcashfoundation/zebra/blob/3e75cb50f66ac783a477eb0db92529e88df69115/zebra-state/src/service/non_finalized_state/tests/prop.rs#L117-L131. It checks that a forked chain has the same state, including note commitment trees (https://github.com/zcashfoundation/zebra/blob/3e75cb50f66ac783a477eb0db92529e88df69115/zebra-state/src/service/non_finalized_state/tests/prop.rs#L224 and https://github.com/zcashfoundation/zebra/blob/3e75cb50f66ac783a477eb0db92529e88df69115/zebra-state/src/service/non_finalized_state/chain.rs#L268-L270), as the original chain after pushing the same blocks.

teor2345 (approved):
Thanks for the tests, that seems like a comprehensive check.

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -521,11 +532,20 @@ impl Chain {
...
-        assert_eq!(
-            self.sprout_trees_by_height.insert(height, tree.clone()),
-            None,
-            "incorrect overwrite of sprout tree: trees must be reverted then inserted",
-        );
+        // Don't add a new tree unless it differs from the previous one or there's no previous tree.
+        if height.is_min()
+            || self
+                .sprout_tree(height.previous().into())
+                .map_or(true, |prev_tree| prev_tree != tree)
```
arya2, line 539: Optional:

I don't think these extra steps are necessary, hashing the roots to check if the ~sprout~ sapling/orchard trees are the same looks expensive, and it would only clone the `Arc` for the previous tree unless it's been modified.

These trees come from `X_trees_by_height.last_key_value().unwrap().clone()`: https://github.com/ZcashFoundation/zebra/blob/release-v1.1.0/zebra-chain/src/parallel/tree.rs#L104-L124

teor2345: The roots for each tree are already calculated and cached, because we use them to verify the history tree. (Or index the sprout tree roots.)

teor2345: So I don't think there's any CPU impact here. However, it is faster to compare the inner frontier if the roots aren't cached. We could fix that in the `PartialEq` impl.

arya2: > The roots for each tree are already calculated and cached

I meant hashing the roots. It's also a second lookup when reading/writing or a bit more searching when reading.

> So I don't think there's any CPU impact here.

I don't think there should be a significant impact either.

upbqdn: Where do you think the roots get hashed?

Unless I miss something, the roots are the resulting hashes of the trees, and the trees cache them.

upbqdn: We basically compare `[u8; 32]` https://github.com/zcashfoundation/zebra/blob/9ebd56092bcdfc1a09062e15a0574c94af37f389/zebra-chain/src/sprout/tree.rs#L100.

upbqdn: > I don't think these extra steps are necessary

I couldn't come up with anything better than comparing the tree being stored to the previous tree. Is there anything else we can do?

upbqdn: > However, it is faster to compare the inner frontier if the roots aren't cached. We could fix that in the PartialEq impl.

Let's do that. I'll open an issue.

arya2: > We basically compare `[u8; 32]` https://github.com/zcashfoundation/zebra/blob/9ebd56092bcdfc1a09062e15a0574c94af37f389/zebra-chain/src/sprout/tree.rs#L100.

Ah, sorry, it's sapling/orchard, not sprout:

https://github.com/ZcashFoundation/zebra/blob/dedup-tree-insertion-fin-state/zebra-chain/src/sapling/tree.rs#L109

https://github.com/ZcashFoundation/zebra/blob/dedup-tree-insertion-fin-state/zebra-chain/src/orchard/tree.rs#L116

The sapling `From` impl calls `Scalar::montgomery_reduce()` and orchard's calls `Fp::montgomery_reduce()`.

arya2: > I couldn't come up with anything better than comparing the tree being stored to the previous tree. Is there anything else we can do?

~It could always clone and insert the value in the `add_x_tree_and_anchors()` methods, even when it's the same as the previous tree. The `BTreeMap`s for `x_trees_by_height` would use a bit more memory, but the value would be an `Arc` referencing the previous tree.~

Update: It could also use [`Arc::ptr_eq`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq) or [`Arc::as_ptr()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.as_ptr). (It's not possible to add a commitment and get the same tree, and it will only create a new `Arc` when there are new note commitments for that pool, see https://github.com/ZcashFoundation/zebra/pull/7259#discussion_r1268735828)

Update: Ignore my first suggestion, I hadn't considered that deduplication has to happen before writing to disk.

teor2345: A check for `Arc` equality works if they are equal, but if the `Arc`s aren't equal, we still need to compare either the cached root or entire frontier.

This logic is already implemented in `Arc::eq()`, so we don't need to do anything extra to take advantage of it:
https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-PartialEq%3CArc%3CT%3E%3E-for-Arc%3CT%3E

We can compare `Arc<Tree>` or `Arc<Root>` and it will just work.

arya2: > we still need to compare either the cached root or entire frontier.

I think it should be okay to only check if it's stored in the same allocation because it should always be equal when there are no new note commitments applied to the previous tree or unequal otherwise.

teor2345: It's possible to create a fresh clone of a note commitment tree from the finalized state at any time. If that tree has the same notes as an existing tree, it must compare equal, but it won't be in the same `Arc`.

(It's also possible to clone the inner value of an `Arc` into a new allocation using the `Arc` methods.)

upbqdn: Thanks for this discussion.

PR #7241, opened 2023-07-18
change(deps): Change dependabot actions schedule and tweak groups

teor2345:
## Motivation

Currently we open Rust and GitHub Actions dependabot PRs on the same day, but it would be easier to review them in two separate groups. It would also stop us hitting Google Cloud rate limits.

serde_json is in two groups, so it opens two PRs when it updates.

## Review

This is a routine fix.

### Reviewer Checklist


arya2 (approved):
Thank you!

PR #7247, opened 2023-07-18
change(deps): Use larger dependabot groups

teor2345:
## Motivation

I'm seeing multiple rate-limit errors in CI, I think we can fix them by having fewer dependabot PRs.

## Solution

Use larger dependabot groups

We could also reduce the number of dependabot PRs that get opened at one time.

## Review

This is a low priority devops fix. We might want to wait to merge it until after most of the open dependency PRs merge. Otherwise it will close and reopen some of them.

### Reviewer Checklist

## Follow Up Work

We could also stop dependabot auto-rebasing PRs, but that's much less convenient.

We might be able to find other ways of reducing our GitHub actions or Google Cloud jobs. For example, doing more quick tests in the same job or step.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,8 +7,8 @@ updates:
       interval: weekly
       day: monday
       timezone: America/New_York
-    # Limit dependabot to 2 PRs per reviewer, but assume one reviewer is busy or away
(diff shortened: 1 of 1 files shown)


upbqdn (approved):
utACK

PR #7255, opened 2023-07-18
build(deps): Update dependencies for release

arya2:
## Motivation

We want to update Zebra's dependencies for the release.

Part of #7136.

## Solution

- Run `cargo update`
- Update deny.toml

## Review

Anyone can review.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #7257, opened 2023-07-18
change(deps): Remove OpenSSL dependency from dev tools

teor2345:
## Motivation

PR #7047 removed OpenSSL in experimental features for security reasons.

This PR removes it entirely, which reduces the dependency tree and security alerts.

Depends-On: #7255

## Solution

Switch from OpenSSL to rustls in Zebra dev tools.

## Review

### Reviewer Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -884,16 +884,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "21a53c0a4d288377e7415b53dcfc3c04da5cdc2cc95c8d5ac178b58f0b861ad6"
 
 [[package]]
-name = "core-foundation"
-version = "0.9.3"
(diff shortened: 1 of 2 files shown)


teor2345:
This PR will probably conflict with PR #7255 because it removes some of the same dependencies.

I can update the lock file after PR #7255 merges.

arya2 (dismissed):

teor2345:
> I can update the lock file after PR #7255 merges.

This is done now.

arya2 (approved):

arya2:
Great catch by the way!

teor2345:
This got auto-labelled with "extra-reviews" because I said "remove" in the title. But it's developer-only code and a dependency, so it doesn't need multiple reviews.

Maybe we could turn that auto-label off 😶‍🌫️

PR #7262, opened 2023-07-19
change(test): Refactor the tests of non-finalized state

upbqdn:
## Motivation

I noticed some inaccuracies in our tests of the non-finalized state when I was looking at the test coverage of #7239.

## Solution

- Allow `proptest` to generate blocks containing only the coinbase tx.
- Don't skip some checks of the genesis block in one of the tests.

### Reviewer Checklist


teor2345 (approved):
Thanks!

PR #7265, opened 2023-07-19
fix(docker): Fix confusing Docker git commit hash default in Zebra logs

teor2345:
## Motivation

PR #7200 adds an incorrect default for the git commit hash in Zebra's logs in Docker builds.

This bug doesn't have a significant impact because vergen uses the `.git` directory, which is currently ignored by `.dockerignore`. So we wouldn't get a fallback hash anyway.

But this change logs "zebrad=unknown" on every single Zebra log line, which could be confusing for users.

## Solution

- Use an empty string if the git commit hash is not specified in Docker

This change will not be user visible if we get this fix merged before the release.

## Review

@arya2 I'm not sure if this is a release blocker or not? What do you think?

### Reviewer Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -80,7 +80,9 @@ ARG COLORBT_SHOW_HIDDEN
 ENV COLORBT_SHOW_HIDDEN=${COLORBT_SHOW_HIDDEN:-1}
 
 ARG SHORT_SHA
-ENV SHORT_SHA=${SHORT_SHA:-unknown}
+# If this is not set, it must be the empty string, so Zebra can try an alternative git commit source:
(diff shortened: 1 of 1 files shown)


gustavovalverde:
PR https://github.com/ZcashFoundation/zebra/pull/7200 added a default, but the hash hasn't been available for a while in the deployed instances.

As we're passing this as build ARG for all images,  I wonder why it's not being set. I'd like to test this before merging.

teor2345:
> PR #7200 added a default, but the hash hasn't been available for a while in the deployed instances.

I've seen that it's missing, but I just wanted to fix the confusing logs in this PR. Because we might want that fix in the 1.1.0 release.

I've updated the PR title and description so it's clearer.

> As we're passing this as build ARG for all images, I wonder why it's not being set. I'd like to test this before merging.

Would you mind if we fixed the missing commit hash in another PR? It doesn't need to go in this release.

gustavovalverde (approved):

PR #7266, opened 2023-07-19
change(state):  Insert only the first tree in each series of identical trees into finalized state

### LINKED ISSUE #4784: Store only the first tree state in each series of identical tree states

upbqdn:
## Motivation

Address part of #4784.

## Solution

- Insert only the first Sapling and Orchard tree in each series of identical trees into the finalized state.
- We don't store Sprout trees by height, only a single tree for the tip, so this PR doesn't optimize the storing of Sprout trees.
- Update snapshots.
- Simplify the methods that handle the trees.
- Correctly retrieve the trees when retrieving by hash or height.

### Reviewer Checklist


zebra-state/src/service/finalized_state/zebra_db/block/tests/snapshots/orchard_trees@mainnet_2.snap
```diff
@@ -13,9 +13,7 @@ expression: stored_orchard_trees
     inner: Frontier(
       frontier: None,
     ),
-    cached_root: Some(Root(Base(
-      bytes: (174, 41, 53, 241, 223, 216, 162, 74, 237, 124, 112, 223, 125, 227, 166, 104, 235, 122, 73, 177, 49, 152, 128, 221, 226, 187, 217, 3, 26, 229, 216, 47),
-    ))),
+    cached_root: None,
```
teor2345: Optional: It might reduce CPU usage if we didn't calculate the cached root for block 2. Is that calculation happening in the state production code, or snapshot test code?

(It's not being stored in the raw data, so it's not a disk usage issue, and the tree is the same, so it's not a consensus issue.)

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -147,34 +147,28 @@ impl ZebraDb {
...
         &self,
         hash_or_height: HashOrHeight,
     ) -> Option<Arc<orchard::tree::NoteCommitmentTree>> {
         let height = hash_or_height.height_or_else(|hash| self.height(hash))?;
 
-        let orchard_tree_handle = self.db.cf_handle("orchard_note_commitment_tree").unwrap();
-
-        self.db.zs_get(&orchard_tree_handle, &height)
+        self.orchard_tree_by_height(&height)
     }
```
teor2345, line 174: Thanks for catching this! I didn't notice this direct access to these column families when I made the compatibility fixes for this format change in `Zebra 1.0.0-rc.9`.

What this means is that the `z_gettreestate` RPC can return incorrect responses with missing trees if:
1. A state is upgraded by Zebra 1.2.0 (assuming this PR is merged after the next release), then
2. A previous Zebra version is launched with that upgraded state

That's ok because backwards compatibility is best-effort, so we can just add a note to the changelog.

We could make this better by splitting this PR, merging these compatibility fixes in `1.1.0`, and then making the format changes in `1.2.0`. So we could tell users `1.1.0` is compatible, but older versions aren't.

Do you think it is worth the effort of doing this?

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -275,60 +275,56 @@ impl DiskWriteBatch {
...
 
-        self.zs_insert(
-            &orchard_note_commitment_tree_cf,
-            height,
-            note_commitment_trees.orchard,
-        );
+        // Store the Orchard tree only if it is not already present at the previous height.
+        if height.is_min() || zebra_db.orchard_tree() != trees.orchard {
+            self.zs_insert(&orchard_tree_cf, height, trees.orchard);
+        }
```
teor2345: This is a state format change for new blocks, so let's bump the state minor version?
https://github.com/ZcashFoundation/zebra/blob/9ebd56092bcdfc1a09062e15a0574c94af37f389/zebra-state/src/constants.rs#L44-L55

teor2345 (dismissed):
Thanks for making this change, it looks great!

Do you think it's worth splitting this PR so Zebra `1.2.0` states will be compatible with older Zebra `1.2.0` versions?

zebra-state/src/service/finalized_state/zebra_db/block.rs
teor2345: I'd also like to get @arya2's feedback here, because if we split the PR, the compatibility fix becomes a release blocker.

zebra-state/src/service/finalized_state/zebra_db/block/tests/snapshots/orchard_trees@mainnet_2.snap
teor2345: Just to be clear, I'm not concerned about CPU usage in tests!

zebra-state/src/service/finalized_state/zebra_db/block.rs
arya2: I think we should instead add config/CLI options behind a compilation feature for ignoring optional format upgrades and a subcommand for migrating down for users intending to switch back to an older version of Zebra.

Maybe Zebra could ask for confirmation before format upgrades too.

> Do you think it is worth the effort of doing this?

I think it's worth it if this is to be a required format upgrade in v1.2.0.

teor2345: > > Do you think it is worth the effort of doing this?
> 
> I think it's worth it if this is to be a required format upgrade in v1.2.0.

Just checking we're talking about the same thing here: this PR currently contains a required format upgrade.

After it is merged, if blocks are finalized by a newer Zebra version, then read by an older Zebra version, some z_gettreestate responses will be incorrect.

So this is a question that I'd like to answer before the `1.1.0` release:

> > We could make this better by splitting this PR, merging these compatibility fixes in 1.1.0, and then making the format changes in 1.2.0. So we could tell users 1.1.0 is compatible, but older versions aren't.

I would be ok with just breaking compatibility of new states with old Zebra versions, and adding a note to the changelog, because running a newer state with an older Zebra version is an uncommon use case.

So do we want to split this PR and wait for a compatibility fix to merge before tagging the release, or just merge this PR and break compatibility for an uncommon use case?

(I asked Arya because he is the release manager and so he might need to delay the release.)

-----

This is a question that might be able to wait for a later release:

> I think we should instead add config/CLI options behind a compilation feature for ignoring optional format upgrades and a subcommand for migrating down for users intending to switch back to an older version of Zebra.
> 
> Maybe Zebra could ask for confirmation before format upgrades too.

All of our scheduled format upgrades are required to support the "spend before sync" RPCs and get reasonable performance. Since there are no optional format upgrades yet, can this design wait for the first release where there is an optional format upgrade?

At that time, we might want to consider how much benefit users get from compatibility with older Zebra versions, the extra implementation/test cost, ongoing maintenance cost / bug risk, and if an interactive prompt breaks existing non-interactive use cases.

teor2345: As an example, `zcashd`'s database capability flags cause significant ongoing user confusion. You can see some recent user questions about them in Discord.

So before we add optional state features or state upgrades, let's check there's a compelling use case?

zebra-state/src/service/finalized_state/zebra_db/block/tests/snapshots/orchard_trees@mainnet_2.snap
teor2345: I think this is ok, it's either due to the test, or due to a state API where the root is used.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/orchard_note_commitment_tree_raw_data@mainnet_1.snap
```diff
@@ -7,8 +7,4 @@ expression: cf_data
     k: "000000",
     v: "0000",
   ),
```
teor2345: Optional: it might reduce CPU usage to calculate and cache the root hash of the empty tree in the genesis block. Currently, every time the empty tree is accessed, the RPCs will have to calculate the root.

This can be implemented by calling `tree.root()` before writing the genesis tree to the state.

We could make this change for new trees in this PR, and by overwriting the genesis entry in the PR that upgrades existing states.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/orchard_note_commitment_tree_raw_data@mainnet_2.snap
```diff
@@ -7,12 +7,4 @@ expression: cf_data
...
     v: "0000",
   ),
-  KV(
-    k: "000001",
-    v: "0001ae2935f1dfd8a24aed7c70df7de3a668eb7a49b1319880dde2bbd9031ae5d82f",
-  ),
-  KV(
-    k: "000002",
-    v: "0001ae2935f1dfd8a24aed7c70df7de3a668eb7a49b1319880dde2bbd9031ae5d82f",
-  ),
```
teor2345: This is the raw data snapshot. (Please feel free to resolve this note.)

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/orchard_note_commitment_tree_raw_data@mainnet_1.snap
upbqdn: We compute (and cache) the roots of the trees of the genesis block already, but then we store `default()` trees anyway: https://github.com/zcashfoundation/zebra/blob/9ebd56092bcdfc1a09062e15a0574c94af37f389/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L365-L379. It should be an easy fix. ~I'll open an issue.~

teor2345: This is technically a database format change, but since the deserialized trees are indistinguishable, I think we can do the fix at any time.

teor2345: Maybe with a patch format version 🙂

upbqdn: I just realized I misread the last sentence in your comment. I'll do it in this PR.

zebra-state/src/service/finalized_state/zebra_db/block.rs
arya2: > Just checking we're talking about the same thing here: this PR currently contains a required format upgrade.

I think it's worth the effort to split up this PR so the format upgrade is backwards compatible with v1.1.0 if this is to be a required format upgrade.

But I think we should make this an optional format upgrade with a CLI option and/or config field behind a compilation feature instead. Perhaps requiring user confirmation just before beginning the migration.

teor2345: > But I think we should make this an optional format upgrade with a CLI option and/or config field behind a compilation feature instead.

I'm a bit confused by this discussion, I feel like we're working towards different goals here. What goals are satisfied by making format upgrades optional and interactive? Are these goals important to users? Are they required for the 1.2.0 release?

We haven't scoped this work, so there aren't any tickets for it. It would significantly increase the complexity of our code, and our build and test matrixes, and ongoing maintenance. So I'm looking for a use case that justifies this extra effort.

> Perhaps requiring user confirmation just before beginning the migration.

Zebra currently doesn't ask for user input, so this would be a breaking change for all non-interactive use cases. I think we should wait for a major version release to break these common use cases.

-----

Here are the goals of requiring the two currently scheduled format upgrades:
* support "spend before sync" RPCs
* fast creation of the new database column family for the "z_getsubtreeindexes" RPC

They are important to users because they significantly improve light wallet sync time. They are required for the 1.2.0 or 1.3.0 release because zcashd already supports these RPC fields. So light wallet server operators and users are less likely to migrate to Zebra until they are done.

The user experience of requiring these changes is that most users won't notice. If users upgrade Zebra to 1.2.0 then downgrade to 1.0.1 or earlier, then some of the outputs of the "z_gettreestate" RPC might be wrong. But since we currently have 1 known user running a light wallet server, we can just tell them not to do that.

teor2345: I'm going to hand this over to @oxarbitrage and @mpguerra for a final decision, hopefully you can all discuss it at the engineering sync next week.

arya2: > We haven't scoped this work, so there aren't any tickets for it.

I think it may be worth the effort to split this PR so v1.1.0 works with the new db format in that case.

Update: Or to add a subcommand for downgrading.

teor2345: > > We haven't scoped this work, so there aren't any tickets for it.
> 
> I think it may be worth the effort to split this PR so v1.1.0 works with the new db format in that case.

Yep, I think that's the only decision we have to make for 1.1.0.

All the other decisions can wait for 1.2.0 or later.

zebra-state/src/service/finalized_state/zebra_db/block/tests/snapshots/orchard_trees@mainnet_2.snap
upbqdn: > Is that calculation happening in the state production code, or snapshot test code?

We calculate the roots of each tree for each block, including genesis, in the state. We also use the cached roots multiple times in the state. The main branch currently doesn't store the roots for the genesis block but does store them for each subsequent block.

This PR doesn't store the roots (more precisely, the trees) if they match the previous ones, but the state still calculates them and uses them. However, the root for height two in `orchard_trees@mainnet_2.snap` shows up because it's calculated in the test code, not in the state, as you pointed out! Huge thanks for that.

So the answer to the question is that the calculation happens in both the state production code and snapshot test code. Since this PR doesn't currently store effectively any Sapling and Orchard roots in the database until their activation, the state would need to recompute the genesis root for each block until their first change. This would be a performance regression for the initial sync, and we should store the genesis roots in the database, as you also pointed out!

zebra-state/src/service/finalized_state/zebra_db/block.rs
mpguerra: If possible, let's do the simplest and quickest thing for now to not block the next release and to give us time for a more in depth discussion later.

upbqdn: I suggest we just merge this PR as is without splitting it.

Reasons:

- It is the simplest and quickest solution.
- The incompatibility impact is very low since someone would have to run a new Zebra version, which wouldn't insert some trees into the database, and then start an older version using the same database. This older version would end up returning `None` for heights for which the trees were not inserted into the database in the `z_gettreestate` RPC.
- This behavior is not unexpected since, as Teor said, backward compatibility is best-effort.

Edit: Trees wouldn't be removed from the database but some new trees wouldn't be inserted.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -275,60 +275,56 @@ impl DiskWriteBatch {
...
-        );
+        self.zs_insert(&sprout_tree_cf, height, trees.sprout);
 
-        self.zs_insert(
-            &sapling_note_commitment_tree_cf,
-            height,
-            note_commitment_trees.sapling,
-        );
+        // Store the Sapling tree only if it is not already present at the previous height.
+        if height.is_min() || zebra_db.sapling_tree() != trees.sapling {
```
arya2: Optional: we could cache the previous note commitment trees and pass them in as an argument here, see #7276.

mpguerra:
this one seems ready to merge? @upbqdn I'll leave it up to you how you want to resolve the final comments :)

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Merged in [5c022aa](https://github.com/ZcashFoundation/zebra/pull/7266/commits/5c022aa99aefbf87be3baa3fb07eec970bbddd55).

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Done in [ef6a863](https://github.com/ZcashFoundation/zebra/pull/7266/commits/ef6a863d9ec12930ae14fc628d6aa9a3a1fa2176).

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/orchard_note_commitment_tree_raw_data@mainnet_1.snap
upbqdn: Done in [e19f365](https://github.com/ZcashFoundation/zebra/pull/7266/commits/e19f365d98f240bdfbe7a570bbf65059945e532d).

upbqdn:
All comments are addressed. I noticed there's a bug in the new code that upgrades the database. I'll push a fix in a separate PR.

upbqdn:
I added a check that checks if the tree roots of the genesis block match the roots of empty trees, and I noticed our proptests currently generate genesis blocks with shielded data, so the trees aren't empty. I'll push a fix in a separate PR.

arya2:
This again:

>   Message:  can't upgrade a database that has already been upgraded, or is newer:
>   disk: 25.0.2
>    upgrade: 25.0.2
>    running: 25.1.0
>    Location: zebra-state/src/service/finalized_state/disk_format/upgrade.rs:390
> 
> https://github.com/ZcashFoundation/zebra/actions/runs/5800049030/job/15723531329?pr=7266#step:8:206

upbqdn:
That should stop after #7304.

arya2 (approved):

arya2:
@Mergifyio update

PR #7268, opened 2023-07-20
fix(feature): elasticsearch for the Mainnet

### LINKED ISSUE #7060: zebrad with elasticsearch sync error
I' m sure you will need more info, here is a start

## Error
```
ES response parsing to a json_body should never fail: Error { kind: Http(reqwest::Error { kind: Decode, source: Error("EOF while parsing a value", line: 1, column: 0) }) }
```

## Metadata
|key|value|
|--|--|
|**version**|1.0.0|
|**Zcash network**|Mainnet|
|**state version**|25|
|**features**|default,elasticsearch,release_max_level_info|
|**branch**|master|
|**git commit**|0e0ee8d|
|**commit timestamp**|2023-06-23T22:17:36.000000000Z|
|**rust compiler**|1.70.0|
|**rust release date**|2023-05-31|
|**optimization level**|3|
|**debug checks**|false|
|**location**|zebra-state/src/service/finalized_state.rs:416:26|

oxarbitrage:
## Motivation

The elasticsearch zebra feature has problems in the Mainnet while it can sync up to tip in the Testnet.

Fix https://github.com/ZcashFoundation/zebra/issues/7060
Part of #7270.

## Solution

The problem is that the amount of data we are sending in bulk calls is greater than 100M. We were sending 400 blocks at each bulk and this PR reduce this to 50.

There is no recommendation about how much bulk data is good enough so in general we have to look for a balance. Querying it too often will slow down the process and too far away means more data until the limit is reached.

The elasticsearch suggestion is to reduce bulk size and do not increase the limit: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html (see http.max_content_length)

This PR reduce the number of blocks sent to elastic from 400 to 50. With that i can synchronize much more than i was before, i am unsure of the performance decrease. I got that number assuming all blocks are full which i think could happen in spam times.

If the limit is too tight and fails then i think a change to 49 or 48 blocks per block will make it because i am not considering a very small header line added on top of each block.

It seems elasticsearch allow the bulk to be compressed (for example https://github.com/elastic/go-elasticsearch/issues/298#issuecomment-876595767) which can help with the performance however i had not explored it yet.

I think we might want to ask the bug report author (@dismad) to try again after we get this merged.

## Review

Happy to get feedback and make changes. I think anyone can review.

### Reviewer Checklist


zebra-state/src/service/finalized_state.rs
```diff
@@ -373,12 +373,30 @@ impl FinalizedState {
             let block_time = block.header.time.timestamp();
             let local_time = chrono::Utc::now().timestamp();
 
-            const AWAY_FROM_TIP_BULK_SIZE: usize = 800;
+            // Mainnet bulk size is small enough to avoid the elasticsearch 100mb content
+            // length limitation. MAX_BLOCK_BYTES = 2M. Each block count as 2 as we send
+            // them with a operation/header line. A value of 100 is 50 blocks.
+            const MAINNET_AWAY_FROM_TIP_BULK_SIZE: usize = 100;
```
teor2345: Block data is hex-encoded, so a 2MB binary block actually creates 4MB of JSON. Some block and transaction data is repeated multiple times in JSON, so I'd suggest allowing a little bit more space as well.

```suggestion
            const MAINNET_AWAY_FROM_TIP_BULK_SIZE: usize = 45;
```

zebra-state/src/service/finalized_state.rs
```diff
@@ -373,12 +373,30 @@ impl FinalizedState {
...
+
+            // Testnet bulk size is larger as blocks are generally smaller in the testnet.
+            // A value of 800 is 400 blocks as we are not counting the operation line.
+            const TESTNET_AWAY_FROM_TIP_BULK_SIZE: usize = 800;
```
teor2345, line 384: This could break if someone adds 50 large blocks to testnet, which could happen at any time. So we could just use the same limit as mainnet?

```
            const TESTNET_AWAY_FROM_TIP_BULK_SIZE: usize = 45;
```

oxarbitrage: I will prefer to don't make changers now in the testnet as it was working. This change affects performance.

zebra-state/src/service/finalized_state.rs
oxarbitrage: I will like to test as it is and if it not work maybe suggest the users to increase ` http.max_content_length` in elasticsearch to `200` or try compressing the data. 

Doing such small blocks will impact the performance of the sync (i am not really sure if the current change will make the sync in lets say 1 week, i can't imagine doing it with bulks of 20 blocks).

By the way, using an odd number there will break the program as each block is 2 lines.

oxarbitrage: Maybe i will add this value to the config so users can test and make changes without changing the code. If zebra panics and the user change the config value it can restart pretty much where it was, elastic will just replace repeated entries.

zebra-state/src/service/finalized_state.rs
teor2345: No worries, I opened #7270 so we remember to fix this security issue before we stabilise this feature in production.

dismad:
Testing now :+1:

dismad:
Progress, ran this overnight and woke up to another error, but I synced up to 79% which is an improvement.

`ES response parsing error. Maybe we are sending more than 100 mb of data (`http.max_content_length`): Error { kind: Http(reqwest::Error { kind: Decode, source: Error("EOF while parsing a value", line: 1, column: 0) }) }
`

Screenshot_2023-07-21_05-52-35

oxarbitrage:
thank you @dismad  for trying it. So i think we will have no option at least for now that to reduce the bulk size for the mainnet to just 24 blocks. Ill make the change in this PR.

That should do it, i am not sure in how much time. We can make something smarter in the future like checking the size of the collected blocks and send to elastic when we are close to be in the limit.

arya2 (approved):

oxarbitrage:
@dismad i made the change for the Mainnet, please give it a new try when you can. 

You can try with the `main` branch soon as this is going to be merged thanks to @arya2 . The merge will close the ticket but we should reopen if it is still happening.

Thanks again for your patience here.

PR #7276, opened 2023-07-21
Reduce disk reads by caching trees.

arya2:
This is a suggestion for #7266 to cache the finalized tip's note commitment trees in the write block task:

- Caches note commitment trees of finalized tip in memory when writing blocks to disk
- Skips disk reads when there's a cached copy of the note commitment trees


upbqdn (approved):
utACK. Looks great, thank you for this optimization.

PR #7288, opened 2023-07-25
fix(state): Avoid temporary failures verifying the first non-finalized block

### LINKED ISSUE #5125: Avoid temporary failures verifying the first non-finalized block

arya2:
## Motivation

This PR fixes a bug where the first block past the final checkpoint will fail verification if it arrives in zebra-state before the final checkpoint block is written to disk.

Close #5125.
Part of #5709.

Depends-On: #6810

## Solution

- Use the finalized tip hash instead of the parent hash when dropping the finalized block write sender

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -907,6 +907,14 @@ impl Service<Request> for StateService {
             } else {
                 tracing::debug!(len = ?old_len, ?tip, "no utxo requests needed pruning");
             }
+
+            if self.finalized_block_write_sender.is_some()
+                && self
+                    .non_finalized_state_queued_blocks
+                    .has_queued_children(self.finalized_block_write_last_sent_hash)
+            {
+                // TODO: process queue
+            }
         }
 
         poll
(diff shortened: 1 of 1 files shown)


zebra-state/src/service.rs
```diff
@@ -689,34 +689,7 @@ impl StateService {
...
-            self.prune_by_height();
-            // Mark `SentHashes` as usable by the `can_fork_chain_at()` method.
-            self.block_write_sent_hashes.can_fork_chain_at_hashes = true;
-
-            // We've finished committing checkpoint verified blocks to finalized state, so drop any repeated queued blocks.
-            self.clear_finalized_block_queue(
-                "already finished committing checkpoint verified blocks: dropped duplicate block, \
-                 block is already committed to the state",
-            );
-        }
```
arya2, line 719: Every line in this block was moved to new methods below.

arya2:
I'll do this in https://github.com/ZcashFoundation/zebra/pull/6810 instead.

Issue #7118, opened 2023-06-30
Log a warning when the loaded config includes fields used by elasticsearch feature


arya2:
## Motivation

Zebra currently warns about unused config fields for the flamegraph and metrics features, but panics when compiled without the elasticsearch feature and trying to a load a configuration file with a elasticsearch fields

The existing warnings are logged once Zebra instantiates the components that would use them, they may be more noticeable if logged immediately after it has loaded the config file.

## Possible Design

- Remove feature flags on elasticsearch config fields
- Log warnings about unused config fields after loading the config, either:
    - while loading the config
    - in [`Component::after_config()`](https://docs.rs/abscissa_core/0.7.0/abscissa_core/component/trait.Component.html#method.after_config) on each component
    - in `register_components()` after `load_config()` (maybe in methods on Config structs)
    - in any other code that's shortly after config loading

### Optional Changes

  - Move warnings about unused config fields for flamegraph or prometheus after loading the config

## Drawbacks

Warning and ignoring the unused config fields would be better for some users.

But other users who expect Zebra to fail if the Zebra build doesn't support the config, or the config fields are wrong, will be surprised.

teor2345:
What about the filter-reload feature?

I'd like to split this ticket into separate tickets for mining and elasticsearch, because they are likely to stabilise at different times.

arya2:
> What about the filter-reload feature?

It looks like it would log a warning and skip the filter-reload feature: 

https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/components/tracing/endpoint.rs#L40
https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/components/tracing/endpoint.rs#L60

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7118?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

teor2345:
> What about the filter-reload feature?
> 
> I'd like to split this ticket into separate tickets for mining and elasticsearch, because they are likely to stabilise at different times.

@mpguerra since we're about to stabilise `getblocktemplate-rpcs`, I'd like to remove it from this ticket.

mpguerra:
un-scheduling for now

PR #7290, opened 2023-07-25
fix(panic): Log a warning instead of panicking for unused mining configs

arya2:
## Motivation

This PR will log a warning if there is a non-default mining section in the loaded config when compiled without the "getblocktemplate-rpcs" feature, instead of panicking.

Part of #7118.

## Solution

- Move the getblocktemplate config to `config/mining.rs`
- Log a warning when there's a mining section in a config without the getblocktemplate-rpcs feature

## Review

Anyone can review.

### Reviewer Checklist


oxarbitrage:
@arya2 do you need someone to finish this work ? I can find some time to bump it up if you want.

arya2:
> @arya2 do you need someone to finish this work ? I can find some time to bump it up if you want.

Yes, please.

teor2345:
If we make `getblocktemplate-rpcs` into a default feature, this might change priority.

teor2345 (approved):
Looks good, let's get it merged before any more RPC work.

I committed a quick tweak to the mining config argument passing, because we don't need to pass `()` now the confit is always enabled.

PR #7302, opened 2023-08-07
fix(test): Don't include shielded data in genesis blocks

upbqdn:
## Motivation

Our proptests were generating genesis blocks containing transactions with joinsplits and shielded data. PR #7266 asserts that the note commitment trees associated with the genesis block are equal to the empty trees. These assertions were failing in that PR because the note commitment trees for genesis blocks weren't empty.

## Solution

- Do not include joinsplits nor shielded data in V4 transactions in the test genesis block.
- Do not include shielded data in V5 transactions in the test genesis block.

### Reviewer Checklist


arya2 (approved):
This looks great! Let's merge this into `dedup-tree-insertion-fin-state` to fix those tests.

The disk format upgrade panicked here and possibly elsewhere:

> can't upgrade a database that has already been upgraded, or is newer

https://github.com/ZcashFoundation/zebra/actions/runs/5798021847/job/15717359090?pr=7302#step:8:215

upbqdn:
Thanks for looking at the tests. I forgot to mention that the bug that this PR addresses is gone—an example of the bug in #7266: https://github.com/ZcashFoundation/zebra/actions/runs/5798018815/job/15717022677.

This PR fails because of an unrelated bug that I'll fix in another PR.

PR #7309, opened 2023-08-09
fix(ci): Specify Testnet in jobs that require it

upbqdn:
## Motivation

The `generate-checkpoints-testnet` job currently fails with https://github.com/ZcashFoundation/zebra/actions/runs/5803425199/job/15732737922 for all PRs that run it.

Note that the equivalent job for Mainnet works fine.

## Solution

The issue seems to be caused by not specifying "Testnet" in the job. This PR fixes that.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -560,6 +560,7 @@ jobs:
       test_id: generate-checkpoints-testnet
       test_description: Generate Zebra checkpoints on testnet
       test_variables: '-e NETWORK=Testnet -e GENERATE_CHECKPOINTS_TESTNET=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache'
+      network: "Testnet"
       needs_zebra_state: true
       # update the disk on every PR, to increase CI speed
       # we don't have a test-update-sync-testnet job, so we need to update the disk here
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -516,6 +516,7 @@ jobs:
       test_description: Test a full sync up to the tip on testnet
       # The value of FULL_SYNC_TESTNET_TIMEOUT_MINUTES is currently ignored.
       test_variables: '-e NETWORK=Testnet -e FULL_SYNC_TESTNET_TIMEOUT_MINUTES=0 -e ZEBRA_FORCE_USE_COLOR=1'
+      network: "Testnet"
       # A full testnet sync could take 2-10 hours in April 2023.
       # The time varies a lot due to the small number of nodes.
       is_long_test: true


.github/workflows/continous-integration-docker.yml
```diff
@@ -560,6 +560,7 @@ jobs:
       test_id: generate-checkpoints-testnet
       test_description: Generate Zebra checkpoints on testnet
       test_variables: '-e NETWORK=Testnet -e GENERATE_CHECKPOINTS_TESTNET=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache'
+      network: "Testnet"
```
gustavovalverde, line 564: I wonder why this is not an issue with `full-sync-to-tip-testnet` or should the fix be applied there too?

gustavovalverde (changes requested):

.github/workflows/continous-integration-docker.yml
upbqdn: Is there a way to find any recent runs of that job? I would check the logs for it.

upbqdn: It looks like there are no recent runs of that job. I think it is likely it will have the same issue.

gustavovalverde: Not an easy way, sadly (I'll be refactoring this to be easier in the following sprints).

But I've just realized we must apply this fix on `full-sync-to-tip-testnet` as we're saving to the disk there, based on this parameter `saves_to_disk: true`.

Which then would corrupt the `Mainnet` disk if this parameter is not specified with the right Network:
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/deploy-gcp-tests.yml#L1734-L1739

upbqdn: Done in [16899b9](https://github.com/ZcashFoundation/zebra/pull/7309/commits/16899b97a7e49a2d594e41f22a62b9218bc42497).

gustavovalverde (approved):
LGTM

oxarbitrage:
Thanks! Waiting for this one to be merged to merge other PRs.

oxarbitrage:
I think this maybe [failed](https://github.com/ZcashFoundation/zebra/actions/runs/5807794507/job/15758325346?pr=7309) because we are building a new lightwalletd cache [here](https://github.com/ZcashFoundation/zebra/actions/runs/5811606354/job/15757542745).

Sorry about that, I will retry this PR after that task finalize.

upbqdn:
We are. The previous workflow got canceled after 6 hours. I re-run it assuming it would pick up the previous state.

gustavovalverde:
I'm re-running the action as it was wrongly triggered the last time: https://github.com/ZcashFoundation/zebra/actions/runs/5807794507/job/15774092033?pr=7309

gustavovalverde:
@mergify refresh

PR #7312, opened 2023-08-09
change(state): Deduplicate note commitment trees stored in the finalized state

### LINKED ISSUE #4784: Store only the first tree state in each series of identical tree states

upbqdn:
## Motivation

Close #4784.

Previous PRs that address parts of #4784:

- #7239.
- #7266.

This is a final PR that prunes the trees stored in the database.

## Solution

- Add support for deleting Sapling and Orchard trees in `DiskWriteBatch`.
- Prune duplicate Sapling and Orchard trees starting from the genesis block to the initial tip.
- Zebra doesn't store Sprout trees by height, it keeps only a single tree for the tip, so this PR doesn't prune Sprout trees.
- I kept the database version at 25.1.0, which was introduced in #7266. The reason is that both this PR and #7266 leave the database in the same format&mdash;this PR affects already stored trees, and #7266 affects incoming trees.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -113,8 +113,14 @@ pub trait WriteDisk {
         K: IntoDisk + Debug,
         V: IntoDisk;
 
-    /// Remove the given key form rocksdb column family if it exists.
+    /// Remove the given key from rocksdb column family if it exists.
     fn zs_delete<C, K>(&mut self, cf: &C, key: K)
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + Debug;
+
+    /// Remove the given key range from rocksdb column family if it exists.
+    fn zs_delete_range<C, K>(&mut self, cf: &C, from: K, to: K)
     where
         C: rocksdb::AsColumnFamilyRef,
         K: IntoDisk + Debug;
@@ -139,6 +145,16 @@ impl WriteDisk for DiskWriteBatch {
     {
         let key_bytes = key.as_bytes();
         self.batch.delete_cf(cf, key_bytes);
+    }
+
+    fn zs_delete_range<C, K>(&mut self, cf: &C, from: K, to: K)
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + Debug,
+    {
+        let from_bytes = from.as_bytes();
+        let to_bytes = to.as_bytes();
+        self.batch.delete_range_cf(cf, from_bytes, to_bytes);
     }
 }
 
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -94,12 +94,10 @@ pub struct DiskDb {
 //       (DiskDb can be cloned),
 //       and make them accessible via read-only methods
 #[must_use = "batches must be written to the database"]
+#[derive(Default)]
 pub struct DiskWriteBatch {
     /// The inner RocksDB write batch.
     batch: rocksdb::WriteBatch,
-
-    /// The configured network.
-    network: Network,
 }
 
 /// Helper trait for inserting (Key, Value) pairs into rocksdb with a consistently
@@ -395,16 +393,10 @@ impl DiskWriteBatch {
     /// Each block must be written to the state inside a batch, so that:
     /// - concurrent `ReadStateService` queries don't see half-written blocks, and
     /// - if Zebra calls `exit`, panics, or crashes, half-written blocks are rolled back.
-    pub fn new(network: Network) -> Self {
+    pub fn new() -> Self {
         DiskWriteBatch {
             batch: rocksdb::WriteBatch::default(),
-            network,
         }
-    }
-
-    /// Returns the configured network for this write batch.
-    pub fn network(&self) -> Network {
-        self.network
     }
 }
 
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -420,7 +420,7 @@ impl DiskDb {
     /// Returns an iterator over the items in `cf` in `range`.
     ///
     /// Holding this iterator open might delay block commit transactions.
-    fn zs_range_iter<C, K, V, R>(&self, cf: &C, range: R) -> impl Iterator<Item = (K, V)> + '_
+    pub fn zs_range_iter<C, K, V, R>(&self, cf: &C, range: R) -> impl Iterator<Item = (K, V)> + '_
     where
         C: rocksdb::AsColumnFamilyRef,
         K: IntoDisk + FromDisk,
@@ -465,7 +465,7 @@ impl DiskDb {
             .map(|result| result.expect("unexpected database failure"))
             .map(|(key, value)| (key.to_vec(), value))
             // Handle Excluded start and the end bound
-            .filter(move |(key, _value)| range.contains(key))
+            .take_while(move |(key, _value)| range.contains(key))
             .map(|(key, value)| (K::from_bytes(key), V::from_bytes(value)))
     }
 
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -308,7 +308,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.sapling_tree_by_height_range(sapling_height..initial_tip_height)
                 {
-                    let _ = sapling_tree_tx.send((height, tree));
+                    let _ = sapling_tree_tx
+                        .send((height, tree))
+                        .map_err(|error| warn!(?error, "unexpected send error"));
                 }
             });
 
@@ -320,7 +322,9 @@ impl DbFormatChange {
                 while let Ok((height, tree)) = sapling_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_sapling_tree != tree {
-                        let _ = unique_sapling_tree_height_tx.send(height);
+                        let _ = unique_sapling_tree_height_tx
+                            .send(height)
+                            .map_err(|error| warn!(?error, "unexpected send error"));
                         prev_sapling_tree = tree;
                     }
                 }
@@ -361,7 +365,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.orchard_tree_by_height_range(orchard_height..initial_tip_height)
                 {
-                    let _ = orchard_tree_tx.send((height, tree));
+                    let _ = orchard_tree_tx
+                        .send((height, tree))
+                        .map_err(|error| warn!(?error, "unexpected send error"));
                 }
             });
 
@@ -373,7 +379,9 @@ impl DbFormatChange {
                 while let Ok((height, tree)) = orchard_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_orchard_tree != tree {
-                        let _ = unique_orchard_tree_height_tx.send(height);
+                        let _ = unique_orchard_tree_height_tx
+                            .send(height)
+                            .map_err(|error| warn!(?error, "unexpected send error"));
                         prev_orchard_tree = tree;
                     }
                 }
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -306,9 +306,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.sapling_tree_by_height_range(sapling_height..initial_tip_height)
                 {
-                    let _ = sapling_tree_tx
-                        .send((height, tree))
-                        .map_err(|error| warn!(?error, "unexpected send error"));
+                    if let Err(error) = sapling_tree_tx.send((height, tree)) {
+                        warn!(?error, "unexpected send error")
+                    }
                 }
             });
 
@@ -320,9 +320,9 @@ impl DbFormatChange {
                 while let Ok((height, tree)) = sapling_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_sapling_tree != tree {
-                        let _ = unique_sapling_tree_height_tx
-                            .send(height)
-                            .map_err(|error| warn!(?error, "unexpected send error"));
+                        if let Err(error) = unique_sapling_tree_height_tx.send(height) {
+                            warn!(?error, "unexpected send error")
+                        }
                         prev_sapling_tree = tree;
                     }
                 }
@@ -361,9 +361,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.orchard_tree_by_height_range(orchard_height..initial_tip_height)
                 {
-                    let _ = orchard_tree_tx
-                        .send((height, tree))
-                        .map_err(|error| warn!(?error, "unexpected send error"));
+                    if let Err(error) = orchard_tree_tx.send((height, tree)) {
+                        warn!(?error, "unexpected send error")
+                    }
                 }
             });
 
@@ -375,9 +375,10 @@ impl DbFormatChange {
                 while let Ok((height, tree)) = orchard_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_orchard_tree != tree {
-                        let _ = unique_orchard_tree_height_tx
-                            .send(height)
-                            .map_err(|error| warn!(?error, "unexpected send error"));
+                        if let Err(error) = unique_orchard_tree_height_tx.send(height) {
+                            warn!(?error, "unexpected send error")
+                        }
+
                         prev_orchard_tree = tree;
                     }
                 }
@@ -407,12 +408,18 @@ impl DbFormatChange {
                 }
             });
 
-            sapling_read_task.join().unwrap();
-            sapling_compare_task.join().unwrap();
-            sapling_delete_task.join().unwrap();
-            orchard_read_task.join().unwrap();
-            orchard_compare_task.join().unwrap();
-            orchard_delete_task.join().unwrap();
+            for task in [
+                sapling_read_task,
+                sapling_compare_task,
+                sapling_delete_task,
+                orchard_read_task,
+                orchard_compare_task,
+                orchard_delete_task,
+            ] {
+                if let Err(error) = task.join() {
+                    warn!(?error, "unexpected join error")
+                }
+            }
 
             // At the end of each format upgrade, we mark the database as upgraded to that version.
             // We don't mark the database if `height` didn't reach the `initial_tip_height` because
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -271,8 +271,6 @@ impl DbFormatChange {
         // Check if we need to prune the note commitment trees in the database.
         if older_disk_version < version_for_pruning_trees {
             let mut height = Height(0);
-            let mut prev_sapling_tree = upgrade_db.sapling_tree_by_height(&height);
-            let mut prev_orchard_tree = upgrade_db.orchard_tree_by_height(&height);
             let sapling_cf = upgrade_db
                 .db
                 .cf_handle("sapling_note_commitment_tree")
@@ -304,6 +302,8 @@ impl DbFormatChange {
             height = sapling_height;
             warn!(?height, "Database upgrade is at:");
 
+            let mut prev_sapling_tree = upgrade_db.sapling_tree_by_height(&height);
+
             while height <= orchard_height {
                 let mut batch = DiskWriteBatch::new();
                 for _ in 0..1_000 {
@@ -328,6 +328,8 @@ impl DbFormatChange {
                     height = height.next();
                 }
             }
+
+            let mut prev_orchard_tree = upgrade_db.orchard_tree_by_height(&height);
 
             // Go through every height from genesis to the tip of the old version. If the state was
             // downgraded, some heights might already be upgraded. (Since the upgraded format is
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -304,7 +304,7 @@ impl DbFormatChange {
 
             // Set up task for reading sapling note commitment trees
             let db = upgrade_db.clone();
-            let sapling_read_task = tokio::spawn(async move {
+            let sapling_read_task = std::thread::spawn(move || {
                 for (height, tree) in
                     db.sapling_tree_by_height_range(sapling_height..initial_tip_height)
                 {
@@ -316,7 +316,7 @@ impl DbFormatChange {
             let (unique_sapling_tree_height_tx, unique_sapling_tree_height_rx) = mpsc::channel();
 
             // Set up task for reading sapling note commitment trees
-            let sapling_compare_task = tokio::spawn(async move {
+            let sapling_compare_task = std::thread::spawn(move || {
                 while let Ok((height, tree)) = sapling_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_sapling_tree != tree {
@@ -328,7 +328,7 @@ impl DbFormatChange {
 
             // Set up task for deleting sapling note commitment trees
             let db = upgrade_db.clone();
-            let sapling_delete_task = tokio::spawn(async move {
+            let sapling_delete_task = std::thread::spawn(move || {
                 let mut prev_height = sapling_height;
                 while let Ok(height) = unique_sapling_tree_height_rx.recv() {
                     let delete_from = (prev_height + 1).expect("should be valid height");
@@ -358,7 +358,7 @@ impl DbFormatChange {
 
             // Set up task for reading orchard note commitment trees
             let db = upgrade_db.clone();
-            let orchard_read_task = tokio::spawn(async move {
+            let orchard_read_task = std::thread::spawn(move || {
                 for (height, tree) in
                     db.orchard_tree_by_height_range(orchard_height..initial_tip_height)
                 {
@@ -370,7 +370,7 @@ impl DbFormatChange {
             let (unique_orchard_tree_height_tx, unique_orchard_tree_height_rx) = mpsc::channel();
 
             // Set up task for reading orchard note commitment trees
-            let orchard_compare_task = tokio::spawn(async move {
+            let orchard_compare_task = std::thread::spawn(move || {
                 while let Ok((height, tree)) = orchard_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_orchard_tree != tree {
@@ -382,7 +382,7 @@ impl DbFormatChange {
 
             // Set up task for deleting orchard note commitment trees
             let db = upgrade_db;
-            let orchard_delete_task = tokio::spawn(async move {
+            let orchard_delete_task = std::thread::spawn(move || {
                 let mut prev_height = orchard_height;
                 while let Ok(height) = unique_orchard_tree_height_rx.recv() {
                     let delete_from = (prev_height + 1).expect("should be valid height");
@@ -407,21 +407,19 @@ impl DbFormatChange {
                 }
             });
 
-            tokio::spawn(async move {
-                sapling_read_task.await.unwrap();
-                sapling_compare_task.await.unwrap();
-                sapling_delete_task.await.unwrap();
-                orchard_read_task.await.unwrap();
-                orchard_compare_task.await.unwrap();
-                orchard_delete_task.await.unwrap();
+            sapling_read_task.join().unwrap();
+            sapling_compare_task.join().unwrap();
+            sapling_delete_task.join().unwrap();
+            orchard_read_task.join().unwrap();
+            orchard_compare_task.join().unwrap();
+            orchard_delete_task.join().unwrap();
 
-                // At the end of each format upgrade, we mark the database as upgraded to that version.
-                // We don't mark the database if `height` didn't reach the `initial_tip_height` because
-                // Zebra wouldn't run the upgrade anymore, and the part of the database above `height`
-                // wouldn't be upgraded.
-                info!(?newer_running_version, "Database has been upgraded to:");
-                Self::mark_as_upgraded_to(&version_for_pruning_trees, &config, network);
-            });
+            // At the end of each format upgrade, we mark the database as upgraded to that version.
+            // We don't mark the database if `height` didn't reach the `initial_tip_height` because
+            // Zebra wouldn't run the upgrade anymore, and the part of the database above `height`
+            // wouldn't be upgraded.
+            info!(?newer_running_version, "Database has been upgraded to:");
+            Self::mark_as_upgraded_to(&version_for_pruning_trees, &config, network);
         }
 
         // End of a database upgrade task.
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -339,9 +339,9 @@ impl DbFormatChange {
                         batch.delete_range_sapling_tree(&db, &delete_from, &height);
                         db.write_batch(batch)
                             .expect("Deleting note commitment trees should always succeed.");
-                    } else if num_entries.map_or(false, |n| n == 0) {
+                    } else if num_entries.map_or(false, |n| n == 1) {
                         let mut batch: DiskWriteBatch = DiskWriteBatch::new();
-                        batch.delete_sapling_tree(&db, &height);
+                        batch.delete_sapling_tree(&db, &delete_from);
                         db.write_batch(batch)
                             .expect("Deleting note commitment trees should always succeed.");
                     }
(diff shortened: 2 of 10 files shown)


arya2 (dismissed):
This looks perfect - thank you!

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -338,4 +338,22 @@ impl DiskWriteBatch {
...
+    }
+
+    /// Deletes the Orchard note commitment tree at the given [`Height`].
+    pub fn delete_orchard_tree(&mut self, zebra_db: &ZebraDb, height: &Height) {
+        let orchard_tree_cf = zebra_db
+            .db
+            .cf_handle("orchard_note_commitment_tree")
+            .unwrap();
+        self.zs_delete(&orchard_tree_cf, height);
+    }
```
arya2: Optional: Add methods for getting the cf handles.

```suggestion
    /// Deletes the Sapling note commitment tree at the given [`Height`].
    pub fn delete_sapling_tree(&mut self, zebra_db: &ZebraDb, height: &Height) {
        self.zs_delete(&zebra_db.sapling_tree_cf(), height);
    }

    /// Deletes the Orchard note commitment tree at the given [`Height`].
    pub fn delete_orchard_tree(&mut self, zebra_db: &ZebraDb, height: &Height) {
        self.zs_delete(&zebra_db.orchard_tree_cf(), height);
    }
```

upbqdn: Oh, nice. I think we can do that in other places as well.

arya2: > Oh, nice. I think we can do that in other places as well.

This came out of mild disappointment seeing the column names defined in several places - I think we should use constants there too. 

Let's merge this PR and create a ticket for that cleanup?

gustavovalverde:
@mergify refresh

arya2:
@Mergifyio update

upbqdn:
I pushed a new version of the tree pruning task. The previous task had two issues:

1. When the task got canceled, and the upgrade height hadn't reached the finalized tip, the database was still marked as upgraded, and Zebra wouldn't launch the upgrade anymore, leaving the rest of the database untouched.
2. The upgrade task was using `DiskWriteBatch`es directly, which works only in tests since we don't impl [Deref](https://doc.rust-lang.org/beta/core/ops/deref/trait.Deref.html)<Target = [DiskDb](https://doc-internal.zebra.zfnd.org/zebra_state/service/finalized_state/disk_db/struct.DiskDb.html)> for `ZebraDb` in production, so I moved the `DiskWriteBatch` instantiation to `ZebraDb`. I removed `Network` from `DiskWriteBatch` to do that.

The current task seems to be working, but it's very slow. Here's the progress by 10_000 heights:

```log
2023-08-10T22:05:13 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 10000
2023-08-10T22:09:26 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 20000
2023-08-10T22:16:42 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 30000
2023-08-10T22:27:03 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 40000
2023-08-10T22:40:30 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 50000
2023-08-10T22:57:12 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 60000
2023-08-10T23:17:32 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 70000
2023-08-10T23:41:36 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 80000
2023-08-11T00:08:32 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 90000
2023-08-11T00:38:12 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 100000
2023-08-11T01:10:45 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 110000
2023-08-11T01:45:54 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 120000
2023-08-11T02:23:46 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 130000
2023-08-11T03:04:15 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 140000
2023-08-11T03:47:28 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 150000
2023-08-11T04:33:31 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 160000
2023-08-11T05:22:14 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 170000
2023-08-11T06:13:25 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 180000
2023-08-11T07:07:06 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 190000
2023-08-11T08:03:14 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 200000
2023-08-11T09:01:50 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 210000
2023-08-11T10:03:08 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 220000
2023-08-11T11:07:05 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 230000
2023-08-11T12:13:44 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 240000
2023-08-11T13:23:02 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 250000
2023-08-11T14:35:07 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 260000
2023-08-11T15:49:41 zebra_state::service::finalized_state::disk_format::upgrade: The upgrade is at height 270000
```

Notice that the upgrade task went through heights 0-10_000 in ~ 4 minutes, but it took more than an hour to go through 260_000-270_000. The task was getting slower as it progressed.

I'm currently testing another version of the upgrade task with batched removals.

arya2:
@Mergifyio update

arya2 (dismissed):

upbqdn:
I'm doing more manual testing, so I temporarily converted this PR to a draft.

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -388,6 +389,11 @@ impl ZebraDb {
 
         Ok(finalized.verified.hash)
     }
+
+    /// Writes the given batch to the database.
+    pub(crate) fn write_batch(&self, batch: DiskWriteBatch) -> Result<(), rocksdb::Error> {
+        self.db.write(batch)
+    }
```
upbqdn, line 396: We're exposing direct write access to the database via arbitrary batches here. It would be better if we prevented that, but I can't think of a better solution.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +263,165 @@ impl DbFormatChange {
...
+            let db = upgrade_db;
+            let orchard_delete_task = std::thread::spawn(move || {
+                let mut prev_height = orchard_height;
+                while let Ok(height) = unique_orchard_tree_height_rx.recv() {
+                    let delete_from = (prev_height + 1).expect("should be valid height");
+                    let num_entries = height.0.checked_sub(delete_from.0);
+
+                    if num_entries.map_or(false, |n| n >= 1) {
+                        let mut batch: DiskWriteBatch = DiskWriteBatch::new();
+                        batch.delete_range_orchard_tree(&db, &delete_from, &height);
```
arya2: Nitpick: it may be slightly faster to use `zs_delete` when the range would only delete a single key.

upbqdn: OK, I'll bring it back.

arya2: Wait but with that said, great cleanup - please bring it back without the duplicate `batch` definitions and `write_batch` calls.

zebra-state/src/service/finalized_state/zebra_db/block.rs
arya2: Would it help to move methods here so this type defines all of the possible db operations that Zebra may make?

upbqdn: I don't know.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Done in [235e43b](https://github.com/ZcashFoundation/zebra/pull/7312/commits/235e43bf8f98a7442cc885147a42c03102d0a07b).

upbqdn:
> The current task seems to be working, but it's very slow.

The migration is now blazing fast.

arya2 (approved):

zebra-state/src/service/finalized_state/zebra_db/block.rs
arya2: Optional: open an issue with the problem? It may be worth fixing ahead of more format upgrades after the subtree indexes, or before stabilizing the zebra-state crate.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -457,7 +465,7 @@ impl DiskDb {
             .map(|result| result.expect("unexpected database failure"))
             .map(|(key, value)| (key.to_vec(), value))
             // Handle Excluded start and the end bound
-            .filter(move |(key, _value)| range.contains(key))
+            .take_while(move |(key, _value)| range.contains(key))
```
teor2345, line 468: Low priority bug:

This will terminate iteration too early if there's an excluded start bound.

I don't think this causes any bugs in existing code, because excluded start bounds don't have a standard Rust iteration syntax. (So you'd have to pass `(Excluded(start), _)` to trigger the bug.)

Instead, we probably want something like:
```rust
// skip excluded start bound (and empty ranges), the start_mode already skips keys before the start bound
.skip_while(move |(key, _value)| !range.contains(key))
// take until the excluded end bound is reached, or we're after the included end bound
.take_while(move |(key, _value)| range.contains(key))
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -115,11 +113,17 @@ pub trait WriteDisk {
...
         V: IntoDisk;
 
-    /// Remove the given key form rocksdb column family if it exists.
+    /// Remove the given key from rocksdb column family if it exists.
     fn zs_delete<C, K>(&mut self, cf: &C, key: K)
     where
         C: rocksdb::AsColumnFamilyRef,
         K: IntoDisk + Debug;
+
+    /// Remove the given key range from rocksdb column family if it exists.
```
teor2345, line 122: Doc Nit: Including `from` but excluding `to`.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
+
+            // Create an unbounded channel for reading note commitment trees
+            let (sapling_tree_tx, sapling_tree_rx) = mpsc::channel();
+
+            // Set up task for reading sapling note commitment trees
+            let db = upgrade_db.clone();
+            let should_cancel_flag = should_cancel_format_change.clone();
+            let sapling_read_task = std::thread::spawn(move || {
+                for (height, tree) in
+                    db.sapling_tree_by_height_range(sapling_height..initial_tip_height)
```
teor2345, line 314: This range should be `0..=initial_tip_height`, for two reasons:
1. If the first shielded transaction is not in the activation block, then there will be duplicate empty note commitment trees for the genesis and activation blocks
2. If the initial tip height is a duplicate, it needs to be deleted (see my detailed comment on the code below)

This is a bug that requires us to re-run the upgrade task after fixing it, by assigning it a new state version.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
+                        prev_sapling_tree = tree;
+                    }
+                }
+            });
+
+            // Set up task for deleting sapling note commitment trees
+            let db = upgrade_db.clone();
+            let sapling_delete_task = std::thread::spawn(move || {
+                let mut delete_from = sapling_height.next();
+                while let Ok(delete_to) = unique_sapling_tree_height_rx.recv() {
```
teor2345, line 348: If the final `delete_to` is not equal to the initial tip height, there needs to be a delete range from the final `delete_from`, up to and including the initial tip height. (So to `initial_tip_height + 1`.)

This is a bug that requires us to re-run the upgrade task after fixing it, by assigning it a new state version.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
+            for task in [
+                sapling_read_task,
+                sapling_compare_task,
+                sapling_delete_task,
+                orchard_read_task,
+                orchard_compare_task,
+                orchard_delete_task,
+            ] {
+                if let Err(error) = task.join() {
+                    warn!(?error, "unexpected join error")
```
teor2345, line 440: Low priority bug:

If a thread errors that means that its code has panicked. So the error handler should resume the panic unwind, just logging a warning can let code in threads panic undetected.

In particular, the database would be marked as upgraded even if a panic terminated the upgrade early, which is a bug.

Instead, let's just simplify things by removing the extra threads?

If there are actually any panics in these threads, then we might need to re-run the format upgrade after fixing the panics and this panic handler.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
-
-            // Go through every height from genesis to the tip of the old version.
-            // If the state was downgraded, some heights might already be upgraded.
-            // (Since the upgraded format is added to the tip, the database can switch between
-            // lower and higher versions at any block.)
-            //
-            // Keep upgrading until the initial database has been upgraded,
-            // or this task is cancelled by a shutdown.
-            while upgrade_height <= initial_tip_height
-                && matches!(cancel_receiver.try_recv(), Err(mpsc::TryRecvError::Empty))
```
teor2345, line 281: Low priority bug:

The channel check cancels the task when the sender is dropped, but the AtomicBool does not.

That might make Zebra hang on shutdown waiting for a format upgrade task to finish.

One way to fix this is to revert to the channel check (we can use a wrapper, trait, or function to simplify it if needed). Another is to set the bool on drop, but in general I try not to use atomic because I don't understand how atomic orderings work and how to avoid concurrency bugs when using them. So it's tricky for me to review this code.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -112,8 +112,7 @@ impl ZebraDb {
             .zs_items_in_range_unordered(&sprout_anchors_handle, ..)
     }
 
-    /// Returns the Sapling note commitment tree of the finalized tip
-    /// or the empty tree if the state is empty.
+    /// Returns the Sapling note commitment trees starting from the given block height up to the chain tip
```
teor2345, line 115: Doc Nit: I think this comment update is a copy-paste error, there's no height passed to this method.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -124,6 +123,33 @@ impl ZebraDb {
             .expect("Sapling note commitment tree must exist if there is a finalized tip")
     }
 
+    /// Returns the Orchard note commitment trees starting from the given block height up to the chain tip
```
teor2345, line 126: Doc Nit: The range can have an end bound, so this doc should be "in the supplied range", and the same for the sapling method.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -338,4 +357,40 @@ impl DiskWriteBatch {
...
+    pub fn delete_sapling_tree(&mut self, zebra_db: &ZebraDb, height: &Height) {
+        let sapling_tree_cf = zebra_db
+            .db
+            .cf_handle("sapling_note_commitment_tree")
+            .unwrap();
+        self.zs_delete(&sapling_tree_cf, height);
+    }
+
+    /// Deletes the range of Sapling note commitment trees at the given [`Height`]s. Doesn't delete the upper bound.
+    pub fn delete_range_sapling_tree(&mut self, zebra_db: &ZebraDb, from: &Height, to: &Height) {
```
teor2345, line 371: Optional:

It seems like this API caused a state format bug, maybe it would be better to write it in terms of a Rust `RangeBounds`?

(Like `sapling_tree_by_height_range`.)

This would be best implemented by changing `zs_delete_range` to take a Rust `RangeBounds`.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
-                // TODO: Do one format upgrade step here
-                //
-                // This fake step just shows how to access the database.
-                let _replace_me_ = upgrade_db.tip();
-
-                upgrade_height = (upgrade_height + 1).expect("task exits before maximum height");
+        // Start of a database upgrade task.
+
+        let version_for_pruning_trees =
+            Version::parse("25.1.0").expect("Hardcoded version string should be valid.");
```
teor2345, line 275: Since PR #7266 was merged into the main branch before this PR, some cached states might have already been marked as "upgraded" to version `25.1.0`, without actually doing the upgrade in this PR's code.

That means that CI could be failing on the note commitment subtrees PRs, because the data format is not as expected.

Let's upgrade to patch version 25.1.1 in the fix PR?

(This is a bug that requires a re-run of the upgrade task, but only in Zebra's CI.)

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
-                "marking database as upgraded"
-            );
-            Self::mark_as_upgraded_to(&database_format_add_format_change_task, &config, network);
+            // At the end of each format upgrade, we mark the database as upgraded to that version.
+            // We don't mark the database if `height` didn't reach the `initial_tip_height` because
+            // Zebra wouldn't run the upgrade anymore, and the part of the database above `height`
+            // wouldn't be upgraded.
+            info!(?newer_running_version, "Database has been upgraded to:");
+            Self::mark_as_upgraded_to(&version_for_pruning_trees, &config, network);
         }
```
teor2345, line 451: Since there have been multiple bugs in this code which the tests didn't catch, let's do a runtime test here that the format is actually as expected?

If we've actually deleted all the duplicates, this should be quick.

I'll write this test in a separate PR.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -261,46 +269,187 @@ impl DbFormatChange {
...
+                .iter()
+                .find(|(_, upgrade)| **upgrade == NetworkUpgrade::Nu5)
+                .expect("there should be Nu5 upgrade");
+
+            // Delete duplicates before sapling and orchard heights
+            let (mut prev_sapling_tree, mut prev_orchard_tree) = {
+                let height = Height(1);
+                let mut batch = DiskWriteBatch::new();
+
+                batch.delete_range_sapling_tree(&upgrade_db, &height, &sapling_height);
```
teor2345, line 294: This code is a nice optimisation, but it can leave a duplicate tree in the network upgrade activation block, if that block contains no shielded transactions. (Which is the typical case, because wallets often wait for activation before generating those transactions.)

I'd suggest deleting this code to reduce code complexity, even if it is slightly slower.

This is a bug that requires us to re-run the upgrade task after fixing it, by assigning it a new state version.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: We might also want to check for a similar missed duplicate bug in the first new block that is committed to the finalized state after Zebra is started.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: Optional:

#### Related Logging Bug

Threads need to be explicitly set up with the logging spans, otherwise their logs can be missing important diagnostic info.

#### Related Performance Question

Are these threads actually more efficient than running all the code in the upgrade thread? Does the difference matter?

Splitting code into separate threads can actually slow it down due to lock contention, context switching overhead, and missed compiler optimisations. (Code that's split across threads can't be optimised using assumptions from code in another thread.)

teor2345 (commented):
I think there are still some duplicate note commitment trees after this state upgrade. See my tests in PR #7365.

I've tried to mark each review comment as a breaking format change, low priority related issue, or nitpick.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: This test is now in PR #7365.

teor2345:
When I run this test on my fully synced `v25.0` state, I get these logs:
```
2023-08-22T05:08:00.359950Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate sapling trees after running de-duplicate tree upgrade height=Height(419200) prev_height=Height(0) tree_root=Root("fbc2f4300c01f0b7820d00e3347c8da4ee614674376cbc45359daa54f9b5493e")

2023-08-22T05:08:02.610711Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(1687104) prev_height=Height(0) tree_root=Root("ae2935f1dfd8a24aed7c70df7de3a668eb7a49b1319880dde2bbd9031ae5d82f")
2023-08-22T05:08:03.271360Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199893) prev_height=Height(2199892) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271380Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199894) prev_height=Height(2199893) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
```

<details>
<summary>8 duplicate orchard trees from 2199895-2199902</summary>
```
2023-08-22T05:08:03.271385Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199895) prev_height=Height(2199894) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271388Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199896) prev_height=Height(2199895) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271392Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199897) prev_height=Height(2199896) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271395Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199898) prev_height=Height(2199897) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271399Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199899) prev_height=Height(2199898) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271402Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199900) prev_height=Height(2199899) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271406Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199901) prev_height=Height(2199900) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271412Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199902) prev_height=Height(2199901) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
```
</details>

```
2023-08-22T05:08:03.271416Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199903) prev_height=Height(2199902) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")
2023-08-22T05:08:03.271420Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate orchard trees after running de-duplicate tree upgrade height=Height(2199904) prev_height=Height(2199903) tree_root=Root("ffb1812f72a3888d8ccc307ca4701078daa6df664e5766ccc3ead1764ed29727")

Message:  found duplicate sapling or orchard trees after running de-duplicate tree upgrade
Location: zebra-state/src/service/finalized_state/disk_format/upgrade.rs:489

Metadata:
version: 1.1.0+41.g54ce3ad
Zcash network: Mainnet
running state version: 25.1.0
initial disk state version: 25.0.2
```

<details>
<summary>extra metadata</summary>

```
branch: de-duplicate-test
git commit: 54ce3ad
commit timestamp: 1980-01-01T00:00:00.000000000Z
rust compiler: 1.74.0-nightly
rust release date: 2023-08-21
optimization level: 3
debug checks: true
```
</details>

Testnet has about 30 sapling duplicates, and about 10,000 orchard duplicates. (There might be an early termination or concurrency bug on orchard, or it might just be the same bugs.)

upbqdn:
Thanks for the review and the test. I'm going to open another PR to address the issues.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Won't the take_while predicate always return true if it skips elements that would return false?

teor2345: I'm pretty sure the code works as documented. But it's a good question, because there aren't any tests. We might want to add them as part of PR #7365 or Marek's upcoming fix PR. (If the tests in those PRs don't cover this code already.)

### Detailed Background

Have you read the `RangeBounds`, `skip_while()`, and `take_while()` docs?

A `RangeBounds` identifies a continuous sequence of ordered keys, but it can exclude some keys at the start and end of the column family:
https://doc.rust-lang.org/std/ops/trait.RangeBounds.html

`skip_while()` runs the predicate on each item it gets, skipping items until it fails, then yields the rest of the iterator (without running the predicate):
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while

`take_while()` runs the predicate on each item it gets, yielding items until it fails, then ignores the rest of the iterator:
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while

<details>

<summary>detailed explanation of this implementation</summary>

There are 3 kinds of key exclusion that happen in this code:
At the database level:
- `start_mode` can skip some keys at the start, but always includes the start bound, even if it should be excluded by the range
At the iterator level:
- `skip_while()` skips the start bound if it should be excluded, then includes everything from the first key that is contained in the range (ignoring the predicate). It should only skip zero or one key (because start_mode skips most of them). Then it includes all the remaining keys, regardless of the end of the range.
- `take_while()` includes everything until the end bound is reached, then stops the iterator (correctly implementing inclusion or exclusion). It can include any number of keys, then stop before any number of keys. 

</details>

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: Edit: `RangeBounds` is the generic trait, `Range` is a particular type of range that implements it.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Ah, for some reason I was imagining it doing the same thing as filter. I got it now.

upbqdn: Addressed in [0b3a1b7](https://github.com/ZcashFoundation/zebra/pull/7379/commits/0b3a1b7ffe333b29c7ed4bfcb26c1453d4c2fa08).

zebra-state/src/service/finalized_state/disk_db.rs
upbqdn: Addressed in [cd3bff6](https://github.com/ZcashFoundation/zebra/pull/7379/commits/cd3bff6668afe0917d23a4ccd5cb119660d9b70a).

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Addressed in [cd4be3e](https://github.com/ZcashFoundation/zebra/pull/7379/commits/cd4be3e2f6829994c7b7c2989f5b39fd06141ecf).

> We might also want to check for a similar missed duplicate bug in the first new block that is committed to the finalized state after Zebra is started.

I don't think this can happen because duplicate trees aren't committed to the database anymore: https://github.com/zcashfoundation/zebra/blob/ead854b587d9a64ef1ecd5c4971b3a4736c04a91/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L339-L356.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Addressed in https://github.com/zcashfoundation/zebra/blob/ead854b587d9a64ef1ecd5c4971b3a4736c04a91/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L339-L356.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: [cd4be3e](https://github.com/ZcashFoundation/zebra/pull/7379/commits/cd4be3e2f6829994c7b7c2989f5b39fd06141ecf) removes the threads.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: [cd4be3e](https://github.com/ZcashFoundation/zebra/pull/7379/commits/cd4be3e2f6829994c7b7c2989f5b39fd06141ecf) uses `mpsc`.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: I tried implementing this but ended up not including it in PR #7379 because it would make the current pruning a bit more complicated than it is.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Addressed in [5c299cc](https://github.com/ZcashFoundation/zebra/pull/7379/commits/5c299cc6355cd435ee14a2ace778400849239f0d).

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Addressed in [cd4be3e](https://github.com/ZcashFoundation/zebra/pull/7379/commits/cd4be3e2f6829994c7b7c2989f5b39fd06141ecf).

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Addressed in [c831a74](https://github.com/ZcashFoundation/zebra/pull/7379/commits/c831a74d9481cf830309c83a8e06e6578796628d).

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Addressed in [9dba249](https://github.com/ZcashFoundation/zebra/pull/7379/commits/9dba2492d920ef9e5728b5543427aa2b63ba0079).

upbqdn:
I opened PR #7379 which addresses all comments.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: No worries! Happy to open a ticket or add a TODO or just do nothing until the refactor is needed. Whatever you think is best.

PR #7313, opened 2023-08-09
fix(doc): consolidate firebase hosting in prod project

gustavovalverde:
## Motivation

Each documentation we're hosting using Firebase, is deployed to a different GCP project each. We're now using our `zfnd-prod-zebra` to deploy all Zebra services targeting users. And we'd like to have all documentation consolidated under a single GCP project.

A new action is being used, and a custom made step was required as the Workload Identity Provider is not supported, and we're no longer allowed to created long-lived JSON keys from our GCP projects.

## Solution

- Use `FirebaseExtended/action-hosting-deploy` action
- Deploy to new a single GCP project in `zfnd-prod-zebra` 
- Create a `production` environment in GitHub to use the same variable name for `$GCP_PROJECT`
- Deploy the different documentations in parallel
- PRs changing the documentation create a temporal URL to validate the changes

## Review

Anyone from DevOps

### Reviewer Checklist

## Follow Up Work
- Do not redeploy all docs if just one is needed


gustavovalverde:
@Mergify refresh

gustavovalverde:
@arya2 @upbqdn do we actually need the `internal` docs published at https://doc-internal.zebra.zfnd.org/zebrad/ additionally to https://doc.zebra.zfnd.org/zebrad/ ?

I might be ignoring something, but I don't see much value added there.

arya2:
> @arya2 @upbqdn do we actually need the `internal` docs published at https://doc-internal.zebra.zfnd.org/zebrad/ additionally to https://doc.zebra.zfnd.org/zebrad/ ?
> 
> I might be ignoring something, but I don't see much value added there.

I haven't found it very useful but it could be convenient for investigating Zebra without access to a dev environment.

We could publish the internal docs on https://doc.zebra.zfnd.org/zebrad/ so it's all in one place? I wouldn't mind just dropping the internal docs page altogether and asking developers to run `cargo doc open` if they want to use it.

upbqdn:
Let's keep it. I use it every now and then, even today.

Having it only locally makes one unable to share links, for example.

arya2 (approved):
Looks good!

PR #7321, opened 2023-08-12
Suggestion for Batch tree removal when migrating the database

arya2:
This is a suggestion for #7320.

It deletes the ranges before we can expect unique note commitment trees without any prior reads or comparisons and adds other optimizations.

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -113,8 +113,14 @@ pub trait WriteDisk {
         K: IntoDisk + Debug,
         V: IntoDisk;
 
-    /// Remove the given key form rocksdb column family if it exists.
+    /// Remove the given key from rocksdb column family if it exists.
     fn zs_delete<C, K>(&mut self, cf: &C, key: K)
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + Debug;
+
+    /// Remove the given key range from rocksdb column family if it exists.
+    fn zs_delete_range<C, K>(&mut self, cf: &C, from: K, to: K)
     where
         C: rocksdb::AsColumnFamilyRef,
         K: IntoDisk + Debug;
@@ -139,6 +145,16 @@ impl WriteDisk for DiskWriteBatch {
     {
         let key_bytes = key.as_bytes();
         self.batch.delete_cf(cf, key_bytes);
+    }
+
+    fn zs_delete_range<C, K>(&mut self, cf: &C, from: K, to: K)
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + Debug,
+    {
+        let from_bytes = from.as_bytes();
+        let to_bytes = to.as_bytes();
+        self.batch.delete_range_cf(cf, from_bytes, to_bytes);
     }
 }
 
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -420,7 +420,7 @@ impl DiskDb {
     /// Returns an iterator over the items in `cf` in `range`.
     ///
     /// Holding this iterator open might delay block commit transactions.
-    fn zs_range_iter<C, K, V, R>(&self, cf: &C, range: R) -> impl Iterator<Item = (K, V)> + '_
+    pub fn zs_range_iter<C, K, V, R>(&self, cf: &C, range: R) -> impl Iterator<Item = (K, V)> + '_
     where
         C: rocksdb::AsColumnFamilyRef,
         K: IntoDisk + FromDisk,
@@ -465,7 +465,7 @@ impl DiskDb {
             .map(|result| result.expect("unexpected database failure"))
             .map(|(key, value)| (key.to_vec(), value))
             // Handle Excluded start and the end bound
-            .filter(move |(key, _value)| range.contains(key))
+            .take_while(move |(key, _value)| range.contains(key))
             .map(|(key, value)| (K::from_bytes(key), V::from_bytes(value)))
     }
 
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -308,7 +308,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.sapling_tree_by_height_range(sapling_height..initial_tip_height)
                 {
-                    let _ = sapling_tree_tx.send((height, tree));
+                    let _ = sapling_tree_tx
+                        .send((height, tree))
+                        .map_err(|error| warn!(?error, "unexpected send error"));
                 }
             });
 
@@ -320,7 +322,9 @@ impl DbFormatChange {
                 while let Ok((height, tree)) = sapling_tree_rx.recv() {
                     let tree = Some(tree);
                     if prev_sapling_tree != tree {
-                        let _ = unique_sapling_tree_height_tx.send(height);
+                        let _ = unique_sapling_tree_height_tx
+                            .send(height)
+                            .map_err(|error| warn!(?error, "unexpected send error"));
                         prev_sapling_tree = tree;
                     }
                 }
@@ -361,7 +365,9 @@ impl DbFormatChange {
                 for (height, tree) in
                     db.orchard_tree_by_height_range(orchard_height..initial_tip_height)
                 {
-                    let _ = orchard_tree_tx.send((height, tree));
+                    let _ = orchard_tree_tx
+                        .send((height, tree))
+                        .map_err(|error| warn!(?error, "unexpected send error"));
                 }
             });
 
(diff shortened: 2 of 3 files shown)


zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
-                        break;
-                    }
-                    let sapling_tree = upgrade_db.sapling_tree_by_height(&height);
-                    let orchard_tree = upgrade_db.orchard_tree_by_height(&height);
 
-                    if prev_sapling_tree == sapling_tree {
-                        batch.delete_sapling_tree(&sapling_cf, &height);
-                    }
+                batch.delete_range_sapling_tree(&upgrade_db, &height, &sapling_height);
+                batch.delete_range_orchard_tree(&upgrade_db, &height, &orchard_height);
```
arya2, line 289: Should these be `sapling_height - 1` and `orchard_height - 1`?

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
+
+            // Create an unbounded channel for duplicate sapling note commitment tree heights
+            let (unique_sapling_tree_height_tx, unique_sapling_tree_height_rx) = mpsc::channel();
+
+            // Set up task for reading sapling note commitment trees
+            let sapling_compare_task = std::thread::spawn(move || {
+                while let Ok((height, tree)) = sapling_tree_rx.recv() {
+                    let tree = Some(tree);
+                    if prev_sapling_tree != tree {
+                        let _ = unique_sapling_tree_height_tx.send(height);
```
arya2: These sends should  panic if they fail

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -449,7 +465,7 @@ impl DiskDb {
             .map(|result| result.expect("unexpected database failure"))
             .map(|(key, value)| (key.to_vec(), value))
             // Handle Excluded start and the end bound
-            .filter(move |(key, _value)| range.contains(key))
```
upbqdn, line 452: I tried searching for the reasons to use `take_while`, but I didn't find anything. Why the switch to `take_while`?

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: I don't know how to come up with a definite answer, but I'm convinced they should. We can always continue the pruning at `height -1` without any performance loss.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
-                    if prev_sapling_tree == sapling_tree {
-                        batch.delete_sapling_tree(&sapling_cf, &height);
-                    }
+                batch.delete_range_sapling_tree(&upgrade_db, &height, &sapling_height);
+                batch.delete_range_orchard_tree(&upgrade_db, &height, &orchard_height);
+                upgrade_db
+                    .write_batch(batch)
+                    .expect("Deleting note commitment trees should always succeed.");
+
+                warn!(?sapling_height, "Database upgrade is at:");
```
upbqdn: We should change these `warn`s to `info`s before merging the final PR, or even remove them or adjust them. I've been using them for benchmarking the migration.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
+                }
+            });
+
+            // Set up task for deleting sapling note commitment trees
+            let db = upgrade_db.clone();
+            let sapling_delete_task = std::thread::spawn(move || {
+                let mut prev_height = sapling_height;
+                while let Ok(height) = unique_sapling_tree_height_rx.recv() {
+                    let delete_from = (prev_height + 1).expect("should be valid height");
+                    let delete_to = (height - 1).expect("should be a valid height");
```
upbqdn: I don't think we should subtract here since `delete_range_cf` excludes `delete_to`: https://docs.rs/rocksdb/latest/rocksdb/struct.WriteBatchWithTransaction.html#method.delete_range_cf.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
                 }
+            });
+
+            // Set up task for deleting orchard note commitment trees
+            let db = upgrade_db;
+            let orchard_delete_task = std::thread::spawn(move || {
+                let mut prev_height = orchard_height;
+                while let Ok(height) = unique_orchard_tree_height_rx.recv() {
+                    let delete_from = (prev_height + 1).expect("should be valid height");
+                    let delete_to = (height - 1).expect("should be a valid height");
```
upbqdn: I don't think we should subtract here since `delete_range_cf` excludes `delete_to`: https://docs.rs/rocksdb/latest/rocksdb/struct.WriteBatchWithTransaction.html#method.delete_range_cf.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -270,69 +270,156 @@ impl DbFormatChange {
...
+                    let num_entries = delete_to.0.checked_sub(delete_from.0);
+
+                    if num_entries.map_or(false, |n| n >= 1) {
+                        let mut batch: DiskWriteBatch = DiskWriteBatch::new();
+                        batch.delete_range_orchard_tree(&db, &delete_from, &delete_to);
+                        db.write_batch(batch)
+                            .expect("Deleting note commitment trees should always succeed.");
+                    } else if num_entries.map_or(false, |n| n == 0) {
+                        let mut batch: DiskWriteBatch = DiskWriteBatch::new();
+                        batch.delete_orchard_tree(&db, &height);
```
upbqdn: I think this should be `height -1`. However, I think we can remove this `else if` and use only `delete_range_orchard_tree` since it doesn't delete the upper bound: https://docs.rs/rocksdb/latest/rocksdb/struct.WriteBatchWithTransaction.html#method.delete_range_cf?

Same for Sapling.

upbqdn (approved):
This looks awesome, thank you for writing it.

I haven't tested it yet. I wonder how it's going to behave if the migration is interrupted.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: > I don't know how to come up with a definite answer, but I'm convinced they should.

~I think they should too.~

`delete_range_cf` doesn't delete the upper bound (thank you).

zebra-state/src/service/finalized_state/disk_db.rs
arya2: [`take_while`](https://doc.rust-lang.org/src/core/iter/adapters/take_while.rs.html#42-54
) will return `None` but [`filter`](https://doc.rust-lang.org/src/core/iter/adapters/filter.rs.html#59) will try to find the next element, so, it would get every key-value after `initial_tip_height` and check that they're above the initial tip before dropping the iterator.

This wasn't an issue before because the method was only used for getting sprout anchors.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7321/commits/f922234ec825432f8aa59ec3a2e28ba0ca607015.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7321/commits/acce514c7057b06985a8ef1df97b887e20d256f5.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7321/commits/acce514c7057b06985a8ef1df97b887e20d256f5.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Removed in https://github.com/ZcashFoundation/zebra/pull/7321/commits/764d84613e62e7bb16b5c9b52033617e0851dc8b.

I think the info log at the end should be enough.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Logged warning if one of these sends fail in https://github.com/ZcashFoundation/zebra/pull/7321/commits/3aa19d2084f3b9dc1cbfec4f6c49c21a79c491a3

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: Oh, at that time of the review, I also assumed the upper bound would be deleted. It occurred to me later on to check the bounds.

arya2:
> I haven't tested it yet. I wonder how it's going to behave if the migration is interrupted.

It should start from the beginning if interrupted, which should be fine because it's only reading/iterating through existing keys and I think delete_range is implemented as a soft deletion.

PR #7330, opened 2023-08-15
Suggestions for Deduplicate note commitment trees stored in the finalized state

arya2:
This changes replaces the mpsc cancel_receiver with an `AtomicBool` and stops the upgrade once the new flag is set to true.

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -2,7 +2,10 @@
 
 use std::{
     cmp::Ordering,
-    sync::{mpsc, Arc},
+    sync::{
+        atomic::{self, AtomicBool},
+        mpsc, Arc,
+    },
     thread::{self, JoinHandle},
 };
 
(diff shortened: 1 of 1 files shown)


zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -353,10 +366,17 @@ impl DbFormatChange {
...
             // Set up task for reading orchard note commitment trees
             let db = upgrade_db.clone();
+            let should_cancel_flag = should_cancel_format_change;
             let orchard_read_task = std::thread::spawn(move || {
                 for (height, tree) in
                     db.orchard_tree_by_height_range(orchard_height..initial_tip_height)
                 {
+                    // Breaking from this loop and dropping the orchard_tree channel
+                    // will cause the orchard compare and delete tasks to finish.
+                    if should_cancel_flag.load(atomic::Ordering::Relaxed) {
```
arya2, line 376: Using `Relaxed` ordering here because it doesn't matter in this case if the cancel signal is seen out of order.

upbqdn (approved):
Thank you for solving this.

Issue #6953, opened 2023-06-14
Add note subtree indexes to the finalized and non-finalized state


teor2345:
## Motivation

We want to create a new database column family to support the `z_getsubtreesbyindex` RPC.

## Specifications

See the [state upgrade design docs](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#design).

`zcashd` implemented support for rebuilding and regenerating their subtree database based on existing blocks during initialization. We may not want to reproduce this exactly but it may be informative: https://github.com/zcash/zcash/pull/6694

### State Format Design

- We'll need a new column family that maps:
    - `subtree_index` to `(end_height, subtree_root)`

### Calculating subtree indexes

We need to calculate fixed-size note commitment subtrees for existing and new blocks.

Here is an algorithm and examples of the calculation:

<details>

<summary>implementation details</summary>

`zcashd`'s best chain / subtrees state PR:
https://github.com/zcash/zcash/pull/6649

In particular:
https://github.com/zcash/zcash/pull/6649/files#diff-31ef5cdda38b86238aa5fa0745b7843ebd43f11d65583fb9cfec23f7225e5f77

For each subtree in the history of the note commitment tree (Sapling, Orchard):
- every time a block that completes a subtree is appended to the chain, we will store a new subtree (a block can only complete one subtree due to block size limits)
    - some blocks will complete a subtree and start a new one
    - some blocks will complete a subtree, and the next block will start a new one
- if a reorg happens (not relevant to Zebra)
- light clients need to be able to fetch any subtrees from `current_height - 100` onwards, using bitwise math (still to be specified)
    - this will probably require scanning through the database for trees, or building an index of note commitment counts per block

- We can calculate the subtree roots efficiently using: https://github.com/zcash/zcash/pull/6431/commits/1e23185ffa3b3e623f588524aa8bf3d4cd4abbb2#diff-dd19eaef96b055fb2c94eaa1a7bcda8074fb912e2fefc0695e0b4c05a76d533eR121-L108 see the discussion at https://discord.com/channels/809218587167293450/809251012610752513/1110592469575422016

</details>

### Complex Code or Requirements

State lookups are concurrent, so we'll need to deal with concurrent database updates where the database overlaps the cloned non-finalized state.

## Implementation Steps

 - [ ] update the [database format](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#current) in the Zebra docs
[...]

PR #7334, opened 2023-08-15
change(state): Add note subtree index handling to zebra-state, but don't write them to the finalized state yet

arya2:
## Motivation

This is needed to support "spend before sync" in lightwalletd.

Part of #6953.

## Solution

- Add `is_complete_subtree()`, `subtree_index`, and `subtree_root` methods to Sapling and Orchard `NoteCommitmentTree`s.
- Find end heights of subtrees
- Find start heights of the next subtree when a subtree is completed
- Find the start height of the first subtree in the pool if subtree_index is 0
- Insert start heights without roots or end heights
- Update subtree_indexes with end heights and subtree roots with `merge_cf`
- Increment database format minor version

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Add subtrees below initial tip height

### DIFF
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -92,10 +92,12 @@ We use the following rocksdb column families:
 | `sapling_nullifiers`               | `sapling::Nullifier`   | `()`                          | Create  |
 | `sapling_anchors`                  | `sapling::tree::Root`  | `()`                          | Create  |
 | `sapling_note_commitment_tree`     | `block::Height`        | `sapling::NoteCommitmentTree` | Create  |
+| `sapling_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
 | *Orchard*                          |                        |                               |         |
 | `orchard_nullifiers`               | `orchard::Nullifier`   | `()`                          | Create  |
 | `orchard_anchors`                  | `orchard::tree::Root`  | `()`                          | Create  |
 | `orchard_note_commitment_tree`     | `block::Height`        | `orchard::NoteCommitmentTree` | Create  |
+| `orchard_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
 | *Chain*                            |                        |                               |         |
 | `history_tree`                     | `block::Height`        | `NonEmptyHistoryTree`         | Delete  |
 | `tip_chain_value_pool`             | `()`                   | `ValueBalance`                | Update  |
@@ -118,6 +120,8 @@ Block and Transaction Data:
   used instead of a `BTreeSet<OutputLocation>` value, to improve database performance
 - `AddressTransaction`: `AddressLocation \|\| TransactionLocation`
   used instead of a `BTreeSet<TransactionLocation>` value, to improve database performance
+- `NoteCommitmentSubtreeIndex`: 16 bits, big-endian, unsigned
+- `NoteCommitmentSubtree`: `Height \|\| {sapling, orchard}::tree::Node`
 
 We use big-endian encoding for keys, to allow database index prefix searches.
 
@@ -333,6 +337,9 @@ So they should not be used for consensus-critical checks.
   state for every height, for the specific pool. Each tree is stored
   as a "Merkle tree frontier" which is basically a (logarithmic) subset of
   the Merkle tree nodes as required to insert new items.
+
+- The `{sapling, orchard}_note_commitment_subtree` stores the completion height and
+  root for every completed level 16 note commitment subtree, for the specific pool.
 
 - `history_tree` stores the ZIP-221 history tree state at the tip of the finalized
   state. There is always a single entry for it. The tree is stored as the set of "peaks"
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -92,7 +92,7 @@ We use the following rocksdb column families:
 | `sapling_nullifiers`               | `sapling::Nullifier`   | `()`                          | Create  |
 | `sapling_anchors`                  | `sapling::tree::Root`  | `()`                          | Create  |
 | `sapling_note_commitment_tree`     | `block::Height`        | `sapling::NoteCommitmentTree` | Create  |
-| `sapling_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
+| `sapling_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtreeData`   | Create  |
 | *Orchard*                          |                        |                               |         |
 | `orchard_nullifiers`               | `orchard::Nullifier`   | `()`                          | Create  |
 | `orchard_anchors`                  | `orchard::tree::Root`  | `()`                          | Create  |
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -97,7 +97,7 @@ We use the following rocksdb column families:
 | `orchard_nullifiers`               | `orchard::Nullifier`   | `()`                          | Create  |
 | `orchard_anchors`                  | `orchard::tree::Root`  | `()`                          | Create  |
 | `orchard_note_commitment_tree`     | `block::Height`        | `orchard::NoteCommitmentTree` | Create  |
-| `orchard_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
+| `orchard_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtreeData`   | Create  |
 | *Chain*                            |                        |                               |         |
 | `history_tree`                     | `block::Height`        | `NonEmptyHistoryTree`         | Delete  |
 | `tip_chain_value_pool`             | `()`                   | `ValueBalance`                | Update  |
@@ -121,7 +121,7 @@ Block and Transaction Data:
 - `AddressTransaction`: `AddressLocation \|\| TransactionLocation`
   used instead of a `BTreeSet<TransactionLocation>` value, to improve database performance
 - `NoteCommitmentSubtreeIndex`: 16 bits, big-endian, unsigned
-- `NoteCommitmentSubtree`: `Height \|\| {sapling, orchard}::tree::Node`
+- `NoteCommitmentSubtreeData<{sapling, orchard}::tree::Node>`: `Height \|\| {sapling, orchard}::tree::Node`
 
 We use big-endian encoding for keys, to allow database index prefix searches.
 
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -27,8 +27,11 @@ use zcash_primitives::merkle_tree::{write_commitment_tree, HashSer};
 
 use super::sinsemilla::*;
 
-use crate::serialization::{
-    serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
+use crate::{
+    serialization::{
+        serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
+    },
+    subtree::{TRACKED_SUBTREE_HEIGHT, TRACKED_SUBTREE_SIZE},
 };
 
 pub mod legacy;
@@ -170,6 +173,13 @@ impl ZcashDeserialize for Root {
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct Node(pallas::Base);
 
+impl Node {
+    /// Calls `to_repr()` on inner value.
+    pub fn to_repr(&self) -> [u8; 32] {
+        self.0.to_repr()
+    }
+}
+
 /// Required to convert [`NoteCommitmentTree`] into [`SerializedTree`].
 ///
 /// Zebra stores Orchard note commitment trees as [`Frontier`][1]s while the
@@ -221,6 +231,18 @@ impl Hashable for Node {
 impl From<pallas::Base> for Node {
     fn from(x: pallas::Base) -> Self {
         Node(x)
+    }
+}
+
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(
+            bytes.try_into().map_err(|_| "wrong byte slice len")?,
+        ))
+        .map(Node)
+        .ok_or("invalid Pallas field element")
     }
 }
 
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
         }
     }
 
+    /// Returns true if the most recently appended leaf completes the subtree
+    pub fn is_complete_subtree(&self) -> bool {
+        self.inner.value().map_or(false, |x| {
+            x.position()
+                .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
+        })
+    }
+
+    /// Returns subtree index at [`TRACKED_SUBTREE_HEIGHT`]
+    pub fn subtree_index(&self) -> u16 {
+        (self.position() >> TRACKED_SUBTREE_HEIGHT)
+            .try_into()
+            .expect("always within u16")
+    }
+
+    /// Returns subtree root, if any
+    pub fn subtree_root(&self) -> Option<Node> {
+        self.is_complete_subtree().then_some(())?;
+        Some(
+            self.inner
+                .value()?
+                .root(Some(TRACKED_SUBTREE_HEIGHT.into())),
+        )
+    }
+
+    /// Returns the index of note completing the next subtree
+    pub fn next_complete_subtree_note_index(&self) -> usize {
+        // # Correctness
+        //
+        // 2^16 - (position % 2^16) should always be between 1..=2^16
+        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+            .try_into()
+            .expect("should fit in usize")
+    }
+
+    /// Counts of note commitments added to the current subtree, max 2^16
+    fn subtree_note_count(&self) -> u64 {
+        self.count() % TRACKED_SUBTREE_SIZE
+    }
+
     /// Returns the current root of the tree, used as an anchor in Orchard
     /// shielded transactions.
     pub fn root(&self) -> Root {
@@ -370,6 +432,11 @@ impl NoteCommitmentTree {
     /// Uncommitted^Orchard = I2LEBSP_l_MerkleOrchard(2)
     pub fn uncommitted() -> pallas::Base {
         pallas::Base::one().double()
+    }
+
+    /// Position of final leaf added to the tree.
+    fn position(&self) -> u64 {
+        self.inner.value().map_or(0, |x| u64::from(x.position()))
     }
 
     /// Count of note commitments added to the tree.
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -484,7 +484,13 @@ impl Eq for NoteCommitmentTree {}
 
 impl PartialEq for NoteCommitmentTree {
     fn eq(&self, other: &Self) -> bool {
-        self.hash() == other.hash()
+        if let (Some(root), Some(other_root)) = (self.cached_root(), other.cached_root()) {
+            // Use cached roots if available
+            root == other_root
+        } else {
+            // Avoid expensive root recalculations which use multiple cryptographic hashes
+            self.inner == other.inner
+        }
     }
 }
 
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -184,11 +184,19 @@ impl TryFrom<&[u8]> for Node {
     type Error = &'static str;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        Option::<pallas::Base>::from(pallas::Base::from_repr(
-            bytes.try_into().map_err(|_| "wrong byte slice len")?,
-        ))
-        .map(Node)
-        .ok_or("invalid Pallas field element")
+        <[u8; 32]>::try_from(bytes)
+            .map_err(|_| "wrong byte slice len")?
+            .try_into()
+    }
+}
+
+impl TryFrom<[u8; 32]> for Node {
+    type Error = &'static str;
+
+    fn try_from(bytes: [u8; 32]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(bytes))
+            .map(Node)
+            .ok_or("invalid Pallas field element")
     }
 }
 
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -178,18 +178,17 @@ impl Node {
     pub fn to_repr(&self) -> [u8; 32] {
         self.0.to_repr()
     }
+}
 
-    /// Converts a byte representation of a field element into an element of
-    /// this prime field without checking if the input is not canonical
-    pub fn from_repr_unchecked(bytes: &[u8]) -> Self {
-        let mut tmp = [0, 0, 0, 0];
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
 
-        tmp[0] = u64::from_le_bytes(bytes[0..8].try_into().unwrap());
-        tmp[1] = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
-        tmp[2] = u64::from_le_bytes(bytes[16..24].try_into().unwrap());
-        tmp[3] = u64::from_le_bytes(bytes[24..32].try_into().unwrap());
-
-        pallas::Base::from_raw(tmp).into()
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(
+            bytes.try_into().map_err(|_| "wrong byte slice len")?,
+        ))
+        .map(Node)
+        .ok_or("invalid Pallas field element")
     }
 }
 
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -18,7 +18,7 @@ use std::{
 };
 
 use bitvec::prelude::*;
-use bridgetree;
+use bridgetree::{self, NonEmptyFrontier};
 use halo2::pasta::{group::ff::PrimeField, pallas};
 use incrementalmerkletree::Hashable;
 use lazy_static::lazy_static;
@@ -31,7 +31,7 @@ use crate::{
     serialization::{
         serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
     },
-    subtree::{TRACKED_SUBTREE_HEIGHT, TRACKED_SUBTREE_SIZE},
+    subtree::TRACKED_SUBTREE_HEIGHT,
 };
 
 pub mod legacy;
@@ -178,6 +178,19 @@ impl Node {
     pub fn to_repr(&self) -> [u8; 32] {
         self.0.to_repr()
     }
+
+    /// Converts a byte representation of a field element into an element of
+    /// this prime field without checking if the input is not canonical
+    pub fn from_repr_unchecked(bytes: &[u8]) -> Self {
+        let mut tmp = [0, 0, 0, 0];
+
+        tmp[0] = u64::from_le_bytes(bytes[0..8].try_into().unwrap());
+        tmp[1] = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+        tmp[2] = u64::from_le_bytes(bytes[16..24].try_into().unwrap());
+        tmp[3] = u64::from_le_bytes(bytes[24..32].try_into().unwrap());
+
+        pallas::Base::from_raw(tmp).into()
+    }
 }
 
 /// Required to convert [`NoteCommitmentTree`] into [`SerializedTree`].
@@ -231,18 +244,6 @@ impl Hashable for Node {
 impl From<pallas::Base> for Node {
     fn from(x: pallas::Base) -> Self {
         Node(x)
-    }
-}
-
-impl TryFrom<&[u8]> for Node {
-    type Error = &'static str;
-
-    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        Option::<pallas::Base>::from(pallas::Base::from_repr(
-            bytes.try_into().map_err(|_| "wrong byte slice len")?,
-        ))
-        .map(Node)
-        .ok_or("invalid Pallas field element")
     }
 }
 
@@ -340,43 +341,46 @@ impl NoteCommitmentTree {
     }
 
     /// Returns true if the most recently appended leaf completes the subtree
-    pub fn is_complete_subtree(&self) -> bool {
-        self.inner.value().map_or(false, |x| {
-            x.position()
-                .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
-        })
+    pub fn is_complete_subtree(tree: &NonEmptyFrontier<Node>) -> bool {
+        tree.position()
+            .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
     }
 
-    /// Returns subtree index at [`TRACKED_SUBTREE_HEIGHT`]
-    pub fn subtree_index(&self) -> u16 {
-        (self.position() >> TRACKED_SUBTREE_HEIGHT)
-            .try_into()
-            .expect("always within u16")
-    }
-
-    /// Returns subtree root, if any
-    pub fn subtree_root(&self) -> Option<Node> {
-        self.is_complete_subtree().then_some(())?;
-        Some(
-            self.inner
-                .value()?
-                .root(Some(TRACKED_SUBTREE_HEIGHT.into())),
+    /// Returns subtree address at [`TRACKED_SUBTREE_HEIGHT`]
+    pub fn subtree_address(&self) -> incrementalmerkletree::Address {
+        incrementalmerkletree::Address::above_position(
+            TRACKED_SUBTREE_HEIGHT.into(),
+            self.position().into(),
         )
     }
 
-    /// Returns the index of note completing the next subtree
-    pub fn next_complete_subtree_note_index(&self) -> usize {
-        // # Correctness
-        //
-        // 2^16 - (position % 2^16) should always be between 1..=2^16
-        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+    /// Accepts a mutable reference to Orchard notes that are to be appended for a block,
+    /// checks if they will complete the current subtree, and splits off any notes that should
+    /// be appended to the next subtree.
+    ///
+    /// Returns current subtree index and notes in the next subtree if the notes complete
+    /// the current subtree.
+    ///
+    /// Returns None otherwise.
+    #[allow(clippy::unwrap_in_result)]
+    pub fn subtree_index_and_notes_in_next_subtree(
+        &self,
+        notes: &mut Vec<NoteCommitmentUpdate>,
+    ) -> Option<(u16, Vec<NoteCommitmentUpdate>)> {
+        let address = self.subtree_address();
+        let index = || address.index().try_into().expect("should fit in u16");
+        let empty_leaf_count = (u64::from(address.max_position() - self.position()))
             .try_into()
-            .expect("should fit in usize")
+            .expect("should fit in usize");
+
+        (notes.len() > empty_leaf_count).then(|| (index(), notes.split_off(empty_leaf_count)))
     }
 
-    /// Counts of note commitments added to the current subtree, max 2^16
-    fn subtree_note_count(&self) -> u64 {
-        self.count() % TRACKED_SUBTREE_SIZE
+    /// Returns subtree root if the most recently appended leaf completes the subtree
+    pub fn subtree_root(&self) -> Option<Node> {
+        let value = self.inner.value()?;
+        Self::is_complete_subtree(value).then_some(())?;
+        Some(value.root(Some(TRACKED_SUBTREE_HEIGHT.into())))
     }
 
     /// Returns the current root of the tree, used as an anchor in Orchard
@@ -436,7 +440,7 @@ impl NoteCommitmentTree {
 
     /// Position of final leaf added to the tree.
     fn position(&self) -> u64 {
-        self.inner.value().map_or(0, |x| u64::from(x.position()))
+        self.inner.value().map_or(0, |x| x.position().into())
     }
 
     /// Count of note commitments added to the tree.
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -346,40 +346,23 @@ impl NoteCommitmentTree {
     }
 
     /// Returns subtree address at [`TRACKED_SUBTREE_HEIGHT`]
-    pub fn subtree_address(&self) -> incrementalmerkletree::Address {
+    pub fn subtree_address(tree: &NonEmptyFrontier<Node>) -> incrementalmerkletree::Address {
         incrementalmerkletree::Address::above_position(
             TRACKED_SUBTREE_HEIGHT.into(),
-            self.position().into(),
+            tree.position(),
         )
     }
 
-    /// Accepts a mutable reference to Orchard notes that are to be appended for a block,
-    /// checks if they will complete the current subtree, and splits off any notes that should
-    /// be appended to the next subtree.
-    ///
-    /// Returns current subtree index and notes in the next subtree if the notes complete
-    /// the current subtree.
-    ///
-    /// Returns None otherwise.
+    /// Returns subtree index and root if the most recently appended leaf completes the subtree
     #[allow(clippy::unwrap_in_result)]
-    pub fn subtree_index_and_notes_in_next_subtree(
-        &self,
-        notes: &mut Vec<NoteCommitmentUpdate>,
-    ) -> Option<(u16, Vec<NoteCommitmentUpdate>)> {
-        let address = self.subtree_address();
-        let index = || address.index().try_into().expect("should fit in u16");
-        let empty_leaf_count = (u64::from(address.max_position() - self.position()))
-            .try_into()
-            .expect("should fit in usize");
-
-        (notes.len() > empty_leaf_count).then(|| (index(), notes.split_off(empty_leaf_count)))
-    }
-
-    /// Returns subtree root if the most recently appended leaf completes the subtree
-    pub fn subtree_root(&self) -> Option<Node> {
+    pub fn completed_subtree_index_and_root(&self) -> Option<(u16, Node)> {
         let value = self.inner.value()?;
         Self::is_complete_subtree(value).then_some(())?;
-        Some(value.root(Some(TRACKED_SUBTREE_HEIGHT.into())))
+        let address = Self::subtree_address(value);
+        let index = address.index().try_into().expect("should fit in u16");
+        let root = value.root(Some(TRACKED_SUBTREE_HEIGHT.into()));
+
+        Some((index, root))
     }
 
     /// Returns the current root of the tree, used as an anchor in Orchard
@@ -435,11 +418,6 @@ impl NoteCommitmentTree {
     /// Uncommitted^Orchard = I2LEBSP_l_MerkleOrchard(2)
     pub fn uncommitted() -> pallas::Base {
         pallas::Base::one().double()
-    }
-
-    /// Position of final leaf added to the tree.
-    fn position(&self) -> u64 {
-        self.inner.value().map_or(0, |x| x.position().into())
     }
 
     /// Count of note commitments added to the tree.
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -159,7 +159,7 @@ impl NoteCommitmentTrees {
     #[allow(clippy::unwrap_in_result)]
     fn update_sapling_note_commitment_tree(
         mut sapling: Arc<sapling::tree::NoteCommitmentTree>,
-        mut sapling_note_commitments: Vec<sapling::tree::NoteCommitmentUpdate>,
+        sapling_note_commitments: Vec<sapling::tree::NoteCommitmentUpdate>,
     ) -> Result<
         (
             Arc<sapling::tree::NoteCommitmentTree>,
@@ -168,12 +168,6 @@ impl NoteCommitmentTrees {
         NoteCommitmentTreeError,
     > {
         let sapling_nct = Arc::make_mut(&mut sapling);
-        let subtree_index_and_notes_in_next_subtree =
-            sapling_nct.subtree_index_and_notes_in_next_subtree(&mut sapling_note_commitments);
-
-        for sapling_note_commitment in sapling_note_commitments {
-            sapling_nct.append(sapling_note_commitment)?;
-        }
 
         // It is impossible for blocks to contain more than one level 16 sapling root:
         // > [NU5 onward] nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16.
@@ -186,17 +180,12 @@ impl NoteCommitmentTrees {
         // <https://zips.z.cash/protocol/protocol.pdf#txnencoding>
         let mut subtree_root = None;
 
-        if let Some((subtree_index, sapling_note_commitments)) =
-            subtree_index_and_notes_in_next_subtree
-        {
-            let node: sapling::tree::Node = sapling_nct
-                .subtree_root()
-                .expect("should be frontier after appending a note");
-            subtree_root = Some((subtree_index, node));
-
-            for sapling_note_commitment in sapling_note_commitments {
-                sapling_nct.append(sapling_note_commitment)?;
+        for sapling_note_commitment in sapling_note_commitments {
+            if let Some(index_and_node) = sapling_nct.completed_subtree_index_and_root() {
+                subtree_root = Some(index_and_node);
             }
+
+            sapling_nct.append(sapling_note_commitment)?;
         }
 
         // Re-calculate and cache the tree root.
@@ -209,7 +198,7 @@ impl NoteCommitmentTrees {
     #[allow(clippy::unwrap_in_result)]
     fn update_orchard_note_commitment_tree(
         mut orchard: Arc<orchard::tree::NoteCommitmentTree>,
-        mut orchard_note_commitments: Vec<orchard::tree::NoteCommitmentUpdate>,
+        orchard_note_commitments: Vec<orchard::tree::NoteCommitmentUpdate>,
     ) -> Result<
         (
             Arc<orchard::tree::NoteCommitmentTree>,
@@ -218,29 +207,18 @@ impl NoteCommitmentTrees {
         NoteCommitmentTreeError,
     > {
         let orchard_nct = Arc::make_mut(&mut orchard);
-        let subtree_index_and_notes_in_next_subtree =
-            orchard_nct.subtree_index_and_notes_in_next_subtree(&mut orchard_note_commitments);
-
-        for orchard_note_commitment in orchard_note_commitments {
-            orchard_nct.append(orchard_note_commitment)?;
-        }
 
         // It is impossible for blocks to contain more than one level 16 orchard root:
         // > [NU5 onward] nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16.
         // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
         let mut subtree_root = None;
 
-        if let Some((subtree_index, orchard_note_commitments)) =
-            subtree_index_and_notes_in_next_subtree
-        {
-            let node = orchard_nct
-                .subtree_root()
-                .expect("should be frontier after appending a note");
-            subtree_root = Some((subtree_index, node));
-
-            for orchard_note_commitment in orchard_note_commitments {
-                orchard_nct.append(orchard_note_commitment)?;
+        for orchard_note_commitment in orchard_note_commitments {
+            if let Some(index_and_node) = orchard_nct.completed_subtree_index_and_root() {
+                subtree_root = Some(index_and_node);
             }
+
+            orchard_nct.append(orchard_note_commitment)?;
         }
 
         // Re-calculate and cache the tree root.
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -1,14 +1,10 @@
 //! Parallel note commitment tree update methods.
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::sync::Arc;
 
 use thiserror::Error;
 
-use crate::{
-    block::{Block, Height},
-    orchard, sapling, sprout,
-    subtree::NoteCommitmentSubtree,
-};
+use crate::{block::Block, orchard, sapling, sprout, subtree::NoteCommitmentSubtree};
 
 /// An argument wrapper struct for note commitment trees.
 #[derive(Clone, Debug)]
@@ -60,7 +56,6 @@ impl NoteCommitmentTrees {
         let height = block
             .coinbase_height()
             .expect("height was already validated");
-        let block_list: BTreeMap<Height, Arc<Block>> = [(height, block)].into_iter().collect();
 
         // Prepare arguments for parallel threads
         let NoteCommitmentTrees {
@@ -70,21 +65,21 @@ impl NoteCommitmentTrees {
             ..
         } = self.clone();
 
-        let sprout_note_commitments: Vec<_> = block_list
-            .values()
-            .flat_map(|block| block.transactions.iter())
+        let sprout_note_commitments: Vec<_> = block
+            .transactions
+            .iter()
             .flat_map(|tx| tx.sprout_note_commitments())
             .cloned()
             .collect();
-        let sapling_note_commitments: Vec<_> = block_list
-            .values()
-            .flat_map(|block| block.transactions.iter())
+        let sapling_note_commitments: Vec<_> = block
+            .transactions
+            .iter()
             .flat_map(|tx| tx.sapling_note_commitments())
             .cloned()
             .collect();
-        let orchard_note_commitments: Vec<_> = block_list
-            .values()
-            .flat_map(|block| block.transactions.iter())
+        let orchard_note_commitments: Vec<_> = block
+            .transactions
+            .iter()
             .flat_map(|tx| tx.orchard_note_commitments())
             .cloned()
             .collect();
@@ -164,7 +159,7 @@ impl NoteCommitmentTrees {
     #[allow(clippy::unwrap_in_result)]
     fn update_sapling_note_commitment_tree(
         mut sapling: Arc<sapling::tree::NoteCommitmentTree>,
-        sapling_note_commitments: Vec<sapling::tree::NoteCommitmentUpdate>,
+        mut sapling_note_commitments: Vec<sapling::tree::NoteCommitmentUpdate>,
     ) -> Result<
         (
             Arc<sapling::tree::NoteCommitmentTree>,
@@ -173,7 +168,12 @@ impl NoteCommitmentTrees {
         NoteCommitmentTreeError,
     > {
         let sapling_nct = Arc::make_mut(&mut sapling);
-        let next_complete_subtree_note_index = sapling_nct.next_complete_subtree_note_index();
+        let subtree_index_and_notes_in_next_subtree =
+            sapling_nct.subtree_index_and_notes_in_next_subtree(&mut sapling_note_commitments);
+
+        for sapling_note_commitment in sapling_note_commitments {
+            sapling_nct.append(sapling_note_commitment)?;
+        }
 
         // It is impossible for blocks to contain more than one level 16 sapling root:
         // > [NU5 onward] nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16.
@@ -186,19 +186,14 @@ impl NoteCommitmentTrees {
         // <https://zips.z.cash/protocol/protocol.pdf#txnencoding>
         let mut subtree_root = None;
 
-        if sapling_note_commitments.len() > next_complete_subtree_note_index {
-            for (note_idx, sapling_note_commitment) in
-                sapling_note_commitments.into_iter().enumerate()
-            {
-                sapling_nct.append(sapling_note_commitment)?;
-                if note_idx == next_complete_subtree_note_index {
-                    let node = sapling_nct
-                        .subtree_root()
-                        .expect("should be frontier after appending a note");
-                    subtree_root = Some((sapling_nct.subtree_index(), node));
-                }
-            }
-        } else {
+        if let Some((subtree_index, sapling_note_commitments)) =
+            subtree_index_and_notes_in_next_subtree
+        {
+            let node: sapling::tree::Node = sapling_nct
+                .subtree_root()
+                .expect("should be frontier after appending a note");
+            subtree_root = Some((subtree_index, node));
+
             for sapling_note_commitment in sapling_note_commitments {
                 sapling_nct.append(sapling_note_commitment)?;
             }
@@ -214,7 +209,7 @@ impl NoteCommitmentTrees {
     #[allow(clippy::unwrap_in_result)]
     fn update_orchard_note_commitment_tree(
         mut orchard: Arc<orchard::tree::NoteCommitmentTree>,
-        orchard_note_commitments: Vec<orchard::tree::NoteCommitmentUpdate>,
+        mut orchard_note_commitments: Vec<orchard::tree::NoteCommitmentUpdate>,
     ) -> Result<
         (
             Arc<orchard::tree::NoteCommitmentTree>,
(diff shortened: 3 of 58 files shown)


arya2:
This PR is ready for review, but it needs the migration before merging (it needs to suspend finalized block commits until it has the start height for the first subtrees after the initial tip).

The migration should be quick, it can find `subtree_index`, `next_start_height` and `end_height` by checking only unique note commitment tree frontiers, and should only have to read the transactions for ~1400 blocks to find the subtree roots.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -138,6 +146,17 @@ impl WriteDisk for DiskWriteBatch {
...
 
+    fn zs_merge<C, K, V>(&mut self, cf: &C, key: K, value: V)
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + Debug,
+        V: IntoDisk,
+    {
+        let key_bytes = key.as_bytes();
+        let value_bytes = value.as_bytes();
+        self.batch.merge_cf(cf, key_bytes, value_bytes);
```
arya2: This PR may need to set a merge operator for this to work.

zebra-chain/src/parallel/tree.rs
```diff
@@ -44,37 +52,35 @@ impl NoteCommitmentTrees {
...
+            .expect("height was already validated");
+        let block_list: BTreeMap<Height, Arc<Block>> = [(height, block)].into_iter().collect();
+
         // Prepare arguments for parallel threads
         let NoteCommitmentTrees {
             sprout,
             sapling,
             orchard,
+            sapling_subtree: _,
+            orchard_subtree: _,
```
arya2:
```suggestion
            ..
```

zebra-chain/src/parallel/tree.rs
```diff
@@ -132,12 +138,44 @@ impl NoteCommitmentTrees {
...
+            self.orchard = orchard;
+            self.orchard_subtree =
+                subtree_root.map(|(subtree_root, starts_subtree, subtree_index)| {
+                    orchard::tree::NoteCommitmentSubtree {
+                        index: subtree_index,
+                        end: height,
+                        next_start: if starts_subtree {
+                            height
+                        } else {
+                            height.next()
```
arya2: This should be the next block with notes, not always the next block height.

I'll make this a flag and get the height of the next note commitment tree before writing to disk.

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +380,23 @@ impl NoteCommitmentTree {
...
+
+    /// Returns subtree index
+    pub fn subtree_index<const DEPTH: u8>(&self) -> u64 {
+        self.position() >> DEPTH
+    }
+
+    /// Returns subtree root, if any
+    pub fn subtree_root<const DEPTH: u8>(&self) -> Option<Node> {
+        Some(self.inner.value()?.root(Some(DEPTH.into())))
+    }
```
upbqdn: This should check if the frontier is a complete subtree before attempting to calculate the root

zebra-chain/src/parallel/tree.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7334/commits/5dd93ec4f0cd3eb3da9bac87b5bf751e72cfc780.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Removed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/5dd93ec4f0cd3eb3da9bac87b5bf751e72cfc780.

zebra-chain/src/parallel/tree.rs
arya2: Converted to a flag in https://github.com/ZcashFoundation/zebra/pull/7334/commits/5dd93ec4f0cd3eb3da9bac87b5bf751e72cfc780.

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,116 @@ impl NoteCommitmentTrees {
...
+            sapling_note_commitments.len() as u64 >= distance_to_complete_subtree;
+
+        // MAX_BLOCK_BYTES is less than 2^21 bytes, notes are 2^5 bytes, and a full subtree has
+        // 2^16 notes. Blocks should not contain multiple level 16 roots.
+        let mut subtree_root = None;
+
+        if should_find_subtree_roots {
+            let starts_new_subtree =
+                sapling_note_commitments.len() as u64 != distance_to_complete_subtree;
+            let complete_subtree_index = distance_to_complete_subtree as usize - 1;
```
upbqdn: I find the `complete_subtree_index` name confusing since it isn't the index of the subtree.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -178,6 +178,44 @@ impl ZebraDb {
...
+    #[allow(clippy::unwrap_in_result)]
+    pub fn sapling_subtree_start_height_by_index(&self, index: u16) -> Option<Height> {
+        let sapling_subtrees = self
+            .db
+            .cf_handle("sapling_note_commitment_subtree")
+            .unwrap();
+
+        let start: Height = self.db.zs_get(&sapling_subtrees, &index)?;
+
+        Some(start)
```
arya2:
```suggestion
        self.db.zs_get(&sapling_subtrees, &index)
```

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -190,6 +228,44 @@ impl ZebraDb {
...
+    #[allow(clippy::unwrap_in_result)]
+    pub fn orchard_subtree_start_height_by_index(&self, index: u16) -> Option<Height> {
+        let orchard_subtrees = self
+            .db
+            .cf_handle("orchard_note_commitment_subtree")
+            .unwrap();
+
+        let start: Height = self.db.zs_get(&orchard_subtrees, &index)?;
+
+        Some(start)
```
arya2:
```suggestion
        self.db.zs_get(&orchard_subtrees, &index)
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
+#[derive(Clone, Debug)]
+pub struct NoteCommitmentSubtree {
+    /// Index of this subtree
+    pub index: u16,
+    /// Start boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// Start boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+    pub start: Option<block::Height>,
+    /// End boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// End boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+    pub height: block::Height,
+}
+
+/// The block height of the last leaf of a subtree and the subtree root
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct NoteCommitmentSubtreeEnd {
+    /// Index of this subtree
+    #[serde(skip_serializing)]
+    pub index: u16,
+    /// End boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// End boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -287,6 +290,71 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
+#[derive(Clone, Debug)]
+pub struct NoteCommitmentSubtree {
+    /// Index of this subtree
+    pub index: u16,
+    /// Start boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// Start boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -287,6 +290,71 @@ pub struct NoteCommitmentTree {
...
+    pub start: Option<block::Height>,
+    /// End boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// End boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -287,6 +290,71 @@ pub struct NoteCommitmentTree {
...
+    pub height: block::Height,
+}
+
+/// The block height of the last leaf of a subtree and the subtree root
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct NoteCommitmentSubtreeEnd {
+    /// Index of this subtree
+    #[serde(skip_serializing)]
+    pub index: u16,
+    /// End boundary of the this subtree, the block height of its last leaf.
```
upbqdn:
```suggestion
    /// End boundary of this subtree, the block height of its last leaf.
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -367,6 +452,11 @@ impl NoteCommitmentTree {
         jubjub::Fq::one().to_bytes()
     }
 
+    /// Position of final leaf added to the tree.
+    pub fn position(&self) -> u64 {
```
upbqdn:
```suggestion
    fn position(&self) -> u64 {
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -372,6 +457,11 @@ impl NoteCommitmentTree {
         pallas::Base::one().double()
     }
 
+    /// Position of final leaf added to the tree.
+    pub fn position(&self) -> u64 {
```
upbqdn:
```suggestion
    fn position(&self) -> u64 {
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -312,6 +380,23 @@ impl NoteCommitmentTree {
...
+            .map_or(false, |x| x.position().is_complete_subtree(DEPTH.into()))
+    }
+
+    /// Returns subtree index
+    pub fn subtree_index<const DEPTH: u8>(&self) -> u64 {
+        self.position() >> DEPTH
+    }
+
+    /// Returns subtree root, if any
+    pub fn subtree_root<const DEPTH: u8>(&self) -> Option<Node> {
```
upbqdn: This should check if the frontier is a complete subtree before attempting to calculate the root

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -113,7 +113,7 @@ pub trait WriteDisk {
         K: IntoDisk + Debug,
         V: IntoDisk;
 
-    /// Remove the given key from rocksdb column family if it exists.
+    /// Remove the given key form rocksdb column family if it exists.
```
upbqdn: Shouldn't it be from?

```suggestion
    /// Remove the given key from rocksdb column family if it exists.
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -663,6 +677,22 @@ impl Chain {
             .map(|(_height, tree)| tree.clone())
     }
 
+    /// Returns the Sapling [`NoteCommitmentSubtree`](sapling::tree::NoteCommitmentSubtree) specified
+    /// by a [`HashOrHeight`], if it exists in the non-finalized [`Chain`].
+    pub fn sapling_subtree(
```
upbqdn, line 683: Will we have subtrees in the non-finalized state? Given their width at the leaf level, we would have at most one at a time.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -190,6 +228,44 @@ impl ZebraDb {
...
+            index,
+            start: Some(start),
+            end,
+            begins_next_subtree: None,
+            subtree_root,
+        }))
+    }
+
+    /// Returns the Orchard note commitment tree of the finalized tip
+    /// or the empty tree if the state is empty.
```
upbqdn: This should be refactored.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -190,6 +228,44 @@ impl ZebraDb {
             .expect("Orchard note commitment tree must exist if there is a finalized tip")
     }
 
+    /// Returns the Orchard note commitment tree of the finalized tip
+    /// or the empty tree if the state is empty.
```
upbqdn: This should be refactored.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -178,6 +178,44 @@ impl ZebraDb {
...
+            index,
+            start: Some(start),
+            end,
+            begins_next_subtree: None,
+            subtree_root,
+        }))
+    }
+
+    /// Returns the Sapling note commitment tree of the finalized tip
+    /// or the empty tree if the state is empty.
```
upbqdn: This should be refactored.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -178,6 +178,44 @@ impl ZebraDb {
         Some(Arc::new(tree))
     }
 
+    /// Returns the Sapling note commitment tree of the finalized tip
+    /// or the empty tree if the state is empty.
```
upbqdn: This should be refactored.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +436,92 @@ impl DiskWriteBatch {
...
+        if let Some(subtree) = trees.sapling_subtree {
+            let (subtree_end, subtree_start) = subtree.bounds();
+
+            let start_height = if subtree_end.index == 0 {
+                let (start_height, _) = zebra_db
+                    .sapling_tree_by_height_range(Height(1)..)
+                    .next()
+                    .expect("must have sapling notes on disk");
+
+                start_height
```
upbqdn: I think this might be wrong. We need to check at which height the subtrees actually start.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +436,92 @@ impl DiskWriteBatch {
...
+                (start_height, subtree_end.height, subtree_end.root),
+            );
+
+            if let Some(subtree_start) = subtree_start {
+                self.zs_insert(
+                    &sapling_subtree_cf,
+                    subtree_start.index,
+                    subtree_start.height,
+                );
+            }
```
upbqdn: Is it OK to write two different data types into the same column family?

upbqdn (commented):
I'll need to give this another pass. I'm posting my notes so far.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +432,92 @@ impl DiskWriteBatch {
...
+                    .sapling_subtree_by_index(subtree_end.index - 1)
+                    .expect("must have previous subtree")
+                    .end;
+
+                let (start_height, _) = zebra_db
+                    .sapling_tree_by_height_range(prev_subtree_end_height.next()..)
+                    .next()
+                    .expect("must have previous subtree");
+
+                start_height
```
upbqdn: Unfortunately, I can't see how this determines the right start height. Won't `sapling_tree_by_height_range` return the given height?

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: The idea here is that if it's the first sapling subtree, its start height is the first height with a unique sapling note commitment treestate?

If we're not storing duplicate note commitment trees, the first entry after height(0) should be the height of the first sapling note.

I'll add this in a comment if that's right.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: Ah this definitely needs a comment.

So if there isn't an entry at this index with just a block height (the start height, entered when a block completes a subtree and starts a new one), then the start height of this subtree should be after the end height of the previous subtree and it looks for the height of the first note commitment tree entry after the end height of the previous subtree.

> Won't sapling_tree_by_height_range return the given height

It should return the height of the first sapling note at or above the given height.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: I think it should be okay.

It should fail to deserialize and return None if we use the wrong method.

zebra-state/src/service/non_finalized_state/chain.rs
arya2: > Will we have subtrees in the non-finalized state?

Only for finalizing after they reach the root of the non-finalized state.

It was easier/faster to get the root while updating the treestate.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Oh, I forgot we deduplicated the trees. LOL

zebra-chain/src/sapling/tree.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/bccc85a938272bddbe0f476e6e44ffde640c5641.

arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/bccc85a938272bddbe0f476e6e44ffde640c5641.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/1bd86bcb3cdbcf2326318d5157c2ea10dd1b453e.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: Added comments in https://github.com/ZcashFoundation/zebra/pull/7334/commits/aea9320805ed46f7fd58866644b792267592ea7b.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: Added comments in https://github.com/ZcashFoundation/zebra/pull/7334/commits/aea9320805ed46f7fd58866644b792267592ea7b.

zebra-chain/src/orchard/tree.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/bccc85a938272bddbe0f476e6e44ffde640c5641.

zebra-chain/src/parallel/tree.rs
arya2: How about `complete_subtree_note_index`?

upbqdn: Sounds good. Or we could even compute the value directly in the `if` condition since we don't use it anywhere else?

upbqdn: > Or we could even compute the value directly in the if condition since we don't use it anywhere else?

Assuming the compiler will optimize the subtraction and do it only once.

zebra-chain/src/sapling/tree.rs
```diff
@@ -287,6 +290,71 @@ pub struct NoteCommitmentTree {
...
+                begins_next_subtree.then_some(NoteCommitmentSubtreeStart {
+                    index: index + 1,
+                    height: end,
+                })
+            }),
+        )
+    }
+}
+
+/// The block height of the first leaf of a subtree
```
upbqdn: I think `NoteCommitmentSubtreeStart` always refers to the next subtree.

Maybe something like this?

```suggestion
/// The index and block height of the next subtree.
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+                begins_next_subtree.then_some(NoteCommitmentSubtreeStart {
+                    index: index + 1,
+                    height: end,
+                })
+            }),
+        )
+    }
+}
+
+/// The block height of the first leaf of a subtree
```
upbqdn: I think `NoteCommitmentSubtreeStart` always refers to the next subtree.

Maybe something like this?

```suggestion
/// The index and block height of the next subtree.
```

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
upbqdn: Should we add an assertion that checks that the index of the subtree start is always greater than any other index in the database so that we don't overwrite stored subtrees?

zebra-chain/src/parallel/tree.rs
arya2: Renamed in https://github.com/ZcashFoundation/zebra/pull/7334/commits/70387762886e8af443972ee5629363fbdac52b70.

zebra-chain/src/sapling/tree.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7334/commits/654b40fc16e2088e85b9d4f3dcbd453c123d0154.

zebra-chain/src/orchard/tree.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7334/commits/654b40fc16e2088e85b9d4f3dcbd453c123d0154.

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
+#[derive(Clone, Debug)]
+pub struct NoteCommitmentSubtree {
+    /// Index of this subtree
+    pub index: u16,
+    /// Start boundary of this subtree, the block height of its last leaf.
+    pub start: Option<block::Height>,
+    /// End boundary of this subtree, the block height of its last leaf.
+    pub end: block::Height,
```
teor2345: There seem to be some inconsistencies in the field names and docs in this struct:
1. "last leaf" in `start`'s docs looks like a copy-paste typo.
2. Either the `start` field name is wrong, or the struct documentation is wrong. If the field is the start of the next subtree, I'd suggest putting it after the end of the current subtree, and renaming both fields to avoid ambiguity.

I'd suggest also making these changes to the start and end structs.

```suggestion
    /// End boundary of the current subtree, the block height of its last leaf.
    pub current_tree_end: block::Height,
    /// Start boundary of the next subtree, the block height of the next subtree's first leaf.
    pub next_tree_start: Option<block::Height>,
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+    /// Whether this subtree's end height is the start height of the next subtree.
+    pub begins_next_subtree: Option<bool>,
```
teor2345: Nit: Can `begins_next_subtree` be a method instead?

```rust
fn begins_next_subtree(&self) -> Option<bool> {
    let next_tree_start = self.next_tree_start?;
    Some(self.current_tree_end == next_tree_start)
}
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
+#[derive(Clone, Debug)]
```
teor2345: Nit: Please derive all useful traits on new structs.

I'm not sure if `PartialEq` should compare the entire struct, or just the `index` (or just `index` and `root`). But if we're mainly going to use it in tests, maybe it should compare all the fields.

(Maybe it would help to create a self-review checklist for these kinds of things?)

```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+pub struct NoteCommitmentSubtree {
```
teor2345: Optional: Do `next_tree_start` and `begins_next_subtree` need to be optional?

#### `next_tree_start`

If `next_tree_start` is always either `current_tree_end` or `current_tree_end + 1`, then it doesn't have to be optional. Because we always have the complete block in the state, we can decide which value we need even if we don't have the next block.

Either way, please document the constraints on the value of `next_tree_start`. (Is it `next_tree_start == current_tree_end || next_tree_start == current_tree_end + 1` or `next_tree_start >= next_tree_start`.)

#### `begins_next_subtree`

Since we always know if `next_tree_start == current_tree_end `, even if `current_tree_end` is optional, we should always be able to answer `begins_next_subtree` with `true` (they are equal) or `false` (either they aren't equal, or `next_tree_start` is `None`, but still known to be not equal to `current_tree_end`).

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
```
teor2345: Nit: This documentation seems incomplete. Can you start it with a one-sentence summary of the concept the struct represents or how it is typically used?

(I remember mentioning this in a recent review, maybe it would help to create a self-review checklist for these kinds of things?)

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+                    index: index + 1,
+                    height: end,
+                })
+            }),
+        )
+    }
+}
+
+/// The index and block height of the next subtree.
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
teor2345: Nit:

```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -292,6 +295,71 @@ pub struct NoteCommitmentTree {
...
+pub struct NoteCommitmentSubtreeStart {
+    /// Index of this subtree
+    #[serde(skip_serializing)]
+    pub index: u16,
+    /// Start boundary of the next subtree, the block height of its first leaf.
+    pub height: block::Height,
+}
+
+/// The block height of the last leaf of a subtree and the subtree root
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
teor2345: Nit:

```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +385,24 @@ impl NoteCommitmentTree {
...
+    /// Returns subtree index
+    pub fn subtree_index<const DEPTH: u8>(&self) -> u64 {
+        self.position() >> DEPTH
+    }
+
+    /// Returns subtree root, if any
+    pub fn subtree_root<const DEPTH: u8>(&self) -> Option<Node> {
+        self.is_complete_subtree::<DEPTH>().then_some(())?;
+        Some(self.inner.value()?.root(Some(DEPTH.into())))
+    }
```
teor2345: Optional: Since `DEPTH` is always the constant `TRACKED_SUBTREE_HEIGHT`, it doesn't seem necessary to make it a generic argument. Can we use `TRACKED_SUBTREE_HEIGHT` directly in these methods?

If we don't, it will be possible to call these functions with an incorrect depth, and reviewers might not pick up the mistake.

zebra-chain/src/sapling/tree.rs
```diff
@@ -287,6 +290,71 @@ pub struct NoteCommitmentTree {
     cached_root: std::sync::RwLock<Option<Root>>,
 }
 
+/// The block height of the final leaf in a level 16 subtree and
+/// the block height of the first leaf of the next subtree
+#[derive(Clone, Debug)]
+pub struct NoteCommitmentSubtree {
```
teor2345: Optional: It seems like these 3 new `sapling` structs are identical to the `orchard` versions, except that the `Node` type is different. Duplicating this code makes them harder to maintain.

Can we move this common code to another module, and add a generic `Node` argument? Then we can add type aliases for `Orchard` and `Sapling` with the correct concrete `Node` type.

To make sure that `sapling::NoteCommitmentSubtreeStart` and `orchard::NoteCommitmentSubtreeStart` are distinct types, you can add a `PhantomData<Node>` field:
https://doc.rust-lang.org/stable/std/marker/struct.PhantomData.html

zebra-state/src/service/finalized_state/disk_format/shielded.rs
```diff
@@ -140,3 +149,39 @@ impl FromDisk for orchard::tree::NoteCommitmentTree {
             .expect("deserialization format should match the serialization format used by IntoDisk")
     }
 }
+
+impl FromDisk for (block::Height, block::Height, sapling::tree::Node) {
```
teor2345: Optional: You could define a named wrapper struct or serialize the `NoteCommitmentSubtree` struct, rather than serializing a tuple. It makes the code easier to review, and avoids mistakes in future refactors.

zebra-state/src/service/finalized_state/disk_format/shielded.rs
```diff
@@ -74,6 +74,15 @@ impl IntoDisk for orchard::tree::Root {
     }
 }
 
+// TODO: Use or find a wrapper type?
```
teor2345: Please create a `NoteCommitmentSubtreeIndex` wrapper type.

Otherwise, any use of `u16` in the database will automatically use this format, even if that's not the most appropriate format for that field.

zebra-state/src/service/finalized_state/disk_format/shielded.rs
```diff
@@ -140,3 +149,39 @@ impl FromDisk for orchard::tree::NoteCommitmentTree {
             .expect("deserialization format should match the serialization format used by IntoDisk")
     }
 }
+
+impl FromDisk for (block::Height, block::Height, sapling::tree::Node) {
+    fn from_bytes(bytes: impl AsRef<[u8]>) -> Self {
+        bincode::DefaultOptions::new()
```
teor2345: I strongly recommend against using `bincode` for serialization. It encodes the field order and wrappers/collections in the structure that's being serialized, even if they aren't significant. It also uses up extra redundant space.

`bincode` was the cause of the breaking change in the `NoteCommitmentTree` state format during the update to the `zcashd` 5.6.0 dependencies. Since the underlying type of `Node` is not under our control, any refactor to the inner fields of that type by the `zcashd` team would also be a breaking state format change.

I think this might be a blocker to merging this PR.

Instead, convert the fields to bytes using a stable serialization format.
For example, see `TransactionLocation`:
https://github.com/ZcashFoundation/zebra/blob/92dd5285e6837eb4f0cafcc25b7b65dc3fe623ec/zebra-state/src/service/finalized_state/disk_format/block.rs#L277-L297

Since this is a format change, we will also need to delete all the cached states with the `v25.2.0` format before pushing the format change:
[here is a list of those states](https://console.cloud.google.com/compute/images?tab=images&hl=en&project=zfnd-dev-zebra&pageState=(%22images%22:(%22s%22:%5B(%22i%22:%22creator%22,%22s%22:%221%22),(%22i%22:%22type%22,%22s%22:%220%22),(%22i%22:%22name%22,%22s%22:%220%22)%5D,%22f%22:%22%255B%257B_22k_22_3A_22Created%2520by_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22zfnd-dev-zebra_5C_22_22_2C_22i_22_3A_22creator_22%257D_2C%257B_22k_22_3A_22Labels_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22v25-2_5C_22_22_2C_22i_22_3A_22labels_22%257D%255D%22)))

I just did that now so the `v25.1` states wouldn't get deleted by the scheduled cleanup task. If you push any other changes to this PR before making the format change, those states will need to be deleted again.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -178,6 +178,40 @@ impl ZebraDb {
...
+
+    /// Returns the Sapling note commitment subtree start height at this index
+    #[allow(clippy::unwrap_in_result)]
+    pub fn sapling_subtree_start_height_by_index(&self, index: u16) -> Option<Height> {
+        let sapling_subtrees = self
+            .db
+            .cf_handle("sapling_note_commitment_subtree")
+            .unwrap();
+
+        self.db.zs_get(&sapling_subtrees, &index)
```
teor2345: Update:

This column family seems to have two different types written to it, but correct usage is not documented, tested, or asserted anywhere.

Whatever we do here needs to be done for orchard as well.

I think this might be a blocker to merging this PR.

-----

#### Potentially Outdated Comments

This implementation will only work if `start` is the first field, and the field encodings are the same.~~

This is not actually true for the implementation in this PR, and there are no tests for it. (`bincode` encodes `Height` in 4 bytes, possibly with a wrapper struct, tuple, or tuple index prefix. `Height as FromDisk` encodes height in 3 bytes with no prefix.)

I think this might be a blocker to merging this PR.

Instead, I'd recommend using a simpler implementation that doesn't rely on undocumented format internals:
```suggestion
        let sapling_subtree = self.sapling_subtree_by_index(index)?;
        Some(sapling_subtree.start)
```

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/empty_column_families@mainnet_2.snap
```diff
@@ -4,7 +4,9 @@ expression: empty_column_families
 ---
 [
   "history_tree: no entries",
+  "orchard_note_commitment_subtree: no entries",
   "orchard_nullifiers: no entries",
+  "sapling_note_commitment_subtree: no entries",
```
teor2345, line 9: The automatic snapshots do not contain any serialized note commitment subtree data, so please add tests for the format:
- [x] a raw bytes data format snapshot test at the `disk_format.rs` level, either using `insta` snapshots, or manually, like the note commitment tree snapshots:
https://github.com/ZcashFoundation/zebra/blob/92dd5285e6837eb4f0cafcc25b7b65dc3fe623ec/zebra-state/src/service/finalized_state/tests/vectors.rs#L142-L177
- [x] a raw bytes data format serialization round-trip test at the `disk_format.rs` level, like:
https://github.com/ZcashFoundation/zebra/blob/92dd5285e6837eb4f0cafcc25b7b65dc3fe623ec/zebra-state/src/service/finalized_state/disk_format/tests/prop.rs#L162-L172
- [x] a structured data format test using `NoteCommitmentSubtree*` type snapshots

zebra-chain/src/sapling/tree.rs
teor2345: If you don't want to make these structs generic over `Node` (which is totally ok), then all my comments about the orchard types also apply to the corresponding sapling types.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: Storing two different types into the same column family will work, but it's a design that's harder to understand and maintain. And this design isn't documented anywhere.

Currently the serialization will panic if the wrong type is read:
https://github.com/ZcashFoundation/zebra/pull/7334/files#diff-59729ef56affcfc84beb78f723e8b00019f4001daa999b69b0e39cbc9a0df657R163-R165
https://github.com/ZcashFoundation/zebra/blob/92dd5285e6837eb4f0cafcc25b7b65dc3fe623ec/zebra-state/src/service/finalized_state/disk_format.rs#L150-L153

Instead, we might be able to read the start of the next subtree from disk using the subtree and tree column families:
1. read the latest complete subtree from the subtree column family
2. if that subtree's `begins_next_subtree` is true, then use its `end` as the start of this subtree
3. otherwise, find the height of the next (full) note commitment tree entry after `end` in the note commitment tree column family, and use that

Or as an alternative, create another column family that stores the start height of the current subtree. (It will always have a single entry.) Column families are cheap, state bugs and maintenance can be expensive.

How does `zcashd` do it?

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +428,102 @@ impl DiskWriteBatch {
...
+            let start_height = if subtree_end.index == 0 {
+                // the first entry after Height(0) should be the height of the first Sapling note
+                let (start_height, _) = zebra_db
+                    .sapling_tree_by_height_range(Height(1)..)
+                    .next()
+                    .expect("must have sapling notes on disk");
+
+                start_height
+            } else if let Some(start_height) =
+                zebra_db.sapling_subtree_start_height_by_index(subtree_end.index)
```
teor2345: This start height won't be present if Zebra has opened an existing `v25.1` database, until that database has been fully upgraded to `v25.2`.

How can we make sure the upgrade runs before this code runs?
Edit: Currently upgrades run in parallel with new block commits, each in their own threads.

Can we merge this PR without that upgrade code? Or do we need to merge the upgrade code first?

This seems like it might be a blocker for this PR.

The details in this comment will become obsolete if we don't store the start height in the column family, but the data dependency is still an issue for the alternative algorithm.

zebra-chain/src/parallel/tree.rs
```diff
@@ -44,37 +52,34 @@ impl NoteCommitmentTrees {
...
-                block
-                    .coinbase_height()
-                    .expect("height was already validated"),
-                block.clone(),
-            )]
-            .into_iter()
-            .collect(),
-        )
+        self.update_trees_parallel_list(block.clone())
     }
```
teor2345: Nit: Since this PR changes this API so it can't take multiple blocks, I'd suggest moving the  contents of `update_trees_parallel_list` into `update_trees_parallel`. (If there's no list of blocks, then we shouldn't have a method called `list`.)

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -133,6 +133,8 @@ pub struct Chain {
     /// When a chain is forked from the finalized tip, also contains the finalized tip root.
     /// This extra root is removed when the first non-finalized block is committed.
     pub(crate) sapling_anchors_by_height: BTreeMap<block::Height, sapling::tree::Root>,
+    /// The Sapling subtree completed in the non-finalized state
```
teor2345: Nit:

```suggestion
    /// A list of Sapling subtrees completed in the non-finalized state
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -144,6 +146,8 @@ pub struct Chain {
     /// When a chain is forked from the finalized tip, also contains the finalized tip root.
     /// This extra root is removed when the first non-finalized block is committed.
     pub(crate) orchard_anchors_by_height: BTreeMap<block::Height, orchard::tree::Root>,
+    /// The Orchard subtree completed in the non-finalized state
```
teor2345: Nit:

```suggestion
    /// A list of Orchard subtrees completed in the non-finalized state
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -133,6 +133,8 @@ pub struct Chain {
     /// When a chain is forked from the finalized tip, also contains the finalized tip root.
     /// This extra root is removed when the first non-finalized block is committed.
     pub(crate) sapling_anchors_by_height: BTreeMap<block::Height, sapling::tree::Root>,
+    /// The Sapling subtree completed in the non-finalized state
+    pub(crate) sapling_subtree: VecDeque<Arc<sapling::tree::NoteCommitmentSubtree>>,
```
teor2345: Optional: Would it be easier to prune the sapling and orchard trees if they were in a `BTreeMap` indexed by end height?

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -663,6 +677,22 @@ impl Chain {
...
+        &self,
+        hash_or_height: HashOrHeight,
+    ) -> Option<Arc<sapling::tree::NoteCommitmentSubtree>> {
+        let height =
+            hash_or_height.height_or_else(|hash| self.height_by_hash.get(&hash).cloned())?;
+
+        self.sapling_subtree
+            .iter()
+            .take(1)
+            .find(|subtree| subtree.end == height)
```
teor2345: Are we guaranteed that there will only ever be one subtree in the non-finalized state?
If we are, please document or assert that guarantee here, and fix the orchard code to match.

If not, let's remove `take(1)`:
```suggestion
        self.sapling_subtree
            .iter()
            .find(|subtree| subtree.end == height)
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -517,10 +517,12 @@ impl DiskDb {
         "sapling_nullifiers",
         "sapling_anchors",
         "sapling_note_commitment_tree",
+        "sapling_note_commitment_subtree",
```
teor2345, line 520: When you add a new column family, please update the database format documentation:

It's ok to just add another row to the table, the detailed format docs don't need to be updated for new column families.

teor2345 (changes requested):
This looks like a good start, thank you for taking care with the details of the implementation.

#### Overall

I think the current code has some blocking bugs, but it's hard for me to tell. I'm not sure if I have the background to review it in detail, particularly without any specifications.

Some of this code is consensus-critical for wallets, so it would be helpful to have links to the protocol documentation, or the corresponding `zcashd` code. (I understand that some of the requirements are undocumented, which makes implementing them tricky.)

This is a complex PR, which has taken me a while to review. I'm not sure if it's worth splitting now, but if you find yourself making changes to the non-finalized state, it might be worth splitting that out and getting it merged first.

It might also be helpful to do the state upgrade as a separate PR, which merges before this PR. (Due to the data dependency between the upgrade and adding subtrees for the latest block.)

#### Background

It might help to re-read the state upgrade documentation before reading my review:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#upgrading-the-state-database

#### Testing

I'm not sure if you missed the testing checklist in ticket #6953, or if you're waiting until after the state upgrade is implemented.

If the implementation and testing steps in the ticket are unclear, or it's missing steps, please let me know, and I'll update the ticket template. (Which I'll be creating today.)

#### Optional Changes & Nitpicks

Since I've made a lot of comments, I'm going to resolve the optional changes and nitpicks that aren't related to any potential bugs. Feel free to make those changes now, in a separate PR, or skip them entirely.

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+        NoteCommitmentTreeError,
+    > {
         let sapling_nct = Arc::make_mut(&mut sapling);
+        let distance_to_complete_subtree =
+            (1 << TRACKED_SUBTREE_HEIGHT) - sapling_nct.count() % (1 << TRACKED_SUBTREE_HEIGHT);
+        let should_find_subtree_roots =
+            sapling_note_commitments.len() as u64 >= distance_to_complete_subtree;
+
+        // MAX_BLOCK_BYTES is less than 2^21 bytes, notes are 2^5 bytes, and a full subtree has
+        // 2^16 notes. Blocks should not contain multiple level 16 roots.
```
teor2345: This limit is an absolute consensus rule requirement. In specifications, "should" means "usually, but there can be exceptions".

Notes are actually 580 bytes, and note commitments are 32 bytes, but their exact sizes are irrelevant due to the consensus rule (after NU5), and the size of Sapling outputs (before NU5).

```suggestion
        // It is impossible for blocks to contain more than one level 16 sapling root:
        // > [NU5 onward] nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16.
        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
        //
        // Before NU5, this limit holds due to the minimum size of Sapling outputs (948 bytes)
        // and the maximum size of a block:
        // > The size of a block MUST be less than or equal to 2000000 bytes.
        // <https://zips.z.cash/protocol/protocol.pdf#blockheader>
        // <https://zips.z.cash/protocol/protocol.pdf#txnencoding>
```

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+        NoteCommitmentTreeError,
+    > {
         let orchard_nct = Arc::make_mut(&mut orchard);
+        let distance_to_complete_subtree =
+            (1 << TRACKED_SUBTREE_HEIGHT) - orchard_nct.count() % (1 << TRACKED_SUBTREE_HEIGHT);
+        let should_find_subtree_roots =
+            orchard_note_commitments.len() as u64 >= distance_to_complete_subtree;
+
+        // MAX_BLOCK_BYTES is less than 2^21 bytes, notes are 2^5 bytes, and a full subtree has
+        // 2^16 notes. Blocks should not contain multiple level 16 roots.
```
teor2345:
```suggestion
        // It is impossible for blocks to contain more than one level 16 orchard root:
        // > [NU5 onward] nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16.
        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
```

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+                sapling_note_commitments.into_iter().enumerate()
+            {
+                sapling_nct.append(sapling_note_commitment)?;
+                if note_idx == complete_subtree_note_in_block_idx {
+                    subtree_root = Some((
+                        sapling_nct
+                            .subtree_root::<TRACKED_SUBTREE_HEIGHT>()
+                            .expect("should be frontier after appending a note"),
+                        starts_new_subtree,
+                        sapling_nct.subtree_index::<TRACKED_SUBTREE_HEIGHT>() as u16,
```
teor2345: This `as` conversion could be incorrect if the underlying function returns a value greater than `u16::MAX`.

I'd suggest doing this conversion inside `subtree_index()`, and using `try_into().expect("always within u16")` to avoid accidental truncation.

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+                orchard_note_commitments.into_iter().enumerate()
+            {
+                orchard_nct.append(orchard_note_commitment)?;
+                if note_idx == complete_subtree_note_in_block_idx {
+                    subtree_root = Some((
+                        orchard_nct
+                            .subtree_root::<TRACKED_SUBTREE_HEIGHT>()
+                            .expect("should be frontier after appending a note"),
+                        starts_new_subtree,
+                        orchard_nct.subtree_index::<TRACKED_SUBTREE_HEIGHT>() as u16,
```
teor2345: Same for orchard, do the conversion in the method.

zebra-chain/src/orchard/tree.rs
arya2: There should only be some `start` height when reading from disk, and it should only have `begins_new_subtree` when updating note commitment trees. When a block finishes a subtree without starting the next one, the start height for the next subtree is the height of the next note.

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+            Arc<sapling::tree::NoteCommitmentTree>,
+            Option<(sapling::tree::Node, bool, u16)>,
+        ),
+        NoteCommitmentTreeError,
+    > {
         let sapling_nct = Arc::make_mut(&mut sapling);
+        let distance_to_complete_subtree =
+            (1 << TRACKED_SUBTREE_HEIGHT) - sapling_nct.count() % (1 << TRACKED_SUBTREE_HEIGHT);
+        let should_find_subtree_roots =
+            sapling_note_commitments.len() as u64 >= distance_to_complete_subtree;
```
teor2345: Optional:

This code would be clearer if:
- `(1 << TRACKED_SUBTREE_HEIGHT)` was a named constant `TRACKED_SUBTREE_SIZE`
- `distance_to_complete_subtree` was a method

This code would be more reliable if:
- we didn't use `as`, because it can truncate large values. Instead, use `try_into().expect()`, ideally via `as_uNN` methods on the underlying type or type wrapper

These are optional because `usize` to `u64` conversions don't wrap as long as they are counting things in memory. (Or as long as we're on a 64 bit architecture.)

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +191,106 @@ impl NoteCommitmentTrees {
...
+        // MAX_BLOCK_BYTES is less than 2^21 bytes, notes are 2^5 bytes, and a full subtree has
+        // 2^16 notes. Blocks should not contain multiple level 16 roots.
+        let mut subtree_root = None;
 
-        for sapling_note_commitment in sapling_note_commitments {
-            sapling_nct.append(sapling_note_commitment)?;
+        if should_find_subtree_roots {
+            let starts_new_subtree =
+                sapling_note_commitments.len() as u64 != distance_to_complete_subtree;
+            let complete_subtree_note_in_block_idx = distance_to_complete_subtree as usize - 1;
```
teor2345: Will this subtraction ever be performed when `distance_to_complete_subtree` is zero?

This underflow could panic in dev builds, or produce incorrect results in release builds.

zebra-chain/src/orchard/tree.rs
teor2345: > There should only be some start height when reading from disk, and it should only have begins_new_subtree when updating note commitment trees.

After reading the docs and your explanation, I'm not quite sure what these fields do or how they are used. As far as I can see, each note commitment subtree should always have a start height. So if there's no start height, the entire subtree could be `None`?

Nit: These details seem like something useful to document.

Optional: It seems like you're trying to implement multiple different pieces of functionality using the same struct. Would an enum (or a struct containing an enum, or an enum containing structs) be a better way to represent those states explicitly?

For example, you might end up with something like:
```rust
enum NoteCommitmentSubtreeState {
    Pending { index },
    Started { index, start },
    Completed { index, start, end, node },
    CompletedOnDisk { index, start, end, begins_next_subtree, node }
}
```

zebra-chain/src/parallel/tree.rs
teor2345: (And the same for Orchard.)

upbqdn:
This PR adds subtrees to https://github.com/zcashfoundation/zebra/blob/654b40fc16e2088e85b9d4f3dcbd453c123d0154/zebra-state/src/service/non_finalized_state/chain.rs#L42, so we should edit https://github.com/zcashfoundation/zebra/blob/654b40fc16e2088e85b9d4f3dcbd453c123d0154/zebra-state/src/service/non_finalized_state/chain.rs#L262. A bunch of tests will probably need edits as well after that since they'll start failing.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +428,102 @@ impl DiskWriteBatch {
             self.zs_insert(&orchard_tree_cf, height, trees.orchard);
         }
 
+        if let Some(subtree) = trees.sapling_subtree {
+            let (subtree_end, subtree_start) = subtree.bounds();
+
+            let start_height = if subtree_end.index == 0 {
+                // the first entry after Height(0) should be the height of the first Sapling note
+                let (start_height, _) = zebra_db
+                    .sapling_tree_by_height_range(Height(1)..)
```
upbqdn: Let's use the height of the genesis block to make the code consistent with the comment above:
```suggestion
                    .sapling_tree_by_height_range(Height(0)..)
```

arya2:
> I'm not sure if you missed the testing checklist in ticket #6953, or if you're waiting until after the state upgrade is implemented.

~I'm waiting for the state upgrade, this PR can't merge without it.~

~It *should* work when creating a new state but would definitely panic when updating an older state - it needs the next PR to pause writing new blocks until it adds the subtrees up to the initial tip height.~

Update: This PR should be fine to merge on its own now that it's not tracking start heights.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: > Currently the serialization will panic if the wrong type is read:

Oh. I'll fix that in the `FromDisk` impl / by using an enum so it's a different variant of the same type.

> How does zcashd do it?

I'm not sure zcashd is tracking the start heights on main or any open PRs.

zebra-chain/src/orchard/tree.rs
arya2: > These details seem like something useful to document

I was hoping to replace it altogether, it was awkward. I like using an enum much better.

Started in https://github.com/ZcashFoundation/zebra/pull/7334/commits/4a52e5c5e2767cff9cb143bef2b09ac4af6f355b, I still need to update the FromDisk/IntoDisk impls but this should help with avoiding deserialization panics too.

zebra-chain/src/orchard/tree.rs
arya2: > Optional: Do next_tree_start and begins_next_subtree need to be optional?

I think they did because I wasn't intending to write those to disk (the design changed a lot during implementation).

> If next_tree_start is always either current_tree_end or current_tree_end + 1, then it doesn't have to be optional.

It's not always `current_tree_end + 1`, it should now use `Pending` when a block completes a subtree without starting the next one, and then it'll look for the first note after the end height of the previous subtree to get the start height.

> `begins_next_subtree`

Removed in favour of a `Started` variant. It wasn't supposed to be written to disk, only used when syncing or upgrading.

zebra-chain/src/sapling/tree.rs
arya2: Could you open a suggestion PR for this?

zebra-state/src/service/finalized_state/disk_format/shielded.rs
arya2: I tried to do this in https://github.com/ZcashFoundation/zebra/pull/7334/commits/62324d6732283fae520a8b4961af5cc809a1578c, but wasn't sure how else to use `Serialize`/`Deserialize`.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: It's not an issue now without start boundaries but it was easy to tell which variant of the enum it is based on the length of the byte slice too.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: > How can we make sure the upgrade runs before this code runs?

We don't have to anymore since we're not tracking start boundaries, if we want to add those later, it could find the start height from the note commitment trees when there's no subtree at the previous index.

zebra-state/src/service/non_finalized_state/chain.rs
arya2: Thank you.

> Are we guaranteed that there will only ever be one subtree in the non-finalized state?

No, I think there could be up to 5. It was originally only for finalizing blocks (which should have been documented).

zebra-chain/src/orchard/tree.rs
teor2345: I would strongly encourage you to delete the start height. It is out of scope for the fast spendability project, and it seems to be adding a lot of complexity to this PR, including multiple states for the subtree data structure.

zebra-chain/src/orchard/tree.rs
teor2345: Is the enum necessary if we are only storing the index, end height, and node?

Is there a way to avoid this state machine (and the associated code complexity) by getting the completed subtree from the existing frontier in the note commitment tree?

How does `zcashd` do this?

zebra-state/src/service/finalized_state/disk_format/shielded.rs
teor2345: `serde` is the problem here, because the data it serializes depends on the inner structure of the types. So please don't use `serde` at all.

(It's ok to use `serde` for RPC types where we control the data structures entirely, and we have snapshot tests for the RPC output format.)

zebra-chain/src/orchard/tree.rs
arya2: > Is the enum necessary if we are only storing the index, end height, and node?

Nope, it's been removed in https://github.com/ZcashFoundation/zebra/commit/62324d6732283fae520a8b4961af5cc809a1578c.


> Is there a way to avoid this state machine (and the associated code complexity) by getting the completed subtree from the existing frontier in the note commitment tree?

Yep, it can search the note commitment trees.

> How does `zcashd` do this?

No enum or start boundaries yet in zcashd.

teor2345:
> it needs the next PR to pause writing new blocks until it adds the subtrees up to the initial tip height.

Hmm, pausing block writes seems complicated, and it could potentially cause hangs. 

#### Finalized State

Can we just keep the upgrade running instead? The upgrade task could run until it gets to the latest finalized tip.

Actually, I don't think that's quite enough, the update task needs to run until the non-finalized state has rebuilt its subtrees. (If we finish the upgrade task at the finalized tip, the next block isn't guaranteed to have subtrees, because the non-finalized state might still be fetching the finalized tip subtree we just wrote.)

So I think the upgrade task can stop when the non-finalized state signals that it is fully rebuilt through a channel. (We could wait to receive a subtree, but that could take forever if a shielded pool isn't being used.)

That seems more robust than making block writes wait, and possibly simpler.

#### Non-Finalized State

The non-finalized state could continually check for an up-to-date finalized tip subtree to build off. When the finalized state has finished rebuilding trees to the tip, the non-finalized state will find that subtree, and start rebuilding off it.

The same code should work for full syncs (or syncs from any checkpoint), upgrades, and non-upgrade restarts, which seems simpler.

#### Rejected Alternatives

Can we read the completed subtree from the current note commitment tree frontier? Or reconstruct it from the previous note commitment tree frontier? That would remove the data dependency, but it also seems very complicated.

<details>

If I understand correctly, when `frontier.position().is_complete_subtree(Level::from(16))` is true, then:
- if `frontier.position().is_complete_subtree(Level::from(17))`, then the subtree root we want has already been hashed into the next largest subtree, and we need to wait for the subtrees to be rebuilt
- otherwise, `frontier.ommers()[0]` is the subtree root we want

</details>

zebra-chain/src/sapling/tree.rs
teor2345: I'm writing tests for a state bug in `main` at the moment, so it's not a priority for me. (It's an optional change that's not necessary to get this PR working.)

I don't want to block this PR on this change. So I think copying and pasting code is fine for now, as long as you can remember to do it each time 😅

zebra-chain/src/orchard/tree.rs
teor2345: Ah, I see this has already happened 🎉

zebra-chain/src/parallel/tree.rs
arya2: > Will this subtraction ever be performed when `distance_to_complete_subtree` is zero?

No, `distance_to_complete_subtree` or `2^16 - (position % 2^16)` will always be between `1..=2^16`.

arya2:
> Can we just keep the upgrade running instead? The upgrade task could run until it gets to the latest finalized tip.

That's a great idea, we can and should do that when adding start boundaries.

It could also skip adding the first start height, inserting only the index, end height, and root of the first subtree completed above the initial tip height, and let the upgrade task stop after updating that.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: It's skipping start boundaries for now, but I think the code was right and the comment was wrong in this case.

zebra-chain/src/parallel/tree.rs
teor2345: That correctness constraint isn't obvious in the code, would you mind documenting it or refactoring it?
(I haven't reviewed the latest code yet, so it might already be clearer.)

arya2: Refactored in https://github.com/ZcashFoundation/zebra/pull/7334/commits/0ae8a47fe8640eed8b132b656973120f259ad13d.

book/src/dev/state-db-upgrades.md
```diff
@@ -92,10 +92,12 @@ We use the following rocksdb column families:
 | `sapling_nullifiers`               | `sapling::Nullifier`   | `()`                          | Create  |
 | `sapling_anchors`                  | `sapling::tree::Root`  | `()`                          | Create  |
 | `sapling_note_commitment_tree`     | `block::Height`        | `sapling::NoteCommitmentTree` | Create  |
+| `sapling_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
```
teor2345:
```suggestion
| `sapling_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtreeData`   | Create  |
```

book/src/dev/state-db-upgrades.md
```diff
@@ -92,10 +92,12 @@ We use the following rocksdb column families:
...
 | *Orchard*                          |                        |                               |         |
 | `orchard_nullifiers`               | `orchard::Nullifier`   | `()`                          | Create  |
 | `orchard_anchors`                  | `orchard::tree::Root`  | `()`                          | Create  |
 | `orchard_note_commitment_tree`     | `block::Height`        | `orchard::NoteCommitmentTree` | Create  |
+| `orchard_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtree`       | Create  |
```
teor2345:
```suggestion
| `orchard_note_commitment_subtree`  | `block::Height`        | `NoteCommitmentSubtreeData`   | Create  |
```

book/src/dev/state-db-upgrades.md
```diff
@@ -118,6 +120,8 @@ Block and Transaction Data:
   used instead of a `BTreeSet<OutputLocation>` value, to improve database performance
 - `AddressTransaction`: `AddressLocation \|\| TransactionLocation`
   used instead of a `BTreeSet<TransactionLocation>` value, to improve database performance
+- `NoteCommitmentSubtreeIndex`: 16 bits, big-endian, unsigned
+- `NoteCommitmentSubtree`: `Height \|\| {sapling, orchard}::tree::Node`
```
teor2345:
```suggestion
- `NoteCommitmentSubtreeData<{sapling, orchard}::tree::Node>`: `Height \|\| {sapling, orchard}::tree::Node`
```

zebra-state/src/service/finalized_state/disk_format/shielded.rs
```diff
@@ -140,3 +152,49 @@ impl FromDisk for orchard::tree::NoteCommitmentTree {
...
+        Self::from_repr_unchecked(bytes.as_ref())
+    }
+}
+
+impl<Node: FromDisk> FromDisk for NoteCommitmentSubtreeData<Node> {
+    fn from_bytes(disk_bytes: impl AsRef<[u8]>) -> Self {
+        let bytes = disk_bytes.as_ref();
+        Self::new(
+            Height::from_bytes(&bytes[..3]),
+            Node::from_bytes(&bytes[3..35]),
```
teor2345: Is there a way to avoid using integer constants here?
It might help to do it like this:
https://github.com/ZcashFoundation/zebra/blob/7fe040ed8ebb2421a920702c98a00dc5beb51953/zebra-state/src/service/finalized_state/disk_format/block.rs#L290

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +396,14 @@ impl DiskWriteBatch {
             self.zs_insert(&orchard_tree_cf, height, trees.orchard);
         }
 
+        if let Some(subtree) = trees.sapling_subtree {
+            self.zs_insert(&sapling_subtree_cf, subtree.index, subtree);
+        }
+
+        if let Some(subtree) = trees.orchard_subtree {
+            self.zs_insert(&orchard_subtree_cf, subtree.index, subtree);
+        }
```
teor2345: To test that this code works after the Sapling or NU5 network upgrades, run a state validity check every time Zebra starts up. Here's an example:
https://github.com/ZcashFoundation/zebra/pull/7365/files#diff-1c066eab26934db6288dddda08c0c7030c8399c4b17a70892b2e988aa6404e90R453-R454

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -178,6 +182,22 @@ impl ZebraDb {
...
+    pub fn sapling_subtree_by_index(
+        &self,
+        index: impl Into<NoteCommitmentSubtreeIndex> + Copy,
+    ) -> Option<Arc<NoteCommitmentSubtree<sapling::tree::Node>>> {
+        let sapling_subtrees = self
+            .db
+            .cf_handle("sapling_note_commitment_subtree")
+            .unwrap();
+
+        let partial_subtree: NoteCommitmentSubtreeData<sapling::tree::Node> =
```
arya2: rename to subtree_data

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: I think this code has now been removed.

zebra-chain/src/orchard/tree.rs
teor2345: This is not important now that there is no start field.

zebra-chain/src/orchard/tree.rs
teor2345: This has been fixed now.

zebra-state/src/service/finalized_state/disk_format/shielded.rs
teor2345: This has been fixed.

zebra-state/src/service/finalized_state/disk_format/tests/snapshots/empty_column_families@mainnet_2.snap
arya2: Raw bytes data format snapshot tests and structured data format tests added in https://github.com/ZcashFoundation/zebra/pull/7334/commits/24ccd4c7a62bb996b1aee506927d938c6068193d, round-trip tests added in https://github.com/ZcashFoundation/zebra/pull/7334/commits/8b531332a490a4b932e63de585cb2fff6c0438ef.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: I'd prefer to add the state validity check in #7350 and defer incrementing the database minor version until Zebra can add subtrees for existing blocks.

I reverted the database format minor version and commented out subtree insertions in https://github.com/ZcashFoundation/zebra/pull/7334/commits/0d974ac578730cb7bb39cb3203d2c5c903b8e214.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,6 +396,16 @@ impl DiskWriteBatch {
...
 
+        // TODO: Increment DATABASE_FORMAT_MINOR_VERSION and uncomment these insertions
+
+        // if let Some(subtree) = trees.sapling_subtree {
+        //     self.zs_insert(&sapling_subtree_cf, subtree.index, subtree.into_data());
+        // }
+
+        // if let Some(subtree) = trees.orchard_subtree {
+        //     self.zs_insert(&orchard_subtree_cf, subtree.index, subtree.into_data());
+        // }
```
teor2345, line 407: Thanks for this! Let's wait until PRs #7385 and #7379 merge, then we'll be ready to go.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: Ok, that also works. I'll try merging this now, but the next PR will have to wait until PRs https://github.com/ZcashFoundation/zebra/pull/7385 and https://github.com/ZcashFoundation/zebra/pull/7379 merge.

teor2345 (approved):
Looks good, let's see if we can get this merged.

PR #7350, opened 2023-08-19
OUTDATED: change(state): Add note subtree indexes to the database, and calculate subtrees for existing blocks

### LINKED ISSUE #6953: Add note subtree indexes to the finalized and non-finalized state

arya2:
## Motivation

This is needed to support "spend before sync" in lightwalletd.

Closes #6953.

Depends-On #7334.

TODO: Add state validity test.

## Solution

- Add subtrees below initial tip height

## Review

Not yet ready for review.

### Reviewer Checklist

## Follow Up Work

- Add state service requests for reading subtrees

### DIFF
--- a/zebra-state/src/constants.rs
+++ b/zebra-state/src/constants.rs
@@ -48,7 +48,7 @@ pub(crate) const DATABASE_FORMAT_VERSION: u64 = 25;
 /// - adding new column families,
 /// - changing the format of a column family in a compatible way, or
 /// - breaking changes with compatibility code in all supported Zebra versions.
-pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 1;
+pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 2;
 
 /// The database format patch version, incremented each time the on-disk database format has a
 /// significant format compatibility fix.
--- a/zebra-state/src/constants.rs
+++ b/zebra-state/src/constants.rs
@@ -48,7 +48,7 @@ pub(crate) const DATABASE_FORMAT_VERSION: u64 = 25;
 /// - adding new column families,
 /// - changing the format of a column family in a compatible way, or
 /// - breaking changes with compatibility code in all supported Zebra versions.
-pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 2;
+pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 1;
 
 /// The database format patch version, incremented each time the on-disk database format has a
 /// significant format compatibility fix.
--- a/zebra-state/src/service/finalized_state/disk_format/tests/prop.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/tests/prop.rs
@@ -6,6 +6,7 @@ use zebra_chain::{
     amount::{Amount, NonNegative},
     block::{self, Height},
     orchard, sapling, sprout,
+    subtree::NoteCommitmentSubtreeData,
     transaction::{self, Transaction},
     transparent,
     value_balance::ValueBalance,
@@ -361,6 +362,16 @@ fn roundtrip_sapling_tree_root() {
     proptest!(|(val in any::<sapling::tree::Root>())| assert_value_properties(val));
 }
 
+#[test]
+fn roundtrip_sapling_subtree_data() {
+    let _init_guard = zebra_test::init();
+
+    proptest!(|(mut val in any::<NoteCommitmentSubtreeData<sapling::tree::Node>>())| {
+        val.end = val.end.clamp(Height(0), MAX_ON_DISK_HEIGHT);
+        assert_value_properties(val)
+    });
+}
+
 // TODO: test note commitment tree round-trip, after implementing proptest::Arbitrary
 
 // Orchard
@@ -434,6 +445,16 @@ fn roundtrip_orchard_tree_root() {
     let _init_guard = zebra_test::init();
 
     proptest!(|(val in any::<orchard::tree::Root>())| assert_value_properties(val));
+}
+
+#[test]
+fn roundtrip_orchard_subtree_data() {
+    let _init_guard = zebra_test::init();
+
+    proptest!(|(mut val in any::<NoteCommitmentSubtreeData<orchard::tree::Node>>())| {
+        val.end = val.end.clamp(Height(0), MAX_ON_DISK_HEIGHT);
+        assert_value_properties(val)
+    });
 }
 
 // TODO: test note commitment tree round-trip, after implementing proptest::Arbitrary
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -27,6 +27,8 @@ use crate::{
     Config,
 };
 
+mod add_subtrees;
+
 /// The kind of database format change we're performing.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum DbFormatChange {
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
@@ -1,0 +1,168 @@
+//! Fully populate the Sapling and Orchard note commitment subtrees for existing blocks in the database.
+
+use std::sync::{
+    atomic::{self, AtomicBool},
+    Arc,
+};
+
+use zebra_chain::{
+    block::Height, orchard::tree::NoteCommitmentTree as OrchardNoteCommitmentTree,
+    sapling::tree::NoteCommitmentTree as SaplingNoteCommitmentTree, subtree::NoteCommitmentSubtree,
+};
+
+use crate::service::finalized_state::{DiskWriteBatch, ZebraDb};
+
+/// Runs disk format upgrade for adding Sapling and Orchard note commitment subtrees to database.
+pub fn _run(
+    initial_tip_height: Height,
+    upgrade_db: &ZebraDb,
+    should_cancel_format_change: Arc<AtomicBool>,
+) {
+    let mut subtree_count = 0;
+    let mut prev_tree: Option<_> = None;
+    for (height, tree) in upgrade_db.sapling_tree_by_height_range(..initial_tip_height) {
+        if should_cancel_format_change.load(atomic::Ordering::Relaxed) {
+            break;
+        }
+
+        let Some(frontier) = tree.frontier() else {
+            prev_tree = Some(tree);
+            continue
+        };
+
+        // Blocks cannot complete multiple level 16 subtrees,
+        // the subtree index can increase by a maximum of 1 every ~20 blocks.
+        let subtree_address = SaplingNoteCommitmentTree::subtree_address(frontier);
+        if subtree_address.index() <= subtree_count {
+            prev_tree = Some(tree);
+            continue;
+        }
+
+        let (index, node) = if SaplingNoteCommitmentTree::is_complete_subtree(frontier) {
+            tree.completed_subtree_index_and_root()
+                .expect("already checked is_complete_subtree()")
+        } else {
+            let mut sapling_nct = Arc::try_unwrap(
+                prev_tree
+                    .take()
+                    .expect("should have some previous sapling frontier"),
+            )
+            .unwrap_or_else(|shared_tree| (*shared_tree).clone());
+
+            let block = upgrade_db
+                .block(height.into())
+                .expect("height with note commitment tree should have block");
+
+            let sapling_note_commitments: Vec<_> = block
+                .transactions
+                .iter()
+                .flat_map(|tx| tx.sapling_note_commitments())
+                .cloned()
+                .collect();
+
+            for sapling_note_commitment in sapling_note_commitments {
+                sapling_nct
+                    .append(sapling_note_commitment)
+                    .expect("finalized notes should append successfully");
+
+                if sapling_nct
+                    .frontier()
+                    .map_or(false, SaplingNoteCommitmentTree::is_complete_subtree)
+                {
+                    break;
+                }
+            }
+
+            sapling_nct
+                .completed_subtree_index_and_root()
+                .expect("already checked is_complete_subtree()")
+        };
+
+        let subtree = NoteCommitmentSubtree::new(index, height, node);
+
+        let mut batch = DiskWriteBatch::new();
+
+        batch.insert_sapling_subtree(upgrade_db, subtree);
+
+        upgrade_db
+            .write_batch(batch)
+            .expect("writing sapling note commitment subtrees should always succeed.");
+
+        subtree_count += 1;
+        prev_tree = Some(tree);
+    }
+
+    let mut subtree_count = 0;
+    let mut prev_tree: Option<_> = None;
+    for (height, tree) in upgrade_db.orchard_tree_by_height_range(..initial_tip_height) {
+        if should_cancel_format_change.load(atomic::Ordering::Relaxed) {
+            break;
+        }
+
+        let Some(frontier) = tree.frontier() else {
+            prev_tree = Some(tree);
+            continue
+        };
+
+        // Blocks cannot complete multiple level 16 subtrees,
+        // the subtree index can increase by a maximum of 1 every ~20 blocks.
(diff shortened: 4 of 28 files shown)


zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -0,0 +1,181 @@
...
+                    )) {
+                        panic!("unexpected send error: {error:?}");
+                    }
+                }
+
+                if subtree_index < tree_subtree_index {
+                    start_height = end_height;
+                    subtree_index = tree_subtree_index;
+                }
+            }
```
arya2: This is wrong.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -74,6 +78,31 @@ pub struct DbFormatChangeThreadHandle {
...
+        version
+            .parse()
+            .expect("Hardcoded version string should be valid.")
+    }
+
+    vec![
+        (parse_expect_version("25.1.0"), Box::new(PruneTrees)),
+        (parse_expect_version("25.2.0"), Box::new(AddSubtrees)),
+    ]
+}
```
teor2345: Turning the upgrades into a list is a good refactor, but large code movements should go in a separate PR, without any changes to functionality.

That makes them much easier to review and test.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -74,6 +78,31 @@ pub struct DbFormatChangeThreadHandle {
...
+fn format_upgrades() -> Vec<(Version, Box<dyn DiskFormatUpgrade>)> {
+    fn parse_expect_version(version: &str) -> Version {
+        version
+            .parse()
+            .expect("Hardcoded version string should be valid.")
+    }
+
+    vec![
+        (parse_expect_version("25.1.0"), Box::new(PruneTrees)),
+        (parse_expect_version("25.2.0"), Box::new(AddSubtrees)),
```
teor2345: Since the `25.2.0` version change is going to be merged in a previous PR, and merging that PR will run CI, we can't re-use that version in this PR. Instead, finish any changes to the format in this PR, and then increment the version again.

Normally we could delete all of the `25.2` cached states in Google Cloud, but I don't think that will work here, because it's a manual step that we can't make Mergify do. (And if we don't do that manual step before each CI run, the upgrade code in this PR will be skipped during the tests.)

```suggestion
        (parse_expect_version("25.3.0"), Box::new(AddSubtrees)),
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -0,0 +1,181 @@
+//! Adds Sapling and Orchard note commitment subtrees to database
```
teor2345: Nit: The previous PR adds the subtrees, this PR fully populates them

```suggestion
//! Fully populate the Sapling and Orchard note commitment subtrees for existing blocks in the database.
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: There's an "implementation details" section of ticket #6953 that might be helpful here, the details are currently behind a disclosure triangle.
Screenshot 2023-08-21 at 14 24 30

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -0,0 +1,181 @@
...
+                }
+
+                if let Err(error) = sapling_tree_tx.send((height, tree)) {
+                    panic!("unexpected send error: {error:?}");
+                }
+            }
+        });
+
+        // Create an unbounded channel for sending sapling note commitment subtrees to be written
+        let (sapling_subtree_tx, sapling_subtree_rx) = mpsc::channel();
```
teor2345: Overall, I feel like this design is more complicated than it needs to be. Is it necessary to launch separate threads to do this upgrade?

It could be easier to implement it without threads, and then see if the performance is acceptable. (Remember, the upgrade task itself is already running in a separate thread.)

#### Security / Reliability

Using an unbounded channel can potentially read all the trees into RAM, if the write thread happens to block. How much memory is that? If it's a large amount, this is a potential hang or crash issue.

(I think it should be ok after duplicates are removed, and the duplicate removal task itself is ok because its compare task drops most of those trees immediately.)

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: > Since the 25.2.0 version change is going to be merged in a previous PR, and merging that PR will run CI, we can't re-use that version in this PR. Instead, finish any changes to the format in this PR, and then increment the version again.

The previous PR shouldn't merge into main without this first merging into it.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: Actually, I think it might be better to delay this refactor for a while, because it will cause merge conflicts with bug fixes to the `25.1` state upgrade.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: Ah, then we'll need to delete all the `25.2` states before doing the final tests and merging both PRs.

In fact, we might need to do that deletion on a regular basis, to get decent CI results from both PRs.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: > Actually, I think it might be better to delay this refactor for a while, because it will cause merge conflicts with bug fixes to the 25.1 state upgrade.

I think it could help with avoiding merge conflicts if we do the upgrade for `25.2.1` in parallel to `25.1.1`.

teor2345: > > Actually, I think it might be better to delay this refactor for a while, because it will cause merge conflicts with bug fixes to the 25.1 state upgrade.
> 
> I think it could help with avoiding merge conflicts if we do the upgrade for `25.2.1` in parallel to `25.1.1`.

I'm not sure what you mean by "in parallel"". Do you mean "concurrently running in a separate thread", "in the same PR", or "in the same release"?

But it's not just merge conflicts that are a factor here, it's technical complexity risk, reviewer attention, testing, and developer time/scope. Reducing code changes helps with all of these.

zebra-state/src/constants.rs
```diff
@@ -48,7 +48,7 @@ pub(crate) const DATABASE_FORMAT_VERSION: u64 = 25;
 /// - adding new column families,
 /// - changing the format of a column family in a compatible way, or
 /// - breaking changes with compatibility code in all supported Zebra versions.
-pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 1;
+pub(crate) const DATABASE_FORMAT_MINOR_VERSION: u64 = 2;
```
teor2345, line 51: Note that the patch version will also need to be reset to zero.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -0,0 +1,168 @@
...
+
+/// Runs disk format upgrade for adding Sapling and Orchard note commitment subtrees to database.
+pub fn _run(
+    initial_tip_height: Height,
+    upgrade_db: &ZebraDb,
+    should_cancel_format_change: Arc<AtomicBool>,
+) {
+    let mut subtree_count = 0;
+    let mut prev_tree: Option<_> = None;
+    for (height, tree) in upgrade_db.sapling_tree_by_height_range(..=initial_tip_height) {
```
teor2345, line 23: Note: I split the lightwalletd back-tracking requirement into ticket #7407.

(Please don't fix it in this PR, it needs its own code and tests, and this PR is big enough already.)

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -355,9 +396,30 @@ impl DiskWriteBatch {
             self.zs_insert(&orchard_tree_cf, height, trees.orchard);
         }
 
+        if let Some(subtree) = trees.sapling_subtree {
+            self.zs_insert(&sapling_subtree_cf, subtree.index, subtree.into_data());
```
teor2345, line 400: Note: Since this PR writes to the database, it can't merge until after the 1.2.0 release.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -0,0 +1,168 @@
...
+        if should_cancel_format_change.load(atomic::Ordering::Relaxed) {
+            break;
+        }
```
teor2345, line 26: Note: the cancel handle got reverted to the channel implementation, because atomics don't correctly handle dropped states.

zebra-state/src/service/finalized_state/disk_format/tests/prop.rs
```diff
@@ -361,6 +362,16 @@ fn roundtrip_sapling_tree_root() {
     proptest!(|(val in any::<sapling::tree::Root>())| assert_value_properties(val));
 }
 
+#[test]
+fn roundtrip_sapling_subtree_data() {
+    let _init_guard = zebra_test::init();
+
+    proptest!(|(mut val in any::<NoteCommitmentSubtreeData<sapling::tree::Node>>())| {
+        val.end = val.end.clamp(Height(0), MAX_ON_DISK_HEIGHT);
```
teor2345, line 370: This might not do what you expect, because almost all the tests will be checking an `end` of `MAX_ON_DISK_HEIGHT`.

This issue is documented here, but for a string regex, rather than a clamp:
> In general, the ideal solution is to find a way to take a seed value and then use prop_map to transform it into the desired, irregular domain.
> This kind of works, but it has problems ... [it] also results in a pretty odd distribution of values. 

https://proptest-rs.github.io/proptest/proptest/tutorial/filtering.html
https://proptest-rs.github.io/proptest/proptest/tutorial/transforming-strategies.html

Instead, transform the input range so its distribution is roughly uniform:
```suggestion
        val.end.0 %= (MAX_ON_DISK_HEIGHT.0 + 1);
```

zebra-state/src/service/finalized_state/disk_format/tests/prop.rs
```diff
@@ -436,6 +447,16 @@ fn roundtrip_orchard_tree_root() {
     proptest!(|(val in any::<orchard::tree::Root>())| assert_value_properties(val));
 }
 
+#[test]
+fn roundtrip_orchard_subtree_data() {
+    let _init_guard = zebra_test::init();
+
+    proptest!(|(mut val in any::<NoteCommitmentSubtreeData<orchard::tree::Node>>())| {
+        val.end = val.end.clamp(Height(0), MAX_ON_DISK_HEIGHT);
```
teor2345, line 455:
```suggestion
        val.end.0 %= (MAX_ON_DISK_HEIGHT.0 + 1);
```

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -224,7 +260,9 @@ impl ZebraDb {
         NoteCommitmentTrees {
             sprout: self.sprout_tree(),
             sapling: self.sapling_tree(),
+            sapling_subtree: None,
             orchard: self.orchard_tree(),
+            orchard_subtree: None,
```
teor2345, line 265: Looks like you might need `*_subtree_for_tip()` methods here.

You can get the tip trees with `zs_last_key_value()`, feel free to grab the `impl FromDisk for NoteCommitmentSubtreeIndex` from PR #7408.

teor2345:
I had to change more code than I expected in the RPC PR, so if you could minimise the changes in this PR, that would be helpful!

Then we can merge them both, and base any fixes on the `main` branch.

zebra-state/src/constants.rs
teor2345: I fixed this in PR #7437.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: I changed the code below this code in PR #7437, so fixing it here would cause merge conflicts.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: I fixed this in PR #7437.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: Let's continue work on this in the ticket.

teor2345:
@arya2 I needed the data from this PR to test the RPC method, so I copied that code across into PR #7437, and did the necessary fixes.

Let's keep working on the database changes in that PR?

I suggested some minor fixes to a test and a production method in this PR. Can you check if anything needs to be done?
Feel free to open new PRs or tickets for those fixes.

zebra-state/src/service/finalized_state/disk_format/tests/prop.rs
teor2345: This is now #7443

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: This is now ticket #7444

PR #7367, opened 2023-08-22
Update issue templates to add a custom blank template

### LINKED ISSUE #7318: Readd blank template to issues
https://github.com/ZcashFoundation/zebra/issues/new/choose

I generally use and prefer an empty template when adding new short issues instead of the ones prepopulated that are more used by external users or special cases. Currently i have to select one of the templates and remove stuff from the url to get to it. 

This is a minor thing but i think we should have the empty issue creation option back to the list.

mpguerra:
Adding in a blank/custom issue template


teor2345 (approved):
Thanks!

PR #7371, opened 2023-08-23
0. Add note commitment subtree types to zebra-chain

teor2345:
## Motivation

We want to review and merge the data structure and cryptographic changes to the subtree code separately.

This is part of ticket #6953.

### PR Construction Process

I made this PR by checking out the zebra-chain changes from @arya2's `subtree-boundaries` branch:
```sh
git checkout -b subtree-boundaries-zebra-chain main
git checkout origin/subtree-boundaries zebra-chain
git commit
```

And then temporarily fixing the compiler errors by using `None` for all the new subtree fields. (Since those fields aren't read by the state yet, any value is ok.)

There's cryptographic code here that I don't fully understand. But we can fix it later during testing if any bugs come up.

## Solution

Add subtree data structures
Add methods for working with subtrees
Add methods for deserialising nodes

## Review

This PR is the first PR for ticket #6953, it blocks all the others.

Teor can't approve this PR, because they opened it based on Arya's branch.

### Reviewer Checklist

## Follow Up Work

Revert the temporary `None` values for subtree fields.

### DIFF
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -27,8 +27,11 @@ use zcash_primitives::merkle_tree::{write_commitment_tree, HashSer};
 
 use super::sinsemilla::*;
 
-use crate::serialization::{
-    serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
+use crate::{
+    serialization::{
+        serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
+    },
+    subtree::{TRACKED_SUBTREE_HEIGHT, TRACKED_SUBTREE_SIZE},
 };
 
 pub mod legacy;
@@ -170,6 +173,13 @@ impl ZcashDeserialize for Root {
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct Node(pallas::Base);
 
+impl Node {
+    /// Calls `to_repr()` on inner value.
+    pub fn to_repr(&self) -> [u8; 32] {
+        self.0.to_repr()
+    }
+}
+
 /// Required to convert [`NoteCommitmentTree`] into [`SerializedTree`].
 ///
 /// Zebra stores Orchard note commitment trees as [`Frontier`][1]s while the
@@ -221,6 +231,18 @@ impl Hashable for Node {
 impl From<pallas::Base> for Node {
     fn from(x: pallas::Base) -> Self {
         Node(x)
+    }
+}
+
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(
+            bytes.try_into().map_err(|_| "wrong byte slice len")?,
+        ))
+        .map(Node)
+        .ok_or("invalid Pallas field element")
     }
 }
 
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
         }
     }
 
+    /// Returns true if the most recently appended leaf completes the subtree
+    pub fn is_complete_subtree(&self) -> bool {
+        self.inner.value().map_or(false, |x| {
+            x.position()
+                .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
+        })
+    }
+
+    /// Returns subtree index at [`TRACKED_SUBTREE_HEIGHT`]
+    pub fn subtree_index(&self) -> u16 {
+        (self.position() >> TRACKED_SUBTREE_HEIGHT)
+            .try_into()
+            .expect("always within u16")
+    }
+
+    /// Returns subtree root, if any
+    pub fn subtree_root(&self) -> Option<Node> {
+        self.is_complete_subtree().then_some(())?;
+        Some(
+            self.inner
+                .value()?
+                .root(Some(TRACKED_SUBTREE_HEIGHT.into())),
+        )
+    }
+
+    /// Returns the index of note completing the next subtree
+    pub fn next_complete_subtree_note_index(&self) -> usize {
+        // # Correctness
+        //
+        // 2^16 - (position % 2^16) should always be between 1..=2^16
+        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+            .try_into()
+            .expect("should fit in usize")
+    }
+
+    /// Counts of note commitments added to the current subtree, max 2^16
+    fn subtree_note_count(&self) -> u64 {
+        self.count() % TRACKED_SUBTREE_SIZE
+    }
+
     /// Returns the current root of the tree, used as an anchor in Orchard
     /// shielded transactions.
     pub fn root(&self) -> Root {
@@ -370,6 +432,11 @@ impl NoteCommitmentTree {
     /// Uncommitted^Orchard = I2LEBSP_l_MerkleOrchard(2)
     pub fn uncommitted() -> pallas::Base {
         pallas::Base::one().double()
+    }
+
+    /// Position of final leaf added to the tree.
+    fn position(&self) -> u64 {
+        self.inner.value().map_or(0, |x| u64::from(x.position()))
     }
 
     /// Count of note commitments added to the tree.
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -178,18 +178,17 @@ impl Node {
     pub fn to_repr(&self) -> [u8; 32] {
         self.0.to_repr()
     }
+}
 
-    /// Converts a byte representation of a field element into an element of
-    /// this prime field without checking if the input is not canonical
-    pub fn from_repr_unchecked(bytes: &[u8]) -> Self {
-        let mut tmp = [0, 0, 0, 0];
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
 
-        tmp[0] = u64::from_le_bytes(bytes[0..8].try_into().unwrap());
-        tmp[1] = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
-        tmp[2] = u64::from_le_bytes(bytes[16..24].try_into().unwrap());
-        tmp[3] = u64::from_le_bytes(bytes[24..32].try_into().unwrap());
-
-        pallas::Base::from_raw(tmp).into()
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(
+            bytes.try_into().map_err(|_| "wrong byte slice len")?,
+        ))
+        .map(Node)
+        .ok_or("invalid Pallas field element")
     }
 }
 
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -18,7 +18,7 @@ use std::{
 };
 
 use bitvec::prelude::*;
-use bridgetree;
+use bridgetree::{self, NonEmptyFrontier};
 use halo2::pasta::{group::ff::PrimeField, pallas};
 use incrementalmerkletree::Hashable;
 use lazy_static::lazy_static;
@@ -31,7 +31,7 @@ use crate::{
     serialization::{
         serde_helpers, ReadZcashExt, SerializationError, ZcashDeserialize, ZcashSerialize,
     },
-    subtree::{TRACKED_SUBTREE_HEIGHT, TRACKED_SUBTREE_SIZE},
+    subtree::TRACKED_SUBTREE_HEIGHT,
 };
 
 pub mod legacy;
@@ -178,6 +178,19 @@ impl Node {
     pub fn to_repr(&self) -> [u8; 32] {
         self.0.to_repr()
     }
+
+    /// Converts a byte representation of a field element into an element of
+    /// this prime field without checking if the input is not canonical
+    pub fn from_repr_unchecked(bytes: &[u8]) -> Self {
+        let mut tmp = [0, 0, 0, 0];
+
+        tmp[0] = u64::from_le_bytes(bytes[0..8].try_into().unwrap());
+        tmp[1] = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+        tmp[2] = u64::from_le_bytes(bytes[16..24].try_into().unwrap());
+        tmp[3] = u64::from_le_bytes(bytes[24..32].try_into().unwrap());
+
+        pallas::Base::from_raw(tmp).into()
+    }
 }
 
 /// Required to convert [`NoteCommitmentTree`] into [`SerializedTree`].
@@ -231,18 +244,6 @@ impl Hashable for Node {
 impl From<pallas::Base> for Node {
     fn from(x: pallas::Base) -> Self {
         Node(x)
-    }
-}
-
-impl TryFrom<&[u8]> for Node {
-    type Error = &'static str;
-
-    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        Option::<pallas::Base>::from(pallas::Base::from_repr(
-            bytes.try_into().map_err(|_| "wrong byte slice len")?,
-        ))
-        .map(Node)
-        .ok_or("invalid Pallas field element")
     }
 }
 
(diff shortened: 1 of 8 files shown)


zebra-chain/src/orchard/tree.rs
```diff
@@ -224,6 +234,18 @@ impl From<pallas::Base> for Node {
     }
 }
 
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<pallas::Base>::from(pallas::Base::from_repr(
```
teor2345:
```suggestion
        // Performance note: If these methods appear in CPU profiles,
        // consider deserializing trusted state data without expensive cryptographic checks.
        Option::<pallas::Base>::from(pallas::Base::from_repr(
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -218,6 +227,18 @@ impl From<jubjub::Fq> for Node {
     }
 }
 
+impl TryFrom<&[u8]> for Node {
+    type Error = &'static str;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        Option::<jubjub::Fq>::from(jubjub::Fq::from_bytes(
```
teor2345, line 233:
```suggestion
        // Performance note: If these methods appear in CPU profiles,
        // consider deserializing trusted state data without expensive cryptographic checks.
        Option::<jubjub::Fq>::from(jubjub::Fq::from_bytes(
```

zebra-chain/src/sapling/tree.rs
```diff
@@ -218,6 +227,18 @@ impl From<jubjub::Fq> for Node {
...
+            bytes.try_into().map_err(|_| "wrong byte slice len")?,
+        ))
+        .map(Node::from)
+        .ok_or("invalid Pallas field element")
```
teor2345: Nit:
```suggestion
        .ok_or("invalid jubjub field element")
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
 
+    /// Returns true if the most recently appended leaf completes the subtree
+    pub fn is_complete_subtree(&self) -> bool {
+        self.inner.value().map_or(false, |x| {
+            x.position()
+                .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
+        })
+    }
+
+    /// Returns subtree index at [`TRACKED_SUBTREE_HEIGHT`]
```
teor2345: Is this the subtree index of the current leaf node, or the most recently completed subtree?

```suggestion
    /// Returns the subtree index for the most recently appended leaf
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
+    pub fn subtree_index(&self) -> u16 {
+        (self.position() >> TRACKED_SUBTREE_HEIGHT)
+            .try_into()
+            .expect("always within u16")
+    }
+
+    /// Returns subtree root, if any
```
teor2345:
```suggestion
    /// If the most recent leaf completed a subtree, returns the root of that subtree
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
+    pub fn subtree_root(&self) -> Option<Node> {
```
teor2345: Optional (?): Zebra uses `Node` to represent a leaf node, and has a separate tree `Root` type. So it would be better to use a `SubtreeRoot` type here, to avoid type confusion.

I am not sure how to prioritise this change, it depends if we find any type confusion bugs during the rest of the implementation and testing. So I'll mark it as optional for now.

<details>

<summary>irrelevant performance note</summary>

Optional: One of the benefits of using a separate type is that it can have a simple `[u8; 32]` inner representation, because we don't do any more cryptographic operations on it. This can improve performance. (Performance optimisations are optional unless there is evidence that it is very slow.)

</details>

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
+        // 2^16 - (position % 2^16) should always be between 1..=2^16
+        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+            .try_into()
+            .expect("should fit in usize")
+    }
+
+    /// Counts of note commitments added to the current subtree, max 2^16
+    fn subtree_note_count(&self) -> u64 {
+        self.count() % TRACKED_SUBTREE_SIZE
+    }
```
teor2345: This will also become unused code which we can delete:

```suggestion
```

#### Obsolete/Incorrect Suggestion

<details>

This is:
```rust
self.position() - Address::above_position(TRACKED_SUBTREE_HEIGHT, self.position()).position_range_start()
```

</details>

zebra-chain/src/parallel/tree.rs
```diff
@@ -49,32 +56,18 @@ impl NoteCommitmentTrees {
...
-            )]
-            .into_iter()
-            .collect(),
-        )
-    }
+        let block = block.clone();
+        let height = block
+            .coinbase_height()
+            .expect("height was already validated");
+        let block_list: BTreeMap<Height, Arc<Block>> = [(height, block)].into_iter().collect();
```
teor2345: Optional: this isn't a block list any more, so we don't need to fake a one-item list.

With one block, this code:
```rust
        block_list
            .values()
            .flat_map(|block| block.transactions.iter())
```

Is the same as this code:
```rust
        block.transactions.iter()
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
+
+    /// Returns the index of note completing the next subtree
+    pub fn next_complete_subtree_note_index(&self) -> usize {
+        // # Correctness
+        //
+        // 2^16 - (position % 2^16) should always be between 1..=2^16
+        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+            .try_into()
+            .expect("should fit in usize")
+    }
```
teor2345: I don't think this method is needed, because we can use other methods to achieve the same result.

```suggestion
```

#### Obsolete/Incorrect Suggestion

<details>

I think this method's documentation is incorrect, because the returned value isn't a position or index.

What this method actually seems to do is return the number of remaining notes before the current subtree is complete. That can also be achieved with library methods.


Here's the documented behaviour:

```rust
        incrementalmerkletree::Address::above_position(
                TRACKED_SUBTREE_HEIGHT,
                self.position()
            )
            .max_position()
            .try_into()
            .expect("should fit in usize")
```

And here are the library APIs that match the current implementation's behaviour:

Untested code:


```suggestion
        let max_position = incrementalmerkletree::Address::above_position(
                TRACKED_SUBTREE_HEIGHT,
                self.position()
            )
            .max_position();
        (max_position - self.position())
            .try_into()
            .expect("should fit in usize")
```

</details>

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +161,96 @@ impl NoteCommitmentTrees {
...
-    ) -> Result<Arc<sapling::tree::NoteCommitmentTree>, NoteCommitmentTreeError> {
+    ) -> Result<
+        (
+            Arc<sapling::tree::NoteCommitmentTree>,
+            Option<(u16, sapling::tree::Node)>,
+        ),
+        NoteCommitmentTreeError,
+    > {
         let sapling_nct = Arc::make_mut(&mut sapling);
+        let next_complete_subtree_note_index = sapling_nct.next_complete_subtree_note_index();
```
teor2345: Consequential dead code:
```suggestion
```

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +161,96 @@ impl NoteCommitmentTrees {
...
+                    let node = sapling_nct
+                        .subtree_root()
+                        .expect("should be frontier after appending a note");
+                    subtree_root = Some((sapling_nct.subtree_index(), node));
+                }
+            }
+        } else {
+            for sapling_note_commitment in sapling_note_commitments {
+                sapling_nct.append(sapling_note_commitment)?;
+            }
```
teor2345: I struggled to work out if this code was correct, and what exactly it does. I found it hard to check the method names, bitwise arithmetic, iterator indexes, loop conditions, and off by one errors at the same time.

But I think it might be equivalent to checking for a complete subtree after every node?

```suggestion
        for sapling_note_commitment in sapling_note_commitments {
            sapling_nct.append(sapling_note_commitment)?;
            if sapling_nct.is_complete_subtree() {
                let subtree_root_node = sapling_nct
                    .subtree_root()
                    .expect("should be frontier after appending a note");
                subtree_root = Some((sapling_nct.subtree_index(), subtree_root_node));
            }
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -317,6 +339,46 @@ impl NoteCommitmentTree {
...
+                .is_complete_subtree(TRACKED_SUBTREE_HEIGHT.into())
+        })
+    }
+
+    /// Returns subtree index at [`TRACKED_SUBTREE_HEIGHT`]
+    pub fn subtree_index(&self) -> u16 {
+        (self.position() >> TRACKED_SUBTREE_HEIGHT)
+            .try_into()
+            .expect("always within u16")
+    }
```
teor2345: I strongly recommend using the same APIs as `zcashd`, rather than creating our own duplicate implementations.

Since we have a subtree index wrapper type, it might help to use it here as well.

Untested code:

```suggestion
    pub fn subtree_index(&self) -> NoteCommitmentSubtreeIndex
        incrementalmerkletree::Address::above_position(
                TRACKED_SUBTREE_HEIGHT,
                self.position()
            )
            .index()
            .try_into()
            .expect("always within u16")
```

zebra-chain/src/orchard/tree.rs
teor2345: Optional: One of the benefits of using a separate type is that it can have a simple `[u8; 32]` inner representation, because we don't do any more cryptographic operations on it. This can improve performance. (Performance optimisations are optional unless there is evidence that it is very slow.)

zebra-chain/src/sapling/tree.rs
```diff
@@ -312,6 +333,46 @@ impl NoteCommitmentTree {
...
+        // 2^16 - (position % 2^16) should always be between 1..=2^16
+        (TRACKED_SUBTREE_SIZE - self.subtree_note_count() - 1)
+            .try_into()
+            .expect("should fit in usize")
+    }
+
+    /// Counts of note commitments added to the current subtree, max 2^16
+    fn subtree_note_count(&self) -> u64 {
+        self.count() % TRACKED_SUBTREE_SIZE
+    }
```
teor2345: Same as for Sapling, and the tree update code needs changes as well.

zebra-chain/src/subtree.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct NoteCommitmentSubtreeIndex(pub u16);
+
+impl From<u16> for NoteCommitmentSubtreeIndex {
+    fn from(value: u16) -> Self {
+        Self(value)
+    }
+}
+
+/// Subtree of Sapling or Orchard note commitment tree
```
teor2345:
```suggestion
/// Subtree root of Sapling or Orchard note commitment tree, with its associated block height and subtree index.
```

zebra-chain/src/subtree.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Creates new [`NoteCommitmentSubtree`]
+    pub fn new(index: impl Into<NoteCommitmentSubtreeIndex>, end: Height, node: Node) -> Arc<Self> {
+        let index = index.into();
+        Arc::new(Self { index, end, node })
+    }
+}
+
+/// Partial subtree of Sapling note commitment tree
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct PartialNoteCommitmentSubtree<Node> {
```
teor2345:
```suggestion
/// Subtree root of Sapling or Orchard note commitment tree, with block height, but without the subtree index.
/// Used for database key-value serialization, where the subtree index is the key, and this struct is the value.
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub struct NoteCommitmentSubtreeData<Node> {
```

teor2345 (commented):
I think we can use the `incrementalmerkletree` crate methods here, rather than creating our own duplicate arithmetic methods. We avoid try to avoid duplicate implementations, because we have to write tests for them as well, and maintain that duplicate code if there are changes.

I also think we can simplify the code a lot. Normally I wouldn't worry about this kind of cleanup, but I found it hard to check if the iterator handling in this PR was correct. We've had similar bugs recently, and we don't have any tests in this PR, so simpler seemed better.

I'm happy to make these changes in a separate PR, I just want to double-check that I understand what the code is meant to do correctly.

zebra-chain/src/orchard/tree.rs
teor2345: This will involve a bunch of renames, so it should be scheduled after the existing code has merged. I'll open a ticket for cleanups if it turns out to be a priority.

zebra-chain/src/subtree.rs
teor2345: This rename should be scheduled after the existing code has merged. I'll open a ticket for cleanups if it turns out to be a priority.

The doc change can be applied in this PR though.

zebra-chain/src/orchard/tree.rs
arya2: This seems excessively verbose:

```suggestion
    pub fn subtree_index(&self) -> u16 {
        incrementalmerkletree::Address::above_position(
            TRACKED_SUBTREE_HEIGHT.into(),
            self.inner.value().map_or(0.into(), |x| x.position()),
        )
        .index()
        .try_into()
        .expect("always within u16")
    }
```

Update: It's less bad if part of it can be reused.

zebra-chain/src/orchard/tree.rs
arya2: > What this method actually seems to do is return the number of remaining notes before the current subtree is complete

I had a lot of trouble naming this because it's the remaining notes before the current subtree is complete except when the last note to be appended completed a subtree, in which case it's 2^16 - 1.

I'll make this change.

zebra-chain/src/parallel/tree.rs
arya2: > But I think it might be equivalent to checking for a complete subtree after every node?

Yep.

arya2:
@teor2345 any chance you'd be available to help write the tests for #7334?

I think we've addressed all other comments from both PRs, and that it would be easier to use the `read::tree` fns in zebra-state when adding the state requests / RPC method.

zebra-chain/src/subtree.rs
```diff
@@ -0,0 +1,63 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct NoteCommitmentSubtreeData<Node> {
+    /// End boundary of this subtree, the block height of its last leaf.
+    pub end: Height,
+    /// Root of this subtree.
+    pub node: Node,
+}
+
+impl<Node> NoteCommitmentSubtreeData<Node> {
+    /// Creates new [`PartialNoteCommitmentSubtree`]
```
teor2345: Here's a guide to doing mass renames without having to check for missed renames like this:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/mass-renames.md

```suggestion
    /// Creates new [`NoteCommitmentSubtreeData`]
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -170,6 +173,26 @@ impl ZcashDeserialize for Root {
...
 
+impl Node {
+    /// Calls `to_repr()` on inner value.
+    pub fn to_repr(&self) -> [u8; 32] {
+        self.0.to_repr()
+    }
+
+    /// Converts a byte representation of a field element into an element of
+    /// this prime field without checking if the input is not canonical
+    pub fn from_repr_unchecked(bytes: &[u8]) -> Self {
```
teor2345: Is this needed?

zebra-chain/src/parallel/tree.rs
```diff
@@ -160,36 +156,96 @@ impl NoteCommitmentTrees {
...
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+        //
+        // Before NU5, this limit holds due to the minimum size of Sapling outputs (948 bytes)
+        // and the maximum size of a block:
+        // > The size of a block MUST be less than or equal to 2000000 bytes.
+        // <https://zips.z.cash/protocol/protocol.pdf#blockheader>
+        // <https://zips.z.cash/protocol/protocol.pdf#txnencoding>
+        let mut subtree_root = None;
+
+        if let Some((subtree_index, sapling_note_commitments)) =
```
teor2345: Can we do the simple loop here?

teor2345:
> @teor2345 any chance you'd be available to help write the tests for #7334?
> 
> I think we've addressed all other comments from both PRs, and that it would be easier to use the `read::tree` fns in zebra-state when adding the state requests / RPC method.

Happy to help write tests, is there a specific way you wanted to split up the test work?

I already wrote a state validity check for the de-duplication upgrade, so if you write that check for the subtrees, then we'll have multiple people on the team who have experience writing them.

How did you want to split the raw/structured snapshot tests, round-trip tests, and unit tests?
Is there any tricky code you specifically want to unit test?

zebra-chain/src/orchard/tree.rs
arya2: Nope, removed in https://github.com/ZcashFoundation/zebra/pull/7371/commits/10289312a42a4d2793845c121122477a1cbbaddd.

zebra-chain/src/subtree.rs
arya2: Ah, thank you, fixed in https://github.com/ZcashFoundation/zebra/pull/7371/commits/b77e050fe11e49afb91b5a58b2975f8eb5fb42d3.

zebra-chain/src/parallel/tree.rs
arya2: Yep, simplified in https://github.com/ZcashFoundation/zebra/pull/7371/commits/b77e050fe11e49afb91b5a58b2975f8eb5fb42d3.

arya2:
> How did you want to split the raw/structured snapshot tests, round-trip tests, and unit tests?

~Could you take the round-trip tests?~

Update: I think I have a good idea of how to write the tests now, and there are existing `Arbitrary` impls for `Node`s.

> Is there any tricky code you specifically want to unit test?

No. Not anymore. 😅

teor2345 (commented):
Thanks for your patience here, I'm happy with this PR and the code looks correct.

@arya2 or @upbqdn can you approve this PR?
I can't, because I opened it.

arya2 (approved):

PR #7377, opened 2023-08-24
change(state): Move format upgrades to separate modules

arya2:
## Scheduling

This change is blocked by #7392.

## Motivation

This PR moves the tree deduplication into a separate module ahead of adding more upgrades and state validity checks.

Depends On: #7379.

## Solution

- Moves upgrade to separate file

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -18,12 +18,42 @@ use zebra_chain::{
 use DbFormatChange::*;
 
 use crate::{
-    config::write_database_format_version_to_disk,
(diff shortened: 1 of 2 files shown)


teor2345:
Just noting that this code movement PR will conflict with PR #7392, which is a low priority security fix.

Refactors are usually lower priority than fixes, so let's mark this as optional?
(Neither of them are on the critical path to getting RPCs working.)

arya2:
> Just noting that this code movement PR will conflict with PR https://github.com/ZcashFoundation/zebra/pull/7392, which is a low priority security fix.

I think the changes for adding subtrees from existing blocks will conflict with that anyway?

PR #7365, opened 2023-08-22
Log errors and panic if duplicate trees are found after the de-duplicating upgrade

teor2345:
## Motivation

There are some duplicate trees left after the version `25.1.0` state upgrade, which de-duplicates note commitment trees.

### Complex Code or Requirements

There seem to be some off-by-one errors in the code that's been merged to the `main` branch, see my upcoming review on PR #7312.

## Solution

This PR just detects the bug, by checking for duplicates after the upgrade has run. (Or if there isn't any upgrade, when the state is opened.)

## Review

This is part of a conversation with @upbqdn, but I'm happy to do any fixes.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

Actually fix the bug.

PR #7379, opened 2023-08-24
fix(state): Fix the deduplication of note commitment trees

upbqdn:
## Motivation & Solution

- Refactor the deduplication.
- Include a test that checks that there are no duplicates.

This PR merges #7365.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -330,20 +330,6 @@ impl DbFormatChange {
                 }
             }
 
-            // The Sapling note commitment trees are pruned if `last_height` has reached the height
-            // right after `initial_tip_height`. If `last_height` has not reached that height, they
-            // have been pruned in a previous upgrade, so the inclusive range between `last_height`
-            // and `initial_tip_height` must be empty. If it is not, there is an error, and the
-            // current upgrade did not succeed.
-            if last_height != initial_tip_height.next()
-                && db
-                    .sapling_tree_by_height_range(last_height..=initial_tip_height)
-                    .next()
-                    .is_some()
-            {
-                panic!("Zebra could not prune all Sapling trees.");
-            }
-
             // Prune duplicate Orchard note commitment trees.
             let mut last_tree = db.orchard_tree_by_height(&Height(0)).expect(
                 "The Orchard note commitment tree for the genesis block should be in the database.",
@@ -393,20 +379,6 @@ impl DbFormatChange {
                     // in the database.
                     last_height = height.next();
                 }
-            }
-
-            // The Sapling note commitment trees are pruned if `last_height` has reached the height
-            // right after `initial_tip_height`. If `last_height` has not reached that height, they
-            // have been pruned in a previous upgrade, so the inclusive range between `last_height`
-            // and `initial_tip_height` must be empty. If it is not, there is an error, and the
-            // current upgrade did not succeed.
-            if last_height != initial_tip_height.next()
-                && db
-                    .orchard_tree_by_height_range(last_height..=initial_tip_height)
-                    .next()
-                    .is_some()
-            {
-                panic!("Zebra could not prune all Orchard trees.");
             }
 
             // Before marking the state as upgraded, check that the upgrade completed successfully.
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -12,9 +12,7 @@ use tracing::Span;
 use zebra_chain::{
     block::Height,
     diagnostic::task::{CheckForPanics, WaitForPanics},
-    orchard,
     parameters::Network,
-    sapling,
 };
 
 use DbFormatChange::*;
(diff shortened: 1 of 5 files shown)


zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -272,180 +275,149 @@ impl DbFormatChange {
...
-                let mut delete_from = sapling_height.next();
-                while let Ok(delete_to) = unique_sapling_tree_height_rx.recv() {
-                    let num_entries = delete_to - delete_from;
-                    if num_entries > 0 {
+                // We delete duplicate trees in batches. A batch is a (possibly empty) series of
+                // identical trees excluding the first tree. We exclude the first tree so that we
+                // don't delete it. We get a batch if we encounter a tree that differs from the
+                // previous one, or if we're at the end of the chain. We get an empty batch if there
+                // are two consecutive differing trees.
+                if last_tree != tree || height == initial_tip_height {
```
teor2345: I think the delete range needs to be different for these two different cases:
1. if the trees aren't equal, we want to keep the tree at `height`
2. if the trees are equal but we've reached the initial tip height, we want to delete the tree at `height`

The code below implements the first case.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -272,180 +275,149 @@ impl DbFormatChange {
...
-                let mut delete_from = orchard_height.next();
-                while let Ok(delete_to) = unique_orchard_tree_height_rx.recv() {
-                    let num_entries = delete_to - delete_from;
-                    if num_entries > 0 {
+                // We delete duplicate trees in batches. A batch is a (possibly empty) series of
+                // identical trees excluding the first tree. We exclude the first tree so that we
+                // don't delete it. We get a batch if we encounter a tree that differs from the
+                // previous one, or if we're at the end of the chain. We get an empty batch if there
+                // are two consecutive differing trees.
+                if last_tree != tree || height == initial_tip_height {
```
teor2345: Same here

teor2345 (commented):
Arya told me there were some performance issues in the initial version of this PR. Have they been fixed? How fast is this new upgrade code on a v25.0.0 state?

(I can test this if you don't have an old state handy.)

If there are still performance issues, we might want to see if it helps to optimise NoteCommitmentTree::eq() to avoid re-calculating the root if it hasn't been cached in memory or on disk.

Because I think previous state versions didn't cache/serialize the root for empty trees? Or was that just the genesis tree?

teor2345:
The checkpoint test failure in CI:
https://github.com/ZcashFoundation/zebra/actions/runs/5965091001/job/16184178963?pr=7379
should be fixed by my PR #7385.

I'm not sure about the RPC conflict test. It might be a performance issue with the upgrade, so I'll push a fix PR to make the upgrade faster. If that doesn't work, we might need to increase the wait time in that test.

teor2345:
PR #7386 might fix the RPC conflict test failure:
https://github.com/ZcashFoundation/zebra/actions/runs/5965091001/job/16183944554?pr=7379

teor2345:
With this PR, #7385, and #7386, it takes about 35 seconds on my machine to upgrade a fully synced mainnet state and do the runtime tests.

The sync is launched concurrently, but the network hasn't even fully started by then, so I think that performance is acceptable.

On testnet it takes about 10 seconds.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: This might be the cause of the panic I get on my testnet state:

```
Message:  Zebra could not prune all Orchard trees.
Location: zebra-state/src/service/finalized_state/disk_format/upgrade.rs:409

Metadata:
version: 1.1.0+62.g961d45c
Zcash network: Testnet
running state version: 25.1.1
initial disk state version: 25.1.0
branch: state-dedup-fixes
git commit: 961d45c
```

You might want to turn this panic into an error log, because the detailed checks show which trees are still duplicates, but this panic doesn't.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: When I run using all 3 PRs on an old state that was synced to 1719530, I get a panic at the initial tip height:
```
2023-08-25T03:06:36.284392Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate sapling trees after running de-duplicate tree upgrade height=Height(1719530) prev_height=Hei
ght(1719517) tree_root=Root("d6efc94648b492e0a8ee310f4032aa618b565cc124e7cc858137f2f36fdb4e70")
Message:  found duplicate sapling or orchard trees after running de-duplicate tree upgrade
Location: zebra-state/src/service/finalized_state/disk_format/upgrade.rs:486

Metadata:
version: 1.1.0+62.g961d45c
Zcash network: Mainnet
running state version: 25.1.1
initial disk state version: 25.0.0
branch: state-dedup-fixes
git commit: 961d45c
```

teor2345:
When I run using all 3 PRs on an old mainnet state that was synced to 1719530, the deduplication takes about 7 seconds.

So I think we're fine for performance here.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: This edge case is addressed in [`4b579a6` (#7379)](https://github.com/ZcashFoundation/zebra/pull/7379/commits/4b579a60786751719110aee6fcfdde9262b02aef).

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: This edge case is addressed in [`4b579a6` (#7379)](https://github.com/ZcashFoundation/zebra/pull/7379/commits/4b579a60786751719110aee6fcfdde9262b02aef).

upbqdn:
> Arya told me there were some performance issues in the initial version of this PR.

The issue was that I was iterating over the trees using `next_back` for each tree. I never ran the upgrade for more than 12 hours, but it would take way longer than that.

> previous state versions didn't cache/serialize the root for empty trees? Or was that just the genesis tree?

That was just the genesis tree. The rest of the trees were stored with their cached roots.

teor2345:
> > Arya told me there were some performance issues in the initial version of this PR.
> 
> The issue was that I was iterating over the trees using `next_back` for each tree. I never ran the upgrade for more than 12 hours, but it would take way longer than that.
> 
> > previous state versions didn't cache/serialize the root for empty trees? Or was that just the genesis tree?
> 
> That was just the genesis tree. The rest of the trees were stored with their cached roots.

Did we update the genesis tree to have a cached root as part of this upgrade or the previous one?

I don't know how often the genesis root is accessed, but it seems like a quick fix for a significant user-triggerable performance/security issue. (It should probably go in a separate PR so we can get this one merged.)

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -284,17 +286,30 @@ impl DbFormatChange {
...
+                Arc::new(sapling::tree::NoteCommitmentTree::default()),
+            ))
+            .into_iter();
+
             // Run through all the trees in the finalized chain.
-            for (height, tree) in db.sapling_tree_by_height_range(Height(1)..=initial_tip_height) {
+            for (height, tree) in db
+                .sapling_tree_by_height_range(Height(1)..=initial_tip_height)
+                .chain(dummy_cap)
+            {
```
teor2345: This looks like a good fix, but if the state hasn't reached the activation height yet, the `initial_tip_height` will remain in the database, because the `dummy_cap` tree is equal to the `initial_tip_height` tree. So neither of these conditions is satisfied:
```rust
if last_tree != tree || height == initial_tip_height { ... }
```

I'll push a suggested refactor to another PR.

teor2345: Once we fix the versioning in PR #7385, this bug would have been caught by the checkpoint test, because the mandatory checkpoint is before NU5 activation

teor2345: PR #7391 should fix this, I tested it on my legacy state with the fix in PR #7385.

upbqdn:
> Did we update the genesis tree to have a cached root as part of this upgrade or the previous one?

We didn't.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -236,8 +241,8 @@ impl DbFormatChange {
         config: Config,
         network: Network,
         initial_tip_height: Option<Height>,
-        upgrade_db: ZebraDb,
-        should_cancel_format_change: Arc<AtomicBool>,
+        db: ZebraDb,
+        cancel_receiver: mpsc::Receiver<CancelFormatChange>,
```
arya2, line 244: ~Optional/Nitpick: I think it may be useful to use a cloneable type here.~

Update: it's good as-is.

arya2 (commented):
This looks good to me once we merge in #7391.

What's this test failure:

> 2023-08-28T02:53:02.693713Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: found duplicate sapling trees after running de-duplicate tree upgrade height=Height(25900) prev_height=Height(25899) tree_root=Root("fbc2f4300c01f0b7820d00e3347c8da4ee614674376cbc45359daa54f9b5493e")
https://github.com/ZcashFoundation/zebra/actions/runs/5994564430/job/16257493713#step:8:26063

(This is from the suggestion PR, I think it's not re-running the upgrade because the disk version was updated earlier.)

upbqdn:
> What's this test failure

Yea, it looks like the upgrade didn't run at all since there are duplicates starting at height 1.

upbqdn:
Images in GCP are still labeled with 25.1.0.

upbqdn:
> What's this test failure

That was due to the bug described in PR #7385. Since the cached state hasn't been upgraded, Zebra incorrectly determined it was a new state, and labeled it as 25.1.1, which was the version in the code. Zebra then skipped the pruning since it's for lower versions, but the test that checks for duplicates runs at each start, so it reported duplicates starting from height 1. This should be resolved by merging PR #7385.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
upbqdn: I thought we were using a single consumer for little overhead.

arya2 (approved):

upbqdn:
@mergifyio update.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: Just a note in case this ever becomes a priority (but it's not at the moment, because we don't need concurrent upgrades yet):

`std::sync::mpsc::Receiver` takes a shared reference, so it can be called on `Arc<Receiver>`:
https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.try_recv

(This is different from `tokio::sync::mpsc::Receiver`, which requires an exclusive reference.)

However, if we were using an `mpsc` channel from multiple threads, only the first thread would get the cancel message. The second thread would keep running until the sender was dropped. (Which doesn't depend on the update thread finishing, so that's ok.)

A correct cloneable implementation would be `matches!(tokio::sync::watch::Receiver::has_changed(), Ok(true) | Err(_))`, with a wrapper that clears the initial change notification:
https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.has_changed

There might also be other ways of doing it.

PR #7380, opened 2023-08-24
fix: Fix a new `cargo fmt` warning

upbqdn:
## Motivation & Solution

- Fix a new `cargo fmt` error: https://github.com/ZcashFoundation/zebra/actions/runs/5965090963/job/16181739018?pr=7379#step:6:11.

This is the new change in rustfmt: https://github.com/rust-lang/rustfmt/pull/5690

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #7381, opened 2023-08-24
fix(clippy): Fix clippy and rustfmt after last releases today

oxarbitrage:
## Motivation

Clippy and rustfmt are broken after recent releases.

Rustfmt was fixed at https://github.com/ZcashFoundation/zebra/pull/7380 but this PR has that again as i think we need them both together to merge.

## Solution

 I don't know how fix the lint (https://rust-lang.github.io/rust-clippy/master/index.html#/arc_with_non_send_sync). Most of the warnings are in `proptest` code, `BoxedStrategy` and the use of `Arc` on it. We might want to wait until they change the `Arc` to `Rc` or ask/find out if they have plans to do so.

In the meantime i disabled the lint everywhere causing problems.  This is the proposed fix of this PR.

### Reviewer Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,17 @@
 # Zebra cargo configuration
 
+# Flags that only apply to test code
+[target.'cfg(any(test, feature = "proptest-impl"))']
+rustflags = [
+    # Rust 1.72 has a false positive when nested generics are used inside Arc.
+    # This makes the `arc_with_non_send_sync` lint trigger on a lot of proptest code.
+    #
+    # TODO: remove this allow when Rust 1.73 is stable, because this lint bug is fixed in that
+    # release:
+    # <https://github.com/rust-lang/rust-clippy/issues/11076>
+    "-Aclippy::arc_with_non_send_sync",
+]
+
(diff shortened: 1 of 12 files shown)


teor2345 (changes requested):
It is ok to ignore this clippy warning in some cases, because it comes from a known bug with generics:
https://github.com/rust-lang/rust-clippy/issues/11076

But because this PR disables the warning on entire modules, I can't see if it's ok to disable the lint in all of those cases.

I'll push a commit that moves these allows closer to the actual code, so it is easier to see they are correct, and we don't end up hiding actual bugs.

teor2345:
Here's another way the same false positive can happen:
https://github.com/rust-lang/rust-clippy/issues/11090

teor2345 (approved):
This lint is just happening in proptest code. The warning is wrong, and it will be fixed in the next Rust release.

Until then, we want to ignore it. I tried adding an exception just for test code in .config/cargo.toml, but I couldn't get it to work. So I added a single exception for test code in each crate that triggers the lint.

That's slightly better than disabling the lint in some production code, which might have the real bug the lint is for.

PR #7385, opened 2023-08-25
fix(db): use the correct state version for databases without a state version file

teor2345:
## Motivation

When Zebra opens a state cache that has never been upgraded, it was reporting the cache as a newly created state, then marking it with the current state version. This incorrectly skips all format upgrades.

Instead, all upgrades should be performed on the state.

### Specifications

https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.is_dir

We missed this bug because there weren't enough tests for state version changes or state upgrades. But there are tests in PR #7379 now.

## Solution

If there's a database directory wit no version file, report its version as `25.0.0`. (Compatible but with no upgrades.)

## Review

This bug fix blocks all the finalized state format changes.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -308,6 +308,12 @@ pub fn database_format_version_in_code() -> Version {
 }
 
 /// Returns the full semantic version of the on-disk database.
+///
+/// Typically, the version is read from a version text file.
+///
+/// If there is an existing on-disk database, but no version file, returns `Ok(Some(major.0.0))`.
+/// (This happens even if the database directory was just newly created.)
+///
 /// If there is no existing on-disk database, returns `Ok(None)`.
 ///
 /// This is the format of the data on disk, the minor and patch versions
@@ -345,13 +351,15 @@ pub fn database_format_version_on_disk(
     // There's no version file on disk, so we need to guess the version
     // based on the database content
     match fs::metadata(db_path) {
-        // But there is a database on disk, so it has the current major version with no upgrades
+        // But there is a database on disk, so it has the current major version with no upgrades.
+        // If the database directory was just newly created, we also return this version.
         Ok(_metadata) => Ok(Some(Version::new(DATABASE_FORMAT_VERSION, 0, 0))),
-        Err(e) if e.kind() == ErrorKind::NotFound => {
-            // There's no version file and no database on disk, so it's a new database.
-            // It will be created with the current version.
-            Ok(None)
-        }
+
+        // There's no version file and no database on disk, so it's a new database.
+        // It will be created with the current version,
+        // but temporarily return the default version above until the version file is written.
+        Err(e) if e.kind() == ErrorKind::NotFound => Ok(None),
+
         Err(e) => Err(e)?,
     }
 }
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -318,25 +318,42 @@ pub fn database_format_version_on_disk(
 ) -> Result<Option<Version>, BoxError> {
     let version_path = config.version_file_path(network);
 
-    let version = match fs::read_to_string(version_path) {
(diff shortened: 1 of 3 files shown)


upbqdn (dismissed):

upbqdn:
@mergifyio update.

upbqdn:
This assert fails https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L481-L486.

It looks like `database_format_version_on_disk` returns `None` here https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/zebra_db.rs#L65-L66, but then `Some` here https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L477-L478, likely due to the new refactor. I don't see why this happens since `database_format_version_on_disk` should return consistent results.

teor2345:
> This assert fails [zcashfoundation/zebra@`f08cd9f`/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L481-L486](https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L481-L486).
> 
> It looks like `database_format_version_on_disk` returns `None` here [zcashfoundation/zebra@`f08cd9f`/zebra-state/src/service/finalized_state/zebra_db.rs#L65-L66](https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/zebra_db.rs#L65-L66), but then `Some` here [zcashfoundation/zebra@`f08cd9f`/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L477-L478](https://github.com/zcashfoundation/zebra/blob/f08cd9fbc6d01f1bae0d9e5b54047a053f36a4e7/zebra-state/src/service/finalized_state/disk_format/upgrade.rs#L477-L478), likely due to the new refactor. I don't see why this happens since `database_format_version_on_disk` should return consistent results.

That's my mistake. As soon as RocksDB creates the database directory, we provide 25.0.0 as the database version, even though the database is empty and newly created. We need to open RocksDB first, because we depend on its lock to protect the version file.

We could delete the "newly created" code path entirely to simplify things, if you want. Because now it's equivalent to an upgrade from 25.0.0 to the current version, but without any data in the database. (We might want to keep logging it differently, to avoid confusing users.) But that's out of scope for this PR, so let's open a ticket if there's enough interest.

zebra-state/src/config.rs
```diff
@@ -318,25 +324,44 @@ pub fn database_format_version_on_disk(
...
+    // There's no version file on disk, so we need to guess the version
+    // based on the database content
+    match fs::metadata(db_path) {
+        // But there is a database on disk, so it has the current major version with no upgrades.
+        // If the database directory was just newly created, we also return this version.
+        Ok(_metadata) => Ok(Some(Version::new(DATABASE_FORMAT_VERSION, 0, 0))),
+
+        // There's no version file and no database on disk, so it's a new database.
+        // It will be created with the current version,
+        // but temporarily return the default version above until the version file is written.
```
upbqdn, line 360: > temporarily return the default version above

I find this comment a bit confusing since we're returning `Ok(None)`, but it makes sense overall.

upbqdn (approved):
> We could delete the "newly created" code path entirely to simplify things, if you want.

I'm fine with leaving it as it is.

PR #7386, opened 2023-08-25
fix(crypt): Avoid expensive note commitment tree root recalculations in eq() methods

### LINKED ISSUE #7272: Compare note commitment trees by their frontiers if no root is present
### Motivation

We currently compare the note commitment trees only by their roots, and if a root is not present, we compute it. This computation is more expensive than comparing the underlying frontiers, and comparing the frontiers is sufficient for comparing the trees.

## Suggested solution

Keep comparing the trees by their roots, but if a root is missing, compare the trees by their frontiers instead of computing the root.

teor2345:
## Motivation

Re-calculating note commitment tree roots can take a long time if the root isn't cached in the state, or already calculated by Zebra.

Instead, it is cheaper to compare the tree frontier, which is at most 32 roots. (Less than 1 kB of byte comparisons.)

### Specifications

https://docs.rs/incrementalmerkletree/0.4.0/incrementalmerkletree/frontier/struct.NonEmptyFrontier.html#impl-PartialEq%3CNonEmptyFrontier%3CH%3E%3E-for-NonEmptyFrontier%3CH%3E

This fix helps avoid weird performance bugs in the next few state upgrades. We've already seen some performance issues in this code.

## Solution

If both roots are cached, compare them. Otherwise, compare the whole frontier.

## Review

This might be needed to get some tests for PR #7379 to pass.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/orchard/tree.rs
+++ b/zebra-chain/src/orchard/tree.rs
@@ -428,7 +428,13 @@ impl Eq for NoteCommitmentTree {}
 
 impl PartialEq for NoteCommitmentTree {
     fn eq(&self, other: &Self) -> bool {
-        self.hash() == other.hash()
+        if let (Some(root), Some(other_root)) = (self.cached_root(), other.cached_root()) {
+            // Use cached roots if available
+            root == other_root
+        } else {
+            // Avoid expensive root recalculations which use multiple cryptographic hashes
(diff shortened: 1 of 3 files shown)


upbqdn (approved):

upbqdn:
@mergifyio update.

Issue #7335, opened 2023-08-15
Fix lightwalletd sync finish detection code in tests


arya2:
### Motivation

zcash/lightwalletd's get_latest_block RPC returns the tip height of the full node.

Zebra's lightwalletd sync test expects for this RPC to return the height of the latest block in the lightwalletd cache to check that the full sync works.

### Suggested Design

- Check lightwalletd logs instead of get_latest_block for cache height
OR
- Check zebrad logs for lightwalletd requests (might need changes to the log level of the request log)

Then revert #7332.

### Other Designs

<details>

<summary>adding new RPCs significantly increases the scope of this ticket</summary>

- Add an RPC method or field to zcash/lightwalletd that returns the sync status

</details>

teor2345:
> * Check lightwalletd logs instead of get_latest_block for cache height

I think there are already some regexes in the tests that check the lightwalletd logs, so we could copy those.

One tricky part of this is that we might need to check the Zebra and lightwalletd logs in concurrently running threads.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7335?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

PR #7390, opened 2023-08-27
fix(ci): Show correct lightwalletd cache heights in CI cached state

teor2345:
## Motivation

`lightwalletd` can have different heights in its logs and cache than Zebra's logs. The tests used to make sure these heights were the same at the end of the test. But that doesn't work with the ECC fork.

This is part of the same underlying problem as ticket #7335, but it's independent of the test fix.

### Specifications

The `height_grep_text` field is used to find the height in the logs, to decide if the cached disk should be replaced. It's also used as a label for the disk.

## Solution

Remove the Zebra log regex and just use the lightwalletd log regex.

## Review

This is a routine fix.

### Reviewer Checklist


gustavovalverde (approved):
LGTM

PR #7392, opened 2023-08-28
fix(security): fix concurrency issues in tree key formats, and CPU usage in genesis tree roots

### LINKED ISSUE #7581: bug: Zebra panics with "Sprout note commitment tree must exist if there is a finalized tip" shortly after launch
### What happened?

Zebra always launches successfully.

Instead, this happened: 

Zebra panics with "Sprout note commitment tree must exist if there is a finalized tip" shortly after launch.

Testing the subtree code in PR #7566 on my local machine
Testing unrelated code on my local machine
Running unrelated PRs in CI

### Zebra logs

Exit Status:
command terminated unsuccessfully by signal 6
Stdout:
Stderr:
Message:  Sprout note commitment tree must exist if there is a finalized tip
Location: zebra-state/src/service/finalized_state/zebra_db/shielded.rs:97
Metadata:
failures:
config_tests

### Zebra Version

zebrad 1.2.0, multiple commits on different PRs

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

various machines: CI, local machine

### Additional information

The panic is from the `expect()` in:
https://github.com/ZcashFoundation/zebra/blob/4f6d28f9b115648f990a7c471ebc5e0b8a4048e6/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L84-L98

This issue should be fixed by PR #7392.

It might be possible for remote nodes to trigger this bug, or make it more likely, because it happens when a block is downloaded during launch. (Or maybe while writing another block.)

teor2345:
## Motivation

Older Zebra versions did not cache the genesis tree roots.
Zebra currently does a lot of unnecessary writes to the anchors, sprout trees, and history trees.

Since block validation, mempool validation, and RPCs can read these trees, and block validation writes them, this is a potential remotely-triggered CPU or disk denial of service.

Part of #7664, alternative strategy 1 (stop deleting keys).
This is also part of a permanent fix for bug #7618.
Adds docs for part of #7737.

## Solution

Improve the database size and performance, particularly during attacks via RPCs, mempool validation, or multiple block writes.

#### Format Upgrades

- Change the sprout and history key format from `tip_height` to `()` (empty key) in the code and docs
- Read the sprout, sapling and orchard genesis trees, then write their cached roots back to the database, and update the docs
- Remove genesis-specific tree code, which uses the wrong format
- Update tests now the genesis block has an entry for an empty tree (rather than using `default()` which is an empty tree)
- Increment the state minor version

#### Format Tests

- Check for incorrect sprout or history key formats, and duplicate tip trees
- Check that there aren't any un-cached tree roots for sprout, sapling, and orchard
- Update database snapshot tests
- Disable format validity checks in some tests that create invalid states

#### Refactors

- Refactor sprout, sapling, and orchard tree and anchor writes so they are all skipped if the trees are the same
    - The history tree changes with every block, so we don't do this check for it
- Refactor sprout, sapling and orchard tree and anchor writes so they automatically cache roots (and use the newer key format)
- This format upgrade is short, so some tests needed changes to wait for a short upgrade, rather than a long upgrade
- Fix metrics for the genesis block

#### Documentation

- Document the underlying performance bug that this PR fixes, by adding a section to the state upgrades doc.

## Review

This is a routine bug fix.

This is a database format change, so it needs 2 reviews.

### Reviewer Checklist


zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -275,16 +276,55 @@ impl DbFormatChange {
         // Start of a database upgrade task.
 
         let version_for_pruning_trees =
-            Version::parse("25.1.1").expect("Hardcoded version string should be valid.");
+            Version::parse("25.1.2").expect("Hardcoded version string should be valid.");
```
teor2345: If this merges after the subtree indexes (and it probably should), update this to a minor version bump

teor2345:
I'll re-open this if we decide these fixes are needed.

teor2345:
> Older Zebra versions used the height as the key for the Sprout tip tree, which risks concurrency bugs.

I have been seen this panic multiple times in CI and on my local machine, so I think we should do this fix eventually. I'm not sure if it is a release blocker, it depends how frequent it is.

book/src/dev/state-db-upgrades.md
```diff
@@ -87,7 +87,7 @@ We use the following rocksdb column families:
 | *Sprout*                           |                        |                               |         |
 | `sprout_nullifiers`                | `sprout::Nullifier`    | `()`                          | Create  |
 | `sprout_anchors`                   | `sprout::tree::Root`   | `sprout::NoteCommitmentTree`  | Create  |
-| `sprout_note_commitment_tree`      | `block::Height`        | `sprout::NoteCommitmentTree`  | Delete  |
+| `sprout_note_commitment_tree`      | `()`                   | `sprout::NoteCommitmentTree`  | Update  |
```
teor2345, line 112: TODO: fix `history_tree` the same way

teor2345:
This PR is needed as a permanent fix for #7618.

teor2345:
CD failed due to #7564, so I restarted it.

teor2345:
This is a database format change, so it needs 2 reviews, @arya2 and @upbqdn are you ok with reviewing it?

teor2345:
I have done a manual upgrade and full sync test for this PR.

The upgrade works fine from a `25.2.2` state on both mainnet and testnet. It takes about 4 seconds to do this upgrade, and a total of 8 seconds to do all the validatation on mainnet. (So it might add 1 second to periodic format validation in CI.)

The testnet full sync works fine. The mainnet full sync is at 80% and still going.

teor2345:
This PR could have an impact on syncing speed, but if it does it will very likely improve it. Currently my local machine syncs mainnet in just under 2 days (around 45-46 hours).

I'll check back after the weekend with the time this PR took locally.

zebrad/tests/common/cached_state.rs
```diff
@@ -39,6 +39,15 @@ pub const ZEBRA_CACHED_STATE_DIR: &str = "ZEBRA_CACHED_STATE_DIR";
 /// but long enough that it doesn't impact performance.
 pub const DATABASE_FORMAT_CHECK_INTERVAL: Duration = Duration::from_secs(5 * 60);
 
+/// Is the current state version upgrade longer than the typical CI sync time?
+///
+/// If is is set to `false`, but the state upgrades finish after zebrad is synced.
```
upbqdn:
```suggestion
/// If this is set to `false`, but the state upgrades finish after zebrad is synced,
```

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
+//! Updating the genesis note commitment trees to cache their roots.
```
arya2, line 1: Nitpick:

```suggestion
//! Updating the genesis note commitment trees and tip sprout tree to cache their roots.
```

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/// Quickly check that the genesis trees and sprout tip tree have cached roots.
+///
+/// This allows us to fail the upgrade quickly in tests and during development,
+/// rather than waiting to see if it failed.
+///
+/// # Panics
+///
+/// If the state is empty.
```
arya2, line 71: ~it looks like it'll return Ok(()) if the db is empty.~

Update: It could feasibly panic.

teor2345:
> This PR could have an impact on syncing speed, but if it does it will very likely improve it. Currently my local machine syncs mainnet in just under 2 days (around 45-46 hours).
> 
> I'll check back after the weekend with the time this PR took locally.

This PR took 44.5 hours to sync, which makes it 1-2 hours faster than the current `main` branch on my machine.

```
rg -e "Diagnostic Metadata" -e "finished initial sync" --max-count 2 ~/.cache/zebra-log/file
2023-10-11T23:43:43.771643Z  INFO zebrad::application: Diagnostic Metadata:
...
2023-10-13T20:11:37.943177Z  INFO {zebrad="62b07d2" net="Main"}: zebrad::components::sync::progress: finished initial sync to chain tip, using gossiped blocks sync_percent=100.000% current_height=Height(2260320) network_upgrade=Nu5 remaining_sync_blocks=0 time_since_last_state_block=0s
```

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+        if tree.cached_root().is_none() {
+            result = Err(format!(
+                "found un-cached sprout tree root after running genesis tree root fix \
+                 {root:?}"
+            ));
+            error!(?result);
+        }
+    }
+
+    for (height, tree) in db.sapling_tree_by_height_range(..) {
```
arya2, line 150: Why check every key if this upgrade only changes the value at `Height(0)`?

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    Ok(())
+}
+
+/// Detailed check that all trees have cached roots.
+///
+/// # Panics
+///
+/// If the state is empty.
```
arya2, line 126: ~It doesn't look like this will panic if the state is empty.~

Update: It could feasibly panic.

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+        .expect("caller has checked for genesis block");
+
+    // Writing the trees back to the database automatically caches their roots.
+    let mut batch = DiskWriteBatch::new();
+
+    // Fix the cached root of the Sprout genesis tree in its anchors column family.
+
+    // It's ok to write the genesis tree to the tip tree index, because is overwritten by
+    // the actual tip before the batch is written to the database.
+    batch.update_sprout_tree(upgrade_db, &sprout_genesis_tree);
```
arya2: Why insert the genesis tree first if `sprout_tree_for_tip()` returns the default tree when empty and the value is always overwritten?

zebra-state/src/service/finalized_state/zebra_db/chain.rs
```diff
@@ -36,41 +39,22 @@ impl ZebraDb {
...
         // versions 1.2.0 and earlier are no longer supported.
         //
         // # Concurrency
         //
-        // There is only one tree in this column family, which is atomically updated by a block
-        // write batch (database transaction). If this update runs between the height read and
-        // the tree read, the height will be wrong, and the tree will be missing.
-        // That could cause consensus bugs.
+        // There is only one entry in this column family, which is atomically updated by a block
+        // write batch (database transaction). If we used a height as the column family tree,
```
arya2:
```suggestion
        // write batch (database transaction). If we used a height as the key in this column family,
```

arya2 (dismissed):

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -456,45 +455,49 @@ impl DiskWriteBatch {
...
-        // (Transaction indexes, note commitments, and UTXOs are committed later.)
+        // Commit block, transaction, and note commitment tree data.
         self.prepare_block_header_and_transaction_data_batch(db, &finalized.verified)?;
 
+        // The consensus rules are silent on shielded transactions in the genesis block,
+        // because there aren't any in the mainnet or testnet genesis blocks.
+        // So this means the genesis anchor is the same as the empty anchor,
+        // which is already present from height 1 to the first shielded transaction.
+        //
+        // In Zebra we include their nullifiers and note commitments because it simplifies our code.
```
upbqdn: Without reading the code below, it wasn't clear to me what "their" refers to. I suggest:

```suggestion
        // In Zebra we include the nullifiers and note commitments in the genesis block because it simplifies our code.
```

zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
```diff
@@ -135,6 +153,16 @@ pub async fn run() -> Result<()> {
         use_internet_connection,
     )?;
 
+    // Wait for the state to upgrade, if the upgrade is long.
+    // If this line hangs, change DATABASE_FORMAT_UPGRADE_IS_LONG to false.
+    if DATABASE_FORMAT_UPGRADE_IS_LONG {
```
upbqdn, line 158: ~Do we have any long upgrades?~

Edit:

If I understand the docs correctly, long upgrades take longer than a full sync. Shouldn't we consider such upgrades impractical since even a full sync would be faster?

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -589,74 +582,112 @@ impl DiskWriteBatch {
...
+        // Sprout lookups need all previous trees by their anchors.
+        // The root must be calculated first, so it is cached in the database.
+        self.zs_insert(&sprout_anchors, tree.root(), tree);
+        self.zs_insert(&sprout_tree_cf, (), tree);
+    }
+
+    /// Legacy method: Deletes the range of Sprout note commitment trees at the given [`Height`]s.
+    /// Doesn't delete anchors from the anchor index. Doesn't delete the upper bound.
+    ///
+    /// From state format 25.3.0 onwards, the history trees are indexed by an empty key,
```
upbqdn:
```suggestion
    /// From state format 25.3.0 onwards, the Sprout trees are indexed by an empty key,
```

book/src/dev/state-db-upgrades.md
```diff
@@ -40,6 +40,28 @@ This means that:
...
+- using an iterator on a column family which also deletes keys
+- creating large numbers of iterators
+- holding an iterator for a long time
+
+See the performance notes and bug reports in:
+- https://github.com/facebook/rocksdb/wiki/Iterator#iterating-upper-bound-and-lower-bound
+- https://tracker.ceph.com/issues/55324
+- https://jira.mariadb.org/browse/MDEV-19670
+
+But need to use iterators for some operations, so our alternatives are (in preferred order):
```
upbqdn:
```suggestion
But we need to use iterators for some operations, so our alternatives are (in preferred order):
```

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+    let orchard_genesis_tree = upgrade_db
+        .orchard_tree_by_height(&Height(0))
+        .expect("caller has checked for genesis block");
+
+    // Writing the trees back to the database automatically caches their roots.
+    let mut batch = DiskWriteBatch::new();
+
+    // Fix the cached root of the Sprout genesis tree in its anchors column family.
+
+    // It's ok to write the genesis tree to the tip tree index, because is overwritten by
```
upbqdn:
```suggestion
    // It's ok to write the genesis tree to the tip tree index, because it's overwritten by
```

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
upbqdn: It's because of this line: https://github.com/zcashfoundation/zebra/blob/af5f97d43400be33a1ea725c1cd3d40a229f4dfa/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L645. 

We need to store the genesis tree at its anchor (together with its cached root). Note that anchors are the tree's roots. We then store the tip's tree again at another anchor. Since the roots of these two trees differ, the batch preserves both trees at their corresponding anchors.

upbqdn (dismissed):

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
upbqdn: It's because of this line: https://github.com/zcashfoundation/zebra/blob/af5f97d43400be33a1ea725c1cd3d40a229f4dfa/zebra-state/src/service/finalized_state/zebra_db/shielded.rs#L645.

We store the genesis and tip trees at their corresponding anchors. Since the anchors differ (note that an anchor is the tree's root), the batch keeps both trees at their anchors. We do this only for Sprout because we need to store a tree for each anchor. For Sapling & Orchard, we store a tree for each height instead.

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
```diff
@@ -0,0 +1,181 @@
...
+///
+/// Returns `Ok` if the upgrade completed, and `Err` if it was cancelled.
+///
+/// # Panics
+///
+/// If the state is empty.
+#[allow(clippy::unwrap_in_result)]
+#[instrument(skip(upgrade_db, cancel_receiver))]
+pub fn run(
+    initial_tip_height: Height,
```
upbqdn: Why don't we need to prefix `initial_tip_height` with an underscore since we don't use it anywhere?

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
teor2345: No, that's not actually what this change does, it only needs to update the tip sprout tree if the tip is at genesis.

Otherwise, the tip already has a cached root.

teor2345: (And whatever we do with the genesis tree in the tip column family is overwritten by the actual tip tree.)

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
teor2345: Because the security issue in the ticket happens for any uncached root, so we need to check them all.

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
teor2345: > Why insert the genesis tree first if `sprout_tree_for_tip()` returns the default tree when empty and the value is always overwritten?

Also this covers the case where the tip is the genesis tree.

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
teor2345: Because it is used by the instrument macro. Happy to use an underscore if you think that would be clearer?

zebrad/tests/common/cached_state.rs
```diff
@@ -39,6 +39,15 @@ pub const ZEBRA_CACHED_STATE_DIR: &str = "ZEBRA_CACHED_STATE_DIR";
 /// but long enough that it doesn't impact performance.
 pub const DATABASE_FORMAT_CHECK_INTERVAL: Duration = Duration::from_secs(5 * 60);
 
+/// Is the current state version upgrade longer than the typical CI sync time?
```
teor2345:
```suggestion
/// Is the current state version upgrade longer than the typical CI update sync time?
/// This is the time it takes Zebra to sync from a previously cached state to the current tip.
```

zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
teor2345: You're right those docs are unclear. A long upgrade takes longer than a sync from a previously cached state to the current tip. The 25.2 upgrade was like this because it took 20 minutes.

A short upgrade finishes before the update sync, for example the 25.1 and 25.3 upgrades.

Hopefully this comment change clears it up:
https://github.com/ZcashFoundation/zebra/pull/7392/files#r1362831651

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
upbqdn: I think it would because I noticed the function takes `initial_tip_height`, and then I noticed it uses `sprout_tree_for_tip()`, and I was thinking if `initial_tip_height` could be obsolete by the time `sprout_tree_for_tip()` is called, resulting in an inconsistency somewhere. Then I noticed we don't use `initial_tip_height` anywhere. Having it prefixed would be more explicit about a situation like this.

upbqdn: ###### Optional Suggestion

```suggestion
    _initial_tip_height: Height,
```

zebra-state/src/service/finalized_state/disk_format/upgrade/fix_tree_key_type.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use super::CancelFormatChange;
+
+/// Runs disk format upgrade for changing the sprout and history tree key types.
+///
+/// Returns `Ok` if the upgrade completed, and `Err` if it was cancelled.
+#[allow(clippy::unwrap_in_result)]
+#[instrument(skip(upgrade_db, cancel_receiver))]
+pub fn run(
+    initial_tip_height: Height,
```
upbqdn: ###### Optional Suggestion

```suggestion
    _initial_tip_height: Height,
```

arya2 (dismissed):

arya2:
`lightwalletd_wallet_grpc_tests` failed because [it marked the database format as upgraded](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:262) before [opening the RPC endpoint](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:278).

teor2345:
> `lightwalletd_wallet_grpc_tests` failed because [it marked the database format as upgraded](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:262) before [opening the RPC endpoint](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:278).

This happened before the other way around, let me work on a better fix.

teor2345:
> `lightwalletd_wallet_grpc_tests` failed because [it marked the database format as upgraded](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:262) before [opening the RPC endpoint](https://github.com/ZcashFoundation/zebra/actions/runs/6553219876/job/17798504721?pr=7392#step:9:278).

@arya2 your (deleted) suggestion here was really good and I ended up implementing it 🙂 

> It would be nice to have a ~expect_stdout_unordered_lines_match() method in TestChild for these cases.

zebra-test/src/command.rs
```diff
@@ -847,6 +848,92 @@ impl<T> TestChild<T> {
         }
     }
 
+    /// Checks each line of the child's stdout, until it finds every regex in `success_regexes`,
```
upbqdn:
```suggestion
    /// Checks each line of the child's stdout, until it finds every regex in `unordered_regexes`,
```

zebra-test/src/command.rs
```diff
@@ -847,6 +848,92 @@ impl<T> TestChild<T> {
...
+            let matched_indices: HashSet<usize> = regex_list.matches(&line).iter().collect();
+            unmatched_indexes = &unmatched_indexes - &matched_indices;
+
+            matched_lines.push(line);
+        }
+
+        Ok(matched_lines)
+    }
+
+    /// Checks each line of the child's stderr, until it finds every regex in `success_regexes`,
```
upbqdn:
```suggestion
    /// Checks each line of the child's stderr, until it finds every regex in `unordered_regexes`,
```

zebra-test/src/command.rs
```diff
@@ -847,6 +848,92 @@ impl<T> TestChild<T> {
         }
     }
 
+    /// Checks each line of the child's stdout, until it finds every regex in `success_regexes`,
+    /// and returns all the lines matched by any regex. Prints all stdout lines.
```
upbqdn: If I understand the method correctly, it would be more accurate to say that it returns the first matching line for each regex:

```suggestion
    /// and returns the first matching line for each regex or a list of unmatched regexes.
    /// Prints all stdout lines.
```

zebra-test/src/command.rs
```diff
@@ -847,6 +848,92 @@ impl<T> TestChild<T> {
...
+            unmatched_indexes = &unmatched_indexes - &matched_indices;
+
+            matched_lines.push(line);
+        }
+
+        Ok(matched_lines)
+    }
+
+    /// Checks each line of the child's stderr, until it finds every regex in `success_regexes`,
+    /// and returns all the lines matched by any regex. Prints all stderr lines.
```
upbqdn: If I understand the method correctly, it would be more accurate to say that it returns the first matching line for each regex:

```suggestion
    /// and returns the first matching line for each regex or a list of unmatched regexes.
    /// Prints all stdout lines.
```

upbqdn (dismissed):

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
upbqdn: "Panics if the state is empty, but not if it's too empty." :D

zebra-test/src/command.rs
teor2345: It's not quite "the first matching line for each regex", it's "all lines matched by any regex until each regex has matched" :

```suggestion
    /// and returns all lines matched by any regex, until each regex has been matched at least once.
    /// If the output finishes or the command times out before all regexes are matched, returns an error with
    /// a list of unmatched regexes. Prints all stdout lines.
```

zebra-test/src/command.rs
teor2345:
```suggestion
    /// and returns all lines matched by any regex, until each regex has been matched at least once.
    /// If the output finishes or the command times out before all regexes are matched, returns an error with
    /// a list of unmatched regexes. Prints all stderr lines.
```

teor2345:
We've had two reviews on the production code, and I've just changed docs and test code. So I've taken the `extra-reviews` label off.

upbqdn (approved):

zebra-state/src/service/finalized_state/disk_format/upgrade/cache_genesis_roots.rs
teor2345: Well technically a half-empty state is a format bug!

PR #7442, opened 2023-08-31
fix(state): If a format change is cancelled, also cancel the format check, and don't mark the database as upgraded 

teor2345:
## Motivation

Zebra can keep running the post-upgrade code when a format upgrade is cancelled.

If the format check is exhaustive, this can cause a panic on shutdown.

If some data is not covered by the format check, it can either:
- cause a panic on shutdown, or
- incorrectly mark the database as upgraded.

### Specifications

don't panic

This is a race condition, so it only happens sometimes, and it depends on other activity within Zebra, and on the machine it's running on.

## Solution

- Return an error when an upgrade is cancelled
- When an upgrade is cancelled, skip the upgrade check, and marking the database as upgraded
- Move an upgrade log to the correct location

## Review

This fix needs to be part of the 1.2.0 release, it is urgent.

### Reviewer Checklist


oxarbitrage (approved):
Looks good, thanks!

Issue #7441, opened 2023-08-31
Decide if we need to speed up adding subtrees for existing database blocks


arya2:
# TODO

- [ ] Make sure the current code is correct
- [ ] Work out if any performance upgrades are needed
- [ ] Split this ticket into multiple tickets, based on which performance upgrades we decide to do.

## Motivation

The database format upgrade for adding subtrees is slow on mainnet. It takes about 20 minutes on a recent machine.

This is already an acceptable time for wallet users to wait for a once-off upgrade, so this ticket is low or optional priority.

### Design Alternatives

Together, these design alternatives could reduce the upgrade time to a few minutes, which is how long it takes Zebra to start up anyway.

These alternatives are sorted in order of complexity. If we do each optimisation in a separate PR, and check the time after each one, we'll know if it worked, and when we can stop doing upgrades.

Many of these optimisations are from the `zcashd` PR in our ticket #6953: https://github.com/zcash/zcash/pull/6694/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R5815

#### Start at the relevant Network Upgrade

It isn't possible to have 

#### Only Read Transactions Ticket

If this fix turns out to be complicated, we should do profiling (for example, a flamegraph) to confirm if it's needed, before we write any complex code.

If reading blocks from the database is slow, we could skip reading the block headers, and only read transactions. There is already a database method (or database code in the existing block method) for this.

We should time the upgrade after this change, and if it isn't any faster, we shouldn't merge the change. It's possible that it's reading the transactions or calculating the note commitments that's the slow part.

#### Incremental Upgrade Ticket

If we start the upgrade from the first missing subtree, then we don't waste the work done by previously interrupted upgrades.

This reduces the time by the time spent on previous incomplete upgrades.

Implementing this ticket makes testing other changes more complicated, so we might want to do it last. (If we do incremental upgrades, they skip updating some subtrees, so our validity tests won't be testing the latest code any more.)

#### Testing Even Indexes Ticket & Re-Using Even Ommers Ticket

This depends on ticket #7446.

Before we start this change, we also need to add another 8 tests for fixed subtrees from `zcashd`:
* Odd subtree indexes & even subtree indexes

Because of the way the frontier is structured, the ommers of incomplete odd-numbered subtrees contain the root of the previous even-numbered subtree. The subtree's root stays in the ommers until the odd-numbered subtree is completed.

[...]

PR #7437, opened 2023-08-31
change(state): Add note subtree indexes for new and existing blocks

teor2345:
## Motivation

This is needed to support "spend before sync" in lightwalletd.

Closes https://github.com/ZcashFoundation/zebra/issues/6953.

### Specifications

See https://github.com/zcash/zcash/pull/6677/files#diff-decae4be02fb8a47ab4557fe74a9cb853bdfa3ec0fa1b515c0a1e5de91f4ad0bR1458-R1475

The specs for subtrees aren't in a Zcash specification yet.

### Complex Code or Requirements

The upgrade runs concurrently with writing new blocks and subtrees, but the subtree indexes they write are different.

This code runs in separate threads, so we're careful to handle cancel, drop, and shutdowns.

Rebuilding subtrees could be CPU-intensive and take a long time.

## Solution

- Write subtrees to the finalized state for new blocks
- Upgrade existing finalized states to calculate and add subtrees
    - Assert that subtree indexes are correct
    - Log when each subtree is added during an upgrade, because upgrades are slow
- Bump the database format version to 25.2.0
- Check the subtree upgrade is valid after it has completed, and on every Zebra launch

Bug fixes:
- Prevent a concurrency bug by using the correct API (it's probably impossible anyway)

Refactors:
- Simplify subtree APIs
- Simplify subtree upgrade code
- Clarify some comments

### Testing

This PR has a state validity test. There are also asserts that subtrees are added in order and continuously.

## Review

This PR is needed to test the subtree RPC method in PR #7436.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Related Work

This is a copy of PR #7350 with some fixes to make it work with the latest `main` branch.

PR #7445, opened 2023-09-01
change(state): Speed up note subtree indexes upgrade

arya2:
## Motivation

It would be nice for this format upgrade to be faster.

Part of #7441.

Depends-On #7437.

## Solution

- Skip appending notes for right hand subtrees where the previous subtree root should be in the ommers

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block.rs
+++ b/zebra-chain/src/block.rs
@@ -2,6 +2,8 @@
 
 use std::{collections::HashMap, fmt, ops::Neg, sync::Arc};
 
+use halo2::pasta::pallas;
+
 use crate::{
     amount::NegativeAllowed,
     block::merkle::AuthDataRoot,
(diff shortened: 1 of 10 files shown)


teor2345 (changes requested):
I think this change might be blocked by testing that the subtree roots calculated by our simpler code are correct:
- #7446

If our roots are correct, but this PR makes them incorrect, it will be tricky to track down that bug without any tests.

I'm also not sure about the priority of this performance improvement, because it's not scheduled in this sprint. There are 4 bug or testing tickets in the tracking issue that could be a higher priority. Let's talk with the team and decide if any performance improvements are needed?

(In my opinion, a 20 minute once-off upgrade could be acceptable performance already.)

PR #7452, opened 2023-09-04
RPC testing branch with `z_getsubtreesbyindex` RPC and state data

teor2345:
## Motivation

This is a branch that we can use to test the `z_getsubtreesbyindex ` RPC method with real state data.

We can:
- do manual tests on this branch, or
- write snapshot or integration tests based on this branch


teor2345:
#7437 is about to merge, so this isn't needed any more.

Issue #7450, opened 2023-09-04
Manually test `z_getsubtreesbyindex` using `zcash-rpc-diff`


teor2345:
# Motivation

We want to test that Zebra matches `zcashd`'s output.

### Specifications

`zcashd` implemented `z_getsubtreesbyindex` in https://github.com/zcash/zcash/pull/6677

New fields are listed in https://github.com/zcash/zcash/pull/6677/files#diff-decae4be02fb8a47ab4557fe74a9cb853bdfa3ec0fa1b515c0a1e5de91f4ad0bR1458-R1475

### Test Commands

We can compare the entire list of subtrees with these commands:
```sh
zcash-rpc-diff $ZEBRA_RPC_PORT z_getsubtreesbyindex sapling 0
zcash-rpc-diff $ZEBRA_RPC_PORT z_getsubtreesbyindex orchard 0
ZCASHD_EXTRA_ARGS="-testnet" zcash-rpc-diff $ZEBRA_TESTNET_RPC_PORT z_getsubtreesbyindex sapling 0
ZCASHD_EXTRA_ARGS="-testnet" zcash-rpc-diff $ZEBRA_TESTNET_RPC_PORT z_getsubtreesbyindex orchard 0
```

This requires fully synced and upgraded Zebra and `zcashd` instances. The `zcashd` instance must have `lightwalletd=1` in its config, and might require a `-reindex`. This can take 24-48 hours.

### Test State

To catch bugs like #7532, we need to do these checks on both:
- an upgraded Zebra state
- a newly created and fully synced Zebra state

### Test Coverage

We can't close this ticket until we've finished making changes to the related RPC and state code, because those changes could break compatibility.

teor2345:
The initial format is looking good so far, but I can't compare data until `zcashd` has finished reindexing:
```
$ zcash-rpc-diff 28232 z_getsubtreesbyindex sapling 2 3
Checking first node release info...
Checking second node release info...
Connected to zebrad v1.2.0+40.gb3d5aac (port 28232) and zcashd v5.6.1-gb7af08e90de (zcash-cli zcash.conf port).

Checking zebrad v1.2.0+40.gb3d5aac network and tip height...
Checking zcashd v5.6.1-gb7af08e90de network and tip height...

WARNING: comparing RPC responses from different heights:
zebrad v1.2.0+40.gb3d5aac is at: 2214786
zcashd v5.6.1-gb7af08e90de is at: 0

Request:
[...]

PR #7500, opened 2023-09-05
change(dev tools): Add an OUTPUT_DATA_LINE_LIMIT to zcash-rpc-diff

teor2345:
## Motivation

Some RPC outputs are very long, making them hard to paste into GitHub comments.

## Solution

When RPC outputs are identical, only show the first 20 and last 20 lines

`git diff` already limits the length of its outputs. (I think?)

## Review

This is a low priority developer tooling change.

### Reviewer Checklist

### DIFF
--- a/zebra-utils/zcash-rpc-diff
+++ b/zebra-utils/zcash-rpc-diff
@@ -22,6 +22,8 @@ JQ="${JQ:-jq}"
 # - Use `-rpccookiefile=your/cookie/file` for a cookie file.
 # - Use `-rpcpassword=your-password` for a password.
 ZCASHD_EXTRA_ARGS="${ZCASHD_EXTRA_ARGS:-}"
+# We show this many lines of data, removing excess lines from the middle of the output.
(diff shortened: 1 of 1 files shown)


oxarbitrage:
Can you show me where (in what RPC method) this modification was useful ?

An example output will be good enough.

teor2345:
> Can you show me where (in what RPC method) this modification was useful ?
> 
> An example output will be good enough.

There are some example outputs in this comment, click the triangles to show them:
https://github.com/ZcashFoundation/zebra/issues/7450#issuecomment-1707448036

The original output was 2000 lines, now it's around 40 lines, so it's much easier to copy.

oxarbitrage (approved):
I like that the new action will only take effect when the outputs are identical. 

Had a hard time trying to find an example, most of our rpc methods have at least some small difference with zcashd.

Examples in https://github.com/ZcashFoundation/zebra/issues/7450#issuecomment-1707448036 didn't worked for me (see https://github.com/ZcashFoundation/zebra/issues/7450#issuecomment-1710082568). 

However i made some dirty hacks here to check it out and everything looks good.

PR #7507, opened 2023-09-06
change(tests): Obtain lightwalletd tip from logs instead of grpc

### LINKED ISSUE #7335: Fix lightwalletd sync finish detection code in tests

oxarbitrage:
## Motivation

Close https://github.com/ZcashFoundation/zebra/pull/7335

## Solution

Get the tip from the lightwalletd logs. Currently a draft to see if the CI passes.

## Review

I guess how we get the block can be improved, it is a bit fragile but i think whatever improve will not remove the fact that we are relying on a specific well formed log line from lightwalletd.

### Reviewer Checklist

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -1835,7 +1835,7 @@ fn lightwalletd_integration_test(test_type: TestType) -> Result<()> {
 
         if test_type.needs_lightwalletd_cached_state() {
             lightwalletd
-                .expect_stdout_line_matches("Done reading [0-9]{1,7} blocks from disk cache")?;
+                .expect_stdout_line_matches("Done reading [0-9]{7} blocks from disk cache")?;
         } else if !test_type.allow_lightwalletd_cached_state() {
             // Timeout the test if we're somehow accidentally using a cached state in our temp dir
             lightwalletd.expect_stdout_line_matches("Done reading 0 blocks from disk cache")?;
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -1835,7 +1835,7 @@ fn lightwalletd_integration_test(test_type: TestType) -> Result<()> {
 
         if test_type.needs_lightwalletd_cached_state() {
             lightwalletd
-                .expect_stdout_line_matches("Done reading [0-9]{7} blocks from disk cache")?;
+                .expect_stdout_line_matches("Done reading [0-9]{1,7} blocks from disk cache")?;
         } else if !test_type.allow_lightwalletd_cached_state() {
             // Timeout the test if we're somehow accidentally using a cached state in our temp dir
             lightwalletd.expect_stdout_line_matches("Done reading 0 blocks from disk cache")?;
--- a/zebrad/tests/common/lightwalletd/sync.rs
+++ b/zebrad/tests/common/lightwalletd/sync.rs
@@ -11,11 +11,7 @@ use tempfile::TempDir;
 use zebra_node_services::rpc_client::RpcRequestClient;
 use zebra_test::prelude::*;
 
-use crate::common::{
-    launch::ZebradTestDirExt,
-    lightwalletd::wallet_grpc::{connect_to_lightwalletd, ChainSpec},
-    test_type::TestType,
-};
+use crate::common::{launch::ZebradTestDirExt, test_type::TestType};
 
 /// The amount of time we wait between each tip check.
 pub const TIP_CHECK_RATE_LIMIT: Duration = Duration::from_secs(60);
@@ -110,7 +106,7 @@ pub fn wait_for_zebrad_and_lightwalletd_sync<
         }
 
         tracing::info!(?test_type, "waiting for lightwalletd to sync to the tip...");
-        while !are_zebrad_and_lightwalletd_tips_synced(lightwalletd_rpc_port, zebra_rpc_address)? {
+        while !are_zebrad_and_lightwalletd_tips_synced(zebra_rpc_address, lightwalletd_mut)? {
             let previous_check = Instant::now();
 
             // To improve performance, only check the tips occasionally
@@ -171,22 +167,37 @@ pub fn wait_for_zebrad_and_lightwalletd_sync<
 /// Returns `Ok(true)` if zebrad and lightwalletd are both at the same height.
 #[tracing::instrument]
 pub fn are_zebrad_and_lightwalletd_tips_synced(
-    lightwalletd_rpc_port: u16,
     zebra_rpc_address: SocketAddr,
+    lightwalletd: &mut TestChild<TempDir>,
 ) -> Result<bool> {
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()?;
 
     rt.block_on(async {
-        let mut lightwalletd_grpc_client = connect_to_lightwalletd(lightwalletd_rpc_port).await?;
+        // We are going to try getting the current lightwalletd height by reading the lightwalletd logs.
+        let mut lightwalletd_next_height = 1;
 
-        // Get the block tip from lightwalletd
-        let lightwalletd_tip_block = lightwalletd_grpc_client
-            .get_latest_block(ChainSpec {})
-            .await?
-            .into_inner();
-        let lightwalletd_tip_height = lightwalletd_tip_block.height;
+        // Only go forward on getting next height from lightwalletd logs if we find the line we are interested in.
+        if let Ok(line) = lightwalletd.expect_stdout_line_matches("Waiting for block: [0-9]+") {
+            let line_json: serde_json::Value = serde_json::from_str(line.as_str())
+                .expect("captured lightwalletd logs are always valid json");
+            let msg = line_json["msg"]
+                .as_str()
+                .expect("`msg` field is always a valid string");
+
+            // Block number is the last word of the message. We rely on that specific for this to work.
+            let last = msg
+                .split(' ')
+                .last()
+                .expect("always possible to get the last word of a separated by space string");
+            lightwalletd_next_height = last
+                .parse()
+                .expect("the last word is always the block number so it can be parsed to i32 ");
+        }
+
+        // The last height in lightwalletd is the one the program is expecting minus one.
+        let lightwalletd_tip_height = (lightwalletd_next_height - 1) as u64;
 
         // Get the block tip from zebrad
         let client = RpcRequestClient::new(zebra_rpc_address);
--- a/zebrad/tests/common/lightwalletd/sync.rs
+++ b/zebrad/tests/common/lightwalletd/sync.rs
@@ -179,6 +179,10 @@ pub fn are_zebrad_and_lightwalletd_tips_synced(
         let mut lightwalletd_next_height = 1;
 
         // Only go forward on getting next height from lightwalletd logs if we find the line we are interested in.
+        //
+        // TODO: move this blocking code out of the async executor.
+        // The executor could block all tasks and futures while this code is running.
+        // That's ok for now, but it might cause test hangs or failures if we spawn tasks, select(), or join().
         if let Ok(line) = lightwalletd.expect_stdout_line_matches("Waiting for block: [0-9]+") {
             let line_json: serde_json::Value = serde_json::from_str(line.as_str())
                 .expect("captured lightwalletd logs are always valid json");
--- a/.github/workflows/continous-integration-docker.yml
+++ b/.github/workflows/continous-integration-docker.yml
@@ -637,7 +637,7 @@ jobs:
       needs_zebra_state: true
       needs_lwd_state: true
       # since we do a full sync in every PR, the new cached state will only be a few minutes newer than the original one
-      saves_to_disk: false
+      saves_to_disk: true
       disk_prefix: lwd-cache
       disk_suffix: tip
       root_state_path: '/var/cache'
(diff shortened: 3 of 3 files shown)


zebrad/tests/acceptance.rs
```diff
@@ -1835,7 +1835,7 @@ fn lightwalletd_integration_test(test_type: TestType) -> Result<()> {
 
         if test_type.needs_lightwalletd_cached_state() {
             lightwalletd
-                .expect_stdout_line_matches("Done reading [0-9]{1,7} blocks from disk cache")?;
+                .expect_stdout_line_matches("Done reading [0-9]{7} blocks from disk cache")?;
```
teor2345, line 1838: This change will make some tests fail until all the `lightwalletd` cached states are synced properly again. Can you split it into a separate PR, and we'll merge it a week or so after this PR merges?

Until the fix merges, and all the PRs are updated, the fixed code will be fighting with the buggy code to make cached states. If the buggy code wins, this check will fail, and every PR that has the fix will fail.

(We used to do a full sync before every test, so doing it in the same PR wouldn't have been a problem. But that got too slow, so now we only do the full sync every week. So we have to do some fixes in stages. Arya and I both missed this when we wrote the ticket.)

teor2345: Hmm, actually, I think I have a better solution, see my comment below.

So there's no action needed here.

teor2345, line 1838: Some tests in this PR will fail on this line, until we run the fixes below in CI.

But that's good, because when those tests pass, that tells us we've fixed the cached state and can merge this PR.

Edit: We'll need to temporarily undo these changes, just until we fix the cached state.

zebrad/tests/common/lightwalletd/sync.rs
```diff
@@ -171,22 +167,37 @@ pub fn wait_for_zebrad_and_lightwalletd_sync<
...
-        let lightwalletd_tip_block = lightwalletd_grpc_client
-            .get_latest_block(ChainSpec {})
-            .await?
-            .into_inner();
-        let lightwalletd_tip_height = lightwalletd_tip_block.height;
+        // We are going to try getting the current lightwalletd height by reading the lightwalletd logs.
+        let mut lightwalletd_next_height = 1;
+
+        // Only go forward on getting next height from lightwalletd logs if we find the line we are interested in.
+        if let Ok(line) = lightwalletd.expect_stdout_line_matches("Waiting for block: [0-9]+") {
```
teor2345, line 186: I can't tell if this code works, because it never runs. `lightwalletd` is syncing correctly, but CI times out before it fully syncs. It gets very close, but it might need another hour or so.

That is what I understand from these logs:
> {"app":"lightwalletd","level":"info","msg":"Adding block to cache 2061544 0000000000d8142b0034f733372c1e613dcc1cec196077763ccaf2355f3a73bd","time":"2023-09-07T18:41:11Z"}
> {"app":"lightwalletd","level":"info","msg":"Adding block to cache 2061956 00000000014c81044db3de5de5b8d5af62b96e1c9ef62508237bfc673cae7770","time":"2023-09-07T18:41:15Z"}
> {"app":"lightwalletd","level":"info","msg":"Adding block to cache 2062341 0000000000f3670f1968818f5555ba355bccaecfee6a83237bbedeb55d5800d6","time":"2023-09-07T18:41:19Z"}
> Error: The operation was canceled.

https://github.com/ZcashFoundation/zebra/actions/runs/6102690617/job/16581662340?pr=7507#step:8:5532

Here's how to fix that (I can do this if you want):

<details>
<summary>outdated steps</summary>

1. Open another PR that disables saving to disk in the full sync. That is the only job that saves to disk. So if we turn it off, this PR controls the cached `lightwalletd` state for all of Zebra. This line controls saving to disk:
https://github.com/ZcashFoundation/zebra/blob/5b0bc700981c39989f46a099d328e00b2166ebac/.github/workflows/continous-integration-docker.yml#L602

2. In this PR, temporarily stop checking that the update sync has a good cached state, by undoing this change:
https://github.com/ZcashFoundation/zebra/pull/7507/files#r1321104076

3. In this PR, temporarily force the update sync to save to disk, and remove its time limit:
https://github.com/ZcashFoundation/zebra/blob/5b0bc700981c39989f46a099d328e00b2166ebac/.github/workflows/continous-integration-docker.yml#L639-L640

The settings we need are:
```
saves_to_disk: true
force_save_to_disk: true
is_long_test: true
```

Then undo most of what we've just done:

4. After that's run, undo step 2, and make sure all the jobs pass. This means the cached state is fixed.

5. Now keep the update sync saving to disk, but undo the force and the long job from step 3:
```
force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
is_long_test: false
```

6. After this PR merges and the fix is in the scheduled job, undo step 1.

</details>

I don't think Arya or I realised it would be this complicated when we opened the ticket. (We were so close, if the sync had taken 1 less hour, everything would work.)

I am happy to make all these changes, or you can do some of them if you want. Let me know what you decide.

oxarbitrage: The step 3 is not clear. It says to update the `saves_to_disk` of line 640 but but not which `force_save_to_disk` or `is_long_test` to change. There are several occurrences of those in this file and i am not sure which ones i should change.

Overall i think there should be a better way for us to make stuff like this happen. I don't have a solution but this does not look as a good approach in general.

teor2345: > The step 3 is not clear. It says to update the `saves_to_disk` of line 640 but but not which `force_save_to_disk` or `is_long_test` to change. There are several occurrences of those in this file and i am not sure which ones i should change.

After having a look at what's possible in more detail, I think we can skip most of this step, if we get everything done before Friday. That's risky but seems worth it if you want to make it easier.

> Overall i think there should be a better way for us to make stuff like this happen. I don't have a solution but this does not look as a good approach in general.

I also want an easier way, and we can make it easier going forward. I don't know what tickets to open to make it happen, but someone should think about that. Maybe you could open a bug ticket, and we can work out the possible solutions later?

Here's what I'm thinking about the alternatives, what do you think?

The easiest alternatives were only available a month ago when the tests started failing. We could have either:
* not disabled the checks that stop invalid cached states from being saved, and fixed the sync issue straight away, or
* disabled the checks, and turned off saving the states we knew were invalid at the same time.

But that decision was made a month ago, so we are stuck with invalid cached states now. (But I think #7390 will stop this happening again, because it will skip saving lightwalletd cached state heights if there is already a higher one.)

Anyway, since there are already invalid cached states, that means our options to fix them without failing all the other PRs are more complicated.

The easiest way now is to make the sync fixes in this PR, merge them to main, delete all the invalid lightwalletd cached states, and re-run the full sync. But that either makes all the other PRs fail, or makes them all run a very slow full sync at the same time. Which is disruptive or expensive. And if we get it wrong we have to do it all over again. So that doesn't seem like an option.

But it's the usually way we'd do things. Pause everything, delete the bad caches, and rebuild them.

So I'm trying to work out the easiest way to do that manually, without breaking everything. It's a tricky problem, so please be patient with me.

teor2345: Here's a shorter set of steps that should do a manual full rebuild in about 8-16 hours of CI time:

1. Run a manual full lightwalletd sync on this PR's branch, and force it to save a working cached state:

2. After that's run, re-run all the jobs in this PR, and make sure they all pass. This means the cached state is fixed.

3. Make the update sync save to disk. This is an existing bug from when we moved the full sync to a scheduled job, but it made the invalid cached states much worse. Fixing it will stop that happening again.
https://github.com/ZcashFoundation/zebra/blob/5b0bc700981c39989f46a099d328e00b2166ebac/.github/workflows/continous-integration-docker.yml#L639-L640

If we get this all done before the scheduled job runs, we don't have to turn it off. Otherwise we need to merge PR #7522 before the scheduled job runs at 1200 UTC Friday:
https://github.com/ZcashFoundation/zebra/blob/e5a2f307a89c65a28ac1beb177bf78cbc0931c61/.github/workflows/continous-integration-docker.yml#L11-L15

teor2345: I started step 1 here:
https://github.com/ZcashFoundation/zebra/actions/runs/6151464098

Using these settings:
IMG_4697

teor2345: Before we do the new step 2, we'll need to revert commit [`0fb1093` (#7507)](https://github.com/ZcashFoundation/zebra/pull/7507/commits/0fb10938d9903d59c338b93a6fad4911aa2d969e).

Sorry about the extra work, sometimes it takes a while to think of shorter alternatives.

zebrad/tests/acceptance.rs
teor2345: Before we do the new step 2, we'll need to revert commit [`0fb1093` (#7507)](https://github.com/ZcashFoundation/zebra/pull/7507/commits/0fb10938d9903d59c338b93a6fad4911aa2d969e).

Sorry about the extra work, sometimes it takes a while to think of shorter alternatives.

teor2345: I just pushed a commit that reverts this temporary change.

zebrad/tests/common/lightwalletd/sync.rs
teor2345: The job finished successfully and created this cached state:
https://console.cloud.google.com/compute/imagesDetail/projects/zfnd-dev-zebra/global/images/lwd-cache-issue7335-0fb1093-v25-mainnet-tip-042848?hl=en&project=zfnd-dev-zebra

It has the correct heights (valid heights in green squares, invalid heights in red circles):
Screenshot 2023-09-12 at 15 24 28

teor2345: If CI passes on this PR with the new cached state, we can do the final step:

> 3. Make the update sync save to disk. This is an existing bug from when we moved the full sync to a scheduled job, but it made the invalid cached states much worse. Fixing it will stop that happening again.
> https://github.com/ZcashFoundation/zebra/blob/5b0bc700981c39989f46a099d328e00b2166ebac/.github/workflows/continous-integration-docker.yml#L639-L640

oxarbitrage: Trying the last step at https://github.com/ZcashFoundation/zebra/pull/7507/commits/67e513f2359fb9e5038d6e9207cd46c5d1315302

oxarbitrage, line 186: It is possible to go after the `Adding block to cache XXXX` instead of the one that waits for the next one and substract one as we are doing now. I don't see much difference on using one or the other.

teor2345: I am running the ECC `lightwalletd` fork, and I don't see any "Adding block to cache" messages once it reaches the tip. So I think the test will be more reliable with "Waiting for block".

teor2345: This seems to almost work, but it is failing when it tries to save the disk with:
> Missing sync height in logs: 

https://github.com/ZcashFoundation/zebra/actions/runs/6161817584/job/16724405147?pr=7507#step:13:307

So we need to use a different `height_grep_text` (the default is for `zebrad`).

It looks like the log that reliably has the `lightwalletd` height is:
> {"app":"lightwalletd","level":"info","msg":"Waiting for block: 2224633","time":"2023-09-12T17:02:08Z"}

https://github.com/ZcashFoundation/zebra/actions/runs/6161817584/job/16724405147?pr=7507#step:13:260

I fixed this in commit 8bad89f3c467f9744d2ef24ac957390fe6ca7079 by fixing the full sync, then copying most of the arguments across into the update sync. (Maybe there's a way to combine them to make this easier, because they should always have the same settings, except for long test.)

teor2345, line 186:
```suggestion
        //
        // TODO: move this blocking code out of the async executor.
        // The executor could block all tasks and futures while this code is running.
        // That's ok for now, but it might cause test hangs or failures if we spawn tasks, select(), or join().
        if let Ok(line) = lightwalletd.expect_stdout_line_matches("Waiting for block: [0-9]+") {
```

teor2345 (approved):
Thanks for your work on this. We almost got to the end of the test, it just failed at the last step when it was writing the state. And I think I fixed that.

So I'm going to approve this PR now, and hope it passes and merges. That way it should all be fixed before the scheduled full sync runs on Friday.

teor2345:
> So I'm going to approve this PR now, and hope it passes and merges.

The update sync passed CI and saved a cached state:
https://console.cloud.google.com/compute/imagesDetail/projects/zfnd-dev-zebra/global/images/lwd-cache-7507-merge-3df0524-v25-mainnet-tip-u-000145?hl=en&project=zfnd-dev-zebra

The heights were updated and recorded in the labels correctly:
Screenshot 2023-09-13 at 10 32 19

So this PR should merge once the other tests pass.

PR #7510, opened 2023-09-07
change(tools): Show zcash-cli version at the start of zcash-rpc-diff

teor2345:
## Motivation

We're seeing inconsistent behaviour from `zcash-rpc-diff` on different developers machines.

For example:
https://github.com/ZcashFoundation/zebra/issues/7450#issuecomment-1710717743

## Solution

- Show the `zcash-cli` version when `zcash-rpc-diff` runs
- Show the `bash` shell version when `zcash-rpc-diff` runs

## Review

This is a low priority test tool change.

### Reviewer Checklist


oxarbitrage (approved):
Thanks, i like it. This will help to avoid some confusion as there are some differences between versions, for example https://github.com/ZcashFoundation/zebra/issues/7450#issuecomment-1710082568

Just tested it and works.

PR #7466, opened 2023-09-04
fix(rpc): Format subtree roots in little-endian order

teor2345:
## Motivation

I assumed that subtree roots would be in big-endian order, but `zcashd` formats them in little-endian order.

### Logs

<details>

<summary>detailed logs showing a diff with reversed byte order</summary>

```
$ zcash-rpc-diff 28232 z_getsubtreesbyindex sapling 0 3                            
Checking first node release info...                                                                  
Checking second node release info...            
Connected to zebrad v1.2.0+40.gb3d5aac (port 28232) and zcashd v5.6.1-gb7af08e90de (zcash-cli zcash.conf port).
                                                                                                     
Checking zebrad v1.2.0+40.gb3d5aac network and tip height... 
Checking zcashd v5.6.1-gb7af08e90de network and tip height...
                                                                                                     
WARNING: comparing RPC responses from different heights:                                  
zebrad v1.2.0+40.gb3d5aac is at: 2215759                                                                                                                                                                   
zcashd v5.6.1-gb7af08e90de is at: 1867054                                                                                                                                                                  
                                                                                                                                                                                                           
Request:                                                                                                                                                                                                   
z_getsubtreesbyindex sapling 0 3                                                                     
                                                                                                     
Querying zebrad v1.2.0+40.gb3d5aac main chain at height >=2215759...                           
                                                                                                     
real    0m0.021s                                                                                     
user    0m0.004s                                                                                                                                                                                           
sys     0m0.003s                                                                                                                                                                                           
                                                                                                     
Querying zcashd v5.6.1-gb7af08e90de main chain at height >=1867054...                                                                                                                                      
                                                                                                                                                                                                           
real    0m0.149s                                                                                                                                                                                           
user    0m0.004s                                                                                     
sys     0m0.004s                                                                                     
                                                                                                     
                                                                                                     
Response diff between zebrad v1.2.0+40.gb3d5aac and zcashd v5.6.1-gb7af08e90de:      
--- /run/user/1000/tmp.KsyUuhrBLV.rpc-diff/zebrad-main-2215759-z_getsubtreesbyindex.json        2023-09-05 09:11:38.020957578 +1000
+++ /run/user/1000/tmp.KsyUuhrBLV.rpc-diff/zcashd-main-1867054-z_getsubtreesbyindex.json        2023-09-05 09:11:38.169956040 +1000
@@ -3,15 +3,15 @@                                                                                    
   "start_index": 0,                                                                                 
   "subtrees": [                                                                                     
[...]

PR #7514, opened 2023-09-08
Add snapshot tests for the `z_getsubtreesbyindex` RPC

### LINKED ISSUE #7513: Add snapshot tests for the `z_getsubtreesbyindex` RPC
## Suggested solution

- Add snapshot tests that check the response format of the `z_getsubtreesbyindex` RPC separately for Sapling and Orchard.

upbqdn:
## Motivation

Close #7513.

## Solution

- Test the response format from `z_getsubtreesbyindex` for Sapling and Orchard.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods/tests/snapshot.rs
+++ b/zebra-rpc/src/methods/tests/snapshot.rs
@@ -5,16 +5,18 @@
 //! cargo insta test --review
 //! ```
 
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
 
 use insta::dynamic_redaction;
 
 use zebra_chain::{
     block::Block,
+    chain_tip::mock::MockChainTip,
     parameters::Network::{Mainnet, Testnet},
     serialization::ZcashDeserializeInto,
+    subtree::NoteCommitmentSubtreeData,
 };
-use zebra_state::MAX_ON_DISK_HEIGHT;
+use zebra_state::{ReadRequest, ReadResponse, MAX_ON_DISK_HEIGHT};
 use zebra_test::mock_service::MockService;
 
 use super::super::*;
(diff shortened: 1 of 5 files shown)


zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -316,6 +322,88 @@ async fn test_rpc_response_data_for_network(network: Network) {
...
+        Buffer::new(mempool, 1),
+        state.clone(),
+        latest_chain_tip,
+    );
+
+    // Test the response format from `z_getsubtreesbyindex` for Sapling.
+
+    // Mock the data for the response.
+    let mut subtrees = BTreeMap::new();
+    let subtree_root = [0u8; 32].as_slice().try_into().expect(
```
teor2345, line 350: Nitpick: If we use a fake root value that isn't symmetrical, then this test will catch bugs like #7466.

For example:
```rust
let mut subtree_root = [0u8; 32];
subtree_root[0] = 0xff;
```

Feel free to submit this in another PR, but I don't think it's a blocker because one of the other tests should cover it.

teor2345 (approved):
Thanks, looks great!

Edit: I made a note about test coverage, but I resolved it because PR #7515 already covers it.

teor2345:
Failed in the merge queue when installing protoc with:
> Error: unable to get latest version

https://github.com/ZcashFoundation/zebra/actions/runs/6139283627/job/16657035984#step:4:13

@gustavovalverde we're seeing a lot of different errors installing protoc, maybe we could replace the action with a manual install script?

teor2345:
@mergifyio refresh

zebra-rpc/src/methods/tests/snapshot.rs
teor2345: PR #7515 already covers this issue, so there's no need to open another PR.

PR #7349, opened 2023-08-17
change(tests): update tests for new lightwalletd version used in Zebra

oxarbitrage:
## Motivation

We switched lightwalletd versions used in Zebra in https://github.com/ZcashFoundation/zebra/pull/7307

This should allow us to readd some asserts about the mempool that were disabled in the aditya version.

## Solution

- Change some asserts due to missing rpc methods in the previous used version. 

## Review

This PR is a draft until i can check if the modified asserts pass the CI.

After that, anyone can review. 

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #7519, opened 2023-09-11
change(dev tools): Limit diff length, and mempool check tx count and length

teor2345:
## Motivation

I was using `getrawmempool` to check why PR #7349 wasn't working. It didn't help.

But I found out that `zcash-rpc-diff` outputs hundreds of thousands of lines when the mempool is full of transactions (8000 transactions).

## Solution

Limit diff lengths
Limit number of mempool transactions checked
Limit transaction data length

### Testing

I manually ran:
```
zcash-rpc-diff 28232 getrawmempool
```

On mainnet when the mempool was full, and the output was only a few hundred lines long.

## Review

This is a low priority developer cleanup.

### Reviewer Checklist

### DIFF
--- a/zebra-utils/zcash-rpc-diff
+++ b/zebra-utils/zcash-rpc-diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -uo pipefail
 
 # Sends a `zcash-cli` command to a Zebra and zcashd instance,
 # and compares the results.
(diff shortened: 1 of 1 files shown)


oxarbitrage:
I tried something like this at https://github.com/ZcashFoundation/zebra/pull/7349 as well using `getrawmempool` but not with `zcash-rpc-diff`.

I noticed the huge amount of transactions the mempool seems to be holding probably all the time ? I think this could deserves some closer monitoring at some point.

oxarbitrage (approved):
There is no point on seeing so many transactions normally unless you really want it, the developer can change the limit if so.

I am not able to test it because i am still syncronizing a new zcashd instance but it seems like it should work, and be helpful in some cases to have it this way.

teor2345:
> I noticed the huge amount of transactions the mempool seems to be holding probably all the time ? I think this could deserves some closer monitoring at some point.

Yeah I was wondering about this, maybe a new spam attack.

conradoplg:
> > I noticed the huge amount of transactions the mempool seems to be holding probably all the time ? I think this could deserves some closer monitoring at some point.
> 
> Yeah I was wondering about this, maybe a new spam attack.

Yep, the mempool is constantly full now due to a spam attack. We might want to reduce its size in our tests.

PR #7520, opened 2023-09-11
feat(ci): add self hosted runner for long tests

### LINKED ISSUE #5572: ci: use self-hosted runners for CI pipelines
## Motivation

Some of our tests, mainly our full sync test, are way above [GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) limits, being the 6 hours timeout the main limitation. 

We might want to start testing GitHub [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) to raise the limits.

### Designs

- Deploy a Self-hosted runner to GCP using terraform
- Confirm how the cached state is handled with a self-hosted runner (SHR) 
- Validate how SSH connections will be handled in a SHR
- Move the full-sync job to this runner
- Validate time, reliability and cost of hosting the runner

## Related Work

Another approach would be using #5551, if we get through their beta waitlist. But this might won't get rid from the 6 hours limit.

gustavovalverde:
## Motivation

We've been creating several jobs to workaround GitHub's 6 hour time limit for hosted runners, but this approach has been hard to maintain, created several edge cases (issues) with SSH connections and made the workflow way too verbose.

Closes: #5572
Depends-On: #7507

### Specifications

- Uses self-hosted runners, which were deployed using ACR: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller

- We need to confirm the regex in the `test-result` job works for all tests

## Solution

- Use self-hosted runner instead of GitHub Hosted runner
- Remove all intermediate jobs which were meant to extend the run of the test over the 6 hours limit
- Move the startup validation and test run steps into a single job
- Combine setup and launch jobs in a single job (2 jobs total)

## Review
- This should execute over 6 hours for a Zebra and LWD sync
- The sync should end successfully

### Reviewer Checklist


gustavovalverde:
Running test: https://github.com/ZcashFoundation/zebra/actions/runs/6145420016

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -104,12 +104,12 @@ env:
...
-  # set up the test, if it doesn't use any cached state
+  # set up and launch the test, if it doesn't use any cached state
   # each test runs one of the *-with/without-cached-state job series, and skips the other
-  setup-without-cached-state:
-    name: Setup ${{ inputs.test_id }} test
+  launch-without-cached-state:
+    name: Launch ${{ inputs.test_id }} test
     if: ${{ !inputs.needs_zebra_state }}
-    runs-on: ubuntu-latest
+    runs-on: zfnd-runners
```
teor2345, line 112: Nitpick: There's an actionlint warning on this line about custom runner names, with instructions on how to resolve it.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -145,7 +145,9 @@ jobs:
           known_hosts: unnecessary
 
       - name: Generate public SSH key
-        run: ssh-keygen -y -f ~/.ssh/google_compute_engine > ~/.ssh/google_compute_engine.pub
+        run: |
+          sudo apt-get update && sudo apt-get -qq install -y --no-install-recommends openssh-client \
+          && ssh-keygen -y -f ~/.ssh/google_compute_engine > ~/.ssh/google_compute_engine.pub
```
teor2345: Optional Nitpick: "run" checks each line for success, so you don't need to use `\` and `&&` here.

This applies to every new script in this PR.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -575,17 +488,21 @@ jobs:
...
-
-  # check the logs of the test we just launched for zebrad startup messages
+  # Show all the test logs, then follow the logs of the test we just launched, until it finishes.
+  # Then check the result of the test.
   #
-  # this step makes sure `zebrad` is running, and configured for `inputs.network`.
-  logs-startup:
-    name: Check startup for ${{ inputs.test_id }}
+  # If `inputs.is_long_test` is `false`, the Rust test harness mostly runs in this job.
+  # Otherwise, it mostly runs in the "logs" jobs.
```
teor2345: I don't think these comments are accurate any more, because this PR deletes the logs jobs.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -575,17 +488,21 @@ jobs:
...
+  test-result:
+    name: Run ${{ inputs.test_id }} test
     # We run exactly one of without-cached-state or with-cached-state, and we always skip the other one.
+    # We also keep the setup dependencies, so a re-run of this job will re-run the setup jobs.
```
teor2345: In my experience, retrying GitHub actions doesn't work like these comments say. Or maybe I understood the comments wrong.

The setup jobs are only re-run if they fail, if they are the specific job that's selected to be re-run, or if all jobs are re-run.

Re-running later jobs doesn't re-run earlier jobs. Instead, it re-uses their results, which fails because these jobs have the side-effect of launching an instance.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -575,17 +488,21 @@ jobs:
...
     needs: [ launch-with-cached-state, launch-without-cached-state ]
-    # If the previous job fails, we still want to show the logs.
+    # If the previous job fails, we also want to run and fail this job,
+    # so that the branch protection rule fails in Mergify and GitHub.
     if: ${{ !cancelled() }}
-    runs-on: ubuntu-latest
+    timeout-minutes: 7200 # 5 days
```
teor2345: Do we want to adjust this timeout depending on `is_long_test`?

Most of our tests are short and should finish well before 6 hours, currently 3 hours should be enough. (Our whole CI finishes in just over 2 hours.)

This seems important, because otherwise a test hang will confuse developers, make the long logs really hard to load in GitHub Actions, and cost us 5 days of Google Cloud machine time.

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -654,81 +573,50 @@ jobs:
...
         run: |
           gcloud compute ssh ${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }} \
           --zone ${{ vars.GCP_ZONE }} \
           --ssh-flag="-o ServerAliveInterval=5" \
           --ssh-flag="-o ConnectionAttempts=20" \
           --ssh-flag="-o ConnectTimeout=5" \
-          --command \
-          "\
+          --command=' \
+          set -e;
```
teor2345, line 592: Just checking why `-e` is needed? It seemed to work fine without it before, and the step already handles the test exit status correctly.

I remember that `tee` sometimes exits with a non-zero status when the logs end, and I'm not sure about the exit status of `grep`.

Maybe we could just stick to what was originally in the run step?

.github/workflows/deploy-gcp-tests.yml
```diff
@@ -654,81 +573,50 @@ jobs:
...
           --tail all \
           --follow \
           ${{ inputs.test_id }} | \
           tee --output-error=exit /dev/stderr | \
           grep --max-count=1 --extended-regexp --color=always \
-          -e 'estimated progress.*network_upgrade.*=.*Canopy' \
-          -e 'estimated progress.*network_upgrade.*=.*Nu5' \
-          -e 'test result:.*finished in' \
-          "
+          "test result: .*ok.* [1-9][0-9]* passed.*finished in"; \
```
teor2345, line 599: You mentioned in the PR that this is a complex part of the code. It looks like the regex is the same as in the previous Run job, so it should work. Just checking if you made any changes to it?

teor2345 (commented):
This looks great, thank you so much for simplifying this. I have a bunch of questions, but I think the most important one is about the long test setting and timeouts.

Everything else should still work the way it is, or if it doesn't work, we'll pick it up in the tests.

Would you mind delaying merging this PR until after #7507 merges?
@oxarbitrage and I are working on a complicated cached state rebuild, which needs to finish before the scheduled jobs run on Friday. So I'd like to get it done first.
https://github.com/ZcashFoundation/zebra/pull/7507#discussion_r1322038826

(And if you have any ideas for making that easier, feel free to comment on that PR, or open a ticket to make rebuilds work better.)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: That diff is very confusing, as I did not add the `set -e`, it's there in `main`:
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/deploy-gcp-tests.yml#L1684

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: There were some logs for LWD in previous jobs (for specific checkpoints). I'm not completely sure if those should be added here somehow, or if this regex will capture both Zebra and LWD

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: Added with: [fix: apply review recommendations](https://github.com/ZcashFoundation/zebra/pull/7520/commits/e0091969dbb3068085b155f65b890ea8fd6bb299)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: Removed in: [fix: apply review recommendations](https://github.com/ZcashFoundation/zebra/pull/7520/commits/e0091969dbb3068085b155f65b890ea8fd6bb299)

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: Was hoping to have an inline ignore for this one, but that's pending in `actionlint` 😩
- https://github.com/rhysd/actionlint/issues/237 
- https://github.com/reviewdog/action-actionlint/issues/80  

This would require adding a new file to our repo (I'm not keen on the idea)

.github/workflows/deploy-gcp-tests.yml
teor2345: This is a Rust test runner log, which is printed at the end of every Rust test. So we don't need to specifically check for `zebrad` or `lightwalletd` messages any more. (The Rust test does that.)

I was only checking for those messages as a way to split the test into roughly 6 hour jobs.

.github/workflows/deploy-gcp-tests.yml
teor2345: Great, happy to go with whatever works here!

.github/workflows/deploy-gcp-tests.yml
teor2345: No worries, it's just a lint.

teor2345 (approved):
Thanks for this, and for the quick changes! I have just approved the dependent PR, so this one will hopefully merge before tomorrow.

(I love deleting complicated code that isn't needed any more.)

PR #7515, opened 2023-09-09
tests(rpc): Add fixed test vectors for `z_getsubtreesbyindex` from zcashd to zebra

oxarbitrage:
Depends-On: #7555

## Motivation

We want to test the `z_getsubtreesbyindex` against fixed test vectors obtained from zcashd and see if they match zebra.

Close https://github.com/ZcashFoundation/zebra/issues/7446

## Solution

The idea is to add json outputs from `zcashd` into the `zebra-test` crate and then create a test that could compare this fixed outputs with answers from zebrad made in real time.

In order to test this in high heights we need a pretty much synchronized cache state for zebra.

Locally, i run this test as:

```
export ZEBRA_CACHED_STATE_DIR="/media/oxarbitrage/4eb53770-4e4f-4d80-a830-0914f3f5b89a/chain/zebra"
export RUST_LOG="info"
cargo test rpc_z_getsubtreesbyindex_zcashd_test_vectors -- --nocapture --ignored
```

## Review

This is not ready yet because i don't have the right outputs from zcashd. In order to use the `getsubtreesbyindex` i had to reindex and that actually ended up downloading all the blocks again so i have a couple of days at least to be in sync again with zcashd.

But i pushed so other people can take a look to the overall structure and let me know if this is a good or bad idea. Thanks!

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

PR #7521, opened 2023-09-11
 Test `z_getsubtreesbyindex` using a lightwalletd gRPC request

### LINKED ISSUE #7451: Test `z_getsubtreesbyindex` using a `lightwalletd` gRPC request
## Scheduling

This might be blocked by #7349.

## Motivation

We want to make sure `lightwalletd` can use our new `z_getsubtreesbyindex` JSON-RPC method successfully.

This test makes sure we continue working with `lightwalletd`, even if they make changes without telling us.

upbqdn:
## Motivation

Close #7451.

### Specifications

I got the docs for the protobuf message types from https://github.com/zcash/lightwalletd/blob/8269810eeefac56ce326b0f6878c1ea3821b8ad5/docs/rtd/index.html.

## Solution

- Add protobuf types for the lightwalletd's `GetSubtreeRoots` RPC.
- Call the RPC, and check the results.

I got bitten by this bug: https://github.com/tokio-rs/prost/issues/332 when I added the enum `ShieldedProtocol` to the file `service.proto`. The problem is that `prost` implicitly derives `Eq` for enums, so deriving it explicitly via `type_attribute` causes a conflict. Lukily, there is another method `message_attribute` that operates only on messages and not enums.

### Reviewer Checklist

### DIFF
--- a/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
+++ b/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
@@ -50,7 +50,8 @@ use crate::common::{
         sync::wait_for_zebrad_and_lightwalletd_sync,
         wallet_grpc::{
             connect_to_lightwalletd, Address, AddressList, BlockId, BlockRange, ChainSpec, Empty,
-            GetAddressUtxosArg, TransparentAddressBlockFilter, TxFilter,
+            GetAddressUtxosArg, GetSubtreeRootsArg, ShieldedProtocol,
+            TransparentAddressBlockFilter, TxFilter,
         },
     },
     test_type::TestType::UpdateCachedState,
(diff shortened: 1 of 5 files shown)


zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
```diff
@@ -337,6 +338,74 @@ pub async fn run() -> Result<()> {
...
+        .into_inner();
+
+    let mut counter = 0;
+    while let Some(_subtree) = subtrees.message().await? {
+        match counter {
+            0 => {
+                // TODO: add asserts that check:
+                // 1. The subtree hash.
+                // 2. The completing block hash.
+                // 3. The completing block height.
```
upbqdn: @teor2345 do you think it's worth adding and checking concrete vectors here?

My answer would be yes, but I haven't added them yet because I have to obtain them first.

teor2345: A big list of test vectors will be added in PR #7515, but feel free to add the first subtree here if you'd like. It's ok to just check mainnet and to use Zebra to generate the test vectors. (Ticket #7450 checks that Zebra matches `zcashd` for the first subtree.)

upbqdn: OK. I'm going to add two subtrees for Sapling and two for Orchard to test the RPC returns two consecutive subtrees.

upbqdn: Done in [06d4484](https://github.com/ZcashFoundation/zebra/pull/7521/commits/06d448425fbc54ec127ec37c4e8931e4883a899c).

teor2345 (approved):
Thanks, looks good!

PR #7552, opened 2023-09-13
imp(doc): add missing Docker variables and examples

### LINKED ISSUE #7008: Document Docker build-time & run-time config params

gustavovalverde:
## Motivation

We did not have documentation on the new variables and arguments which were introduced and changed in:
- #7200

Closes: #7008 

## Solution
- Document these arguments and variables
- Add examples for each one, and a mining example using some of these values

## Review

@upbqdn might want to have a look at this one

### Reviewer Checklist

## Follow Up Work

We should have a `docker-compose.yml` with some scenarios already set-up so it's easier for people to have a running node, for example, with non-volatile storage for their cache state.

### DIFF
--- a/book/src/user/docker.md
+++ b/book/src/user/docker.md
@@ -28,12 +28,15 @@ See [Building Zebra](https://github.com/ZcashFoundation/zebra#building-zebra) fo
 
 ## Advanced usage
 
-You're able to specify various parameters when building or launching the Docker image, which are meant to be used by developers and CI pipelines. For example, specifying the Network where Zebra will run (Mainnet, Testnet, etc), or enabling features like mining.
+You're able to specify various parameters when building or launching the Docker image, which are meant to be used by developers and CI pipelines. For example, specifying the Network where Zebra will run (Mainnet, Testnet, etc), or enabling features like metrics with Prometheus.
 
-For example, if we'd like to enable mining on the image, we'd build it using the following `build-arg`:
+For example, if we'd like to enable metrics on the image, we'd build it using the following `build-arg`:
+
+> [!WARNING]
+> To fully use and display the metrics, you'll need to run a Prometheus and Grafana server, and configure it to scrape and visualize the metrics endpoint. This is explained in more detailed in the [Metrics](https://zebra.zfnd.org/user/metrics.html#zebra-metrics) section of the User Guide.
 
 ```shell
-docker build -f ./docker/Dockerfile --target runtime --build-arg FEATURES='default-release-binaries getblocktemplate-rpcs' --tag local/zebra.mining:latest .
+docker build -f ./docker/Dockerfile --target runtime --build-arg FEATURES='default-release-binaries prometheus' --tag local/zebra.mining:latest .
 ```
 
 To increase the log output we can optionally add these `build-arg`s:
@@ -42,13 +45,13 @@ To increase the log output we can optionally add these `build-arg`s:
 --build-arg RUST_BACKTRACE=full --build-arg RUST_LOG=debug --build-arg COLORBT_SHOW_HIDDEN=1
 ```
 
-And after our image has been built, we can run it on `Mainnet` with the following command:
+And after our image has been built, we can run it on `Mainnet` with the following command, which will expose the metrics endpoint on port `9999` and force the logs to be colored:
 
 ```shell
-docker run --env NETWORK="Mainnet" --env RPC_PORT="8232" --env MINER_ADDRESS="t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc" -p 8232:8232 local/zebra.mining
+docker run --env LOG_COLOR="true" -p 9999:9999 local/zebra.mining
 ```
 
-Based on our actual `entrypoint.sh` script, the following configuration file will be generated (on the fly, at startup) and used by Zebra:
+Based on our actual `runtime-entrypoint.sh` script, the following configuration file will be generated (on the fly, at startup) and used by Zebra:
 
 ```toml
 [network]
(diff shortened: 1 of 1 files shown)


book/src/user/docker.md
```diff
@@ -26,8 +26,96 @@ docker run --detach zebra:local
 
 See [Building Zebra](https://github.com/ZcashFoundation/zebra#building-zebra) for more information.
 
+## Advanced usage
+
+You're able to specify various parameters when building or launching the Docker image, which are meant to be used by developers and CI pipelines. For example, specifying the Network where Zebra will run (Mainnet, Testnet, etc), or enabling features like mining.
+
+For example, if we'd like to enable mining on the image, we'd build it using the following `build-arg`:
```
upbqdn: I would omit the mining example because it requires users to submit their own Zcash address for rewards. Otherwise, Zebra won't start.

book/src/user/docker.md
```diff
@@ -26,8 +26,96 @@ docker run --detach zebra:local
...
+docker build -f ./docker/Dockerfile --target runtime --build-arg FEATURES='default-release-binaries getblocktemplate-rpcs'  --tag local/zebra.mining:latest .
+```
+
+To increase the log output we can optionally add these `build-arg`s:
+
+```shell
+--build-arg RUST_BACKTRACE=full --build-arg RUST_LOG=debug --build-arg COLORBT_SHOW_HIDDEN=1
+```
+
+And after our image has been built, we can run it in the `Mainnet` with the following command:
```
upbqdn: ###### Optional Suggestion

```suggestion
And after our image has been built, we can run it on `Mainnet` with the following command:
```

book/src/user/docker.md
```diff
@@ -26,8 +26,96 @@ docker run --detach zebra:local
...
+use_color = false
+[mining]
+miner_address = "t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc"
+```
+
+### Build time arguments
+
+#### Configuration
+
+- `FEATURES`: Specifies the features to build `zebrad` with. Example: `"default-release-binaries getblocktemplate-rpcs"`
```
upbqdn, line 70: ###### Optional Suggestion

It would be great if we could enumerate all options for all variables instead of examples, but that would be a lot more work, plus it's not possible with some variables.

book/src/user/docker.md
```diff
@@ -26,8 +26,96 @@ docker run --detach zebra:local
...
+
+#### Logging
+
+- `RUST_LOG`: Sets the trace log level. Example: `"debug"`
+- `RUST_BACKTRACE`: Enables or disables backtraces. Example: `"full"`
+- `RUST_LIB_BACKTRACE`: Enables or disables library backtraces. Example: `1`
```
upbqdn, line 76: ###### Optional Suggestion

Do you think we could unify the usage of `1` and `true`? E.g., use only one?

book/src/user/docker.md
```diff
@@ -26,8 +26,96 @@ docker run --detach zebra:local
...
+cache_dir = "/var/cache/zebrad-cache"
+[rpc]
+listen_addr = "0.0.0.0:8232"
+[tracing]
+use_color = false
+[mining]
+miner_address = "t1XhG6pT9xRqRQn3BHP7heUou1RuYrbcrCc"
+```
+
+### Build time arguments
```
upbqdn, line 66: ###### Optional Suggestion

This title uses "_arguments_", but the next one uses "_variables_".

upbqdn (dismissed):

book/src/user/docker.md
gustavovalverde: Any other good example that we can use and it might be something non-dev users would use?

book/src/user/docker.md
gustavovalverde: I used different titles considering that at build time we use `--build-arg` (arguments) and at runtime we use `--env` (environment variables).

book/src/user/docker.md
gustavovalverde: Yes, but we should do that in a different PR.

book/src/user/docker.md
upbqdn: Oh, I see.

book/src/user/docker.md
gustavovalverde: Let me know of any other example you'd like to use, or if we can continue with this one and improve later on.

upbqdn: A good one is `prometheus`: https://zebra.zfnd.org/user/metrics.html, but that requires running Prometheus and Grafana, as described in the link.

Another one is `progress-bar`. That one doesn't require anything extra, but I don't know if users will be interested in that since they will run Zebra in Docker. We might also remove the feature if we decide we want the progress bars by default.

Also, the `getblocktemplate-rpcs` feature might be removed once mining is not experimental.

I think the best way forward would be to add docs for running Zebra in Docker here: https://zebra.zfnd.org/user/metrics.html. We would then have a nice example feature for Docker, which would be `prometheus`.

upbqdn: The docs for Prometheus already describe how to run it together with Grafana in separate Docker containers. Having a third container with Zebra would be a nice fit.

upbqdn: When looking for examples, I noticed a few bugs in the docs. I fixed them in #7567.

gustavovalverde: This([fix(doc): use a better example for advanced usage](https://github.com/ZcashFoundation/zebra/pull/7552/commits/9d7516daef9441f748bb294321f7fe268758785f)) should be good now. When we have at least a `docker-compose`, then we can make a full example with it (as it's way better than handling individual containers)

upbqdn (approved):

book/src/user/docker.md
upbqdn: Sounds good.

Issue #7532, opened 2023-09-12
bug: some subtrees have incorrect end heights after a full sync


teor2345:
### Steps to fix

- [x] Fix the bug
- [x] Remove different code for full sync and upgrading legacy states
- [ ] ~~Remove different code for end of block and middle of block~~ Edit: this is tricky to do
- [x] Test using sapling and orchard
- [ ] Optionally: test using full sync (+restart Zebra) and legacy upgrades
- [x] Revert PR #7555

### What happened?

I expected to see this happen:

Subtree end heights should be the block they are completed in.
Subtrees should match `zcashd`.

Instead, this happened: 

Subtree end heights were 1 more than they should be, under these conditions:
- the subtree was completed at the end of a block, and
- the subtree was created by a newly synced tip block (not an upgrade).

Some subtrees won't match `zcashd` after a full Zebra sync.

### What were you doing when the issue happened?

Testing PR #7531 on my local machine.

### Zebra logs

```
The application panicked (crashed).
Message:  already checked that the block completed a subtree: 
updated subtree:
index: Some(NoteCommitmentSubtreeIndex(0))
remaining notes: 0
is complete: true
original subtree:
index: NoteCommitmentSubtreeIndex(0)
remaining notes: 3
is complete: false
Location: zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs:599
Metadata:
version: 1.2.0+46.g8e74e49.modified
Zcash network: Mainnet 
running state version: 25.2.1
initial disk state version: 25.2.0                                                                                                                                                                         
features: default,getblocktemplate_rpcs,metrics_exporter_prometheus,prometheus,release_max_level_info 
branch: upgrade-subtrees-backwards
git commit: 8e74e49
```

### Zebra Version

zebrad 1.2.0 commit 8e74e49

### Which operating systems does the issue happen on?

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

local machine

### Additional information

This bug would have been caught by ticket https://github.com/ZcashFoundation/zebra/issues/7446, but only if we found a test vector for "Completed Subtree at the end of the block", and "subtrees added with new tip blocks", and ran it on a full sync Zebra state.

The best way to detect this bug is a manual `zcash-rpc-diff` after a full sync of a new state, I'll add a check to ticket #7450.

[...]

PR #7555, opened 2023-09-13
fix(ci): Temporarily disable panic on subtree validation failure

teor2345:
## Motivation

This PR disabled subtree validation panics until we fix subtree bug #7532.

## Solution

Replace a panic with an error

## Review

This is an urgent fix that is blocking unrelated PRs from merging.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
@@ -183,7 +183,8 @@ pub fn check(db: &ZebraDb) {
     let check_sapling_subtrees = check_sapling_subtrees(db);
     let check_orchard_subtrees = check_orchard_subtrees(db);
     if !check_sapling_subtrees || !check_orchard_subtrees {
-        panic!("missing or bad subtree(s)");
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
This looks like at least something we will have to do now while we figure out the root cause.

teor2345:
This test is failing because the latest image is invalid in another way - it finished at 2 million blocks, rather than 2.2 million. 

This is either a bug in the self-hosted runner CI workflow (which might have been fixed since), or an unrelated bug in Zebra. The simplest way to fix it is to delete the image. (We don't need it any more, because we can reproduce the bug.)

Here's how I searched for the latest image:
https://console.cloud.google.com/compute/images?hl=en&project=zfnd-dev-zebra&tab=images&pageState=(%22images%22:(%22f%22:%22%255B%257B_22k_22_3A_22Created%2520by_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22zfnd_5C_22_22_2C_22i_22_3A_22creator_22%257D%255D%22,%22s%22:%5B(%22i%22:%22creationTimestamp%22,%22s%22:%221%22),(%22i%22:%22type%22,%22s%22:%220%22),(%22i%22:%22name%22,%22s%22:%220%22)%5D))

PR #7562, opened 2023-09-14
fix(state): Compare subtrees in non-finalized chains when evaluating chain equality

### LINKED ISSUE #7447: Compare subtrees in non-finalized chains when evaluating chain equality
## Motivation

We added roots of note commitment subtrees into the non-finalized chain in PR #7334, but we forgot to take them into account when evaluating equality of chains in [`eq_internal_state`](https://github.com/zcashfoundation/zebra/blob/6f503049c631804aca7d99615becb8be5cc1bd7a/zebra-state/src/service/non_finalized_state/chain.rs#L263).

## Solution

Compare the subtree roots in [`eq_internal_state`](https://github.com/zcashfoundation/zebra/blob/6f503049c631804aca7d99615becb8be5cc1bd7a/zebra-state/src/service/non_finalized_state/chain.rs#L263).

Adding the comparison will likely cause some tests to fail. We'll need to adjust those as well.

upbqdn:
## Motivation

Close #7447.

## Solution

- Compare the subtrees in [eq_internal_state](https://github.com/zcashfoundation/zebra/blob/6f503049c631804aca7d99615becb8be5cc1bd7a/zebra-state/src/service/non_finalized_state/chain.rs#L263).
- This fix turned out to be trivial since none of our tests create subtrees in non-finalized chains.
- I thought of moving the chain comparison to `Eq`, but we already use `Eq` for another purpose, so I left the comparison as it is.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/non_finalized_state/chain.rs
+++ b/zebra-state/src/service/non_finalized_state/chain.rs
@@ -281,6 +281,10 @@ impl Chain {
             // history trees
             self.history_trees_by_height == other.history_trees_by_height &&
 
+            // note commitment subtrees
+            self.sapling_subtrees == other.sapling_subtrees &&
+            self.orchard_subtrees == other.orchard_subtrees &&
+
             // anchors
             self.sprout_anchors == other.sprout_anchors &&
             self.sprout_anchors_by_height == other.sprout_anchors_by_height &&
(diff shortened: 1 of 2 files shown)


teor2345 (approved):
Thanks for the fix!

I think the ideal fix would be to put the fields in eq_internal_state() in an inner struct, where we derive PartialEq. Then we could just call inner.eq() from there.

Then in the outer struct, we could keep the existing manual implementation of PartialEq, which compares tip hashes. (Because that's what it means for chains to be equal.)

upbqdn:
> I think the ideal fix would be to put the fields in eq_internal_state() in an inner struct, where we derive PartialEq. Then we could just call inner.eq() from there.

This seems quick and beneficial because we won't forget to add new fields as we already did, so I'm going to do that.

Issue #7446, opened 2023-09-01
Add format checks with fixed subtree values from `zcashd`


teor2345:
## Motivation

We are calculating subtree roots, but we don't know if they match `zcashd`.

### Specifications

The RPC format isn't on the RPC website yet, but it is in the `zcashd` PR at https://github.com/zcash/zcash/pull/6677/files#diff-decae4be02fb8a47ab4557fe74a9cb853bdfa3ec0fa1b515c0a1e5de91f4ad0bR1458-R1475

## Suggestion

Add a list of subtree indexes and subtree roots for testing, and check that they match here:
https://github.com/ZcashFoundation/zebra/pull/7437/files#diff-55a5fe0b23f2b84d8ef7fbd30e96d1abade1bfe4980ddf5a61abd8065da18012R209-R214

### Testing Matrix

We need to test the 8 combinations of these different code paths:
* Sapling & Orchard
* Completed Subtree at the end of the block & Incomplete subtree at the end of the block
* Subtrees upgraded from existing blocks & subtrees added with new tip blocks (one way to do this is to test indexes near zero, and indexes near the tip)

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7446?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

oxarbitrage:
Do we know what argument combination we should use to generate this test vectors:

> - Completed Subtree at the end of the block & Incomplete subtree at the end of the block
> - Subtrees upgraded from existing blocks & subtrees added with new tip blocks (one way to do this is to test indexes near zero, and indexes near the tip)

teor2345:
> Do we know what argument combination we should use to generate this test vectors:

I don't know, sorry. Finding them is part of the work in this ticket.

The upgrade code changes in PR #7531 will make it a lot easier to search the entire database for specific test vectors.  I can help with that after I get that PR working. But it might not be until later this week, because I have to do ZIP editor work today.

oxarbitrage:
That is ok, i can do the searching. What is not clear to me is what i am looking for.

teor2345:
You could log the subtree pool and index when Zebra runs the 8 different kinds of subtree code, then use the RPC to get the test vector from `zcashd`:
```sh
[...]

PR #7566, opened 2023-09-15
fix(state): Use correct end heights for end of block subtrees during the full sync

### LINKED ISSUE #7532: bug: some subtrees have incorrect end heights after a full sync

teor2345:
## Reviewer Questions

Does this PR have any data bugs?
Are there any other blocking issues? (For example, security bugs.)

Since this PR is on the critical path, I'd like other suggestions to go in other PRs or tickets.

## Motivation

When a subtree is created during the full sync, and the final note commitment is at the end of a block, the end height is one higher than it should be.

Close #7532

### Specifications

`zcashd` uses the height of the block the subtree ends in:
https://github.com/zcash/zcash/pull/6694

Because this bug fix is complex, I added some extra tests, and bumped the state version so the whole database will be upgraded again. I also cleared any data from the previous upgrade. This will make sure all our tests run on the new data.

## Solution

Refactor the subtree upgrade so it uses block validation code that has already been tested
Bump the state version and delete previous data, so the tests run on the data from this upgrade only

Use correct heights for subtrees completed at the end of a block. Previously, these subtrees had a height that was 1 higher than the block their final note was in. This bug only happened for subtrees created by new blocks. (And not for upgraded blocks.)

This bug would have been caught by ticket #7446, but only if we found a test vector for "Completed Subtree at the end of the block", and "subtrees added with new tip blocks", and ran it on a full sync Zebra state.

I refactored the upgrade and tip block subtrees to use the same code, which is how I found this bug. So a similar bug shouldn't happen again.

### Related Changes

Wait for the state upgrade in tests:
- before subtree gRPC tests
- before writing updated cached states (full syncs are ok, so are tests that don't write states)

Refactor the subtree upgrade so the upgrade can be done in reverse order, but don't reverse the order yet.

Fix some clippy lints.

### Testing

Add extra state validity tests, including a quick check before the upgrade runs for 20 minutes. The quick check runs on every Zebra launch, even in production.

## Review

This is a routine change, but it is on the critical path for getting the RPCs finished and tested. It is also blocking all the rest of the test tickets.

I'd like non-blocking suggestions to go in other PRs or tickets.

### Reviewer Checklist


zebrad/tests/common/lightwalletd/wallet_grpc_test.rs
```diff
@@ -339,6 +352,59 @@ pub async fn run() -> Result<()> {
...
+    if state_format_change.contains("launching upgrade task") {
+        let latest_version_for_adding_subtrees = latest_version_for_adding_subtrees();
+        tracing::info!(
+            ?test_type,
+            ?zebra_rpc_address,
+            %state_format_change,
+            %latest_version_for_adding_subtrees,
+            "waiting for zebrad subtree state upgrade..."
+        );
+        zebrad.expect_stdout_line_matches(&format!("marked database format as upgraded.*format_upgrade_version.*=.*{latest_version_for_adding_subtrees}"))?;
```
teor2345:
```suggestion
        //
        // TODO: this line will hang if the state upgrade finishes before the subtree tests start.
        // But that is unlikely with the 25.2 upgrade, because it takes 20+ minutes.
        // If it happens for a later upgrade, this code can be removed, as long as all the
        // lightwalletd cached states are version 25.2.2 or later.
        zebrad.expect_stdout_line_matches(&format!("marked database format as upgraded.*format_upgrade_version.*=.*{latest_version_for_adding_subtrees}"))?;
```

teor2345:
> Error: Error: API rate limit exceeded for installation ID 1523599. If you reach out to GitHub Support for help, please include the request ID B04D:8776:1FF995:41593F:65040607.

https://github.com/ZcashFoundation/zebra/actions/runs/6195144535/job/16819307823?pr=7566#step:4:13

The tests should succeed if you re-run them in an hour or so.

zebra-chain/src/orchard/tree.rs
```diff
@@ -394,6 +394,64 @@ impl NoteCommitmentTree {
...
+
+        Some(tree.position().into())
+    }
+
+    /// Returns true if this tree has at least one new subtree, when compared with `prev_tree`.
+    pub fn contains_new_subtree(&self, prev_tree: &Self) -> bool {
+        // A new subtree was completed by the last note commitment in this tree.
+        if self.is_complete_subtree() {
+            return true;
+        }
```
upbqdn: This makes the function return `true` even when `self` equals `prev_tree` or the index of `self` is not greater than the index of `prev_tree`.

zebra-chain/src/orchard/tree.rs
```diff
@@ -394,6 +394,64 @@ impl NoteCommitmentTree {
...
+
+        // Check for spurious index differences.
+        //
+        // There is one new subtree somewhere in the trees. It is either:
+        // - a new subtree at the end of the previous tree, or
+        // - a new subtree in this tree (but not at the end).
+        //
+        // This happens because the subtree index only increases when the first note is added to
```
upbqdn: ###### Doc nit (optional)

It's not clear to me what "_This_" refers to. I assume it's the first item in the list above?

zebra-chain/src/sapling/tree.rs
```diff
@@ -404,6 +462,48 @@ impl NoteCommitmentTree {
...
+            // If the subtree has no nodes, the remaining number of nodes is the number of nodes in
+            // a subtree.
+            None => {
+                // This position is guaranteed to be in the first subtree.
+                let first_position = 0.into();
+
+                let subtree_address = incrementalmerkletree::Address::above_position(
+                    TRACKED_SUBTREE_HEIGHT.into(),
+                    first_position,
+                );
```
upbqdn: ###### Optional Simplifiaction

```suggestion
                let subtree_address = incrementalmerkletree::Address::above_position(
                    TRACKED_SUBTREE_HEIGHT.into(),
                    // This position is guaranteed to be in the first subtree.
                    0.into(),
                );
```

zebra-chain/src/orchard/tree.rs
```diff
@@ -423,6 +481,48 @@ impl NoteCommitmentTree {
...
+            // If the subtree has no nodes, the remaining number of nodes is the number of nodes in
+            // a subtree.
+            None => {
+                // This position is guaranteed to be in the first subtree.
+                let first_position = 0.into();
+
+                let subtree_address = incrementalmerkletree::Address::above_position(
+                    TRACKED_SUBTREE_HEIGHT.into(),
+                    first_position,
+                );
```
upbqdn: ###### Optional Simplifiaction

```suggestion
                let subtree_address = incrementalmerkletree::Address::above_position(
                    TRACKED_SUBTREE_HEIGHT.into(),
                    // This position is guaranteed to be in the first subtree.
                    0.into(),
                );
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -14,161 +21,264 @@ use crate::service::finalized_state::{
 
 /// Runs disk format upgrade for adding Sapling and Orchard note commitment subtrees to database.
 ///
+/// Trees are added to the database in reverse height order, so that wallets can sync correctly
+/// while the upgrade is running.
+///
```
upbqdn: ###### Note (optional)

This doesn't hold for this PR, but I guess we can leave it here.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -387,46 +496,294 @@ fn check_orchard_subtrees(db: &ZebraDb) -> bool {
...
+#[must_use = "subtree should be written to the database after it is calculated"]
+#[instrument(skip(read_db, prev_tree, tree))]
+fn calculate_sapling_subtree(
+    read_db: &ZebraDb,
+    prev_end_height: Height,
+    prev_tree: Arc<sapling::tree::NoteCommitmentTree>,
+    end_height: Height,
+    tree: Arc<sapling::tree::NoteCommitmentTree>,
+) -> NoteCommitmentSubtree<sapling::tree::Node> {
+    // If a subtree is completed by a note commentment in the block at `end_height`,
```
upbqdn: ###### Optional Suggestion

```suggestion
    // If a subtree is completed by a note commitment in the block at `end_height`,
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -387,46 +496,294 @@ fn check_orchard_subtrees(db: &ZebraDb) -> bool {
...
+#[must_use = "subtree should be written to the database after it is calculated"]
+#[instrument(skip(read_db, prev_tree, tree))]
+fn calculate_orchard_subtree(
+    read_db: &ZebraDb,
+    prev_end_height: Height,
+    prev_tree: Arc<orchard::tree::NoteCommitmentTree>,
+    end_height: Height,
+    tree: Arc<orchard::tree::NoteCommitmentTree>,
+) -> NoteCommitmentSubtree<orchard::tree::Node> {
+    // If a subtree is completed by a note commentment in the block at `end_height`,
```
upbqdn: ###### Optional Suggestion

```suggestion
    // If a subtree is completed by a note commitment in the block at `end_height`,
```

teor2345:
## Questions for reviewers

Does this PR have any data bugs?
Are there any other blocking issues? (For example, security bugs.)

Since this PR is on the critical path, I'd like other suggestions to go in other PRs or tickets.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -180,76 +290,78 @@ pub fn run(
...
+                result = Err("completed subtree roots should match");
+                error!(?result);
             }
         }
         // Check that the final note has a greater subtree index if it didn't complete a subtree.
         else {
             let Some(prev_tree) = db.sapling_tree_by_height(&subtree.end.previous()) else {
-                error!(?subtree.end, "missing note commitment tree at subtree completion height");
-                is_valid = false;
+                result = Err("missing note commitment tree at subtree completion height");
```
upbqdn: ###### Optional Suggestion

```suggestion
                result = Err("missing note commitment tree below subtree completion height");
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -272,99 +384,96 @@ fn check_sapling_subtrees(db: &ZebraDb) -> bool {
...
+                result = Err("completed subtree roots should match");
+                error!(?result);
             }
         }
         // Check that the final note has a greater subtree index if it didn't complete a subtree.
         else {
             let Some(prev_tree) = db.orchard_tree_by_height(&subtree.end.previous()) else {
-                error!(?subtree.end, "missing note commitment tree at subtree completion height");
-                is_valid = false;
+                result = Err("missing note commitment tree at subtree completion height");
```
upbqdn: ###### Optional Suggestion

```suggestion
                result = Err("missing note commitment tree below subtree completion height");
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -374,7 +483,7 @@ fn check_orchard_subtrees(db: &ZebraDb) -> bool {
         .orchard_tree_by_height_range(..)
         // Exclude empty orchard tree and add subtree indexes
         .filter_map(|(height, tree)| Some((tree.subtree_index()?, height, tree)))
-        // Exclude heights that don't complete a subtree and count completed subtree
+        // Exclude heights that don't complete a subtree and count completed subtrees
         .filter_map(|(subtree_index, height, tree)| {
             if tree.is_complete_subtree() || subtree_index.0 > subtree_count {
```
upbqdn, line 491: ###### Optional

I just noticed this filter skips the tree with subtree index 0 if that tree is not a complete subtree. It might be worth mentioning that in a comment.

upbqdn (dismissed):

upbqdn:
I just reviewed this PR. Feel free to close all comments labeled as optional right away without addressing them.

zebra-chain/src/orchard/tree.rs
teor2345:
```suggestion
        // Spurious index differences happen because the subtree index only increases when the first note is added to
```

zebra-chain/src/orchard/tree.rs
teor2345: Ah, good catch! ~~I think a way to fix this bug is to make `self.is_complete_subtree()` the last line of this method.~~

This bug doesn't have any practical impact yet, but it's a trap for future uses of this method.

zebra-chain/src/sapling/tree.rs
teor2345: This will be fixed when I copy-paste-replace the orchard code:
https://github.com/ZcashFoundation/zebra/pull/7566/files/82bf50a37f5da716957e54fa85c205eee555869e#r1328982047

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: Oops, I thought I'd reverted all that code, but I must have missed this comment.. Happy to fix this since I have to edit the file anywa.

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: This will get fixed when I copy-paste-replace the sapling comments:
https://github.com/ZcashFoundation/zebra/pull/7566#discussion_r1329046372

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
teor2345: I think I might leave this comment for @arya2 to clarify, because he wrote this code.

teor2345:
My first fix in [5a1deb1](https://github.com/ZcashFoundation/zebra/pull/7566/commits/5a1deb1865071977df36dcb10e64c9e328f8c322) was wrong, but I think I fixed it in [41eef09](https://github.com/ZcashFoundation/zebra/pull/7566/commits/41eef096793dca9ec4d94318f117ec2cae675ad5).

There was also a bug in the quick check which skipped the same subtrees, which is why it didn't detect this issue on my local machine.

Edit: ok let's try again with [54f2bba](https://github.com/ZcashFoundation/zebra/pull/7566/commits/54f2bba601b55575b5748f2636ab5aafd54ca433). Good news is that the quick check detected that bug.

oxarbitrage (approved):
approving without reviewing so it can be merged fast, @upbqdn approved previously.

zebra-chain/src/orchard/tree.rs
```diff
@@ -394,6 +394,72 @@ impl NoteCommitmentTree {
...
+        //
+        // Spurious index differences happen because the subtree index only increases when the
+        // first note is added to the new subtree. So we need to exclude subtrees completed by the
+        // last note commitment in the previous tree.
+        //
+        // We also need to exclude empty previous subtrees, because the index changes to zero when
+        // the first note is added, but a subtree wasn't completed.
+        if prev_tree.is_complete_subtree() || prev_index == -1 {
+            return false;
+        }
```
upbqdn, line 457: I noticed two issues:

1. The method returns `false` here when `prev_tree` is empty, so `prev_index` is -1, but `self` is a complete subtree with index 0. I think the returned value should be `true` for this case. This case was covered correctly before the refactor.

2. The method returns `false` here when `index_difference` is 1, and both `prev_tree` and `self` are complete subtrees. This case was covered correctly before the refactor.

I think we need to refactor this method again.

zebra-chain/src/orchard/tree.rs
```diff
@@ -394,6 +394,72 @@ impl NoteCommitmentTree {
...
+        // If the indexes are equal, there can only be a new subtree if `self` just completed it.
+        if index == prev_index {
+            return self.is_complete_subtree();
+        }
+
+        // If `self` is the next index, check for spurious index differences.
+        //
+        // There is one new subtree somewhere in the trees. It is either:
+        // - a new subtree at the end of the previous tree, or
+        // - a new subtree in this tree (but not at the end).
```
upbqdn, line 447: We don't know here that the subtree is not at the end because we don't check the output of `self.is_complete_subtree()` after the refactor.

zebra-chain/src/orchard/tree.rs
```diff
@@ -394,6 +394,72 @@ impl NoteCommitmentTree {
...
+        // There is at least one new subtree, even if there is a spurious index difference.
+        if index_difference > 1 {
+            return true;
+        }
+
+        // If the indexes are equal, there can only be a new subtree if `self` just completed it.
+        if index == prev_index {
+            return self.is_complete_subtree();
+        }
+
```
teor2345, line 442: I think both issues can be resolved by adding this code here:

```rust
// If `self` has the next index, check if it completed a subtree.
if self.is_complete_subtree() {
    return true;
}
```

zebra-chain/src/orchard/tree.rs
teor2345: See my suggestion above.

zebra-chain/src/orchard/tree.rs
teor2345: See my suggestion above, which covers both cases.

zebra-chain/src/orchard/tree.rs
teor2345: I can do this change in another PR.

upbqdn: Looks good. I also think the suggested solution will resolve both issues.

PR #7580, opened 2023-09-18
fix(ci): handle `tee` trying to write to a closed pipe

### LINKED ISSUE #7564: devops: CI failure in `tee` when Rust tests pass
### Describe the issue or request

It seems this is a new issue in the CI. I am not sure how often it is happening, i only saw it once by now but it might worth to have a tciket in case it gets repeated.

A pull request failed with 

```
tee: 'standard output': Broken pipe
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 44 filtered out; finished in 2.11s
sudo docker exit status: 101
Error: Process completed with exit code 101.
```

at https://github.com/ZcashFoundation/zebra/actions/runs/6165092427/job/16733752210?pr=7515

Apparently, "The Rust test passes, but Docker returns an error status. This is probably an issue with the self-hosted runner changes." (that was mentioned by @teor2345 in slack).

### Expected Behavior

No failure or more details on what is the failure inside docker.

### Current Behavior

Failure with `sudo docker exit status: 101`

### Possible Solution

We need to figure out if this was a temporal issue, we might have to do nothing to resolve the ticket other than just a short comment of what went wrong, provide what to do if it happens again.

### Additional Information/Context

It looks like the failure is unrelated to what the pull request was doing.

### Is this happening on PRs?

yes

### Is this happening on the main branch?

not seen yet

gustavovalverde:
## Motivation
This issue happens even if there's not an actual failure in integration tests.

```
tee: 'standard output': Broken pipe
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 44 filtered out; finished in 2.11s
sudo docker exit status: 101
Error: Process completed with exit code 101.
```

Fixes #7564

`trap '' PIPE` is a not a very intuitive command.

## Solution

Use `trap '' PIPE` as a command in the pipeline might exit early and we want the script to continue executing, as using `grep --max-count=1` might exit after finding the first match, causing a broken pipe error with `tee`. This is being combined with  `set -o pipefail` to have strict error handling, as we want the script to fail if any command in the pipeline fails.

## Review

We should confirm this part of CI is being tested and passes as expected.

### Reviewer Checklist

## Follow Up Work

N/A

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -592,6 +592,9 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command=' \
           set -e;
+          set -o pipefail;
+          trap '' PIPE;
+
           sudo docker logs \
           --tail all \
           --follow \
@@ -599,11 +602,13 @@ jobs:
           tee --output-error=exit /dev/stderr | \
           grep --max-count=1 --extended-regexp --color=always \
           "test result: .*ok.* [1-9][0-9]* passed.*finished in"; \
+
           EXIT_STATUS=$( \
           sudo docker wait ${{ inputs.test_id }} || \
           sudo docker inspect --format "{{.State.ExitCode}}" ${{ inputs.test_id }} || \
           echo "missing container, or missing exit status for container" \
           ); \
+
           echo "sudo docker exit status: $EXIT_STATUS"; \
           exit "$EXIT_STATUS" \
           '


teor2345:
> This is being combined with `set -o pipefail` to have strict error handling, as we want the script to fail if any command in the pipeline fails.

This is just for interest, I want to avoid any other changes in a bug fix PR:
http://redsymbol.net/articles/unofficial-bash-strict-mode/

Other possible settings are:
- `set -u`: fail if variables are unset
- `set -x`: show all the code that is executed (GitHub already does this, but only for top-level commands), this will also interfere with the exit status check

teor2345:
We just agreed that I wouldn't do DevOps reviews any more 😅

arya2 (approved):
Looks good!

PR #7591, opened 2023-09-20
fix(chain): Return errors instead of panicking in methods for `Height`s

### LINKED ISSUE #7263: Return errors instead of panicking in methods for  `Height`s
### Motivation

Resolve the issues raised in https://github.com/ZcashFoundation/zebra/pull/7239#discussion_r1266136500.

upbqdn:
## Motivation

Close #7263.

## Solution

- Don't panic in:
	- `Height::next()`, and
	- `Height::previous()`.

And instead:

- Create a new error type `HeightError`.
- Return the new error type wrapped in `Result`.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/height.rs
+++ b/zebra-chain/src/block/height.rs
@@ -1,6 +1,7 @@
 //! Block height.
 
 use std::ops::{Add, Sub};
+use thiserror::Error;
 
 use crate::{serialization::SerializationError, BoxError};
 
@@ -23,6 +24,14 @@ pub mod json_conversion;
 /// `ZcashSerialize` or `ZcashDeserialize` for `Height`.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct Height(pub u32);
+
+#[derive(Error, Debug)]
+pub enum HeightError {
+    #[error("The resulting height would ovferflow Height::MAX.")]
+    Overflow,
+    #[error("The resulting height would underflow Height::MIN.")]
+    Underflow,
+}
 
 impl std::str::FromStr for Height {
     type Err = SerializationError;
@@ -72,8 +81,8 @@ impl Height {
     ///
     /// - If the current height is at its maximum.
     // TODO Return an error instead of panicking #7263.
-    pub fn next(self) -> Self {
-        (self + 1).expect("Height should not be at its maximum.")
+    pub fn next(self) -> Result<Self, HeightError> {
+        (self + 1).ok_or(HeightError::Overflow)
     }
 
     /// Returns the previous [`Height`].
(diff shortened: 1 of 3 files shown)


zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -556,7 +556,7 @@ impl Chain {
         // Don't add a new tree unless it differs from the previous one or there's no previous tree.
         if height.is_min()
             || self
-                .sprout_tree(height.previous().into())
+                .sprout_tree(height.previous().unwrap().into())
```
upbqdn: It's OK to use `unwrap()` here because `previous()` gets executed only when `height.is_min()` returns `false` at line 557. Likewise for all the other `unwrap()`s in this PR.

zebra-chain/src/block/height.rs
```diff
@@ -24,6 +25,14 @@ pub mod json_conversion;
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct Height(pub u32);
 
+#[derive(Error, Debug)]
+pub enum HeightError {
+    #[error("The resulting height would ovferflow Height::MAX.")]
```
teor2345:
```suggestion
    #[error("The resulting height would overflow Height::MAX.")]
```

zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs
```diff
@@ -353,7 +353,12 @@ fn check_sapling_subtrees(db: &ZebraDb) -> Result<(), &'static str> {
         }
         // Check that the final note has a greater subtree index if it didn't complete a subtree.
         else {
-            let Some(prev_tree) = db.sapling_tree_by_height(&subtree.end.previous()) else {
+            let prev_height = subtree
+                .end
+                .previous()
+                .expect("Height should not underflow here.");
```
teor2345: Usually we put a reason in the `expect()` string, so it's easier to fix bugs.

```suggestion
                .expect("underflow is impossible: Sapling activation is a long way from genesis");
```

teor2345: This applies to all the `expect()` and `unwrap()` in this PR.

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: Usually in this situation we would say `expect("already checked for underflow")` so if we're wrong it's easier to diagnose the bug.

teor2345 (commented):
Would you mind putting a reason in every `expect()`, and turning every `unwrap()` into an `expect()` with a reason?

That way the PR will be easier to review, and if the `expect()` fails we'll be able to fix it.

zebra-chain/src/block/height.rs
upbqdn: This was a handy typo because it made the length of the line match the one below for underflows. Goodbye, typo.

upbqdn:
Thanks for noting that I should improve the error messages. It didn't occur to me that I didn't actually explain why we can always expect an `Ok`.

I generally try to follow the "expect as precondition" style: https://doc.rust-lang.org/std/error/index.html#common-message-styles.

zebra-state/src/service/non_finalized_state/chain.rs
upbqdn: Done in [bad9fbb](https://github.com/ZcashFoundation/zebra/pull/7591/commits/bad9fbbaf644248ba687a4a5f7da816c5abbfa57).

teor2345 (approved):
Thanks!

PR #7602, opened 2023-09-21
change(state): Check database format is valid every 5 minutes, to catch format errors in new block code

### LINKED ISSUE #7570: Test that "sync from empty" and "upgrade from existing" state code produces the same data
### Motivation

We want to prevent bugs like #7532 happening again. Expecting developers to run manual tests on all the different ways of creating a state isn't reliable.

#### Run subtree tests on both kinds of states

One option is to run the subtree tests in tickets #7450, #7451, and #7446 on both a "sync from new" state and an "upgrade from existing" state. But to do that, we'd need to:
1. change the state selection script so it accepts a parameter for new/upgraded state kinds
2. change the state deletion script so it always keeps both state kinds
3. Run all those tests twice
4. Remember to do this for every future state upgrade

#### Re-run state validity checks while Zebra is running

Another option is to repeat the state validity check in PR #7566:
- [x] periodically while Zebra is running
- [x] before Zebra shuts down (or any time after the full sync finishes)
- [ ] fail CI if the format is wrong

Then that check will run:
- on upgraded states: PRs and main branch merges (this already happens once),
- on "sync from empty" states: the weekly full sync (this is the missing check),
- before we write the cached state in CI.

Then as soon as we merge PR #7565 we'll automatically get a ticket when these checks fail on `main` merges or scheduled full syncs.

This seems like the more reliable option, because it will automatically test all state upgrade code, not just subtrees. 

Now we have two different code paths for creating state data, we need to check that they create exactly the same data.

But unfortunately, we can't run a full and an upgrade sync on each PR, because that would take 3 days. So this ticket is about finding another way to reduce the risk.

### Testing

This is a testing ticket.

teor2345:
## Motivation

We want to catch format errors in new block code. But currently that code is only tested when Zebra adds new blocks, then shuts down, then opens the same state. So PRs can merge with bugs in that code.

Instead, we check the format is valid every 5 minutes while Zebra is running.

This is the periodic part of #7570.

We can't run this code too often, or it will impact performance. But we want to make sure it gets run during most tests, so we catch bugs early.

## Solution

Continue running the format change thread after the initial change finishes, then periodically run a format check:
- Turn the format change task into a continuous loop
- Move the max height check into the format checks
- Make slower format checks cancellable on shutdown

Related changes:
- Refactor the format change/check code so that "no change" has an enum variant
- Add a variant for checking the format of new blocks
- Refactor the format checks so they can be called using a single method
- Make all the format checks run completely, and log their errors, then panic
- Time format upgrades and format checks
- Store the format change flag in the database struct for debugging

### Example Logs

On startup, it takes 2 seconds to open the database, and 7 seconds to check the format is valid:
```
2023-09-21T02:45:10.692975Z  INFO zebra_state::service::finalized_state::disk_format::upgrade: trying to open newer database format: data should be compatible running_version=25.2.1 disk_version=25.2.2
2023-09-21T02:45:12.388329Z  INFO zebra_state::service::finalized_state::disk_format::upgrade: marked database format as downgraded running_version=25.2.1 disk_version=25.2.2 
2023-09-21T02:45:19.485921Z  INFO zebra_state::service::finalized_state::disk_format::upgrade: database format is valid running_version=25.2.1 inital_disk_version=25.2.2
```

So the format check is running approximately 2% of the time. Since it runs in parallel with the rest of the code, and it will be faster when the data is in the database memory cache, this is acceptable performace.

## Review

This is on the critical path for finishing off the subtree work. It's not strictly required for the 1.3.0 release, but it would help to find any other data bugs.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -1,7 +1,8 @@
-//! In-place format upgrades for the Zebra state database.
+//! In-place format upgrades and format validity checks for the Zebra state database.
 
 use std::{
     cmp::Ordering,
+    convert::Infallible,
     sync::{mpsc, Arc},
     thread::{self, JoinHandle},
 };
@@ -22,7 +23,9 @@ use DbFormatChange::*;
 
 use crate::{
     config::write_database_format_version_to_disk,
-    constants::{latest_version_for_adding_subtrees, DATABASE_FORMAT_VERSION},
+    constants::{
+        latest_version_for_adding_subtrees, DATABASE_FORMAT_CHECK_INTERVAL, DATABASE_FORMAT_VERSION,
+    },
     database_format_version_in_code, database_format_version_on_disk,
     service::finalized_state::{DiskWriteBatch, ZebraDb},
     Config,
@@ -30,30 +33,11 @@ use crate::{
 
 pub(crate) mod add_subtrees;
 
-/// The kind of database format change we're performing.
+/// The kind of database format change or validity check we're performing.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum DbFormatChange {
-    /// Mark the format as newly created by `running_version`.
-    ///
-    /// Newly created databases are opened with no disk version.
-    /// It is set to the running version by the format change code.
-    NewlyCreated { running_version: Version },
-
-    /// Check that the database from a previous instance has the current `running_version` format.
-    ///
-    /// Current version databases have a disk version that matches the running version.
-    /// No upgrades are needed, so we just run a format check on the database.
-    /// The data in that database was created or updated by a previous Zebra instance.
-    OpenCurrent { running_version: Version },
-
-    /// Check that the database from this instance has the current `running_version` format.
-    ///
-    /// The data in that database was created or updated by the currently running Zebra instance.
-    /// So we periodically check for data bugs, which can happen if the upgrade and new block
-    /// code produce different data. (They can also be caused by disk corruption.)
-    #[allow(dead_code)]
-    CheckNewBlocksCurrent { running_version: Version },
-
+    // Data Format Changes
+    //
     /// Upgrade the format from `older_disk_version` to `newer_running_version`.
     ///
     /// Until this upgrade is complete, the format is a mixture of both versions.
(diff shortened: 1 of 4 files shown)


mpguerra:
Does this run on CI only or on any running zebra instance?

teor2345:
> Does this run on CI only or on any running zebra instance?

This runs on all Zebra instances, because if the database format is wrong, then we can break a consensus rule, or send incorrect data to wallets.

(We already run these checks at startup, but that's not enough to detect all format bugs.)

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -81,52 +91,44 @@ impl ZebraDb {
...
-            // Do the quick checks first, then the slower checks.
-            upgrade::add_subtrees::quick_check(&db);
-
-            DbFormatChange::check_for_duplicate_trees(db.clone());
-            upgrade::add_subtrees::check(&db);
-        }
-
-        db
+        // We have to get this height before we spawn the upgrade task, because threads can take
+        // a while to start, and new blocks can be committed as soon as we return from this method.
```
upbqdn, line 112: What would be the problem if we got the height after committing new blocks?

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -62,39 +89,39 @@ pub enum DbFormatChange {
...
-    update_task: Option<Arc<JoinHandle<Result<(), CancelFormatChange>>>>,
+    update_task: Option<Arc<JoinHandle<Result<Infallible, CancelFormatChange>>>>,
 
-    /// A channel that tells the running format thread to finish early.
     /// A channel that tells the running format thread to finish early.
     cancel_handle: mpsc::SyncSender<CancelFormatChange>,
 }
 
-/// Marker for cancelling a format upgrade.
+/// Marker type that is sent to cancel a format upgrade, and returned as a error on cancellation.
```
upbqdn, line 107: ###### Optional Suggestion
```suggestion
/// Marker type that is sent to cancel a format upgrade, and returned as an error on cancellation.
```

upbqdn (approved):

zebra-state/src/service/finalized_state/zebra_db.rs
teor2345: It think it depends on exactly how the upgrade code is written.

For deduplicating trees if a duplicate tree for the new block wasn't written, then it won't appear in the iterator, and nothing happens. If it's a unique block then it will appear in the iterator and it will get kept.

For the subtrees it might overwrite the subtree from a new block with the same subtree data from the upgrade. Which is completely ok.

A partial upgrade is very similar to adding new blocks using the latest state code. In both cases the database is a mix of versions. So it should usually work, but we don't test it very well, because it's an edge case.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
teor2345: I'm going to fix this in the next PR instead.

Issue #7463, opened 2023-09-04
Rename subtree methods and types to "subtree root"


teor2345:
## Schedule

This cleanup should happen last, because it will conflict with almost every other subtree change.

## Motivation

We are storing subtree roots and end heights in the subtree indexes, not the subtree trees themselves.

This looks like the sort of information that gets lost and causes misunderstandings in future :) Can we rename the methods and update comments to correctly reflect what we are storing? Since we want this PR to merge soon we should do it in a follow up PR

_Originally posted by @mpguerra in https://github.com/ZcashFoundation/zebra/pull/7437#discussion_r1314853877_
            

### Suggested Changes

Open 2 PRs, one for mass renames, and one for comments.

#### Mass Renames

- Rename types
- Rename methods and their arguments

Make a list of renames and then do a mass rename:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/mass-renames.md#universal-renames-with-sed

Don't rename the database column families, because that requires a state upgrade, and they are technically correct (every value is an index to a subtree):
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#current-state-database-format

#### Comments

Update the comments of each subtree type and method to make it clear that the subtree indexes contain subtree heights and roots.

teor2345:
@mpguerra this might need to move to the next sprint again due to more important tickets and open PRs.

PR #7608, opened 2023-09-22
change(state): Put chain data into an inner struct

upbqdn:
## Motivation

The non-finalized [`Chain`](https://github.com/zcashfoundation/zebra/blob/7348d080d6dd0bb2e7153569ce4158d86861524f/zebra-state/src/service/non_finalized_state/chain.rs#L43) contains many fields that we need to manually compare in [`eq_internal_state`](https://github.com/zcashfoundation/zebra/blob/7348d080d6dd0bb2e7153569ce4158d86861524f/zebra-state/src/service/non_finalized_state/chain.rs#L265) when comparing two chains. We can't use regular `PartialEq` because we already use that for comparing only the partial cumulative work between chains: https://github.com/zcashfoundation/zebra/blob/7348d080d6dd0bb2e7153569ce4158d86861524f/zebra-state/src/service/non_finalized_state/chain.rs#L2189.

We recently added new fields into `Chain`, and we forgot to compare them in `eq_internal_state`, which was a bug that is likely to happen again.

## Solution

- Put the fields used in `eq_internal_state` into an inner struct, and derive and use `PartialEq` on this struct in `eq_internal_state`.

### Reviewer Checklist


upbqdn:
I labeled this PR as a draft because wrapping the fields of `Chain` into an inner struct causes ~ 290 errors. I would like approval of the current code before I fix all the spots where we access the fields. The fixes will be trivial but tedious, so I would like to avoid them if they wouldn't be right anyway.

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -48,6 +48,29 @@ pub struct Chain {
...
+    pub(super) last_fork_height: Option<Height>,
+    // # Note
+    //
+    // Most diagnostics are implemented on the NonFinalizedState, rather than each chain.
+    // Some diagnostics only use the best chain, and others need to modify the Chain state,
+    // but that's difficult with `Arc<Chain>`s.
+}
+
+/// The internal state of [`Chain`].
+#[derive(PartialEq, Clone, Debug)]
```
teor2345: `Eq` is also guaranteed for this type, as long as all the types are `Eq` (which is checked by the derive).

https://doc.rust-lang.org/std/cmp/trait.Eq.html

```suggestion
#[derive(Clone, Debug, PartialEq, Eq)]
```

teor2345:
> I labeled this PR as a draft because wrapping the fields of `Chain` into an inner struct causes ~ 290 errors. I would like approval of the current code before I fix all the spots where we access the fields. The fixes will be trivial but tedious, so I would like to avoid them if they wouldn't be right anyway.

The code looks fine to me.

If you `impl Deref for Chain { Target: ChainInner ... }` then most of those errors will go away, and we can see if the tests pass. You might also need to `impl DerefMut` for mutable accesses.

https://doc.rust-lang.org/std/ops/trait.Deref.html

Then we can do the mass replace in a later PR, because it's easier to review that way. (And the mass replace is likely to conflict with #7463 in this sprint.)

teor2345 (approved):

teor2345:
Failed due to issue #7659, I'll restart when it's finished.

PR #7610, opened 2023-09-22
chore: Release v1.3.0

upbqdn:
---
name: 'Release Checklist Template'
about: 'Checklist to create and publish a Zebra release'
title: 'Release Zebra (version)'
labels: 'A-release, C-trivial, P-Critical :ambulance:'
assignees: ''

---

# Prepare for the Release

- [x] Make sure the PRs with the new checkpoint hashes and missed dependencies are already merged.
      (See the release ticket checklist for details)

# Summarise Release Changes

These steps can be done a few days before the release, in the same PR:

## Change Log

**Important**: Any merge into `main` deletes any edits to the draft changelog.
Once you are ready to tag a release, copy the draft changelog into `CHANGELOG.md`.

We use [the Release Drafter workflow](https://github.com/marketplace/actions/release-drafter) to automatically create a [draft changelog](https://github.com/ZcashFoundation/zebra/releases). We follow the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format.

To create the final change log:
- [x] Copy the **latest** draft changelog into `CHANGELOG.md` (there can be multiple draft releases)
- [x] Delete any trivial changes
    - [x] Put the list of deleted changelog entries in a PR comment to make reviewing easier
- [x] Combine duplicate changes
- [x] Edit change descriptions so they will make sense to Zebra users
- [x] Check the category for each change
  - Prefer the "Fix" category if you're not sure

## README

README updates can be skipped for urgent releases.

Update the README to:
- [x] Remove any "Known Issues" that have been fixed since the last release.
- [x] Update the "Build and Run Instructions" with any new dependencies.
      Check for changes in the `Dockerfile` since the last tag: `git diff <previous-release-tag> docker/Dockerfile`.
- [x] If Zebra has started using newer Rust language features or standard library APIs, update the known working Rust version in the README, book, and `Cargo.toml`s

You can use a command like:
```sh
fastmod --fixed-strings '1.58' '1.65'
```

## Create the Release PR

- [x] Push the updated changelog and README into a new branch
      for example: `bump-v1.0.0` - this needs to be different to the tag name
- [x] Create a release PR by adding `&template=release-checklist.md` to the comparing url ([Example](https://github.com/ZcashFoundation/zebra/compare/bump-v1.0.0?expand=1&template=release-checklist.md)).
- [x] Freeze the [`batched` queue](https://dashboard.mergify.com/github/ZcashFoundation/repo/zebra/queues) using Mergify.
[...]

PR #7615, opened 2023-09-25
change(ui): Enable the progress bar feature by default, but only show progress bars when the config is enabled

### LINKED ISSUE #7363: Make a config for progress bars, rather than having the feature activate them by default
### Motivation

Currently the `progress-bars` feature turns on progress bars and file logging immediately.

But if we want users to test them, we should make the progress bars configurable with `zebrad.toml`. This allows us to compile the `progress-bar` feature by default, but not activate it for every user by default.

(Activating progress bars by default could be a breaking change for some users that expect Zebra to write logs to standard output.)

### Specifications

We probably want to have this in the tracing config, and then use it to activate the progress bar display loop.

If the display loop isn't active, progress bar operations are ignored, so it's ok for the progress bar update code to always be on. 

We might want to make the config flexible, so that we can have expert and default progress bar configs, once we know which progress bars are a priority for users.

### Testing

We can manually test that progress bars are off by default, off when the feature is active, but on when the feature is active and they are configured on.

### LINKED ISSUE #7364: Make progress-bars a default feature
### Motivation

Once we're happy to support progress bars, we can compile them into all Zebra builds, by adding that feature to the default feature list:
https://github.com/ZcashFoundation/zebra/blob/31426f3b0e38bb681df3b82a38dad07405a876a5/zebrad/Cargo.toml#L53-L55

### Specifications

https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature

This is a well-supported Cargo configuration.

### Testing

We can manually test that progress bars are off by default, off when the feature is active, but on when the feature is active and they are configured on.

### Related Work

Depends on ticket #7363.

teor2345:
## TODO

After this PR merges, add:
- [x] A changelog feature entry asking for user testing
- [ ] A post on the forums encouraging users to test it after it is released, and mentioning [the known issue with estimates](https://github.com/console-rs/indicatif/issues/556)
- [ ] A mention in the Arborist call

## Motivation

Currently the `progress-bars` feature turns on progress bars and file logging when it is enabled.

To make it easier for users to test progress bars, we can make the progress bars configurable with `zebrad.toml`. This allows us to compile the `progress-bar` feature by default, but not activate it for every user by default.

(Activating progress bars by default could be a breaking change for some users that expect Zebra to write logs to standard output.)

Close #7363
Close #7364

When the progress bar feature is activated, the log file also needs to be activated (unless the user has specifically turned it off).

I added a summary and detailed config, because we might want to have an "expert" mode. But we need to know what users want first!

## Solution

- Add a config for progress bars, which is only used when the `progress-bars` feature is enabled
    - If they are not enabled, log a message telling users how to enable them in the config
    - Document how to enable them in the README
- Enable the `progress-bars` feature by default in Zebra builds

### Testing

- Add a config with progress bars enabled to the config tests
- Make the config tests handle log files correctly
- Check the config path in the config tests

## Review

This is a low priority feature, it doesn't need to go in the 1.3.0 release.

### Reviewer Checklist

## Follow Up Work

Work out which progress bars users want.


zebrad/src/components/tracing.rs
```diff
@@ -108,10 +108,16 @@ pub struct Config {
     /// replaced with `.folded` and `.svg` for the respective files.
     pub flamegraph: Option<PathBuf>,
 
+    /// Shows progress bars for block syncing, and mempool transactions, and peer networking.
```
oxarbitrage, line 164:
```suggestion
    /// Shows progress bars for block syncing, mempool transactions, and peer networking.
```

zebrad/src/components/tracing.rs
```diff
@@ -152,25 +158,56 @@ impl Config {
...
+
+    if progress_bar.is_some() {
         let default_log_file = dirs::state_dir()
             .or_else(dirs::data_local_dir)
             .map(|dir| dir.join("zebrad.log"));
 
+        return default_log_file;
+    }
+
+    None
```
arya2: Optional/Nitpick:

```suggestion
    log_file.or(progress_bar.and(
        dirs::state_dir()
            .or_else(dirs::data_local_dir)
            .map(|dir| dir.join("zebrad.log")),
    ))
```

zebrad/src/components/tracing.rs
```diff
@@ -152,25 +158,56 @@ impl Config {
...
             filter: None,
             buffer_limit: 128_000,
             endpoint_addr: None,
             flamegraph: None,
-            #[cfg(not(feature = "progress-bar"))]
-            log_file: None,
-            #[cfg(feature = "progress-bar")]
-            log_file: default_log_file,
+            progress_bar,
+            log_file: default_log_file(None, progress_bar),
```
arya2: Optional/Nitpick: Could we use lazy_static for the path and get rid of `default_log_file()`

```rust
            log_file: progress_bar.and(LOG_FILE_PATH),
```

zebrad/src/components/tracing/component.rs
```diff
@@ -86,7 +86,7 @@ impl Tracing {
         let use_color = config.use_color_stdout();
         let use_color_stderr = config.use_color_stderr();
 
-        let filter = config.filter.unwrap_or_default();
+        let filter = config.filter.clone().unwrap_or_default();
```
arya2, line 89: Could we use `default_log_file()` on line 117:

```
let writer = if let Some(log_file) = config.log_file.as_ref() {
```

arya2 (commented):
This looks great, I think there's a small mistake in the second commit not using the default log file path when none is provided but progress bars are enabled.

`indicatif-log-bridge` looks like a promising alternative if it works with `howudoin`:

> [Integrations](https://github.com/console-rs/indicatif#integrations)
>
> You can use [indicatif-log-bridge](https://crates.io/crates/indicatif-log-bridge) to integrate with the [log crate](https://crates.io/crates/log) and avoid having both fight for your terminal.

oxarbitrage (commented):
This seems to be not working as i will expect. 

I compiled in release mode and then executed the generate command and i get:

```
$ ./target/release/zebrad generate
...
[tracing]
buffer_limit = 128000
force_use_color = false
use_color = true
use_journald = false
```

The same in debug mode:

```
$ ./target/debug/zebrad generate
...
[tracing]
buffer_limit = 128000
force_use_color = false
use_color = true
use_journald = false
```

The `progress_bar` or `log_file` fields are not there.

If i add `progress_bar = "summary"` to the config i also need to add `log_file = "zebra.log"` to make it work properly.

We could add a new `Disabled` member to the `ProgressConfig` enum and make that the default to fix this problem.

Another option is to document very well how do we expect users to use this feature in this stage. I will like to know that too.

zebrad/src/components/tracing.rs
teor2345: I find this harder to read, so I don't think it's worth applying.

zebrad/src/components/tracing.rs
teor2345: We could use a lazy static, but since this code is only run once, it adds more code complexity without much benefit.

I also find option-based logic operations harder to read.

zebrad/src/components/tracing/component.rs
teor2345: I don't have enough context to understand why we would want to make this change. Is it optional, or does it fix a bug?

teor2345:
Thanks for the reviews!

> The `progress_bar` or `log_file` fields are not there.

This is the intended behaviour, because those fields are optional. Fields only get added to the default config if they are required.

The default for `progress_bar` is currently "no progress bars", and the default for `log_file` is "no log file" (log to the terminal). This PR doesn't change that.

> This looks great, I think there's a small mistake in the second commit not using the default log file path when none is provided but progress bars are enabled.

> If i add `progress_bar = "summary"` to the config i also need to add `log_file = "zebra.log"` to make it work properly.

- [x] I think this is the same bug, and I'd like to fix it in deserialization, so we don't need to remember to call a special method to get the correct value for `log_file`.

> We could add a new `Disabled` member to the `ProgressConfig` enum and make that the default to fix this problem.

I tried that first, but then I realised that `Some(Disabled)` was the same config as `None`. So the setting for a disabled progress bar is `None` (no setting). This lets us change the default later if we want to, without asking everyone to modify their configs.

> indicatif-log-bridge looks like a promising alternative if it works with howudoin

We don't use the `log` crate, we use `tracing`, so that integration won't work for us. (In particular, we use `tracing-subscriber` and `tracing-appender`, rather than [the `tracing` feature `log`](https://docs.rs/tracing/latest/tracing/#emitting-log-records).) 

> Another option is to document very well how do we expect users to use this feature in this stage. I will like to know that too.

- [x] I had planned on doing this in the changelog, forums, and Arborist call, but you're right, some user documentation is needed.

zebrad/src/components/tracing/component.rs
arya2: It's so that it uses a default log file path when no path is provided in the config but progress bars are enabled.

To "avoid having both [progress bars and logs] fight for your terminal"

arya2:
> We don't use the log crate, we use tracing, so that integration won't work for us. (In particular, we use tracing-subscriber and tracing-appender, rather than [the tracing feature log](https://docs.rs/tracing/latest/tracing/#emitting-log-records).)

The tracing-appender shouldn't be a problem, we could pass it a different writer that calls `suspend(|| write(msg))` if we add `suspend()` to `howudoin` or switch to using indicatif directly. I'll create an issue for this, users should at least be able to see warnings in their terminal alongside progress bars.

> I'd like to fix it in deserialization

🎉 - that's a good spot for it.

teor2345:
> > We don't use the log crate, we use tracing, so that integration won't work for us. (In particular, we use tracing-subscriber and tracing-appender, rather than [the tracing feature log](https://docs.rs/tracing/latest/tracing/#emitting-log-records).)
> 
> The tracing-appender shouldn't be a problem, we could pass it a different writer that calls `suspend(|| write(msg))` if we add `suspend()` to `howudoin` or switch to using indicatif directly. I'll create an issue for this, users should at least be able to see warnings in their terminal alongside progress bars.

Or we could have a tracing subscriber that calls [add_warn](https://docs.rs/howudoin/latest/howudoin/struct.Tx.html#method.add_warn), instead of writing to the terminal.

zebrad/src/components/tracing/component.rs
teor2345: This is fixed during deserialization in e154a5a4a6 using a wrapper type. This is the recommended workaround for a missing post-deserialization feature in serde.

teor2345:
> Another option is to document very well how do we expect users to use this feature in this stage. I will like to know that too.

This is documented in the README and API reference in commit 9e82ebd748.

zebrad/src/components/tracing.rs
```diff
@@ -166,11 +238,50 @@ impl Default for Config {
...
+    if progress_bar.is_some() {
+        return default_log_file();
+    }
+
+    None
+}
+
+/// Returns the configured log file path using the runtime `log_file` and `progress_bar` config.
+///
+/// This is the inverse of [`deserialize_default_log_file()`].
```
teor2345:
```suggestion
/// This is the inverse of [`runtime_default_log_file()`].
```

arya2 (dismissed):

teor2345:
Looks like one of the config_tests is hanging, probably because activating the progress bars using that config doesn't show the logs that the test is looking for.

I'll see if I can fix the test.

teor2345:
> Looks like one of the config_tests is hanging, probably because activating the progress bars using that config doesn't show the logs that the test is looking for.
> 
> I'll see if I can fix the test.

Ok that should be fixed now!

zebra-test/src/command/to_regex.rs
```diff
@@ -135,15 +136,20 @@ pub trait CollectRegexSet {
...
+            .map(|item| item.to_regex_set())
+            .try_collect();
         let regexes = regexes?;
 
         // This conversion discards flags and limits from Regex and RegexBuilder.
-        let regexes = regexes.iter().map(|regex| regex.as_str());
+        let regexes = regexes
+            .iter()
+            .map(|regex_set| regex_set.patterns())
+            .flatten();
```
teor2345:
```suggestion
            .flat_map(|regex_set| regex_set.patterns());
```

arya2 (approved):

teor2345:
> ## TODO
> After this PR merges, add:
> 
> * [x]  A changelog feature entry asking for user testing
> * [ ]  A post on the forums encouraging users to test it after it is released, and mentioning [the known issue with estimates](https://github.com/console-rs/indicatif/issues/556)
> * [ ]  A mention in the Arborist call

I have done the first task in PR #7610, I'll put the rest in the appropriate Slack channels.

PR #7616, opened 2023-09-25
ref(workflows): consolidate workflows based on their purpose

### LINKED ISSUE #6166: actions: use a name convention for workflow files/names
## Motivation

We'd like to standardize the way the workflows are named, and align their names with what's actually being executed inside the workflow.

### Specifications

When we do the redesign of Zebra CI, can we use a naming standard for workflow files?
For example, we could name them something like: `(purpose)-(when they run)-(where they run)-(what they do)-(patch?).yml`

So we'd end up with names like:
- `sub-build-docker-image.yml`
- `ci-pr-gcp-continuous-integration-test-docker.yml`
- `ci-pr-gh-continuous-integration-test-os.yml`
- `cd-release-gcp-continuous-delivery-deploy-docker.yml`

### LINKED ISSUE #6167: actions: segregate workflows based on their purpose
## Motivation

Some workflows have different requirements like resources, time and dependencies, and even have jobs with different objectives (in some cases, not related to other workflow jobs).

### Specifications

If multiple workflows require the same setup, this should be split into a dependable workflow which sets all the information of the following workflows. And if this workflows are not dependable on each other or have different objectives, those should also live in their own workflow files.

Sometimes we require to restart a specific workflow, which also triggers a whole dependency tree which makes re-running time consuming and almost as similar as depending on all previous workflows if there's a false/positive.

- We can't make a dependable workflow depend on another dependable workflow
- Reduce workflow nesting as much as possible
- Be able to cancel specific time consuming jobs without affecting other jobs which ran successfully
- Resource intensive workflows/jobs should be split apart if other workflows don't have this resource heavy requirements

gustavovalverde:
## Motivation

Some workflows have different requirements like resources, time and dependencies, and even have jobs with different objectives (in some cases, not related to other workflow jobs).

Fixes: #6166
Fixes: #6167
Depends-On: #7660

To comply with the naming convention, some jobs had to be moved from one workflow to another, and we have a completely new workflow, but no new jobs, which consolidates integration tests (sync tests) which required to be deployed in GCP 

## Solution

- Rename the workflows to make their naming more consistent by adding a naming convention
- Move jobs to the workflow which is more consistent with their objective

## Review

- Confirm that all workflows are working as expected (running and not failing)
- Validate if the naming convention makes sense

### Reviewer Checklist


teor2345:
I really like the new names!
https://github.com/ZcashFoundation/zebra/tree/refactor-workflows/.github/workflows

But as I said in the meeting I am the worst person to review them, because I know what they do already.

teor2345:
I think maybe this one is a typo?
https://github.com/ZcashFoundation/zebra/blob/refactor-workflows/.github/workflows/ci-unites-tests-os.patch.yml

The rest say `unit-tests`.

.github/workflows/ci-integration-tests-gcp.yml
```diff
@@ -397,7 +204,7 @@ jobs:
   test-full-sync:
```
teor2345, line 204: Can you please run a manual full sync on this PR before merging it?
We'll need to fix bug #7618 to make full syncs work again.

(We forgot to run a full sync for the self-hosted runners change, and so I'm currently checking all the PRs back past that change as part of diagnosing bug #7618.)

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,344 @@
...
+    steps:
+      - uses: r7kamura/rust-problem-matchers@v1.4.0
+
+      - name: Inject slug/short variables
+        uses: rlespinasse/github-slug-action@v4
+        with:
+          short-length: 7
+
+      # Make sure Zebra can sync the genesis block on testnet
+      - name: Run tests using a testnet config
```
teor2345: This job is timing out because it's configured for mainnet, not testnet. This seems like a bug in the workflow, because the Dockerfile hasn't changed.

teor2345 (dismissed):
There's a bug in a testnet workflow that's making the tests fail.

.github/workflows/ci-integration-tests-gcp.yml
```diff
@@ -101,20 +101,20 @@ jobs:
   # The default network is mainnet unless a manually triggered workflow or repository variable
   # is configured differently.
   #
-  # The outputs for this job have the same names as the workflow outputs in find-cached-disks.yml
+  # The outputs for this job have the same names as the workflow outputs in manual-find-cached-disks.yml
   get-available-disks:
     name: Check if cached state disks exist for ${{ inputs.network || vars.ZCASH_NETWORK }}
-    uses: ./.github/workflows/find-cached-disks.yml
+    uses: ./.github/workflows/manual-find-cached-disks.yml
```
teor2345: manual-find-cached-disks.yml is a sub-workflow, not a manual workflow.

gustavovalverde: Fixed with [chore: use better name for find-disks](https://github.com/ZcashFoundation/zebra/pull/7616/commits/3db9e6c89733c5baf6c573a3d555a4f8c9178016)

.github/workflows/ci-unit-tests-docker.yml
gustavovalverde: This should fix it: [fix(ci): use the entrypoint.sh to change the Network](https://github.com/ZcashFoundation/zebra/pull/7616/commits/eac74b8e20f06826a546716f51044af111ae3dcd)

gustavovalverde:
This one is ready @arya2

gustavovalverde:
This should be ready

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -1,4 +1,4 @@
-name: CD
+name: Deploy Nodes to GCP
```
teor2345, line 1: Do we want to delay this large CI change until after the release?
I'll leave it to you and @upbqdn to decide, because he is the release manager.

gustavovalverde: Same as in https://github.com/ZcashFoundation/zebra/pull/7682#discussion_r1357444420

gustavovalverde: @upbqdn if by any means this delays you considerably for the release, just go ahead without these 2 PRs.

teor2345:
It looks like this PR is missing some patch jobs, or they need to be renamed, or the branch protection rules need to be updated.

Different rules can be handled differently:
1. I don't think we need a `Run sync-to-checkpoint test` branch protection rule or patch job, because it almost never runs on PRs. (It doesn't even run in the scheduled job.)
2. The CD Docker rules are important for PRs, because we don't want to merge PRs that break Docker.

Screenshot 2023-10-13 at 13 25 27

.github/workflows/cd-deploy-nodes-gcp.yml
upbqdn: This release will be big enough even without this change, so let's delay?

teor2345:
> Different rules can be handled differently:
> 
> 1. I don't think we need a `Run sync-to-checkpoint test` branch protection rule or patch job, because it almost never runs on PRs. (It doesn't even run in the scheduled job.)

Admin: I have removed this useless branch protection rule, it's not actually checking anything in any PRs.

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -108,110 +88,14 @@ jobs:
...
-      - name: Inject slug/short variables
-        uses: rlespinasse/github-slug-action@v4
-        with:
-          short-length: 7
-
-      # Make sure Zebra can sync at least one full checkpoint on mainnet
-      - name: Run tests using the default config
-        shell: /usr/bin/bash -exo pipefail {0}
-        run: |
-          docker pull ${{ vars.GAR_BASE }}/zebrad@${{ needs.build.outputs.image_digest }}
```
teor2345: This test is checking the production Docker image...

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,355 @@
...
+      - uses: r7kamura/rust-problem-matchers@v1.4.0
+
+      - name: Inject slug/short variables
+        uses: rlespinasse/github-slug-action@v4
+        with:
+          short-length: 7
+
+      - name: Run tests using the default config
+        run: |
+          docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
```
teor2345: But when it is moved to here, it is testing the test (CI) docker image.

Changing the image build will run the correct commands, because they now use the same `entrypoint.sh`. But they use different Docker scripts and Rust features, so we need these tests to run on the production Docker image.

(Otherwise we risk breaking the production image and not finding out until a user reports it.)

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,355 @@
...
+          fi
+
+          # Handle other potential errors here
+          echo "An error occurred while processing the logs.";
+          exit 1;
+        env:
+          NETWORK: Mainnet
+
+  # Test reconfiguring the docker image for testnet.
+  test-configuration-file-testnet:
```
teor2345: This also applies to the testnet version of these tests.

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,355 @@
...
+          # Handle other potential errors here
+          echo "An error occurred while processing the logs.";
+          exit 1;
+        env:
+          ZEBRA_CONF_PATH: '/etc/zebrad/zebrad.toml'
+          NETWORK: Testnet
+
+  # Test that Zebra works using the $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test Zebra custom Docker config file
```
teor2345: Even if it didn't before, this test should also be running on the production Docker image. We don't care how the test image handles custom configs, because we don't do configs that way in tests.

teor2345 (changes requested):
I think there's an issue with the test coverage after this refactor. Some tests changed from running on the production image to the test image. But we need them to run on the production image, because that's the image that users use.

teor2345:
> 2\. The CD Docker rules are important for PRs, because we don't want to merge PRs that break Docker.

These rules are not being satisfied because the new jobs are running on the wrong Docker images. Fixing that will make this PR merge.

gustavovalverde:
> think there's an issue with the test coverage after this refactor. Some tests changed from running on the production image to the test image. But we need them to run on the production image, because that's the image that users use.

This has been fixed with:  [refactor: test config file in CI and CD with a reusable workflow](https://github.com/ZcashFoundation/zebra/pull/7616/commits/43ae06e7387d5b34b8c16f141b1017a595b5bd52)

gustavovalverde:
This is ready to go!

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -118,99 +98,37 @@ jobs:
...
-          exit 1;
+    uses: ./.github/workflows/sub-test-zebra-config.yml
+    with:
+      test_id: 'default-conf'
+      grep_patterns: '-e "net.*=.*Main.*estimated progress to chain tip.*BeforeOverwinter"'
+      test_variables: '-e NETWORK'
+      network: 'Mainnet'
 
-  # Test reconfiguring the docker image for testnet.
+  # Test reconfiguring the the docker image for tesnet.
```
arya2:
```suggestion
  # Test reconfiguring the docker image for testnet.
```

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,266 @@
...
+        run: |
+          docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }}
+          docker run -e NETWORK -e ZEBRA_TEST_LIGHTWALLETD --name lightwalletd-tests -t ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}:sha-${{ env.GITHUB_SHA_SHORT }} cargo test --locked --release --features lightwalletd-grpc-tests --package zebrad --test acceptance -- --nocapture --include-ignored lightwalletd_integration
+        env:
+          ZEBRA_TEST_LIGHTWALLETD: '1'
+          NETWORK: ${{ inputs.network || vars.ZCASH_NETWORK }}
+
+  # Test that Zebra works using the default config with the latest Zebra version.
+  test-configuration-file:
+    name: Test default config file
```
teor2345: Since these jobs run on both the CI (test) and CD (release) images, let's give them more specific names. The names also have to match the patch jobs.

```suggestion
    name: Test CI default Docker config file
```

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,266 @@
...
+    uses: ./.github/workflows/sub-test-zebra-config.yml
+    with:
+      test_id: 'default-conf'
+      grep_patterns: '-e "net.*=.*Main.*estimated progress to chain tip.*BeforeOverwinter"'
+      test_variables: '-e NETWORK'
+      network: 'Mainnet'
+
+  # Test reconfiguring the the docker image for tesnet.
+  test-configuration-file-testnet:
+    name: Test testnet config file
```
teor2345:
```suggestion
    name: Test CI testnet Docker config file
```

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,266 @@
...
+    with:
+      test_id: 'testnet-conf'
+      grep_patterns: '-e "net.*=.*Test.*estimated progress to chain tip.*Genesis" -e "net.*=.*Test.*estimated progress to chain tip.*BeforeOverwinter"'
+      # TODO: improve the entrypoint to avoid using `ENTRYPOINT_FEATURES=""`
+      test_variables: '-e NETWORK -e ZEBRA_CONF_PATH="/etc/zebrad/zebrad.toml" -e ENTRYPOINT_FEATURES=""'
+      network: 'Testnet'
+
+  # Test that Zebra works using $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test custom config file
```
teor2345:
```suggestion
    name: Test CI custom Docker config file
```

.github/workflows/ci-unit-tests-os.yml
```diff
@@ -31,12 +29,17 @@ on:
       - '.cargo/config.toml'
       - '**/clippy.toml'
       # workflow definitions
-      - '.github/workflows/continous-integration-os.yml'
-  pull_request:
+      - '.github/workflows/ci-unit-tests-os.yml'
+      - '.github/workflows/sub-build-docker-image.yml'
```
teor2345: The OS workflow does not use the Docker image, so it doesn't need to run when that file changes

teor2345:
```suggestion
```

.github/workflows/ci-unit-tests-docker.patch.yml
```diff
@@ -0,0 +1,82 @@
...
+      - run: 'echo "No build required"'
+
+  test-lightwalletd-integration:
+    name: Test integration with lightwalletd
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-configuration-file:
+    name: Test Zebra default Docker config file
```
teor2345:
```suggestion
    name: Test CI default Docker config file
```

.github/workflows/ci-unit-tests-docker.patch.yml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-zebra-conf-path:
+    name: Test Zebra custom Docker config file
```
teor2345:
```suggestion
    name: Test CI custom Docker config file
```

.github/workflows/docs-deploy-firebase.patch.yml
```diff
@@ -18,7 +18,7 @@ on:
       - '.cargo/config.toml'
       - '**/clippy.toml'
       # workflow definitions
-      - '.github/workflows/docs.yml'
+      - '.github/workflows/docs-deploy-firebase.yml'
```
teor2345, line 21: When we change the workflow file name it also needs to change in the patch workflow.

I applied this change to the branch because I can't comment on unchanged files.

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -118,99 +98,37 @@ jobs:
 
   # Test that Zebra works using the default config with the latest Zebra version.
   test-configuration-file:
-    name: Test Zebra CD Docker config file
-    timeout-minutes: 15
-    runs-on: ubuntu-latest
+    name: Test Zebra default Docker config file
```
teor2345: We can either change the branch protection rules to use the new name (I prefer this one for consistency)

```suggestion
    name: Test CD default Docker config file
```

Or change the job name to match the branch protection rules

```suggestion
    name: Test Zebra default Docker config file
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -118,99 +98,37 @@ jobs:
...
+      test_variables: '-e NETWORK'
+      network: 'Mainnet'
 
-  # Test reconfiguring the docker image for testnet.
+  # Test reconfiguring the the docker image for tesnet.
   test-configuration-file-testnet:
-    name: Test testnet Zebra CD Docker config file
-    timeout-minutes: 15
-    runs-on: ubuntu-latest
+    name: Test Zebra testnet Docker config file
```
teor2345: The production image name needs to be different from the CI (test) image jobs.

```suggestion
    name: Test CD testnet Docker config file
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -118,99 +98,37 @@ jobs:
...
+    uses: ./.github/workflows/sub-test-zebra-config.yml
+    with:
+      test_id: 'testnet-conf'
+      grep_patterns: '-e "net.*=.*Test.*estimated progress to chain tip.*Genesis" -e "net.*=.*Test.*estimated progress to chain tip.*BeforeOverwinter"'
+      test_variables: '-e NETWORK -e ZEBRA_CONF_PATH="/etc/zebrad/zebrad.toml"'
+      network: 'Testnet'
+
+  # Test that Zebra works using $ZEBRA_CONF_PATH config
+  test-zebra-conf-path:
+    name: Test Zebra custom Docker config file
```
teor2345: We can either change the branch protection rules to use the new name (I prefer this one for consistency)

```suggestion
    name: Test CD custom Docker config file
```

Or change the job name to match the branch protection rules

```suggestion
    name: Test Zebra custom Docker config file
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -48,32 +48,12 @@ on:
   #    - 'docker/**'
   #    - '.dockerignore'
   #    - '.github/workflows/continous-delivery.yml'
-  #    - '.github/workflows/build-docker-image.yml'
-
-  # Only runs the Docker image tests, doesn't deploy any instances
-  pull_request:
```
teor2345, line 54: The Docker release image checks need to run on PRs so that we don't merge PRs that break the release images.

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -225,14 +143,14 @@ jobs:
...
-  # - on every push/merge to the `main` branch
+  # - on every push to the `main` branch
   # - on every release, when it's published
   deploy-nodes:
     strategy:
       matrix:
         network: [Mainnet, Testnet]
     name: Deploy ${{ matrix.network }} nodes
-    needs: [ build, test-configuration-file, versioning ]
+    needs: [ build, versioning, test-configuration-file, test-configuration-file-testnet, test-zebra-conf-path ]
```
teor2345: We can't depend on testnet jobs because testnet is unreliable.

```suggestion
    needs: [ build, versioning, test-configuration-file, test-zebra-conf-path ]
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -334,7 +252,7 @@ jobs:
   # Note: this instances are not automatically replaced or deleted
   deploy-instance:
     name: Deploy single ${{ inputs.network }} instance
-    needs: [ build, test-configuration-file ]
+    needs: [ build ]
```
teor2345: This should depend on the 2 mainnet config file tests, because we don't want to deploy nodes if the image fails.

.github/workflows/continous-delivery.patch.yml
```diff
@@ -1,35 +0,0 @@
-name: CD
-
-on:
-  # Only patch the Docker image test jobs
-  pull_request:
-    paths-ignore:
```
teor2345, line 6: This patch file is still required because we want the release image jobs to run on PRs. It will also need to be updated with the new set of jobs.

.github/workflows/ci-unit-tests-docker.yml
teor2345: The 2 new mainnet jobs in this file will also need new branch protection rules, because we don't want PRs to merge if the CI images stop working.

teor2345:
We're almost there with the set of PR jobs and the branch protection rules, but we need to check the release images in every PR, just like we did before.

Otherwise we risk releasing an image that doesn't work.

teor2345:
I think it's ok to change the branch protection rules now, as long as we get this PR in soon. But doing that will stop all the other PRs from merging.

gustavovalverde:
**EDIT**: Actually, let's keep the test coverage as is, I'll make the workflow segregation in another PR [fix: revert test coverage on CD](https://github.com/ZcashFoundation/zebra/pull/7616/commits/100f95ed606afd1cdd485c86b78dc4ee1dabeba0)


> The Docker release image checks need to run on PRs so that we don't merge PRs that break the release images.

> we need to check the release images in every PR, just like we did before.

If we don't want to release a production image with issues, I think running this in the `main` branch should be a good starting point, as we'll be warned when this test fails.

A more "elegant" solution is also making the push on `Docker Hub` a dependency on this tests, at **release** time (but that's a refactor on its own). Otherwise, if we're going to test this in PRs I think this file should be segregated between a workflow that is meant to deploy instances (CD), and other that is meant to run tests on the test or release image (CI)

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: Reverted

.github/workflows/continous-delivery.patch.yml
gustavovalverde: Reverted

gustavovalverde:
All recommendations were applied

gustavovalverde:
I made a last fix to the docker config test to accept different images. Even though this was (mainly) required in this spots: https://github.com/ZcashFoundation/zebra/pull/7616/commits/fd6aaf217d5cce9b932cf8e8636e937e207a1700#diff-d43d0df5f20e7f637ae74715b01cc8a0b644574a546a5242c78b9b8d4bab9571R217

I added it to other places for consistency

gustavovalverde:
If this test fails https://github.com/ZcashFoundation/zebra/actions/runs/6553940936/job/17800228563?pr=7616 after [fix(dockerfile): use variables or default for config path and file](https://github.com/ZcashFoundation/zebra/pull/7616/commits/3757299fb02209654c024f999eef07de14b60be6) it might because this file does not exists in the release image: `zebrad/tests/common/configs/v1.0.0-rc.2.toml`

If that's the case, we can remove this test from CD, or use a file generated by Zebra, different than the one created by the entrypoint

.github/workflows/cd-deploy-nodes-gcp.patch.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-configuration-file-testnet:
+    name: Test CD testnet Docker config file
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
```
teor2345: Optional: testnet doesn't need a patch job, because we can't use it in branch protection rules, because it is unreliable

```suggestion
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -225,14 +165,14 @@ jobs:
...
-  # - on every push/merge to the `main` branch
+  # - on every push to the `main` branch
   # - on every release, when it's published
   deploy-nodes:
     strategy:
       matrix:
         network: [Mainnet, Testnet]
     name: Deploy ${{ matrix.network }} nodes
-    needs: [ build, test-configuration-file, versioning ]
+    needs: [ build, versioning, test-configuration-file, test-zebra-conf-path ]
```
teor2345, line 175:
```suggestion
    needs: [ build, versioning, test-configuration-file, test-zebra-conf-path, test-configuration-file ]
```

.github/workflows/ci-unit-tests-docker.yml
```diff
@@ -0,0 +1,269 @@
...
+      test_variables: '-e NETWORK -e ZEBRA_CONF_PATH="zebrad/tests/common/configs/v1.0.0-rc.2.toml"'
+      network: ${{ inputs.network || vars.ZCASH_NETWORK }}
+
+  failure-issue:
+    name: Open or update issues for main branch failures
+    # When a new test is added to this workflow, add it to this list.
+    #
+    # This list is for reliable tests that are run on the `main` branch.
+    # Testnet jobs are not in this list, because we expect testnet to fail occasionally.
+    needs: [ test-all, test-all-getblocktemplate-rpcs, test-fake-activation-heights, test-empty-sync, test-lightwalletd-integration, test-configuration-file, test-zebra-conf-path, test-configuration-file-testnet ]
```
teor2345: We don't want to open issues for testnet jobs, because they are unreliable

```suggestion
    needs: [ test-all, test-all-getblocktemplate-rpcs, test-fake-activation-heights, test-empty-sync, test-lightwalletd-integration, test-configuration-file, test-zebra-conf-path ]
```

teor2345:
I'm going to delete the old branch protection rules now, so this PR and other PRs can merge.

Admin: This PR needs 4 new branch protection rules after it merges:
- [ ] Test CD default Docker config file / Test default-conf in Docker
- [ ] Test CD custom Docker config file / Test custom-conf in Docker
- [ ] Test CI default Docker config file / Test default-conf in Docker
- [ ] Test CI custom Docker config file / Test custom-conf in Docker

I'll go update the patch jobs to match in this PR now.

.github/workflows/cd-deploy-nodes-gcp.patch.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  build:
+    name: Build CD Docker / Build images
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-configuration-file:
+    name: Test CD default Docker config file
```
teor2345:
```suggestion
    name: Test CD default Docker config file / Test default-conf in Docker
```

.github/workflows/cd-deploy-nodes-gcp.patch.yml
```diff
@@ -0,0 +1,45 @@
...
+      - run: 'echo "No build required"'
+
+  test-configuration-file-testnet:
+    name: Test CD testnet Docker config file
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-zebra-conf-path:
+    name: Test CD custom Docker config file
```
teor2345:
```suggestion
    name: Test CD custom Docker config file / Test custom-conf in Docker
```

.github/workflows/ci-unit-tests-docker.patch.yml
```diff
@@ -0,0 +1,82 @@
...
+      - run: 'echo "No build required"'
+
+  test-lightwalletd-integration:
+    name: Test integration with lightwalletd
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-configuration-file:
+    name: Test CI default Docker config file
```
teor2345:
```suggestion
    name: Test CI default Docker config file / Test default-conf in Docker
```

.github/workflows/ci-unit-tests-docker.patch.yml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No build required"'
+
+  test-zebra-conf-path:
+    name: Test CI custom Docker config file
```
teor2345:
```suggestion
    name: Test CI custom Docker config file / Test custom-conf in Docker
```

teor2345 (approved):
Thank you for all your work on this, I think we're good to go here

teor2345:
> If we don't want to release a production image with issues, I think running this in the `main` branch should be a good starting point, as we'll be warned when this test fails.
> 
> A more "elegant" solution is also making the push on `Docker Hub` a dependency on this tests, at **release** time (but that's a refactor on its own).

We like to stop these kinds of bugs before we merge a PR, because fixing them after we merge or tag the release takes extra time. This is particularly important if we're trying to get a release out quickly to meet a deadline or fix a security issue.

It's ok to also test in the `main` branch if you want, but this seems less important (maybe optional?). It helps us work out if an issue is in multiple PRs, or the `main` branch and all PRs.

> Otherwise, if we're going to test this in PRs I think this file should be segregated between a workflow that is meant to deploy instances (CD), and other that is meant to run tests on the test or release image (CI)

That seems like a great idea! There's no need to have the release depend on the release image tests, as long as we're sure the tested release image is always the same as the DockerHub push released image. (If we're testing every PR, it's unlikely that the test will fail during the release itself.)

PR #7617, opened 2023-09-25
ref(docker): use cache mounts for build cache

### LINKED ISSUE #6169: docker: improve Rust dependency caching in CI to avoid rebuilding
## Motivation

When files are copied into the Dockerfile with the `COPY` command, the cache can get invalidated. We could improve this by using the `--link` option which creates more independent layers to avoid rebuilding when not needed

### Specifications

Documentation: https://docs.docker.com/engine/reference/builder/#copy---link

gustavovalverde:
## Motivation

Caching Rust is not straightforward and we've been using Cargo Chef so far to accomplish that, but build caches are being invalidated most of the time, not making it very useful.

This is considerably documented:
- https://hackmd.io/@kobzol/S17NS71bh
- https://github.com/moby/moby/issues/39530
- https://github.com/rust-lang/cargo/issues/2644
- https://github.com/rust-lang/cargo/issues/6529
- https://github.com/Swatinem/rust-cache/issues/155

Depends-On: #7616
Fixes: #6169

### Specifications

- https://docs.docker.com/build/cache/
- https://docs.docker.com/engine/reference/builder/#copy---link
- https://docs.docker.com/language/rust/develop/

This changes are advanced Docker use cases, the above documentation might make it easier to understand what's going on.

This aldo depends on #7616, but the only change to focus here is the `Dockerfile`

## Solution

Implement the approach in https://docs.docker.com/language/rust/develop/ with `COPY --link`

## Review

We must confirm with several changes:
- Rust code
- Workflows (which has been invalidating the image, recently)
- etc

### Reviewer Checklist


gustavovalverde:
I'll reopen this when needed.

PR #7627, opened 2023-09-26
change(state): Only do continuous format checks in CI

teor2345:
## Motivation

We don't want to run continuous state format checks on all running Zebra instances.

It's also possible these checks are making syncing slow, but we won't know until the jobs in #7618 finish.

We need to enable this config in CI, but disable it by default. Almost all the CI configs call the default config method, then make changes. So these changes should be enough.

I want to check the CI logs to make sure:
- [ ] The checks are happening in CI 
- [ ] The shutdown check happens soon enough to fail the test (if not, I will change some tests to wait for it)

## Solution

- Only check the state format at startup and shutdown
    - Add a debug config for a state format check interval, which is off by default
- Enable state format checks in CI every 5 minutes

## Review

This PR blocks the release.

### Reviewer Checklist

## Follow Up Work

This code might need more changes based on ticket #7618. We can make that fix in another PR if needed.


zebrad/tests/common/cached_state.rs
```diff
@@ -30,9 +30,15 @@ use crate::common::{
     test_type::TestType,
 };
 
-/// Path to a directory containing a cached Zebra state.
+/// The environmental variable that holds the path to a directory containing a cached Zebra state.
```
arya2, line 33:
```suggestion
/// The environment variable that holds the path to a directory containing a cached Zebra state.
```

arya2 (approved):

PR #7629, opened 2023-09-27
change(state): Check for mistakenly deleted note commitment trees

### LINKED ISSUE #7461: Check that there are no missing note commitment trees after deduplication
## Motivation

Test that the note commitment tree deduplication format upgrade keeps all unique note commitment trees.

### Possible Design

- Find every height with a tree change
- Check that the number of notes added to that tree equals the number of notes in the transactions of the block at that height

teor2345:
## TODO

- [ ] Decide how to deal with the slow test
- [ ] Implement that solution

## Motivation

We want to make sure we haven't deleted any unique note commitment trees as part of the tree de-duplication upgrade.

Close #7461

This check is very slow, because it reads a lot of blocks:
- 33 seconds on testnet
- TODO: over 10 minutes on mainnet

So alternatives are:
- run a quick partial check in all CI jobs
- run the full check in a separate test function, ideally as part of an existing (quick) CI job
- run on some random heights in a separate test function, ideally as part of an existing (quick) CI job
- only run the full check manually, when we change this code
- TODO: are there any other alternatives

I don't think we can decide until we know how slow it is, but let's list the alternatives.

## Solution

- [Add quick genesis tree note count checks](https://github.com/ZcashFoundation/zebra/commit/8326e4378aeb103a482ed2ff23a3422f2a6a0fe8)
- [Check that no unique trees were accidentally deleted as part of the state validity checks](https://github.com/ZcashFoundation/zebra/commit/b9b70618e25f28bc6705593a5611fd0f30ebe318)

## Review

This PR is not ready for review yet.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade.rs
@@ -16,7 +16,9 @@ use zebra_chain::{
         task::{CheckForPanics, WaitForPanics},
         CodeTimer,
     },
+    orchard,
     parameters::Network,
+    sapling,
 };
 
 use DbFormatChange::*;
@@ -628,9 +630,12 @@ impl DbFormatChange {
 
         if !db.is_empty() {
             let genesis = Height(0);
-            if let Err(err) = Self::check_tree_note_counts_for_height(db, genesis) {
-                result = Err(err);
-            }
+            let sapling_tree = db.sapling_tree_by_height(&genesis);
+            let orchard_tree = db.orchard_tree_by_height(&genesis);
+
+            let count_result =
+                Self::check_tree_note_counts_for_height(db, genesis, sapling_tree, orchard_tree);
+            result = result.and(count_result);
         }
 
         // We don't need to check for an empty database, because the trees should be missing either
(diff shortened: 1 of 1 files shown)


zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -613,17 +620,300 @@ impl DbFormatChange {
...
+            match sapling_height.cmp(&orchard_height) {
+                Ordering::Equal => {
+                    count_result = Self::check_tree_note_counts_for_height(
+                        db,
+                        sapling_height,
+                        Some(sapling_tree.clone()),
+                        prev_sapling_tree,
+                        Some(orchard_tree.clone()),
+                        prev_orchard_tree,
+                    );
```
teor2345, line 756: This is the loop that calls the slow code, or the iterators might be slow.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -613,17 +620,300 @@ impl DbFormatChange {
...
+        prev_orchard_tree: Option<Arc<orchard::tree::NoteCommitmentTree>>,
+    ) -> Result<(), String> {
+        // Check both pools before returning any errors.
+        let mut result = Ok(());
+
+        // Deserializing the block is the most expensive part of this function,
+        // so it is ok to do a redundant check if there are no sapling or orchard notes in it.
+        let block = db
+            .block(height.into())
+            .expect("caller should check for block or tree at height");
```
teor2345, line 860: I think this is the slowest part of the code inside that loop, but I haven't done any profiling.

teor2345:
This check took about 2 hours and 40 minutes on my machine with mirrored fast SSDs. It timed out at 3 hours in CI.

Since this code reads a lot of large blocks, I'm not sure if we'll be able to make it much faster, unless we split shielded transactions or outputs into their own column families in the database. Even then, it might not improve the speed.

We might also be seeing slow deserialization because it does cryptographic operations (ticket #4788, which was only partially fixed).

PR #7636, opened 2023-09-27
fix(note-commitment-trees): Populate subtree fields in the NoteCommitmentTrees structs

### LINKED ISSUE #7444: Correctly populate subtree values in ZebraDb::note_commitment_trees()
## Motivation

The subtrees returned by `ZebraDb::note_commitment_trees()` are currently `None`.

### Suggested Solution

Populate the subtree fields with the tip subtrees, or document that they are always `None` in the function documentation.

This change might need `*_subtree_for_tip()` methods.

You can get the tip trees with `zs_last_key_value()`, PR #7408 has the required `impl FromDisk for NoteCommitmentSubtreeIndex`.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/7350#discussion_r1309528107_

oxarbitrage:
## Motivation

The subtree fields of `ZebraDb::note_commitment_trees()` and `Chain::update_chain_tip_with_block_parallel()` are always `None`.

Close https://github.com/ZcashFoundation/zebra/issues/7444

## Solution

Populate fields when we have data.

## Review

Open to suggestions, i was not able to figure out a quick way to add testing for this.

### Reviewer Checklist


zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -289,6 +290,32 @@ impl ZebraDb {
...
+        if tip_height >= sapling_activation_height {
+            let sapling_nct_handle = self.db.cf_handle("sapling_note_commitment_tree").unwrap();
+
+            let tip_note_commitment_tree: Arc<sapling::tree::NoteCommitmentTree> = self
+                .zs_last_key_value(&sapling_nct_handle)
+                .map(|(_key, value): (RawBytes, _)| Arc::new(value))
+                .expect("Sapling note commitment tree must exist if there is a finalized tip");
+
+            if let Some((subtree_index, node)) =
+                tip_note_commitment_tree.completed_subtree_index_and_root()
```
teor2345: This method might not do what you think it does, you might want to check the docs. (@arya2 is the best person to explain the details, because he wrote it.)

mpguerra: if the details are not clear from the code/documentation we should make sure we capture those also

oxarbitrage: I think we might need to use `subtree_index()` to get the `subtree_index` but unsure how to get the `node` yet.

oxarbitrage:
EDIT: There was actually an error on my code causing the failures. This is fixed in https://github.com/ZcashFoundation/zebra/pull/7636/commits/f339bd62b8feb5302f090e1ffb9dcb844f62e976

----------------------------------------------------------------------------
<strike>This CD issue seems unrelated</strike>

```
Error: Fail to evaluate job outputs
Error: The template is not valid. ZcashFoundation/zebra/.github/workflows/build-docker-image.yml@cda56f1d4d415beee17d[3](https://github.com/ZcashFoundation/zebra/actions/runs/6331470384/job/17196096691?pr=7636#step:44:3)78503f57ad4573505ca (Line: 59, Col: 19): Error reading JToken from JsonReader. Path '', line 0, position 0.
```

https://github.com/ZcashFoundation/zebra/actions/runs/6331470384/job/17196096691?pr=7636

<strike>I am re running all workflows as i think that might was a temporal problem.</strike>

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -289,6 +290,33 @@ impl ZebraDb {
...
+            {
+                return Some(NoteCommitmentSubtree::new(subtree_index, tip_height, node));
+            }
+        }
+
+        None
```
arya2: This should work:

```suggestion
        let sapling_subtrees = self
            .db
            .cf_handle("sapling_note_commitment_subtree")
            .unwrap();

        let (index, subtree_data): (
            NoteCommitmentSubtreeIndex,
            NoteCommitmentSubtreeData<sapling::tree::Node>,
        ) = self.db.zs_last_key_value(&sapling_subtrees)?;

        Some(subtree_data.with_index(index))
```

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -435,15 +463,42 @@ impl ZebraDb {
...
+                .expect("Orchard note commitment tree must exist if there is a finalized tip");
+
+            if let Some((subtree_index, node)) =
+                tip_note_commitment_tree.completed_subtree_index_and_root()
+            {
+                return Some(NoteCommitmentSubtree::new(subtree_index, tip_height, node));
+            }
+        }
+
+        None
```
arya2:
```suggestion
        let orchard_subtrees = self
            .db
            .cf_handle("orchard_note_commitment_subtree")
            .unwrap();

        let (index, subtree_data): (
            NoteCommitmentSubtreeIndex,
            NoteCommitmentSubtreeData<orchard::tree::Node>,
        ) = self.db.zs_last_key_value(&orchard_subtrees )?;

        Some(subtree_data.with_index(index))
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -737,6 +737,16 @@ impl Chain {
...
     }
 
+    /// Returns the Sapling note commitment subtree of the tip of this [`Chain`],
+    pub fn sapling_subtree_for_tip(&self) -> Option<NoteCommitmentSubtree<sapling::tree::Node>> {
+        if !self.is_empty() {
+            let tip = self.non_finalized_tip_height();
+            Some(self.sapling_subtree(tip.into())?)
+        } else {
+            None
+        }
```
arya2: Optional/Nitpick:

```suggestion
        let (&index, data) = self.sapling_subtrees.last_key_value()?;
        Some(data.with_index(index))
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -939,6 +949,16 @@ impl Chain {
...
     }
 
+    /// Returns the Orchard note commitment subtree of the tip of this [`Chain`],
+    pub fn orchard_subtree_for_tip(&self) -> Option<NoteCommitmentSubtree<orchard::tree::Node>> {
+        if !self.is_empty() {
+            let tip = self.non_finalized_tip_height();
+            Some(self.orchard_subtree(tip.into())?)
+        } else {
+            None
+        }
```
arya2:
```suggestion
        let (&index, data) = self.orchard_subtrees.last_key_value()?;
        Some(data.with_index(index))
```

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: > This method might not do what you think it does, you might want to check the docs. (@arya2 is the best person to explain the details, because he wrote it.)

`completed_subtree_index_and_root()` was meant for finding subtree roots during a format upgrade or when adding verified blocks to the state, could it get the last subtree on disk instead?

teor2345: > if the details are not clear from the code/documentation we should make sure we capture those also

What the method does is clear to me, but the intended usage is not mentioned in the docs:
https://github.com/ZcashFoundation/zebra/blob/cdddbcc7fe1d114fcfc8ce169bf93fdfd2ac4a67/zebra-chain/src/orchard/tree.rs#L537-L538

It might also help to say that the method returns `None` most of the time. (It only returns `Some` for `1 / 2^16` notes.)

@arya2 is this documentation something you could fix?

teor2345: > could it get the last subtree on disk instead?

Is this what we actually want though?

We don't know what the method is meant to do, because the documentation on `note_commitment_trees()` is unclear and needs to be updated.

But looking at how `note_commitment_trees()` is called, it is only used in this one place, where the subtrees are ignored:
https://github.com/ZcashFoundation/zebra/blob/cdddbcc7fe1d114fcfc8ce169bf93fdfd2ac4a67/zebra-state/src/service/finalized_state.rs#L245-L250

But it looks like it is meant to be equivalent to `prev_note_commitment_trees`, which is populated by `Chain::treestate()` (which also needs documentation). So I had to look all the way back to:
https://github.com/ZcashFoundation/zebra/blob/cdddbcc7fe1d114fcfc8ce169bf93fdfd2ac4a67/zebra-state/src/service/non_finalized_state/chain.rs#L694-L701

So let's assume `note_commitment_trees()` is like `treestate()`, and it should return the trees for the tip block, and the subtrees that end at the tip block. So the subtrees will usually be `None`. (These methods could use the same name to make their matching behaviour clearer.)

If we make the suggested changes to alway get the last subtree, the code will be incorrect, because the last subtree usually ends before the tip block.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: I think this code is functional, but doesn't actually do what we want (or what this method is documented to do):
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340708466

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: Let's hold off on Orchard changes until we've got the Sapling changes worked out?
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340708466

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -435,15 +463,42 @@ impl ZebraDb {
...
+    }
+
     /// Returns the shielded note commitment trees of the finalized tip
     /// or the empty trees if the state is empty.
     pub fn note_commitment_trees(&self) -> NoteCommitmentTrees {
```
teor2345: This documentation needs to be updated to say which subtrees are being returned, and we might also want to rename this method to say `tip`. (And rename `Chain::treestate()` to `Chain::note_commitment_trees()` to match this method.)
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340708466

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -737,6 +737,16 @@ impl Chain {
             .collect()
     }
 
+    /// Returns the Sapling note commitment subtree of the tip of this [`Chain`],
```
teor2345: This documentation is ambiguous, because it doesn't say "latest subtree" or "completed subtree.

Orchard also needs the same changes.

Here's an edited copy of the docs for `Chain::sapling_subtree()`:
```suggestion
    /// Returns the Sapling [`NoteCommitmentSubtree`] if it was completed at the tip height.
```

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: This suggestion changes the behaviour of the function from "completed at tip" to "most recently completed". I'm not sure that's what we want:
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340708466

zebra-state/src/service/non_finalized_state/chain.rs
teor2345: This is a behaviour-changing suggestion:
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340711814

teor2345 (commented):
Before we make any more detailed suggestions, let's decide if the method's behaviour is:
1. returns a subtree if it was completed in the block at the tip, or
2. returns the last completed subtree regardless of its end height.

I think it's 1 because that's what we do in `Chain`:
https://github.com/ZcashFoundation/zebra/pull/7636/files#r1340708466

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: I think documenting that only used for writing and always `None` on read would be good too.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: It could filter the return value unless the end height matches the finalized tip height.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
oxarbitrage: I don't think we should change `Chain::treestate()` to `Chain::note_commitment_trees()` as this method not just returns `NoteCommitmentTrees` but also `HistoryTree` which is wrapped as a `Treestate`.
This is not the same as the `ZebraDb::note_commitment_tree` which just return `NoteCommitmentTrees`.

In addition, `Chain::treestate()` takes a height, so we should not add `_for_tip` suffix as I did for others in https://github.com/ZcashFoundation/zebra/pull/7636/commits/d78d8d7814bf6e6bfd42fe6dee74f61c52dfdf1a

Please also note `ZebraDb::note_commitment_trees` was changed to `ZebraDb::note_commitment_trees_for_tip` in https://github.com/ZcashFoundation/zebra/pull/7636/commits/d45243e658fa8bed5de563ff96b224f1aabedd1e

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
oxarbitrage: Code suggestion was added in https://github.com/ZcashFoundation/zebra/pull/7636/commits/84e79009ecb7ffbc41e056278d99151018070430
Filter was added in https://github.com/ZcashFoundation/zebra/pull/7636/commits/5866bbba3f9b4e025c91b6c150c10e9ff937d494

zebra-state/src/service/non_finalized_state/chain.rs
oxarbitrage: I think we should not apply this suggestion and return the subtree that was completed at tip (current behaviour), which is the same we did for `ZebraDb`.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -289,11 +289,32 @@ impl ZebraDb {
...
+            .db
+            .cf_handle("sapling_note_commitment_subtree")
+            .unwrap();
+
+        let (index, subtree_data): (
+            NoteCommitmentSubtreeIndex,
+            NoteCommitmentSubtreeData<sapling::tree::Node>,
+        ) = self.db.zs_last_key_value(&sapling_subtrees)?;
+
+        let tip_height = self.finalized_tip_height().unwrap_or(Height(0));
```
teor2345: Optional: This can be changed in Orchard as well.

```suggestion
        let tip_height = self.finalized_tip_height()?;
```

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -737,6 +737,16 @@ impl Chain {
             .collect()
     }
 
+    /// Returns the Sapling [`NoteCommitmentSubtree`] if it was completed at the tip height.
+    pub fn sapling_subtree_for_tip(&self) -> Option<NoteCommitmentSubtree<sapling::tree::Node>> {
+        if !self.is_empty() {
+            let tip = self.non_finalized_tip_height();
+            Some(self.sapling_subtree(tip.into())?)
```
teor2345: Optional: Also can be done for Orchard.

```suggestion
            self.sapling_subtree(tip.into())
```

teor2345 (dismissed):
Looks great!

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -302,7 +302,7 @@ impl ZebraDb {
             NoteCommitmentSubtreeData<sapling::tree::Node>,
         ) = self.db.zs_last_key_value(&sapling_subtrees)?;
```
teor2345, line 317: Marek has discovered that `zs_last_key_value()` might be causing the performance problems during syncing, so when we find a solution to that issue, we'll need to apply it here. Let's wait to merge this PR until we have a fix, because it will get confusing if we merge more slow code.

Here's the original analysis, we are still working on a solution:
https://github.com/ZcashFoundation/zebra/issues/7618#issuecomment-1745607787

teor2345: It looks like this might only be a problem for iterator lookups on column families with lots of deleted keys. Since we only delete subtrees once during an upgrade (which should all be done now), and we don't look up subtrees very often, this issue does not impact this PR.

We'll do some longer-term fixes in ticket #7664, and we can fix this code there if we need to.

teor2345 (approved):

teor2345:
@mergifyio update

teor2345:
@mergifyio refresh

teor2345:
Failed in the merge queue in PR #7692 due to issue #7659, I've re-queued it.

PR #7600, opened 2023-09-20
fix(test): Remove intermittent asserts in test

oxarbitrage:
## Motivation

Close https://github.com/ZcashFoundation/zebra/issues/7598

## Solution

Remove intermittent asserts as at least a temporary fix.

## Review

Anyone can review

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

More research about this problem at https://github.com/ZcashFoundation/zebra/issues/7529

PR #7644, opened 2023-09-28
fix(tests): Checking if transactions get into the mempool fails in lightwalletd tests

### LINKED ISSUE #7529: Checking if transactions get into the mempool fails in lightwalletd tests
In `send_transaction_tests` we send transactions using grpc and then attempt to check if they are inside the mempool using grpc calls as well. This should be more or less what a wallet using lightwalletd may want to do.

However, the check of the mempool works in some local environments (i was able to make that work reliable locally) but fails in the CI for unknown reasons.

In the past, when we were using `adityapk00` lightwalletd fork the mempool calls were commented out so an empty response was expected.  When we changed to the `zcash` lightwalletd we were expecting this test to start working so they were enabled in https://github.com/ZcashFoundation/zebra/pull/7349, however this is not the case so we disabled them back again in the same PR.

We don't know the reason of the failure. More details can be found in https://github.com/ZcashFoundation/zebra/pull/7349

arya2:
## Motivation

Zebra should be using the zcash/lightwalletd service.proto.

Close #7529.

## Solution

- Use updated service.proto from [zcash/lightwalletd](https://github.com/zcash/lightwalletd/blob/master/walletrpc/service.proto)
  - Where height has been changed from -1 to 0, lightwalletd ignores that field anyways.

Related Cleanups:
- Remove redundant truncate in test code

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebrad/tests/common/lightwalletd/proto/service.proto
+++ b/zebrad/tests/common/lightwalletd/proto/service.proto
@@ -36,7 +36,7 @@ message TxFilter {
 // by GetMempoolStream(), the latest block height.
 message RawTransaction {
     bytes data = 1;     // exact data returned by Zcash 'getrawtransaction'
-    int64 height = 2;  // height that the transaction was mined (or -1)
+    uint64 height = 2;  // height that the transaction was mined (or -1)
 }
 
 // A SendResponse encodes an error code and a string. It is currently used
@@ -110,11 +110,12 @@ message Exclude {
 
 // The TreeState is derived from the Zcash z_gettreestate rpc.
 message TreeState {
-    string network = 1; // "main" or "test"
-    uint64 height = 2;
-    string hash = 3;    // block id
-    uint32 time = 4;    // Unix epoch time when the block was mined
-    string tree = 5;    // sapling commitment tree state
+    string network = 1;     // "main" or "test"
+    uint64 height = 2;      // block height
+    string hash = 3;        // block id
+    uint32 time = 4;        // Unix epoch time when the block was mined
+    string saplingTree = 5; // sapling commitment tree state
+    string orchardTree = 6; // orchard commitment tree state
 }
 
 enum ShieldedProtocol {
@@ -122,16 +123,11 @@ enum ShieldedProtocol {
     orchard = 1;
 }
 
-// Request type for `GetSubtreeRoots`
 message GetSubtreeRootsArg {
-    uint32 startIndex = 1;                  // Index identifying where to start returning subtree roots.
-    ShieldedProtocol shieldedProtocol = 2;  // Shielded protocol to return subtree roots for.
+    uint32 startIndex = 1;                  // Index identifying where to start returning subtree roots
+    ShieldedProtocol shieldedProtocol = 2;  // Shielded protocol to return subtree roots for
     uint32 maxEntries = 3;                  // Maximum number of entries to return, or 0 for all entries.
 }
-
-
-// Response type for `GetSubtreeRoots`.
-// The actual response contains a stream of `SubtreeRoot`s.
 message SubtreeRoot {
     bytes rootHash = 2;                     // The 32-byte Merkle root of the subtree.
     bytes completingBlockHash = 3;          // The hash of the block that completed this subtree.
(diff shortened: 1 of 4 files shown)


zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -148,9 +147,6 @@ pub async fn run() -> Result<()> {
 
     let mut rpc_client = connect_to_lightwalletd(lightwalletd_rpc_port).await?;
 
-    // To avoid filling the mempool queue, limit the transactions to be sent to the RPC and mempool queue limits
-    transactions.truncate(max_sent_transactions());
```
teor2345, line 152: Note: Removing redundant code like this is great. But if you remember, please mention it in the PR description, so reviewers don't have to search for the reason for the code change.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -202,15 +198,14 @@ pub async fn run() -> Result<()> {
...
-    //    "all transactions from future blocks failed to send to an isolated mempool"
-    //);
+    // This check will fail if there are no non-coinbase transactions in the first 50 non-finalized blocks
+    assert!(
+        counter >= 1,
+        "all transactions from future blocks failed to send to an isolated mempool"
+    );
 
     // GetMempoolTx: make sure at least one of the transactions were inserted into the mempool.
     tracing::info!("calling GetMempoolStream gRPC to fetch transactions...");
```
teor2345, line 239: Optional: Does updating the service file also allow the `GetMempoolStream` check to be re-enabled?

zebrad/tests/common/lightwalletd/proto/service.proto
```diff
@@ -36,7 +36,7 @@ message TxFilter {
 // by GetMempoolStream(), the latest block height.
 message RawTransaction {
```
teor2345, line 37: Updating this file seems like something that's easy to forget. Is there somewhere we can document that it needs to be done?

teor2345 (dismissed):
Thank you for this fix. Is there a way to avoid this mistake next time we update `lightwalletd` forks?

zebrad/tests/common/lightwalletd/proto/service.proto
arya2: Added a comment in https://github.com/ZcashFoundation/zebra/pull/7644/commits/b4e6761673620a5720f0bdcbe9b2ae27944cd43b.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: I knew I was forgetting something.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: No, but the other assertion failed and had the message, "developers: should fail if `get_mempool_stream` start working": 

https://github.com/ZcashFoundation/zebra/pull/7600/files#diff-91452b1b52c32dc72de4dd903dd355f10ffc99e30d7217df595bb0177f88bd27L234-L237

teor2345 (approved):
This code looks good to me but CI seems to be having issues.

teor2345 (changes requested):
The test that this PR changes is still failing in CI, maybe we need to disable that part of the test:
https://github.com/ZcashFoundation/zebra/actions/runs/6433900141/job/17472517050?pr=7644#step:9:416

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -148,8 +147,21 @@ pub async fn run() -> Result<()> {
 
     let mut rpc_client = connect_to_lightwalletd(lightwalletd_rpc_port).await?;
 
-    // To avoid filling the mempool queue, limit the transactions to be sent to the RPC and mempool queue limits
-    transactions.truncate(max_sent_transactions());
+    // Call GetMempoolTx so lightwalletd caches the empty mempool state.
+    // This is a workaround for a bug where lightwalletd will skip calling `get_raw_transaction`
+    // the first time GetMempoolTx is called because it replaces the cache early and only calls the
+    // RPC method for transaction ids that are missing in the old cache as keys.
+    //
```
arya2, line 155:
```suggestion
    // <https://github.com/zcash/lightwalletd/blob/master/frontend/service.go#L495-L502>
    //
```

teor2345 (approved):
This seems to pass CI now

Issue #7541, opened 2023-09-13
devops: Align tests names in CI, Docker and Rust code


gustavovalverde:
## Motivation

Our test system is confusing. When you want to run a test, you need to understand three different naming systems:

1. **GitHub Workflows** use friendly names like `"Zebra checkpoint"` and `"lightwalletd tip update"`
2. **Docker** uses environment variables like `TEST_DISK_REBUILD=1` or `FULL_SYNC_MAINNET_TIMEOUT_MINUTES`
3. **Rust code** uses function names like `sync_to_mandatory_checkpoint_mainnet`

This makes simple things hard:

- **Finding tests**: You can't easily figure out which workflow runs which test
- **Debugging failures**: Tracing from a failed workflow to the actual Rust function takes forever
- **Running tests locally**: You need to guess the right environment variables
- **Adding new tests**: You have to update multiple files and remember the mapping rules
- **Getting started**: New people get lost trying to understand how tests work

**Example of the current complexity:**

```text
Workflow: "Zebra checkpoint"
    ↓
Docker: TEST_DISK_REBUILD=1,NETWORK=Mainnet
    ↓
Rust: sync_to_mandatory_checkpoint_mainnet
```

**Additional examples:**

- CI job: `test-empty-sync` → Rust test: `sync_large_checkpoints_mainnet`
- CI job: `test-lightwalletd-integration` → needs `ZEBRA_TEST_LIGHTWALLETD` + `TEST_LWD_INTEGRATION=1` → Rust test: `lightwalletd_integration`

This should just be one simple command.

## Current Test Mappings

Here's how everything maps together right now:

### Integration Tests (GCP)

| **Workflow Job**               | **test_id**           | **Environment Variables**                          | **Rust Function**                              |
| ------------------------------ | --------------------- | -------------------------------------------------- | ---------------------------------------------- |
| regenerate-stateful-disks      | sync-to-checkpoint    | TEST_DISK_REBUILD=1                                | sync_to_mandatory_checkpoint_mainnet/testnet   |
| test-stateful-sync             | sync-past-checkpoint  | TEST_CHECKPOINT_SYNC=1                             | sync_past_mandatory_checkpoint_mainnet/testnet |
| test-full-sync                 | full-sync             | FULL_SYNC_MAINNET_TIMEOUT_MINUTES                  | full_sync_mainnet                              |
[...]

PR #7660, opened 2023-10-02
ref(docker): combine test and prod entrypoints into one

### LINKED ISSUE #7639: devops: Use the same `entrypoint.sh` for all Docker stages
### Describe the issue or request

We're using a different entrypoint for tests and for our user image, and both have different behaviors and approaches when running Zebra. This causes some tests to behave or be called differently, not allowing a single standard way to interact with the resulting artifact.

### Expected Behavior

Handle production and non-production use-cases with the same parameters/variables.

### Current Behavior

Some tests fail if we use the same approach as with the production image, like passing the `NETWORK` environment variable, for example:
- https://github.com/ZcashFoundation/zebra/actions/runs/6323068078/job/17170285661

This is caused by the entrypoint, as one considers this changes, and the other doesn't.

### Possible Solution

Have a single entrypoint, with all the scenarios combined.

### Additional Information/Context

_No response_

### Is this happening on PRs?

Just on #7616

### Is this happening on the main branch?

No

gustavovalverde:
## Motivation

Maintaining both entrypoint is not necessary, and having a single one help with refactors like:
- https://github.com/ZcashFoundation/zebra/pull/7616 

Fixes: #7639

- A whole file disappears and it's merged into another
- Some extra changes were added to make the whole file less verbose and fix linting issues

## Solution

- Uses `-x` instead of echoing the variables values
- Sets default values where required
- Create a function to list directories
- Create a function to run cargo tests
- Use a better approach to handle different options in the `case` manegement for tests and production
- Replaces all instances of `runtime-entrypoint.sh` with `entrypoint.sh`

## Review

- Verify all tests are passing in this PR
- A long lived instance must be deployed from this PR to validate CD configuration

### Reviewer Checklist

## Follow Up Work

Use arguments instead of variables, to run tests, based on the following comment:
- https://github.com/ZcashFoundation/zebra/issues/7541#issuecomment-1717060738


teor2345:
It might help to run a full sync with these script changes, to check that it works. But that's also a very slow test. What do you think?

I don't mind either way, as long as the full sync is working on `main`, we can wait for the end of the week. (But if it's not working on `main` that's a blocker.)

docker/entrypoint.sh
```diff
@@ -1,128 +1,292 @@
...
+        # List directory used by test
+        list_directory "${ZEBRA_CACHED_STATE_DIR}"
+        run_cargo_test "${ENTRYPOINT_FEATURES}" "generate_checkpoints_testnet"
+
+      elif [[ "${TEST_LWD_RPC_CALL}" -eq "1" ]]; then
+        # Starting at a cached Zebra tip, test a JSON-RPC call to Zebra.
+        list_directory "${ZEBRA_CACHED_STATE_DIR}"
+        # Run both the fully synced RPC test and the subtree snapshot test, one test at a time.
+        # Since these tests use the same cached state, a state problem in the first test can fail the second test.
+        run_cargo_test "${ENTRYPOINT_FEATURES}" "fully_synced_rpc_"
```
teor2345: This test command is missing the test-threads argument. It's ok to add that argument to run_cargo_test, because all the other tests only run one test function. (So the number of threads doesn't matter.)

gustavovalverde: Thank you! Fixed with [fix: add missing test-threads to fully_synced_rpc_](https://github.com/ZcashFoundation/zebra/pull/7660/commits/0f0c175ded0f1e79a20a9f9dcb5f193b9aa838c5)

gustavovalverde:
> It might help to run a full sync with these script changes, to check that it works. But that's also a very slow test. What do you think?

If other tests are passing, it should be the same behavior with the full sync. If my any means this causes an odd behavior (which I wouldn't expect) it should be quick to fix or revert.

teor2345:
> > It might help to run a full sync with these script changes, to check that it works. But that's also a very slow test. What do you think?
> 
> If other tests are passing, it should be the same behavior with the full sync. If my any means this causes an odd behavior (which I wouldn't expect) it should be quick to fix or revert.

Ok, but the full sync hasn't passed since 1 September, before the self-hosted runners changes. So did you want to wait for a successful full sync on `main` before merging these changes?

I am concerned we might end up with a stack of changes which might all contribute to the ongoing failures, and it's difficult to work out which ones are causing it. (Like we have in ticket #7618 in the Zebra Rust code right now!)

docker/entrypoint.sh
```diff
@@ -246,7 +246,7 @@ case "$1" in
         list_directory "${ZEBRA_CACHED_STATE_DIR}"
         # Run both the fully synced RPC test and the subtree snapshot test, one test at a time.
         # Since these tests use the same cached state, a state problem in the first test can fail the second test.
-        run_cargo_test "${ENTRYPOINT_FEATURES}" "fully_synced_rpc_"
+        run_cargo_test "${ENTRYPOINT_FEATURES}" "--test-threads 1 fully_synced_rpc_"
```
teor2345: This is probably going to cause argument quoting issues, because `cargo test ... "--test-threads 1 fully_synced_rpc_"` is one argument, but we need `cargo test ... --test-threads 1 fully_synced_rpc_` (3 arguments).

If this test fails, I'd suggest adding the arguments inside the `run_cargo_test()` function. And if it doesn't fail, let's check the test logs to make sure it's actually running.

Here's an example of the difference:
```sh
$ cargo test -- --test-threads 1 test_name
    Finished test [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests src/main.rs (target/debug/deps/empty_project-dd6fff5bb63559a4)

running 1 test
test test_name ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ cargo test -- "--test-threads 1 test_name"
    Finished test [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests src/main.rs (target/debug/deps/empty_project-dd6fff5bb63559a4)
error: Unrecognized option: 'test-threads 1 test_name'
error: test failed, to rerun pass `--bin empty-project`
Exit 101
```

gustavovalverde:
> So did you want to wait for a successful full sync on main before merging these changes?

Oh yeah, that's ok, this doesn't has to merged beforehand as we'll all be focusing on fixing the full sync test.

In another note, the long-live instance is working.

Here's the workflow run: https://github.com/ZcashFoundation/zebra/actions/runs/6386506331

And the instance logs.
image

docker/entrypoint.sh
gustavovalverde: Fixed with [fix: allow 3 arguments instead of 2 to handle fully_synced_rpc_ cor…](https://github.com/ZcashFoundation/zebra/pull/7660/commits/fed758bc44cb1e73620dc82b7c3536a8f61f8a72)

.github/workflows/ci-sync-tests-gcp.yml
```diff
@@ -0,0 +1,616 @@
+name: CI Docker
```
teor2345: This is waiting for a successful full sync, which probably needs some other fixes.

teor2345: I really think we should do a full sync, unless this change is urgent?
If we make multiple significant changes to CI, it will be hard to work out which code is causing any failures.

A Zebra testnet and lightwalletd mainnet full sync should be enough, both of them complete within about 8 hours.

.github/workflows/ci-sync-tests-gcp.yml
```diff
@@ -0,0 +1,616 @@
...
+      test_variables: '-e NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }} -e TEST_SUBMIT_BLOCK=1 -e ZEBRA_FORCE_USE_COLOR=1 -e ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache'
+      needs_zebra_state: true
+      needs_lwd_state: false
+      saves_to_disk: false
+      disk_suffix: tip
+      root_state_path: '/var/cache'
+      zebra_state_dir: 'zebrad-cache'
+    secrets: inherit
+
+  failure-issue:
```
teor2345: The dependencies of failure-issue need to be updated.

docker/entrypoint.sh
```diff
@@ -1,128 +1,292 @@
 #!/usr/bin/env bash
```
teor2345, line 1: Optional: some developers have asked for a short summary of what each CI / develops file does at the top of the file. One sentence is enough.

docker/entrypoint.sh
```diff
@@ -1,128 +1,292 @@
...
-        fi
+  --* | -*)
+    if [[ -n "${ZEBRA_CONF_PATH}" ]]; then
+        exec zebrad -c "${ZEBRA_CONF_PATH}" "$@" || { echo "Execution with custom configuration failed"; exit 1; }
+    else
+        exec zebrad "$@" || { echo "Execution failed"; exit 1; }
+    fi
+    ;;
+  "")
+    if [[ -n "${ZEBRA_CONF_PATH}" ]]; then
```
teor2345: It looks like the ZEBRA_CONF_PATH is always set, so all the test workflows are running zebrad with the default config, rather than running the test commands below.

Maybe the test environment variable checks could go first, before we even build the default config?
(Generating the default config and logging it during the tests will make them harder to debug, because developers will assume the default config is being used by the tests.)

.github/workflows/ci-sync-tests-gcp.yml
gustavovalverde: This file was not meant to be here.

docker/entrypoint.sh
gustavovalverde: Fixed here [chore: add comments to main parts of the file](https://github.com/ZcashFoundation/zebra/pull/7660/commits/9f7a92cd4a0c3e1839df30f9f3416e3c3dbb3e81)

docker/entrypoint.sh
gustavovalverde: Thanks! This might solve that: [ref: handle tests better if $ENTRYPOINT_FEATURES is set](https://github.com/ZcashFoundation/zebra/pull/7660/commits/93c4077792f3f68947cb26780d087f6def1bf05d)

arya2 (dismissed):

arya2 (approved):

Issue #7661, opened 2023-10-03
Fix invalid cached states which are making `lightwalletd` tests fail


mpguerra:
## Motivation

It looks like the lightwalletd tests are failing because a previous lightwalletd job didn't write a full cache:

```
{"app":"lightwalletd","level":"info","msg":"Done reading 88745 blocks from disk cache","time":"2023-09-30T21:11:05Z"}
{"app":"lightwalletd","level":"info","msg":"Adding block to cache 509062
```

https://github.com/ZcashFoundation/zebra/actions/runs/6365517114/job/17282788077?pr=7636#step:9:228

## Suggested Fix

- [x] Find all the lightwalletd cached states with heights less than 2 million
- [ ] Work out which PR, branch, and CI job created them, and open a ticket to fix it
- [ ] Optional: stop creating new invalid images, by stopping running CI on that PR, or disabling that scheduled CI job (if that's possible)
- [x] Delete all the lightwalletd cached states with heights less than 2 million (they are all invalid)
- [ ] Repeat these steps until the original bug is fixed, merged, and all PRs are updated with that fix

### Background

This ticket has some notes about how we fixed this last time: [dev docs: Document how to fix cached state issues](https://github.com/ZcashFoundation/zebra/issues/7558)

gustavovalverde:
Deleted this disk created by:
- https://github.com/ZcashFoundation/zebra/commit/6f50304
```
Description: Created from commit 6f50304 with height 506458 and database format v25-1-1
Selected Disk: lwd-cache-v1-2-0-6f50304-v25-mainnet-tip-031355
```

teor2345:
Since this cached state was created by Zebra `v1.2.0`, and we've already fixed bug #7335 in the `main` branch, we don't need to do anything more here.

PR #7662, opened 2023-10-03
fix(ci): wait for disk to be mounted in VM

### LINKED ISSUE #7659: CI failing with mounting volume error
https://github.com/ZcashFoundation/zebra/actions/runs/6382508836/job/17321768540?pr=7653#step:13:188

```
docker: Error response from daemon: error while mounting volume '/var/lib/docker/volumes/fully-synced-rpc-85b855b/_data': failed to mount local volume: mount /dev/sdb:/var/lib/docker/volumes/fully-synced-rpc-85b855b/_data: device or resource busy.
Error: Process completed with exit code 125.
```

This seems to be happening for every PR we have opened *recently*. We can post more links to failures and more information in this ticket.

gustavovalverde:
## Motivation

Lately, mounting the disk by Docker has been unreliable, the root cause might be different things, but in the end we should confirm the disk is mounted before running the container with disk being mounted or _busy_

Fixes: #7659

## Solution

Add a wait command: `while [[ ! -e /dev/sdb ]]; do sleep 1; done`

## Review

We might want to run this PR multiple times if it works at first, just to confirm multiple runs (4+) will output the same result.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -432,6 +432,8 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command \
           "\
+          # Wait for the disk to be attached
+          while [[ ! -e /dev/sdb ]]; do sleep 1; done && \
           sudo docker run \
           --name ${{ inputs.test_id }} \
           --tty \
@@ -480,6 +482,8 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command \
           "\
+          # Wait for the disk to be attached
+          while [[ ! -e /dev/sdb ]]; do sleep 1; done && \
           sudo docker run \
           --name ${{ inputs.test_id }} \
           --tty \


.github/workflows/deploy-gcp-tests.yml
```diff
@@ -432,6 +432,8 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command \
           "\
+          # Wait for the disk to be attached
+          while [[ ! -e /dev/sdb ]]; do sleep 1; done && \
```
teor2345, line 436: Optional: Can we do `set -x` at the start of these scripts, so if there is a hang, we can see which script line it is on?

teor2345 (approved):
Thanks for the quick fix, looks good to me!

I am happy to try it on the `main` branch, and do extra fixes in another PR if needed.

teor2345:
This is failing due to bug #7661, so I think we should admin-merge it, so we can fix the `lightwalletd` cached state and `zebrad` sync bugs.

.github/workflows/deploy-gcp-tests.yml
gustavovalverde: I'll apply this later on another PR

gustavovalverde:
Admin-merged

PR #7665, opened 2023-10-04
fix(ci): disk validation for docker volume mount

### LINKED ISSUE #7659: CI failing with mounting volume error

gustavovalverde:
## Reviewer Instructions

Since this change is urgent, please put suggested cleanups in ticket #7677.

## Motivation

PR #7662 did not work as expected, as the validation was too simplistic.

@teor2345 also found that Google recently updated their `cos-stable` image and this upgraded Docker considerably (by 4 major versions), which might be the reason of this appearing recently with no changes to the code. 

Fixes: #7659

## Solution

- Revert to use `busybox` instead of `ubuntu`, as this was causing errors in the VM logs
- Before formatting, validate the disk is not being used with `lsof`
- Do not mount the Docker volume before running Docker. Instead mount it in the `docker run` command
  -  Based on new Docker versions, this approach can now be used: https://docs.docker.com/storage/volumes/#how-mounting-block-devices-works

## Review

- We did not run this multiple times in the previous PR, we should re-run the tests multiple times to confirm.
- Since this change is urgent, please put suggested cleanups in ticket #7677

### Reviewer Checklist

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -193,15 +193,15 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command \
           " \
-          set -x \
+          set -x; \
           while ! \
           sudo mkfs.ext4 -v /dev/sdb \
           && \
           sudo docker volume create --driver local --opt type=ext4 --opt device=/dev/sdb \
           ${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }}; \
           do \
-          sudo journalctl -u docker.service \
-          sleep 5 \
+          sudo journalctl -u docker.service; \
+          sleep 5; \
           done \
           "
 
@@ -215,7 +215,7 @@ jobs:
           --ssh-flag="-o ConnectTimeout=5" \
           --command \
           "\
-          set -x \
+          set -x; \
           while ! \
           sudo docker run \
           --name ${{ inputs.test_id }} \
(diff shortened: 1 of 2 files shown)


teor2345:
@gustavovalverde since this is an urgent fix, do you mind if I work on it today?
It would be helpful to get it merged before the full sync jobs start tomorrow, so we can test all these bug fixes.

teor2345:
## Reviewer Instructions

Since this change is urgent, please put suggested cleanups in ticket #7677

gustavovalverde:
@teor2345 we can also output the Docker daemon logs after the failure, to confirm if it gives more information about the error:

```shell
sudo journalctl -u docker.service

```

teor2345:
> @teor2345 we can also output the Docker daemon logs after the failure, to confirm if it gives more information about the error:
> 
> ```shell
> sudo journalctl -u docker.service
> ```

Thanks, I gave this a try.

teor2345:
I'm currently retrying the commands if they fail, I also need to delete the containers that were created without the correct mounts.

teor2345:
I'm trying lsof for debugging. If that doesn't work, fuser, losetup, or lsof | grep a_inode might help:
https://stackoverflow.com/questions/624154/linux-which-process-is-causing-device-busy-when-doing-umount/45782183#45782183

This could be caused by an automatic fsck:
https://unix.stackexchange.com/questions/179855/device-already-mounted-or-resource-is-busy/342464#342464

teor2345:
So far lsof isn't showing much: there aren't any files open on the device. Maybe one of the other commands will help more.

gustavovalverde:
I canceled [fix: remove extra ;] (https://github.com/ZcashFoundation/zebra/pull/7665/commits/3c29dab5c349acf3ac44dea4ce5c446ba91daef3) with a new commit

But it did not fail to mount the volumes

oxarbitrage (approved):

PR #7682, opened 2023-10-05
ref(workflows): use a single job to run GCP tests

gustavovalverde:
## Motivation

To be able to handle edge-cases and complex scenarios (launch step no failing, but test-result failint and not being able to re-run it), and for the sake of simplicity, this PR proposes using a single job for handling all the following scenarios:
- Test without cached states
- Test with Zebra cached state
- Test with Lightwalletd cached state

This is a very delicate change, even though is simplifying the process, as it has to consider multiple scenarios which were using multiple jobs and steps, and now it's using a single job and some steps have been merged.

## Solution

- Use variables like `$DISK_OPTION` and `$MOUNT_FLAGS` to allow flexibility when mounting the volumes.

## Review

- We should run most tests across this changes

### Reviewer Checklist

## Follow Up Work

Move the scripts out of the workflow.


teor2345 (dismissed):
This needs a bug fix in the base branch.

gustavovalverde:
This should be good to go!

teor2345:
Would it help to check these workflows manually, to make sure they are still using the expected cached states? And writing the expected cached states?

What's the trickiest part of this change? (Or the most likely bug in it.)

.github/workflows/ci-integration-tests-gcp.yml
```diff
@@ -596,7 +596,7 @@ jobs:
     #
     # This list is for reliable tests that are run on the `main` branch.
     # Testnet jobs are not in this list, because we expect testnet to fail occasionally.
-    needs: [ regenerate-stateful-disks, test-full-sync, lightwalletd-full-sync, test-stateful-sync, test-update-sync, checkpoints-mainnet, lightwalletd-update-sync, lightwalletd-rpc-test, lightwalletd-transactions-test, lightwalletd-grpc-test, get-block-template-test, submit-block-test ]
+    needs: [ regenerate-stateful-disks, test-stateful-sync, test-full-sync, test-update-sync, checkpoints-mainnet, test-full-sync-testnet, checkpoints-testnet, lightwalletd-full-sync, lightwalletd-update-sync, lightwalletd-rpc-test, lightwalletd-transactions-test, lightwalletd-grpc-test, get-block-template-test, submit-block-test ]
```
teor2345: Optional: This change seems to just be re-ordering the jobs, can we remove it from this PR?
(This applies to the next change in the PR diff as well.)

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -473,8 +304,8 @@ jobs:
       # VM and to the container might require more steps in this workflow, and additional
       # considerations.
       #
-      # The disk mounted in the VM is located at /dev/$DISK_NAME, we want the root `/` of this disk to be
-      # available in the docker container at two different paths:
+      # The disk mounted in the VM is located at /dev/sdb, we mount the root `/` of this disk to the docker
+      # container, and might have two different paths (if lightwalletd state is needed):
```
teor2345: Nitpick: This change seems to be a mistake in the merge, the old docs were correct.

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -484,44 +315,47 @@ jobs:
       # subdirectories for their data. (But Zebra, lightwalletd, and the test harness must not
       # delete the whole cache directory.)
       #
-      # This paths must match the variables used by the tests in Rust, which are also set in
+      # This path must match the variable used by the tests in Rust, which are also set in
```
teor2345: Nitpick: Some tests use a Zebra and lightwalletd variable, so the plural is correct

```suggestion
      # These paths must match the variables used by the tests in Rust, which are also set in
```

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -484,44 +315,47 @@ jobs:
...
-          set -ex;
+
           # Extract the correct disk name based on the device-name
-          export DISK_NAME=$(ls -l /dev/disk/by-id | grep -oE "google-${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }} -> ../../[^ ]+" | grep -oE "/[^/]+$" | cut -c 2-); \
+          DISK_NAME=$(ls -l /dev/disk/by-id | grep -oE "google-${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }} -> ../../[^ ]+" | grep -oE "/[^/]+$" | cut -c 2-)
+
+          MOUNT_FLAGS="--mount type=volume,volume-driver=local,volume-opt=device=/dev/$DISK_NAME,volume-opt=type=ext4,dst=${{ inputs.root_state_path }}/${{ inputs.zebra_state_dir }}"
+
+          # Check if we need to mount for Lightwalletd state
+          if [[ "${{ inputs.needs_lwd_state }}" == "true" || "${{ inputs.test_id }}" == "lwd-full-sync" ]]; then
```
teor2345: Nitpick: This comment accidentally got deleted rather than being moved.

```suggestion
          # Check if we need to mount for Lightwalletd state
          # lightwalletd-full-sync reads Zebra and writes lwd, so it is handled specially.
          if [[ "${{ inputs.needs_lwd_state }}" == "true" || "${{ inputs.test_id }}" == "lwd-full-sync" ]]; then
```

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -535,58 +369,6 @@ jobs:
           sudo journalctl -b \
```
teor2345, line 370: 2 lines above this line we use `$DISK_NAME`, but it hasn't been written to `$GITHUB_ENV` by any previous step.

teor2345 (commented):
This looks good! Thank you for this cleanup.

Is there a description of how the workflow works now? Can we add that to the top of this workflow?

It looks like you merged all 3 of the cases in the PR description into the same job, but also made some other changes to simplify pipe handling. Is there any other significant change here?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -435,36 +288,14 @@ jobs:
...
-          --ssh-flag="-o ServerAliveInterval=5" \
-          --ssh-flag="-o ConnectionAttempts=20" \
-          --ssh-flag="-o ConnectTimeout=5" \
-          --command=' \
-          lsblk;
-          sudo lsof /dev/$DISK_NAME;
-          sudo dmesg;
-          sudo journalctl -b \
+          DISK_NAME=$(ls -l /dev/disk/by-id | grep -oE "google-${{ inputs.test_id }}-${{ env.GITHUB_SHA_SHORT }} -> ../../[^ ]+" | grep -oE "/[^/]+$" | cut -c 2-);
+          sudo mkfs.ext4 -v /dev/$DISK_NAME \
```
teor2345, line 292: This change might need a full sync test, I think the fastest test that uses this script is a testnet full sync.

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: Yeah, this comes from the main branch. I'll create another PR to improve this. I don't think using $GITHUB_ENV will be the right approach, as this is executing in the the GCP VM context.

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: This behavior is already in the `main` branch: https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/deploy-gcp-tests.yml#L196-L197

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: Fixed with: [fix(ci): use better comment](https://github.com/ZcashFoundation/zebra/pull/7682/commits/a32b3742ebf75221d3a60d7a1c033c7fe3e0a789)

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: By the way, I can't trigger a sync test unless we merge (as these are "new" workflows) https://github.com/ZcashFoundation/zebra/actions/workflows/ci-integration-tests-gcp.yml

teor2345: So do we want to delay this large CI change until after the release?
I'll leave it to you and @upbqdn to decide, because he is the release manager.

gustavovalverde: I'd prefer not to as this PR won't have an impact on the built image. But I don't mind if @upbqdn prefers not to include this PR; that's ok for me as we can merge it just after the release. We can have it approved and labeled to not merge. Whatever is less traumatic.

teor2345: My concern was more about unexpected CI failures delaying the release, but I also think we've done good testing here.

upbqdn: This release will be big enough even without this change, so let's delay?

teor2345:
I removed the `Run lwd-full-sync-test` branch protection rule because it very rarely runs in PRs, it runs in a scheduled job now.

teor2345 (approved):

teor2345:
@gustavovalverde I just saw this $DISK_NAME fix follow-up again in my final review:
https://github.com/ZcashFoundation/zebra/pull/7682/files#r1357377327

It would be ok to copy the same script you use to find the disk name into that SSH command.

PR #7684, opened 2023-10-05
Increase cached state disk sizes to 400 GB

teor2345:
## Motivation

We're at 280 GB of 300 GB, and getting much closer could cause weird errors or intermittent failures.

## Solution

Increase cached state disk sizes to 400 GB.

We don't need big boot disks, they just cost us money, so decrease them to 50 GB.

## Review

This PR only needs to be merged within the next few months, so it's a low priority.

### Reviewer Checklist


teor2345:
Failed due to issue #7659, I've restarted it.

gustavovalverde (approved):
LGTM

PR #7688, opened 2023-10-06
fix(ci): Don't wait before launching GCP tests

teor2345:
## Motivation

Removing the wait times in the job seems to make it more reliable.

## Review

This is urgent.

### Reviewer Checklist


teor2345:
These runs were successful, except for the known issue with the checkpoint rebuild:
- https://github.com/ZcashFoundation/zebra/actions/runs/6426622891/attempts/1?pr=7688
- https://github.com/ZcashFoundation/zebra/actions/runs/6426622891/attempts/2?pr=7688
    - TODO: check for failures
- TODO: do a third run

This seems to be better than the current `main` branch, so maybe it's worth merging this change.

teor2345:
This doesn't seem to help consistently

PR #7690, opened 2023-10-08
fix(ci): handle disk mounting and logs reading edge-cases

gustavovalverde:
## Motivation

Some tests are still failing with `tee: 'standard output': Broken pipe`, recently it has been more frequently with the `sync-to-checkpoint` test, but it could happen on other tests too.

Also, mounting errors are still happening, and we'd like to capture the `dmesg` when this happens at `launch`, for better debugging. Other options like `set -x` will also help debugging and replicating some commands locally.

**Edit:**
In the end we had this different outputs for disks from the same VM:
```shell
NAME      MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda         8:0    0  300G  0 disk 
sdb         8:16   0   50G  0 disk 
|-sdb1      8:17   0 45.8G  0 part /var/lib/toolbox
|                                  /var/lib/google
|                                  /var/lib/docker
|                                  /var/lib/containerd
|                                  /var
|                                  /home
|                                  /mnt/stateful_partition
|-sdb2      8:18   0   16M  0 part 
|-sdb3      8:19   0    2G  0 part 
| `-vroot 253:0    0  1.9G  1 dm   /
|-sdb4      8:20   0   16M  0 part 
|-sdb5      8:21   0    2G  0 part 
|-sdb6      8:22   0  512B  0 part 
|-sdb7      8:23   0  512B  0 part 
|-sdb8      8:24   0   16M  0 part /usr/share/oem
|-sdb9      8:25   0  512B  0 part 
|-sdb10     8:26   0  512B  0 part 
|-sdb11     8:27   0    8M  0 part 
`-sdb12     8:28   0   32M  0 part 
```

```shell
NAME      MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda         8:0    0   50G  0 disk 
|-sda1      8:1    0 45.8G  0 part /var/lib/toolbox
|                                  /var/lib/google
|                                  /var/lib/docker
|                                  /var/lib/containerd
|                                  /var
|                                  /home
|                                  /mnt/stateful_partition
|-sda2      8:2    0   16M  0 part 
|-sda3      8:3    0    2G  0 part 
| `-vroot 253:0    0  1.9G  1 dm   /
|-sda4      8:4    0   16M  0 part 
|-sda5      8:5    0    2G  0 part 
|-sda6      8:6    0  512B  0 part 
|-sda7      8:7    0  512B  0 part 
|-sda8      8:8    0   16M  0 part /usr/share/oem
[...]

PR #7703, opened 2023-10-09
fix(ci): Retry launching failed docker test instances

teor2345:
## Motivation

We can't merge PRs because sometimes launching the Docker test image fails.

This PR stops #7659 being a critical issue by allowing PRs to merge anyway. So we can keep that ticket open, but fix it over the next week or two while still merging other PRs.

### Action Reference

The `retry-step` arguments are listed here:
https://github.com/marketplace/actions/retry-step

I deleted some parts of the script that seemed to be redundant, and I deleted the redundant "no `lightwalletd path`" step. Instead, we always mount both paths, some tests don't use the `lightwalletd` path.

(It's also possible for tests to use the `lightwalletd` paths as temporary storage, but we never see them do it because their states aren't written as a cache image.)

## Solution

- Make create instance and launch Docker test into a single step
- Retry that step up to 10 times if it fails

## Review

This is blocking all the other PRs merging reliably. It's not a long-term solution, but it would let us merge other PRs while we're fixing the actual bug.

To reduce the diff, I used one-space indentation rather than two-space indentation. I'll fix that in another PR.

### Reviewer Checklist

## Follow Up Work

Fix the underlying bug, and once we're sure it's fixed, remove the retries.

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -167,23 +167,25 @@ jobs:
       - name: Launch ${{ inputs.test_id }} GCP test instance
         id: launch-instance
         uses: nick-fields/retry@v2.9.0
-        # This step usually finishes in around 5 minutes.
-        timeout_minutes: 10
-        # We use a large number of attempts because the failure rate can be up to 50%.
-        max_attempts: 10
-        # We want to fail on timeouts but retry on errors.
-        retry_on: error
-        # It can take a few minutes to delete a container, but the command should wait.
-        retry_wait_seconds: 60
-        # These commands are slow, so we don't need to check their results every second.
-        polling_interval_seconds: 10
-        # TODO: shell: bash (arguments)
-        # TODO: make a cleanup script and call it from both launch steps and delete instance.
-        on_retry_command: |
+        with:
+         # TODO: fix intendation
+         # This step usually finishes in around 5 minutes.
+         timeout_minutes: 10
+         # We use a large number of attempts because the failure rate can be up to 50%.
+         max_attempts: 10
+         # We want to fail on timeouts but retry on errors.
+         retry_on: error
+         # It can take a few minutes to delete a container, but the command should wait.
+         retry_wait_seconds: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-gcp-tests.yml
```diff
@@ -180,27 +201,8 @@ jobs:
...
-          --ssh-flag="-o ServerAliveInterval=5" \
-          --ssh-flag="-o ConnectionAttempts=20" \
-          --ssh-flag="-o ConnectTimeout=5" \
-          --command \
-          "\
-          while sudo lsof /dev/sdb; do \
-            echo 'Waiting for /dev/sdb to be free...'; \
-            sleep 10; \
-          done; \
-          sudo mkfs.ext4 -v /dev/sdb \
```
teor2345, line 198: If waiting for the disk or initialising the disk is needed, we can add it inside the ssh command below.

teor2345:
If we need to wait or initalise the `without-cached-state` disk, we can do that inside the ssh command. But I'm not sure if that's needed, because there's no reference to `/dev/sdb` until we initialise it.

Using block devices also isn't recommended, so that could be part of our issue here:
https://docs.docker.com/storage/volumes/#block-storage-devices

Here's a manual Zebra full sync for testnet, and lightwalletd full sync for mainnet:
- https://github.com/ZcashFoundation/zebra/actions/runs/6453236984
- https://github.com/ZcashFoundation/zebra/actions/runs/6453240025

<details>

<summary>older buggy syncs</summary>

Here's a manual Zebra full sync for testnet, and lightwalletd full sync for mainnet, with a shell command bug!
- https://github.com/ZcashFoundation/zebra/actions/runs/6452973318
- https://github.com/ZcashFoundation/zebra/actions/runs/6452974255

Here's a manual Zebra full sync for testnet, and lightwalletd full sync for mainnet, with a bug that stops the checkpoint syncs working:
- https://github.com/ZcashFoundation/zebra/actions/runs/6452808563

</details>

gustavovalverde:
This should no longer be needed after #7690

PR #7704, opened 2023-10-09
fix(ci): wait for mounting and show system logs if fail

gustavovalverde:
## Motivation

Disk mounting is failing. It seems to be related to the disk not being ready, and we might just need to wait longer. But this might help confirming if that's the case.

## Solution

- Wait for the disk to be ready before mounting docker
-  If even after waiting the disk fails to be mounted, print all required logs

### Reviewer Checklist


upbqdn (commented):
The PR description says:

> Wait for the disk to be ready before mounting docker

Where does the waiting happen?

PR #7713, opened 2023-10-09
fix(ci): `grep` the Zcash network correctly

gustavovalverde:
## Motivation

Some tests fail to grep `Zcash network: ${{ inputs.network }}`, and fail, timeout or take longer than needed.

## Solution

- Use the same approach as in the `Result of ${{ inputs.test_id }} test` step, which is handling odd use cases nicely.

## Review

- Validate the startup logs take less than 1-2 minutes

### Reviewer Checklist


teor2345 (approved):
Thanks for this fix!

PR #7722, opened 2023-10-10
fix(ci): do not keep logging if `grep` times out or finds the string

gustavovalverde:
## Motivation

Even though we tried to fix this in #7713, it was still not working on that PR.

## Solution

- Use `>(...)` which is a form of process substitution in Bash, which avoids the "broken pipe" error as the `tee` command isn't directly writing to a closed pipe

## Review

If added a 5 min timeout here, so if this passes it should be good to go. 

### Reviewer Checklist

### DIFF
--- a/.github/workflows/deploy-gcp-tests.yml
+++ b/.github/workflows/deploy-gcp-tests.yml
@@ -595,11 +595,14 @@ jobs:
       # lines before the startup logs. So that's what we use here.
       #
       # The log pipeline ignores the exit status of `docker logs`.
-      # It also ignores the expected 'broken pipe' error from `tee`,
-      # which happens when `grep` finds a matching output and moves on to the next job.
+      #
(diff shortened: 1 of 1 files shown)


upbqdn (approved):

gustavovalverde:
I changed this back to Draft as it's still not working.

I'll be testing a new approach locally and in CI using other Docker options for logging
- https://docs.docker.com/engine/reference/commandline/logs/#options
- https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers

teor2345:
If we install `ripgrep` on our Google Cloud machine, we could use `rg --passthru ...` rather than `tee ... | grep ...`.

`rg` accepts the same arguments as `grep`.

mpguerra:
@gustavovalverde do we still need this one?

PR #7724, opened 2023-10-10
fix(ci): Handle new cargo machete output format for unused dependencies

teor2345:
## Motivation

`cargo machete` changed its output format, so we need to change our CI to process it.

Until we do that, every PR will fail to merge with an unused dependencies error.

## Solution

Just include the dependency lines, rather than trying to exclude the other output lines.

## Review

This is critical because it blocks everything merging.

### Reviewer Checklist

### DIFF
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -312,8 +312,8 @@ jobs:
           echo "-- full cargo machete output, including ignored dependencies --"
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
-          if (cargo machete --skip-target-dir 2>/dev/null || true) | \
-          (grep -e '^\t' || true) | \
+          if cargo machete --skip-target-dir 2>/dev/null | \
+          grep --extended-regexp -e '^\t' | \
           grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils; then
               echo "New unused dependencies were found, please remove them!"
               exit 1
--- a/.github/workflows/continous-integration-os.yml
+++ b/.github/workflows/continous-integration-os.yml
@@ -313,7 +313,7 @@ jobs:
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
           if (cargo machete --skip-target-dir 2>/dev/null || true) | \
-          grep -e '\t' | \
+          grep -e '^\t' -e '^ ' | \
           grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils; then
               echo "New unused dependencies were found, please remove them!"
               exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/continous-integration-os.yml
```diff
@@ -313,7 +313,8 @@ jobs:
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
           if (cargo machete --skip-target-dir 2>/dev/null || true) | \
-          grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils -e "found the following" -e Cargo.toml -e Done; then
+          grep -e '\t' | \
```
teor2345:
```suggestion
          grep -e '^\t' -e '^ ' | \
```

.github/workflows/continous-integration-os.yml
```diff
@@ -313,7 +313,8 @@ jobs:
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
           if (cargo machete --skip-target-dir 2>/dev/null || true) | \
-          grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils -e "found the following" -e Cargo.toml -e Done; then
+          grep -e '^\t' -e '^ ' | \
```
teor2345:
```suggestion
          (grep -e '^\t' || true) | \
```

.github/workflows/continous-integration-os.yml
```diff
@@ -313,7 +313,8 @@ jobs:
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
           if (cargo machete --skip-target-dir 2>/dev/null || true) | \
```
teor2345:
```suggestion
          if cargo machete --skip-target-dir 2>/dev/null | \
```

.github/workflows/continous-integration-os.yml
```diff
@@ -313,7 +313,8 @@ jobs:
...
-          grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils -e "found the following" -e Cargo.toml -e Done; then
+          (grep -e '^\t' || true) | \
```
teor2345:
```suggestion
          grep --extended-regexp -e '^\t' | \
```

.github/workflows/continous-integration-os.yml
```diff
@@ -312,8 +312,9 @@ jobs:
           echo "-- full cargo machete output, including ignored dependencies --"
           cargo machete --skip-target-dir || true
           echo "-- unused dependencies are below this line, full output is above --"
-          if (cargo machete --skip-target-dir 2>/dev/null || true) | \
-          grep -v -e gumdrop -e humantime-serde -e tinyvec -e zebra-utils -e "found the following" -e Cargo.toml -e Done; then
+          if cargo machete --skip-target-dir 2>/dev/null | \
+          grep --extended-regexp -e '^\t' | \
```
teor2345:
```suggestion
          grep --extended-regexp -e '^\\t' | \
```

teor2345:
I admin-merged this urgent CI fix, because it was a very simple workflow change.

PR #7731, opened 2023-10-11
change(state): Set iterator read bounds where possible in DiskDb

### LINKED ISSUE #7664: Set iterator read bounds on deleting column families to avoid very slow syncs
### Motivation

Iterating on a column family that deletes lots of keys can be very slow.

We fixed some of this code in PRs #7663 and #7392, but we need a general fix to stop this issue happening again.

This currently impacts the `utxo_loc_by_transparent_addr_loc` column family, and any future uses of iterators with `zs_delete()`. It can be remotely triggered.

### Specifications

The RocksDB documentation recommends specifying lower and upper bounds on iterators:
https://github.com/facebook/rocksdb/wiki/Iterator#iterating-upper-bound-and-lower-bound
https://docs.rs/rocksdb/latest/rocksdb/struct.ReadOptions.html#method.set_iterate_range
https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.iterator_cf_opt

This is how these bugs in other projects...
https://tracker.ceph.com/issues/55324
https://jira.mariadb.org/browse/MDEV-19670

Were fixed in those projects:
https://github.com/ceph/ceph/pull/46096/files
https://github.com/facebook/rocksdb/pull/5403

We need to use iterators for some operations, so our alternatives are (in preferred order):
1. Minimise the number of keys we delete, and how often we delete them (currently only UTXOs require key deletion)
2. Avoid using iterators on column families where we delete keys (currently only one column family does this, and it won't be necessary after PR #7392)
3. If we must use iterators on those column families, set read bounds to minimise the amount of deleted data that is read  (this ticket)

If this solution passes our benchmark tests, we should document it in:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md

### Testing

When we implement the test in ticket #7649, it will also cover this change.

### Documentation

We should document this issue in the developer state upgrade docs.

### Related Work

This is a long-term fix for #7618.

PR #7663 implements alternative 2 for some column families.
PR #7392 implements alternative 1 for some column families.

arya2:
## Motivation

Unbounded RocksDB iterators are slow in column families with many delete tombstones.

This PR updates `DiskDb` to bound iterators when reading a range of keys.

Close #7664.

## Solution

- Set lower and upper bounds on rocksdb iterators where possible in `DiskDb` methods

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Refactor `address_utxo_locations()`

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -359,7 +359,7 @@ impl ReadDisk for DiskDb {
         V: FromDisk,
     {
         // Reading individual values from iterators does not seem to cause database hangs.
-        self.zs_range_iter(cf, ..=upper_bound, false).next()
+        self.zs_range_iter(cf, ..=upper_bound, true).next()
     }
 
     fn zs_items_in_range_ordered<C, K, V, R>(&self, cf: &C, range: R) -> BTreeMap<K, V>
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -534,7 +534,13 @@ impl DiskDb {
                 // Increment the last byte in the vector that is not u8::MAX, or
                 // skip adding an upper bound if every byte is u8::MAX
                 if let Some(increment_idx) = bound.iter().rposition(|&v| v != u8::MAX) {
-                    bound[increment_idx] += 1;
+                    let increment_byte = bound
+                        .get_mut(increment_idx)
+                        .expect("index should be in bounds");
+                    *increment_byte = increment_byte
+                        .checked_add(1)
+                        .expect("adding 1 should succeed");
+
                     opts.set_iterate_upper_bound(bound);
                 }
             }
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -401,6 +401,10 @@ impl DiskWriteBatch {
 impl DiskDb {
     /// Returns an iterator over the items in `cf` in `range`.
     ///
+    /// Accepts a `reverse` argument and creates the iterator with an [`IteratorMode`](rocksdb::IteratorMode)
+    /// of [`End`](rocksdb::IteratorMode::End), or [`From`](rocksdb::IteratorMode::From)
+    /// with [`Direction::Reverse`](rocksdb::Direction::Reverse).
+    ///
     /// Holding this iterator open might delay block commit transactions.
     pub fn zs_range_iter<C, K, V, R>(
         &self,
@@ -496,15 +500,33 @@ impl DiskDb {
     where
         R: RangeBounds<Vec<u8>>,
     {
-        use std::ops::Bound::*;
-
         let mut opts = ReadOptions::default();
+        let (lower_bound, upper_bound) = Self::zs_iter_bounds(range);
 
-        if let Included(bound) | Excluded(bound) = range.start_bound() {
-            opts.set_iterate_lower_bound(bound.clone());
+        if let Some(bound) = lower_bound {
+            opts.set_iterate_lower_bound(bound);
         };
 
-        match range.end_bound().cloned() {
+        if let Some(bound) = upper_bound {
+            opts.set_iterate_upper_bound(bound);
+        };
+
+        opts
+    }
+
+    /// Returns a lower and upper iterate bounds for a range.
+    fn zs_iter_bounds<R>(range: &R) -> (Option<Vec<u8>>, Option<Vec<u8>>)
+    where
+        R: RangeBounds<Vec<u8>>,
+    {
+        use std::ops::Bound::*;
+
+        let lower_bound = match range.start_bound() {
+            Included(bound) | Excluded(bound) => Some(bound.clone()),
+            Unbounded => None,
+        };
+
+        let upper_bound = match range.end_bound().cloned() {
             Included(mut bound) => {
                 // Skip adding an upper bound if every byte is u8::MAX, or
                 // increment the last byte in the upper bound that is less than u8::MAX,
@@ -515,17 +537,13 @@ impl DiskDb {
                     v == &0
                 });
 
-                if !is_max_key {
-                    opts.set_iterate_upper_bound(bound);
-                }
+                (!is_max_key).then_some(bound)
             }
-            Excluded(bound) => {
-                opts.set_iterate_upper_bound(bound);
-            }
-            Unbounded => {}
+            Excluded(bound) => Some(bound),
+            Unbounded => None,
         };
 
-        opts
+        (lower_bound, upper_bound)
     }
 
     /// Returns the RocksDB iterator "from" mode for `range`.
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -195,7 +195,7 @@ pub trait ReadDisk {
     fn zs_first_key_value<C, K, V>(&self, cf: &C) -> Option<(K, V)>
     where
         C: rocksdb::AsColumnFamilyRef,
-        K: FromDisk,
+        K: IntoDisk + FromDisk,
         V: FromDisk;
 
     /// Returns the highest key in `cf`, and the corresponding value.
@@ -204,7 +204,7 @@ pub trait ReadDisk {
     fn zs_last_key_value<C, K, V>(&self, cf: &C) -> Option<(K, V)>
     where
         C: rocksdb::AsColumnFamilyRef,
-        K: FromDisk,
+        K: IntoDisk + FromDisk,
         V: FromDisk;
 
     /// Returns the first key greater than or equal to `lower_bound` in `cf`,
@@ -322,34 +322,22 @@ impl ReadDisk for DiskDb {
     fn zs_first_key_value<C, K, V>(&self, cf: &C) -> Option<(K, V)>
     where
         C: rocksdb::AsColumnFamilyRef,
-        K: FromDisk,
+        K: IntoDisk + FromDisk,
         V: FromDisk,
     {
         // Reading individual values from iterators does not seem to cause database hangs.
-        self.db
-            .iterator_cf(cf, rocksdb::IteratorMode::Start)
-            .next()?
-            .map(|(key_bytes, value_bytes)| {
-                Some((K::from_bytes(key_bytes), V::from_bytes(value_bytes)))
-            })
-            .expect("unexpected database failure")
+        self.zs_range_iter(cf, .., false).next()
     }
 
     #[allow(clippy::unwrap_in_result)]
     fn zs_last_key_value<C, K, V>(&self, cf: &C) -> Option<(K, V)>
     where
         C: rocksdb::AsColumnFamilyRef,
-        K: FromDisk,
+        K: IntoDisk + FromDisk,
         V: FromDisk,
     {
         // Reading individual values from iterators does not seem to cause database hangs.
-        self.db
-            .iterator_cf(cf, rocksdb::IteratorMode::End)
-            .next()?
-            .map(|(key_bytes, value_bytes)| {
-                Some((K::from_bytes(key_bytes), V::from_bytes(value_bytes)))
-            })
-            .expect("unexpected database failure")
+        self.zs_range_iter(cf, .., true).next()
     }
 
     #[allow(clippy::unwrap_in_result)]
@@ -359,17 +347,8 @@ impl ReadDisk for DiskDb {
         K: IntoDisk + FromDisk,
         V: FromDisk,
     {
-        let lower_bound = lower_bound.as_bytes();
-        let from = rocksdb::IteratorMode::From(lower_bound.as_ref(), rocksdb::Direction::Forward);
-
         // Reading individual values from iterators does not seem to cause database hangs.
-        self.db
-            .iterator_cf(cf, from)
-            .next()?
-            .map(|(key_bytes, value_bytes)| {
-                Some((K::from_bytes(key_bytes), V::from_bytes(value_bytes)))
-            })
-            .expect("unexpected database failure")
+        self.zs_range_iter(cf, lower_bound.., false).next()
     }
 
     #[allow(clippy::unwrap_in_result)]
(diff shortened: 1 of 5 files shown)


zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -514,6 +491,43 @@ impl DiskDb {
...
+            Included(mut bound) => {
+                // Increment the last byte in the vector that is not u8::MAX, or
+                // skip adding an upper bound if every byte is u8::MAX
+                if let Some(increment_idx) = bound.iter().rposition(|&v| v != u8::MAX) {
+                    let increment_byte = bound
+                        .get_mut(increment_idx)
+                        .expect("index should be in bounds");
+                    *increment_byte = increment_byte
+                        .checked_add(1)
+                        .expect("adding 1 should succeed");
```
arya2: This is already tested by `all_upgrades_and_wrong_commitments_with_fake_activation_heights` using `zs_next_key_value_from()`.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -431,14 +402,19 @@ impl DiskDb {
     /// Returns an iterator over the items in `cf` in `range`.
     ///
     /// Holding this iterator open might delay block commit transactions.
-    pub fn zs_range_iter<C, K, V, R>(&self, cf: &C, range: R) -> impl Iterator<Item = (K, V)> + '_
+    pub fn zs_range_iter<C, K, V, R>(
+        &self,
+        cf: &C,
+        range: R,
+        reverse: bool,
```
teor2345, line 413: This new argument needs an update to the method docs.

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: This code is technically correct, but might not achieve the performance goals, because the upper bound is larger than the minimum required upper bound.

I'd suggest using a well-tested big integer library function to do this.

Please also add a test case for the minimal upper bound. For example, the minimal upper bound of `0x00ff` is `0x0100`, not `0x01ff`.

teor2345 (dismissed):
This refactor is nice and simple, but unfortunately the upper bound calculation is not minimal.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7731/commits/a703d1ce996256b02f13d707d51cde17d8d33c64.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/7731/commits/a703d1ce996256b02f13d707d51cde17d8d33c64.

zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -24,3 +24,28 @@ impl DiskDb {
...
 }
+
+/// Check that the sprout tree database serialization format has not changed.
+#[test]
+fn zs_iter_opts_increments_key_by_one() {
+    let _init_guard = zebra_test::init();
+
+    // let (config, network) = Default::default();
+    // let db = DiskDb::new(&config, network);
+
```
arya2:
```suggestion
```

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: But unfortunately that test does not have sufficient coverage for the goals of the ticket.

zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -24,3 +24,25 @@ impl DiskDb {
...
     }
 }
+
+/// Check that zs_iter_opts returns an upper bound one greater than provided inclusive end bounds.
+#[test]
+fn zs_iter_opts_increments_key_by_one() {
+    let _init_guard = zebra_test::init();
+
+    let keys: [u32; 13] = [
+        0, 1, 200, 255, 256, 257, 65535, 65536, 65537, 16777215, 16777216, 16777217, 16777218,
```
teor2345: This is missing a test case for u32::MAX

teor2345: Otherwise this test has good coverage and maybe we don't need a proptest, since we know the structure of the code.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -514,6 +495,62 @@ impl DiskDb {
...
+                // Skip adding an upper bound if every byte is u8::MAX, or
+                // increment the last byte in the upper bound that is less than u8::MAX,
+                // and clear any bytes after it to increment the big-endian number this
+                // Vec represents to RocksDB.
+                let is_max_key = bound.iter_mut().rev().all(|v| {
+                    *v = v.wrapping_add(1);
+                    v == &0
+                });
+
+                (!is_max_key).then_some(bound)
```
teor2345: What happens if the key is variable-length?

teor2345 (commented):
This looks great!

But there's one edge case we still need to cover: variable-length keys.

We don't have any at the moment, but PR #7392 changes the size of a key, and its validation runs an iterator with one key size on a column family which has been changed to a different key size.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: It would sometimes (probably <1%) fail to put an upper bound on iterators, which is still constrained by `take_while` but might unnecessarily read-ahead.

Its not an issue for the sprout/history tree change where the key size changes across db format versions, but if key structs had variable length serialized representations, like if heights 0-255 were just 1 byte, this would need something like a method in `IntoDisk` for checking the max serialization size of a struct.

teor2345: Zebra's data access patterns aren't random, so I'm not sure if that `<1%` probability will hold under deliberate attack, or with certain key formats. Here are two examples:

#### Attackers using maximum keys

In an attack or overload scenario, if the keys depend on mempool or block data, then many of them could be at the maximum value.

But this is unlikely to happen with the current UTXO key format, because the height comes first, and it is always less than the maximum. So if we decide to accept this risk, we should document it. (But see my alternative solution below.)

#### Zebra using maximum keys

In the sprout and history tree changes, the new key is `()`, which has an empty array as its serialization. This is the maximum value for an empty array. 

An earlier PR fixed this issue by using `zs_get()` to access keys, rather than an iterator.

But without that fix, every iterator that used `()` would read some tombstones from the previous height-based key format. This happens because the byte array increment code always returns `None` for `[0; u8]`. The missing bound would temporarily reduce performance after the format change, until those tombstones were compacted. (Sometimes that doesn't happen for a long time.)

#### Proposed Solution

If the key is the maximum for its current length, add an extra `0x01` byte to the front of the array. This will give you the  minimal bound.

arya2: > If the key is the maximum for its current length, add an extra 0x01 byte to the front of the array. This will give you the minimal bound.

Good idea.

Applied in https://github.com/ZcashFoundation/zebra/pull/7731/commits/fcf171f10be906f0c77041401ad1b59d5adadbaf.

zebra-state/src/service/finalized_state/disk_db/tests.rs
arya2: > This is missing a test case for u32::MAX

Added in https://github.com/ZcashFoundation/zebra/pull/7731/commits/600aa5f6087dc0b6af9d9fcce54b614cedaee5c4.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -533,16 +549,20 @@ impl DiskDb {
...
-                let is_max_key = bound.iter_mut().rev().all(|v| {
+                let is_zero = bound.iter_mut().rev().all(|v| {
                     *v = v.wrapping_add(1);
                     v == &0
                 });
 
-                (!is_max_key).then_some(bound)
+                if is_zero {
+                    bound.push(0);
+                    *bound.get_mut(0).expect("should have at least 1 element") += 1;
```
arya2:
```suggestion
                    *bound.get_mut(0).expect("should have at least 1 element") = 1;
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -321,34 +322,20 @@ impl ReadDisk for DiskDb {
     fn zs_first_key_value<C, K, V>(&self, cf: &C) -> Option<(K, V)>
     where
         C: rocksdb::AsColumnFamilyRef,
-        K: FromDisk,
+        K: IntoDisk + FromDisk,
```
teor2345, line 325: Nitpick: This IntoDisk bound is no longer needed. (Same for all the methods that use seek without using a key.)

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -321,34 +322,20 @@ impl ReadDisk for DiskDb {
...
     {
-        // Reading individual values from iterators does not seem to cause database hangs.
-        self.db
-            .iterator_cf(cf, rocksdb::IteratorMode::Start)
-            .next()?
-            .map(|(key_bytes, value_bytes)| {
-                Some((K::from_bytes(key_bytes), V::from_bytes(value_bytes)))
-            })
-            .expect("unexpected database failure")
+        self.zs_seek(cf, |iter| iter.seek_to_first())
```
teor2345: This is a substantial change which could change performance. Can you do a full sync before we merge this PR? Or split this change into a separate PR?

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -430,15 +397,40 @@ impl DiskWriteBatch {
 impl DiskDb {
     /// Returns an iterator over the items in `cf` in `range`.
     ///
+    /// Accepts a `reverse` argument and creates the iterator with an [`IteratorMode`](rocksdb::IteratorMode)
```
teor2345: Nitpick: potentially confusing documentation.

```suggestion
    /// Accepts a `reverse` argument. If it is `true`, creates the iterator with an [`IteratorMode`](rocksdb::IteratorMode)
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -430,15 +397,40 @@ impl DiskWriteBatch {
...
+        self.zs_range_iter_with_direction(cf, range, reverse)
+    }
+
+    /// Returns the sought item in `cf`.
+    ///
+    /// Accepts a closure for seeking a key in a raw iterator over the items in `cf`.
+    pub fn zs_seek<C, K, V, F>(&self, cf: &C, seek_fn: F) -> Option<(K, V)>
+    where
+        C: rocksdb::AsColumnFamilyRef,
+        K: IntoDisk + FromDisk,
```
teor2345: Nitpick: The `IntoDisk` bound is not needed here.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -514,6 +507,65 @@ impl DiskDb {
...
+
+        let lower_bound = match range.start_bound() {
+            Included(bound) | Excluded(bound) => Some(bound.clone()),
+            Unbounded => None,
+        };
+
+        let upper_bound = match range.end_bound().cloned() {
+            Included(mut bound) => {
+                // Increment the last byte in the upper bound that is less than u8::MAX,
+                // and clear any bytes after it to increment the big-endian number this
```
teor2345: Optional Nitpick: RocksDB's documentation refers to keys as "lexicographically ordered bytes" rather than "big-endian numbers":
https://docs.rs/rocksdb/latest/rocksdb/struct.DBRawIteratorWithThreadMode.html#method.seek

```suggestion
                // and clear any bytes after it to increment the next key in lexicographic order (next big-endian number) this
```

zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -24,3 +24,47 @@ impl DiskDb {
...
+            "the upper bound should be 1 greater than the original key"
+        );
+
+        if expected_upper_bound == 0 {
+            assert_eq!(
+                bytes,
+                vec![1],
+                "there should be an extra byte with a value of 1"
+            );
+        }
```
teor2345, line 71: Please check the "no added bytes" case here as well.

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: Does each of the methods that calls `zs_seek()` have enough test coverage to make sure it works correctly? Or do we need to add `Some(_)` and `None` return value tests for the methods that aren't called from production code?

zebra-state/src/service/finalized_state/disk_db.rs
teor2345: Can't we just push 0x01 in the line before?

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -514,6 +507,65 @@ impl DiskDb {
...
+                // Vec represents to RocksDB.
+                let is_zero = bound.iter_mut().rev().all(|v| {
```
teor2345: Optional Nitpick: "is_zero" is technically correct, but it makes the code hard to read, because the bound is actually beyond the maximum value for the original length of the key.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: > Can't we just push 0x01 in the line before?

I think that would put the 1 in the last element and that we need it in the first element.

teor2345: `bound.insert(0, 0x01)` achieves the same thing as these 2 lines, but it's much less verbose.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Split out to #7759.

teor2345 (dismissed):
I'm happy to merge if these extra checks pass

zebra-state/src/service/finalized_state/disk_db/tests.rs
teor2345: I added this check

zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -24,3 +24,60 @@ impl DiskDb {
...
+
+        if expected_upper_bound == 0 {
+            assert_eq!(
+                bytes,
+                vec![1],
+                "there should be an extra byte with a value of 1"
+            );
+        } else {
+            assert_eq!(
+                key.len(),
```
teor2345:
```suggestion
                key.to_be_bytes().len(),
```

zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -24,3 +24,60 @@ impl DiskDb {
...
+                bytes.len(),
+                "there should be no extra bytes"
+            );
+        }
+
+        assert_ne!(
+            bytes[0],
+            0x00,
```
teor2345:
```suggestion
            bytes[0], 0x00,
```

teor2345 (dismissed):

teor2345 (dismissed):

gustavovalverde:
@mergifyio refresh

teor2345 (approved):
Thanks!

PR #7732, opened 2023-10-11
change(state): Set upper bound when reading from deleting column family tx_loc_by_transparent_addr_loc

### LINKED ISSUE #7664: Set iterator read bounds on deleting column families to avoid very slow syncs

arya2:
## Motivation

This PR updates `address_transaction_locations()` to avoid using multiple iterators to read data from disk and to set an upper iterate bound.

This function is currently used by `zebra-rpc` for the `get_address_tx_ids` method.

Part of #7664.

## Solution

- Refactor `address_transaction_locations()` to use `zs_range_iter` instead of `zs_next_key_value_from`

Related cleanups:
- Speed up `rpc_getaddresstxids_response`

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/transparent.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/transparent.rs
@@ -433,13 +433,13 @@ impl AddressTransaction {
         address_location: AddressLocation,
         query_start: Height,
     ) -> std::ops::RangeInclusive<AddressTransaction> {
+        // Iterating from the start height filters out transactions that aren't needed.
+        let start_height = max(query_start, address_location.height());
+
         // Iterating from the lowest possible transaction location gets us the first transaction.
         //
         // The address location is the output location of the first UTXO sent to the address,
         // and addresses can not spend funds until they receive their first UTXO.
-        //
-        // Iterating from the start height filters out transactions that aren't needed.
-        let start_height = max(query_start, address_location.height());
         let first_utxo_idx = address_location.transaction_index().0;
 
         let tx_loc = |tx_idx| TransactionLocation::from_index(start_height, tx_idx);
@@ -457,6 +457,7 @@ impl AddressTransaction {
     /// existing (valid) value.
     ///
     /// [1]: super::super::disk_db::ReadDisk::zs_next_key_value_from
+    #[allow(dead_code)]
     pub fn address_iterator_next(&mut self) {
         // Iterating from the next possible output location gets us the next output,
         // even if it is in a later block or transaction.
--- a/zebra-state/src/service/finalized_state/disk_format/transparent.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/transparent.rs
@@ -416,36 +416,36 @@ impl AddressTransaction {
         }
     }
 
-    /// Create an [`AddressTransaction`] which starts iteration for the supplied
+    /// Create a range of [`AddressTransaction`]s which starts iteration for the supplied
     /// address. Starts at the first UTXO, or at the `query_start` height,
-    /// whichever is greater.
+    /// whichever is greater. Ends at the last transaction index for an [`AddressLocation`].
     ///
-    /// Used to look up the first transaction with
-    /// [`ReadDisk::zs_next_key_value_from`][1].
+    /// Used to look up transactions with
+    /// [`DiskDb::zs_range_iter`][1].
     ///
     /// The transaction location might be invalid, if it is based on the
     /// `query_start` height. But this is not an issue, since
-    /// [`ReadDisk::zs_next_key_value_from`][1] will fetch the next existing
-    /// (valid) value.
+    /// [`DiskDb::zs_range_iter`][1] will fetch all existing
+    /// (valid) values in the range.
     ///
-    /// [1]: super::super::disk_db::ReadDisk::zs_next_key_value_from
-    pub fn address_iterator_start(
+    /// [1]: super::super::disk_db::DiskDb
+    pub fn address_iterator_range(
         address_location: AddressLocation,
         query_start: Height,
-    ) -> AddressTransaction {
+    ) -> std::ops::RangeInclusive<AddressTransaction> {
         // Iterating from the lowest possible transaction location gets us the first transaction.
         //
         // The address location is the output location of the first UTXO sent to the address,
         // and addresses can not spend funds until they receive their first UTXO.
-        let first_utxo_location = address_location.transaction_location();
-
+        //
         // Iterating from the start height filters out transactions that aren't needed.
-        let query_start_location = TransactionLocation::from_usize(query_start, 0);
+        let start_height = max(query_start, address_location.height());
+        let first_utxo_idx = address_location.transaction_index().0;
 
-        AddressTransaction {
-            address_location,
-            transaction_location: max(first_utxo_location, query_start_location),
-        }
+        let tx_loc = |tx_idx| TransactionLocation::from_index(start_height, tx_idx);
+        let addr_tx = |tx_idx| AddressTransaction::new(address_location, tx_loc(tx_idx));
+
+        addr_tx(first_utxo_idx)..=addr_tx(u16::MAX)
     }
 
     /// Update the transaction location to the next possible transaction for the
--- a/zebra-state/src/service/finalized_state/disk_format/transparent.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/transparent.rs
@@ -431,21 +431,21 @@ impl AddressTransaction {
     /// [1]: super::super::disk_db::DiskDb
     pub fn address_iterator_range(
         address_location: AddressLocation,
-        query_start: Height,
+        query: std::ops::RangeInclusive<Height>,
     ) -> std::ops::RangeInclusive<AddressTransaction> {
-        // Iterating from the start height filters out transactions that aren't needed.
-        let start_height = max(query_start, address_location.height());
-
         // Iterating from the lowest possible transaction location gets us the first transaction.
         //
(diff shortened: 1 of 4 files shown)


zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -235,44 +235,25 @@ impl ZebraDb {
...
-
-            addr_transactions.insert(transaction_location);
-
-            // A potentially invalid key representing the next possible output
-            transaction_location.address_iterator_next();
-        }
+        let last_transaction_location = AddressTransaction::new(
+            address_location,
+            TransactionLocation::from_usize(*query_height_range.end(), u16::MAX.into()),
+        );
```
teor2345: The constant here should be `MAX_ON_DISK_OUTPUT_INDEX`, any higher indexes are invalid in the database.

Optional: Can you move this limit code into an `AddressTransaction::address_iterator_end()` method?

Unlike `address_iterator_start()`, we can't use the `address_location` to improve performance here. But it's helpful to have all the range logic in the same place.

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -235,44 +235,25 @@ impl ZebraDb {
         let tx_loc_by_transparent_addr_loc =
             self.db.cf_handle("tx_loc_by_transparent_addr_loc").unwrap();
 
-        // Manually fetch the entire addresses' transaction locations
-        let mut addr_transactions = BTreeSet::new();
-
         // A potentially invalid key representing the first UTXO send to the address,
         // or the query start height.
-        let mut transaction_location = AddressTransaction::address_iterator_start(
+        let transaction_location = AddressTransaction::address_iterator_start(
```
teor2345: Optional Nitpick: since we're no longer modifying this variable, it is always the `first_transaction_location`.

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -235,44 +235,25 @@ impl ZebraDb {
...
-            if transaction_location.transaction_location().height > *query_height_range.end() {
-                break;
-            }
-
-            addr_transactions.insert(transaction_location);
-
-            // A potentially invalid key representing the next possible output
-            transaction_location.address_iterator_next();
-        }
+        let last_transaction_location = AddressTransaction::new(
```
teor2345: Optional Nitpick: This is technically the `max_transaction_location`, because unlike the first transaction, we don't know the last transaction in the database until the iterator runs.

teor2345 (dismissed):
This is good except for an incorrect constant (which is currently technically valid due to a workaround in other code).

The names here could do with some changes to avoid confusion when we modify this code again, but that is optional.

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -235,44 +235,25 @@ impl ZebraDb {
...
-            // We're past the end height, so we're finished with this query
-            if transaction_location.transaction_location().height > *query_height_range.end() {
-                break;
-            }
-
-            addr_transactions.insert(transaction_location);
-
-            // A potentially invalid key representing the next possible output
-            transaction_location.address_iterator_next();
-        }
```
teor2345: It is really nice to delete all this complex code, thanks for that!

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
arya2: > The constant here should be `MAX_ON_DISK_OUTPUT_INDEX`, any higher indexes are invalid in the database.

`OutputIndex` is in `address_location`, this one is a `u16`: https://github.com/ZcashFoundation/zebra/blob/use-range-iter/zebra-state/src/service/finalized_state/disk_format/block.rs#L70-L87

arya2: > Optional: Can you move this limit code into an AddressTransaction::address_iterator_end() method?

Yep, moved in https://github.com/ZcashFoundation/zebra/pull/7732/commits/7662b6ba9e27d21bd531565719587938b7bb3337.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -416,36 +416,36 @@ impl AddressTransaction {
         }
     }
 
-    /// Create an [`AddressTransaction`] which starts iteration for the supplied
+    /// Create a range of [`AddressTransaction`]s which starts iteration for the supplied
     /// address. Starts at the first UTXO, or at the `query_start` height,
-    /// whichever is greater.
+    /// whichever is greater. Ends at the last transaction index for an [`AddressLocation`].
```
teor2345: This documentation is not quite right: the range ends at the maximum possible transaction index for the end height.

This is different from the last actual UTXO, which is stored in the database and requires an iterator query to discover.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -416,36 +416,36 @@ impl AddressTransaction {
...
     ///
-    /// Used to look up the first transaction with
-    /// [`ReadDisk::zs_next_key_value_from`][1].
+    /// Used to look up transactions with
+    /// [`DiskDb::zs_range_iter`][1].
     ///
     /// The transaction location might be invalid, if it is based on the
```
teor2345, line 426:
```suggestion
    /// The transaction locations might be invalid, if it is based on the
```

teor2345: This part of the docs needs updating because there are two locations now: the start can be invalid, the end is always invalid.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -416,36 +416,36 @@ impl AddressTransaction {
...
-    ) -> AddressTransaction {
+        query: std::ops::RangeInclusive<Height>,
+    ) -> std::ops::RangeInclusive<AddressTransaction> {
         // Iterating from the lowest possible transaction location gets us the first transaction.
         //
         // The address location is the output location of the first UTXO sent to the address,
         // and addresses can not spend funds until they receive their first UTXO.
         let first_utxo_location = address_location.transaction_location();
 
         // Iterating from the start height filters out transactions that aren't needed.
```
teor2345:
```suggestion
        // Iterating from the start height to the end height filters out transactions that aren't needed.
```

teor2345 (dismissed):
Thank you! I think the changes are good, but the documentation is now outdated.

teor2345:
Can you also update the PR title? There aren't any read bounds being set in this PR, it's just a refactor that allows them to be set in the next PR.

It might also help to use the name of the column family or RPC, because there's just one of them.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7732/commits/cb9a1f66c53758597329b2f36278df3c8126b818.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/7732/commits/cb9a1f66c53758597329b2f36278df3c8126b818.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7732/commits/cb9a1f66c53758597329b2f36278df3c8126b818.

teor2345 (approved):

arya2:
@Mergifyio refresh

PR #7734, opened 2023-10-12
fix(state): Limit the retrieval of note commitment subtrees

### LINKED ISSUE #7708: bug: z_getsubtreesbyindex can return 2 subtrees when 1 was requested
### What happened?

When 1 subtree is requested, 0 or 1 subtrees are returned.

Instead, this happened: 

Sometimes 2 subtrees are returned, and there is a gap between them.

Running tests in CI on a cached state:
https://github.com/ZcashFoundation/zebra/actions/runs/6453229873/job/17516699670?pr=7703#step:9:355
https://github.com/ZcashFoundation/zebra/actions/runs/6453229873/job/17516707077?pr=7703#step:9:302

#### Snapshot test

```
Snapshot file: zebrad/tests/snapshots/z_getsubtreesbyindex_mainnet_sapling_0_1.snap
Snapshot: z_getsubtreesbyindex_mainnet_sapling_0_1
Source: zebrad/tests/acceptance.rs:2723
────────────────────────────────────────────────────────────────────────────────
Expression: parsed
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    5     5 │     "subtrees": [
    6     6 │       {
    7     7 │         "root": "754bb593ea42d231a7ddf367640f09bbf59dc00f2c1d2003cc340e0c016b5b13",
    8     8 │         "end_height": 558822
          9 │+      },
         10 │+      {
         11 │+        "root": "ffdd1a0628296bf274daa5ea13fb57818d678957d49f91ccd69efae0bac24934",
         12 │+        "end_height": 2255074
    9    13 │       }
   10    14 │     ]
   11    15 │   },
```

#### gRPC test

```
Message:  The response from the `z_getsubtreesbyindex` RPC contains a wrong number of Sapling subtrees.
Location: zebrad/tests/common/lightwalletd/wallet_grpc_test.rs:429
```

### Zebra Version

zebrad 1.2.0 PR #7703

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

CI

### Additional information

This seems to happen when there is a new subtree in the non-finalised state, so the bug is probably in the state code that merges finalised and non-finalised subtrees.

upbqdn:
## Motivation

Close #7708.

The issue in #7708 was that when retrieving the subtrees from the finalized and non-finalized state, we always added all subtrees from the non-finalized state to the response without paying attention to the requested limit.

## Solution

- Apply the given limit when retrieving subtrees.
- In non-release builds, make sure the limit applies.
- Add tests that test if we combine the subtrees from finalized and non-finalized states correctly. In order to implement the tests:
    - Derive `Default` for `Chain`.
    - Derive `Default` for Sapling & Orchard note commitment tree `Node`.
    - Allow direct insertions of subtrees into `Chain` for tests.
    - Set the visibility of `DiskWriteBatch` to `pub(super)`.

### Unrelated Clean-ups

- Use `default()` for subtree roots in other tests.
- Clarify some docs.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/read/tree.rs
+++ b/zebra-state/src/service/read/tree.rs
@@ -190,6 +190,19 @@ where
 {
     let mut db_list = db.orchard_subtree_list_by_index_for_rpc(start_index, limit);
 
+    if let Some(limit) = limit {
+        let subtrees_num = u16::try_from(db_list.len())
+            .expect("There can't be more than `u16::MAX` Orchard subtrees.");
+
+        // Return the subtrees if the amount of them reached the given limit.
+        if subtrees_num == limit.0 {
+            return db_list;
+        }
+
+        // If not, make sure the amount is below the limit.
+        debug_assert!(subtrees_num < limit.0);
+    }
+
     // If there's no chain, then we have the complete list.
     let Some(chain) = chain else {
         return db_list;
@@ -209,6 +222,17 @@ where
         // If there's no matching index, just update the list of trees.
         let Some(db_subtree) = db_list.get(&fork_index) else {
             db_list.insert(fork_index, fork_subtree);
+
+            // Stop adding new subtrees once their amount reaches the given limit.
+            if let Some(limit) = limit {
+                let subtrees_num = u16::try_from(db_list.len())
+                    .expect("There can't be more than `u16::MAX` Orchard subtrees.");
+
+                if subtrees_num == limit.0 {
+                    break;
+                }
+            }
+
             continue;
         };
 
@@ -218,6 +242,12 @@ where
         }
 
         // Otherwise, the subtree is already in the list, so we don't need to add it.
+    }
+
+    // Make sure the amount of retrieved subtrees does not exceed the given limit.
+    #[cfg(debug_assertions)]
+    if let Some(limit) = limit {
+        assert!(db_list.len() <= limit.0.into());
     }
 
     // Check that we got the start subtree from the non-finalized or finalized state.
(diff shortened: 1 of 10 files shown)


zebra-state/src/service/read/tree.rs
```diff
@@ -74,6 +74,19 @@ where
 {
     let mut db_list = db.sapling_subtree_list_by_index_for_rpc(start_index, limit);
 
+    if let Some(limit) = limit {
+        let subtrees_num = u16::try_from(db_list.len())
+            .expect("There can't be more than `u16::MAX` Sapling subtrees.");
+
+        // Return the subtrees if the amount of them reached the given limit.
+        if subtrees_num == limit.0 {
```
arya2, line 82: Could we update the implementation of (`sapling_subtrees_in_range`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-state/src/service/non_finalized_state/chain.rs#L710) to take a range?

(And update the docs for `sapling_subtree_list_by_index_for_rpc`)

arya2 (approved):
This looks great, thank you!

zebra-state/src/service/read/tree.rs
teor2345: This change is trickier than it looks, because turning `start, len` into `start..(start + len)` could cause integer overflows or off-by-one errors.

PR #7739, opened 2023-10-13
cleanup(state): Refactor subtree read functions to accept ranges and check memory first

arya2:
## Motivation

The `read::{sapling, orchard}_subtrees` functions are reading from disk before checking the non-finalized state and minor cleanup of inconsistent behaviour between the `Chain` methods for retrieving subtrees from the non-finalized state compared to those of `DiskDb`.

## Solution

- Check if all of the subtrees in the range exist in the chain before reading from disk
- Use `RangeBounds` instead of start/end bounds
- Update `Chain::{sapling, orchard}_subtrees_in_range` to only query for the provided range
- Move and consolidate checks for results containing `start_index`
- Update docs for `DiskDb::{sapling, orchard}_subtree_list_by_index_for_rpc()` and `Chain::{sapling, orchard}_subtrees_in_range`
 
## Review

Anyone can review.

### Reviewer Checklist


zebra-state/src/service/read/tree.rs
```diff
@@ -182,78 +150,46 @@ where
...
-        assert!(db_list.len() <= limit.0.into());
-    }
-
-    // Check that we got the start subtree from the non-finalized or finalized state.
-    // (The non-finalized state doesn't do this check.)
-    if db_list.get(&start_index).is_some() {
-        db_list
+    // Check that we got the start subtree
+    if results.contains_key(&start_index) {
+        results
```
upbqdn, line 194: If the start bound of `range` is `Excluded`, then we'll always return an empty map here. Should we also prohibit `Excluded` start bounds of `range`?

upbqdn (commented):
Thanks for this refactor. It looks great.

upbqdn (dismissed):

zebra-state/src/service/read/tree.rs
upbqdn: If we will, let's mention it in the docs?

zebra-state/src/service/read/tree.rs
```diff
@@ -66,78 +66,46 @@ where
...
-    // Unlike the other methods, this returns any trees in the range,
-    // even if there is no tree for start_index.
-    let fork_list = chain.as_ref().sapling_subtrees_in_range(start_index, limit);
-
-    // If there's no subtrees in chain, then we have the complete list.
-    if fork_list.is_empty() {
-        return db_list;
+    let start_index = match range.start_bound().cloned() {
+        Included(start_index) | Excluded(start_index) => start_index,
+        Unbounded => panic!("range provided to sapling_subtrees() must have start bound"),
```
teor2345: Runtime API restrictions are risky, because they can introduce panic bugs, particularly if the code does not have enough test coverage.

This also seems like an unnecessary restriction: if the range doesn't have a start bound, there is only one sensible start index, which is zero. So we could allow an unlimited range here.

However, this will need at least one extra test.

zebra-state/src/service/read/tree.rs
```diff
@@ -66,78 +66,46 @@ where
...
-
-    // Unlike the other methods, this returns any trees in the range,
-    // even if there is no tree for start_index.
-    let fork_list = chain.as_ref().sapling_subtrees_in_range(start_index, limit);
-
-    // If there's no subtrees in chain, then we have the complete list.
-    if fork_list.is_empty() {
-        return db_list;
+    let start_index = match range.start_bound().cloned() {
+        Included(start_index) | Excluded(start_index) => start_index,
```
teor2345: Given the restriction that the start index must be included, I'd suggest using `start_index + 1` for the `Excluded` case.

However, this will need at least one extra test.

zebra-state/src/service/read/tree.rs
```diff
@@ -66,78 +66,46 @@ where
...
-            db_list.insert(fork_index, fork_subtree);
-
-            // Stop adding new subtrees once their amount reaches the given limit.
-            if let Some(limit) = limit {
-                let subtrees_num = u16::try_from(db_list.len())
-                    .expect("There can't be more than `u16::MAX` Sapling subtrees.");
-
-                if subtrees_num == limit.0 {
+    let results = match chain.map(|chain| chain.as_ref().sapling_subtrees_in_range(range.clone())) {
+        Some(chain_results) if chain_results.contains_key(&start_index) => return chain_results,
```
teor2345: The memory-only case is currently rare, because the probability of there being a single subtree in the non-finalised state is currently low, and clients often request multiple subtrees. So I'm not sure that this optimisation will achieve any practical performance gain.

However, the change is still correct, and it does simplify some code.

And after this change, the concurrency comment in this function's documentation is outdated. It might be worth moving it next to the concurrent code, so we remember to update it.

And this case needs an extra test.

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
```diff
@@ -261,52 +260,14 @@ impl ZebraDb {
     #[allow(clippy::unwrap_in_result)]
     pub fn sapling_subtree_list_by_index_for_rpc(
```
teor2345: This method's name and documentation are now outdated, please update them. (It doesn't contain any RPC-specific code any more, so it can just use a similar name to the other range methods.)

zebra-state/src/service.rs
```diff
@@ -1507,14 +1508,29 @@ impl Service<ReadRequest> for ReadStateService {
 
                 tokio::task::spawn_blocking(move || {
                     span.in_scope(move || {
+                        let end_index = limit
+                            .and_then(|limit| start_index.0.checked_add(limit.0))
```
teor2345, line 1512: Does the overflow case have tests?

zebra-state/src/service/read/tree.rs
```diff
@@ -66,78 +66,46 @@ where
...
+        Some(chain_results) => {
```
teor2345: Does the "combine chain with finalized" case have any tests?

We know it didn't previously have enough tests, because it had a limit bug in it.

teor2345 (commented):
This looks like a good rewrite, and I appreciate how it simplifies the code.

All my comments on the Sapling methods also apply to Orchard.

I'm concerned that this change could lead to unexpected bugs, or significant extra testing. Did Pili approve a scope for this work, or is there a ticket with the scope of the work? In particular, supporting some range types might need extra tests, because the start bound gets special handling.

Since this is a substantial rewrite, do we need to re-do the manual tests with `zcashd` from #7450? Which unit tests do we need to add?

A few of your recent PRs have missed documentation updates and test coverage. Would it help if you make a personal checklist that you go through before you submit a PR? Or could we add a checklist for PR authors to the PR template?

(I am working on PR template updates at the moment.)

zebra-state/src/service/read/tree.rs
teor2345: It would be unusual for us to reject some range types, and only do it at runtime (and not at compile time). For every other API that accepts a range, Zebra handles all range types correctly. But I understand this is a scope increase.

See my suggestions for how to handle `Unbounded` and `Excluded` on that code, and my detailed concerns here:
https://github.com/ZcashFoundation/zebra/pull/7739#discussion_r1359946646

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
teor2345: This method can also be used in the state validity tests, which would improve test coverage:
https://github.com/ZcashFoundation/zebra/blob/c43142332ae503323e46b082cba20f4c7f72ca78/zebra-state/src/service/finalized_state/disk_format/upgrade/add_subtrees.rs#L349

zebra-state/src/service/read/tree.rs
```diff
@@ -66,78 +66,46 @@ where
...
-    };
-
-    // Unlike the other methods, this returns any trees in the range,
-    // even if there is no tree for start_index.
-    let fork_list = chain.as_ref().sapling_subtrees_in_range(start_index, limit);
-
-    // If there's no subtrees in chain, then we have the complete list.
-    if fork_list.is_empty() {
-        return db_list;
+    let start_index = match range.start_bound().cloned() {
```
teor2345: If the goal is to find the first included index, we're effectively doing something like:

```rust
let start_index = (0..=u16::MAX).iter().find(|index| range.contains(NoteCommitmentSubtreeIndex(index)));
let Some(start_index) = start_index else {
    // Range is mathematically empty
    return BTreeMap::new();
};
```

Ideally writing that could would optimise to start_index, start_index + 1, or 0 based on the bound. But that doesn't seem to happen yet, maybe Rust iterators don't optimise constant iterators with closures well.

You can see the assembly here, it's significantly longer than just matching the bound:
 https://play.rust-lang.org/?version=stable&mode=release&edition=2021

zebra-state/src/service.rs
arya2: There's test coverage for using the `read::{sapling, orchard}_subtree` methods with a `RangeFrom` as would be used here when this overflows.

zebra-state/src/service/read/tree.rs
arya2: Tested here: https://github.com/ZcashFoundation/zebra/pull/7739/files#diff-c6180374b618db7b086470ac69e69dafaba94e7b39019a1d41f2a3857fb54086R159

zebra-state/src/service/read/tree.rs
arya2: Tested here: https://github.com/ZcashFoundation/zebra/pull/7739/files#diff-c6180374b618db7b086470ac69e69dafaba94e7b39019a1d41f2a3857fb54086R202

zebra-state/src/service/read/tree.rs
arya2: > So I'm not sure that this optimisation will achieve any practical performance gain.

It probably won't, that was a nitpick-y cleanup.

Tested here: https://github.com/ZcashFoundation/zebra/pull/7739/files#diff-c6180374b618db7b086470ac69e69dafaba94e7b39019a1d41f2a3857fb54086R175

zebra-state/src/service/read/tree.rs
arya2: Tested here: https://github.com/ZcashFoundation/zebra/pull/7739/files#diff-c6180374b618db7b086470ac69e69dafaba94e7b39019a1d41f2a3857fb54086R159-R169

zebra-state/src/service/finalized_state/zebra_db/shielded.rs
arya2: Renamed in https://github.com/ZcashFoundation/zebra/pull/7739/commits/0d7b13b089c7b43069956293d665db6d79534e54.

mpguerra:
> I'm concerned that this change could lead to unexpected bugs, or significant extra testing. Did Pili approve a scope for this work, or is there a ticket with the scope of the work? In particular, supporting some range types might need extra tests, because the start bound gets special handling.

I was wondering which issue this PR related to... :P We should finish off the rest of the issues in this sprint and merge their associated PRs before doing any more work on this one

zebra-state/src/service.rs
teor2345: I was asking if the overflow case of this statement ever runs in our tests. Do our RPC integration tests have test cases that call it?

As far as I remember they mainly use indexes and limits 1 and 2. But it would be quick to add.

zebra-state/src/service/read/tree.rs
teor2345: Thank you!

zebra-state/src/service/read/tests/vectors.rs
```diff
@@ -100,6 +100,115 @@ async fn populated_read_state_responds_correctly() -> Result<()> {
...
+    impl<T> std::ops::RangeBounds<T> for RangeExclusiveStart<T> {
+        fn start_bound(&self) -> Bound<&T> {
+            Excluded(&self.start)
+        }
+        fn end_bound(&self) -> Bound<&T> {
+            self.end.as_ref().map(Excluded).unwrap_or(Unbounded)
+        }
+    }
+
+    let range = RangeExclusiveStart::default();
```
teor2345: These 15 lines of code can be replaced with one, because RangeBounds is implemented for a tuple pair:
https://doc.rust-lang.org/std/ops/trait.RangeBounds.html#impl-RangeBounds%3CT%3E-for-(Bound%3CT%3E,+Bound%3CT%3E)

Normally I'd say this simplification is optional, but it's actually tricky to read the test without it.

```rust
    let range = (Excluded(0.into()), Unbounded);
```

zebra-state/src/service.rs
arya2: > Do our RPC integration tests have test cases that call it?

No, there are snapshot tests for `z_get_subtrees_by_index()` but they use a mock state service with low indexes/limits.

zebra-chain/src/subtree.rs
```diff
@@ -14,7 +14,7 @@ pub const TRACKED_SUBTREE_HEIGHT: u8 = 16;
 
 /// A note commitment subtree index, used to identify a subtree in a shielded pool.
 /// Also used to count subtrees.
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Default)]
```
teor2345: Optional: In general I prefer to see an explicit `0` or `zero()` method in code, rather than `default()`. It is easier for reviewers to find incorrect uses of `zero()` than `default()`, particularly in complicated nested types.

zebra-state/src/service/read/tree.rs
```diff
@@ -164,96 +89,93 @@ where
...
+    read_chain: F1,
+    read_disk: F2,
+) -> BTreeMap<NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<N>>
+where
+    C: AsRef<Chain>,
+    N: PartialEq,
+    R: std::ops::RangeBounds<NoteCommitmentSubtreeIndex> + Clone,
+    F1: FnOnce(&Chain, R) -> BTreeMap<NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<N>>,
+    F2: FnOnce(R) -> BTreeMap<NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<N>>,
+{
```
teor2345: Optional Nitpick: As a reviewer, I prefer complicated generic functions to have meaningful names for their generic types. Otherwise I spend a lot of time working out what each type does.

For example, it took me more than a minute to work out that `N` was the type of the tree note. This could be avoided by renaming `N` to `Note`, or whatever the generic is called in `NoteCommitmentSubtreeData<_>`.

I'd also suggest renaming:
* `F1` to `ChainSubtreeFn`
* `F2` to `DbSubtreeFn`

Optional:
* `C` to `Chain`
* `R` to `Range`

(The last two are common usage in Zebra or the Rust standard library, so they are easy to guess.)

zebra-state/src/service/read/tree.rs
```diff
@@ -164,96 +89,93 @@ where
...
+        range,
+        |chain, range| chain.orchard_subtrees_in_range(range),
+        |range| db.orchard_subtree_list_by_index_range(range),
+    )
+}
 
-    // Unlike the other methods, this returns any trees in the range,
-    // even if there is no tree for start_index.
-    let fork_list = chain.as_ref().orchard_subtrees_in_range(start_index, limit);
+/// Returns a list of [`NoteCommitmentSubtree`]s starting at `start_index`.
```
teor2345: This comment needs updating, please document the current arguments of the function.

zebra-state/src/service/read/tree.rs
```diff
@@ -164,96 +89,93 @@ where
...
+///
+/// If there is no subtree at the first index in the range, the returned list is empty.
+/// Otherwise, subtrees are continuous up to the finalized tip.
+///
+/// # Correctness
+///
+/// APIs that return single subtrees can't be used here, because they can create
```
teor2345:
```suggestion
/// APIs that return single subtrees can't be used for `read_chain` and `read_disk`, because they can create
```

zebra-state/src/service.rs
teor2345: I am happy to call this test optional and resolve this comment, because it's an edge case that seems correct from the code.

Are you ok with that, or is the test important enough to spend time on?

zebra-chain/src/subtree.rs
arya2: It doesn't need it since the tuple implements `RangeBounds`, removed in https://github.com/ZcashFoundation/zebra/pull/7739/commits/668462d586cfe5a409c56b189478c374296e6b2d.

zebra-state/src/service/read/tree.rs
arya2: > As a reviewer, I prefer complicated generic functions to have meaningful names for their generic types. Otherwise I spend a lot of time working out what each type does.

Whoops. Noted.

Renamed in https://github.com/ZcashFoundation/zebra/pull/7739/commits/668462d586cfe5a409c56b189478c374296e6b2d.

zebra-state/src/service/read/tree.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7739/commits/668462d586cfe5a409c56b189478c374296e6b2d.

zebra-state/src/service/read/tree.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7739/commits/668462d586cfe5a409c56b189478c374296e6b2d.

zebra-state/src/service.rs
arya2: > Are you ok with that, or is the test important enough to spend time on?

I'm happy with that :)

teor2345:
> > I'm concerned that this change could lead to unexpected bugs, or significant extra testing. Did Pili approve a scope for this work, or is there a ticket with the scope of the work? In particular, supporting some range types might need extra tests, because the start bound gets special handling.
> 
> I was wondering which issue this PR related to... :P We should finish off the rest of the issues in this sprint and merge their associated PRs before doing any more work on this one

This PR blocks ticket #7463, because they both change the subtree code.

We can start work on #7463, but this PR should merge before the PR for that ticket, otherwise there will be a lot of rework to merge them. Since #7463 is a fully scripted change, we can just re-run the script after this PR merges. (Any non-scripted parts of that ticket should be in a separate PR based on this PR.)

Conflicts like this are another reason why scheduling work is important!

zebra-state/src/service/read/tree.rs
```diff
@@ -50,97 +50,24 @@ where
 
 /// Returns a list of Sapling [`NoteCommitmentSubtree`]s starting at `start_index`.
```
teor2345: There is no `start_index` argument any more.

Please remember to update documentation!

zebra-state/src/service/read/tree.rs
```diff
@@ -164,96 +91,108 @@ where
...
+/// Otherwise, subtrees are continuous up to the finalized tip.
+///
+/// Accepts a `chain` from the non-finalized state, a `range` of subtree indexes to retrieve,
+/// a `read_chain` function for retrieving the `range` of subtrees from `chain`, and
+/// a `read_disk` function for retrieving the `range` from [`ZebraDb`].
+///
+/// Returns without calling `read_disk` if `read_chain` returns a subtree at the first subtree index
+/// in the provided `range`. Otherwise, if `read_chain` and `read_disk` return different subtrees for
+/// the same subtree index, ignores all the trees in `chain` after the first inconsistent tree.
+///
```
arya2, line 127: Should we move the correctness comment back up instead of this note, or do would it help to add:

```suggestion
/// See correctness comment in function body for more details.
///
```

?

teor2345 (dismissed):
Ok, this looks great, pending the minor function doc update.

zebra-state/src/service/read/tree.rs
teor2345: I don't think we should duplicate this level of detail.

How about a summary of the goals in the function docs, and the details of how that is achieved inside the function in the existing detailed comment?

For example, the function doc could say:
> Returns a consistent set of subtrees for the supplied chain fork and database.
> Avoids reading the database if the trees are present in memory.

arya2: That's better, applied in https://github.com/ZcashFoundation/zebra/pull/7739/commits/3416c11264c6c7f4d0cfb3a47429b73f7dde3600.

zebra-state/src/service/read/tree.rs
arya2: Ah, thank you, fixed in https://github.com/ZcashFoundation/zebra/pull/7739/commits/3416c11264c6c7f4d0cfb3a47429b73f7dde3600.

teor2345 (approved):
Thanks!

PR #7740, opened 2023-10-13
feature(mining): Enable mining RPCs by default in production builds

teor2345:
## Post-Merge TODOs

- [ ] Describe this newly stabilised feature in the CHANGELOG
- [x] Arborist announcement
- [ ] Forum update

## Motivation

We want to enable mining by default in production builds.

Close #7361
Depends-On: #7744

## Solution

- Enable the feature by default
- Update the documentation to remove custom builds and docker images to enable the feature
- Update the README to point to the Docker mining instructions
  - Consistently use ZF funding stream addresses in examples

### Testing

- Always set the miner address in tests based on the network

This is a very verbose change, but all it does is pass the network from all the tests down to the base test config method, then set the miner address for mainnet or testnet there.

## Review

This is a pretty straightforward change. It doesn't matter which release it goes in.

I'd like extra documentation cleanups to go in another PR or ticket, that's what we're doing with test cleanups.

### Reviewer Checklist

  - [ ] Will the PR name make sense to users?
    - [ ] Does it need extra CHANGELOG info? (new features, breaking changes, large changes)
  - [ ] Are the PR labels correct?
  - [ ] Does the code do what the ticket and PR says?
    - [ ] Does it change concurrent code, unsafe code, or consensus rules?
  - [ ] How do you know it works? Does it have tests?

## Follow Up Work

These follow-up changes will be made in separate PRs:
- #7362
- any other documentation changes
- any other cleanups

PR #7744, opened 2023-10-15
fix(ci): Skip mempool checks if it wasn't queried by lightwalletd

teor2345:
## Motivation

CI timed out waiting for Zebra to log a RPC call that was never made:
https://github.com/ZcashFoundation/zebra/actions/runs/6503861323/job/17665273725?pr=7740#step:9:473

This gRPC is extremely hard to test and maybe we should just give up on the unreliable parts of the test.

## Solution

lightwalletd just isn't checking the mempool, no matter how long we wait before the RPC call.

Another possible issue is that using exactly the same wait time as lightwalletd creates a race condition between the zebrad and lightwalletd timers.

Changes:
- skip the mempool checks if lightwalletd didn't query the mempool
- increase the wait times
- add extra logging to make future bugs easier to diagnose

## Review

This will continue to occasionally fail PRs until it is fixed.

It doesn't need to be part of the release freeze because CI fixes don't go in the changelog.

### Reviewer Checklist

## Follow Up Work

If this test continues to be unreliable, disable those parts of the test.


teor2345:
This has had a CI run in this PR and it has passed.

PR #7740 seems to change the timing or cached state, so it has kept failing.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -159,10 +159,12 @@ pub async fn run() -> Result<()> {
...
         .into_inner();
 
-    // Lightwalletd won't call `get_raw_mempool` again until 2 seconds after the last call
+    // Lightwalletd won't call `get_raw_mempool` again until 2 seconds after the last call:
     // <https://github.com/zcash/lightwalletd/blob/master/frontend/service.go#L482>
+    //
+    // So we need to wait longer than that here.
     let sleep_until_lwd_last_mempool_refresh =
-        tokio::time::sleep(std::time::Duration::from_secs(2));
+        tokio::time::sleep(std::time::Duration::from_secs(3));
```
teor2345:
```suggestion
        tokio::time::sleep(std::time::Duration::from_secs(5));
```

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -159,10 +159,12 @@ pub async fn run() -> Result<()> {
...
         .into_inner();
 
-    // Lightwalletd won't call `get_raw_mempool` again until 2 seconds after the last call
+    // Lightwalletd won't call `get_raw_mempool` again until 2 seconds after the last call:
     // <https://github.com/zcash/lightwalletd/blob/master/frontend/service.go#L482>
+    //
+    // So we need to wait longer than that here.
     let sleep_until_lwd_last_mempool_refresh =
-        tokio::time::sleep(std::time::Duration::from_secs(2));
+        tokio::time::sleep(std::time::Duration::from_secs(5));
```
teor2345, line 167:
```suggestion
        tokio::time::sleep(std::time::Duration::from_secs(10));
```

arya2 (approved):

PR #7754, opened 2023-10-17
fix(ci): Limit parallel jobs launched by the crates workflow

### LINKED ISSUE #6702: CI fails because the API rate limit is exceeded during protoc install
## Motivation

While installing protoc:
> Error: API rate limit exceeded for installation ID 1523599.

https://github.com/ZcashFoundation/zebra/actions/runs/4997840627/jobs/8952981811?pr=6699#step:3:13

### Possible Solutions

1. In the "build crates individually" job, make the different crates into steps in the same job on the same machine, and `cargo clean` between each step. Then we only have to install `protoc` once, rather than 14 times.

2. Manually install protoc without using the GitHub API.

## Related Work

This is a regression to:
- #3712 
- #5819

teor2345:
## Motivation

Some workflows launch a lot of jobs, but those jobs are fast. So they can wait until later to run.

This also avoids hitting some rate limits.

Close #6702 

Depends-On: #7616

## Solution

Limit the number of parallel crate build jobs to avoid reaching rate-limits.

Limiting the number of parallel jobs also makes CI faster overall, because other long jobs can start earlier.

## Review

This is a routine fix.

### Reviewer Checklist

## Follow Up Work

We could stop installing protoc that way if the bug happens again.

### DIFF
--- a/.github/workflows/ci-build-crates.yml
+++ b/.github/workflows/ci-build-crates.yml
@@ -102,6 +102,8 @@ jobs:
     needs: [ matrix, check-matrix ]
     runs-on: ubuntu-latest
     strategy:
+      # avoid rate-limit errors by only launching a few of these jobs at a time
+      max-parallel: 2
       fail-fast: true
       matrix: ${{ fromJson(needs.matrix.outputs.matrix) }}
 


teor2345:
I have manually confirmed that this change works: it is currently running and only 2 build crates jobs are active:
Screenshot 2023-10-19 at 08 29 53

oxarbitrage (approved):
looks good, thanks.

PR #7758, opened 2023-10-17
poc(scanner): get started with the blockchain scanner proof of concept

### LINKED ISSUE #7727: feature: Blockchain scanner PoC
### Motivation

We want to investigate an approach to blockchain scanning in zebra using existing crates.

### Specifications

One potential approach would be to send a single `compactblock` from a test function to the existing crate and check the results.

We can then try with multiple blocks or other more complex scenarios.

### Possible Design

- Read blocks from disk and convert them into `CompactBlock`s - https://github.com/ZcashFoundation/zebra/pull/7758
- [Parse a hard-coded viewing key](https://docs.rs/zcash_address/latest/zcash_address/unified/struct.Ufvk.html) - https://github.com/ZcashFoundation/zebra/issues/7903

## Follow Up Task

- [ ] Admin: Add a branch protection rule for `Build crates individually / Build zebra-scan crate`

oxarbitrage:
## Motivation

Create a new crate `zebra-scan` and put proof of concept tests into it.

Part of https://github.com/ZcashFoundation/zebra/issues/7727

## Follow Up Task

- [ ] Admin: Add a branch protection rule for `Build crates individually / Build zebra-scan crate`


zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,313 @@
...
+    let dfvk: DiversifiableFullViewingKey = extsk.to_diversifiable_full_viewing_key();
+    let vks: Vec<(&AccountId, &SaplingIvk)> = vec![];
+    let nf = Nullifier([7; 32]);
+
+    let (state_config, network) = Default::default();
+    let (_state_service, read_only_state_service, _latest_chain_tip, _chain_tip_change) =
+        zebra_state::spawn_init(state_config, network, Height::MAX, 3000).await?;
+    let db = read_only_state_service.db();
+
+    let mut height = Height(0);
```
oxarbitrage: This will fail with:

```
---- tests::scanning_from_zebra stdout ----
thread 'tests::scanning_from_zebra' panicked at zebra-scanner/src/tests.rs:85:10:
called `Result::unwrap()` on an `Err` value: TreeSizeUnknown { protocol: Sapling, at_height: BlockHeight(0) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So it seems that we need to start scanning in sapling:

```suggestion
    let mut height = zebra_chain::parameters::NetworkUpgrade::Sapling.activation_height(network).unwrap();
```

But, the state is empty in this test so we need a synchronized chain to make this loop run or some alternative.

oxarbitrage: Ok, so i fixed this at https://github.com/ZcashFoundation/zebra/pull/7758/commits/cee1db69864ff1b820893b868008d0b0ff0a4155 with a fake sapling size of 1. 

Also, i changed to use a populated cache where we have 11 blocks inside, this allow us to test these 11 blocks are converted to compactblock and scanned. No relevant transactions should be found.

I also separated the test in 2 tests where one do the above and another use fake generated blocks to get a successful case.

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+readme = "../README.md"
+homepage = "https://zfnd.org/zebra/"
+# crates.io is limited to 5 keywords and categories
+keywords = ["zebra", "zcash"]
+# Must be one of <https://crates.io/category_slugs>
+categories = ["cryptography::cryptocurrencies"]
+
+[features]
+
+# Production features that activate extra dependencies, or extra features in dependencies
```
teor2345: Optional: This decision is not a blocker.

Team Decision: Should the scanner be off by default, and enabled by a Rust feature? How will the feature work? (Which crates need the feature, what is it called, etc.)

Should the name be different to the crate name?
For example: `external-wallet` or `wallet-support`

`cargo` automatically enables optional crate dependencies with the same name as features. But how it works is changing in 2024, so we might not want to rely on it.

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+
+[dependencies]
+zcash_client_backend = "0.10.0-rc.1"
+zcash_primitives = "=0.13.0-rc.1"
```
teor2345: This strict version requirement doesn't match our other crates. You could add a comment explaining why it's needed, or change it to match.

```suggestion
zcash_primitives = "0.13.0-rc.1"
```

zebra-scanner/Cargo.toml
teor2345: In particular, I'd like to avoid compiling `zcash_client_backend` into production `zebrad` releases until it does something useful. Unused dependencies are a minor security risk, and they also slow down builds.

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+zcash_note_encryption = "0.4.0"
+
+rand = "0.8.5"
+
+bls12_381 = "0.8.0"
+jubjub = "0.10.0"
+ff = "0.13.0"
+group = "0.13.0"
+
+tokio = { version = "1.33.0", features = ["full", "tracing", "test-util"] }
```
teor2345: We avoid depending on all of `tokio`'s features in production code. And there's no need to depend on test features in production builds.

Here's how to fix this:

In production (here):
```toml
tokio = { version = "1.33.0", features = ["tracing"] }
```

And whatever else is needed from this list:
https://docs.rs/tokio/latest/tokio/#feature-flags

We usually have something like:
https://github.com/ZcashFoundation/zebra/blob/bb7eeb6f76a91d8e6659891ff6d8bd0e055d2b74/zebra-consensus/Cargo.toml#L56

And then in `[dev-dependencies]`:
```toml
tokio = { version = "1.33.0", features = ["full", "tracing", "test-util"] }
```

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+
+zebra-state = { path = "../zebra-state", version = "1.0.0-beta.30" }
```
teor2345: Is `zebra-state` a test dependency?

Team Design Question: If our design calls the scanner from `zebra-state`, this will result in a circular dependency. They are annoying to get right so we try to avoid them. Is this dependency necessary, or should we use a different design?

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+zebra-consensus = { path = "../zebra-consensus", version = "1.0.0-beta.30" }
```
teor2345: Is `zebra-consensus` a test dependency?

If not, let's talk about how it is used with the team. (It might have similar circular dependency issues to `zebra-state`.)

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+
+zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.30", features = ["async-error"] }
+
+zebra-test = { path = "../zebra-test", version = "1.0.0-beta.30" }
```
teor2345: This is a test-only dependency.

zebra-scanner/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Blockchain scanner code for Zebra. 🦓
```
teor2345: Please don't put emoji in source code files, they make some editor / language server combinations panic.

```suggestion
//! Blockchain scanner code for Zebra.
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
+//! Test that we can scan blocks.
+//!
```
teor2345: Incomplete comment, or extra trailing empty line.

Optional: It could be helpful to summarise the tests and how they work here. 1-2 sentences is enough.

```suggestion
```

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "zebra-scanner"
+version = "1.0.0-beta.30"
+authors = ["Zcash Foundation <zebra@zfnd.org>"]
+description = "Blockchain scanner for the Zebra Zcash node"
```
teor2345: Optional:

```suggestion
description = "Shielded transaction scanner for the Zcash blockchain"
```

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
```
teor2345: Adding another crate requires a new branch protection rule to make sure that crate builds individually.

Which admin will add that rule after the PR merges?

zebra-scanner/Cargo.toml
teor2345: It looks like this dependency is unused.

```suggestion
```

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+jubjub = "0.10.0"
+ff = "0.13.0"
+group = "0.13.0"
+
+tokio = { version = "1.33.0", features = ["full", "tracing", "test-util"] }
+
+zebra-state = { path = "../zebra-state", version = "1.0.0-beta.30" }
+zebra-consensus = { path = "../zebra-consensus", version = "1.0.0-beta.30" }
+
+zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.30", features = ["async-error"] }
```
teor2345: This `zebra-chain` feature is unused:

```suggestion
zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.30" }
```

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+# crates.io is limited to 5 keywords and categories
+keywords = ["zebra", "zcash"]
+# Must be one of <https://crates.io/category_slugs>
+categories = ["cryptography::cryptocurrencies"]
+
+[features]
+
+# Production features that activate extra dependencies, or extra features in dependencies
+
+[dependencies]
```
teor2345: Maybe it would be better to move all these dependencies to `[dev-dependencies]`, then move them back when they are needed for production APIs?

zebra-scanner/Cargo.toml
teor2345: Replaced by https://github.com/ZcashFoundation/zebra/pull/7758/files#r1370890516

zebra-scanner/Cargo.toml
teor2345: Resolved by https://github.com/ZcashFoundation/zebra/pull/7758/files#r1370890516

zebra-scanner/Cargo.toml
teor2345: This can wait until we have production APIs.

zebra-scanner/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "zebra-scanner"
+version = "1.0.0-beta.30"
```
teor2345: Team Decision: Should this unaudited and untested crate have an alpha version instead? Should it have a prerelease version?

```suggestion
version = "0.1.0-alpha.0"
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+use group::GroupEncoding;
+use zebra_chain::{
+    block::Block,
+    chain_tip::ChainTip,
+    serialization::{ZcashDeserializeInto, ZcashSerialize},
+    transaction::Transaction,
+};
+
+#[tokio::test]
+async fn scanning_from_populated_zebra_state() -> Result<()> {
```
teor2345: The tests in this file are missing function comments that summarise what they do. 1 sentence is enough.

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+            .header
+            .previous_block_hash
+            .bytes_in_display_order()
+            .to_vec(),
+        time: block
+            .header
+            .time
+            .timestamp()
+            .try_into()
+            .expect("should work during 21st century"),
```
teor2345: Nitpick: Why is `try_into()` needed if both types are `u32`? Is there something weird going on here that isn't in the Rust API docs?

https://docs.rs/zebra-chain/1.0.0-beta.30/zebra_chain/serialization/struct.DateTime32.html#method.timestamp
https://docs.rs/zcash_client_backend/latest/zcash_client_backend/proto/compact_formats/struct.CompactBlock.html#structfield.time

Nitpick: in some C APIs, times are i32, so let's be extra clear that's not happening here:
https://en.wikipedia.org/wiki/Year_2038_problem

```suggestion
            .try_into()
            .expect("unsigned 32-bit times should work until 2105"),
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+        vtx: block
+            .transactions
+            .iter()
+            .cloned()
+            .enumerate()
+            .map(transaction_to_compact)
+            .collect(),
+        chain_metadata: Some(chain_metadata),
+
+        ..Default::default()
```
teor2345: Nitpick: Supplying default values for all new fields can cause errors or consensus bugs if the data structure changes later. Instead, supply specific placeholder values, so new fields cause compiler errors when the dependency is upgraded.

```suggestion
        // The protocol version is used for the gRPC wire format, so it isn't needed here.
        proto_version: 0,
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+
+// Convert a zebra transaction into a compact transaction.
+fn transaction_to_compact((index, tx): (usize, Arc<Transaction>)) -> CompactTx {
+    CompactTx {
+        index: index
+            .try_into()
+            .expect("tx index in block should fit in u64"),
+        hash: tx.hash().bytes_in_display_order().to_vec(),
+
+        // `fee` is not checked by the `scan_block` function.
```
teor2345: Optional:

```suggestion
        // `fee` is not checked by the `scan_block` function. It is allowed to be unset.
        // <https://docs.rs/zcash_client_backend/latest/zcash_client_backend/proto/compact_formats/struct.CompactTx.html#structfield.fee>
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+        fee: 0,
+
+        spends: tx
+            .sapling_nullifiers()
+            .map(|nf| CompactSaplingSpend {
+                nf: <[u8; 32]>::from(*nf).to_vec(),
+            })
+            .collect(),
+        outputs: tx
```
teor2345:
```suggestion
        // > output encodes the cmu field, ephemeralKey field, and a 52-byte prefix of the encCiphertext field of a Sapling Output
        //
        // <https://docs.rs/zcash_client_backend/latest/zcash_client_backend/proto/compact_formats/struct.CompactSaplingOutput.html>
        outputs: tx
```

zebra-scanner/src/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+        &zcash_primitives::consensus::MainNetwork,
+        cb,
+        &vks[..],
+        &[(account, nf)],
+        None,
+    )
+    .unwrap();
+
+    // The response should have one transaction relevant to the key we provided.
+    assert_eq!(res.transactions().len(), 1);
```
teor2345: Should we check that the correct transaction is found, and it's for the correct block hash?

zebra-scanner/Cargo.toml
arya2: > Team Decision: Should this unaudited and untested crate have an alpha version instead? Should it have a prerelease version?

An alpha version would be good, no opinion on the prerelease version.

teor2345: I feel like it's not really `1.0.0` quality yet, and we expect that the API might change. So that's why I suggested `0.1.0`.

But for simplicity we could use `1.0.0`?

arya2: > I feel like it's not really `1.0.0` quality yet, and we expect that the API might change. So that's why I suggested `0.1.0`.
> 
> But for simplicity we could use `1.0.0`?

`0.1.0` feels more consistent with the versioning schemes of other crates.

teor2345: I think that was my phrasing, `prerelease` can mean `alpha` or "lower than `1.0.0`"

zebra-scanner/Cargo.toml
oxarbitrage: This is a typo, no need to be the strict version. Fixed in https://github.com/ZcashFoundation/zebra/pull/7758/commits/2ca1c93ce72456d06e0d2ca26ed7bcff0f946522

zebra-scanner/src/lib.rs
oxarbitrage: Really ? It is strange nobody complained ever https://github.com/ZcashFoundation/zebra/blob/main/zebra-state/src/lib.rs#L1

oxarbitrage:
@teor2345 i think you made a very detailed review for a draft PR where most of the comments made could had been fixed before marking it as ready for review. These kind of comments in this stage are very distractive. Please remember draft PRs will generally not be that ready in regards to typos, spaces, comments, unused dependencies, etc. 

Next time please save those kind of comments for when the PR is ready for review. Thank you.

zebra-scanner/src/tests.rs
oxarbitrage: `try_into` is needed because the result from `timestamp()` is `i64`.

teor2345: Ah right it's DateTime rather than DateTime32, because we never changed that.

zebra-scanner/src/lib.rs
teor2345: Maybe no-one has ever tried to edit that line of that specific file with emacs and rust-analyzer. But it does happen, and lots of people have complained about it in other projects:
https://github.com/ZcashFoundation/zebra/blob/0e00ef4f4f20e40d61f9b7788af8bffb169cbed7/zebra-chain/src/transparent.rs#L58-L62

teor2345:
> @teor2345 i think you made a very detailed review for a draft PR where most of the comments made could had been fixed before marking it as ready for review. These kind of comments in this stage are very distractive. Please remember draft PRs will generally not be that ready in regards to typos, spaces, comments, unused dependencies, etc.
> 
> Next time please save those kind of comments for when the PR is ready for review. Thank you.

You're right, I could have saved some of these comments for a detailed review. That's tricky on GitHub, but I'll try to remember to do that next time.

Maybe you didn't see my comment on Slack, but I did a detailed review based on your questions at yesterday's planning meeting. You asked what decisions we needed to make at this stage of the project. The comments I marked with "Team Decision" are things that will be difficult to change once we merge this PR, or once we do the first release after the merge.

zebra-scanner/src/tests.rs
arya2: > Should we check that the correct transaction is found, and it's for the correct block hash?

It found the correct transaction in the manual test I ran.

zebra-scanner/Cargo.toml
teor2345: We resolved this by deciding not to call the scanner inside `zebra-state`.

zebra-scanner/Cargo.toml
teor2345: I added a follow up task to the PR and crate for this.

oxarbitrage:
Should be merged after the release so i added the do not merge label.

deny.toml
```diff
@@ -75,6 +75,9 @@ skip-tree = [
     # wait for indexmap, toml_edit, serde_json, tower to upgrade
     { name = "hashbrown", version = "=0.12.3" },
 
+    # wait for jsonrpc-http-server to upgrade
+    { name = "redox_syscall", version = "=0.2.16" },
+
```
teor2345: Nitpick: This duplicate dependency exception is already present at line 70. Was this accidentally caused by a git merge?

teor2345:
```suggestion
```

zebra-scan/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+rand = "0.8.5"
+bls12_381 = "0.8.0"
+jubjub = "0.10.0"
+ff = "0.13.0"
+group = "0.13.0"
+tokio = { version = "1.33.0", features = ["test-util"] }
+
+zebra-state = { path = "../zebra-state", version = "1.0.0-beta.30" }
+zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.30" }
+zebra-test = { path = "../zebra-test", version = "1.0.0-beta.30" }
```
teor2345: We don't use versions for internal dev dependencies, to avoid release issues with circular dependencies.

For example:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/tower-batch-control/Cargo.toml#L49-L50

```suggestion
zebra-state = { path = "../zebra-state" }
zebra-chain = { path = "../zebra-chain" }
zebra-test = { path = "../zebra-test" }
```

zebra-scan/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Shielded transaction scanner for the Zcash blockchain.
+
+#![doc(html_favicon_url = "https://zfnd.org/wp-content/uploads/2022/03/zebra-favicon-128.png")]
+#![doc(html_logo_url = "https://zfnd.org/wp-content/uploads/2022/03/zebra-icon.png")]
+#![doc(html_root_url = "https://doc.zebra.zfnd.org/zebra_state")]
```
teor2345: This URL was removed by PR #7886 and the site no longer exists. The crate name is also wrong.

```suggestion
#![doc(html_root_url = "https://docs.rs/zebra_scan")]
```

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+    },
+    zip32::{AccountId, DiversifiableFullViewingKey, ExtendedSpendingKey},
+};
+
+use color_eyre::Result;
+
+use rand::{rngs::OsRng, RngCore};
+
+use ff::{Field, PrimeField};
+use group::GroupEncoding;
```
teor2345, line 35: Optional: our usual dependency groupings are: std, external dependencies, zebra dependencies, crate dependencies, module dependencies. But this is optional.

I like the split of `zcash_` and I think we should keep it.

```suggestion
use color_eyre::Result;
use ff::{Field, PrimeField};
use group::GroupEncoding;
use rand::{rngs::OsRng, RngCore};

use zcash_client_backend::{
    proto::compact_formats::{
        self as compact, ChainMetadata, CompactBlock, CompactSaplingOutput, CompactSaplingSpend,
        CompactTx,
    },
    scanning::scan_block,
};
use zcash_note_encryption::Domain;
use zcash_primitives::{
    block::BlockHash,
    consensus::{BlockHeight, Network},
    constants::SPENDING_KEY_GENERATOR,
    memo::MemoBytes,
    sapling::{
        note_encryption::{sapling_note_encryption, SaplingDomain},
        util::generate_random_rseed,
        value::NoteValue,
        Note, Nullifier, SaplingIvk,
    },
    zip32::{AccountId, DiversifiableFullViewingKey, ExtendedSpendingKey},
};

```

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    let db = read_only_state_service.db();
+
+    // use the tip as starting height
+    let mut height = latest_chain_tip.best_tip_height().unwrap();
+
+    let mut transactions_found = 0;
+    let mut transactions_scanned = 0;
+    let mut blocks_scanned = 0;
+    while let Some(block) = db.block(height.into()) {
```
teor2345: Optional/Nitpick: Accessing the state database directly is ok in the tests, but it won't work in production code.

Using `Request::Block` or a watch channel on the state service is the correct way to get all blocks in production code. This change is optional in this test, but if you copied this code into production in a later PR, it will need to be fixed there.

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+
+        // `fee` is not checked by the `scan_block` function. It is allowed to be unset.
+        // <https://docs.rs/zcash_client_backend/latest/zcash_client_backend/proto/compact_formats/struct.CompactTx.html#structfield.fee>
+        fee: 0,
+
+        spends: tx
+            .sapling_nullifiers()
+            .map(|nf| CompactSaplingSpend {
+                nf: <[u8; 32]>::from(*nf).to_vec(),
+            })
```
teor2345, line 225: Optional: this would be easier to read and more reusable as a `Nullifier::to_bytes()` method in
https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/note/nullifiers.rs

Example:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-chain/src/value_balance.rs#L334-L335

### Things we can't do

It would be nice to `impl From<sapling::Nullifier> for CompactSaplingSpend`, and `impl From<sapling::Output> for CompactSaplingOutput`, but that would move these dependencies to `zebra-chain`. (Due to Rust's orphan trait implementation rules.)

So we could do that behind a feature in `zebra-chain`, but it seems out of scope. Let's see if it's needed later.

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+        // `actions` is not checked by the `scan_block` function.
+        actions: vec![],
+    }
+}
+
+/// Create a fake compact block with provided fake account data.
+// This is a copy of zcash_primitives `fake_compact_block` where the `value` argument was changed to
+// be a number for easier conversion:
+// https://github.com/zcash/librustzcash/blob/zcash_primitives-0.13.0/zcash_client_backend/src/scanning.rs#L635
+// We need to copy because this is a test private function upstream.
```
teor2345, line 258: Note: this should be ok because the licenses are the same, and we've credited the original authors by saying where we got the code from.

teor2345 (dismissed):
Looks great, thanks!

Using the correct docs URL is a blocker. There are some other things that could be cleaned up to avoid confusion later.

I've resolved most of my optional suggestions, please only look at them if you have time and motivation to polish this code. (I'm guessing some parts are going to end up as production code.)

teor2345:
I don't need to review this again, so @arya2 feel free to approve it after any minor changes.

zebra-scan/src/tests.rs
teor2345: This is going to cause a merge conflict, so let's not.

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    let mut transactions_found = 0;
+    let mut transactions_scanned = 0;
+    let mut blocks_scanned = 0;
+    while let Some(block) = db.block(height.into()) {
+        // We fake the sapling tree size to 1 because we are not in Sapling heights.
+        let sapling_tree_size = 1;
+        let orchard_tree_size = db
+            .orchard_tree_by_hash_or_height(height.into())
+            .expect("should exist")
```
teor2345: Nitpick: `expect()` messages should explain what condition is broken when the test panics

```suggestion
            .expect("each state block must have a sapling tree")
```

zebra-scan/src/tests.rs
```diff
@@ -0,0 +1,371 @@
...
+            .count();
+
+        let chain_metadata = ChainMetadata {
+            sapling_commitment_tree_size: sapling_tree_size
+                .try_into()
+                .expect("position should fit in u32"),
+            orchard_commitment_tree_size: orchard_tree_size
+                .try_into()
+                .expect("position should fit in u32"),
```
teor2345: Nitpick: This is part of the protocol spec, so it's not a "should", it's a requirement.

```suggestion
                .expect("sapling position is limited to u32::MAX"),
            orchard_commitment_tree_size: orchard_tree_size
                .try_into()
                .expect("orchard position is limited to u32::MAX"),
```

teor2345 (approved):
Thanks!

PR #7770, opened 2023-10-18
fix(state): Avoid panicking in zebra-state initialization when checkpoints are missing

arya2:
## Motivation

Zebra should not panic without checkpoints.

Qedit asked about how to set up a local testnet with a different genesis block and a different network ID for local testing, one approach would be to modify the `genesis_hash()` return value for `Testnet` and clear initial/cached testnet peers.

## Solution

- Always store queued UTXOs when there are no checkpoints 
  (or when the `max_checkpoint_height `is lower than `checkpoint_verify_concurrency_limit`)

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service.rs
+++ b/zebra-state/src/service.rs
@@ -389,8 +389,7 @@ impl StateService {
         let full_verifier_utxo_lookahead = max_checkpoint_height
             - HeightDiff::try_from(checkpoint_verify_concurrency_limit)
                 .expect("fits in HeightDiff");
-        let full_verifier_utxo_lookahead =
-            full_verifier_utxo_lookahead.expect("unexpected negative height");
+        let full_verifier_utxo_lookahead = full_verifier_utxo_lookahead.unwrap_or(block::Height(0));
 
         let non_finalized_state_queued_blocks = QueuedBlocks::default();
         let pending_utxos = PendingUtxos::default();


zebra-state/src/service.rs
```diff
@@ -389,8 +389,7 @@ impl StateService {
         let full_verifier_utxo_lookahead = max_checkpoint_height
             - HeightDiff::try_from(checkpoint_verify_concurrency_limit)
                 .expect("fits in HeightDiff");
-        let full_verifier_utxo_lookahead =
-            full_verifier_utxo_lookahead.expect("unexpected negative height");
+        let full_verifier_utxo_lookahead = full_verifier_utxo_lookahead.unwrap_or(block::Height(0));
```
teor2345, line 393: > Zebra should not panic without checkpoints.

I think this code change is helpful, but the explanation isn't correct.

Zebra must have checkpoints up to `mandatory_checkpoint_height()`. If it doesn't, then it could break consensus rules:
https://github.com/ZcashFoundation/zebra/blob/290ccf2d94e68b80008517614b4f3873e06a93fd/zebra-chain/src/parameters/network.rs#L96-L102

Is there a specific situation or bug that this change fixes?
That might be helpful context to put in the PR description.

teor2345, line 393: Also, there is a significant amount of consensus code that isn't implemented due to the mandatory checkpoint, including:
* difficulty adjustment for heights less than 17
* slow start block subsidy
* activating NU5 or NU6 at genesis or block 1 (which allows transactions v4,v5,v6)
* contextual validation of the genesis block (it has to be a checkpoint)

If the eventual goal of this PR series is to make all those things work, then this list should become a tracking issue. Until we do all those things, Zebra won't be compatible with `zcashd` in an alternate testnet.

However, if the eventual goal is to test ZSAs, then forking the current testnet chain by adding NU6 at a nearby height might be more reliable (and much faster to implement, although slower to sync).

mpguerra:
Would it be worth setting up a call with them to discuss potential approaches? Seems like this is a problem that could benefit from some brainstorming with the whole team. Also, if we need to do anything on our side we should plan for it and schedule accordingly :)

teor2345:
> Would it be worth setting up a call with them to discuss potential approaches? Seems like this is a problem that could benefit from some brainstorming with the whole team. Also, if we need to do anything on our side we should plan for it and schedule accordingly :)

Yes, I think a good next step is to check what their goals actually are, then offer them some alternatives. Until we know that, it's tricky to plan for these kinds of changes.

mpguerra:
Are there any next steps for this PR or are we just closing without merging?

arya2:
> Are there any next steps for this PR or are we just closing without merging?

Let's close without merging and revisit this if needed.

teor2345:
> Are there any next steps for this PR or are we just closing without merging?

The next step is a meeting or discussion with the developers working on ZSAs, so we can help them work through their testing needs for their Zebra integration.

Issue #7613, opened 2023-09-24
feature: Compile sprout and sapling parameters into the Zebra binary


teor2345:
### Motivation

The `zcashd` team have recently started compiling the Sprout and Sapling parameters into their binary:
> What we are now doing is we have bundled the Sprout verification and the sapling proving and verification parameters into the Zcashd binary.
> 
> Whereas now the largest component of these is the sapling spend parameters, which are about 50 megabytes.
> 
> If you take the existing release binaries and add about 16 megabytes, that's going to be approximately the right ballpark.

https://github.com/ZcashCommunityGrants/arboretum-notes/blob/main/AllArboristCallNotes/Arborist%20Call%2061-Notes.md#1-zebrad-status---zebra-update

We've had issues with the parameter download server being unreliable or rate-limited before. They are likely to get worse now that ECC is no longer using it on a regular basis.

### Specifications

This `bundled-prover` feature allows access to bundled Sapling parameters:
https://docs.rs/zcash_proofs/latest/zcash_proofs/prover/struct.LocalTxProver.html#method.bundled

I'm not sure if there is a Rust crate for bundled Sprout verification parameters.

### Complex Code or Requirements

This might make our CI builds slower to link, or use more disk space.

### Testing

Our existing unit tests cover this code, but we might want to check that our integration tests have run on Sprout and Sapling transactions. We can do that by checking the block heights from the cached state to the tip had those transactions.

Sapling is common, but Sprout is rare now.

### Related Work

We might need to do this as after the zcashd 5.7.0 release, depending on ECC's API deprecations and download server support guarantees.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7613?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

teor2345:
What I still need to do today is:
- load Sprout parameters from the binary. I'm going to do this the same way as `zcashd`
- delete unused parameter download code and dependency features, to resolve unused code warnings
- remove the known download issue from the README
- update any docs that mention parameter downloads
- Remove parameter cache steps in CI Multi-OS test and coverage workflows:
  - `ci-unit-tests-os.yml`
  - `ci-coverage.yml`

(These changes were all required to clean up the Rust code, because they depended on the download examples, which depended on the download methods.)

[...]

PR #7797, opened 2023-10-22
change(devops): Use correct name for cd-build-zcash-params.yml

teor2345:
## Motivation

This isn't a reusable workflow, so it should either start with `ci` or `cd`. Since it is only run on the `main` branch, `cd` seems better.

## Review

### Reviewer Checklist


teor2345:
This workflow will be deleted by PR #7613 or the PR after it, so changing its name will just cause merge conflicts.

Issue #7738, opened 2023-10-12
docs: Add ticket templates for RPC methods and state format upgrades 


teor2345:
### Motivation

We have learnt a lot about reliably implementing RPCs and in-place state format upgrades over the past few years. We have a well-defined process for RPCs, and we need a well-defined process for state upgrades.

It might also help to upgrade the PR template to add a checklist for the PR author, because we often forget tests or doc updates in PRs.

### Rough Plan

- [x] #7833
- [ ] RPC ticket template:

Checklist:
- optional: unit tests on a fake state
- integration tests: either gRPC tests or tests on a populated state (not a fake state)
- snapshot tests
- manual diff tests

- [ ] State Format Upgrade template:

- integration tests for RPCs or validation that uses new format
- add these implementation and testing checklists:
  - https://github.com/ZcashFoundation/zebra/issues/7738#issuecomment-1774217120
  - https://github.com/ZcashFoundation/zebra/issues/7738#issuecomment-1774219966

- [x] PR template:

Author:
- add or update tests
- update function documentation

Reviewer:
- edit tests to say: "do the tests cover the goals of the ticket?"
- does it match the ticket scope?
- what is needed to merge this PR?
- can the PR be split? Can we open other PRs for cleanup?

### Review

We should get multiple developers to read and review these templates, to make sure we can all understand and use them.

### Related Work

The state upgrade template might reference the docs in #7737.

teor2345:
Here is an updated checklist which was taken out of the state docs:

### Implementation Steps

- [ ] update the [database format](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#current) in the Zebra docs
- [ ] increment the state minor version
- [ ] create a new module for the upgrade and its format checks
- [ ] refactor the write methods so they can be used for the upgrade and block writes
  - [ ] write the new format in the block write task
  - [ ] update older formats in the format upgrade task
- [ ] write detailed format checks for the new format (quick checks are optional)
- [ ] test that the new format works when fully syncing a new state, and updating an older state

See the [upgrade design docs](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#design) for more details.

teor2345:
For the state format upgrade template:

### Required Tests

- [ ] Randomised property tests on any new format serializations
- [ ] Unit tests for any calculations or data processing
- [ ] A state validity check that checks the entire format, or if that would take hours:
[...]

PR #7799, opened 2023-10-22
change(doc): Document state format upgrade implementation and testing

### LINKED ISSUE #7737: doc: Document how to implement state format upgrades
### Motivation

We have learnt a lot about in-place state format upgrades over the past few months, let's write the most important parts down.

### Rough Plan

Background:
- [x] explain two different ways to sync, full and upgrade

Design:
- [x] upgrade
  - [x] read methods must be able to read both formats
  - [x] write methods should be refactored so they are called from both the full sync and upgrade code, this prevents data inconsistencies
- [x] quick and detailed format checks
  - always runs on startup, shutdown, and after upgrades
  - runs periodically in CI using a custom config

Development:
- [x] how to edit or reset the state version file during development for manual testing

### Review

We should get multiple developers to read and review this documentation, to make sure we all understand and agree with it.

teor2345:
## Motivation

We want to document what we've learned from the last 3 in-place state upgrades.

Close #7737

## Solution

Document:
- how the state works
- block verification modes
- why state upgrades are needed
- why they are tricky and need special care
- how to implement them more reliably (see also the checklist link below)
- what testing is needed

Also update some outdated format docs, and add some headings.

There are also ticket templates coming in #7738, here is a draft implementation checklist for state upgrades:
https://github.com/ZcashFoundation/zebra/issues/7738#issuecomment-1774217120

## Review

We should get two reviews on this one.

### Reviewer Checklist

## Follow Up Work

Reference these docs from a state upgrade ticket template.

### DIFF
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -1,5 +1,23 @@
 # Zebra Cached State Database Implementation
 
+## Current Implementation
+
+### Verification Modes
+[verification]: #verification
+
+Zebra's state has two verification modes:
+- block hash checkpoints, and
+- full verification.
+
+This means that verification uses two different codepaths, and they must produce the same results.
+
+By default, Zebra uses as many checkpoints as it can, because they are more secure against rollbacks
+(and some other kinds of chain attacks). Then it uses full verification for the last few thousand
+blocks.
+
+When Zebra gets more checkpoints in each new release, it checks the previously verified cached
+state against those checkpoints. This checks that the two codepaths produce the same results.
+
 ## Upgrading the State Database
 
 For most state upgrades, we want to modify the database format of the existing database. If we
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -52,7 +52,8 @@ To reduce code and testing complexity:
 
 Here are the goals of in-place upgrades:
 - avoid a full download and rebuild of the state
-- the previous state format must be able to be loaded by the new state
+- Zebra must be able to upgrade the format from previous minor or patch versions of its disk format
+  (Major disk format versions are breaking changes. They create a new empty state and re-sync the whole chain.)
   - this is checked the first time CI runs on a PR with a new state version.
     After the first CI run, the cached state is marked as upgraded, so the upgrade doesn't run
     again. If CI fails on the first run, any cached states with that version should be deleted.
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -30,7 +30,7 @@ which can take days.
 
 Blocks can be written to the database via two different code paths, and both must produce the same results:
 
-- Upgrading pre-existing zebra-state databases/cache to the latest format
+- Upgrading a pre-existing database to the latest format
 - Writing newly-synced blocks in the latest format
 
 This code is high risk, because discovering bugs is tricky, and fixing bugs can require a full reset
--- a/book/src/dev/state-db-upgrades.md
+++ b/book/src/dev/state-db-upgrades.md
@@ -213,8 +213,13 @@ We use the following rocksdb column families:
 | `history_tree`                     | `()`                   | `NonEmptyHistoryTree`         | Update  |
 | `tip_chain_value_pool`             | `()`                   | `ValueBalance`                | Update  |
 
-Zcash structures are encoded using `ZcashSerialize`/`ZcashDeserialize`.
-Other structures are encoded using `IntoDisk`/`FromDisk`.
+### Data Formats
+[rocksdb-data-format]: #rocksdb-data-format
+
+We use big-endian encoding for keys, to allow database index prefix searches.
+
+Most Zcash protocol structures are encoded using `ZcashSerialize`/`ZcashDeserialize`.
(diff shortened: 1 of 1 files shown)


book/src/dev/state-db-upgrades.md
```diff
@@ -1,18 +1,63 @@
...
 which can take days.
 
+### Writing Blocks to the State
+[write-block]: #write-block
+
+Zebra's state has two methods of getting to the latest format:
+- syncing from empty with the latest format, and
+- upgrading from a previous format to the latest format.
+
+This means that writing blocks uses two different codepaths, and they must produce the same results.
```
arya2: Optional/Nitpick: I think "both must produce the same results" should go in the first line of this section.

```suggestion
Blocks can be written to the database via two different code paths, and both must produce the same results:

- Upgrading pre-existing zebra-state databases/cache to the latest format
- Writing newly-synced blocks in the latest format

```

book/src/dev/state-db-upgrades.md
```diff
@@ -1,18 +1,63 @@
...
+To reduce code and testing complexity:
+- when a previous Zebra version opens a newer state, the entire state is considered to have that lower version, and
+- when a newer Zebra version opens an older state, each required upgrade is run on the entire state.
+
 ### In-Place Upgrade Goals
+[upgrade-goals]: #upgrade-goals
 
+Here are the goals of in-place upgrades:
 - avoid a full download and rebuild of the state
 - the previous state format must be able to be loaded by the new state
```
arya2: Nitpick/Optional:

```suggestion
- Zebra's state must be able to upgrade data from the previous major version of its disk format
  and load data from the previous minor or patch versions of its disk format
```

book/src/dev/state-db-upgrades.md
```diff
@@ -62,17 +116,50 @@ But we need to use iterators for some operations, so our alternatives are (in pr
...
+#### Manually Triggering a Format Upgrade
+[manual-upgrade]: #manual-upgrade
+
+Zebra stores the current state minor and patch versions in a `version` file in the database
+directory. This path varies based on the OS, major state version, network, and config.
+
+For example, the default mainnet state version on Linux is at:
+`~/.cache/zebra/state/v25/mainnet/version`
+
+To upgrade from `v25.0.0`, delete the version file.
```
arya2: Nitpick/Optional:

```suggestion
To upgrade a cached Zebra state from `v25.0.0` to the latest disk format, delete the version file.
```

book/src/dev/state-db-upgrades.md
```diff
@@ -62,17 +116,50 @@ But we need to use iterators for some operations, so our alternatives are (in pr
...
+To upgrade from a specific version `v25.x.y`, edit the file so it contains `x.y`.
+
+Editing the file will trigger a re-upgrade over an existing state. Re-upgrades can hide format bugs.
+For example, if the old code was correct, and the new code skips blocks, the validity checks won't
+find that bug.
```
arya2: Optional/Nitpick:

```suggestion
Editing the file and running Zebra will trigger a re-upgrade over an existing state.
Re-upgrades can hide format bugs. For example, if the old code was correct, and the
new code skips blocks, the validity checks won't find that bug.
```

book/src/dev/state-db-upgrades.md
```diff
@@ -200,6 +296,7 @@ Each column family handles updates differently, based on its specific consensus
   - Each key-value entry is created once.
   - Keys can be deleted, but values are never updated.
   - Code called by ReadStateService must ignore deleted keys, or use a read lock.
+  - We avoid deleting keys, and avoid using iterators on Delete column families, for performance.
```
arya2:
```suggestion
  - We avoid deleting keys, and avoid using iterators on `Delete` column families, for performance.
```

arya2 (dismissed):
Looks good!

book/src/dev/state-db-upgrades.md
teor2345: Thank you for the detailed review and feedback!

This really good feedback, even though it is not quite right - that means we need to clearly document the current behaviour.

```suggestion
- Zebra must be able to upgrade the format from previous minor or patch versions of its disk format
  (Major disk format versions are breaking changes. They create a new empty state and re-sync the whole chain.)
```

upbqdn (approved):

book/src/dev/state-db-upgrades.md
```diff
@@ -1,18 +1,64 @@
...
 For most state upgrades, we want to modify the database format of the existing database. If we
 change the major database version, every user needs to re-download and re-verify all the blocks,
 which can take days.
 
+### Writing Blocks to the State
+[write-block]: #write-block
+
+Blocks can be written to the database via two different code paths, and both must produce the same results:
+
+- Upgrading pre-existing zebra-state databases/cache to the latest format
```
upbqdn: ###### Optional Suggestion

Could we simplify this line to something like:

```suggestion
- Upgrading a pre-existing database to the latest format
```

upbqdn (dismissed):

book/src/dev/state-db-upgrades.md
teor2345: Sure!

teor2345:
This has had two full reviews now.

upbqdn (approved):

Issue #7680, opened 2023-10-05
Automatically open an issue when main branch release-related merges fail


mpguerra:
Releases:
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/docs-deploy-firebase.yml
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/docs-dockerhub-description.yml
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/release-drafter.yml

teor2345:
I removed https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/cd-build-zcash-params.yml because we deleted it when we put parameters in the Zebra binary,

mpguerra:
Not a priority

PR #7817, opened 2023-10-25
change(release): Make the latest tag point to the production build, rather than the build with experimental features

### LINKED ISSUE #7415: devops: Remove redundant tags in Docker Hub
## Motivation

Docker Hub currently displays six tags for the most recent release: https://hub.docker.com/r/zfnd/zebra/tags. Also, notice that various previous releases contain various numbers of redundant tags.

### Suggested Fixes

- We should figure out which ones are needed and remove the rest.
- We should make `latest` point to the default build, not the `experimental` build. But only after we've made `getblocktemplate-rpcs` the default build, to avoid breaking existing users.

teor2345:
## Motivation

We want to remove redundant DockerHub tags, and fix incorrect tags.

Close #7415

### Specifications

https://github.com/docker/metadata-action#flavor-input

## Solution

- ~~Stop publishing the `v1.x.y` tag for future releases~~
- ~~Stop trying to publish the `edge` tag (it never actually got published)~~ it only gets published in CI on `main` branch builds, which is what we want
- Make `latest` point to the production image, not experimental
- Make overlapping prod/experimental tags point to the production image
- Add a deprecation warning and its section to the changelog

Related Cleanups:
- Rearrange some changelog items
- Remove an OS CI workflow trigger when the docker build workflow changed

### Testing

We won't be able to test this until the release.

The expected behaviour is in the tag workflow comments:
https://github.com/ZcashFoundation/zebra/pull/7817/files#diff-5a5111df219975ea6b3b69567752b89fa0c670d5a67598674ad411da2ed96985R93-R102

And it should appear here:
https://hub.docker.com/r/zfnd/zebra/tags

## Review

This should get merged this week because it blocks the release.

### Reviewer Checklist

## Follow Up Work

- #7891
- #7892


.github/workflows/release-binaries.yml
```diff
@@ -18,29 +18,31 @@ jobs:
   # Each time this workflow is executed, a build will be triggered to create a new image
   # with the corresponding tags using information from git
 
-  # The image will be named `zebra:<semver>`
-  build:
-    name: Build Release Docker
+  # The image will be named `zebra:<semver>.experimental`
+  build-experimental:
```
teor2345, line 22: These are swapped to make sure tags with the same name point to the production image.

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,30 @@ jobs:
...
+            # Stopping publishing these tags is a breaking change for these use cases:
+            # - `latest`: always use the latest Zebra release when you pull or update
+            # - `1`: use the latest Zebra release, but require manual intervention for the next network upgrade
+            # - `1.x`: update to bug fix releases, but don't add any new features or incompatibilities
+            #
+            # semver automatically adds a "latest" tag, but only on stable releases
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=ref,event=tag
```
teor2345, line 98: `tag` is deleted to remove `v1.x.y`

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,30 @@ jobs:
...
-            type=ref,event=branch
-            type=ref,event=pr
+            # DockerHub release and CI tags.
+            # This tag makes sure tests are using exactly the right image, even when multiple PRs run at the same time.
             type=sha
-            # edge is the latest commit on the default branch.
-            type=edge,enable={{is_default_branch}}
```
teor2345, line 103: `edge` never worked

teor2345: This is wrong, `edge` is only for branch events.

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,30 @@ jobs:
...
+            # These CI-only tags support CI on PRs, the main branch, and scheduled full syncs. 
+            # These tags do not appear on DockerHub, because DockerHub images are only published on the release event.
+            type=ref,event=pr
+            type=ref,event=branch
```
teor2345, line 120:
```suggestion
            type=ref,event=branch
            type=edge,enable={{is_default_branch}}
```

gustavovalverde (commented):
I did not deep-dive into `uses: docker/metadata-action@v5` options, but can't this be accomplished without adding the extra `build-experimental`?

If possible, is there any extra complexity we're trying to avoid?

.github/workflows/release-binaries.yml
```diff
@@ -18,29 +18,31 @@ jobs:
...
+      features: ${{ format('{0} {1}', vars.RUST_PROD_FEATURES, vars.RUST_EXPERIMENTAL_FEATURES) }}
       rust_log: ${{ vars.RUST_LOG }}
     # This step needs access to Docker Hub secrets to run successfully
     secrets: inherit
 
-  # The image will be named `zebra:<semver>.experimental`
-  build-experimental:
-    name: Build Experimental Features Release Docker
+  # The image will be named `zebra:<semver>`
+  # It should be built last, so overlapping tags point to the production build, not the experimental build.
```
teor2345:
```suggestion
  # It should be built last, so tags with the same name point to the production build, not the experimental build.
```

.github/workflows/release-binaries.yml
```diff
@@ -18,29 +18,31 @@ jobs:
...
+  build:
+    name: Build Release Docker
```
teor2345, line 38:
```suggestion
    name: Build Release Docker
    # Run this build last, regardless of whether experimental worked
    needs: build-experimental
    if: always()
```

teor2345:
> I did not deep-dive into `uses: docker/metadata-action@v5` options, but can't this be accomplished without adding the extra `build-experimental`?

This change does not add an extra job, it makes the production build run after the experimental build.

Previously, the `latest` tag pointed to whichever job took longer, which is usually experimental because it compiles more code. Now the `latest` tag is only generated for the production build.

But I also changed the order just in case we add a tag with the same name in future.

(I forgot to use "needs" to make sure they run in the correct order, but that's fixed now.)

> If possible, is there any extra complexity we're trying to avoid?

Is there a specific change you want to make to avoid complexity?

Removing the experimental build is out of scope, and we'll want an experimental build again as soon as we have some blockchain scanning in Zebra. So I'm not sure that's necessary. (Also it's a breaking change for anyone using experimental tags.)

gustavovalverde:
> This change does not add an extra job, it makes the production build run after the experimental build.

Yeah, my bad. I just got confused with the GitHub diff

> Is there a specific change you want to make to avoid complexity?

No, this is ok. I'll leave the final review to the assigned reviewer.

oxarbitrage (commented):
I think that the current and proposed tags for this PR might be excessive when compared to the zcash docker image, which maintains a smaller set of tags, as seen here: [https://hub.docker.com/r/electriccoinco/zcashd/tags](https://hub.docker.com/r/electriccoinco/zcashd/tags).

My suggestion would be to simplify the tag structure. I propose the following tags:

- `latest`
- `v1.4.0`
- `v1.4.0-experimental`

Here are my reasons for this recommendation:

1. **Minimizing Tag Count**: Fewer tags make it easier to manage and navigate, reducing potential confusion.

2. **Avoiding Ambiguity**: Using tags like `1.4` or `1` can lead to confusion.

3. **Clarifying `sha-13230d0.experimental`**: If the purpose of tags like `sha-13230d0.experimental` is unclear, it's best to either rename them to something more descriptive or avoid publishing them.

4. **Prefixing with `v`**: Prefixing version tags with `v` (e.g., `v1.4.0` instead of `1.4.0`) seems to be a common convention and looks better to me.

5. **Docker-Specific Information**: Consider including Docker-specific details in the [Docker Hub overview](https://hub.docker.com/r/zfnd/zebra) page instead of the README.

teor2345:
> I think that the current and proposed tags for this PR might be excessive when compared to the zcash docker image, which maintains a smaller set of tags, as seen here: [hub.docker.com/r/electriccoinco/zcashd/tags](https://hub.docker.com/r/electriccoinco/zcashd/tags).
> 
> My suggestion would be to simplify the tag structure.

I like the idea of this change, and I agree with the goal of simplifying things for users.

But it seems like a significant change that might be beyond the scope of the ticket or PR. It could also be a breaking change for some users.

To find out if it is a breaking change, we can check which tags are actually being used:
https://docs.docker.com/trusted-content/insights-analytics/#view-the-analytics-data

I don't have access to that data, but @gustavovalverde does, or he can give someone an account.

Then if it tag isn't used much, we can just delete it. If it is used, we can announce its deprecation in one release, and delete it in the next. That way users have time to change their scripts.

I also don't want to block the release on a longer investigation, so I suggest we make these changes in this PR:
- Stop publishing the `1.x.y` tag for future releases (this can't be a silently breaking change, because the tag never updates)
- Announce the deprecation of the other tags we want to delete in the changelog
- Make `latest` point to the production image, not experimental
- Make overlapping prod/experimental tags point to the production image

teor2345:
Actions from the meeting about this PR:

Minimise this PR.
Make a note in the changelog about proposed changes below (based on user feedback and docker image pull statistics).

Open tickets for:

Check docker pull statistics and consider deleting:
- `1.x.y` (use `v1.x.y` instead)
- `1.x`
- `1`
- `sha-xxxxxxx`

And add:
- `latest-experimental`

And Change:
- `.experimental` to `-experimental`

Deleting release tags is blocked by #7680 so we can see release failures easily.

Changing features is blocked by changing the tags, because we want the new features to appear under the new tag names:
- add all off-by-default features to the experimental build: `elasticsearch`, `journald`, `prometheus`, `filter-reload`
- add an `experimental` feature to `zebrad`
- add the `experimental` feature to the GitHub experimental feature variable, and the Docker builds/tests. Consider removing experimental from the OS builds because they are slow already.

Final set of tags:
- `latest`
- `v1.x.y`
- `latest-experimental`
- `v1.x.y-experimental`

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,31 @@ jobs:
...
-            type=ref,event=tag
-            type=ref,event=branch
-            type=ref,event=pr
+            # DockerHub release and CI tags.
+            # This tag makes sure tests are using exactly the right image, even when multiple PRs run at the same time.
             type=sha
-            # edge is the latest commit on the default branch.
+            # These CI-only tags support CI on PRs, the main branch, and scheduled full syncs. 
+            # These tags do not appear on DockerHub, because DockerHub images are only published on the release event.
+            type=ref,event=pr
```
teor2345, line 116:
```suggestion
            type=ref,event=tag
            type=ref,event=pr
```

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,31 @@ jobs:
...
+            # - `1.x.y`: always use the exact version, don't automatically upgrade
+            # - `sha-zzzzzz`: always use the exact commit (the same as `1.x.y`, but also used in CI and production image tests)
+            #
+            # Stopping publishing these tags is a breaking change for these use cases:
+            # - `latest`: always use the latest Zebra release when you pull or update
+            # - `1`: use the latest Zebra release, but require manual intervention for the next network upgrade
+            # - `1.x`: update to bug fix releases, but don't add any new features or incompatibilities
+            #
+            # semver automatically adds a "latest" tag, but only on stable releases
             type=semver,pattern={{version}}
```
teor2345, line 110:
```suggestion
            type=ref,event=tag
            type=semver,pattern={{version}}
```

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,31 @@ jobs:
...
+            # These DockerHub release tags support the following use cases:
+            # - `1.x.y`: always use the exact version, don't automatically upgrade
+            # - `sha-zzzzzz`: always use the exact commit (the same as `1.x.y`, but also used in CI and production image tests)
+            #
+            # Stopping publishing these tags is a breaking change for these use cases:
+            # - `latest`: always use the latest Zebra release when you pull or update
+            # - `1`: use the latest Zebra release, but require manual intervention for the next network upgrade
+            # - `1.x`: update to bug fix releases, but don't add any new features or incompatibilities
+            #
+            # semver automatically adds a "latest" tag, but only on stable releases
```
teor2345:
```suggestion
            # These DockerHub release tags support the following use cases:
            # - `latest`: always use the latest Zebra release when you pull or update
            # - `1`: use the latest Zebra release, but require manual intervention for the next network upgrade
            # - `1.x`: update to bug fix releases, but don't add any new features or incompatibilities
            # - `v1.x.y` or `1.x.y`: always use the exact version, don't automatically upgrade
            # - `sha-zzzzzz`: always use the exact commit (the same as `1.x.y`, but also used in CI and production image tests)
            #
            # Stopping publishing some tags is a silently breaking change:
            # - `1`: doesn't get expected new consensus-compatible releases or bug fixes
            # - `1.x`: doesn't get expected bug fixes
            #
            # `semver` adds a "latest" tag if `inputs.latest_tag` is `true`.
```

.github/workflows/sub-build-docker-image.yml
```diff
@@ -88,19 +92,34 @@ jobs:
...
+            #
+            # Stopping publishing some tags is a silently breaking change:
+            # - `1`: doesn't get expected new consensus-compatible releases or bug fixes
+            # - `1.x`: doesn't get expected bug fixes
+            #
+            # `semver` adds a "latest" tag if `inputs.latest_tag` is `true`.
+            type=ref,event=tag
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
```
teor2345:
```suggestion
            type=semver,pattern={{version}}
            type=semver,pattern={{major}}.{{minor}}
            type=semver,pattern={{major}}
            type=ref,event=tag
```

CHANGELOG.md
```diff
@@ -44,6 +44,13 @@ so it can't be used to retry failed downloads in `zebrad` 1.3.0 and earlier.
 
 We recommend upgrading to the latest Zebra release to avoid download issues in new installs.
 
+### DockerHub image tags
+
+Zebra 1.4.0 and later will not have a `v1.4.0` tag on [`zebrad` Docker images](https://hub.docker.com/r/zfnd/zebra/tags).
+If you need to stay on a particular version, use the `1.4.0`, `1.4` (patch releases only), or `1` (until NU6) tags.
```
teor2345:
```suggestion
Zebra currently publishes 11 [DockerHub tags](https://hub.docker.com/r/zfnd/zebra/tags) for each new release.
We want to reduce the number of DockerHub tags we publish at the next minor Zebra release 1.5.0.

Based on usage and user feedback, we could stop publishing:
- The `1` tag, which updates each release until NU6
- The `1.x` tag, which updates each patch release until the next minor release
- The `1.x.y` tag, which is the same as `v1.x.y`
- The `sha-xxxxxxx` tag, which is the same as `v1.x.y` (for production releases)

We also want to standardise experimental image tags to `-experimental`, rather than `.experimental`.

So for release 1.5.0, we might only publish these tags:
- `latest`
- `latest-experimental` (a new tag)
- `v1.5.0`
- `v1.5.0-experimental`

Please let us know if you need any other tags by [opening a GitHub ticket](https://github.com/ZcashFoundation/zebra/issues/new?assignees=&labels=C-enhancement%2CS-needs-triage&projects=&template=feature_request.yml&title=feature%3A+).
```

teor2345:
The new tickets are #7891 and #7892, and this PR is minimal.

gustavovalverde:
> To find out if it is a breaking change, we can check which tags are actually being used:
> https://docs.docker.com/trusted-content/insights-analytics/#view-the-analytics-data

This is the month with the most downloads:
image

oxarbitrage (approved):
I think this is good enough for now, thank you.

PR #7818, opened 2023-10-25
cleanup(deps): Remove unused deny.toml duplicate dependency exceptions

teor2345:
## Motivation

These dependency exceptions are unused because Zebra no longer depends on those crate versions. (Some other dependency has stopped using them.)

Details are here:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-unmatched-skip-root-warnings-in-check-denytoml-bans


arya2 (approved):
Thank you!

PR #7843, opened 2023-10-26
Make macOS a tier 2 supported platform in the docs

### LINKED ISSUE #6812: Process abort in state tests on some macOS versions with Rust 1.70

### LINKED ISSUE #7824: Tracking: security: Limit ability of synthetic nodes to spread throughout network. Credit: Ziggurat Team
### Motivation

"RT-S1 f1" from "Red Team report on Zcash testnet"

> Zebra accepts peers from its peers much more easily when compared with the zcashd node. This behavior allows synthetic nodes (which do not do much in the network) to spread quite fast throughout the testnet network.
> But at least Zebra doesn’t share all peers in the Addr message. It always shares a portion of its peers.

And

"RT-S1 f2" from "Red Team report on Zcash testnet"

> Bad behavior
> Zcashd also spreads synthetic nodes, but more slowly.
> The node will always share all IPs to which it has connections.
>
> Good behavior
> Time to accept new peers is much slower because:
> - a zcashd node randomly processes a single node per Addr message.
> - A zcashd node spreads peers more carefully - it will never share more nodes from the same IP. Once it decides which node (identified by IP and port) will be shared, it always shares that same node. Even if zcashd has more connections to nodes from the same IP, it will only share a single node in the peer list.

There is nothing particularly complex here.

## Potential Fixes

Before starting work, discuss the impact of each potential fix on this vulnerability with the team, and decide on 1-3 fixes that have the greatest impact.

### Required Changes

These changes must be made to fix the issues:
- [x] Limit the number of addresses taken from any one peer to less than half the outbound connection limit (#1869)
    - benefits: one peer can only provide roughly half of the address used for outbound connections (75 by default), simple change that can be made in one location
    - drawbacks: ignoring addresses means that startup is slower, and some nodes might struggle to connect or sync. This change needs a full sync speed test.
- [x] #7823
    - benefits: peers can't make repeated requests to get our entire address book, improves performance under load
    - drawbacks: it takes longer for peers to get valid addresses
- [x] #7951
    - benefits: prevents an attack where peers can repeatedly connect, supply an address in the version message, and fill the address book
    - drawbacks: we might lose those addresses when the peer sends an addr message (but addr messages should always contain that peer's address, so in some cases we might lose the remote address of inbound connections), these addresses are treated like gossiped addresses, so they have slightly higher priority

To deal with the drawbacks of these other changes, make this small change in the same PR as #1869:
  - [x] When limiting the number of taken addresses, keep the un-taken addresses in the connection cache in case we need more
    - Advantages: Zebra will find peers faster if we keep all of the addresses
    - Disadvantages: Addresses could become outdated

## Accepted Changes

These changes are useful but not strictly required:
- [ ] Remove the unused NeverAttemptedAlternate peer status
    - benefits: code is simpler
    - drawbacks: none (except for those already accepted in #7951)

## Needs More Analysis

These changes were accepted, but they've since turned out to be more complicated than expected:
- [ ] Always keep the latest unique addresses from each connected peer, regardless of which address message they were sent in
    - required analysis: before making this change, check how often the cache is truncated. If it is only rarely truncated, this  change isn't needed.
    - benefits: we always have the latest addresses from each peer, so we only drop excess addresses. There are no duplicate addresses in the cache. This improves connectivity.
    - drawbacks: complicates the code
- [ ] Remove `Option` from `MetaAddr.{services,untrusted_last_seen}`
    - required analysis: before making this change, list every MetaAddrChange that uses `None` for either of these fields, and decide on acceptable defaults
    - benefits: code is simpler
    - drawbacks: we have to supply defaults, but those defaults are probably correct (or at least acceptable)

## Rejected Changes

<details>
<summary>Changes that aren't needed or don't work</summary>

- [ ] When a peer is disconnected, send its cached addresses to the address book as `NeverAttemptedAlternate` (or a new status that is a lower priority than failed, or move `NeverAttemptedAlternate` to a lower priority than failed)
    - benefits: we don't drop addresses, so we can use them later if we can't connect to any other addresses
    - drawbacks: peers can fill our address book by connecting, sending addresses, then disconnecting
- [ ] send any excess addresses to the address book as NeverAttemptedAlternate when they're received, but ignore them in the outbound connector until there haven't been any other candidates for a few minutes
    - benefits: never loses addresses, seeks new addresses from different peers before using all the addresses from one peer
    - drawbacks: imitates the existing reconnection attempt ordering, delays could cause hangs, complicates outbound connector

</details>

### Testing

- Check that we are only taking a few addresses from each peer
- Check that repeated getaddr requests return the same results
- Manual inspection of the version address sending code
- Manual inspection of the address cache code
- Existing integration testing

### Related Work

- #7822

### LINKED ISSUE #7842: docs: Make macOS a tier 2 supported platform in the docs

mpguerra:
## Motivation

We recently re-enabled macos tests in #7834 

## Solution

Move relevant table row from "Tier 3" to "Tier 2"

### Reviewer Checklist

### DIFF
--- a/book/src/user/supported-platforms.md
+++ b/book/src/user/supported-platforms.md
@@ -34,6 +34,7 @@ For the full requirements, see [Tier 2 platform policy](platform-tier-policy.md#
 | -------|-------|-------|-------|-------
 | `x86_64-unknown-linux-gnu` | [GitHub ubuntu-latest](https://github.com/actions/virtual-environments#available-environments) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | N/A
 | `x86_64-unknown-linux-gnu` | [GitHub ubuntu-latest](https://github.com/actions/virtual-environments#available-environments) | 64-bit | [latest beta release](https://github.com/rust-lang/rust/blob/beta/src/version) | N/A
+| `x86_64-apple-darwin` | [GitHub macos-latest](https://github.com/actions/virtual-environments#available-environments) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | N/A
 
 ## Tier 3
 
@@ -46,5 +47,3 @@ For the full requirements, see [Tier 3 platform policy](platform-tier-policy.md#
 | platform | os | notes | rust | artifacts
 | -------|-------|-------|-------|-------
 | `aarch64-unknown-linux-gnu` | [Debian 11](https://www.debian.org/releases/bullseye/) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | N/A
-| `x86_64-apple-darwin` | [GitHub macos-latest](https://github.com/actions/virtual-environments#available-environments) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | N/A
-


arya2 (approved):
🎉

PR #7848, opened 2023-10-26
change(devops): Split dependency upgrades into 4 large categories

### LINKED ISSUE #7809: Split dependency upgrades into 4 large categories
## Motivation

Some of our Rust dependency groups have only one dependency upgrade in them in most weeks.

### Suggested Solution

Our plan is to limit dependency upgrades tp 4 PRs per week:
* ECC dependencies
* production Rust dependencies
* test/developer tool Rust dependencies, and
* DevOps dependencies

The DevOps changes will be auto-assigned to DevOps reviewers by our current CODEOWNERS file.

If there are only a few Rust updates, the PR will be auto-assigned to the people who usually review the crates that depend on those dependencies. That's the best we can do for now.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/issues/7421#issuecomment-1777890008_

arya2:
## Motivation

This PR merges the dependabot groups for crypto, async, log-time, concurrency, progress-bar, app, formats, and data-structures into prod.

~~It's still missing a group for DevOps dependencies.~~

Closes #7809.

### Specifications

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

## Solution

- Merge the groups for production dependencies

## Review

Not yet ready for review.

### Reviewer Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,6 @@
 version: 2
 updates:
+  # Rust section
   - package-ecosystem: cargo
     directory: '/'
     # serde, clap, and other dependencies sometimes have multiple updates in a week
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -28,8 +28,10 @@ updates:
             - "bs58"
             - "ripemd"
         # groups are limited to 10 items
-        crypto:
+        prod:
           patterns:
+        # crypto:
+        #   patterns:
             - "bellman"
```
arya2: This may get around a limit on the number of patterns:

```suggestion
            - "bellman|red*|jubjub|group|bls12_381|blake*|secp256k1|sha2|*25519*|rand*"
```

teor2345:
I was imagining a rewrite into something much simpler. Something like:

```yaml
# Rust section
ecc:
  patterns:
    # ecc but not address patterns
prod:
  dependency-type: "production"
  exclude-patterns:
    # ecc but not address patterns
dev:
  dependency-type: "development"
# Devops section
devops:
  patterns:
    - "*"
```

.github/dependabot.yml
teor2345: Why not use `dependency-type: "production"` with `exclude-patterns` for the ECC dependencies (excluding the address dependencies).

For example:
https://github.com/ZcashFoundation/zebra/pull/7848#issuecomment-1781955833

arya2: ~Dependabot doesn't list support for setting dependency-type to production or development with cargo, and~ I'm not sure if it ignores patterns after the first match.

Update: It was `allow` that doesn't support production/development dependency types with cargo. Let's try it and see if it works for groups.

Applied in https://github.com/ZcashFoundation/zebra/commit/b4486304baea6784647d97745d2eda3a4e0ee22.

teor2345 (approved):
Let's see how it goes?

We can test now because we have open dependency PRs.

teor2345:
This seems to have worked for devops:
- #7882

PR #7853, opened 2023-10-26
fix(network): Reconnect with peers after brief network interruption

### LINKED ISSUE #7772: bug: zebrad will not reconnect after an internet connection failure and restore
### What happened?

 I expected to see this happen: zebrad re-establishes a working connection after the internet connection is re-established.

Instead, this happened: zebrad continuously warns that 'chain updates have stalled, state height has not increased for # minutes'. Restarting the node fixes the issue.

Disconnecting and reconnecting the router attached to the devices from the internet (I'm filing this issue from the device in question at the same time zebrad is not reconnecting).

### Zebra logs

https://gist.github.com/autotunafish/30959aa159bf301c38af8f4480a206e2

peers
https://gist.github.com/autotunafish/79fdcdfcb20ce1b4eb6a27f24885babd

### Zebra Version

1.3.0

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

Linux john-Satellite-L775D 5.4.0-164-generic #181-Ubuntu SMP Fri Sep 1 13:41:22 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Additional information

_No response_

arya2:
## Motivation

Zebra currently will not reconnect after an internet connection failure.

Part of #7772.

## Solution

- Always attempt to make an outbound connection in `TimerCrawl`

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Investigate why it isn't sending a `MorePeers` message in [`PeerSet::poll_ready()`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/peer_set/set.rs#L1047)
- Fix that and revert this PR

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -910,6 +910,7 @@ where
             Ok(TimerCrawl { tick }) => {
                 let candidates = candidates.clone();
                 let demand_tx = demand_tx.clone();
+                let should_always_dial = active_outbound_connections.update_count() == 0;
 
                 let crawl_handle = tokio::spawn(
                     async move {
@@ -918,7 +919,7 @@ where
                             "crawling for more peers in response to the crawl timer"
                         );
 
-                        crawl(candidates, demand_tx, true).await?;
+                        crawl(candidates, demand_tx, should_always_dial).await?;
 
                         Ok(TimerCrawlFinished)
                     }
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -896,7 +896,7 @@ where
                             // There weren't any peers, so try to get more peers.
                             debug!("demand for peers but no available candidates");
 
-                            crawl(candidates, demand_tx).await?;
+                            crawl(candidates, demand_tx, false).await?;
 
                             Ok(DemandCrawlFinished)
                         }
@@ -918,7 +918,7 @@ where
                             "crawling for more peers in response to the crawl timer"
                         );
 
-                        crawl(candidates, demand_tx).await?;
+                        crawl(candidates, demand_tx, true).await?;
 
                         Ok(TimerCrawlFinished)
                     }
@@ -957,11 +957,12 @@ where
 }
 
 /// Try to get more peers using `candidates`, then queue a connection attempt using `demand_tx`.
-/// If there were no new peers, the connection attempt is skipped.
+/// If there were no new peers and `should_always_dial` is false, the connection attempt is skipped.
 #[instrument(skip(candidates, demand_tx))]
 async fn crawl<S>(
     candidates: Arc<futures::lock::Mutex<CandidateSet<S>>>,
     mut demand_tx: futures::channel::mpsc::Sender<MorePeers>,
+    should_always_dial: bool,
 ) -> Result<(), BoxError>
 where
     S: Service<Request, Response = Response, Error = BoxError> + Send + Sync + 'static,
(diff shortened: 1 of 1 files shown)


zebra-network/src/peer_set/initialize.rs
```diff
@@ -918,7 +918,7 @@ where
                             "crawling for more peers in response to the crawl timer"
                         );
 
-                        crawl(candidates, demand_tx).await?;
+                        crawl(candidates, demand_tx, true).await?;
```
teor2345: I think this solution would fix this specific bug, but it would also cause Zebra to reach its maximum number of outbound connections after 1-2 hours. (The limit is 75 and there is a crawl every minute.)

At the moment, Zebra only makes new connections when it gets new peers or doesn't have enough peers for its requests. So that limits its outbound connections.

Instead, we could always add demand when `active_outbound_connections.update_count()` is below some fixed number of peers. The existing initial peer connection limit seems like a good number, or maybe half of that (rounded up).

zebra-network/src/peer_set/initialize.rs
```diff
@@ -896,7 +896,7 @@ where
                             // There weren't any peers, so try to get more peers.
```
teor2345, line 896: This PR doesn't fix these underlying bugs in the DemandHandshakeOrCrawl, or the tasks that trigger it:
- the syncer is not logging syncs or timeouts, it seems to be hung
- the mempool crawler isn't logging. Is it hung, has it stopped because we've fallen behind the tip (I didn't check the logs in the ticket for this), or do we need to add logs so we can see it being active?

If there is demand from the mempool crawler, then there aren't any valid peers, but there should be. So that could be another bug.

teor2345 (commented):
This fix might work, but there are no new tests, and no comments about manual tests. So I can't be sure. It also has some network load drawbacks that we should fix.

There are other bugs we might need to follow up, but some haven't been fully analysed, and they don't have tickets yet. Can you work out if we need to add logs to diagnose them? This PR would be a good place to add those logs.

zebra-network/src/peer_set/initialize.rs
teor2345: To be clear, we don't need to fix everything in this PR, just enough to resolve the disconnection bug. But let's open tickets for the rest, so we don't forget.

zebra-network/src/peer_set/initialize.rs
arya2: It checks that it's less than `peerset_outbound_connection_limit()` here: https://github.com/ZcashFoundation/zebra/blob/fix-reconnect-bug/zebra-network/src/peer_set/initialize.rs#L838

> Zebra only makes new connections when it gets new peers or doesn't have enough peers for its requests.

It was already demanding `MorePeers` whenever the crawl finds some peer addresses, I was hoping not to change its typical behaviour.

arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/7853/commits/5f3bf474307b9935ad409b536719690f2d9062fe.

arya2: I wonder if it updates the connection tracker correctly.

teor2345 (approved):
Thanks! We might want to raise the minimum number of peers later, but 1 is an acceptable minimum for now.

zebra-network/src/peer_set/initialize.rs
teor2345: That requires the peer set to drop failed connections, I'll open a PR for that soon.

PR #7855, opened 2023-10-26
change(state): Refactor the naming of note commitment subtrees

### LINKED ISSUE #7463: Rename subtree methods and types to "subtree root"

upbqdn:
## Motivation

Close #7463.

## Solution

#7463 mentions that "_We are storing subtree roots and end heights in the subtree indexes, not the subtree trees themselves._" I thought of renaming `NoteCommitmentSubtree` to `SubtreeRoot`, but there were at least two problems with that:

1. There is also the struct `NoteCommitmentSubtreeData`, which I wasn't sure how to rename.
2. `NoteCommitmentSubtree` has a field `node`, which I had renamed to `root`, so we would end up with something like `subtree_root.root`.

Given the issues above, I decided to do a simpler refactor:

1. Rename the generic type `Node` in `NoteCommitmentSubtree` to `Root`.
2. Rename the field `node` in `NoteCommitmentSubtree` to `root`.
3. Rename the field `end` in `NoteCommitmentSubtree` to `end_height` to match the name in the RPC so that the field doesn't need to be renamed for the RPC.

After this refactor, I didn't find any docs that would require changes. 

Also, if we ever need to store whole subtrees instead of only roots, we can use the already existing struct `NoteCommitmentSubtree`, but keep the RPC that currently uses this struct intact because it actually uses `NoteCommitmentSubtreeData`. 

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/subtree.rs
+++ b/zebra-chain/src/subtree.rs
@@ -55,21 +55,21 @@ impl From<NoteCommitmentSubtreeIndex> for u64 {
 /// with its associated block height and subtree index.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub struct NoteCommitmentSubtree<Root> {
+pub struct NoteCommitmentSubtree<SubtreeRoot> {
     /// Index of this subtree
     pub index: NoteCommitmentSubtreeIndex,
     /// Root of this subtree.
-    pub root: Root,
+    pub root: SubtreeRoot,
     /// End boundary of this subtree, the block height of its last leaf.
     pub end_height: Height,
 }
 
-impl<Root> NoteCommitmentSubtree<Root> {
+impl<SubtreeRoot> NoteCommitmentSubtree<SubtreeRoot> {
     /// Creates new [`NoteCommitmentSubtree`]
     pub fn new(
         index: impl Into<NoteCommitmentSubtreeIndex>,
         end_height: Height,
-        root: Root,
+        root: SubtreeRoot,
     ) -> Self {
         let index = index.into();
         Self {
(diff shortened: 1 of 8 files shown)


upbqdn:
I'm not actually sure about the solution in this PR. Do others have suggestions for better solutions?

arya2 (dismissed):
Looks good, thank you!

zebra-chain/src/subtree.rs
```diff
@@ -50,61 +50,64 @@ impl From<NoteCommitmentSubtreeIndex> for u64 {
...
 // - consider defining sapling::SubtreeRoot and orchard::SubtreeRoot types or type wrappers,
 //   to avoid type confusion between the leaf Node and subtree root types.
-// - rename the `Node` generic to `SubtreeRoot`
 
 /// Subtree root of Sapling or Orchard note commitment tree,
 /// with its associated block height and subtree index.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub struct NoteCommitmentSubtree<Node> {
+pub struct NoteCommitmentSubtree<Root> {
```
arya2: Optional/Nitpick: Rename this to `SubtreeRoot` to avoid confusion with other sapling/orchard `Root` types.

upbqdn: Done in [a278777](https://github.com/ZcashFoundation/zebra/pull/7855/commits/a278777decfe04dfc265badd6c1049ab2697b83e).

arya2 (approved):
Thank you!

Issue #7822, opened 2023-10-25
Tracking: security: Limit ability of synthetic nodes to take up connection slots. Credit: Ziggurat Team 


mpguerra:
### Motivation

"RT-S1 f3" from "Red Team report on Zcash testnet"

> Once a connection is established with a synthetic node, a connection can remain active forever. That could easily lead to denial of service when all connection slots would be occupied by synthetic nodes and rightful users wouldn’t have access to the node.
> The problem arises when we take into account that synthetic nodes can do almost nothing so it’s cheap (that’s why it can increase the risk of exploitation to medium) to set up thousands of nodes just to connect to many nodes and occupy the connection slots.
 
Also "Observation #5" from "Red Team report on Zcash testnet":

> Synthetic nodes can keep the connection alive without doing any actual blockchain work forever. Sometimes a synthetic node does get disconnected, but a quick reconnect solves the issue.

### Complex Code or Requirements

There are two scenarios here:
1. no load: peers that do nothing, provide no useful information, and aren't syncing from us should get disconnected
2. overload: peers that block readiness by constantly sending inbound requests should get disconnected (follow up to #7772, existing overload defence is global, the only peer-specific defence is the heartbeat)

## Potential Fixes

Before starting work, discuss the impact of each potential fix on this vulnerability with the team, and decide on 1-3 fixes that have the greatest impact:
- [ ] Increase the minimum number of connected peers to the initial target peerset size, so that Zebra always has a variety of peers, and the first/newest/fastest peers don't take up all the slots (also helps with #7824)
 https://github.com/ZcashFoundation/zebra/blob/7e7f989545bac06ee674547425ac0ac907cc0bce/zebra-network/src/peer_set/initialize.rs#L913
- [ ] Retry failed peers only if we've successfully connected to them, don't believe `untrusted_last_seen` https://github.com/ZcashFoundation/zebra/blob/7e7f989545bac06ee674547425ac0ac907cc0bce/zebra-network/src/meta_addr.rs#L639-L640
- [ ] Retry other failed peers (untrusted or no last seen time) if their last attempt is longer than `MAX_PEER_ACTIVE_FOR_GOSSIP` (or maybe just 10+ minutes?) but less than `MAX_RECENT_PEER_AGE` 
- [ ] #7852
- [x] #1875
- [ ] If we're not getting any new blocks in the progress task, send a `ChangePeerConnections` request to the network to disconnect some peers
[...]

PR #7859, opened 2023-10-27
fix(net): Fix potential network hangs, and reduce code complexity

### LINKED ISSUE #7858: security: avoid hangs in the peer set and related code
### Motivation

There are still some potential hangs in the peer set and related code.

### Specifications

Rust futures need to give a copy of their task waker to any code that could potentially wake the task.

Method calls shouldn't be skipped based on the outcomes of previous methods, unless the connection is shutting down.

### Testing

This is fairly straightforward, we need to check that each method that could wake the service or future is called.

We have existing tests that cover this code.

### Related Work

Follow up to #7772.

teor2345:
## Motivation

Zebra's peer handling code is complex, and it contains some potential hangs. This PR fixes some hangs, and simplifies the code.

Follow up to #7772
Close #7858

### Specifications

Rust futures wake a task when any waker cloned from the most recently passed Context is woken:
> When a future is not ready yet, poll returns Poll::Pending and stores a clone of the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) copied from the current [Context](https://doc.rust-lang.org/std/task/struct.Context.html). This [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) is then woken once the future can make progress.
> ...
> Note that on multiple calls to poll, only the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) from the [Context](https://doc.rust-lang.org/std/task/struct.Context.html) passed to the most recent call should be scheduled to receive a wakeup.

https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll

This PR contains a lot of refactors to make future changes easier.
It also eliminates some wakers entirely to simplify the code and waking logic.

## Solution

Bug fixes:
- Pass context correctly to the peer set background task oneshot. (We're not waking on it now, but we might want to in future.)
- Pass context correctly to the peer client task, and the peer client request sender, so they always wake the client task. (The client task is the task that is spawned by the peer set buffer.)
- Check all ready peers for errors before sending requests to them.

Code simplification:
- Remove preselected peers, just select the peer when we have a new request. This makes waking much less complex.
- Use `futures::ready!` (the macro) and `?` to simplify polling

Refactors:
- Refactor polling methods to return `Poll`, and `Ok` if they do something
- Make panic checking method names clearer (this can be split into its own PR)

### Testing

The existing tests cover this behaviour fairly well. It is hard to check for hangs in tests.

Some tests have been updated for the new polling behaviour.

## Review

This might still need some test fixes.

### Reviewer Checklist

## Follow Up Work

- #7858 
- #7822

### DIFF
--- a/zebra-chain/src/diagnostic/task/thread.rs
+++ b/zebra-chain/src/diagnostic/task/thread.rs
@@ -18,9 +18,11 @@ where
 {
     type Output = T;
 
-    /// Panics if the thread panicked.
-    /// If `panic_on_unexpected_termination` is true, and Zebra is not shutting down, also panics
-    /// if the thread exits.
+    /// # Panics
+    ///
+    /// - if the thread panicked.
+    /// - if the thread is cancelled, `panic_on_unexpected_termination` is true, and
+    ///   Zebra is not shutting down.
     ///
     /// Threads can't be cancelled except by using a panic, so there are no thread errors here.
     /// `panic_on_unexpected_termination` is
--- a/zebra-chain/src/diagnostic/task/thread.rs
+++ b/zebra-chain/src/diagnostic/task/thread.rs
@@ -8,19 +8,42 @@ use std::{
     thread::{self, JoinHandle},
 };
 
+use crate::shutdown::is_shutting_down;
+
 use super::{CheckForPanics, WaitForPanics};
 
-impl<T> CheckForPanics for thread::Result<T> {
+impl<T> CheckForPanics for thread::Result<T>
+where
+    T: std::fmt::Debug,
+{
     type Output = T;
 
     /// Panics if the thread panicked.
+    /// If `panic_on_unexpected_termination` is true, and Zebra is not shutting down, also panics
+    /// if the thread exits.
     ///
     /// Threads can't be cancelled except by using a panic, so there are no thread errors here.
+    /// `panic_on_unexpected_termination` is
     #[track_caller]
-    fn check_for_panics(self) -> Self::Output {
+    fn check_for_panics_with(self, panic_on_unexpected_termination: bool) -> Self::Output {
         match self {
             // The value returned by the thread when it finished.
-            Ok(thread_output) => thread_output,
+            Ok(thread_output) => {
+                if !panic_on_unexpected_termination {
+                    debug!(?thread_output, "ignoring expected thread exit");
+
+                    thread_output
+                } else if is_shutting_down() {
+                    debug!(
+                        ?thread_output,
+                        "ignoring thread exit because Zebra is shutting down"
+                    );
+
+                    thread_output
+                } else {
+                    panic!("thread unexpectedly exited with: {:?}", thread_output)
+                }
+            }
 
             // A thread error is always a panic.
             Err(panic_payload) => panic::resume_unwind(panic_payload),
@@ -28,17 +51,24 @@ impl<T> CheckForPanics for thread::Result<T> {
     }
 }
 
-impl<T> WaitForPanics for JoinHandle<T> {
+impl<T> WaitForPanics for JoinHandle<T>
+where
+    T: std::fmt::Debug,
+{
     type Output = T;
 
     /// Waits for the thread to finish, then panics if the thread panicked.
     #[track_caller]
-    fn wait_for_panics(self) -> Self::Output {
-        self.join().check_for_panics()
+    fn wait_for_panics_with(self, panic_on_unexpected_termination: bool) -> Self::Output {
+        self.join()
+            .check_for_panics_with(panic_on_unexpected_termination)
     }
 }
 
-impl<T> WaitForPanics for Arc<JoinHandle<T>> {
+impl<T> WaitForPanics for Arc<JoinHandle<T>>
+where
+    T: std::fmt::Debug,
+{
     type Output = Option<T>;
 
     /// If this is the final `Arc`, waits for the thread to finish, then panics if the thread
@@ -46,7 +76,7 @@ impl<T> WaitForPanics for Arc<JoinHandle<T>> {
     ///
     /// If this is not the final `Arc`, drops the handle and immediately returns `None`.
     #[track_caller]
-    fn wait_for_panics(self) -> Self::Output {
+    fn wait_for_panics_with(self, panic_on_unexpected_termination: bool) -> Self::Output {
         // If we are the last Arc with a reference to this handle,
         // we can wait for it and propagate any panics.
         //
@@ -56,14 +86,17 @@ impl<T> WaitForPanics for Arc<JoinHandle<T>> {
         // This is more readable as an expanded statement.
         #[allow(clippy::manual_map)]
         if let Some(handle) = Arc::into_inner(self) {
-            Some(handle.wait_for_panics())
+            Some(handle.wait_for_panics_with(panic_on_unexpected_termination))
         } else {
             None
         }
     }
 }
 
-impl<T> CheckForPanics for &mut Option<Arc<JoinHandle<T>>> {
+impl<T> CheckForPanics for &mut Option<Arc<JoinHandle<T>>>
+where
+    T: std::fmt::Debug,
+{
     type Output = Option<T>;
 
     /// If this is the final `Arc`, checks if the thread has finished, then panics if the thread
@@ -71,14 +104,14 @@ impl<T> CheckForPanics for &mut Option<Arc<JoinHandle<T>>> {
     ///
     /// If the thread has not finished, or this is not the final `Arc`, returns `None`.
     #[track_caller]
-    fn check_for_panics(self) -> Self::Output {
+    fn check_for_panics_with(self, panic_on_unexpected_termination: bool) -> Self::Output {
         let handle = self.take()?;
 
         if handle.is_finished() {
             // This is the same as calling `self.wait_for_panics()`, but we can't do that,
             // because we've taken `self`.
             #[allow(clippy::manual_map)]
-            return handle.wait_for_panics();
+            return handle.wait_for_panics_with(panic_on_unexpected_termination);
         }
 
         *self = Some(handle);
@@ -87,7 +120,10 @@ impl<T> CheckForPanics for &mut Option<Arc<JoinHandle<T>>> {
     }
 }
 
-impl<T> WaitForPanics for &mut Option<Arc<JoinHandle<T>>> {
+impl<T> WaitForPanics for &mut Option<Arc<JoinHandle<T>>>
+where
+    T: std::fmt::Debug,
+{
     type Output = Option<T>;
 
     /// If this is the final `Arc`, waits for the thread to finish, then panics if the thread
@@ -95,10 +131,13 @@ impl<T> WaitForPanics for &mut Option<Arc<JoinHandle<T>>> {
     ///
     /// If this is not the final `Arc`, drops the handle and returns `None`.
     #[track_caller]
-    fn wait_for_panics(self) -> Self::Output {
+    fn wait_for_panics_with(self, panic_on_unexpected_termination: bool) -> Self::Output {
         // This is more readable as an expanded statement.
         #[allow(clippy::manual_map)]
-        if let Some(output) = self.take()?.wait_for_panics() {
+        if let Some(output) = self
+            .take()?
+            .wait_for_panics_with(panic_on_unexpected_termination)
+        {
             Some(output)
         } else {
             // Some other task has a reference, so we should give up ours to let them use it.
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -441,6 +441,8 @@ impl Client {
 
         let result = match self.heartbeat_task.poll_unpin(cx) {
             Poll::Pending => {
+                // The heartbeat task returns `Pending` while it continues to run.
+                // But if it has dropped its receiver, it is shutting down, and we should also shut down.
                 if is_canceled {
                     self.set_task_exited_error(
                         "heartbeat",
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -468,7 +468,7 @@ impl Client {
                 )
             }
             Poll::Ready(Err(error)) => {
-                // Heartbeat task stopped with panic.
+                // Heartbeat task panicked.
                 let error = error.panic_if_task_has_panicked();
 
                 // Heartbeat task was cancelled.
@@ -489,7 +489,14 @@ impl Client {
         }
     }
 
-    /// Check if the connection's task has exited.
+    /// Check if the connection's request/response task has exited.
+    ///
+    /// Returns an error if the connection task exited. Otherwise, schedules the client task for
+    /// wakeup when the connection task finishes, and returns `Pending`.
+    ///
+    /// # Panics
+    ///
+    /// If the connection task panicked.
     fn check_connection(&mut self, context: &mut Context<'_>) -> Result<(), SharedPeerError> {
         match self.connection_task.poll_unpin(context) {
             Poll::Pending => {
@@ -501,21 +508,20 @@ impl Client {
                 self.set_task_exited_error("connection", PeerError::ConnectionTaskExited)
             }
             Poll::Ready(Err(error)) => {
-                // Heartbeat task was cancelled.
+                // Connection task panicked.
+                let error = error.panic_if_task_has_panicked();
+
+                // Connection task was cancelled.
                 if error.is_cancelled() {
                     self.set_task_exited_error(
-                        "heartbeat",
+                        "connection",
                         PeerError::HeartbeatTaskExited("Task was cancelled".to_string()),
                     )
                 }
-                // Heartbeat task stopped with panic.
-                else if error.is_panic() {
-                    panic!("heartbeat task has panicked: {error}");
-                }
-                // Heartbeat task stopped with error.
+                // Connection task stopped with another kind of task error.
                 else {
                     self.set_task_exited_error(
-                        "heartbeat",
+                        "connection",
                         PeerError::HeartbeatTaskExited(error.to_string()),
                     )
                 }
@@ -546,6 +552,9 @@ impl Client {
     }
 
     /// Poll for space in the shared request sender channel.
+    ///
+    /// Returns an error if the sender channel is closed. If there is no space in the channel,
+    /// returns `Pending`, and schedules the task for wakeup when there is space available.
     fn poll_request(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), SharedPeerError>> {
         let server_result = ready!(self.server_tx.poll_ready(cx));
         if server_result.is_err() {
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -18,6 +18,8 @@ use futures::{
 use tokio::{sync::broadcast, task::JoinHandle};
 use tower::Service;
 
+use zebra_chain::diagnostic::task::CheckForPanics;
+
 use crate::{
     peer::{
         error::{AlreadyErrored, ErrorSlot, PeerError, SharedPeerError},
@@ -421,6 +423,13 @@ impl MissingInventoryCollector {
 
 impl Client {
     /// Check if this connection's heartbeat task has exited.
+    ///
+    /// Returns an error if the heartbeat task exited. Otherwise, schedules the client task for
+    /// wakeup when the heartbeat task finishes, or the channel closes, and returns `Pending`.
+    ///
+    /// # Panics
+    ///
+    /// If the heartbeat task panicked.
     #[allow(clippy::unwrap_in_result)]
     fn check_heartbeat(&mut self, cx: &mut Context<'_>) -> Result<(), SharedPeerError> {
         let is_canceled = self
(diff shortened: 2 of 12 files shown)


zebra-network/src/peer_set/set.rs
```diff
@@ -441,14 +424,24 @@ where
...
-                Poll::Pending | Poll::Ready(None) => return,
+    ///
+    /// Never returns an error. If there are no unready peers, returns `Ok` immediately.
+    /// Otherwise, returns `Poll::Pending`, and registers a wakeup for the next task that becomes
+    /// ready.
+    fn poll_unready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), BoxError>> {
+        if self.unready_services.is_empty() {
+            // Don't replace the previous waker, because we have no peers to wake the task.
+            return Poll::Ready(Ok(()));
+        }
```
teor2345: This is one half of the fix: don't replace the existing waker with a waker that will never wake.

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +997,74 @@ where
...
+        // This waker skips itself if it is empty and will never wake.
+        // Otherwise, each connected peer should become ready or timeout within a few minutes.
+        //
+        // TODO: drop peers that overload us with inbound messages and never become ready (#7822)
+        let _ = self.poll_unready(cx)?;
+
+        // Correctness: the waker in the context is waiting on peers. If it is replaced later in
+        // the function, the network could hang. We drop the reference here so it can't be used.
+        #[allow(dropping_references)]
+        std::mem::drop(cx);
```
teor2345: This is the other half of the fix: wake on the first ready existing peer, unless there aren't any, then wake on the first connected peer.

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +997,74 @@ where
...
-                    Poll::Pending => {
-                        trace!("preselected service is no longer ready, moving to unready list");
-                        self.push_unready(key, service);
-                    }
-                    Poll::Ready(Err(error)) => {
-                        trace!(%error, "preselected service failed, dropping it");
-                        std::mem::drop(service);
-                    }
-                }
-            }
```
teor2345, line 1034: Because this code sets a waker, it was easiest to just delete it.

zebra-network/src/peer_set/inventory_registry.rs
```diff
@@ -283,11 +283,15 @@ impl InventoryRegistry {
...
+    /// Rotates the inventory HashMaps on every timer tick.
+    /// Drains the inv_stream channel and registers all advertised inventory.
     ///
-    /// - rotates HashMaps based on interval events
-    /// - drains the inv_stream channel and registers all advertised inventory
-    pub fn poll_inventory(&mut self, cx: &mut Context<'_>) -> Result<(), BoxError> {
+    /// Returns an error if the inventory channel is closed. Otherwise returns `Poll::Pending`, and
+    /// registers a wakeup the next time there is new inventory.
+    ///
+    /// TODO: also register a wakeup for the next timer tick.
```
teor2345: I think this is incorrect: a `Waker` gives out the ability to wake the calling task to one or more other tasks or executor functions:
> When a future is not ready yet, poll returns Poll::Pending and stores a clone of the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) copied from the current [Context](https://doc.rust-lang.org/std/task/struct.Context.html). This [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) is then woken once the future can make progress.

https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll

More background:
https://doc.rust-lang.org/std/task/struct.Context.html
https://doc.rust-lang.org/std/task/struct.Waker.html
https://doc.rust-lang.org/std/task/struct.Waker.html#method.wake

zebra-network/src/peer_set/set.rs
```diff
@@ -265,7 +254,11 @@ where
     C: ChainTip,
 {
     fn drop(&mut self) {
-        self.shut_down_tasks_and_channels()
+        // We don't want to wake the current task, we just want to drop everything we can.
```
teor2345:
```suggestion
        // We don't have access to the current task, and we're shutting down, so we just want to drop everything we can.
```

zebra-network/src/peer_set/set.rs
```diff
@@ -479,14 +474,19 @@ where
...
                     debug!(%error, "service failed while unready, dropping service");
 
                     let cancel = self.cancel_handles.remove(&key);
                     assert!(cancel.is_some(), "missing cancel handle");
                 }
             }
         }
+
+        // There are no unready peers, so the unready_peers waker will never wake pending tasks.
+        *cx = Context::from_waker(original_waker);
```
teor2345: This line does nothing, because the task is the same, so the context and waker are the same:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-network/src/peer_set/set.rs
```diff
@@ -441,14 +424,26 @@ where
...
-            match Pin::new(&mut self.unready_services).poll_next(cx) {
-                // No unready service changes, or empty unready services
-                Poll::Pending | Poll::Ready(None) => return,
+    ///
+    /// Never returns an error. If there are no unready peers, returns `Ok`, and preserves the
+    /// original waker. Otherwise, returns `Poll::Pending`, and registers a wakeup for the next
+    /// peer that becomes ready.
+    fn poll_unready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), BoxError>> {
+        if self.unready_services.is_empty() {
+            // Don't replace the previous waker, because we have no peers to wake the task.
```
teor2345: This is unnecessary, because the waker is a copy, so it can be woken by the spawned task for unready peers, or the inventory timer:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-network/src/peer_set/set.rs
```diff
@@ -441,14 +424,26 @@ where
     ///
     /// Move newly ready services to the ready list if they are for peers with supported protocol
     /// versions, otherwise they are dropped. Also drop failed services.
-    fn poll_unready(&mut self, cx: &mut Context<'_>) {
-        loop {
-            match Pin::new(&mut self.unready_services).poll_next(cx) {
-                // No unready service changes, or empty unready services
-                Poll::Pending | Poll::Ready(None) => return,
+    ///
+    /// Never returns an error. If there are no unready peers, returns `Ok`, and preserves the
```
teor2345: Update this comment

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1004,74 @@ where
...
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.poll_background_errors(cx)?;
-
-        // Update peer statuses
+        // Update service and peer statuses.
+        //
+        // # Correctness
+        //
+        // If we poll these futures in series, then only the last polled future will wake
```
teor2345: This is wrong:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1004,74 @@ where
...
+        // or no peers in the peer set at all.
+        //
+        // As a temporary fix, we try to pick the most likely waker, and skip wakers that will
+        // never wake.
+
+        // We can tolerate not being woken for these tasks, because we only return `Pending`
+        // when we have no ready peers. Delays in background error checking are acceptable,
+        // and we can't do anything useful with inventory until we have peers. (The oldest
+        // inventory will get automatically dropped.)
+        let _ = self.poll_background_errors(cx)?;
```
teor2345: We should match on `Poll` here, to make sure that's what we're ignoring even if the function types change.

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1004,74 @@ where
...
+        let _ = self.inventory_registry.poll_inventory(cx)?;
+
+        // We must be woken if there are new peers, but new peers can be infrequent if our
+        // connection slots are full, or we're connected to all available/useful peers.
+        //
+        // TODO: wake waiting tasks if there are new peers or unready peers become ready (#7858)
```
teor2345: This is already happening, same with the TODO below:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1004,74 @@ where
...
+        // This method preserves the original waker if it has no unready peers and will never wake.
+        // Otherwise, each connected peer should become ready or timeout within a few minutes.
+        //
+        // TODO: drop peers that overload us with inbound messages and never become ready (#7822)
+        let _ = self.poll_unready(cx)?;
+
+        // Correctness: the waker in the context is waiting on peers. If it is replaced later in
+        // the function, the network could hang. We drop the reference here so it can't be used.
+        #[allow(dropping_references)]
+        std::mem::drop(cx);
```
teor2345: Not needed:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-network/src/peer_set/set.rs
teor2345: This wasn't necessary for the waker, but reducing the complexity of this code helps avoid other logic bugs around waking, pending, and similar issues:
https://github.com/ZcashFoundation/zebra/pull/7859/files#r1380722199

zebra-chain/src/diagnostic/task/thread.rs
```diff
@@ -8,45 +8,75 @@ use std::{
...
-impl<T> CheckForPanics for thread::Result<T> {
+impl<T> CheckForPanics for thread::Result<T>
+where
+    T: std::fmt::Debug,
+{
     type Output = T;
 
     /// Panics if the thread panicked.
+    /// If `panic_on_unexpected_termination` is true, and Zebra is not shutting down, also panics
+    /// if the thread exits.
```
arya2: Optional/Nitpick:

```suggestion
    /// # Panics
    ///
    /// - if the thread panicked.
    /// - if the thread is cancelled, `panic_on_unexpected_termination` is true, and
    ///   Zebra is not shutting down.
```

zebra-network/src/peer_set/set.rs
arya2: I found this a lot easier to understand than the old code.

zebra-network/src/peer/client.rs
```diff
@@ -421,26 +423,33 @@ impl MissingInventoryCollector {
...
-                PeerError::HeartbeatTaskExited("Task was cancelled".to_string()),
-            );
-        }
-
-        match self.heartbeat_task.poll_unpin(cx) {
+        let result = match self.heartbeat_task.poll_unpin(cx) {
             Poll::Pending => {
-                // Heartbeat task is still running.
-                Ok(())
+                if is_canceled {
```
arya2, line 446: Optional:

```suggestion
                // The heartbeat task will return `Pending` when cancelled but unable to immediately
                // acquire a permit to send a message to the `address_book_updater`.
                if is_canceled {
```

zebra-network/src/peer/client.rs
```diff
@@ -536,6 +570,33 @@ impl Client {
...
+        // if either task exits, or if the heartbeat task drops the cancel handle.
+        //
+        //`ready!` returns `Poll::Pending` when `server_tx` is unready, and
+        // schedules this task for wakeup.
+        //
+        // It's ok to exit early and skip wakeups when there is an error, because the connection
+        // and its tasks are shut down immediately on error.
+
+        let _heartbeat_pending: Poll<()> = self.poll_heartbeat(cx)?;
+        let _connection_pending: Poll<()> = self.poll_connection(cx)?;
```
arya2, line 596: Good idea explicitly typing these.

zebra-network/src/peer/client.rs
```diff
@@ -459,44 +468,66 @@ impl Client {
...
+                    self.set_task_exited_error(
+                        "connection",
+                        PeerError::HeartbeatTaskExited(error.to_string()),
+                    )
+                }
             }
-        }
+        };
+
+        Poll::Ready(result)
```
arya2, line 532: Optional/Nitpick/Subjective: This wasn't changed in this PR, but return an error from `set_task_exited_error()` and put it in a `Result::Err` here?

```rust
        Poll::Ready(Err(result))
```

zebra-network/src/peer_set/set.rs
```diff
@@ -393,31 +382,25 @@ where
...
-                    None
+                    Poll::Ready(Ok(()))
                 }
 
-                // The tasks have been sent and consumed, but then they exited.
-                //
-                // Correctness: the peer set must receive at least one task.
-                //
-                // TODO: refactor `handle_rx` and `guards` into an enum
-                //       for the background task state: Waiting/Running/Shutdown.
```
arya2, line 404: Do we still want this TODO?

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1009,70 @@ where
...
-                    Poll::Ready(Err(error)) => {
-                        trace!(%error, "preselected service failed, dropping it");
-                        std::mem::drop(service);
-                    }
-                }
-            }
+        if self.ready_services.is_empty() {
+            // # Correctness
+            //
+            // If the channel is full, drop the demand signal rather than waiting.
```
arya2:
```suggestion
            // If the channel is full, send the demand signal rather than waiting.
```

zebra-network/src/peer_set/set.rs
```diff
@@ -441,14 +424,30 @@ where
...
+    /// doesn't register any wakeups. (Since wakeups come from peers, there needs to be at least one
+    /// peer to register a wakeup.)
+    fn poll_unready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), BoxError>> {
+        // # Correctness
+        //
+        // `poll_next()` must always be called, because `self.unready_services` could have been
+        // empty before the call to `self.poll_ready()`.
+        //
+        // > When new futures are added, `poll_next` must be called in order to begin receiving
+        // > wake-ups for new futures.
```
arya2, line 447: 🎉

arya2 (dismissed):

zebra-network/src/peer_set/set.rs
teor2345: Yeah I'm really not sure why we needed to preselect a peer. It seems to work fine just selecting it at random at request time.

zebra-network/src/peer/client.rs
teor2345: That's not quite what is happening here: there are no messages being sent in this code (until shutdown).

```suggestion
                // The heartbeat task returns `Pending` while it continues to run.
                // But if it has dropped its receiver, it is shutting down, and we should also shut down.
                if is_canceled {
```

zebra-network/src/peer/client.rs
teor2345: I think that would make it more difficult to say `set_task_exited_error(...)?`.

We don't do that here, but we probably could.

zebra-network/src/peer_set/set.rs
teor2345: I'm not sure if it makes sense any more. We might want to create `TaskGuard` and `TaskGuardSet` types, or make what we've done in the state with the update task more generic, and move it to `zebra_chain::task`.

I'm not sure, let's see what's actually needed later.

zebra-network/src/peer_set/set.rs
teor2345: That's not what's happening here:
> Attempts to send a message on this Sender, returning the message if there was an error.

https://docs.rs/futures/latest/futures/channel/mpsc/struct.Sender.html#method.try_send

teor2345:
I still need to fix some tests that are hanging, but they all seem to be inventory tests, so that should be an easy fix.

zebra-network/src/peer_set/set.rs
arya2: Ah, that makes more sense.

teor2345:
This should be ready to go now, there was a minor issue with the `Update` future for the inventory registry, but it seemed like a good idea to make all the poll methods consistent.

teor2345:
Does this need a full sync before it merges, or can we wait until Friday's scheduled full sync?

If we do one now, the previous time was 2d 3h 15m:
https://github.com/ZcashFoundation/zebra/actions/runs/6824538443/job/18561008809

zebra-network/src/peer_set/set.rs
```diff
@@ -479,14 +502,65 @@ where
...
+    /// because the peer set has exclusive access to send requests to each peer. (If an inbound
+    /// request is in progress, it will be handled, then our request will be sent by the connection
+    /// task.)
+    ///
+    /// Returns `Poll::Ready` if there are some ready peers, and `Poll::Pending` if there are no
+    /// ready peers. Never registers any task wakeups.
+    ///
+    /// # Panics
+    ///
+    /// If any peers somehow became unready. This indicates a bug in the peer set, where requests
```
arya2: Optional/Nitpick:

```suggestion
    /// If any peers somehow became unready without being sent a request. This indicates a bug in the peer set, where requests
```

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1108,71 @@ where
...
-                    }
-                    Poll::Ready(Err(error)) => {
-                        trace!(%error, "preselected service failed, dropping it");
-                        std::mem::drop(service);
-                    }
-                }
-            }
+        // Check for failures in ready peers, removing newly errored or disconnected peers.
+        // So it needs to run after `poll_unready()`.
+        let ready_peers: Poll<()> = self.poll_ready_peer_errors(cx);
```
arya2, line 1150: Optional:

```suggestion
        //
        // TODO: Store an internal context with a noop_waker in a field for methods that never register wakers?
        let ready_peers: Poll<()> = self.poll_ready_peer_errors(cx);
```

zebra-network/src/peer_set/set.rs
```diff
@@ -995,78 +1108,71 @@ where
...
+
+        // Check for newly ready peers, including newly added peers (which are added as unready).
+        // So it needs to run after `poll_discover()`. Registers a wakeup if there are any unready
+        // peers.
+        //
+        // Each connected peer should become ready within a few minutes, or timeout, close the
+        // connection, and release its connection slot.
+        //
+        // TODO: drop peers that overload us with inbound messages and never become ready (#7822)
+        let _poll_pending_or_ready: Poll<Option<()>> = self.poll_unready(cx)?;
```
arya2, line 1137: Could we add a correctness comment here like: "Must never register any task wakeups on this context after this call to `poll_unready()`"

arya2 (dismissed):
> Does this need a full sync before it merges, or can we wait until Friday's scheduled full sync?

I think we can wait until the next scheduled full sync.

zebra-network/src/peer_set/set.rs
teor2345: That's not how task wakeups work. (It's how I thought they worked when I initially opened this PR, but I was wrong.)

*Every* potential waking event gets a reference to the waker via the context, and *any* of those events can wake the task.

zebra-network/src/peer_set/set.rs
```diff
@@ -479,14 +502,65 @@ where
...
+
+    /// Checks previously ready peer services for errors.
+    ///
+    /// The only way these peer `Client`s can become unready is when we send them a request,
+    /// because the peer set has exclusive access to send requests to each peer. (If an inbound
+    /// request is in progress, it will be handled, then our request will be sent by the connection
+    /// task.)
+    ///
+    /// Returns `Poll::Ready` if there are some ready peers, and `Poll::Pending` if there are no
+    /// ready peers. Never registers any task wakeups.
```
teor2345:
```suggestion
    /// ready peers. Registers a wakeup if any peer has failed due to a disconnection, hang, or protocol error.
```

zebra-network/src/peer_set/set.rs
teor2345: This method does register wakeups for ready peer failures, I have edited the method comment.

And the suggested change is not needed, because the waker is shared among all the events that could wake the task.

arya2 (approved):

zebra-network/src/peer_set/set.rs
arya2: > any of those events can wake the task

Like a select?

teor2345: Sort of, but much lower level. The context contains a waker, which gives the ability to wake the task.

Technically there is nothing stopping a future from storing a waker when it returns `Poll::Ready`, and in some cases in Zebra we do this in inner poll methods, in case the outer poll function returns `Pending` due to outer state or other inner poll methods.

> The typical life of a Waker is that it is constructed by an executor, wrapped in a [Context](https://docs.rs/futures/latest/futures/task/struct.Context.html), then passed to [Future::poll()](https://docs.rs/futures/latest/futures/future/trait.Future.html#tymethod.poll). Then, if the future chooses to return [Poll::Pending](https://docs.rs/futures/latest/futures/task/enum.Poll.html#variant.Pending), it must also store the waker somehow and call [Waker::wake()](https://docs.rs/futures/latest/futures/task/struct.Waker.html#method.wake) when the future should be polled again.

https://docs.rs/futures/latest/futures/task/struct.Waker.html

> When a future is not ready yet, poll returns Poll::Pending and stores a clone of the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) copied from the current [Context](https://doc.rust-lang.org/std/task/struct.Context.html). This [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) is then woken once the future can make progress. For example, a future waiting for a socket to become readable would call .clone() on the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) and store it. When a signal arrives elsewhere indicating that the socket is readable, [Waker::wake](https://doc.rust-lang.org/std/task/struct.Waker.html#method.wake) is called and the socket future’s task is awoken. Once a task has been woken up, it should attempt to poll the future again, which may or may not produce a final value.

https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll

Issue #7816, opened 2023-10-25
devops: Build CI Docker runs twice for every PR


teor2345:
### Describe the issue or request

On every PR, we run the same Docker build twice, with exactly the same settings:

https://github.com/ZcashFoundation/zebra/blob/aad883b68ae1fb7a98aa2868bc1bd4a34954bd13/.github/workflows/ci-unit-tests-docker.yml#L84-L86

https://github.com/ZcashFoundation/zebra/blob/aad883b68ae1fb7a98aa2868bc1bd4a34954bd13/.github/workflows/ci-integration-tests-gcp.yml#L126-L128

### Expected Behavior

This seems redundant, we only need to build one image.

### Current Behavior

The same image is built twice.

### Possible Solution

Make the jobs in one workflow wait on the build job in the other workflow, or re-combine the workflows into a single workflow.

### Additional Information/Context

This costs us extra build machine time, and extra storage costs.

### Is this happening on PRs?

Yes

### Is this happening on the main branch?

Not really, de-duplicating the `main` merge and release builds is more complicated, and there is only one duplicate every few weeks

teor2345:
Cross-workflow dependencies can be created in GitHub actions:
https://stackoverflow.com/questions/63148639/create-dependencies-between-jobs-in-github-actions#63148947

teor2345:
Seems like these duplicate jobs might be causing CI failures:
> #1 ERROR: Error response from daemon: error parsing HTTP 429 response body: invalid character 'T' looking for beginning of value: "Too Many Requests (HAP429).\n"

https://github.com/ZcashFoundation/zebra/actions/runs/6646940836/job/18061386992?pr=7831#step:10:209

teor2345:
Seems like this can also allow PRs to merge even if one of the duplicate jobs fails. That's definitely a bug.

This happened on PR #7831 but it still seems to be merging?
<img width="913" alt="Screenshot 2023-10-26 at 09 21 37" src="https://github.com/ZcashFoundation/zebra/assets/8951843/b5055b03-c36f-4f83-8cbc-d1b421951748">

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7816?planning-poker) with Zenhub @arya2 @gustavovalverde @oxarbitrage @teor2345 @upbqdn

PR #7869, opened 2023-10-30
fix(net): Add outer timeouts for critical network operations to avoid hangs

### LINKED ISSUE #7772: bug: zebrad will not reconnect after an internet connection failure and restore

teor2345:
## Motivation

This is a quick fix for Zebra hanging rather than disconnecting when the peer set is empty. It prevents the requesting service side of the hang.

Close #7772

### Specifications

Sometimes a hang can happen if the called service doesn't correctly set up its waker (a bug), the request is really expensive, or the service is full of requests.

It seems unnecessary to have timeouts within service implementations, and in their callers. But it's very easy to add code that doesn't have a timeout, or has a subtle hang or slowness bug under some conditions. So we should move timeouts as far out of the service stack as possible, to include more code.

For now this PR just adds timeouts to the outermost layer of critical network services, or explains how they are already implemented.

## Solution

Timeouts:
- Add a timeout to syncer obtain, extend, and genesis block downloads and verifies
- Add a timeout to mempool crawls
- Add a timeout to mempool download and verify

Hang fixes:
- Remove a potential hang from address book requests
- Remove a potential hang when the lookahead limit is reached, then the pending blocks drop below the limit
- Remove a potential hang when the lookahead limit is reached, then the syncer is completely reset

Documentation:
- Explain why the inbound service never hangs

### Testing

- [x] TODO: Manually disconnect Zebra from the network for more than 10 minutes, then restore the network, and make sure that it reconnects within 10 minutes

## Review

This PR would be good to get in the release. It is low risk, because it just adds timeouts to existing code. (Or ignores inbound peers requests under load.)

### Reviewer Checklist

## Follow Up Work

PR #7859 is a more complicated fix to an underlying peer set bug. But it should go in the next release to get more testing.

### DIFF
--- a/zebrad/Cargo.toml
+++ b/zebrad/Cargo.toml
@@ -168,6 +168,7 @@ toml = "0.8.3"
 futures = "0.3.29"
 rayon = "1.7.0"
 tokio = { version = "1.33.0", features = ["time", "rt-multi-thread", "macros", "tracing", "signal"] }
+tokio-stream = { version = "0.1.14", features = ["time"] }
 tower = { version = "0.4.13", features = ["hedge", "limit"] }
 pin-project = "1.1.3"
 
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -9,7 +9,7 @@ use std::{
     collections::HashSet,
     future::Future,
     pin::Pin,
-    sync::Arc,
+    sync::{Arc, TryLockError},
     task::{Context, Poll},
     time::Duration,
 };
@@ -278,7 +278,11 @@ impl Service<zn::Request> for Inbound {
                     }
                 }
                 Err(TryRecvError::Empty) => {
-                    // There's no setup data yet, so keep waiting for it
+                    // There's no setup data yet, so keep waiting for it.
+                    //
+                    // We could use Future::poll() to get a waker and return Poll::Pending here.
+                    // But we want to drop excess requests during startup instead. Otherwise,
+                    // the inbound service gets overloaded, and starts disconnecting peers.
                     result = Ok(());
                     Setup::Pending {
                         full_verify_concurrency_limit,
@@ -307,6 +311,11 @@ impl Service<zn::Request> for Inbound {
                 mempool,
                 state,
             } => {
+                // # Correctness
+                //
+                // Clear the stream but ignore the final Pending return value.
+                // If we returned Pending here, and there were no waiting block downloads,
+                // then inbound requests would wait for the next block download, and hang forever.
                 while let Poll::Ready(Some(_)) = block_downloads.as_mut().poll_next(cx) {}
 
                 result = Ok(());
@@ -366,20 +375,35 @@ impl Service<zn::Request> for Inbound {
                 //
                 // # Correctness
                 //
-                // Briefly hold the address book threaded mutex while
-                // cloning the address book. Then sanitize in the future,
-                // after releasing the lock.
-                let peers = address_book.lock().unwrap().clone();
+                // If the address book is busy, try again inside the future. If it can't be locked
+                // twice, ignore the request.
+                let address_book = address_book.clone();
+
+                let get_peers = move || match address_book.try_lock() {
+                    Ok(address_book) => Some(address_book.clone()),
+                    Err(TryLockError::WouldBlock) => None,
+                    Err(TryLockError::Poisoned(_)) => panic!("previous thread panicked while holding the address book lock"),
+                };
+
+                let peers = get_peers();
 
                 async move {
-                    // Correctness: get the current time after acquiring the address book lock.
+                                        // Correctness: get the current time inside the future.
                     //
-                    // This time is used to filter outdated peers, so it doesn't really matter
+                    // This time is used to filter outdated peers, so it doesn't matter much
                     // if we get it when the future is created, or when it starts running.
                     let now = Utc::now();
 
+                    // If we didn't get the peers when the future was created, wait for other tasks
+                    // to run, then try again when the future first runs.
+                    if peers.is_none() {
+                        tokio::task::yield_now().await;
+                    }
+                    let peers = peers.or_else(get_peers);
+                    let is_busy = peers.is_none();
+
                     // Send a sanitized response
-                    let mut peers = peers.sanitized(now);
+                    let mut peers = peers.map_or_else(Vec::new, |peers| peers.sanitized(now));
 
                     // Truncate the list
                     let address_limit = div_ceil(peers.len(), ADDR_RESPONSE_LIMIT_DENOMINATOR);
@@ -387,8 +411,20 @@ impl Service<zn::Request> for Inbound {
                     peers.truncate(address_limit);
 
                     if peers.is_empty() {
-                        // We don't know if the peer response will be empty until we've sanitized them.
-                        debug!("ignoring `Peers` request from remote peer because our address book is empty");
+                        // Sometimes we don't know if the peer response will be empty until we've
+                        // sanitized them.
+                        if is_busy {
+                            debug!(
+                                "ignoring `Peers` request from remote peer because our address \
+                                 book has no available peers"
+                            );
+                        } else {
+                            info!(
+                                "ignoring `Peers` request from remote peer because our address \
+                                 book is busy"
+                            );
+                        }
+
                         Ok(zn::Response::Nil)
                     } else {
                         Ok(zn::Response::Peers(peers))
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -388,7 +388,7 @@ impl Service<zn::Request> for Inbound {
                 let peers = get_peers();
 
                 async move {
-                                        // Correctness: get the current time inside the future.
+                    // Correctness: get the current time inside the future.
                     //
                     // This time is used to filter outdated peers, so it doesn't matter much
                     // if we get it when the future is created, or when it starts running.
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -414,14 +414,14 @@ impl Service<zn::Request> for Inbound {
                         // Sometimes we don't know if the peer response will be empty until we've
                         // sanitized them.
                         if is_busy {
-                            debug!(
-                                "ignoring `Peers` request from remote peer because our address \
-                                 book has no available peers"
-                            );
-                        } else {
                             info!(
                                 "ignoring `Peers` request from remote peer because our address \
                                  book is busy"
+                            );
+                        } else {
+                            debug!(
+                                "ignoring `Peers` request from remote peer because our address \
+                                 book has no available peers"
                             );
                         }
 
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -2,13 +2,17 @@
 //!
 //! It is used when Zebra is a long way behind the current chain tip.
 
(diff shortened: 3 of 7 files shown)


zebrad/src/components/sync.rs
```diff
@@ -550,76 +557,103 @@ where
...
-                self.handle_block_response(rsp)?;
-            }
-            self.update_metrics();
+            // Avoid hangs due to service readiness or other internal operations
+            extra_hashes = timeout(BLOCK_VERIFY_TIMEOUT, self.try_to_sync_once(extra_hashes))
+                .await
+                .map_err(Into::into)
+                // TODO: replace with flatten() when it stabilises (#70142)
+                .and_then(convert::identity)?;
+        }
```
teor2345, line 578: This is the only change in the diff for this loop, the rest is code movement and removing an indent.

teor2345, line 579: This is the only change in the diff for this loop, the rest is code movement and removing an indent.

zebrad/src/components/sync.rs
```diff
@@ -550,76 +557,103 @@ where
...
         self.update_metrics();
 
         while !self.prospective_tips.is_empty() || !extra_hashes.is_empty() {
-            // Check whether any block tasks are currently ready:
-            while let Poll::Ready(Some(rsp)) = futures::poll!(self.downloads.next()) {
-                self.handle_block_response(rsp)?;
-            }
-            self.update_metrics();
+            // Avoid hangs due to service readiness or other internal operations
+            extra_hashes = timeout(BLOCK_VERIFY_TIMEOUT, self.try_to_sync_once(extra_hashes))
```
arya2, line 574: Optional/Nitpick: Could we leave a TODO for replacing this with a constant + some duration per future in `downloads`?

arya2 (dismissed):
Looks good!

> Manually disconnect Zebra from the network for more than 10 minutes, then restore the network, and make sure that it reconnects within 10 minutes

This worked as expected.

upbqdn:
> Manually disconnect Zebra from the network

I'm just wondering how you did this.

teor2345:
> > Manually disconnect Zebra from the network
> 
> I'm just wondering how you did this.

I would remove the ethernet cable on the machine, or turn off wifi. If I wanted to keep working on other things on the same machine, I'd make the Linux firewall block or drop all packets into and out of a process.

Dropping all packets to or from the Zcash network port number doesn't work, because peers can have other ports.

Here is a GUI that does that:
https://github.com/evilsocket/opensnitch

Or you can use `zebrad &; iptables -m owner --pid-owner $! -j REJECT`:
https://stackoverflow.com/questions/4314163/create-iptables-rule-per-process-service

(This used to require another user account, but recent versions can do it by process ID.)

zebrad/src/components/sync.rs
teor2345: Some of these timeouts already exist, but based on the logs in the ticket, they aren't enough to stop all of the syncer hangs.

For example, until we merge this PR, the following code won't have a timeout:

Called by `download_and_verify()`:
https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync/downloads.rs#L333-L336

Called inside the spawned `downloads` future:
https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync/downloads.rs#L503-L514

Called by the `syncer` module:
https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync.rs#L629-L635

https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync.rs#L659

https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync.rs#L705

https://github.com/ZcashFoundation/zebra/blob/b0555c0e61cd6576234e39170d1accdda342c293/zebrad/src/components/sync.rs#L780

And similar code in `extend_tips()` and `request_genesis()`.

And any code we add in future without explicitly thinking about timeouts every time. So an outer wrapper seems easier to maintain.

teor2345: I started trying to do each one, and by the time I got to changing the third low-level statement, I decided it was much easier to use an outer timeout.

teor2345:
I pushed another significant hang fix in commit [b2385d9](https://github.com/ZcashFoundation/zebra/pull/7869/commits/b2385d932167bbf33bf2840e2b2af593529c26b2). 

Previously we weren't resetting the lookahead pause flag when we were above the lookahead limit, then the syncer was reset. We could also skip resetting the flag if the mutex was locked when the final block was downloaded.

This could be the cause of some of the hangs we see during syncing. It's better for a few blocks to wait for a short time than hang the entire syncer.

teor2345:
> ERROR: failed to solve: rust:bullseye: unexpected status from HEAD request to https://registry-1.docker.io/v2/library/rust/manifests/bullseye: 429 Too Many Requests

https://github.com/ZcashFoundation/zebra/actions/runs/6700359536/job/18206164708?pr=7869#step:10:230

This rate-limit will be hit less if we fix #7816.

zebrad/src/components/sync.rs
```diff
@@ -541,7 +548,8 @@ where
     /// following a fork. Either way, Zebra should attempt to obtain some more tips.
     ///
     /// Returns `Err` if there was an unrecoverable error and restarting the synchronization is
-    /// necessary.
+    /// necessary. This includes outer timeouts, where an entire syncing step takes an extrememly
```
upbqdn:
```suggestion
    /// necessary. This includes outer timeouts, where an entire syncing step takes an extremely
```

zebrad/src/components/inbound.rs
```diff
@@ -366,29 +375,56 @@ impl Service<zn::Request> for Inbound {
...
+                            debug!(
+                                "ignoring `Peers` request from remote peer because our address \
+                                 book has no available peers"
+                            );
+                        } else {
+                            info!(
+                                "ignoring `Peers` request from remote peer because our address \
+                                 book is busy"
+                            );
+                        }
```
upbqdn: Shouldn't we swap these two messages?

```suggestion
                        // Sometimes we don't know if the peer response will be empty until we've
                        // sanitized them.
                        if is_busy {
                            info!(
                                "ignoring `Peers` request from remote peer because our address \
                                 book is busy"
                            );
                        } else {
                            debug!(
                                "ignoring `Peers` request from remote peer because our address \
                                 book has no available peers"
                            );
                        }
```

arya2 (dismissed):

zebrad/src/components/inbound.rs
teor2345: Thank you, I got them the wrong way around because I got the logic wrong. (I was thinking `!is_busy`, and I got rid of the `!` but didn't swap the messages).

arya2 (dismissed):

zebrad/Cargo.toml
```diff
@@ -168,6 +168,7 @@ toml = "0.8.3"
 futures = "0.3.29"
 rayon = "1.7.0"
 tokio = { version = "1.33.0", features = ["time", "rt-multi-thread", "macros", "tracing", "signal"] }
+tokio-stream = { version = "0.1.14", features = ["time"] }
```
upbqdn, line 171: I'd like a bit more time to finish reviewing this PR before it gets merged, so I'm opening this dummy comment. I'll resolve it manually soon.

teor2345: We won't be doing anything with the release until after tomorrow's meeting about PR #7817, and after the [full sync](https://github.com/ZcashFoundation/zebra/actions/runs/6686504232) finishes.

So you have plenty of time.

teor2345:
This failure looks like an infrastructure issue, compiling hung part of the way through:
```
Tue, 31 Oct 2023 21:07:01 GMT
   Compiling zebra-chain v1.0.0-beta.30 (/opt/zebrad/zebra-chain)
Tue, 31 Oct 2023 21:07:11 GMT
   Compiling metrics v0.21.1
Tue, 31 Oct 2023 21:07:12 GMT
   Compiling humantime-serde v1.1.1
Tue, 31 Oct 2023 21:07:12 GMT
   Compiling bincode v1.3.3
Tue, 31 Oct 2023 21:11:58 GMT
Error: The operation was canceled.
```

https://github.com/ZcashFoundation/zebra/actions/runs/6711815526/job/18240439236?pr=7869

teor2345:
Here's a comment from Slack that might be useful here:

The network & state release freeze applies to high-risk PRs between the start of the last full sync and the release.

I'm thinking of allowing [fix(net): Add outer timeouts for critical network operations to avoid hangs](https://github.com/ZcashFoundation/zebra/pull/7869), because it helps fix an important network hang bug.

Here's my risk analysis of each change in the PR:

- adding timeouts to syncer and mempool crawler functions that usually don't time out
	- in the syncer a timeout resets the syncer, this is a standard operation
	- in the mempool it just skips the crawl and does another one after the rate-limit timer, the exact timing of crawls is not important
- removing a potential deadlock or hang from the block downloader, but we know it doesn't usually deadlock (though it might or might not be the source of the regular 10 minute hangs in the logs)
	- this might slightly delay some block verify futures, but that delay is much shorter than a hang
- removing a potential hang from the inbound service by ignoring the request, but we know it usually works
	- ignoring inbound requests is allowed by the protocol, and zcashd does it all the time, so we know Zebra and zcashd can handle it

I think the biggest risk here is a panic or hang, but I can't see how that would happen in these code changes. And if it does and it's frequent, it is likely to get picked up by the CI tests before we release.

zebrad/src/components/sync/downloads.rs
```diff
@@ -471,17 +482,11 @@ where
...
-                        },
-                        Err(TryLockError::Poisoned(_)) => panic!("thread panicked while holding the past_lookahead_limit_sender mutex guard"),
-                        // We'll try allowing new downloads when we get the next block
-                        Err(TryLockError::WouldBlock) => {}
-                    }
+                    // Reset the watched value to false, since we're well under the limit.
+                    // We need to block here, because if we don't the syncer can hang.
+
+                    // But if Zebra is shutting down, ignore the send error.
+                    let _ = past_lookahead_limit_sender.lock().expect("thread panicked while holding the past_lookahead_limit_sender mutex guard").send(false);
```
teor2345, line 489: Oops, I accidentally deleted this metric:

```suggestion
                    let _ = past_lookahead_limit_sender.lock().expect("thread panicked while holding the past_lookahead_limit_sender mutex guard").send(false);
                    metrics::counter!("sync.max.height.limit.reset.count", 1);
```

upbqdn (approved):

Issue #7787, opened 2023-10-19
Investigate why there are more than 2000 network peer addresses


upbqdn:
## Important - User Privacy

Do not post node IP addresses on this ticket, because they can be used to locate Zcash users.

## Motivation

`zebrad` on my local machine had recently 2244 "known peers":

![image](https://github.com/ZcashFoundation/zebra/assets/8863307/449f4d69-d541-4eb9-99d3-ddb02bf4293e)

There can't be that many Zcash nodes. We should confirm if these are peers from other networks, peers on rapidly changing addresses (credit to Teor for pointing this possibility out), or some other sort of peers.

teor2345:
This ticket is tricky to analyse because we don't want to post user IP addresses publicly. They can be used to find Zcash users, which is a privacy issue.

Instead, I would suggest looking for any obvious patterns in failed peer IP addresses:
- adding failed peers to the `getpeerinfo` RPC
- adding a log that lists the IP addresses

Also, consider logging the `VersionMessage` when a peer fails.

For example:
* lots of IP addresses in the same IPv4 /16, or IPv6 /24 or /32
* lots of addresses with the same IP but different ports (unstable ports or many instances behind a NAT)
* lots of addresses with the same port (possibly a Zcash fork, we can ignore these ports in our address book code)

It might also help to:
* add logging to work out why the failed addresses were disconnected, for example, they have a different chain fork or incompatible messages
* see what `zcashd` does with these addresses

Ultimately I don't think there's much to worry about here, but it is helpful to remove useless addresses where we can.

teor2345:
JSON `getpeerinfo` in peer IP address order:
```sh
 zcash-cli -rpcport=28232 getpeerinfo | jq 'map(.addr)|sort'
```

Count ports:
```sh
zcash-cli -rpcport=28232 getpeerinfo | jq 'map(.addr)|sort' | rg ':' | cut -d: -f2 | cut -d '"' -f1 | sort -n | uniq -c
```

teor2345:
Class B IPv4 address blocks:

```sh
zcash-cli -rpcport=28232 getpeerinfo | jq 'map(.addr)|sort' | rg ':' | cut -d: -f1 | cut -d '"' -f2 | cut -d. -f1-2 | sort -n | uniq -c
```

Change `cut -d. -f1-2` to `1` for class A or `1-3` for class C.

teor2345:
[...]

PR #7902, opened 2023-11-03
fix(net): Fix a potential hang caused by accessing the address book directly

teor2345:
## Motivation

The outbound connector accesses the address book directly, potentially blocking other connections or async tasks. Instead, we can send the failure to the channel.

Also there are some unnecessary services arguments in address book updates.

These fixes happened as part of #7787.

  - ~~Have you added or updated tests?~~ Hangs are extremely difficult to test for.

Sending to channels is much faster and cheaper than locking a mutex, which can be delayed for a long time under heavy load.

## Solution

- Send failures to the address book update channel
- Fix a (possibly unreachable) panic with a zero channel size
- Remove unnecessary `services` arguments from some address book updates
- Update tests

### Testing

Existing tests cover this code, but hangs are hard to test for.

## Review

This is a low priority bug fix.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/address_book_updater.rs
+++ b/zebra-network/src/address_book_updater.rs
@@ -1,6 +1,6 @@
 //! The timestamp collector collects liveness information from peers.
 
-use std::{net::SocketAddr, sync::Arc};
+use std::{cmp::max, net::SocketAddr, sync::Arc};
 
 use thiserror::Error;
 use tokio::{
@@ -12,6 +12,9 @@ use tracing::{Level, Span};
 use crate::{
     address_book::AddressMetrics, meta_addr::MetaAddrChange, AddressBook, BoxError, Config,
 };
+
+/// The minimum size of the address book updater channel.
+pub const MIN_CHANNEL_SIZE: usize = 10;
 
 /// The `AddressBookUpdater` hooks into incoming message streams for each peer
 /// and lets the owner of the sender handle update the address book. For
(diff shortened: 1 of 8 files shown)


zebra-network/src/address_book_updater.rs
```diff
@@ -13,6 +13,9 @@ use crate::{
     address_book::AddressMetrics, meta_addr::MetaAddrChange, AddressBook, BoxError, Config,
 };
 
+/// The minimum size of the address book updater channel.
+pub const MIN_CHANNEL_SIZE: usize = 10;
```
arya2, line 17: Optional/Nitpick: Use a power of 2 as the default minimum for channel capacities where appropriate if you think that would be a good convention.

zebra-network/src/peer_set/candidate_set.rs
```diff
@@ -414,6 +414,8 @@ where
     }
 
     /// Returns the address book for this `CandidateSet`.
```
arya2, line 416:
```suggestion
    /// Returns the address book for this `CandidateSet`.
    ///
    /// TODO: Consider converting the candidate set to a tower service that owns the
    ///       address book and handles periodic tasks in a `poll_ready` method before using
    ///       this method in production.
```

arya2 (approved):
This is really nice.

zebra-network/src/address_book_updater.rs
teor2345: I don't really mind either way.

zebra-network/src/peer_set/candidate_set.rs
teor2345: I think this TODO is already in this file at least once, or in a nearby file

PR #7923, opened 2023-11-08
change(state): Make the types for finalized blocks consistent

### LINKED ISSUE #7273: Make the types for finalized blocks consistent
### Motivation

Address the discussion in https://github.com/ZcashFoundation/zebra/pull/7035#discussion_r1239046571.

upbqdn:
## Motivation

Close #7273. I initially wanted to summarize the discussion that the issue points to in the issue description, but I ended up fixing the code itself because I thought it was easier.

The issue was that we were using `SemanticallyVerifiedBlock` and `SemanticallyVerifiedBlockWithTrees` for blocks that the finalized state commits to its database. Doing so is not what we want because, as Teor said:

> Also, I don't think we should be downgrading verification status from contextual or checkpoint to semantic, unless it's a quick downgrade using into(), and then calling a common verification function. Most of the time we've already done that verification on those types though!

### PR Author Checklist

## Solution

- Use a new structure called `FinalizedBlock` instead of `SemanticallyVerifiedBlockWithTrees`, and also use it when writing a finalized block into the database.
- This PR doesn't really change any logic in the state. It is only a structural refactor.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 
 /// An argument wrapper struct for note commitment trees.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct NoteCommitmentTrees {
     /// The sprout note commitment tree.
     pub sprout: Arc<sprout::tree::NoteCommitmentTree>,
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 
 /// An argument wrapper struct for note commitment trees.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct NoteCommitmentTrees {
     /// The sprout note commitment tree.
     pub sprout: Arc<sprout::tree::NoteCommitmentTree>,
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -11,6 +11,8 @@ use crate::{
 };
 
 /// An argument wrapper struct for note commitment trees.
+///
+/// The default instance represents the trees and subtrees that correspond to the genesis block.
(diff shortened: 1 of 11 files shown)


zebra-chain/src/parallel/tree.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
 };
 
 /// An argument wrapper struct for note commitment trees.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
teor2345: Let's document that the default represents the genesis treestate. (Which is how other tree defaults work elsewhere in production code.)

Optional: We usually derive a bunch of standard traits, not sure why that didn't happen here.

```suggestion
#[derive(Clone, Debug, Default, Eq, PartialEq)]
```

zebra-state/src/request.rs
```diff
@@ -224,6 +224,7 @@ pub struct ContextuallyVerifiedBlock {
 }
 
 /// Wraps note commitment trees and the history tree together.
+#[derive(Default)]
```
teor2345: Optional: We usually derive a bunch of standard traits, not sure why that didn't happen here.

Let's also document that the default represents the genesis treestate.

```suggestion
#[derive(Clone, Debug, Default, Eq, PartialEq)]
```

teor2345 (dismissed):
Thanks for this, I have some optional suggestions about standard trait derives (and documenting them).

zebra-chain/src/parallel/tree.rs
upbqdn: > Let's document that the default represents the genesis treestate. (Which is how other tree defaults work elsewhere in production code.)

Done in [71fbc1d](https://github.com/ZcashFoundation/zebra/pull/7923/commits/71fbc1d3d38c5a79e8f7322660b32d9185d0826e). I also documented the difference between the default root on its own, and the default root of the default tree in [2087424](https://github.com/ZcashFoundation/zebra/pull/7923/commits/2087424df095a2f928f8597cceddeea3fd800e0e).

teor2345:
I added the CI failure to ticket #7898 because it seemed very similar.

teor2345 (approved):

Issue #7845, opened 2023-10-26
Tracking: Add parameters to `Network::Testnet` variant


arya2:
## Motivation

Zebra currently hard-codes some consensus parameters that may need to be changed for testing.

## Scope

To be discussed in https://github.com/ZcashFoundation/zebra/discussions/7967.

#### Preparatory Refactors:
- [x] #8325
- [x] #8326
- [x] #7968

After these changes are merged, we should decide which parameters we want to add and create more issues like #7970 for each of them.

#### Adding Parameters:
- [x] #7970
  - [x] #7840
- [x] #8361
- [x] #8362
- [x] #8327
- [x] #8412
- [ ] Add a `default_port` field to `NetworkParameters` (needed for custom Testnets but not Regtest)
- [x] #8364 
- [x] #8365
- [x] #8366
- [x] #8367

### Possible Design

- Turn the `Testnet` variant into:
 ```rust
Testnet(Arc<NetworkParameters { 
    cache_name,
    genesis_hash,
    network_id,
    default_port,
    activation_heights,
    ..
>)
```
- Consider adding fields for:
  - `slow_start_interval` (nSubsidySlowStartInterval)
  - `target_difficulty_limit` (powLimit)
  - `key_prefixes`
  - `debug_mandatory_checkpoint_height: Option<Height>`
  - `checkpoints: Path`
  - `pow_avg_window` (update logic to use `min(num_blocks_in_cache, 17)`)
- Add a `new_testnet()` method or `DEFAULT_TESTNET` constant with the existing testnet defaults, to replace literal `Testnet` variant.
- Replace matching against `Testnet` with an `is_mainnet()` method. (This is only required when we're matching against `Testnet`, we can change the `Mainnet` matches gradually.)

mpguerra:
Lowering priority as it's not clear that it's a blocker for ZSA testing anymore

arya2:
Elevating priority so ZSAs can be tested with a new genesis block.

teor2345:
> Elevating priority so ZSAs can be tested with a new genesis block.

Mining a new genesis block is out of scope for this ticket. It's an extremely tricky process, let's check what `zcashd` does, and open another ticket?

(It's possible they just re-use the mainnet or testnet block for all networks.)

daira:
zcashd uses a separate genesis block for testnet, and does not support multiple testnets without modifying code and rebuilding. Is it sufficient to just use that block and upgrade at height 1?

[...]

PR #7924, opened 2023-11-08
change(network): Add configurable network parameters to `Network::Testnet`

arya2:
## Motivation

This PR is a step towards adding the `Regtest` network to Zebra, adding optional `cache_name`, `genesis_hash`, `network_id`, `default_port`, and `activation_heights` fields to `Network::Testnet`.

For testing ZSAs, it's not yet clear what other fields will be needed or whether the scope of this approach will be viable.

Part of #7845.

## Solution

- Adds new fields to `TestnetParameters` in `Testnet`
- Uses values in `TestnetParameters`, if any, instead of hard-coded defaults
- Updates references

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/tests/vectors.rs
+++ b/zebra-chain/src/block/tests/vectors.rs
@@ -202,7 +202,7 @@ fn block_test_vectors_height_mainnet() {
 fn block_test_vectors_height_testnet() {
     let _init_guard = zebra_test::init();
 
-    block_test_vectors_height(Testnet(None.into()));
+    block_test_vectors_height(Network::new_testnet());
 }
 
 /// Test that the block test vector indexes match the heights in the block data,
@@ -254,7 +254,7 @@ fn block_commitment_mainnet() {
 fn block_commitment_testnet() {
     let _init_guard = zebra_test::init();
 
-    block_commitment(Testnet(None.into()));
+    block_commitment(Network::new_testnet());
 }
 
 /// Check that the block commitment field parses without errors.
--- a/zebra-chain/src/block/tests/vectors.rs
+++ b/zebra-chain/src/block/tests/vectors.rs
@@ -202,7 +202,7 @@ fn block_test_vectors_height_mainnet() {
 fn block_test_vectors_height_testnet() {
     let _init_guard = zebra_test::init();
 
-    block_test_vectors_height(Testnet);
+    block_test_vectors_height(Testnet(None.into()));
 }
 
 /// Test that the block test vector indexes match the heights in the block data,
@@ -213,7 +213,7 @@ fn block_test_vectors_height(network: Network) {
             zebra_test::vectors::MAINNET_BLOCKS.iter(),
             zebra_test::vectors::MAINNET_FINAL_SAPLING_ROOTS.clone(),
         ),
-        Testnet => (
+        Testnet(_) => (
             zebra_test::vectors::TESTNET_BLOCKS.iter(),
             zebra_test::vectors::TESTNET_FINAL_SAPLING_ROOTS.clone(),
         ),
@@ -254,7 +254,7 @@ fn block_commitment_mainnet() {
 fn block_commitment_testnet() {
     let _init_guard = zebra_test::init();
 
-    block_commitment(Testnet);
+    block_commitment(Testnet(None.into()));
 }
 
 /// Check that the block commitment field parses without errors.
@@ -267,7 +267,7 @@ fn block_commitment(network: Network) {
             zebra_test::vectors::MAINNET_BLOCKS.iter(),
             zebra_test::vectors::MAINNET_FINAL_SAPLING_ROOTS.clone(),
         ),
-        Testnet => (
+        Testnet(_) => (
             zebra_test::vectors::TESTNET_BLOCKS.iter(),
             zebra_test::vectors::TESTNET_FINAL_SAPLING_ROOTS.clone(),
         ),
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -49,14 +49,14 @@ mod tests;
 const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
 
 /// An enum describing the possible network choices.
-#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub enum Network {
     /// The production mainnet.
     #[default]
     Mainnet,
 
     /// The oldest public test network.
-    Testnet(TestnetParameters),
+    Testnet(#[serde(skip)] TestnetParameters),
 }
 
 /// Configures testnet chain parameters to use instead of default values.
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -218,7 +218,8 @@ impl Network {
 
     /// Return the network cache name.
     pub fn cache_name(&self) -> String {
-        self.with_testnet_parameters(NetworkParameters::cache_name)
+        self.testnet_parameters()
+            .cache_name()
             .map_or_else(|| self.lowercase_name(), ToString::to_string)
     }
 
@@ -233,14 +234,11 @@ impl Network {
     }
 
     /// Returns testnet parameters, if any.
-    pub fn with_testnet_parameters<T, F>(&self, f: F) -> Option<T>
-    where
-        F: FnOnce(NetworkParameters) -> Option<T>,
-    {
+    pub fn testnet_parameters(&self) -> TestnetParameters {
         if let Network::Testnet(testnet_parameters) = *self {
-            testnet_parameters.0.and_then(f)
+            testnet_parameters
         } else {
-            None
+            None.into()
         }
     }
 }
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -80,27 +80,27 @@ impl From<Option<NetworkParameters>> for TestnetParameters {
 impl TestnetParameters {
     /// Returns `network_id` in network parameters, if any.
     pub fn network_id(self) -> Option<[u8; 4]> {
-        self.0.and_then(NetworkParameters::network_id)
+        self.0?.network_id()
     }
 
     /// Returns `default_port` in network parameters, if any.
     pub fn default_port(self) -> Option<u16> {
-        self.0.and_then(NetworkParameters::default_port)
+        self.0?.default_port()
     }
 
     /// Returns `cache_name` in network parameters, if any.
     pub fn cache_name(self) -> Option<&'static str> {
-        self.0.and_then(NetworkParameters::cache_name)
+        self.0?.cache_name()
     }
 
     /// Returns `genesis_hash` in network parameters, if any.
     pub fn genesis_hash(self) -> Option<&'static str> {
-        self.0.and_then(NetworkParameters::genesis_hash)
+        self.0?.genesis_hash()
     }
 
     /// Returns `activation_heights` in network parameters, if any.
     pub fn activation_heights(self) -> Option<&'static [(Height, NetworkUpgrade)]> {
-        self.0.and_then(NetworkParameters::activation_heights)
+        self.0?.activation_heights()
     }
 }
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -239,7 +239,7 @@ impl Network {
     }
 
     /// Returns `true` if this network is a testing network.
-    pub fn is_testnet(&self) -> bool {
+    pub fn is_a_test_network(&self) -> bool {
         matches!(*self, Network::Testnet(_))
     }
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -6,11 +6,8 @@ use thiserror::Error;
 
 use crate::{
     block::{Height, HeightDiff},
-    parameters::NetworkUpgrade::Canopy,
+    parameters::NetworkUpgrade::{self, Canopy},
 };
-
-#[cfg(any(test, feature = "proptest-impl"))]
-use proptest_derive::Arbitrary;
 
 #[cfg(test)]
 mod tests;
@@ -52,22 +49,70 @@ mod tests;
 const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
 
 /// An enum describing the possible network choices.
-#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
-#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 pub enum Network {
     /// The production mainnet.
     #[default]
     Mainnet,
 
     /// The oldest public test network.
-    Testnet,
+    Testnet(TestnetParameters),
+}
+
+/// Configures testnet chain parameters to use instead of default values.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct TestnetParameters(Option<ChainParameters>);
+
+/// Configures chain parameters to use instead of default values.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct ChainParameters {
+    network_id: Option<NetworkId>,
+    default_port: Option<u16>,
+    cache_name: Option<&'static str>,
+    genesis_hash: Option<&'static str>,
+    activation_heights: Option<&'static [(Height, NetworkUpgrade)]>,
+}
+
+/// The network id to expect in AddrV2 messages.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct NetworkId {
+    ipv4: u8,
+    ipv6: u8,
+}
+
+impl TestnetParameters {
+    /// Accepts a closure that takes `ChainParameters` and returns an `Option`, and
+    /// a `default` to return if there are no chain parameters or if the
+    /// provided closure returns `None`.
+    ///
+    /// Returns value from the provided closure if it is `Some`, or `None`.
+    pub fn get_or<T, F>(self, f: F, default: T) -> T
+    where
+        F: FnOnce(ChainParameters) -> Option<T>,
+    {
+        self.0.and_then(f).unwrap_or(default)
+    }
+}
+
+impl std::ops::Deref for TestnetParameters {
+    type Target = Option<ChainParameters>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<Option<ChainParameters>> for TestnetParameters {
+    fn from(value: Option<ChainParameters>) -> Self {
+        Self(value)
+    }
 }
 
 impl From<Network> for &'static str {
     fn from(network: Network) -> &'static str {
         match network {
             Network::Mainnet => "Mainnet",
-            Network::Testnet => "Testnet",
+            Network::Testnet(_) => "Testnet",
         }
     }
 }
@@ -85,11 +130,26 @@ impl fmt::Display for Network {
 }
 
 impl Network {
+    /// Creates a new `Testnet` with no parameters.
+    pub fn new_testnet() -> Self {
+        Self::new_testnet_with_parameters(None)
+    }
+
+    /// Creates a new `Testnet` with the provided parameters.
+    pub fn new_testnet_with_parameters(parameters: impl Into<TestnetParameters>) -> Self {
+        Self::Testnet(parameters.into())
+    }
+
+    /// Creates a new `Mainnet`.
+    pub fn new_mainnet() -> Self {
+        Self::Mainnet
+    }
+
     /// Get the default port associated to this network.
     pub fn default_port(&self) -> u16 {
         match self {
             Network::Mainnet => 8233,
-            Network::Testnet => 18233,
+            Network::Testnet(params) => params.get_or(|p| p.default_port, 18233),
         }
     }
 
@@ -116,7 +176,8 @@ impl Network {
     pub fn bip70_network_name(&self) -> String {
         match self {
             Network::Mainnet => "main".to_string(),
-            Network::Testnet => "test".to_string(),
+            // TODO: add network name field?
+            Network::Testnet { .. } => "test".to_string(),
         }
     }
 
@@ -125,9 +186,37 @@ impl Network {
         self.to_string().to_ascii_lowercase()
     }
 
+    /// Return the network cache name.
+    pub fn cache_name(&self) -> String {
+        self.with_testnet_parameters(|p| p.cache_name)
+            .map_or_else(|| self.lowercase_name(), |s| s.to_string())
+    }
+
+    /// Returns `true` if this network is `Mainnet`.
+    pub fn is_mainnet(&self) -> bool {
+        *self == Network::Mainnet
+    }
+
     /// Returns `true` if this network is a testing network.
-    pub fn is_a_test_network(&self) -> bool {
-        *self != Network::Mainnet
+    pub fn is_testnet(&self) -> bool {
+        matches!(*self, Network::Testnet(_))
+    }
+
+    /// Returns testnet parameters, if any.
+    pub fn testnet_parameters(&self) -> Option<ChainParameters> {
+        if let Network::Testnet(testnet_parameters) = *self {
+            *testnet_parameters
+        } else {
+            None
+        }
+    }
+
+    /// Returns testnet parameter if it is some.
+    pub fn with_testnet_parameters<T, F>(&self, f: F) -> Option<T>
+    where
+        F: FnOnce(ChainParameters) -> Option<T>,
+    {
+        self.testnet_parameters().and_then(f)
     }
 }
 
@@ -137,7 +226,7 @@ impl FromStr for Network {
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         match string.to_lowercase().as_str() {
             "mainnet" => Ok(Network::Mainnet),
-            "testnet" => Ok(Network::Testnet),
+            "testnet" => Ok(Network::new_testnet()),
             _ => Err(InvalidNetworkError(string.to_owned())),
         }
     }
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -78,17 +78,17 @@ impl From<Option<NetworkParameters>> for TestnetParameters {
 }
 
 impl TestnetParameters {
-    /// Returns `network_id` in chain parameters, if any.
+    /// Returns `network_id` in network parameters, if any.
     pub fn network_id(self) -> Option<NetworkId> {
         self.0.and_then(NetworkParameters::network_id)
     }
 
-    /// Returns `default_port` in chain parameters, if any.
+    /// Returns `default_port` in network parameters, if any.
     pub fn default_port(self) -> Option<u16> {
         self.0.and_then(NetworkParameters::default_port)
     }
 
-    /// Returns `cache_name` in chain parameters, if any.
+    /// Returns `cache_name` in network parameters, if any.
     pub fn cache_name(self) -> Option<&'static str> {
         self.0.and_then(NetworkParameters::cache_name)
     }
@@ -98,13 +98,13 @@ impl TestnetParameters {
         self.0.and_then(NetworkParameters::genesis_hash)
     }
 
-    /// Returns `activation_heights` in chain parameters, if any.
+    /// Returns `activation_heights` in network parameters, if any.
     pub fn activation_heights(self) -> Option<&'static [(Height, NetworkUpgrade)]> {
         self.0.and_then(NetworkParameters::activation_heights)
     }
 }
 
-/// Configures chain parameters to use instead of default values.
+/// Configures network parameters to use instead of default values.
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 pub struct NetworkParameters {
     network_id: Option<NetworkId>,
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -107,10 +107,20 @@ impl TestnetParameters {
 /// Configures network parameters to use instead of default values.
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 pub struct NetworkParameters {
+    /// Used as the network magic, Zebra will reject messages and connections from peers
+    /// with a different network magic.
     network_id: Option<[u8; 4]>,
+
+    /// Default port for the network
     default_port: Option<u16>,
+
+    /// Network portion of zebra-state cache path
     cache_name: Option<&'static str>,
+
+    /// Genesis hash for this testnet
     genesis_hash: Option<&'static str>,
+
+    /// Activation heights for this testnet
     activation_heights: Option<&'static [(Height, NetworkUpgrade)]>,
 }
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -79,7 +79,7 @@ impl From<Option<NetworkParameters>> for TestnetParameters {
 
 impl TestnetParameters {
     /// Returns `network_id` in network parameters, if any.
-    pub fn network_id(self) -> Option<NetworkId> {
+    pub fn network_id(self) -> Option<[u8; 4]> {
         self.0.and_then(NetworkParameters::network_id)
     }
 
@@ -107,7 +107,7 @@ impl TestnetParameters {
 /// Configures network parameters to use instead of default values.
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 pub struct NetworkParameters {
-    network_id: Option<NetworkId>,
+    network_id: Option<[u8; 4]>,
     default_port: Option<u16>,
     cache_name: Option<&'static str>,
     genesis_hash: Option<&'static str>,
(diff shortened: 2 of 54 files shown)


teor2345:
> For testing ZSAs, it's not yet clear what other fields will be needed or whether the scope of this approach will be viable.

It sounds like this ticket should be `do-not-merge` until we find that out?

zebra-chain/src/block/tests/vectors.rs
```diff
@@ -202,7 +202,7 @@ fn block_test_vectors_height_mainnet() {
 fn block_test_vectors_height_testnet() {
     let _init_guard = zebra_test::init();
 
-    block_test_vectors_height(Testnet);
+    block_test_vectors_height(Testnet(None.into()));
```
teor2345: For ease of implementation, and minimising changes, this could be replaced with importing a `Testnet` constant at the top of the file.

(And the appropriate lint exceptions to allow the constant to have non-standard capitalisation.)

zebra-chain/src/block/tests/vectors.rs
```diff
@@ -213,7 +213,7 @@ fn block_test_vectors_height(network: Network) {
             zebra_test::vectors::MAINNET_BLOCKS.iter(),
             zebra_test::vectors::MAINNET_FINAL_SAPLING_ROOTS.clone(),
         ),
-        Testnet => (
+        Testnet(_) => (
```
teor2345, line 216: I don't think we should be matching on `Mainnet` or `Testnet` directly, because it's only going to make these kinds of changes harder. Every time we change the structure of the `Network` enum, we'll have to change all 400 lines across 50+ files.

Instead, let's use the existing `is_a_test_network()` method, switch all this code to it in a PR, then implement the rest of this PR on top of that PR?

arya2: > Instead, let's create an is_mainnet() method, switch all this code to it in a PR, then implement the rest of this PR on top of that PR?

Tsk. I added that method and forgot to use it, I'll fix that.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -271,7 +271,9 @@ impl NetworkUpgrade {
         };
         match network {
             Mainnet => mainnet_heights,
-            Testnet => testnet_heights,
+
+            // TODO: use params if available
+            Testnet(params) => params.activation_heights().unwrap_or(testnet_heights),
```
teor2345, line 276: Seems like this TODO is implemented?

zebra-chain/src/primitives/address.rs
```diff
@@ -57,7 +59,7 @@ impl From<Network> for zcash_address::Network {
     fn from(network: Network) -> Self {
         match network {
             Network::Mainnet => zcash_address::Network::Main,
-            Network::Testnet => zcash_address::Network::Test,
+            Network::Testnet(_) => zcash_address::Network::Test,
```
teor2345, line 62: This is incorrect, this conversion is only valid if `network == Network::new_testnet()`.

Otherwise, it is an unsupported network.

zebra-chain/src/primitives/zcash_note_encryption.rs
```diff
@@ -34,7 +34,7 @@ pub fn decrypts_successfully(transaction: &Transaction, network: Network, height
                         output,
                     )
                 }
-                Network::Testnet => {
+                Network::Testnet(_) => {
```
teor2345, line 37: This is incorrect, use the conversion in `primitives::address` instead.

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
     Mainnet,
 
     /// The oldest public test network.
-    Testnet,
+    Testnet(#[serde(skip)] TestnetParameters),
+}
+
+/// Configures testnet network parameters to use instead of default values.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct TestnetParameters(Option<NetworkParameters>);
```
teor2345, line 64: This will resolve the enum size lints, and also make argument passing and cloning much more efficient.

However, it will also make `Network` into a non-Copy type, so we should save it for another PR.

```suggestion
pub struct TestnetParameters(Option<Arc<NetworkParameters>>);
```

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+impl From<Option<NetworkParameters>> for TestnetParameters {
+    fn from(value: Option<NetworkParameters>) -> Self {
+        Self(value)
+    }
+}
+
+impl TestnetParameters {
+    /// Returns `network_id` in network parameters, if any.
+    pub fn network_id(self) -> Option<[u8; 4]> {
+        self.0.and_then(NetworkParameters::network_id)
```
teor2345: This might be easier to read as:

```rust
        self.0?.network_id()
```

This applies to all these implementations.

zebra-chain/src/parameters/network.rs
teor2345, line 64: I don't think it's clear what `TestnetParameters(None)` means, and it makes the code more confusing and complex.

Instead, I suggest using a method that creates a testnet with the default parameters, wherever you would have used `None`.

This also improves test coverage, because we're using the same codepath for the standard testnet and alternate testnets.

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+    default_port: Option<u16>,
+
+    /// Network portion of zebra-state cache path
+    cache_name: Option<&'static str>,
+
+    /// Genesis hash for this testnet
+    genesis_hash: Option<&'static str>,
+
+    /// Activation heights for this testnet
+    activation_heights: Option<&'static [(Height, NetworkUpgrade)]>,
```
teor2345, line 124: I suggest making all of these parameters required, and using the default for the public testnet if the parameter isn't supplied. Then we can delete `TestnetParameters` and just use `NetworkParameters` instead.

This avoids:
- the optional-of-optionals design, which is difficult to understand
- re-implementing each method twice for NetworkParameters and TestnetParameters
- poor test coverage due to a large number of code paths

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+        self.activation_heights
+    }
 }
 
+/// The network id to expect in AddrV2 messages.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct NetworkId {
+    ipv4: u8,
+    ipv6: u8,
+}
```
teor2345, line 150: The network ID inside messages does not change based on the Zcash network.

Here is the specification for `AddrV2` address IDs, which does not depend on the Zcash network:
https://zips.z.cash/zip-0155#specification

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+    genesis_hash: Option<&'static str>,
+
+    /// Activation heights for this testnet
+    activation_heights: Option<&'static [(Height, NetworkUpgrade)]>,
+}
+
+impl NetworkParameters {
+    fn network_id(self) -> Option<[u8; 4]> {
+        self.network_id
+    }
```
teor2345, line 130: I think this might be a misunderstanding of how the Zcash network protocol works, or I might have used the wrong words to describe it.

What you're looking for is the network magic number:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-network/src/constants.rs#L418-L425

Which is checked in every network message's header:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-network/src/protocol/external/codec.rs#L166
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-network/src/protocol/external/codec.rs#L392-L394

And you'll need to replace this code with the relevant method:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-network/src/protocol/external/types.rs#L27-L35

This should be documented here, but it hasn't been documented yet:
https://github.com/zcash/zips/blob/b1410c0f42c0f08429046309393038bb6f275113/zip-0204.rst#L4

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+pub struct NetworkId {
+    ipv4: u8,
+    ipv6: u8,
+}
 impl From<Network> for &'static str {
     fn from(network: Network) -> &'static str {
         match network {
             Network::Mainnet => "Mainnet",
-            Network::Testnet => "Testnet",
+            Network::Testnet(_) => "Testnet",
```
teor2345, line 155: This is sort of correct, but it replaces custom testnets with the default in the config. Let's document that behaviour.

zebra-chain/src/parameters/network.rs
```diff
@@ -116,7 +216,8 @@ impl Network {
     pub fn bip70_network_name(&self) -> String {
         match self {
             Network::Mainnet => "main".to_string(),
-            Network::Testnet => "test".to_string(),
+            // TODO: add network name field?
```
teor2345, line 219: Isn't this just `cache_name`, without the trailing "net"?

zebra-chain/src/block/tests/vectors.rs
teor2345: This method already exists:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-chain/src/parameters/network.rs#L129

zebra-chain/src/parameters/network.rs
```diff
@@ -125,9 +226,30 @@ impl Network {
...
+
+    /// Returns `true` if this network is `Mainnet`.
+    pub fn is_mainnet(&self) -> bool {
+        *self == Network::Mainnet
+    }
+
     /// Returns `true` if this network is a testing network.
-    pub fn is_a_test_network(&self) -> bool {
-        *self != Network::Mainnet
+    pub fn is_testnet(&self) -> bool {
```
teor2345: The method name `is_a_test_network()` was chosen specifically so that reviewers could see that it was `true` for any testnet, not just the default public testnet. So let's not change it?

zebra-chain/src/sprout/tests/tree.rs
```diff
@@ -101,7 +101,7 @@ fn incremental_roots_with_blocks_for_network(network: Network) -> Result<()> {
             &*vectors::MAINNET_FINAL_SPROUT_ROOTS,
             MAINNET_FIRST_JOINSPLIT_HEIGHT,
         ),
-        Network::Testnet => (
+        Network::Testnet(_) => (
             &*vectors::TESTNET_BLOCKS,
             &*vectors::TESTNET_FINAL_SPROUT_ROOTS,
             TESTNET_FIRST_JOINSPLIT_HEIGHT,
```
teor2345, line 107: It seems like this code has been copied around a lot, maybe we should replace it with methods in `test::vectors` that take `is_mainnet: bool`?

This can happen in an earlier PR, and then all this test code doesn't need to change at all in this PR.

zebra-chain/src/work/difficulty.rs
```diff
@@ -394,7 +394,7 @@ impl ExpandedDifficulty {
             /* 2^243 - 1 */
             Network::Mainnet => (U256::one() << 243) - 1,
             /* 2^251 - 1 */
-            Network::Testnet => (U256::one() << 251) - 1,
+            Network::Testnet(_) => (U256::one() << 251) - 1,
```
teor2345, line 397: This should become a `NetworkParameter`, because it is different for regtest.

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -95,7 +95,7 @@ fn funding_stream_address_period(height: Height, network: Network) -> u32 {
 fn funding_stream_address_index(height: Height, network: Network) -> usize {
     let num_addresses = match network {
         Network::Mainnet => FUNDING_STREAMS_NUM_ADDRESSES_MAINNET,
-        Network::Testnet => FUNDING_STREAMS_NUM_ADDRESSES_TESTNET,
+        Network::Testnet(_) => FUNDING_STREAMS_NUM_ADDRESSES_TESTNET,
```
teor2345, line 98: Let's check what `zcashd` does here.

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -60,7 +60,7 @@ pub fn height_for_first_halving(network: Network) -> Height {
         Network::Mainnet => Canopy
             .activation_height(network)
             .expect("canopy activation height should be available"),
-        Network::Testnet => FIRST_HALVING_TESTNET,
+        Network::Testnet(_) => FIRST_HALVING_TESTNET,
```
teor2345, line 63: Let's check what `zcashd` does here.

zebra-consensus/src/checkpoint/list.rs
```diff
@@ -55,16 +54,33 @@ const TESTNET_CHECKPOINTS: &str = include_str!("test-checkpoints.txt");
...
+        };
+
+        match checkpoint_list.hash(block::Height(0)) {
+            Some(hash) if hash == genesis_hash(network) => checkpoint_list,
+            Some(_) => {
+                panic!("The hard-coded genesis checkpoint does not match the network genesis hash")
+            }
+            None => unreachable!("Parser should have checked for a missing genesis checkpoint"),
+        }
+    }
```
teor2345, line 75: This is only correct if `network == Network::new_testnet()`. If it's not, then the list should consist of only the checkpoints supplied by the custom network. Currently that's a single item list containing the genesis block hash.

~~If we implement it that way, then there's no need to check the genesis checkpoint in production code, and that check can be moved to a unit test.~~

zebra-consensus/src/checkpoint/list.rs
```diff
@@ -122,11 +116,9 @@ impl CheckpointList {
...
-            Some((block::Height(0), hash))
-                if (hash == &genesis_hash(Network::Mainnet)
-                    || hash == &genesis_hash(Network::Testnet)) => {}
+            Some((block::Height(0), hash)) if hash == &genesis_hash(network) => {}
             Some((block::Height(0), _)) => {
-                Err("the genesis checkpoint does not match the Mainnet or Testnet genesis hash")?
+                Err("the genesis checkpoint does not match the network genesis hash")?
             }
             Some(_) => Err("checkpoints must start at the genesis block height 0")?,
             None => Err("there must be at least one checkpoint, for the genesis block")?,
```
teor2345, line 124: Rather than complicating this code by adding a network parameter, let's move the checks that depend on `network` into `new()`. (And replace the similar check already in `new()`.)

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+        self.0.and_then(NetworkParameters::activation_heights)
+    }
+}
+
+/// Configures network parameters to use instead of default values.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct NetworkParameters {
+    /// Used as the network magic, Zebra will reject messages and connections from peers
+    /// with a different network magic.
+    network_id: Option<[u8; 4]>,
```
teor2345, line 112: This is going to make the code a bit easier to read, maybe we could move the type from `zebra-network` to `zebra-chain` in a PR, then base this PR on it?

```suggestion
    network_magic: Option<network::Magic>,
```

zebra-network/src/config.rs
```diff
@@ -225,7 +225,7 @@ impl Config {
     pub fn initial_peer_hostnames(&self) -> &IndexSet<String> {
         match self.network {
             Network::Mainnet => &self.initial_mainnet_peers,
-            Network::Testnet => &self.initial_testnet_peers,
+            Network::Testnet(_) => &self.initial_testnet_peers,
```
teor2345, line 228: This is incorrect and could cause issues with the reliability of the public testnet. A custom testnet needs to supply its own peers if it changes certain parameters.

Currently those are the genesis hash, the activation heights, and the network magic. Let's write code that checks if these parameters match the default testnet parameters?

There's a specific way to do that so that each newly added parameter field causes a compile error, and we have to add it to the check.

zebra-network/src/protocol/external/codec.rs
```diff
@@ -628,6 +628,9 @@ impl Codec {
     /// Currently, Zebra parses received `addrv2`s, ignoring some address types.
     /// Zebra never sends `addrv2` messages.
     pub(super) fn read_addrv2<R: Read>(&self, reader: R) -> Result<Message, Error> {
+        // TODO: Deserialize this field without `ZcashDeserialize`, check if there is some network id
+        //       in `self.with_testnet_parameters(NetworkParameters::network_id)`, and use that instead
+        //       of the default value. Check that serialization uses configured values too.
         let addrs: Vec<AddrV2> = reader.zcash_deserialize_into()?;
```
teor2345: This is incorrect and does not follow the specification in https://zips.z.cash/zip-0155

zebra-chain/src/parameters/network.rs
arya2: Yeah it is, I'll change that to `network_name` and add "net" in the method.

zebra-network/src/protocol/external/types.rs
```diff
@@ -101,16 +101,16 @@ impl Version {
...
             (Mainnet, Blossom) => 170_009,
-            (Testnet, Heartwood) => 170_010,
+            (Testnet(_), Heartwood) => 170_010,
             (Mainnet, Heartwood) => 170_011,
-            (Testnet, Canopy) => 170_012,
+            (Testnet(_), Canopy) => 170_012,
             (Mainnet, Canopy) => 170_013,
-            (Testnet, Nu5) => 170_050,
+            (Testnet(_), Nu5) => 170_050,
             (Mainnet, Nu5) => 170_100,
```
teor2345, line 114: This is probably incorrect for non-default testnets. What does `zcashd` do here?

We could require the `CURRENT_NETWORK_PROTOCOL_VERSION` for any non-default testnet, because that's usually what they actually want (only connect to up to date peers).

zebra-state/src/service/finalized_state.rs
```diff
@@ -428,7 +428,7 @@ impl FinalizedState {
 
             let mut blocks_size_to_dump = match self.network {
                 Network::Mainnet => MAINNET_AWAY_FROM_TIP_BULK_SIZE,
-                Network::Testnet => TESTNET_AWAY_FROM_TIP_BULK_SIZE,
+                Network::Testnet(_) => TESTNET_AWAY_FROM_TIP_BULK_SIZE,
             };
```
teor2345, line 432: This code is incorrect, and will fail if the public or custom testnet has a significant number of 2MB blocks. Instead, the value of `MAINNET_AWAY_FROM_TIP_BULK_SIZE` should be used for all networks.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -504,8 +504,8 @@ fn address_variant(address: &transparent::Address) -> u8 {
     match (address.network(), address) {
         (Mainnet, PayToPublicKeyHash { .. }) => 0,
         (Mainnet, PayToScriptHash { .. }) => 1,
-        (Testnet, PayToPublicKeyHash { .. }) => 2,
-        (Testnet, PayToScriptHash { .. }) => 3,
+        (Testnet(_), PayToPublicKeyHash { .. }) => 2,
+        (Testnet(_), PayToScriptHash { .. }) => 3,
```
teor2345, line 508: This might be incorrect for custom testnets if they change their address formats. What does `zcashd` do here?

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -107,7 +107,7 @@ lazy_static! {
     pub static ref FUNDING_STREAM_HEIGHT_RANGES: HashMap<Network, std::ops::Range<Height>> = {
         let mut hash_map = HashMap::new();
         hash_map.insert(Network::Mainnet, Height(1_046_400)..Height(2_726_400));
-        hash_map.insert(Network::Testnet, Height(1_028_500)..Height(2_796_000));
+        hash_map.insert(Network::new_testnet(), Height(1_028_500)..Height(2_796_000));
```
teor2345, line 110: This will cause a panic with a cryptic error message for custom testnets: 
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-consensus/src/block/subsidy/funding_streams.rs#L32

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -127,7 +127,7 @@ lazy_static! {
         testnet_addresses.insert(FundingStreamReceiver::Ecc, FUNDING_STREAM_ECC_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
         testnet_addresses.insert(FundingStreamReceiver::ZcashFoundation, FUNDING_STREAM_ZF_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
         testnet_addresses.insert(FundingStreamReceiver::MajorGrants, FUNDING_STREAM_MG_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
-        addresses_by_network.insert(Network::Testnet, testnet_addresses);
+        addresses_by_network.insert(Network::new_testnet(), testnet_addresses);
```
teor2345, line 130: This will cause a panic here for custom testnets:
https://github.com/ZcashFoundation/zebra/blob/2efc6bf4b2875f16a987bcbff0f90cff98eda3fa/zebra-consensus/src/block/subsidy/funding_streams.rs#L126-L128

zebra-chain/src/parameters/network.rs
teor2345: Let's check what `zcashd` does for "regtest" first, because I don't think it will be "regtestnet".

zebra-network/src/config.rs
teor2345: I think the specific check that we want here is:
If the network is an incompatible custom testnet (that is, it changes any incompatible parameters from the default testnet values), and the configured initial testnet peers contains any of the default testnet peers, then panic.

zebra-chain/src/parameters/network.rs
arya2: I wanted to leave placeholders and didn't look at it close enough on the first pass. Fixed in https://github.com/ZcashFoundation/zebra/pull/7924/commits/8fa3a533283d8043ce01d4d690af6cdc4b4eb8db.

zebra-chain/src/block/tests/vectors.rs
teor2345: Ah actually I don't think this is a good idea, because then every reviewer has to notice the difference between `Testnet` (variant) and `Testnet` (default public testnet constant).

Instead, we could have a `PUBLIC_TESTNET` constant, and do a mass replace?

zebra-chain/src/parameters/network.rs
```diff
@@ -85,11 +170,26 @@ impl fmt::Display for Network {
 }
 
 impl Network {
+    /// Creates a new `Testnet` with no parameters.
+    pub fn new_testnet() -> Self {
```
teor2345, line 174: It looks like we're using this method for both:
- the canonical public testnet used in tests
- the custom testnet parameters used by `zebrad` when configured with Testnet

They can be the same, but we also want to make it easy for them to be different. And make the code easy to maintain.

So here's what I suggest to make the code clearer and easier to maintain:
- use a PUBLIC_TESTNET constant in tests, and make it test-only/proptest-impl using #cfg
- use a new_configured_testnet() method in production code, it should only be called once when the network config is loaded, so it can be pub(crate)
- make all the other methods that can create a testnet into test-only/proptest-impl using #cfg 

After that change, anything that doesn't compile is an incorrect use of the public testnet in production code, or an incorrect use of the configured testnet in test code.

We'll also need tests for custom parameter changes, so we can't make those methods private, they need to be called from tests.

```suggestion
    #[cfg(any(test, feature = "proptest-impl"))]
    pub(crate) fn new_public_testnet() -> Self {
```

zebra-chain/src/parameters/network.rs
```diff
@@ -85,11 +170,26 @@ impl fmt::Display for Network {
...
+        Self::new_testnet_with_parameters(None)
+    }
+
+    /// Creates a new `Testnet` with the provided parameters.
+    pub fn new_testnet_with_parameters(parameters: impl Into<TestnetParameters>) -> Self {
```
teor2345, line 179: Except I think this one needs to be private so it can be called from both new_public_testnet() and new_configured_testnet(). (But not anything else.)

```suggestion
    fn new_testnet_with_parameters(parameters: impl Into<TestnetParameters>) -> Self {
```

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+    network_id: Option<[u8; 4]>,
+
+    /// Default port for the network
+    default_port: Option<u16>,
+
+    /// Network portion of zebra-state cache path
+    cache_name: Option<&'static str>,
+
+    /// Genesis hash for this testnet
+    genesis_hash: Option<&'static str>,
```
teor2345, line 121: This should be a block::Hash, because it has a fixed length:
```suggestion
    genesis_hash: Option<block::Hash>,
```

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct NetworkParameters {
+    /// Used as the network magic, Zebra will reject messages and connections from peers
+    /// with a different network magic.
+    network_id: Option<[u8; 4]>,
+
+    /// Default port for the network
+    default_port: Option<u16>,
+
+    /// Network portion of zebra-state cache path
```
teor2345, line 117: Note for future reference or for the ticket you're going to open if this is out of scope:

```rust
    /// Network portion of zebra-state, zebra-network, and zebra-scan cache paths
```

zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 9787acf42b2d604cd553a8b2bb0b59552f213587fb162e15f36c66a1147b0986 # shrinks to total_number_of_peers = 2
```
teor2345, line 7: This shouldn't be committed unless is represents a genuine failure. But it most likely represents a bug in this PR.

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+    fn default_port(self) -> Option<u16> {
+        self.default_port
+    }
+    fn cache_name(self) -> Option<&'static str> {
+        self.cache_name
+    }
+    fn genesis_hash(self) -> Option<&'static str> {
+        self.genesis_hash
+    }
+    fn activation_heights(self) -> Option<&'static [(Height, NetworkUpgrade)]> {
```
teor2345, line 140: This list should eventually include network protocol versions, which requires the Version type to be moved from zebra-network. Let's open a ticket?

I think the Version is our of scope because the network magic prevents network crossover, and there won't be multiple significant new network upgrades during testing.

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+
+impl From<Option<NetworkParameters>> for TestnetParameters {
+    fn from(value: Option<NetworkParameters>) -> Self {
+        Self(value)
+    }
+}
+
+impl TestnetParameters {
+    /// Returns `network_id` in network parameters, if any.
+    pub fn network_id(self) -> Option<[u8; 4]> {
```
teor2345, line 82: Taking self by value only works while this is a Copy type. Instead, take &self (by reference) to make a future transition to a non-Copy type easier.

zebra-chain/src/parameters/network.rs
teor2345: And regardless of the internally stored type, I would suggest a return value that represents the actual requirements of the type: each network upgrade is unique in the list.
```suggestion
    fn activation_heights(self) -> HashMap<NetworkUpgrade, Height> {
```

zebra-network/src/protocol/external/types.rs
teor2345: I think this is out of scope for now, let's open another ticket.

zebra-chain/src/parameters/network.rs
```diff
@@ -125,9 +226,30 @@ impl Network {
         self.to_string().to_ascii_lowercase()
```
teor2345, line 226: This might need to depend on cache_name, but that belongs in another ticket.

zebra-chain/src/parameters/network.rs
```diff
@@ -85,11 +170,26 @@ impl fmt::Display for Network {
...
+    }
+
     /// Get the default port associated to this network.
     pub fn default_port(&self) -> u16 {
         match self {
             Network::Mainnet => 8233,
-            Network::Testnet => 18233,
+            Network::Testnet(params) => params.default_port().unwrap_or(18233),
         }
     }
```
teor2345, line 194: mandatory_checkpoint_height should be modified to avoid a panic.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -388,9 +390,11 @@ impl NetworkUpgrade {
         height: block::Height,
     ) -> Option<Duration> {
         match (network, height) {
-            (Network::Testnet, height) if height < TESTNET_MINIMUM_DIFFICULTY_START_HEIGHT => None,
+            (Network::Testnet(_), height) if height < TESTNET_MINIMUM_DIFFICULTY_START_HEIGHT => {
```
teor2345, line 393: Let's open a ticket to make this constant and TESTNET_MAX_TIME_START_HEIGHT into a parameter?

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+    default_port: Option<u16>,
+
+    /// Network portion of zebra-state cache path
+    cache_name: Option<&'static str>,
+
+    /// Genesis hash for this testnet
+    genesis_hash: Option<&'static str>,
+
+    /// Activation heights for this testnet
+    activation_heights: Option<&'static [(Height, NetworkUpgrade)]>,
```
teor2345, line 124: Usually the unique type is listed first.
```suggestion
    activation_heights: Option<&'static [(NetworkUpgrade, Height)]>,
```

zebra-chain/src/parameters/network.rs
teor2345: I think the Version is our of scope because the network magic prevents network crossover, and there won't be multiple significant new network upgrades during testing.

zebra-network/src/protocol/external/types.rs
teor2345: I think CURRENT_NETWORK_PROTOCOL_VERSION is fine for now, there will only be one node version running on custom testnets.

zebra-chain/src/parameters/network.rs
teor2345: I think this is out of scope, the testnet cache is fine for most use cases.

zebra-chain/src/work/difficulty.rs
teor2345: I think this could be out of scope for now, we could open a separate ticket for all the difficulty changes.

zebra-consensus/src/checkpoint/list.rs
teor2345: If changing genesis is out of scope, then this should be the hard-coded testnet genesis block hash.

zebra-network/src/config.rs
teor2345: A simpler check would be to just check against the default testnet for now.

zebra-chain/src/parameters/network.rs
```diff
@@ -53,21 +50,109 @@ const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
...
+
+    /// Returns `activation_heights` in network parameters, if any.
+    pub fn activation_heights(self) -> Option<&'static [(Height, NetworkUpgrade)]> {
+        self.0?.activation_heights()
+    }
+}
+
+/// Configures network parameters to use instead of default values.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
+pub struct NetworkParameters {
```
teor2345, line 109: I think the first useful parameter to add is `activation_heights`, let's open a ticket?
In another ticket it would be useful to add an NU6 variant and a random consensus branch ID.

teor2345 (commented):
Draft review, I'll resolve most of it.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: > What does zcashd do here?

I'm not sure, I imagine it may be storing the entire prefix, I'll ask.

I think this could be fine as long as we use always different cache directories, but it would be safer to use different values for `Regtest` or custom testnets (without differentiating between different kinds of custom testnets).

arya2:
The changes in this PR and the requirements of the issue it was intended to close have diverged.

PR #7935, opened 2023-11-10
feat(zebra-state): Adds a method to ChainTipChange for getting unseen blocks when the best chain changes

### LINKED ISSUE #7927: feat(zebra-state): Send a reference to latest non-finalized state in TipAction::Reset
## Motivation

This is needed for the scanner task in zebra-scan (https://github.com/ZcashFoundation/zebra/issues/7905) to get any blocks in the non-finalized portion of the best chain when there is a `Reset` action without repeated calls to the state service.

### Possible Design

- Add the `latest_non_finalized_state` watch channel receiver
- Add a method on `LatestTipChange` that returns all the blocks in the latest non-finalized chain

This is acceptable because the worst-case scanning performance is 3 blocks per second, which means we can re-scan an entire non-finalized chain before the next block arrives. (Forks usually only happen every 300+ blocks.)

### Alternative Solution

This solution is more complex than we need.

<details>
<summary>Too complicated</summary>

Add a method on `LatestTipChange`:
  - accepting a previous block hash
  - check if it matches any previous_block_hash in the non-finalized best chain (follow the old chain from the tip until reaching a common ancestor block)
  - return all blocks after and including the block with a matching previous_block_hash, or 
  - return all blocks in the non-finalized best chain if it doesn't find the provided hash

</details>

## Related Work

This ticket is required for:
- #7934

arya2:
## Motivation

This PR still needs test coverage.

This can be used by a task in zebra-scan to wait for unseen blocks in the best chain tip.

Closes #7927.

## Solution

- Adds reference to tip block in `ChainTipBlock`
- Adds `non_finalized_state_receiver` to `ChainTipChange`
- Adds `wait_for_blocks` method on `ChainTipChange` for getting any unseen blocks when the best chain changes

### Testing

This PR needs tests.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/chain_tip.rs
+++ b/zebra-state/src/service/chain_tip.rs
@@ -13,7 +13,7 @@ use tokio::sync::watch;
 use tracing::{field, instrument};
 
 use zebra_chain::{
-    block::{self, Block, Height},
+    block::{self, Block},
     chain_tip::{BestTipChanged, ChainTip},
     parameters::{Network, NetworkUpgrade},
     transaction::{self, Transaction},
@@ -484,6 +484,9 @@ pub enum TipAction {
         ///
         /// Mainly useful for logging and debugging.
         hash: block::Hash,
+
+        /// The best chain tip block.
+        block: Arc<block::Block>,
     },
 }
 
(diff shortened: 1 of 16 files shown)


zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,94 @@ impl ChainTipChange {
...
+
+        // Return block if previous block hash matches last tip block hash
+        // TODO: Add a new variant for NetworkActivation upgrades?
+        if block.previous_block_hash() == last_change_hash {
+            return Ok(vec![block.block]);
+        }
+
+        // There's a new best chain
+
+        let non_finalized_state = self.non_finalized_state_receiver.cloned_watch_data();
```
arya2: This may need a correctness comment about it relying on the block write task sending the updated non-finalized state before updating the chain tip sender. (Other parts of Zebra rely on this too, is it obvious enough already, or would it be better to add the comment in the block write task?)

zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,113 @@ impl ChainTipChange {
...
+            return Ok(vec![block.block]);
+        };
+
+        // Return block if previous block hash matches last tip block hash
+        // TODO: Add a new variant for NetworkActivation upgrades?
+        if block.previous_block_hash() == last_change_hash {
+            return Ok(vec![block.block]);
+        }
+
+        // There's a new best chain
```
arya2:
```suggestion
        // There's a new best chain or this instance of `ChangeTipChange` missed a tip block change.
```

zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,114 @@ impl ChainTipChange {
...
+                    }
+                }
+            }
+        });
+
+        rx
+    }
+
+    /// Returns the new best chain tip when its previous block hash matches the last_change_hash,
+    /// or any blocks in the new best chain that it hasn't returned already otherwise.
```
arya2:
```suggestion
    /// otherwise returns any blocks in the new best chain that it hasn't returned already.
```

zebra-state/src/service/non_finalized_state.rs
```diff
@@ -463,11 +461,10 @@ impl NonFinalizedState {
             .find_map(|chain| chain.created_utxo(outpoint))
     }
 
-    /// Returns the `block` with the given hash in any chain.
-    #[allow(dead_code)]
-    pub fn any_block_by_hash(&self, hash: block::Hash) -> Option<Arc<Block>> {
+    /// Returns the `block` with the given hash in any side chain.
+    pub fn side_chain_block_by_hash(&self, hash: block::Hash) -> Option<Arc<Block>> {
```
teor2345, line 465: Please don't replace these unused methods, they exist to make sure the API is complete. Instead, add the new methods you need. That way, when someone looks at the API, they know they have to choose between alternative methods with different behaviour. 

Also, it is important to explain what a side chain is in the method documentation.

zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,114 @@ impl ChainTipChange {
         Ok(action)
     }
 
+    /// Spawns a task that listens for any new blocks on the best chain tip, sending
```
teor2345, line 527: Why spawn a new task when we can create the list when a "changed blocks" method is called on a Reset?

if I understand this code correctly, this task holds a lot of implicit state about which blocks it has and hasn't seen. (And which blocks its receiver has seen.) But that seems like an inflexible design, because what if we want multiple receivers?

What if the receiver is slow and misses two chain fork changes? (How do we delete or ignore the irrelevant blocks in the channel from the first change?)

zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,114 @@ impl ChainTipChange {
...
+    /// ## Note
+    ///
+    /// Task will exit if:
+    /// - there is a watch channel error from `LatestChainTip`
+    /// - the mpsc receiver is dropped
+    ///
+    /// # Panics
+    ///
+    /// The spawned task will panic if messages are not read from the returned channel
+    /// receiver and the channel buffer is full when the task tries to send a block.
```
teor2345, line 541: Note: Panicking when reading stops is not a robust design. But I'm not sure if that will matter, because I don't think the task is needed.

zebra-state/src/service/chain_tip.rs
```diff
@@ -509,6 +524,114 @@ impl ChainTipChange {
...
+        rx
+    }
+
+    /// Returns the new best chain tip when its previous block hash matches the last_change_hash,
+    /// otherwise returns any blocks in the new best chain that it hasn't returned already.
+    ///
+    /// # Panics
+    ///
+    /// This method is meant to be used with `spawn_wait_for_blocks` and could panic when called
+    /// after the `ChangeTipChange` instance misses a tip block change.
```
teor2345, line 580: Note: again, panicking because events don't happen in the expected order seems risky.

teor2345 (changes requested):
I understand this PR is still in progress, but I wanted to suggest a significant design change: rather than keeping an implied list of the blocks seen by the receiver using a task, have the receiver supply an explicit list of seen blocks to a method on the Reset variant.

This provides the following advantages:
- there can be multiple receivers, and they don't have to be in sync with each other
- we only do the work to find the changed blocks when a receiver asks for them
- the blocks seen by the receiver are an explicit list, rather than being implied by the structure of the task and its channels
- the receiver doesn't have to have processed the entire previous chain to get the correct set of changed blocks
- the code is much simpler

So I think we could replace the seen blocks / reset task with:
- putting the best chain's blocks and block hashes (not just the tip) in the `Reset` variant
- a "changed blocks" method called by the receiving task which takes a list of recent block hashes it has processed

For example, it could look like:
```rust
fn new_best_chain_blocks(&self, seen_blocks: &OrderedMap<block::Hash>) -> Vec<Arc<Block>> {
    // if the variant is Grow, return the tip block
    // find block hashes in the best chain that aren't in seen_blocks
    // return those blocks
}
```

Because forks can be different lengths, we need to keep 100 blocks (the non-finalised best chain length), plus 100 times the maximum change in difficulty per block. We have two constants for it in Zebra (increase and decrease).

So when we process a block we add it to seen_blocks, then remove any blocks over that limit.

zebra-state/src/service/non_finalized_state.rs
teor2345: Or maybe these methods aren't needed if the design changes?

arya2:
This is not needed for now.

PR #7940, opened 2023-11-14
cleanup(clippy): Remove unnecessary `try_into()` with `cargo clippy --fix`

teor2345:
## Motivation

This is a routine cleanup of nightly clippy warnings.

  - ~~Will the PR name make sense to users?~~
  - ~~Have you added or updated tests?~~
  - ~~Is the documentation up to date?~~

Not a significant or user change, no test or doc changes needed.

Is the code change in Orchard correct? Is every diversifier value a valid point?

## Solution

Remove unnecessary `try_from()` using Clippy.

### Testing

This already has test coverage.

## Review

### Reviewer Checklist

  - [x] Does the PR scope match the ticket?
  - [x] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [x] Are all the PR blockers dealt with?

### DIFF
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -525,8 +525,7 @@ mod tests {
             INVENTORY_ROTATION_INTERVAL
                 < Duration::from_secs(
                     POST_BLOSSOM_POW_TARGET_SPACING
-                        .try_into()
-                        .expect("non-negative"),
+                        .into(),
                 ),
             "we should expire inventory every time 1-2 new blocks get generated"
         );
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -523,7 +523,7 @@ mod tests {
 
         assert!(
             INVENTORY_ROTATION_INTERVAL
-                < Duration::from_secs(POST_BLOSSOM_POW_TARGET_SPACING.into(),),
+                < Duration::from_secs(POST_BLOSSOM_POW_TARGET_SPACING.into()),
             "we should expire inventory every time 1-2 new blocks get generated"
         );
     }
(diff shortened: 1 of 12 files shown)


zebra-network/src/constants.rs
```diff
@@ -523,11 +523,7 @@ mod tests {
 
         assert!(
             INVENTORY_ROTATION_INTERVAL
-                < Duration::from_secs(
-                    POST_BLOSSOM_POW_TARGET_SPACING
-                        .try_into()
-                        .expect("non-negative"),
-                ),
+                < Duration::from_secs(POST_BLOSSOM_POW_TARGET_SPACING.into(),),
```
teor2345:
```suggestion
                < Duration::from_secs(POST_BLOSSOM_POW_TARGET_SPACING.into()),
```

zebra-state/src/service/finalized_state/disk_format/block.rs
```diff
@@ -91,9 +91,7 @@ impl TransactionIndex {
 
     /// Returns this index as a `usize`
     pub fn as_usize(&self) -> usize {
-        self.0
-            .try_into()
-            .expect("the maximum valid index fits in usize")
+        self.0.into()
     }
 
     /// Creates a transaction index from a `u64`.
```
upbqdn, line 97: Why don't implement the standard `From` trait for all the methods in this `impl` block?

upbqdn (approved):
> Is the code change in Orchard correct?

Yes.


> Is every diversifier value a valid point?

Yes.

zebra-state/src/service/finalized_state/disk_format/block.rs
teor2345: Here are some possible reasons:
- integer constants can't be used to convert into the type, because there are multiple possible From methods for a generic integer
- it's hard to see what integer types are actually being converted in reviews, and sometimes that is important due to their byte size or potential overflow
- the code is outdated and could be cleaned up

In high-level code using From is better, but in the database sometimes it's important to know how big things are. I'm not sure if that applies here.

Issue #7941, opened 2023-11-14
feat(config): Add keys to be scanned to Zebra config file 


oxarbitrage:
We want to add an array of keys to be scanned to the zebra config so we can add them to the key storage at startup. 

The keys field can be a vector of tuples  `Vec<(String, Option<height>)>` where the first argument is the viewing key as a string and height is an optional birth date. Or it can just be `Vec<String>` if we want to add the birth height later. 

Tasks:

- [ ] Add key list to zebra config. (New "scan" section)

teor2345:
> This is blocked on #7904 as the keys present in the config should be added to the storage at startup.

If we need to unblock this ticket, we can split connecting the keys up to storage into another PR or ticket. Totally up to you.

> Tasks:
> 
> * [ ]  Add key list to zebra config. (New "scanner" section?)

I'd suggest `scan`, because most of our [config names match the crate names](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/tests/common/configs/v1.4.0.toml)

oxarbitrage:
I think you are right, i unblocked it, created https://github.com/ZcashFoundation/zebra/issues/7946 which is blocked on this one and in #7904

Issue #7928, opened 2023-11-08
feat(zebra-scan): Make the key storage database persistent


oxarbitrage:
We have a RAM key database as part of https://github.com/ZcashFoundation/zebra/issues/7904

We want to make this persistent, this task is blocked on https://github.com/ZcashFoundation/zebra/issues/7904 and https://github.com/ZcashFoundation/zebra/issues/7926

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/7928?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

teor2345:
This was already done in PR #8036

PR #7942, opened 2023-11-14
feat(scanner): Add a very basic RAM database to store keys and scan results

### LINKED ISSUE #7904: feat(zebra-scan): Create a key storage database in RAM
As part of the scanning project, we are going to be holding user defined viewing keys and scanning results in a storage database.

Proposed structure:

```
pub struct Storage {
    /// The key and an optional birthday for it.
    keys: HashMap<String, Option<Height>>,

    /// The key and the related transaction id.
    results: HashMap<String, Vec<Hash>>
}
```

- [ ] Create the storage module.
- [ ] Create a basic test for it.

This key storage database should be always ready for being traversed by the scanning task.

As the keys are initially added to the config no delete methods should be implemented as part of this ticket.

oxarbitrage:
## Motivation

We want to have a database with viewing keys and scanning results. The scanner task can get the keys to scan from this database and insert results of found transactions.

This is the most basic version i was able to do with a simple test. 

Should close https://github.com/ZcashFoundation/zebra/issues/7904, unblock https://github.com/ZcashFoundation/zebra/issues/7941 and partially unblock https://github.com/ZcashFoundation/zebra/issues/7928

## Solution

Add a basic storage for the blockchain scanner project.

### Testing

A new test that uses the new database was added.

## Review

I am open to suggestions if we keep the PR simple, i prefer to open new tickets for more complex tasks.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/storage.rs
+++ b/zebra-scan/src/storage.rs
@@ -4,14 +4,18 @@ use std::collections::HashMap;
 
 use zebra_chain::{block::Height, transaction::Hash};
 
+/// The type used in Zebra to store Sapling scanning keys.
+/// It can represent a full viewing key or an individual viewing key.
+pub type SaplingScanningKey = String;
+
 /// Store key info and results of the scan.
 #[allow(dead_code)]
 pub struct Storage {
     /// The key and an optional birthday for it.
-    keys: HashMap<String, Option<Height>>,
+    keys: HashMap<SaplingScanningKey, Option<Height>>,
 
     /// The key and the related transaction id.
-    results: HashMap<String, Vec<Hash>>,
+    results: HashMap<SaplingScanningKey, Vec<Hash>>,
 }
 
 #[allow(dead_code)]
@@ -25,12 +29,12 @@ impl Storage {
     }
 
     /// Add a key to the storage.
-    pub fn add_key(&mut self, key: String, birthday: Option<Height>) {
+    pub fn add_key(&mut self, key: SaplingScanningKey, birthday: Option<Height>) {
         self.keys.insert(key, birthday);
     }
 
     /// Add a result to the storage.
-    pub fn add_result(&mut self, key: String, txid: Hash) {
+    pub fn add_result(&mut self, key: SaplingScanningKey, txid: Hash) {
         if let Some(results) = self.results.get_mut(&key) {
             results.push(txid);
         } else {
@@ -44,7 +48,7 @@ impl Storage {
     }
 
     /// Get all keys.
-    pub fn get_keys(&self) -> Vec<String> {
+    pub fn get_keys(&self) -> Vec<SaplingScanningKey> {
         self.keys.keys().cloned().collect()
     }
 }
--- a/zebra-scan/src/storage.rs
+++ b/zebra-scan/src/storage.rs
@@ -1,0 +1,50 @@
+//! Store viewing keys and results of the scan.
+
+use std::collections::HashMap;
+
+use zebra_chain::{block::Height, transaction::Hash};
+
+/// Store key info and results of the scan.
+#[allow(dead_code)]
+pub struct Storage {
+    /// The key and an optional birthday for it.
+    keys: HashMap<String, Option<Height>>,
+
+    /// The key and the related transaction id.
+    results: HashMap<String, Vec<Hash>>,
+}
+
+#[allow(dead_code)]
+impl Storage {
+    /// Create a new storage.
+    pub fn new() -> Self {
+        Self {
+            keys: HashMap::new(),
+            results: HashMap::new(),
+        }
+    }
+
+    /// Add a key to the storage.
+    pub fn add_key(&mut self, key: String, birthday: Option<Height>) {
+        self.keys.insert(key, birthday);
+    }
+
+    /// Add a result to the storage.
+    pub fn add_result(&mut self, key: String, txid: Hash) {
+        if let Some(results) = self.results.get_mut(&key) {
+            results.push(txid);
+        } else {
+            self.results.insert(key, vec![txid]);
+        }
+    }
+
+    /// Get the results of a key.
+    pub fn get_results(&self, key: &str) -> Option<&Vec<Hash>> {
+        self.results.get(key)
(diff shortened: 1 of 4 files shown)


zebra-scan/src/storage.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use std::collections::HashMap;
+
+use zebra_chain::{block::Height, transaction::Hash};
+
+/// Store key info and results of the scan.
+#[allow(dead_code)]
+pub struct Storage {
+    /// The key and an optional birthday for it.
+    keys: HashMap<String, Option<Height>>,
```
oxarbitrage: Should we use something else instead of `String`?

zebra-scan/src/storage.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// The key and the related transaction id.
+    results: HashMap<String, Vec<Hash>>,
```
oxarbitrage: Is `Hash` a reasonable thing to store here? Should we also store block height where the transaction is ?

zebra-scan/src/storage.rs
teor2345: I think `String` is the best type for now, because:
- there are [3 different types of keys](https://docs.rs/zcash_client_backend/latest/zcash_client_backend/scanning/trait.ScanningKey.html#foreign-impls) we can use [scan_block](https://docs.rs/zcash_client_backend/latest/zcash_client_backend/scanning/fn.scan_block.html) with: full viewing keys, individual transaction viewing keys, and individual viewing keys with a scope and nullifier deriving key.
- To use `HashMap`, the type needs to implement `Hash`, but [SaplingIvk](https://docs.rs/zcash_primitives/0.13.0/x86_64-unknown-linux-gnu/zcash_primitives/sapling/keys/struct.SaplingIvk.html) and [DiversifiableFullViewingKey](https://docs.rs/zcash_client_backend/latest/zcash_client_backend/keys/sapling/struct.DiversifiableFullViewingKey.html) don't do that.

I think we could easily support full and individual transaction keys using `String`. I put the scope/nullifier deriving key as a separate item on the tracking issue, because it seemed complicated.

**Tracking Issue:** If we want a more specific type that is guaranteed to be valid, we can open a separate ticket for an enum across all the key types. I put that on the tracking issue.

**Optional:** For now we could make a type alias to `SaplingScanningKey`, so it's easier to read now, and change later?
```rust
/// The type used in Zebra to store Sapling scanning keys.
/// It can represent a full viewing key or an individual viewing key.
pub type SaplingScanningKey = String;
```

zebra-scan/src/storage.rs
teor2345: **Optional:** Using "sapling" in names will be less confusing, particularly when we add Orchard.

This would also need changes to all the method names.

```rust
    /// The key and an optional birthday for it. If the birthday is not supplied, the Sapling activation height is used.
    sapling_keys: HashMap<SaplingScanningKey, Option<Height>>,

    /// The key and the related transaction id.
    sapling_results: HashMap<SaplingScanningKey, Vec<Hash>>,
```

teor2345: I don't think we need to store the block height or hash. Using the transaction hash, we can get the block height and hash from the state, so there's no need to store it here as well.

**New Ticket:** I opened ticket #7943 with the details of the new state request.

teor2345: **New ticket:** Do we want to store the decrypted results of scanning in this PR, or do that in a later PR?

These types seem pretty complicated for a first PR, so I will split this into another ticket. It would also be helpful to add type aliases for `SaplingSpend` and `SaplingOutput`, so we can change them later.

```suggestion
    /// The new outputs created in each transaction, for each scanning key. 
    sapling_spend_results: HashMap<String, HashMap<Hash, WalletSaplingSpend>>,

    /// The spent outputs by each transaction, for each scanning key.
    /// If the scanning key is an individual viewing key (without a nullifier deriving key), there are no nullifiers in those results.
    sapling_spend_results: HashMap<String, HashMap<Hash, WalletSaplingOutput<Option<Nullifier>>>>,
```

zebra-scan/src/storage.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// Get the results of a key.
+    pub fn get_results(&self, key: &str) -> Option<&Vec<Hash>> {
+        self.results.get(key)
+    }
+
+    /// Get all keys.
+    pub fn get_keys(&self) -> Vec<String> {
+        self.keys.keys().cloned().collect()
+    }
```
teor2345: Optional: Did you want to return the birthdays from this method as well? Or from another method (`get_key_birthdays()`)?

That could be another ticket, but it seems simple enough to do here:
```suggestion
    /// Get all keys and their birthdays.
    pub fn get_keys(&self) -> HashMap<String, Option<Height>> {
        self.keys.clone()
    }
```

zebra-scan/src/storage.rs
```diff
@@ -0,0 +1,50 @@
...
+            results.push(txid);
+        } else {
+            self.results.insert(key, vec![txid]);
+        }
+    }
+
+    /// Get the results of a key.
+    pub fn get_results(&self, key: &str) -> Option<&Vec<Hash>> {
+        self.results.get(key)
+    }
```
teor2345: Optional: This return type can be simplified to `Vec`:
```suggestion
    /// Get the results for a key.
    pub fn get_results(&self, key: &str) -> Vec<Hash> {
        self.results.get(key).cloned().unwrap_or_default()
    }
```

teor2345 (dismissed):
This looks great!

All of my suggestions are optional, but some of them are so detailed that they belong in other tickets. So I'll go and open those tickets now. (And resolve those comments.)

zebra-scan/src/storage.rs
teor2345: The state request is now ticket #7943.

teor2345: This is now ticket #7944.

teor2345 (approved):
Thanks!

Issue #7892, opened 2023-11-01
change(release): Enable all user features in the `experimental` build


teor2345:
### Motivation

We want to be able to point users to an experimental build for all new user features, not just one or two features we've been focused on recently. This makes using those features easier for users, because they don't have to build Zebra themselves.

It also makes our documentation simpler, because we can just say "use the `latest-experimental` image", rather than explaining which features are in it.

### Suggested Solution:

First PR:
- [ ] Add an `experimental` feature to `zebrad` that just activates `default-release-binaries`
- [ ] Admin: Set the [`RUST_EXPERIMENTAL_FEATURES` GitHub Actions variable](https://github.com/ZcashFoundation/zebra/settings/variables/actions) to `experimental`

Second PR:
- [ ] Add all user features to the `experimental` feature: `elasticsearch`, `journald`, `prometheus`, `filter-reload`, `shielded-scan`
- [ ] Add a config test for the `experimental` feature (as well as `getblockchain-rpcs` and `shielded-scan`), and add a config with all the experimental features enabled

#### Optional

If the `Test all` job is slow:

<details>

<summary>Not needed any more - job is fast</summary>

- [ ] Split it into an experimental job and a default features job
https://github.com/ZcashFoundation/zebra/blob/afbe80706092de2287417bd5cab248a9f8167065/.github/workflows/ci-unit-tests-docker.yml#L122-L125
- [ ] Consider building an experimental image so that we get build caching for that image
- [ ] Admin: Add branch protection rules for new jobs

</details>

### Complex Code or Requirements

We don't want to make CI fail when an admin changes the settings, so it's better to split this ticket into 2 PRs.

### Testing

Make sure the tests are actually running in CI.

teor2345:
This is blocked by #7891 because we want the new features to appear in the new tags, after the breaking change.

teor2345:
I've updated this ticket, it looks like we'll need an admin for parts of it, so let's check when @gustavovalverde is around.

mpguerra:
@gustavovalverde I'm unscheduling this one because there are more pressing things this sprint and I'm a bit lost with this one... Is there anything left to do here?

arya2:
> Is there anything left to do here?

The experimental build was added to CI in #8175 and is testing every PR update: https://github.com/ZcashFoundation/zebra/actions/runs/7997001572/job/21840879753#step:5:105

[...]

PR #7949, opened 2023-11-15
feat(config): Allow to add keys to be scanned by the zebra-scan crate to config

### LINKED ISSUE #7941: feat(config): Add keys to be scanned to Zebra config file

oxarbitrage:
## Motivation

We want the user to add sapling keys to the config file so Zebra can scan the blockchain with them.

Close https://github.com/ZcashFoundation/zebra/issues/7941

## Solution

Add a config type to the zebra-scan crate with a `sapling_keys_to_scan` field of type `IndexMap<String, u32>` where the string is the key and the number the birthday height.

This field is used as:

```
[scan.sapling_keys_to_scan]
"somekey" = 1
"someotherkey" = 100000
```

The configuration is behind a `zebra-scan` rust feature so build should be done as follows to see the new section:

```
$ cargo build --features="zebra-scan"
```

### Testing

This PR has been tested manually.

## Review

This is a very basic version introducing the less amount of code possible, open to make changes.

### Reviewer Checklist


zebra-scan/src/config.rs
```diff
@@ -0,0 +1,21 @@
...
+use indexmap::IndexMap;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, default)]
+/// Configuration for scanning.
+pub struct Config {
+    /// The sapling keys to scan for and the birthday height of each of them.
+    // TODO: any value below sapling activation as the birthday height should default to sapling activation.
+    pub sapling_keys_to_scan: IndexMap<String, u32>,
```
teor2345: Optional:
- Is there any reason why this is `IndexMap` rather than `HashMap`? (If we change it we can also drop the dependency.)
- Can we use `zebra_scan::SaplingScanningKey` here?
- It might be less confusing to use the same variable name as the storage, but it really doesn't matter

```rust
    /// The sapling keys to scan for and the birthday height of each of them.
    // TODO: any value below sapling activation as the birthday height should default to sapling activation.
    //       use SaplingScanningKey and Height here
    pub sapling_scanning_keys: HashMap<String, u32>,
```

zebra-scan/src/config.rs
```diff
@@ -0,0 +1,21 @@
+//! Configuration for blockchain scanning tasks.
+
+use indexmap::IndexMap;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
teor2345: Optional: These traits are useful for testing.

```suggestion
#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
```

zebra-scan/src/config.rs
oxarbitrage: Yea, the only reason i saw is that `IndexMap` guarantees the order of the list which can be useful if we make tests.

I am happy to change it if you think it do not worth it.

zebrad/Cargo.toml
```diff
@@ -68,6 +68,9 @@ getblocktemplate-rpcs = [
     "zebra-chain/getblocktemplate-rpcs",
 ]
 
+# Experimental zebra-scan indexing
+zebra-scan = []
```
teor2345: Optional: In 2024 there will be some changes to feature names that are the same as dependency names, so we might want to pick a different feature name to the crate. (We can always fix this later if it causes problems.)
https://github.com/orgs/rust-lang/projects/30?pane=issue&itemId=28974333

This would also help avoid developer confusion between shielded and transparent scanning.

```suggestion
# Experimental shielded blockchain scanning
shielded-scan = ["zebra-scan"]
```

zebra-scan/src/config.rs
teor2345: That seems like a good reason, so let's keep `IndexMap` and say that in the field docs. So we don't accidentally "simplify" it later.

It will also guarantee the order if we read and then write the config, or if there are a large number of keys and we choose some keys to scan first.

zebrad/Cargo.toml
```diff
@@ -151,6 +154,7 @@ zebra-network = { path = "../zebra-network", version = "1.0.0-beta.31" }
 zebra-node-services = { path = "../zebra-node-services", version = "1.0.0-beta.31" }
 zebra-rpc = { path = "../zebra-rpc", version = "1.0.0-beta.31" }
 zebra-state = { path = "../zebra-state", version = "1.0.0-beta.31" }
+zebra-scan = { path = "../zebra-scan", version = "0.1.0-alpha.0" }
```
teor2345: This will enable the dependency in Zebra's production builds, but we only want it in experimental builds. Here's the first part of that change, I added the second part to ticket #7892.

```suggestion

# Experimental shielded-scan feature
zebra-scan = { path = "../zebra-scan", version = "0.1.0-alpha.0", optional = true }
```

zebrad/src/config.rs
```diff
@@ -43,4 +43,8 @@ pub struct ZebradConfig {
     #[serde(skip_serializing_if = "zebra_rpc::config::mining::Config::skip_getblocktemplate")]
     /// Mining configuration
     pub mining: zebra_rpc::config::mining::Config,
+
+    #[cfg(feature = "zebra-scan")]
+    /// Scanner configuration
+    pub scan: zebra_scan::config::Config,
```
teor2345: Note: Using the feature to enable the config is ok for now.

I added a task to the tracking issue to warn when the config is set but the feature is not. (Currently we panic, which is ok for an MVP.)

I also added a task to ticket #7892 to test parsing experimental configs. So we don't have to worry about config tests for the MVP.

teor2345: Optional: We've had some user and developer confusion between shielded scanning (which is this feature), and transparent scanning (which we already do).

Would it help to rename the config?
```suggestion
    #[cfg(feature = "zebra-scan")]
    /// Scanner configuration
    pub shielded_scan: zebra_scan::config::Config,
```

teor2345 (dismissed):
Looks great!

Adding `zebra-scan` as a dependency in production `zebrad` builds might be a blocker. Everything else is optional, I've tried to move as much work out of the MVP and onto the tracking issue as possible.

zebra-scan/src/config.rs
oxarbitrage: > Can we use zebra_scan::SaplingScanningKey here?

Ill add this ass soon as https://github.com/ZcashFoundation/zebra/pull/7942 gets merged.

zebrad/Cargo.toml
oxarbitrage: Thanks for that one.

teor2345 (approved):
Looks good to me, thanks for the quick changes

teor2345:
CI failed due to #7898:
https://github.com/ZcashFoundation/zebra/actions/runs/6884108042/job/18726073944?pr=7949#step:5:199

So I'll restart it.

PR #7952, opened 2023-11-16
security: Limit how many addresses we use from each peer address message

### LINKED ISSUE #1869: Zebra should limit the number of addresses it uses from a single Addrs response, to avoid address book takeover
## Motivation

Zebra accepts up to 1000 addresses from each peer address request. But the Zcash mainnet only has around 200 peers, and the `peerset_initial_target_size` defaults to 75.

These limits will be particularly important once we limit the size of the address book (#1873), so that peers can't flood or replace our entire address book.

## Solution

Zebra should limit the number of peers it accepts in response to each peer request.

Zebra should choose peers from each peer request using the following algorithm:
- [ ] take `peerset_initial_target_size / 2` peers at random, and leave the rest in the connection peer cache

These designs have security implications, so we shouldn't do them:
- ~~remove addresses that are already in the address book~~ this can be exploited to fill the address book with new malicious addresses
- ~~sort peers by their `untrusted_last_seen_time`, newest first~~ choosing peers based on untrusted data makes it easier to add new malicious addresses
- ~~Increase the default `crawl_and_dial` interval, to compensate for the peers we're discarding~~ we already crawl in response to unmet demand, so this isn't needed

This algorithm makes it much less likely that a majority of Zebra's peers will come from a single peer request, while also limiting the number of extra peer requests that Zebra makes.

## Context

`zcashd` does not have this issue.

teor2345:
## Motivation

We want to limit how many addresses we use from each peer, so that our outbound connections are based on peers shared by most of our peers.

Close #1869

We might want to have an overall changelog entry for the tracking issue, but it doesn't need to be added here.

We need to limit in two places: existing cached peers, and responses based on a request. These limits operate slightly differently, because only a new response can cause an overflow.

## Solution

- When we have cached peers, only take some of them to answer a request
- When we send a peer request, only take some of the response to answer a request, and put the rest in the cache
- In response to unsolicited address messages, add the latest addresses to the cache, and discard the oldest ones
- Choose response addresses at random
- Update outdated docs referring to this ticket

### Testing

Add a new proptest for the cache update method.

The remaining parts of the PR are simple and can be checked by manual review, or by existing tests (including integration tests).

## Review

This is a routine fix.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -84,6 +84,10 @@ pub const DEFAULT_MAX_CONNS_PER_IP: usize = 1;
 /// This will be used as `Config.peerset_initial_target_size` if no valid value is provided.
 pub const DEFAULT_PEERSET_INITIAL_TARGET_SIZE: usize = 25;
 
+/// The maximum number of peers we will add to the address book after each `getaddr` request.
+pub const PEER_ADDR_RESPONSE_LIMIT: usize =
+    DEFAULT_PEERSET_INITIAL_TARGET_SIZE * OUTBOUND_PEER_LIMIT_MULTIPLIER / 2;
+
 /// The buffer size for the peer set.
 ///
 /// This should be greater than 1 to avoid sender contention, but also reasonably
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -27,8 +27,8 @@ use zebra_chain::{
 
 use crate::{
     constants::{
-        self, MAX_OVERLOAD_DROP_PROBABILITY, MIN_OVERLOAD_DROP_PROBABILITY,
-        OVERLOAD_PROTECTION_INTERVAL,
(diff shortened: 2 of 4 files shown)


mpguerra:
Is this PR linked to #1869 ?

teor2345:
> Is this PR linked to #1869 ?

Yes, I've updated the ticket.

arya2 (approved):
This looks great!

I opened an optional suggestion PR (#7960), but this is good to merge as-is, feel free to remove the do-not-merge tag and close #7960 if you disagree with the suggestion.

teor2345:
> This looks great!
> 
> I opened an optional suggestion PR (#7960), but this is good to merge as-is, feel free to remove the do-not-merge tag and close #7960 if you disagree with the suggestion.

I am happy to deal with #7960 separately. (Since it is an optional suggestion, this PR does not need to be blocked by it.)

PR #7955, opened 2023-11-16
security: Rate limit GetAddr responses

### LINKED ISSUE #7823: security: Rate limit `GetAddr` responses. Credit: Ziggurat Team
### Motivation

Observation #2 from Red Team report on Zcash testnet

> it's possible to constantly request an infinite amount of GetAddr requests and Zebra will always reply.

### Specifications

_No response_

There are many alternative ways to fix this issue. One of the simpler ways is to:
- In the inbound service, generate a random list of peers once, and send a copy to every peer that requests addresses
- When the rate-limit time is reached, generate another random list, and send copies of it

This means that Zebra controls what addresses it sends out, rather than peers controlling how many addresses they can get using repeated requests, multiple connections, or multiple reconnections.

#### Rough Calculations for Refresh Period

Current behaviour:
- Zebra requests new peers from 3 of its peers at least every minute
- Zebra only provides peers that have been live in the last 3 hours

Goal:
- Only want to give out 80% of our address book every hour
- We want our peers to have a reasonable chance of getting live peers

Parameters:
- Only give out 1/4 of our available peers (reduced from 1/3)
- Rotate every 10 minutes - this also happens to be close to the block timeout

### Testing

_No response_

### Related Work

_No response_

### LINKED ISSUE #7968: Add an empty `NetworkParameters` struct to `Network::Testnet`
## Motivation 

Depends-On: #8325, #8326

### Possible Design

- Add an empty and optional `TestnetParameters` struct in `Network::Testnet`
- Add a `Network::new_testnet()` method that returns `Network::Testnet { ..default_testnet_params }`
  - Mass replace any remaining instances of `Testnet` in tests with `Network::new_testnet()`

Example (with fields that shouldn't be added yet):
 ```rust
    cache_name,
    genesis_hash,
    network_id,
    default_port,
    activation_heights
>)
```

### PR DESCRIPTION
**Pull request by arya2**
## Motivation

We want to avoid giving out Zebra's entire address book over a short duration.

Closes #7823.

## Solution

- Updates `ADDR_RESPONSE_LIMIT_DENOMINATOR` to 4
- Adds a 10 minute `INBOUND_CACHED_ADDRS_REFRESH_INTERVAL` constant
- Adds a `CachedPeerAddrs` struct with a reference to the address book
- Replaces the `address_book` field in `Setup::Initialized` with an instance of `CachedPeerAddrs`
- Refreshes the cached addresses in the inbound service's `poll_ready()` method if they are stale
- Returns `cached_addrs` in response to `inbound::Peers` requests instead of reading from the address book
- Logs a warning if it hasn't been able to get a lock on the address book to update the cached_addrs in an unexpectedly long time

Related cleanups:

- Moves logic for truncating list of peers for `GetAddr` response to a `sanitized_window()` method on `AddressBook`.

### Testing

TODO: This PR still needs tests.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -281,7 +281,7 @@ impl AddressBook {
 
     /// Get the active addresses in `self` in random order with sanitized timestamps,
     /// including our local listener address.
-    pub fn sanitized(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
+    pub(crate) fn sanitized(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
         use rand::seq::SliceRandom;
         let _guard = self.span.enter();
 
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -270,7 +270,7 @@ impl AddressBook {
     /// including our local listener address.
     ///
     /// Limited to a the number of peer addresses Zebra should give out per `GetAddr` request.
-    pub fn sanitized_window(&self) -> Vec<MetaAddr> {
+    pub fn fresh_get_addr_response(&self) -> Vec<MetaAddr> {
         let now = Utc::now();
         let mut peers = self.sanitized(now);
         let address_limit = peers.len().div_ceil(ADDR_RESPONSE_LIMIT_DENOMINATOR);
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -18,7 +18,7 @@ use tracing::Span;
 use zebra_chain::{parameters::Network, serialization::DateTime32};
 
 use crate::{
-    constants,
+    constants::{self, ADDR_RESPONSE_LIMIT_DENOMINATOR, MAX_ADDRS_IN_MESSAGE},
     meta_addr::MetaAddrChange,
     protocol::external::{canonical_peer_addr, canonical_socket_addr},
     types::MetaAddr,
@@ -264,6 +264,18 @@ impl AddressBook {
     /// Get the local listener [`SocketAddr`].
     pub fn local_listener_socket_addr(&self) -> SocketAddr {
         self.local_listener
+    }
+
+    /// Get the active addresses in `self` in random order with sanitized timestamps,
+    /// including our local listener address.
+    pub fn sanitized_window(&self) -> Vec<MetaAddr> {
+        let now = Utc::now();
+        let mut peers = self.sanitized(now);
+        let address_limit = peers.len().div_ceil(ADDR_RESPONSE_LIMIT_DENOMINATOR);
+        let address_limit = MAX_ADDRS_IN_MESSAGE.min(address_limit);
+        peers.truncate(address_limit);
+
+        peers
     }
 
     /// Get the active addresses in `self` in random order with sanitized timestamps,
--- a/zebra-network/src/address_book.rs
+++ b/zebra-network/src/address_book.rs
@@ -268,12 +268,13 @@ impl AddressBook {
 
     /// Get the active addresses in `self` in random order with sanitized timestamps,
     /// including our local listener address.
+    ///
+    /// Limited to a the number of peer addresses Zebra should give out per `GetAddr` request.
     pub fn sanitized_window(&self) -> Vec<MetaAddr> {
         let now = Utc::now();
         let mut peers = self.sanitized(now);
         let address_limit = peers.len().div_ceil(ADDR_RESPONSE_LIMIT_DENOMINATOR);
-        let address_limit = MAX_ADDRS_IN_MESSAGE.min(address_limit);
-        peers.truncate(address_limit);
+        peers.truncate(MAX_ADDRS_IN_MESSAGE.min(address_limit));
 
         peers
     }
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -515,7 +515,7 @@ mod tests {
         );
 
         assert!(
-            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE,
+            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE * 2,
             "the address book limit should actually be used"
         );
     }
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -309,7 +309,7 @@ pub const MAX_ADDRS_IN_MESSAGE: usize = 1000;
 ///
 /// This limit makes sure that Zebra does not reveal its entire address book
 /// in a single `Peers` response.
-pub const ADDR_RESPONSE_LIMIT_DENOMINATOR: usize = 3;
+pub const ADDR_RESPONSE_LIMIT_DENOMINATOR: usize = 4;
 
 /// The maximum number of addresses Zebra will keep in its address book.
 ///
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -501,6 +501,10 @@ mod tests {
     fn ensure_address_limits_consistent() {
         // Zebra 1.0.0-beta.2 address book metrics in December 2021.
         const TYPICAL_MAINNET_ADDRESS_BOOK_SIZE: usize = 4_500;
+        // Minimum address book size that allows Zebra to fill the max_addrs_in_message
+        // without giving out more than `1 / ADDR_RESPONSE_LIMIT_DENOMINATOR` of its peers
+        const MINIMUM_MAINNET_ADDRESS_BOOK_SIZE: usize =
+            MAX_ADDRS_IN_MESSAGE * (ADDR_RESPONSE_LIMIT_DENOMINATOR + 1);
 
         let _init_guard = zebra_test::init();
 
@@ -515,7 +519,8 @@ mod tests {
         );
 
         assert!(
-            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE * 2,
+            MAX_ADDRS_IN_ADDRESS_BOOK
+                <= TYPICAL_MAINNET_ADDRESS_BOOK_SIZE.max(MINIMUM_MAINNET_ADDRESS_BOOK_SIZE),
             "the address book limit should actually be used"
         );
     }
--- a/zebra-network/src/constants.rs
+++ b/zebra-network/src/constants.rs
@@ -499,12 +499,9 @@ mod tests {
     #[test]
     #[allow(clippy::assertions_on_constants)]
     fn ensure_address_limits_consistent() {
-        // Zebra 1.0.0-beta.2 address book metrics in December 2021.
-        const TYPICAL_MAINNET_ADDRESS_BOOK_SIZE: usize = 4_500;
-        // Minimum address book size that allows Zebra to fill the max_addrs_in_message
-        // without giving out more than `1 / ADDR_RESPONSE_LIMIT_DENOMINATOR` of its peers
-        const MINIMUM_MAINNET_ADDRESS_BOOK_SIZE: usize =
-            MAX_ADDRS_IN_MESSAGE * (ADDR_RESPONSE_LIMIT_DENOMINATOR + 1);
+        // Estimated network address book size in November 2023, after the address book limit was increased.
+        // Zebra 1.0.0-beta.2 address book metrics in December 2021 showed 4500 peers.
+        const TYPICAL_MAINNET_ADDRESS_BOOK_SIZE: usize = 5_500;
 
         let _init_guard = zebra_test::init();
 
@@ -519,8 +516,7 @@ mod tests {
         );
 
         assert!(
-            MAX_ADDRS_IN_ADDRESS_BOOK
-                <= TYPICAL_MAINNET_ADDRESS_BOOK_SIZE.max(MINIMUM_MAINNET_ADDRESS_BOOK_SIZE),
+            MAX_ADDRS_IN_ADDRESS_BOOK <= TYPICAL_MAINNET_ADDRESS_BOOK_SIZE,
             "the address book limit should actually be used"
         );
     }
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -124,28 +124,21 @@ pub struct CachedPeerAddrs {
 }
 
 impl CachedPeerAddrs {
-    /// Creates a new [`CachedPeerAddrs`].
+    /// Creates a new empty [`CachedPeerAddrs`].
     fn new(address_book: Arc<Mutex<AddressBook>>) -> Self {
-        let cached_addrs = address_book
-            .lock()
-            .expect("previous thread panicked while holding the address book lock")
-            .sanitized_window();
-
-        let refresh_time = Instant::now() + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
-
         Self {
             address_book,
-            cached_addrs,
-            refresh_time,
+            cached_addrs: Vec::new(),
+            refresh_time: Instant::now(),
         }
     }
 
-    /// Refreshes the `cached_addrs` if the time has past `refresh_time`
+    /// Refreshes the `cached_addrs` if the time has past `refresh_time` or the cache is empty
     fn try_refresh(&mut self) {
         let now = Instant::now();
 
-        // return early if the cached addresses are still fresh
-        if now < self.refresh_time {
+        // return early if there are some cached addresses, and they are still fresh
+        if now < self.refresh_time && !self.cached_addrs.is_empty() {
             return;
         }
 
@@ -457,7 +450,7 @@ impl Service<zn::Request> for Inbound {
                     if peers.is_empty() {
                         debug!(
                             "ignoring `Peers` request from remote peer because our address \
-                                book has no available peers"
+                             book has no available peers"
                         );
 
                         Ok(zn::Response::Nil)
--- a/zebrad/src/components/inbound.rs
+++ b/zebrad/src/components/inbound.rs
@@ -9,17 +9,15 @@ use std::{
     collections::HashSet,
     future::Future,
     pin::Pin,
-    sync::{Arc, TryLockError},
+    sync::{Arc, Mutex, TryLockError},
     task::{Context, Poll},
-    time::Duration,
+    time::{Duration, Instant},
 };
 
-use chrono::Utc;
 use futures::{
     future::{FutureExt, TryFutureExt},
     stream::Stream,
 };
-use num_integer::div_ceil;
 use tokio::sync::oneshot::{self, error::TryRecvError};
 use tower::{buffer::Buffer, timeout::Timeout, util::BoxService, Service, ServiceExt};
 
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: zebra-network/src/address_book.rs [hunk 1]
```diff
@@ -266,6 +266,18 @@ impl AddressBook {
...
+        let address_limit = peers.len().div_ceil(ADDR_RESPONSE_LIMIT_DENOMINATOR);
+        let address_limit = MAX_ADDRS_IN_MESSAGE.min(address_limit);
+        peers.truncate(address_limit);
+
+        peers
+    }
+
     /// Get the active addresses in `self` in random order with sanitized timestamps,
     /// including our local listener address.
     pub fn sanitized(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
```
teor2345: It's confusing to have two public methods with the same docs and similar names. Let's make the less secure one private, and update the docs to make it clear how they are different?

zebrad/src/components/inbound.rs
```diff
@@ -111,6 +111,73 @@ pub struct InboundSetupData {
     pub latest_chain_tip: zs::LatestChainTip,
 }
 
+/// Caches and refreshes a partial list of peer addresses to be returned as a `GetAddr` response.
+pub struct CachedPeerAddrs {
```
teor2345: Optional: We already have two peer address caches in Zebra, the per-connection address message cache, and the address book. So it might be clearer to call this `CachedPeerAddrResponse`. (And change the field name to match.)

zebrad/src/components/inbound.rs
```diff
@@ -111,6 +111,73 @@ pub struct InboundSetupData {
...
+    }
+}
+
+impl std::ops::Deref for CachedPeerAddrs {
+    type Target = Vec<MetaAddr>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.cached_addrs
+    }
+}
```
teor2345: This is clever, but in general `Deref` is meant for smart pointer types, and other uses can result in surprising and hard to debug behaviour:
> the rules regarding Deref and [DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, Deref **should only be implemented for smart pointers** to avoid confusion.

https://doc.rust-lang.org/std/ops/trait.Deref.html

We occasionally ignore this rule to make testing or refactors more convenient, but we try to document the reasons why we did it. There doesn't seem to be any reason to use `Deref` here - and it is confusing, because there are two address lists inside `CachedPeerAddrs`.

An explicit `cached_addr_response()` method would make the behaviour of the code much clearer.

zebra-network/src/constants.rs
```diff
@@ -511,7 +511,7 @@ mod tests {
         );
 
         assert!(
-            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE,
+            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE * 2,
             "the address book limit should actually be used"
         );
```
teor2345: I think this change makes it so the assertion predicate is different to the assert message. Maybe you could update the text?

Optional: Or you could update the typical address book size based on the 2023 network? You'll need to run Zebra for a few days with a very large address book size. (And ideally without PR #7952.)

zebrad/src/components/inbound.rs
```diff
@@ -111,6 +111,73 @@ pub struct InboundSetupData {
...
+                self.cached_addrs = address_book.sanitized_window();
+                self.refresh_time = now + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
+            }
+
+            Err(TryLockError::WouldBlock) => {
+                let next_refresh_time = self.refresh_time + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
+
+                if now > next_refresh_time {
+                    warn!("getaddrs response hasn't been refreshed in some time");
+                };
```
teor2345: Note: Locking for 10 minutes is extremely unlikely during normal operation, but a warning is a good idea in case we have a bug that causes it. Thank you for thinking of this!

zebrad/src/components/inbound.rs
```diff
@@ -317,11 +390,12 @@ impl Service<zn::Request> for Inbound {
                 // If we returned Pending here, and there were no waiting block downloads,
                 // then inbound requests would wait for the next block download, and hang forever.
                 while let Poll::Ready(Some(_)) = block_downloads.as_mut().poll_next(cx) {}
+                cached_peer_addrs.try_refresh();
```
teor2345: If Zebra's address book is under heavy load, I don't think we should try to refresh the cache every time there is a (possibly unrelated) inbound request. That would just make the load worse.

Instead, we could just try the refresh when we get an address request. It's ok to do this outside the future, because it just tries locking, and won't ever block.

This will also require doc updates (and a comment about why we're doing it there).

zebrad/src/components/inbound.rs
```diff
@@ -111,6 +111,73 @@ pub struct InboundSetupData {
...
+    }
+
+    /// Refreshes the `cached_addrs` if the time has past `refresh_time`
+    fn try_refresh(&mut self) {
+        let now = Instant::now();
+
+        // return early if the cached addresses are still fresh
+        if now < self.refresh_time {
+            return;
+        }
```
teor2345: Previously we've had network issues where we tried to do too much during startup, the inbound service queue got full, and then we disconnected some peers. So I'd like to minimise locking at startup.

Also, this code has a potential bug where it runs when the address book is empty, then keeps an empty cache for 10 minutes. This is much more likely to happen right at startup.

So I suggest making these changes:
- create an empty address book at startup
- try to refresh if the cache is stale or empty

This also simplifies the code, so we don't have to keep the sanitization and refresh intervals in `new()` and `try_refresh()` in sync.

```suggestion
    /// Creates a new empty [`CachedPeerAddrs`].
    fn new(address_book: Arc<Mutex<AddressBook>>) -> Self {
        Self {
            address_book,
            cached_addrs: Vec::new(),
            refresh_time: Instant::now(),
        }
    }

    /// Refreshes the `cached_addrs` if the time has past `refresh_time` or the cache is empty
    fn try_refresh(&mut self) {
        let now = Instant::now();

        // return early if there are some cached addresses, and they are still fresh
        if now < self.refresh_time && !self.cached_addrs.is_empty() {
            return;
        }
```

zebrad/src/components/inbound.rs
```diff
@@ -377,53 +451,14 @@ impl Service<zn::Request> for Inbound {
...
-                            );
-                        } else {
-                            debug!(
-                                "ignoring `Peers` request from remote peer because our address \
-                                 book has no available peers"
-                            );
-                        }
+                        debug!(
+                            "ignoring `Peers` request from remote peer because our address \
+                                book has no available peers"
```
teor2345: Nitpick: Rustfmt should have caught this, but it seems like it preserves spaces that are deleted in Rust strings.

```suggestion
                            "ignoring `Peers` request from remote peer because our address \
                             book has no available peers"
```

teor2345 (commented):
This looks good, I think it would be enough to just test that we get peers when we send an address message.

zebra-network/src/address_book.rs
arya2: > Let's make the less secure one private

It would require some test changes and there are other methods that give out the same information, could we create an issue for it?

Renamed to `fresh_get_addr_response()`.

zebrad/src/components/inbound.rs
arya2: I think the only way this could happen currently would be by queuing RPC requests to hold onto the lock, which isn't a concern. But yeah, can't hurt if something changes.

zebrad/src/components/inbound.rs
```diff
@@ -111,6 +110,74 @@ pub struct InboundSetupData {
     pub latest_chain_tip: zs::LatestChainTip,
 }
 
+/// Caches and refreshes a partial list of peer addresses to be returned as a `GetAddr` response.
+pub struct CachedPeerAddrResponse {
```
arya2: Move this to its own module.

zebrad/src/components/inbound.rs
teor2345: It could also happen if Zebra is under load, and peers rapidly connect and disconnect. We have rate-limits, but for outbound connections they are still quite fast.

zebra-network/src/address_book.rs
```diff
@@ -266,6 +266,19 @@ impl AddressBook {
...
+        let mut peers = self.sanitized(now);
+        let address_limit = peers.len().div_ceil(ADDR_RESPONSE_LIMIT_DENOMINATOR);
+        peers.truncate(MAX_ADDRS_IN_MESSAGE.min(address_limit));
+
+        peers
+    }
+
     /// Get the active addresses in `self` in random order with sanitized timestamps,
     /// including our local listener address.
     pub fn sanitized(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
```
teor2345: Let's see if this works, feel free to revert if it doesn't.

```suggestion
    pub(crate) fn sanitized(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
```

zebra-network/src/address_book.rs
teor2345: See commit [87a226f](https://github.com/ZcashFoundation/zebra/pull/7955/commits/87a226f3e4b898f89258eca42ec60902fce58bfc) for a compatible method visibility change. (I hope!)

If we've changed the method used by GetAddr, do those tests also need to use the new method?

I think this test does:
https://github.com/ZcashFoundation/zebra/blob/c3b11e325bccd24c16a993d46db7f847303c970b/zebra-network/src/address_book/tests/prop.rs#L39

But the other test that uses `sanitized()` doesn't. And maybe we need another test for this method specifically?

zebra-network/src/constants.rs
```diff
@@ -511,7 +515,8 @@ mod tests {
...
 
         assert!(
-            MAX_ADDRS_IN_ADDRESS_BOOK < TYPICAL_MAINNET_ADDRESS_BOOK_SIZE,
+            MAX_ADDRS_IN_ADDRESS_BOOK
+                <= TYPICAL_MAINNET_ADDRESS_BOOK_SIZE.max(MINIMUM_MAINNET_ADDRESS_BOOK_SIZE),
             "the address book limit should actually be used"
```
teor2345: `MINIMUM_MAINNET_ADDRESS_BOOK_SIZE ` is the same value as `MAX_ADDRS_IN_ADDRESS_BOOK`, so the check will always be true, regardless of the values of `MAX_ADDRS_IN_MESSAGE` or `ADDR_RESPONSE_LIMIT_DENOMINATOR`.

Instead, I suggest modifying the check so it is still meaningful, and we have to ask ourselves "is the limit too big to actually be used?" next time we change the peer set size. (That's the point of these kinds of checks.)

See my suggestion above as well.
```suggestion
            MAX_ADDRS_IN_ADDRESS_BOOK <= TYPICAL_MAINNET_ADDRESS_BOOK_SIZE,
            "the address book limit should actually be used"
```

zebra-network/src/constants.rs
```diff
@@ -497,6 +497,10 @@ mod tests {
     fn ensure_address_limits_consistent() {
         // Zebra 1.0.0-beta.2 address book metrics in December 2021.
         const TYPICAL_MAINNET_ADDRESS_BOOK_SIZE: usize = 4_500;
+        // Minimum address book size that allows Zebra to fill the max_addrs_in_message
+        // without giving out more than `1 / ADDR_RESPONSE_LIMIT_DENOMINATOR` of its peers
+        const MINIMUM_MAINNET_ADDRESS_BOOK_SIZE: usize =
+            MAX_ADDRS_IN_MESSAGE * (ADDR_RESPONSE_LIMIT_DENOMINATOR + 1);
```
teor2345:
```suggestion
        // Estimated network address book size in November 2023, after the address book limit was increased.
        // Zebra 1.0.0-beta.2 address book metrics in December 2021 showed 4500 peers.
        const TYPICAL_MAINNET_ADDRESS_BOOK_SIZE: usize = 5_500;
```

teor2345 (commented):
I think we might be over-complicating this change a bit. Just change the constant?

teor2345 (dismissed):
Nothing here is blocking, so we're good to go when you're happy with the remaining test and module changes.

zebrad/src/components/inbound.rs
arya2: > It could also happen if Zebra is under load, and peers rapidly connect and disconnect. We have rate-limits, but for outbound connections they are still quite fast.

You're right, I think that's worth "fixing", I'll open a separate PR to await the lock and a candidate outside the spawned tokio thread here: https://github.com/ZcashFoundation/zebra/blob/rate-limit-getaddr/zebra-network/src/peer_set/initialize.rs#L881

zebrad/src/components/inbound/cached_peer_addr_response.rs
```diff
@@ -0,0 +1,78 @@
...
+                self.refresh_time = now + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
+                self.value = zn::Response::Peers(peers);
+            }
+
+            Ok(_) => {
+                debug!(
+                    "could not refresh cached response because our address \
+                     book has no available peers"
+                );
+            }
```
teor2345: If our address book stays empty for a long time, I think we should stop sharing those outdated peers. An empty address book can happen if Zebra is disconnected for a long time, and all its peers become outdated. Or if it's an old version that's no longer compatible.

```suggestion
            // Update cached value and refresh_time, even if the address book is empty.
            //
            // Security: this avoids outdated gossiped peers. Outdated Zebra binaries will gradually lose all their peers,
            // because those peers refuse to connect to outdated versions. So we don't want those outdated Zebra
            // versions to keep gossiping old peer information either.
            Ok(peers) => {
                self.refresh_time = now + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
                self.value = zn::Response::Peers(peers);
            }
```

teor2345 (commented):
From the CI test results, it looks like some more test or production updates are needed.

zebrad/src/components/inbound.rs
teor2345: That fix isn't needed, because the mutex is futures-aware.

arya2:
> From the CI test results, it looks like some more test or production updates are needed.

Yeah, that `real_peer_set` test seems to be broken in CI (but works locally).

I think this needs another test too.

teor2345:
The docker CI failure is #7898

zebrad/src/components/inbound/cached_peer_addr_response.rs
arya2: Updated to stop sharing outdated peers in https://github.com/ZcashFoundation/zebra/pull/7955/commits/7c843983cd0932a6f50ffa6c0dd92de8919e8a67.

Added the comment in https://github.com/ZcashFoundation/zebra/pull/7955/commits/f72a52bf1fa7a1034b00ba19985d5a67c8948120.

zebra-network/src/address_book.rs
arya2: > If we've changed the method used by GetAddr, do those tests also need to use the new method?

I think so, some of the new method wasn't covered. Fixed in https://github.com/ZcashFoundation/zebra/pull/7955/commits/2ee9a207342a94aed5d3e871c1304673b4839e00.

zebrad/src/components/inbound/cached_peer_addr_response.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::{
+    sync::{Mutex, TryLockError},
+    time::Instant,
+};
+
+use super::*;
+
+/// The maximum duration that a `CachedPeerAddrResponse` is considered fresh before the inbound service
+/// should get new peer addresses from the address book to send as a `GetAddr` response.
+pub const INBOUND_CACHED_ADDRS_REFRESH_INTERVAL: Duration = Duration::from_secs(10 * 60);
```
teor2345: Optional Nitpick: I think this constant name could be confusing, because it's used in the inbound service, but the address book contains peer addresses for outbound connections.

Similarly for the next constant.

```rust
pub const CACHED_ADDRS_REFRESH_INTERVAL: Duration = Duration::from_secs(10 * 60);
```

zebrad/src/components/inbound/cached_peer_addr_response.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// The maximum duration that a `CachedPeerAddrResponse` is considered fresh before the inbound service
```
teor2345: These docs have been copied and pasted, so they aren't accurate for this constant.

zebrad/src/components/inbound/cached_peer_addr_response.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        let max_fresh_time = self.refresh_time + INBOUND_CACHED_ADDRS_MAX_FRESH_DURATION;
+
+        // try getting a lock on the address book if it's time to refresh the cached addresses
+        match self
+            .address_book
+            .try_lock()
+            .map(|book| book.fresh_get_addr_response())
+        {
+            // Update cached value and refresh_time, even if the address book is empty.
```
teor2345: This comment doesn't match what the code does.

```suggestion
            // Update cached value and refresh_time if there are some gossipable peers in the  address book.
```

zebrad/src/components/inbound/cached_peer_addr_response.rs
```diff
@@ -0,0 +1,96 @@
...
+            //
+            // Security: this avoids outdated gossiped peers. Outdated Zebra binaries will gradually lose all their peers,
+            // because those peers refuse to connect to outdated versions. So we don't want those outdated Zebra
+            // versions to keep gossiping old peer information either.
+            Ok(peers) if !peers.is_empty() => {
+                self.refresh_time = now + INBOUND_CACHED_ADDRS_REFRESH_INTERVAL;
+                self.value = zn::Response::Peers(peers);
+            }
+
+            Ok(_) if now > max_fresh_time => {
```
teor2345: I think there's a bug here, the max fresh time is 1 minute from the old refresh time, which is approximately `now`. But the next refresh time is 10 minutes from now. This will cause a lot of empty responses.

I think simpler code is better here, to avoid subtle bugs. And document what the code does in code comments.

zebrad/src/components/inbound/cached_peer_addr_response.rs
arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/7955/commits/1e3fb0eb7d389c697586d07bd5394ee1f9dee5db.

zebrad/src/components/inbound/cached_peer_addr_response.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/7955/commits/a4a8963c70058a16fa8502d1703e431c0f05141b.

zebrad/src/components/inbound/cached_peer_addr_response.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/7955/commits/f902adb2edae8e3827614897eed7933b62f15ec8.

zebrad/src/components/inbound/cached_peer_addr_response.rs
arya2: I agree, updated in https://github.com/ZcashFoundation/zebra/pull/7955/commits/120b6a5e1a1c29cb557d95144bea1d322d348d5c and https://github.com/ZcashFoundation/zebra/pull/7955/commits/b4dd424ba84752136eed6e540afb8fb68787ad71.

arya2:
I used up the Github actions limit with frequent commits. 😞

teor2345 (approved):
Thanks for all the changes, this looks good for now

PR #7960, opened 2023-11-17
Suggestion for 'Limit how many addresses we use from each peer address message'

arya2:
This is an optional refactor for #7952.

@teor2345 can review and merge this change if they agree with it.

### DIFF
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -27,8 +27,8 @@ use zebra_chain::{
 
 use crate::{
     constants::{
-        self, MAX_OVERLOAD_DROP_PROBABILITY, MIN_OVERLOAD_DROP_PROBABILITY,
-        OVERLOAD_PROTECTION_INTERVAL,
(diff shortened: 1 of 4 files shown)


zebra-network/src/peer/connection.rs
```diff
@@ -843,6 +829,25 @@ where
...
+                                        &mut self.cached_addrs,
+                                        new_addrs,
+                                        PEER_ADDR_RESPONSE_LIMIT,
+                                    );
+
+                                    Message::Addr(response_addrs)
+                                } else {
+                                    msg
+                                }
+                            };
```
teor2345, line 849: Maintainability Issue: Putting this closure in the middle of a 300 line method makes both that method and the closure harder to read.

teor2345 (commented):
I'm not sure if this PR is an improvement, because every time we call `process_message()`, we have to remember to use `with_cached_addrs()` on its argument. But there is no type signature or documentation that makes this call required.

This change also makes it harder to understand `process_message()` by reading its code, because the message it is being passed has been pre-filtered. (And that isn't documented anywhere.)

PR #7965, opened 2023-11-20
change(systemd): Send all zebrad logs to the journal under systemd

teor2345:
## Motivation

In bug #7961 a user reported that they only get panic messages in the systemd journal. But we'd like all Zebra logs to go there.

This does not close that ticket, but it will be useful to help diagnose future issues like it.

### Specifications

https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#StandardOutput=

The `journald` feature should also do something similar, but it will provide structured logs, rather than just log lines.

## Solution

Send all zebrad output to the journal under systemd

### Testing

This is difficult to test unless someone is already running Zebra with systemd.

## Review

This is a routine fix.

### Reviewer Checklist

### DIFF
--- a/zebrad/systemd/zebrad.service
+++ b/zebrad/systemd/zebrad.service
@@ -22,6 +22,8 @@ ExecStart=/usr/bin/zebrad start
 Restart=always
 PrivateTmp=true
 NoNewPrivileges=true
+StandardOutput=journal
+StandardError=journal
 
 [Install]
 Alias=zebrad


oxarbitrage (commented):
I think this changes make sense, will probably work but i will like  to test them locally by someone before merging. should not be that hard, could be us or we can ask the ticket creator to test it.

> This is difficult to test unless someone is already running Zebra with systemd.

It is that difficult ? If so, we should think on writing a book page on how to achieve this.

teor2345:
> > This is difficult to test unless someone is already running Zebra with systemd.
> 
> It is that difficult ? If so, we should think on writing a book page on how to achieve this.

I think it's the systemd part that's hard, adding Zebra if you're already running systemd should be pretty easy.

oxarbitrage (approved):
utACK. I assume this will not make any harm.

PR #7969, opened 2023-11-21
cleanup(zebra-network): Avoid address book lock contention in some rare cases

arya2:
## Motivation

This PR limits how often Zebra's outbound connector can try to get a lock on the address book.

The outbound connector will wait for a lock on the candidate set and address book with a short timeout before continuing with the next `crawler_action`.

## Solution

- Try to find a candidate address in the outbound connector task instead of a spawned task

## Review

Anyone can review, this PR is optional-priority and should be reviewed last.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -868,18 +868,18 @@ where
                     "opening an outbound peer connection"
                 );
 
+                // Try to get the next available peer for a handshake.
+                //
+                // candidates.next() has a short timeout, and briefly holds the address
+                // book lock, so it shouldn't hang.
+                //
+                // Hold the lock for as short a time as possible.
+                let candidate = { candidates.lock().await.next().await };
+
                 // Spawn each handshake or crawl into an independent task, so handshakes can make
                 // progress while crawls are running.
                 let handshake_or_crawl_handle = tokio::spawn(
                     async move {
-                        // Try to get the next available peer for a handshake.
-                        //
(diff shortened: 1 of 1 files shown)


zebra-network/src/peer_set/initialize.rs
```diff
@@ -868,18 +868,18 @@ where
                     "opening an outbound peer connection"
                 );
 
+                // Try to get the next available peer for a handshake.
+                //
+                // candidates.next() has a short timeout, and briefly holds the address
+                // book lock, so it shouldn't hang.
+                //
+                // Hold the lock for as short a time as possible.
+                let candidate = { candidates.lock().await.next().await };
```
teor2345, line 877: I think this specific fix might revert a previously fixed deadlock or performance issue.

The crawler needs to run independently of candidate set locks and rate-limits, so that it can get new peer addresses to add to the address book. But this change blocks the crawler loop on the candidate set rate-limit.

Instead, I would suggest documenting the existing calling code and called method using a security comment.

teor2345 (changes requested):
The goal of this PR is already satisfied by the candidate set rate-limit here:
https://github.com/ZcashFoundation/zebra/blob/3be22b2cdaa59cd1c028432764f718a3cbef370d/zebra-network/src/peer_set/candidate_set.rs#L406

Is any other change needed to satisfy this goal?

arya2:
> Is any other change needed to satisfy this goal?

No, I missed that.

PR #7977, opened 2023-11-22
security(net): Stop sending peer addresses from handshakes directly to the address book

### LINKED ISSUE #7951: Put addresses from the peer's version message and remote IP in the connection cache, don't send them directly to the address book
## Motivation

This prevents an attack where peers can repeatedly connect, supply an address in the version message, and fill the address book.

### Complex Requirements or Code

Potential drawbacks:

We might lose these addresses when the peer sends an addr message (but addr messages should always contain that peer's address, so in some cases we might lose the remote address of inbound connections). This is fixed in #7952 by adding all new addresses to the cache, taking any that are needed, then truncating the rest if needed.

These addresses are treated like gossiped addresses, so they have slightly higher priority. This is unlikely to have a significant impact, because their times will quickly become outdated by new addresses from this peer or other peers.

teor2345:
## Motivation

We want to stop sending connection addresses directly to the address book, because that's insecure.

Close #7951

This is actually simpler than the previous code, because it doesn't have concurrency issues with other updates for the same address.

Using gossiped addresses changes the outbound connection priority slightly.

## Solution

- Send handshake peer addresses to the per-connection address cache
- Update documentation

### Testing

@arya2 what tests should we write here?
Is inspection enough, or should we have a "not sent to the address book" or "sent to the cache" test?

## Review

This is a routine fix.

### Reviewer Checklist

  - [x] Does the PR scope match the ticket?
  - [x] Are all the PR blockers dealt with?

## Follow Up Work

There are some cleanups in #7824 that depend on this.

### DIFF
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -609,6 +609,7 @@ where
             .field("error_slot", &self.error_slot)
             .field("metrics_label", &self.metrics_label)
             .field("last_metrics_state", &self.last_metrics_state)
+            .field("last_overload_time", &self.last_overload_time)
             .finish()
     }
 }
--- a/zebra-network/src/peer/connection.rs
+++ b/zebra-network/src/peer/connection.rs
@@ -541,13 +541,15 @@ where
     /// other state handling.
     pub(super) request_timer: Option<Pin<Box<Sleep>>>,
 
-    /// A cached copy of the last unsolicited `addr` or `addrv2` message from this peer.
+    /// Unused peers from recent `addr` or `addrv2` messages from this peer.
+    /// Also holds the initial addresses sent in `version` messages, or guessed from the remote IP.
     ///
-    /// When Zebra requests peers, the cache is consumed and returned as a synthetic response.
-    /// This works around `zcashd`'s address response rate-limit.
+    /// When peers send solicited or unsolicited peer advertisements, Zebra puts them in this cache.
     ///
-    /// Multi-peer `addr` or `addrv2` messages replace single-peer messages in the cache.
-    /// (`zcashd` also gossips its own address at regular intervals.)
+    /// When Zebra's components request peers, some cached peers are consumed and returned as a
+    /// synthetic response. This works around `zcashd`'s address response rate-limit.
+    ///
+    /// The cache is limited to avoid denial of service attacks.
     pub(super) cached_addrs: Vec<MetaAddr>,
 
     /// The `inbound` service, used to answer requests from this connection's peer.
@@ -618,7 +620,7 @@ impl<S, Tx> Connection<S, Tx>
 where
     Tx: Sink<Message, Error = SerializationError> + Unpin,
 {
-    /// Return a new connection from its channels, services, and shared state.
+    /// Return a new connection from its channels, services, shared state, and metadata.
     pub(crate) fn new(
         inbound_service: S,
         client_rx: futures::channel::mpsc::Receiver<ClientRequest>,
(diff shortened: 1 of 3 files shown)


zebra-network/src/peer/connection.rs
```diff
@@ -541,13 +541,15 @@ where
...
+    /// Also holds the initial addresses sent in `version` messages, or guessed from the remote IP.
     ///
-    /// When Zebra requests peers, the cache is consumed and returned as a synthetic response.
-    /// This works around `zcashd`'s address response rate-limit.
+    /// When peers send solicited or unsolicited peer advertisements, Zebra puts them in this cache.
     ///
-    /// Multi-peer `addr` or `addrv2` messages replace single-peer messages in the cache.
-    /// (`zcashd` also gossips its own address at regular intervals.)
+    /// When Zebra's components request peers, some cached peers are consumed and returned as a
+    /// synthetic response. This works around `zcashd`'s address response rate-limit.
```
upbqdn: What are synthetic responses? I didn't find any information in the repo or on the web.

upbqdn (dismissed):
All changes in this PR look good to me, so I'm approving. However, I don't know the implications of sending handshake peer addresses to the per-connection address cache.

zebra-network/src/peer/connection.rs
teor2345: I think it might be clearer to say "modified response" here. I'll make a suggestion.

zebra-network/src/peer/connection.rs
```diff
@@ -541,13 +541,15 @@ where
...
+    ///
+    /// The cache is limited to avoid denial of service attacks.
```
teor2345:
```suggestion
    /// When Zebra's components request peers, some cached peers are randomly selected,
    /// consumed, and returned as a modified response. This works around `zcashd`'s address
    /// response rate-limit.
    ///
    /// The cache size is limited to avoid denial of service attacks.
```

teor2345: @upbqdn feel free to apply this if you're happy with it.

teor2345:
> However, I don't know the implications of sending handshake peer addresses to the per-connection address cache.

This PR prevents two attacks:
- peers modifying each others' addresses using `version` messages
- peers filling the address book by repeatedly connecting and sending different addresses in each `version` message

Here are some potential drawbacks:
- We might lose these addresses when the peer sends an addr message (but addr messages should always contain that peer's address, so in some cases we might lose the remote address of inbound connections). This is fixed in https://github.com/ZcashFoundation/zebra/pull/7952 by adding all new addresses to the cache, taking any that are needed, then truncating the rest if needed.
- These addresses are treated like gossiped addresses, so they have slightly higher priority. This is unlikely to have a significant impact, because their times will quickly become outdated by new addresses from this peer or other peers.

upbqdn (approved):

teor2345:
The macOS job seems to have failed due to a GitHub runner error, this is an issue with their infrastructure. There is no output in the failed steps on the GUI or in the raw logs, there's just this:
> 2023-11-23T13:58:23.3683400Z Requested labels: macos-latest
> 2023-11-23T13:58:23.3683747Z Job defined at: ZcashFoundation/zebra/.github/workflows/ci-unit-tests-os.yml@refs/pull/7977/merge
> 2023-11-23T13:58:23.3683911Z Waiting for a runner to pick up this job...
> 2023-11-23T13:58:25.8364996Z Job is waiting for a hosted runner to come online.
> 2023-11-23T13:58:29.5183703Z Job is about to start running on the hosted runner: GitHub Actions 22 (hosted)

https://github.com/ZcashFoundation/zebra/actions/runs/6970673248/job/18969099332?pr=7977

Screenshot 2023-11-24 at 09 01 07

PR #7989, opened 2023-11-23
feat(zebra-scan): Connect with zebrad

### LINKED ISSUE #7946: feat(scanner): Read and store keys from config at zebra startup
After we have a key storage and the keys in the config file we need to read them and add them to the database at zebra startup.

Tasks:

- [ ] Create a `Storage` instance and add the config keys to it.

Blocked on https://github.com/ZcashFoundation/zebra/issues/7941 and https://github.com/ZcashFoundation/zebra/issues/7904

oxarbitrage:
## Motivation

We want to pass the keys that are in the config from zebrad to zebra-scan at node startup.

Close https://github.com/ZcashFoundation/zebra/issues/7946

## Solution

- In `zebrad/src/comamnds/start.rs` a `Storage` RAM database is created and the keys are dumped into it form config. Then the scan task is spawned with the created storage instance and the state.
- The scan task is at `zebra-scan/src/scan.rs`. The task just print some logs at the moment for testing purposes.

### Testing

- A test was added to `zebrad/test/accpentance.rs`.

## Review

I think anyone can review.

### Reviewer Checklist


zebrad/src/commands/start.rs
```diff
@@ -385,6 +400,8 @@ impl StartCmd {
                     exit_when_task_finishes = false;
                     Ok(())
                 }
+
+                // TODO: add scan task which is tricky because it needs to be behind a feature.
```
oxarbitrage: I was not able to make a version for the scanner inside the `select!` macro that works behind the `zebra-scan` feature. Any input is appreciated.

zebra-scan/src/scan.rs
```diff
@@ -0,0 +1,58 @@
...
+const CHECK_INTERVAL: Duration = Duration::from_secs(10);
+
+/// Start the scan task given state and storage.
+///
+/// - This function is dummy at the moment. It just makes sure we can read the storage and the state.
+/// - Modificatiuons here might have an impact in the `scan_task_starts` test.
+/// - Real scanning code functionality will be added in the future here.
+pub async fn start(mut state: State, storage: Storage) -> Result<(), Report> {
+    // We want to make sure the state has a tip height available before we start scanning.
+    std::thread::sleep(INITIAL_WAIT);
```
teor2345: Using `thread::sleep()` in an async function will pause the whole async executor. It can also lead to compiler errors, because the future is not able to hold some kinds of data over a sleep. I'm not sure if that's the error you're having?

Try using `tokio::time::sleep(...).await` instead:
https://docs.rs/tokio/latest/tokio/time/fn.sleep.html

This also applies to the other sleep.

zebra-scan/src/scan.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        // Read keys from the storage
+        let available_keys = storage.get_sapling_keys();
+
+        for key in available_keys {
+            info!(
+                "Scanning the blockchain for key {} from block 1 to {}",
+                key.0, tip_height.0
+            );
+        }
```
teor2345: This `unwrap()` will cause a panic when the state is empty, which can happen on startup and during some tests. I'm not sure if that's the error you're having?

Since this is a just dummy code, we don't need to unwrap anything, we should just be able to print the whole tip:
```suggestion
        let tip = match request? {
            zebra_state::Response::Tip(tip) => tip,
            _ => unreachable!("unmatched response to a state::Tip request"),
        };

        // Read keys from the storage
        let available_keys = storage.get_sapling_keys();

        for key in available_keys {
            info!(
                "Scanning the blockchain for key {} from block 1 to {}",
                key.0, tip,
            );
        }
```

zebrad/src/commands/start.rs
teor2345: What compiler error do you get when you try it, and what code have you tried?

I'll make a suggestion above which might fix it, but it's hard to be sure without the actual code and error messages.

zebrad/src/commands/start.rs
```diff
@@ -285,6 +286,18 @@ impl StartCmd {
...
+        // Spawn never ending scan task.
+        let scan_task_handle = {
+            info!("spawning zebra_scanner");
+            let mut storage = zebra_scan::storage::Storage::new();
+            for (key, birthday) in config.shielded_scan.sapling_keys_to_scan.iter() {
+                storage.add_sapling_key(key.clone(), Some(zebra_chain::block::Height(*birthday)));
+            }
+
+            tokio::spawn(zebra_scan::scan::start(state, storage).in_current_span())
+        };
```
teor2345, line 299: If you need a dummy handle, here's how to make one:
https://doc.rust-lang.org/std/future/fn.pending.html

And also remove the feature from every use of the task handle, because we will always have a task handle.

```suggestion
            tokio::spawn(zebra_scan::scan::start(state, storage).in_current_span())
        };

        #[cfg(not(feature = "zebra-scan"))]
        // Spawn a dummy scan task which doesn't do anything and never finishes.
        let scan_task_handle = tokio::spawn(std::future::pending().in_current_span());
```

teor2345 (commented):
I'm not sure what errors you're getting, so I've tried to make some suggestions about what it could be.

I can also look at the specific error if you post the code or the error.

zebra-scan/src/scan.rs
oxarbitrage: Thank you for that, i don't do async code since a while. The initial sleep is just to make sure we have a tip but it can be removed. I changed it to tokio sleep by now. https://github.com/ZcashFoundation/zebra/pull/7989/commits/27d29364677a28bd572cd84a9261f5ef3876b8e8

zebra-scan/src/scan.rs
oxarbitrage: Applied this at https://github.com/ZcashFoundation/zebra/pull/7989/commits/27d29364677a28bd572cd84a9261f5ef3876b8e8

zebrad/src/commands/start.rs
oxarbitrage: Yea, this is what is needed. Thanks!

https://github.com/ZcashFoundation/zebra/pull/7989/commits/cd8b0fd36845e1f96112bb95b6c6455c03a6a737

zebra-scan/src/scan.rs
```diff
@@ -0,0 +1,58 @@
...
+        let request = state
+            .ready()
+            .await
+            .map_err(|e| eyre!(e))?
+            .call(zebra_state::Request::Tip)
+            .await
+            .map_err(|e| eyre!(e));
+
+        let tip = match request? {
+            zebra_state::Response::Tip(tip) => tip.unwrap().0,
```
teor2345: Let's remove the panic on an empty state for now.
```suggestion
            zebra_state::Response::Tip(tip) => tip,
```

teor2345 (approved):
Looks good, since this is blocking other work, let's merge and replace things later.

I just quickly removed the panic that happens when the state is empty.

PR #8010, opened 2023-11-27
dependabot(change): Exclude hyper from production dependabot upgrades

arya2:
## Motivation

We want to exclude hyper from future dependabot PRs until `jsonrpc-http-server` starts using v1 so that they can be merged without changes.

#### Check before marking the PR as ready for review:
  - [x] Will the PR name make sense to users?
  - [x] Does the PR have a priority label?
  - [x] Have you added or updated tests?
  - [x] Is the documentation up to date?

##### For significant changes:
  - [x] Is there a summary in the CHANGELOG?
  - [x] Can these changes be split into multiple PRs?

_If a checkbox isn't relevant to the PR, mark it as done._

## Solution

- Add `|hyper` to the exclude pattern for production dependencies.

## Review

Anyone can review.

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

## Follow Up Work

- Revert this PR when `jsonrpc-http-server` starts using hyper v1.x.x

PR #8016, opened 2023-11-28
Revert "dependabot(change): Exclude hyper from production dependabot upgrades"

teor2345:
Reverts ZcashFoundation/zebra#8010

## Motivation

We want to stop `hyper` dependabot PRs until we manually upgrade it.

`groups.exclude-patterns` from PR #8010 doesn't seem to do what we want:
> exclude-patterns | Use to exclude certain dependencies from the group. If a dependency is excluded from a group, Dependabot will continue to raise single pull requests to update the dependency to its latest version.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

For example, see https://github.com/ZcashFoundation/zebra/pull/8015.

`ignore` would be closer, but it requires another PR to undo, so let's avoid that:
> By default all dependencies that are explicitly defined in a manifest are kept up to date by Dependabot version updates. In addition, Dependabot security updates also update vulnerable dependencies that are defined in lock files. You can use allow and ignore to customize which dependencies to maintain. 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

Instead, we can just manually tell dependabot to ignore `hyper` 1.x.x:
> @dependabot ignore <dependency name> major version will close this group update PR and stop Dependabot creating any more for the specific dependency's major version (unless you unignore this specific dependency's major version or upgrade to it yourself)

And it will automatically do upgrades after our first manual `hyper` upgrade.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,8 +23,7 @@ updates:
         prod:
           dependency-type: "production"
           exclude-patterns:
-            # TODO: remove hyper from pattern after jsonrpc-http-server upgrades to v1
-            - "zcash_*|orchard|halo2*|incrementalmerkletree|bridgetree|equihash|hyper"
+            - "zcash_*|orchard|halo2*|incrementalmerkletree|bridgetree|equihash"
         dev:
           dependency-type: "development"
   # Devops section


teor2345:
> Instead, we can just manually tell dependabot to ignore `hyper` 1.x.x:
> 
> > @dependabot ignore  major version will close this group update PR and stop Dependabot creating any more for the specific dependency's major version (unless you unignore this specific dependency's major version or upgrade to it yourself)
> 
> And it will automatically do upgrades after our first manual `hyper` upgrade.

I've already done that here:
https://github.com/ZcashFoundation/zebra/pull/8015#issuecomment-1828928358

So apart from reverting the changes there's nothing else to do.

teor2345:
This has done what we want in PR https://github.com/ZcashFoundation/zebra/pull/8017

And there is no separate hyper PR.

arya2 (approved):

PR #8034, opened 2023-11-29
change(zebra-scan): Scan only one key per backend call

### LINKED ISSUE #8018: Scan keys one at a time, so we know which key(s) decrypt each transaction

oxarbitrage:
## Motivation

We want to scan only one key and one block per call to the `zcash_client_backend::scanning::scan_block` function.

The call support scanning a list of keys which it should be more efficient but we can't relate transaction with key in the results.

Close https://github.com/ZcashFoundation/zebra/issues/8018

## Solution

Replace the list argument of the zebra zcan_block function with single key element.

The `scanning_from_populated_zebra_state` test case was removed because after the change no empty key will be allowed and the account/nullifier method to scan is not supported; by that the test ends up being the same as `scanning_zecpages_from_populated_zebra_state`

### Testing

Change covered by existing `scanning_zecpages_from_populated_zebra_state` test, after minor modifications to the call argument.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -68,7 +68,7 @@ pub async fn start(mut state: State, storage: Storage) -> Result<(), Report> {
     }
 }
 
-/// Returns transactions belonging to any of the given [`ScanningKey`]s.
+/// Returns transactions belonging to the given `ScanningKey`.
 ///
 /// TODO:
 /// - Remove the `sapling_tree_size` parameter or turn it into an `Option` once we have access to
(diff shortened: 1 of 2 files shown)


teor2345 (approved):
Thanks!

teor2345:
The lightwalletd test failed with:
> {"app":"lightwalletd","bind_addr":"127.0.0.1:40611","error":"listen tcp 127.0.0.1:40611: bind: address already in use","level":"fatal","msg":"couldn't create listener","time":"2023-11-29T21:21:38Z"}

https://github.com/ZcashFoundation/zebra/actions/runs/7038410028/job/19155707096?pr=8034#step:4:305

This is one of those bugs that happens occasionally when we use the same port as another test or running CI server process.

Issue #8037, opened 2023-11-30
Run Zebra's unit tests with the scanner enabled


teor2345:
## Motivation

We're not running the shielded scanner tests in CI, leading to bugs like:
https://github.com/ZcashFoundation/zebra/pull/8036/commits/9a7848ee08c7ac8f470a9f43e63f42571b421874

### Required Changes

- [x] Fix test failures when the `shielded-scan` feature is enabled: https://github.com/ZcashFoundation/zebra/issues/8037#issuecomment-1834692864
- [x] Admin: add `shielded-scan` to the RUST_EXPERIMENTAL_FEATURES GitHub repository variable

https://github.com/ZcashFoundation/zebra/blob/1708f9d94679870d76cd3906b250756cbc4af07c/.github/workflows/ci-unit-tests-docker.yml#L133-L135

teor2345:
Let's see how it goes:
<img width="1156" alt="Screenshot 2023-12-01 at 07 57 33" src="https://github.com/ZcashFoundation/zebra/assets/8951843/16e9ad22-c39f-481e-89bf-1b562d69ece5">

teor2345:
Instructions to revert:
1. Get a repository or organisation admin
2. Go to https://github.com/ZcashFoundation/zebra/settings/variables/actions
3. Change the value of `RUST_EXPERIMENTAL_FEATURES` to ` ` (a single space)

teor2345:
It looks like this is going to take a significant number of changes:
> failures:
>     config_tests
>     downgrade_state_format
>     end_of_support_is_checked_at_start
>     ephemeral_existing_directory
>     ephemeral_missing_directory
>     misconfigured_ephemeral_existing_directory
>     misconfigured_ephemeral_missing_directory
>     non_blocking_logger
>     persistent_mode
>     restart_stop_at_height
>     start_args
>     start_no_args
>     sync_large_checkpoints_mainnet
>     sync_large_checkpoints_mempool_mainnet
>     sync_one_checkpoint_mainnet
>     update_state_format
>     zebra_rpc_conflict
>     zebra_state_conflict
>     zebra_zcash_listener_conflict

https://github.com/ZcashFoundation/zebra/actions/runs/7052897741/job/19199392237?pr=8038#step:4:38827

So I'll revert the experimental features variable for now.

teor2345:
Some of these failures could be caused by a panic in one test, let's try to fix the config test first.

mpguerra:
[...]

Issue #8022, opened 2023-11-28
feat(scanner): Start scanning where it left off last time


oxarbitrage:
## Motivation

We discussed in the Zebra team for the need of restarting a scan that was interrupted, left behind or if new keys are added to the config after a restart. Ideally the scan for a key should start around where it was left when applicable. 

### Suggested Design

A possible design might be:

- [x] Insert new keys into the database with the height before their birthday height (done in PR #8036)
- [x] Add a database method to get the last scanned height for each key in the database, by getting the last value in each key's range (based on the existing code for getting the first value in each key's range)
- [x] When scanning blocks:
  - Start from the lowest height for any key.
  - Only scan with the keys that haven't scanned that height already.

#### Testing

- [ ] Make sure cached blocks at or below the last scanned height aren't scanned
- [ ] Make sure cached blocks above the last scanned height are scanned
- [ ] Verify that all incoming blocks within `zebrad` are scanned
- [ ] Test with different heights for different keys

teor2345:
I updated this design after the changes in PR #8036

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/8022?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

teor2345:
I updated the tests based on what we skipped for other tickets. Feel free to open more tickets for those tests.

Issue #7813, opened 2023-10-24
test(scan): Scan blocks containing ZECpages transactions with the ZECpages shielded key


teor2345:
### Motivation

When we have a draft API, let's try scanning some real blocks with a real key.

ZECpages publishes their key here:
https://zecpages.com/boardinfo

### Complex Code or Requirements

If we run this in CI we will need cached state. But to start with we can just run it manually and skip the test unless a specific env var is set.

### Implementation

To verify the test worked, we might need a method that gets all the sapling results for a key in the database, in a specific height range.

### Testing

Unit test:
- [ ] scan a real block test vector that contains a zecpages post
- [ ] scan one that doesn't

Integration test:
- [ ] scan the entire blockchain for zecpages posts - this test currently takes 15 hours, so it needs to be a manual-only test, or a scheduled weekly test, or we should scan batches of 100 blocks in parallel
- [ ] check that the results match zecpages posts

### Related Work

This is a test for the blockchain scanning work.

teor2345:
I am doing part of this testing, but I think it would be good to assign multiple people to it.

teor2345:
I manually tested this over the weekend, here are some results:

The scanner works and there were no panics or hangs.

For a full scan with a synced state using `du -shc <db path>`
10M     .cache/zebra-custom/private-scan/v0/mainnet
245G    .cache/zebra-custom/state/v25/mainnet

For an incremental scan as the state is being synced:
12M     .cache/zebra-tmp/zebra-private-scan-v0-mainnet-0JYZye
262G   .cache/zebra-tmp/zebra-state-v25-mainnet-SLEwyi

The size is acceptable for a few keys, but adding the key to every row might be an issue for a user with lots of keys or shielded transactions. (This is out of scope for the MVP.)

It took about 15 hours to scan the entire chain using 1 thread and 1 key in 3 different variants (full, individual spend, and individual change). Performance improvements are out of scope for the MVP.

This is about 42 blocks per second, which is already fast enough to make some complicated changes unnecessary.

```
2023-12-01T04:48:03.819888Z  INFO {zebrad="850bd6a" net="Main"}: zebra_scan::scan: Scanning the blockchain: now at block Height(419200), current tip Some((Height(2315738), block::Hash("0000000000773b953a99b4a115b479a31fb674060160faa843a845451c4e4582")))
[...]

PR #8040, opened 2023-12-01
Scan blocks with sapling keys and write the results to the database

### LINKED ISSUE #7953: Scan verified blocks from sapling activation to the tip
## Motivation

We want to scan all the blocks in the state, this is the ticket for scanning verified blocks from sapling activation to the tip.

### Suggested Design

Count from sapling activation height to `LatestChainTip.best_tip_height()`, ask the state for each block, then scan it. If there is no block for a height, wait 10 seconds, then try that height again in a loop.

Outside the state there is no distinction between finalised and non-finalised blocks, so this design will scan all of them.

## Out of Scope

- #7905

### LINKED ISSUE #8021: Write scanning results to database
After we have a database created in https://github.com/ZcashFoundation/zebra/issues/8019 and we have results from https://github.com/ZcashFoundation/zebra/issues/8018, we need to insert them to the database. 

This will be the issue that makes the connection between the scanning results and the persistent storage. Results can be written using the `Storage::add_sapling_result()` method.

Writing results to database files can block, so it should be wrapped in a `spawn_blocking()` function.

### LINKED ISSUE #8049: Check if any of the security issues are blockers for the MVP, and document the remaining security issues

teor2345:
## Motivation

This PR scans the blockchain and adds the scanner results to the scanner database.

Close #8021

I think it might also
close #7953
but @upbqdn feel free to replace that code with your PR, and re-open that ticket if it's not finished.

The changelog summary for 1.5.0 already talks about scanning.

I tried scanning with all the keys to make sure it would work, but we only really need to scan with the full viewing key. I couldn't find the encoding for individual viewing keys, so I left that until later.

There's also a bunch of `spawn_blocking()` calls in here.

## Solution

Changes:
- Parse keys once at the start of the scan
  - Create a function that parses a key into a list of keys
  - Take a list of keys from the same key in the scanner for now
- Pass a ChainTipChange to the scanner function, this turned out to be mostly useless but it's good for progress
- Get a block from the state instead of the tip height
- Use a dummy sapling tree size
- Scan each block and add the results to storage
  - Convert a scanned block to a sapling result

Fixes:
- Fix availability of tokio::time in scanner
- Make it easier to pass keys and blocks to conversion functions
- Increase scanner wait times
- Don't log secret keys, only log every 100,000 blocks
- Move blocking tasks into spawn_blocking()

### Testing

- Update the acceptance test

I manually tested this, getting it in CI is ticket #8037

## Review

@oxarbitrage might want to review the scanner bits, and @upbqdn the block/state bits?

### Reviewer Checklist

## Follow Up Work

#8022 then lots of testing:
- #7813 
- #8037 
- Round-trip tests for the database formats

### DIFF
--- a/zebra-scan/Cargo.toml
+++ b/zebra-scan/Cargo.toml
@@ -22,6 +22,7 @@ categories = ["cryptography::cryptocurrencies"]
 
 color-eyre = "0.6.2"
 indexmap = { version = "2.0.1", features = ["serde"] }
+itertools = "0.12.0"
 semver = "1.0.20"
 serde = { version = "1.0.193", features = ["serde_derive"] }
 tokio = { version = "1.34.0", features = ["time"] }
--- a/zebra-scan/Cargo.toml
+++ b/zebra-scan/Cargo.toml
@@ -24,7 +24,7 @@ color-eyre = "0.6.2"
 indexmap = { version = "2.0.1", features = ["serde"] }
 semver = "1.0.20"
 serde = { version = "1.0.193", features = ["serde_derive"] }
-tokio = "1.34.0"
+tokio = { version = "1.34.0", features = ["time"] }
 tower = "0.4.13"
 tracing = "0.1.39"
 
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -84,17 +84,23 @@ pub async fn start(
     tokio::time::sleep(INITIAL_WAIT).await;
 
     loop {
-        // Make sure we can query the state
-        let request = state
+        // Get a block from the state.
+        // We can't use ServiceExt::oneshot() here, because it causes lifetime errors in init().
+        let block = state
             .ready()
             .await
             .map_err(|e| eyre!(e))?
-            .call(zebra_state::Request::Tip)
+            .call(zebra_state::Request::Block(height.into()))
             .await
-            .map_err(|e| eyre!(e));
+            .map_err(|e| eyre!(e))?;
 
-        let tip = match request? {
-            zebra_state::Response::Tip(tip) => tip,
+        let block = match block {
+            zebra_state::Response::Block(Some(block)) => block,
+            zebra_state::Response::Block(None) => {
+                // If we've reached the tip, sleep for a while then try and get the same block.
+                tokio::time::sleep(CHECK_INTERVAL).await;
+                continue;
+            }
             _ => unreachable!("unmatched response to a state::Tip request"),
         };
 
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -104,11 +104,18 @@ pub async fn start(
             _ => unreachable!("unmatched response to a state::Tip request"),
         };
 
+        // Only log at info level every 100,000 blocks
+        let is_info_log =
+            height == storage.min_sapling_birthday_height() || height.0 % INFO_LOG_INTERVAL == 0;
 
-        for key in available_keys {
+        // TODO: add debug logs?
+        if is_info_log {
             info!(
-                "Scanning the blockchain for key {} from block {:?} to {:?}",
-                key.0, key.1, tip,
+                "Scanning the blockchain: now at block {:?}, current tip {:?}",
+                height,
+                chain_tip_change
+                    .latest_chain_tip()
+                    .best_tip_height_and_hash(),
             );
         }
 
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -119,7 +119,39 @@ pub async fn start(
             );
         }
 
-        tokio::time::sleep(CHECK_INTERVAL).await;
+        for (key_num, (sapling_key, birthday_height)) in key_birthdays.iter().enumerate() {
+            // # Security
+            //
+            // We can't log `sapling_key` here because it is a private viewing key. Anyone who reads
+            // the logs could use the key to view those transactions.
+            if is_info_log {
+                info!(
+                    "Scanning the blockchain for key {}, started at block {:?}",
+                    key_num, birthday_height,
+                );
+            }
+
+            // Scan the block, which blocks async execution until the scan is complete.
+            //
+            // TODO: skip scanning before birthday height (#8022)
+            // TODO: scan each key in parallel (after MVP?)
+            let (dfvks, ivks) = parsed_keys.get(sapling_key).cloned().unwrap_or_default();
+
+            let dfvk_results =
+                scan_block(storage.network(), &block, 0, &dfvks).map_err(|e| eyre!(e))?;
+            let ivk_results =
+                scan_block(storage.network(), &block, 0, &ivks).map_err(|e| eyre!(e))?;
+
+            let dfvk_results = scanned_block_to_db_result(dfvk_results);
+            let ivk_results = scanned_block_to_db_result(ivk_results);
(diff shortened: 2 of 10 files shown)


teor2345:
Seems to work in my manual tests, and quickly too:
> 2023-12-01T04:57:54.983238Z  INFO {zebrad="850bd6a" net="Main"}: zebra_scan::scan: Scanning the blockchain: now at block Height(600000), current tip Some((Height(2315746), block::Hash("00000000017abab20936b148ac1ed54a38cb87019f7ac489baeab2db256ea59e")))
> 2023-12-01T04:57:54.983251Z  INFO {zebrad="850bd6a" net="Main"}: zebra_scan::scan: Scanning the blockchain for key 0, started at block Height(419200)

This is only running on one core, and doing each key twice (dfvk and ivks).

zebra-scan/src/scan.rs
```diff
@@ -1,81 +1,184 @@
...
+                    .best_tip_height_and_hash(),
             );
         }
 
-        tokio::time::sleep(CHECK_INTERVAL).await;
+        for (key_num, (sapling_key, birthday_height)) in key_birthdays.iter().enumerate() {
+            // # Security
+            //
+            // We can't log `sapling_key` here because it is a private viewing key. Anyone who reads
+            // the logs could use the key to view those transactions.
```
oxarbitrage, line 127: Thanks for the fix. I initially added the pk to the log.

zebra-scan/src/scan.rs
```diff
@@ -88,9 +191,9 @@ pub async fn start(mut state: State, storage: Storage) -> Result<(), Report> {
 /// - Add prior block metadata once we have access to Zebra's state.
 pub fn scan_block<K: ScanningKey>(
     network: Network,
-    block: Arc<Block>,
+    block: &Arc<Block>,
     sapling_tree_size: u32,
-    scanning_key: &K,
+    scanning_keys: &[K],
```
oxarbitrage, line 196: I don't understand why it was better to reverse this change.

zebrad/tests/acceptance.rs
```diff
@@ -2805,8 +2805,8 @@ async fn fully_synced_rpc_z_getsubtreesbyindex_snapshot_test() -> Result<()> {
 }
 
 /// Test that the scanner task gets started when the node starts.
-#[cfg(feature = "shielded-scan")]
 #[test]
+#[cfg(feature = "shielded-scan")]
```
oxarbitrage, line 2809: Some tests do this before the `[test]` macro (ex: https://github.com/ZcashFoundation/zebra/blob/main/zebrad/tests/acceptance.rs#L1305) and others do this after. What is the best way and why? Should we normalize the codebase? We should open a separated cleanup ticket for this if so.

zebrad/tests/acceptance.rs
```diff
@@ -2832,13 +2832,16 @@ fn scan_task_starts() -> Result<()> {
     let output = child.wait_with_output()?;
 
     output.stdout_line_contains("spawning shielded scanner with configured viewing keys")?;
+    /*
+    TODO: these lines only happen when we reach sapling activation height
```
oxarbitrage, line 2836: optional:

I think we should print a message as soon as we know we are behind the sapling activation block saying something as:

"scan task will not get started until the blockchain is at sapling activation height (block number: {sapling activation_height})"

This way users will know the scanner will do nothing until that.

We can look for that message in this test so it will not require further modifications. A  new test will be done with state.

zebra-scan/Cargo.toml
```diff
@@ -22,9 +22,10 @@ categories = ["cryptography::cryptocurrencies"]
 
 color-eyre = "0.6.2"
 indexmap = { version = "2.0.1", features = ["serde"] }
+itertools = "0.12.0"
 semver = "1.0.20"
 serde = { version = "1.0.193", features = ["serde_derive"] }
-tokio = "1.34.0"
+tokio = { version = "1.34.0", features = ["time"] }
```
oxarbitrage, line 28: Is this change needed?

zebra-scan/src/scan.rs
```diff
@@ -1,81 +1,184 @@
...
+            let mut storage = storage.clone();
+
+            // We use a dummy size of the Sapling note commitment tree.
+            //
+            // We can't set the size to zero, because the underlying scanning function would return
+            // `zcash_client_backeng::scanning::ScanError::TreeSizeUnknown`.
+            //
+            // And we can't set them close to 0, because the scanner subtracts the number of notes
+            // in the block, and panics with "attempt to subtract with overflow". The number of
+            // notes in a block must be less than this value, this is a consensus rule.
```
oxarbitrage, line 153: Optional:

Quote the consensus rule. Use `# Consensus` doc tag as we use everywhere else.

zebra-scan/src/scan.rs
```diff
@@ -1,81 +1,184 @@
...
+            let sapling_tree_size = 1 << 16;
+
+            tokio::task::spawn_blocking(move || {
+                let dfvk_res =
+                    scan_block(network, &block, sapling_tree_size, &dfvks).map_err(|e| eyre!(e))?;
+                let ivk_res =
+                    scan_block(network, &block, sapling_tree_size, &ivks).map_err(|e| eyre!(e))?;
```
oxarbitrage, line 160: Ok, i see this is for testing purposes so it should be fine as it for now.

oxarbitrage (approved):
utACK. Looks really good to me, i left a few questions and optionals.

zebra-scan/src/scan.rs
teor2345: If we have a diversifiable full viewing key, there will only be one key. But if we have an individual viewing key, there will be two keys for the same account or transaction, the payment key and the change key (which are called external and internal).

So when I was testing using individual viewing keys, I needed to pass two of them into the one scan.

zebrad/tests/acceptance.rs
teor2345: It doesn't matter, it works the same way in either order, I just found it easier to read.

zebrad/tests/acceptance.rs
teor2345: Great, let's do this in a separate ticket

teor2345: This is now #8041

zebra-scan/Cargo.toml
teor2345: Yes, if the crate is built individually with no features, it needs access to the tokio time APIs.

Sometimes this works anyway in CI, I'm not sure why. Possibly because we're building all targets.

zebra-scan/src/scan.rs
teor2345: Since this is a dummy value, I expect we'll replace it later with the actual value from the state. So I don't think it will go to the audit like this.

zebra-scan/Cargo.toml
teor2345: This is now tested by #8042, which might fail the first time due to this bug in main

zebra-scan/src/scan.rs
teor2345: Added TODO in 247684c492 in PR #8045

zebra-scan/src/scan.rs
teor2345: Documented in 329c59dd1a in PR #8045

PR #8045, opened 2023-12-04
change(scan): Create a function that scans one block by height, and stores the results in the database

teor2345:
## Motivation

This is a quick follow up to a bunch of recent scanner changes.

@upbqdn or @oxarbitrage this might conflict with what you're working on, but I'm happy to re-do it, because your PRs are actually changing things.

## Solution

- Create a function that scans one block by height, gets it from the state, and puts the results in the database
- Add some TODOs and update some function comments

Most of these arguments are cheap to clone, but if we have a lot of keys the key sets could be very large, so I used `Arc`.

### Testing

This is just a cleanup of the previous PR.

## Review

This is a routine cleanup.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -22,7 +22,7 @@ use zcash_primitives::{
 };
 
 use zebra_chain::{
-    block::Block,
+    block::{Block, Height},
     chain_tip::ChainTip,
     diagnostic::task::WaitForPanics,
     parameters::Network,
@@ -55,7 +55,7 @@ const INFO_LOG_INTERVAL: u32 = 100_000;
 /// Start a scan task that reads blocks from `state`, scans them with the configured keys in
 /// `storage`, and then writes the results to `storage`.
 pub async fn start(
-    mut state: State,
+    state: State,
     chain_tip_change: ChainTipChange,
     storage: Storage,
 ) -> Result<(), Report> {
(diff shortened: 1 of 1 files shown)


zebra-scan/src/scan.rs
```diff
@@ -80,115 +81,159 @@ pub async fn start(
...
+            .expect("a valid blockchain never reaches the max height");
+    }
+}
+
+/// Get the block at `height` from `state`, scan it with the keys in `parsed_keys`, and store the
+/// results in `storage`. If `height` is lower than the `key_birthdays` for that key, skip it.
+///
+/// Returns:
+/// - `Ok(Some(height))` if the height was scanned,
+/// - `Ok(None)` if the height was not in the database, and
```
upbqdn:
```suggestion
/// - `Ok(None)` if the height was not in the state, and
```

upbqdn (dismissed):

zebra-scan/src/scan.rs
teor2345: Fair point, since we have multiple databases now.

upbqdn (dismissed):

teor2345:
Just an import merge conflict, should be fixed now.

zebra-scan/src/scan.rs
```diff
@@ -22,8 +22,12 @@ use zcash_primitives::{
...
 
 use zebra_chain::{
-    block::Block, chain_tip::ChainTip, diagnostic::task::WaitForPanics, parameters::Network,
-    serialization::ZcashSerialize, transaction::Transaction,
+    block::{Block, Height},
+    chain_tip::ChainTip,
+    diagnostic::task::WaitForPanics,
+    parameters::Network,
+    serialization::ZcashSerialize,
+    transaction::{self, Transaction},
```
teor2345:
```suggestion
    transaction::Transaction,
```

upbqdn (approved):

PR #8059, opened 2023-12-05
fix(scan): Start scanning task only if there are keys to scan

### LINKED ISSUE #7892: change(release): Enable all user features in the `experimental` build

oxarbitrage:
## Motivation

Part of https://github.com/ZcashFoundation/zebra/issues/7892

When we run the Zebra test suite some tests will fail with database already in use. It seems a bug in shutdown. We want to run the scan tests and config tests so we can lock that functionality and keep making progress.

## Solution

- Start scan only if there are keys.
- Fix config test for new config file that is generated when `shielded-scan` feature is enabled.
- Add ephemeral database in tests by default.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Re enable feature in CI: https://github.com/ZcashFoundation/zebra/issues/7892
- Create a new ticket top test scan database conflict. Deal with the bug there.

### DIFF
--- a/zebrad/src/commands/start.rs
+++ b/zebrad/src/commands/start.rs
@@ -289,10 +289,12 @@ impl StartCmd {
         let syncer_task_handle = tokio::spawn(syncer.sync().in_current_span());
 
         #[cfg(feature = "shielded-scan")]
-        // Spawn never ending scan task.
-        let scan_task_handle = {
+        // Spawn never ending scan task only if we have keys to scan for.
+        let scan_task_handle = if !config.shielded_scan.sapling_keys_to_scan.is_empty() {
             info!("spawning shielded scanner with configured viewing keys");
             zebra_scan::spawn_init(&config.shielded_scan, config.network.network, state)
+        } else {
+            tokio::spawn(std::future::pending().in_current_span())
         };
 
         #[cfg(not(feature = "shielded-scan"))]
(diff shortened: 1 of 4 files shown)


teor2345 (approved):
Thanks for these fixes!

Issue #8046, opened 2023-12-04
Document how to run and query the results of the blockchain scanning MVP


mpguerra:
## Motivation

We want a "naive" zebra user to be able to use `zebra-scan` to query the Zcash blockchain with a viewing key and get a list of the transactions related with it

## Scope

The documentation should provide details of:
- how to compile and run zebra with the `zebra-scan` feature enabled
- how to specify the viewing key(s), e.g location and format for the config file containing the keys
- how to query the database to retrieve data about the transactions found

It would be nice if we could also post the same instructions to the blog and include screenshots and examples, e.g using the ZECPages viewing key.

The incomplete draft document is here:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/user/shielded-scan.md

teor2345:
Before we do the instructions, let's decide if #8048 is a blocker for the MVP?

If we put secret keys in the existing config file, we might need to change its permissions, but that is a breaking change if someone has deliberately made their file group or world (all users) readable.

teor2345:
So far I think there are two ways to read the file:
- using the read results method in the Rust API
    - this code needs to run inside Zebra, or while Zebra is shut down
    - if we want it to run while Zebra is running, we can add a zebra-state argument that opens a secondary database
    - we could write a zebrad command that takes a key (or a key index for security), and prints the heights and transaction IDs

- using `ldb` to dump the database
    - open the database in secondary read-only mode using command-line arguments
    - use --from and --to arguments to get the entire range of results for the key, by changing the height to ffffff
    - text processing to split up the stored data into individual 64-hex digit hashes
    - this is a less usable method

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/zebra/8046?planning-poker) with Zenhub @arya2 @oxarbitrage @teor2345 @upbqdn

teor2345:
In the Zebra sprint planning we decided on `ldb` and #8050.

teor2345:
Here's some experiments with the zecpages key:

Showing all the transactions:
```sh
ldb --try_load_options --db="$HOME/.cache/zebra-custom/private-scan/v1/mainnet" --secondary_path= --column_family=sapling_tx_ids --hex scan
...
[...]

PR #8062, opened 2023-12-06
change(scan): Store one transaction ID per database row, to make queries easier

### LINKED ISSUE #8050: Change scanner database format so it is easier for users to query for transaction IDs
## Motivation

We want technical users to be able to query the scanner database more easily.

This means:
- [x] each transaction ID should be in its own database row/entry
- [x] #8056

teor2345:
## Motivation

We want technical users to be able to use `ldb` to query the scanner database. This means we need one transaction ID per row.

Close #8050.

There is already a summary for this project in the changelog.

This changes the format from:
`ScannerKey | Height -> Vec<transaction::Hash>` to:
`ScannerKey | Height | TransactionIndex -> Option<transaction::Hash>`.

`None` can be used to represent birthday heights and scanner progress in #8022.

This format isn't ideal, because handing `None` is harder to code and review. We can change it to use a custom enum later, or track keys, birthdays, and progress in a separate column family. But that's for after the MVP.

## Solution

- Change the scanner database format (see above)
- Change the serialization formats
- Update the major version to 1
- Update format docs

### Testing

I updated the existing tests where needed, and they pass.

## Review

This is blocking finishing off the scanner, so it seems like a high priority for review.

### Reviewer Checklist

## Follow Up Work

- #8022
- #8046
- re-do #7813

### DIFF
--- a/zebra-scan/src/storage/db/sapling.rs
+++ b/zebra-scan/src/storage/db/sapling.rs
@@ -4,16 +4,19 @@
 //!
 //! | name             | key                           | value                    |
 //! |------------------|-------------------------------|--------------------------|
-//! | `sapling_tx_ids` | `SaplingScannedDatabaseIndex` | `transaction::Hash`      |
+//! | `sapling_tx_ids` | `SaplingScannedDatabaseIndex` | `Option<SaplingScannedResult>`      |
 //!
 //! And types:
+//! `SaplingScannedResult`: same as `transaction::Hash`, but with bytes in display order.
+//! `None` is stored as a zero-length array of bytes.
+//!
 //! `SaplingScannedDatabaseIndex` = `SaplingScanningKey` | `TransactionLocation`
 //! `TransactionLocation` = `Height` | `TransactionIndex`
 //!
 //! This format allows us to efficiently find all the results for each key, and the latest height
 //! for each key.
 //!
-//! If there are no results for a height, we store an empty list of results for the coinbase
+//! If there are no results for a height, we store `None` as the result for the coinbase
 //! transaction. This allows is to scan each key from the next height after we restart. We also use
 //! this mechanism to store key birthday heights, by storing the height before the birthday as the
 //! "last scanned" block.
--- a/zebra-scan/src/storage/db/sapling.rs
+++ b/zebra-scan/src/storage/db/sapling.rs
@@ -18,10 +18,13 @@
 //! this mechanism to store key birthday heights, by storing the height before the birthday as the
 //! "last scanned" block.
 
-use std::collections::{BTreeMap, HashMap};
+use std::{
+    collections::{BTreeMap, HashMap},
+    ops::RangeBounds,
+};
 
 use itertools::Itertools;
-use zebra_chain::block::Height;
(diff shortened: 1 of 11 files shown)


zebra-scan/src/storage/db/sapling.rs
```diff
@@ -4,24 +4,31 @@
 //!
 //! | name             | key                           | value                    |
 //! |------------------|-------------------------------|--------------------------|
-//! | `sapling_tx_ids` | `SaplingScannedDatabaseIndex` | `Vec<transaction::Hash>` |
+//! | `sapling_tx_ids` | `SaplingScannedDatabaseIndex` | `transaction::Hash`      |
 //!
 //! And types:
```
teor2345:
```suggestion
//! | `sapling_tx_ids` | `SaplingScannedDatabaseIndex` | `Option<SaplingScannedResult>`      |
//!
//! And types:
//! `SaplingScannedResult`: same as `transaction::Hash`, but with bytes in display order.
//! `None` is stored as a zero-length array of bytes.
//!
```

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -4,24 +4,31 @@
...
+//! `SaplingScannedDatabaseIndex` = `SaplingScanningKey` | `TransactionLocation`
+//! `TransactionLocation` = `Height` | `TransactionIndex`
 //!
 //! This format allows us to efficiently find all the results for each key, and the latest height
 //! for each key.
 //!
-//! If there are no results for a height, we store an empty list of results. This allows is to scan
-//! each key from the next height after we restart. We also use this mechanism to store key
-//! birthday heights, by storing the height before the birthday as the "last scanned" block.
+//! If there are no results for a height, we store an empty list of results for the coinbase
```
teor2345:
```suggestion
//! If there are no results for a height, we store `None` as the result for the coinbase
```

zebra-state/src/service/finalized_state/disk_format/block.rs
```diff
@@ -103,11 +103,21 @@ impl TransactionIndex {
...
     pub fn as_u64(&self) -> u64 {
         self.0.into()
     }
+
+    /// The minimum value of a transaction index.
+    ///
+    /// This value corresponds to the coinbase transaction.
+    pub const MIN: Self = Self(u16::MIN);
+
+    /// The maximun value of a transaction index.
```
teor2345:
```suggestion
    /// The maximum value of a transaction index.
```

zebra-state/src/service/finalized_state/disk_format/block.rs
```diff
@@ -150,6 +165,42 @@ impl TransactionLocation {
...
+
+    /// The minimum value of a transaction location.
+    ///
+    /// This value corresponds to the genesis coinbase transaction.
+    pub const MIN: Self = Self {
+        height: Height::MIN,
+        index: TransactionIndex::MIN,
+    };
+
+    /// The maximun value of a transaction location.
```
teor2345:
```suggestion
    /// The maximum value of a transaction location.
```

oxarbitrage (approved):

Issue #8019, opened 2023-11-28
Implement scanning results storage format


teor2345:
## Motivation

To store results in the database, we need to serialize them and put them in a RocksDB column family.

## Suggested Design

- [x] Create a new `ZebraDb` in `zebra-scan`.

### Column Family

- [x] Create a new column family
- [x] Write its format:

name | key | value
------|-----|------
`sapling_tx_ids` | `SaplingScanningKey` + `Height` | `Vec<transaction::Hash>`

This format makes it easy to:
- [x] get the results for each key using the whole key range
- [ ] get the last scanned height that each key actually had results using the key range and reading the last value
- [x] get the result for a specific key and block
- [ ] get all the results in the database

And also to mark scanner progress:
- [x] If there are no scanning results, insert an empty list of transaction IDs:
       `SaplingScanningKey` + `Height` -> _empty list of transaction IDs_

This lets us find the most recently scanned height for each key, even if it didn't have any transactions in that block.

### Serialization

These serializations should be behind the scanner feature, because they could be risky for production use:
- [x] `SaplingScanningKey` is currently a string, so we can serialize it using `as_bytes()`. This will be replaced later, it's more efficient and correct to serialize parsed binary types.
- [x] `Vec` can be serialized by serializing each item and combining them together. This only works if the items are fixed size, and it can also cause performance issues for large numbers of items.

### Documentation

- [x] The scanner database should be documented in its Rust module, because it's easier to keep the docs in sync.

## Out of Scope

We can use a binary serialization later when we create an enum for the key types.

We can use `TransactionLocation` rather than `Height`, and look up the transaction ID in the state. But the `Height` is simpler to start with.

We can increment the database version whenever we change the format. We have the code for deleting previous versions, but we need to launch the task.

PR #7628, opened 2023-09-26
fix(test): `roundtrip_sapling_tree_root` and `roundtrip_orchard_tree_root`

oxarbitrage:
## Motivation

The round trip tree root proptests almost always test the same end height:

```
running 1 test
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
...
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 3342791
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:467] &val.end.0 = 16777215
```
 
Close https://github.com/ZcashFoundation/zebra/issues/7443

## Solution

Add suggested fix, when applied the tests will cover much more values:

```
running 1 test
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 13521308
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 13932221
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 15280320
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 5739578
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 4218397
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 10812062
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 5608137
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 12093244
[zebra-state/src/service/finalized_state/disk_format/tests/prop.rs:466] &val.end.0 = 1981967
[...]

Issue #7443, opened 2023-09-01
Fix proptest coverage of note commitment subtree serialization


teor2345:
## Motivation

Almost all the subtree proptest cases are checking an `end` of `MAX_ON_DISK_HEIGHT`.

### Specifications

This issue is documented here, but for a string regex, rather than a clamp:
> In general, the ideal solution is to find a way to take a seed value and then use prop_map to transform it into the desired, irregular domain.
> This kind of works, but it has problems ... [it] also results in a pretty odd distribution of values. 

https://proptest-rs.github.io/proptest/proptest/tutorial/filtering.html
https://proptest-rs.github.io/proptest/proptest/tutorial/transforming-strategies.html

## Suggested Solution

Instead, transform the input range so its distribution is roughly uniform:
```suggestion
        val.end.0 %= (MAX_ON_DISK_HEIGHT.0 + 1);
```

This fix is required for both Sapling and Orchard proptests.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/7350#discussion_r1309512808_

mpguerra:
@oxarbitrage can you please add an estimate to this one?

oxarbitrage:
Estimate added.

PR #8071, opened 2023-12-06
scan(test): Implement scanner format round-trip tests

### LINKED ISSUE #8050: Change scanner database format so it is easier for users to query for transaction IDs

teor2345:
## Motivation

We skipped these tests in tickets #8019 and #8050. Now the format has settled it's a good time to do them.

### Test APIs

Proptests generate random typed data, then run it through a test closure or function. In this case, our `assert_value_properties()` makes sure that the round-trip format is correct.

We can't use `clamp()` with proptests because it sets the value to the end of the range, but we want it distributed over the range.

## Solution / Testing

- Add round-trip format tests for all the scanner formats
- Add `Arbitrary` implementations for all the scanner formats

Related changes:
- Derive `Default` in test-only code to make testing easier
- Clean up an existing round-trip test

Related fixes:
- Increase proptest coverage of height serialization

(The fix for ticket #7443 in PR #7628 was incomplete.)

## Review

This is a lower priority routine testing change.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/block/height.rs
+++ b/zebra-chain/src/block/height.rs
@@ -23,6 +23,7 @@ pub mod json_conversion;
 /// There are multiple formats for serializing a height, so we don't implement
 /// `ZcashSerialize` or `ZcashDeserialize` for `Height`.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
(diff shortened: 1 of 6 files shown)


teor2345:
This is failing because I need to limit the heights to the database format range.

oxarbitrage (approved):
Thanks!

PR #8072, opened 2023-12-06
change(log): Silence verbose failed connection logs

teor2345:
## Motivation

1. Having lots of failed connection logs makes Zebra less usable, but when there are only a few connections users need to know why
2. It's also causing GitHub to be very slow showing logs in tests

## Solution

- In production, log failed connections at info level at startup, and when there are less than 5 connections
- In tests, always log failed connections at debug level (except maybe for startup)

## Review

This is a low priority logging fix.

### Reviewer Checklist

### DIFF
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -863,8 +863,9 @@ where
 
                 // Increment the connection count before we spawn the connection.
                 let outbound_connection_tracker = active_outbound_connections.track_connection();
+                let outbound_connections = active_outbound_connections.update_count();
                 debug!(
-                    outbound_connections = ?active_outbound_connections.update_count(),
+                    ?outbound_connections,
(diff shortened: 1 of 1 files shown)


zebra-network/src/peer_set/initialize.rs
```diff
@@ -1048,6 +1052,9 @@ where
         + 'static,
     C::Future: Send + 'static,
 {
+    // The maximum number of open or pending connections before we log an info-level message.
```
arya2:
```suggestion
    // The maximum number of open or pending outbound connections Zebra can have while logging an info-level error message.
```

arya2 (dismissed):
Looks good!

zebra-network/src/peer_set/initialize.rs
teor2345: I did this differently in commit 7a51714

arya2 (approved):

PR #8075, opened 2023-12-07
test(scan): Add raw database format snapshots to the scanner

### LINKED ISSUE #8069: test: Add database snapshot tests for the scanner database
### Motivation

Add these snapshot tests to the scanner:
- [x] snapshot tests for the raw database format
- [x] snapshot tests for the typed database format
  - [x] after PR #8080 merges, add scanning progress markers to both snapshot tests

These tests already exist for the state, we just need to copy and paste and change the config to the scanner.

### Testing

These are new tests.

### Related Work

Split off #8037

teor2345:
## Motivation

These are part of the standard database tests in #8069.

Not really, it's just the same as the state.

## Solution

- Add snapshot tests for the raw scanner database format 
- Expose database test support code in zebra-state
- Expose test support code in zebra-scan

Related changes:
- Make scanner storage methods clearer and easier to use
- Move scanner tests to a submodule

### Testing

These are extra tests.

I have manually checked the data formats in the snapshots and they appear correct.

## Review

Because of the code movement, the state db test APIs, and the storage API changes, this PR blocks other tests. So it would be good to get it merged soon. It doesn't need to be in the release.

### Reviewer Checklist

## Follow Up Work

The rest of the tests in #8069


oxarbitrage:
codespell failed, restarting: https://github.com/ZcashFoundation/zebra/actions/runs/7141093973/job/19447724265?pr=8075

teor2345:
Thanks for fixing up the imports, that was a bit of a mistake from me!

zebrad/Cargo.toml
```diff
@@ -118,6 +118,7 @@ proptest-impl = [
     "zebra-state/proptest-impl",
     "zebra-network/proptest-impl",
     "zebra-chain/proptest-impl",
+    "zebra-scan?/proptest-impl",
```
oxarbitrage, line 121:
```suggestion
    "zebra-scan/proptest-impl",
```

typo here?

oxarbitrage (approved):
looks good to me.

zebrad/Cargo.toml
teor2345: This means "only enable the given feature if something else enables the optional dependency":
https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

That's what we want here, it means that `zebra-scan` is only enabled if we turn on its feature.

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -47,13 +47,15 @@ impl Storage {
...
+    /// Returns `None` if the result is missing or an empty marker for a birthday or progress
+    /// height.
     //
     // TODO: add tests for this method
     pub fn sapling_result_for_index(
         &self,
         index: &SaplingScannedDatabaseIndex,
     ) -> Option<SaplingScannedResult> {
-        self.db.zs_get(&self.sapling_tx_ids_cf(), &index)
+        self.db.zs_get(&self.sapling_tx_ids_cf(), &index).flatten()
```
arya2, line 58: Optional: Could we return the nested option and change `SaplingScannedResult([u8; 32])` to `SaplingScannedResult(Option<[u8; 32]>)` to keep information about the empty result inside the new type?

arya2 (approved):
Looks good!

zebra-scan/src/storage/db/sapling.rs
teor2345: We could, but I think it would be much easier to keep information about keys/birthdays/progress and results in separate column families. (After the MVP.)

If that happens then there will be no empty results, and the signature of this method won't change.

Nothing uses this method at the moment except the snapshot tests.

PR #8076, opened 2023-12-08
fix(rpc): Make the `verbose` argument of the `getrawtransaction` RPC optional

upbqdn:
## Motivation

The `verbose` argument of the `getrawtransaction` RPC is specified as optional in the documentation: https://zcash.github.io/rpc/getrawtransaction.html. It's also specified as optional in Zebra's docs: https://github.com/ZcashFoundation/zebra/blob/5dd33d7265b9a8afe7f6b84264cc4faa2a5bfada/zebra-rpc/src/methods.rs#L215, but not implemented as such.

### Specifications

## Solution

- Use `Option<u8>` instead of `u8` for the `verbose` argument.

### Testing

- I adjusted existing tests and didn't add any new tests.
- I manually tested that the argument is optional with this PR and that its value defaults to 0.
- I also manually tested that Zebra required this argument without this PR, which didn't comply with the spec.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -226,7 +226,7 @@ pub trait Rpc {
     fn get_raw_transaction(
         &self,
         txid_hex: String,
-        verbose: u8,
+        verbose: Option<u8>,
     ) -> BoxFuture<Result<GetRawTransaction>>;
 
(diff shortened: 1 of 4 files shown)


teor2345 (approved):
Thanks!

PR #8077, opened 2023-12-08
change(scan): Impl `From<SaplingScannedResult>` for `[u8; 32]`

upbqdn:
## Motivation

This is a trivial PR that adds `From<SaplingScannedResult>` for `[u8; 32]`.

## Solution

- Impl `From<SaplingScannedResult>` for `[u8; 32]`.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/scan.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/scan.rs
@@ -43,6 +43,12 @@ impl From<&[u8; 32]> for SaplingScannedResult {
     }
 }
 
+impl From<SaplingScannedResult> for [u8; 32] {
+    fn from(scanned_result: SaplingScannedResult) -> Self {
+        scanned_result.0
+    }
+}
+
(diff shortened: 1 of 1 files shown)


zebra-state/src/service/finalized_state/disk_format/scan.rs
```diff
@@ -43,6 +43,12 @@ impl From<&[u8; 32]> for SaplingScannedResult {
     }
 }
 
+impl From<SaplingScannedResult> for [u8; 32] {
```
teor2345, line 46: This is going to conflict with:
https://github.com/ZcashFoundation/zebra/pull/8075/commits/ea3ca722b2cf71261b2c735211e2f8242fe4d076

Since there's two different ways to turn a [u8; 32] into a transaction ID, I think it would be better to use explicit method names here. (Which is what that PR does.)

upbqdn: Makes sense. Thanks for pointing that out. I'm closing this PR without merging.

PR #8079, opened 2023-12-08
change(state): Allow opening the database in a read-only mode

### LINKED ISSUE #8078: feature: Consider allowing opening the database in "read-only" or "secondary" mode
### Motivation

Zebra currently uses the [open_cf_descriptors](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors) function [here](https://github.com/ZcashFoundation/zebra/blob/5dd33d7265b9a8afe7f6b84264cc4faa2a5bfada/zebra-state/src/service/finalized_state/disk_db.rs#L647) to open the database. 

I was using the `zebra-scan` crate in a separate binary to access the scanning results while running Zebra. The crate tried to open the same database as Zebra in read-write mode, and since Zebra was running, I couldn't access the database.

RocksDB [supports](https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances) opening the database in "read-only" or "secondary" mode, which would allow separate binaries that use, for example, the `zebra_scan` crate, to read the scanning results while Zebra is running.

### Specifications

RocksDB's "read-only" and "secondary" modes: https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances

Rust bindings:

- [open_cf_descriptors_as_secondary](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_as_secondary)
- [open_cf_descriptors_read_only](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_read_only)

upbqdn:
## Motivation

Close #8078.

Depends-On: #8080

### Specifications

RocksDB's "read-only" and "secondary" modes: https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances

Rust bindings:

- [open_cf_descriptors_read_only](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_read_only)

## Solution

- Add a read-only option when opening the database.

### Notes

- Writing functionality is still exposed even though the database can be opened in read-only mode, so I opened #8098.
- The "secondary" mode seems like a better alternative because the "read-only" mode seems only to open a snapshot of the database, but I didn't manage to figure out what the `secondary_path` in [open_cf_descriptors_as_secondary](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_as_secondary) should be.

### Testing

- I manually accessed the database using the `zebra-scan` crate in a separate binary while Zebra was running.

### Reviewer Checklist


zebra-scan/src/storage/db.rs
```diff
@@ -46,11 +46,11 @@ impl Storage {
     /// If there is no existing database, creates a new database on disk.
     ///
     /// New keys in `config` are not inserted into the database.
-    pub(crate) fn new_db(config: &Config, network: Network) -> Self {
+    pub(crate) fn new_db(config: &Config, network: Network, read_only: bool) -> Self {
         Self::new_with_debug(
             config, network,
             // TODO: make format upgrades work with any database, then change this to `false`
```
teor2345: This TODO is now ambiguous

```suggestion
            // TODO: make format upgrades work with any database, then change debug_skip_format_upgrades to `false`
```

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -96,6 +96,7 @@ impl ZebraDb {
         network: Network,
         debug_skip_format_upgrades: bool,
         column_families_in_code: impl IntoIterator<Item = String>,
+        read_only: bool,
```
teor2345, line 99: Format upgrades try to write to the database, so the value of debug_skip_format_upgrades in the struct must be true when read_only is true.

teor2345:
> * Writing functionality is still exposed even though the database can be opened in read-only mode.

We can open another ticket to clean that up. It's not really a problem because trying to write will fail. (And either panic or return an error, depending on how we handle RocksDB write errors.)

One possible solution is having a type that implements all the read methods, but only implements the two write methods: `write(batch)` and `spawn_format_change()` when it is a read-write database. This can be implemented using a generic parameter that allows writing.

For example, the second generic parameter here is required to be DBWithThreadModeInner, we could do something similar with a ReadWriteDatabase unit struct:
https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#impl-DBCommon%3CT,+DBWithThreadModeInner%3E

And here are the methods that work regardless of the generic type (for us that would be read methods):
https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#impl-DBCommon%3CT,+D%3E

> * The "secondary" mode seems like a better alternative because the "read-only" mode seems only to open a snapshot of the database, but I didn't manage to figure out what the `secondary_path` in [open_cf_descriptors_as_secondary](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_as_secondary) should be.

I don't think the difference is that important, but a random temporary directory would avoid locking conflicts with multiple secondaries:
https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances

When using ldb I've just been giving an empty string, which I assume uses the current directory, or a temporary directory, or no directory. It hasn't caused any errors or created any files yet.

teor2345 (commented):
Thanks! This is some great work!

I'm not sure if it's out of scope for the MVP, happy to let the team make a call on that.

zebra-state/src/service/finalized_state/zebra_db.rs
upbqdn: Done in [06904cb](https://github.com/ZcashFoundation/zebra/pull/8079/commits/06904cbbc4c283c7d771099e8adba47806414228).

upbqdn:
I'm putting this PR out of the draft since it looks like we can merge it.

teor2345 (dismissed):
Thanks! I've scheduled this after the scanner tests merge, because they are blocking other work.

teor2345 (dismissed):

zebra-scan/src/tests/vectors.rs
```diff
@@ -157,7 +157,7 @@ fn scanning_fake_blocks_store_key_and_results() -> Result<()> {
         zcash_client_backend::encoding::encode_extended_full_viewing_key("zxviews", &extfvk);
 
     // Create a database
-    let mut s = crate::storage::Storage::new(&Config::ephemeral(), Network::Mainnet);
+    let mut s = crate::storage::Storage::new(&Config::ephemeral(), Network::Mainnet, false);
```
teor2345:
```suggestion
    let mut s = crate::storage::tests::new_test_storage(Network::Mainnet);
```

teor2345 (dismissed):
I did a bit of a merge and cleanup, let's see how it goes

upbqdn:
There was still an issue with path resolution that should be fixed now.

I also updated the description and labels. All should be good to go.

teor2345 (approved):
Looks great! I've made this merge wait until after #8080, because that's blocking some final tests.

upbqdn:
How did you do that?

teor2345:
Using `Depends-On: #8080` in the PR description:
https://docs.mergify.com/workflow/schedule-merge/#using-depends-on-to-manage-merge-dependencies

They also have a `Merge-After: time` field now.

teor2345:
@mergifyio refresh

teor2345:
This only failed the merge queue due to #7898

Issue #7728, opened 2023-10-11
Epic: Blockchain Scanning


mpguerra:
# Motivation

We want zebra to be able to scan shielded transactions using viewing keys. This issue will track the steps and work necessary in order for zebra to implement blockchain scanning.

# Scope

![scope diagram](https://github.com/ZcashFoundation/zebra/assets/1311133/f59b8436-a19a-4f94-a1db-087e1caab7e6)

## Feature: De-couple zebra-scan from zebrad

- [ ] The scanner will communicate with Zebra for the finalized part of the blockchain with a `ReadStateService::init` method: https://github.com/ZcashFoundation/zebra/issues/8520
- [ ] We need a replacement for `ChainTipChange` so we can use it in `zebra_scanner::spawn_init`: https://github.com/ZcashFoundation/zebra/issues/8580
- [ ] Create a new binary for the new process: https://github.com/ZcashFoundation/zebra/issues/8581
- [ ] Get rid of scanner startup code from zebrad: https://github.com/ZcashFoundation/zebra/issues/8582
- [ ] Move or remove the possibility of keys in the scanner: https://github.com/ZcashFoundation/zebra/issues/8584
- [ ] Testing: 
  - [ ] #8585,
  - [x] #8593
- [ ] Documentation: https://github.com/ZcashFoundation/zebra/issues/8586

Non finalized state:

- [ ] https://github.com/ZcashFoundation/zebra/issues/8576 

## RPC Interface improvements

- [ ] #8154
- [ ] #8156 
- [ ] Optional `ScanService` Requests 
    - [x] #8226
    - [x] #8202 
- [ ] Optional RPCs
    - [x] #8164
    - [ ] #8245
    - [x] #8233 
- [x] #8250
- [x] #8236 
- [x] #8243 

#### Key and scanning results database

- [x] #7932
- [ ] #7926
- [ ] #7937
- [x] #7944
- [ ] Launch a task to clean up old scanner database versions
- [ ] Delete scanner results with empty transaction ID lists, except for the highest one for each key
- [ ] Delete results for keys that have been removed from the config (security)
- [ ] Store keys in the database with account IDs and birthday heights 
- [ ] Reduce database size using small integer key IDs in the results table

#### Scanner

- [x] #8221
    - [x] #8229
- [x] #7905
- [x] #7934
    - [x] #7927
- [x] #8231
- [x] #8232

#### Integration with zebrad

- [x] #7943
- [x] #8101 

#### Security

[...]

PR #8080, opened 2023-12-09
feat(scanner): Restart scanning where left

### LINKED ISSUE #8022: feat(scanner): Start scanning where it left off last time

oxarbitrage:
## Motivation

We want the scanner to have the ability of resume the scanning where it was left after a restart. 

Close  https://github.com/ZcashFoundation/zebra/issues/8022

## Solution

I am pushing for a solution that involves adding a control empty result entry to the database for each key every X blocks. This is because a very probable scenario is that the key do not have any real transaction close to sapling (all new accounts).

### Testing

The functionality is tested by adding an acceptance test that needs a cache state close to tip.

## Review

@teor2345 had been working with me in this.

### Reviewer Checklist

## Follow Up Work

This PR does not include the cases of new keys/deleted keys in the config that are part of https://github.com/ZcashFoundation/zebra/issues/8022

It will be done in a separated PR.


zebra-scan/src/scan.rs
```diff
@@ -54,7 +54,7 @@ const INITIAL_WAIT: Duration = Duration::from_secs(15);
 const CHECK_INTERVAL: Duration = Duration::from_secs(30);
 
 /// We log an info log with progress after this many blocks.
-const INFO_LOG_INTERVAL: u32 = 100_000;
+const INFO_LOG_INTERVAL: u32 = 1_000;
```
oxarbitrage: Changed for testing. We can get back to 100_000.

zebra-scan/src/storage.rs
```diff
@@ -19,6 +19,10 @@ pub use db::{SaplingScannedResult, SaplingScanningKey};
 
 use self::db::ScannerWriteBatch;
 
+/// We insert an empty results entry to the database every this interval for each stored key,
+/// so we can track progress.
+const INSERT_CONTROL_INTERVAL: u32 = 1_000;
```
oxarbitrage, line 24: Does not need to be the same as `INFO_LOG_INTERVAL` but can be, to make things a bit simpler.

zebra-scan/src/storage.rs
```diff
@@ -89,10 +93,20 @@ impl Storage {
     ///
     /// This method can block while reading database files, so it must be inside spawn_blocking()
     /// in async code.
-    pub fn sapling_keys(&self) -> HashMap<SaplingScanningKey, Height> {
+    pub fn sapling_keys_birthdays(&self) -> HashMap<SaplingScanningKey, Height> {
```
oxarbitrage: I think this function can be removed.

zebra-scan/src/storage.rs
```diff
@@ -123,6 +141,11 @@ impl Storage {
             batch.insert_sapling_result(self, entry);
         }
 
+        // Add tracking entry for key.
+        if is_control_time {
+            batch.insert_sapling_height(self, &sapling_key, height);
```
oxarbitrage: We can delete lower heights controls that were added before. That way we will only have at most 1 control result at a time in the database for each key.

oxarbitrage: I am also unsure if the control block will show as a transaction found when getting results (via rocksdb tool or @upbqdn script).

zebra-state/src/service/finalized_state/disk_format/scan.rs
```diff
@@ -163,7 +163,8 @@ impl IntoDisk for SaplingScannedResult {
 
 impl FromDisk for SaplingScannedResult {
     fn from_bytes(bytes: impl AsRef<[u8]>) -> Self {
-        SaplingScannedResult(bytes.as_ref().try_into().unwrap())
+        // TODO: Change of confirm the `unwrap_or` is good enough.
+        SaplingScannedResult(bytes.as_ref().try_into().unwrap_or([0; 32]))
```
oxarbitrage: I am not sure about this.

zebrad/tests/acceptance.rs
```diff
@@ -2860,3 +2860,100 @@ fn scan_task_starts() -> Result<()> {
...
+///
+/// The test will run zebrad with a key to scan, scan the first few blocks after sapling and then stops.
+/// Then it will restart zebrad and check that it resumes scanning where it was left.
+///
+/// Note: This test will remove all the contents you may have in the ZEBRA_CACHED_STATE_DIR/private-scan directory
+/// so it can start with an empty scanning state.
+#[ignore]
+#[test]
+#[cfg(feature = "shielded-scan")]
+fn scan_start_where_left() -> Result<()> {
```
oxarbitrage, line 2879: I toke a few shortcuts doing the test instead of adding code in `launch`, `test_type`, etc.

oxarbitrage (commented):
A few comments and doubts i have.

zebra-scan/src/storage.rs
teor2345: It can take 9 hours to scan the most shielded-heavy 100,000 blocks, which is 4 blocks per second:
https://github.com/ZcashFoundation/zebra/issues/7813#issuecomment-1837756994

So I would recommend keeping this around 1000, regardless of the log interval.

zebra-scan/src/storage.rs
teor2345: > We can delete lower heights controls that were added before. That way we will only have at most 1 control result at a time in the database for each key.

I think this might be out of scope for the MVP, it's in the Future Iterations list in #7728 as:
> Delete scanner results with empty transaction ID lists, except for the highest one for each key

Deleting results needs extra testing, so it might be easier to put them all in a later iteration.

teor2345: > I am also unsure if the control block will show as a transaction found when getting results (via rocksdb tool or @upbqdn script).

The control blocks will show as an empty transaction ID, one that is zero bytes long. So it's almost impossible to interpret them as a real transaction ID.

For the RocksDB tool we will need to specifically ignore them using a command line tool like grep.

Marek's tool calls a method which already ignores them here:
https://github.com/ZcashFoundation/zebra/blob/5dd33d7265b9a8afe7f6b84264cc4faa2a5bfada/zebra-scan/src/storage/db/sapling.rs#L89-L95

zebrad/tests/acceptance.rs
teor2345: That's fine in tests 🙂

zebra-state/src/service/finalized_state/disk_format/scan.rs
teor2345: I think this might cause the control block transaction ID problem you were concerned about here:
https://github.com/ZcashFoundation/zebra/pull/8080#discussion_r1421816842

Instead, deserialize `Option<SaplingScannedResult>`, which will correctly turn an empty transaction ID into `None`.

I will push a PR that removes this implementation, because it should never be used - the format is always `Option<SaplingScannedResult>`. But that was only changed in a recent PR.
https://github.com/ZcashFoundation/zebra/blob/5dd33d7265b9a8afe7f6b84264cc4faa2a5bfada/zebra-scan/src/storage/db/sapling.rs#L5-L7

zebra-scan/src/storage.rs
teor2345: Do we ever want to know the birthday heights?

I think it might be useful later, but it will also be a pain to update when the format changes. So let's delete it and the function it calls?

zebrad/tests/acceptance.rs
```diff
@@ -2860,3 +2860,100 @@ fn scan_task_starts() -> Result<()> {
 
     Ok(())
 }
+
+/// Test that the scanner can continue scanning where it was left when zebrad restarts.
+///
+/// Needs a cache state close to the tip. A possible way to run it locally is:
```
teor2345, line 2866: Do we want to open a ticket to run this test in CI? It seems important enough.

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -141,6 +141,54 @@ impl Storage {
         keys
     }
 
+    /// Returns all the keys and their last scanned heights.
+    pub fn sapling_keys_and_last_scanned_heights(&self) -> HashMap<SaplingScanningKey, Height> {
+        let sapling_tx_ids = self.sapling_tx_ids_cf();
+        let mut keys = HashMap::new();
+
+        let last_stored_record: Option<(SaplingScannedDatabaseIndex, SaplingScannedResult)> =
```
teor2345: This is the other half of the serialization type fix in:
https://github.com/ZcashFoundation/zebra/pull/8080/files#r1421817246

```suggestion
        let last_stored_record: Option<(SaplingScannedDatabaseIndex, Option<SaplingScannedResult>)> =
```

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -141,6 +141,54 @@ impl Storage {
...
+
+        let mut last_stored_record_index = last_stored_record
+            .expect("checked this is `Some` in the code branch above")
+            .0;
+
+        loop {
+            // Find the previous key, and the last height we have for it.
+            let Some(entry) = self
+                .db
+                .zs_prev_key_value_back_from(&sapling_tx_ids, &last_stored_record_index)
```
teor2345: I think the first run through the loop can skip a key here.

I'm going to add a bunch of comments like this, so I don't lose track. But I think it would be better to talk through them in a video review.

zebra-scan/src/storage/db.rs
```diff
@@ -78,7 +78,17 @@ impl Storage {
...
 
-        // TODO: report the last scanned height here?
+        // Report where we are for each key in the database.
+        let keys = new_storage.sapling_keys_last_heights();
+        for (key_num, (_key, height)) in keys.iter().enumerate() {
+            tracing::info!(
+                "Last scanned height for key number {} is {}, resuming at {}",
+                key_num,
+                height.previous().expect("height is not genesis").as_usize(),
+                height.as_usize(),
```
teor2345, line 90: I think this previous() doesn't match the height we're inserting into the database.

Another one for a video review?

zebra-scan/src/scan.rs
```diff
@@ -168,24 +157,29 @@ pub async fn scan_height_and_store_results(
     // Scan it with all the keys.
     //
     // TODO: scan each key in parallel (after MVP?)
-    for (key_num, (sapling_key, birthday_height)) in key_birthdays.iter().enumerate() {
+    for (key_num, (sapling_key, last_scanned_height)) in key_last_scanned_heights.iter().enumerate()
+    {
+        // Only scan what was not scanned for each key
+        if height <= *last_scanned_height {
```
teor2345, line 162: This is good and matches what we write to the database.

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -141,6 +141,54 @@ impl Storage {
...
+            let _last_result: Option<SaplingScannedResult> = entry.1;
+
+            let height_results = self.sapling_results_for_key_and_height(&sapling_key, height);
+
+            // If there are no results for this block, then it's a "skip up to height" marker, and
+            // the target height is the next height. If there are some results, it's the actual
+            // target height.
+            if height_results.values().all(Option::is_none) {
+                height = height
+                    .next()
```
teor2345: We put the scanned height in the database, and skip heights equal to this height in the scanner. So I think we will skip a block here, but we don't actually want to.

This is different from birthday heights, where we want to scan the birthday height.

Another one for a video review.

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -141,6 +141,54 @@ impl Storage {
...
+                    .expect("results should only be stored for validated block heights");
+            }
+
+            keys.insert(sapling_key.clone(), height);
+
+            // Skip all the results after the next key.
```
teor2345: Nitpick: "after" could mean "including the next key".

The wording of the other function you got this from could also be clearer, but I think we're deleting that function.

```suggestion
            // Skip all the results until the next key.
```

teor2345 (commented):
Most of this looks good, but I think we need to count our keys and heights carefully. This might need some unit tests to be sure.

Let's do a video review and talk through the database reading loop?

zebra-state/src/service/finalized_state/disk_format/scan.rs
teor2345: This implementation is removed in PR #8075, so the change in this PR isn't needed:
https://github.com/ZcashFoundation/zebra/pull/8075/commits/105d29c4c01aa43ac6941b186ee1828a2ad26c3a

zebra-scan/src/storage.rs
teor2345: After PR #8075 this is fine and incorrect code won't compile any more.

teor2345:
I left some TODOs in PR #8075 and #8083 about testing this code using snapshots. So let's open another ticket to do that once they all merge?

I think that would cover most of the testing for this PR, except maybe for some unit tests for skipped keys and skipped heights.

zebra-scan/src/storage.rs
oxarbitrage: Removed in https://github.com/ZcashFoundation/zebra/pull/8080/commits/53dcb31882b12f954e8ee4a4fa742249f3dd1faa

zebra-scan/src/storage/db/sapling.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8080/commits/68bd53f0a0e11168c4765bbc20d21e7b70a0a517

zebra-state/src/service/finalized_state/disk_format/scan.rs
oxarbitrage: I approved https://github.com/ZcashFoundation/zebra/pull/8075, ill rebase this PR when that gets merged and do fixes needed here.

zebrad/tests/acceptance.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/issues/8085

zebra-scan/src/scan.rs
teor2345:
```suggestion
const INFO_LOG_INTERVAL: u32 = 10_000;
```

zebra-scan/src/scan.rs
```diff
@@ -68,17 +68,17 @@ pub async fn start(
```
teor2345: We're reading all the blocks even if there isn't a key that needs them. So let's set `height` to the minimum height in `key_heights` instead?

zebra-scan/src/storage/db.rs
teor2345: We start at the next height after the last scanned height:
```suggestion
                height.as_usize(),
                height.next().expect("height is not maximum").as_usize(),
```

zebra-scan/src/storage/db/sapling.rs
teor2345: I think the way to fix this is to move the `zs_prev_key_value_back_from()` to the bottom of the loop. So it only gets called once on the last key.

zebra-scan/src/storage/db/sapling.rs
teor2345: I think this isn't needed at all, because the database already has the last scanned heights.

This is the other half of the logging fix here:
https://github.com/ZcashFoundation/zebra/pull/8080/files#r1423194715

zebra-scan/src/storage/db/sapling.rs
```diff
@@ -95,37 +95,39 @@ impl Storage {
...
 
-    /// Returns all the keys and their birthday heights.
-    pub fn sapling_keys_and_birthday_heights(&self) -> HashMap<SaplingScanningKey, Height> {
-        // This code is a bit complex because we don't have a separate column family for keys
-        // and their birthday heights.
-        //
-        // TODO: make a separate column family after the MVP.
-
+    /// Returns all the keys and their last scanned heights.
+    pub fn sapling_keys_and_last_scanned_heights(&self) -> HashMap<SaplingScanningKey, Height> {
```
teor2345, line 101: In a new PR we can add a test that compares `sapling_results_for_key(key).keys().max()` with `sapling_keys_and_last_scanned_heights().get(key)`, and makes sure they match.

zebra-scan/src/storage/db/sapling.rs
oxarbitrage: Done as part of https://github.com/ZcashFoundation/zebra/pull/8080/commits/ae5bc94e54f94fd9cb49e95ead4cf62184edaadb

teor2345 (commented):
Looking good now, let me know if you want help with the height we start fetching blocks.

zebra-scan/src/storage/db/sapling.rs
teor2345: I have code for this now, I'll push it when it works.

zebra-scan/src/scan.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8080/commits/0a5c08628abb412515acc4bd6c3b10ad3be3dc34

teor2345 (approved):
Thanks for working through this, let's see how it goes!

PR #8083, opened 2023-12-11
test(scan): Add typed database format snapshots to the scanner

### LINKED ISSUE #8069: test: Add database snapshot tests for the scanner database

teor2345:
## Motivation

These are part of the standard database tests in #8069.

Depends-On: #8075

It's actually much simpler than the existing state tests.

## Solution

- Add snapshot tests for every scanner database method, using the typed format
- Expose more test support code in zebra-scan

Related changes:
- Clean up `hex` implementation in `transaction::Hash`

### Testing

These are extra tests.

I have manually checked the data formats in the snapshots and they appear correct.

## Review

This would be helpful to merge soon, because it tests other PRs.

### Reviewer Checklist

## Follow Up Work

Database conflict and ephemeral tests in #8069, or whatever acceptance tests we decide are useful.

### DIFF
--- a/zebra-scan/src/storage/db/tests.rs
+++ b/zebra-scan/src/storage/db/tests.rs
@@ -1,3 +1,58 @@
 //! General scanner database tests.
 
+use std::sync::Arc;
+
+use zebra_chain::{
+    block::{Block, Height},
+    parameters::Network::{self, *},
+    serialization::ZcashDeserializeInto,
+};
+use zebra_state::TransactionIndex;
+
+use crate::{
+    storage::Storage,
+    tests::{FAKE_SAPLING_VIEWING_KEY, ZECPAGES_SAPLING_VIEWING_KEY},
+    Config,
+};
+
+#[cfg(test)]
 mod snapshot;
+
+/// Returns an empty `Storage` suitable for testing.
+pub fn new_test_storage(network: Network) -> Storage {
+    Storage::new(&Config::ephemeral(), network)
+}
+
+/// Add fake keys to `storage` for testing purposes.
+pub fn add_fake_keys(storage: &mut Storage) {
+    // Snapshot a birthday that is automatically set to activation height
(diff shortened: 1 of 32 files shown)


zebra-scan/src/storage/db/tests.rs
```diff
@@ -1,3 +1,58 @@
...
+/// Add fake results to `storage` for testing purposes.
+pub fn add_fake_results(storage: &mut Storage, network: Network, height: Height) {
+    let blocks = match network {
+        Mainnet => &*zebra_test::vectors::CONTINUOUS_MAINNET_BLOCKS,
+        Testnet => &*zebra_test::vectors::CONTINUOUS_TESTNET_BLOCKS,
+    };
+
+    let block: Arc<Block> = blocks
+        .get(&height.0)
+        .expect("block height has test data")
```
upbqdn, line 43: Since `blocks` has a constant length, currently set to 11, and `height` is a parameter the caller sets arbitrarily, it's easy to trigger a panic here.

I personally still struggle with formulating error messages from time to time. The message here says "block height has test data", which is the opposite of what's happening when displayed to the user. On the other hand, the message explains to the developer why we can expect an `Ok` here. Do you recommend using this style more often than the other?

upbqdn (dismissed):

zebra-scan/src/storage/db/tests.rs
teor2345: I think usual Rust style is "unexpected (bad thing): (good thing) must happen". That helps remove the ambiguity.

teor2345:
I rebased this PR.

upbqdn (approved):

PR #8104, opened 2023-12-13
Add a scanning results reader

### LINKED ISSUE #8046: Document how to run and query the results of the blockchain scanning MVP

upbqdn:
## Motivation

So far, we have only used `ldb` to print the contents of the storage with the scanning results. Since the results contain only TXIDs, we couldn't straightforwardly tell much about their properties.

Depends-On: #8079.

## Solution

- Add a separate binary into `zebra-utils` that prints Zebra's scanning results:

    1. Open Zebra's scanning storage and read the results containing scanning keys and TXIDs.
    2. Fetch the transactions by their TXIDs from Zebra using the `getrawtransaction` RPC.
    3. Decrypt the TX outputs using the corresponding scanning key.
    4. Print the memos in the outputs.

- Add a tutorial describing how to use the tool.

### Testing

I didn't add or change any tests.

### Reviewer Checklist

### DIFF
--- a/zebra-utils/Cargo.toml
+++ b/zebra-utils/Cargo.toml
@@ -67,6 +67,7 @@ getblocktemplate-rpcs = [
 ]
 
 shielded-scan = [
+    "itertools",
     "jsonrpc",
     "zcash_primitives",
     "zcash_client_backend",
--- a/zebra-utils/Cargo.toml
+++ b/zebra-utils/Cargo.toml
@@ -36,6 +36,11 @@ name = "block-template-to-proposal"
 path = "src/bin/block-template-to-proposal/main.rs"
 required-features = ["getblocktemplate-rpcs"]
 
+[[bin]]
+name = "scanning-results-reader"
+path = "src/bin/scanning-results-reader/main.rs"
+required-features = ["shielded-scan"]
+
 [features]
 default = []
 
@@ -61,6 +66,13 @@ getblocktemplate-rpcs = [
     "zebra-chain/getblocktemplate-rpcs",
 ]
 
+shielded-scan = [
+    "jsonrpc",
+    "zcash_primitives",
+    "zcash_client_backend",
+    "zebra-scan"
+]
+
 [dependencies]
 color-eyre = "0.6.2"
 # This is a transitive dependency via color-eyre.
@@ -76,6 +88,7 @@ thiserror = "1.0.48"
 
 zebra-node-services = { path = "../zebra-node-services", version = "1.0.0-beta.32" }
 zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.32" }
+zebra-scan = { path = "../zebra-scan", version="0.1.0-alpha.1",optional = true }
 
 # These crates are needed for the block-template-to-proposal binary
 zebra-rpc = { path = "../zebra-rpc", version = "1.0.0-beta.32", optional = true }
@@ -90,3 +103,8 @@ reqwest = { version = "0.11.22", default-features = false, features = ["rustls-t
 
 # These crates are needed for the zebra-checkpoints and search-issue-refs binaries
 tokio = { version = "1.34.0", features = ["full"], optional = true }
+
+jsonrpc = { version="0.16.0", optional = true }
+
+zcash_primitives = { version="0.13.0-rc.1", optional = true }
+zcash_client_backend = {version="0.10.0-rc.1", optional = true}
--- a/zebra-utils/Cargo.toml
+++ b/zebra-utils/Cargo.toml
@@ -88,7 +88,7 @@ thiserror = "1.0.48"
 
 zebra-node-services = { path = "../zebra-node-services", version = "1.0.0-beta.32" }
 zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.32" }
-zebra-scan = { path = "../zebra-scan", version="0.1.0-alpha.1",optional = true }
+zebra-scan = { path = "../zebra-scan", version = "0.1.0-alpha.1", optional = true }
 
 # These crates are needed for the block-template-to-proposal binary
 zebra-rpc = { path = "../zebra-rpc", version = "1.0.0-beta.32", optional = true }
@@ -104,7 +104,7 @@ reqwest = { version = "0.11.22", default-features = false, features = ["rustls-t
 # These crates are needed for the zebra-checkpoints and search-issue-refs binaries
 tokio = { version = "1.34.0", features = ["full"], optional = true }
 
-jsonrpc = { version="0.16.0", optional = true }
+jsonrpc = { version = "0.16.0", optional = true }
 
-zcash_primitives = { version="0.13.0-rc.1", optional = true }
-zcash_client_backend = {version="0.10.0-rc.1", optional = true}
+zcash_primitives = { version = "0.13.0-rc.1", optional = true }
+zcash_client_backend = {version = "0.10.0-rc.1", optional = true}
--- a/zebra-utils/README.md
+++ b/zebra-utils/README.md
@@ -1,17 +1,16 @@
 # Zebra Utilities
 
-This crate contains tools for zebra maintainers.
-
-## Programs
+Tools for maintaining and testing Zebra:
 
 - [zebra-checkpoints](#zebra-checkpoints)
 - [zebrad-hash-lookup](#zebrad-hash-lookup)
 - [zebrad-log-filter](#zebrad-log-filter)
 - [zcash-rpc-diff](#zcash-rpc-diff)
+- [scanning-results-reader](#scanning-results-reader)
 
 Binaries are easier to use if they are located in your system execution path.
 
-### zebra-checkpoints
+## zebra-checkpoints
 
 This command generates a list of zebra checkpoints, and writes them to standard output. Each checkpoint consists of a block height and hash.
 
@@ -93,7 +92,7 @@ Then use the commands above to regenerate the checkpoints.
 - Open a pull request with the updated Mainnet and Testnet lists at:
   https://github.com/ZcashFoundation/zebra/pulls
 
-### zebrad-hash-lookup
+## zebrad-hash-lookup
 
 Given a block hash the script will get additional information using `zcash-cli`.
 
@@ -108,7 +107,7 @@ $
 ```
 This program is commonly used as part of `zebrad-log-filter` where hashes will be captured from `zebrad` output.
 
-### zebrad-log-filter
+## zebrad-log-filter
 
 The program is designed to filter the output from the zebra terminal or log file. Each time a hash is seen the script will capture it and get the additional information using `zebrad-hash-lookup`.
 
@@ -127,7 +126,7 @@ next: 00000001436277884eef900772f0fcec9566becccebaab4713fd665b60fab309
 ...
 ```
 
-### zcash-rpc-diff
+## zcash-rpc-diff
 
 This program compares `zebrad` and `zcashd` RPC responses.
 
--- a/zebra-utils/README.md
+++ b/zebra-utils/README.md
@@ -187,3 +187,52 @@ You can override the binaries the script calls using these environmental variabl
 - `$ZCASH_CLI`
 - `$DIFF`
 - `$JQ`
+
+## Scanning Results Reader
+
+A utility for displaying Zebra's scanning results.
+
+### How It Works
+
+1. Opens Zebra's scanning storage and reads the results containing scanning keys
+   and TXIDs.
+2. Fetches the transactions by their TXIDs from Zebra using the
+   `getrawtransaction` RPC.
+3. Decrypts the tx outputs using the corresponding scanning key.
+4. Prints the memos in the outputs.
+
+### How to Try It
+
+#### Scan the Block Chain with Zebra
+
+1. Add a viewing key to your Zebra config file. For example:
+
+   ``` toml
+   [shielded_scan.sapling_keys_to_scan]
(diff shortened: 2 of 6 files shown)


zebra-utils/Cargo.toml
```diff
@@ -76,6 +88,7 @@ thiserror = "1.0.48"
 
 zebra-node-services = { path = "../zebra-node-services", version = "1.0.0-beta.32" }
 zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.32" }
+zebra-scan = { path = "../zebra-scan", version="0.1.0-alpha.1",optional = true }
```
teor2345: Optional Nitpick: spacing

```suggestion
zebra-scan = { path = "../zebra-scan", version = "0.1.0-alpha.1", optional = true }
```

zebra-utils/Cargo.toml
```diff
@@ -90,3 +103,8 @@ reqwest = { version = "0.11.22", default-features = false, features = ["rustls-t
 
 # These crates are needed for the zebra-checkpoints and search-issue-refs binaries
 tokio = { version = "1.34.0", features = ["full"], optional = true }
+
+jsonrpc = { version="0.16.0", optional = true }
+
+zcash_primitives = { version="0.13.0-rc.1", optional = true }
+zcash_client_backend = {version="0.10.0-rc.1", optional = true}
```
teor2345: Optional Nitpick:

```suggestion
jsonrpc = { version = "0.16.0", optional = true }

zcash_primitives = { version = "0.13.0-rc.1", optional = true }
zcash_client_backend = {version = "0.10.0-rc.1", optional = true}
```

zebra-utils/README.md
```diff
@@ -188,3 +187,52 @@ You can override the binaries the script calls using these environmental variabl
...
+    [network]
+    network = 'Mainnet'
+
+    [rpc]
+    listen_addr = "127.0.0.1:8232"
+    ```
+
+3. Compile and run Zebra with `--features "shielded-scan"` and your config file.
+   Zebra will start scanning the block chain and inform you about its progress
+   each 100 000 blocks in the log.
```
teor2345: A recent PR changed this to 10,000:
https://github.com/ZcashFoundation/zebra/blob/92758a0d9ff6299e92c0c212b97eff1a30febf59/zebra-scan/src/scan.rs#L57

```suggestion
   each 10 000 blocks in the log.
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+use jsonrpc::Client;
+
+use serde_json::value::RawValue;
+use zcash_client_backend::decrypt_transaction;
+use zcash_client_backend::keys::UnifiedFullViewingKey;
+use zcash_primitives::consensus::{BlockHeight, BranchId};
+use zcash_primitives::transaction::Transaction;
+use zcash_primitives::zip32::AccountId;
+use zebra_scan::scan::sapling_key_to_scan_block_keys;
+use zebra_scan::{storage::Storage, Config};
```
teor2345: Optional Nitpick: standard grouping

```suggestion
use jsonrpc::simple_http::SimpleHttpTransport;
use jsonrpc::Client;
use serde_json::value::RawValue;

use zcash_client_backend::decrypt_transaction;
use zcash_client_backend::keys::UnifiedFullViewingKey;
use zcash_primitives::consensus::{BlockHeight, BranchId};
use zcash_primitives::transaction::Transaction;
use zcash_primitives::zip32::AccountId;

use zebra_scan::scan::sapling_key_to_scan_block_keys;
use zebra_scan::{storage::Storage, Config};
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+///
+/// - The Sapling key from the storage is not valid.
+/// - There is no diversifiable full viewing key (dfvk) available.
+/// - The RPC response cannot be decoded from a hex string to bytes.
+/// - The transaction fetched via RPC cannot be deserialized from raw bytes.
+#[allow(clippy::print_stdout)]
+pub fn main() {
+    let network = zcash_primitives::consensus::Network::MainNetwork;
+    let storage = Storage::new(&Config::default(), zebra_network(&network), true);
+    let mut prev_memo = "".to_owned();
```
teor2345, line 48: Optional Style: Strictly we should use `None` for a missing value, but that might complicate the code a lot.

This doesn't change functionality and is probably fine for a basic tool, but it might help to add a comment.

```suggestion
    // If the first memo is empty, it doesn't get printed. But we never print empty memos anyway.
    let mut prev_memo = "".to_owned();
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    for (key, _) in storage.sapling_keys_last_heights().iter() {
+        let dfvk = sapling_key_to_scan_block_keys(key, zebra_network(&network))
+            .expect("Scanning key from the storage should be valid")
+            .0
+            .into_iter()
+            .next()
```
teor2345: Optional: There's a method for this:
https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one

It also needs a `use itertools::Itertools` and maybe a dependency on `itertools`.

```suggestion
            .exactly_one()
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+            .expect("There should be exactly one dfvk");
+
+        let ufvk_with_acc_id = HashMap::from([(
+            AccountId::from(1),
+            UnifiedFullViewingKey::new(Some(dfvk), None).expect("`dfvk` should be `Some`"),
+        )]);
+
+        for (height, txids) in storage.sapling_results(key) {
+            let height = BlockHeight::from_u32(height.0);
```
teor2345: Optional: Would this be more useful as a `From` impl in `zebra_chain`?

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+
+            for txid in txids.iter() {
+                let tx = Transaction::read(
+                    &hex::decode(&get_tx_via_rpc(hex::encode(txid.bytes_in_display_order())))
```
teor2345: Optional: This got added recently. You might also need to `use hex::ToHex`.

```suggestion
                    &hex::decode(&get_tx_via_rpc(txid.encode_hex()))
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Trims trailing zeroes from a memo, and returns the memo as a string.
+fn memo_bytes_to_string(memo: &[u8; 512]) -> String {
+    match memo.iter().rposition(|&byte| byte != 0) {
+        Some(i) => std::str::from_utf8(&memo[..=i]).unwrap_or("").to_owned(),
```
teor2345: Optional: We could ignore some invalid UTF-8 sequences, which might or might not be what we want. Let's test and see?

It might do weird things with binary data, so maybe we could impose a minimum length on the resulting string? (Specifically, a length of adjacent characters that are not the UTF replacement character.)

```suggestion
        Some(i) => std::str::from_utf8_lossy(&memo[..=i]).into_owned(),
```

zebra-utils/src/bin/scanning-results-reader/main.rs
```diff
@@ -0,0 +1,128 @@
...
+) -> zebra_chain::parameters::Network {
+    match network {
+        zcash_primitives::consensus::Network::MainNetwork => {
+            zebra_chain::parameters::Network::Mainnet
+        }
+        zcash_primitives::consensus::Network::TestNetwork => {
+            zebra_chain::parameters::Network::Testnet
+        }
+    }
+}
```
teor2345: Optional: This seems like it would be useful in `zebra_chain`, if we don't have this `From` impl there already.

teor2345 (dismissed):
Looks great!

I wonder if we could move some of this code into `scan.rs`, or maybe a separate `decrypt.rs` module? I could imagine two APIs:
- high-level: key string, transaction data bytes, Zebra network, height, returns Zebra typed memo data
- low-level: zcash_primitives key type, transaction type, network, heights, returns raw or zcash_primitives memo data

This could happen in a separate ticket.

zebra-utils/Cargo.toml
upbqdn: I formatted the file manually because I didn't have any tool for it. I tried to imitate the formatting based on existing parts. Did you notice this bug just by looking at the file, or did you use some tool?

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Yes, this will keep excluding empty memos until the first non-empty one. I kept many things basic when writing this.

zebra-utils/Cargo.toml
teor2345: Just noticed it by reading, I'm not sure if there's a lint tool. But here is the convention:
https://doc.rust-lang.org/nightly/style-guide/cargo.html

teor2345 (dismissed):
Happy to merge this now and do any other changes in another PR. Or do the changes here. Totally up to you!

zebra-utils/Cargo.toml
upbqdn: Thanks for sharing those.

upbqdn:
I'd prefer to make the changes here. I should be done by tomorrow.

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Do you think we could simplify the return type of `sapling_key_to_scan_block_keys`: https://github.com/zcashfoundation/zebra/blob/a3af2a702d2eb043e89e9b0b8e31e963cb10aeb8/zebra-scan/src/scan.rs#L291-L296 instead?

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: You're right. We even have `impl From<Network> for zcash_primitives::consensus::Network`
 in Zebra already. I couldn't use it when the reader was in its own crate. I forgot about that when I was moving the code here.

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: I used it in [9e16ad5](https://github.com/ZcashFoundation/zebra/pull/8104/commits/9e16ad569d13d5a8299d1d0a6a4c791f29b513bb).

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Done in [c0c2101](https://github.com/ZcashFoundation/zebra/pull/8104/commits/c0c21016f13f91b63060152cdaabd982c5901354).

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Done in [c0c2101](https://github.com/ZcashFoundation/zebra/pull/8104/commits/c0c21016f13f91b63060152cdaabd982c5901354).

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Done in [c0c2101](https://github.com/ZcashFoundation/zebra/pull/8104/commits/c0c21016f13f91b63060152cdaabd982c5901354).

zebra-utils/src/bin/scanning-results-reader/main.rs
upbqdn: Done in [c0c2101](https://github.com/ZcashFoundation/zebra/pull/8104/commits/c0c21016f13f91b63060152cdaabd982c5901354).

zebra-utils/src/bin/scanning-results-reader/main.rs
teor2345: > Do you think we could simplify the return type of `sapling_key_to_scan_block_keys`: [zcashfoundation/zebra@`a3af2a7`/zebra-scan/src/scan.rs#L291-L296](https://github.com/zcashfoundation/zebra/blob/a3af2a702d2eb043e89e9b0b8e31e963cb10aeb8/zebra-scan/src/scan.rs#L291-L296) instead?

Yes, I think we could remove IVKs from the MVP production code. We don't know the standard IVK format and we don't have code to parse them, so it seems better to simplify it.

Is there a way to move the IVK code into a test? Then we can be sure we have support without running it in production.

teor2345 (approved):
Let's go! Then we can do some cleanups later.

zebra-utils/README.md
```diff
@@ -188,3 +187,52 @@ You can override the binaries the script calls using these environmental variabl
 - `$ZCASH_CLI`
 - `$DIFF`
 - `$JQ`
+
+## Scanning Results Reader
```
mpguerra, line 191: We should probably link to this from the zebra book

Issue #7833, opened 2023-10-25
Document how to design state queries (and writes)


teor2345:
This needs to be done before #7738, but it's not a ticket template.

Everything that goes in the docs doesn't need to go in the ticket template, I can just link to the docs from the template.

This should document the typed column families from PR #8112.

teor2345:
From PR #7972:

Document how to avoid this happening again in the state upgrade docs.

In particular, don't concurrently read and write the same keys in the same column family:
- in an upgrade, and
- in the sync.

If we need to do that, we should make the upgrade happen before syncing. (That is, block on the upgrade when the state is opened.)

And make sure the read and write methods have compatible conditions.

teor2345:
The typed column families from PR #8112 might also help with some common database bugs.

PR #8112, opened 2023-12-15
Make sure scanner database is accessed using the correct types

teor2345:
## Motivation

During the scanner MVP, it was easy to change the column family's types, but forget to update the code everywhere. There were at least two bugs like this, from two different developers.

Instead, we can make each column family strongly typed, so it can only be accessed using the correct types.

Preparation for #7833, let's document the easier way this PR implements.

### API & Rust Reference

https://docs.rs/rocksdb/latest/rocksdb/struct.WriteBatchWithTransaction.html

https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html
https://doc.rust-lang.org/std/marker/struct.PhantomData.html

## Solution

Production:
- Define types for reading and writing sapling transaction ID results
- Simplify implementations using typed column families
- Make sure each method writes its own batch
- Implement common methods as a trivial trait

Docs & debugging:
- impl Debug and PartialEq on some database types to improve debugging
- Update docs & parameter names
- Deprecate some traits that we shouldn't use for new column families

Unrelated cleanups:
- impl TryFrom<zcash_primitives::BlockHeight> for Height
- Simplify logging macros in the scanner

We can't make the column family name string constant part of the type, because Rust doesn't support String const generics yet.

### Testing

This PR is well covered by existing tests, which pass.

## Review

This is a routine cleanup/internal enhancement.

### Reviewer Checklist

## Follow Up Work

Use the new `TypedColumnFamily` and `WriteTypedBatch` for existing state column families. We might discover some bugs this way.

Document how to set up a new column family using these types - #7833

### DIFF
--- a/zebra-chain/src/block/height.rs
+++ b/zebra-chain/src/block/height.rs
@@ -112,6 +112,15 @@ impl From<Height> for BlockHeight {
     }
 }
 
+impl TryFrom<BlockHeight> for Height {
+    type Error = &'static str;
+
+    /// Checks that the `height` is within the valid [`Height`] range.
+    fn try_from(height: BlockHeight) -> Result<Self, Self::Error> {
+        Self::try_from(u32::from(height))
+    }
+}
+
 /// A difference between two [`Height`]s, possibly negative.
 ///
 /// This can represent the difference between any height values,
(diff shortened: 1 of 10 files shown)


upbqdn (approved):
utACK.

Looks great!

Issue #7937, opened 2023-11-10
Create a `zebra-db` crate


mpguerra:
## Motivation

We want to re-using RocksDB for our scanner work, because we already have well-tested low level interfaces to it.

### Suggested Chanhes

We can split out a new `zebra-db` crate, which `zebra-state` and `zebra-scan` can depend on.

That crate should contain the low-level database code from ticket #7926.

### Related Work

This was mentioned in https://github.com/ZcashFoundation/zebra/issues/7904#issuecomment-1802586921 and earlier comments on that ticket.

natalieesk:
Closed as it was deprioritised, but then reopened as we will still do this work.

PR #8115, opened 2023-12-18
Move history tree and value balance to typed column families

teor2345:
## Motivation

This is another part of ticket #7833, making sure that the new typed column families work for more complex situations in the state. These commits or PRs will become examples in the documentation.

This PR converts the history tree and value balance to typed column families, and cleans up some risky and redundant code.

### Database Format Docs

Please check the types in the PR match the existing types in the database:

This PR removes complexity by using consistent types for reading and writing column families.

## Solution

- Convert history trees to typed column families
  - Create methods for legacy (compatibility with 1.2.0 and earlier) and raw types (upgrade from 1.2.0 to 1.3.0)
  - Delete redundant format code that makes risky assumptions about history tree formats on disk
- Convert value balance to typed column families
- Add typed methods for writing into existing batches (owned and borrowed)
- Make it easier to pass references when writing typed column families (by taking `&Key` and `&Value` and relying on auto-deref)
- Always pass `ZebraDb` to database methods

Related changes:
- Clean up and document `zebra-state` exports

Unrelated cleanups:
- Remove a 1.72 lint exception now 1.74 is stable

### Reviewer Checklist


upbqdn (approved):
The solution in this PR contains this item:

> - Always pass `ZebraDb` to database methods

I reverted this change because it caused a very tricky bug where Zebra would compile with `cargo build`, but not, for example, with `cargo build --package zebrad` or in other cases. We'll need a bigger refactor if we want to do that change.

I also manually resolved merge conflicts and fixed some API bugs that the merge introduced.

Otherwise, this PR looks good.

upbqdn:
The cause of the bug was that passing `ZebraDb` to functions that expect `DiskDb` created a reliance on a `Deref` impl for `ZebraDb` targeting `DiskDb`, which is conditionally compiled if the `proptest-impl` or `shielded-scan` features are enabled during the compilation of `zebra-state`. By a weird coincidence, running `cargo build` or `cargo build --release` seems to enable `shielded-scan` when compiling `zebra-state`, so the `Deref` impl gets compiled in that case. It looks like this is a bug on its own because `shielded-scan` should not be enabled. Running, for example, `cargo build --package zebrad` doesn't enable the feature, so the `Deref` was missing, and Zebra didn't compile.

teor2345:
> The cause of the bug was that passing `ZebraDb` to functions that expect `DiskDb` created a reliance on a `Deref` impl for `ZebraDb` targeting `DiskDb`, which is conditionally compiled if the `proptest-impl` or `shielded-scan` features are enabled during the compilation of `zebra-state`. By a weird coincidence, running `cargo build` or `cargo build --release` seems to enable `shielded-scan` when compiling `zebra-state`, so the `Deref` impl gets compiled in that case. It looks like this is a bug on its own because `shielded-scan` should not be enabled. Running, for example, `cargo build --package zebrad` doesn't enable the feature, so the `Deref` was missing, and Zebra didn't compile.

We could have enabled the `Deref` impl in all builds, or called the `db()` method where needed.

But we'll deal with this when we split ZebraDb and DiskDb properly in #7937.

PR #8121, opened 2023-12-20
feat(scanner): Don't scan and log if we are below sapling height

### LINKED ISSUE #8041: Scanner should log a message when Zebra is before the Sapling activation block
I think we should print a message as soon as we know we are behind the sapling activation block saying something as:

"scan task will not get started until the blockchain is at sapling activation height (block number: {sapling activation_height})"

This way users will know the scanner will do nothing until that.

We can look for that message in this test so it will not require further modifications. A  new test will be done with state.

_Originally posted by @oxarbitrage in https://github.com/ZcashFoundation/zebra/pull/8040#discussion_r1413121675_

oxarbitrage:
## Motivation

We want to log when the tip is below sapling and the scanner is activated. 

Close https://github.com/ZcashFoundation/zebra/issues/8041

## Solution

Get the current tip at each scanning loop iteration and compare it with sapling activation height. Do not scan if the tip is below the sapling activation height.

### Testing

The `scan_task_starts` was modified to test this feature.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/scan.rs
+++ b/zebra-scan/src/scan.rs
@@ -70,7 +70,8 @@ pub async fn start(
         .await;
     let key_heights = Arc::new(key_heights);
 
-    let mut height = get_min_height(&key_heights).unwrap_or(storage.min_sapling_birthday_height());
+    let sapling_activation_height = storage.min_sapling_birthday_height();
+    let mut height = get_min_height(&key_heights).unwrap_or(sapling_activation_height);
 
     // Parse and convert keys once, then use them to scan all blocks.
     // There is some cryptography here, but it should be fast even with thousands of keys.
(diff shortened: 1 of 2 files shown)


zebra-scan/src/scan.rs
```diff
@@ -90,6 +91,14 @@ pub async fn start(
...
 
     loop {
+        // Do not scan and notify if we are below sapling activation height.
+        let tip_height = tip_height(state.clone()).await?;
+        if tip_height < sapling_activation_height {
+            info!("scanner is waiting for sapling activation. Current tip: {}, Sapling activation: {}", tip_height.0, sapling_activation_height.0);
+            tokio::time::sleep(CHECK_INTERVAL).await;
+            continue;
+        }
+
```
upbqdn: I suggest we move this tip-checking logic above the scanning loop into its own loop to:

1. keep the scanning loop simple, and
2. not slow it down by requesting the tip in each iteration.

oxarbitrage: Thank you for that, was done in https://github.com/ZcashFoundation/zebra/pull/8121/commits/d6c39846cf1e72318d66cec7952fab02035fd133

upbqdn (approved):

PR #8127, opened 2023-12-29
fix(elasticsearch): Make bulk size the same for testnet and mainnet

### LINKED ISSUE #7270: Avoid panics by always limiting elasticsearch bulk loads to 100MB
## Motivation

Zebra's elasticsearch code will panic if it sends a request containing more than 100 MB of JSON.

Currently we limit elasticsearch to approximately 101 MB of JSON on mainnet, and 1.6 GB of JSON on testnet. Each block uses around 4.1 MB of JSON, because its data is hex-encoded, and then extra fields are added.

The current settings work for existing blocks. But anyone can submit enough transactions to create 50 max-size blocks to mainnet or testnet, and make this code panic.

_Originally posted by @teor2345 in https://github.com/ZcashFoundation/zebra/pull/7268#discussion_r1268941742_
            

### Priority

This is currently an experimental feature, so this security fix is optional.

oxarbitrage:
## Motivation

The bulk size for testnet is greater than mainnet which should synchronize the chain faster, however Elasticsearch has a limit of 100M per bulk. A stream of big blocks in the testnet can crash the node. 

Close https://github.com/ZcashFoundation/zebra/issues/7270

## Solution

Make the testnet bulk size equal to the mainnet. This is 24 blocks per bulk.

### Testing

I manually tested running the testnet, it syncing and adding records to the database as expected.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-state/src/service/finalized_state.rs
+++ b/zebra-state/src/service/finalized_state.rs
@@ -458,15 +458,11 @@ impl FinalizedState {
             let block_time = block.header.time.timestamp();
             let local_time = chrono::Utc::now().timestamp();
 
-            // Mainnet bulk size is small enough to avoid the elasticsearch 100mb content
-            // length limitation. MAX_BLOCK_BYTES = 2MB but each block use around 4.1 MB of JSON.
+            // Bulk size is small enough to avoid the elasticsearch 100mb content length limitation.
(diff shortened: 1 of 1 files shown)


teor2345 (approved):
Thanks!

PR #8005, opened 2023-11-27
ref(workflow): move most scripts to their own executables

gustavovalverde:
## Motivation

Some CI files are hard/understand as they have bash scripts included, some of this scripts are also hard to maintain or lint considering they're inside a `.yml` and not in an actual `.sh`

Some of these scripts might also be useful as a separate file to run independently for specific use cases, like searching for disks in GCP (but that's not part of the scope of this PR)/

Fixes #6168

Depends-On: #8083

#### Check before marking the PR as ready for review:
  - [X] Will the PR name make sense to users?
  - [X] Does the PR have a priority label?
  - [ ] Have you added or updated tests?
  - [X] Is the documentation up to date?

##### For significant changes:
  - [ ] Is there a summary in the CHANGELOG?
  - [ ] Can these changes be split into multiple PRs?

_If a checkbox isn't relevant to the PR, mark it as done._

### Complex Code or Requirements

Some scripts required refactoring for consistency and better use outside the workflow YAML

## Solution

- Move most scripts to the `./github/workflows/scripts` directory
- Lint the scripts to comply with bash linting
- Refactor the scripts so it's easier to make further updates.

## Review

We need to confirm the right disk are being pulled, comparing with recent PRs

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

## Follow Up Work

Make this scripts usable as a standalone script

PR #8132, opened 2024-01-02
Revert "ref(workflow): move most scripts to their own executables"

### LINKED ISSUE #8123: main branch CI failed: "GitHub Actions has encountered an internal error when running your job" in Integration Tests on GCP
GitHub Actions workflow [Integration Tests on GCP #664](https://github.com/ZcashFoundation/zebra/actions/runs/7291110237) failed.

Event: push
Branch: [main](https://github.com/ZcashFoundation/zebra/tree/main)
Commit: [d1e72c114faff3eab212c90a21a7258f4e5b11ed](https://github.com/ZcashFoundation/zebra/commit/d1e72c114faff3eab212c90a21a7258f4e5b11ed)

<sup><i>Created by [jayqi/failed-build-issue-action](https://github.com/jayqi/failed-build-issue-action)</i></sup>

teor2345:
## Motivation

CI has been failing with an internal GitHub error since PR #8005 merged. Since people are away, the simplest way to fix this is to revert it for now.

Reverts ZcashFoundation/zebra#8005
Close #8123


teor2345:
If this CI passes without errors, we can merge this PR. We need to manually scroll down and check it for errors, because they don't reliably show up in the CI summary or auto-opened tickets:
https://github.com/ZcashFoundation/zebra/actions/runs/7383079119?pr=8132

teor2345:
That worked, and I manually checked all of the other CI runs for this PR, they also passed.

gustavovalverde (changes requested):
Fixing it this way would revert improvements made to workflows that are not failing. From the 15+ impacted workflow files here, 1 workflow is failing. I'd prefer to tackle this specific issue and not revert the whole PR. I'll prioritize this

teor2345:
> Fixing it this way would revert improvements made to workflows that are not failing. From the 15+ impacted workflow files here, 1 workflow is failing. I'd prefer to tackle this specific issue and not revert the whole PR. I'll prioritize this

My goal was restoring test coverage as reliably and quickly as possible. This PR does that.

If we merge this PR, then the rest of the team doesn't need to worry about this bug, and you can take whatever time you need to fix the bugs in that workflow and its sub-workflows.

So I'd prefer a quick solution that makes all workflows functional. We can always add back the improvements when they are working correctly.

Would you be ok with reverting the changes to `ci-integration-tests-gcp.yml` and the workflows it depends on instead?

gustavovalverde:
> My goal was restoring test coverage as reliably and quickly as possible. This PR does that.

The odd thing here is that those failures are not being reflected in PR or main _checks_
image

As far as I've been able to identify the workflow is getting a failure, but all jobs are ending successfully (so the final check result is a successful run) even on the commit that opened the issue: https://github.com/ZcashFoundation/zebra/commit/d1e72c114faff3eab212c90a21a7258f4e5b11ed 

> Would you be ok with reverting the changes to ci-integration-tests-gcp.yml and the workflows it depends on instead?

Yes, I'd be ok reverting the changes in `sub-deploy-integration-tests-gcp.yml` (which is the sub-workflow being executed), but I'd prefer to try a fix with tomorrow as a deadline, and if I can't fix it, we merge the partial revert.

teor2345:
Ok, happy to leave the fix or the partial revert to you.

teor2345:
We don't need reminders for this, we're still deciding how to fix it.

PR #8134, opened 2024-01-05
Update copyright year to 2024

sandakersmann:
Update copyright year to 2024


teor2345 (approved):
Just for clarity, this is not legally required, but we appreciate the contribution.

teor2345:
It's been a while since we've run our CI on an external PR, so don't worry if it fails. We'll take a look and fix the bug in CI.

PR #8136, opened 2024-01-08
feat(mine): Add an internal Zcash miner to Zebra

teor2345:
## Motivation

It would be easier to test Zebra if we could mine blocks without external mining software.

This PR depends on changes to `equihash`:
https://github.com/zcash/librustzcash/pull/1088
https://github.com/zcash/librustzcash/pull/1083

This is an experimental "free sprint" project.

    - [x] List of features in `zebrad/src/lib.rs` 
    - [x] List of tasks in `zebrad/src/commands/start.rs` 

### Specifications

This code obeys the consensus rules by generating valid blocks, but it uses the existing verifiers. So there are no consensus-critical changes here.

This code launches concurrent async tasks and OS threads. We should check that panics and shutdowns are handled correctly. In particular, miner threads should restart mining with each new block template.

We should check that any common or expected errors are handled by retrying rather than panicking.

We should check that the threads are actually doing *different* work, they should have nonces with large gaps between them.

## Solution

Rust Features:
- add an `internal-miner` feature and dependencies

Solver:
- depend on the new solver code in `equihash`
- add an `equihash::Solution::solve()` method with difficulty checks

Miner:
- add a miner task to the start command
- wait until a valid template is available
- spawn to a blocking async executor thread 
- spawn to a low-priority OS thread
- handle shutdowns by cancelling the miner

Parallel Mining:
- use a different nonce for each thread
- spawn multiple miner threads based on a new config, use `available_parallelism()` by default
- `select! {}` on solutions or new block templates using `long_poll_id`
- cancel and restart the solver if there's a new template
- rate-limit template changes

Related Changes:
- Fix a bug in the `get_block_template` RPC change detection
- make RPC instances cloneable and LongPollId copyable
- use the same generic constraints for the RPC struct and impls
- move some code around so it's easier to re-use
- add utility and convenience methods to existing types

### Testing

I manually verified that:
- the mining threads are running, and there are approximately the configured number of threads
- they are using 100% CPU
- they are low priority (blue in `htop`)
- they produce valid blocks (according to Zebra's verifier)
- the miner runs overnight successfully
- the blocks are mined off the tip
- the blocks are accepted by other nodes
- the [block nonces](https://testnet.zcashblockexplorer.com/blocks/000365cd36db8fcee91bbe4de75eac1a191ee5ede9a915f6f41fb636d7aec9f3) match [the pattern used by Zebra](https://github.com/ZcashFoundation/zebra/pull/8136/files#diff-d73b289b9e87396ee873d64e4bdd5f1d9e57f820adf73df42be03040a3341e89R487-R491), and it's not just the first thread successfully mining 

## Review

This PR is ready for review.

I'd like to focus on bugs and overall code structure in this review.

### Reviewer Checklist

## Follow Up Work

- Get the `equihash` changes merged and released, and update that dependency
- Check we have at least one peer before starting mining (the existing sync to tip check partly covers this)
- Optional: Add a cached state test that mines a block on testnet?

### DIFF
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -61,6 +61,8 @@ impl Solution {
             .zcash_serialize(&mut input)
             .expect("serialization into a vec can't fail");
 
+        // The part of the header before the nonce and solution.
+        // This data is kept constant during solver runs, so the verifier API takes it separately.
         let input = &input[0..Solution::INPUT_LENGTH];
 
         equihash::is_valid_solution(n, k, input, nonce.as_ref(), solution)?;
@@ -116,6 +118,8 @@ impl Solution {
         header
             .zcash_serialize(&mut input)
             .expect("serialization into a vec can't fail");
+        // Take the part of the header before the nonce and solution.
+        // This data is kept constant for this solver run.
         let input = &input[0..Solution::INPUT_LENGTH];
 
         while !is_shutting_down() {
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -109,6 +109,12 @@ impl Solution {
             for solution in solutions {
                 header.solution = Self::from_bytes(solution)
                     .expect("unexpected invalid solution: incorrect length");
+                // TODO: only run this redundant check in tests
+                header
+                    .solution
+                    .check(&header)
+                    .expect("unexpected invalid solution: invalid solution for header");
+
                 if Self::difficulty_is_valid(&header) {
                     return header;
                 }
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -109,7 +109,7 @@ impl Solution {
         let input = &input[0..Solution::INPUT_LENGTH];
 
         while !is_shutting_down() {
-            let solutions = equihash::tromp::solve_200_9_compressed(input, || {
+            let solutions = equihash_solver::tromp::solve_200_9_compressed(input, || {
                 // This skips the first nonce, which doesn't matter in practice.
                 Self::next_nonce(&mut header.nonce);
                 Some(*header.nonce)
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -19,7 +19,6 @@ use crate::{
 pub struct Error(#[from] equihash::Error);
 
 /// The error type for Equihash solving.
-#[non_exhaustive]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, thiserror::Error)]
 #[error("solver was cancelled")]
 pub struct SolverCancelled;
@@ -91,6 +90,8 @@ impl Solution {
     /// Mines and returns one or more [`Solution`]s based on a template `header`.
     /// The returned header contains a valid `nonce` and `solution`.
     ///
+    /// If `cancel_fn()` returns an error, returns early with `Err(SolverCancelled)`.
+    ///
     /// The `nonce` in the header template is taken as the starting nonce. If you are running multiple
     /// solvers at the same time, start them with different nonces.
     /// The `solution` in the header template is ignored.
@@ -99,7 +100,10 @@ impl Solution {
     /// It can run for minutes or hours if the network difficulty is high.
     #[cfg(feature = "internal-miner")]
     #[allow(clippy::unwrap_in_result)]
-    pub fn solve(mut header: Header) -> Result<Header, SolverCancelled> {
+    pub fn solve<F>(mut header: Header, mut cancel_fn: F) -> Result<Header, SolverCancelled>
+    where
+        F: FnMut() -> Result<(), SolverCancelled>,
+    {
         use crate::shutdown::is_shutting_down;
 
         let mut input = Vec::new();
@@ -109,12 +113,22 @@ impl Solution {
         let input = &input[0..Solution::INPUT_LENGTH];
 
         while !is_shutting_down() {
+            // Don't run the solver if we'd just cancel it anyway.
+            cancel_fn()?;
+
             let solutions = equihash_solver::tromp::solve_200_9_compressed(input, || {
+                // Cancel the solver if we have a new template.
+                if cancel_fn().is_err() {
+                    return None;
+                }
+
                 // This skips the first nonce, which doesn't matter in practice.
                 Self::next_nonce(&mut header.nonce);
                 Some(*header.nonce)
             });
 
+            // If we got any solutions, try submitting them, because the new template might just
+            // contain some extra transactions. Mining extra transactions is optional.
             for solution in &solutions {
                 header.solution = Self::from_bytes(solution)
                     .expect("unexpected invalid solution: incorrect length");
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -145,7 +145,7 @@ impl Solution {
                 }
             }
 
-            info!(
+            debug!(
                 solutions = ?solutions.len(),
                 "found valid solutions which did not pass the validity or difficulty checks"
             );
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -21,7 +21,7 @@ pub struct Error(#[from] equihash::Error);
 /// The size of an Equihash solution in bytes (always 1344).
 pub(crate) const SOLUTION_SIZE: usize = 1344;
 
-/// Equihash Solution.
+/// Equihash Solution in compressed format.
 ///
 /// A wrapper around [u8; 1344] because Rust doesn't implement common
 /// traits like `Debug`, `Clone`, etc for collections like array
@@ -60,10 +60,86 @@ impl Solution {
         Ok(())
     }
 
-    #[cfg(feature = "getblocktemplate-rpcs")]
+    /// Returns a [`Solution`] containing the bytes from `solution`.
+    /// Returns an error if `solution` is the wrong length.
+    pub fn from_bytes(solution: Vec<u8>) -> Result<Self, SerializationError> {
+        if solution.len() != SOLUTION_SIZE {
+            return Err(SerializationError::Parse(
+                "incorrect equihash solution size",
+            ));
+        }
+
+        let mut bytes = [0; SOLUTION_SIZE];
+        // Won't panic, because we just checked the length.
+        bytes.copy_from_slice(&solution);
+
+        Ok(Self(bytes))
+    }
+
     /// Returns a [`Solution`] of `[0; SOLUTION_SIZE]` to be used in block proposals.
+    #[cfg(feature = "getblocktemplate-rpcs")]
     pub fn for_proposal() -> Self {
         Self([0; SOLUTION_SIZE])
+    }
+
+    /// Mines and returns one or more [`Solution`]s based on a template `header`.
+    /// The returned header contains a valid `nonce` and `solution`.
+    ///
+    /// The `nonce` in the header template is taken as the starting nonce. If you are running multiple
+    /// solvers at the same time, start them with different nonces.
+    /// The `solution` in the header template is ignored.
+    ///
+    /// This method is CPU and memory-intensive. It uses 144 MB of RAM and one CPU core while running.
+    /// It can run for minutes or hours if the network difficulty is high.
+    #[cfg(feature = "internal-miner")]
+    pub fn solve(mut header: Header) -> Header {
+        let mut input = Vec::new();
+        header
+            .zcash_serialize(&mut input)
+            .expect("serialization into a vec can't fail");
+        let input = &input[0..Solution::INPUT_LENGTH];
+
+        loop {
+            let solutions = equihash::tromp::solve_200_9_compressed(input, || {
+                // This skips the first nonce, which doesn't matter in practice.
+                Self::next_nonce(&mut header.nonce);
+                Some(*header.nonce)
+            });
+
+            for solution in solutions {
+                header.solution = Self::from_bytes(solution)
+                    .expect("unexpected invalid solution: incorrect length");
+                if Self::difficulty_is_valid(&header) {
+                    return header;
+                }
+            }
+        }
+    }
+
+    /// Modifies `nonce` to be the next integer in big-endian order.
+    /// Wraps to zero if the next nonce would overflow.
+    #[cfg(feature = "internal-miner")]
+    fn next_nonce(nonce: &mut [u8; 32]) {
+        let _ignore_overflow = crate::primitives::byte_array::increment_big_endian(&mut nonce[..]);
+    }
+
+    /// Returns `true` if the `nonce` and `solution` in `header` meet the difficulty threshold.
+    ///
+    /// Assumes that the difficulty threshold in the header is valid.
+    #[cfg(feature = "internal-miner")]
+    fn difficulty_is_valid(header: &Header) -> bool {
+        // Simplified from zebra_consensus::block::check::difficulty_is_valid().
+        let difficulty_threshold = header
+            .difficulty_threshold
+            .to_expanded()
+            .expect("unexpected invalid header template: invalid difficulty threshold");
+
+        // TODO: avoid calculating this hash multiple times
+        let hash = header.hash();
+
+        // Note: this comparison is a u256 integer comparison, like zcashd and bitcoin. Greater
+        // values represent *less* work.
+        hash <= difficulty_threshold
     }
 }
 
@@ -109,17 +185,6 @@ impl ZcashSerialize for Solution {
 impl ZcashDeserialize for Solution {
     fn zcash_deserialize<R: io::Read>(mut reader: R) -> Result<Self, SerializationError> {
         let solution: Vec<u8> = (&mut reader).zcash_deserialize_into()?;
-
-        if solution.len() != SOLUTION_SIZE {
-            return Err(SerializationError::Parse(
-                "incorrect equihash solution size",
-            ));
-        }
-
-        let mut bytes = [0; SOLUTION_SIZE];
-        // Won't panic, because we just checked the length.
-        bytes.copy_from_slice(&solution);
-
-        Ok(Self(bytes))
+        Self::from_bytes(solution)
     }
 }
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -118,11 +118,12 @@ impl Solution {
             for solution in &solutions {
                 header.solution = Self::from_bytes(solution)
                     .expect("unexpected invalid solution: incorrect length");
-                // TODO: only run this redundant check in tests
-                header
-                    .solution
-                    .check(&header)
-                    .expect("unexpected invalid solution: invalid solution for header");
+
+                // TODO: work out why we sometimes get invalid solutions here
+                if let Err(error) = header.solution.check(&header) {
+                    info!(?error, "found invalid solution for header");
+                    continue;
+                }
 
                 if Self::difficulty_is_valid(&header) {
                     info!("found valid solution and difficulty");
@@ -132,7 +133,7 @@ impl Solution {
 
             info!(
                 solutions = ?solutions.len(),
-                "found valid solutions which did not pass the difficulty check"
+                "found valid solutions which did not pass the validity or difficulty checks"
             );
         }
 
--- a/zebra-chain/src/work/equihash.rs
+++ b/zebra-chain/src/work/equihash.rs
@@ -12,11 +12,17 @@ use crate::{
     },
 };
 
-/// The error type for Equihash
+/// The error type for Equihash validation.
 #[non_exhaustive]
 #[derive(Debug, thiserror::Error)]
 #[error("invalid equihash solution for BlockHeader")]
 pub struct Error(#[from] equihash::Error);
+
+/// The error type for Equihash solving.
+#[non_exhaustive]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, thiserror::Error)]
+#[error("solver was cancelled")]
+pub struct SolverCancelled;
 
 /// The size of an Equihash solution in bytes (always 1344).
 pub(crate) const SOLUTION_SIZE: usize = 1344;
@@ -62,7 +68,7 @@ impl Solution {
 
     /// Returns a [`Solution`] containing the bytes from `solution`.
     /// Returns an error if `solution` is the wrong length.
-    pub fn from_bytes(solution: Vec<u8>) -> Result<Self, SerializationError> {
+    pub fn from_bytes(solution: &[u8]) -> Result<Self, SerializationError> {
         if solution.len() != SOLUTION_SIZE {
             return Err(SerializationError::Parse(
                 "incorrect equihash solution size",
@@ -71,7 +77,7 @@ impl Solution {
 
         let mut bytes = [0; SOLUTION_SIZE];
         // Won't panic, because we just checked the length.
-        bytes.copy_from_slice(&solution);
+        bytes.copy_from_slice(solution);
 
         Ok(Self(bytes))
     }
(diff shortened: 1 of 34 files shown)


teor2345:
This seems to work on testnet, but it's currently causing orphaned chain forks due to some missing code:
> - cancel and restart the solver if there's a new template

Screenshot 2024-01-08 at 13 42 42

teor2345:
If you ever need to debug a C memory error:
https://askubuntu.com/questions/1349047/where-do-i-find-core-dump-files-and-how-do-i-view-and-analyze-the-backtrace-st
https://stackoverflow.com/questions/50458782/why-do-i-not-get-a-backtrace-when-my-program-has-a-segmentation-fault-even-thoug

```
$ zebrad
...
2024-01-10T00:54:45.837069Z  INFO {zebrad="b729c0f" net="Test"}:{peer=Out("v4redacted:18233")}:msg_as_req{msg="inv"}:inbound:download_and_verify{hash=0026877df340a76c5e858428a68413ea75d600a95dc74e161bc254257db90704}: zebrad::components::inbound::downloads: downloaded and verified gossiped block height=Height(2667601)
2024-01-10T00:54:49.415856Z  INFO {zebrad="b729c0f" net="Test"}:spawn_init:generate_block_templates: zebrad::components::miner: mining with an updated block template height=2667602 transactions=0
Segmentation fault      (core dumped)
$ ulimit -c
unlimited
$ cat /proc/sys/kernel/core_pattern 
|/nix/store/a55d0mc43m0ibl328yxd5dggrr7r4f1h-systemd-250.9/lib/systemd/systemd-coredump %P %u %g %s %t %c %h
$ coredumpctl list
...
Wed 2024-01-10 10:54:55 AEST 3579213 1000 100 SIGSEGV present  /home/dev/zebra/target/release/zebrad                                       5.4G                                                            
$ coredumpctl info 3579213
           PID: 3579213 (zebrad)                                                                                                                                                                                                  
        Signal: 11 (SEGV)
     Timestamp: Wed 2024-01-10 10:54:49 AEST (7min ago)
  Command Line: /home/dev/bin/zebrad -c /home/dev/.config/zebra/zebrad-testnet.toml
    Executable: /home/dev/zebra/target/release/zebrad
       Storage: /var/lib/systemd/coredump/core.zebrad.1000.3c86c05a0ae149aea8afd4adc336ff77.3579213.1704848089000000 (present)
     Disk Size: 5.4G
       Message: Process 3579213 (zebrad) of user 1000 dumped core.
...
```

Looks like C to me:
```
$ coredumpctl debug 3579213
...
                Stack trace of thread 3615860:
                #0  0x00007fa26fab845a free (libc.so.6 + 0x9545a)
                #1  0x000055991d928df9 n/a (/home/dev/zebra/target/release/zebrad + 0x1366df9)
                #2  0x000055991d166c63 n/a (/home/dev/zebra/target/release/zebrad + 0xba4c63)
                ELF object binary architecture: AMD x86-64
...
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa26fab845a in free () from /nix/store/b2hc0i92l22ir2kavnjn3z5z6mzabbvm-glibc-2.34-210/lib/libc.so.6
```

teor2345:
After reducing optimisations and increasing debug info, I couldn't reproduce the error.
https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level

I tried both:
```
$ emacs ~/.cargo/config.toml
[profile.release]
debug = 2
strip = false

opt-level = 0
```
and
```
[profile.release.package.equihash]
opt-level = 0
```

I modified the C code to prevent double-frees, and expose use-after-free. That gave me `free(): invalid pointer` errors. So the bug is somewhere in the C code. I also fixed a memory leak in the C code.

The new debugging output is:
```
$ coredumpctl debug ...
                Stack trace of thread 4112578:                                                       
                #0  0x00007fd03c05f45a free (libc.so.6 + 0x9545a)
                #1  0x000055fa12f2136d n/a (/home/dev/zebra/target/release/zebrad + 0x136d36d)
                #2  0x000055fa129febf3 n/a (/home/dev/zebra/target/release/zebrad + 0xe4abf3)
                #3  0x000055fa11f5a420 n/a (/home/dev/zebra/target/release/zebrad + 0x3a6420)
```

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+        Ok(Self(bytes))
+    }
+
     /// Returns a [`Solution`] of `[0; SOLUTION_SIZE]` to be used in block proposals.
+    #[cfg(feature = "getblocktemplate-rpcs")]
     pub fn for_proposal() -> Self {
         Self([0; SOLUTION_SIZE])
     }
+
+    /// Mines and returns one or more [`Solution`]s based on a template `header`.
```
arya2, line 95:
```suggestion
    /// Mines and returns a [`Solution`] based on a template `header`.
```

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+        header
+            .zcash_serialize(&mut input)
+            .expect("serialization into a vec can't fail");
+        let input = &input[0..Solution::INPUT_LENGTH];
+
+        while !is_shutting_down() {
+            // Don't run the solver if we'd just cancel it anyway.
+            cancel_fn()?;
+
+            let solutions = equihash_solver::tromp::solve_200_9_compressed(input, || {
```
arya2, line 129: Optional/Nitpick:

```suggestion
        while !is_shutting_down() {
            // Don't run the solver if we'd just cancel it anyway.
            cancel_fn()?;

            let solutions = equihash_solver::tromp::solve_200_9_compressed(&input, || {
```

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+                // Cancel the solver if we have a new template.
+                if cancel_fn().is_err() {
+                    return None;
+                }
+
+                // This skips the first nonce, which doesn't matter in practice.
```
arya2, line 135: Optional: Note that it'll eventually try the first nonce (after trying every other nonce)?

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+        while !is_shutting_down() {
+            // Don't run the solver if we'd just cancel it anyway.
+            cancel_fn()?;
+
+            let solutions = equihash_solver::tromp::solve_200_9_compressed(input, || {
+                // Cancel the solver if we have a new template.
+                if cancel_fn().is_err() {
+                    return None;
+                }
+
```
arya2, line 134: Optional: 

```rust

/// Decrements `byte_array` by 1, interpreting it as a big-endian integer.
/// If the big-endian integer overflowed, sets all the bytes to one, and returns `true`.
pub fn decrement_big_endian(byte_array: &mut [u8]) -> bool {
    // Decrement the last byte in the array that is less than u8::MAX, and clear any bytes after it
    // to decrement the next value in big-endian (lexicographic) order.
    let is_wrapped_overflow = byte_array.iter_mut().rev().all(|v| {
        *v = v.wrapping_sub(1);
        v == &u8::MAX
    });

    is_wrapped_overflow
}

/// Modifies `nonce` to be the previous integer in big-endian order.
/// Wraps to max value if the previous nonce would overflow.
#[cfg(feature = "internal-miner")]
fn prev_nonce(nonce: &mut [u8; 32]) {
    let _ignore_overflow = crate::primitives::byte_array::decrement_big_endian(&mut nonce[..]);
}

// above the while loop:
let mut last_nonce = header.nonce;
Self::prev_nonce(&mut last_nonce);
```

```suggestion
                if header.nonce == last_nonce {
                    return None;
                }

```

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+                Some(*header.nonce)
+            });
+
+            // If we got any solutions, try submitting them, because the new template might just
+            // contain some extra transactions. Mining extra transactions is optional.
+            for solution in &solutions {
+                header.solution = Self::from_bytes(solution)
+                    .expect("unexpected invalid solution: incorrect length");
+
+                // TODO: work out why we sometimes get invalid solutions here
```
arya2, line 148: Optional: If checking a solution is significant compared to calculating/mining a solution, leave a TODO for checking `difficulty_is_valid()` first?

zebra-chain/src/work/equihash.rs
```diff
@@ -60,11 +65,120 @@ impl Solution {
...
+    #[cfg(feature = "internal-miner")]
+    fn next_nonce(nonce: &mut [u8; 32]) {
+        let _ignore_overflow = crate::primitives::byte_array::increment_big_endian(&mut nonce[..]);
+    }
+
+    /// Returns `true` if the `nonce` and `solution` in `header` meet the difficulty threshold.
+    ///
+    /// Assumes that the difficulty threshold in the header is valid.
+    #[cfg(feature = "internal-miner")]
+    fn difficulty_is_valid(header: &Header) -> bool {
```
arya2, line 182: Optional/Nitpick: Could we use the same method for this and the block verifier check here: https://github.com/ZcashFoundation/zebra/blob/internal-mine/zebra-consensus/src/block/check.rs#L100

(Or maybe just a reference in the method docs?)

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -585,6 +660,7 @@ where
                 ));
 
                 // Return immediately if the chain tip has changed.
+                latest_chain_tip.mark_best_tip_seen();
```
arya2: I think this method may be marking the best tip as *unseen* (decrementing the watch channel receiver instance's version so the next call to `changed` is immediately ready in most cases).

Update: It may need to still call `borrow_and_update()` [above](https://github.com/ZcashFoundation/zebra/pull/8136/files#diff-c0bec6265a8d68255f80da0e5f499035c5c0181cc218c3d68805094e2c7815f5R579), but this looks correct here.

arya2 (commented):
This looks wonderful so far, thank you!

zebra-state/src/service/watch_receiver.rs
```diff
@@ -99,15 +99,25 @@ where
...
+    /// Returns `true` when the inner value has been updated, even if the old and new values are equal.
+    ///
+    /// Does not mark the watched data as seen.
+    pub fn has_changed(&self) -> Result<bool, watch::error::RecvError> {
+        self.receiver.has_changed()
+    }
+
     /// Marks the watched data as seen.
+    /// Calls [`watch::Receiver::mark_changed()`].
     pub fn mark_as_seen(&mut self) {
```
arya2, line 119: Optional: Could we rename this to match the method it calls on `receiver` like `changed()` and `has_changed()`

```suggestion
    pub fn mark_changed(&mut self) {
```

zebrad/src/commands/start.rs
```diff
@@ -308,6 +309,31 @@ impl StartCmd {
...
+                address_book,
+            );
+
+            crate::components::miner::spawn_init(&config.mining, rpc)
+        };
+
+        #[cfg(not(feature = "internal-miner"))]
+        // Spawn a dummy miner task which doesn't do anything and never finishes.
+        let miner_task_handle: tokio::task::JoinHandle<Result<(), Report>> =
+            tokio::spawn(std::future::pending().in_current_span());
```
arya2, line 347: Just curious, why use pending futures rather than the original miner/scan tasks?

Update: Ah, I missed the `not()`, thank you.

zebra-chain/src/work/equihash.rs
teor2345: Not before the heat death of the universe 🙂

Zcash uses a 256 bit nonce.

zebra-chain/src/work/equihash.rs
teor2345: This is incorrect, because INPUT_LENGTH is shorter than the length of the serialized header. (It excludes the nonce, solution length, and solution.)

I'll add a comment instead.

zebra-chain/src/work/equihash.rs
teor2345: What does this change do, and why is it useful?
I can't evaluate your suggestion without that.

Also, suggestions that are this complex are better in another PR.

zebra-chain/src/work/equihash.rs
teor2345: Actually I think I'll fix the return type here, not the docs.

zebra-chain/src/work/equihash.rs
teor2345: No, because the original method checks the difficult threshold as well, which is unnecessary here. (We assume that the caller specified the threshold correctly in the header template, and if they didn't, the block will get rejected during full validation.)

Also, the error type returned by the original method is only in the state crate.

We could extract out this check into a separate method, but trying to get the APIs to match could be tricky. So it seemed easiest just to copy 2 lines of code and say where they came from.

The reference you asked for is already in a comment on line 169 of this file.

teor2345 (commented):
Thanks for the review, but I think I'll fix some of these a little differently.

zebra-chain/src/work/equihash.rs
teor2345: Validation is *always* much faster than generation in a proof of work system.

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Ah you're right, I didn't read the method docs, and caused this bug myself. I'll go revert those changes.

zebra-state/src/service/watch_receiver.rs
teor2345: This change will be reverted because it's a bug.

We can add a new method if it's ever needed, but we haven't needed it so far.

zebrad/src/commands/start.rs
teor2345: When the feature isn't enabled, the functions that create the miner/scan tasks aren't compiled, so we can't call them.

zebra-chain/src/work/equihash.rs
arya2: > What does this change do

Hypothetically, the idea was to avoid retrying nonces until there's a new template.

Practically, nothing at all.

> and why is it useful?

It's not, I hadn't thought it through in context.

teor2345: Yeah given the size of the nonce space, and that they're arbitrary, it really doesn't matter if we skip a few.

That's why each solver gets initial nonces like:
- [0, 0, ... , 0, 1]
- [1, 0, ... , 0, 2]
- ...
- [254, 0, ... , 0, 255]
- [255, 0, ... , 0, 0]

It's a mess in the data but it's easy to code.

zebra-chain/src/work/equihash.rs
teor2345: Fixed in bccf803f2b

zebra-chain/src/work/equihash.rs
teor2345: Documented in 6e6bc01c19

zebra-rpc/src/methods/get_block_template_rpcs.rs
teor2345: Fixed in 8656076e39

zebra-state/src/service/watch_receiver.rs
teor2345: Fixed in 8656076e39, I added the new method anyway for clarity.

teor2345:
This should be feature-complete. There aren't any new tests, but the existing tests have good coverage, and I tested it manually.

I am just writing the docs now.

teor2345:
Now the docs should be complete as well.

zebrad/src/components/miner.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    // These tasks run forever unless there is a fatal error or shutdown.
+    // When that happens, the first task to error returns, and the other JoinHandle futures are
+    // cancelled.
+    let first_result;
+    select! {
+        result = template_generator => { first_result = result; }
+        result = mining_solvers.try_next() => {
+            first_result = result
+                .transpose()
```
arya2: Nitpick: Use `futures::StreamExt` instead of `TryStreamExt`

```suggestion
        result = mining_solvers.next() => {
            first_result = result
```

zebrad/src/components/miner.rs
```diff
@@ -0,0 +1,527 @@
...
+        + Clone
+        + Send
+        + Sync
+        + 'static,
+    <BlockVerifierRouter as Service<zebra_consensus::Request>>::Future: Send,
+    SyncStatus: ChainSyncStatus + Clone + Send + Sync + 'static,
+    AddressBook: AddressBookPeers + Clone + Send + Sync + 'static,
+{
+    // Shut down the task when the template sender is dropped, or Zebra shuts down.
+    while template_receiver.has_changed().is_ok() && !is_shutting_down() {
```
arya2, line 350:
```suggestion
    // TODO: Add an `is_closed` method and replace the `has_changed().is_ok()` with `!is_closed()`
    while template_receiver.has_changed().is_ok() && !is_shutting_down() {
```

Update: Nevermind, there's no `is_closed()` method on `watch::Receiver`

zebrad/src/components/miner.rs
```diff
@@ -0,0 +1,527 @@
...
+    SyncStatus: ChainSyncStatus + Clone + Send + Sync + 'static,
+    AddressBook: AddressBookPeers + Clone + Send + Sync + 'static,
+{
+    // Pass the correct arguments, even if Zebra currently ignores them.
+    let mut parameters = get_block_template::JsonParameters {
+        mode: Template,
+        data: None,
+        capabilities: vec![LongPoll, CoinbaseTxn],
+        long_poll_id: None,
+        _work_id: None,
```
arya2, line 247: Subjective/Nitpick:

```suggestion
        capabilities: vec![LongPoll, CoinbaseTxn],
        ..Default::default()
```

zebrad/src/components/miner.rs
teor2345: Happy to do this, but I'd like to focus on bugs or wider code structure in this review.

zebrad/src/components/miner.rs
```diff
@@ -0,0 +1,527 @@
...
+                return false;
+            }
+            *old_block = Some(Arc::new(block));
+            true
+        });
+
+        // If the blockchain is changing rapidly, limit how often we'll update the template.
+        // But if we're shutting down, do that immediately.
+        if !template_sender.is_closed() && !is_shutting_down() {
+            sleep(Duration::from_secs(
```
arya2, line 298: Optional: Put this in a `future::join` so it's awaited in parallel to the long poll result, with the rest of the while loop in an async block.

I'll do this in another PR if it sounds good.

zebrad/src/components/miner.rs
teor2345: If you'd like to do this in another PR, go for it!

zebrad/src/components/miner.rs
teor2345: I think it's clearer if we're explicit about the `mode` (because we want a template), and `long_poll_id` (because we change it later).

I would be ok with replacing the other fields with defaults if you submitted another PR.

zebrad/src/components/miner.rs
teor2345: Can you explain why this change would be helpful? Is the difference important enough to complicate the code?

It is usually clearer to split code into another function, rather than assigning an async block to a named variable.

zebrad/src/components/miner.rs
teor2345: > Update: Nevermind, there's no `is_closed()` method on `watch::Receiver`

It would make the code more readable to add an `is_closed()` method to our `WatchReceiver`, where we control the API.

zebrad/src/components/miner.rs
arya2: > Can you explain why this change would be helpful?

If a long polling request returns right before a new block is added to the chain, it could delay mining by a few seconds.

I think we want to check if that the new template's previous block hash is the same as before, or that it's been longer than `GET_BLOCK_TEMPLATE_MEMPOOL_LONG_POLL_INTERVAL` since a new template was sent to the watch channel?

> Is the difference important enough to complicate the code?

I'm hoping it'll be relatively simple, but otherwise no, it's a marginal efficiency gain, and this intended to be used for testing.

> It is usually clearer to split code into another function, rather than assigning an async block to a named variable.

On second thought, a join isn't a great solution here, but if it were, that would be better.

zebrad/src/components/miner.rs
arya2: > It would make the code more readable to add an `is_closed()` method to our `WatchReceiver`, where we control the API.

Yes it will, okay, I'll go for it in another PR :D

zebrad/src/components/miner.rs
teor2345: > > Can you explain why this change would be helpful?
> 
> If a long polling request returns right before a new block is added to the chain, it could delay mining by a few seconds.

Since this code is only supported on testnet, and isn't intended for high-efficiency mining, I think that's ok.

> I think we want to check if that the new template's previous block hash is the same as before,

If you want to do this check to avoid sending useless identical updates, that is implemented here:
https://github.com/ZcashFoundation/zebra/pull/8136/files/1b180e1925762c99d094e3faf536f483d5366590#diff-d73b289b9e87396ee873d64e4bdd5f1d9e57f820adf73df42be03040a3341e89R289

The block hash is the hash of the header, and checking for header equality is actually faster than hashing. (And the block bytes are cryptographically bound to the header, so it's also equivalent to checking for block equality.)

If you want to check if pending solutions should still be sent, that's unlikely to happen, because we submit all solutions as soon as there are one or more solutions from a solver run.

There is a miner hint for sending old blocks in `GetBlockTemplate.submit_old`, and a TODO for it here:
https://github.com/ZcashFoundation/zebra/pull/8136/files/1b180e1925762c99d094e3faf536f483d5366590#diff-d73b289b9e87396ee873d64e4bdd5f1d9e57f820adf73df42be03040a3341e89R438-R440

Since it's rare, and invalid blocks are verified in parallel then correctly rejected, I don't think it's needed.

> or that it's been longer than `GET_BLOCK_TEMPLATE_MEMPOOL_LONG_POLL_INTERVAL` since a new template was sent to the watch channel?

I'm not sure what this means or how it's relevant, sorry!

> > Is the difference important enough to complicate the code?
> 
> I'm hoping it'll be relatively simple, but otherwise no, it's a marginal efficiency gain, and this intended to be used for testing.
> 
> > It is usually clearer to split code into another function, rather than assigning an async block to a named variable.
> 
> On second thought, a join isn't a great solution here, but if it were, that would be better.

A simpler alternative is storing `Instant::now() + GET_BLOCK_TEMPLATE_MEMPOOL_LONG_POLL_INTERVAL` before the RPC call, then using `sleep_until()` here. That could happen in another PR.

zebra-chain/src/work/equihash.rs
arya2: Just curious, why change the last byte?

teor2345: As I wrote in the code comments, I don't want to have to care if the endianness of `next_nonce()` changes, particularly because it's defined in another crate.

It's also helpful to have a record of which solver thread mined each block for testing:
> We should check that the threads are actually doing different work, they should have nonces with large gaps between them.
> 
> the mining threads are running, and there are approximately the configured number of threads
> the [block nonces](https://testnet.zcashblockexplorer.com/blocks/000365cd36db8fcee91bbe4de75eac1a191ee5ede9a915f6f41fb636d7aec9f3) match [the pattern used by Zebra](https://github.com/ZcashFoundation/zebra/pull/8136/files#diff-d73b289b9e87396ee873d64e4bdd5f1d9e57f820adf73df42be03040a3341e89R487-R491), and it's not just the first thread successfully mining

arya2 (approved):
This looks excellent, thank you.

PR #8157, opened 2024-01-12
fix(scan): Fix the RPC tx fetching in the scanning results reader

upbqdn:
## Motivation

Fix the scanning results reader since it started sending RPC requests that Zebra doesn't understand.

### PR Author Checklist

### Specifications

https://github.com/apoelstra/rust-jsonrpc/blob/master/CHANGELOG.md

## Solution

We recently updated the `jsonrpc` dependency from 0.16 to 0.17. The former version used to implicitly send an array of params, even if a single param was passed to the request builder. The new version doesn't do that anymore, but Zebra expects an array, so this commit explicitly passes an array consisting of a single param to the request builder.

### Reviewer Checklist


arya2 (approved):
Thank you!

Issue #7534, opened 2023-09-12
Make it possible to run CI tests locally for developers


oxarbitrage:
It will be nice that the developers could run a similar version of the Zebra upstream CI does on their local machines. There are several examples on why this will be a useful tool to have, some of them are:

- There are changes that one could do to CI that will affect all the pull requests and other part of the systems. - For people that are not very familiar it is risky to experiment in the real live thing.
- Sometimes i see changes to CI settings in PRs that are experiments, trying this or that. We could avoid this noise a bit by being able to test before pushing.
- In general the developers don't use `entrypoint.sh` or docker to run tests locally which is what the CI does, we need a way to run the same thing both places so we can find certain problems easier.

I think we can assume that the developers has (or can easy obtain) an up to date zebra blockchain and a lightwalletd somewhere in the local hard drive.

This can maybe be done by just doing a tutorial, i am not sure, we can discuss more before doing anything.

### Suggestions

- Have a script to pass the ENV variables and/or arguments to run specific tests from CI, and be able to match each Rust test name from CI with the variables.
- Change instructions in acceptance.rs to use the entrypoint.sh script to run tests
- Make CI test IDs and Rust test names the same, pass the ID to the script instead of using environment variables

PR #8175, opened 2024-01-22
ref(docker): Consolidate all tests in the `entrypoint.sh` script

### LINKED ISSUE #7892: change(release): Enable all user features in the `experimental` build

gustavovalverde:
## Motivation

There were some pending TODOs in CI of tests that were not available in out Docker `entrypoint.sh`, which allows more flexibility for running tests using environment variables.

This had an impact on further implementations of #7534
This also sets some ground for #7892 and #7541

- Not all tests are run with the same parameters, and the `run_cargo_test()` function could not be used.

## Solution

- Move the test execution from CI to the `entrypoint.sh`
- Create new execution environment variables for missing tests
- Add an `EXPERIMENTAL_FEATURES` ARG and ENV to the Dockerfile, to allow the execution of tests with Experimental Features.

### Testing

- Tests should give the same output, and also take a similar amount of time to run
- We should confirm _rebuilding_ is not happening on different tests (except, maybe, the Experimental one)

## Review

Anyone from DevOps. @upbqdn had an specific need for the TEST_FEATURES to be changed.

### Reviewer Checklist

## Follow Up Work

This will also allow to build the experimental docker image easier (more of that work in another PR)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -143,7 +143,7 @@ RUN chmod u+x /entrypoint.sh
 # Entrypoint environment variables
 ENV ENTRYPOINT_FEATURES=${ENTRYPOINT_FEATURES}
 # We repeat the ARGs here, so they are available in the entrypoint.sh script for $RUN_ALL_EXPERIMENTAL_TESTS
-ARG EXPERIMENTAL_FEATURES="elasticsearch journald prometheus filter-reload"
+ARG EXPERIMENTAL_FEATURES="shielded-scan journald prometheus filter-reload"
 ENV ENTRYPOINT_FEATURES_EXPERIMENTAL="${ENTRYPOINT_FEATURES} ${EXPERIMENTAL_FEATURES}"
 
 # By default, runs the entrypoint tests specified by the environmental variables (if any are set)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@
 # Keep these argument defaults in sync with GitHub vars.RUST_PROD_FEATURES and vars.RUST_TEST_FEATURES
 # https://github.com/ZcashFoundation/zebra/settings/variables/actions
 ARG FEATURES="default-release-binaries"
-ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints shielded-scan"
+ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints"
 ARG EXPERIMENTAL_FEATURES=""
 
 # This stage implements cargo-chef for docker layer caching
@@ -114,7 +114,7 @@ ARG FEATURES
 ARG TEST_FEATURES
 ARG EXPERIMENTAL_FEATURES
 # ${EXPERIMENTAL_FEATURES} is empty, so this is equivalent to ${FEATURES} ${TEST_FEATURES} by default
-ARG ENTRYPOINT_FEATURES="${FEATURES} ${TEST_FEATURES} ${EXPERIMENTAL_FEATURES}"
+ARG ENTRYPOINT_FEATURES="${FEATURES} ${TEST_FEATURES}"
 
 # Re-hydrate the minimum project skeleton identified by `cargo chef prepare` in the planner stage,
 # over the top of the original source files,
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -16,7 +16,8 @@
 # Keep these argument defaults in sync with GitHub vars.RUST_PROD_FEATURES and vars.RUST_TEST_FEATURES
 # https://github.com/ZcashFoundation/zebra/settings/variables/actions
 ARG FEATURES="default-release-binaries"
-ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints"
+ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints shielded-scan"
```
upbqdn: I think we are better off not including `shielded-scan` for now, actually.

We will use it at some point, for example, when we make it a part of the default features, but we now use it only for a single test and enable it for the test separately.

```suggestion
ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints"
```

gustavovalverde: Solved with [fix(docker): do not add shielded-scan as part of TEST_FEATURES](https://github.com/ZcashFoundation/zebra/pull/8175/commits/c9f5118ee3b7edf5bdf39e9c35d79679d89841a0)

arya2 (dismissed):

docker/Dockerfile
```diff
@@ -140,6 +142,9 @@ RUN chmod u+x /entrypoint.sh
 
 # Entrypoint environment variables
 ENV ENTRYPOINT_FEATURES=${ENTRYPOINT_FEATURES}
+# We repeat the ARGs here, so they are available in the entrypoint.sh script for $RUN_ALL_EXPERIMENTAL_TESTS
+ARG EXPERIMENTAL_FEATURES="elasticsearch journald prometheus filter-reload"
```
arya2:
```suggestion
ARG EXPERIMENTAL_FEATURES="shielded-scan journald prometheus filter-reload"
```

Some tests rely on these features being disabled.

arya2: These are the test failures:
- `config_tests`: expects config fields that aren't set, we should fix this to emit a warning instead of panicking
- `end_of_support_is_checked_at_start`: The logs with these features are unexpected or the timeout becomes too short, it should be enough for this test to pass without extra features
- `ephemeral_missing/existing_directory`/`misconfigured_ephemeral_missing/existing_directory`: the command exits without a kill where the test expected a kill exit, seemingly because it expects the http request for inserting data into elasticsearch to succeed and panics early since there's no elasticsearch server.
- `persistent_mode`/`sync_large_checkpoints_x`/`start_args`/`tracing_endpoint`: Have the same issue as the ephemeral directory tests, the spawned process exits before doing what's expected
    
These tests expect the second zebrad testchild to fail, but either could open port 9200 or fail first: 
- zebra_metrics_conflict
- zebra_rpc_conflict
- zebra_state_conflict
- zebra_tracing_conflict

arya2 (approved):

PR #8178, opened 2024-01-22
feat(grpc): Add initial `Getinfo` grpc

### LINKED ISSUE #8162: feature: Implement `getinfo` method for new gRPC interface
### Motivation

In order to get to grips with the new grpc library we want to implement a grpc server and a `getinfo` method.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

This is part of #7728

oxarbitrage:
## Motivation

We want to get started with the grpc methods by adding a first dummy call that will work as the framework to add other more interesting ones.

Part of https://github.com/ZcashFoundation/zebra/issues/8162

Depend on https://github.com/ZcashFoundation/zebra/pull/8167

## Solution

I followed the hello world example of tonic (https://github.com/hyperium/tonic/blob/master/examples/helloworld-tutorial.md) and adapted to our needs. 

Access to the scanner database will be done by a service at https://github.com/ZcashFoundation/zebra/pull/8185 . We want to add some real info to this method after that.

This is the most basic and smaller piece i was able to build. 

### Testing

No testing was done here, only manual testing.

In the server side the grpc endpoint can be started with:

```
cargo run --bin scanner-grpc-server
```

Then the client can make a call with:

```

$grpcurl -plaintext -import-path ./proto -proto scanner.proto '[::1]:50051' scanner.Scanner/GetInfo
{
  "minSaplingBirthdayHeight": 419200
}
$ 
```

I think we should make tests by using client code at https://github.com/hyperium/tonic/blob/master/examples/helloworld-tutorial.md#writing-our-client

## Review

I think anyone that will be working in the grpc methods we are about to develop can review.

### Reviewer Checklist

## Follow Up Work

- Create test
- Run with zebra instead of separated binary
- Add configuration (endpoint socket/addr, port, etc).
- Decide what to add that is actually useful in the getinfo response.


zebra-grpc/Cargo.toml
```diff
@@ -14,8 +14,19 @@ keywords = ["zebra", "zcash"]
...
+name = "scanner-grpc-server"
+path = "src/server.rs"
+
 [dependencies]
 
 tonic = "0.10.2"
+prost = "0.12.3"
+tokio = { version = "1.35.1", features = ["macros", "rt-multi-thread"] }
 
 zebra-scan = { path = "../zebra-scan", version = "0.1.0-alpha.1" }
```
arya2: I think we should reverse this dependency to use zebra-scan as the entry point for the scanner process.

zebra-grpc/src/server.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct ScannerRPC {}
+
+#[tonic::async_trait]
+impl Scanner for ScannerRPC {
+    async fn get_info(
+        &self,
+        _request: tonic::Request<Empty>,
+    ) -> Result<Response<InfoReply>, Status> {
+        let network = zcash_primitives::consensus::Network::MainNetwork;
+        let storage = Storage::new(&Config::default(), network.into(), true);
```
arya2: We could access `Storage` through a service if we reverse the dependency.

zebra-grpc/Cargo.toml
oxarbitrage: Can you explain the reasons of this ?

zebra-grpc/src/server.rs
oxarbitrage: A service sounds good but i can't see the benefit apart from more work. If the service is inside `zebra-scan` we might need to depend on it anyway? Can you explain more the rationality of the suggested change ?

What i understand is that we will eventually want the grpc server to get started when zebra starts. If it is the grpc calling the scanner or the reverse is the same to me right now but i will like to see what i am not seeing here and make a team decision.

I will be happy to make the change if that is what we decide to do.

arya2: > A service sounds good but i can't see the benefit apart from more work. If the service is inside zebra-scan we might need to depend on it anyway? Can you explain more the rationality of the suggested change ?

I think it'll help with maintainability, using zebra-scan without zebra-grpc (if users want to read results directly from the db, running zebra-scan in its own process, or adding another API), separating concerns so we can work on features in parallel more easily, and could make it easier to reference past design decisions when making new ones because it's more similar to zebra-rpc.

We could use generics to access `Storage` (by implementing a trait that's defined in `zebra-grpc` or a common dependency, then passing an implementation of that trait to an `init()` or `start_server()` function in `zebra-grpc`).

> What i understand is that we will eventually want the grpc server to get started when zebra starts. If it is the grpc calling the scanner or the reverse is the same to me right now but i will like to see what i am not seeing here and make a team decision.

Both should work, and if zebra-scan depends on zebra-grpc, we could define traits instead of implementing the `Service` trait. I think it's a good separation of concerns to use zebra-scan as an entrypoint that spawns its own API server, and that using the `Service` trait is simpler than defining our own.

zebra-grpc/Cargo.toml
arya2: To avoid closely coupling zebra-scan/zebra-grpc so that zebra-scan is more extensible, and the reasons in [the comment below ](https://github.com/ZcashFoundation/zebra/pull/8178#discussion_r1462346185).

zebra-grpc/src/server.rs
oxarbitrage: Ok, fair enough, thanks. I think we can do this with little changes to this PR. I made the mods at https://github.com/ZcashFoundation/zebra/pull/8178/commits/92c1b7d731f20701e633dcdf128a818cf4a63381

oxarbitrage (commented):
Looks good, i am not really sure if the failing test is related to this PR. https://github.com/ZcashFoundation/zebra/actions/runs/7661775200/job/20881794125?pr=8178

arya2 (approved):
Looks good!

Issue #8180, opened 2024-01-22
Remove or comment out zebra internal miner


oxarbitrage:
In https://github.com/ZcashFoundation/zebra/pull/8136 we added an internal miner to zebra that depends on changes made to the ECC `equihash` crates. The problem is that this changes are currently in the form of pull requests, unpublished in crates.io. For this reason the release publishing to crates.io of Zebra is failing as all dependencies must be published. More info in https://github.com/ZcashFoundation/zebra/pull/8165#issuecomment-1902616143

We need to fix this in order to complete the current ongoing release (1.5.1) which was made in github but was not published to crates.io.

We decided to remove the internal miner code as a temporal solution until the needed PRs (https://github.com/zcash/librustzcash/pull/1088 and https://github.com/zcash/librustzcash/pull/1083) are merged and a new version is published. 

This ticket is to:
- [ ] remove the dependency and related code that depends on it.
- [x] open a ticket to restore once stuff gets merged and new version published.

## Follow up work

Create a hotfix release (1.5.2) with the changes done.

PR #8187, opened 2024-01-23
chore: Hotfix v1.5.2

### LINKED ISSUE #8181: Create a 1.5.2 hotfix release
We had problems with our last release (https://github.com/ZcashFoundation/zebra/pull/8165#issuecomment-1902616143)

- [ ] After we do https://github.com/ZcashFoundation/zebra/issues/8180 we need to create a 1.5.2 release and finish the release publishing process with that.
- [ ] Document removal of the feature in the hotfix changelog.

Blocked by https://github.com/ZcashFoundation/zebra/issues/8180

oxarbitrage:
## Motivation

This is a hotfix release. We had problems with the crates.io publishing (https://github.com/ZcashFoundation/zebra/issues/8180).

Close https://github.com/ZcashFoundation/zebra/issues/8181

It bumps everything to a new version instead of just some dependencies. This might not be needed but i think it can avoid future confusion.

## Follow up work

- [ ] Publish the new crate versions to crates.io after this get merged into main.
- [ ] Create the hotfix release in github.

### DIFF
--- a/zebra-test/Cargo.toml
+++ b/zebra-test/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zebra-test"
-version = "1.0.0-beta.33"
+version = "1.0.0-beta.34"
 authors = ["Zcash Foundation <zebra@zfnd.org>"]
 description = "Test harnesses and test vectors for Zebra"
 license = "MIT OR Apache-2.0"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to Zebra are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org).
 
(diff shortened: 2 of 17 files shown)


zebra-test/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zebra-test"
-version = "1.0.0-beta.33"
+version = "1.0.0-beta.34"
```
arya2, line 3: Optional:

We could skip bumping the version for this crate since there haven't been any code or dependency changes.

It's also reasonable to bump the version to keep it consistent with the other crates.

This PR looks good either way.

```suggestion
version = "1.0.0-beta.33"
```

arya2 (approved):
Looks good!

zebra-test/Cargo.toml
oxarbitrage: Yea, i decided to update everything because:
- The command in the release drafter is one to update all the betas.
- I think we want to keep all the betas the same number.
- With that , we could still skip the upgrade of `tower-*` crates but i think it might be confusing to upgrade some crates and not others.

Also for future hotfixes i think we should always bump everything or change the drafter.

PR #8192, opened 2024-01-24
fix(CI): Pin `cargo-deny-action` to v1.5.10

upbqdn:
## Motivation

We use `cargo-deny-action@v1` in CI, which currently defaults to v1.5.11. This version of the action fails with this error: https://github.com/ZcashFoundation/zebra/actions/runs/7639791083/job/20814504342?pr=8185#step:5:14.

## Solution

- Pin `cargo-deny-action` to v1.5.10.

### Reviewer Checklist

## Follow Up Work

Revert this PR by addressing #8193.

### DIFF
--- a/.github/workflows/ci-unit-tests-os.yml
+++ b/.github/workflows/ci-unit-tests-os.yml
@@ -255,7 +255,7 @@ jobs:
       - uses: r7kamura/rust-problem-matchers@v1.4.0
 
       - name: Check ${{ matrix.checks }} with features ${{ matrix.features }}
-        uses: EmbarkStudios/cargo-deny-action@v1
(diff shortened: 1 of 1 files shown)


oxarbitrage:
Do you think we should open a ticket in https://github.com/EmbarkStudios/cargo-deny-action . It seems to be a regression. We might want to get back to what we had if they fix. If so, we should open a ticket in zebra to follow up.

upbqdn:
Yes, I've already reported the issue upstream, check #8193.

upbqdn:
(I opened this PR first so I would get its number, and then I opened its reciprocal issue and linked them together. I wonder if there's a better way to do it, for example, by being able to open a linked issue and PR together.)

arya2 (approved):
Thank you!

PR #8198, opened 2024-01-25
add(chain): Adds `ViewingKey` type in zebra-chain

arya2:
## Motivation

This PR adds a `ViewingKey` type to `zebra-chain` behind the `shielded-scan` feature for parsing viewing keys into.

Safely hashing viewing keys may need special attention.

## Solution

Adds:
- `shielded-scan` feature to zebra-chain to depend on `zcash_client_backend`
- `ViewingKey`, `SaplingViewingKey`, and `OrchardViewingKey` types
- `parse()` method to `ViewingKey` for decoding keys

### Testing

- Tests that zecpages sapling efvk parses successfully

## Review

Anyone can review.

### Reviewer Checklist


zebra-chain/src/primitives/viewing_key.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl From<DiversifiableFullViewingKey> for ViewingKey {
+    fn from(value: DiversifiableFullViewingKey) -> Self {
+        Self::DiversifiableFullViewingKey(Box::new(value))
+    }
+}
+
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq)]
+/// The hash of a viewing key for use as an identifier and for authorizing remote access
```
upbqdn: ###### Optional Suggestion

Let's swap the attribute and comment as per https://github.com/rust-lang/style-team/issues/72.

```suggestion
/// The hash of a viewing key for use as an identifier and for authorizing remote access
#[derive(Debug, PartialOrd, Ord, PartialEq, Eq)]
```

zebra-chain/src/primitives/viewing_key.rs
```diff
@@ -0,0 +1,119 @@
...
+pub mod sapling;
+
+#[cfg(test)]
+mod tests;
+
+const KEY_HASH_BYTE_SIZE: usize = 32;
+
+#[derive(Debug, Clone)]
+/// A Zcash Sapling or Orchard viewing key
+// TODO: Add Orchard types and any other Sapling key types
```
upbqdn: ###### Optional Suggestion

Let's swap the attribute and comment as per https://github.com/rust-lang/style-team/issues/72.

```suggestion
/// A Zcash Sapling or Orchard viewing key
// TODO: Add Orchard types and any other Sapling key types
#[derive(Debug, Clone)]
```

zebra-chain/src/primitives/viewing_key.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Returns an encoded byte representation of the viewing key
+    pub fn to_bytes(&self) -> Vec<u8> {
+        match self {
+            Self::SaplingIvk(ivk) => ivk.to_repr().to_vec(),
+            Self::DiversifiableFullViewingKey(dfvk) => dfvk.to_bytes().to_vec(),
+        }
+    }
+
+    /// Parses an encoded viewing key and returns it as a [`ViewingKey`] type.
+    pub fn parse(
```
upbqdn: The doc comment says that the fn parses a viewing key, but since the fn relies on https://github.com/zcashfoundation/zebra/blob/297576f78ce1135d6db17d6e2f8b03d387672ff7/zebra-chain/src/primitives/viewing_key/sapling.rs#L32, it won't parse https://github.com/zcashfoundation/zebra/blob/297576f78ce1135d6db17d6e2f8b03d387672ff7/zebra-chain/src/primitives/viewing_key.rs#L20-L21.

upbqdn (commented):
@arya2 could you please describe how we are going to use the key hashes?

zebra-chain/src/primitives/viewing_key.rs
arya2: ~I don't think the `scan_block()` function in `zcash_client_backend` currently supports incoming viewing keys yet, I'll remove this for now and replace it with fvk / vk variants.~

I was wrong, `ScanningKey` is implemented for `SaplingIvk`.

arya2: Added more parse methods in https://github.com/ZcashFoundation/zebra/pull/8198/commits/b1d9723844be8cc42628a72beea0d7575ab8925b, mostly still unimplemented.

arya2:
> please describe how we are going to use the key hashes?

I added to its docs that it's for authorizing access via the gRPC interface, it's so that the client doesn't have to keep re-sending their viewing keys once they're registered in the scanner.

zebra-chain/src/primitives/viewing_key.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Accepts an encoded Orchard viewing key to decode
+    ///
+    /// Returns a [`ViewingKey`] if successful, or None otherwise
+    fn parse_orchard(sapling_key: &str, network: Network) -> Option<Self> {
+        OrchardViewingKey::parse(sapling_key, network).map(Self::Orchard)
+    }
+
+    /// Parses an encoded viewing key and returns it as a [`ViewingKey`] type.
+    pub fn parse(key: &str, network: Network) -> Option<Self> {
+        // TODO: Try types with prefixes first if some don't have prefixes?
```
arya2:
```suggestion
```

zebra-chain/Cargo.toml
```diff
@@ -34,6 +34,12 @@ getblocktemplate-rpcs = [
     "zcash_address",
 ]
 
+# Experimental shielded scanning support
+shielded-scan = [
+    "zcash_client_backend"
+]
+
+
```
arya2:
```suggestion
```

zebra-chain/src/primitives/viewing_key/sapling.rs
```diff
@@ -0,0 +1,120 @@
...
+            Self::Dfvk(dfvk) => dfvk.to_bytes().to_vec(),
+        }
+    }
+
+    /// Accepts an encoded Sapling extended full viewing key to decode
+    ///
+    /// Returns a [`SaplingViewingKey::Dfvk`] if successful, or None otherwise
+    fn parse_extended_full_viewing_key(sapling_key: &str, network: Network) -> Option<Self> {
+        decode_extended_full_viewing_key(network.sapling_efvk_hrp(), sapling_key)
+            .ok()
```
arya2, line 33:
```suggestion
            // this should fail often, so a debug-level log is okay
            .map_err(|err| debug!(?err, "could not decode Sapling extended full viewing key"))
            .ok()
```

zebra-chain/src/primitives/viewing_key.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub use hash::{ViewingKeyHash, ViewingKeyWithHash};
+use orchard::OrchardViewingKey;
+use sapling::SaplingViewingKey;
+
+#[cfg(test)]
+mod tests;
+
+/// A Zcash Sapling or Orchard viewing key
+// TODO: Add Orchard types and any other Sapling key types
```
arya2:
```suggestion
```

zebra-chain/src/primitives/viewing_key/tests.rs
```diff
@@ -0,0 +1,42 @@
+//! Tests for Zebra history trees
```
arya2:
```suggestion
//! Tests for zebra-chain viewing key hashes
```

upbqdn (commented):
I still struggle to see the advantages of hashing the keys instead of using them directly. What would be worse if we didn't hash the keys?

arya2:
> I still struggle to see the advantages of hashing the keys instead of using them directly. What would be worse if we didn't hash the keys?

It could be useful if the private key for the TLS encryption were compromised.

zebra-chain/src/primitives/viewing_key/sapling.rs
```diff
@@ -0,0 +1,122 @@
...
+    zip32::DiversifiableFullViewingKey as SaplingDfvk,
+};
+
+use crate::parameters::Network;
+
+/// A Zcash Sapling viewing key
+#[derive(Debug, Clone)]
+pub enum SaplingViewingKey {
+    /// A viewing key for Sapling
+    Vk(Box<SaplingVk>),
```
upbqdn: ###### Optional Suggestion

I think we shouldn't need this key as it is only a building block for `SaplingFvk`, and users shouldn't have access to it.

upbqdn (dismissed):
Looks great.

We'll need tests that check the "to bytes" and "from bytes" round-trips, but we can do them later.

zebra-chain/src/primitives/viewing_key/sapling.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8198/commits/2dee022f87e1c065cf952b451ff42f406eece404

upbqdn (approved):

PR #8215, opened 2024-01-30
feat(docker): allow users to use Zebra + LWD with persistent states

### LINKED ISSUE #8023: doc: Explain how to use a persistent state disk with Docker
### Motivation

A user was surprised when their Docker run stopped and they couldn't find Zebra's state:
> Hi guys, do you know how can I resume Zebrad's sync? I've installed it trough docker and stopped at block 550000. But now I can't find Zebra's files/folders.

https://discord.com/channels/809218587167293450/809251029673312267/1179106203868090378

It would be helpful to tell users how to set up a persistent state disk in the Docker guide:
https://github.com/ZcashFoundation/zebra/blob/main/book/src/user/docker.md

If we're going to change any configured paths inside the Docker image, we might want to do that first.

### Specifications

https://docs.docker.com/storage/volumes/

_No response_

### Testing

Manually run the commands and check that the state is persistent.

### Related Work

_No response_

gustavovalverde:
## Motivation

We want to make it easy for users to use Zebra and Lightwalletd without the risk of losing their cached states, and avoiding complexity when running both tools together.

Fixes #8023

> Thanks to @AArnott for putting up [this guide](https://blog.nerdbank.net/2022/12/hosting-zcash-litewalletd-via-zebra), as it served as the foundation for building these compose files.

- A Health-check was needed for Zebra, so the Lightwalletd service doesn't start until Zebra is ready
- The default LWD has to be run **unsecurely** with `--no-tls-very-insecure` as we can't create certificates for the user
- `curl` is required in our production image to allow the health-check to work

## Solution

- Create a `docker-compose.yml` to run an isolated Zebra node
- Create a `docker-compose.lwd.yml` to run a lightwalletd server alongside Zebra
- Create a `zcash.conf` file for lightwalletd
- Move unused _production_ environment variables from `test.env` to `.env`

### Testing

This can be tested running:
```shell
docker-compose -f docker/docker-compose.yml up
```
For a Zcash Node, and change its behavior in `.env`

```shell
docker-compose -f docker/docker-compose.yml -f docker/docker-compose.lwd.yml up
``` 
To run LWD + Zcash

## Review

**Note:** This is not ready until respective user documentation is updated 

### Reviewer Checklist


docker/docker-compose.lwd.yml
```diff
@@ -0,0 +1,61 @@
...
+    ports:
+      # Zebra uses the following inbound and outbound TCP ports
+      - "8233:8233"
+      - "8232:8232"
+    healthcheck:
+      start_period: 1m
+      interval: 15s
+      timeout: 10s
+      retries: 3
+      test: ["CMD-SHELL", "curl --data-binary '{\"jsonrpc\": \"1.0\", \"id\":\"curltest\", \"method\": \"getinfo\", \"params\": [] }' -H 'content-type: application/json' http://127.0.0.1:8232/ || exit 1"]
```
AArnott: Shouldn't this health-check be on *zebra* rather than the lightwalletd service?
Note, this file _says_ it creates a service called `zebra`, but in fact it defines the lightwalletd service) as shown below. So is this healthcheck simply on the wrong file?

docker/docker-compose.lwd.yml
```diff
@@ -0,0 +1,61 @@
+version: "3.8"
+
+services:
+  zebra:
+    ports:
+      # Zebra uses the following inbound and outbound TCP ports
+      - "8233:8233"
+      - "8232:8232"
```
AArnott: These are zebra ports, but this file actually describes the lightwalletd service.

docker/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    #! NOTE: This will override the zebrad.toml in the image and make some variables irrelevant
+    # configs:
+    #   - source: zebra_config
+    #     target: /etc/zebrad/zebrad.toml
+    #     uid: '2001' # Rust's container default user uid
+    #     gid: '2001' # Rust's container default group gid
+    #     mode: 0440
+    volumes:
+      - zebrad-cache:/var/cache/zebrad-cache
+      - lwd-cache:/var/cache/lwd-cache
```
AArnott: Why does this file (which apparently describes the zebra service) include information about the lwd cache? Shouldn't that be in the lwd yml file?

AArnott (commented):
I'm confused at how you've created two docker-compose files and how you've distributed the yaml across them. In my blog post I used just one file. I've never seen it broken out like this, so I'm probably just ignorant and my comments may be wrong. But hopefully at least one of us will come out having learned something. :)

docker/docker-compose.lwd.yml
gustavovalverde: Before these lines you'll see this:
```yaml
services:
  zebra:
```

This indicates that, when running this YAML, it will override the Zebra service with these values, so the healthcheck is actually added to the Zebra service

docker/docker-compose.lwd.yml
gustavovalverde: Same as https://github.com/ZcashFoundation/zebra/pull/8215#discussion_r1473431847

docker/docker-compose.yml
gustavovalverde: You're actually right, this is just needed for our CI

gustavovalverde:
> I'm confused at how you've created two docker-compose files and how you've distributed the yaml across them

@AArnott that's a really good question, this is a docker compose functionality https://docs.docker.com/compose/multiple-compose-files/merge/

As our users might have different needs, we've created 2 docker-compose yaml:

1. `docker-compose.yml` which runs the Zebra node completely isolated, and
2. `docker-compose.lwd.yml` which inherits and overrides the first one

A user might decide to just run the Zebra node with `docker-compose -f docker/docker-compose.yml up` or running Zebra + LWD together, with `docker-compose -f docker/docker-compose.yml -f docker/docker-compose.lwd.yml up` using both files.

docker/docker-compose.yml
gustavovalverde: Fixed here: https://github.com/ZcashFoundation/zebra/pull/8215

docker/docker-compose.lwd.yml
AArnott: Interesting. Why not just add them to the zebra service directly?

AArnott:
> a user might decide to just run the Zebra node

That sounds good. Then why are zebra-specific things only put into the lwd file? A zebra-only container should still know its own ports and healthcheck, shouldn't it?

gustavovalverde:
> A zebra-only container should still know its own ports and healthcheck, shouldn't it?

@AArnott  Yes, I did this as the healtcheck depends on port `8232` (the RPC endpoint) to be opened. But we don't always want this port opened **by default** on a Zebra node. Though it's required if we want to have lightwalletd.

AArnott (approved):

docker/.env
```diff
@@ -0,0 +1,34 @@
...
+ZEBRA_FORCE_USE_COLOR=1
+LOG_COLOR=true
+
+###
+# Configuration Variables
+# These variables are used to configure the zebra node
+# Check the entrypoint.sh script for more details
+###
+
+# Path and name of the config file. These two have defaults set in the Dockerfile.
```
upbqdn: ###### Optional Suggestion

```suggestion
# The config file full path used in the Dockerfile.
```

docker/.env
```diff
@@ -0,0 +1,34 @@
...
+ZEBRA_CHECKPOINT_SYNC=true
+# [state]
+# Set this to change the default cached state directory
+ZEBRA_CACHED_STATE_DIR=/var/cache/zebrad-cache
+LIGHTWALLETD_DATA_DIR=/var/cache/lwd-cache
+# [metrics]
+METRICS_ENDPOINT_ADDR=0.0.0.0
+METRICS_ENDPOINT_PORT=9999
+# [tracing]
+LOG_COLOR=false
```
upbqdn: `LOG_COLOR` is set to `true` at line 4.

```suggestion
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    # configs:
+    #   - source: zebra_config
+    #     target: /etc/zebrad/zebrad.toml
+    #     uid: '2001' # Rust's container default user uid
+    #     gid: '2001' # Rust's container default group gid
+    #     mode: 0440
+    volumes:
+      - zebrad-cache:/var/cache/zebrad-cache
+    ports:
+      # Zebra uses the following inbound and outbound TCP ports
```
upbqdn: ###### Optional Suggestion

```suggestion
      # Zebra uses the following default inbound and outbound TCP ports
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      target: runtime
+    restart: unless-stopped
+    deploy:
+      resources:
+        reservations:
+          cpus: "4"
+          memory: 16G
+    env_file:
+      - .env
+    #! Uncomment the following line to use a zebrad.toml from the host machine
```
upbqdn:
```suggestion
    #! Uncomment the `configs` mapping below to use the `zebrad.toml` config file from the host machine
```

docker/docker-compose.lwd.yml
```diff
@@ -0,0 +1,59 @@
+version: "3.8"
+
+services:
+  zebra:
+    ports:
+      - "8232:8232" # Opens an RPC endpoint (for wallet storing and mining)
```
upbqdn: ###### Optional Suggestion

```suggestion
      - "8232:8232" # Opens an RPC endpoint (for lightwalletd and mining)
```

docker/docker-compose.lwd.yml
```diff
@@ -0,0 +1,59 @@
...
+    healthcheck:
+      start_period: 1m
+      interval: 15s
+      timeout: 10s
+      retries: 3
+      test: ["CMD-SHELL", "curl --data-binary '{\"jsonrpc\": \"1.0\", \"id\":\"curltest\", \"method\": \"getinfo\", \"params\": [] }' -H 'content-type: application/json' http://127.0.0.1:8232/ || exit 1"]
```
upbqdn: ###### Optional Suggestion

We can simplify the `curl` command:

```suggestion
      test: ["CMD-SHELL", "curl --data-binary '{\"id\":\"curltest\", \"method\": \"getinfo\"}' -H 'content-type: application/json' 127.0.0.1:8232 || exit 1"]
```

upbqdn (dismissed):

upbqdn (approved):

PR #8217, opened 2024-01-31
add(scan): Implement `DeleteKeys` `ScanService` request

### LINKED ISSUE #8204: Implement `DeleteKeys` scan service request to delete registered keys and any cached results for those keys
## Motivation

We want to clear cached results and stop scanning for a key with a request to the scan service

## Possible Design

- Send a `ScanTaskCommand::RemoveKeys {}` message to the scanner task
  - Remove the key from the `parsed_keys` collection in the scan task and send a message notifying the caller that it's been removed
- Wait for the scan task to confirm that the key has been removed
- Implement and call a method on Storage to delete results for a key

arya2:
## Motivation

This PR implements the `DeleteKeys` request for `ScanService` to:
- drop keys from the scanner task
- delete keys and their results from the database

Closes #8204.

## Solution

- Updates request/response variant types
- Adds a new argument to `scan::start()` for the scan task channel receiver
- Adds `ScanTask::remove_keys()` method for sending a `RemoveKeys` message
- Adds `ScanTask::process_msgs()` and calls it from the loop in `scan::start()`
- Adds a `delete_sapling_keys()` method to `Storage`
- Calls `remove_keys()` and `delete_sapling_keys()` from the `ScanService` match arm for `Request::DeleteKeys`

### Testing

- Tests that the new `Storage::delete_sapling_results()` method deletes all results for a provided key
- Tests that the ScanService sends a message to the `ScanTask` to remove the keys and that it deletes the results before responding

## Review

Anyone can review. 

### Reviewer Checklist

## Follow Up Work

- Add a test for `ScanTask::process_msgs()`

### DIFF
--- a/zebra-scan/src/init.rs
+++ b/zebra-scan/src/init.rs
@@ -1,6 +1,9 @@
 //! Initializing the scanner.
 
-use std::sync::{mpsc, Arc};
+use std::sync::{
+    mpsc::{self},
+    Arc,
+};
 
 use color_eyre::Report;
 use tokio::{sync::oneshot, task::JoinHandle};
@@ -23,7 +26,7 @@ pub enum ScanTaskCommand {
         done_tx: oneshot::Sender<()>,
 
         /// Key hashes that are to be removed
-        key_hashes: Vec<()>,
+        keys: Vec<String>,
     },
 
     /// Start sending results for key hashes to `result_sender`
@@ -43,7 +46,7 @@ pub struct ScanTask {
     pub handle: JoinHandle<Result<(), Report>>,
 
     /// Task command channel sender
-    cmd_sender: mpsc::Sender<ScanTaskCommand>,
+    pub cmd_sender: mpsc::Sender<ScanTaskCommand>,
 }
 
 impl ScanTask {
@@ -65,10 +68,10 @@ impl ScanTask {
         chain_tip_change: ChainTipChange,
     ) -> Self {
         // TODO: Pass `_cmd_receiver` to `scan::start()` to pass it new keys after it's been spawned
-        let (cmd_sender, _cmd_receiver) = mpsc::channel();
+        let (cmd_sender, cmd_receiver) = mpsc::channel();
 
         Self {
-            handle: scan::spawn_init(config, network, state, chain_tip_change),
+            handle: scan::spawn_init(config, network, state, chain_tip_change, cmd_receiver),
             cmd_sender,
         }
     }
@@ -90,8 +93,8 @@ pub fn spawn_init(
     network: Network,
     state: scan::State,
     chain_tip_change: ChainTipChange,
-) -> JoinHandle<Result<(), Report>> {
-    scan::spawn_init(config, network, state, chain_tip_change)
+) -> ScanTask {
+    ScanTask::spawn(config, network, state, chain_tip_change)
 }
 
 /// Initialize [`ScanService`] based on its config.
--- a/zebra-scan/src/init.rs
+++ b/zebra-scan/src/init.rs
@@ -39,11 +39,11 @@ pub enum ScanTaskCommand {
     },
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 /// Scan task handle and command channel sender
 pub struct ScanTask {
     /// [`JoinHandle`] of scan task
-    pub handle: JoinHandle<Result<(), Report>>,
+    pub handle: Arc<JoinHandle<Result<(), Report>>>,
 
     /// Task command channel sender
     pub cmd_sender: mpsc::Sender<ScanTaskCommand>,
@@ -55,7 +55,7 @@ impl ScanTask {
         let (cmd_sender, _cmd_receiver) = mpsc::channel();
 
         Self {
-            handle: tokio::spawn(std::future::pending()),
+            handle: Arc::new(tokio::spawn(std::future::pending())),
             cmd_sender,
         }
     }
@@ -71,7 +71,13 @@ impl ScanTask {
         let (cmd_sender, cmd_receiver) = mpsc::channel();
 
         Self {
-            handle: scan::spawn_init(config, network, state, chain_tip_change, cmd_receiver),
+            handle: Arc::new(scan::spawn_init(
+                config,
+                network,
+                state,
+                chain_tip_change,
+                cmd_receiver,
+            )),
             cmd_sender,
         }
     }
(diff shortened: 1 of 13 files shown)


zebra-scan/src/init.rs
```diff
@@ -65,10 +71,16 @@ impl ScanTask {
         chain_tip_change: ChainTipChange,
     ) -> Self {
         // TODO: Pass `_cmd_receiver` to `scan::start()` to pass it new keys after it's been spawned
```
oxarbitrage:
```suggestion
```

Is this TODO outdated ? We are passing `cmd_receiver` to `scan_spawn_init` now.

zebra-scan/src/init.rs
```diff
@@ -23,7 +26,7 @@ pub enum ScanTaskCommand {
         done_tx: oneshot::Sender<()>,
 
         /// Key hashes that are to be removed
-        key_hashes: Vec<()>,
+        keys: Vec<String>,
```
oxarbitrage, line 26: Should we use `ViewingKey` introduced in https://github.com/ZcashFoundation/zebra/pull/8198/ and in other places or `String` is good enough ?

zebra-scan/src/init.rs
```diff
@@ -36,25 +39,28 @@ pub enum ScanTaskCommand {
...
-    pub handle: JoinHandle<Result<(), Report>>,
+    pub handle: Arc<JoinHandle<Result<(), Report>>>,
 
     /// Task command channel sender
-    cmd_sender: mpsc::Sender<ScanTaskCommand>,
+    pub cmd_sender: mpsc::Sender<ScanTaskCommand>,
 }
 
 impl ScanTask {
     /// Spawns a new [`ScanTask`] for tests.
```
oxarbitrage, line 50: If this is only used in tests i was wondering if we should add a `#[cfg(test)]` or similar to it. I understand this is used in the binary too so it might complicate things to do that now.

zebra-scan/src/scan.rs
```diff
@@ -130,6 +137,56 @@ pub async fn start(
...
+
+                Err(TryRecvError::Empty) => break,
+                Err(TryRecvError::Disconnected) => {
+                    // Return early if the sender has been dropped.
+                    return Err(eyre!("command channel disconnected"));
+                }
+            };
+
+            match cmd {
+                ScanTaskCommand::RemoveKeys { done_tx, keys } => {
```
oxarbitrage, line 167: Should we add a limit on the number of keys users can send for deletion in a single call ? Depending in the use case this limit can be greater so i think this can be configurable but for now a const limit for all the keys that can be in any of this calls will be good enough.

zebra-scan/src/service/tests.rs
```diff
@@ -0,0 +1,79 @@
+//! Tests for ScanService.
```
oxarbitrage, line 1: This tests are great, thank you creating them. One thing i noticed is that we can run this tests without the `shielded-scan` feature flag. Is this something we might want to change?

zebra-scan/src/service/tests.rs
```diff
@@ -0,0 +1,79 @@
...
+    assert!(
+        scan_service
+            .db
+            .sapling_results(&zec_pages_sapling_efvk)
+            .is_empty(),
+        "all results for this key should have been deleted"
+    );
+
+    Ok(())
+}
```
oxarbitrage, line 79: I will like to add more tests or maybe just a `TODO` in this PR to delete some keys but make sure other results and keys remain there. The current test add a key and some results and delete them. We also need to make sure that when we have multiple stuff, only what we want is deleted.

oxarbitrage (dismissed):
I think this looks really good. I left some of comments and questions but i don't think any of those are blockers.

zebra-scan/src/init.rs
arya2: I think a string is good enough for now. The db methods assume a constant length for the encoded keys, so we may want to use ids or hashes later. We could also serialize the keys into a fixed size to continue using them as the first part of the key.

zebra-scan/src/init.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8217/commits/0e1061d5ebd95e72dd6ed5f69bcd72038489d58e.

oxarbitrage: Thanks, that looks good.

zebra-scan/src/service/tests.rs
arya2: I think it's okay for the `zebra-scan` unit tests to always run and for the `shielded-scan` feature to enable integration tests with `zebrad`.

zebra-scan/src/service/tests.rs
arya2: Updated the test in https://github.com/ZcashFoundation/zebra/pull/8217/commits/dc4e8dd7548f5e7bb851bb1d6671dc86aadbdb89 and https://github.com/ZcashFoundation/zebra/pull/8217/commits/b406ef78743aaa1b160328d3077041c767735b18 to add results for a second key and check that there all of the expected entries/results are there after the first key has been deleted.

zebra-scan/src/scan.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8217/commits/d35ad90fd2feda7d635bf9c11af99202f3d71617.

oxarbitrage (approved):
thanks for the changes.

zebra-scan/src/service/tests.rs
```diff
@@ -0,0 +1,79 @@
...
+    let expected_keys = vec![zec_pages_sapling_efvk.clone()];
+    let cmd_handler_fut = tokio::task::spawn_blocking(move || {
+        let Ok(ScanTaskCommand::RemoveKeys { done_tx, keys }) = cmd_receiver.recv() else {
+            panic!("should successfully receive RemoveKeys message");
+        };
+
+        assert_eq!(keys, expected_keys, "keys should match the request keys");
+
+        done_tx.send(()).expect("send should succeed");
+    });
```
upbqdn, line 59: ###### Optional Suggestion

This looks great. I wonder if we could use mocked services from `zebra-test`, for example: https://github.com/ZcashFoundation/zebra/blob/main/zebra-test/src/mock_service.rs and https://github.com/ZcashFoundation/zebra/blob/main/zebra-test/src/transcript.rs. I haven't used them yet, but I'll look into it.

zebra-scan/src/init.rs
```diff
@@ -80,18 +89,21 @@ impl ScanTask {
...
-/// TODO: add a test for this function.
-pub fn spawn_init(
-    config: &Config,
-    network: Network,
-    state: scan::State,
-    chain_tip_change: ChainTipChange,
-) -> JoinHandle<Result<(), Report>> {
-    scan::spawn_init(config, network, state, chain_tip_change)
+    /// Sends a message to the scan task to remove the provided viewing keys.
+    pub fn remove_keys(
```
upbqdn, line 94: ###### Optional Suggestion

I think it would be good to place this function into a different file since the filename is `init.rs`, but let's do that in a different PR?

zebra-scan/src/scan.rs
```diff
@@ -130,6 +137,56 @@ pub async fn start(
...
+                    // TODO: Replace with Arc::unwrap_or_clone() when it stabilises:
+                    // https://github.com/rust-lang/rust/issues/93610
+                    let mut updated_parsed_keys =
+                        Arc::try_unwrap(parsed_keys).unwrap_or_else(|arc| (*arc).clone());
```
upbqdn, line 171: ###### Optional Suggestion

`Arc::get_mut` might also have been an option here.

Issue #8163, opened 2024-01-17
feature: Add an RPC method to retrieve scan results for a viewing key


mpguerra:
### Motivation

We want to be able to request results for a given viewing key via an RPC method.

### Specifications

- [ ] retrieve all scanned results for a set of viewing keys

We probably want to start doing this without authenticating the requester and add authentication later

### Related Work

This is part of #7728 and the RPC Interface

PR #8219, opened 2024-01-31
add(scan): Implement `ClearResults` `ScanService` request

arya2:
## Motivation

This PR implements the `ClearResults` request for ScanService to delete the results for a key without removing the key itself.

Closes #8207.

#### Check before marking the PR as ready for review:
  - [x] Will the PR name make sense to users?
  - [x] Does the PR have a priority label?
  - [x] Have you added or updated tests?
  - [x] Is the documentation up to date?

##### For significant changes:
  - [x] Is there a summary in the CHANGELOG?
  - [x] Can these changes be split into multiple PRs?

_If a checkbox isn't relevant to the PR, mark it as done._

## Solution

- Renames `delete_sapling_results` to `delete_sapling_keys`
- Adds a `delete_sapling_results` that adds back an empty entry after deleting the results for a sapling key
- Calls the new storage method from the `ScanService::call()` match arm for `Request::ClearResults`

Related cleanups:
- Moves check for max number of keys to a request method

### Testing

- Tests that `sapling_delete_results()` deletes results but leaves an empty entry with the key
- Tests that the scan service `ClearResults` request does the same

## Review

Anyone can review.

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

## Follow Up Work

Test that the scan service responds with an error for requests with an invalid number of keys.

PR #8224, opened 2024-02-01
add(scan): Implement `Results` request

### LINKED ISSUE #8205: Implement `Results` scan service request to read any cached results for a set of keys
## Motivation

We want to be able to request results for a given viewing key set from the scan service (to be used via an RPC method).

## Possible Design

- Accept a `String` in the `ScanService` `Results` request variant
- Clone `db` and spawn a blocking thread
  - Call `db.sapling_results(key)` and return the transaction ids in the result

oxarbitrage:
## Motivation

We need to implement the `Results` request in order to later do the `get_results` grpc.

Close https://github.com/ZcashFoundation/zebra/issues/8205

## Solution

Instead of just returning ids, this PR returns all data we have in a nested Tree. This data includes the corresponding key and height for each transaction hash found. A nested Tree structure is used to avoid repeating any data.

This version just returns all the results we have for the given keys. Height ranges or other tweaks were not included.

This request is less complex than others as no command needs to be sent to be scanner.

### Testing

Test for the service is added, test for the `sapling_results` database function was not added as it is already tested by `deletes_keys_and_results_correctly`.

## Review

Anyone can review. I am open to suggestions to improve it.

### Reviewer Checklist

## Follow Up Work

https://github.com/ZcashFoundation/zebra/issues/8163

### DIFF
--- a/zebra-scan/src/service.rs
+++ b/zebra-scan/src/service.rs
@@ -144,8 +144,13 @@ impl Service<Request> for ScanService {
 
                     let mut final_result = BTreeMap::new();
                     for key in keys {
+                        let db = db.clone();
                         let mut heights_and_transactions = BTreeMap::new();
-                        let txs = db.sapling_results_for_key(&key);
+                        let txs = {
+                            let key = key.clone();
+                            tokio::task::spawn_blocking(move || db.sapling_results_for_key(&key))
+                        }
+                        .await?;
                         txs.iter().for_each(|(k, v)| {
                             heights_and_transactions
                                 .entry(*k)
--- a/zebra-scan/src/service.rs
+++ b/zebra-scan/src/service.rs
@@ -1,11 +1,12 @@
 //! [`tower::Service`] for zebra-scan.
 
-use std::{future::Future, pin::Pin, task::Poll, time::Duration};
+use std::{collections::BTreeMap, future::Future, pin::Pin, task::Poll, time::Duration};
 
 use futures::future::FutureExt;
 use tower::Service;
 
-use zebra_chain::parameters::Network;
+use zebra_chain::{parameters::Network, transaction::Hash};
+
 use zebra_state::ChainTipChange;
 
 use crate::{init::ScanTask, scan, storage::Storage, Config, Request, Response};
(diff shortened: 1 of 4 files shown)


zebra-scan/src/service.rs
```diff
@@ -130,8 +131,33 @@ impl Service<Request> for ScanService {
...
+            Request::Results(keys) => {
+                let db = self.db.clone();
+
+                return async move {
+                    if keys.len() > MAX_REQUEST_KEYS {
+                        return Err(format!(
+                            "maximum number of keys per request is {MAX_REQUEST_KEYS}"
+                        )
+                        .into());
+                    }
```
oxarbitrage: Will be removed after https://github.com/ZcashFoundation/zebra/pull/8219 gets merged.

zebra-scan/src/service.rs
```diff
@@ -130,8 +131,33 @@ impl Service<Request> for ScanService {
...
+
+                    let mut final_result = BTreeMap::new();
+                    for key in keys {
+                        let mut heights_and_transactions = BTreeMap::new();
+                        let txs = db.sapling_results_for_key(&key);
```
arya2: We should call database methods from blocking threads to avoid blocking the async executor.

arya2:
```suggestion
                        let db = db.clone();
                        let mut heights_and_transactions = BTreeMap::new();
                        let txs = {
                            let key = key.clone();
                            tokio::task::spawn_blocking(move || db.sapling_results_for_key(&key))
                        }
                        .await?;
```

arya2 (dismissed):
This looks good, just needs a blocking thread for the disk reads.

zebra-scan/src/service.rs
oxarbitrage: Thanks for pointing to that. Fixed in https://github.com/ZcashFoundation/zebra/pull/8224/commits/80a429d727100ac3832031c97d3111e1b757ab4a

oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8224/commits/80a429d727100ac3832031c97d3111e1b757ab4a

zebra-scan/src/service.rs
oxarbitrage: Removed as part of https://github.com/ZcashFoundation/zebra/pull/8224/commits/b6d7003f569d1198ffb7d1d2630499627d7883a3

arya2 (approved):

PR #8237, opened 2024-02-06
add(scan): Add `ClearResults` and `DeleteKeys` gRPC methods

### LINKED ISSUE #8235: Add `ClearResults` and `DeleteKeys` gRPC methods
## Motivation

We want to be able to clear the results for a given viewing key and to delete the key and stop scanning for its transactions via RPC methods.

arya2:
## Motivation

We want to clear results and delete keys in zebra-scan via gRPC requests.

Closes #8235

## Solution

- Adds the gRPC methods to the proto definition
- Implements the methods on `ScannerRPC` to call the `ScanService`

### Testing

These methods still need tests, but they're thin wrappers around scan service requests that do have tests.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Add tests to `zebra-grpc`

### DIFF
--- a/zebra-grpc/proto/scanner.proto
+++ b/zebra-grpc/proto/scanner.proto
@@ -8,8 +8,12 @@ service Scanner {
     // Get information about the scanner service.
     rpc GetInfo (Empty) returns (InfoReply);
 
+    // Clear results for a set of keys without removing the keys from the scanner.
+    // This request does not stop the scanner from scanning blocks for these keys, it
+    // only clears past results.
     rpc ClearResults(ClearResultsRequest) returns (Empty);
 
+    // Deletes a set of keys and their results from the scanner
     rpc DeleteKeys(DeleteKeysRequest) returns (Empty);
 }
 
@@ -19,10 +23,14 @@ message InfoReply {
     uint32 min_sapling_birthday_height = 1;
 }
 
+// A request for clearing past results from the scanner cache.
 message ClearResultsRequest {
+    // Keys for which to clear results.
     repeated string keys = 1;
 }
 
+// A request to delete keys, delete their results, and stop scanning for their results.
 message DeleteKeysRequest {
+    // Keys to delete from scanner.
     repeated string keys = 1;
 }
--- a/zebra-grpc/proto/scanner.proto
+++ b/zebra-grpc/proto/scanner.proto
@@ -9,6 +9,8 @@ service Scanner {
     rpc GetInfo (Empty) returns (InfoReply);
 
     rpc ClearResults(ClearResultsRequest) returns (Empty);
+
+    rpc DeleteKeys(DeleteKeysRequest) returns (Empty);
 }
 
 // A response to a GetInfo call.
@@ -18,5 +20,9 @@ message InfoReply {
 }
 
 message ClearResultsRequest {
+    repeated string keys = 1;
+}
+
+message DeleteKeysRequest {
     repeated string keys = 1;
 }
--- a/zebra-grpc/proto/scanner.proto
+++ b/zebra-grpc/proto/scanner.proto
@@ -7,10 +7,16 @@ message Empty {}
 service Scanner {
     // Get information about the scanner service.
     rpc GetInfo (Empty) returns (InfoReply);
+
+    rpc ClearResults(ClearResultsRequest) returns (Empty);
 }
 
 // A response to a GetInfo call.
 message InfoReply {
     // The minimum sapling height allowed.
     uint32 min_sapling_birthday_height = 1;
+}
+
+message ClearResultsRequest {
+    repeated string keys = 1;
 }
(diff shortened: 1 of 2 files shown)


zebra-grpc/src/server.rs
```diff
@@ -67,6 +67,60 @@ where
...
+    }
+
+    async fn delete_keys(
+        &self,
+        request: tonic::Request<DeleteKeysRequest>,
+    ) -> Result<Response<Empty>, Status> {
+        let keys = request.into_inner().keys;
+
+        if keys.is_empty() {
+            let msg = "must provide the keys for which to clear results";
```
arya2:
```suggestion
            let msg = "must provide at least 1 key to delete";
```

zebra-grpc/src/server.rs
```diff
@@ -67,6 +67,60 @@ where
...
     }
+
+    async fn clear_results(
+        &self,
+        request: tonic::Request<ClearResultsRequest>,
+    ) -> Result<Response<Empty>, Status> {
+        let keys = request.into_inner().keys;
+
+        if keys.is_empty() {
+            let msg = "must provide the keys for which to clear results";
```
arya2:
```suggestion
            let msg = "must provide at least 1 key for which to clear results";
```

zebra-grpc/proto/scanner.proto
```diff
@@ -7,10 +7,30 @@ message Empty {}
 service Scanner {
     // Get information about the scanner service.
     rpc GetInfo (Empty) returns (InfoReply);
+
+    // Clear results for a set of keys without removing the keys from the scanner.
+    // This request does not stop the scanner from scanning blocks for these keys, it
+    // only clears past results.
+    rpc ClearResults(ClearResultsRequest) returns (Empty);
+
+    // Deletes a set of keys and their results from the scanner
```
oxarbitrage:
```suggestion
    // Deletes a set of keys and their results from the scanner.
    // This request stop the scanner from scanning blocks for the these keys.
```

zebra-grpc/src/server.rs
```diff
@@ -67,6 +67,60 @@ where
...
+            .and_then(|service| service.call(ScanServiceRequest::ClearResults(keys)))
+            .await
+            .map_err(|err| Status::unknown(format!("scan service returned error: {err}")))?
+        else {
+            return Err(Status::unknown(
+                "scan service returned an unexpected response",
+            ));
+        };
+
+        Ok(Response::new(Empty {}))
```
oxarbitrage, line 95: Do we want to return the removed results ?

zebra-grpc/src/server.rs
```diff
@@ -67,6 +67,60 @@ where
...
+            .and_then(|service| service.call(ScanServiceRequest::DeleteKeys(keys)))
+            .await
+            .map_err(|err| Status::unknown(format!("scan service returned error: {err}")))?
+        else {
+            return Err(Status::unknown(
+                "scan service returned an unexpected response",
+            ));
+        };
+
+        Ok(Response::new(Empty {}))
```
oxarbitrage, line 122: Do we want to return removed keys?

oxarbitrage (dismissed):
Looks good to me but i think we might want tests. We can do all the grpc testing (including snapshots) for all grpc in a separated ticket or add tests here so the other calls can follow.

Can you make a manual test and post the output ? A manual test should be to start the server and run these calls, make sure they reply as expected.

zebra-grpc/src/server.rs
arya2: I think we should later accept an optional height range or a stop height for clearing only some of the results, but only return an empty response if it's successful or an error from this RPC, assuming that if they're clearing the results, they already have them or are no longer interested in them.

zebra-grpc/src/server.rs
arya2: I don't think it's necessary, it would just be sending back the keys that the client provided.

arya2:
> Looks good to me but i think we might want tests. We can do all the grpc testing (including snapshots) for all grpc in a separated ticket or add tests here so the other calls can follow.

I'll open an issue for adding tests for the gRPC methods, I think it's okay to add them later since the RPC methods are thin wrappers around the scan service calls.

> Can you make a manual test and post the output ? A manual test should be to start the server and run these calls, make sure they reply as expected.

Yep, I'll start the server from zebrad in a follow-up PR targeting this branch for the manual test.

zebra-grpc/src/server.rs
oxarbitrage: ok, that makes sense, thank you.

arya2:
Manual test of these methods:

```console

~/github/zebra$ $HOME/go/bin/grpcurl -plaintext -import-path ./zebra-grpc/proto -proto zebra-grpc/proto/scanner.proto -d '{"keys": [""]}' '[::1]:50051' scanner.Scanner/ClearResults
{}
~/github/zebra$ $HOME/go/bin/grpcurl -plaintext -import-path ./zebra-grpc/proto -proto zebra-grpc/proto/scanner.proto -d '{"keys": [""]}' '[::1]:50051' scanner.Scanner/DeleteKeys
{}

```

It seems strange to have a successful response when the provided keys were not registered in the scanner, we may want to return an error in that case later.

oxarbitrage (approved):

PR #8241, opened 2024-02-06
add(scan): Start scanner gRPC server with `zebrad`

### LINKED ISSUE #8242: add(grpc): Start the grpc server when zebrad starts, add config options
- [ ] We want to configure the address and port of the grpc.
- [ ] Start grpc server when zebrad start.

arya2:
## Motivation

We want the zebrad start command to start the scanner gRPC server.

Closes #8242

## Solution

- Add a `spawn_init()` fn for the root `init()` fn in `zebra_scan`
- Replace the call to `ScanTask::spawn()` in zebrad start command with the new `spawn_init()` fn

### Testing

- This was for manual testing of the ClearResults and DeleteKeys RPC methods and worked as expected.
- Tests that the scanner RPC server is running, accepts connections, and responds to the GetInfo method.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/bin/rpc_server.rs
+++ b/zebra-scan/src/bin/rpc_server.rs
@@ -14,7 +14,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let scan_service = ServiceBuilder::new().buffer(10).service(scan_service);
 
     // Start the gRPC server.
-    zebra_grpc::server::init(scan_service).await?;
+    zebra_grpc::server::init("127.0.0.1:8231".parse()?, scan_service).await?;
 
     Ok(())
 }
--- a/zebra-scan/src/config.rs
+++ b/zebra-scan/src/config.rs
@@ -41,6 +41,8 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             sapling_keys_to_scan: IndexMap::new(),
+
+            // TODO: Add a const generic for specifying the default cache_dir path, like 'zebra' or 'zebra-scan'?
             db_config: DbConfig::default(),
         }
     }
--- a/zebra-scan/src/config.rs
+++ b/zebra-scan/src/config.rs
@@ -29,7 +29,7 @@ pub struct Config {
     /// listen_addr = '127.0.0.1:8231'
     /// ```
     ///
-    /// The recommended ports for the RPC server are:
+    /// The recommended ports for the gRPC server are:
     /// - Mainnet: 127.0.0.1:8231
     /// - Testnet: 127.0.0.1:18231
     pub listen_addr: Option<SocketAddr>,
--- a/zebra-scan/src/config.rs
+++ b/zebra-scan/src/config.rs
@@ -1,6 +1,6 @@
 //! Configuration for blockchain scanning tasks.
 
-use std::fmt::Debug;
+use std::{fmt::Debug, net::SocketAddr};
 
 use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
@@ -19,6 +19,20 @@ pub struct Config {
     //
     // TODO: allow keys without birthdays
     pub sapling_keys_to_scan: IndexMap<SaplingScanningKey, u32>,
+
+    /// IP address and port for the zebra-scan gRPC server.
+    ///
+    /// Note: The gRPC server is disabled by default.
+    /// To enable the gRPC server, set a listen address in the config:
+    /// ```toml
+    /// [shielded-scan]
+    /// listen_addr = '127.0.0.1:8231'
+    /// ```
+    ///
+    /// The recommended ports for the RPC server are:
+    /// - Mainnet: 127.0.0.1:8231
+    /// - Testnet: 127.0.0.1:18231
+    pub listen_addr: Option<SocketAddr>,
 
     /// The scanner results database config.
     //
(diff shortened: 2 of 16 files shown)


zebra-scan/src/config.rs
```diff
@@ -20,6 +20,20 @@ pub struct Config {
...
+    /// IP address and port for the zebra-scan gRPC server.
+    ///
+    /// Note: The gRPC server is disabled by default.
+    /// To enable the gRPC server, set a listen address in the config:
+    /// ```toml
+    /// [shielded-scan]
+    /// listen_addr = '127.0.0.1:8231'
+    /// ```
+    ///
+    /// The recommended ports for the RPC server are:
```
oxarbitrage:
```suggestion
    /// The recommended ports for the gRPC server are:
```

zebra-scan/src/bin/rpc_server.rs
```diff
@@ -14,7 +14,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let scan_service = ServiceBuilder::new().buffer(10).service(scan_service);
 
     // Start the gRPC server.
-    zebra_grpc::server::init(scan_service).await?;
+    zebra_grpc::server::init("127.0.0.1:8231".parse()?, scan_service).await?;
```
oxarbitrage, line 17: Please note https://github.com/ZcashFoundation/zebra/pull/8241 can't be closed until this is coming from the config file. We can do another ticket for that or make sure the merge of this PR will not close the ticket.

oxarbitrage (changes requested):
This looks really good however i found a bug that i think we should fix.

Zebra crashes if the shielded-scan database is not ephemeral with error:

```
Message:  Opening database "/home/alfredo/.cache/zebra/private-scan/v1/mainnet" failed: Error { message: "IO error: lock hold by current process, acquire time 1707322113 acquiring thread 359173: /home/alfredo/.cache/zebra/private-scan/v1/mainnet/LOCK: No locks available" }. Hint: Check if another zebrad process is running. Try changing the state cache_dir in the Zebra config.
```

Works fine if the `shielded-scan.ephemeral` config field is `true`. This is the reason why the test does not detect this.

I am not sure why this is happening and if it was introduced in this PR but the ephemeral is false by default so we should fix before merging.

zebra-scan/src/bin/rpc_server.rs
arya2: This is the binary for testing, I think it's okay for now (until zebra-scan works without the state service). The configured listen_addr is used in the init function in zebra-grpc and by zebrad when starting zebra-scan.

arya2:
> Works fine if the `shielded-scan.ephemeral` config field is `true`. This is the reason why the test does not detect this.
> 
> I am not sure why this is happening and if it was introduced in this PR but the ephemeral is false by default so we should fix before merging.

I think it's because the default 'cache_dir' is the same for zebra-state and zebra-scan's Storage, I added a TODO in the zebra-scan config to add a const generic to `DbConfig' for specifying the default cache_dir.

zebra-scan/src/config.rs
```diff
@@ -41,6 +55,9 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             sapling_keys_to_scan: IndexMap::new(),
+            listen_addr: None,
+
+            // TODO: Add a const generic for specifying the default cache_dir path, like 'zebra' or 'zebra-scan'?
```
arya2, line 60: ~@oxarbitrage I think this TODO would solve the panic. I'd prefer to do this in a follow-up PR if that's okay, I don't think the bug was introduced in this PR.~

Nevermind, it was introduced in ~this PR~ recently, I'll fix it here.

oxarbitrage:
> I think it's because the default 'cache_dir' is the same for zebra-state and zebra-scan's Storage, I added a TODO in the zebra-scan config to add a const generic to `DbConfig' for specifying the default cache_dir.

I think this is not the case, i tried used different directories and the same error happens. Relevant config parts:

```
[shielded_scan]
cache_dir = "/home/alfredo/.cache/zebra2"
delete_old_database = true
ephemeral = false
listen_addr = "127.0.0.1:8231"

[shielded_scan.sapling_keys_to_scan]

[state]
cache_dir = "/home/alfredo/.cache/zebra"
delete_old_database = true
ephemeral = false

```

Starting zebra with:

```
alfredo@spaceship:~/zebra/pr8241/zebra$ ./target/debug/zebrad -c myconf.toml start
```

zebra-scan/src/bin/rpc_server.rs
oxarbitrage: Oh, sorry about that. You are right.

zebra-scan/src/config.rs
oxarbitrage: ~Ok, fair enough. Thanks.~

Thanks!

oxarbitrage (dismissed):

zebra-scan/src/config.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8241/commits/3ca3af0ec5e491e9ef4f52e52b3d821b711faf1c.

oxarbitrage (approved):
That worked, thanks again.

Issue #8250, opened 2024-02-08
Refactor ScanTask in zebra-scan


arya2:
## Motivation

The scan task in zebra-scan is could be simplified and more maintainable if it were refactored, possibly into a new tower service or by adding some of its logic to the existing `ScanService`.

This would make it easier to:
- Implement efficient batch scanning
- Register keys immediately (it's currently only checking for new registered keys before scanning a block in the root scan task, but not *while* it's scanning a block)
- Subscribe to results for a block height while it's still being scanned (so that simultaneous `Results` and `SubscribeResults` requests will catch any results for a given key without extra logic)

## Possible Design

A design meeting could be worthwhile to discuss possible designs. We may want to consider the `Downloads` service in the inbound service and/or the `QueuedBlocks` struct in the state service.

Example:

We could replace the `scan_task` field on the `ScanService` with `scan_tasks`

```rust

struct ScanTasks {
    /// Watch channel of senders for subscribed clients
    result_senders: tokio::sync::watch::Sender<HashMap<String, Sender<()>>>,

    /// Scan tasks by id
    tasks_by_id: HashMap<ScanTaskId, ScanTask>,

    /// tasks scanning transactions for a given key
    by_key: HashMap<String, Vec<ScanTaskId>>,

    /// tasks by the next block height that they will scan
    by_height: BTreeMap<Height, ScanTaskId>,
}
```

arya2:
We may want to do #7934 in the same PR as this one to avoid changing a lot of the scan task twice, we can do #7927 first to prepare for #7934.

upbqdn:
There's a discussion about the new design in #8271.

mpguerra:
@arya2 is this still relevant in light of #8581 ?

arya2:
> @arya2 is this still relevant in light of #8581 ?

Yeah, it's not affected by #8581, though it doesn't seem like a priority right now.

mpguerra:
We won't be doing any more work on the scanner

PR #8253, opened 2024-02-08
add(scan): Implement SubscribeResults request for scan service

### LINKED ISSUE #8206: Implement `SubscribeResults` scan service request to subscribe to any new results for a set of keys
## Motivation

We want to be able to subscribe to new scanner results that can be streamed to RPC clients.

arya2:
## Motivation

We want to be able to subscribe to new scan results to stream to RPC clients.

Closes #8206.

## Solution

- Updates `process_messages()` function to return a list of subscribed keys and their result senders
- Adds a watch channel for sending `subscribed_keys` to `scan_range()` tasks
- Checks if there's a `result_sender` for a key and sends it new results if there are any

### Testing

- Updates test vector for `process_messages()` to check that it processes `SubscribeResults` messages correctly
- Adds a test for the `ScanService` to see that it sends the message correctly
- Adds an acceptance test for checking that the results channel receives new relevant transactions

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- #8250
- #8259
- #8256

### DIFF
--- a/zebra-node-services/src/scan_service/request.rs
+++ b/zebra-node-services/src/scan_service/request.rs
@@ -25,7 +25,7 @@ pub enum Request {
     /// Accept keys and return transaction data
     Results(Vec<String>),
 
-    /// TODO: Accept `KeyHash`es and return a channel receiver
+    /// Accept keys and return a channel receiver for transaction data
     SubscribeResults(HashSet<String>),
 
     /// Clear the results for a set of viewing keys
--- a/zebra-node-services/src/scan_service/request.rs
+++ b/zebra-node-services/src/scan_service/request.rs
@@ -1,5 +1,7 @@
 //! `zebra_scan::service::ScanService` request types.
 
+use std::collections::HashSet;
+
 use crate::BoxError;
 
 /// The maximum number of keys that may be included in a request to the scan service
@@ -24,7 +26,7 @@ pub enum Request {
     Results(Vec<String>),
 
     /// TODO: Accept `KeyHash`es and return a channel receiver
-    SubscribeResults(Vec<()>),
+    SubscribeResults(HashSet<String>),
 
     /// Clear the results for a set of viewing keys
     ClearResults(Vec<String>),
--- a/zebra-node-services/src/scan_service/response.rs
+++ b/zebra-node-services/src/scan_service/response.rs
@@ -1,6 +1,6 @@
 //! `zebra_scan::service::ScanService` response types.
 
-use std::{collections::BTreeMap, sync::mpsc};
+use std::collections::BTreeMap;
 
 use zebra_chain::{block::Height, transaction};
 
@@ -30,5 +30,5 @@ pub enum Response {
     ClearedResults,
 
     /// Response to [`SubscribeResults`](super::request::Request::SubscribeResults) request
-    SubscribeResults(mpsc::Receiver<transaction::Hash>),
+    SubscribeResults(tokio::sync::mpsc::Receiver<transaction::Hash>),
 }
--- a/zebra-node-services/src/scan_service/response.rs
+++ b/zebra-node-services/src/scan_service/response.rs
@@ -2,7 +2,7 @@
 
 use std::{collections::BTreeMap, sync::mpsc};
 
-use zebra_chain::{block::Height, transaction::Hash};
+use zebra_chain::{block::Height, transaction};
 
 #[derive(Debug)]
 /// Response types for `zebra_scan::service::ScanService`
@@ -21,7 +21,7 @@ pub enum Response {
     /// Response to [`Results`](super::request::Request::Results) request
     ///
     /// We use the nested `BTreeMap` so we don't repeat any piece of response data.
-    Results(BTreeMap<String, BTreeMap<Height, Vec<Hash>>>),
+    Results(BTreeMap<String, BTreeMap<Height, Vec<transaction::Hash>>>),
 
     /// Response to [`DeleteKeys`](super::request::Request::DeleteKeys) request
     DeletedKeys,
@@ -29,6 +29,6 @@ pub enum Response {
     /// Response to [`ClearResults`](super::request::Request::ClearResults) request
     ClearedResults,
 
-    /// Response to `SubscribeResults` request
-    SubscribeResults(mpsc::Receiver<Hash>),
+    /// Response to [`SubscribeResults`](super::request::Request::SubscribeResults) request
+    SubscribeResults(mpsc::Receiver<transaction::Hash>),
 }
--- a/zebra-node-services/src/scan_service/response.rs
+++ b/zebra-node-services/src/scan_service/response.rs
@@ -4,6 +4,20 @@ use std::collections::BTreeMap;
 
 use zebra_chain::{block::Height, transaction};
 
+/// A relevant transaction for a key and the block height where it was found.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ScanResult {
+    /// The key that successfully decrypts the transaction
+    pub key: String,
+
+    /// The height of the block with the transaction
+    pub height: Height,
+
+    /// A transaction ID, which uniquely identifies mined v5 transactions,
+    /// and all v1-v4 transactions.
+    pub tx_id: transaction::Hash,
+}
+
 #[derive(Debug)]
 /// Response types for `zebra_scan::service::ScanService`
 pub enum Response {
@@ -30,5 +44,5 @@ pub enum Response {
     ClearedResults,
 
     /// Response to [`SubscribeResults`](super::request::Request::SubscribeResults) request
-    SubscribeResults(tokio::sync::mpsc::Receiver<transaction::Hash>),
+    SubscribeResults(tokio::sync::mpsc::Receiver<ScanResult>),
 }
--- a/zebra-node-services/src/scan_service/response.rs
+++ b/zebra-node-services/src/scan_service/response.rs
@@ -1,9 +1,6 @@
 //! `zebra_scan::service::ScanService` response types.
 
-use std::{
-    collections::BTreeMap,
-    sync::{mpsc, Arc},
-};
+use std::{collections::BTreeMap, sync::mpsc};
 
 use zebra_chain::{block::Height, transaction::Hash};
 
@@ -28,5 +25,5 @@ pub enum Response {
     ClearedResults,
 
     /// Response to SubscribeResults request
-    SubscribeResults(mpsc::Receiver<Arc<Hash>>),
+    SubscribeResults(mpsc::Receiver<Hash>),
 }
--- a/zebra-scan/src/service/scan_task/commands.rs
+++ b/zebra-scan/src/service/scan_task/commands.rs
@@ -1,12 +1,12 @@
 //! Types and method implementations for [`ScanTaskCommand`]
 
-use std::{
-    collections::{HashMap, HashSet},
-    sync::mpsc::{self, Receiver},
-};
+use std::collections::{HashMap, HashSet};
 
 use color_eyre::{eyre::eyre, Report};
-use tokio::sync::{mpsc::error::TrySendError, oneshot};
+use tokio::sync::{
+    mpsc::{error::TrySendError, Receiver, Sender},
+    oneshot,
+};
 
 use zcash_primitives::{sapling::SaplingIvk, zip32::DiversifiableFullViewingKey};
 use zebra_chain::{block::Height, parameters::Network, transaction};
@@ -15,6 +15,8 @@ use zebra_state::SaplingScanningKey;
 use crate::scan::sapling_key_to_scan_block_keys;
 
 use super::ScanTask;
+
+const RESULTS_SENDER_BUFFER_SIZE: usize = 100;
 
 #[derive(Debug)]
 /// Commands that can be sent to [`ScanTask`]
@@ -39,7 +41,7 @@ pub enum ScanTaskCommand {
     /// Start sending results for key hashes to `result_sender`
     SubscribeResults {
         /// Sender for results
-        result_sender: mpsc::Sender<transaction::Hash>,
+        result_sender: Sender<transaction::Hash>,
 
         /// Key hashes to send the results of to result channel
         keys: HashSet<String>,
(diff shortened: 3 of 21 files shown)


arya2:
The acceptance test should pass now, the test was blocking the async executor with a blocking `.recv_timeout()` call and there weren't any relevant transactions in the height range that was being scanned before the timeout.

zebrad/tests/common/shielded_scan/scans_for_new_key.rs
```diff
@@ -4,9 +4,9 @@
 //! Sapling activation height and [`REQUIRED_MIN_TIP_HEIGHT`]
 //!
 //! export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/state"
-//! cargo test scans_for_new_key --features="shielded-scan" -- --ignored --nocapture
+//! cargo test scans_for_new_key --release --features="shielded-scan" -- --ignored --nocapture
```
upbqdn, line 7: Why use `--release` here?

zebra-scan/src/service/tests.rs
```diff
@@ -77,6 +77,50 @@ pub async fn scan_service_deletes_keys_correctly() -> Result<()> {
...
+        .call(Request::SubscribeResults(keys.iter().cloned().collect()));
+
+    let expected_keys = keys.iter().cloned().collect();
+    let cmd_handler_fut = tokio::task::spawn_blocking(move || {
+        let Ok(ScanTaskCommand::SubscribeResults {
+            result_sender: _,
+            keys,
+        }) = cmd_receiver.recv()
+        else {
+            panic!("should successfully receive RemoveKeys message");
```
upbqdn:
```suggestion
            panic!("should successfully receive `SubscribeResults` message");
```

zebra-node-services/src/scan_service/response.rs
```diff
@@ -33,5 +30,5 @@ pub enum Response {
     ClearedResults,
 
     /// Response to `SubscribeResults` request
```
upbqdn: ###### Optional Suggestion

```suggestion
    /// Response to [`SubscribeResults`](super::request::Request::SubscribeResults) request
```

zebra-node-services/src/scan_service/request.rs
```diff
@@ -24,7 +26,7 @@ pub enum Request {
     Results(Vec<String>),
 
     /// TODO: Accept `KeyHash`es and return a channel receiver
```
upbqdn: Missing docs

zebra-node-services/src/scan_service/response.rs
```diff
@@ -33,5 +30,5 @@ pub enum Response {
...
-    SubscribeResults(mpsc::Receiver<Arc<Hash>>),
+    SubscribeResults(mpsc::Receiver<Hash>),
```
upbqdn: ###### Optional Suggestion

It might be useful to use `transaction::Hash` to distinguish the type from `block::Hash`.

zebra-scan/src/service/scan_task/scan.rs
```diff
@@ -125,31 +139,56 @@ pub async fn start(
...
-                .expect("scan_until_task channel should not be closed");
+            let start_height = new_keys
+                .iter()
+                .map(|(_, (_, _, height))| *height)
+                .min()
+                .unwrap_or(sapling_activation_height);
+
+            if parsed_keys.len() == new_keys.len() {
+                info!(?start_height, "setting new start height");
+                height = start_height;
```
upbqdn: Why do we do this?

zebra-scan/src/service/tests.rs
```diff
@@ -77,6 +77,50 @@ pub async fn scan_service_deletes_keys_correctly() -> Result<()> {
...
+    join_result?;
+
+    let results_receiver = match response.map_err(|err| eyre!(err))? {
+        Response::SubscribeResults(results_receiver) => results_receiver,
+        _ => panic!("scan service returned unexpected response variant"),
+    };
+
+    results_receiver
+        .try_recv()
+        .expect_err("channel should be disconnected");
```
upbqdn: Is that because the sender is thrown away here https://github.com/zcashfoundation/zebra/blob/82803cbe592e2fc78eef548ddc37a4e6709f4491/zebra-scan/src/service/scan_task/commands.rs#L143?

zebra-node-services/src/scan_service/response.rs
oxarbitrage: Do we need to return `Height` as well ?

zebra-scan/src/service/scan_task/commands.rs
```diff
@@ -43,10 +40,11 @@ pub enum ScanTaskCommand {
     // TODO: Implement this command (#8206)
     SubscribeResults {
         /// Sender for results
-        result_sender: mpsc::Sender<Arc<Transaction>>,
+        // TODO: Update type to return full `WalletTx`
```
oxarbitrage: If we decide to return this type, should we also return it in the `Result` service ? If so, please add a `TODO` comment there as well. If we are unsure (I am unsure if that is what we will eventually do) lets remove the `TODO`.

zebra-scan/src/service/scan_task/commands.rs
```diff
@@ -43,10 +40,11 @@ pub enum ScanTaskCommand {
...
+        result_sender: mpsc::Sender<transaction::Hash>,
```
oxarbitrage: Do we need to return `Height` as well?

zebra-scan/src/service/scan_task/commands.rs
```diff
@@ -173,4 +193,21 @@ impl ScanTask {
...
     }
+
+    /// Sends a message to the scan task to start sending the results for the provided viewing keys to a channel.
+    ///
+    /// Returns the channel receiver.
+    pub fn subscribe(
+        &mut self,
+        keys: HashSet<SaplingScanningKey>,
+    ) -> Result<Receiver<transaction::Hash>, mpsc::SendError<ScanTaskCommand>> {
+        // TODO: Use a bounded channel
```
oxarbitrage, line 207: I think we should bound now.

zebra-scan/src/service/scan_task/executor.rs
```diff
@@ -1,45 +1,54 @@
...
 
 const EXECUTOR_BUFFER_SIZE: usize = 100;
 
-pub fn spawn_init() -> (
-    Sender<ScanRangeTaskBuilder>,
-    JoinHandle<Result<(), BoxError>>,
-) {
+pub fn spawn_init(
+    subscribed_keys_receiver: tokio::sync::watch::Receiver<
+        HashMap<String, std::sync::mpsc::Sender<transaction::Hash>>,
```
oxarbitrage: It is possible to use tokio everywhere ?

zebra-scan/src/service/scan_task/executor.rs
```diff
@@ -1,45 +1,54 @@
...
+    >,
+) -> (Sender<ScanRangeTaskBuilder>, JoinHandle<Result<(), Report>>) {
     // TODO: Use a bounded channel.
```
oxarbitrage:
```suggestion
    // TODO: Use a bounded channel.
```

It looks like already bounded.

zebrad/tests/common/shielded_scan/subscribe_results.rs
```diff
@@ -0,0 +1,106 @@
+//! Test registering and subscribing to the results for a new key in the scan task while zebrad is running.
+//!
+//! This test requires a cached chain state that is partially synchronized past the
+//! Sapling activation height and [`REQUIRED_MIN_TIP_HEIGHT`]
+//!
+//! export ZEBRA_CACHED_STATE_DIR="/path/to/zebra/state"
+//! cargo test scan_subscribe_results --release --features="shielded-scan" -- --ignored --nocapture
```
oxarbitrage: If we change https://github.com/ZcashFoundation/zebra/pull/8253/files#r1486053216 we should also change here.

zebrad/tests/common/shielded_scan/scans_for_new_key.rs
arya2: I was expecting it to be a slow test, but it's not needed.

zebra-scan/src/service/scan_task/scan.rs
arya2: > Why do we do this?

Earlier, it wasn't starting the scan task unless there were keys in the config, so there was always a start height. 

Now, it may start the scan task when there aren't any keys, and it would request blocks from the state service until it reaches the lowest start height of a parsed key.

This skips straight to the new min height without looping through all of the intermediate heights.

It also skips spawning a new task if the root scan task is free to scan the entire block height range for the new keys.

oxarbitrage (commented):
It looks good to me, i made some small suggestions and some questions.

zebra-scan/src/service/tests.rs
arya2: Yep, I meant to check that it's the receiver for the same channel. Checked for the specific error type in https://github.com/ZcashFoundation/zebra/pull/8253/commits/c45fbe1ca16869fb5b45131460256d29fe1369b3.

zebra-scan/src/service/scan_task/executor.rs
arya2: Yep. I'll do that here. Updated in https://github.com/ZcashFoundation/zebra/pull/8253/commits/c45fbe1ca16869fb5b45131460256d29fe1369b3 and https://github.com/ZcashFoundation/zebra/pull/8253/commits/5f6f4b3d040200f224aef5e287f25e9125b14837.

zebra-scan/src/service/scan_task/scan.rs
upbqdn: I still don't see why we set the new `height` when the number of `parsed_keys` matches `new_keys`. What's different when the numbers of  keys differ? Also, could this not set `height` too high if a user sets `start_height` to a recent tip, for example?

arya2: New keys are excluded from the return value of `process_messages()` if they were already in `parsed_keys`, so it was checking if `parsed_keys` *was* empty before the call to `process_message()`, see https://github.com/ZcashFoundation/zebra/pull/8253/commits/4d3099a9f4c9967fbfbf88bb1aaf6e8b8500355c.

> could this not set height too high if a user sets start_height to a recent tip, for example?

It could, but it'll spawn a task to scan for the queried height range of the next registered key anyway, and the current behaviour is the same as if the registered keys were in the config, so I think we should fix this in the refactor.

zebra-scan/src/service/tests.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

zebra-node-services/src/scan_service/response.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

zebra-node-services/src/scan_service/request.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

zebra-scan/src/service/scan_task/commands.rs
arya2: I think it's okay to make an RPC call from the gRPC method to get the full raw transaction for both the cached and new results from the `scan` RPC (as the 'verbose' response).

It may be be okay to return different types for cached results and newly scanned results, so it doesn't have to cache the full `WalletTx` but returns it for newly scanned results when it's there, but it may also be better to always return the full raw transaction (as the `verbose` response) so it'll be the same for detection keys later.

zebra-node-services/src/scan_service/response.rs
arya2: > It might be useful to use `transaction::Hash` to distinguish the type from `block::Hash`.

Agreed, updated in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

> Do we need to return Height as well ?

Yeah, I think adding height and key is a good idea. I'll do that here.

Update: Added in https://github.com/ZcashFoundation/zebra/pull/8253/commits/d39790e800a4d0e29538a81ac2087988f39a89c1.

zebra-scan/src/service/scan_task/commands.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8253/commits/5f6f4b3d040200f224aef5e287f25e9125b14837 and https://github.com/ZcashFoundation/zebra/pull/8253/commits/c45fbe1ca16869fb5b45131460256d29fe1369b3.

zebra-scan/src/service/scan_task/executor.rs
arya2: Yep, thank you, removed the TODO in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

zebrad/tests/common/shielded_scan/subscribe_results.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8253/commits/b6f5461389d5af6a4c7e427f6a866b75a177b5c2.

zebra-scan/src/service/scan_task/commands.rs
arya2: Yeah, I think so, added in https://github.com/ZcashFoundation/zebra/pull/8253/commits/d39790e800a4d0e29538a81ac2087988f39a89c1.

zebra-scan/src/service/scan_task/scan.rs
upbqdn: Ah, sorry. I should've seen it earlier. The original comparison makes sense to me now. Feel free to revert https://github.com/ZcashFoundation/zebra/commit/4d3099a9f4c9967fbfbf88bb1aaf6e8b8500355c.

Thank you for the answers.

upbqdn (approved):

PR #8273, opened 2024-02-14
ci: Update Debian from Bullseye to Bookworm

### LINKED ISSUE #8272: The version of Protocol Buffers in the Docker file is too old
The implementation of the new gRPC endpoint contains a `proto3` type definition with an optional field. Support for optional fields was enabled by default in Protocol Buffers v3.15.0: https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0. However, our Docker file uses Debian Bullseye with the `protobuf-compiler` package containing Protocol Buffers v3.12: https://packages.debian.org/bullseye/protobuf-compiler, which doesn't support optional fields by default.

upbqdn:
## Motivation

Close #8272. I think now is a good time to update Debian to Bookworm since it's causing issues with the current development instead of coming up with workarounds for Bullseye.

### Specifications

- Debian Bookworm contains the package `protobuf-compiler` with Protocol Buffers v3.21: https://packages.debian.org/bookworm/protobuf-compiler, which supports optional fields in type definitions.

## Solution

- Replace all occurrences of `bullseye` or `Bullseye` with `bookworm` or `Bookworm`.

### Testing

Let's see if anything breaks in the CI.

### Reviewer Checklist

### DIFF
--- a/book/src/user/supported-platforms.md
+++ b/book/src/user/supported-platforms.md
@@ -17,7 +17,7 @@ For the full requirements, see [Tier 1 platform policy](platform-tier-policy.md#
 
 | platform | os | notes | rust | artifacts
 | -------|-------|-------|-------|-------
-| `x86_64-unknown-linux-gnu` | [Debian 11](https://www.debian.org/releases/bullseye/) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | Docker
+| `x86_64-unknown-linux-gnu` | [Debian 11](https://www.debian.org/releases/bookworm/) | 64-bit | [latest stable release](https://github.com/rust-lang/rust/releases) | Docker
 
 ## Tier 2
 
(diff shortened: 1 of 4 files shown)


arya2 (approved):
Looks good!

oxarbitrage:
Sorry to be late, i think this change is ok but leaving the optional as it is will break the build in a lot of linux distributions.

For example i use Linux Mint 21.2 Cinnamon which is not that old and it is shipped with `ibprotoc 3.12.4`.

Upgrading does not seems to be that straightforward, i think we should use the alternative way to write the optionals as @arya2 suggested.

upbqdn:
Ah, so I guess you can't compile Zebra now? Do you mean upgrading to Linux Mint 21.3 Virginia?

oxarbitrage:
I mean upgrading protoc alone. But still, i don't think we should force a version that current supported OSes do not have. I have the impression some users will give up right away if they have to upgrade their systems or the library.

In the other hand i think making the change to Bookworm for the CI as done in this PR is ok.

oxarbitrage:
I opened an issue for this at https://github.com/ZcashFoundation/zebra/issues/8278

PR #8281, opened 2024-02-16
add(scan): Test the `RegisterKeys` scan service call

### LINKED ISSUE #8176: add(scan): `RegisterKeys` gRPC
### Motivation

We want to be able to register viewing keys over the gRPC interface.

Depends on #8203.

Tracked in #7728.

upbqdn:
## Motivation

We are missing tests for the `RegisterKeys` service call.

## Solution

This PR adds tests for `RegisterKeys`, and contains a bunch of additional clean-ups:

- Don't wait for Sapling activation for tests.
- Set the sleep interval for the scan service to ten secs to:
  - make tests faster, and
  - make the service respond faster to requests.
- Enable mocking Sapling scanning keys for networks other than Mainnet.

### Reviewer Checklist

### DIFF
--- a/zebra-scan/src/service/scan_task/scan.rs
+++ b/zebra-scan/src/service/scan_task/scan.rs
@@ -21,7 +21,6 @@ use zcash_client_backend::{
     scanning::{ScanError, ScanningKey},
 };
 use zcash_primitives::{
-    constants::*,
     sapling::SaplingIvk,
     zip32::{AccountId, DiversifiableFullViewingKey, Scope},
 };
@@ -407,19 +406,11 @@ pub fn scan_block<K: ScanningKey>(
 //       performance: stop returning both the dfvk and ivk for the same key
 // TODO: use `ViewingKey::parse` from zebra-chain instead
 pub fn sapling_key_to_scan_block_keys(
-    sapling_key: &SaplingScanningKey,
+    key: &SaplingScanningKey,
     network: Network,
 ) -> Result<(Vec<DiversifiableFullViewingKey>, Vec<SaplingIvk>), Report> {
-    let hrp = if network.is_a_test_network() {
-        // Assume custom testnets have the same HRP
-        //
-        // TODO: add the regtest HRP here
-        testnet::HRP_SAPLING_EXTENDED_FULL_VIEWING_KEY
-    } else {
-        mainnet::HRP_SAPLING_EXTENDED_FULL_VIEWING_KEY
-    };
-
-    let efvk = decode_extended_full_viewing_key(hrp, sapling_key).map_err(|e| eyre!(e))?;
+    let efvk =
+        decode_extended_full_viewing_key(network.sapling_efvk_hrp(), key).map_err(|e| eyre!(e))?;
 
     // Just return all the keys for now, so we can be sure our code supports them.
     let dfvk = efvk.to_diversifiable_full_viewing_key();
--- a/zebra-scan/src/service/scan_task/scan.rs
+++ b/zebra-scan/src/service/scan_task/scan.rs
@@ -61,7 +61,7 @@ const INITIAL_WAIT: Duration = Duration::from_secs(15);
 /// The amount of time between checking for new blocks and starting new scans.
 ///
 /// TODO: The current value is set to 10 so that tests don't sleep for too long and finish faster.
-/// Set it to 30 after #8250 gets addressed or remove this const completely in the refactor.
+///       Set it to 30 after #8250 gets addressed or remove this const completely in the refactor.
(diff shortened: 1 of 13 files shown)


zebra-scan/src/service/scan_task/scan.rs
```diff
@@ -61,8 +60,9 @@ const INITIAL_WAIT: Duration = Duration::from_secs(15);
 
 /// The amount of time between checking for new blocks and starting new scans.
 ///
-/// This is just under half the target block interval.
-pub const CHECK_INTERVAL: Duration = Duration::from_secs(30);
+/// TODO: The current value is set to 10 so that tests don't sleep for too long and finish faster.
+/// Set it to 30 after #8250 gets addressed or remove this const completely in the refactor.
```
arya2: Optional/Nitpick:

```suggestion
/// TODO: The current value is set to 10 so that tests don't sleep for too long and finish faster.
///       Set it to 30 after #8250 gets addressed or remove this const completely in the refactor.
```

zebra-scan/src/service/tests.rs
```diff
@@ -254,3 +256,70 @@ pub async fn scan_service_get_results_for_key_correctly() -> Result<()> {
 
     Ok(())
 }
+
+/// Tests that the scan service registers keys correctly.
+#[tokio::test]
+pub async fn scan_service_registers_keys_correctly() -> Result<()> {
+    for network in Network::iter() {
```
arya2, line 262: I like this change, but I think we should avoid adding the `strum`/`strum_macros` and add our own method to `Network`:

```rust
    /// Returns an iterator with an item for each variant of `Network`
    // TODO: Use default values of `Testnet` variant when adding fields for #7845
    pub fn iter_defaults() -> impl Iterator<Item = Self> {
        [Self::Mainnet, Self::Testnet].into_iter()
    }
```

zebra-scan/src/service/tests.rs
```diff
@@ -254,3 +256,70 @@ pub async fn scan_service_get_results_for_key_correctly() -> Result<()> {
...
+        .ready()
+        .await
+        .map_err(|err| eyre!(err))?
+        .call(Request::RegisterKeys(keys_to_register[..1].to_vec()))
+        .await
+        .map_err(|err| eyre!(err))?
+    {
+        Response::RegisteredKeys(registered_keys) => {
+            // The key should be registered.
+            assert_eq!(registered_keys, mocked_keys[..1]);
```
arya2:
```suggestion
            assert_eq!(
                registered_keys,
                mocked_keys[..1],
                "response should match newly registered key"
            );
```

zebra-scan/src/service/tests.rs
```diff
@@ -254,3 +256,70 @@ pub async fn scan_service_get_results_for_key_correctly() -> Result<()> {
...
+        .await
+        .map_err(|err| eyre!(err))?
+        .call(Request::RegisterKeys(keys_to_register))
+        .await
+        .map_err(|err| eyre!(err))?
+    {
+        Response::RegisteredKeys(registered_keys) => {
+            // Only the last two keys should be registered in this service call since the first one
+            // was registered in the previous call.
+            assert_eq!(registered_keys, mocked_keys[1..3]);
```
arya2:
```suggestion
            assert_eq!(
                registered_keys,
                mocked_keys[1..3],
                "response should match newly registered keys"
            );
```

arya2 (commented):
This looks great, thank you for also updating the `CHECK_INTERVAL`.

I'm not sure about adding the `strum` crate, I think we should add our own code where needed to avoid reviewing the added dependency, but I like the idea of iterating through `Network` variants in tests. We may want to cleanup other tests that check both networks to iterator through the default variants before adding fields to the `Testnet` variant so that they won't need to be updated later.

zebra-scan/src/service/tests.rs
upbqdn: OK, I'll use a manual implementation of the iterator. However, I remember I wanted to keep a manual implementation of something else, and Teor said it was better to have it as a dependency because it was easier to review a dependency bump than to maintain a bit of code. I don't remember what it was, though. 

The burden here will be that we'll need to keep the implementation of the iterator in sync with the struct fields when everyone has forgotten about it or there are other devs. We have other spots like that in Zebra, and we occasionally run into issues.

It's not a big deal either way.

upbqdn: Done in [6a49f16](https://github.com/ZcashFoundation/zebra/pull/8281/commits/6a49f16c2efc720b1088dbab5cd2e7e26b3ec698).

arya2 (approved):

PR #8282, opened 2024-02-19
Enable optional fields for Protocol Buffers older than v3.15.0

### LINKED ISSUE #8278: Use an alternative to optional fields in protocol buffer files
In https://github.com/ZcashFoundation/zebra/pull/8266 we added an `optional` field to the spec, this is not supported in older versions of protoc. Several linux distributions are shipped with older versions of this software (for example Linux Mint 21.2 Cinnamon use the `libprotoc 3.12.4` version), which do not support the new feature. Upgrading the library is not that straightforward and if we decide to keep the optionals as it is, we should mention the change in the README as some users will have the following build error:

```
  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: proto/scanner.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }

```

## Possible Design

One alternative is to use https://github.com/ZcashFoundation/zebra/pull/8266#issuecomment-1942973687 or remove the optional from the spec.

## Related Work

- #8272

upbqdn:
## Motivation

Close #8278.

## Solution

- Enable the `experimental_allow_proto3_optional` flag for `protoc`.

### Reviewer Checklist

### DIFF
--- a/zebra-grpc/build.rs
+++ b/zebra-grpc/build.rs
@@ -3,6 +3,7 @@
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     tonic_build::configure()
         .btree_map(["."])
+        .protoc_arg("--experimental_allow_proto3_optional")
         .type_attribute(".", "#[derive(serde::Deserialize, serde::Serialize)]")
         .compile(&["proto/scanner.proto"], &[""])?;
     Ok(())


oxarbitrage (approved):
That is good enough for me, thanks for the fix!

PR #8301, opened 2024-02-20
ci: Fix creating GCP instances

### LINKED ISSUE #8151: main branch CI failed: push in Integration Tests on GCP
GitHub Actions workflow [Integration Tests on GCP #725](https://github.com/ZcashFoundation/zebra/actions/runs/7496385890) failed.

Event: push
Branch: [main](https://github.com/ZcashFoundation/zebra/tree/main)
Commit: [9a7dc4b405e6e89e289e6d36fd040735865f2592](https://github.com/ZcashFoundation/zebra/commit/9a7dc4b405e6e89e289e6d36fd040735865f2592)

<sup><i>Created by [jayqi/failed-build-issue-action](https://github.com/jayqi/failed-build-issue-action)</i></sup>

upbqdn:
## Motivation

- Close #8151.
- We're about to release Zebra 1.6, and a necessary precondition is to: "_Make sure there has been [at least one successful full sync test](https://github.com/ZcashFoundation/zebra/actions/workflows/ci-integration-tests-gcp.yml?query=event%3Aschedule) since the last state change, [...]"_, and there's no such test.

## Solution

- Remove the `DISK_OPTION` env var, which causes this error: https://github.com/ZcashFoundation/zebra/actions/runs/7930252459/job/21652230260#step:11:76. The problem is that the variable is set in another job with id `get-disk-name`, and evaluates to an empty string in other jobs.

### Reviewer Checklist

## Follow Up Work

Refactor the cached disk selection. TODO: Replace the previous sentence with an issue number.

### DIFF
--- a/.github/workflows/sub-deploy-integration-tests-gcp.yml
+++ b/.github/workflows/sub-deploy-integration-tests-gcp.yml
@@ -197,8 +197,6 @@ jobs:
           echo "cached_disk_name=${CACHED_DISK_NAME}" >> "${GITHUB_OUTPUT}"
           echo "STATE_VERSION=${LOCAL_STATE_VERSION}" >> "${GITHUB_ENV}"
           echo "CACHED_DISK_NAME=${CACHED_DISK_NAME}" >> "${GITHUB_ENV}"
-          echo "DISK_OPTION=image=${CACHED_DISK_NAME}" >> "${GITHUB_ENV}"
-
 
       # Create a Compute Engine virtual machine and attach a cached state disk using the
       # $CACHED_DISK_NAME variable as the source image to populate the disk cached state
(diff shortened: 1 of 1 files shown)


upbqdn:
~~Removing the variable effectively turns off using cached disks, but I don't think that worked before because we had a syntactic error in the `--create-disk` flag. We were missing a comma, which was added in [`516fe1c` (#8005)](https://github.com/ZcashFoundation/zebra/pull/8005/commits/516fe1cc8ed4b758e47e74e868e8669ffe2f2bd7). Adding the comma likely started causing the current issue since the `--create-disk` flag now evaluates to `--create-disk=,`, which seems to be the bug. Without the comma, `DISK_OPTION` evaluated to an empty string, and the `name` property directly followed, so the `--create-disk` flag was syntactically correct, but the `image=$CACHED_DISK_NAME` propery, set in `DISK_OPTION`, was missing.~~

Edit: the above is not quite right. I opened #8302.

upbqdn:
This PR didn't address the underlying issue, so I'm closing it. We should address #8302 instead.

PR #8342, opened 2024-03-11
feat(rpc): OpenAPI spec

### LINKED ISSUE #8324: add(rpc): Create documentation resource for all RPCs
## Motivation

Zebra now contains RPCs that are only documented in the API docs rendered at https://docs.rs/zebra-grpc/0.1.0-alpha.2/zebra_grpc/server/struct.ScannerRPC.html. These docs are Rust-specific and not suitable as general documentation for an RPC endpoint.

Zebra also implements JSON RPCs for lightwalletd: https://docs.rs/zebra-rpc/1.0.0-beta.35/zebra_rpc/methods/index.html; and JSON RPCs for mining, which are currently not rendered in the public API due to #8169. Note that the JSON RPCs are in a different crate than the gRPCs, which complicates discoverability.

`zcashd` renders its RPCs here: https://zcash.github.io/rpc/. Ideally, we should create a similar resource that would:

- be language-agnostic,
- contain the docs for all RPCs under one roof,
- render the docs directly from the source code.

It would be great if we found a tool that would fulfill the requirements above, and we could integrate it into CD.

oxarbitrage:
## Motivation

As part of the hackaton sprint i went over developing a swagger API for zebra RPC methods. I tried different rust tools initially but they didn't went too well so i decided to create an utility that can extract the information needed from the documentation in the Rust code and create a specification with that.

The result of that is in this PR, it is a bit fragile and needs some more work but i think is a good start.

## Solution

The generator will scrape data from the rust docs files to get the information needed to generate the open api spec. This code is a bit fragile and can be improved. This results in a binary called `openapi-generator` in the `zebra-utils` crate.  

There is a  lot of changes to method documentations to make them fit in the needed format, i tried to not make it very different than what we already had.

There are some defaults added to types so we can show example responses.

### Testing

Manually. You can run the generator and paste the output into https://editor.swagger.io/

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- There are some TODOs in the code.
- The gRPC methods from the scanner are not included. For these the documentation is in the proto files and there are some tools to transform that to an openapi spec, so it should be a matter of adding that command at the end of the script and append the output.


gustavovalverde:
This is lovely! ❤️

zebra-utils/README.md
```diff
@@ -232,3 +233,80 @@ A utility for displaying Zebra's scanning results.
...
+## OpenAPI generator
+
+This utility generates an `openapi.yaml` specification by extracting information from RPC method documentation in the `zebra-rpc` crate code.
+
+### Usage
+
+To use the generator tool, build and run it with the following command:
+
+```console
+cargo run --bin openapi-generator --features="openapi-generator
```
upbqdn:
```suggestion
cargo run --bin openapi-generator --features="openapi-generator"
```

upbqdn (dismissed):
Amazing. A great foundation for #8324.

upbqdn (approved):
Also, thanks for the thorough comments describing the parsing process.

Issue #8327, opened 2024-02-26
Add a `name` field to `NetworkParameters` to be used at the end of the cache dir path


arya2:
## Motivation

We want to ensure that the cached state matches the consensus rules for a network when adding `Regtest` or custom testnets.

Some users may want to use part of the Testnet chain for custom testnets, this could be done with the `CopyState` command:
`zebrad copy-state --max-source-height 1200000 --target-config-path path/to/custom-testnet-config`. This won't work for `Regtest` because the db format is different.

### Design

- Add the network name to the `cache_dir` path to store the Regtest state separately from the `Testnet` state. The chainstate is stored in a `regtest` directory in `zcashd`.
- Panic with a "reserved network name" message if the value is set to "Mainnet", "Testnet", or "Regtest" for a configured testnet

PR #4289, opened 2022-05-04
feat(util): add a `zebrad tip-height` utility command

jvff:
## Motivation

Currently CI parses the output from tests to determine the chain tip block height. This is not great, and it would be better to read the chain tip block height directly from the cached state directory.

## Solution

Create a `zebra-tip-height` utility that can be used to read the chain tip block height from a specified Zebra cached state directory.

```
cargo run --bin zebra-tip-height -- -c path/to/zebra/cached/state
```

or

```
cargo build --release --bin zebra-tip-height
./target/release/zebra-tip-height -c /home/user/.cache/zebra/ -n testnet
```

### Reviewer Checklist

  - [ ] Code implements Specs and Designs
  - [ ] Tests for Expected Behaviour
  - [ ] Tests for Errors

PR #8368, opened 2024-03-20
add(consensus/network): Add an empty `Parameters` struct in `Network::Testnet`

### LINKED ISSUE #7968: Add an empty `NetworkParameters` struct to `Network::Testnet`

arya2:
## Motivation

We want to add configurable network parameters to `Network` for supporting `Regtest` and custom Testnets to be used in integration tests.

Closes #7968.

The db format for `HistoryTree`s and `transparent::Address`es are the same across any test networks, this could mean that a `Regtest` cache or a custom `Testnet` cache could be used on the default `Testnet`. We may want to follow up this PR with a change that ensures Zebra panics if it's configured to run on the default public testnet with a `Regtest` cache, or we may want to make that change in this PR.

I'd like special attention to changes in:
- The db format upgrade code where a `network == Testnet` condition was replaced with `network.is_a_test_network()`
- Tests where `Network::iter()` was used instead of running with `Mainnet` and `Testnet` being applicable to `Regtest`

## Solution

- Adds a `NetworkParameters` struct and includes it in `Network::Testnet`
- Adds a `new_default_testnet()` method for creating the default public `Testnet`
- Updates a condition in the db format upgrade to be true for any test network
- Updates the key types in:
  - `FUNDING_STREAM_HEIGHT_RANGES`, 
  - `FUNDING_STREAM_ADDRESSES`, and
  - `INITIAL_MIN_NETWORK_PROTOCOL_VERSION`
- Updates production code matching on `Testnet` to return appropriate values
- Adds TODOs to production code where changes are needed when adding new fields
- Uses `Network::iter()` in tests that run equivalent code for both `Mainnet` and `Testnet`
- Replaces instances of the `Testnet` constructor with `Network::new_default_testnet()`
- Changes type of `network` field on `transparent::Address` and `HistoryTreeParts` to `NetworkKind`

### Testing

`Network::iter()` was used in most tests that check both `Mainnet` and `Testnet` so they'll cover any new networks we return from that function.

This PR may still need a test for the custom deserialization of `Network` or other changes.

## Review

Anyone can review.

This PR may be easier to review by commit:
- [Minor/trivial cleanup](https://github.com/ZcashFoundation/zebra/commit/1c9dc314b1c6cf9bbc85d5481d031e6d990bd20b)
- [Changes to production code](https://github.com/ZcashFoundation/zebra/commit/bf9cbf34f3b4e707bbce431d152e7e25980ada91)
- [Test updates](https://github.com/ZcashFoundation/zebra/commit/4ed8a71035b3c980af486b190dac47bf706da635)
- [Adds `NetworkKind` and uses it instead of `Network` in `transparent::Address` and `HistoryTreeParts`
](https://github.com/ZcashFoundation/zebra/pull/8368/commits/dbf32960639e5741bd9f02ec163534de18c45c5b)
### Reviewer Checklist

## Follow Up Work

- Start adding fields to `NetworkParameters`, starting with `activation_heights` (#7970)

### DIFF
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -4,7 +4,7 @@ use std::{fmt, str::FromStr, sync::Arc};
 
 use thiserror::Error;
 
-use zcash_primitives::consensus::{Network as ZcashPrimitivesNetwork, Parameters as _};
+use zcash_primitives::constants;
 
 use crate::{
     block::{self, Height, HeightDiff},
@@ -107,13 +107,19 @@ impl NetworkKind {
     /// Returns the human-readable prefix for Base58Check-encoded transparent
     /// pay-to-public-key-hash payment addresses for the network.
     pub fn b58_pubkey_address_prefix(self) -> [u8; 2] {
-        <ZcashPrimitivesNetwork>::from(self).b58_pubkey_address_prefix()
+        match self {
+            Self::Mainnet => constants::mainnet::B58_PUBKEY_ADDRESS_PREFIX,
+            Self::Testnet | Self::Regtest => constants::testnet::B58_PUBKEY_ADDRESS_PREFIX,
+        }
     }
 
     /// Returns the human-readable prefix for Base58Check-encoded transparent pay-to-script-hash
     /// payment addresses for the network.
     pub fn b58_script_address_prefix(self) -> [u8; 2] {
-        <ZcashPrimitivesNetwork>::from(self).b58_script_address_prefix()
+        match self {
+            Self::Mainnet => constants::mainnet::B58_SCRIPT_ADDRESS_PREFIX,
+            Self::Testnet | Self::Regtest => constants::testnet::B58_SCRIPT_ADDRESS_PREFIX,
+        }
     }
 
     /// Return the network name as defined in
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -135,7 +135,9 @@ impl NetworkKind {
 
 impl From<NetworkKind> for &'static str {
     fn from(network: NetworkKind) -> &'static str {
-        // These should be different from the `Display` impl for `Network`
+        // These should be different from the `Display` impl for `Network` so that its lowercase form
+        // can't be parsed as the default Testnet in the `Network` `FromStr` impl, it's easy to
+        // distinguish them in logs, and so it's generally harder to confuse the two.
         match network {
             NetworkKind::Mainnet => "MainnetKind",
             NetworkKind::Testnet => "TestnetKind",
@@ -271,6 +273,7 @@ impl Network {
     }
 }
 
+// This is used for parsing a command-line argument for the `TipHeight` command in zebrad.
 impl FromStr for Network {
     type Err = InvalidNetworkError;
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -65,11 +65,11 @@ impl NetworkParameters {
     }
 }
 
-#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 /// An enum describing the kind of network, whether it's the production mainnet or a testnet.
 // Note: The order of these variants is important for correct bincode (de)serialization
 //       of history trees in the db format.
 // TODO: Replace bincode (de)serialization of `HistoryTreeParts` in a db format upgrade?
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub enum NetworkKind {
     /// The production mainnet.
     #[default]
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -11,6 +11,8 @@ use crate::{
     parameters::NetworkUpgrade::Canopy,
 };
 
+pub mod testnet;
+
 #[cfg(any(test, feature = "proptest-impl"))]
 use proptest_derive::Arbitrary;
 
@@ -53,18 +55,6 @@ mod tests;
 /// after the grace period.
 const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
 
-/// Network consensus parameters for test networks such as Regtest and the default Testnet.
-#[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
-#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub struct NetworkParameters {}
-
-impl NetworkParameters {
-    /// Returns true if the instance of [`NetworkParameters`] represents the default public Testnet.
-    pub fn is_default_testnet(&self) -> bool {
-        self == &Self::default()
-    }
-}
-
 /// An enum describing the kind of network, whether it's the production mainnet or a testnet.
 // Note: The order of these variants is important for correct bincode (de)serialization
 //       of history trees in the db format.
@@ -100,7 +90,7 @@ pub enum Network {
 
     /// A test network such as the default public testnet,
     /// a configured testnet, or Regtest.
-    Testnet(Arc<NetworkParameters>),
+    Testnet(Arc<testnet::Parameters>),
 }
 
 impl NetworkKind {
@@ -172,13 +162,13 @@ impl fmt::Display for Network {
 }
 
 impl Network {
-    /// Creates a new [`Network::Testnet`] with the default Testnet [`NetworkParameters`].
+    /// Creates a new [`Network::Testnet`] with the default Testnet [`testnet::Parameters`].
     pub fn new_default_testnet() -> Self {
-        Self::Testnet(Arc::new(NetworkParameters::default()))
+        Self::Testnet(Arc::new(testnet::Parameters::default()))
     }
 
-    /// Creates a new configured [`Network::Testnet`] with the provided Testnet [`NetworkParameters`].
-    pub fn new_configured_testnet(params: NetworkParameters) -> Self {
+    /// Creates a new configured [`Network::Testnet`] with the provided Testnet [`testnet::Parameters`].
+    pub fn new_configured_testnet(params: testnet::Parameters) -> Self {
         Self::Testnet(Arc::new(params))
     }
 
@@ -217,7 +207,7 @@ impl Network {
     pub fn is_max_block_time_enforced(&self, height: block::Height) -> bool {
         match self {
             Network::Mainnet => true,
-            // TODO: Move `TESTNET_MAX_TIME_START_HEIGHT` to a field on NetworkParameters (#8364)
+            // TODO: Move `TESTNET_MAX_TIME_START_HEIGHT` to a field on testnet::Parameters (#8364)
             Network::Testnet(_params) => height >= super::TESTNET_MAX_TIME_START_HEIGHT,
         }
     }
@@ -226,7 +216,7 @@ impl Network {
     pub fn default_port(&self) -> u16 {
         match self {
             Network::Mainnet => 8233,
-            // TODO: Add a `default_port` field to `NetworkParameters` to return here. (zcashd uses 18344 for Regtest)
+            // TODO: Add a `default_port` field to `testnet::Parameters` to return here. (zcashd uses 18344 for Regtest)
             Network::Testnet(_params) => 18233,
         }
     }
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -66,8 +66,10 @@ impl NetworkParameters {
 }
 
 #[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
-#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 /// An enum describing the kind of network, whether it's the production mainnet or a testnet.
+// Note: The order of these variants is important for correct bincode (de)serialization
+//       of history trees in the db format.
+// TODO: Replace bincode (de)serialization of `HistoryTreeParts` in a db format upgrade?
 pub enum NetworkKind {
     /// The production mainnet.
     #[default]
@@ -75,6 +77,9 @@ pub enum NetworkKind {
 
     /// A test network.
     Testnet,
+
+    /// Regtest mode
+    Regtest,
 }
 
 impl From<Network> for NetworkKind {
@@ -92,7 +97,8 @@ pub enum Network {
     #[default]
     Mainnet,
 
-    /// The oldest public test network.
+    /// A test network such as the default public testnet,
+    /// a configured testnet, or Regtest.
     Testnet(Arc<NetworkParameters>),
 }
 
@@ -112,26 +118,21 @@ impl NetworkKind {
     /// Return the network name as defined in
     /// [BIP70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#paymentdetailspaymentrequest)
     pub fn bip70_network_name(&self) -> String {
-        match self {
-            Self::Mainnet => "main".to_string(),
-            Self::Testnet => "test".to_string(),
-        }
-    }
-
-    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
-    pub fn from_zcash_address(network: zcash_address::Network) -> Self {
-        match network {
-            zcash_address::Network::Main => NetworkKind::Mainnet,
-            zcash_address::Network::Test | zcash_address::Network::Regtest => NetworkKind::Testnet,
+        if *self == Self::Mainnet {
+            "main".to_string()
+        } else {
+            "test".to_string()
         }
     }
 }
 
 impl From<NetworkKind> for &'static str {
     fn from(network: NetworkKind) -> &'static str {
+        // These should be different from the `Display` impl for `Network`
         match network {
             NetworkKind::Mainnet => "MainnetKind",
             NetworkKind::Testnet => "TestnetKind",
+            NetworkKind::Regtest => "RegtestKind",
         }
     }
 }
@@ -148,8 +149,7 @@ impl From<&Network> for &'static str {
             Network::Mainnet => "Mainnet",
             // TODO:
             // - Add a `name` field to use here instead of checking `is_default_testnet()`
-            // - Find out what zcashd calls the regtest cache dir for the `Network::new_regtest()` method, or
-            //   if it always uses an ephemeral db, and do the same for Regtest in Zebra (#8327).
+            // - zcashd calls the Regtest cache dir 'regtest' (#8327).
             Network::Testnet(params) if params.is_default_testnet() => "Testnet",
             Network::Testnet(_params) => "UnknownTestnet",
         }
@@ -186,15 +186,8 @@ impl Network {
     pub fn kind(&self) -> NetworkKind {
         match self {
             Network::Mainnet => NetworkKind::Mainnet,
+            // TODO: Return `NetworkKind::Regtest` if the parameters match the default Regtest params
             Network::Testnet(_) => NetworkKind::Testnet,
-        }
-    }
-
-    /// Returns the default [`Network`] for a [`NetworkKind`].
-    pub fn from_kind(kind: NetworkKind) -> Self {
-        match kind {
-            NetworkKind::Mainnet => Self::Mainnet,
-            NetworkKind::Testnet => Self::new_default_testnet(),
         }
     }
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -66,6 +66,17 @@ impl NetworkParameters {
     }
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
+/// An enum describing the kind of network, whether it's the production mainnet or a testnet.
+pub enum NetworkKind {
+    /// The production mainnet.
+    Mainnet,
+
+    /// A test network.
+    Testnet,
+}
+
 /// An enum describing the possible network choices.
 #[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
@@ -106,40 +117,49 @@ impl<'de> Deserialize<'de> for Network {
     }
 }
 
-impl Network {
+impl NetworkKind {
     /// Returns the human-readable prefix for Base58Check-encoded transparent
     /// pay-to-public-key-hash payment addresses for the network.
-    pub fn b58_pubkey_address_prefix(&self) -> [u8; 2] {
-        <ZcashPrimitivesNetwork>::try_from(self)
-            // This prefix is the same for Testnet and Regtest in zcashd.
-            // TODO: Use the constants directly when implementing `Parameters` for `Network` (#8365)
-            .unwrap_or(ZcashPrimitivesNetwork::TestNetwork)
-            .b58_pubkey_address_prefix()
+    pub fn b58_pubkey_address_prefix(self) -> [u8; 2] {
+        <ZcashPrimitivesNetwork>::from(self).b58_pubkey_address_prefix()
     }
 
     /// Returns the human-readable prefix for Base58Check-encoded transparent pay-to-script-hash
     /// payment addresses for the network.
-    pub fn b58_script_address_prefix(&self) -> [u8; 2] {
-        <ZcashPrimitivesNetwork>::try_from(self)
-            // This prefix is the same for Testnet and Regtest in zcashd.
-            // TODO: Use the constants directly when implementing `Parameters` for `Network` (#8365)
-            .unwrap_or(ZcashPrimitivesNetwork::TestNetwork)
-            .b58_script_address_prefix()
+    pub fn b58_script_address_prefix(self) -> [u8; 2] {
+        <ZcashPrimitivesNetwork>::from(self).b58_script_address_prefix()
     }
-    /// Returns true if the maximum block time rule is active for `network` and `height`.
-    ///
-    /// Always returns true if `network` is the Mainnet.
-    /// If `network` is the Testnet, the `height` should be at least
-    /// TESTNET_MAX_TIME_START_HEIGHT to return true.
-    /// Returns false otherwise.
-    ///
-    /// Part of the consensus rules at <https://zips.z.cash/protocol/protocol.pdf#blockheader>
-    pub fn is_max_block_time_enforced(&self, height: block::Height) -> bool {
+
+    /// Return the network name as defined in
+    /// [BIP70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#paymentdetailspaymentrequest)
+    pub fn bip70_network_name(&self) -> String {
         match self {
-            Network::Mainnet => true,
-            // TODO: Move `TESTNET_MAX_TIME_START_HEIGHT` to a field on NetworkParameters (#8364)
-            Network::Testnet(_params) => height >= super::TESTNET_MAX_TIME_START_HEIGHT,
+            Self::Mainnet => "main".to_string(),
+            Self::Testnet => "test".to_string(),
         }
+    }
+
+    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
+    pub fn from_zcash_address(network: zcash_address::Network) -> Self {
+        match network {
+            zcash_address::Network::Main => NetworkKind::Mainnet,
+            zcash_address::Network::Test | zcash_address::Network::Regtest => NetworkKind::Testnet,
+        }
+    }
+}
+
+impl From<NetworkKind> for &'static str {
+    fn from(network: NetworkKind) -> &'static str {
+        match network {
+            NetworkKind::Mainnet => "MainnetKind",
+            NetworkKind::Testnet => "TestnetKind",
+        }
+    }
+}
+
+impl fmt::Display for NetworkKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str((*self).into())
     }
 }
 
@@ -178,17 +198,41 @@ impl Network {
         }
     }
 
+    /// Returns the [`NetworkKind`] for this network.
+    pub fn kind(&self) -> NetworkKind {
+        match self {
+            Network::Mainnet => NetworkKind::Mainnet,
+            Network::Testnet(_) => NetworkKind::Testnet,
+        }
+    }
+
     /// Returns an iterator over [`Network`] variants.
     pub fn iter() -> impl Iterator<Item = Self> {
         // TODO: Use default values of `Testnet` variant when adding fields for #7845.
         [Self::Mainnet, Self::new_default_testnet()].into_iter()
     }
 
+    /// Returns true if the maximum block time rule is active for `network` and `height`.
+    ///
+    /// Always returns true if `network` is the Mainnet.
+    /// If `network` is the Testnet, the `height` should be at least
+    /// TESTNET_MAX_TIME_START_HEIGHT to return true.
+    /// Returns false otherwise.
+    ///
+    /// Part of the consensus rules at <https://zips.z.cash/protocol/protocol.pdf#blockheader>
+    pub fn is_max_block_time_enforced(&self, height: block::Height) -> bool {
+        match self {
+            Network::Mainnet => true,
+            // TODO: Move `TESTNET_MAX_TIME_START_HEIGHT` to a field on NetworkParameters (#8364)
+            Network::Testnet(_params) => height >= super::TESTNET_MAX_TIME_START_HEIGHT,
+        }
+    }
+
     /// Get the default port associated to this network.
     pub fn default_port(&self) -> u16 {
         match self {
             Network::Mainnet => 8233,
-            // TODO: Add a `default_port` field to `NetworkParameters` to return here.
+            // TODO: Add a `default_port` field to `NetworkParameters` to return here. (zcashd uses 18344 for Regtest)
             Network::Testnet(_params) => 18233,
         }
     }
@@ -214,10 +258,7 @@ impl Network {
     /// Return the network name as defined in
     /// [BIP70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#paymentdetailspaymentrequest)
     pub fn bip70_network_name(&self) -> String {
-        match self {
-            Network::Mainnet => "main".to_string(),
-            Network::Testnet(_params) => "test".to_string(),
-        }
+        self.kind().bip70_network_name()
     }
 
     /// Return the lowercase network name.
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -78,7 +78,8 @@ pub enum NetworkKind {
     /// A test network.
     Testnet,
 
-    /// Regtest mode
+    /// Regtest mode, not yet implemented
+    // TODO: Add `new_regtest()` and `is_regtest` methods on `Network`.
     Regtest,
 }
 
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -2,7 +2,6 @@
 
 use std::{fmt, str::FromStr, sync::Arc};
 
-use serde::{Deserialize, Deserializer};
 use thiserror::Error;
 
 use zcash_primitives::consensus::{Network as ZcashPrimitivesNetwork, Parameters as _};
@@ -66,20 +65,28 @@ impl NetworkParameters {
     }
 }
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 /// An enum describing the kind of network, whether it's the production mainnet or a testnet.
 pub enum NetworkKind {
     /// The production mainnet.
+    #[default]
     Mainnet,
 
     /// A test network.
     Testnet,
 }
 
+impl From<Network> for NetworkKind {
+    fn from(network: Network) -> Self {
+        network.kind()
+    }
+}
+
 /// An enum describing the possible network choices.
 #[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
+#[serde(into = "NetworkKind")]
 pub enum Network {
     /// The production mainnet.
     #[default]
@@ -87,34 +94,6 @@ pub enum Network {
 
     /// The oldest public test network.
     Testnet(Arc<NetworkParameters>),
-}
-
-impl<'de> Deserialize<'de> for Network {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        #[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
-        #[serde(deny_unknown_fields)]
-        enum DNetwork {
-            #[default]
-            Mainnet,
-            #[serde(alias = "Testnet")]
-            DefaultTestnet,
-            Regtest,
-            #[serde(untagged)]
-            ConfiguredTestnet(NetworkParameters),
-        }
-
-        let network = match DNetwork::deserialize(deserializer)? {
-            DNetwork::Mainnet => Network::Mainnet,
-            DNetwork::DefaultTestnet => Network::new_default_testnet(),
-            DNetwork::Regtest => unimplemented!("Regtest is not yet implemented"),
-            DNetwork::ConfiguredTestnet(params) => Network::Testnet(Arc::new(params)),
-        };
-
-        Ok(network)
-    }
 }
 
 impl NetworkKind {
@@ -184,9 +163,14 @@ impl fmt::Display for Network {
 }
 
 impl Network {
-    /// Creates a new [`Network::Testnet`] with the default Testnet network parameters.
+    /// Creates a new [`Network::Testnet`] with the default Testnet [`NetworkParameters`].
     pub fn new_default_testnet() -> Self {
         Self::Testnet(Arc::new(NetworkParameters::default()))
+    }
+
+    /// Creates a new configured [`Network::Testnet`] with the provided Testnet [`NetworkParameters`].
+    pub fn new_configured_testnet(params: NetworkParameters) -> Self {
+        Self::Testnet(Arc::new(params))
     }
 
     /// Returns true if the network is the default Testnet, or false otherwise.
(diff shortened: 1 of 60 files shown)


zebra-chain/src/parameters/network.rs
```diff
@@ -108,17 +164,32 @@ impl fmt::Display for Network {
...
-        [Self::Mainnet, Self::Testnet].into_iter()
+        [Self::Mainnet, Self::new_default_testnet()].into_iter()
     }
 
     /// Get the default port associated to this network.
     pub fn default_port(&self) -> u16 {
         match self {
             Network::Mainnet => 8233,
-            Network::Testnet => 18233,
+            // TODO: Add a `default_port` field to `NetworkParameters` to return here.
```
idky137: In zcashd, Regtest has its own default port (nDefaultPort = 18344; from zcash/src/chainparams.cpp:714). It could be good to add this to the TODO?

idky137: Looking good though!

arya2: Thank you, added in c342e5fb121b1218e999d876a1fd577ff1cc02f9 (here: [`zebra_chain::parameters::network:235`](https://github.com/ZcashFoundation/zebra/pull/8368/commits/c342e5fb121b1218e999d876a1fd577ff1cc02f9#diff-4dcd6cfa50ff2b87d1dc68673b1ec3a39b816a0473d15a17423a80ae80029bcaR235)).

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,105 @@ mod tests;
...
+
 impl From<&Network> for &'static str {
     fn from(network: &Network) -> &'static str {
         match network {
             Network::Mainnet => "Mainnet",
-            Network::Testnet => "Testnet",
+            // TODO:
+            // - Add a `name` field to use here instead of checking `is_default_testnet()`
+            // - Find out what zcashd calls the regtest cache dir for the `Network::new_regtest()` method, or
+            //   if it always uses an ephemeral db, and do the same for Regtest in Zebra (#8327).
```
arya2:
```suggestion
            // - zcashd calls the Regtest cache dir 'regtest' (#8327).
```

zebra-chain/src/parameters/network.rs
```diff
@@ -108,17 +163,69 @@ impl fmt::Display for Network {
...
+    /// Returns the [`NetworkKind`] for this network.
+    pub fn kind(&self) -> NetworkKind {
+        match self {
+            Network::Mainnet => NetworkKind::Mainnet,
+            Network::Testnet(_) => NetworkKind::Testnet,
+        }
+    }
+
+    /// Returns the default [`Network`] for a [`NetworkKind`].
+    pub fn from_kind(kind: NetworkKind) -> Self {
```
arya2: This conversion should be moved to the config deserialization since it shouldn't be used outside that context.

zebra-chain/src/parameters/network.rs
```diff
@@ -173,7 +277,7 @@ impl FromStr for Network {
     fn from_str(string: &str) -> Result<Self, Self::Err> {
         match string.to_lowercase().as_str() {
             "mainnet" => Ok(Network::Mainnet),
-            "testnet" => Ok(Network::Testnet),
+            "testnet" => Ok(Network::new_default_testnet()),
```
arya2, line 273: Consider reserving "Mainnet", "Testnet", and "Regtest" network names (panicking if a configured testnet uses them) so we don't return a default testnet here for configured testnets.

TODO: Check where this conversion is used.

Update: The `FromStr` impl doesn't seem to be used anywhere, we may want to remove it?

I added panicking if a reserved network name is used to #8327.

zebra-chain/src/primitives/address.rs
```diff
@@ -14,17 +17,17 @@ pub enum Address {
 
     /// Sapling address
     Sapling {
-        /// Address' network
-        network: Network,
+        /// Address' network kind
+        network: NetworkKind,
```
arya2, line 18: I think this enum is only used by the RPCs. Do we want to try using the network name here instead? There's no way to distinguish between testnet and regtest for transparent addresses, but that's only the case for transparent addresses, not Sapling or Orchard.

Update: I added a `Regtest` variant to `NetworkKind` instead for more robust conversions to/from `zcash_address::Network`.

zebra-chain/src/primitives/address.rs
```diff
@@ -47,17 +50,34 @@ impl TryFrom<zcash_address::Network> for Network {
     fn try_from(network: zcash_address::Network) -> Result<Self, Self::Error> {
         match network {
             zcash_address::Network::Main => Ok(Network::Mainnet),
-            zcash_address::Network::Test => Ok(Network::Testnet),
+            zcash_address::Network::Test => Ok(Network::new_default_testnet()),
             zcash_address::Network::Regtest => Err("unsupported Zcash network parameters".into()),
```
arya2:
```suggestion
     // TODO: Add conversion to regtest (#7839)
zcash_address::Network::Regtest => Err("unsupported Zcash network parameters".into()),
```

zebra-chain/src/primitives/address.rs
```diff
@@ -47,17 +50,34 @@ impl TryFrom<zcash_address::Network> for Network {
...
         }
     }
 }
 
-impl From<&Network> for zcash_address::Network {
-    fn from(network: &Network) -> Self {
+impl TryFrom<&Network> for zcash_address::Network {
+    type Error = UnsupportedNetwork;
+
+    fn try_from(network: &Network) -> Result<Self, Self::Error> {
```
arya2: This is inconsistent with the `From<NetworkKind>` impl below. Check if this is still needed with the next version of librustzcash at all. If it _is_ still needed, consider adding an `UnknownTestnet` variant of NetworkKind.

Update: Removed the `From<NetworkKind> impl`.

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,105 @@ mod tests;
...
+    /// Returns true if the instance of [`NetworkParameters`] represents the default public Testnet.
+    pub fn is_default_testnet(&self) -> bool {
+        self == &Self::default()
+    }
+}
+
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
+/// An enum describing the kind of network, whether it's the production mainnet or a testnet.
+pub enum NetworkKind {
```
arya2, line 63: Add a correctness comment about the order of NetworkKind variants being important to bincode (de)serialization of the db format.

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,105 @@ mod tests;
...
         match self {
-            Network::Mainnet => true,
-            Network::Testnet => height >= super::TESTNET_MAX_TIME_START_HEIGHT,
+            Self::Mainnet => "main".to_string(),
+            Self::Testnet => "test".to_string(),
+        }
+    }
+
+    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
+    pub fn from_zcash_address(network: zcash_address::Network) -> Self {
```
arya2: Consider adding a `Regtest` variant to NetworkKind to use here and in `From<NetworkKind> for zcash_address::Network impl.

Consider moving this method impl to the address mod where it's used and making it a private method.

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,105 @@ mod tests;
...
+            zcash_address::Network::Test | zcash_address::Network::Regtest => NetworkKind::Testnet,
+        }
+    }
+}
+
+impl From<NetworkKind> for &'static str {
+    fn from(network: NetworkKind) -> &'static str {
+        match network {
+            NetworkKind::Mainnet => "MainnetKind",
+            NetworkKind::Testnet => "TestnetKind",
```
arya2, line 133: Add a comment saying these values should differ from the `Display` impl for `Network`.

Consider adding a test verifying that this variant can't be parsed or converted `.into()` the default testnet.

Update: Added the comment, the test felt unnecessary.

zebra-chain/src/parameters/network.rs
```diff
@@ -108,17 +163,69 @@ impl fmt::Display for Network {
...
+    /// If `network` is the Testnet, the `height` should be at least
+    /// TESTNET_MAX_TIME_START_HEIGHT to return true.
+    /// Returns false otherwise.
+    ///
+    /// Part of the consensus rules at <https://zips.z.cash/protocol/protocol.pdf#blockheader>
+    pub fn is_max_block_time_enforced(&self, height: block::Height) -> bool {
+        match self {
+            Network::Mainnet => true,
+            // TODO: Move `TESTNET_MAX_TIME_START_HEIGHT` to a field on NetworkParameters (#8364)
+            Network::Testnet(_params) => height >= super::TESTNET_MAX_TIME_START_HEIGHT,
```
arya2, line 211: Consider adding a test that checks that this returns the correct value for the default testnet, either in this PR or the one that moves this to a field on `NetworkParameters`.

Update: This is already being tested, albeit strangely, we could update the test in the PR that moves this to a field.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -328,20 +328,39 @@ pub(crate) fn transparent_output_address(
...
 
-impl From<&Network> for zcash_primitives::consensus::Network {
-    fn from(network: &Network) -> Self {
+impl TryFrom<&Network> for zcash_primitives::consensus::Network {
+    type Error = UnsupportedNetwork;
+
+    fn try_from(network: &Network) -> Result<Self, Self::Error> {
+        match network {
+            Network::Mainnet => Ok(zcash_primitives::consensus::Network::MainNetwork),
+            // Note: There are differences between the `TestNetwork` parameters and those returned by
```
arya2: Consider making this a correctness comment.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -350,7 +369,7 @@ impl From<zcash_primitives::consensus::Network> for Network {
     fn from(network: zcash_primitives::consensus::Network) -> Self {
         match network {
             zcash_primitives::consensus::Network::MainNetwork => Network::Mainnet,
-            zcash_primitives::consensus::Network::TestNetwork => Network::Testnet,
+            zcash_primitives::consensus::Network::TestNetwork => Network::new_default_testnet(),
```
arya2: This _may_ cause problems, because `TestNetwork` could also mean Regtest. Double-check where it's used, and consider replacing this trait impl with a regular method.

Update: Removed this `From` impl and returned b58 constants directly in https://github.com/ZcashFoundation/zebra/pull/8368/commits/4ddb8cdb875cd41440ea6fd72da4611859547f11.

zebra-chain/src/transparent/address.rs
```diff
@@ -183,26 +183,26 @@ impl ToAddressWithNetwork for PublicKey {
...
-            network: network.clone(),
+            network,
             script_hash,
         }
     }
 
-    /// Returns the network for this address.
-    pub fn network(&self) -> Network {
+    /// Returns the network kind for this address.
+    pub fn network(&self) -> NetworkKind {
```
arya2: Consider changing this method name, field names, and variables of this type to `network_kind`.

Update: It's still easy enough to search for `NetworkKind`, we may want to do this later.

zebra-consensus/src/checkpoint/list.rs
```diff
@@ -57,19 +57,26 @@ impl ParameterCheckpoint for Network {
...
         .parse()
         .expect("hard-coded hash parses")
     }
 
     fn checkpoint_list(&self) -> CheckpointList {
         // parse calls CheckpointList::from_list
+        // TODO:
+        // - Add a `genesis_hash` field to `NetworkParameters` and return it here (#8366)
+        // - Consider adding a `CUSTOM_TESTNET_CHECKPOINTS` constant to enable building with another checkpoints list
+        //   when using a configured testnet?
```
arya2: We don't need checkpoints for regtest or configured testnets.

```suggestion
        // TODO: Disable checkpoints for regtest and custom testnets
```

zebra-consensus/src/checkpoint/list.rs
```diff
@@ -142,9 +149,11 @@ impl CheckpointList {
 
         // Check that the list starts with the correct genesis block
         match checkpoints.iter().next() {
+            // TODO: Move this check to `<Network as ParameterCheckpoint>::checkpoint_list(&network)` method above (#8366),
+            //       See <https://github.com/ZcashFoundation/zebra/pull/7924#discussion_r1385865347>
```
arya2: Consider removing this TODO (we don't need checkpoints for regtest or configured testnets).

zebra-network/src/config.rs
```diff
@@ -176,6 +178,10 @@ pub struct Config {
     pub max_connections_per_ip: usize,
 }
 
+lazy_static! {
+    static ref EMPTY_INITIAL_REGTEST_PEERS: IndexSet<String> = IndexSet::new();
```
arya2: Consider cloning the initial peers from the config and returning an owned `IndexSet` to avoid using a lazy_static here.

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -104,30 +104,32 @@ lazy_static! {
     /// as described in [protocol specification §7.10.1][7.10.1].
     ///
     /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
-    pub static ref FUNDING_STREAM_HEIGHT_RANGES: HashMap<Network, std::ops::Range<Height>> = {
+    // TODO: Move the value here to a field on `NetworkParameters` (#8367)
+    pub static ref FUNDING_STREAM_HEIGHT_RANGES: HashMap<String, std::ops::Range<Height>> = {
```
arya2: It should be fine to use the testnet values on regtest, double-check chainparams in zcashd, doesn't matter either way except for interoperability or compatibility between zcashd and zebrad (which may be important for existing tests that rely on `Regtest` in zcashd).

We could also use `NetworkKind` as the key type here, but these values should likely be moved to `Network` fields/methods soon anyway.

Update: There aren't any funding stream addresses for Regtest in zcashd, we should do the same for compatibility.

zebra-network/src/config.rs
```diff
@@ -660,6 +673,18 @@ impl<'de> Deserialize<'de> for Config {
             max_connections_per_ip,
         } = DConfig::deserialize(deserializer)?;
 
+        let network = if let Some(network_params) = testnet_parameters {
+            assert_eq!(
+                network_kind,
+                NetworkKind::Testnet,
+                "set network to 'Testnet' to use testnet parameters"
```
arya2:
```suggestion
                "set network to 'Testnet' to use configured testnet parameters"
```

zebra-network/src/config.rs
```diff
@@ -660,6 +673,18 @@ impl<'de> Deserialize<'de> for Config {
             max_connections_per_ip,
         } = DConfig::deserialize(deserializer)?;
 
+        let network = if let Some(network_params) = testnet_parameters {
```
arya2, line 672:
```suggestion
        // TODO: Panic here if the initial testnet peers are the default
        //       initial testnet peers on an incompatible custom testnet.        
        let network = if let Some(network_params) = testnet_parameters {
```

zebra-network/src/protocol/external/types.rs
```diff
@@ -80,7 +81,7 @@ impl Version {
     /// - after Zebra's local network is slow or shut down.
     fn initial_min_for_network(network: &Network) -> Version {
         *constants::INITIAL_MIN_NETWORK_PROTOCOL_VERSION
-            .get(network)
+            .get(&network.bip70_network_name())
             .expect("We always have a value for testnet or mainnet")
```
arya2, line 85:
```suggestion
            .expect("we always have a value for BIP 70 network names")
```

Update: Neither message is correct anymore, but we'll be moving this value to a field soon anyway.

zebra-consensus/src/parameters/subsidy.rs
idky137: In Zcashd regtest [src/chainparams.cpp:751], there is differing logic for loading funding streams. It is configurable in [src/init.cpp:1328].

zebra-consensus/src/checkpoint/list.rs
idky137: In Zcashd, regtest doesn't use checkpoints as chain is fully configurable.

zebra-consensus/src/checkpoint/list.rs
idky137: Similar to above, in Zcashd, regtest doesn't use checkpoints as chain is fully configurable.

zebra-chain/src/primitives/address.rs
```diff
@@ -47,17 +50,34 @@ impl TryFrom<zcash_address::Network> for Network {
...
+    }
+}
+
+impl From<NetworkKind> for zcash_address::Network {
+    fn from(network: NetworkKind) -> Self {
         match network {
-            Network::Mainnet => zcash_address::Network::Main,
-            Network::Testnet => zcash_address::Network::Test,
+            NetworkKind::Mainnet => zcash_address::Network::Main,
+            NetworkKind::Testnet => zcash_address::Network::Test,
```
idky137: should we match against [DefaultRegtestParams] and set to zcash_address::Network::Regtest?

idky137 (commented):
Should we have new_default_regtest and is_a_regtest methods similar to testnet? This would be useful for wallets to use in testing and simplify migration. They could also be used where we need to distinguish between public_testnet and regtest.

zebra-chain/src/primitives/address.rs
arya2: We need to add the required fields on `NetworkParameters` and the regtest constructor first.

arya2:
> Should we have new_default_regtest and is_a_regtest methods similar to testnet?

Yes, we should add a `is_regtest()` method, but not in this PR, we want to minimize the changes here so it's easy to review, so we're only adding an empty struct to Testnet where we can start adding fields that will be needed for `Regtest`.

zebra-state/src/service/finalized_state/disk_format/chain.rs
```diff
@@ -39,42 +43,54 @@ impl FromDisk for ValueBalance<NonNegative> {
...
+impl HistoryTreeParts {
+    /// Converts [`HistoryTreeParts`] to a [`NonEmptyHistoryTree`].
+    pub(crate) fn with_network(
+        self,
+        network: &Network,
+    ) -> Result<NonEmptyHistoryTree, HistoryTreeError> {
+        assert_eq!(
+            self.network,
+            network.kind(),
+            "history tree network kind should match current network"
```
arya2, line 62:
```suggestion
            "history tree network kind should match current configured network"
```

zebra-state/src/service/finalized_state/disk_format/chain.rs
```diff
@@ -39,42 +43,54 @@ impl FromDisk for ValueBalance<NonNegative> {
...
+            current_height: history_tree.current_height(),
+        }
+    }
+}
+
+impl IntoDisk for HistoryTreeParts {
     type Bytes = Vec<u8>;
 
     fn as_bytes(&self) -> Self::Bytes {
-        let data = HistoryTreeParts {
```
arya2, line 53: `Network` is now always converted into a `NetworkKind` before being serialized anyway, but we still need an `IntoDisk` impl for `HistoryTreeParts` since we can't deserialize the full network on history tree and using `HistoryTreeParts` in the column family type requires both `IntoDisk` and `FromDisk` impls.

Check if it's possible to use `NetworkKind` in the `HistoryTree` type to revert some of these changes?

Update: It's needed for `NetworkUpgrade::current()` calls, we could replace it with `NetworkKind` + activation heights if we want to later.

zebra-consensus/src/parameters/subsidy.rs
arya2: Updated the TODO to reflect that in https://github.com/ZcashFoundation/zebra/pull/8368/commits/8b7455990c338e017c4278d3ea2cfe68a39471ca.

I left "if required" because there is _some_ (very minimal) checkpoint data in zcashd (even for Regtest), so I'm not yet sure if we can get rid of the checkpoints altogether here.

zebra-chain/src/primitives/address.rs
```diff
@@ -185,10 +163,34 @@ impl Address {
...
+        }
+    }
+
+    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
+    ///
+    /// This method is meant to be used for encoding Zcash addresses in zebra-rpc methods.
+    fn to_zcash_address(self) -> zcash_address::Network {
+        match self {
+            NetworkKind::Mainnet => zcash_address::Network::Main,
+            NetworkKind::Testnet => zcash_address::Network::Test,
```
arya2, line 192: This may still need another `NetworkKind` variant (which includes a String for the network part of the address prefix), or it may be fine as it is, let's ask if it's acceptable to use the `ztest` part of the prefix for configurable testnet addresses.

If it needs a configurable prefix, it may still be okay to change that in another PR since this method is currently only used by `zebra-rpc`.

zebra-chain/src/transparent/address.rs
upbqdn: ###### Optional Suggestion


I'd lean towards doing the refactors in this PR.

zebra-consensus/src/checkpoint/tests.rs
```diff
@@ -205,18 +205,30 @@ async fn multi_item_checkpoint_list() -> Result<(), Report> {
...
+            .await?;
+        }
+
+        for height in 0..zebra_test::vectors::CONTINUOUS_TESTNET_BLOCKS.len() {
+            continuous_blockchain(
+                Some(block::Height(height.try_into().unwrap())),
+                network.clone(),
+            )
+            .await?;
+        }
```
upbqdn: We iterate over `Network`s here, and we use both Mainnet and Testnet heights in each iteration. We should use the heights only for the network associated with them.

zebra-consensus/src/parameters/subsidy.rs
```diff
@@ -226,7 +231,9 @@ impl ParameterSubsidy for Network {
             Network::Mainnet => NetworkUpgrade::Canopy
                 .activation_height(self)
                 .expect("canopy activation height should be available"),
-            Network::Testnet => FIRST_HALVING_TESTNET,
+
```
upbqdn: ###### Optional Suggestion

```suggestion
```

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -498,14 +498,17 @@ impl AddressTransaction {
...
     // Return smaller values for more common variants.
     //
     // (This probably doesn't matter, but it might help slightly with data compression.)
     match (address.network(), address) {
         (Mainnet, PayToPublicKeyHash { .. }) => 0,
         (Mainnet, PayToScriptHash { .. }) => 1,
-        (Testnet, PayToPublicKeyHash { .. }) => 2,
-        (Testnet, PayToScriptHash { .. }) => 3,
+        // There's no way to distinguish between Regtest and Testnet for encoded transparent addresses,
+        // so the network kind should always be `Mainnet` or `Testnet`.
```
upbqdn: Shouldn't the following line say:

```suggestion
        // so the network kind should always be `Testnet` or `Regtest` for network kinds other than Mainnet.
```
?

zebra-chain/src/primitives/zcash_note_encryption.rs
```diff
@@ -25,8 +25,24 @@ pub fn decrypts_successfully(transaction: &Transaction, network: &Network, heigh
...
+                    // `CRegTestParams()` in zcashd, so this function can't return `TestNetwork` unless
+                    // Zebra is using the default public Testnet.
+                    //
+                    // TODO: Remove this conversion by implementing `zcash_primitives::consensus::Parameters`
+                    //       for `Network` (#8365).
+                    assert!(params.is_default_testnet(), "could not convert configured testnet to zcash_primitives::consensus::Network");
+                    zcash_primitives::consensus::Network::TestNetwork
+                }
+            };
+
```
upbqdn: It looks like we can move this conversion out of the loop, after the `if let`.

 Doing so would be worth it since we call this code for each Sapling output once the full semantic verification kicks in.

zebra-chain/src/primitives/address.rs
arya2: It's okay to use the Testnet HRPs for now, we should add configurable HRPs once https://github.com/zcash/librustzcash/issues/1344 is implemented.

zebra-chain/src/primitives/address.rs
```diff
@@ -185,10 +163,34 @@ impl Address {
...
             Self::Unified { .. } => None,
         }
     }
 }
+
+impl NetworkKind {
+    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
+    ///
+    /// This method is meant to be used for decoding Zcash addresses in zebra-rpc methods.
+    fn from_zcash_address(network: zcash_address::Network) -> Self {
```
upbqdn, line 178: Why don't we go with the `From` trait?

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
 /// after the grace period.
 const ZIP_212_GRACE_PERIOD_DURATION: HeightDiff = 32_256;
 
-/// An enum describing the possible network choices.
+/// Network consensus parameters for test networks such as Regtest and the default Testnet.
 #[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub enum Network {
+pub struct NetworkParameters {}
```
upbqdn: ###### Optional Suggestion

We could call the struct `Parameters`, and prefix it with `network::` when using it outside of this module, similarly to, for example, `block::Height`s.

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
...
+    pub fn is_default_testnet(&self) -> bool {
+        self == &Self::default()
+    }
+}
+
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+/// An enum describing the kind of network, whether it's the production mainnet or a testnet.
+// Note: The order of these variants is important for correct bincode (de)serialization
+//       of history trees in the db format.
+// TODO: Replace bincode (de)serialization of `HistoryTreeParts` in a db format upgrade?
```
upbqdn: ###### Optional Suggestion

Attributes usually go below comments.

```suggestion
/// An enum describing the kind of network, whether it's the production mainnet or a testnet.
// Note: The order of these variants is important for correct bincode (de)serialization
//       of history trees in the db format.
// TODO: Replace bincode (de)serialization of `HistoryTreeParts` in a db format upgrade?
#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
```

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
...
+pub enum NetworkKind {
```
upbqdn, line 63: ###### Optional Suggestion

Similarly to `NetworkParameters`, we could go with `Kind`, and prefix it with `network::` when referring to it.

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
...
+/// Network consensus parameters for test networks such as Regtest and the default Testnet.
 #[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub enum Network {
+pub struct NetworkParameters {}
+
+impl NetworkParameters {
+    /// Returns true if the instance of [`NetworkParameters`] represents the default public Testnet.
+    pub fn is_default_testnet(&self) -> bool {
+        self == &Self::default()
```
upbqdn: ###### Optional Suggestion

```suggestion
        // TODO: Change this comparison once `Self` gets populated, since
        // it currently uses `default`, but the default `NetworkKind` is Mainnet.
        self == &Self::default()
```

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
...
+// Note: The order of these variants is important for correct bincode (de)serialization
+//       of history trees in the db format.
+// TODO: Replace bincode (de)serialization of `HistoryTreeParts` in a db format upgrade?
+pub enum NetworkKind {
     /// The production mainnet.
     #[default]
     Mainnet,
 
-    /// The oldest public test network.
+    /// A test network.
```
upbqdn, line 68: q: Are custom testnets public as well?

zebra-chain/src/parameters/network.rs
```diff
@@ -51,52 +53,112 @@ mod tests;
...
+            "main".to_string()
+        } else {
+            "test".to_string()
+        }
+    }
+}
+
+impl From<NetworkKind> for &'static str {
+    fn from(network: NetworkKind) -> &'static str {
+        // These should be different from the `Display` impl for `Network`
```
upbqdn: Is it so that we don't confuse `NetworkKind` with `Network` in logs?

I think it would be worth adding the reason to the comment.

zebra-chain/src/parameters/network.rs
```diff
@@ -108,17 +170,62 @@ impl fmt::Display for Network {
...
+    pub fn is_default_testnet(&self) -> bool {
+        if let Self::Testnet(params) = self {
+            params.is_default_testnet()
+        } else {
+            false
+        }
+    }
+
+    /// Returns the [`NetworkKind`] for this network.
+    pub fn kind(&self) -> NetworkKind {
```
upbqdn, line 185: ###### Optional Suggestion

This fn essentially converts `&Network` to `NetworkKind`. Wouldn't it be more handy to implement it using the `From` trait and then go with the implicit `into()` where needed?

zebra-state/src/service/finalized_state/disk_format/chain.rs
```diff
@@ -39,42 +43,54 @@ impl FromDisk for ValueBalance<NonNegative> {
 // https://docs.rs/bincode/1.3.3/bincode/config/index.html#options-struct-vs-bincode-functions
 
 #[derive(serde::Serialize, serde::Deserialize)]
-struct HistoryTreeParts {
-    network: Network,
+pub struct HistoryTreeParts {
+    network: NetworkKind,
```
upbqdn: Could we rename this field to `network_kind`?

Calling the field `network` made me think `HistoryTreeParts` comes with an instance of `Network`, and I was wondering why we needed the `with_network` method.

upbqdn (dismissed):
This PR looks great. I left some comments and questions.

It's a bit unfortunate we need to overload the word "network" since we also have a crate called zebra-network, which is almost entirely unrelated to the `add(network):` prefix in the name of this PR.

zebra-chain/src/primitives/address.rs
```diff
@@ -185,10 +163,34 @@ impl Address {
...
+            zcash_address::Network::Main => NetworkKind::Mainnet,
+            zcash_address::Network::Test => NetworkKind::Testnet,
+            zcash_address::Network::Regtest => NetworkKind::Regtest,
+        }
+    }
+
+    /// Converts a [`zcash_address::Network`] to a [`NetworkKind`].
+    ///
+    /// This method is meant to be used for encoding Zcash addresses in zebra-rpc methods.
+    fn to_zcash_address(self) -> zcash_address::Network {
```
upbqdn, line 189: > Why don't we go with the `From` trait?

Similarly here.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: Yeah, I meant to convey something like "Regtest uses Testnet transparent addresses", I'm happy to go with your suggestion, or a slightly more verbose:

```suggestion
        // we can consider `Regtest` to use `Testnet` transparent addresses, so it's okay to use the `Testnet`
        // address variant for `Regtest` transparent addresses in the db format
```

I'll apply one of these tomorrow, let me know if you have a preference between them.

zebra-chain/src/primitives/zcash_note_encryption.rs
arya2: Nice catch!

zebra-chain/src/primitives/address.rs
arya2: It was to keep it private to this module and so it would be easier to search for where it's used to avoid using it outside of RPC methods, but since it's okay for now to use `Testnet` HRPs for custom Testnets, I'll update this and go with the `From` trait.

Update: Let's keep it, it's still easier to search for and seems fine either way.

zebra-chain/src/parameters/network.rs
arya2: The default network is also Mainnet, we're only using `NetworkParameters` for Testnet, do we want to rename `NetworkParameters` to `TestnetParameters` or move it to a sub-module to make it `network::testnet::Parameters`?

zebra-chain/src/parameters/network.rs
arya2: Yep, except `Regtest` which would be a special case of the `Testnet` variant.

zebra-chain/src/parameters/network.rs
arya2: > Is it so that we don't confuse NetworkKind with Network in logs?

Yeah, I was thinking so its lowercase form can't be parsed as the default Testnet in the `FromStr` trait impl and so it's generally harder to confuse the two, distinguishing them in logs is a good reason too, I'll add both.

Do you think we could also remove the `FromStr` trait impl, or do you know why we have it? I haven't seen it being used anywhere.

zebra-chain/src/parameters/network.rs
arya2: I wanted to use the method so it's easier to find where the conversion is used. I was also just copying the pattern of [`io::Error.kind()`](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind). There's also a `From` trait impl that's used for serde serialization of the owned `Network`.

I don't have a strong opinion if you'd prefer the `From` impl, but I have a weak preference for the `kind()` method otherwise.

upbqdn: Sure, let's keep it as it is.

zebra-chain/src/parameters/network.rs
upbqdn: > Do you think we could also remove the FromStr trait impl, or do you know why we have it?

We need it because we derive a command-line arg [`Parser`](https://docs.rs/clap/latest/clap/trait.Parser.html) here: https://github.com/zcashfoundation/zebra/blob/05e41a28037ca4288a61b7a7e9a4e0ce40e40b98/zebrad/src/commands/tip_height.rs#L21-L31, which needs to parse the network from a command-line arg. 

I guess the macro impls https://docs.rs/clap/latest/clap/builder/struct.ValueParser.html# for `Network` through https://docs.rs/clap/latest/clap/macro.value_parser.html#.

I guess we could impl any other trait for `Network` that is listed among the "Supported types" here: https://docs.rs/clap/latest/clap/macro.value_parser.html#, but that probably wouldn't help anything.

We introduced the parser in #4289 as a Zebra utility, then moved it to a Zebra command.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
upbqdn: Go ahead with yours, it makes more sense.

zebra-chain/src/primitives/address.rs
upbqdn: That makes sense, feel free to keep it as it is. I was just wondering what the reason was.

zebra-chain/src/parameters/network.rs
upbqdn: > move it to a sub-module to make it `network::testnet::Parameters`?

This sounds great to me.

upbqdn: Sorry for the confusing comment, I forgot the params are only related to Testnet.

upbqdn: If we can't go with `testnet::Parameters` for any reason, it think it would be good to at least rename the struct to `TestnetParameters`.

zebra-chain/src/transparent/address.rs
arya2: Renamed in https://github.com/ZcashFoundation/zebra/pull/8368/commits/f04b94b13afc061bcedee9b44d54b467c0e71fc6 and https://github.com/ZcashFoundation/zebra/pull/8368/commits/86cd919135c40e5b2dd512c1ed14aa0963465262.

zebra-consensus/src/checkpoint/tests.rs
arya2: Nice catch, fixed in https://github.com/ZcashFoundation/zebra/pull/8368/commits/ea2d001fdcf3fc145081e3570b49fa6804566a29.

zebra-chain/src/primitives/zcash_note_encryption.rs
arya2: Moved in https://github.com/ZcashFoundation/zebra/pull/8368/commits/86cd919135c40e5b2dd512c1ed14aa0963465262.

zebra-chain/src/parameters/network.rs
arya2: Renamed in https://github.com/ZcashFoundation/zebra/pull/8368/commits/365b58ea3f874d09f54de898f3c1dc8455f2bb77.

zebra-chain/src/parameters/network.rs
arya2: `Kind` feels too generic, I'd prefer to keep it as `NetworkKind`, though I like this idea.

zebra-chain/src/parameters/network.rs
arya2: Moved to a `testnet` module in https://github.com/ZcashFoundation/zebra/pull/8368/commits/365b58ea3f874d09f54de898f3c1dc8455f2bb77. This helps move the `default_testnet_params()` method in the following PR to a `Default` trait impl too, good idea!

zebra-chain/src/parameters/network.rs
arya2: Added the reason to the comment and also a note about where the `FromStr` impl is used in https://github.com/ZcashFoundation/zebra/pull/8368/commits/86cd919135c40e5b2dd512c1ed14aa0963465262#diff-4dcd6cfa50ff2b87d1dc68673b1ec3a39b816a0473d15a17423a80ae80029bca.

Thank you for finding where that's used.

arya2:
> It's a bit unfortunate we need to overload the word "network" since we also have a crate called zebra-network

Yep. It caused some indecisiveness when I was naming the PR and choosing labels, since the PR is related to which Zcash network the node is in, but I think "consensus" actually fits better, I'll update the PR title to include both.

Running a full sync test here: 
https://github.com/ZcashFoundation/zebra/actions/runs/8651283991

zebra-state/src/service/finalized_state/disk_format/chain.rs
arya2: Yep, renamed in https://github.com/ZcashFoundation/zebra/pull/8368/commits/86cd919135c40e5b2dd512c1ed14aa0963465262 (I initially skipped it because I wasn't sure how bincode would serialize it, but it's ignored, so this is better).

zebra-chain/src/primitives/zcash_note_encryption.rs
```diff
@@ -22,11 +22,29 @@ pub fn decrypts_successfully(transaction: &Transaction, network: &Network, heigh
 
     let alt_height = height.0.into();
     let null_sapling_ovk = zcash_primitives::keys::OutgoingViewingKey([0u8; 32]);
+    let network = match network {
```
upbqdn, line 25: ###### Optional Suggestion & Optimization Nitpick

We can move all these three variables inside the `if let` below. We call the `decrypts_successfully` fn for each tx during full semantic validation, and having the variables in the `if let` would mean they would be instantiated only if the tx contains shielded Sapling data. The variables would also be closer to where we use them, with a more limited scope.

This suggestion is not relevant to this PR, but it's a quick one while we're at it. Feel free to skip it, though.

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -450,11 +450,11 @@ where
         // Prevent loss of miner funds due to an unsupported or incorrect address type.
         if let Some(miner_address) = mining_config.miner_address.clone() {
             assert_eq!(
-                miner_address.network(),
-                network.clone(),
+                miner_address.network_kind(),
+                network.kind(),
                 "incorrect miner address config: {miner_address} \
                          network.network {network} and miner address network {} must match",
```
upbqdn, line 456: ###### Optional Suggestion

```suggestion
                "Incorrect mining address config: Zebra is configured to run on {}, but the \
                         configured mining address is for {}. The network kinds must match.",
                network.kind(),
```

upbqdn (approved):

upbqdn:
I changed `NetworkParameters` to `Parameters` in the PR name.

zebra-chain/src/primitives/zcash_note_encryption.rs
upbqdn: I just noticed you removed the conversion in #8383, so we can ignore this.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Oh. This could cause problems on Regtest, since the miner address will always be `NetworkKind::Testnet` (I updated this code before adding the `NetworkKind::Regtest` variant). I'll update the assertion in the activation heights PR if that's okay, there isn't a Regtest yet so it won't cause problems until later.

Update: I've fixed it on the `activ-heights-params-impl` branch with:
```rust
        if let Some(miner_address) = mining_config.miner_address.clone() {
            match network.kind() {
                network_kind @ NetworkKind::Mainnet => assert_eq!(
                    miner_address.network_kind(),
                    NetworkKind::Mainnet,
                    "Incorrect mining address config: Zebra is configured to run on a {network_kind} network, \
                    but the configured mining address is for {}. The network kinds must match.",
                    miner_address.network_kind(),
                ),
                // `Regtest` uses `Testnet` transparent addresses.
                network_kind @ (NetworkKind::Testnet | NetworkKind::Regtest) => assert_eq!(
                    miner_address.network_kind(),
                    NetworkKind::Testnet,
                    "Incorrect mining address config: Zebra is configured to run on a {network_kind} network, \
                    but the configured mining address is for {}. The network kinds must match.",
                    miner_address.network_kind(),
                ),
            }
        }
```

I'll push the commit after fixing some other issues.

arya2:
Added a do-not-merge label to wait on the full sync test to finish.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8383/commits/325a6c615491d79fdc699f7bdf3cddbcb9c06a71.

arya2:
@Mergifyio refresh

arya2:
@Mergifyio refresh

arya2:
@Mergifyio refresh

PR #8376, opened 2024-03-23
fix(rpc): Await state service requests in `getblock` method in parallel

arya2:
## Motivation

This is cleanup in the `getblock` method ahead of adding any additional service calls, adding other fields to the `getblock` response that are provided in `zcashd` would likely require another state service call to request the block header, and we may also want to look up the height if the method was called with a block hash.

This could change the order in which the state service receives and responds to these requests, which should be okay.

## Solution

- Calls `state.oneshot()` instead of `.ready().and_then(svc.call())` for getting the transactions ids, sapling/orchard trees, and block depth
- Awaits the futures in a `FuturesOrdered`

### Testing

Existing tests should cover this change.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Look up block heights and headers in the state service
- Add fields in zcashd to `GetBlock::Object` verbose response of the `getblock` method
- We may also want to:
  - Use a `join!()` instead of a `FuturesOrdered` (not done here due to `rustfmt` issues)
  - Move the error conversion to a trait method + `From` impl to avoid duplicate code

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -9,7 +9,7 @@
 use std::{collections::HashSet, default::Default, fmt::Debug, sync::Arc};
 
 use chrono::Utc;
-use futures::{FutureExt, TryFutureExt};
+use futures::{stream::FuturesOrdered, FutureExt, StreamExt, TryFutureExt};
 use hex::{FromHex, ToHex};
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
Looks a lot better. Thank you.

PR #8383, opened 2024-04-03
add(consensus): Add `activation_heights` field to `NetworkParameters` and implement `Parameters` for `Network`

### LINKED ISSUE #7840: Add config field for accepting network upgrade activation heights
## Motivation

This field is needed for configuring custom testnet activation heights without hard-coded changes or compilation features.

Part of #7970

### LINKED ISSUE #7970: Add a field for the network upgrade activation heights to `TestnetParameters`
## Motivation

This field is needed for configuring custom testnet activation heights without hard-coded changes or compilation features.

Depends-On: #7968

It addresses part of item 2 in https://github.com/ZcashFoundation/zebra/issues/8280:

> First of all TESTNET_ACTIVATION_HEIGHTS[Nu5] constant is modified in network_upgrade.rs#L113 from 1_842_420 to 1_060_755. If there are plans to make this list a config rather than const or somehow use FAKE_TESTNET_ACTIVATION_HEIGHTS

## Possible Design

- Add an `activation_heights` field to the `TestnetParameters` from #7968
- Add a configuration field in the `consensus` section for defining custom testnet activation heights or for enabling the [`ZFuture` network upgrade](https://docs.rs/zcash_primitives/latest/zcash_primitives/consensus/enum.NetworkUpgrade.html#variant.ZFuture) 
  - This could also be done separately as #7840

## Testing

Check that the parameters have not changed for Network::Mainnet or the default Testnet

### LINKED ISSUE #8365: Implement `zcash_primitives::consensus::Parameters` for `Network` to pass `network` to `try_sapling_output_recovery()` directly
## Motivation

Some network parameters in Zebra may not be convertible to `zcash_primitives::consensus::Network`, which Zebra currently uses to provide parameters to `try_sapling_output_recovery()` (and elsewhere). 

There also isn't a `Regtest` variant in `zcash_primitives::consensus::Network`, and not all the parameters from `Regtest` match those in `zcash_primitives::consensus::Network::TestNetwork`.

## Note

Search for this issue number and remove conversions of zebra-chain's `Network` to `zcash_primitives::consensus::Network`.

arya2:
## Motivation

We want to have an `activation_heights` field on `NetworkParameters` so that we can use different values on `Regtest` and so that it can be configured for custom Testnets.

Closes #7970.
Closes #8365.
Closes #7840.

Depends-On: #8368

## Solution

- Adds an `activation_heights` field to `NetworkParameters`
- Adds an intermediate struct for deserializing `testnet_parameters` from Zebra's config
- Implements `Parameters` for `Network` to avoid tricky conversions to `zcash_primitives::consensus::Network`
  - Uses placeholder values for unknown testnets, we should discuss what to do there before merging this PR

### Testing

- Adds a `stored_configs_parsed_correctly` test to check if new configs are valid more quickly
- Updates the stored config for v1.7.0
- Checks that the zcash_primitives `Parameters` impl for zebra-chain's `Network`'s method outputs match those of the `Parameters` impl for `zcash_primitives::consensus::Network`.

## Review

Anyone can review.

### Reviewer Checklist


zebrad/tests/common/configs/v1.7.0.toml
```diff
@@ -61,6 +61,16 @@ network = "Testnet"
 peerset_initial_target_size = 25
 
 [network.testnet_parameters]
+activation_heights = [
+    [0, "Genesis"],
+    [1, "BeforeOverwinter"],
```
arya2: Genesis is certainly required, `BeforeOverwinter` is probably the same as `BASE_SPROUT` in zcashd so it probably also has to always be 1?

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -1,12 +1,30 @@
...
-#[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
-pub struct Parameters {}
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct Parameters {
+    /// The network upgrade activation heights for this network.
+    ///
+    /// Note: This value is ignored by `Network::activation_list()` when `zebra-chain` is
+    ///       compiled with the `zebra-test` feature flag AND the `TEST_FAKE_ACTIVATION_HEIGHTS`
+    ///       environment variable is set.
+    pub activation_heights: BTreeMap<Height, NetworkUpgrade>,
```
arya2: ~We may want/need to activate multiple network upgrades at a given height, maybe this should be a struct/array instead? Or a `BTreeMap` of `Vec`s?~

~Let's leave it as a BTreeMap and require network upgrade activations to be at unique heights until we're sure it's needed.~

~Does anyone know if network upgrades rely on prior network upgrades, or, could NU5 be activated without Overwinter for example?~

Update: [They do rely on prior network upgrades](https://github.com/zcash/zcash/blob/master/src/init.cpp#L1298), I'll update the methods instead of the data struct to implicitly activate missing network upgrades when a later network upgrade is activated and add a check in the `Parameters` construction to ensure that network upgrades are in the right order.

Follow-up question, does `BeforeOverwinter` always need to be activated at `Height(1)`? I haven't found any issues with it being activated at a later/higher height yet, but I'll make it required for now, just in case.

zebrad/tests/acceptance.rs
```diff
@@ -922,6 +922,71 @@ fn invalid_generated_config() -> Result<()> {
...
+
+        // ignore files starting with shieldedscan prefix
+        // if we were not built with the shielded-scan feature.
+        #[cfg(not(feature = "shielded-scan"))]
+        if config_file_name.starts_with(SHIELDED_SCAN_CONFIG_PREFIX) {
+            tracing::info!(?config_file_path, "skipping shielded-scan config file path");
+            continue;
+        }
+
+        let stored_config_path = config_file_full_path(config_file.path());
```
arya2: Most of this code is the same as `stored_configs_work()`, do we want to refactor the common parts into another function here?

zebra-chain/src/parameters/network/tests/vectors.rs
```diff
@@ -0,0 +1,151 @@
...
+        assert_eq!(
+            activation_height, Height(expected_activation_height),
+            "activation height for all networks after Genesis and BeforeOverwinter \
+            should match NU5 activation height, network_upgrade: {nu}, activation_height: {activation_height:?}"
+        );
+    }
+}
+
+/// Checks that there are no duplicate activation heights when using configured activation heights.
+// TODO: Convert this to a proptest.
```
arya2: Actually, this seems unnecessary, it's really only here to make sure we don't remove the filter for `BeforeOverwinter` without updating `.split_off(&Height(2))` to `Height(1)`.

```suggestion
// TODO: Find out if it's okay for `BeforeOverwinter` to be activated at a later height, and remove this test if it is, 
//       see <https://github.com/ZcashFoundation/zebra/pull/8383/files#r1563215645>
```

zebra-chain/src/parameters/network.rs
```diff
@@ -279,3 +275,84 @@ impl FromStr for Network {
...
+        }
+    }
+
+    fn b58_pubkey_address_prefix(&self) -> [u8; 2] {
+        self.kind().b58_pubkey_address_prefix()
+    }
+
+    fn b58_script_address_prefix(&self) -> [u8; 2] {
+        self.kind().b58_script_address_prefix()
+    }
```
arya2, line 343: Do we want to return these constants directly if this is called very frequently, or do we trust that rustc will optimize it?

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -1,16 +1,168 @@
...
+        // if multiple network upgrades are configured with the same activation height.
+        let activation_heights: BTreeMap<_, _> = before_overwinter
+            .into_iter()
+            .map(|h| (h, BeforeOverwinter))
+            .chain(overwinter.into_iter().map(|h| (h, Overwinter)))
+            .chain(sapling.into_iter().map(|h| (h, Sapling)))
+            .chain(blossom.into_iter().map(|h| (h, Blossom)))
+            .chain(heartwood.into_iter().map(|h| (h, Heartwood)))
+            .chain(canopy.into_iter().map(|h| (h, Canopy)))
+            .chain(nu5.into_iter().map(|h| (h, Nu5))) // TODO: Find out if `BeforeOverwinter` is required at Height(1), remove this filter if it's not required to be at Height(1)
```
arya2, line 86:
```suggestion
            .chain(nu5.into_iter().map(|h| (h, Nu5))) 
            // TODO: Find out if `BeforeOverwinter` is required at Height(1), remove this filter if it's not required to be at Height(1)
```

zebrad/tests/acceptance.rs
```diff
@@ -922,6 +922,71 @@ fn invalid_generated_config() -> Result<()> {
...
+
+        tracing::info!(
+            ?stored_config_path,
+            "testing old config can be parsed by current zebrad"
+        );
+
+        ZebradApp::default()
+            .load_config(&stored_config_path)
+            .expect("config should parse");
```
upbqdn: ###### Optional Suggestion

We can simplify:

```suggestion
        tracing::info!(
            ?config_file_path,
            "testing old config can be parsed by current zebrad"
        );

        ZebradApp::default()
            .load_config(&config_file_path)
            .expect("config should parse");
```

zebra-chain/src/parameters/network.rs
upbqdn: I guess that's hard to tell. I also often wonder what gets optimized away and what stays. I would leave it as it is. I guess there are slower parts in Zebra that are worth optimizing.

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -1,16 +1,168 @@
...
+    }
+}
+
+impl ParametersBuilder {
+    /// Checks that the provided network upgrade activation heights are in the correct order, then
+    /// sets them as the new network upgrade activation heights.
+    pub fn activation_heights(
+        mut self,
+        ConfiguredActivationHeights {
+            // TODO: Find out if `BeforeOverwinter` is required at Height(1), remove this filter if it's not required to be at Height(1)
```
arya2:
```suggestion
            // TODO: Find out if `BeforeOverwinter` is required at Height(1), allow for configuring its activation height if it's not required to be at Height(1)
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -449,13 +449,23 @@ where
...
+                    "Incorrect mining address config: Zebra is configured to run on a {network_kind} network, \
+                    but the configured mining address is for {}. The network kinds must match.",
+                    miner_address.network_kind(),
+                ),
+                // `Regtest` uses `Testnet` transparent addresses.
+                network_kind @ (NetworkKind::Testnet | NetworkKind::Regtest) => assert_eq!(
+                    miner_address.network_kind(),
+                    NetworkKind::Testnet,
+                    "Incorrect mining address config: Zebra is configured to run on a {network_kind} network, \
+                    but the configured mining address is for {}. The network kinds must match.",
```
upbqdn: This error message says that the network kinds must match, but when Zebra is configured for Regtest, the kinds won't match because the address needs to be for Testnet.

I tried to clarify it below.

```suggestion
                    "Incorrect config: Zebra is configured to run on a {network_kind} network, \
                    which implies the configured mining address needs to be for Testnet, \
                    but the provided address is for {}.",
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -449,13 +449,23 @@ where
...
-                "incorrect miner address config: {miner_address} \
-                         network.network {network} and miner address network {} must match",
-                miner_address.network(),
-            );
+            match network.kind() {
+                network_kind @ NetworkKind::Mainnet => assert_eq!(
+                    miner_address.network_kind(),
+                    NetworkKind::Mainnet,
+                    "Incorrect mining address config: Zebra is configured to run on a {network_kind} network, \
+                    but the configured mining address is for {}. The network kinds must match.",
```
upbqdn: We have a bug in the message for non-Mainnet network kinds, so I tried to adjust this one so that they look similar.

```suggestion
                    "Incorrect config: Zebra is configured to run on a Mainnet network, \
                    which implies the configured mining address needs to be for Mainnet, \
                    but the provided address is for {}.",
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -449,13 +449,23 @@ where
...
         if let Some(miner_address) = mining_config.miner_address.clone() {
-            assert_eq!(
-                miner_address.network(),
-                network.clone(),
-                "incorrect miner address config: {miner_address} \
-                         network.network {network} and miner address network {} must match",
-                miner_address.network(),
-            );
+            match network.kind() {
+                network_kind @ NetworkKind::Mainnet => assert_eq!(
```
upbqdn: If we refactor the error message below, we can simplify:

```suggestion
                NetworkKind::Mainnet => assert_eq!(
```

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -7,14 +7,25 @@ use crate::parameters::{Network, Network::*};
...
-use std::ops::Bound::*;
 
 use chrono::{DateTime, Duration, Utc};
 use hex::{FromHex, ToHex};
 
 #[cfg(any(test, feature = "proptest-impl"))]
 use proptest_derive::Arbitrary;
 
+/// A list of network upgrades in the order that they must be activated.
+pub const NETWORK_UPGRADES_IN_ORDER: [NetworkUpgrade; 8] = [
```
arya2, line 18: @upbqdn `EnumIter` is looking pretty good now, do we want to open an issue for replacing this with the macro?

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -310,9 +327,12 @@ impl NetworkUpgrade {
...
             .activation_list()
             .iter()
-            .filter(|(_, nu)| nu == &self)
+            .find(|(_, nu)| nu == &self)
             .map(|(height, _)| *height)
-            .next()
+            .or_else(|| {
+                self.next_upgrade()
+                    .and_then(|next_nu| next_nu.activation_height(network))
+            })
```
upbqdn, line 342: I think this change merits a docs update.

upbqdn (dismissed):

zebra-chain/src/parameters/network.rs
```diff
@@ -279,3 +275,84 @@ impl FromStr for Network {
...
+            zp_consensus::NetworkUpgrade::Overwinter => NetworkUpgrade::Overwinter,
+            zp_consensus::NetworkUpgrade::Sapling => NetworkUpgrade::Sapling,
+            zp_consensus::NetworkUpgrade::Blossom => NetworkUpgrade::Blossom,
+            zp_consensus::NetworkUpgrade::Heartwood => NetworkUpgrade::Heartwood,
+            zp_consensus::NetworkUpgrade::Canopy => NetworkUpgrade::Canopy,
+            zp_consensus::NetworkUpgrade::Nu5 => NetworkUpgrade::Nu5,
+        };
+
+        // Heights are hard-coded below Height::MAX or checked when the config is parsed.
+        target_nu
```
arya2, line 294: This is called at least twice for every valid Sapling output in coinbase transactions (`zcash_note_encryption::try_sapling_output_recovery()` eventually calls `plaintext_version_is_valid`). It's slightly worse for configured activation heights where some network upgrades are omitted, because it'll iterate through the list for every network upgrade after `self`.

I think this is fine because:
- It doesn't seem significant relative to the decryption and key derivations in `try_sapling_output_recovery()`
- There should be few cache-misses if it's called frequently since it's the same memory location across calls
- It should only be ~40B so it won't take up much space in the cache, and
- Semantic block validation is done in parallel and sync performance is likely constrained by the sequential contextual validation in the state service write task

So it seems unlikely to measurably affect sync performance for most users.

We could open an issue to benchmark this and other factors in sync performance, if the performance impact _is_ measurable, we could use a struct instead (like the one used for configuring activation heights).

zebrad/tests/acceptance.rs
upbqdn: Let's do that some other time?

zebra-chain/src/parameters/network_upgrade.rs
upbqdn: Sounds good.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8383/commits/fb2d13a7062d5d65db49cf89a472425c9efd59bd.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8383/commits/fb2d13a7062d5d65db49cf89a472425c9efd59bd.

zebra-rpc/src/methods/get_block_template_rpcs.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8383/commits/fb2d13a7062d5d65db49cf89a472425c9efd59bd.

zebra-chain/src/parameters/network_upgrade.rs
arya2: Updated the docs in https://github.com/ZcashFoundation/zebra/pull/8383/commits/fb2d13a7062d5d65db49cf89a472425c9efd59bd.

zebrad/tests/acceptance.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8383/commits/fb2d13a7062d5d65db49cf89a472425c9efd59bd.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -290,29 +290,56 @@ impl NetworkUpgrade {
...
             .next()
     }
 
-    /// Returns the activation height for this network upgrade on `network`.
+    /// Returns the activation height for this network upgrade on `network`, or
+    ///
+    /// Returns the activation height of the first network upgrade that relies
+    /// on this network upgrade if there is no activation height for this network upgrade
+    /// such as on Regtest or a configured Testnet where multiple network upgrades have the
+    /// same activation height, or if one is omitted when others that rely on it are included.
```
upbqdn: ###### Superfluous Nitpick

I haven't encountered a statement claiming the network upgrades actually rely on the preceding ones, so I'd say we're better off avoiding it: 

```suggestion
    /// Returns the activation height of the first network upgrade that follows
    /// this network upgrade if there is no activation height for this network upgrade
    /// such as on Regtest or a configured Testnet where multiple network upgrades have the
    /// same activation height, or if one is omitted when others that follow it are included.
```

upbqdn (dismissed):

upbqdn (approved):

arya2:
@Mergifyio refresh

PR #8384, opened 2024-04-04
add(rpc): Add block times to verbose output of `getblock` RPC method

arya2:
## Motivation

This field is provided in zcashd and was asked about on Discord.

## Solution

- Adds `BlockHeader` state requests/responses (there were already `FindBlockHeaders` requests/responses, this one is simpler)
- Requests the block header from the state service in the `getblock()` RPC method
- Adds the timestamp (seconds since the unix epoch) as a `time` field to the verbose return value from the `getblock()` RPC

### Testing

Update the existing snapshot and vectors tests to include block time.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Adding other fields from the block header to the `getblock` RPC's verbose output

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -682,9 +682,6 @@ where
     // - use `height_from_signed_int()` to handle negative heights
     //   (this might be better in the state request, because it needs the state height)
     // - create a function that handles block hashes or heights, and use it in `z_get_treestate()`
-    // - Type second argument as an enum with variant for each verbosity level and a `Fields` variant
-    //   for specifying exactly which fields should be included in the return value. Consider replacing
-    //   `ReadRequest::BlockHeader` with a `PartialBlockData` request that accepts the fields?
     fn get_block(
         &self,
         hash_or_height: String,
@@ -733,7 +730,7 @@ where
                     }),
                     _ => unreachable!("unmatched response to a block request"),
                 }
-            } else if verbosity == 1 {
+            } else if verbosity == 1 || verbosity == 2 {
                 // # Performance
                 //
                 // This RPC is used in `lightwalletd`'s initial sync of 2 million blocks,
@@ -753,6 +750,8 @@ where
                 // All possible responses are valid, even if the best chain changes. Clients
                 // must be able to handle chain forks, including a hash for a block that is
                 // later discovered to be on a side chain.
+
+                let should_read_block_header = verbosity == 2;
 
                 let hash = match hash_or_height {
                     HashOrHeight::Hash(hash) => hash,
@@ -790,7 +789,7 @@ where
                 //
                 // We look up by block hash so the hash, transaction IDs, and confirmations
                 // are consistent.
-                let requests = [
+                let mut requests = vec![
                     // Get transaction IDs from the transaction index by block hash
                     //
                     // # Concurrency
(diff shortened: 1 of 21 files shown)


zebra-state/src/service.rs
```diff
@@ -1288,6 +1289,31 @@ impl Service<ReadRequest> for ReadStateService {
...
 
+            // Used by the get_block (verbose) RPC and the StateService.
+            ReadRequest::BlockHeader(hash_or_height) => {
+                let state = self.clone();
+
+                tokio::task::spawn_blocking(move || {
+                    span.in_scope(move || {
+                        let header = state.non_finalized_state_receiver.with_watch_data(
+                            |non_finalized_state| {
+                                read::block_header(
```
arya2, line 1300: There's very similar code in some of these read state requests that we could probably make more generic.

arya2:
We may want to get the full block data from the state service for `verbosity = 1` instead, because it'll be needed for returning the serialized block sizes and auth data roots (see https://github.com/zcash/zcash/blob/master/src/rpc/blockchain.cpp#L225).

Or we may not need to match zcashd's output exactly.

Update: 
There's a note about this RPC being used for lightwalletd's initial sync, so, for now it needs to load all its fields very efficiently. 

If we do want to match zcashd's output, we could:
- Move or duplicate the `debug_like_zcashd` config from the `mining` section to the `rpc` section and only read the entire block when it's enabled, and/or
- Add an optional argument for callers to provide a list of fields to be included in the return value.

zebra-rpc/src/methods.rs
```diff
@@ -675,10 +675,12 @@ where
     }
 
     // TODO:
-    // - use a generic error constructor (#5548)
     // - use `height_from_signed_int()` to handle negative heights
     //   (this might be better in the state request, because it needs the state height)
     // - create a function that handles block hashes or heights, and use it in `z_get_treestate()`
+    // - Type second argument as an enum with variant for each verbosity level and a `Fields` variant
+    //   for specifying exactly which fields should be included in the return value. Consider replacing
+    //   `ReadRequest::BlockHeader` with a `PartialBlockData` request that accepts the fields?
```
arya2: ~It would be nice to make this change in this PR to avoid always reading block headers while lightwalletd is syncing.~

Let's read the header when `verbosity` is 2 for now, I'll open an issue for allowing callers to specify exact fields.

Opened https://github.com/ZcashFoundation/zebra/issues/8390 for possibly allowing callers to specify exact fields, omitted `BlockHeader` request for verbosity 1 in https://github.com/ZcashFoundation/zebra/pull/8384/commits/c4fe4739aada4c5ab329829df1e386e14f875fd2.

mpguerra:
It seems like the scope for this PR is possibly expanding. Should we create an issue for all of the things we may want to do in order to support lightwalletd and potentially also match zcashd behaviour?

zebra-state/src/request.rs
```diff
@@ -614,6 +614,16 @@ pub enum Request {
...
 
+    /// Looks up a block header by hash or height in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// [`Response::BlockHeader(block::Header)`](Response::BlockHeader).
+    ///
+    /// Note: the [`HashOrHeight`] can be constructed from a [`block::Hash`] or
+    /// [`block::Height`] using `.into()`.
+    BlockHeader(HashOrHeight),
```
oxarbitrage, line 625: Why do we need `BlockHeader` requests and responses in the `Request` enum ? Isn't enough to have one in `ReadRequest`?

arya2: It's so the state service can handle a superset of the read state service requests and to fill in the `TryFrom` impls from `Request` to `ReadRequest` and `ReadResponse` to `Response`.

oxarbitrage: got it, thanks.

oxarbitrage (approved):

PR #8393, opened 2024-04-11
bump(zcash_script): Bump zcash script v0.1.15 and restore Windows support

oxarbitrage:
## Motivation

Updates `zcash_script` to `0.1.15` and `zcash_primitives` to `0.13.0`.

## Solution

Update zcash_script, restore windows support to CI and disable tests were needed.

### Testing

Windows CI and @arya2 made some local checks in windows.

## Review

Anyone can review.

### Reviewer Checklist


zebrad/tests/common/configs/v1.6.0-windows.toml
arya2, line 1: Why do we need this? If we get this PR in to restore Windows support before the release, do we want to increment the minor version and make it 1.7.0?

oxarbitrage: This is just a test but it seems in windows the `cache_dir` fields are in single quotes.

oxarbitrage: Apart from that(and some deny.toml issues) it seems to be working, last version running at https://github.com/ZcashFoundation/zebra/actions/runs/8653105561/job/23727560101?pr=8393

Ill confirm if that ends up being a successful run.

oxarbitrage: Ok, so tried several changes to a windows specific config however it seems there is something i am not seeing in regards to the format. I think a config file will need to be generated from a windows machine in order to fix that.

Anyway, i don't think is relevant to show that the tests for windows are actually working with the bumped zcash_script version. I will leave this PR open for discussion and then close.

arya2: Generating the config file from Windows didn't work either.

I think we should add `#[cfg(not(target_os = "windows"))]` above the call to `last_config_is_stored()` in `config_tests()`, we won't lose any test coverage if it's still running on other systems, and we won't have to add a Windows version for every new version of the config.

mpguerra: whatever we do with this PR, windows support should wait until @gustavovalverde has re-enabled any relevant CI for windows, so not in this release :)

oxarbitrage: Just for testing, after removing some of the test cases for Windows(https://github.com/ZcashFoundation/zebra/pull/8393/commits/d2c0af02b767ae381244abf0dc704cc61f0a9da4, https://github.com/ZcashFoundation/zebra/pull/8393/commits/a584822b1c4acb95d5be2998907fc840f59af640 and https://github.com/ZcashFoundation/zebra/pull/8393/commits/beec6cd65597fb4bcf597da4e24ddae2dd671b6f) the CI passed: https://github.com/ZcashFoundation/zebra/actions/runs/8711206097/job/23894807945?pr=8393

zebra-state/src/service/non_finalized_state/tests/prop.rs
```diff
@@ -451,6 +451,7 @@ fn finalized_equals_pushed_history_tree() -> Result<()> {
 /// Check that rejected blocks do not change the internal state of a genesis chain
 /// in a non-finalized state.
 #[test]
+#[cfg(not(target_os = "windows"))]
```
arya2: Did this one fail? It's passing for me locally on WIndows.

zebra-rpc/src/server/tests/vectors.rs
```diff
@@ -23,6 +23,7 @@ use super::super::*;
 
 /// Test that the JSON-RPC server spawns when configured with a single thread.
 #[test]
+#[cfg(not(target_os = "windows"))]
 fn rpc_server_spawn_single_thread() {
```
arya2, line 27: Optional/Nitpick: This test worked for me locally if we want to instead disable the `rpc_sever_spawn_parallel_threads()` on Windows and run this one? 

I think it's fine either way, but:
- If we leave the parallel one enabled, we'll be able to check if a warning is removed in a future version of Zebra
- If we enable this one instead, the test name will be more accurate if it's falling back on a single-threaded server.

I don't think it's important to run both tests on Windows, and I suspect there'll be slightly more frequent port conflicts in the tests on Windows if both are enabled because it uses a different [`random_known_port()`](https://github.com/ZcashFoundation/zebra/blob/test_zcash_script_v0.1.15/zebra-test/src/net.rs#L63).

Optional/Nitpick: Could we also fix the typo in the name? (it should be "rpc_server_spawn_parallel_threads", it's been there for a while, doesn't feel like it's worth its own PR but it would make it easier to search through the code or test results).

Not important, but, interestingly, it's logging that, ["Multi-threaded server is not available on Windows. Falling back to single thread."](https://github.com/ZcashFoundation/zebra/actions/runs/8653751415/job/23729618725#step:12:4667). I haven't found the log in Zebra's code or the relevant dependencies, I think I saw a merged PR in the JSON-RPC github repo for supporting multi-threaded servers on Windows, I'm not sure what's happening there.

I don't think we need to support multi-threaded RPC servers on Windows in Zebra for now, we could open an issue for re-enabling the test if we move away from JSON-RPC (and potentially reading the port from return value of `RpcServer::spawn` to avoid port conflicts).

arya2 (commented):
We may be able to re-enable the `rejection_restores_internal_state_genesis()` test. 

The logs suggest that the `rpc_server_spawn_single_thread` test is redundant on Windows, so I think it's fine to skip it, or to skip the multi-thread version of the test, either should work.

zebra-state/src/service/non_finalized_state/tests/prop.rs
oxarbitrage: It seems i got confused, this test is passing the widows CI. I restored it at https://github.com/ZcashFoundation/zebra/pull/8393/commits/60b43c0ff4a8c4c8305141c18ac3709712f75886

zebra-rpc/src/server/tests/vectors.rs
oxarbitrage: Disabling multi threads and leaving single ones seems to fail: https://github.com/ZcashFoundation/zebra/actions/runs/8784436404/job/24102673174?pr=8393

I reverted back to disabling single thread test in https://github.com/ZcashFoundation/zebra/pull/8393/commits/941a082de1a63fe2c82c69739baa5a5de9cac710

arya2 (approved):
Looks good

PR #8413, opened 2024-04-18
add(consensus/network): Adds `new_regtest()` constructors to `testnet::Parameters` and `Network`

arya2:
## Motivation

We want to support `Regtest` in Zebra.

This PR adds methods for creating a new `Network` or `testnet::Parameters` with the `Regtest` parameters as the values of fields that exist on `testnet::Parameters` so far. 

We'll need to add at least a few more fields (see 'Follow Up Work') that are needed for `Regtest` to `testnet::Parameters` and set them as their `Regtest` values in these methods for Zebra to support a minimal version of `Regtest`. Creating the genesis block may also rely on adding more parameters.

Closes #7839.

Depends-On: #8411.

#### Check before marking the PR as ready for review:
  - [x] Will the PR name make sense to users?
  - [x] Does the PR have a priority label?
  - [x] Have you added or updated tests?
  - [x] Is the documentation up to date?

##### For significant changes:
  - [x] Is there a summary in the CHANGELOG?
  - [x] Can these changes be split into multiple PRs?

_If a checkbox isn't relevant to the PR, mark it as done._

## Solution

- Adds `new_regtest()` constructor methods to `testnet::Parameters` and `zebra_chain::Network`
- Calls `Network::new_regtest()` instead of `unimplemented!()` in the `zebra_network::Config` deserialization
- Adds `with_sapling_hrps()` method on `ParametersBuilder` and uses it to set the `Regtest` HRPs in `new_regtest()`

Related Cleanups:
- Returns an error when deserializing the config if it uses any default initial peers for a configured testnet
- Allows for defining `testnet_parameters` in the config on `Mainnet` and `Regtest` (and the Regtest activation heights on other networks)
- Refactors Testnet parameters and Regtest activation heights part of  `zebra_network::Config` deserialization
- Uses default Testnet activation heights as the default configured Testnet activation heights
- Uses the activation heights from `ParametersBuilder` in `testnet::Parameters::default()`

### Testing

- Updates the latest stored config
- Checks the default `Regtest` activation heights in `activates_network_upgrades_correctly()`
- Adds a test for checking that `ParametersBuilder` validates configured Sapling HRPs as expected

## Review

Anyone can review.

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

[...]

PR #8421, opened 2024-04-20
add(consensus): Adds `disable_pow` and `genesis_hash` fields, converts `equihash::Solution` to an enum

### LINKED ISSUE #7846: Use network genesis_hash as first checkpoint when checkpoints are missing
## Motivation

The genesis block must be a checkpoint, because it can't be contextually validated.

### Possible Design

If Zebra doesn't find any checkpoints for the network, use the `genesis_hash` as the first and only checkpoint.

### LINKED ISSUE #8362: Add a `debug_disable_pow_check` field to `Network`
## Motivation

We want to use this field to disable the equihash and target difficulty checks when validating blocks for `Regtest`.

### LINKED ISSUE #8366: Add a `genesis_hash` field to `NetworkParameters`
## Motivation

We want to allow for configuring the genesis hash in `NetworkParameters` for supporting `Regtest` or custom testnets.

### LINKED ISSUE #8412: Copy or generate `Regtest` genesis block from zcashd
## Motivation

Zebra can't sync the `Regtest` genesis block from peers since `Regtest` is isolated, so we need to either generate the genesis block or copy it from zcashd if that works.

arya2:
## Motivation

We want to support Regtest in Zebra by committing the Regtest genesis block.

Closes #8412.
Closes #8366.
Closes #8362.
Closes #7846.

Depends-On: #8413.

## Solution

- Adds `genesis_hash` and `disable_pow` fields
- Adds a copy of the Regtest genesis block from zcashd
- Converts `equihash::Solution` to an enum to represent the smaller solutions for Regtest
- Updates `Network::checkpoint_list()` method to return the genesis checkpoint for Regtest

Related Cleanups:
- Moves the check for the genesis hash in `CheckpointList::from_list()` to `Network::checkpoint_list()` and check that its the first item in the checkpoints that's the genesis hash rather than the first item in the `BTreeMap`.

### Testing

- Adds a test to check that the Regtest genesis block from zcashd can be validated and committed.
- Updates a test to check that getting the Regtest checkpoint list won't panic

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Add a `target_difficulty_limit` field to `testnet::Parameters` and set it as `U256::from_big_endian(&[0x0f].repeat(8))` for Regtest


zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -27,6 +27,14 @@ pub const MAX_NETWORK_NAME_LENGTH: usize = 30;
 /// Maximum length for a configured human-readable prefix.
 pub const MAX_HRP_LENGTH: usize = 30;
 
+/// The block hash of the Testnet genesis block, `zcash-cli -regtest getblockhash 0`
```
upbqdn:
```suggestion
/// The block hash of the Regtest genesis block, `zcash-cli -regtest getblockhash 0`
```

zebra-consensus/src/checkpoint/list.rs
```diff
@@ -55,41 +55,57 @@ impl ParameterCheckpoint for Network {
...
-                .parse()
-                .expect("Hard-coded Testnet checkpoint list parses and validates"),
+        let checkpoints_for_network = match self {
+            Network::Mainnet => MAINNET_CHECKPOINTS,
+            Network::Testnet(params) if !params.is_regtest() => TESTNET_CHECKPOINTS,
+            // Returns only the genesis checkpoint for Regtest
+            Network::Testnet(_params) => {
+                // It's okay to skip checking the hard-coded genesis checkpoint is the genesis checkpoint
+                return CheckpointList::from_list([(block::Height(0), self.genesis_hash())])
+                    .expect("hard-coded checkpoint list parses and validates");
```
arya2: This should be moved to the match statement below when it's using any configured Testnet so that it's general to when the genesis hash from the `TESTNET_CHECKPOINTS` doesn't match the first checkpoint on Regtest or a configured Testnet.

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/8421/commits/d4f6a0ef6ba5b1ff2db8b25f5cb5a5c89956c155

Related: Should it require another configuration field for skipping validation before Canopy (if there are no checkpoints), or should it disallow pre-Canopy network upgrades altogether on Regtest or custom Testnets?

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -144,6 +161,15 @@ impl ParametersBuilder {
         self
     }
 
+    /// Parses a hex-encoded block hash and sets the genesis hash to be used in the [`Parameters`] being built.
```
upbqdn: ###### Optional Suggestion

I'm not even sure if the following suggestion makes sense.  Feel free to dismiss it right away. 

```suggestion
    /// Parses the hex-encoded block hash and sets it as the genesis hash in the [`Parameters`] being built.
```

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -144,6 +161,15 @@ impl ParametersBuilder {
...
+    pub fn with_genesis_hash(mut self, genesis_hash: impl fmt::Display) -> Self {
+        self.genesis_hash = genesis_hash
+            .to_string()
+            .parse()
+            .expect("configured genesis hash must parse");
```
upbqdn, line 169: ###### Optional Suggestion

I got a feedback from Deirdre and Teor that it's better to shift the responsibility for errors to the caller instead of panicking in the fn. 

I guess we could even refactor the fn so it takes hashes directly?

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -340,6 +340,9 @@ impl NetworkUpgrade {
                 self.next_upgrade()
                     .and_then(|next_nu| next_nu.activation_height(network))
             })
+            // TODO: Remove `.expect()` calls on the return value of this function (except for Genesis), or
+            //       update this method's return type
+            .or(Some(block::Height::MAX))
```
upbqdn: I'd say this merits a docs update.

zebra-chain/src/work/equihash.rs
```diff
@@ -29,16 +29,26 @@ pub struct SolverCancelled;
...
 pub(crate) const SOLUTION_SIZE: usize = 1344;
 
+/// The size of an Equihash solution in bytes on Regtest (always 36).
+pub(crate) const REGTEST_SOLUTION_SIZE: usize = 36;
+
 /// Equihash Solution in compressed format.
 ///
 /// A wrapper around [u8; 1344] because Rust doesn't implement common
 /// traits like `Debug`, `Clone`, etc for collections like array
 /// beyond lengths 0 to 32.
```
upbqdn: ###### Optional Suggestion


```suggestion
/// A wrapper around `[u8; n]` where `n` is the solution size because
/// Rust doesn't implement common traits like `Debug`, `Clone`, etc.
/// for collections like arrays beyond lengths 0 to 32.
```

zebra-chain/src/block/genesis/block-regtest-0-000-000.txt
```diff
@@ -0,0 +1 @@
+040000000000000000000000000000000000000000000000000000000000000000000000db4d7a85b768123f1dff1d4c4cece70083b2d27e117b4ac2e31d087988a5eac40000000000000000000000000000000000000000000000000000000000000000dae5494d0f0f0f2009000000000000000000000000000000000000000000000000000000000000002401936b7db1eb4ac39f151b8704642d0a8bda13ec547d54cd5e43ba142fc6d8877cab07b30101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff4d04ffff071f0104455a6361736830623963346565663862376363343137656535303031653335303039383462366665613335363833613763616331343161303433633432303634383335643334ffffffff010000000000000000434104678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5fac00000000
```
upbqdn, line 1: Let's move this file to `zebra-test`?

upbqdn (approved):

zebra-chain/src/parameters/network/testnet.rs
arya2: This could be good cleanup, it's probably less important since the builder struct is only used at startup, so I'd prefer to defer it until after the PRs have merged, I opened #8432 so we don't forget. I think the intention of panicking in these methods was conciseness where the methods are called and to avoid duplicate code, but in hindsight, that doesn't seem important.

zebra-chain/src/block/genesis/block-regtest-0-000-000.txt
arya2: It would be nice to have them all in one place, but it still seems wrong to put it in `zebra_test::vectors` since it's being used outside of Zebra's tests, @oxarbitrage could we leave this for now and potentially move them all to a more generically named folder like `block_data` later?

zebra-chain/src/parameters/network/testnet.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8421/commits/0f91dd03df5bc3c64977c18dffbf6ecfa4bcba96.

zebra-chain/src/parameters/network/testnet.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8421/commits/0f91dd03df5bc3c64977c18dffbf6ecfa4bcba96.

zebra-chain/src/work/equihash.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8421/commits/0f91dd03df5bc3c64977c18dffbf6ecfa4bcba96.

zebra-chain/src/parameters/network_upgrade.rs
arya2: Reverted this change in https://github.com/ZcashFoundation/zebra/pull/8421/commits/479b0b272eba902374e941bc225db58691ca71e4.

This is still going to cause all sorts of problems when at least a Sapling activation height isn't set, but the changes to fix that are somewhat substantial, most of the calls to `activation_height()` need to be updated, as do many calls to `Network.sapling_activation_height()`, perhaps more, so I'll do that in another PR.

zebra-chain/src/parameters/network/testnet.rs
upbqdn: > the intention of panicking in these methods was conciseness where the methods are called and to avoid duplicate code

This was my motivation the other day as well, and I also opened an issue to fix it.

Let's go with the current implementation.

upbqdn (dismissed):

zebra-chain/src/block/genesis/block-regtest-0-000-000.txt
upbqdn: > it still seems wrong to put it in `zebra_test::vectors` since it's being used outside of Zebra's tests

I agree. I didn't realize this.

zebra-chain/src/history_tree.rs
```diff
@@ -455,9 +455,15 @@ impl HistoryTree {
...
-            .activation_height(network)
-            .expect("Heartwood height is known");
+        let Some(heartwood_height) = NetworkUpgrade::Heartwood.activation_height(network) else {
+            assert!(
+                self.0.is_none(),
+                "history tree must not exist pre-Heartwood"
+            );
+
+            return Ok(());
+        };
```
upbqdn, line 465: @arya2 can you share which test was failing?

arya2: It was the new acceptance test, validating the regtest genesis block

arya2:
Looks like a test failed due to a port conflict: https://github.com/ZcashFoundation/zebra/actions/runs/8820097707/job/24212898628?pr=8421#step:12:2817

I'll re-run it once the rest of the workflow is done.

upbqdn (approved):

PR #8428, opened 2024-04-23
Adds `debug_validate_without_checkpoints` field to `testnet::Parameters` and ignores the ZIP-212 grace period on Regtest

arya2:
## Motivation

This PR allows for committing unvalidated blocks below the mandatory checkpoint height on Regtest and custom Testnets so their transaction outputs could be used in later blocks that _are_ validated.

~We may want to allow for committing pre-Canopy blocks on Regtest and custom Testnets, or we may want to disable pre-Canopy blocks on Regtest and custom Testnets.~

We may want to skip the ZIP-212 grace period on Regtest and custom Testnets, or we may want to add full semantic validation for blocks during the ZIP-212 grace period.

### Specifications

[ZIP-212](https://zips.z.cash/zip-0212#changes-to-the-process-of-receiving-sapling-or-orchard-notes)

## Solution

- Adds a `debug_validate_without_checkpoints` field to `testnet::Parameters`
- In the checkpoint verifier, sends the first queued block for a block height to the state service to be contextually validated and committed when the `debug_validate_without_checkpoints` field is set
- Skips the ZIP-212 grace period in the `mandatory_checkpoint_height()` method for non-default networks, this means Zebra likely won't support the ZIP-212 grace period on Regtest/custom-Testnets.

Related:
- Clears any initial peers when the network is set to `Regtest` so Zebra is isolated.

### Testing

This PR doesn't have any tests yet, I'll add tests if we want to go in this direction.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/parameters/network/testnet.rs
+++ b/zebra-chain/src/parameters/network/testnet.rs
@@ -73,6 +73,9 @@ pub struct ParametersBuilder {
     hrp_sapling_payment_address: String,
     /// A flag for disabling proof-of-work checks when Zebra is validating blocks
     disable_pow: bool,
+    /// A flag for allowing Zebra to try contextually validating the first queued block
+    /// for a given height below the mandatory checkpoint height.
+    debug_validate_without_checkpoints: bool,
 }
 
 impl Default for ParametersBuilder {
@@ -94,6 +97,7 @@ impl Default for ParametersBuilder {
                 .parse()
                 .expect("hard-coded hash parses"),
             disable_pow: false,
+            debug_validate_without_checkpoints: false,
         }
     }
 }
@@ -240,6 +244,15 @@ impl ParametersBuilder {
         self
     }
 
+    /// Sets the `debug_validate_without_checkpoints` flag to be used in the [`Parameters`] being built.
+    pub fn with_debug_validate_without_checkpoints(
+        mut self,
+        debug_validate_without_checkpoints: bool,
+    ) -> Self {
+        self.debug_validate_without_checkpoints = debug_validate_without_checkpoints;
+        self
+    }
+
     /// Converts the builder to a [`Parameters`] struct
     pub fn finish(self) -> Parameters {
         let Self {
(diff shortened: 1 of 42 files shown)


arya2:
We agreed to:
- Try contextually validating and committing the first queued block at a height without checkpoints on Regtest and custom Testnets
- Fully validate post-Canopy blocks with the ZIP-212 grace period rules (#8430)

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -73,6 +73,9 @@ pub struct ParametersBuilder {
     hrp_sapling_payment_address: String,
     /// A flag for disabling proof-of-work checks when Zebra is validating blocks
     disable_pow: bool,
+    /// A flag for allowing Zebra to try contextually validating the first queued block
+    /// for a given height below the mandatory checkpoint height.
+    debug_validate_without_checkpoints: bool,
```
arya2, line 78: Does anyone have an idea for a better name here?

zebra-state/src/service/read/find.rs
```diff
@@ -607,6 +607,8 @@ fn best_relevant_chain(
         .into_iter()
         .take(POW_MEDIAN_BLOCK_SPAN)
         .collect();
+
+    // TODO: Find out what zcashd does here.
```
arya2, line 611: It's in the [protocol spec](https://zips.z.cash/protocol/protocol.pdf), section 7.7.3, Difficulty Adjustment (p. 132):
`MedianTime(height ◦ N) := median([ nTime(𝑖) for 𝑖 from max(0, height − PoWMedianBlockSpan) up to height − 1 ])`

So it's the past 11 blocks, or if the chain is shorter than 12 blocks, it's every past block where the median is:
`median(𝑆) := sorted(𝑆)ceiling((length(𝑆)+1)/2)`

arya2:
I'll open PRs for:
- Serializing the parameters in the config without losing information first to make testing easier (creating a tokio runtime and calling `block_on()` to run `StartCmd.start()` would require adding a shutdown channel to kill the task, and using a regular tokio green thread won't work either).
- Making block template generation work when there are fewer than 11 blocks in the chain,

And then recreate this PR with more organized commits.

PR #8459, opened 2024-04-24
fix(state): Always return updated trees in `NoteCommitmentTree` update methods

arya2:
## Motivation

This PR makes the code a bit more clear.

~`Arc::make_mut()` will clone the value if there are other references to the value, so if a note commitment tree is being updated after its been requested from the state service but before its been dropped, it could, in rare cases, drop the updated note commitment tree and add the new block to the state with the previous note commitment tree.~

~This only affects the non-finalized state, and note commitment trees are only read by the `getblock` and `z_gettreestate` RPC methods.~

## Solution

- Use `.unwrap_or_clone()` instead to get a copy of the note commitment trees
- Update and return the local copy of the note commitment trees

Related Cleanups:
- Avoids creating a rayon scope if there are no note commitments in the block

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/parallel/tree.rs
+++ b/zebra-chain/src/parallel/tree.rs
@@ -132,10 +132,10 @@ impl NoteCommitmentTrees {
     /// Update the sprout note commitment tree.
     /// This method modifies the tree inside the `Arc`, if the `Arc` only has one reference.
     fn update_sprout_note_commitment_tree(
-        mut sprout: Arc<sprout::tree::NoteCommitmentTree>,
+        sprout: Arc<sprout::tree::NoteCommitmentTree>,
         sprout_note_commitments: Vec<sprout::NoteCommitment>,
     ) -> Result<Arc<sprout::tree::NoteCommitmentTree>, NoteCommitmentTreeError> {
-        let sprout_nct = Arc::make_mut(&mut sprout);
+        let mut sprout_nct = Arc::unwrap_or_clone(sprout);
 
         for sprout_note_commitment in sprout_note_commitments {
             sprout_nct.append(sprout_note_commitment)?;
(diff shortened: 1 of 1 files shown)


zebra-chain/src/parallel/tree.rs
```diff
@@ -79,8 +79,19 @@ impl NoteCommitmentTrees {
         let mut sapling_result = None;
         let mut orchard_result = None;
 
+        // Note: Only updating the note commitment trees when there are more notes that
+        //       need to be appended to the tree prevents unnecessarily duplicating note commitment
+        //       tree data in the non-finalized state.
```
arya2: This could maybe use some re-wording or be moved

zebra-chain/src/parallel/tree.rs
```diff
@@ -79,8 +79,19 @@ impl NoteCommitmentTrees {
...
+        let has_sprout_notes = !sprout_note_commitments.is_empty();
+        let has_sapling_notes = !sapling_note_commitments.is_empty();
+        let has_orchard_notes = !orchard_note_commitments.is_empty();
+
+        if !(has_sprout_notes || has_sapling_notes || has_orchard_notes) {
+            return Ok(());
+        }
```
arya2, line 88: This seems like an unnecessary refactor if we want to revert it.

arya2:
There was a known test failure on Windows: https://github.com/ZcashFoundation/zebra/actions/runs/8820859044/job/24215403949?pr=8459#step:12:4697

mpguerra:
> There was a known test failure on Windows: https://github.com/ZcashFoundation/zebra/actions/runs/8820859044/job/24215403949?pr=8459#step:12:4697

Do we need to disable some tests on Windows?

upbqdn:
@arya2, what issue is this PR supposed to fix?

arya2:
> Do we need to disable some tests on Windows?

@mpguerra That seems like a good idea for now, I opened #8468 for disabling the test that's failing and #8461 for checking if it's failing because of a bug in the production code on Windows or if it's a bug in the test.

> what issue is this PR supposed to fix?

@upbqdn It doesn't close any issues, I noticed this while double-checking that the `z_get_treestate` RPC method output includes notes in the block height at the hash or height provided as an argument, and it usually does, unless there are other `Arc` pointers to the same `NoteCommitmentTree` allocation.

Only the first commit is necessary, the second one, I thought for a moment that it would cause duplicate allocations for the same data, realized it won't because it only clones the `NoteCommitmentTree` if there are new notes that need to be appended, and added the comment for clarity around that.

Update: 

This was not an issue, I misunderstood what `make_mut()` was doing, the new code is equivalent to what it was doing before.

We also noticed that it's always cloning the note commitment trees that are being updated, I'll see if I can fix that quickly in this PR.

zebra-chain/src/parallel/tree.rs
arya2: There's always another Arc that owns these, the value will always be cloned. That's fine, we want to always clone the inner value, otherwise it would be updating the note commitment tree for the previous chain tip.

```suggestion
```

arya2:
This PR doesn't really change anything.

PR #8460, opened 2024-04-24
fix(rpc): Avoid possibly returning data from different blocks in `z_get_treestate` RPC method

arya2:
## Motivation

This avoids a concurrency bug where the `z_get_treestate` RPC method could return the note commitment trees for a different block than the one used to return the block hash and time.

## Solution

- Use the block hash to read Sapling/Orchard trees in `z_get_treestate` RPC method.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- BlockQuery state request?

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -1098,6 +1098,11 @@ where
                     data: None,
                 })?;
 
+            // # Concurrency
+            //
+            // For consistency, this lookup must be performed first, then all the other
+            // lookups must be based on the hash.
+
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
thanks!

upbqdn (commented):
I can't see how the original code could return the wrong data. @arya2 or @oxarbitrage, could you provide more context?

arya2:
> I can't see how the original code could return the wrong data.

If the best chain changes between calls for the block data and its Sapling/Orchard trees when queried by height, it could've returned the trees for the new best chain at that height while returning the block data from the old best chain at that height.

Issue #7119, opened 2023-06-30
Tracking: Add regtest network


arya2:
## Motivation

This has been requested, and could be useful for testing.

Depends-On: #7845

## Possible Design

- [x] #7839
- [x] #8327
- [x] Disable peer connections (or only connect to nodes with the same network magic)
- Skip loading checkpoints and update `NetworkUpgrade::activation_height` to use zcashd defaults or user-defined heights
  - [x] #7846
- [x] #8479
  - [x] remove the implementation constraint that `PoWLimit` must be less than or equal to `(2^256 - 1) / PoWAveragingWindow` (approximately `2^251`) to avoid overflow
  - [x] calculate difficulty adjustment for heights less than 17
- [x] Skip `SLOW_START_INTERVAL` checks ([`nSubsidySlowStartInterval` is 0 for regtest in zcashd](https://github.com/zcash/zcash/blob/master/src/chainparams.cpp#L639))
- [x] #8478
- [x] #8361
- [x] https://github.com/ZcashFoundation/zebra/issues/9182
- [ ] https://github.com/ZcashFoundation/zebra/issues/8367
- [ ] Update prefixes (optional)
- Match other differences between zcashd testnet vs regtest in [chainparams](https://github.com/zcash/zcash/blob/master/src/chainparams.cpp)

teor2345:
These are just some notes for if we eventually do this ticket. No action is needed at this time.

-----

We might also want to take a look at:
- #1135

It seems like some of these changes would benefit from the Regtest variant containing a consensus rule configuration. (Like zcashd has for all its network variants.)

It would also be a lot easier to make this change if we refactored all our network conditions into methods on the Network enum. For example, a has_slow_start() method, rather than a match statement buried in the zebra-consensus code.

> * update `From<Network> for &'static str` for `db_path()`

We either need to work out how to make sure the consensus rules match the cached state, or make regtest ephemeral by default.

This isn't an issue for mainnet or testnet because their rules are fixed and can't be configured.

> * Set `peerset_initial_target_size` to 0 (Zebra should not make any peer connections)

This change isn't enough to stop Zebra making outbound connections, because Zebra will try to connect to new peers every minute or so. Instead, the initial seed peers list needs to be empty for regtest.

> * Update prefixes

Which prefixes?

[...]

Issue #8430, opened 2024-04-23
Add semantic block validation during the ZIP-212 grace period after Canopy activation


arya2:
## Motivation

We want to fully validate Canopy blocks and abide by the ZIP-212 grace period where note plaintext lead bytes are allowed to be either `0x01` or `0x02` so that Regtest is as close to the Mainnet consensus rules as possible. Prior to Canopy activation those lead bytes must be `0x01`, and after the grace period, they must be `0x02`.

This rule only applies to nodes and not wallets, and librustzcash may only allow for lead bytes of `0x02` after Canopy activation, so Zebra is currently mandating checkpoint validation until the end of the ZIP-212 grace period.

### Possible Design

The grace period after Canopy should only apply until the next network upgrade (NU5) is activated, so it should be the lesser of `Canopy Activation Height + ZIP_212_DURATION` or `NU5 Activation Height` 

### Testing 

- Ensure that Zebra can validate transactions with note plaintexts that have lead bytes of `0x01` as well as `0x02` during the ZIP-212 grace period
- Ensure that Zebra will only validate transactions with note plaintexts that have lead bytes of `0x02` after the ZIP-212 grace period

arya2:
Both the latest version of `zcash_primitives` and the version Zebra currently uses (`0.13.0`) allow for plaintext lead bytes of `0x01` or `0x02` during the grace period, see:
- [latest version](https://github.com/zcash/sapling-crypto/blob/main/src/note_encryption.rs#L401)
- [0.13.0](https://github.com/zcash/librustzcash/blob/zcash_primitives-0.13.0/zcash_primitives/src/sapling/note_encryption.rs#L380)

arya2:
Let's wait until Zebra is using `zcash_primitives` `v0.14.0` and then pass `Zip212Enforcement::On` in the call to `try_sapling_note_decryption()` in `zcash_note_encryption`

mpguerra:
depends-on: #8540

PR #8475, opened 2024-04-30
add(consensus): Adds block construction and validation for NU5 from block height 1 for Regtest

### LINKED ISSUE #8280: Tracking: Enable short-chain single-node network for testing purposes
### Motivation

In ‘Zcash-Tx-Tool’ we use a custom branch of Zebra. 

Here is an overview of current changes we made split by feature ([Draft PR](https://github.com/ZcashFoundation/zebra/pull/8279)):

0) Configuration

There is a modified configuration file ‘singlenode-config.toml’ that sets peer lists to empty ones,  provides a miner address and sets the listen address to 0.0.0.0. This config file can be easily moved to ‘zcash-tx-tool’ repo to be used only in respective dockerfile

1) Disable proof-of-work

This one looks like the easiest one to be implemented as a build feature. There are 2 changes in zebra-consensus crate: block.rs#L182 and checkpoint.rs#L604 where we basically leave out the PoW check. Currently we do it always for simplification but this can be changed to only work if a feature flag ‘tx-tool-test’ (or similar) was used

2) Enable short chain

First of all TESTNET_ACTIVATION_HEIGHTS[Nu5] constant  is modified in network_upgrade.rs#L113 from 1_842_420 to 1_060_755. If there are plans to make this list a config rather than const or somehow use FAKE_TESTNET_ACTIVATION_HEIGHTS - it would be the best to use that, but currently the only way seems to be a config option around this const which looks rather ugly but will work

Secondly in get_block_template.rs#116 and get_block_template_rpcs.rs#L499 check_synced_to_tip is disabled. This can be hidden behind a feature flag as well

Last one is file ‘test-checkpoint.txt’ which is shortened to be able to use a block verifier instead of checkpoint verifier. Unfortunately this file is not provided to a running Zebra as external but instead included as a const string during build (see list.ts#L44). This complicates it a bit. again we can select between short and normal version based on a feature flag

### Scope

- [x] https://github.com/ZcashFoundation/zebra/issues/8363
- [x] https://github.com/ZcashFoundation/zebra/issues/7970
- [ ] https://github.com/ZcashFoundation/zebra/issues/8361, or
  - [ ] Otherwise avoid returning initial or cached peers
- [x] https://github.com/ZcashFoundation/zebra/issues/8362

### LINKED ISSUE #8363: Skip call to `check_synced_to_tip()` in getblocktemplate method on Regtest
## Motivation

This check can cause problems for when trying to mine blocks on a shortened chain in integration tests

### Possible Design

We could skip the check altogether for test networks.

arya2:
## Motivation

We want to support Regtest in Zebra for integration testing.

Closes #8363.
Closes #8280.
Part of #7119.

~This PR shouldn't change much, the `regtest_nu5_activation_height` field in the config should be removed, but that's a minor change.~

Update: This PR shouldn't change anymore except to address review comments.

### Specifications

- [ZIP 212](https://zips.z.cash/zip-0212#changes-to-the-process-of-receiving-sapling-or-orchard-notes)
- Mean target difficulty on p. 132 of the spec:
> MeanTarget(height ◦ N) :=
> PoWLimit, if height ≤ PoWAveragingWindow
> mean([ToTarget(nBits(𝑖)) for 𝑖 from height−PoWAveragingWindow up to height−1]),
> otherwise.

- ~I couldn't find the `POW_ADJUSTMENT_BLOCK_SPAN` in the protocol spec, I don't think this should be a problem because there are always more than that many blocks where a check was replaced with `relevant_chain.is_empty()` by Canopy activation.~

Update: 
It's based on MedianTime and ActualTimespan in the [specification](https://zips.z.cash/protocol/protocol.pdf), PoWAveragingWindow is also used in MeanTarget.

## Solution

- Checks if the network `is_regtest()` in the start command, and if it is, commits the Regtest genesis block and skips starting the syncer
- Skips the ZIP 212 grace period for non-default networks
- Sets the mandatory checkpoint height at one block below `Canopy` activation for non-default networks (this may need some follow up changes, or changes in this PR)
- Adds funding streams and minimal protocol version for `NetworkKind::Regtest`
- Returns the reserved activation chain history root hash in the `getblocktemplate` method for blocks that should activate `Heartwood` network upgrade
- Checks Heartwood was activated at a given block height even if the current network upgrade is NU5
- Uses a slow start interval of `Height(0)` if PoW is disabled (this should be changed in a follow up PR or in this PR)
- Uses fewer than 11 blocks for calculating the median time past
- Uses fewer than 17 blocks for calculating the mean target difficulty

### Testing

Added a test that gets block templates and submits them as blocks

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

- Some changes from https://github.com/ZcashFoundation/zebra/pull/8428
- Adding a fields for:
  - Slow start interval
  - Disabling peer connections
  - Funding streams
  - Minimum protocol version
- Using `zcash_address::Network` instead of `NetworkKind` in transparent addresses?


zebra-chain/src/block/commitment.rs
```diff
@@ -115,7 +115,7 @@ impl Commitment {
                 Ok(root) => Ok(FinalSaplingRoot(root)),
                 _ => Err(InvalidSapingRootBytes),
             },
-            Heartwood if Some(height) == Heartwood.activation_height(network) => {
+            Heartwood | Nu5 if Some(height) == Heartwood.activation_height(network) => {
```
upbqdn, line 122: Why did we not need Nu5 here before?

zebra-state/src/service/check.rs
```diff
@@ -80,23 +80,25 @@ where
...
-        // verified blocks, so there will be at least 1 million blocks in the state when it is
-        // called. So this error should never happen.
-        #[cfg(not(test))]
+    // In production, blocks without enough context are invalid.
+    //
+    // The BlockVerifierRouter makes sure that the first 1 million blocks (or more) are
+    // checkpoint verified. The state queues and block write task make sure that blocks are
+    // committed in strict height order. But this function is only called on semantically
+    // verified blocks, so there will be at least 1 million blocks in the state when it is
+    // called. So this error should never happen.
```
arya2:
```suggestion
    // called. So this error should never happen on Mainnet or the default Testnet.
```

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -77,7 +85,9 @@ pub fn block_subsidy(height: Height, network: &Network) -> Result<Amount<NonNega
         return Ok(Amount::zero());
     };
 
-    if height < SLOW_START_INTERVAL {
+    // TODO: Add this as a field on `testnet::Parameters` instead of checking `disable_pow()`, this is 0 for Regtest in zcashd,
+    //       see <https://github.com/zcash/zcash/blob/master/src/chainparams.cpp#L640>
+    if height < SLOW_START_SHIFT && !network.disable_pow() {
```
arya2:
```suggestion
    if height < SLOW_START_INTERVAL && !network.disable_pow() {
```

zebra-state/src/service/check/difficulty.rs
```diff
@@ -64,14 +64,14 @@ pub(crate) struct AdjustedDifficulty {
     /// The `header.difficulty_threshold`s from the previous
     /// `PoWAveragingWindow + PoWMedianBlockSpan` (28) blocks, in reverse height
     /// order.
-    relevant_difficulty_thresholds: [CompactDifficulty; POW_ADJUSTMENT_BLOCK_SPAN],
+    relevant_difficulty_thresholds: Vec<CompactDifficulty>,
```
arya2, line 68: I was wondering about using `Vec`s, fixed-size arrays of `Option`s, `AtLeastOne`, or a new type that allows arbitrary constraints on minimum/maximum `Vec` lengths, any preferences? If we want to change this, we could do it in this PR or in a follow up PR.

Update: Opened #8484 to make this change later.

zebra-chain/src/parameters/network.rs
```diff
@@ -251,8 +254,19 @@ impl Network {
...
+        };
+
+        if is_a_default_network {
+            (canopy_activation + ZIP_212_GRACE_PERIOD_DURATION)
+                .expect("ZIP-212 grace period ends at a valid block height")
+        } else {
+            canopy_activation
+                .previous()
+                .expect("Canopy activation must be above Genesis height")
+        }
```
arya2, line 269: Update: This isn't necessary, we can remove the ZIP-212 grace period altogether, and the rule seems to apply after NU5 activation, see https://github.com/ZcashFoundation/zebra/issues/8430#issuecomment-2087870452 and #8485.

This would work too, optional:

```suggestion
        let default_mandatory_checkpoint_height = (canopy_activation
            + ZIP_212_GRACE_PERIOD_DURATION)
            .expect("ZIP-212 grace period ends at a valid block height");

        if is_a_default_network {
            default_mandatory_checkpoint_height
        } else {
            let nu5_activation_height = NetworkUpgrade::Nu5
                .activation_height(self)
                .expect("Canopy activation height must be present for both networks");

            default_mandatory_checkpoint_height.min(
                nu5_activation_height
                    .previous()
                    .expect("NU5 activation must be above Genesis height"),
            )
        }
```

arya2:
~Running full sync test here: https://github.com/ZcashFoundation/zebra/actions/runs/8904378011~

Update: 

Applied the correct slow start interval, cancelled the first full sync test, and running a new full sync test here: https://github.com/ZcashFoundation/zebra/actions/runs/8904708364

I'm also running one locally with checkpoint sync disabled on the `lower-mandatory-checkpoint-height` branch that's based on this branch.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
```diff
@@ -305,7 +305,15 @@ pub fn generate_coinbase_and_roots(
     // Calculate block default roots
     //
     // TODO: move expensive root, hash, and tree cryptography to a rayon thread?
-    let default_roots = calculate_default_root_hashes(&coinbase_txn, mempool_txs, history_tree);
+    let chain_history_root = history_tree
+        .hash()
+        .or_else(|| {
+            (NetworkUpgrade::Heartwood.activation_height(network) == Some(height))
```
upbqdn: Why do we need this condition, and why `Heartwood`?

zebra-state/src/service/check.rs
```diff
@@ -80,23 +80,25 @@ where
...
-        #[cfg(not(test))]
+    // In production, blocks without enough context are invalid.
+    //
+    // The BlockVerifierRouter makes sure that the first 1 million blocks (or more) are
+    // checkpoint verified. The state queues and block write task make sure that blocks are
+    // committed in strict height order. But this function is only called on semantically
+    // verified blocks, so there will be at least 1 million blocks in the state when it is
+    // called. So this error should never happen on Mainnet or the default Testnet.
+    #[cfg(not(test))]
+    if relevant_chain.is_empty() {
```
upbqdn, line 107: The previous implementation used to return `Err` when `relevant_chain.len() < POW_ADJUSTMENT_BLOCK_SPAN`, but we're now changing the condition to `relevant_chain.is_empty()`. I didn't find any relevant docs for `POW_ADJUSTMENT_BLOCK_SPAN`, and I haven't figured out why exactly we were doing the check, and if we can change it.

The docs for this fn say it panics when "the state contains less than 28 (`POW_ADJUSTMENT_BLOCK_SPAN`) blocks", but I don't see the panic.

zebra-state/src/service/check/difficulty.rs
```diff
@@ -314,22 +315,30 @@ impl AdjustedDifficulty {
...
-    ) -> DateTime<Utc> {
+    ///
+    /// # Panics
+    ///
+    /// If provided an empty Vec
+    pub(crate) fn median_time(mut median_block_span_times: Vec<DateTime<Utc>>) -> DateTime<Utc> {
         median_block_span_times.sort_unstable();
-        median_block_span_times[POW_MEDIAN_BLOCK_SPAN / 2]
+
+        // > median(𝑆) := sorted(𝑆)ceiling((length(𝑆)+1)/2)
```
upbqdn: Let's put the `ceiling` into a subscript since its output is an index.


```suggestion
        // > median(𝑆) := sorted(𝑆)_{ceiling((length(𝑆)+1)/2)}
```

zebra-state/src/service/check/difficulty.rs
```diff
@@ -314,22 +315,30 @@ impl AdjustedDifficulty {
...
+        // <https://zips.z.cash/protocol/protocol.pdf>, section 7.7.3, Difficulty Adjustment (p. 132)
+        let median_idx = (POW_MEDIAN_BLOCK_SPAN / 2).min(median_block_span_times.len() / 2);
```
upbqdn: I'm guessing we're doing it this way because zcashd does it like this?

upbqdn (dismissed):

zebra-chain/src/block/commitment.rs
arya2: I think we always want to check if a height is the Heartwood activation height, but until now, we've known that it's only possible if the current network upgrade is Heartwood.

With the hard-coded activation heights we were using before, there were no cases where multiple network upgrades were activated at the same height, but currently on Regtest, `Genesis` is activated at height 0 and every other network upgrade is activated at height 1.

NetworkUpgrade::current() returns the latest network upgrade that's activated at the provided height, so on Regtest for heights above height 0, it returns NU5, and it's possible for the current network upgrade to be NU5 (or Canopy, or any network upgrade above Heartwood) at the Heartwood activation height.

We may want to check if it's the Heartwood activation height for Canopy as well, I only added NU5 because the check is inefficient (though it may not have a measurable impact on performance).

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: Chain history roots were activated at Heartwood, it may be useful to add a comment here.

The null root hash is a reserved value for the chain history activation.

zebra-state/src/service/check.rs
arya2: > The docs for this fn say it panics when "the state contains less than 28 (POW_ADJUSTMENT_BLOCK_SPAN) blocks", but I don't see the panic.

I don't think I've seen it panic either, it may have just returned an error from the `submitblock` RPC method. We may want to update the comment.

zebra-state/src/service/check/difficulty.rs
```diff
@@ -314,22 +315,30 @@ impl AdjustedDifficulty {
...
+    /// # Panics
+    ///
+    /// If provided an empty Vec
+    pub(crate) fn median_time(mut median_block_span_times: Vec<DateTime<Utc>>) -> DateTime<Utc> {
         median_block_span_times.sort_unstable();
-        median_block_span_times[POW_MEDIAN_BLOCK_SPAN / 2]
+
+        // > median(𝑆) := sorted(𝑆)_{ceiling((length(𝑆)+1)/2)}
+        // <https://zips.z.cash/protocol/protocol.pdf>, section 7.7.3, Difficulty Adjustment (p. 132)
+        let median_idx = (POW_MEDIAN_BLOCK_SPAN / 2).min(median_block_span_times.len() / 2);
```
arya2:
```suggestion
        let median_idx = median_block_span_times.len() / 2;
```

zebra-state/src/service/check/difficulty.rs
arya2: > I'm guessing we're doing it this way because zcashd does it like this?

It wasn't actually necessary to get the minimum of the two values, so long as the max length of the provided Vec is 11. Fixed in https://github.com/ZcashFoundation/zebra/pull/8475/commits/5be5ee1aeb7e2889bed34b87cd30810a5136e29d.

zebra-state/src/service/check.rs
upbqdn: Yea, let's remove that bit. I guess this PR is a good place to do that.

upbqdn: Regarding the original check, should we keep it for non-regtest networks?

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
upbqdn: I looked at this again. If we got no history tree for the Heartwood activation height, shouldn't we consider that an error and abort instead of using the null hash? (Even though we know that the null hash corresponds to the Heartwood activation height.)

zebra-chain/src/block/commitment.rs
upbqdn: This is a great description of the general situation. I think you should include it somewhere in the codebase.

upbqdn (dismissed):

zebra-state/src/service/check.rs
arya2: Ah, I think `POW_ADJUSTMENT_BLOCK_SPAN` is defined as PoWAveragingWindow + PoWMedianBlockSpan here (page 132 of the spec): 
`ActualTimespan(height : N) := MedianTime(height) − MedianTime(height − PoWAveragingWindow)`
`MedianTime(height : N) := median([[ nTime(𝑖) for 𝑖 from max(0, height − PoWMedianBlockSpan) up to height − 1 ]])`

MedianTime has a floor of 0, so, I think we're okay to use this for Mainnet and the default Testnet too.

It's also used in MeanTarget unless `height ≤ PoWAveragingWindow`

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: I think this part is confusing because the `height` parameter should be `next_height`, the condition in the `or_else()` would be true if we got no history tree for the Genesis activation height.

zebra-state/src/service/check.rs
arya2: > Yea, let's remove that bit. I guess this PR is a good place to do that.

Ah, it would've panicked at the time that the doc was written because it was using an assertion instead of returning an error.

zebra-chain/src/block/commitment.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/commit/2ad072a19f1713090bbbaccabb5d3a7d381a55f1.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
arya2: Renamed the parameter in https://github.com/ZcashFoundation/zebra/commit/2ad072a19f1713090bbbaccabb5d3a7d381a55f1.

zebra-state/src/service/check.rs
arya2: Removed the incorrect docs about the function panicking and added a comment about why it's okay to use a relevant chain that's shorter than the `POW_ADJUSTMENT_BLOCK_SPAN` in https://github.com/ZcashFoundation/zebra/commit/2ad072a19f1713090bbbaccabb5d3a7d381a55f1.

zebra-rpc/src/methods/get_block_template_rpcs/get_block_template.rs
upbqdn: > I think this part is confusing because the `height` parameter should be `next_height`, the condition in the `or_else()` would be true if we got no history tree for the Genesis activation height.

Thanks for the clarification. I didn't realize this.

upbqdn (approved):

arya2:
@Mergifyio refresh

PR #8485, opened 2024-05-01
change(consensus): Lower mandatory checkpoint height

### LINKED ISSUE #8430: Add semantic block validation during the ZIP-212 grace period after Canopy activation

arya2:
## Motivation

We want to support full validation of blocks from Canopy activation onwards, particularly for full validation of Canopy blocks on Regtest (though this use case also relies on supporting block construction for Canopy blocks).

Closes #8430.

Depends-On: #8475.

### Specifications

https://zips.z.cash/zip-0212

## Solution

Lowers mandatory checkpoint height to the block before Canopy activation.

### Testing

Running a sync test locally with checkpoint sync disabled and an added checkpoint at height `1_046_399` until it gets past the grace period.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -5,7 +5,7 @@ use std::{fmt, str::FromStr, sync::Arc};
 use thiserror::Error;
 
 use crate::{
-    block::{self, Height, HeightDiff},
+    block::{self, Height},
     parameters::NetworkUpgrade,
 };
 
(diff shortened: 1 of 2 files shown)


upbqdn (commented):
I don't think we can do this change. https://github.com/ZcashFoundation/zebra/issues/8430#issuecomment-2087870452 implies that Zebra would allow the 0x01 prefix for coinbase txns during the grace period if we removed checkpointing. However, ZIP-212 says:

> #### Consensus rule change for coinbase transactions
> 
> _After the activation of this ZIP, any Sapling output of a coinbase transaction that is decrypted to a note plaintext as specified in [17](https://zips.z.cash/zip-0212#zip-0213), MUST have note plaintext lead byte equal to 0x02. This applies even during the “grace period”, and also applies to funding stream outputs [16](https://zips.z.cash/zip-0212#zip-0207) sent to shielded payment addresses, if there are any. Since NU5 activates after the end of the grace period [19](https://zips.z.cash/zip-0212#zip-0252), Orchard outputs will always require a note plaintext lead byte equal to 0x02._ 


Note that Zebra is not able to decrypt notes of non-coinbase txns, but decrypts coinbase ones because of this rule:

> [Heartwood onward] All Sapling and Orchard outputs in coinbase transactions MUST decrypt to a note plaintext , i.e. the procedure in § 4.19.3 ‘Decryption using a Full Viewing Key (Sapling and Orchard)’ on p. 67 does not return ⊥, using a sequence of 32 zero bytes as the outgoing viewing key . (This implies that before Canopy activation, Sapling outputs of a coinbase transaction MUST have note plaintext lead byte equal to 0x01.)

arya2:
> I don't think we can do this change. 

Yeah, I misunderstood the doc comment on that constant, we can do this easily once Zebra's using zcash_primitives v0.15.0, see https://github.com/ZcashFoundation/zebra/issues/8430#issuecomment-2093509723

PR #8505, opened 2024-05-03
change(consensus): Allow configurable NU5 activation height on Regtest

arya2:
## Motivation

We want to allow configuring an NU5 activation height on Regtest. This was removed in a recent PR because I thought Zebra didn't support Canopy block construction, it does, it was the `proposal_block_from_template()` function used in tests and the internal-miner that only worked with NU5.

## Solution

- Accept a `current_network_upgrade` argument in `proposal_block_from_template()`
- Select which field from the template should be used in the block based on the current network upgrade
- Use the `testnet_parameters.activation_heights.nu5` field as the NU5 activation height on Regtest

Related Changes:
- Updates `with_activation_heights()` to overwrite activation heights for all network upgrades except Genesis
- Adds `current_with_activation_height()` method and uses it in `Commitment::from_bytes()`

### Testing

The `regtest_submit_blocks()` test was updated to submit Canopy blocks and switch to NU5 blocks at height 100

## Review

Anyone can review.

### Reviewer Checklist


oxarbitrage (approved):
looks good to me, thanks!

Issue #8183, opened 2024-01-22
Restore internal miner


oxarbitrage:
We removed the equihash custom dependency in Zebra that was part of the internal miner support. Restore the function once this pull requests are merged and a new version of the `equihash` is published:

- https://github.com/zcash/librustzcash/pull/1088
- https://github.com/zcash/librustzcash/pull/1083

Internal miner support was removed in https://github.com/ZcashFoundation/zebra/pull/8184

PR #8506, opened 2024-05-04
change(mining): Restores parts of the internal-miner feature for use on Regtest

arya2:
## Motivation

We want Zebra to have an internal miner for use on Testnet and Regtest.

Part of #8183.

Depends-On: #8505.

## Solution

Restores parts of the internal-miner feature and uses a empty `Solution` instead of the equihash tromp solver.

### Testing

Manually tested.

## Review

Anyone can review.

### Reviewer Checklist

### DIFF
--- a/zebra-rpc/src/config/mining.rs
+++ b/zebra-rpc/src/config/mining.rs
@@ -29,6 +29,15 @@ pub struct Config {
     ///
     /// This developer-only config is not supported for general use.
     pub debug_like_zcashd: bool,
+
+    /// Mine blocks using Zebra's internal miner, without an external mining pool or equihash solver.
+    ///
+    /// This experimental feature is only supported on testnet.
+    /// Mainnet miners should use a mining pool with GPUs or ASICs designed for efficient mining.
+    ///
+    /// The internal miner is off by default.
+    #[cfg(feature = "internal-miner")]
+    pub internal_miner: bool,
 }
 
 impl Default for Config {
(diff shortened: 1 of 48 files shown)


mpguerra:
@arya2 will this close #8183 ?

arya2:
> @arya2 will this close #8183 ?

Not yet. This doesn't mine valid equihash solutions, so it won't work on the default testnet, and it's single-threaded since multiple threads are unnecessary while it skips checking the proof of work.

zebra-rpc/src/config/mining.rs
```diff
@@ -29,6 +29,15 @@ pub struct Config {
     ///
     /// This developer-only config is not supported for general use.
     pub debug_like_zcashd: bool,
+
+    /// Mine blocks using Zebra's internal miner, without an external mining pool or equihash solver.
+    ///
+    /// This experimental feature is only supported on testnet.
+    /// Mainnet miners should use a mining pool with GPUs or ASICs designed for efficient mining.
```
oxarbitrage: I can't see where in the code we are not allowing the miner to run in mainnet.

I also think the text should say the feature is supported in public testnet and custom testnets ?

arya2: > I can't see where in the code we are not allowing the miner to run in mainnet.

It's not explicitly disallowed anywhere, it just wouldn't work very well with the CPU equihash solver and won't work at all until that's restored.

> I also think the text should say the feature is supported in public testnet and custom testnets ?

Updated the docs in https://github.com/ZcashFoundation/zebra/pull/8506/commits/752949e8f4c58c64a0cb77e6d820b758624d3c2a to say that it's currently only supported on Regtest since it's currently the only one that disables proof of work validation.

oxarbitrage: do you think we should panic/error if the network is not regtest ?

arya2: It wouldn't hurt to log a warning and skip spawning the miner on unsupported networks (if `disable_pow` is false) until the equihash solver is restored, but I don't think it's necessary for an experimental feature.

Once the equihash solver is restored, CPU mining on Mainnet still isn't worthwhile, but I see no harm in allowing it.

I do think we should assign templates a random nonce for networks with higher difficulties where multiple zebra nodes may be mining.

oxarbitrage (approved):

oxarbitrage:
@mergifyio refresh

PR #8522, opened 2024-05-13
deps(ecc): Update `zcash_primitives` and `orchard` only for zebra-chain crate

oxarbitrage:
## Motivation

We are trying to split the dependency upgrades needed for zcashd 5.9.0. As doing everything at once is too complicated we are trying to split it up in multiple pull requests. This pull request do the upgrades only for `zebra-script` crate so is not mergable.

## Solution

By now, this PR is just compare with what @upbqdn was working on in regards to upgrade the `zebra-scan` crate.

### Testing

- `cargo test --package zebra-chain` pass.
- `cargo test --package zebra-consensus` pass.

### Reviewer Checklist


zebra-chain/src/primitives/zcash_note_encryption.rs
```diff
@@ -20,16 +20,14 @@ pub fn decrypts_successfully(transaction: &Transaction, network: &Network, heigh
...
 
     if let Some(bundle) = alt_tx.sapling_bundle() {
         for output in bundle.shielded_outputs().iter() {
-            let recovery = zcash_primitives::sapling::note_encryption::try_sapling_output_recovery(
-                network,
-                alt_height,
+            let recovery = sapling::note_encryption::try_sapling_output_recovery(
                 &null_sapling_ovk,
                 output,
+                sapling::note_encryption::Zip212Enforcement::GracePeriod,
```
oxarbitrage: This was because the `coinbase_outputs_are_decryptable_for_historical_blocks` test in `zebra-consensus` pass only if this variable isin the grace period, however i think we need to compare with the height of the output to decide what goes here ?

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -29,11 +29,11 @@ impl zp_tx::components::transparent::Authorization for TransparentAuth<'_> {
 // In this block we convert our Output to a librustzcash to TxOut.
 // (We could do the serialize/deserialize route but it's simple enough to convert manually)
 impl zp_tx::sighash::TransparentAuthorizingContext for TransparentAuth<'_> {
-    fn input_amounts(&self) -> Vec<zp_tx::components::amount::Amount> {
+    fn input_amounts(&self) -> Vec<zp_tx::components::amount::NonNegativeAmount> {
         self.all_prev_outputs
             .iter()
             .map(|prevout| {
-                zp_tx::components::amount::Amount::from_nonnegative_i64_le_bytes(
+                zp_tx::components::amount::NonNegativeAmount::from_nonnegative_i64_le_bytes(
```
oxarbitrage: We need to convert here from zebra `Amount<NonNegative>` type to a zcash_primitives `NonNegativeAmount` here. I think `from_nonnegative_i64_le_bytes` is good enough here as zebra `Amount` is `i64`.

oxarbitrage (commented):
Almost everything here is renames and reimport except for some comments i added to the code as review comments.

zebra-chain/src/primitives/zcash_note_encryption.rs
arya2: Oh good, they've changed this already, I thought this was in 0.15.0.

ZIP-212 enforcement can always be on here because we only call this function after Canopy activation. We should leave a comment about needing to check the height if Zebra's mandatory checkpoint height ever falls below the last block before Canopy activation (or we could just double check the height here anyways).

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -269,10 +261,10 @@ pub(crate) fn sighash(
...
                 script_code: &script,
                 script_pubkey: &script,
-                value: output
-                    .value
-                    .try_into()
-                    .expect("amount was previously validated"),
+                value: zp_tx::components::amount::NonNegativeAmount::from_nonnegative_i64_le_bytes(
+                    output.value.to_bytes(),
+                )
+                .expect("amount was previously validated"),
```
arya2: Nitpick: Could we update the `TryFrom` impl above (line ~209) to:

```rust
/// Convert a Zebra non-negative Amount into a librustzcash one.
impl TryFrom<Amount<NonNegative>> for zp_tx::components::amount::NonNegativeAmount {
    type Error = ();

    fn try_from(amount: Amount<NonNegative>) -> Result<Self, Self::Error> {
        zp_tx::components::amount::NonNegativeAmount::from_nonnegative_i64(amount.into())
    }
}
```

And use `try_into()` here and on line 36?

```suggestion
                value: output
                    .value
                    .try_into()
                    .expect("amount was previously validated"),
```

zebra-chain/src/transaction/serialize.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 };
 
 use super::*;
-use sapling::{Output, SharedAnchor, Spend};
+use crate::sapling::SharedAnchor;
```
arya2: Nitpick/Optional: Could we revert the other changes in this file and update this to:

```suggestion
use crate::sapling::{Output, SharedAnchor, Spend};
```

zebra-chain/src/primitives/zcash_primitives.rs
oxarbitrage: Thanks, i used this at https://github.com/ZcashFoundation/zebra/pull/8522/commits/617fc03b5780cd322de41bb273507d811b53afd0

zebra-chain/src/primitives/zcash_note_encryption.rs
oxarbitrage: If we change to `sapling::note_encryption::Zip212Enforcement::On` then the test will fail with:

```
Message:  coinbase outputs must be decryptable with an all-zero key: CoinbaseOutputsNotDecryptable
Location: zebra-consensus/src/transaction/tests.rs:2796
```

We might need to fix the test.

zebra-chain/src/transaction/serialize.rs
oxarbitrage: Thanks, fixed this at https://github.com/ZcashFoundation/zebra/pull/8522/commits/1da71d8888782d25f82118f6810421075f31c246

zebra-chain/src/primitives/zcash_note_encryption.rs
arya2: So the test will pass, and so we don't have to update this code if Zebra's mandatory checkpoint height is ever lowered below the last block before Canopy activation, we could derive `Ord`/`PartialOrd` impls on `NetworkUpgrade` and do:

```rust
    let zip_212_enforcement = if network_upgrade >= NetworkUpgrade::Canopy {
        sapling::note_encryption::Zip212Enforcement::On
    } else {
        sapling::note_encryption::Zip212Enforcement::Off
    };

    if let Some(bundle) = alt_tx.sapling_bundle() {
        for output in bundle.shielded_outputs().iter() {
            let recovery = sapling::note_encryption::try_sapling_output_recovery(
                &null_sapling_ovk,
                output,
                zip_212_enforcement,
            );
            if recovery.is_none() {
                return false;
            }
        }
    }

```

It could also just get the Canopy activation height and check that it's past it.

Zebra skips the check in production because of the mandatory checkpoint height but does check historical blocks before Canopy in the test, so the test actually covers more than strictly necessary here.

This is the relevant section of the ZIP: https://zips.z.cash/zip-0212#consensus-rule-change-for-coinbase-transactions

Note: This seems to be wrong in zcashd, see https://github.com/zcash/zcash/issues/6890. ~It may be a good idea to do a sync test from Canopy activation until the end of the grace period with checkpoint sync disabled to make sure that all of the blocks in the grace period follow the ZIP, though I think it was implemented correctly in zcashd prior to v5.9.0.~

Update: Actually, there's no point in a sync test, the mandatory checkpoint height is still after the grace period. We can do a sync test with checkpoint sync disabled when we're lowering the mandatory checkpoint height.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -29,11 +29,11 @@ impl zp_tx::components::transparent::Authorization for TransparentAuth<'_> {
...
 impl zp_tx::sighash::TransparentAuthorizingContext for TransparentAuth<'_> {
-    fn input_amounts(&self) -> Vec<zp_tx::components::amount::Amount> {
+    fn input_amounts(&self) -> Vec<zp_tx::components::amount::NonNegativeAmount> {
         self.all_prev_outputs
             .iter()
             .map(|prevout| {
-                zp_tx::components::amount::Amount::from_nonnegative_i64_le_bytes(
+                zp_tx::components::amount::NonNegativeAmount::from_nonnegative_i64_le_bytes(
                     prevout.value.to_bytes(),
                 )
```
arya2:
```suggestion
                prevout
                    .value
                    .try_into()
```

mpguerra:
@oxarbitrage should we close this one in favour of #8548 ?

oxarbitrage:
> @oxarbitrage should we close this one in favour of #8548 ?

No, that suggestion PR was already merged into [here](https://github.com/ZcashFoundation/zebra/pull/8527) and already in main. This one is different and needs to be merged with zebra-scan upgrades that are being worked on [here](https://github.com/ZcashFoundation/zebra/pull/8560) and in some code i have not pushed yet.

oxarbitrage:
Closing as the commits from here were included into https://github.com/ZcashFoundation/zebra/pull/8568

PR #8524, opened 2024-05-14
add(network): Adds a network magic field to `testnet::Parameters` and the config

### LINKED ISSUE #7845: Tracking: Add parameters to `Network::Testnet` variant

arya2:
## Motivation

We want to allow for configuring the network magic on custom testnets.

Part of #7845.
Depends-On: #8518.

## Solution

- Moves network `Magic` type from zebra-network to zebra-chain
- Adds a `network_magic` field on `testnet::Parameters` and uses it in the network codec
- Reads a network magic from the config

### Testing

- Adds a stored test config
- Updates a `Magic` debug impl test
- Adds a test to check that the codec returns an error when messages have unexpected network magics

We could also add a test for the constraints in `testnet::ParametersBuilder::with_network_magic()` (can't be the network magic of a default network) if we want.

## Review

Anyone can review.

### Reviewer Checklist

## Follow Up Work

Accept peer IPs in the config as initial peers.

### DIFF
--- a/zebra-chain/src/parameters/constants.rs
+++ b/zebra-chain/src/parameters/constants.rs
@@ -15,3 +15,13 @@ pub const SLOW_START_INTERVAL: Height = Height(20_000);
 ///
 /// This calculation is exact, because `SLOW_START_INTERVAL` is divisible by 2.
 pub const SLOW_START_SHIFT: Height = Height(SLOW_START_INTERVAL.0 / 2);
+
+/// Magic numbers used to identify different Zcash networks.
+pub mod magics {
+    use crate::parameters::network::magic::Magic;
+
+    /// The production mainnet.
+    pub const MAINNET: Magic = Magic([0x24, 0xe9, 0x27, 0x64]);
+    /// The testnet.
+    pub const TESTNET: Magic = Magic([0xfa, 0x1a, 0xf9, 0xbf]);
+}
--- a/zebra-chain/src/parameters/constants.rs
+++ b/zebra-chain/src/parameters/constants.rs
@@ -24,6 +24,6 @@ pub mod magics {
     pub const MAINNET: Magic = Magic([0x24, 0xe9, 0x27, 0x64]);
     /// The testnet.
     pub const TESTNET: Magic = Magic([0xfa, 0x1a, 0xf9, 0xbf]);
-    /// The regtest.
+    /// The regtest, see <https://github.com/zcash/zcash/blob/master/src/chainparams.cpp#L716-L719>
     pub const REGTEST: Magic = Magic([0xaa, 0xe8, 0x3f, 0x5f]);
 }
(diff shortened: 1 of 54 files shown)


zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -139,6 +144,20 @@ impl ParametersBuilder {
         self
     }
 
+    /// Sets the network name to be used in the [`Parameters`] being built.
+    pub fn with_network_magic(mut self, network_magic: Magic) -> Self {
+        assert!(
+            [magics::MAINNET, magics::TESTNET, magics::REGTEST]
```
arya2: The testnet magic should be allowed (it's the default value anyways if this method isn't called before `.finish()`), but maybe only on compatible testnets? (currently where the activation heights are the same as the default testnet, as we add more fields, the target difficulty, equihash parameters, and other parameters may also be relevant).

```suggestion
            [magics::MAINNET, magics::REGTEST]
```

zebra-chain/src/parameters/constants.rs
```diff
@@ -15,3 +15,15 @@ pub const SLOW_START_INTERVAL: Height = Height(20_000);
 ///
 /// This calculation is exact, because `SLOW_START_INTERVAL` is divisible by 2.
 pub const SLOW_START_SHIFT: Height = Height(SLOW_START_INTERVAL.0 / 2);
+
+/// Magic numbers used to identify different Zcash networks.
```
oxarbitrage, line 19: Should we add a link pointing to where did we got these ?

arya2: I'm not sure where these are specified (I didn't find it in the protocol specification), but I added a link to where I found it in the zcashd code in https://github.com/ZcashFoundation/zebra/pull/8524/commits/f2bef3a8bfbb8f15f97178eaad04ef66bf7e5469.

oxarbitrage (approved):
Looks good. Thanks.

PR #8527, opened 2024-05-14
deps(zcash_proofs): Update zcash_proofs from 0.13.0-rc.1 to 0.14.0

### LINKED ISSUE #8540: deps: Upgrade zcash_proofs and zcash_primitives in Zebra

oxarbitrage:
## Motivation

This PR updates `zcash_proofs` and introduces `sapling-crypto` to `zebra-consensus`.

## Solution

Copies some code from zcash_proofs into zebra, this is done in `parse_parameters.rs`. This will save us from making changes upstream.

### Testing

All tests are passing locally.

### Reviewer Checklist

### DIFF
--- a/zebra-consensus/Cargo.toml
+++ b/zebra-consensus/Cargo.toml
@@ -59,9 +59,9 @@ tracing = "0.1.39"
 tracing-futures = "0.2.5"
 
 orchard = "0.6.0"
-sapling = { package = "sapling-crypto", git = "https://github.com/oxarbitrage/sapling-crypto", branch = "add-verifying-key-methods" }
+sapling = { package = "sapling-crypto", version = "0.1" }
 
-zcash_proofs = { package = "zcash_proofs", git = "https://github.com/oxarbitrage/librustzcash", branch = "use-custom-sapling" }
+zcash_proofs = { version = "0.14.0", features = ["multicore" ] }
 wagyu-zcash-parameters = "0.2.0"
 
 tower-fallback = { path = "../tower-fallback/", version = "0.2.41-beta.13" }
--- a/zebra-consensus/Cargo.toml
+++ b/zebra-consensus/Cargo.toml
@@ -58,8 +58,8 @@ tower = { version = "0.4.13", features = ["timeout", "util", "buffer"] }
 tracing = "0.1.39"
 tracing-futures = "0.2.5"
 
-orchard = "0.6.0"
 sapling = { package = "sapling-crypto", version = "0.1" }
+orchard = "0.7.0"
 
 zcash_proofs = { version = "0.14.0", features = ["multicore" ] }
 wagyu-zcash-parameters = "0.2.0"
--- a/zebra-consensus/Cargo.toml
+++ b/zebra-consensus/Cargo.toml
@@ -59,8 +59,9 @@ tracing = "0.1.39"
 tracing-futures = "0.2.5"
 
 orchard = "0.6.0"
+sapling = { package = "sapling-crypto", git = "https://github.com/oxarbitrage/sapling-crypto", branch = "add-verifying-key-methods" }
 
-zcash_proofs = { version = "0.13.0-rc.1", features = ["multicore" ] }
+zcash_proofs = { package = "zcash_proofs", git = "https://github.com/oxarbitrage/librustzcash", branch = "use-custom-sapling" }
 wagyu-zcash-parameters = "0.2.0"
 
 tower-fallback = { path = "../tower-fallback/", version = "0.2.41-beta.13" }
--- a/zebra-consensus/src/primitives/groth16.rs
+++ b/zebra-consensus/src/primitives/groth16.rs
@@ -70,18 +70,11 @@ pub type ItemVerifyingKey = PreparedVerifyingKey<Bls12>;
 
 lazy_static::lazy_static! {
     /// The Sapling Groth16 verifying key for spends.
-    pub static ref SAPLING_SPEND_VERIFYING_KEY: std::sync::OnceLock<crate::groth16::VerifyingKey<Bls12>> = {
-        let cell = std::sync::OnceLock::new();
-        let _ = cell.set(GROTH16_PARAMETERS.sapling.spend.verifying_key().inner().clone());
-        cell
-    };
-
+    pub static ref SAPLING_SPEND_VERIFYING_KEY: crate::groth16::VerifyingKey<Bls12> =
+        GROTH16_PARAMETERS.sapling.spend.verifying_key().inner().clone();
      /// The Sapling Groth16 verifying key for outputs.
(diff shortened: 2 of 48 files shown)


zebra-consensus/src/primitives/groth16.rs
```diff
@@ -68,6 +68,22 @@ pub type BatchVerifyingKey = VerifyingKey<Bls12>;
...
+        cell
+    };
+
+     /// The Sapling Groth16 verifying key for outputs.
+    pub static ref SAPLING_OUTPUT_VERIFYING_KEY: std::sync::OnceLock<crate::groth16::VerifyingKey<Bls12>> = {
+        let cell = std::sync::OnceLock::new();
+        let _ = cell.set(GROTH16_PARAMETERS.sapling.output.verifying_key().inner().clone());
+        cell
+    };
+}
```
oxarbitrage: I had problems with the static lifetimes so i had to add this but there is maybe an easier way.

zebra-consensus/src/primitives/groth16.rs
```diff
@@ -102,7 +122,10 @@ pub static SPEND_VERIFIER: Lazy<
             (|item: Item| {
                 Verifier::verify_single_spawning(
                     item,
-                    &GROTH16_PARAMETERS.sapling.spend_prepared_verifying_key,
+                    &GROTH16_PARAMETERS
+                        .sapling
+                        .spend_prepared_verifying_key
+                        .inner(),
```
oxarbitrage: This is one of the functions added to `sapling-crypto`: https://github.com/oxarbitrage/sapling-crypto/blob/add-verifying-key-methods/src/circuit.rs#L593

zebra-consensus/src/primitives/groth16.rs
```diff
@@ -139,7 +166,10 @@ pub static OUTPUT_VERIFIER: Lazy<
             (|item: Item| {
                 Verifier::verify_single_spawning(
                     item,
-                    &GROTH16_PARAMETERS.sapling.output_prepared_verifying_key,
+                    &GROTH16_PARAMETERS
+                        .sapling
+                        .output_prepared_verifying_key
+                        .inner(),
```
oxarbitrage: https://github.com/oxarbitrage/sapling-crypto/blob/add-verifying-key-methods/src/circuit.rs#L604

zebra-consensus/Cargo.toml
```diff
@@ -59,8 +59,9 @@ tracing = "0.1.39"
 tracing-futures = "0.2.5"
 
 orchard = "0.6.0"
+sapling = { package = "sapling-crypto", git = "https://github.com/oxarbitrage/sapling-crypto", branch = "add-verifying-key-methods" }
```
oxarbitrage: This is the custom `sapling-crypto` version, specifically i am using custom inner functions to get inner types, instead we need to to something as described in https://github.com/zcash/sapling-crypto/issues/136
By doing it this way we are skipping type safety so i don't think upstream will allow the way i did it here but i think it helps to show the goal.

zebra-consensus/Cargo.toml
```diff
@@ -59,8 +59,9 @@ tracing = "0.1.39"
...
 
-zcash_proofs = { version = "0.13.0-rc.1", features = ["multicore" ] }
+zcash_proofs = { package = "zcash_proofs", git = "https://github.com/oxarbitrage/librustzcash", branch = "use-custom-sapling" }
```
oxarbitrage: This just points to a custom version of zcash_proofs 0.14.0 that use the sapling version we modified above, Zebra has troubles if there are 2 versions of sapling running.

oxarbitrage (commented):
Some comments.

zebra-consensus/src/primitives/groth16.rs
arya2: I think `lazy_static` is the way to go here, the `OnceLock` seems unnecessary (and may add a bit of synchronization overhead):

```suggestion
lazy_static::lazy_static! {
    /// The Sapling Groth16 verifying key for spends.
    pub static ref SAPLING_SPEND_VERIFYING_KEY: crate::groth16::VerifyingKey<Bls12> =
        GROTH16_PARAMETERS.sapling.spend.verifying_key().inner().clone();

     /// The Sapling Groth16 verifying key for outputs.
    pub static ref SAPLING_OUTPUT_VERIFYING_KEY: crate::groth16::VerifyingKey<Bls12> =
        GROTH16_PARAMETERS.sapling.output.verifying_key().inner().clone();
}
```

`Box::leak()` would work too (though it would have to be repeated elsewhere, so lazy_static seems better):

```rust
let vk = Box::new(
    GROTH16_PARAMETERS
        .sapling
        .spend
        .verifying_key()
        .inner()
        .clone(),
);

let vk: &VerifyingKey<_> = Box::leak(vk);
```

arya2: This looks like good use of `lazy_static`, ~I can't think of an easier way (or at least not one that requires fewer changes), but since these values are being cloned anyways and are only used once, using references doesn't seem to have any advantage over providing the batch verifier with an owned copy of the verifying key, so that's an option too.~

Update: Actually it's using the reference across threads, `lazy_static` is the best option.

zebra-consensus/src/primitives/groth16.rs
arya2: I think we should copy the `parse_parameters()` function from `zcash_proofs` and update the `SaplingParameters.spend_prepared_verifying_key` field type to `groth16::PreparedVerifyingKey<Bls12>` instead, I don't see an alternative to Zebra's `verify_proof()` function.

Same for the `output_prepared_verifying_key` field.

We'll need to copy `parse_parameters()`, `hashreader::HashReader` and `verify_hash()` too.

zebra-consensus/src/primitives/groth16.rs
oxarbitrage: Thanks, it seems the `OnceLock` was left there from a different attempt i did before. Fixed in https://github.com/ZcashFoundation/zebra/pull/8527/commits/d7ad18976facdeee9fb533fdda8ec4aec56ad0fb

zebra-consensus/src/primitives/groth16.rs
oxarbitrage: I will definitely try any alternative but I am not understanding this approach. Can you explain more or differently ?

arya2: I'll open a suggestion PR, note that @upbqdn would prefer we wait for the next version, and if we're okay waiting that could work too, but my concern is that we'll have to wait for the next zcashd release so there aren't dependency minor/patch version conflicts.

arya2: Opened https://github.com/ZcashFoundation/zebra/pull/8548 with the changes I had in mind.

upbqdn (approved):

PR #8548, opened 2024-05-17
Suggestion for "Update zcash_proofs from 0.13.0-rc.1 to 0.14.0"

### LINKED ISSUE #8540: deps: Upgrade zcash_proofs and zcash_primitives in Zebra

arya2:
## Motivation

This is a suggestion PR for https://github.com/ZcashFoundation/zebra/pull/8522 that copies the `parse_parameters()` function into Zebra to provide access to the inner values of the new types, `sapling::circuit::{SpendParameters, PreparedSpendVerifyingKey, OutputParameters, PreparedOutputVerifyingKey}`.

## Review

@oxarbitrage can review and merge.


oxarbitrage (approved):
That makes sense, thanks for the PR! 

This avoid us making changes in the upstream crate. I am not sure if we will do this forever but i think is good enough for now.

PR #8566, opened 2024-05-23
zebra_script: update to new zcash_script callback API

### LINKED ISSUE #3436: Create wrapper for `precomputed` in `CachedFfiTransaction`
## Motivation

`CachedFfiTransaction` has a `unsafe impl Send/Sync` since it has a FFI pointer (`precomputed`) which is neither, and we need both in order to be stored in a `Box` and sent across threads. This is fine as long as we take the proper precautions, as already extensively documented in its source code.

However, we could easily add another field which is not Send/Sync, which would be a problem since the whole struct is being forced into implementing Send/Sync and it would compile without errors while still being unsafe.

The solution is to create a wrapper for the `precomputed` and only make it `unsafe impl Send/Sync`. Since a struct is `Send/Sync` as long as all its fields are too, if we add a field that is not `Send/Sync` then the compiler would detected that.

### Specifications

[This is the approach suggested by the Rustonomicon](https://doc.rust-lang.org/nomicon/send-and-sync.html):

> Most uses of raw pointers should be encapsulated behind a sufficient abstraction that Send and Sync can be derived. For instance all of Rust's standard collections are Send and Sync (when they contain Send and Sync types) in spite of their pervasive use of raw pointers to manage allocations and complex ownership.

### Designs

- [ ] Create a wrapper for `precomputed` that has  `unsafe impl Send/Sync`
- [ ] Remove  `unsafe impl Send/Sync` from `CachedFfiTransaction`
- [ ] Update comments to reflect the changes

conradoplg:
## Motivation

See https://github.com/ZcashFoundation/zcash_script/pull/157 for context.

Currently no matching issue.

Closes https://github.com/ZcashFoundation/zebra/issues/3436 because makes it moot.

## Solution

This changes zebra_script to use the new zcash_script API.

This is a bit bigger than expected because I did a small refactoring that made the implementation easier: passing a BranchID instead of a NetworkUpgrade to the sighasher (which is simpler since all callers converted a BranchID to a NetworkUpgrade before calling it)

This also changes the SigHasher to allow precomputation, which is used with the new API. This way we keep the same performance since the old API also supported precomputation. But this will also allow us to share the precomputation with the shielded sighash implementation, but I'll do that in a follow up PR.

### Testing

I just updated the tests for the new SigHasher API where applicable.

## Review

This depends on #8568 since I wanted to make sure I was using up-to-date APIs.

### Reviewer Checklist

## Follow Up Work

We can now use the new SigHasher to share precomputation with the shielded sighash, but that's optional.


zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -242,25 +278,23 @@ impl From<Script> for zcash_primitives::legacy::Script {
...
-            let output = all_previous_outputs[input_index].clone();
-            script = output.lock_script.into();
+            let output = &precomputed_tx_data.all_previous_outputs[input_index];
+            lock_script = output.lock_script.clone().into();
+            unlock_script = zcash_primitives::legacy::Script(script_code.unwrap());
             zp_tx::sighash::SignableInput::Transparent {
                 hash_type: hash_type.bits() as _,
                 index: input_index,
-                script_code: &script,
-                script_pubkey: &script,
```
conradoplg, line 263: Amusingly this was wrong, but didn't cause problems because sighash for transparent txs was being computed by zcash_script and this path was never being taken.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -242,25 +279,23 @@ impl From<Script> for zcash_primitives::legacy::Script {
...
-    let script: zcash_primitives::legacy::Script;
+    let lock_script: zcash_primitives::legacy::Script;
+    let unlock_script: zcash_primitives::legacy::Script;
     let signable_input = match input_index {
         Some(input_index) => {
-            let output = all_previous_outputs[input_index].clone();
-            script = output.lock_script.into();
+            let output = &precomputed_tx_data.all_previous_outputs[input_index];
+            lock_script = output.lock_script.clone().into();
+            unlock_script = zcash_primitives::legacy::Script(script_code.unwrap());
```
upbqdn: Are we OK with the `unwrap`?

zebra-chain/src/transaction.rs
```diff
@@ -192,14 +194,21 @@ impl Transaction {
...
 
-    /// Calculate the sighash for the current transaction
+    /// Calculate the sighash for the current transaction.
+    ///
+    /// If you need to compute multiple sighashes for the same transactions,
+    /// it's more efficient to use [`Transaction::sighasher()`].
     ///
     /// # Details
     ///
     /// The `input` argument indicates the transparent Input for which we are
```
upbqdn:
```suggestion
    /// The `input` argument indicates the transparent input for which we are
```

zebra-chain/src/transaction.rs
```diff
@@ -192,14 +194,21 @@ impl Transaction {
...
-    /// producing a sighash. It is comprised of the index identifying the
-    /// transparent::Input within the transaction and the transparent::Output
-    /// representing the UTXO being spent by that input.
+    /// producing a sighash, or None if it's a shielded input. It is comprised
+    /// of the index identifying the transparent::Input within the transaction
+    /// and the transparent::Output representing the UTXO being spent by that
+    /// input.
```
upbqdn:
```suggestion
    /// producing a sighash, or `None` if it's a shielded input. The
    /// `input` represents the index of the [`transparent::Input`]
    /// within the transaction, and `all_previous_outputs` represents
    /// the UTXOs being spent by the input.
```

zebra-chain/src/transaction/sighash.rs
```diff
@@ -40,50 +39,55 @@ impl AsRef<[u8]> for SigHash {
...
+    /// and the transparent::Output representing the UTXO being spent by that
+    /// input.
+    ///
+    /// The `script_code` argument indicates the script code being validated
+    /// for transparent inputs, or None if it's a shielded input.
+    ///
+    /// # Panics
+    ///
+    /// - if passed in any NetworkUpgrade from before NetworkUpgrade::Overwinter
+    /// - if called on a v1 or v2 transaction
```
upbqdn: These seem outdated. Are they correct?

zebra-chain/src/transaction/sighash.rs
```diff
@@ -40,50 +39,55 @@ impl AsRef<[u8]> for SigHash {
...
-    fn hash_sighash_librustzcash(&self) -> SigHash {
+    /// Calculate the sighash for the current transaction.
+    ///
+    /// # Details
+    ///
+    /// The `input` argument indicates the transparent Input for which we are
+    /// producing a sighash, or None if it's a shielded input. It is comprised
+    /// of the index identifying the transparent::Input within the transaction
+    /// and the transparent::Output representing the UTXO being spent by that
+    /// input.
```
upbqdn: There's only `input_index`, and there are no outputs.

upbqdn (dismissed):

zebra-chain/src/transaction/sighash.rs
conradoplg: Thanks, fixed in e868a8184

zebra-chain/src/transaction/sighash.rs
conradoplg: Good point, fixed in e868a8184

zebra-chain/src/primitives/zcash_primitives.rs
conradoplg: I changed to a `expect()` in e868a8184 to document why it should be OK.

Alternatively, we could change the arguments to take an `Option<(usize, Vec<u8>)>` which would ensure that both are passed when needed. What do you think?

upbqdn: Yea, I'm in favor of the optional tuple.

conradoplg: Done in ef5a60280

upbqdn (approved):

PR #8568, opened 2024-05-26
build(ecc):Upgrade remaining ECC dependencies

oxarbitrage:
## Motivation

Put everything together. Replaces:

https://github.com/ZcashFoundation/zebra/pull/8522
https://github.com/ZcashFoundation/zebra/pull/8560
https://github.com/ZcashFoundation/zebra/pull/8567

Close https://github.com/ZcashFoundation/zebra/issues/8541
Close https://github.com/ZcashFoundation/zebra/issues/8540

The commit history is messy, should be ignored. 

#### Check before marking the PR as ready for review:
  - [ ] Will the PR name make sense to users?
  - [ ] Does the PR have a priority label?
  - [ ] Have you added or updated tests?
  - [ ] Is the documentation up to date?

##### For significant changes:
  - [ ] Is there a summary in the CHANGELOG?
  - [ ] Can these changes be split into multiple PRs?

_If a checkbox isn't relevant to the PR, mark it as done._

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

PR #8577, opened 2024-06-04
change(scan): Refactor scanning keys

upbqdn:
## Motivation

Address https://github.com/ZcashFoundation/zebra/pull/8568#discussion_r1617380954.

### PR Author Checklist

## Solution

- Don't use `UnifiedFullViewingKey::new` to construct ufvks so we don't need to use the `test-dependencies` feature with `zcash_client_backend` since that feature should not be used in production code.
- Refactor the construction of ufvks and their use for scanning. We were picking only the first key for scanning, which was a bug.
- Stop using `SaplingIvk` since they are covered by the new `ScanningKeys`.

### Reviewer Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6251,6 +6251,7 @@ dependencies = [
  "tokio",
  "tower",
  "tracing",
+ "zcash_address",
  "zcash_client_backend",
  "zcash_keys",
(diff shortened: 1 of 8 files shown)


oxarbitrage (approved):
That looks great, thanks!

PR #8624, opened 2024-06-17
fix(tests): Address port conflict issues and enable related Windows tests

### LINKED ISSUE #8553: Re enable port conflict related Windows tests
In https://github.com/ZcashFoundation/zebra/pull/8551 we disabled some Windows tests that were failing with port conflicts to often.

**Possible design:**

Use OS assigned ports whenever binding a `TcpListener` and get the socket address directly from the `TcpListener`:
- Return the listen address from `open_listener()` in the peer set `init()` function and use it instead of the random known port
- Use the `address()` method on `Server` in zebra-rpc (`server_instance.address()`)
- Bind a `TcpListener` in zebra-grpc, return the address (maybe via a oneshot channel? Or by spawning a tokio task and returning the value from the init function) and replace the call to `Router.serve()` with `serve_with_incoming()`
- Use port 0 for the RPC listen addr and read the RPC socket address from the Zebra's logs when spawning `TestChild`s instead of using the config value

arya2:
## Motivation

We want to re-enable these tests on Windows while avoiding spurious panics due to port conflicts.

Closes #8553.

## Solution

- Bind a `TcpListener` in zebra-grpc, return the address by spawning a tokio task and returning the value from the init function
- Replace the call to `Router.serve()` with `serve_with_incoming()`
- Use port 0 for the RPC `listen_addr` in RPC endpoint acceptance tests and read the RPC socket address from Zebra's logs when spawning `TestChild`s instead of using the configured listen address

Omitted:
- Returning the listen address from `open_listener()` in the peer set `init()` function and use it instead of the random known port
  - It's already possible to get the peer set's listen address from the address book returned by `open_listener()`
  - The affected tests are deliberately checking fixed ports, there are already versions of them using OS-assigned ports
- Returning the local listen address from the `Server` instance in zebra-rpc 
  - The affected tests don't need to know the listen address, it's fine to use an unknown OS-assigned port to check that the RPC server spawns successfully

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-grpc/src/server.rs
+++ b/zebra-grpc/src/server.rs
@@ -2,9 +2,14 @@
 
 use std::{collections::BTreeMap, net::SocketAddr, pin::Pin};
 
+use color_eyre::eyre::eyre;
 use futures_util::future::TryFutureExt;
+use tokio::task::JoinHandle;
 use tokio_stream::{wrappers::ReceiverStream, Stream};
-use tonic::{transport::Server, Request, Response, Status};
+use tonic::{
+    transport::{server::TcpIncoming, Server},
+    Request, Response, Status,
+};
 use tower::{timeout::error::Elapsed, ServiceExt};
 
 use zebra_chain::{block::Height, transaction};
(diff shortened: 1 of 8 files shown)


oxarbitrage (approved):
excellent, thank you!

PR #8629, opened 2024-06-20
change(consensus): Lower mandatory checkpoint height

### LINKED ISSUE #8430: Add semantic block validation during the ZIP-212 grace period after Canopy activation

arya2:
## Motivation

We want to support full validation of blocks from Canopy activation onwards, particularly for full validation of Canopy blocks on Regtest (though this use case also relies on supporting block construction for Canopy blocks).

Closes https://github.com/ZcashFoundation/zebra/issues/8430.

https://zips.z.cash/zip-0212

## Solution

Lowers mandatory checkpoint height to the block before Canopy activation.

### Tests

The mandatory checkpoint height test was updated to expect the last pre-Canopy block height.

Running a sync test locally with checkpoint sync disabled and an added checkpoint at height `1_046_399` until it gets past the grace period

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -198,7 +198,7 @@ impl Network {
     /// Mandatory checkpoints are a Zebra-specific feature.
     /// If a Zcash consensus rule only applies before the mandatory checkpoint,
     /// Zebra can skip validation of that rule.
-    ///
+    /// This is necessary because Zebra can't fully validate the blocks prior to Canopy.
     // TODO:
     // - Support constructing pre-Canopy coinbase tx and block templates and return `Height::MAX` instead of panicking
     //   when Canopy activation height is `None` (#8434)
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -5,7 +5,7 @@ use std::{fmt, str::FromStr, sync::Arc};
 use thiserror::Error;
 
 use crate::{
-    block::{self, Height, HeightDiff},
+    block::{self, Height},
     parameters::NetworkUpgrade,
 };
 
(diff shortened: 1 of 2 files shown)


arya2:
The sync test with checkpoint sync disabled passed the end of the grace period.

zebra-chain/src/parameters/network/tests/prop.rs
```diff
@@ -1,32 +1,27 @@
...
+    /// Check that the mandatory checkpoint is immediately before Canopy activation.
     ///
-    /// This is necessary because Zebra can't fully validate the blocks during the grace period due
-    /// to a limitation of `librustzcash`.
-    ///
-    /// See [`ZIP_212_GRACE_PERIOD_DURATION`] for more information.
+    /// This is necessary because Zebra can't fully validate the blocks prior to Canopy.
     #[test]
-    fn mandatory_checkpoint_is_after_zip212_grace_period(network in any::<Network>()) {
+    fn mandatory_checkpoint_is_after_canopy(network in any::<Network>()) {
```
upbqdn:
```suggestion
    fn mandatory_checkpoint_is_immediately_before_canopy(network in any::<Network>()) {
```

zebra-chain/src/parameters/network/tests/prop.rs
```diff
@@ -1,32 +1,27 @@
...
+use super::super::Network;
 use crate::{
     block::Height,
     parameters::{NetworkUpgrade, TESTNET_MAX_TIME_START_HEIGHT},
 };
 
 proptest! {
-    /// Check that the mandatory checkpoint is after the ZIP-212 grace period.
+    /// Check that the mandatory checkpoint is immediately before Canopy activation.
     ///
```
upbqdn:
```suggestion
```

zebra-chain/src/parameters/network/tests/prop.rs
```diff
@@ -1,32 +1,27 @@
...
-    /// This is necessary because Zebra can't fully validate the blocks during the grace period due
-    /// to a limitation of `librustzcash`.
-    ///
-    /// See [`ZIP_212_GRACE_PERIOD_DURATION`] for more information.
+    /// This is necessary because Zebra can't fully validate the blocks prior to Canopy.
```
upbqdn:
```suggestion
```

zebra-chain/src/parameters/network.rs
```diff
@@ -235,33 +199,16 @@ impl Network {
     /// If a Zcash consensus rule only applies before the mandatory checkpoint,
     /// Zebra can skip validation of that rule.
     ///
```
upbqdn:
```suggestion
    /// This is necessary because Zebra can't fully validate the blocks prior to Canopy.
```

upbqdn (dismissed):

upbqdn (approved):

PR #8594, opened 2024-06-06
change(scan): Remove support for starting the scanner at `zebrad` startup

upbqdn:
## Motivation

Close #8582.

#### Check before marking the PR as ready for review:
  - [x] Will the PR name make sense to users?
  - [x] Does the PR have a priority label?
  - [ ] Have you added or updated tests?
  - [x] Is the documentation up to date?
  - [x] Is there a summary in the CHANGELOG?
  - [x] Can these changes be split into multiple PRs?

## Solution

- Remove the code that starts the scanner from the start command.

### Testing

Let's move the the integration tests out of `zebrad` in #8593 once we have the new binary.

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

## Follow Up Work

- #8593
- #8586
- #8585

Issue #8585, opened 2024-06-04
Testing the new `zebra-scanner` binary


oxarbitrage:
We have some good test coverage for our blockchain scanner and most of them will keep working after we introduce a new binary (https://github.com/ZcashFoundation/zebra/issues/8581) however tests that relies on the scanner being started by zebrad will fail. We need to replace these by specific tests in the new binary scope. We may also want to add more tests and add the build of this binary into the CI.

More details to be added.

Note: https://github.com/ZcashFoundation/zebra/pull/8631 should be reverted if/when the `scan-start-where-left-test` test is restored.

mpguerra:
@oxarbitrage what else do we need to do in order to close off this issue?

oxarbitrage:
I think what we wanted to do here is actually done. 

What i will like to do at some point (and we can mark it as optional for now) is to add a test that checks the scanner behavior when we are synced up to the tip. This can be complicated to do so we will probably not do now. Do you want me to close this issue and open another one just for that ?

mpguerra:
yes please!

oxarbitrage:
Closing and opening https://github.com/ZcashFoundation/zebra/issues/8697 for the optional remaining test.

PR #8631, opened 2024-06-20
fix(ci): do not silently fail integration tests

gustavovalverde:
## Motivation

After #8594 some test might silently fail as a job dependency was removed for `failure-issue`, but it was still in the `needs` list.

## Solution

Comment the non existent dependency from `failure-issue` 

### Tests

- Integration tests should run here without issue

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/.github/workflows/sub-ci-integration-tests-gcp.yml
+++ b/.github/workflows/sub-ci-integration-tests-gcp.yml
@@ -589,7 +589,7 @@ jobs:
         lightwalletd-grpc-test,
         get-block-template-test,
         submit-block-test,
-        scan-start-where-left-test,
+        # scan-start-where-left-test,
         scan-task-commands-test,
       ]
     # Only open tickets for failed scheduled jobs, manual workflow runs, or `main` branch merges.


oxarbitrage (commented):
I don't understand the fix. It is enough to have the test commented out instead of just being removed ? Please also note this might be restored in the context of https://github.com/ZcashFoundation/zebra/issues/8585

arya2 (approved):
This looks good, I left a note in https://github.com/ZcashFoundation/zebra/issues/8585 about reverting this if/when the test is restored so we don't forget.

gustavovalverde:
> It is enough to have the test commented out instead of just being removed ?

@oxarbitrage I commented the test from the list, instead of removing it, as the test itself was commented out and not removed. It was, mainly, to keep it consistent with the previous change

oxarbitrage:
> > It is enough to have the test commented out instead of just being removed ?
> 
> @oxarbitrage I commented the test from the list, instead of removing it, as the test itself was commented out and not removed. It was, mainly, to keep it consistent with the previous change

Got it, thank you.

PR #8636, opened 2024-06-21
add(book): Describes custom Testnets in the Zebra book

### LINKED ISSUE #8588: add(docs): Describe custom testnets and their difference from Regtest
Zebra Book contains user info on Regtest: https://zebra.zfnd.org/user/regtest.html, but doesn't mention custom testnets, which Zebra supports.

We should:
- Add a description of custom testnets and their difference from Regtest.
- Describe how closely these testnets are to Mainnet as regards consensus validation.
- Describe how the custom testnets behave as regards connecting to peers.

We should also enhance the docs for Regtest with information mentioned in the list above.

arya2:
## Motivation

We want to document Zebra's custom Testnets in the Zebra book.

Closes #8588.

## Solution

- Adds a page to the Zebra book about custom Testnets and how they compare to Mainnet, the default Testnet, or Regtest.

Related changes:
- Adds the expected `genesis_hash` to Zebra's configuration.
- Uses the default Testnet if a `testnet_parameters` section is included in the config with default testnet values and no network name.

### Tests

Updates the latest stored config with a configured genesis hash.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+- The network name must:
+    - not be any of the reserved network names: `Testnet`, `Mainnet`, and `Regtest`,
+    - contain only alphanumeric characters and underscores, and
+    - be shorter than the `MAX_NETWORK_NAME_LENGTH` of `30`.
+- The network magic must not be any of the reserve network magics: `[36, 233, 39, 100]` and `[170, 232, 63, 95]`, these are the `Mainnet` and `Regtest` network magics respectively.
+- The network upgrade activation heights must be in order, such that the activation height for every network upgrade is at or above the activation height of every preceding network upgrade.
+
+## Comparison To Mainnet and Default Public Testnet Consensus Rules
+
+Aside from the configurable parameters, custom Testnets in Zebra validate all of same consensus rules as Testnet.
```
upbqdn:
```suggestion
Aside from the configurable parameters, custom Testnets in Zebra validate the same consensus rules as Testnet.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+network = "Testnet"
+
+[network.testnet_parameters.activation_heights]
+network_name = "ConfiguredTestnet_1"
+network_magic = [0, 1, 0, 255]
+slow_start_interval = 0
+target_difficulty_limit = "0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f"
+disable_pow = true
+
+# Configured activation heights must be greater than or equal to 1, and less than
```
upbqdn:
```suggestion
# Configured activation heights must be greater than 0, and less than
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+# or equal to Zebra's max block height of 2^31 - 1, block height 0 is reserved
+# for the Genesis network upgrade in Zebra.
```
upbqdn: Zebra's max block height in the state is actually 16777215: https://github.com/zcashfoundation/zebra/blob/d1a9004b3290add477f9a83ce959b7538725be1d/zebra-state/src/service/finalized_state/disk_format/block.rs#L35, but I don't think it applies here.

Maybe let's go with the following:

```suggestion
# 2^31. Block height 0 is reserved for the Genesis network upgrade in Zebra.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+# Heartwood, and Canopy at block height 1 as well.
+[network.testnet_parameters.activation_heights]
+NU5 = 1
+
+# This section may be omitted if it's not necessary to 
+# add transactions to Zebra's mempool
+[rpc]
+listen_addr = "0.0.0.0:18232"
+```
+
```
upbqdn, line 60:
```suggestion
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+miner_address = 't27eWDgjFYJGVXmzrXeVjnb5J3uXDM9xH9v'
+    
+[network]
+network = "Testnet"
+
+[network.testnet_parameters.activation_heights]
+# The Mainnet, Testnet, and Regtest network names are reserved.
+network_name = "ConfiguredTestnet_2"
+
+# The Mainnet and Regtest network magics are reserved.
```
upbqdn, line 76: Should we mention Testnet as well?

```suggestion
# The Mainnet, Testnet and Regtest network magics are reserved.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+network_magic = [0, 1, 0, 255]
```
upbqdn, line 77: We mention above that the parameters correspond to Mainnet, but the magic doesn't seem like any of the ones we have here https://github.com/zcashfoundation/zebra/blob/d1a9004b3290add477f9a83ce959b7538725be1d/zebra-chain/src/parameters/constants.rs#L24.

Shouldn't the magic be like this for Testnet instead?

```suggestion
network_magic = [250, 26, 249, 191]
```

Or we can't use the one for Testnet because we're using the default initial peers?

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+NU5 = 1
+
+# This section may be omitted if it's not necessary to 
+# add transactions to Zebra's mempool
+[rpc]
+listen_addr = "0.0.0.0:18232"
+```
+
+
+Relevant parts of the configuration file with Mainnet parameters:
```
upbqdn: We mention here that the parameters correspond to Mainnet, but the config seems to be for Testnet.

```suggestion
Relevant parts of the configuration file with Testnet parameters:
```

I think it's better to go with Testnet because the mining address below is for Testnet.

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+disable_pow = false
+
+[network.testnet_parameters.activation_heights]
+BeforeOverwinter = 1
+Overwinter = 347_500
+Sapling = 419_200
+Blossom = 653_600
+Heartwood = 903_000
+Canopy = 1_046_400
+NU5 = 1_687_104
```
upbqdn, line 100: These are for Mainnet, so if we apply the other suggestions, we should change these so they match Testnet.

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+# add transactions to Zebra's mempool
+[rpc]
+listen_addr = "0.0.0.0:18232"
+```
+
+
+## Caveats and Restrictions
+
+There are a few caveats:
+- Configured network upgrade activation heights must be above the genesis block height, which is reserved for Zebra's `Genesis` network upgrade, and must not be above Zebra's max block height of `2^31 -1`.
```
upbqdn:
```suggestion
- Configured network upgrade activation heights must be above the genesis block height, which is reserved for Zebra's `Genesis` network upgrade, and must not be above Zebra's max block height of `2^31 - 1`.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+- While it's possible to activate Canopy and later network upgrades after height 1, Zebra cannot currently produce pre-Canopy block templates, so the `getblocktemplate` RPC method and Zebra's internal miner which depends on the `getblocktemplate` method won't work until Canopy is activated. An alternative block source will be required to mine pre-Canopy blocks onto Zebra's chain.
+- While it's possible to use the default Testnet network magic with a configured Testnet, Zebra will panic when configured to use the default initial Testnet peers and Testnet parameters that are incompatible with the default public Testnet.
```
upbqdn: > ... Zebra will panic when configured to use the [...] Testnet parameters that are incompatible with the default public Testnet.

Should we document what those are?

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+- If the genesis hash is configured, a genesis block will need to be copied into the custom Testnet state or submitted via the `submitblock` RPC method, Zebra cannot currently generate genesis blocks. See the `CreateGenesisBlock()` function in `zcashd/src/chainparams.cpp` for use cases that require a new genesis block.
+
+There are also a few other restrictions on these parameters:
+- The network name must:
+    - not be any of the reserved network names: `Testnet`, `Mainnet`, and `Regtest`,
+    - contain only alphanumeric characters and underscores, and
+    - be shorter than the `MAX_NETWORK_NAME_LENGTH` of `30`.
+- The network magic must not be any of the reserve network magics: `[36, 233, 39, 100]` and `[170, 232, 63, 95]`, these are the `Mainnet` and `Regtest` network magics respectively.
```
upbqdn:
```suggestion
- The network magic must not be any of the reserved network magics: `[36, 233, 39, 100]` and `[170, 232, 63, 95]`, these are the `Mainnet` and `Regtest` network magics respectively.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,166 @@
...
+
+Zebra's Regtest network is a special case of a custom Testnet that:
+- Won't make peer connections[^fn1],
+- Skips Proof-of-Work validation,
+- Uses a reserved network magic and network name,
+- Activates network upgrades up to and including Canopy at block height 1,
+- Tries to closely match the `zcashd` Regtest parameters, and
+- Expects the Regtest genesis hash.
+
+Once Zebra's internal miner can mine Equihash solutions configurable parameters, Regtest should validate Proof-of-Work with the zcashd Equihash parameters unless it's disabled in its configuration, and custom Testnets should allow for configuring their Equihash parameters as well.
```
upbqdn:
```suggestion
Once Zebra's internal miner can mine Equihash solutions with configurable parameters, Zebra's Regtest should validate Proof-of-Work with the zcashd Regtest Equihash parameters unless it's disabled in its configuration, and custom Testnets should allow for configuring their Equihash parameters as well.
```

upbqdn (dismissed):

book/src/user/custom-testnets.md
arya2: > Zebra's max block height in the state is actually 16777215

That's a good point, I'll add a footnote about the max on-disk serialized block height currently being `2^24 - 1`.

Update: Added in https://github.com/ZcashFoundation/zebra/pull/8636/commits/14147916707645cfee7c653abb569a1dbb5f871c.

book/src/user/custom-testnets.md
arya2: The Testnet network magic actually isn't reserved, though it may be a good idea to only allow it for compatible custom Testnets (currently configuring anything except the network name will result in an incompatible Testnet).

Update: Added a note about not recommending usage of the default Testnet network magic with incompatible parameters in https://github.com/ZcashFoundation/zebra/pull/8636/commits/14147916707645cfee7c653abb569a1dbb5f871c.

book/src/user/custom-testnets.md
arya2: It is a Testnet, only the activation heights, slow start interval, genesis hash, and target difficulty limit match the Mainnet parameters, because the other ones are reserved. It's mentioned in the "Configuring More Testnet Parameters To Match Mainnet Consensus Rules" section that the Mainnet address prefixes and coin type should be reserved for only Mainnet, and it's mentioned in "Caveats and Restrictions" that the Mainnet network name and magic are reserved as well.

I'll list those parameters explicitly here so it's clear that the network name, network magic, address prefixes, and coin type still do not match Mainnet.

> I think it's better to go with Testnet because the mining address below is for Testnet

The first config has more practical parameters for a custom Testnet, this one is meant to be an example of how to configure a custom Testnet to more closely match Mainnet parameters. I'll note above that the first config is more practical.

Added "with practical parameters" to the heading before the first config and a footnote about what parameters in the second config actually match Mainnet in https://github.com/ZcashFoundation/zebra/pull/8636/commits/14147916707645cfee7c653abb569a1dbb5f871c.

book/src/user/custom-testnets.md
arya2: > Or we can't use the one for Testnet because we're using the default initial peers?

We could use the default Testnet magic here, but can't use the default initial peers because the parameters in both of these configs are incompatible with the default Testnet. I don't think we should use the default Testnet magic because the other parameters are incompatible, but it shouldn't make much difference in practice.

book/src/user/custom-testnets.md
arya2: I'd like to keep these so that it's possible to copy the Mainnet best chain and checkpoints over to a custom Testnet with this config and disable the Proof-of-Work check to add blocks to it.

Added a comment above this config field about when to use these activation heights in https://github.com/ZcashFoundation/zebra/pull/8636/commits/14147916707645cfee7c653abb569a1dbb5f871c.

book/src/user/custom-testnets.md
arya2: > Should we document what those are?

Yes! Added a footnote and a line in the introduction about which fields will result in an incompatible Testnet in https://github.com/ZcashFoundation/zebra/pull/8636/commits/14147916707645cfee7c653abb569a1dbb5f871c.

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,180 @@
...
+
+##### Footnotes
+
+[^fn1]: Only the network upgrade activation heights, target difficulty limit, slow start interval, and genesis hash of a custom Testnet may currently be configured in Zebra to match Mainnet, other differences between Mainnet and Testnet consensus rules still remain with any custom Testnet configuration. More parameters may become configurable in the future, but the Mainnet Zcash address prefixes, coin type, network name, and network magic should remain reserved only for Mainnet.
+
+[^fn2]: Zebra's max on-disk serialized block height is currently `2^24 - 1`, the max block height of `2^31 - 1` can only be represented in-memory, so while an activation height of `2^31 - 1` is valid, Zebra's best chain would not currently be able to reach that activation height.
+
+[^fn3]: Configuring any of the Testnet parameters that are currently configurable except the network name will result in an incompatible custom Testnet, these are: the network magic, network upgrade activation heights, slow start interval, genesis hash, and target difficulty limit.
```
arya2:
```suggestion
[^fn3]: Configuring any of the Testnet parameters that are currently configurable except the network name will result in an incompatible custom Testnet, these are: the network magic, network upgrade activation heights, slow start interval, genesis hash, disabled Proof-of-Work and target difficulty limit.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,180 @@
...
+
+Custom Testnets in Zebra enable testing consensus rule changes on a public, configured Testnet independent of the default public Zcash Testnet.
+
+Zebra's Testnet can be configured with custom:
+- Network upgrade activation heights, 
+- Network names, 
+- Network magics, 
+- Slow start intervals, 
+- Genesis hashes, and
+- Target difficulty limits.
```
arya2, line 11:
```suggestion
- Target difficulty limits.

It can also be configured to disable Proof-of-Work validation.
```

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,180 @@
...
+
+Configuring any of those Testnet parameters except the network name with non-default values will result in an incompatible custom Testnet.
+
+It's also possible to disable Proof-of-Work validation by setting `disable_pow` to `true` so that blocks can be mined onto the chain without valid Equihash solutions, nor block hashes below their target difficulties.
```
arya2:
```suggestion
It's also possible to disable Proof-of-Work validation by setting `disable_pow` to `true` so that blocks can be mined onto the chain without valid Equihash solutions, nor block hashes below their target difficulties.

Configuring any of those Testnet parameters except the network name with non-default values will result in an incompatible custom Testnet.
```

book/src/user/custom-testnets.md
upbqdn: I see. Thanks for the clarification. Let's not use the Testnet magic.

book/src/user/custom-testnets.md
```diff
@@ -0,0 +1,180 @@
...
+- Network upgrade activation heights, 
+- Network names, 
+- Network magics, 
+- Slow start intervals, 
+- Genesis hashes, and
+- Target difficulty limits.
+
+It's also possible to disable Proof-of-Work validation by setting `disable_pow` to `true` so that blocks can be mined onto the chain without valid Equihash solutions, nor block hashes below their target difficulties.
+
+Configuring any of those Testnet parameters except the network name with non-default values will result in an incompatible custom Testnet.
```
upbqdn: Can we also clarify what an "incompatible Testnet" means?

upbqdn (dismissed):

book/src/user/custom-testnets.md
upbqdn: We mention it a few times later on, so I think it would be useful to have explained what we mean by the incompatibility.

arya2:
```suggestion
Configuring any of those Testnet parameters except the network name with non-default values will result in an incompatible custom Testnet. Incompatible Testnets will fail to successfully complete peer handshakes with one another, or could provide one another with invalid blocks or invalid mempool transactions. Peer node connections that consistently provide invalid blocks or mempool transactions should be considered misbehaving peer connections and dropped.
```

arya2: Described what's meant by "incompatible Testnet" in https://github.com/ZcashFoundation/zebra/pull/8636/commits/0842950c785dca8826127029a9cf26470214dfa5.

upbqdn (approved):

PR #8656, opened 2024-06-28
fix(test): Increase expected height in sync-past-checkpoint test

arya2:
## Motivation

This may help the `sync-past-checkpoint` test pass in CI. This test stopped working when we lowered the mandatory checkpoint height by the ZIP-212 grace period because the tip height in our cached state is now above the new mandatory checkpoint height.

## Solution

Increases the expected height in the affected checkpoint sync test back to where it was before the mandatory checkpoint height was lowered.

### Follow-up Work

Revert this change after updating the cached state to lower its tip height in CI?

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -1267,7 +1267,7 @@ fn create_cached_database(network: Network) -> Result<()> {
 
 #[tracing::instrument]
 fn sync_past_mandatory_checkpoint(network: Network) -> Result<()> {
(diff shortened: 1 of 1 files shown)


arya2:
Seems to work: https://github.com/ZcashFoundation/zebra/actions/runs/9715002966/job/26815976764?pr=8656

upbqdn (approved):

PR #8674, opened 2024-07-12
add(rpc): Add a tonic server in zebra-rpc

### LINKED ISSUE #8618: Add a tonic RPC server to `zebra-rpc` with a read state service
## Motivation

`zebra-rpc` needs to notify clients when there have been changes to Zebra's best chain tip or mempool with server-sent events or a socket. 

This would also be a step towards supporting multiple formats in `zebra-rpc`'s current JSON RPC methods (this would also require adding a JSON codec to the tonic server, see https://github.com/ZcashFoundation/zebra/issues/8245)

## Design

Add a tonic server in `zebra-rpc` behind a feature flag for implementing methods with streaming responses and spawn it when Zebra starts. This should look a lot like the scanner gRPC server in `zebra-grpc`.

arya2:
## Motivation

We want to implement the indexer RPC methods in a tonic server to stream responses to its clients.

Closes https://github.com/ZcashFoundation/zebra/issues/8618.

## Solution

- Adds a tonic server similar to the one in `zebra-grpc`
- Implements a simple `chain_tip_change` method which notifies clients when the best chain tip in Zebra changes

### Tests

Tests that the server starts and the `chain_tip_change` method notifies the client when the best chain tip changes.

### Follow-up Work

Implement the rest of the `chain_tip_change` method

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/config.rs
```diff
@@ -30,6 +30,22 @@ pub struct Config {
...
     pub listen_addr: Option<SocketAddr>,
 
+    /// IP address and port for the indexer RPC server.
+    ///
+    /// Note: The indexer RPC server is disabled by default.
+    /// To enable the indexer RPC server, compile `zebrad` with the
+    /// `indexer` feature flag and set a listen address in the config:
+    /// ```toml
+    /// [rpc]
+    /// indexer_listen_addr = '127.0.0.1:8232'
```
oxarbitrage: We use port 8232 for the normal RPC server. Should we suggest a different port for the indexer?

oxarbitrage (dismissed):
Looks good to me, just some comment in the default port chosen.

zebra-rpc/src/config.rs
arya2: > Should we suggest a different port for the indexer?

Yeah, I think we should, how about `8231`?

```suggestion
    /// indexer_listen_addr = '127.0.0.1:8231'
```

arya2: Yeah, I think we should suggest a different port there, fixed in https://github.com/ZcashFoundation/zebra/pull/8674/commits/63d1e438e6e3b9dd55f1f01e425beee1121347b2.

oxarbitrage: We use 8231 as the grpc server port of the scanner, which is in a different binary so is maybe ok.

oxarbitrage (approved):

PR #8714, opened 2024-07-23
Update lightwalletd.md

therealyingtong:
Change the testnet default RPC port from `18323` to `18232`


oxarbitrage (approved):
Hey, you are right here. Thanks for sending us the change.

PR #8694, opened 2024-07-17
change(consensus): Add lockbox funding stream

arya2:
## Motivation

We likely want to add a lockbox funding stream.

Will close #8686, and close #8687 once parameters/addresses are updated.

Part of #8695.

This PR will need to be updated once there's more information about which dev fund proposal we should implement, and, if it's alternative 2 in [this draft ZIP](https://zips.z.cash/draft-nuttycom-funding-allocation), also what funding stream addresses to use for ZCG.

### Specifications & References

Alternatives 2 and 3 in https://zips.z.cash/draft-nuttycom-funding-allocation
https://zips.z.cash/draft-nuttycom-lockbox-streams

## Solution

- Adds a new funding stream receiver for the "deferred pool"
- Checks funding streams after the next halving
- Generates coinbase transactions with new funding streams after NU6 is active 
- Checks that block subsidies are fully claimed after NU6 is active
- Checks that coinbase transactions spend less than:
   `block_subsidy + miner_fees - expected_lockbox_funding_stream_amount`
- Adds a `lockbox_input_value()` function 

Related changes:
- Factors `Halving(height)` calculation into its own `num_halvings()` function
- In `block::check::subsidy_is_valid()`:
  - Removes unnecessary check that `halving_div.count_ones() == 1`
  - Removes redundant check for `height > slow_start_shift`
  - Checks `num_halvings()` instead of `halving_divisor()`
  - Minor simplifications
 
### Tests

- Adds an acceptance test that relies on configurable funding streams on Testnet
- Updates test to check NU6 funding stream values
- Adds a `check_lockbox_input_value()` test for the `lockbox_input_value()` fn

### Follow Up Work

- https://github.com/ZcashFoundation/zebra/pull/8729
- https://github.com/ZcashFoundation/zebra/pull/8727

Blocked on ZIP-253 status being changed to 'Proposed':
- Set an NU6 activation height for Testnet 
- [Update minimum network protocol versions for NU6 on Testnet/Mainnet
](https://github.com/ZcashFoundation/zebra/pull/8733)
- Update initial current minimum network protocol version
- Update post-NU6 funding stream height ranges
- Update the NU6 consensus branch id
- [Update the end of support halt time](https://github.com/ZcashFoundation/zebra/pull/8734)

This PR also adds some TODOs for documentation updates that should be addressed promptly.

### PR Author's Checklist

- [x] The PR name will make sense to users.
- [ ] The PR provides a CHANGELOG summary.
- [x] The solution is tested.
[...]

PR #8718, opened 2024-07-23
change(network-params): Configurable Testnet funding streams

### LINKED ISSUE #8367: Add protocol version information to fields on `NetworkParameters`
## Motivation

We want to make these values configurable on Testnets:
  - [`INITIAL_MIN_NETWORK_PROTOCOL_VERSION`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/constants.rs#L395)
  - [`CURRENT_NETWORK_PROTOCOL_VERSION`](https://github.com/ZcashFoundation/zebra/blob/restore-internal-miner/zebra-network/src/constants.rs#L340)
  - [`min_specified_for_upgrade`](https://github.com/ZcashFoundation/zebra/blob/restore-internal-miner/zebra-network/src/protocol/external/types.rs#L75)

See https://github.com/ZcashFoundation/zebra/pull/7924#discussion_r1385877140.

<details>

<summary> Outdated:
</summary>
These have already been added.

  - [FUNDING_STREAM_HEIGHT_RANGES](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/parameters/subsidy.rs#L107)
  - [FUNDING_STREAM_ADDRESSES](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/parameters/subsidy.rs#L116)
</details>

### Possible Design

Move these values to fields/methods on `NetworkParameters`.

arya2:
## Motivation

We want configurable funding streams to test consensus rule changes in NU6.

Closes #8367.

Depends-On: #8723.

## Solution

- Moves `subsidy.rs` from zebra-consensus to zebra-chain
- Moves `funding_stream_address_period()` fn to zebra-chain
- Refactors funding stream information into structs
- Splits funding streams into pre/post NU6 funding streams
  - Adds expected height ranges for post-NU6 funding streams with empty recipients list
- Adds a `funding_streams` field on `testnet::Parameters`

Related changes:
- Returns early from `is_regtest()` after checking the network magic
- Moves funding stream receiver names to a method
- Adds missing `#[serde(deny_unknown_fields)]` on Testnet parameters struct in zebra-network config deserialize impl

### Tests

This PR adds a vector test for the constraints on `ParametersBuilder::with_funding_streams()` setter methods and updates the v1.9.0 stored config.

### PR Author's Checklist

- [x] The solution is tested.

## Follow Up Work

Move the `zebra_consensus::block::subsidy` module to `zebra-chain`.

### PR Reviewer's Checklist


zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -60,76 +88,144 @@ pub const FUNDING_STREAM_RECEIVER_DENOMINATOR: u64 = 100;
...
-    // TODO: Move the value here to a field on `testnet::Parameters` (#8367)
-    pub static ref FUNDING_STREAM_HEIGHT_RANGES: HashMap<NetworkKind, std::ops::Range<Height>> = {
-        let mut hash_map = HashMap::new();
-        hash_map.insert(NetworkKind::Mainnet, Height(1_046_400)..Height(2_726_400));
-        hash_map.insert(NetworkKind::Testnet, Height(1_028_500)..Height(2_796_000));
-        hash_map.insert(NetworkKind::Regtest, Height(1_028_500)..Height(2_796_000));
-        hash_map
+    /// The post-NU6 funding streams for Mainnet
+    pub static ref POST_NU6_FUNDING_STREAMS_MAINNET: FundingStreams = FundingStreams {
+        height_range: Height(2_726_400)..Height(3_146_400),
```
oxarbitrage, line 202: The start range is currently defined in the ZIP (https://github.com/zcash/zips/pull/878/files) but the end range is not. Where is this number coming from ? Should we clarify ?

oxarbitrage: Another thing that had my attention is that there might be no funding stream at all after NU6. Should we still want to have a `POST_NU6_FUNDING_STREAMS_MAINNET` around if that is the case?

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -60,76 +88,144 @@ pub const FUNDING_STREAM_RECEIVER_DENOMINATOR: u64 = 100;
...
-        let mut regtest_addresses = HashMap::with_capacity(3);
-        regtest_addresses.insert(FundingStreamReceiver::Ecc, FUNDING_STREAM_ECC_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
-        regtest_addresses.insert(FundingStreamReceiver::ZcashFoundation, FUNDING_STREAM_ZF_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
-        regtest_addresses.insert(FundingStreamReceiver::MajorGrants, FUNDING_STREAM_MG_ADDRESSES_TESTNET.iter().map(|a| a.to_string()).collect());
-        addresses_by_network.insert(NetworkKind::Testnet, regtest_addresses);
-
-        addresses_by_network
+    /// The post-NU6 funding streams for Testnet
+    pub static ref POST_NU6_FUNDING_STREAMS_TESTNET: FundingStreams = FundingStreams {
+        height_range: Height(2_942_000)..Height(3_362_000),
```
oxarbitrage, line 232: Is `3_362_000` this the next halving ? Current proposals are for 1 or 2 years.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -233,6 +318,13 @@ impl ParameterSubsidy for Network {
         }
     }
 }
+
```
oxarbitrage, line 326:
```suggestion

// TODO: Try to unify in one implementation with the above `impl ParameterSubsidy for &Network`
```

oxarbitrage (dismissed):
Looks good, i think we can clarify some of the block numbers maybe adding them as constants so we can change them more easy.

zebra-chain/src/parameters/network/subsidy.rs
arya2: > The start range is currently defined in the ZIP (https://github.com/zcash/zips/pull/878/files) but the end range is not. Where is this number coming from ? Should we clarify ?

These heights and the recipients in the next PR are based on alternative 2 in the lockbox funding allocation draft ZIP, updated the docs to explain where it's from in https://github.com/ZcashFoundation/zebra/pull/8718/commits/852d63438288cd433e6c2cbd88c4326ba4765e56.

> Another thing that had my attention is that there might be no funding stream at all after NU6.

An array would probably work better than pre/post-NU6 funding stream constants/fields if there's another funding stream after this one, this seems fine for now.

This should continue to apply the funding stream until the next halving or the end height, whichever comes first (for alternative 2, that's the end height), it just won't start applying it until after NU6.

zebra-chain/src/parameters/network/subsidy.rs
arya2: These heights and the recipients in the next PR are based on alternative 2 in the lockbox funding allocation draft ZIP (for the number of blocks, assuming the spacing is about the same, I wasn't sure how to translate 2 years into a number of blocks, I don't think it's very consistent on Testnet) and draft ZIP 253's current Testnet activation height (though that will likely change and require an update here).

Updated the doc comment and added a TODO for leaving a link to the lockbox funding allocation ZIP in https://github.com/ZcashFoundation/zebra/pull/8718/commits/852d63438288cd433e6c2cbd88c4326ba4765e56.

zebra-chain/src/parameters/network/subsidy.rs
arya2: I figured out how to get rid of it, fixed in https://github.com/ZcashFoundation/zebra/pull/8718/commits/852d63438288cd433e6c2cbd88c4326ba4765e56.

oxarbitrage (dismissed):
Thanks for the updates.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -60,76 +88,149 @@ pub const FUNDING_STREAM_RECEIVER_DENOMINATOR: u64 = 100;
...
-        let mut hash_map = HashMap::new();
-        hash_map.insert(FundingStreamReceiver::Ecc, "Electric Coin Company");
-        hash_map.insert(FundingStreamReceiver::ZcashFoundation, "Zcash Foundation");
-        hash_map.insert(FundingStreamReceiver::MajorGrants, "Major Grants");
-        hash_map
-    };
+    /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
+    height_range: std::ops::Range<Height>,
+    /// Funding stream recipients by [`FundingStreamReceiver`].
+    recipients: HashMap<FundingStreamReceiver, FundingStreamRecipient>,
```
upbqdn, line 100: I find the mapping from `Receivers` to `Recipients` rather confusing. The protocol spec also doesn't contain the term "receiver". I'll try to think of some alternative.

conradoplg: My suggestion: FundingStreamRecipient -> FundingStreamParams

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -48,49 +48,34 @@ pub fn funding_stream_values(
...
+    // the number of addresses for each recipient should be validated for a configured height range
+    // when configured Testnet parameters are built.
+    let num_addresses = funding_streams
+        .recipients()
+        .values()
+        .next()
+        // TODO: Return an Option from this function and replace `.unwrap()` with `?`
+        .unwrap()
+        .addresses()
+        .len();
```
conradoplg, line 77: I'm a bit confused by this code. What is the purpose of the `next()` call?

zebra-chain/src/parameters/network/subsidy.rs
arya2: I find it confusing too, ~but I'd like to do any cleanups in https://github.com/ZcashFoundation/zebra/pull/8719.~

Update: FundingStreamRecipient -> FundingStreamParams works, I'll do that here.

Update: Updating it here causes a lot of conflicts in the next PR, I'd still prefer to do any cleanups in a follow-up PR.

conradoplg (commented):
Left some comments, also, are the TODOs added supposed to be addressed in a later PR or in this one still?

zebra-consensus/src/block/subsidy/funding_streams.rs
arya2: It's to get the first item in the recipients `HashMap`, any of them would be fine since they should all be the same length on Mainnet and Testnet, or they should all be checked when testnet::Parameters is being built on configured Testnets.

I don't really like this if you have a suggestion for how to do it differently.

arya2:
> Left some comments, also, are the TODOs added supposed to be addressed in a later PR or in this one still?

They're all meant to be addressed in a later PR, the ones related to referencing the final versions of draft ZIPs are meant to be addressed when those draft ZIPs are finalized, the rest are addressed in #8694.

zebra-consensus/src/block/subsidy/funding_streams.rs
conradoplg: Got it. I think this might return the wrong thing since the first element could be the Deferred recipient, whose address length is 0, which will lead to a panic in the assert below

arya2: Thank you! Fixed in https://github.com/ZcashFoundation/zebra/pull/8694/commits/7df2e2863f75c2b85323e8906f9f30ee0bacc73f.

oxarbitrage (approved):

PR #8719, opened 2024-07-24
change(consensus): Refactor `block::check::subsidy_is_valid()`

arya2:
## Motivation

This is an optional refactor of the `subsidy_is_valid()` function and could include other cleanups following the NU6 consensus rule changes.

Depends-On: #8694.

## Solution

- Adds a `num_halvings()` fn
- Calls `num_halvings()` instead of `halving_divisor()` in `subsidy_is_valid()`

### Tests

Zebra has several vectors tests for this function.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-consensus/src/block/check.rs
```diff
@@ -148,77 +148,58 @@ pub fn subsidy_is_valid(block: &Block, network: &Network) -> Result<(), BlockErr
...
-    } else {
-        network.slow_start_interval()
-    };
-
-    if height < slow_start_interval {
-        unreachable!(
-            "unsupported block height: callers should handle blocks below {:?}",
-            slow_start_interval
-        )
-    } else if halving_div.count_ones() != 1 {
```
arya2, line 173: `height < slow_start_interval` is checked in `num_halvings()` (and was checked in `halving_divisor()` before this change)

`halving_div.count_ones() != 1` where `halving_div` is `1 << num_halvings` should always be false if `num_halvings <= 2` (and should have always been false before this change since it would return early `if 1u64.checked_shl(num_halvings) == None`).

arya2:
Closing this for now, it's easier to recreate than to maintain

PR #8729, opened 2024-07-29
change: Track the balance of the deferred chain value pool

### LINKED ISSUE #8688: Add lockbox pool in zebra-state
## Motivation

All of the current dev fund proposals include a lockbox pool for a reserve, which will require some changes in zebra-state to track the lockbox pool value balance.

## Possible Design

- Add a `deferred` field on `ValueBalance` in zebra-chain
- Add the expected lockbox funding stream value balance to the deferred value balance in the `chain_value_pools` field on non-finalized chains when committing blocks (`block_subsidy(height) * FS_DEFERRED_NUMERATOR / 100`)
- Update the `ValueBalance::to_bytes()` and `from_bytes()` methods to encode/decode the deferred field (this may involve a minor db version bump)

upbqdn:
## Motivation

Close #8688.

Depends-On: #8694.

## Solution

There are two main changes in this PR:

- Calculate and store deferred value balances for each block in the non-finalized state, and
- track the total amount of the deferred chain value pool in the finalized state.

This PR also bumps the major database format version.

#### High-level Description

- The amount of the deferred balance per block is calculated in `zebra-consensus` during the semantic validation of both checkpointed and non-checkpointed blocks.
- The deferred balances are passed to `zebra-state` in the validated blocks.
- The balances are added to the chain value pool of each chain in the non-finalized state upon committing the block, and subtracted upon removing the block.
- Zebra stores the balances of individual blocks in the non-finalized state but doesn't do so in the finalized one. It only tracks the total amount in the deferred chain value pool of the tip of the finalized state by adding the deferred balances to the pool upon committing blocks coming from the non-finalized state. Therefore, retrieving the deferred values for individual blocks from the finalized state doesn't currently work. If we ever need this functionality, I think it will be better to calculate it dynamically based on height and network.
- The `ValueBalance` struct has a new field, so it has a new serialization format for storing the struct in the database. This new format is backward-compatible but not forward-compatible with the currently supported Zebra versions, so this PR bumps the major database format version from 25.3.0 to 26.0.0.

## Tests

Current tests should suffice.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


zebra-chain/src/block.rs
```diff
@@ -208,8 +208,8 @@ impl Block {
     /// Get the overall chain value pool change in this block,
     /// the negative sum of the transaction value balances in this block.
     ///
-    /// These are the changes in the transparent, sprout, sapling, and orchard
-    /// chain value pools, as a result of this block.
+    /// These are the changes in the transparent, Sprout, Sapling, and  Orchard chain value pools, as
+    /// a result of this block.
```
arya2:
```suggestion
    /// These are the changes in the transparent, Sprout, Sapling, Orchard, and
    /// Deferred chain value pools, as a result of this block.
```

zebra-chain/src/value_balance.rs
```diff
@@ -166,60 +180,6 @@ impl ValueBalance<NegativeAllowed> {
...
-    /// In Zebra, we define this pool as the sum of all unspent transaction outputs.
-    /// (Despite their encoding as an `int64`, transparent output values must be non-negative.)
-    ///
-    /// This is a consensus rule derived from Bitcoin:
-    ///
-    /// > because a UTXO can only be spent once,
-    /// > the full value of the included UTXOs must be spent or given to a miner as a transaction fee.
-    ///
-    /// <https://developer.bitcoin.org/devguide/transactions.html#transaction-fees-and-change>
-    pub fn add_block(
```
arya2, line 212: It could be nice to keep this method just for the documentation (with an added argument for the block's deferred pool value balance change).

zebra-chain/src/value_balance.rs
```diff
@@ -333,55 +287,72 @@ impl ValueBalance<NonNegative> {
...
-                "Four [u8; 8] should always concat with no error into a single [u8; 32]"
+                "five [u8; 8] should always concat with no error into a single [u8; 40]"
             ),
         }
     }
 
     /// From byte array
     #[allow(clippy::unwrap_in_result)]
-    pub fn from_bytes(bytes: [u8; 32]) -> Result<ValueBalance<NonNegative>, ValueBalanceError> {
+    pub fn from_bytes(bytes: &[u8]) -> Result<ValueBalance<NonNegative>, ValueBalanceError> {
```
arya2, line 342: This method will now panic if the slice is too short, could we keep the argument typed as an array but update the array's length to 40 so it panics in the `FromDisk` impl instead?

Update: This comment is addressed by the suggestion in https://github.com/ZcashFoundation/zebra/pull/8729/files#r1698699350.

zebra-consensus/src/block.rs
```diff
@@ -271,13 +283,29 @@ where
                 })?;
             }
 
+            // TODO: Add link to lockbox stream ZIP
+            let expected_deferred_amount = subsidy::funding_streams::funding_stream_values(
```
arya2, line 287: It would be nice to get this amount from `subsidy_is_valid()` in a follow-up PR for cleanups.

arya2 (commented):
This is looking good so far, the only blocker seems to be a db format upgrade. The format change won't be backwards compatible, so it may need a major version bump, though it should be straightforward to add a format downgrade for this so users won't have to re-sync the entire chain state to use a previous version of Zebra after upgrading their state cache db format.

There are some suggestions for cleanups that are probably best to leave for a follow-up PR.

zebra-chain/src/value_balance.rs
upbqdn: Would it suffice to move and keep the docs elsewhere, but not the method? It was only used in one place, adding one more layer to the call stack, and I didn't feel like it was representative of the logic we needed for the deferred balances.

zebra-chain/src/value_balance.rs
upbqdn: I did it this way so we can deserialize both 32 and 40 bytes, which makes it backward-compatible, see #8732. I'm happy to go down a different path, though.

upbqdn: > so we can deserialize both 32 and 40 bytes

I was referring to this code: https://github.com/zcashfoundation/zebra/blob/9c0c7ffc89a44b0d2a19e42031775dd63a4fc3c5/zebra-chain/src/value_balance.rs#L340-L344

upbqdn: To clarify, the implementation in #8732 is merely a trimmed-down version of the one in this PR.

zebra-consensus/src/block.rs
upbqdn: I think an even better place might be `check::miner_fees_are_valid`. I'll try doing it tomorrow.

zebra-chain/src/value_balance.rs
upbqdn: Addressed in a1b3ca618fd63af845b7838a8785680adfaf8e6d.

zebra-chain/src/value_balance.rs
```diff
@@ -20,13 +19,15 @@ mod tests;
 
 use ValueBalanceError::*;
 
+// TODO: Update the docs.
```
upbqdn: Address this TODO.

zebra-chain/src/value_balance.rs
```diff
@@ -333,55 +321,72 @@ impl ValueBalance<NonNegative> {
...
 
         let orchard = Amount::from_bytes(
             bytes[24..32]
                 .try_into()
-                .expect("Extracting the last quarter of a [u8; 32] should always succeed"),
+                .expect("orchard amount should be parsable"),
         )
         .map_err(Orchard)?;
 
+        let deferred = match bytes.len() {
```
upbqdn: Move this check to the beginning of the function.

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -93,6 +91,7 @@ pub fn funding_stream_address(
     funding_streams.recipient(receiver)?.addresses().get(index)
 }
 
+// TODO: Move this fn to [`FundingStreamReceiver`].
```
upbqdn: Address this TODO.

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -26,10 +26,7 @@ pub fn halving_divisor(height: Height, network: &Network) -> Option<u64> {
         .expect("blossom activation height should be available");
 
     if height < network.slow_start_shift() {
-        panic!(
-            "unsupported block height {height:?}: checkpoints should handle blocks below {:?}",
-            network.slow_start_shift()
-        )
+        None
```
upbqdn, line 29: We should not return `None` here.

upbqdn: Remove this condition entirely.

zebra-consensus/src/checkpoint.rs
```diff
@@ -607,8 +608,15 @@ where
             crate::block::check::equihash_solution_is_valid(&block.header)?;
         }
 
+        let expected_block_subsidy = block_subsidy(height, &self.network)?;
+
+        // TODO: Add link to lockbox stream ZIP
+        let expected_deferred_amount =
+            funding_stream_values(height, &self.network, expected_block_subsidy)?
+                .remove(&FundingStreamReceiver::Deferred);
```
arya2: This could cause a [panic](https://github.com/ZcashFoundation/zebra/actions/runs/10181559322/job/28162027058?pr=8729) because `block_subsidy()` expects to be called with heights after the slow start interval.

```suggestion
        // TODO: Add link to lockbox stream ZIP
        let expected_deferred_amount = if height > self.network.slow_start_shift() {
            funding_stream_values(height, &self.network, block_subsidy(height, &self.network)?)?
                .remove(&FundingStreamReceiver::Deferred)
        } else {
            None
        };
```

zebra-chain/src/value_balance.rs
upbqdn: Done in [d7f3df9](https://github.com/ZcashFoundation/zebra/pull/8729/commits/d7f3df9f43be4cec604a37d49499883c22ea32d5).

zebra-consensus/src/block/subsidy/funding_streams.rs
upbqdn: [16eeee4](https://github.com/ZcashFoundation/zebra/pull/8729/commits/16eeee445b22e76235ad2145718a30de667a2ad1)

zebra-chain/src/value_balance.rs
upbqdn: [16eeee4](https://github.com/ZcashFoundation/zebra/pull/8729/commits/16eeee445b22e76235ad2145718a30de667a2ad1)

zebra-chain/src/value_balance.rs
upbqdn: cc @arya2

zebra-consensus/src/checkpoint.rs
```diff
@@ -607,8 +608,16 @@ where
             crate::block::check::equihash_solution_is_valid(&block.header)?;
         }
 
+        let expected_deferred_amount = if height > self.network.slow_start_shift() {
```
arya2:
```suggestion
        let expected_deferred_amount = if height > self.network.slow_start_interval() {
```

zebra-consensus/src/block/subsidy/general.rs
upbqdn: I'll do this change as a separate PR.

upbqdn: Done in #8735.

zebra-state/src/constants.rs
```diff
@@ -46,7 +46,7 @@ pub const STATE_DATABASE_KIND: &str = "state";
 ///
 /// Instead of using this constant directly, use [`constants::state_database_format_version_in_code()`]
 /// or [`config::database_format_version_on_disk()`] to get the full semantic format version.
-const DATABASE_FORMAT_VERSION: u64 = 25;
+const DATABASE_FORMAT_VERSION: u64 = 26;
```
arya2, line 49: Could we bump the minor db version and store the new `ValueBalance` disk format with a different key?

zebra-chain/src/value_balance.rs
```diff
@@ -333,55 +319,80 @@ impl ValueBalance<NonNegative> {
...
     /// From byte array
     #[allow(clippy::unwrap_in_result)]
     pub fn from_bytes(bytes: &[u8]) -> Result<ValueBalance<NonNegative>, ValueBalanceError> {
+        let bytes_length = bytes.len();
+
+        // Return an error early if bytes don't have the right lenght instead of panicking later.
+        match bytes_length {
+            32 | 40 => {}
+            _ => return Err(Unparsable),
+        };
```
arya2, line 349: ###### Optional / Nitpick

We could combine the match statements in this method if we're okay with assigning bytes out of order:

```rust
let deferred = match bytes.len() {
    32 => Amount::zero(),
    40 => Amount::from_bytes(
        bytes[32..40]
            .try_into()
            .expect("deferred amount should be parsable"),
    )
    .map_err(Deferred)?,
    _ => return Err(Unparsable),
};
```

upbqdn: I prefer having the check first and keeping reading the bytes in order.

zebra-state/src/constants.rs
upbqdn: Bumping the database version seems like the right solution to me because it doesn't require workarounds such as storing the new `ValueBalance` under a new key. ~~We also did previous database format upgrades such as pruning the note commitment trees where previous Zebra versions couldn't read the new format anymore.~~

oxarbitrage:
Can we rebase to get rid of extra commits ?

zebra-state/src/constants.rs
arya2: > it doesn't require workarounds such as storing the new ValueBalance under a new key

We could make it temporary and add another database format change that moves it back to the `()` key now that the code is compatible with both formats, it would be nice to avoid requiring zebrad to sync from scratch for such a small format change, especially just before a network upgrade.

arya2:
> Can we rebase to get rid of extra commits ?

I'd prefer that we avoid a rebase since there would be a lot of major conflicts with a relatively high chance of one being resolved incorrectly. We could try squashing the changes? (Though I think this is fine as-is).

arya2 (approved):

arya2:
Thank you for this change!

The full sync on Mainnet with almost all of the changes [succeeded here](https://github.com/ZcashFoundation/zebra/actions/runs/10201923807), and we just checked that the Testnet sync passed as well in GCP (though Github is having trouble updating the test's status).

It looks like the full sync tests that are still pending in Github aren't required for merging into the main branch anyway, so this should merge soon.

PR #8737, opened 2024-08-02
suggestion: Maintain backwards-compatible database format for `ValueBalance`

arya2:
This is a suggestion for @upbqdn for PR #8729


arya2:
We decided to do #8745 instead.

PR #8746, opened 2024-08-06
deps(librustzcash): Upgrade zcash dependencies for NU6

### LINKED ISSUE #8713: Upgrade `zcash_protocol`/`zcash_primitives` to a version that includes the NU6 NetworkUpgrade variant without `zcash_unstable = "nu6"`

oxarbitrage:
## Motivation

We want to use the current main version of librustzcash so we can remove unstable features.

We will get back to depend on a release after librustzcash components are released.

Close https://github.com/ZcashFoundation/zebra/issues/8713

## Solution

Upgrade, fix problems.

Draft until CI pass as there might be issues for example in the deny.toml.

### Tests

No tests were added as our current suite should be enough coverage.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -109,6 +109,7 @@ unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 # List of URLs for allowed Git repositories
 allow-git = [
+    https://github.com/zcash/librustzcash.git
 ]
 
 [sources.allow-org]
--- a/deny.toml
+++ b/deny.toml
@@ -12,7 +12,7 @@
 multiple-versions = "deny"
 
 # Don't allow wildcard ("any version") dependencies
-# TODO: Change to `deny` after weupgrade to crates.io version of the ECC dependencies (#8746).
+# TODO: Change to `deny` after we upgrade to crates.io version of the ECC dependencies (#8749).
 wildcards = "allow"
 # Allow private and dev wildcard dependencies.
 # Switch this to `false` when #6924 is implemented.
@@ -94,7 +94,7 @@ skip-tree = [
     { name = "http-body", version = "=0.4.6" },
     { name = "hyper", version = "=0.14.30" },
 
-    # Remove this after we upgrade ECC dependencies to a crates.io version (#8746)
+    # Remove this after we upgrade ECC dependencies to a crates.io version (#8749)
     { name = "equihash", version = "=0.2.0" },
     { name = "f4jumble", version = "=0.1.0" },
     { name = "secp256k1", version = "=0.26.0" },
--- a/deny.toml
+++ b/deny.toml
@@ -94,7 +94,7 @@ skip-tree = [
     { name = "http-body", version = "=0.4.6" },
     { name = "hyper", version = "=0.14.30" },
 
-    # Remove this after we upgrade ECC dependencies to a crates.io version (#8749)
+    # TODO: Remove this after we upgrade ECC dependencies to a crates.io version (#8749)
     { name = "equihash", version = "=0.2.0" },
     { name = "f4jumble", version = "=0.1.0" },
     { name = "secp256k1", version = "=0.26.0" },
--- a/deny.toml
+++ b/deny.toml
@@ -12,7 +12,8 @@
 multiple-versions = "deny"
 
 # Don't allow wildcard ("any version") dependencies
-wildcards = "deny"
+# TODO: Change to `deny` after weupgrade to crates.io version of the ECC dependencies (#8746).
+wildcards = "allow"
 # Allow private and dev wildcard dependencies.
 # Switch this to `false` when #6924 is implemented.
 allow-wildcard-paths = true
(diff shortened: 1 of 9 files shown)


zebra-scan/src/service/scan_task/scan.rs
```diff
@@ -555,3 +555,11 @@ pub fn dfvk_to_ufvk(dfvk: &DiversifiableFullViewingKey) -> Result<UnifiedFullVie
     ))?])?)
     .map_err(|e| eyre!(e))
 }
+
+/// Returns the `zcash_primitives::consensus::Network` for this network.
```
upbqdn: ###### Optional

```suggestion
/// Returns the [`zcash_primitives::consensus::Network`] for this network.
```

oxarbitrage:
The "Release crates / Check crate release dry run (pull_request)" check will not pass for this PR because we are relying on a GitHub repository for `librustzcash` instead of a version from crates.io. This is not allowed in our release process.

If this PR gets approved, we will need to perform an admin merge. As a side effect, all future pull requests will also require admin merges until we can revert to a proper release. An alternative is to make the "Release crates / Check crate release dry run (pull_request)" check not required. The test will fail, but we will still be able to merge. This should be a temporary measure, but it seems better than having to admin merge every subsequent pull request.

deny.toml
```diff
@@ -92,6 +93,15 @@ skip-tree = [
     { name = "http", version = "=0.2.12" },
     { name = "http-body", version = "=0.4.6" },
     { name = "hyper", version = "=0.14.30" },
+
+    # Remove this after we upgrade ECC dependencies to a crates.io version (#8749)
```
upbqdn:
```suggestion
    # TODO: Remove this after we upgrade ECC dependencies to a crates.io version (#8749)
```

deny.toml
```diff
@@ -12,7 +12,8 @@
 multiple-versions = "deny"
 
 # Don't allow wildcard ("any version") dependencies
-wildcards = "deny"
+# TODO: Change to `deny` after we upgrade to crates.io version of the ECC dependencies (#8749).
+wildcards = "allow"
```
upbqdn, line 16: Why do we need this?

upbqdn (dismissed):

deny.toml
oxarbitrage: When we do for example `zcash_address = { git = "https://github.com/zcash/librustzcash/", branch = "main"}` we are actually using a wildcard that will match any version `main` has last. `cargo-deny` will fail if we don't make this change.

upbqdn (approved):

mpguerra:
>An alternative is to make the "Release crates / Check crate release dry run (pull_request)" check not required. The test will fail, but we will still be able to merge. This should be a temporary measure, but it seems better than having to admin merge every subsequent pull request.

Let's go ahead and do this then. Does anyone know how/where to do this?

zebra-chain/Cargo.toml
```diff
@@ -97,10 +97,13 @@ orchard = "0.8.0"
...
-zcash_primitives = { version = "0.15.0", features = ["transparent-inputs"] }
+# TODO: Revert to a release once librustzcash is released (#8749).
+zcash_primitives = { git = "https://github.com/zcash/librustzcash/", branch = "main", features = ["transparent-inputs"] }
 sapling = { package = "sapling-crypto", version = "0.1" }
-zcash_protocol = { version = "0.1.1" }
-zcash_address = { version = "0.3.2" }
+# TODO: Revert to a release once librustzcash is released (#8749).
+zcash_protocol = { git = "https://github.com/zcash/librustzcash/", branch = "main"}
+# TODO: Revert to a release once librustzcash is released (#8749).
+zcash_address = { git = "https://github.com/zcash/librustzcash/", branch = "main"}
```
conradoplg, line 106: I think it's preferrable to point to a commit in the main branch, and not the main branch itself, e.g. 

`{ git = "https://github.com/zcash/librustzcash/", rev = "5a4a3e06dcd2cf5bdf79a7cd48709b58693c65f0" }`

This will make it less likely to suddenly break if something incompatible is merged into their `main`.

conradoplg (changes requested):
I think it's preferrable to point to a commit in the main branch, and not the main branch itself, e.g. 

`{ git = "https://github.com/zcash/librustzcash/", rev = "5a4a3e06dcd2cf5bdf79a7cd48709b58693c65f0" }`

This will make it less likely to suddenly break if something incompatible ends up being merged into their `main`.

oxarbitrage:
> I think it's preferrable to point to a commit in the main branch, and not the main branch itself, e.g.
> ...

I think we can do that for sure but by the talks we had with the ECC engineers in multiple meetings we decided to use the last `main` as all that it is and will be there is part of the next release, reducing the delta.

> This will make it less likely to suddenly break if something incompatible ends up being merged into their main.

My understanding is that we actually want to break if something like that happens.

I don't mind changing this up to a commit but just wanted to point that out.

conradoplg:
> > I think it's preferrable to point to a commit in the main branch, and not the main branch itself, e.g.
> > ...
> 
> I think we can do that for sure but by the talks we had with the ECC engineers in multiple meetings we decided to use the last `main` as all that it is and will be there is part of the next release, reducing the delta.
> 
> > This will make it less likely to suddenly break if something incompatible ends up being merged into their main.
> 
> My understanding is that we actually want to break if something like that happens.
> 
> I don't mind changing this up to a commit but just wanted to point that out.

I realized this is all moot since `Cargo.lock` already ensures the same commit will be used even if `main` is updated, so we're good. (I'm so used to working on a library that I forget about this :laughing: )

conradoplg (approved):

oxarbitrage:
> I realized this is all moot since Cargo.lock already ensures the same commit will be used even if main is updated, so we're good. (I'm so used to working on a library that I forget about this 😆 )

That's good to know, thanks for pointing it out.

oxarbitrage:
@mergifyio refresh

PR #8748, opened 2024-08-08
change(state): Support in-place disk format upgrades for major database version bumps

### LINKED ISSUE #8745: Support in-place disk format upgrades for major database version bumps
## Motivation

We want to avoid requiring a full sync for major Zebra database format version bumps

upbqdn:
## Motivation

Close #8745.

## Solution

- Try to move an existing cached state to a new location so it can be used again.

### Tests

- I manually tested the solution.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -15,7 +15,7 @@ use tracing::Span;
 use zebra_chain::parameters::Network;
 
 use crate::{
-    constants::{DATABASE_FORMAT_VERSION_FILE_NAME, STATE_DATABASE_KIND},
+    constants::{DATABASE_FORMAT_VERSION_FILE_NAME, RESTORABLE_DB_VERSIONS, STATE_DATABASE_KIND},
     state_database_format_version_in_code, BoxError,
 };
 
@@ -314,6 +314,15 @@ fn check_and_delete_database(
     let dir_major_version = parse_major_version(&dir_name)?;
 
     if dir_major_version >= major_version {
+        return None;
+    }
+
+    // Don't delete databases that can be reused.
+    if RESTORABLE_DB_VERSIONS
+        .iter()
+        .map(|v| v - 1)
+        .any(|v| v == dir_major_version)
+    {
         return None;
     }
 
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -1000,6 +1000,9 @@ impl DiskDb {
             let new_db_exists = DB::list_cf(&opts, &new_path).is_ok_and(|cf| !cf.is_empty());
 
             if old_db_exists && !new_db_exists {
+                // Create the required folder path for the destination.
+                // This is because we can't directly rename e.g. `state/v25/mainnet/` to `state/v26/mainnet/`
+                // with `fs::rename()` because `state/v26/` does not exist.
                 match fs::create_dir_all(&new_path) {
                     Ok(()) => info!("created new directory for state cache at {new_path:?}"),
                     Err(e) => {
@@ -1021,6 +1024,8 @@ impl DiskDb {
                             }
                         }
 
+                        // Get the parent of the old path, e.g. `state/v25/` and delete it if it is empty.
+                        
                         let old_path = old_path
                             .parent()
                             .expect("old state cache must have parent path");
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -1000,9 +1000,9 @@ impl DiskDb {
             let new_db_exists = DB::list_cf(&opts, &new_path).is_ok_and(|cf| !cf.is_empty());
 
             if old_db_exists && !new_db_exists {
-                // Create the required folder path for the destination.
-                // This is because we can't directly rename e.g. `state/v25/mainnet/` to `state/v26/mainnet/`
-                // with `fs::rename()` because `state/v26/` does not exist.
+                // Create the required folder path for the destination. This is because we can't
+                // directly rename e.g. `state/v25/mainnet/` to `state/v26/mainnet/` with
+                // `fs::rename()` because `state/v26/` does not exist.
                 match fs::create_dir_all(&new_path) {
                     Ok(()) => info!("created new directory for state cache at {new_path:?}"),
                     Err(e) => {
@@ -1024,8 +1024,8 @@ impl DiskDb {
                             }
                         }
 
-                        // Get the parent of the old path, e.g. `state/v25/` and delete it if it is empty.
-                        
+                        // Get the parent of the old path, e.g. `state/v25/` and delete it if it is
+                        // empty.
                         let old_path = old_path
                             .parent()
                             .expect("old state cache must have parent path");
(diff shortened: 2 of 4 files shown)


zebra-state/src/config.rs
```diff
@@ -317,6 +317,15 @@ fn check_and_delete_database(
         return None;
     }
 
+    // Don't delete databases that can be reused.
```
arya2, line 320: There are some extra checks in `try_reusing_previous_db_after_major_upgrade()` that would be nice to have here as well.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -940,6 +943,111 @@ impl DiskDb {
...
+                        );
+                        return;
+                    }
+                };
+
+                match fs::rename(&old_path, &new_path) {
+                    Ok(()) => {
+                        info!("moved state cache from {old_path:?} to {new_path:?}");
+
+                        match fs::remove_file(new_path.join(DATABASE_FORMAT_VERSION_FILE_NAME)) {
```
arya2, line 1024: We should keep the version file for when multiple format upgrades need to be applied.

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -106,6 +107,14 @@ impl ZebraDb {
         )
         .expect("unable to read database format version file");
 
+        DiskDb::try_reusing_previous_db_after_major_upgrade(
```
arya2, line 110: It would be nice to:
- apply any required format upgrades before returning the db when the code format version is a higher major version than the disk format version, 
- apply any format upgrades for the previous major version at the old path, and,
- apply the major format upgrade in a temp directory, then move it to the new db path once it's been upgraded.

arya2 (commented):
This is looking great, thank you!

zebra-state/src/service/finalized_state/disk_db.rs
upbqdn: The version file doesn't contain the major version, only the minor and patch versions. When we do a major version bump, we need to set the minor and patch versions to zero. If we transfer the db with its original version file containing non-zero versions, we'll end up with a malformed db version. Zebra recreates the version file if it's missing and sets the minor and patch versions to zero, so deleting the file is a simple means to reset the counters for minor and patch versions.

zebra-state/src/config.rs
upbqdn: Do you have any specific ones in mind? If yes, let's add them to this PR, otherwise let's do it in a future one?

zebra-state/src/service/finalized_state/zebra_db.rs
upbqdn: This sounds like it would make opening the db much slower than it is. Currently, opening the db is fast because db upgrades run concurrently, and moving the db is instant because Zebra just renames the path.

I'm now thinking we should delete all previous db upgrades, and keep only the framework for doing them. The reason is that we don't need them anymore as there aren't and won't be any Zebra instances that could use them. If we ever need another DB upgrade, we can either:

- transfer the old database and use the framework for upgrades, or
- apply the framework for upgrades to the current db version as we did in the past, or
- start syncing from scratch.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -940,6 +943,111 @@ impl DiskDb {
...
+                    Ok(()) => info!("created new directory for state cache at {new_path:?}"),
+                    Err(e) => {
+                        warn!(
+                            "could not create new directory for state cache at {new_path:?}: {e}"
+                        );
+                        return;
+                    }
+                };
+
+                match fs::rename(&old_path, &new_path) {
```
conradoplg, line 1020: From the `fs::rename()` docs, it seems this will fail if `new_path` exists which it should since it was just created. So I think you will need to remove the leaf folder from `new_path` above.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -940,6 +943,111 @@ impl DiskDb {
...
+                info!("skipped reusing previous state cache: state is outside cache directory");
+                return;
+            }
+
+            let opts = DiskDb::options();
+            let old_db_exists = DB::list_cf(&opts, &old_path).is_ok_and(|cf| !cf.is_empty());
+            let new_db_exists = DB::list_cf(&opts, &new_path).is_ok_and(|cf| !cf.is_empty());
+
+            if old_db_exists && !new_db_exists {
+                match fs::create_dir_all(&new_path) {
```
conradoplg:
```suggestion
                // Create the required folder path for the destination.
                // This is because we can't directly rename e.g. `state/v25/mainnet/` to `state/v26/mainnet/`
                // with `fs::rename()` because `state/v26/` does not exist.
                match fs::create_dir_all(&new_path) {
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -940,6 +943,111 @@ impl DiskDb {
...
+                        match fs::remove_file(new_path.join(DATABASE_FORMAT_VERSION_FILE_NAME)) {
+                            Ok(()) => info!("removed version file at {new_path:?}"),
+                            Err(e) => {
+                                warn!("could not remove version file at {new_path:?}: {e}")
+                            }
+                        }
+
+                        let old_path = old_path
+                            .parent()
+                            .expect("old state cache must have parent path");
```
conradoplg, line 1035:
```suggestion
                        // Get the parent of the old path, e.g. `state/v25/` and delete it if it is empty.
                        
                        let old_path = old_path
                            .parent()
                            .expect("old state cache must have parent path");
```

zebra-state/src/service/finalized_state/zebra_db.rs
conradoplg: Note that it is tricky to use temp directories because they can be in another device and the move won't work, unless we make sure we create the temp directory alongisde the db folders (which may have unintended consequences?)

conradoplg (changes requested):
There is a small fix required for Windows.

I was confused by some of the code but I think I got it. I suggested some comments which should hopefully help whoever reads the code in the future, feel free to tweak them.

Also I noticed the upgrade itself is not here, was it planned to be done in another PR?

zebra-state/src/service/finalized_state/disk_db.rs
upbqdn: Ah, you're referring to this part in the docs?

> In contrast, on Windows, `from` can be anything, but `to` must not be a directory.

upbqdn:
> I was confused by some of the code but I think I got it. I suggested some comments which should hopefully help whoever reads the code in the future, feel free to tweak them.

I originally had similar comments in the code, but I deleted them because I thought the `info` and `warn` messages were descriptive enough. Thanks for adding them.

zebra-state/src/service/finalized_state/disk_db.rs
upbqdn: Addressed in [99cf990](https://github.com/ZcashFoundation/zebra/pull/8748/commits/99cf99065d394ed535a2e8d575e0a62c341f3522).

upbqdn:
> Also I noticed the upgrade itself is not here, was it planned to be done in another PR?

Yes, this PR only moves the old db to a new path. Zebra then uses the db and rewrites the `ValueBalance`, stored under the same key as the original one. Current Zebra can read both the original and new format of `ValueBalance` from the db, but Zebra 1.8 can't, so we needed to bump the major db version. If we didn't move the db, Zebra 1.9 would start syncing the cached state from scratch.

The new `ValueBalance` has 40 bytes, and the original one has 32. It was trivial to implement deserialization of both formats, so we got backward compatibility with previous Zebra versions, but previous Zebra versions are not forward-compatible with the change since they'll panic when they try to deserialize the new format.

conradoplg (approved):
Ah, got it, thanks. LGTM!

PR #8750, opened 2024-08-08
chore(actions): remove warnings related to gcp and docker steps

gustavovalverde:
## Motivation

Having warnings that are not helpful in our workflows summaries, makes warning that we should care about less relevant or harder to find.

image

## Solution

- Remove deprecated arguments
- Deactivate Docker summary (not supported by Docker Build Cloud, yet)
  - Use a variable `DOCKER_BUILD_SUMMARY ` and set it to false, as this feature might be available in the future

### Follow-up Work

I also tried solving this warning:
- `Unable to extend GITHUB_TOKEN expiration time due to: GITHUB_TOKEN has expired.`

But it seems we won't be able to do it, based on this discussion: 
- https://github.com/orgs/community/discussions/50472

### PR Author's Checklist

- [X] The solution is tested.

### DIFF
--- a/.github/workflows/cd-deploy-nodes-gcp.yml
+++ b/.github/workflows/cd-deploy-nodes-gcp.yml
@@ -230,7 +230,6 @@ jobs:
         id: auth
         uses: google-github-actions/auth@v2.1.3
         with:
-          retries: '3'
           workload_identity_provider: '${{ vars.GCP_WIF }}'
           service_account: '${{ vars.GCP_DEPLOYMENTS_SA }}'
 
@@ -332,7 +331,6 @@ jobs:
         id: auth
         uses: google-github-actions/auth@v2.1.3
         with:
-          retries: '3'
           workload_identity_provider: '${{ vars.GCP_WIF }}'
           service_account: '${{ vars.GCP_DEPLOYMENTS_SA }}'
 
--- a/.github/workflows/sub-build-docker-image.yml
+++ b/.github/workflows/sub-build-docker-image.yml
@@ -75,6 +75,8 @@ jobs:
     permissions:
       contents: 'read'
       id-token: 'write'
+    env:
+      DOCKER_BUILD_SUMMARY: ${{ vars.DOCKER_BUILD_SUMMARY }}
     steps:
       - uses: actions/checkout@v4.1.7
         with:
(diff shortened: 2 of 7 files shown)


.github/workflows/ci-test-librustzcash.yml
mpguerra, line 1: This seems unrelated to the PR title. Should we split this out into another PR?

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -230,7 +230,6 @@ jobs:
         id: auth
         uses: google-github-actions/auth@v2.1.4
         with:
-          retries: '3'
```
mpguerra, line 233: Are these the deprecated arguments we are removing? Otherwise I'd like to understand why we are we removing the retries

.github/workflows/sub-build-docker-image.yml
```diff
@@ -75,6 +75,8 @@ jobs:
     permissions:
       contents: 'read'
       id-token: 'write'
+    env:
+      DOCKER_BUILD_SUMMARY: ${{ vars.DOCKER_BUILD_SUMMARY }}
```
mpguerra, line 79: In your PR comment you said this is not supported by docker build cloud. Why are we adding it here?

mpguerra (dismissed):
Looks good to me, I just have some questions to try to understand the changes.

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: Yes, those have been deprecated 
image

.github/workflows/sub-build-docker-image.yml
gustavovalverde: Oh, I'm adding it as a variable, which defaults to false, just in case is supported later on and we'd like to activate it.

.github/workflows/ci-test-librustzcash.yml
gustavovalverde: Oh, it was late and I did not see I added this extra file 😅  . I'll fix this and force-push it.

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: I've update the PR comment with this info.

.github/workflows/sub-build-docker-image.yml
gustavovalverde: I've update the PR comment with this info.

upbqdn (approved):

gustavovalverde:
@Mergifyio refresh

mpguerra (approved):

PR #8751, opened 2024-08-08
Use zcash_script’s new `Script` trait

sellout:
## Motivation

This is a precursor to testing the Rust implementation of Zcash Script.

Blocked-On: https://github.com/ZcashFoundation/zcash_script/pull/171 and publishing a new version of `zcash_script` with those changes.

## Solution

This uses a trait that wraps the C++ Zcash Script implementation. As we progress toward cutting over to a Rust implementation, this trait will additionally have impls for the Rust version and a variant that runs both C++ & Rust, comparing the results.

Additionally, this eliminates a few cases from `zebra_script::Error` that can never be produced.

### Tests

The tests are the same as the previous zebra_script tests – there should be no change in behavior.

### Follow-up Work

This is blocked on a release of zcash_script containing ZcashFoundation/zcash_script#171.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


sellout:
This depends on ZcashFoundation/zcash_script#171.

zebra-script/src/lib.rs
```diff
@@ -202,33 +149,30 @@ impl CachedFfiTransaction {
...
+        let ret = zcash_script::Cxx::verify_callback(
+            &|script_code, hash_type| {
+                let script_code_vec = script_code.to_vec();
+                Some(
+                    (*ctx).sighasher.sighash(
+                        HashType::from_bits_truncate(hash_type.bits() as u32),
+                        Some(((*ctx).input_index, script_code_vec)),
+                    ).0
+                )
+            },
```
sellout: I currently get some lifetime errors around this callback. They make sense, but I don’t know how to avoid them.
```
error: lifetime may not live long enough
   --> zebra-script/src/lib.rs:153:13
    |
111 |       pub fn is_valid(&self, branch_id: ConsensusBranchId, input_index: usize) -> Result<(), Error> {
    |                       - let's call the lifetime of this reference `'1`
...
153 | /             &|script_code, hash_type| {
154 | |                 let script_code_vec = script_code.to_vec();
155 | |                 Some(
156 | |                     (*ctx).sighasher.sighash(
...   |
160 | |                 )
161 | |             },
    | |_____________^ cast requires that `'1` must outlive `'static`

error[E0597]: `ctx.input_index` does not live long enough
   --> zebra-script/src/lib.rs:158:31
    |
153 |               &|script_code, hash_type| {
    |               ------------------------- value captured here
    |  _____________|
    | |
154 | |                 let script_code_vec = script_code.to_vec();
155 | |                 Some(
156 | |                     (*ctx).sighasher.sighash(
157 | |                         HashType::from_bits_truncate(hash_type.bits() as u32),
158 | |                         Some(((*ctx).input_index, script_code_vec)),
    | |                               ^^^^^^^^^^^^^^^^^^ borrowed value does not live long enough
159 | |                     ).0
160 | |                 )
161 | |             },
    | |_____________- cast requires that `ctx.input_index` is borrowed for `'static`
...
170 |       }
    |       - `ctx.input_index` dropped here while still borrowed
```

sellout: `verify_callback` constrains the closure with `Fn`. I feel like I should be able to weaken the constraint to `FnOnce`, but I end up with different errors (on the zcash_script side, rather than the zebra side) for both `FnOnce` and `FnMut`.

mpguerra:
Thank you for your PR! We're focusing on changes for NU6 testnet activation at the moment so we won't be able to prioritise this review for another week or two.

zebra-script/src/lib.rs
arya2, line 1: This file needs formatting.

zebra-script/src/lib.rs
arya2: ~It might work with `move |script_code, hash_type| {`, if `verify_callback()` can take an owned closure, I don't think making it a `FnOnce` or `FnMut` will help here since it's not consuming or mutating anything as far as I can tell, it looks like `ctx` just might be dropped early if it's not moved and if the closure is returned from this function or passed to another thread.~

Update: I'll give this another go shortly.

zebra-script/src/lib.rs
```diff
@@ -6,19 +6,13 @@
...
 
-use zcash_script::{
-    zcash_script_error_t, zcash_script_error_t_zcash_script_ERR_OK,
-    zcash_script_error_t_zcash_script_ERR_TX_DESERIALIZE,
-    zcash_script_error_t_zcash_script_ERR_TX_INDEX,
-    zcash_script_error_t_zcash_script_ERR_TX_SIZE_MISMATCH,
-};
+use zcash_script;
+use zcash_script::api as zscript;
+use zcash_script::api::Script;
```
```suggestion
use zcash_script as zscript;
use zcash_script::ZcashScript;
```

zebra-script/Cargo.toml
```diff
@@ -15,7 +15,7 @@ keywords = ["zebra", "zcash"]
 categories = ["api-bindings", "cryptography::cryptocurrencies"]
 
 [dependencies]
-zcash_script = "0.2.0"
```
nuttycom: This change is unnecessary, use a `[patch.crates-io]` section.

zebra-script/src/lib.rs
sellout: Thanks for your assistance with this!

sellout:
Re: the unchecked checkboxes
- there are no visible changes to add to the CHANGELOG and
- I don’t have permissions to add labels.

arya2:
> Re: the unchecked checkboxes
> * there are no visible changes to add to the CHANGELOG and

We usually check boxes if they don't apply to the PR.

> * I don’t have permissions to add labels.

Added a priority label.

Added a `do-not-merge` label as well, we should wait until the changes in `zcash_script` have been published before merging this PR.

mpguerra:
> Added a `do-not-merge` label as well, we should wait until the changes in `zcash_script` have been published before merging this PR.

What's next here? I see https://github.com/ZcashFoundation/zcash_script/pull/171 has merged

conradoplg:
> What's next here? I see [ZcashFoundation/zcash_script#171](https://github.com/ZcashFoundation/zcash_script/pull/171) has merged

We should double check this PR works with zcash_script `main` branch, and then I can make a zcash_script release and we point to that instead.

sellout:
> We should double check this PR works with zcash_script `main` branch, and then I can make a zcash_script release and we point to that instead.

I can do this, but wondering how to show that it’s successful – should I put up a temporary PR that just fetches zcash_script from GH instead of the crate registry and shows that everything still passes? And we’ll just close it afterward?

Also, I’m happy for there to be a zcash_script release now, but I figured it would be fine to leave it until the next PR or two land on zcash_script, which contain the actual Rust implementation.

conradoplg:
> I can do this, but wondering how to show that it’s successful – should I put up a temporary PR that just fetches zcash_script from GH instead of the crate registry and shows that everything still passes? And we’ll just close it afterward?
> 
> Also, I’m happy for there to be a zcash_script release now, but I figured it would be fine to leave it until the next PR or two land on zcash_script, which contain the actual Rust implementation.

You can simply change Cargo.toml in this PR to point `zcash_script` to a specific commit in its repo (i.e. the one corresponding to its current `main`), which I think would also make CI pass here. (I can make the change if you prefer, let me know)

This issue with delaying the zcash_script release is that if this PR merges before the next Zebra release, this would prevent us from publishing the zebra-script crate (which we usually do at the same time) since it would depend on a git repo and crates.io doesn't allow that. We have several options though: don't merge this PR until all changes in zcash_script are done, or just make a zcash_script release with what it currently has and point to that. We'll figure it out, I think the important thing now is to just make sure this PR works with the current zcash_script.

sellout:
> You can simply change Cargo.toml in this PR to point `zcash_script` to a specific commit in its repo (i.e. the one corresponding to its current `main`), which I think would also make CI pass here.

Ok, this now points at the zcash_script master branch. I also made a PR (#8896, to be discarded) that shows that the current Zebra main branch (without this PR) will also work with the new zcash_script.

mpguerra:
> > What's next here? I see [ZcashFoundation/zcash_script#171](https://github.com/ZcashFoundation/zcash_script/pull/171) has merged
> 
> We should double check this PR works with zcash_script `main` branch, and then I can make a zcash_script release and we point to that instead.

I guess @sellout has done this already in #8896? 

@conradoplg shall we do the zcash_script release and merge this?

sellout:
I should have ZcashFoundation/zcash_script#175 merged today, which is effectively an addendum to ZcashFoundation/zcash_script#171, so it’s probably worth getting that in before the release … and I think it will involve at least a small change to this PR as a consequence.

conradoplg:
I went ahead and updated this for the latest zcash_script which should be now working.

CI should fail only the cargo deny sources check since it's pointing to `zcash_script` master branch.

sellout:
I wonder if the `ZcashScript` instance that’s used should be handled by a feature flag? I intended the `CxxRustComparisonInterpreter` to be mostly for testing, expecting one or the other to be in the release.

conradoplg:
> I wonder if the `ZcashScript` instance that’s used should be handled by a feature flag? I intended the `CxxRustComparisonInterpreter` to be mostly for testing, expecting one or the other to be in the release.

A feature flag sounds good. I seem to recall that the plan was to run both together for a while until we have assurance that the Rust one is equivalent, and then switch to the Rust one and get rid of the C++ one.

nuttycom:
> > I wonder if the `ZcashScript` instance that’s used should be handled by a feature flag? I intended the `CxxRustComparisonInterpreter` to be mostly for testing, expecting one or the other to be in the release.
> 
> A feature flag sounds good. I seem to recall that the plan was to run both together for a while until we have assurance that the Rust one is equivalent, and then switch to the Rust one and get rid of the C++ one.

Yeah, I think it's important for some nodes to be running both simultaneously, taking the results from the C++ interpreter but loudly announcing any divergences. Probably even halting if there's a divergence, so that the node is left in a state where it's debuggable.

zebra-script/src/lib.rs
```diff
@@ -198,31 +128,31 @@ impl CachedFfiTransaction {
...
+                    .0,
             )
         };
-
-        if ret == 1 {
-            Ok(())
-        } else {
-            Err(Error::from(err))
-        }
+        zcash_script::CxxRustComparisonInterpreter::verify_callback(
```
nuttycom: Does the comparison interpreter always perform the comparisons, or is this controlled by feature flags?

sellout: The comparison interpreter always performs the comparison. I think we would want a feature flag in Zebra to select the interpreter that’s used. Seems like a reasonable set of cases would be
- c++
- rust
- comparison
- stepwise comparison

Unfortunately, I can’t think of a comprehensible way to break it down into two orthogonal features. One clear dimension is whether or not it’s a comparison.

sellout: Actually, a slightly different suggestion:

I think zcash_script should have a `rust_interpreter` feature (@ebfull had this in his draft, but I had removed it). This fits with the future organization, IMO, when the rust impl will live in zcash_transparent, and so the dependency on zcash_transparent would be conditional on the feature.

If that feature is disabled, then the C++ interpreter is the _only_ option.

Then in zebra … should it be a feature or configuration to select the particular interpreter (if the zcash_script `rust_interpreter` feature is disabled, then you’ll get the C++ interpreter no matter what the configuration says).

mpguerra:
Ideally we should resolve conflicts here before the next review

conradoplg:
I fixed conflicts and tried to update the existing code but Rust is not happy the change from `SighashCalculator` to `&SighashCalculator` (to be honest I never understood [why it even worked before](https://users.rust-lang.org/t/passing-a-closure-to-an-external-c-ffi-library/100271)). I'm not sure what the proper fix is or if it might require changing the zcash_script API.

This didn't turn up in zcash_script tests because there it always passes a regular function and not a closure.

@sellout is this something you can look into, or would you like for us to investigate?

sellout:
> This didn't turn up in zcash_script tests because there it always passes a regular function and not a closure.
> 
> @sellout is this something you can look into, or would you like for us to investigate?

Yeah, I’ll look at this. I’m making other changes in zcash_script now, so I’ll add a test with a closure and make sure that’s working.

sellout:
@conradoplg I put up ZcashFoundation/zcash_script#216. I tried it against this branch (with some additional changes here) and it worked.

I think that counts as a bugfix release, as it doesn’t affect the cases that already worked. But there are changes in ZcashFoundation/zcash_script#209 (and others coming) that do affect the interface (although in fairly minor ways – like the set of `Error` variants).

conradoplg:
This should be good to go. I did the simplest thing and just added a `comparison-interpreter` feature that enables the comparison interpreter.

I tested by manually pointing to my local copy of `zcash_script`, and adding a warning there just before the comparison check, regardless of the result, and confirmed the warning appeared if and only if the feature was enabled.

oxarbitrage (dismissed):
please fix conflicts.

conradoplg:
> please fix conflicts.

Oops, done

oxarbitrage (approved):

conradoplg:
I'm admin-merging this, because there is some CI issue making some jobs to be forever pending. We'll investigate the [issue](https://github.com/ZcashFoundation/zebra/issues/9622) later with a external PR of our own. (My guess is some patch job issue)

str4d (commented):
Post-hoc utACK (working through my review backlog)

PR #8761, opened 2024-08-14
cleanup(state): Run possible minor database format version upgrades before deleting the disk version

arya2:
## Motivation

Users could manually change the end of support time in code, use a version of Zebra published without an end-of-support halt, or have an outdated state cache produced by a version of Zebra that has reached its end-of-support but hasn’t been run in a long time (perhaps on Testnet).

We may also publish minor version upgrades in the future shortly before a major version upgrade that can be restored from a previous database format.

## Solution

- Run minor database format upgrades before deleting the older disk version

Related changes:
- Delete any old cache directories without checking if they're restorable after any restorable version should have been renamed

### Tests

Needs manual testing:
- starting a sync in a new cache directory with an old version of Zebra and checking that minor db format upgrades are applied
- starting a sync with the most recent release of Zebra and checking that it skips applying minor format upgrades

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


mpguerra:
How important is this to merge soon? I would like @upbqdn to review it but he's not going to be back for a couple of weeks.

arya2:
> How important is this to merge soon? I would like @upbqdn to review it but he's not going to be back for a couple of weeks.

It's not very important to merge soon. We can wait.

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -128,25 +120,63 @@ impl ZebraDb {
...
+                    read_only,
+                ),
+            }
         };
 
+        match format_change.clone() {
+            DbFormatChange::Upgrade {
+                older_disk_version,
+                newer_running_version,
+            } if !debug_skip_format_upgrades
```
arya2, line 149: This needs a correctness note about not running `if read_only` so there's no corruption if multiple Zebra instances try to read and write the same files.

upbqdn (commented):
Could we open an issue and do a refactor like this with any upcoming database upgrade instead? The upgrade will give us an opportunity to do thorough manual testing. I can open the issue.

arya2:
> Could we open an issue and do a refactor like this with any upcoming database upgrade instead? The upgrade will give us an opportunity to do thorough manual testing. I can open the issue.

I'd prefer to do thorough manual testing sooner to make outdated caches more usable, but we could split out everything except checking `if !debug_skip_format_upgrades` before `try_reusing_previous_db_after_major_upgrade()`.

zebra-state/src/config.rs
```diff
@@ -317,15 +317,6 @@ fn check_and_delete_database(
...
     }
 
-    // Don't delete databases that can be reused.
-    if RESTORABLE_DB_VERSIONS
-        .iter()
-        .map(|v| v - 1)
-        .any(|v| v == dir_major_version)
-    {
-        return None;
-    }
```
upbqdn: Why are we removing this?

arya2: It was because this function's only called from the start command where we can expect Zebra to rename the directory before it's called, and there are a couple other conditions that need to be met for it to be reusable, but in hindsight, it's better to keep it anyways, in part so that nothing breaks if the function is ever used differently,

upbqdn: Thanks for the comment. I thought there was some deeper reason I didn't understand.

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -128,25 +120,63 @@ impl ZebraDb {
...
+                && newer_running_version.major == (older_disk_version.major + 1)
+                && RESTORABLE_DB_VERSIONS.contains(&newer_running_version.major) =>
+            {
+                warn!("upgrading database format to the next major version");
```
upbqdn: ###### Optional

I'd say this log entry is part of normal operation.

```suggestion
                info!("upgrading database format to the next major version");
```

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -128,25 +120,63 @@ impl ZebraDb {
...
+                let db = new_db(Version::new(older_disk_version.major, u64::MAX, u64::MAX));
+                let finalized_tip = db.finalized_tip_height();
+
+                // It's fine to mark the database format as upgraded later if the database is empty
+                if finalized_tip.is_some() {
+                    warn!("applying minor database format upgrades ahead of major format version bump, this may take a few minutes");
+                    let (_cancel_tx, cancel_rx) = std::sync::mpsc::channel();
```
upbqdn, line 161: I think not using the sender will make Zebra irresponsive to shutdowns when the upgrade is running.

zebra-state/src/service/finalized_state/zebra_db.rs
```diff
@@ -128,25 +120,63 @@ impl ZebraDb {
...
+        match format_change.clone() {
+            DbFormatChange::Upgrade {
+                older_disk_version,
+                newer_running_version,
+            } if !debug_skip_format_upgrades
+                && newer_running_version.major == (older_disk_version.major + 1)
+                && RESTORABLE_DB_VERSIONS.contains(&newer_running_version.major) =>
+            {
+                warn!("upgrading database format to the next major version");
+                let db = new_db(Version::new(older_disk_version.major, u64::MAX, u64::MAX));
```
upbqdn, line 155: Let's document why we're using `u64::MAX`.

zebra-state/src/service/finalized_state/zebra_db.rs
arya2: Something like:

```rust
std::thread::spawn(move || {
    while !is_shutting_down() {
        if cancel_tx.is_closed() {
            return;
        }

        std::thread::sleep(Duration::from_millis(1));
    }
    cancel_tx.send(CancelFormatChange);
});
```

might work here, but there's no `is_closed()` method on this channel, so we'd need to replace it with a watch channel first.

zebra-state/src/service/finalized_state/zebra_db.rs
arya2, line 155:
```suggestion
                // use the latest possible minor version to trigger all remaining format upgrades for the older major version
                let db = new_db(Version::new(older_disk_version.major, u64::MAX, u64::MAX));
```

mpguerra:
@arya2 do we still want to do this one soon or should we just close for now?

arya2:
> @arya2 do we still want to do this one soon or should we just close for now?

It's not a priority, let's close it for now.

PR #8762, opened 2024-08-14
fix(test): Renames `regtest_submit_blocks` test to avoid an RPC port conflict

arya2:
## Motivation

The `submit_block` test [failed here](https://github.com/ZcashFoundation/zebra/actions/runs/10379987688/job/28739428347#step:16:758), it happened [before the timeout](https://github.com/ZcashFoundation/zebra/actions/runs/10379987688/job/28739428347#step:16:342), so it must've panicked. The last log was the RPC listen address, and the test passed when it was run again, so it was likely due to a port conflict.

The `regtest_submit_blocks` test shouldn't be running there, and it could occasionally cause a port conflict since long-running tests like `submit_block` that use `TestType::zebrad_config()` are still using an unbound RPC port.

## Solution

Renames the `regtest_submit_blocks` test so it doesn't contain `"submit_block"`. 

### Tests

Only the `submit_block` test should run in that CI job in this PR.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -2957,7 +2957,7 @@ fn external_address() -> Result<()> {
 // TODO: Test this with an NU5 activation height too once config can be serialized.
 #[tokio::test]
 #[cfg(feature = "getblocktemplate-rpcs")]
-async fn regtest_submit_blocks() -> Result<()> {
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
Nice one. Thanks!

Issue #8730, opened 2024-07-30
Add lockbox information to `getblock` RPC method


oxarbitrage:
We need to add a lockbox value pool field to the `getblock` RPC method.

https://github.com/zcash/zcash/pull/6912/files#diff-decae4be02fb8a47ab4557fe74a9cb853bdfa3ec0fa1b515c0a1e5de91f4ad0bR276

This is the value going to the lockbox in the requested block.

Update: Zebra do not keep track of the needed value balabce amounts for each block but only for the tip. This tracking is needed to create the `valuePools` object efficiently, doing this is a lot more work than expected. More info in https://github.com/ZcashFoundation/zebra/pull/8765. In discussions with the zcashd team we concluded this will be nice to have but is not critical. I changed the labels and the milestone to reflect this decisions and closed the corresponding PR attempt.

mpguerra:
not a priority

PR #8765, opened 2024-08-14
rpc(getblock): Add value pools output

oxarbitrage:
## Motivation

We want the lockbox value pool information in the `getblock` output for a given block

The lockbox information was added in zcashd at https://github.com/zcash/zcash/pull/6912,  It is part of a list of pools that this PR adds to Zebra.

https://github.com/ZcashFoundation/zebra/issues/8730

https://github.com/zcash/zcash/pull/6912

## Solution

The solution implements the `valuePools` object and some of the fields. In zcashd this is like:

```
...
  "valuePools": [
    {
      "id": "transparent",
      "monitored": true,
      "chainValue": 0.03437500,
      "chainValueZat": 3437500,
      "valueDelta": 0.00625000,
      "valueDeltaZat": 625000
    },
 ...
 ```

Where `valueDelta` is the value pool for the requested block but `chainValue` is the value pool up to the requested blocks. Zebra do not keep track of value pools at specific heights for the finalized chain so i think we can't have that value. I am open to suggestions here. 

Solution implements `id`, `valueDelta` and `valueDeltaZat` for each pool including the lockbox

### Tests

Snapshot tests added by now.

### Follow-up Work

Need zcashd diff tests.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -10,6 +10,7 @@ use std::{collections::HashSet, fmt::Debug, sync::Arc};
 
 use chrono::Utc;
 use futures::{stream::FuturesOrdered, FutureExt, StreamExt, TryFutureExt};
+use get_block_template_rpcs::types::zec::Zec;
 use hex::{FromHex, ToHex};
 use indexmap::IndexMap;
 use jsonrpc_core::{self, BoxFuture, Error, ErrorCode, Result};
@@ -20,6 +21,7 @@ use tracing::Instrument;
 
 use zcash_primitives::consensus::Parameters;
 use zebra_chain::{
+    amount::{Amount, NegativeAllowed},
     block::{self, Height, SerializedBlock},
     chain_tip::ChainTip,
     parameters::{ConsensusBranchId, Network, NetworkUpgrade},
@@ -27,6 +29,7 @@ use zebra_chain::{
     subtree::NoteCommitmentSubtreeIndex,
     transaction::{self, SerializedTransaction, Transaction, UnminedTx},
     transparent::{self, Address},
+    value_balance::ValueBalance,
 };
 use zebra_node_services::mempool;
 use zebra_state::{HashOrHeight, MinedTx, OutputIndex, OutputLocation, TransactionLocation};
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: zebra-rpc/src/methods.rs [hunk 1]
```diff
@@ -1548,6 +1574,7 @@ impl Default for GetBlock {
             time: None,
             tx: Vec::new(),
             trees: GetBlockTrees::default(),
+            value_pools: vec![],
```
arya2, line 1577: Optional/Nitpick: This `Default` impl could be derived.

zebra-state/src/service.rs
```diff
@@ -1873,6 +1874,33 @@ impl Service<ReadRequest> for ReadStateService {
...
+                                read::block(
+                                    non_finalized_state.best_chain(),
+                                    &state.db,
+                                    hash_or_height,
+                                )
+                            },
+                        );
+
+                        let value_pool_change = block
+                            .map(|b| b.chain_value_pool_change(&HashMap::new(), None))
```
arya2, line 1893: In `zcashd`, this seems to be the total pool value balances. Zebra currently only tracks those for non-finalized and finalized chain tips, so we may need to update the database format again to avoid reading up to half the chain to compute these totals.

Update: We can leave this for now, the field names are distinct from those in zcashd, querying the value pool changes is still useful, and it'll be easier to update this code to return the chain value balance totals at a given height once Zebra begins tracking those than it would be to write it from scratch.

arya2 (commented):
This PR looks great, though it seems zcashd is returning the total chain value balances at the provided heights, so we may need another PR to track those in Zebra first.

oxarbitrage:
Closed as it is not critical and much more work than expected.

zebra-rpc/src/methods.rs
```diff
@@ -10,6 +10,7 @@ use std::{collections::HashSet, fmt::Debug, sync::Arc};
 
 use chrono::Utc;
 use futures::{stream::FuturesOrdered, FutureExt, StreamExt, TryFutureExt};
+use get_block_template_rpcs::types::zec::Zec;
```
arya2, line 13: This type needs to be moved out from behind the `getblocktemplate-rpcs` feature flag.

arya2:
This looks good to merge once the `Zec` type has been moved.

PR #8769, opened 2024-08-16
change(rpc): Add value pool balances to `getblockchaininfo` RPC method response

### LINKED ISSUE #8767: Add `valuePools` object to `getblockchaininfo` RPC
In https://github.com/ZcashFoundation/zebra/pull/8765 we tried to add `valuePools` object to the `getblock` RPC method but we realized we can't do it without changing the database which we prefer to defer. https://github.com/ZcashFoundation/zebra/issues/8730

However, the [getblockchaininfo](https://zcash.github.io/rpc/getblockchaininfo.html) RPC method provides value pool information for the tip, something that we can provide using similar code introduced in https://github.com/ZcashFoundation/zebra/pull/8765

arya2:
## Motivation

We want to add value pool balances at the chain tip to Zebra's `getblockchaininfo` RPC method response.

Closes #8767.

https://zcash.github.io/rpc/getblockchaininfo.html

## Solution

- Moves the `Zec` type in `zebra-rpc` out from behind the `getblocktemplate-rpcs` feature flag.
- Adds a `ValuePoolBalance` type to a new `types::get_blockchain_info` module with a fn for creating one from a `ValueBalance`.
- Updates the `getblockchaininfo` RPC method return type to a `BoxFuture` to make async calls to the state service.
- Adds a new state service request for getting the tip height, tip hash, and the value balance.
  - Checks that the tip hasn't changed while reading the value balance from the finalized state. This could cause the RPC method to return an error after a few retries, which could be avoided by adding the value balance information to the `ChainTipBlock` instead of adding a new state service request, but in practice, the tip and value balance should be in-memory when there is no non-finalized best chain and blocks are being committed rapidly, so it should be unlikely to return that error. Still, let me know if we want to add it to the `ChainTipBlock` instead, we could do that here or in a follow-up PR.
- Updates the `getblockchaininfo` method to get the chain tip height, hash, value balance, and block time from state requests.
- Updates snapshots

Related Changes:
- Minor refactor of `estimated_height` assignment
- Uses generic error construction in the last couple places where possible and removes the outdated TODOs

### Tests

Updates existing snapshots (and some prop tests to handle the `BoxFuture` wrapped return type).

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/methods/types/get_blockchain_info.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use super::*;
+
+/// A value pool's balance in Zec and Zatoshis
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ValuePoolBalance {
+    /// Name of the pool
+    id: String,
+    /// Total amount in the pool, in ZEC
```
arya2, line 16: Do we want to include `monitored` here? It would always be true in Zebra.

oxarbitrage: Yea, let's don't add it by now.

oxarbitrage (dismissed):
Looks really good to me, great work, thanks!

oxarbitrage:
It seems the tests are hang somewhere so the CI is giving up. https://github.com/ZcashFoundation/zebra/actions/runs/10414872254/job/28844504197?pr=8769

I am having the same issue locally when running the suite:

```
cargo test
```

arya2:
> It seems the tests are hang somewhere so the CI is giving up

Ah, yes, I forgot to handle the mock service requests in the prop tests.

oxarbitrage (approved):

PR #8788, opened 2024-08-20
fix(rpc): Return verification errors from `sendrawtransaction` RPC method

### LINKED ISSUE #8778: Return verification errors from `sendrawtransaction` RPC method
## Motivation

Zebra's `sendrawtransaction` method always returns a txid for submitted txs even if they end up not being accepted. 

We want to return any transaction verification errors to the client, instead of just confirming that a transaction has been queued for verification

## Possible Design

We could add a new mempool request to verify transactions immediately when received from the RPC server. This could block other calls to the mempool service, so we likely want to continue using the existing request for verifying gossiped transactions.

arya2:
## Motivation

We want to return an error from the `sendrawtransaction` RPC method if a transaction cannot be verified and added to the mempool.

Closes #8778.

## Solution

Adds a `QueueRpc` mempool request to wait for a transaction verification result and return an error message to the client if the transaction was not verified and added to Zebra's mempool.

### Tests

This PR still needs tests.

### Follow-up Work

- @hhanh00 reported that Zebra was failing to add transactions to its mempool or return an error describing why transactions weren't added to the mempool, we want to return a specific error so we can fix the issue.
- We likely need an update to lightwalletd to handle these errors correctly.
- Handle verification errors in inbound service by incrementing peer connection "bad score" when tracking misbehaving nodes

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-node-services/src/mempool.rs
+++ b/zebra-node-services/src/mempool.rs
@@ -4,6 +4,7 @@
 
 use std::collections::HashSet;
 
+use tokio::sync::oneshot;
 use zebra_chain::transaction::{self, UnminedTx, UnminedTxId};
 
 #[cfg(feature = "getblocktemplate-rpcs")]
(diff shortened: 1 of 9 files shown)


zebra-node-services/src/mempool.rs
```diff
@@ -58,6 +58,11 @@ pub enum Request {
     /// The transaction downloader checks for duplicates across IDs and transactions.
     Queue(Vec<Gossip>),
 
+    /// Queue a list of unmined transactions submitted via the `sendrawtransaction` RPC method,
+    /// and wait for them to be verified, rejected, or for their verification task to be cancelled if
+    /// the transaction id is mined into the best chain.
+    QueueRpc(Vec<UnminedTx>),
```
arya2: It's probably best to return the verification error for all `Queue` requests, (or the `oneshot::Receiver`s so the inbound service can still respond to peers quickly) so it's easier to link transactions to peers for scoring misbehaving peers. Typed errors will also be helpful there, so it would be better to move `TransactionDownloadVerifyError` to `zebra-node-services` than to revert the changes for making it clone-able.

Update: Addressed in https://github.com/ZcashFoundation/zebra/pull/8788/commits/acdd8dcb0b2927f8212b03d8186db9bf472b68ec and https://github.com/ZcashFoundation/zebra/pull/8788/commits/4bbac10ef5992bc039a2867be2bbacf539456f1d.

upbqdn (approved):

PR #8792, opened 2024-08-21
build(deps): make librustzcash crates workspace dependencies

### LINKED ISSUE #8791: Move common dependencies to workspace dependencies
## Motivation

We want to reduce repetitive dependency definitions in our crates so changes are easier to review and so that it's harder to use different versions of dependencies in different crates.

tomekpiotrowski:
## Motivation

Right now versions of some of the dependencies, particularly `librustzcash` crates and some ECC dependencies, are repeated in multiple `Cargo.toml` files which makes those versions difficult to manage.

Fixes #8791

## Solution

This PR introduces workspace dependencies to the workspace `Cargo.toml` file. Right now the `librustzcash` dependencies point to the `main` branch of the zcash/librustzcash repo as that was the newest version used.

This also removes all renames of `sapling-crypto` to `sapling` as package renames are not currently handled by workspace dependencies. In the code `use sapling::` was substituted with `use sapling_crypto::`.

### Tests

This PR passes all `cargo test` tests.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,20 @@ resolver = "2"
 
 # `cargo release` settings
 
+[workspace.dependencies]
+incrementalmerkletree = "0.6.0"
+orchard = "0.9.0"
+sapling-crypto = "0.2.0"
+# TODO: Revert to a release once librustzcash is released (#8749).
+zcash_address = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_history = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash/", branch = "main" }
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -21,6 +21,20 @@ resolver = "2"
 
 # `cargo release` settings
 
+[workspace.dependencies]
+incrementalmerkletree = "0.6.0"
+orchard = "0.9.0"
+sapling-crypto = "0.2.0"
```
arya2, line 27: These dependency versions have been bumped, which seems fine if the tests pass, we trust the ECC dependencies.

arya2 (commented):
Looks good to me! Thank you for making this change.

It looks like the `cargo-deny` check is failing due to duplicate versions of some packages being used now, see https://zebra.zfnd.org/dev/continuous-integration.html#fixing-duplicate-dependencies-in-check-denytoml-bans for fixing that.

Cargo.toml
tomekpiotrowski: Ah, yes. Should've mentioned that in the PR description.

zebra-scan/src/tests.rs
```diff
@@ -1,4 +1,3 @@
-//! Test that we can scan the Zebra blockchain using the external `zcash_client_backend` crate
```
arya2, line 1: Let's revert this change.

arya2 (approved):
This looks good, thank you!

A line in the zebra-scan test module docs was deleted that we want to keep, but the rest of the changes in this PR will be useful for updating our librustzcash dependencies quickly before the release, so we can revert that in a follow-up PR.

PR #8794, opened 2024-08-21
fix(docs): sample configurations in the custom testnet book page

oxarbitrage:
## Motivation

I tried to launch some custom testnets and i found a few small issues in the docs, with the sample configurations provided. 

## Solution

Small modifications so the provided configs can be just pasted into a file and used to launch zebra. Initial peers field was added, with an empty list or:

```
error: zebrad fatal error: parse error: cannot use default initials peers with incompatible testnet for key `network` at line 44 column 1
```
### Tests

Manual, both configurations now are valid and start zebrad.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2 (approved):
Thank you!!

PR #8796, opened 2024-08-22
ref(docker): use cache mounts for build cache

### LINKED ISSUE #6169: docker: improve Rust dependency caching in CI to avoid rebuilding

gustavovalverde:
## Motivation

We've been using external tools and custom approaches to improve the caching layers in Rust, and avoid invalidating the caches if no Rust-code is changed.

This update eliminates the need for external tools like cargo-chef to leverage caching layers, resulting in an average build time reduction of 4m30s (~36% improvement).

While this solution doesn't fully resolve the issues mentioned in https://github.com/ZcashFoundation/zebra/issues/6169#issuecomment-1712776391, it represents the best possible approach without resorting to custom solutions, which we'd prefer to avoid.

Depends-On: #8817
Closes #6169

- https://docs.docker.com/language/rust/develop/

## Solution

- Use cache mounts instead of `cargo-chef`
- Remove custom `rsync` steps

### Others
- Add a `RUST_VERSION` arg so we don't invalidate our cache so frequently with Rust upgrades
- Use `--link` in our `runtime` stage, to reduce the resulting artifact
- Remove the installation of Google OS Config agent, as this is no longer required

### Tests
- Tests are passing for CI and CD deployments
- Rebuilding is taking less time (**~8 minutes** instead of **~13 minutes**)

### Follow-up Work
- Do not run our `runtime` stage with the root user
- Add SBOM and provenance attestations https://docs.docker.com/build/ci/github-actions/attestations/ 

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


gustavovalverde:
@arya2 this one is ready now

mpguerra:
Do we want to close #6169 with this PR or are we planning on making any other improvements in the short term?

gustavovalverde:
I've just linked #6169, for automatic closure.

This is not solving the underlying issue, but this is as far as we can get right now, until the `mtime` issue is solved in Rust.

docker/Dockerfile
```diff
@@ -224,14 +209,15 @@ ARG FEATURES
 ENV FEATURES=${FEATURES}
 
 # Path and name of the config file
+# This are set here to always this variable set, even if the user does not set it
```
arya2: Nitpick:

```suggestion
# These are set to a default value when not defined in the environment
```

arya2 (dismissed):
This looks good!

mpguerra (approved):

PR #8803, opened 2024-08-26
fix(docker): do not run the Zebra nodes with the root user

gustavovalverde:
## Motivation

We should not run our nodes with a privileged user inside the container. If an attacker manages to break out of the container, they will have root access to the host.

- https://docs.docker.com/build/building/best-practices/#user
- https://www.docker.com/blog/understanding-the-docker-user-instruction/

## Solution

- Create an user to run `zebrad`

### Tests

- CI and CD tests should pass

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


oxarbitrage (approved):
Looks good, thanks for the fix!

PR #8826, opened 2024-09-02
add(network): Check in proptest seeds for `zebra-network` tests

upbqdn:
## Motivation

I started using `cargo fix` more regularly, and running it without any flags returns:

```log
error: the working directory of this package has uncommitted changes, and `cargo fix` can potentially perform destructive changes; if you'd like to suppress this error pass `--allow-dirty`, `--allow-staged`, or commit the changes to these files:

  * zebra-network/proptest-regressions/peer_set/set/ (dirty)
```

I've had the proptest file dangling among unstaged changes in git's workdir for a while. I assume it gets generated for others as well. Having it checked in is recommended in the file itself.

## Solution

- Track the file mentioned above in git.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
+++ b/zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
@@ -1,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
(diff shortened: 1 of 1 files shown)


arya2:
I think we want to keep running the network prop tests without this file because they depend on the network as well as code changes.

We may want to ask the technical advisory board about this one.

PR #8839, opened 2024-09-04
feat(rpc): Add a `stop` rpc method

### LINKED ISSUE #8836: Implement `stop` RPC method
We want to make the zcashd rpc tests work with Zebrad as the backend.
https://github.com/ZcashFoundation/zebra/issues/8779

In those tests nodes are stopped by a `stop` RPC method: https://zcash.github.io/rpc/stop.html

We need to implement it for tests to work, we might want to make it regtest only similar to https://github.com/ZcashFoundation/zebra/issues/8834 even if that does not seems to be how it is implemented in zcashd:

https://github.com/zcash/zcash/blob/master/src/rpc/server.cpp#L281

oxarbitrage:
## Motivation

Close https://github.com/ZcashFoundation/zebra/issues/8836

## Solution

Add an rpc that terminate the running process. Only work if the network is the regtest.

### Tests

Manually tested.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -301,6 +301,18 @@ pub trait Rpc {
         &self,
         address_strings: AddressStrings,
     ) -> BoxFuture<Result<Vec<GetAddressUtxos>>>;
+
+    #[rpc(name = "stop")]
+    /// Stop the running zebrad process.
+    ///
+    /// # Notes
+    ///
+    /// Only works if the network of the running zebrad process is `Regtest`.
+    ///
+    /// zcashd reference: [`stop`](https://zcash.github.io/rpc/stop.html)
+    /// method: post
+    /// tags: control
+    fn stop(&self) -> Result<()>;
 }
 
 /// RPC method implementations.
@@ -1343,6 +1355,18 @@ where
             Ok(response_utxos)
         }
         .boxed()
+    }
+
+    fn stop(&self) -> Result<()> {
+        if self.network.is_regtest() {
+            std::process::exit(0);
+        } else {
+            Err(Error {
+                code: ErrorCode::MethodNotFound,
+                message: "stop is only available on regtest networks".to_string(),
+                data: None,
+            })
+        }
     }
 }
 
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -1359,6 +1359,7 @@ where
 
     fn stop(&self) -> Result<()> {
         if self.network.is_regtest() {
+            // TODO: Use a controlled shutdown.
             std::process::exit(0);
         } else {
             Err(Error {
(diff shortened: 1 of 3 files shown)


zebra-rpc/src/methods.rs
```diff
@@ -1344,6 +1356,18 @@ where
         }
         .boxed()
     }
+
+    fn stop(&self) -> Result<()> {
+        if self.network.is_regtest() {
+            std::process::exit(0);
```
upbqdn, line 1363: This function doesn't provide a controlled shutdown. We implement a graceful shutdown here: https://github.com/zcashfoundation/zebra/blob/d9fca01d4a5993391dfa591f179bf2217a6a2c75/zebrad/src/application.rs#L513. I think it would be safer to use it instead of direct termination. `zcashd` also uses a controlled shutdown when calling the `stop` RPC.

upbqdn (dismissed):

zebra-rpc/src/methods.rs
oxarbitrage: You're right; I considered that as well. However, we don't have access to the application through the RPC methods, so while it's definitely possible, it’s not as straightforward to implement. Since this is currently only for the regtest environment, I believe the direct termination approach should suffice for now. We can implement a proper solution once we remove the regtest limitation. I added a TODO comment to do that.

upbqdn: Sounds good.

upbqdn: Instead of having a handle to `zebrad`, we could send a signal to the process itself.

- https://docs.rs/nix/0.29.0/nix/sys/signal/fn.raise.html

zebra-rpc/src/methods.rs
```diff
@@ -1344,6 +1356,19 @@ where
         }
         .boxed()
     }
+
+    fn stop(&self) -> Result<()> {
+        if self.network.is_regtest() {
+            // TODO: Use a controlled shutdown.
```
upbqdn:
```suggestion
            // TODO: Use graceful termination in `stop` RPC (#8850)
```

upbqdn (dismissed):

zebra-rpc/src/methods.rs
arya2: Could we either use a signal or call [`set_shutting_down()`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/shutdown.rs#L27) and check for it in [`shutdown()`](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/components/tokio.rs#L96)?

upbqdn (approved):

oxarbitrage:
@mergifyio refresh

PR #8857, opened 2024-09-09
add(mempool): Verify transactions with unmined inputs in the mempool

### LINKED ISSUE #8777: feature: support transactions with unmined inputs in the mempool
### Motivation

We don't allow txs with unmined inputs in the mempool. However, supporting them can be useful, in particular for the tex address support which creates two transactions, one z->t and another t->t and submits them sequentially. Currently the second one is rejected in zebrad because of the unknown input since the first one wasn't mined yet.

### Specifications

No spec for the mempool :smiling_face_with_tear: 

We need to consider that this changes the mempool criteria from "any txs that could be mined on their own" to "a collection of txs that possibly need to be mined together" and check if this has unintended consequences. (In particular I think this will stop https://github.com/ZcashFoundation/zebra/issues/5674 from being implementable)

### Testing

Probably trying to submit two txs where the second spends something from the first

### Related Work

See discussion in https://discord.com/channels/809218587167293450/809251029673312267/1275193366740602993

arya2:
## Motivation

We want to verify transactions in the mempool that spend the transparent outputs of other transactions in the mempool.

Closes #8777.

## Solution

- Adds a handle to the mempool service in the transaction verifier
- Adds `AwaitOutput` request in the mempool service
- Updates the `spent_utxos()` method in the transaction verifier to try querying the mempool for spent outpoints before returning a `TransparentInputNotFound` error.
- Polls the mempool service from the transaction verifier after verifying a mempool transaction that creates new UTXOs (in case they're spent by other mempool transactions that are still being verified)
- Updates `getblocktemplate` method and ZIP-317 transaction selection to include transactions that spent the outputs of other transactions in the mempool when all of their dependencies have been selected

### Tests

Added tests to check that:
- Inserting a transaction also updates the transaction dependencies tracked in the verified set
- The `Storage.remove()` method works as expected (the rest of `evict_one()` should be covered already)
- The mempool responds to `AwaitOutput` requests when the output for the queried outpoint is:
  - Added to the mempool
  - Already available in the mempool
- The transaction verifier:
  - Calls the mempool with `AwaitOutput` requests as expected and processes `UnspentOutput` responses correctly
  - Polls the mempool after verifying a mempool transaction that creates new transparent outputs
  - Includes all of the expected `spent_mempool_outpoints` in its response
- Long poll ids remaining consistent as long the mempool contents are unchanged (and everything else is unchanged)
- ZIP-317 mempool transaction selection excludes transactions that have dependencies unless their dependencies have already been selected

If needed, this PR could also add tests for:
- ZIP-317 mempool transaction selection always returns the right order of transactions (this could be unnecessary since they’re sorted afterwards anyway but it could be a good sanity check)
- The `getblocktemplate` method returns transactions with dependencies in the right order when the transaction hashes would have them sorted in the wrong order
- Ideally at least a basic acceptance test that checks that it all works via RPCs

## Follow Up Work

The mempool's `Downloads` sends a result to the response channel too early, it should wait until the transaction is being inserted into the mempool's verified set.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -188,23 +255,90 @@ fn checked_add_transaction_weighted_random(
...
+        selected_txs.push((0, candidate_tx));
+
+        for dependent_candidate_tx_ids_by_level in
+            mempool_tx_deps.all_dependents_leveled(&selected_tx_id)
+        {
+            for dependent_candidate_tx_id in &dependent_candidate_tx_ids_by_level {
+                let mempool_tx_deps = mempool_tx_deps.dependencies();
+                let candidate_tx_deps = mempool_tx_deps.get(dependent_candidate_tx_id);
+
+                if has_direct_dependencies(candidate_tx_deps, selected_txs) {
```
arya2: ~If anyone has any ideas on how to address this, there's an issue here that I'm not sure how to fix this yet where tx3 won't be selected if tx1 is independent, and tx2 depends on tx1, but tx3 depends on both tx1 and tx2, because it's checked before tx2 is selected.~

~I thought about storing those in another collection to try selecting them after the rest of the dependent txs, but that still won't work if there's a tx4 that depends on tx1 and tx3.~

Update: There's a simple fix here. I think it may have been a non-issue to begin with, but the new code looks simpler and more efficient anyways.

zebra-node-services/src/mempool.rs
```diff
@@ -14,6 +17,12 @@ use crate::BoxError;
 
 mod gossip;
 
+#[cfg(feature = "getblocktemplate-rpcs")]
+mod transaction_dependencies;
+
+#[cfg(feature = "getblocktemplate-rpcs")]
+pub use transaction_dependencies::TransactionDependencies;
```
arya2:
```suggestion
mod transaction_dependencies;

pub use transaction_dependencies::TransactionDependencies;
```

zebra-consensus/src/transaction.rs
```diff
@@ -473,7 +547,17 @@ where
...
-                    Response::Mempool { transaction }
+
+                    if let Some(mut mempool) = mempool {
+                        if !transaction.transaction.transaction.outputs().is_empty() {
+                            tokio::spawn(async move {
+                                tokio::time::sleep(POLL_MEMPOOL_DELAY).await;
+                                mempool.ready().await.expect("mempool poll_ready() method should not return an error");
+                            });
+                        }
+                    }
```
arya2: This should be moved to the mempool's `Downloads` component, see [this comment](https://github.com/ZcashFoundation/zebra/pull/8901/files#r1794389135).

The changes in this PR seem excessively substantial already, so I'll do this in a follow up PR.

mpguerra:
@upbqdn can you please review this one?

zebrad/src/components/mempool/storage.rs
```diff
@@ -116,12 +127,15 @@ pub struct Storage {
...
+    /// The set of outpoints with pending requests for their associated transparent::Output.
+    pub(super) pending_outputs: PendingOutputs,
+
     /// The set of transactions rejected due to bad authorizations, or for other
     /// reasons, and their rejection reasons. These rejections only apply to the
     /// current tip.
     ///
     /// Only transactions with the exact [`UnminedTxId`] are invalid.
-    tip_rejected_exact: HashMap<UnminedTxId, ExactTipRejectionError>,
+    tip_rejected_exact: HashMap<transaction::Hash, ExactTipRejectionError>,
```
conradoplg: What is the reason for this change? The semantics of `tip_rejected_exact` is for it to incude transactions that have failed to validate e.g. due to an invalid signature. If this is changed to a Hash, then a transaction with e.g. the same data but a valid signature would be rejected when it shouldn't.

zebrad/src/components/mempool/storage.rs
```diff
@@ -180,11 +195,16 @@ impl Storage {
     /// If inserting this transaction evicts other transactions, they will be tracked
     /// as [`SameEffectsChainRejectionError::RandomlyEvicted`].
     #[allow(clippy::unwrap_in_result)]
-    pub fn insert(&mut self, tx: VerifiedUnminedTx) -> Result<UnminedTxId, MempoolError> {
+    pub fn insert(
+        &mut self,
+        tx: VerifiedUnminedTx,
+        spent_mempool_outpoints: Vec<transparent::OutPoint>,
```
conradoplg, line 205: Please document the new argument

conradoplg (commented):
I'm still reviewing but I added a comment

zebrad/src/components/mempool/storage.rs
arya2: Thank you for catching that, documented in https://github.com/ZcashFoundation/zebra/pull/8857/commits/cc76b40885215fa1903ad26349583ac863f22f54.

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -126,15 +143,35 @@ impl VerifiedSet {
...
+        spent_mempool_outpoints: Vec<transparent::OutPoint>,
+        pending_outputs: &mut PendingOutputs,
     ) -> Result<(), SameEffectsTipRejectionError> {
         if self.has_spend_conflicts(&transaction.transaction) {
             return Err(SameEffectsTipRejectionError::SpendConflict);
         }
 
-        self.cache_outputs_from(&transaction.transaction.transaction);
+        // This likely only needs to check that the transaction hash of the outpoint is still in the mempool,
+        // bu it's likely rare that a transaction spends multiple transparent outputs of
```
arya2:
```suggestion
        // but it's likely rare that a transaction spends multiple transparent outputs of
```

zebrad/src/components/mempool/storage.rs
arya2: > What is the reason for this change?

I think the broader change to mined ids was to avoid iterating through the verified set to find the auth digest when adding transaction dependencies (since `OutPoint` only includes the mined id).

The change to `tip_rejected_exact` was a mistake, it would work fine for the `sendrawtransaction` RPC, but I see it would incorrectly reject gossiped transaction ids when it should try to download and verify them. 

Restored its type to the full `UnminedTxId` in https://github.com/ZcashFoundation/zebra/pull/8857/commits/ce6d169a9a0a1848b1d50a9eb29b7710f48043da.

conradoplg (dismissed):
This is a complex PR (not your fault, it's a complex issue). I've been reviewing it for a while but it feels I could spend ages reviewing this to be 100% sure. I'm OK with approving it but do you think we should ask for someone else to take a look?

Also, did you manage to test it for the use case we wish to address (i.e. sending to tex addresses?)

arya2:
> This is a complex PR (not your fault, it's a complex issue). I've been reviewing it for a while but it feels I could spend ages reviewing this to be 100% sure. I'm OK with approving it but do you think we should ask for someone else to take a look?

In hindsight, it probably should've been split into multiple PRs. A second review would be ideal since it involves concurrency and consensus changes, though it has also gone through a few rounds of self-review.

I think a second review of just the changes to ZIP 317 block production would be okay.

> Also, did you manage to test it for the use case we wish to address (i.e. sending to tex addresses?)

I haven't, and there's no acceptance test that tries to submit an orphaned transaction, I'll try doing that now.

upbqdn:
I'm also still reviewing. A bunch of other things took precedence in the meantime, unfortunately.

zebra-consensus/src/transaction.rs
```diff
@@ -52,31 +60,79 @@ mod tests;
...
     ZS::Future: Send + 'static,
+    Mempool: Service<mempool::Request, Response = mempool::Response, Error = BoxError>
+        + Send
+        + Clone
+        + 'static,
+    Mempool::Future: Send + 'static,
 {
     /// Create a new transaction verifier.
-    pub fn new(network: &Network, state: ZS) -> Self {
+    pub fn new(network: &Network, state: ZS, mempool_setup_rx: oneshot::Receiver<Mempool>) -> Self {
```
upbqdn, line 107: Can we not go without a channel?

zebra-consensus/src/transaction.rs
```diff
@@ -523,19 +617,25 @@ where
     /// in the same order as the matching inputs in the transaction.
     async fn spent_utxos(
```
upbqdn, line 621: We need to update the fn docs.

zebra-consensus/src/transaction.rs
arya2: It needs some way to get it concurrently since the mempool also has a handle to the transaction verifier. It doesn't have to be a channel, but a channel seemed like the simplest option and consistent with how the inbound service is set up.

zebra-consensus/src/transaction.rs
```diff
@@ -473,7 +547,22 @@ where
...
+
+                    if let Some(mut mempool) = mempool {
+                        if !transaction.transaction.transaction.outputs().is_empty() {
+                            tokio::spawn(async move {
+                                tokio::time::sleep(POLL_MEMPOOL_DELAY).await;
+                                let _ = mempool
+                                    .ready()
+                                    .await
+                                    .expect("mempool poll_ready() method should not return an error")
+                                    .call(mempool::Request::CheckForVerifiedTransactions)
```
upbqdn, line 559: Why do we need to do this call here?

upbqdn (dismissed):
What do you mean by "orphaned" transactions in the title?

zebra-consensus/src/transaction.rs
```diff
@@ -570,7 +676,41 @@ where
                 continue;
             }
         }
-        Ok((spent_utxos, spent_outputs))
+
+        if let Some(mempool) = mempool {
```
upbqdn, line 683: Edit: disregard this comment.

This new implementation changes the lookup logic. 

In the old implementation, if `is_mempool` was set, and `UnspentBestChainUtxo` returned `None`, the fn returned `TransactionError::TransparentInputNotFound`.

In the new implementation, if the same conditions hold, and `mempool` is `None`, no error is returned.

upbqdn: Oh no, the above is wrong, closing.

zebra-consensus/src/transaction.rs
arya2: The docs for that variant say there are issues with polling a tower service with a `Buffer` layer without following that with a call.

arya2:
> What do you mean by "orphaned" transactions in the title?

Transactions that spend outputs that are missing in the chain state, though I'm not sure if "orphan transactions" is the right term if those outputs can still be found in the mempool.

This PR is still missing an acceptance test for the use case we're trying to address (testing it manually with zcashd seems difficult, and a test that creates transactions could provide a valuable reference for other tests in the future), but I think it's fine to merge it without an acceptance test since each component has unit tests (though I'm happy to prioritize writing that test if anyone disagrees).

zebra-consensus/src/transaction.rs
arya2: Updated in https://github.com/ZcashFoundation/zebra/pull/8857/commits/257567b559dba0359519d61684e80daac52cc434.

upbqdn: I was referring to:

- The description of `known_utxos`, since the fn doesn't take that anymore. We could move that comment somewhere else.
- The returned values, since the fn doesn't return a tuple anymore.

zebra-consensus/src/transaction.rs
```diff
@@ -514,28 +608,38 @@ where
         }
     }
 
-    /// Wait for the UTXOs that are being spent by the given transaction.
+    /// Waits for the UTXOs that are being spent by the given transaction to arrive in
+    /// the state for [`Block`](Request::Block) requests.
     ///
     /// `known_utxos` are additional UTXOs known at the time of validation (i.e.
     /// from previous transactions in the block).
```
upbqdn: ###### Optional

```suggestion
```

zebra-consensus/src/transaction.rs
```diff
@@ -514,28 +608,38 @@ where
...
     ) -> Result<
         (
             HashMap<transparent::OutPoint, transparent::Utxo>,
             Vec<transparent::Output>,
+            Vec<transparent::OutPoint>,
         ),
         TransactionError,
     > {
+        let is_mempool = req.is_mempool();
+        let known_utxos = req.known_utxos();
```
upbqdn, line 637: ###### Optional

```suggestion
        // Additional UTXOs known at the time of validation,
        // i.e., from previous transactions in the block.
        let known_utxos = req.known_utxos();
```

zebra-consensus/src/transaction.rs
```diff
@@ -514,28 +608,38 @@ where
...
+    /// the state for [`Block`](Request::Block) requests.
     ///
     /// `known_utxos` are additional UTXOs known at the time of validation (i.e.
     /// from previous transactions in the block).
     ///
+    /// Looks up UTXOs that are being spent by the given transaction in the state or waits
+    /// for them to be added to the mempool for [`Mempool`](Request::Mempool) requests.
+    ///
     /// Returns a tuple with a OutPoint -> Utxo map, and a vector of Outputs
     /// in the same order as the matching inputs in the transaction.
```
upbqdn: ###### Optional

```suggestion
    /// Returns a triple containing:
    /// - `OutPoint` -> `Utxo` map, 
    /// - vec of `Output`s in the same order as the matching inputs in the `tx`,
    /// - vec of `Outpoint`s spent by a mempool `tx` that were not found in the best chain's utxo set.
```

upbqdn:
> > What do you mean by "orphaned" transactions in the title?

> Transactions that spend outputs that are missing in the chain state, though I'm not sure if "orphan transactions" is the right term if those outputs can still be found in the mempool.

This RFC, https://github.com/zcashfoundation/zebra/blob/257567b559dba0359519d61684e80daac52cc434/book/src/dev/rfcs/0005-state-updates.md#L62, defines orphaned blocks as those that are no longer included in the best chain, which implies they were mined at some point. I'd assume orphaned txs are those in such blocks, but this PR doesn't apply to them. Maybe something like "Verify transactions spending mempool UTXOs"?

zebrad/src/components/mempool/pending_outputs.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use tokio::sync::broadcast;
+
+use tower::BoxError;
+use zebra_chain::transparent;
+
+use zebra_node_services::mempool::Response;
+
+#[derive(Debug, Default)]
+pub struct PendingOutputs(HashMap<transparent::OutPoint, broadcast::Sender<transparent::Output>>);
```
upbqdn, line 15: Missing docs

zebra-node-services/src/mempool/transaction_dependencies.rs
```diff
@@ -0,0 +1,124 @@
+//! Representation of mempool transactions' dependencies on other transactions in the mempool.
+
+use std::collections::{HashMap, HashSet};
+
+use zebra_chain::{transaction, transparent};
+
+/// Representation of mempool transactions' dependencies on other transactions in the mempool.
+#[derive(Default, Debug, Clone)]
+pub struct TransactionDependencies {
+    /// Lists of mempool transactions that create UTXOs spent by
```
upbqdn:
```suggestion
    /// Lists of mempool transaction ids that create UTXOs spent by
```

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -238,27 +243,38 @@ impl GetBlockTemplate {
...
             // `zcashd` sometimes seems to do this, but other times the order is arbitrary.
-            mempool_txs_with_templates.sort_by_key(|(tx_template, _tx)| tx_template.data.clone());
+            mempool_txs_with_templates.sort_by_key(|(min_tx_index, tx_template, _tx)| {
+                (*min_tx_index, tx_template.data.clone())
+            });
         } else {
             // Sort by hash, this is faster.
-            mempool_txs_with_templates
-                .sort_by_key(|(tx_template, _tx)| tx_template.hash.bytes_in_display_order());
+            mempool_txs_with_templates.sort_by_key(|(min_tx_index, tx_template, _tx)| {
```
upbqdn: Do we need to sort at all?

arya2: Probably not, I think it was there to make comparisons with the zcashd getblocktemplate method easier. I wouldn't mind removing it.

upbqdn: I'm in favor of removing it as well then.

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -158,6 +178,52 @@ fn setup_fee_weighted_index(transactions: &[VerifiedUnminedTx]) -> Option<Weight
...
+///
+/// Requires items in `selected_txs` to be unique to work correctly.
+fn has_direct_dependencies(
+    candidate_tx_deps: Option<&HashSet<transaction::Hash>>,
+    selected_txs: &Vec<(InBlockTxDependenciesDepth, VerifiedUnminedTx)>,
+) -> bool {
+    let Some(deps) = candidate_tx_deps else {
+        return true;
+    };
+
```
upbqdn, line 200: ###### Optional

```suggestion

if selected_txs.len() < deps.len() {
    return false;
}

```

zebra-rpc/src/methods/get_block_template_rpcs/zip317.rs
```diff
@@ -181,30 +251,110 @@ fn checked_add_transaction_weighted_random(
...
+    selected_txs.push((0, candidate_tx));
+
+    // Try adding any dependent transactions if all of their dependencies have been selected.
+
+    let mut current_level_dependents = mempool_tx_deps.direct_dependents(selected_tx_id);
+    while !current_level_dependents.is_empty() {
+        let mut next_level_dependents = HashSet::new();
+
+        for dependent_tx_id in &current_level_dependents {
+            if has_direct_dependencies(tx_dependencies.get(dependent_tx_id), selected_txs) {
```
upbqdn, line 304:
```suggestion
            // ## Note
            //
            // A necessary condition for adding the dependent tx is that it spends unmined outputs coming only from
            // the selected txs, which come from the mempool. If the tx also spends in-chain outputs, it won't
            // be added. This behavior is not specified by consensus rules and can be changed at any time,
            // meaning that such txs could be added.
            if has_direct_dependencies(tx_dependencies.get(dependent_tx_id), selected_txs) {
```

zebrad/src/components/mempool/tests/vector.rs
```diff
@@ -956,6 +1083,8 @@ async fn setup(
         chain_tip_change.clone(),
     );
 
+    tokio::spawn(async move { while mempool_transaction_receiver.recv().await.is_ok() {} });
```
upbqdn, line 1086: Why do we need this?

arya2: Dropping the receiver at the end of the `setup()` function could cause the mempool's `poll_ready()` method to return an error when trying to send transaction ids to a closed channel.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: I've limited it to sorting only in test compilations in https://github.com/ZcashFoundation/zebra/pull/8857/commits/42dfc394501fa2a48dc2243682d94fadb37720b1 (it's still needed for the snapshot tests but I'd prefer to keep it out of release compilations).

zebrad/src/components/mempool/pending_outputs.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/pull/8857/commits/5043adf9e97e1e5e872ed77197b27e454f17efa5.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
```diff
@@ -237,28 +243,45 @@ impl GetBlockTemplate {
...
+        let (mempool_tx_templates, mempool_txs): (Vec<_>, Vec<_>) = {
+            let mut mempool_txs_with_templates: Vec<(
+                InBlockTxDependenciesDepth,
+                TransactionTemplate<amount::NonNegative>,
+                VerifiedUnminedTx,
+            )> = mempool_txs
+                .into_iter()
+                .map(|(min_tx_index, tx)| (min_tx_index, (&tx).into(), tx))
+                .collect();
+            #[cfg(test)]
```
upbqdn:
```suggestion
```

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -161,49 +198,51 @@ impl VerifiedSet {
     ///
     /// [ZIP-401]: https://zips.z.cash/zip-0401
     #[allow(clippy::unwrap_in_result)]
-    pub fn evict_one(&mut self) -> Option<VerifiedUnminedTx> {
+    pub fn evict_one(&mut self) -> Vec<VerifiedUnminedTx> {
```
upbqdn: Returning a vec doesn't semantically correspond to the name of the fn and its docs.

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -161,49 +198,51 @@ impl VerifiedSet {
     ///
```
upbqdn, line 204: ###### Optional

```suggestion
    ///
    /// This fn is equivalent to `EvictTransaction` in [ZIP-401].
    ///
```

zebrad/src/components/mempool/storage.rs
```diff
@@ -243,22 +270,25 @@ impl Storage {
...
+            assert!(
+                !victim_txs.is_empty(),
+                "mempool is empty, but was expected to be full"
             );
 
-            // If this transaction gets evicted, set its result to the same error
-            // (we could return here, but we still want to check the mempool size)
-            if victim_tx.transaction.id == tx_id {
-                result = Err(SameEffectsChainRejectionError::RandomlyEvicted.into());
+            for victim_tx in victim_txs {
```
upbqdn: Evicting multiple txs at once doesn't seem to correspond to ZIP-401.

zebrad/src/components/mempool/storage.rs
```diff
@@ -605,31 +648,31 @@ impl Storage {
...
         &mut self,
         tip_height: zebra_chain::block::Height,
-    ) -> HashSet<UnminedTxId> {
-        let mut txid_set = HashSet::new();
+    ) -> HashSet<transaction::Hash> {
+        let mut tx_ids = HashSet::new();
         // we need a separate set, since reject() takes the original unmined ID,
         // then extracts the mined ID out of it
-        let mut unmined_id_set = HashSet::new();
+        let mut mined_tx_ids = HashSet::new();
```
upbqdn: Let's document somewhere why we're using "mined" tx ids.

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -23,14 +26,26 @@ use zebra_chain::transaction::MEMPOOL_TRANSACTION_COST_THRESHOLD;
...
+    transactions: HashMap<transaction::Hash, VerifiedUnminedTx>,
+
+    /// A map of dependencies between transactions in the mempool that
+    /// spend or create outputs of other transactions in the mempool.
+    transaction_dependencies: TransactionDependencies,
+
+    /// The [`transparent::Utxo`]s created by verified transactions in the mempool
+    ///
+    /// Note that these UTXOs may not be unspent.
+    /// Outputs can be spent by later transactions or blocks in the chain.
```
upbqdn: Shouldn't we go with outputs instead of UTXOs?

```suggestion
    /// The [`transparent::Output`]s created by verified transactions in the mempool
    ///
    /// Note that these outputs may not be unspent.
    /// They can be spent by later transactions or blocks in the chain.
```

upbqdn: How can they be spent by transactions in the chain?

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -241,8 +288,19 @@ impl VerifiedSet {
             || Self::has_conflicts(&self.orchard_nullifiers, tx.orchard_nullifiers().copied())
     }
 
-    /// Inserts the transaction's outputs into the internal caches.
-    fn cache_outputs_from(&mut self, tx: &Transaction) {
+    /// Inserts the transaction's outputs into the internal caches and responds to pending output requests.
+    fn cache_outputs_and_respond_to_pending_output_requests_from(
+        &mut self,
+        tx_hash: transaction::Hash,
```
upbqdn: ###### Optional

We're passing `tx_hash` separately even though it's part of `tx`. Also, we call this fn only in one spot, and the body is trivial, so we could just remove the fn altogether.

upbqdn (commented):
I'm trying to figure out what happens if someone submits txs containing a circular dependency. We should make sure we can handle such cases well.

zebrad/src/components/mempool/storage/verified_set.rs
upbqdn: https://discord.com/channels/809218587167293450/809251029673312267/1304517119823712349

zebrad/src/components/mempool/storage.rs
upbqdn: Str4d made additional clarification: https://discord.com/channels/809218587167293450/809251029673312267/1304534822999556176.

arya2:
> I'm trying to figure out what happens if someone submits txs containing a circular dependency. We should make sure we can handle such cases well.

The transaction verifier would return an `TransparentInputNotFound` error for both because (unlike pending_utxos in the state service) the mempool only responds to `AwaitOutput` requests once the transaction creating that output has been added to the mempool's verified set.

A similar question around ordering the transactions in block templates caused me some strife (the code should work as expected there too, and it's been excluded from release compilations, but it was difficult to reason about). That exact question may have also caused me some strife a little earlier, as well as what happens to long chains of dependent transactions.

zebrad/src/components/mempool/storage.rs
arya2: > Str4d made additional clarification: https://discord.com/channels/809218587167293450/809251029673312267/1304534822999556176.

There doesn't seem to be any issue with adding rejection errors for all of the dependents, but it also doesn't need to allocate rejections for all of the dependents or complicate the code unnecessarily. ~I'll push a fix soon.~

Fixed in https://github.com/ZcashFoundation/zebra/pull/8857/commits/df9c1c1894885e16fb02194130f8dca40097074d.

zebrad/src/components/mempool/storage.rs
arya2: We actually don't need to use the unmined transaction ids here. We're using just the transaction hashes elsewhere in this PR since transactions that spend the outputs of other transactions in the mempool don't have the auth digest of those transactions in their transparent inputs, only the `OutPoint` (tx hash and output index).

Updated to avoid using the extra collection and with a comment on why mined ids are sufficient in  https://github.com/ZcashFoundation/zebra/pull/8857/commits/874372a8bfeabf08ce9aa903e62dfff8b97a566d.

zebrad/src/components/mempool/storage/verified_set.rs
arya2: They can't, the documentation here was wrong, thank you for catching that. Fixed in https://github.com/ZcashFoundation/zebra/pull/8857/commits/874372a8bfeabf08ce9aa903e62dfff8b97a566d.

zebra-rpc/src/methods/get_block_template_rpcs/types/get_block_template.rs
arya2: Thank you, and nice catch! Removed in https://github.com/ZcashFoundation/zebra/pull/8857/commits/df9c1c1894885e16fb02194130f8dca40097074d.

zebrad/src/components/mempool/storage/verified_set.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/8857/commits/df9c1c1894885e16fb02194130f8dca40097074d.

zebrad/src/components/mempool/storage/verified_set.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/8857/commits/df9c1c1894885e16fb02194130f8dca40097074d.

zebrad/src/components/mempool/storage/verified_set.rs
arya2: Yes we should. Applied in https://github.com/ZcashFoundation/zebra/pull/8857/commits/874372a8bfeabf08ce9aa903e62dfff8b97a566d.

zebrad/src/components/mempool/storage/verified_set.rs
arya2: > so we could just remove the fn altogether

Ah, thank you! The updated method name was far too verbose. Removed in https://github.com/ZcashFoundation/zebra/pull/8857/commits/874372a8bfeabf08ce9aa903e62dfff8b97a566d

zebrad/src/components/mempool/storage/verified_set.rs
```diff
@@ -23,14 +26,25 @@ use zebra_chain::transaction::MEMPOOL_TRANSACTION_COST_THRESHOLD;
 /// This also caches the all the spent outputs from the transactions in the mempool. The spent
 /// outputs include:
 ///
+/// - the dependencies of transactions that spent the outputs of other transactions in the mempool
+/// - the UTXOs of transactions in the mempool
```
upbqdn: Shouldn't this be outputs as well?


```suggestion
/// - the outputs of transactions in the mempool
```

upbqdn (dismissed):

upbqdn (approved):

gustavovalverde:
@mergify refresh

gustavovalverde:
@mergify requeue

gustavovalverde:
@mergify queue

PR #8872, opened 2024-09-18
Introduce initial ZSA support in Zebra (draft PR for CI checks)

dmidem:
This draft PR introduces initial support for Zcash Shielded Assets (ZSA) in the Zebra codebase. The primary goal of this PR is to run CI checks provided by the ZcashFoundation repository to verify compilation correctness and execute unit and integration tests.


mpguerra:
@dmidem as discussed during our call, this will need to have its conflicts resolved before we can run CI on it :)

arya2:
It's not possible to fix all of the complaints from cargo-deny until the changes in librustzcash have been published, and not urgent to fix any of them, but see [Continuous Integration in the Zebra book for how to fix duplicate dependencies in the deny.toml checks](https://zebra.zfnd.org/dev/continuous-integration.html#fixing-duplicate-dependencies-in-check-denytoml-bans) for later.

Note: at least the `windows-sys` entry should be updated when updating Zebra's dependencies on main.

Building `zebra-chain` without any default features seems to be the only failure that needs fixing:
https://github.com/ZcashFoundation/zebra/actions/runs/11015467979/job/30589574007?pr=8872

@gustavovalverde Is there anything else we could try to get [this one](https://github.com/ZcashFoundation/zebra/actions/runs/11015468066/job/30602874752?pr=8872) working?

mpguerra:
I think probably it's good enough to see where it's failing now and see if there are any actionable failures on the QEDIT side, e.g maybe https://github.com/ZcashFoundation/zebra/actions/runs/11015467979/job/30589574007?pr=8872

zebra-chain/src/orchard_zsa/burn.rs
```diff
@@ -0,0 +1,80 @@
...
+        ))
+    }
+}
+
+impl TrustedPreallocate for BurnItem {
+    fn max_allocation() -> u64 {
+        // FIXME: is this a correct calculation way?
+        // The longest Vec<BurnItem> we receive from an honest peer must fit inside a valid block.
+        // Since encoding the length of the vec takes at least one byte, we use MAX_BLOCK_BYTES - 1
+        (MAX_BLOCK_BYTES - 1) / BURN_ITEM_SIZE
```
arya2, line 49: 👍 This looks perfect, it just needs to be higher than the protocol limit without being unreasonably large, and this should be just ~0.1% more than the protocol limit.

zebra-chain/src/transaction/arbitrary.rs
```diff
@@ -918,6 +920,26 @@ pub fn transaction_to_fake_v5(
...
+            ..
+        } => V5 {
+            network_upgrade: block_nu,
+            inputs: inputs.clone(),
+            outputs: outputs.clone(),
+            lock_time: *lock_time,
+            expiry_height: height,
+            sapling_shielded_data: sapling_shielded_data.clone(),
+            // FIXME: is it possible to convert V6 shielded data to V5?
+            // FIXME: add another function for V6, like transaction_to_fake_v6?
```
arya2, line 939: There's probably no need to convert v6 transactions to v5, it won't be used until test blocks with v6 transactions have been added, which may never happen. A function for converting to fake v6 transactions may be useful for testing or might not be necessary either.

zebra-chain/src/transaction/serialize.rs
```diff
@@ -673,6 +688,59 @@ impl ZcashSerialize for Transaction {
...
+                        .branch_id()
+                        .expect("valid transactions must have a network upgrade with a branch id"),
+                ))?;
+
+                // Denoted as `lock_time` in the spec.
+                lock_time.zcash_serialize(&mut writer)?;
+
+                // Denoted as `nExpiryHeight` in the spec.
+                writer.write_u32::<LittleEndian>(expiry_height.0)?;
+
```
arya2, line 723: Nitpick: ZIP 230 is still a draft, but, if the explicit fee ends up being included it should be added here and as a field on the `Transaction::V6` variant.

Everything else here looks good.

zebra-chain/src/orchard_zsa/issuance.rs
```diff
@@ -0,0 +1,260 @@
...
+        // Since encoding the length of the vec takes at least one byte, we use MAX_BLOCK_BYTES - 1
+        (MAX_BLOCK_BYTES - 1) / NOTE_SIZE
+    }
+}
+
+impl ZcashSerialize for IssueAction {
+    fn zcash_serialize<W: io::Write>(&self, mut writer: W) -> Result<(), io::Error> {
+        writer.write_u8(self.is_finalized().as_u8())?;
+        self.notes().zcash_serialize(&mut writer)?;
+        self.asset_desc().zcash_serialize(&mut writer)?;
```
arya2, line 180: These look like they're in the opposite order as they're listed in ZIP 230, but otherwise looks good, as does the serialization for `IssueNote`.

zebra-chain/src/orchard_zsa/issuance.rs
```diff
@@ -0,0 +1,260 @@
...
+
+// Sizes of the serialized values for types in bytes (used for TrustedPreallocate impls)
+// FIXME: are those values correct (43, 32 etc.)?
+//const ISSUANCE_VALIDATING_KEY_SIZE: u64 = 32;
+const ADDRESS_SIZE: u64 = 43;
+const NULLIFIER_SIZE: u64 = 32;
+const NOTE_VALUE_SIZE: u64 = 4;
+const RANDOM_SEED_SIZE: u64 = 32;
+// FIXME: is this a correct way to calculate (simple sum of sizes of components)?
+const NOTE_SIZE: u64 =
```
arya2, line 44: It says [147 here](https://github.com/zcash/zips/pull/854/files#diff-57dfe6dfbd7af7e21e6f4731a1c1af0028c106e8dd61122cf64d22ffd0474fa9R326) but these add up to 143, this or the ZIP may need an update (probably this implementation, using a u64 for note value seems good).

zebra-chain/src/orchard_zsa/issuance.rs
```diff
@@ -0,0 +1,260 @@
...
+        let rseed = zcash_deserialize_random_seed(&mut reader, &rho)?;
+
+        Option::from(Note::from_parts(recipient, value, asset, rho, rseed))
+            .ok_or_else(|| SerializationError::Parse("Invalid orchard_zsa Note components!"))
+    }
+}
+
+impl TrustedPreallocate for Note {
+    fn max_allocation() -> u64 {
+        // FIXME: is this a correct calculation way?
```
arya2, line 169: Yep! This looks good.

zebra-chain/src/orchard_zsa/issuance.rs
```diff
@@ -0,0 +1,260 @@
...
+        let notes = (&mut reader).zcash_deserialize_into()?;
+        let asset_descr = (&mut reader).zcash_deserialize_into()?;
+        Ok(IssueAction::from_parts(asset_descr, notes, finalize))
+    }
+}
+
+impl TrustedPreallocate for IssueAction {
+    fn max_allocation() -> u64 {
+        // FIXME: impl correct calculation
+        10
```
arya2, line 197: This looks good, the function that calls `max_allocation()` ([`zcash_deserialize_external_count()`](https://github.com/QED-it/zebra/blob/switch-to-zsa-crates-nu6-txv6-gen/zebra-chain/src/serialization/zcash_deserialize.rs#L79-L100)) will deserialize and pre-allocate sequentially based on the actual counts, so until it goes through each item and reads their `CompactSize`, it'll just be pre-allocating memory for the pointers, if one of them is shorter than expected it'll return an error, and their combined real size can't be longer than the `Codec`'s `builder.max_len` field, which is `MAX_PROTOCOL_MESSAGE_LEN`.

```suggestion
        (MAX_BLOCK_BYTES - 1) / NOTE_SIZE
```

arya2 (commented):
I haven't finished reviewing yet, but this is looking great so far. The well-factored code is appreciated.

Issue #8434, opened 2024-04-24
Support constructing pre-Canopy block templates


arya2:
## Motivation

This is needed to mine pre-Canopy blocks on Regtest and some custom Testnets (except forks of the default Testnet)

Supporting pre-Canopy blocks on Regtest will also require generating coinbase transactions with founders' rewards until the first halving in order to be compatible with zcashd. Ideally it would also validate the founders' rewards in the `subsidy_is_valid()` check. See [this comment](https://github.com/ZcashFoundation/zebra/pull/8888#discussion_r1788342311) for more details.

upbqdn:
This issue is currently tracked in the "Zebra Ready for zcashd Deprecation" milestone: https://github.com/ZcashFoundation/zebra/milestone/41. I don't think we need it for the deprecation. @arya2, do you agree?

arya2:
Definitely not needed for zcashd deprecation.

natalieesk:
Closed as not needed for zcashd deprecation

PR #8888, opened 2024-09-23
feat(regtest): Add regtest halving interval and port test

### LINKED ISSUE #8880: Make the halving interval configurable for custom networks and regtests
There had been a request for making the halving heights configurable in Regtest mode: 

https://discordapp.com/channels/809218587167293450/1285973514964439121/1285975294997434369

For the first halving, we have a `TODO` here to do this: https://github.com/ZcashFoundation/zebra/blob/v1.9.0/zebra-chain/src/parameters/network/subsidy.rs#L394

Also, this feature will be useful if we plan to merge https://github.com/zcash/zcash/blob/v5.10.0/qa/rpc-tests/nuparams.py into the zebra rpc test suite.

oxarbitrage:
## Motivation

We want to be able to configure the halving interval when we are in custom testnet or use the fixed default in regtest.

Close https://github.com/ZcashFoundation/zebra/issues/8880 

## Solution

- Added `FIRST_HALVING_REGTEST` at height 144 to match zcashd regtest mode.
- Added configuration option `halving_interval` for custom testnets.
- Ported the nuparams.py tests from zcashd
- Fixed inconsistency between `Nu6` and `NU6` in rpc outputs.

### Tests

`./qa/pull-tester/rpc-tests.py nuparams`

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -413,8 +413,7 @@ impl ParameterSubsidy for Network {
                 } else if params.is_default_testnet() {
                     FIRST_HALVING_TESTNET
                 } else {
-                    height_for_halving_index(1, self)
-                        .expect("first halving height should be available")
+                    height_for_halving(1, self).expect("first halving height should be available")
                 }
             }
         }
@@ -558,8 +557,8 @@ pub fn funding_stream_address_period<N: ParameterSubsidy>(height: Height, networ
 /// See `Halving(height)`, as described in [protocol specification §7.8][7.8]
 ///
 /// [7.8]: https://zips.z.cash/protocol/protocol.pdf#subsidies
-pub fn height_for_halving_index(halving_index: u32, network: &Network) -> Option<Height> {
-    if halving_index == 0 {
+pub fn height_for_halving(halving: u32, network: &Network) -> Option<Height> {
+    if halving == 0 {
         return Some(Height(0));
     }
 
@@ -571,7 +570,7 @@ pub fn height_for_halving_index(halving_index: u32, network: &Network) -> Option
             .0,
     );
     let pre_blossom_halving_interval = network.pre_blossom_halving_interval();
-    let halving_index = i64::from(halving_index);
+    let halving_index = i64::from(halving);
 
     let unscaled_height = halving_index
         .checked_mul(pre_blossom_halving_interval)
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -40,10 +40,6 @@ pub const BLOSSOM_POW_TARGET_SPACING_RATIO: u32 = 2;
 /// `(60 * 60 * 24 * 365 * 4) / 150 = 840960`
 pub const PRE_BLOSSOM_HALVING_INTERVAL: HeightDiff = 840_000;
 
-/// The halving height interval in the regtest is 6 hours.
-/// [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
-pub(crate) const PRE_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff = 144;
-
 /// After Blossom the block time is reduced to 75 seconds but halving period should remain around 4 years.
 pub const POST_BLOSSOM_HALVING_INTERVAL: HeightDiff =
     PRE_BLOSSOM_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff);
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -42,24 +42,20 @@ pub const PRE_BLOSSOM_HALVING_INTERVAL: HeightDiff = 840_000;
 
 /// The halving height interval in the regtest is 6 hours.
 /// [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
-pub const PRE_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff = 144;
+pub(crate) const PRE_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff = 144;
 
 /// After Blossom the block time is reduced to 75 seconds but halving period should remain around 4 years.
 pub const POST_BLOSSOM_HALVING_INTERVAL: HeightDiff =
     PRE_BLOSSOM_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff);
 
-/// After Blossom the block time is reduced to 75 seconds but halving period should remain 6 hours.
-pub const POST_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff =
-    PRE_BLOSSOM_REGTEST_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff);
-
 /// The first halving height in the testnet is at block height `1_116_000`
 /// as specified in [protocol specification §7.10.1][7.10.1]
 ///
 /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
-pub const FIRST_HALVING_TESTNET: Height = Height(1_116_000);
+pub(crate) const FIRST_HALVING_TESTNET: Height = Height(1_116_000);
 
 /// The first halving height in the regtest is at block height `287`.
-pub const FIRST_HALVING_REGTEST: Height = Height(287);
+const FIRST_HALVING_REGTEST: Height = Height(287);
 
 /// The funding stream receiver categories.
 #[derive(Deserialize, Clone, Copy, Debug, Eq, Hash, PartialEq)]
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -50,6 +50,11 @@ pub const POST_BLOSSOM_HALVING_INTERVAL: HeightDiff =
 /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
 pub const FIRST_HALVING_TESTNET: Height = Height(1_116_000);
 
+/// The first halving height in the regtest is at block height `114`
+/// as [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
+/// is 114 blocks.
+pub const FIRST_HALVING_REGTEST: Height = Height(114);
+
 /// The funding stream receiver categories.
 #[derive(Deserialize, Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum FundingStreamReceiver {
@@ -391,8 +396,15 @@ impl ParameterSubsidy for Network {
             Network::Mainnet => NetworkUpgrade::Canopy
                 .activation_height(self)
                 .expect("canopy activation height should be available"),
-            // TODO: Check what zcashd does here, consider adding a field to `testnet::Parameters` to make this configurable.
-            Network::Testnet(_params) => FIRST_HALVING_TESTNET,
+            Network::Testnet(params) => {
+                if params.is_regtest() {
+                    FIRST_HALVING_REGTEST
+                } else if params.is_default_testnet() {
+                    FIRST_HALVING_TESTNET
+                } else {
+                    params.halving_interval()
+                }
+            }
         }
     }
 }
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -410,11 +410,7 @@ impl ParameterSubsidy for Network {
                 } else if params.is_default_testnet() {
                     FIRST_HALVING_TESTNET
                 } else {
-                    Height(
-                        self.post_blossom_halving_interval()
-                            .try_into()
-                            .expect("post blossom halving interval should fit in a Height"),
-                    )
+                    height_for_halving_index(1, self).unwrap()
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -50,6 +50,11 @@ pub const POST_BLOSSOM_HALVING_INTERVAL: HeightDiff =
 /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
 pub const FIRST_HALVING_TESTNET: Height = Height(1_116_000);
 
+/// The first halving height in the regtest is at block height `114`
+/// as [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
+/// is 114 blocks.
+pub const FIRST_HALVING_REGTEST: Height = Height(114);
```
arya2: The first halving height should depend on the Blossom activation height on Regtest (it should be between 144-287), and also the slow start and halving intervals on custom Testnets, so Zebra can't calculate the halving heights until after it has read its configuration:

image

(page 134 of the [protocol specification](https://zips.z.cash/protocol/protocol.pdf))

It _could_ use a `OnceCell` for custom Testnets/Regtest and calculate them at compile time for Mainnet and the default Testnet (or use hard-coded values), but it should be quick enough that there's still lower-hanging fruit for improving performance.

Update: There's no need for a OnceCell, it could be added as a field on `testnet::Parameters` or as a constant for Mainnet.

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -629,6 +646,11 @@ impl Parameters {
     pub fn disable_pow(&self) -> bool {
         self.disable_pow
     }
+
+    /// Returns the halving interval for this network
+    pub fn halving_interval(&self) -> Height {
```
arya2: This should be used where `PRE_BLOSSOM_HALVING_INTERVAL` or `POST_BLOSSOM_HALVING_INTERVAL` are currently being used (multiplied or divided by `BLOSSOM_POW_TARGET_SPACING_RATIO`).

zebra-network/src/config.rs
```diff
@@ -597,6 +597,7 @@ impl<'de> Deserialize<'de> for Config {
             activation_heights: Option<ConfiguredActivationHeights>,
             pre_nu6_funding_streams: Option<ConfiguredFundingStreams>,
             post_nu6_funding_streams: Option<ConfiguredFundingStreams>,
+            halving_interval: Option<u32>,
```
arya2: Could we rename this to `pre_blossom_halving_interval`? The post-Blossom halving interval should be double the pre-Blossom interval, and they're integers, so the pre-Blossom value seems slightly easier to make configurable or to configure.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -40,16 +40,27 @@ pub const BLOSSOM_POW_TARGET_SPACING_RATIO: u32 = 2;
...
+    PRE_BLOSSOM_REGTEST_HALVING_INTERVAL * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff);
+
 /// The first halving height in the testnet is at block height `1_116_000`
 /// as specified in [protocol specification §7.10.1][7.10.1]
 ///
 /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
 pub const FIRST_HALVING_TESTNET: Height = Height(1_116_000);
 
+/// The first halving height in the regtest is at block height `287`.
+pub const FIRST_HALVING_REGTEST: Height = Height(287);
```
arya2: This actually works for now, since the blossom activation height is not configurable on Regtest.

We need an update to the first halving height on custom Testnets when the blossom activation height is different from the default Testnet, but that can be done in another PR.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -391,8 +408,38 @@ impl ParameterSubsidy for Network {
...
+                if params.is_regtest() {
+                    FIRST_HALVING_REGTEST
+                } else if params.is_default_testnet() {
+                    FIRST_HALVING_TESTNET
+                } else {
+                    Height(
+                        self.post_blossom_halving_interval()
+                            .try_into()
+                            .expect("post blossom halving interval should fit in a Height"),
+                    )
```
arya2: This part's not right, but neither is returning `FIRST_HALVING_TESTNET` as it currently is on main.

We could leave it as it is (returning `FIRST_HALVING_TESTNET`) or add it as a field on `testnet::Parameters`. It could use the `height_for_halving_index()` in #8894 to calculate the value of the field if #8894 is merged into this PR.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -391,8 +408,38 @@ impl ParameterSubsidy for Network {
...
+                }
+            }
+        }
+    }
+
+    fn post_blossom_halving_interval(&self) -> HeightDiff {
+        self.pre_blossom_halving_interval() * (BLOSSOM_POW_TARGET_SPACING_RATIO as HeightDiff)
```
arya2: Optional/Nitpick/Subjective: Could we calculate this when building `testnet::Parameters` and add it as a field?

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -391,8 +408,38 @@ impl ParameterSubsidy for Network {
...
+            Network::Mainnet => PRE_BLOSSOM_HALVING_INTERVAL,
+            Network::Testnet(params) => {
+                if params.is_regtest() {
+                    PRE_BLOSSOM_REGTEST_HALVING_INTERVAL
+                } else if params.is_default_testnet() {
+                    PRE_BLOSSOM_HALVING_INTERVAL
+                } else {
+                    params.pre_blossom_halving_interval()
+                }
+            }
```
arya2:
```suggestion
            Network::Testnet(params) => params.pre_blossom_halving_interval(),
```

arya2 (commented):
This is looking good, the only blocker is an update to the `height_for_first_halving()` method so it returns `FIRST_HALVING_TESTNET` on custom testnets, or ideally the correct value. I'll try writing a `halving_for_height()` function so we can use the correct value there.

Update: `FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL` and the `funding_stream_address_period()` are also using the constant post blossom halving interval.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -40,16 +40,27 @@ pub const BLOSSOM_POW_TARGET_SPACING_RATIO: u32 = 2;
 /// `(60 * 60 * 24 * 365 * 4) / 150 = 840960`
 pub const PRE_BLOSSOM_HALVING_INTERVAL: HeightDiff = 840_000;
 
+/// The halving height interval in the regtest is 6 hours.
+/// [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
+pub const PRE_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff = 144;
```
arya2: Could we make these constants private to the module to ensure they can't be used directly? (I just made this mistake now in a suggestion PR).

zebra-consensus/src/block/subsidy/general.rs
```diff
@@ -23,40 +23,71 @@ use crate::{block::SubsidyError, funding_stream_values};
...
+            .0,
+    );
+    let pre_blossom_halving_interval = network.pre_blossom_halving_interval();
+    let halving_index = i64::from(halving_index);
+
+    let unscaled_height = halving_index * pre_blossom_halving_interval;
+    let pre_blossom_height = unscaled_height.min(blossom_height) + slow_start_shift;
+    let post_blossom_height = 0.max(unscaled_height - blossom_height)
+        * i64::from(BLOSSOM_POW_TARGET_SPACING_RATIO)
+        + slow_start_shift;
```
arya2: Using checked operations here could help if the pre-Blossom halving interval is configured to be excessively large, even if it panics with a message to reduce the interval.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -379,6 +386,15 @@ pub trait ParameterSubsidy {
...
     /// [7.10]: <https://zips.z.cash/protocol/protocol.pdf#fundingstreams>
     fn height_for_first_halving(&self) -> Height;
+
+    /// Returns the halving interval after Blossom
+    fn post_blossom_halving_interval(&self) -> HeightDiff;
+
+    /// Returns the halving interval before Blossom
+    fn pre_blossom_halving_interval(&self) -> HeightDiff;
+
+    /// Returns the address change interval for funding streams
```
arya2, line 392: The [reference to the spec here is also wrong/outdated](https://github.com/ZcashFoundation/zebra/blob/regtest-halving-interval/zebra-chain/src/parameters/network/subsidy.rs#L317-L321) if we could update it to refer to 7.10 instead of 7.10.1 there too.

```suggestion
    /// Returns the address change interval for funding streams
    /// as described in [protocol specification §7.10][7.10].
    ///
    /// > FSRecipientChangeInterval := PostBlossomHalvingInterval / 48
    ///
    /// [7.10]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
```

zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -57,6 +60,17 @@ impl ParameterSubsidy for TestnetParameterSubsidyImpl {
...
+        POST_BLOSSOM_HALVING_INTERVAL
+    }
+
+    fn pre_blossom_halving_interval(&self) -> block::HeightDiff {
+        PRE_BLOSSOM_HALVING_INTERVAL
+    }
+
+    fn funding_stream_address_change_interval(&self) -> HeightDiff {
+        FUNDING_STREAM_ADDRESS_CHANGE_INTERVAL
+    }
```
arya2: These should be based on the `pre_blossom_halving_interval` config field and the Blossom activation height, and `testnet::ParametersBuilder` should panic or return an error when:
- Setting the halving interval or activation heights after setting the funding streams, or
- Calling `finish()` with a non-default halving interval or Blossom activation height but the default Testnet funding streams (this will affect `new_regtest()`, zcashd has no funding streams on Regtest, we should clear the funding streams for Regtest in Zebra as well, though in practice it seems unlikely anyone would reach the funding stream start heights anyway).

Related: zcashd seems to have founders rewards in their Regtest until the `first_halving_height - 1`, we may want to add that in Zebra or remove it in zcashd, we should ask the ECC what they think would be best here.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -391,10 +407,36 @@ impl ParameterSubsidy for Network {
...
-            // TODO: Check what zcashd does here, consider adding a field to `testnet::Parameters` to make this configurable.
-            Network::Testnet(_params) => FIRST_HALVING_TESTNET,
+            Network::Testnet(params) => {
+                if params.is_regtest() {
+                    FIRST_HALVING_REGTEST
+                } else if params.is_default_testnet() {
+                    FIRST_HALVING_TESTNET
+                } else {
+                    height_for_halving_index(1, self)
+                        .expect("first halving height should be available")
```
arya2: Optional: It would be nice to remove any calls to `is_default_testnet()` outside of tests or code that only runs at startup, here we could compute the first halving height when building `testnet::Parameters`, or even the first 3-4 halving heights so they can be read from memory instead of being computed (which could be hard-coded constants for Mainnet). 

I think this is fine for now and that that can be done in another PR since it only affects Testnet, and is only called twice per semantically verified block and once per call to the `get_block_subsidy` RPC method.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -503,10 +545,54 @@ pub fn funding_stream_address_period<N: ParameterSubsidy>(height: Height, networ
...
         .try_into()
         .expect("all values are positive and smaller than the input height")
 }
+
+/// The first block height of the halving at the provided halving index for a network.
+///
+/// See `Halving(height)`, as described in [protocol specification §7.8][7.8]
+///
+/// [7.8]: https://zips.z.cash/protocol/protocol.pdf#subsidies
+pub fn height_for_halving_index(halving_index: u32, network: &Network) -> Option<Height> {
```
arya2: Could we remove `index` from the function/argument names? It seems needlessly confusing (I think I may have named it, but I made a mistake).

```suggestion
pub fn height_for_halving(halving: u32, network: &Network) -> Option<Height> {
```

arya2 (commented):
This is looking great, the only potential blocker is an update to `TestnetParameterSubsidyImpl` to account for the halving interval and Blossom activation height when validating configured funding streams.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -40,6 +40,10 @@ pub const BLOSSOM_POW_TARGET_SPACING_RATIO: u32 = 2;
 /// `(60 * 60 * 24 * 365 * 4) / 150 = 840960`
 pub const PRE_BLOSSOM_HALVING_INTERVAL: HeightDiff = 840_000;
 
+/// The halving height interval in the regtest is 6 hours.
+/// [zcashd regtest halving interval](https://github.com/zcash/zcash/blob/v5.10.0/src/consensus/params.h#L252)
+pub(crate) const PRE_BLOSSOM_REGTEST_HALVING_INTERVAL: HeightDiff = 144;
```
arya2: Nitpick/Subjective: I think it would be nice to include this in the `parameters::network::testnet` module.

zebra-chain/src/parameters/network/testnet.rs
mpguerra: > Related: zcashd seems to have founders rewards in their Regtest until the `first_halving_height - 1`, we may want to add that in Zebra or remove it in zcashd, we should ask the ECC what they think would be best here.

I would say that we probably want to add it to Zebra rather than removing in zcashd

oxarbitrage: Can you do this in a suggestion PR or commit ? It is not fully clear to me how to make it.

arya2: @mpguerra I was wrong about zcashd having founders' rewards in their Regtest, [it only applies until Canopy and after the genesis block](https://github.com/zcash/zcash/blob/master/src/miner.cpp#L122-L155), and Zebra's Regtest [always activates Canopy at `Height(1)`](https://github.com/ZcashFoundation/zebra/blob/custom-testnet-funding-stream-addr-period-check/zebra-chain/src/parameters/network/testnet.rs#L596), so they already match.

Updated https://github.com/ZcashFoundation/zebra/issues/8434 to include adding a founders' rewards if zcashd hasn't been deprecated before it's being implemented.

Opened suggestion PR #8921 to address the rest of this comment.

arya2 (approved):
Thank you!

PR #8894, opened 2024-09-26
Suggestion for "feat(regtest): Add regtest halving interval and port test"

arya2:
This is a suggestion PR for #8888 that:
- Adds a `height_for_halving_index()` function to be used for calculating the height for first index on custom Testnets and Regtest (though it's not needed on Regtest as the relevant parameters are not configurable and a hard-coded value is fine).
- Factors calculating of the number of halvings out of `halving_divisor()` into `num_halvings()` for use in tests

@oxarbitrage can review and optionally merge it into their PR.


oxarbitrage (approved):
thanks, trying it.

PR #8896, opened 2024-09-29
Point zcash_script dependency at master

sellout:
## Motivation

Show that the existing Zebra main branch would work with a new zcash_script release.

This should not be merged.


mpguerra:
PR served its purpose, closing

PR #8897, opened 2024-09-29
Point #8751 at zcash_script master

sellout:
## Motivation

Show that #8751 would work with a new zcash_script release.

This should not be merged.


sellout:
This shouldn’t have been a separate PR, sorry.

PR #8900, opened 2024-09-30
feat(rpc): Cookie auth system for the RPC endpoint

### LINKED ISSUE #8864: Protect the Zebra RPC endpoint
The Zebra RPC endpoint is currently unencrypted and accessible by anyone who knows the address and port. While the current recommendation is to restrict access to localhost or a trusted local network, there are cases where the endpoint needs to be open to the internet in production environments.

Implement one of the following authentication methods:

- Basic HTTP Authentication: This method would require users to provide a username and password when connecting to the RPC endpoint.
- Cookie-based Authentication: Similar to Zcashd, this method uses a randomly generated cookie stored on the file system to authenticate RPC connections.

We could implement just one of these methods or explore alternative solutions.  

Authentication should be enabled by default but turning it off should be an option for testing purposes and setups where the endpoint is restricted to localhost or a private network.

Encryption of RPC traffic (for example using TLS) is not part of this ticket but might be considered in the future.

oxarbitrage:
## Motivation

We want to authenticate the RPC method by the zcashd cookie method.

Close https://github.com/ZcashFoundation/zebra/issues/8864

## Solution

- Added `enable_cookie_auth` config field to use cookie authentication, enabled by default.
- Generate cookie when the RPC endpoint starts.
- Start the RPC endpoint server with that generated cookie.
- In the middleware, if auth is enabled, check for basic HTTP auth header.
  - Redirect to request if the user provided auth matches the one generated that the server has.
  - Else error 401.

### Tests

- Manual
- All tests should remain working as we disabled auth for all of them.

### Follow-up Work

- Make the python RPC tests work with authproxy again, or/and:
  - Create a rust test to query with auth. 

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


oxarbitrage:
Requests from a remote host that don't have the cookie generated at startup will be rejected. In zcashd, the `zcash-cli` can be used in a remote host that has the cookie and authenticate. However, we are not targeting a specific application, we want to use the authentication method in a generic way from a remote client. One option can be to use curl with the cookie file and compare server side. Other options are welcome, i am still brainstorming it.

oxarbitrage:
We had a chat today about this with @upbqdn and @arya2, i am adding some more research here. 

About the cookie auth method:

> ... Read access to this file controls who can access through RPC ...

https://bitcoin.org/en/release/v0.12.0#rpc-random-cookie-rpc-authentication

That means the cookie method is actually for local access. I think we should focus on that in this PR.

For remote access, we thought in username/password over TLS/SSL as an option. Bitcoin supported this for its RPC endpoint in the past however they don't do it anymore claiming that the RPC access should only be shared with trusted environments.

https://en.bitcoin.it/wiki/Enabling_SSL_on_original_client_daemon#:~:text=SSL%20for%20RPC%20in%20Bitcoin,was%20criticized%20for%20some%20time.

https://bitcoin.stackexchange.com/questions/108293/why-does-bitcoin-cores-rpc-interface-not-use-encryption

It seems that the remote access should be a combination of username and password, with the additional `rpcallowip` config option plus a vpn layer.

oxarbitrage:
Continuing with the cookie auth method, the zcash-cli sends the cookie content as basic HTTP credentials to the server: https://github.com/zcash/zcash/blob/master/src/bitcoin-cli.cpp#L251
https://github.com/zcash/zcash/blob/master/src/bitcoin-cli.cpp#L266

We want to do that but just with `curl`, we need to intercept the `Authorization` in the zebra RPC middleware and compare it with the cookie in the server side.

I got confused thinking the cookie method will work for remote access, my apologies for that.

upbqdn:
> That means the cookie method is actually for local access. I think we should focus on that in this PR.

Do we have any use cases that require authentication for local access, though?

oxarbitrage:
> Do we have any use cases that require authentication for local access, though?

It's a security measure. You can't access the resources if you don't have read access to the cookie, even if you are in the same machine.

zebra-rpc/src/config.rs
```diff
@@ -94,6 +99,9 @@ impl Default for Config {
 
             // Debug options are always off by default.
             debug_force_finished_sync: false,
+
+            //
```
upbqdn:
```suggestion
            // Use the default cache dir for the auth cookie.
```

zebra-rpc/src/methods.rs
```diff
@@ -314,6 +314,10 @@ pub trait Rpc {
     /// tags: control
     #[rpc(name = "stop")]
     fn stop(&self) -> Result<String>;
+
+    #[rpc(name = "unauthenticated")]
+    /// A dummy RPC method that just returns a non-authenticated RPC error.
```

```suggestion
    /// A dummy RPC method that just returns a non-authenticated RPC error.
    #[rpc(name = "unauthenticated")]
```

zebra-rpc/src/server/cookie.rs
```diff
@@ -0,0 +1,47 @@
...
+    io::{Read, Write},
+    path::PathBuf,
+};
+
+/// The user field in the cookie (arbitrary, only for recognizability in debugging/logging purposes)
+pub const COOKIEAUTH_USER: &str = "__cookie__";
+/// Default name for auth cookie file */
+pub const COOKIEAUTH_FILE: &str = ".cookie";
+
+/// Generate a new auth cookie and return the encoded password.
```
upbqdn:
```suggestion
/// Generate a new auth cookie and store it in the given `cookie_dir`.
```

zebra-rpc/src/server/cookie.rs
```diff
@@ -0,0 +1,47 @@
...
+pub fn generate(cookie_dir: PathBuf) -> Option<()> {
```
upbqdn: ###### Optional

Returning `Result` would be more expressive.

zebra-rpc/src/server/cookie.rs
```diff
@@ -0,0 +1,47 @@
...
+    let mut file = File::open(cookie_dir.join(COOKIEAUTH_FILE)).ok()?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents).ok()?;
+
+    let parts: Vec<&str> = contents.split(":").collect();
+    Some(parts[1].to_string())
+}
+
+/// Delete the auth cookie.
+pub fn delete() -> Option<()> {
```
upbqdn: ###### Optional

Similarly here.

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -141,4 +149,51 @@ impl FixHttpRequestMiddleware {
...
+            .and_then(|token| URL_SAFE.decode(token).ok())
+            .and_then(|decoded| String::from_utf8(decoded).ok())
+            .and_then(|decoded_str| {
+                decoded_str
+                    .split(':')
+                    .nth(1)
+                    .map(|password| password.to_string())
+            })
+            .map_or(false, |password| {
+                if let Some(cookie_password) = cookie::get(self.0.cookie_dir.clone()) {
```
upbqdn: We read the cookie from the disk on each check here and only rely on the OS caching the disk reads. We could keep the cookie in the memory instead. For example, by using `lazy_static!` in `cookie::get`.

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -141,4 +149,51 @@ impl FixHttpRequestMiddleware {
             );
         }
     }
+
+    /// Change the method name in the JSON request.
```
upbqdn: ###### Optional

Could we move the contents of this fn to `unauthenticated` and remove the fn since we call it only from there, and it only uses a single-use, hard-coded method name?

zebrad/tests/common/checkpoints.rs
```diff
@@ -504,7 +505,8 @@ pub fn wait_for_zebra_checkpoints_generation<
 /// Returns an approximate `zebrad` tip height, using JSON-RPC.
 #[tracing::instrument]
 pub async fn zebrad_tip_height(zebra_rpc_address: SocketAddr) -> Result<Height> {
-    let client = RpcRequestClient::new(zebra_rpc_address);
+    let auth_cookie = cookie::get(RpcConfig::default().cookie_dir).expect("cookie should exist");
+    let client = RpcRequestClient::new(zebra_rpc_address, auth_cookie);
```
upbqdn: We write the cookie to the disk, and then read it even in tests that are not related to the authentication. We could generate the cookie for tests without writing it to the disk at all, and instantiate it only through `lazy_static!`.

upbqdn (dismissed):

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -34,13 +37,18 @@ use jsonrpc_http_server::{
...
+#[derive(Clone, Debug)]
+pub struct FixHttpRequestMiddleware(pub crate::config::Config);
 
 impl RequestMiddleware for FixHttpRequestMiddleware {
     fn on_request(&self, mut request: Request<Body>) -> RequestMiddlewareAction {
         tracing::trace!(?request, "original HTTP request");
 
+        // Check if the request is authenticated
+        if !self.check_credentials(request.headers_mut()) {
+            request = Self::unauthenticated(request);
```
arya2: This should work for responding here instead of changing the method name (though it doesn't return the request id or jsonrpc version, it could but it would have to parse the request body first, which didn't seem worthwhile):

```suggestion
            let error = jsonrpc_core::Error {
                code: jsonrpc_core::ErrorCode::ServerError(401),
                message: "unauthenticated method".to_string(),
                data: None,
            };

            return jsonrpc_http_server::Response {
                code: jsonrpc_http_server::hyper::StatusCode::from_u16(401)
                    .expect("hard-coded status code should be valid"),
                content_type: header::HeaderValue::from_static("application/json; charset=utf-8"),
                content: serde_json::to_string(&jsonrpc_core::Response::from(error, None))
                    .expect("hard-coded result should serialize"),
            }
            .into();
```

zebra-rpc/src/config.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8900/commits/77082947877c025b51de5d61a99e63769eeb9215

zebra-rpc/src/server/cookie.rs
oxarbitrage: Updated.

zebra-rpc/src/server/cookie.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8900/commits/d48d9d4cdfbbd7edac5cd45819f4c22fc1463b70

zebra-rpc/src/server/cookie.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8900/commits/d48d9d4cdfbbd7edac5cd45819f4c22fc1463b70

zebra-rpc/src/methods.rs
oxarbitrage: Do not apply anymore thanks to https://github.com/ZcashFoundation/zebra/pull/8900/commits/963c1d26348c70fdec92955fd1ebb4077c208355

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: `unauthenticated` is gone in https://github.com/ZcashFoundation/zebra/pull/8900/commits/963c1d26348c70fdec92955fd1ebb4077c208355

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: Thanks, that is really good! Added in https://github.com/ZcashFoundation/zebra/pull/8900/commits/963c1d26348c70fdec92955fd1ebb4077c208355

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: Thanks. I tried that approach to address the read/write issue, but I realized we also want authentication to be enabled by default, with the option to disable it. https://github.com/ZcashFoundation/zebra/issues/8864.

To address this, I added a configuration option to toggle authentication and disabled auth for all tests. 

I manually tested the system, but we should add specific tests to verify the cookie-based authentication method.

zebrad/tests/common/checkpoints.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8900#discussion_r1798481844

mpguerra:
Can this be merged once approved?

zebra-rpc/src/server/cookie.rs
```diff
@@ -0,0 +1,48 @@
...
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+
+    let parts: Vec<&str> = contents.split(":").collect();
+    Ok(parts[1].to_string())
+}
+
+/// Delete the auth cookie.
+pub fn delete() -> Result<()> {
+    remove_file(COOKIEAUTH_FILE)?;
```
upbqdn: I don't think the removal works because we're using only the file name without the path to it.

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -141,4 +161,30 @@ impl FixHttpRequestMiddleware {
...
+            .and_then(|token| URL_SAFE.decode(token).ok())
+            .and_then(|decoded| String::from_utf8(decoded).ok())
+            .and_then(|decoded_str| {
+                decoded_str
+                    .split(':')
+                    .nth(1)
+                    .map(|password| password.to_string())
+            })
+            .map_or(false, |password| {
+                if let Ok(cookie_password) = cookie::get(self.0.cookie_dir.clone()) {
```
upbqdn: We're still reading the cookie from the disk on each RPC request.

upbqdn:
> Can this be merged once approved?

Nope.

upbqdn:
This PR is missing a priority label.

zebra-rpc/src/server/cookie.rs
oxarbitrage: Fixed in https://github.com/ZcashFoundation/zebra/pull/8900/commits/43131b8a5645f76c0d3b453a0099d9a7061b5f5a

zebra-rpc/src/server/http_request_compatibility.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/8900/commits/43131b8a5645f76c0d3b453a0099d9a7061b5f5a

upbqdn (approved):

upbqdn:
Should we also update user docs and describe where users can find the cookie and how to use it?

PR #8901, opened 2024-09-30
change(mempool): Return verification result after attempting to insert transactions in the mempool

### LINKED ISSUE #8778: Return verification errors from `sendrawtransaction` RPC method

arya2:
## Motivation

We want to return a verification result from the mempool after a transaction has been inserted into its verified set (or has failed some final checks in the mempool).

## Solution

- Moves timeout from mempool to mempool's `Download` struct
- Sends the final result to the response channel after attempting to insert a transaction into the mempool storage instead of after a transaction is successfully verified

### Tests

Updates the `mempool_responds_to_await_output` test to expect the mempool storage to include a transaction immediately after receiving its verification result.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebrad/src/components/mempool/downloads.rs
+++ b/zebrad/src/components/mempool/downloads.rs
@@ -276,7 +276,7 @@ where
     pub fn download_if_needed_and_verify(
         &mut self,
         gossiped_tx: Gossip,
-        rsp_tx: Option<oneshot::Sender<Result<(), BoxError>>>,
+        mut rsp_tx: Option<oneshot::Sender<Result<(), BoxError>>>,
     ) -> Result<(), MempoolError> {
         let txid = gossiped_tx.id();
 
@@ -402,11 +402,11 @@ where
         // Tack the hash onto the error so we can remove the cancel handle
         // on failure as well as on success.
         .map_err(move |e| (e, txid))
-            .inspect(move |result| {
-                // Hide the transaction data to avoid filling the logs
-                let result = result.as_ref().map(|_tx| txid);
-                debug!("mempool transaction result: {result:?}");
-            })
+        .inspect(move |result| {
+            // Hide the transaction data to avoid filling the logs
+            let result = result.as_ref().map(|_tx| txid);
+            debug!("mempool transaction result: {result:?}");
+        })
         .in_current_span();
 
         let task = tokio::spawn(async move {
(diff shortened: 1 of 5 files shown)


mpguerra:
Is there a related issue for this PR?

arya2:
> Is there a related issue for this PR?

It's related to https://github.com/ZcashFoundation/zebra/issues/8778.

Zebra began returning errors from the `sendrawtransaction` RPC method when verifying the transaction fails or was cancelled in https://github.com/ZcashFoundation/zebra/pull/8788, but in [one of the tests](https://github.com/ZcashFoundation/zebra/pull/8857/commits/2342d75d6d0f6aa007b2278c54704149e59f1857) for https://github.com/ZcashFoundation/zebra/pull/8857, we noticed that it sends the success response too early because:
- There are duplicate spend and mempool size checks that could still fail, or
- Re-verifying the transaction could fail if the chain tip changes during verification and the update best chain spends or reveals relevant outputs or nullifiers.

zebrad/src/components/mempool/downloads.rs
arya2, line 1: To ensure that the Zebra responds with the mempool insertion result as quickly as possible, we should:
- Move the call to the mempool's `poll_ready()` method from the transaction verifier that was added in https://github.com/ZcashFoundation/zebra/pull/8857 to the mempool's `Downloads` here
  -  This seems simpler than polling from the transaction verifier and more precise than using a relatively long sleep duration to wait for the rest of the Downloads task to execute
- Always poll the mempool after a transaction has been downloaded and verified instead of only when polling the mempool after transactions that create new transparent outputs have been verified.
  - We could also add a flag for transactions queued via the `sendrawtransaction` RPC method and only poll the mempool for successfully verified transactions that don't create new transparent outputs when the flag is set, though that may not be a worthwhile optimization.
- Send errors to the responder channel here in the mempool's `Downloads`, only waiting for the insertion result if the transaction was successfully verified so there's no need to poll the mempool for transactions that failed verification

Update: The `Downloads` struct doesn't have a handle to the buffered mempool service so the polling is still in the transaction verifier. The rest of this comment was addressed in https://github.com/ZcashFoundation/zebra/pull/8901/commits/3dc3c1caf69df70c554289f3df227b6380d33b8d.

mpguerra:
Do we still want to do this one? Otherwise we should just close for now

arya2:
> Do we still want to do this one? Otherwise we should just close for now

We should still do it, it's ready for review now it makes it easier to debug the `sendrawtransaction` method (the branch has already been used to rule out a possible category of bug [here](https://discord.com/channels/809218587167293450/809251029673312267/1296588588716130356)).

oxarbitrage (approved):
Looks good to me, thanks.

mpguerra:
@mergify queue

PR #8908, opened 2024-10-03
ref(actions): allow more flexibility on cached states usage

gustavovalverde:
## Motivation

We've added the possibility to deploy instances with a cached state attached, but we might also want to deploy them without it, or change the cached state disk preference depending on the github source, while not affecting our actual deployment process.

## Solution

- Allow to select a cached state based on:
  - Type of cached disk to use (`tip` or `checkpoint`)
  - Source (`main` branch vs `PR` or `any` other available)
  - Network
- Allow to deploy a node without cached state
- Search for branch/PR reference instead of commit (as this is an use case that we haven't used for a while)
- Improve the disk selection logic, by allowing a preference for cached states from the `main` branch, without disallowing finding other available disks

### Tests

- Deployed to Testnet with a cached state: https://github.com/ZcashFoundation/zebra/actions/runs/11162158270
- Deployed to Testnet without a cached state: https://github.com/ZcashFoundation/zebra/actions/runs/11163638750 (get disk is skipped and instance is deployed without a cached state attached).

### Follow-up Work

Support the `matrix` use case when releasing a new Zebra version: https://github.com/ZcashFoundation/zebra/blob/2552c3c94bee3d24db7ae75ae67663400c6fcad6/.github/workflows/cd-deploy-nodes-gcp.yml#L233 

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -250,14 +273,14 @@ jobs:
...
-      # but the implementation is failing as it's requiring the disk names, contrary to what is stated in the official documentation
       - name: Create instance template for ${{ matrix.network }}
         run: |
-          NAME="zebrad-cache-${{ env.GITHUB_HEAD_REF_SLUG_URL || env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}-${NETWORK}"
-          DISK_PARAMS="name=${NAME},device-name=${NAME},size=400GB,type=pd-ssd"
+          DISK_NAME="zebrad-cache-${{ env.GITHUB_HEAD_REF_SLUG_URL || env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}-${NETWORK}"
+          DISK_PARAMS="name=${DISK_NAME},device-name=${DISK_NAME},size=400GB,type=pd-ssd"
           if [ -n "${{ env.CACHED_DISK_NAME }}" ]; then
             DISK_PARAMS+=",image=${{ env.CACHED_DISK_NAME }}"
+          elif [ ${{ !inputs.no_cached_disk && github.event_name == 'workflow_dispatch' }} ]; then
```
upbqdn: Shouldn't this be without the exclamation mark?

```suggestion
          elif [ ${{ inputs.no_cached_disk && github.event_name == 'workflow_dispatch' }} ]; then
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -326,7 +349,8 @@ jobs:
     permissions:
       contents: 'read'
       id-token: 'write'
-    if: github.event_name == 'workflow_dispatch'
+    # Run even if we don't need a cached disk, but only when triggered by a workflow_dispatch
+    if: ${{ !failure(needs.get-disk-name) && github.event_name == 'workflow_dispatch' }}
```
upbqdn: It looks like we can't use `failure` this way: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#failure.

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -363,10 +387,12 @@ jobs:
       # Create instance template from container image
       - name: Manual deploy of a single ${{ inputs.network }} instance running zebrad
         run: |
-          NAME="zebrad-cache-${{ env.GITHUB_HEAD_REF_SLUG_URL || env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}-${NETWORK}"
-          DISK_PARAMS="name=${NAME},device-name=${NAME},size=400GB,type=pd-ssd"
+          DISK_NAME="zebrad-cache-${{ env.GITHUB_HEAD_REF_SLUG_URL || env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}-${NETWORK}"
+          DISK_PARAMS="name=${DISK_NAME},device-name=${DISK_NAME},size=400GB,type=pd-ssd"
           if [ -n "${{ env.CACHED_DISK_NAME }}" ]; then
             DISK_PARAMS+=",image=${{ env.CACHED_DISK_NAME }}"
+          elif [ ${{ !inputs.no_cached_disk && github.event_name == 'workflow_dispatch' }} ]; then
```
upbqdn: Same as above

```suggestion
          elif [ ${{ inputs.no_cached_disk && github.event_name == 'workflow_dispatch' }} ]; then
```

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -20,7 +20,7 @@ echo "Extracting local state version..."
 LOCAL_STATE_VERSION=$(grep -oE "DATABASE_FORMAT_VERSION: .* [0-9]+" "${GITHUB_WORKSPACE}/zebra-state/src/constants.rs" | grep -oE "[0-9]+" | tail -n1)
 echo "STATE_VERSION: ${LOCAL_STATE_VERSION}"
 
-# Function to find a cached disk image based on the git pattern (commit, main, or any branch)
+# Function to find a cached disk image based on the git pattern (branch, main, or any branch)
```
upbqdn, line 23: ###### Optional

I think we can keep the "commit" here.

```suggestion
# Finds a cached disk image based on the given git pattern (commit, main, or any branch)
```

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -43,31 +43,29 @@ find_cached_disk_image() {
...
-        if [[ "${PREFER_MAIN_CACHED_STATE}" == "true" ]]; then
-            CACHED_DISK_NAME=$(find_cached_disk_image "main-[0-9a-f]+" "main branch")
-        # Else, try to find one from any branch
-        else
+    # Check if main branch images are preferred
+    if [[ "${PREFER_MAIN_CACHED_STATE}" == "true" ]]; then
+        CACHED_DISK_NAME=$(find_cached_disk_image "main-[0-9a-f]+" "main branch")
+    # Else, try to find a cached disk image from the current branch (or PR)
+    else
+        CACHED_DISK_NAME=$(find_cached_disk_image ".+-${GITHUB_REF}" "branch")
```
upbqdn: Shouldn't we test if `GITHUB_REF` isn't empty before we pass it to `find_cached_disk_image`?

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -3,9 +3,9 @@
 # This script finds a cached Google Cloud Compute image based on specific criteria.
 #
 # If there are multiple disks:
-# - prefer images generated from the same commit, then
+# - prefer images generated from the same branch, then
```
upbqdn:
```suggestion
# - if `PREFER_MAIN_CACHED_STATE` is "true", then select an image from the `main` branch, else
```

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -3,9 +3,9 @@
...
 # - if prefer_main_cached_state is true, prefer images from the `main` branch, then
```
upbqdn:
```suggestion
# - try to find a cached disk image from the current branch (or PR), else
```

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -3,9 +3,9 @@
...
-# - use any images from any other branch or commit.
+# - use any images from any other branch or branch.
```
upbqdn:
```suggestion
# - try to find an image from any branch.
```

.github/workflows/scripts/gcp-get-cached-disks.sh
gustavovalverde: This is no longer looking for a specific commit but a github reference, which in most cases is the PR branch, or the main branch. https://github.com/ZcashFoundation/zebra/pull/8908/files/2552c3c94bee3d24db7ae75ae67663400c6fcad6#diff-3fe6320dbec0ab834aa0e833b666416668b3538ab84b4ce0d90b817867d6d208L50

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: Seems like I was sleepy when doing this

.github/workflows/scripts/gcp-get-cached-disks.sh
gustavovalverde: I think the only benefit will be that we'd be able to `echo` an error in the logs. But if the `GITHUB_REF` is missing, then there's a bunch of other things that will also fail in other workflows, as there's not supposed to be a commit without a reference.

Unless you're thinking in other future scenarios with different ways to call this script.

upbqdn:
@gustavovalverde can you update this branch? I can't do it.

.github/workflows/sub-find-cached-disks.yml
```diff
@@ -74,20 +74,28 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2.1.1
 
+      # Performs formatting on disk name components.
+      #
       # Disk images in GCP are required to be in lowercase, but the blockchain network
-      # uses sentence case, so we need to downcase ${{ inputs.network }}
+      # uses sentence case, so we need to downcase ${{ inputs.network }}.
+      #
+      # Disk image names in GCP are limited to 63 characters, so we need to limit
```
upbqdn, line 82: Can you add a link to further details or describe the other 51 chars?

gustavovalverde:
@upbqdn this one should be ready to merge

.github/workflows/scripts/gcp-get-cached-disks.sh
```diff
@@ -36,38 +36,36 @@ find_cached_disk_image() {
...
-
-    # If no cached disk image is found
-    if [[ -z "${CACHED_DISK_NAME}" ]]; then
-        # Check if main branch images are preferred
-        if [[ "${PREFER_MAIN_CACHED_STATE}" == "true" ]]; then
-            CACHED_DISK_NAME=$(find_cached_disk_image "main-[0-9a-f]+" "main branch")
-        # Else, try to find one from any branch
-        else
+    # Check if main branch images are preferred
+    if [[ "${PREFER_MAIN_CACHED_STATE}" == "true" ]]; then
```
upbqdn, line 51: If `PREFER_MAIN_CACHED_STATE` is `true` and `find_cached_disk_image` doesn't find anything, then the script fails. Is this the intended behavior? I can open a quick PR to handle this case since I've already given it a thought.

upbqdn: - This is what I had in mind: #8936.

.github/workflows/sub-find-cached-disks.yml
gustavovalverde: Solved with https://github.com/ZcashFoundation/zebra/pull/8908/commits/ba68afde7ccbe46450d6ea092e5546e0889e4a36

.github/workflows/scripts/gcp-get-cached-disks.sh
gustavovalverde: Solved with https://github.com/ZcashFoundation/zebra/pull/8908/commits/50cd029a75d4e8dad88e76b3fc233a173b35e0cd

upbqdn (approved):

PR #8921, opened 2024-10-10
Suggestion for "feat(regtest): Add regtest halving interval and port test"

arya2:
This is a suggestion PR for #8888 that addresses [this comment](https://github.com/ZcashFoundation/zebra/pull/8888#discussion_r1788342311) by:
- Clearing the Regtest funding streams,
- Panicking when there's an attempt to set Testnet parameters in the `ParametersBuilder` that affect the funding stream address period after setting configured funding streams, and
- Checking that the funding streams in a `ParametersBuilder` are valid for the derived network's funding stream address period.

@oxarbitrage can review and optionally merge it into their PR.

### DIFF
--- a/zebra-chain/src/parameters/network/testnet.rs
+++ b/zebra-chain/src/parameters/network/testnet.rs
@@ -15,9 +15,8 @@ use crate::{
 use super::{
     magic::Magic,
     subsidy::{
-        FundingStreamReceiver, FundingStreamRecipient, FundingStreams, ParameterSubsidy,
-        BLOSSOM_POW_TARGET_SPACING_RATIO, FIRST_HALVING_TESTNET,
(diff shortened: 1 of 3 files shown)


zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -231,6 +217,9 @@ pub struct ParametersBuilder {
     pre_nu6_funding_streams: FundingStreams,
     /// Post-NU6 funding streams for this network
     post_nu6_funding_streams: FundingStreams,
+    /// A flag indicating whether to allow changes to fields that affect
+    /// the funding stream address period.
+    should_lock_funding_stream_address_period: bool,
```
arya2, line 222: This field and the related checks aren't strictly necessary if we want `ParametersBuilder` to be more flexible, it could be useful for debugging otherwise.

oxarbitrage (approved):

PR #8923, opened 2024-10-10
ci: Refactor Dockerfile & entrypoint

### LINKED ISSUE #9210: Remove CI functionality related to experimental features
## Rationale

We don't need or use any experimental features anymore.

upbqdn:
## Motivation

CI contains a bunch of nits:

- The entrypoint script quietly uses its own Zebra config if the user passes an invalid path to the config. This causes some of our CI tests to pass, even though they shouldn't since they're passing non-existent configs to Zebra in Docker.
- Some tests for the scanner don't run.
- The suite for all unit tests runs twice unnecessarily.
- We no longer have experimental features in Zebra, but we still have a failing workflow that tries to build a Docker image with experimental features.
- Zebra runs under `root` inside Docker, which is a security risk.
- The `runtime` Docker stage for production images contains unneeded packages, making the images larger than necessary and increasing the surface for security vulnerabilities.
- The `deps` Docker stage for tests contains unneeded packages, which slows CI down.
- The entrypoint script contains unused variables and is rather complex.
- The `EXPOSE` instruction contains two ports, but Zebra uses only one of them at a time, and it doesn't contain some other ports Zebra uses.
- Rust features are passed from CI to Dockerfile and entrypoint in many variables. In the Dockerfile, we currently have:
    - `FEATURES` (which comes from `RUST_PROD_FEATURES` in CI)
    - `TEST_FEATURES` (which comes from `RUST_TEST_FEATURES` in CI)
    - `EXPERIMENTAL_FEATURES` (which should probably come from `RUST_EXPERIMENTAL_FEATURES` in CI, containing `"shielded-scan"`, but it doesn't, so the Dockerfile uses the default value `"journald prometheus filter-reload"`)
    - `ENTRYPOINT_FEATURES = "$FEATURES $TEST_FEATURES"`
    - `ENTRYPOINT_FEATURES_EXPERIMENTAL = $ENTRYPOINT_FEATURES $EXPERIMENTAL_FEATURES`
- CI uses some vars that don't exist in the entrypoint.
- The RPC server is enabled by default in our production Docker images, even though Zebra doesn't enable it by default.

Close #9210.

## Solution

- Fix the tests for the scanner.
- Don't run all unit tests twice.
- Use only one variable to pass Rust features from CI to Dockerfile and entrypoint.
- Remove the workflow for building Docker images with experimental features.
- Create a non-privileged system user in the `runtime` Docker stage and switch to it.
- Don't use `gosu`.
- Remove all packages from the `runtime` stage.
- Remove unneeded packages from the `deps` stage
- Fix some malfunctioning CI tests.
- Don't use the `EXPOSE` instruction in Docker.
- Bump the Rust version in Dockerfile.
- Change the location of the entrypoint in Docker images from `/etc/zebrad` to `/usr/local/bin`.
- Refactor the structure of the entrypoint; remove redundant env vars, and add docs.
- Explicitly specify the location of the conf and cache dirs in Docker according to https://specifications.freedesktop.org/basedir-spec/latest/.
- Prepare the Dockerfile and entrypoint files for deploying a Testnet mining instance.
- Rename the `ZEBRA_CACHED_STATE_DIR` env var to `ZEBRA_CACHE_DIR` since that dir no longer contains only the state but also the network cache and the cookie file.
- Don't enable the RPC server by default. This is a significant breaking change for users.
- Refactor how we configure Zebra inside Docker. This is a breaking change.

## Tests

- Manually test that `zebrad` runs under the new `zebra` user:

  Running
  ```bash
  docker build -f docker/Dockerfile --target runtime -t zebra:local .
  docker run -d --rm --name zebra_local zebra:local
  docker exec -it -u root zebra_local bash
  apt-get update && apt-get install -y procps
  ps aux
  ```
  displays
  ```
  USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
  zebra          1 86.3  2.7 6605720 2691512 ?     Ssl  09:49  31:03 zebrad -c /etc/zebrad/zebrad.toml
  root         150  0.0  0.0   4188  3368 pts/0    Ss   10:23   0:00 bash
  root         438  0.0  0.0   8088  4044 pts/0    R+   10:25   0:00 ps aux
  ```

### PR Checklist

- [x] The PR name is suitable for the change log.
- [x] The solution is tested.


gustavovalverde:
@upbqdn can the _motivation_ be expanded/updated here? For future reference, it might be confusing if someone looks at the PR and understand all these changes were required to fix the use of the `root` user.

docker/Dockerfile
```diff
@@ -161,68 +162,48 @@ RUN --mount=type=bind,source=tower-batch-control,target=tower-batch-control \
...
-
-RUN addgroup --system --gid ${GID} ${USER} \
-    && adduser \
-    --system \
-    --disabled-login \
-    --shell /bin/bash \
-    --home ${APP_HOME} \
-    --uid "${UID}" \
-    --gid "${GID}" \
-    ${USER}
```
gustavovalverde: There might be instances where a user would like to (re-)build the image with a custom UID:GUID, as they might require to mount files from their host, which will be incompatible with the UID:GUID of the container user, and it's also a good Docker practice to specify the UID for the default user for other edge-cases.

Unless there's a good reason to remove this, I'd suggest to keep it.

The PR that added this has some references to the reasoning behind it: https://github.com/ZcashFoundation/zebra/pull/8803#issue-2487686719

docker/Dockerfile
```diff
@@ -161,68 +162,48 @@ RUN --mount=type=bind,source=tower-batch-control,target=tower-batch-control \
...
-# To save space, this step starts from scratch using debian, and only adds the resulting
-# binary from the `release` stage
+# This step starts from scratch using Debian and only adds the resulting binary
+# from the `release` stage.
 FROM debian:bookworm-slim AS runtime
 
-# Set the default path for the zebrad binary
-ARG APP_HOME
-ENV APP_HOME=${APP_HOME}
-WORKDIR ${APP_HOME}
```
gustavovalverde, line 176: We should set a `WORKDIR` otherwise the user will end up in the `/` directory with no permissions, which they might require for testing/personal purposes.

The `ARG + ENV` combination also allows the user to set a custom directory, in case their host permissions does not allow the one we've chosen.

docker/Dockerfile
```diff
@@ -161,68 +162,48 @@ RUN --mount=type=bind,source=tower-batch-control,target=tower-batch-control \
...
-    --disabled-login \
-    --shell /bin/bash \
-    --home ${APP_HOME} \
-    --uid "${UID}" \
-    --gid "${GID}" \
-    ${USER}
-
-# Config settings for zebrad
-ARG FEATURES
-ENV FEATURES=${FEATURES}
```
gustavovalverde, line 209: If the `runtime` stage is built with custom `FEATURES` this will be propagated by default as an `ENV` variable to the `entrypoint.sh`, if we remove this instructions, then we need to always specify the `FEATURES` environment variable when running the image, otherwise it will be empty.

docker/Dockerfile
```diff
@@ -161,68 +162,48 @@ RUN --mount=type=bind,source=tower-batch-control,target=tower-batch-control \
...
 
-# Path and name of the config file
-# These are set to a default value when not defined in the environment
-ENV ZEBRA_CONF_DIR=${ZEBRA_CONF_DIR:-/etc/zebrad}
-ENV ZEBRA_CONF_FILE=${ZEBRA_CONF_FILE:-zebrad.toml}
+RUN adduser --system ${USER}
 
-RUN mkdir -p ${ZEBRA_CONF_DIR} && chown ${UID}:${UID} ${ZEBRA_CONF_DIR} \
-    && chown ${UID}:${UID} ${APP_HOME}
```
gustavovalverde, line 217: Having a `HOME` directory for the application is a good practice, it also starts the container in an empty directory the users can use as they see fit.

docker/Dockerfile
```diff
@@ -161,68 +162,48 @@ RUN --mount=type=bind,source=tower-batch-control,target=tower-batch-control \
...
 
-COPY --from=release /usr/local/bin/zebrad /usr/local/bin
-COPY --from=release /etc/zebrad/entrypoint.sh /etc/zebrad
+# TODO: Shorten `ZEBRA_CACHED_STATE_DIR` to `ZEBRA_CACHE_DIR`
+ARG ZEBRA_CACHED_STATE_DIR="/var/cache/zebrad"
+ENV ZEBRA_CACHED_STATE_DIR=${ZEBRA_CACHED_STATE_DIR}
+RUN mkdir -p ${ZEBRA_CACHED_STATE_DIR} && chown ${USER} ${ZEBRA_CACHED_STATE_DIR}
 
-# Expose configured ports
-EXPOSE 8233 18233
```
gustavovalverde, line 223: This is meant for documentation purposes https://docs.docker.com/reference/dockerfile/#expose

docker/entrypoint.sh
```diff
@@ -195,171 +181,204 @@ run_cargo_test() {
...
+# - TODO: If "$1" is "monitoring", start a monitoring node.
+# - If "$1" doesn't match any of the above, run "$@" directly.
+case "$1" in
+--* | -* | zebrad)
+  shift
+
+  prepare_env_vars
+
+  if [[ ! -f "${ZEBRA_CONF_PATH}" ]] && [[ -d "${ZEBRA_CONF_DIR}" ]]; then
+    ZEBRA_CONF_PATH="${ZEBRA_CONF_DIR}/zebrad.toml"
```
gustavovalverde: This default should be set in the Dockerfile, to keep it as the single source of truth for default variables.

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -151,19 +151,6 @@ jobs:
...
-    # Make sure Zebra can sync the genesis block on testnet
-    uses: ./.github/workflows/sub-test-zebra-config.yml
-    with:
-      test_id: 'testnet-conf'
-      docker_image: ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
-      grep_patterns: '-e "net.*=.*Test.*estimated progress to chain tip.*Genesis" -e "net.*=.*Test.*estimated progress to chain tip.*BeforeOverwinter"'
-      # TODO: improve the entrypoint to avoid using `ENTRYPOINT_FEATURES=""`
-      test_variables: '-e NETWORK -e ZEBRA_CONF_PATH="/etc/zebrad/zebrad.toml" -e ENTRYPOINT_FEATURES=""'
-      network: 'Testnet'
-
```
gustavovalverde: This test was created to confirm that any change we do in CI or in Docker doesn't affect the ability to read the proper `$NETWORK` environment variable. As it had happened before that some changes breaks this behavior, and then the tests are running in `Mainnet` instead of `Testnet`, but we realized too late or had to wait for some tests to run to confirm it.

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -163,18 +163,6 @@ jobs:
...
-    name: Test CD custom Docker config file
-    needs: build
-    uses: ./.github/workflows/sub-test-zebra-config.yml
-    with:
-      test_id: 'custom-conf'
-      docker_image: ${{ vars.GAR_BASE }}/zebrad@${{ needs.build.outputs.image_digest }}
-      grep_patterns: '-e "loaded zebrad config.*config_path.*=.*v1.0.0-rc.2.toml"'
-      test_variables: '-e NETWORK -e ZEBRA_CONF_PATH="zebrad/tests/common/configs/v1.0.0-rc.2.toml"'
-      network: ${{ inputs.network || vars.ZCASH_NETWORK }}
-
```
gustavovalverde: Although very simple, the objective of this test is to confirm that the `entrypoint` is able to handle a custom configuration path (`$ZEBRA_CONF_PATH`), run with it and confirm that path is being used.

This could be extended to validate is running with a mounted file using `--mount type=bind,source="$(pwd)"/target,target=/app` as part of the `test_variables`, but that's out of scope.

gustavovalverde (changes requested):
I haven't fully review the `entrypoint.sh` as some of these requested changes might have a slight impact there. Changes to CI files—unless they're related to the Docker changes—should be in a different PR that depends on this one.

docker/Dockerfile
gustavovalverde: image

docker/entrypoint.sh
upbqdn: Which var do you have in mind?

.github/workflows/cd-deploy-nodes-gcp.yml
upbqdn: The test was using a custom config file set in `test_variables`. However, the file was not included in the Docker image, and the entrypoint script created a new, default one under the original file's path. Zebra then loaded this new file, and the test passed because the pattern in `grep_patterns` matched Zebra's output containing the original path, even though the config file was different.

upbqdn: The test fails in this PR due to the fixes in the entrypoint.

docker/Dockerfile
upbqdn: Note that we're creating a system user. Those users should have UIDs in [1, 999]. 

In Debian, the docs for `adduser` say that the default dynamic range for system UIDs is defined by [FIRST_SYSTEM_UID, LAST_SYSTEM_UID], which is [100, 999] in `etc/adduser.conf`: https://manpages.debian.org/bullseye/adduser/adduser.8.en.html 

Moreover, the Linux spec says that system UIDs in [100, 499] should be reserved for dynamic allocation: https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/uidrange.html

I'm surprised the Docker docs don't mention this.

`adduser` automatically checks if a UID is available before assigning one from the right range. Since I wasn't sure what UID to pick, and no user brought it up, I removed it to avoid scope creep. I'd prefer to address this issue according to the spec once a user brings it up.

docker/Dockerfile
upbqdn: System users should have no home dir and should not even be able to log in to the machine. Our users should use `-u root` when they're logging to the machine. The logic I had in mind is very simple and minimal:

- The whole `runtime` target has its entrypoint executed under the non-privileged `zebra` system user which has no home dir and no login (just as in a regular Linux env).
- The Dockerfile sets up the minimal requirements for the `zebra` user to execute the entrypoint.
- When our user wants to do some tweaks, they explicitly use `-u root`, which they can always do, and which gives them the clarity that they have full privileges.

I wanted to describe this in our docs in a subsequent PR. Is it OK if we do it like that?

docker/Dockerfile
upbqdn: The problem is that Zebra always binds exactly one of these ports. I removed it because there was no way to specify only one of them at build time, and "exposing" both seemed confusing, even in docs. I'd deal with it when a user mentions it to keep the scope tight again.

docker/Dockerfile
upbqdn: System users should have no home dir. Our users should go with `-u root`. Another approach would be to execute the entrypoint under `root`, and then run `zebrad` under the `zebra` system user. Our users could then login implicitly as `root` without `-u root`, but that adds a bit of extra complexity to the entrypoint, which I wanted to keep simple for now.

docker/Dockerfile
upbqdn: Yes, but that var is not used in any meaningful way in `entrypoint.sh`, so I removed it. I would like to refactor the way we configure Zebra inside Docker in follow-up PRs.

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn: This test is failing with this PR, similarly to the other one. I have a better approach in mind, which I didn't do in this PR. Let's add it back in a subsequent PR?

Moreover, there seem to be some parts that are bugs or hard to understand. For example, I couldn't figure out what `-e NETWORK` in `test_variables` is supposed to do. Also, setting `ENTRYPOINT_FEATURES` to the empty string to enable the test in the entrypoint makes it very hard to follow the execution path in the whole pipeline.

docker/Dockerfile
upbqdn: Zebra also optionally binds two more ports for RPCs (or rather, only one of them), and other ports for other services. We don't mention that here, which adds to the confusion.

upbqdn:
> @upbqdn can the motivation be expanded/updated here? For future reference, it might be confusing if someone looks at the PR and understand all these changes were required to fix the use of the root user.

I updated the PR description.

upbqdn:
My yml linter updated the formatting of the `cd-deploy-nodes-gcp.yml` file when merging `main`. I'm happy to revert those changes if they don't fit.

docker/Dockerfile
gustavovalverde: Unfortunately, this is not clear in Docker's documentation. But some of their tools, `docker init` for example, output the recommended approach, which is the following (with `alpine` linux):

```bash
ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/nonexistent" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser
USER appuser
```

In any case, to match their recommendation in `debian`, this would change to:
```
RUN addgroup --system --gid ${GID} ${USER} \
    && useradd \
    --uid "${UID}" \
    --no-create-home \
    --home-dir /nonexistent \
    --shell /usr/sbin/nologin \
    --comment "" \
    appuser
```

They **main** reasons to keep this instruction are:
- **Host and Container Alignment:** When you mount host directories into the container, file permissions are based on `UID` and `GID`. If the container's user ID match those on the host, you avoid permission issues.
  - I've experienced this myself when I was testing the docker-compose and mounting a cached state from my host into the container, which had different UID:GID, and the container was not able to write to it.

- **Predictable IDs**: By specifying the UID and GID, you ensure that the user inside the container has the same IDs across different builds, deployments, and host systems.

Additionally, this is very common in Docker, and plenty of projects use this approach to deal with these and other use-cases: https://github.com/search?q=%22useradd%22+%22--uid%22++language%3ADockerfile&type=code

docker/Dockerfile
gustavovalverde: It's used here:
https://github.com/ZcashFoundation/zebra/pull/8923/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7caR81
and here:
https://github.com/ZcashFoundation/zebra/pull/8923/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7caR100

From[ our documentation](https://github.com/ZcashFoundation/zebra/blob/46c6b6eb38617c07e8a130650b33831759b73263/book/src/user/docker.md?plain=1#L59), if we suggest:
```shell
docker build -f ./docker/Dockerfile --target runtime --build-arg FEATURES='default-release-binaries prometheus' --tag local/zebra.mining:latest .
```

The `entrypoint` will evaluate the `$FEATURES` var and it will be empty, as it was never defined as a variable. This would be confusing for the user as they built it with `--build-arg FEATURES='default-release-binaries prometheus'` but the configuration file is not adding the corresponding section, as that argument was not passed as an Environment variable (at build time) to the container.

docker/Dockerfile
gustavovalverde: If we're not allowing the user to log in with the user running Zebra, then we should document the approach they can use to get a `bash/sh` terminal in the container to troubleshoot in it.

docker/entrypoint.sh
gustavovalverde: We can use the var you defined: `ZEBRA_CONF_PATH`. But have the default value defined in the Dockerfile, as we do here: https://github.com/ZcashFoundation/zebra/blob/afeb05f50b5edd3bff29870d4d3c5a08e4bb9d9a/docker/Dockerfile#L189

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: `-e NETWORK` tells docker to use whatever value the `$NETWORK` env variable is set to, to override any default that was set at build time, or in the Dockerfile. This happens here: https://github.com/ZcashFoundation/zebra/blob/46c6b6eb38617c07e8a130650b33831759b73263/.github/workflows/sub-test-zebra-config.yml#L90-L91 and makes Zebra run with a Testnet configuration, and the test validates that Zebra is correctly running with it.

> This has saved me (and others) from making mistakes multiple times while making CI refactors, so this is very important under those circumstances.

I do agree that setting the `ENTRYPOINT_FEATURES` to an empty string is a dirty hack to make this work, but that's a tech debt that wouldn't justify removing the whole test. In any case, we can remove the use of the `ENTRYPOINT_FEATURES` variables, while keeping this test behavior.

I'd suggest commenting this and adding a TODO in top of it, or creating an issue, instead of removing the test. Just so we don't forget later on, as this is an important validation.

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: I'd suggest commenting this and adding a TODO in top of it, or creating an issue. Just so we don't forget later on.

Either of them I'd suggest indicating something like:

> We need to create a test that validates we can mount a configuration file to a different path that the default used by `ZEBRA_CONF_PATH`, and that Zebra runs using this new file and path.

mpguerra:
Do we want to do anything further here? Otherwise we should either merge or close without merging

gustavovalverde:
> Do we want to do anything further here? Otherwise we should either merge or close without merging

I added some review comments which are pending for a reply, and we should also consider the latest interaction we had with some users in Discord, as some changes related to permission handling, mounting a configuration file, and the use of `curl`, etc, should be considered so we don't break those use-cases.

upbqdn:
I'm planning to address the comments and get the PR approved.

docker/Dockerfile
upbqdn: Addressed in d842ff3108769acc622860fe7d597ae956739c37.

docker/Dockerfile
upbqdn: I added a home dir in d842ff3108769acc622860fe7d597ae956739c37 in the end.

docker/Dockerfile
upbqdn: I added a home dir in d842ff3108769acc622860fe7d597ae956739c37.

docker/Dockerfile
upbqdn: Addressed in e0e204867fddb118f439ec80c58e411485313830.

docker/Dockerfile
upbqdn: Addressed in e2abaf40787f430dd3f93b6ab8cbbca0fa7a7863.

docker/entrypoint.sh
upbqdn: That wouldn't work with the implementation in this PR. If `ZEBRA_CONF_PATH` is set before the entrypoint executes, the entrypoint will pass that path to Zebra. If it is not set, the entrypoint will create its own default config to be passed to Zebra.

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn: > `-e NETWORK` tells docker to use whatever value the `$NETWORK` env variable is set to, to override any default that was set at build time, or in the Dockerfile.

I must not understand the syntax of the `-e` flag then. Shouldn't it be something like `'-e NETWORK="Testnet"'`? The flag is used to set env vars, but what are we setting here? In the enrypoint, the default value is set here: https://github.com/zcashfoundation/zebra/blob/e2abaf40787f430dd3f93b6ab8cbbca0fa7a7863/docker/entrypoint.sh#L23, and the var is used, for example, here https://github.com/zcashfoundation/zebra/blob/e2abaf40787f430dd3f93b6ab8cbbca0fa7a7863/docker/entrypoint.sh#L73-L74. We override the default value here: https://github.com/zcashfoundation/zebra/blob/e2abaf40787f430dd3f93b6ab8cbbca0fa7a7863/.github/workflows/sub-test-zebra-config.yml#L90-L91. So the entrypoint either takes the pre-set value of `NETWORK` from its environment or implicitly sets the var to `"Mainnet"`. To me, passing `'-e NETWORK'` to `docker run` seems void: https://github.com/ZcashFoundation/zebra/actions/runs/12991342019/job/36229077857#step:5:3.

upbqdn: To clarify, the reason why the test currently passes on `main` is that the entrypoint actually ignores the value of `ZEBRA_CONF_PATH`. We can see it by running

```bash
docker build -f docker/Dockerfile --target runtime -t zebra:local .
docker run --rm -e ZEBRA_CONF_PATH="foobar" -e NETWORK="Testnet" zebra:local
```

This works fine, but it shouldn't because we're setting the conf path to `foobar`. In the test, we set the path to `/etc/zebrad/zebrad.toml`, which doesn't exist anywhere, neither in the image nor in the repo, but the test still passes.

We never noticed because the entrypoint is currently too convoluted, and it took me a long time to find out why something like this works. This PR makes the test fail, which should be the correct behavior. I would even suggest we completely remove all "optional" ways for wiggling with any non-essential configs and keep it very simple at the start. We can then add more functionality whenever someone asks for it and document it well.

upbqdn: I commented it out in e6e018788edeff2ab31ddacfc059fdfe278ab4f3.

upbqdn:
@gustavovalverde, this PR is ready for another review.

#### Things to note

- The PR doesn't remove any CI checks anymore.
- All conversations above are either addressed or outdated.
- There might be some missing docs updates.

gustavovalverde (approved):
All my comments have been addressed. LGTM!

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn: I fixed and updated this test in the meantime, so it should work as expected.

PR #8930, opened 2024-10-11
Network Sustainability Mechanism - ZIP 233 implementation

### LINKED ISSUE #9504: Implement ZIP 233: Network Sustainability Mechanism: Burning
### References

- [ZIP 233: Network Sustainability Mechanism: Burning](https://github.com/zcash/zips/blob/main/zip-0233.rst)

mariopil:
## Motivation

This PR implements the ZIP-233, described [here](https://github.com/zcash/zips/pull/913).

[ZIP-233](https://github.com/zcash/zips/pull/913)

## Solution

A new field, `zip233_amount,` is added to the v6 transaction version according to the specification in ZIP-233. 

This implementation uses the `nsm` branch of the librustzcash library.

### Tests

For testing purposes, a local testnet network was set up. It contained two nodes—a zebra and a zcashd one (with the ZIP-233 implemented, too, PR [here](https://github.com/zcash/zcash/pull/6969)). A transaction in a new format containing the `zip233_amount` field was created and added to the block by zcashd. The zebra node synced with zcashd without any errors and accepted the block.

### Follow-up Work

The next PR will contain an implementation of zips [ZIP-234](https://github.com/zcash/zips/pull/914) and [ZIP-235](https://github.com/zcash/zips/pull/915).

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -555,13 +555,22 @@ where
             // Get the `value_balance` to calculate the transaction fee.
             let value_balance = tx.value_balance(&spent_utxos);
 
+            let zip233_amount = match *tx {
+            	#[cfg(feature="tx_v6")]
+                Transaction::V6{ .. } => tx.zip233_amount(),
+                _ => Amount::zero()
+            };
+
             // Calculate the fee only for non-coinbase transactions.
             let mut miner_fee = None;
             if !tx.is_coinbase() {
                 // TODO: deduplicate this code with remaining_transaction_value()?
                 miner_fee = match value_balance {
                     Ok(vb) => match vb.remaining_transaction_value() {
-                        Ok(tx_rtv) => Some(tx_rtv),
+                        Ok(tx_rtv) => match tx_rtv - zip233_amount {
+                            Ok(fee) => Some(fee),
+                            Err(_) => return Err(TransactionError::IncorrectFee),
+                        }
                         Err(_) => return Err(TransactionError::IncorrectFee),
                     },
                     Err(_) => return Err(TransactionError::IncorrectFee),
--- a/zebra-consensus/src/transaction/check.rs
+++ b/zebra-consensus/src/transaction/check.rs
@@ -122,9 +122,15 @@ pub fn lock_time_has_passed(
 ///
 /// This check counts both `Coinbase` and `PrevOut` transparent inputs.
 pub fn has_inputs_and_outputs(tx: &Transaction) -> Result<(), TransactionError> {
+    #[cfg(feature = "tx_v6")]
+    let has_other_circulation_effects = tx.has_zip233_amount();
+
+    #[cfg(not(feature = "tx_v6"))]
+    let has_other_circulation_effects = false;
+
     if !tx.has_transparent_or_shielded_inputs() {
         Err(TransactionError::NoInputs)
-    } else if !tx.has_transparent_or_shielded_outputs() {
+    } else if !tx.has_transparent_or_shielded_outputs() && !has_other_circulation_effects {
         Err(TransactionError::NoOutputs)
     } else {
         Ok(())
--- a/zebra-consensus/src/transaction/tests/prop.rs
+++ b/zebra-consensus/src/transaction/tests/prop.rs
@@ -303,6 +303,9 @@ fn mock_transparent_transaction(
     // Create the mock transaction
     let expiry_height = block_height;
 
+    #[cfg(feature = "tx_v6")]
+    let zip233_amount = Amount::zero();
+
     let transaction = match transaction_version {
         4 => Transaction::V4 {
             inputs,
@@ -317,6 +320,17 @@ fn mock_transparent_transaction(
             outputs,
             lock_time,
             expiry_height,
+            sapling_shielded_data: None,
+            orchard_shielded_data: None,
+            network_upgrade,
+        },
+        #[cfg(feature = "tx_v6")]
+        6 => Transaction::V6 {
+            inputs,
+            outputs,
+            lock_time,
+            expiry_height,
+            zip233_amount,
             sapling_shielded_data: None,
             orchard_shielded_data: None,
             network_upgrade,
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -3517,6 +3517,8 @@ async fn nu6_funding_streams_and_coinbase_balance() -> Result<()> {
         &[],
         chain_history_root,
         vec![],
+        #[cfg(feature = "tx_v6")]
+        None,
     )
     .expect("coinbase transaction should be valid under the given parameters");
 
@@ -3575,6 +3577,8 @@ async fn nu6_funding_streams_and_coinbase_balance() -> Result<()> {
         &[],
         chain_history_root,
         vec![],
+        #[cfg(feature = "tx_v6")]
+        None,
     )
     .expect("coinbase transaction should be valid under the given parameters");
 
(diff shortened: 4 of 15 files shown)


jackgavigan:
@mariopil Does this PR need to be refreshed when the NSM ZIPs are finalized?

mariopil:
> @mariopil Does this PR need to be refreshed when the NSM ZIPs are finalized?

Yes, the feature flag will be removed here.

zebra-rpc/src/methods/tests/snapshots/get_blockchain_info@mainnet_10.snap
```diff
@@ -86,6 +86,11 @@ expression: info
       "name": "NU6",
       "activationheight": 2726400,
       "status": "pending"
+    },
+    "ffffffff": {
```
aphelionz: Will we remember to update this placeholder hex when the time comes?

zebra-rpc/src/methods/tests/snapshots/get_blockchain_info@testnet_10.snap
```diff
@@ -86,6 +86,11 @@ expression: info
       "name": "NU6",
       "activationheight": 2976000,
       "status": "pending"
+    },
+    "ffffffff": {
```
aphelionz: Same question as above

zebra-state/Cargo.toml
```diff
@@ -16,6 +16,8 @@ categories = ["asynchronous", "caching", "cryptography::cryptocurrencies"]
 
 [features]
 
+default = []
```
aphelionz: Is this unneccessary?

shielded-nate:
Note: This PR depends on https://github.com/zcash/librustzcash/pull/1567 being merged in `librustzcash`, and then the `Cargo.toml` dependency metadata here needs to be updated from referring to that PR's git branch to the publicly released Cargo versions.

zebra-consensus/src/transaction.rs
```diff
@@ -554,17 +565,20 @@ where
...
-                miner_fee = match value_balance {
-                    Ok(vb) => match vb.remaining_transaction_value() {
-                        Ok(tx_rtv) => Some(tx_rtv),
-                        Err(_) => return Err(TransactionError::IncorrectFee),
-                    },
-                    Err(_) => return Err(TransactionError::IncorrectFee),
-                };
+                miner_fee = value_balance
+                    .map(|vb| vb.remaining_transaction_value())
+                    .map(|tx_rtv| tx_rtv - zip233_amount)
```
aphelionz: One thing I'm not seeing that I could have missed is a test that exercises the logic here - do we have this anywhere?

zebra-state/Cargo.toml
mariopil: Might be some left-over from previous implementation when feature flags were used.

zebra-rpc/src/methods/tests/snapshots/get_blockchain_info@mainnet_10.snap
mariopil: As I am aware the snap files are auto-generated. Once the code is updated, this should be too.

zebra-consensus/src/transaction.rs
mariopil: There is a `nu7_nsm_transactions()` test which calls the function above.

aphelionz: Ah, great. Thanks.

aphelionz:
@str4d Would it be useful to make a PR containing ONLY the Nu7 and V6 enum handling, without adding any additional functionality? I believe it would make this PR and the other NSM PR _way_ smaller.

mariopil:
Should those changes be guarded by similar [feature](https://github.com/zcash/librustzcash/pull/1718) flags like the [ZIP-233 librustzcash PR](https://github.com/zcash/librustzcash/pull/1567) @daira @nuttycom ?

aphelionz:
Perhaps @arya2 or @oxarbitrage has some thoughts here as well?

oxarbitrage:
> Perhaps @arya2 or @oxarbitrage has some thoughts here as well?

Will be good to rebase this PR with `main`. Right now the diff includes code from https://github.com/ZcashFoundation/zebra/pull/9256 that have already landed, which makes the review harder. Rebasing against the latest main will shrink the diff to only your changes and simplify the review. Thanks!

mariopil:
The PR has been rebased to the latest main.

zebra-consensus/src/transaction/check.rs
```diff
@@ -122,9 +122,15 @@ pub fn lock_time_has_passed(
 ///
 /// This check counts both `Coinbase` and `PrevOut` transparent inputs.
 pub fn has_inputs_and_outputs(tx: &Transaction) -> Result<(), TransactionError> {
+    #[cfg(feature = "tx_v6")]
+    let has_other_outputs = tx.has_zip233_amount();
```
aphelionz: **nitpick:** this variable name threw me off for a second because I mistakenly thought it meant a proper transaction output. `has_other_circulation_effects` or something? Not a blocker for merge, just wanted to note it.

aphelionz:
I like it! Thanks, @mariopil.

Final white-glove check against the ZIP:
- adds the `zip233_amount` field to v6 transactions
- the amount is removed from circulation while still honoring the inputs == outputs rule
- serializes and sends things over to `librustzcash` for SigHash stuff

Speaking of librustzcash, I think we now need to focus on https://github.com/zcash/librustzcash/pull/1567 merged, right?

mpguerra:
@mariopil can you please resolve the conflicts and then we'll review

mariopil:
I've just rebased the PR

zebra-chain/src/parameters/transaction.rs
```diff
@@ -11,3 +11,8 @@ pub const SAPLING_VERSION_GROUP_ID: u32 = 0x892F_2085;
 /// Orchard transactions must use transaction version 5 and this version
 /// group ID. Sapling transactions can use v4 or v5 transactions.
 pub const TX_V5_VERSION_GROUP_ID: u32 = 0x26A7_270A;
+
+// TODO: Change this
```
upbqdn:
```suggestion
// TODO: Update the id once it's known. 
```

Cargo.toml
```diff
@@ -25,15 +25,16 @@ resolver = "2"
...
+# TODO: Revert to a release once librustzcash is released (#8749).
+zcash_address = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_client_backend = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_encoding = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_history = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_keys = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_primitives = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm"}
+zcash_proofs = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_transparent = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
+zcash_protocol = { git = "https://github.com/ShieldedLabs/librustzcash/", branch = "nsm" }
```
upbqdn: Are these still needed?

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -145,6 +145,52 @@ impl<'a> zp_tx::Authorization for PrecomputedAuth<'a> {
 
 // End of (mostly) copied code
 
+impl TryFrom<&Transaction> for zp_tx::Transaction {
```
upbqdn: Where do we need this?

upbqdn (commented):
Thanks for the PR. I left a few questions. Could you update the Solution section in the description? Some parts seem outdated.

Cargo.toml
mariopil: Yes, this PR is related with the ZIP-233 changes done in librustzcash. Until the librustzcash [PR](https://github.com/zcash/librustzcash/pull/1567) is merged, all its dependencies must be accessed via the "nsm" branch.

zebra-chain/src/primitives/zcash_primitives.rs
mariopil: Hmm, that could be some leftover after rebasing. Seems that code was originally removed in February. I'll remove it.

upbqdn: Great, thank you.

mpguerra:
blocked on https://github.com/zcash/librustzcash/pull/1567 for now

zebrad/tests/acceptance.rs
```diff
@@ -3603,6 +3607,162 @@ async fn nu6_funding_streams_and_coinbase_balance() -> Result<()> {
...
+
+    tracing::info!("running nu7_nsm_transactions test");
+
+    let base_network_params = testnet::Parameters::build()
+        // Regtest genesis hash
+        .with_genesis_hash("029f11d80ef9765602235e1bc9727e3eb6ba20839319f761fee920d63401e327")
+        .with_target_difficulty_limit(U256::from_big_endian(&[0x0f; 32]))
+        .with_disable_pow(true)
+        .with_slow_start_interval(Height::MIN)
+        .with_activation_heights(ConfiguredActivationHeights {
```
arya2, line 3670: The `ParametersBuilder` returned by `testnet::Parameters::build()` initially has all of the default Testnet values, but the deferred pool balance may be less on a configured Testnet at NU6.1 activation than it would be on the default Testnet. For this Testnet, the deferred pool balance would be zero when NU6.1 activates, so the lockbox disbursements must be zero as well.

Using Regtest instead could work too.


```suggestion
        .with_lockbox_disbursements(vec![])
        .with_activation_heights(ConfiguredActivationHeights {
```

Cargo.toml
```diff
@@ -221,3 +221,18 @@ panic = "abort"
...
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+# zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "77fee523dc68d5d43f91d47aeb6cfbd40668f698" }
+# END zcash/librustzcash
```
conradoplg: Have those been incorporated into the upstream crates? We usually avoid using git dependencies because that blocks us from making releases.

nuttycom: `zcash_protocol-0.6.0` and later have the `Nu7` feature available when using the `zcash_unstable=nu7` config flag: https://github.com/zcash/librustzcash/blob/11aca78ffabb78183f63edb5ab37aa9c17d8f5e7/components/zcash_protocol/src/consensus.rs#L445-L446

nuttycom: So based on https://github.com/ZcashFoundation/zebra/pull/8930/files/b375c57b084f78ad28b9c6ac9de49221b380027b#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L33 I don't see why these patch dependencies are needed at all.

mariopil: The last released `zcash_primitives` (version 0.24) doesn't contain the v6 methods, which are needed for NSM and are available on the librustzcash main, e.g.:
```rust
 pub fn read<R: Read>(reader: R, consensus_branch_id: BranchId) -> io::Result<Self> {
        let mut reader = HashReader::new(reader);

        let version = TxVersion::read(&mut reader)?;
        match version {
            TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 => {
                Self::read_v4(reader, version, consensus_branch_id)
            }
            TxVersion::V5 => Self::read_v5(reader.into_base_reader(), version),
            #[cfg(zcash_unstable = "nu7")]
            TxVersion::V6 => todo!(),
            #[cfg(zcash_unstable = "zfuture")]
            TxVersion::ZFuture => Self::read_v5(reader.into_base_reader(), version),
        }
    }
```
vs: https://github.com/zcash/librustzcash/blob/67993dc50975d311c54129b83f987980a57becef/zcash_primitives/src/transaction/mod.rs#L711
That's why the patch was used. And I've applied the patch to the whole librustzcash to avoid any compilation/dependencies problems.

conradoplg: Since git patches prevents us from publishing crates, we prefer to get the required changes merged and published in librustzcash first

nuttycom: Given that these changes are only available under a config flag, we can potentially release a point release with them if there are no semver-breaking API changes as of that commit. I will take a look.

nuttycom: Alternately, these PRs could target an `NU7` feature branch? In librustzcash, for a situation like this we would create a feature branch for this sort of development, and regularly merge `main` to it (via normal merges) along with merges from branches specific to the feature development, like this one. Then, eventually, the whole feature branch would be merged back to `main` when it's ready for release; interim states of the feature branch can rely on patch dependencies so long as everything relies on published dependencies once the branch is merged back to `main`.

aphelionz: I agree with @conradoplg regarding waiting for merge and release. It looks like those changes are in `main` and a new `librustzcash` release needs to be cut and the crates published?

nuttycom: Opened https://github.com/zcash/librustzcash/pull/1941

nuttycom: https://crates.io/crates/zcash_primitives 0.24.1 has been released.

mariopil: Thanks! I've updated and rebased the PR.

conradoplg:
My suggestion is to follow what librustzcash does, e.g.

```
#[cfg(all(zcash_unstable = "nu7", feature = "tx_v6"))]
```

Then we'd need to actually add a job in CI to test the gated code but we can do that later.

zebra-consensus/src/transaction/check.rs
```diff
@@ -122,9 +122,15 @@ pub fn lock_time_has_passed(
...
+    #[cfg(feature = "tx_v6")]
+    let has_other_circulation_effects = tx.has_zip233_amount();
+
+    #[cfg(not(feature = "tx_v6"))]
+    let has_other_circulation_effects = false;
+
     if !tx.has_transparent_or_shielded_inputs() {
         Err(TransactionError::NoInputs)
-    } else if !tx.has_transparent_or_shielded_outputs() {
+    } else if !tx.has_transparent_or_shielded_outputs() && !has_other_circulation_effects {
```
conradoplg, line 133: I don't think this rule is actually in the ZIP - in that case the ZIP needs to be updated

zebra-consensus/src/transaction/tests/prop.rs
```diff
@@ -32,7 +32,7 @@ proptest! {
         (network, block_height) in sapling_onwards_strategy(),
         block_time in datetime_full(),
         relative_source_fund_heights in vec(0.0..1.0, 1..=MAX_TRANSPARENT_INPUTS),
-        transaction_version in 4_u8..=5,
+        transaction_version in 4_u8..=6,
```
conradoplg: Doesn't this need to be feature-gated?

zebra-test/src/zip0244.rs
conradoplg, line 1: These seem to be only formatting changes? If so I'd prefer to revert them unless there is a particular reason for it

conradoplg (commented):
Some additional comments, but it's looking good, thanks

aphelionz:
@conradoplg I can't find it written anywhere (it may have been a sync call or an arborist discussion) but I believe that it was deemed the unstable flag was for ZIP-234 and ZIP-235 only since this is going in. That doesn't fix the CI though :/ 

cc @arya2 @nuttycom do you recall?

zebra-consensus/src/transaction/tests/prop.rs
mariopil: I reverted this back

conradoplg:
> @conradoplg I can't find it written anywhere (it may have been a sync call or an arborist discussion) but I believe that it was deemed the unstable flag was for ZIP-234 and ZIP-235 only since this is going in. That doesn't fix the CI though :/
> 
> cc @arya2 @nuttycom do you recall?

You're right, ignore my suggestion. I looked into it again and noticed that the CI failure is just due to missing feature propagation, e.g. in `zebra-rpc`, you need to add `tx_v6 = ["zebra-chain/tx_v6", "zebra-state/tx_v6", "zebra-consensus/tx_v6"]` to make sure the feature is enabled in all relevant dependencies. You will probably need to do the same in other crates.

It works normally (i.e. with `cargo test`) due to feature unification, but fails when building the crates individually (this is why we do that in CI in the first place).

(It's also weird that CI is now stuck instead of failing, I'll look into it)

conradoplg:
Could you please rebase on top of main to solve conflicts and to pick up the new refactored CI? Hopefully that might sort out the stuck CI.

mariopil:
@conradoplg  I've rebased the PR.

conradoplg:
Sorry, you will need to rebase again :sweat_smile:  (I was going to do that but GitHub does not allow it)

mariopil:
@conradoplg Done.

aphelionz:
@conradoplg are you able to approve the pending workflows?

upbqdn:
Approved.

conradoplg:
> I looked into it again and noticed that the CI failure is just due to missing feature propagation, e.g. in `zebra-rpc`, you need to add `tx_v6 = ["zebra-chain/tx_v6", "zebra-state/tx_v6", "zebra-consensus/tx_v6"]` to make sure the feature is enabled in all relevant dependencies. You will probably need to do the same in other crates.
> 
> It works normally (i.e. with `cargo test`) due to feature unification, but fails when building the crates individually (this is why we do that in CI in the first place).

The remaining failure is caused by the issue I pointed out above.

mariopil:
@conradoplg I've updated the Cargo.toml files, all crates should build now.

conradoplg (approved):
I still think we should also gate under `zcash_unstable` but I'll do that in a follow up PR since this has been waiting for ages

upbqdn:
This should be good once we merge #10050.

conradoplg:
@mergifyio requeue

PR #8940, opened 2024-10-17
fix(rpc): Refactor the cookie-based RPC authentication

### LINKED ISSUE #8864: Protect the Zebra RPC endpoint

upbqdn:
## Motivation

Resolve https://github.com/ZcashFoundation/zebra/pull/8900#discussion_r1801089575 and https://github.com/ZcashFoundation/zebra/pull/8900#discussion_r1801090790.

## Solution

- Don't read the cookie from the disk at all.
- Use a path to the cookie file when removing it.
- Rename `FixHttpRequestMiddleware` to `HttpRequestMiddleware` since it doesn't contain only fixes anymore but also serves as an authentication layer.

### Tests

No new tests.

### PR Author's Checklist

- [x] The PR name is suitable for the change log.
- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-rpc/src/server.rs
+++ b/zebra-rpc/src/server.rs
@@ -311,8 +311,12 @@ impl RpcServer {
         let wait_on_shutdown = move || {
             span.in_scope(|| {
                 if config.enable_cookie_auth {
-                    cookie::remove_from_disk(&config.cookie_dir)
-                        .expect("Zebra must be able to remove the auth cookie from the disk");
+                    if let Err(err) = cookie::remove_from_disk(&config.cookie_dir) {
+                        warn!(
+                            ?err,
+                            "unexpectedly could not remove the rpc auth cookie from the disk"
+                        )
+                    }
                 }
 
                 info!("Stopping RPC server");
--- a/zebra-rpc/src/server.rs
+++ b/zebra-rpc/src/server.rs
@@ -9,6 +9,8 @@
 
 use std::{fmt, panic, thread::available_parallelism};
 
+use cookie::Cookie;
+use http_request_compatibility::With;
 use jsonrpc_core::{Compatibility, MetaIoHandler};
 use jsonrpc_http_server::{CloseHandle, ServerBuilder};
 use tokio::task::JoinHandle;
@@ -25,7 +27,7 @@ use crate::{
     config::Config,
     methods::{Rpc, RpcImpl},
     server::{
-        http_request_compatibility::FixHttpRequestMiddleware,
+        http_request_compatibility::HttpRequestMiddleware,
         rpc_call_compatibility::FixRpcResponseMiddleware,
     },
 };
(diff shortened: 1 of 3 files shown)


zebra-rpc/src/server.rs
```diff
@@ -280,32 +286,37 @@ impl RpcServer {
...
-    fn shutdown_blocking_inner(close_handle: CloseHandle) {
+    fn shutdown_blocking_inner(close_handle: CloseHandle, config: Config) {
         // The server is a blocking task, so it can't run inside a tokio thread.
         // See the note at wait_on_server.
         let span = Span::current();
         let wait_on_shutdown = move || {
             span.in_scope(|| {
+                if config.enable_cookie_auth {
+                    cookie::remove_from_disk(&config.cookie_dir)
+                        .expect("Zebra must be able to remove the auth cookie from the disk");
```
arya2: It doesn't seem too important to remove the cookie file at shutdown, it should still work as expected when Zebra is restarted, so it seems better to try logging a warning here instead, at least for release builds.

We could also always skip removing the cookie and overwrite it at startup instead.

```suggestion
                    if let Err(err) = cookie::remove_from_disk(&config.cookie_dir) {
                        warn!(
                            ?err,
                            "unexpectedly could not remove the rpc auth cookie from the disk"
                        )
                    }
```

zebra-rpc/src/server/cookie.rs
```diff
@@ -6,43 +6,47 @@ use rand::RngCore;
...
 }
 
-/// Get the encoded password from the auth cookie.
-pub fn get(cookie_dir: PathBuf) -> Result<String> {
-    let mut file = File::open(cookie_dir.join(COOKIEAUTH_FILE))?;
-    let mut contents = String::new();
-    file.read_to_string(&mut contents)?;
+/// Writes the given cookie to the given dir.
+pub fn write_to_disk(cookie: &Cookie, dir: &Path) -> Result<()> {
+    File::create(dir.join(FILE))?.write_all(format!("__cookie__:{}", cookie.0).as_bytes())?;
```
arya2, line 40:
```suggestion
    // Create the directory if needed.
    std::fs::create_dir_all(&dir)?;
    File::create(dir.join(FILE))?.write_all(format!("__cookie__:{}", cookie.0).as_bytes())?;
```

zebra-rpc/src/server/http_request_compatibility.rs
```diff
@@ -9,38 +9,58 @@ use jsonrpc_http_server::{
...
+#[derive(Clone, Debug, Default)]
+pub struct HttpRequestMiddleware {
+    cookie: Option<Cookie>,
+}
+
+/// A trait for updating an object, consuming it and returning the updated version.
+pub trait With<T> {
+    /// Updates `self` with an instance of type `T` and returns the updated version of `self`.
+    fn with(self, _: T) -> Self;
+}
```
arya2, line 54: This looks great, good idea! 

Optional: This could be useful elsewhere in Zebra as well, such as in the `testnet::ParametersBuilder`, so it might fit better in `zebra-chain`, but it could also be moved in whenever we try implementing it for other structs.

arya2 (commented):
Thank you! This is an excellent refactor, it fixes the issues mentioned in the PR description, and the new `With` trait looks like it'll be very useful later/elsewhere in Zebra (I'm looking forward to using it in the `parameters::network::testnet` module).

zebra-rpc/src/server/http_request_compatibility.rs
upbqdn: I'm glad you like it. :) It pretty much randomly occurred to me to do it this way. It seems to go well with something like `let foo = if cond1 {Foo::...with(...)} else if cond2 {Foo::...with(...)} else {Foo::...with(...)}`.

Yea, let's move it somewhere else if we're using it in other places.

oxarbitrage (approved):
Looks really good and it works. Thanks for the refactor!

zebra-rpc/src/server/http_request_compatibility.rs
upbqdn: We could maybe add a macro which could take an arbitrary number of arguments: `...with!(a, b, c)`.

PR #8950, opened 2024-10-18
fix(clippy): Resolve new clippy lints and use renamed type from Rust std lib

arya2:


oxarbitrage (approved):
Lint was:

```
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
   --> zebra-chain/src/transaction/serialize.rs:625:17
    |
625 | /                 match sapling_shielded_data {
626 | |                     Some(sd) => writer.write_all(&<[u8; 64]>::from(sd.binding_sig)[..])?,
627 | |                     None => {}
628 | |                 }
    | |_________________^ help: try: `if let Some(sd) = sapling_shielded_data { writer.write_all(&<[u8; 64]>::from(sd.binding_sig)[..])? }`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match
    = note: `#[warn(clippy::single_match)]` on by default
```

I confirmed it is fixed in this PR.

PR #8951, opened 2024-10-19
change(rpc): Avoid re-verifying transactions in blocks if those transactions are in the mempool

### LINKED ISSUE #5674: Automatically verify transactions in blocks when the same transaction is already in the mempool
## Motivation

Miners need RPCs and verification to be fast, to improve profits.

When Zebra has already verified a transaction in the mempool, we know it will be valid in a block. (The mempool checks are stricter than the block transaction checks.)

So we can save time by automatically approving any block transaction verify requests that match a transaction already in the mempool.

## Security

We will need to test that the mempool checks are stricter than the block transaction checks. Otherwise, we could approve invalid blocks.

arya2:
## Motivation

We want the transaction verifier to avoid re-verifying transactions for the block verifier that have already been verified for the mempool.

Closes https://github.com/ZcashFoundation/zebra/issues/5674.

## Solution

- Adds `transaction_hash` and `known_outpoint_hashes` fields on `transaction::Request::Block`.
- Adds a mempool request for querying a transaction and its dependencies by its mined id
- Adds and calls a `try_find_verified_unmined_tx()` method on the transaction verifier which queries the mempool for the transaction, then checks that all of its dependencies are present in the set of `known_outpoint_hashes`.

### Tests

- Adds a `skips_verification_of_block_transactions_in_mempool` test

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-consensus/src/transaction/tests.rs
```diff
@@ -693,13 +696,179 @@ async fn mempool_request_with_unmined_output_spends_is_accepted() {
...
+    let (mempool_setup_tx, mempool_setup_rx) = tokio::sync::oneshot::channel();
+    let verifier = Verifier::new(&Network::Mainnet, state.clone(), mempool_setup_rx);
+    let verifier = Buffer::new(verifier, 1);
+
+    mempool_setup_tx
+        .send(mempool.clone())
+        .ok()
+        .expect("send should succeed");
+
+    let height = NetworkUpgrade::Canopy
```
oxarbitrage: Optional: I think we should use Nu6 in new tests.

zebra-consensus/src/transaction/tests.rs
```diff
@@ -693,13 +696,179 @@ async fn mempool_request_with_unmined_output_spends_is_accepted() {
...
+        .expect("Canopy activation height is specified");
+    let fund_height = (height - 1).expect("fake source fund block height is too small");
+    let (input, output, known_utxos) = mock_transparent_transfer(
+        fund_height,
+        true,
+        0,
+        Amount::try_from(10001).expect("invalid value"),
+    );
+
+    // Create a non-coinbase V4 tx with the last valid expiry height.
```
oxarbitrage, line 733: Optional: I think it will be good to use v5 transactions in all new tests as eventually v4 will be probably be deprecated.

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
```diff
@@ -105,7 +105,8 @@ pub async fn test_responses<State, ReadState>(
         extra_coinbase_data: None,
         debug_like_zcashd: true,
         // TODO: Use default field values when optional features are enabled in tests #8183
-        ..Default::default()
```
oxarbitrage, line 108: Is this change needed here ?

oxarbitrage (dismissed):
I like how the problem is resolved in this pull request. The code and logic looks good and the test too. I left a few minor and optional comments and a question.

zebra-rpc/src/methods/tests/snapshot/get_block_template_rpcs.rs
arya2: It's just to fix `cargo clippy --no-default-features`, I think I noticed it while I was on this branch and thought it was small and low-priority enough to include, but it doesn't have to be in this PR.

zebra-consensus/src/transaction/tests.rs
arya2: Good idea, practically free extra test coverage, applied in https://github.com/ZcashFoundation/zebra/pull/8951/commits/af9001dd91fdec765ec4137882bd8831bf0402db.

zebra-consensus/src/transaction/tests.rs
arya2: Good idea, practically free extra test coverage, applied in https://github.com/ZcashFoundation/zebra/pull/8951/commits/af9001dd91fdec765ec4137882bd8831bf0402db.

oxarbitrage (approved):
looks good, i already reviewed most of this code before.

zebra-consensus/src/transaction.rs
```diff
@@ -146,8 +146,12 @@ where
 pub enum Request {
     /// Verify the supplied transaction as part of a block.
     Block {
+        /// The transaction hash.
+        transaction_hash: transaction::Hash,
         /// The transaction itself.
         transaction: Arc<Transaction>,
+        /// Set of transaction hashes that create new transparent outputs.
+        known_outpoint_hashes: Arc<HashSet<transaction::Hash>>,
```
upbqdn, line 154: Do we need to wrap `HashSet` in `Arc`?

upbqdn: We do for cheap immutable cloning of `Request`s.

PR #8956, opened 2024-10-21
fix(build): Update Zebra's MSRV and use the latest Rust Docker image

arya2:
## Motivation

This [test failed](https://github.com/ZcashFoundation/zebra/actions/runs/11440468788/job/31826372026#step:10:846) on main due to a missing `PanicHookInfo` struct in the standard library that wasn't added until [Rust v1.81.0](https://doc.rust-lang.org/stable/releases.html#stabilized-apis-1).

We want to keep Zebra's minimum Rust version up-to-date and use the latest version of Rust in our docker builds.

## Solution

Updates `RUST_VERSION` in Dockerfile and the `rust-version` field in `zebrad/Cargo.toml`

### PR Author's Checklist

### PR Reviewer's Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ ARG TEST_FEATURES="lightwalletd-grpc-tests zebra-checkpoints"
 ARG EXPERIMENTAL_FEATURES=""
 
 ARG APP_HOME="/opt/zebrad"
-ARG RUST_VERSION=1.79.0
+ARG RUST_VERSION=1.82.0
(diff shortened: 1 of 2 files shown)


oxarbitrage (approved):

upbqdn (commented):
Why did we go with 1.81 and 1.82, and not only 1.82?

arya2:
> Why did we go with 1.81 and 1.82, and not only 1.82?


v1.81 is the new minimum rust version because it added the `PanicHookInfo` struct Zebra is using.

PR #8966, opened 2024-10-25
fix(state): Updates `total_size_on_disk` metric to be the total sst file size

arya2:
## Motivation

The db metrics printed by Zebra when it starts up seem too high, it looks like the total sst size in RocksDB includes the live database size.

## Solution

- Avoid adding the live database size to the total db size when printing db metrics

### Tests

Manually tested that the output matches the cache directory size reported by `du -hs`

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


oxarbitrage (approved):
nice one, thanks!

PR #8978, opened 2024-10-28
change(deps): Use ECC deps with activation height for NU6

upbqdn:
## Motivation

Zebra 2.0.0 uses ECC deps that don't have NU6 Mainnet activation height set. We must use new versions of these crates that have the activation height set.

## Solution

- Unpin ECC deps.
- Bump ECC deps. For reference, see https://github.com/zcash/zcash/blob/a0602ebfb765a965acd8de0b012da28f257a6918/Cargo.toml.
- Enable the `legacy-api` feature of `incrementalmerkletree` to be able to serialize note commitment trees using an old serialization format for the `z_gettreestate` RPC.

### Tests

No new tests.

### Follow-up Work

- Do a new release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -566,7 +566,7 @@ version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cef977c7f8e75aa81fc589064c121ab8d32448b7939d34d58df479aa93e65ea5"
 dependencies = [
- "incrementalmerkletree 0.7.0",
+ "incrementalmerkletree",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


arya2 (approved):

arya2:
Looks like the checkpoint job didn't find a cached state: https://github.com/ZcashFoundation/zebra/actions/runs/11559820540/job/32175800783?pr=8978#step:15:3351

Update: I'm running the `sync_to_mandatory_checkpoint_mainnet` test locally since it seems to timeout just before the mandatory checkpoint height it's trying to reach.

Update: The test passes locally, I think we should admin merge this if it times out again. Increasing the timeout should help otherwise.

PR #8987, opened 2024-11-04
fix(ci): fail cache disk creation if no db version is found

gustavovalverde:
In some cases Zebra logs might not output the database version, and thus we should avoid creating a disk without a version.

Before this change, a disk was created without a db version number, just indicating a `-v-`, that caused other tests to fail as an actual version was not found in their regexes.

The expected output in logs is:

```shell
Zcash network: Mainnet
running state version: 26.0.0
initial disk state version: creating.new.database
```

Like in this example: https://github.com/ZcashFoundation/zebra/actions/runs/11584155837/job/32255694034#step:11:614

Sometimes this is not the output as the **Diagnostic Metadata** is not shown: https://github.com/ZcashFoundation/zebra/actions/runs/11526479010/job/32091165009#step:11:632

## Solution

- If the version is not found, exit the `Get database versions from logs` step using an `exit 1`

### Follow-up Work

- We should confirm why the **Diagnostic Metadata** is not shown in some instances
- Manually fix the version numbers in the already created disks in GCP

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/.github/workflows/sub-deploy-integration-tests-gcp.yml
+++ b/.github/workflows/sub-deploy-integration-tests-gcp.yml
@@ -503,18 +503,19 @@ jobs:
           )
 
           if [[ -z "$INITIAL_DISK_DB_VERSION" ]]; then
+          # Check for new database creation
+          if echo "$DOCKER_LOGS" | grep -q "creating.new.database"; then
+              INITIAL_DISK_DB_VERSION="new"
+          else
               echo "Checked logs:"
               echo ""
               echo "$DOCKER_LOGS"
               echo ""
-              echo "Missing initial disk database version in logs: $INITIAL_DISK_DB_VERSION"
+              echo "Missing initial disk database version in logs"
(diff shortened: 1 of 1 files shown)


arya2:
> We should confirm why the Diagnostic Metadata is not shown in some instances

The added build metadata is actually unrelated, it would only happen when Zebra is compiled with the `indexer` feature, which isn't currently happening in CI.

In the case of the `scan_start_where_left` test, the shielded scanner was formerly a component of zebrad (behind the `shielded-scan` feature), but now it's running without starting Zebrad. The test could be updated to spawn a zebrad child (it's currently opening the state and reading blocks from it directly, and opening the state directly will still [log the running version but in an unexpected format](https://github.com/ZcashFoundation/zebra/actions/runs/11526479010/job/32091165009#step:11:634)).

arya2 (approved):

PR #9004, opened 2024-11-07
chore(ci): do not default to `tracing` mode in jobs execution

gustavovalverde:
## Motivation

Some jobs are hard to analyze as some variables are showing their values by default (which in some cases is the full log), multiple times in a single step.

If we'd like to debug with tracing, we can re-run a job with debugging on.

## Solution

- Remove the `shell` option with `-x` in most of the integration testing jobs

### Tests

- Tests should pass in this PR

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2 (approved):
Looks good! Would we be temporarily reverting these changes when we want to re-run a test with debug info?

PR #9005, opened 2024-11-08
feat(actions): migrate Mergify to GitHub Merge Queue

gustavovalverde:
## Motivation

We previously migrated to [GitHub Rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) as the new and preferred way to handle branch protection rules, but Mergify hasn't been updated to use rulesets, and we're also dealing with some deprecations around our actual Mergify configurations.

As GitHub Merge queue has matured enough to comply with most of our use cases, and it's already included in our actual GitHub Plan, we decided to start this migration path.

> This is an initial step before completely removing Mergify, to keep the duality between the two to avoid downtime with queue management

- https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#merge_group

## Solution

- Additionally to the tests that are triggered in a PR, we'll additionally re-run the Integration tests and CD Tests as these are the most relevant tests that must be confirmed when multiple PRs are being merged at the same time.

### Tests

- After merging this PR we should disable Mergify and activate GitHub Merge Queue, to confirm the behavior, as this only gets evaluated once the `merge_group` is available in the default branch.

### Follow-up Work

- Disable Mergify and turn on GitHub Merge Queue
- We should evaluate this new behavior, the time it takes for Merge PRs to get fully evaluated, etc.
- The Merge queue configurations should be as similar as the ones we have in the `mergify.yml` file
- Documentation should be updated in the following PR, once everything has been confirmed and the removal of Mergify is finalized.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/.github/workflows/cd-deploy-nodes-gcp.patch-external.yml
+++ b/.github/workflows/cd-deploy-nodes-gcp.patch-external.yml
@@ -1,7 +1,7 @@
 # Workflow patches for skipping Google Cloud CD deployments on PRs from external repositories.
 name: Deploy Nodes to GCP
 
-# Run on PRs from external repositories, let them pass, and then Mergify will check them.
+# Run on PRs from external repositories, let them pass, and then GitHub's Merge Queue will check them.
 # GitHub doesn't support filtering workflows by source branch names, so we have to do it for each
 # job.
 on:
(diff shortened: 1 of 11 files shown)


oxarbitrage (approved):
Looks good as the first step, let's merge and disable mergify to see if it works.

PR #8967, opened 2024-10-25
add(rpc): Adds `getblockheader` RPC method

arya2:
## Motivation

This is needed for block explorer support and has been requested by a mining pool.

Closes #8442.
Part of #8446 (always includes `height` in response when called with `verbosity = 2`).

### Specifications & References

https://zcash.github.io/rpc/getblockheader.html

## Solution

- Adds a `getblockheader` RPC method like the one in zcashd

Related changes:
- Adds `height` field to `getblock` response when called with a height `if verbosity == 2`

### Tests

- Adds a snapshot test for querying by height/hash and with verbose being true/false
- Adds a vectors test

### PR Author's Checklist

- [x] The PR name will make sense to users.
- [x] The solution is tested.
- [x] The documentation is up to date.
- [x] The PR has a priority label.

### PR Reviewer's Checklist

- [ ] The PR Author's checklist is complete.
- [ ] The PR resolves the issue.

PR #9006, opened 2024-11-08
change(rpc): Update `getblock` RPC to more closely match zcashd

### LINKED ISSUE #8446: Update `getblock` RPC method
## Motivation

This is needed for zcashd deprecation 

https://zcash.github.io/rpc/getblock.html 

## Scope

1. Identify remaining work to fully implement

conradoplg:
## Motivation

Currently zebra has partial support for `getblock()` and it mostly targets lightwalletd usage. For zcashd deprecation we want it to match zcashd more closely.

[zcashd RPC docs](https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L732-L806)

## Solution

Based on #8967

Since `getblock()` returns basically a superset of `getblockheader()`, I change our `getblock()` to call `getblockheader()` for all the shared fields.

Work on returning the additional non-header fields will be done in separate PRs.

I also changed `getblockheader()` return codes since whatever error it returns is now returned by `getblock()`, and we have committed to error codes in our snapshots. I also added the -5 error code which can also be returned by zcashd.

Since the previous behaviour can be still useful for performance reasons, I moved it to be carried out when verbosity=3 is used. It would be a simple change to `lightwalletd`. I'm not sure about the specific number though, should we use something farther off?

This also adds `ZEBRAD_EXTRA_ARGS` to `zcash-rpc-diff` because I was stubborn and wanted to make it work with cookie authentication :laughing: Let me know if you'd prefer it in a separate PR, but it seems simple enough.

### Tests

I adjusted the existing `vectors.rs` tests and updated the snapshots. I also tested it manually and with `zcash-rpc-diff`; the current diff is that `size, blockcommitments authdataroot, chainhistoryroot, chainwork anchor, chainSupply and valuePools` are not yet being filled; and verbosity=2 is still the same as 1 (i.e. `tx` is not expanded)

### Follow-up Work

- Check which of the remaining fields are truly required (some of them are undocumented). I'll ask the deprecation group.
- Created issues for the fields we are going to work on

### Additional information to put in Changelog

```
### Performance Regression

The `getblock()` change described below will impact lightwalletd operators who
back it with Zebra. The initial sync performance will drop unless you change
the `getblock()` call to use `verbosity=3`. See below for details.

### Changed

- The `getblock()` RPC now supports all the fields also returned by `getblockheader()`.
  The old partial support is now moved to `verbosity=3`; it might be useful for
  lightwalletd if you want to keep the same initial sync performance as before
  (you will need to changed lightwalletd source code to use `verbosity=3`)
```

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2:
This is looking good! I left some suggestions in #9008.

> Since the previous behaviour can be still useful for performance reasons, I moved it to be carried out when verbosity=3 is used. It would be a simple change to lightwalletd. I'm not sure about the specific number though, should we use something farther off?

The new behaviour seems performant enough for lightwalletd to continue using `verbosity = 1`. I don't think we need to mention a performance regression in the changelog either.

The performance issue before (fixed in https://github.com/ZcashFoundation/zebra/pull/5307) was that it was reading all of the block's transactions, the database queries added here should be quick, and, if they still have a measurable impact on lightwalletd sync performance, can be optimized easily (like by parallelizing or unifying state requests and maybe using an iterator to get the block hashes at `height`/`height.next()`) .

(Though we should make sure to cache block sizes by hash in a new column family when adding that field to avoid reading, deserializing, and then serializing the entire block to get its size.)

zebra-rpc/src/methods.rs
```diff
@@ -154,16 +157,19 @@ pub trait Rpc {
...
     /// # Notes
     ///
-    /// With verbosity=1, [`lightwalletd` only reads the `tx` field of the
-    /// result](https://github.com/zcash/lightwalletd/blob/dfac02093d85fb31fb9a8475b884dd6abca966c7/common/common.go#L152),
-    /// and other clients only read the `hash` and `confirmations` fields,
-    /// so we only return a few fields for now.
+    /// Zebra previously partially supported verbosity=1 by returning only the
+    /// fields required by lightwalletd ([`lightwalletd` only reads the `tx`
+    /// field of the
+    /// result](https://github.com/zcash/lightwalletd/blob/dfac02093d85fb31fb9a8475b884dd6abca966c7/common/common.go#L152)).
```
```suggestion
    /// field of the result](https://github.com/zcash/lightwalletd/blob/dfac02093d85fb31fb9a8475b884dd6abca966c7/common/common.go#L152)).
```

zebra-rpc/src/methods.rs
```diff
@@ -154,16 +157,19 @@ pub trait Rpc {
     /// # Parameters
     ///
     /// - `hash_or_height`: (string, required, example="1") The hash or height for the block to be returned.
-    /// - `verbosity`: (number, optional, default=1, example=1) 0 for hex encoded data, 1 for a json object, and 2 for json object with transaction data.
+    /// - `verbosity`: (number, optional, default=1, example=1) 0 for hex encoded data, 1 for a json object, and 2 for json object with transaction data, and 3 for a partially filled json object (which is faster and useful for lightwalletd-only usage)
```
oxarbitrage: I am confused here. I only see the verbosity cases 0, 1 and 2 in the method code.

zebra-rpc/src/methods.rs
```diff
@@ -861,65 +840,57 @@ where
...
-                    time,
+                    height: Some(height),
+                    version: Some(version),
+                    merkle_root: Some(merkle_root),
+                    time: Some(time),
+                    nonce: Some(nonce),
+                    solution: Some(solution),
+                    bits: Some(bits),
+                    difficulty: Some(difficulty),
+                    // TODO
```
oxarbitrage: Can you explain more what are the todo here and in the next ?

zebra-rpc/src/methods.rs
```diff
@@ -1750,21 +1737,84 @@ pub enum GetBlock {
...
         #[serde(skip_serializing_if = "Option::is_none")]
         height: Option<Height>,
 
-        /// The height of the requested block.
+        /// The version field of the requested block.
         #[serde(skip_serializing_if = "Option::is_none")]
-        time: Option<i64>,
+        version: Option<u32>,
 
+        /// The merkle root of the requesteed block.
```
```suggestion
        /// The merkle root of the requested block.
```

zebra-rpc/src/methods.rs
```diff
@@ -2156,3 +2223,22 @@ pub fn height_from_signed_int(index: i32, tip_height: Height) -> Result<Height>
         Ok(Height(sanitized_height))
     }
 }
+
```
oxarbitrage, line 2228:
```suggestion

/// A helper module to serialize `Option<T: ToHex>` as a hex string.
```

oxarbitrage (dismissed):
It looks good, i made a few comments and documentation suggestions.

zebra-rpc/src/methods.rs
conradoplg: Sure! I removed the TODOs in 1013611ef to make it less confusing (I forgot to clean them up) but they are:

- Supporting returning detailed transaction objects with verbosity=2 (that will be https://github.com/ZcashFoundation/zebra/issues/9024)
- Actually computing and returning the size of the block (that will be https://github.com/ZcashFoundation/zebra/issues/9020)

zebra-rpc/src/methods.rs
conradoplg: Initially I introduced a new verbosity but Arya pointed out that wasn't needed, and I forgot to update this. Thanks! Fixed in 1013611e

zebra-rpc/src/methods.rs
```diff
@@ -861,65 +839,55 @@ where
...
+                // This could be `None` if there's a chain reorg between state queries.
+                let orchard_tree =
+                    orchard_tree.ok_or_server_error("missing orchard tree for block")?;
+
+                let final_orchard_root =
+                    if nu5_activation.is_some() && height >= nu5_activation.unwrap() {
+                        Some(orchard_tree.root().into())
+                    } else {
+                        None
+                    };
```
arya2: Optional/Nitpick: We could avoid the `unwrap()` here with a match statement.

```suggestion
                let final_orchard_root = match nu5_activation {
                    Some(activation_height) if height >= activation_height => {
                        Some(orchard_tree.root().into())
                    }
                    _other => None,
                };
```

arya2 (dismissed):
Thank you!

zebra-rpc/src/methods.rs
```diff
@@ -995,13 +974,16 @@ where
...
-                    let mut root: [u8; 32] = sapling_tree.root().into();
-                    root.reverse();
-                    root
-                } else {
-                    [0; 32]
-                };
+                let sapling_activation = NetworkUpgrade::Sapling.activation_height(&network);
+                let sapling_tree_size = sapling_tree.count();
+                let final_sapling_root: [u8; 32] =
+                    if sapling_activation.is_some() && height >= sapling_activation.unwrap() {
```
arya2, line 985: `sapling_activation.unwrap()` should be fine since even configured Testnets must have an activation height for Canopy and all prior network upgrades.

arya2:
@Mergify update

conradoplg:
@mergifyio update

mpguerra (approved):

conradoplg:
@mergifyio queue

PR #9027, opened 2024-11-15
try modifying `ligthwalletd-update-sync` test variables

oxarbitrage:
### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


conradoplg:
I think we might want to keep that variable set but remove its `else if` clause in `entrypoint.sh`, see discussion in Slack.

PR #9038, opened 2024-11-18
fix(ci): allow to run all lighwalletd tests in Docker

gustavovalverde:
## Motivation

A LWD test was expecting the `ZEBRA_TEST_LIGHTWALLETD` to be set, but this variable is needed for all LWD tests and not specifically for `lightwalletd_integration`.

We had to rename this variable on a buggy `elif` statement in our Docker entrypoint.

This was avoiding most LWD tests to run correctly.

## Solution

- Use `$TEST_LWD_INTEGRATION` to run `lightwalletd_integration` in our entrypoint and in CI
- Remove `$ZEBRA_TEST_LIGHTWALLETD` as this was an incorrect logic

### Tests

- We must validate this tests are running correctly, by **checking the logs directly** in this PR

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -59,7 +59,7 @@ fi
 : "${RUN_ALL_EXPERIMENTAL_TESTS:=}"
 : "${TEST_FAKE_ACTIVATION_HEIGHTS:=}"
 : "${TEST_ZEBRA_EMPTY_SYNC:=}"
-: "${ZEBRA_TEST_LIGHTWALLETD:=}"
+: "${TEST_LWD_INTEGRATION:=}"
 : "${FULL_SYNC_MAINNET_TIMEOUT_MINUTES:=}"
 : "${FULL_SYNC_TESTNET_TIMEOUT_MINUTES:=}"
 : "${TEST_DISK_REBUILD:=}"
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -304,6 +300,10 @@ case "$1" in
         # Since these tests use the same cached state, a state problem in the first test can fail the second test.
         run_cargo_test "${ENTRYPOINT_FEATURES}" "--test-threads" "1" "fully_synced_rpc_"
 
+      elif [[ "${TEST_LWD_INTEGRATION}" -eq "1" ]]; then
+        # Test launching lightwalletd with an empty lightwalletd and Zebra state.
+        run_cargo_test "${ENTRYPOINT_FEATURES}" "lightwalletd_integration"
```
arya2, line 305: The `lightwalletd_integration` test is the one that doesn't use a cached state. It looks like `lightwalletd_update_sync` is the intended test.

```suggestion
        run_cargo_test "${ENTRYPOINT_FEATURES}" "lightwalletd_update_sync"
```

gustavovalverde: We already run the update here:
https://github.com/ZcashFoundation/zebra/pull/9038/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7caL317

And we trigger it from CI here:
https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L354

arya2 (approved):

gustavovalverde:
@mergify queue

gustavovalverde:
The `lightwalletd_wallet_grpc_tests` are failing in this PR, but that's a different issue. Let's merge this one so we can continue the Rust fix in another one https://github.com/ZcashFoundation/zebra/actions/runs/11895020195/job/33189274878#step:15:721

Issue #9024, opened 2024-11-14
getblock: return transactions details with `verbosity=2`


conradoplg:
### Motivation

For zcashd compatibility. In verbose=2 mode, the transactions details should be returned instead of transaction hashes.

### Specifications

https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L797-L799

### Complex Code or Requirements

Can probably be done by calling getrawtransaction to reuse its logic.

PR #9044, opened 2024-11-19
suggestion to don't clone `self` in `get_block`

oxarbitrage:
## Motivation

In https://github.com/ZcashFoundation/zebra/pull/9006 we are cloning self in the `get_block` implementation which could bring some problems in some tests.

## Solution

This is an alternative to do that that @conradoplg might find useful, feel free to discard.

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -750,9 +750,16 @@ where
 
         let mut state = self.state.clone();
         let verbosity = verbosity.unwrap_or(DEFAULT_GETBLOCK_VERBOSITY);
-        let self_clone = self.clone();
-
+        let network = self.network.clone();
         let original_hash_or_height = hash_or_height.clone();
+
+        // If verbosity requires a call to `get_block_header`, resolve it here
+        let get_block_header_future = if matches!(verbosity, 1 | 2) {
+            Some(self.get_block_header(original_hash_or_height.clone(), Some(true)))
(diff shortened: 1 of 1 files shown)


zebra-rpc/src/methods.rs
```diff
@@ -750,9 +750,16 @@ where
...
-
+        let network = self.network.clone();
         let original_hash_or_height = hash_or_height.clone();
+
+        // If verbosity requires a call to `get_block_header`, resolve it here
+        let get_block_header_future = if matches!(verbosity, 1 | 2) {
+            Some(self.get_block_header(original_hash_or_height.clone(), Some(true)))
+        } else {
+            None
+        };
```
arya2, line 761: Optional/Nitpick/Subjective: This may actually be a little less readable, but it's tempting to use `then()` here.

```suggestion
        let get_block_header_future = matches!(verbosity, 1 | 2)
            .then(|| self.get_block_header(original_hash_or_height.clone(), Some(true)));
```

arya2:
@conradoplg There's very little in `RpcImpl` being cloned but this looks like a good cleanup if you'd like to merge it into your PR. @oxarbitrage This could be a good change to include in your jsonrpsee PR if the base branch merges without it.

conradoplg:
Thanks! I'm not sure if this will solve the issues I'm getting implementing #9024 but I'll give it a shot there, inspired on this approoach.

PR #9049, opened 2024-11-21
fix(rpc): Refactor `getrawtransaction` & RPC error handling

### LINKED ISSUE #8744: bug: `getrawtransaction` output conflicts with `zcashd` output
### What happened?

`zcashd`'s `getrawtransaction` API is inherited from Bitcoin Core, and thus its behaviour around general transaction properties (as opposed to Zcash-specific properties) is likely ossified within the wider ecosystem (in particular for exchanges). In particular, the way it sets the `confirmations` field has been that way for many years.

In https://github.com/zcash/zcash/commit/f381d4e0856b5f6ebba429119f3fb97f83983c51 `zcashd` brought in the Insight Block Explorer patches, which added a `height` field alongside the `confirmations` field. Thus everything I say below about how `height` is set applies equivalently to `confirmations` (just with different values as appropriate).

`zcashd`'s `getrawtransaction` sets `height` in one of three ways:
- If the transaction is in the main chain, `height` is set to the height of the block it was mined in.
  https://github.com/zcash/zcash/blob/0bcddb896c51f58771a20966a0b1f5e5110a7e52/src/main.cpp#L2085-L2088
  https://github.com/zcash/zcash/blob/0bcddb896c51f58771a20966a0b1f5e5110a7e52/src/rpc/rawtransaction.cpp#L322-L329

- If the transaction is in the mempool, `height` is omitted.
  https://github.com/zcash/zcash/blob/0bcddb896c51f58771a20966a0b1f5e5110a7e52/src/main.cpp#L2064-L2069
  https://github.com/zcash/zcash/blob/0bcddb896c51f58771a20966a0b1f5e5110a7e52/src/rpc/rawtransaction.cpp#L322

- If the transaction is _not_ in the mempool and _not_ in the main chain, but is still recognized due to being mined in a block that is currently orphaned (in a non-main chain), `height` is set to -1.
  https://github.com/zcash/zcash/blob/0bcddb896c51f58771a20966a0b1f5e5110a7e52/src/rpc/rawtransaction.cpp#L332-L335

`zebrad` does not implement this API, and instead sets `height` as follows:
- If the transaction is in the main chain, `height` is set to the height of the block it was mined in.
  https://github.com/ZcashFoundation/zebra/blob/16168d762313a9d4c2af4c30cfc3e172c3291145/zebra-rpc/src/methods.rs#L1019-L1028
  https://github.com/ZcashFoundation/zebra/blob/16168d762313a9d4c2af4c30cfc3e172c3291145/zebra-rpc/src/methods.rs#L1692-L1698

- If the transaction is in the mempool, `height` is set to -1.
  https://github.com/ZcashFoundation/zebra/blob/16168d762313a9d4c2af4c30cfc3e172c3291145/zebra-rpc/src/methods.rs#L1004

- If the transaction is _not_ in the mempool and _not_ in the main chain, but is still recognized due to being mined in a block that is currently orphaned (in a non-main chain), `zebrad` reports `Err("transaction not found")` because non-main chain blocks are not included in `zebra_state::ReadResponse::Transaction`.
  https://github.com/ZcashFoundation/zebra/blob/16168d762313a9d4c2af4c30cfc3e172c3291145/zebra-rpc/src/methods.rs#L1029-L1031

Debugging `lightwalletd`, which itself is misusing the output of `getrawtransaction` from `zcashd`: https://github.com/zcash/librustzcash/pull/1473#discussion_r1701063435

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

### LINKED ISSUE #8786: Return -5 if transaction is not found in `getrawtransaction`
See https://github.com/ZcashFoundation/zebra/issues/8782 for context.

Will unblock [this lightwalletd issue](https://github.com/zcash/lightwalletd/issues/497)

### LINKED ISSUE #8945: `getrawtransaction` fixes and updates
## Motivation

This method is required for both mining and block explorer usecases. There are a number of outstanding fixes and updates that are necessary in order for this RPC method to match its zcashd counterpart which are listed as subissues.

upbqdn:
## Motivation

Close #8945.

- https://zcash.github.io/rpc/getrawtransaction.html
- https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/rawtransaction.cpp#L340

## Solution

- Fix the bugs described in #8744. Note that `zcashd` stores transactions in orphaned blocks and responds to queries requesting such transactions with `height` to -1. Since Zebra doesn't keep orphaned blocks below the rollback limit, it now returns the error _"No such mempool or blockchain transaction"_. If we wanted to match `zcashd`, we'd need to start storing orphaned blocks.
- Fix the error codes for `getrawtransaction` and some other RPCs.
- Related cleanups:
  - Refactor the error handling for RPCs.
  - Add references to the original error codes.
  - Simplify prop tests and run them with arbitrary `Network`s.
  - [Don't derive Serialize & Deserialize for txids](https://github.com/ZcashFoundation/zebra/pull/9049/commits/c9088f8f97ea70659ae6f333f2fcb7735ac451b6)

### Tests

Update and add

- snapshot tests,
- unit tests, and
- prop tests

to check the new error codes.

### Follow-up Work

- None

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/server/error.rs
```diff
@@ -0,0 +1,105 @@
...
+/// A trait for conditionally converting a value into a `Result<T, jsonrpc_core::Error>`.
+pub(crate) trait OkOrError<T> {
+    /// Converts the implementing type to `Result<T, jsonrpc_core::Error>`, using an error code and
+    /// message if conversion is to `Err`.
+    fn ok_or_error(
+        self,
+        code: impl Into<jsonrpc_core::ErrorCode>,
+        message: impl ToString,
+    ) -> std::result::Result<T, jsonrpc_core::Error>;
+}
```
arya2: Error constructor methods for commonly-used error codes (or just `Misc`) could be nice here.

```suggestion
/// A trait for mapping errors to [`jsonrpc_core::Error`].
pub(crate) trait MapError<T>: Sized {
    /// Maps errors to [`jsonrpc_core::Error`] with a specific error code.
    fn map_error(
        self,
        code: impl Into<jsonrpc_core::ErrorCode>,
    ) -> std::result::Result<T, jsonrpc_core::Error>;

    /// Maps errors to [`jsonrpc_core::Error`] with a [`LegacyCode::Misc`] error code.
    fn map_misc_error(self) -> std::result::Result<T, jsonrpc_core::Error> {
        self.map_error(LegacyCode::Misc)
    }
}

/// A trait for conditionally converting a value into a `Result<T, jsonrpc_core::Error>`.
pub(crate) trait OkOrError<T>: Sized {
    /// Converts the implementing type to `Result<T, jsonrpc_core::Error>`, using an error code and
    /// message if conversion is to `Err`.
    fn ok_or_error(
        self,
        code: impl Into<jsonrpc_core::ErrorCode>,
        message: impl ToString,
    ) -> std::result::Result<T, jsonrpc_core::Error>;

    /// Converts the implementing type to `Result<T, jsonrpc_core::Error>`, using a [`LegacyCode::Misc`] error code.
    fn ok_or_misc_error(
        self,
        message: impl ToString,
    ) -> std::result::Result<T, jsonrpc_core::Error> {
        self.ok_or_error(LegacyCode::Misc, message)
    }
}
```

zebra-rpc/src/methods.rs
```diff
@@ -1116,73 +1149,80 @@ where
         .boxed()
     }
 
-    // TODO: use HexData or SentTransactionHash to handle the transaction ID
     fn get_raw_transaction(
         &self,
-        txid_hex: String,
+        HexData(txid): HexData,
```
arya2: This could be typed as a `transaction::Hash` (which implements `ToHex` & `FromHex`), though it may return the wrong error code that way.

zebra-rpc/src/methods.rs
```diff
@@ -1116,73 +1149,80 @@ where
...
             };
 
-            // Now check the state
-            let request = zebra_state::ReadRequest::Transaction(txid);
-            let response = state
+            // If the tx wasn't in the mempool, check the state.
+            match state
                 .ready()
-                .and_then(|service| service.call(request))
+                .and_then(|service| service.call(zebra_state::ReadRequest::Transaction(txid)))
```
arya2, line 1154: Optional: To match zcashd's output when a transaction is on a side chain, we either need a new request to check for transactions on side chains, or to update this request so that it checks side chains for the transaction id too and returns an enum variant indicating whether the transaction was found on the best chain or a side chain.

Update: Let's skip this or do it in a follow-up PR if work hasn't already started to try and get https://github.com/ZcashFoundation/zebra/pull/9059 merged sooner.

zebra-rpc/src/methods.rs
```diff
@@ -733,200 +746,212 @@ where
...
         verbosity: Option<u8>,
     ) -> BoxFuture<Result<GetBlock>> {
-        // From <https://zcash.github.io/rpc/getblock.html>
-        const DEFAULT_GETBLOCK_VERBOSITY: u8 = 1;
-
         let mut state = self.state.clone();
-        let verbosity = verbosity.unwrap_or(DEFAULT_GETBLOCK_VERBOSITY);
 
         async move {
-            let hash_or_height: HashOrHeight = hash_or_height.parse().map_server_error()?;
```
arya2, line 764: These changes to the `get_block()` method may conflict with those in https://github.com/ZcashFoundation/zebra/pull/9006.

arya2 (commented):
This is looking good so far, I left an optional suggestion about matching zcashd's output when a transaction is found in a side chain. 

I'd like to review the changes to the `get_block()` method against #9006 if we can target this PR against the `get-block` branch (or wait for #9006 to merge and rebase off of main).

zebra-rpc/src/methods.rs
upbqdn: Let's resolve those when we merge either of the PRs.

zebra-rpc/src/server/error.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Maps errors to [`jsonrpc_core::Error`] with a specific error code.
+    fn map_error(
+        self,
+        code: impl Into<jsonrpc_core::ErrorCode>,
+    ) -> std::result::Result<T, jsonrpc_core::Error>;
+
+    /// Maps errors to [`jsonrpc_core::Error`] with a [`LegacyCode::Misc`] error code.
+    fn map_misc_error(self) -> std::result::Result<T, jsonrpc_core::Error> {
+        self.map_error(LegacyCode::Misc)
+    }
```
oxarbitrage, line 71: Unused method. I got a warning building locally.
```suggestion
```

zebra-rpc/src/server/error.rs
```diff
@@ -0,0 +1,118 @@
...
+        message: impl ToString,
+    ) -> std::result::Result<T, jsonrpc_core::Error>;
+
+    /// Converts the implementing type to `Result<T, jsonrpc_core::Error>`, using a [`LegacyCode::Misc`] error code.
+    fn ok_or_misc_error(
+        self,
+        message: impl ToString,
+    ) -> std::result::Result<T, jsonrpc_core::Error> {
+        self.ok_or_error(LegacyCode::Misc, message)
+    }
```
oxarbitrage, line 90:
```suggestion
```

zebra-rpc/src/methods.rs
```diff
@@ -733,200 +746,212 @@ where
...
+                // Reference for the legacy error code:
+                // <https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L629>
+                .map_error(server::error::LegacyCode::InvalidParameter)?;
+
+            match verbosity.unwrap_or(1) {
+                0 => {
+                    // # Performance
+                    //
+                    // This RPC is used in `lightwalletd`'s initial sync of 2 million blocks,
+                    // so it needs to load block data very efficiently.
```
```suggestion
```

This comment is repeated. I think this one can be removed but please double check before.

zebra-rpc/src/server/error.rs
```diff
@@ -0,0 +1,118 @@
+//! RPC error codes & their handling.
+
+/// Bitcoin RPC error codes
+///
+/// Drawn from <https://github.com/zcash/zcash/blob/master/src/rpc/protocol.h#L32-L80>.
```
oxarbitrage:
```suggestion
/// Drawn from <https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/protocol.h#L32-L80>.
```

It is maybe safer to link to a commit.

zebra-rpc/src/methods/tests/snapshots/getrawtransaction_invalid_txid@mainnet_10.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: zebra-rpc/src/methods/tests/snapshot.rs
+expression: rsp
+snapshot_kind: text
```
oxarbitrage, line 4: Unsure if we should/can just remove this line from the snapshots.

zebra-rpc/src/server/error.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// - All explicit discriminants fit within `i64`.
+#[derive(Default)]
+pub enum LegacyCode {
+    // General application defined errors
+    /// `std::exception` thrown in command handling
+    #[default]
+    Misc = -1,
+    /// Server is in safe mode, and command is not allowed in safe mode
+    ForbiddenBySafeMode = -2,
```
oxarbitrage, line 17: We could just remove the ones we are not using and suggest to be added on demand. However, as i am not sure what is better, i am ok keeping it like it is.

oxarbitrage (commented):
this looks good to me, i left a few minor comments and suggestions, also we need to fix conflicts.

zebra-rpc/src/methods/tests/snapshots/getrawtransaction_invalid_txid@mainnet_10.snap
upbqdn: You mean the `snapshot_kind`? What would be the motivation?

oxarbitrage: Only the snapshots of this PR has the `snapshot_kind`.

zebra-rpc/src/server/error.rs
upbqdn: That was just an interim inconsistency. Addressed in [ed69359](https://github.com/ZcashFoundation/zebra/pull/9049/commits/ed693598c69a0ef3b1e92e982783e0992d61399b).

zebra-rpc/src/server/error.rs
upbqdn: That was just an interim inconsistency. Addressed in [ed69359](https://github.com/ZcashFoundation/zebra/pull/9049/commits/ed693598c69a0ef3b1e92e982783e0992d61399b).

zebra-rpc/src/methods.rs
upbqdn: Thanks, I did that in [bdb8372](https://github.com/ZcashFoundation/zebra/pull/9049/commits/bdb837284d7fe8089b25e19d180b68d82a2b3ef6). The error code is not compatible with `zcashd`, but it wasn't anyway if the submitted string wasn't hex. I'll try to figure out if we need compatibility here and open another PR if we do.

zebra-rpc/src/methods/tests/snapshots/getrawtransaction_invalid_txid@mainnet_10.snap
upbqdn: I think it's fine as it is.

zebra-rpc/src/server/error.rs
upbqdn: Not a biggie, so I'll leave it as it is.

zebra-rpc/src/methods.rs
upbqdn: I didn't find similar comments. Could you share which ones you had in mind?

arya2: I think this comment is true for verbosity 1, but not for verbosity 0 here as lightwalletd doesn't use getblock with verbosity 0.

There is a note about performance [here](https://github.com/ZcashFoundation/zebra/pull/9049/files/475dbc9df3f041d5b6722539d136f8d2b3e56dfe#diff-09daf8f97cb59393d0c24039a43290af0772271b820987ef1a540404a78c44dbR786-R793) for the verbosity 1 arm of this match statement.

arya2 (dismissed):
Looks good to me, it would be nice to apply [this suggestion](https://github.com/ZcashFoundation/zebra/pull/9049#discussion_r1876731824) but it could also be removed in the next PR so I don't think it's a blocker.

zebra-rpc/src/methods.rs
upbqdn: I can't apply this suggestion anymore, so I moved the note at the top of the fn in [2c342f7](https://github.com/ZcashFoundation/zebra/pull/9049/commits/2c342f7e1ad48e261ae6b32ebd423e83b34fe800). I moved it because the verbosity 1 arm is not in the code anymore.

arya2 (dismissed):

zebra-rpc/src/methods.rs
arya2: > I can't apply this suggestion anymore

It's essentially the intended change, I'd describe that as having applied the suggestion, though I'm not sure if that's accurate.

arya2:
The lightwalletd tests seem to be required now, they should pass once https://github.com/ZcashFoundation/zebra/pull/9052 has merged. There was a timeout here: https://github.com/ZcashFoundation/zebra/actions/runs/12265821808/job/34222718084?pr=9049#step:12:2016, 

but the same tests passed with rust stable so it was likely spurious.

zebra-rpc/src/methods.rs
upbqdn: I just meant that the "apply" button here on GitHub wasn't active anymore, so I did the suggested changes locally. The downside is that @oxarbitrage didn't get any credit in the commit history.

arya2:
@Mergify update

conradoplg (changes requested):
I think there might be a bug, I'm looking into it

conradoplg:
Yep, e.g. `zcash-cli -conf=$PWD/zcash.conf getrawtransaction 851bf6fbf7a976327817c738c489d7fa657752445430922d94c983c0b9ed4609 1` is returning

```
error code: -1
error message:
Invalid params: invalid type: string "1e4842fd4962d395cd9905b7030fd5cb54c8c30148468c600416e4f14a264d2c", expected an array of length 32.
```

on `main` it returns

```
{
  "hex": "01000000010000000000000000000000000000000000000000000000000000000000000000ffffffff025100ffffffff0250c30000000000002321027a46eb513588b01b37ea24303f4b628afd12cc20df789fede0921e43cad3e875acd43000000000000017a9147d46a730d31f97b1930d3368a967c309bd4d136a8700000000",
  "height": 1,
  "confirmations": 2740688
}
```

upbqdn:
Thanks for spotting the bug. I'll switch back to `String`s and deserialize them manually instead of using `transaction::Hash` directly.

upbqdn:
> I think there might be a bug, I'm looking into it

Fixed.

zebra-rpc/src/methods.rs
upbqdn: I returned to using `String`s for the TXIDs even though they provide a weaker semantic abstraction. The reason is that `String`s are exactly what the RPC gets from the client. This means we can have greater test coverage (and error control). Using `transaction::Hash` felt great, but then I had to submit that directly to the RPC instead of a `String`, and we couldn't test deserialization properly, which led to bugs like https://github.com/ZcashFoundation/zebra/pull/9049#issuecomment-2536303946.

conradoplg (approved):

PR #9052, opened 2024-11-22
fix(test): Fixes bugs in the lightwalletd integration tests

### LINKED ISSUE #9012: bug: some Zebra and Lightwalletd logs have changed, causing some tests to fail
### What happened?

Tests passing

Instead, this happened:
Cached states are not being created as we're failing the disk creation if the expected logs are not found: 
- #8987

We have multiple regexes in Zebra for disk versioning, that are not being found:
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-deploy-integration-tests-gcp.yml#L493
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-deploy-integration-tests-gcp.yml#L493
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-deploy-integration-tests-gcp.yml#L493

And the [height of the disk](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-deploy-integration-tests-gcp.yml#L493), which varies by test:
- For [test_sync_to_mandatory_checkpoint_${NETWORK}](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L87) is `flushing database to disk .*height.*=.*Height.*\(`
- For [full_sync_mainnet](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L147), [zebrad_update_sync](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L186), [generate_checkpoints_mainnet](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L216) and [full_sync_testnet](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L252) is `current_height.*=.*Height.*\(`
- For [generate_checkpoints_testnet](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L295) is `zebra_tip_height.*=.*Height.*\(`
- For [lightwalletd_full_sync](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L327) and [lightwalletd_update_sync](https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-ci-integration-tests-gcp.yml#L360) is `Waiting for block: `

CI was running

### Zebra logs

This is an example: https://github.com/ZcashFoundation/zebra/actions/runs/11725652272/job/32664982469#step:12:170 

### Zebra Version

_No response_

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

arya2:
## Motivation

This PR fixes a bug in the `sending_transactions_using_lightwalletd` test where transactions from multiple future blocks are being sent to the mempool without committing prior blocks to the chain.

It also fixes a bug in the `lightwalletd_wallet_grpc_tests` where the test expects the ZF funding stream address balance to always be greater than zero.

## Solution

- Submits blocks to the state after sending their transactions to the mempool in the `sending_transactions_using_lightwalletd` test, and
- Only check that _active_ funding stream addresses have a balance that's greater than zero in the `lightwalletd_wallet_grpc_tests` test.

### Tests

The test should pass in CI.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2, line 1: This test could also fail due to transactions having unpaid actions (which are allowed by default in blocks but rejected by default in the mempool).

I'm looking into why it failed here: https://github.com/ZcashFoundation/zebra/actions/runs/11979091682/job/33400891720?pr=9052#step:15:844 where that transaction _did_ pay the conventional fee. 

The log also indicates that lightwalletd can't parse errors returned from Zebra's `sendrawtransaction` method, seemingly because it's expecting a colon-separated string but is receiving JSON.

conradoplg:
@mergifyio update

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -195,20 +265,29 @@ pub async fn run() -> Result<()> {
...
+
+        match rpc_client.send_transaction(request).await {
+            Ok(response) => assert_eq!(response.into_inner(), expected_response),
+            Err(err) => {
+                tracing::warn!(?err, "failed to send transaction");
+                zebrad_rpc_client
+                    .send_transaction(transaction)
+                    .await
+                    .map_err(|e| eyre!(e))?;
+            }
```
conradoplg: What is the reasoning for this? Wouldn't it be better to fail the test if sending a tx through the rpc_client fails?

zebrad/tests/common/lightwalletd/send_transaction_test.rs
```diff
@@ -195,20 +265,29 @@ pub async fn run() -> Result<()> {
...
+        zebrad.expect_stdout_line_matches("sending mempool transaction broadcast")?;
     }
 
-    // Check if some transaction is sent to mempool,
-    // Fails if there are only coinbase transactions in the first 50 future blocks
-    tracing::info!("waiting for mempool to verify some transactions...");
-    zebrad.expect_stdout_line_matches("sending mempool transaction broadcast")?;
     // Wait for more transactions to verify, `GetMempoolTx` only returns txs where tx.HasShieldedElements()
     // <https://github.com/zcash/lightwalletd/blob/master/frontend/service.go#L537>
-    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
conradoplg, line 211: Is this OK to remove? Should we update the comment above?

conradoplg (commented):
Sorry, I missed this PR before. It looks good, I just have some questions about parts of it.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: This was meant to help with debugging since lightwalletd was failing to parse the error, it shouldn't be merged.

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: It sleeps here until 4 seconds after this function is called just below, I think that should be enough.

conradoplg:
FYI I got [this error](https://github.com/ZcashFoundation/zebra/actions/runs/12118226331/job/33782709037#step:15:895) on another PR:

```
  left: SendResponse { error_code: 0, error_message: "failed to validate tx: transaction::Hash(\"private\"), error: transaction did not pass consensus validation: could not validate nullifiers and anchors on best chain: immature transparent coinbase spend: attempt to spend OutPoint { hash: transaction::Hash(\"47bc05a53a5f270604f9d6a52bdc48cb9a5c7a1efc4ff6eb00f8842f287a3f2c\"), index: 0 } at Height(2739104), but spends are invalid before Height(2739107), which is 100 blocks after it was created at Height(2739007)" }
 right: SendResponse { error_code: 0, error_message: "\"0f85fada75a7a59a02e15363f8bd62150c9148d133b50d570952bb89bc713a1f\"" }
 ```
 
 which does seems like something that might be fixed by this PR

**Comment by conradoplg** [comment]
[This failure](https://github.com/ZcashFoundation/zebra/actions/runs/12164273923/job/33926037777?pr=9052#step:15:834) seems new:

```
Error: 
   0: error sending request for url (http://127.0.0.1:37073/): error trying to connect: tcp connect error: Connection refused (os error 111)
   1: error trying to connect: tcp connect error: Connection refused (os error 111)
   2: tcp connect error: Connection refused (os error 111)
   3: Connection refused (os error 111)

Location:
   zebrad/tests/common/regtest.rs:134
```

**Review by conradoplg — DISMISSED** [review]
Looks good, just needs solving conflicts

**Comment by arya2** [comment]
I still need to fix the new test failure before this merges (`error trying to connect` when calling Zebra's RPC server) first, but I'm happy for this to merge without a fix for the invalid script issue.

**Comment by arya2** [comment]
The test is now passing locally.

File: zebrad/tests/common/lightwalletd/send_transaction_test.rs [hunk 3]
```diff
@@ -296,25 +288,6 @@ async fn send_transactions_from_block(
...
-    #[allow(unused_assignments)]
-    {
-        zebrad = zebrad.with_timeout(LARGE_CHECKPOINT_TIMEOUT);
-    }
-
-    if tx_log.is_err() {
-        tracing::info!("lightwalletd didn't query the mempool, skipping mempool contents checks");
-        return Ok((zebrad, has_tx_with_shielded_elements, 0));
-    }
-
```
conradoplg: Could you explain the reasoning why it is OK to remove this part?

conradoplg (dismissed):
Looks good! I left a question just to be sure about the fix of the last issue

zebrad/tests/common/lightwalletd/send_transaction_test.rs
arya2: It looks like lightwalletd is [always querying `getrawmempool` as long as it's been 2s since the last query](https://github.com/zcash/lightwalletd/blob/master/frontend/service.go#L471-L475), and the body of the two loops below shouldn't execute if that ever changes. I think it adds a little test coverage because lightwalletd shouldn't return an error if its copy of the mempool is empty, just an empty response.

There might be another problem here, it's checking that the ids of transactions it gets from lightwalletd are all in `transaction_hashes`, but lightwalletd may return transactions that were added to the mempool from previous blocks, and `transaction_hashes` only includes ids of transactions in the current block. I'll fix that.

Update: Fixed the expected transaction hashes issue in https://github.com/ZcashFoundation/zebra/pull/9052/commits/06d163b410b4ead14c89d65f781eba0923b0d619.

conradoplg (dismissed):

zebrad/src/components/mempool/downloads.rs
```diff
@@ -124,7 +124,7 @@ pub enum TransactionDownloadVerifyError {
     #[error("transaction download / verification was cancelled")]
     Cancelled,
 
-    #[error("transaction did not pass consensus validation: {0}")]
+    #[error("transaction did not pass consensus validation, error: {0:?}")]
```
conradoplg: Optional: revert if you agree with the reasoning here https://github.com/ZcashFoundation/zebra/pull/9066#discussion_r1869582725

```suggestion
    #[error("transaction did not pass consensus validation: {0}")]
```

conradoplg:
(It was dequeued because I noticed that issue I commented just after I approved it)

oxarbitrage (approved):

PR #9083, opened 2024-12-12
add(rpc): getblock: return transaction details with verbosity=2

### LINKED ISSUE #9024: getblock: return transactions details with `verbosity=2`

conradoplg:
## Motivation

We want getblock to match zcashd. With verbosity=2 we returned a list of hashes, but a list of transaction objects is expected.

Closes #9024 

https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L735

## Solution

If `verbosity=2`, call `getrawtransaction` for each tx in the block internally and use its output to fill the transaction details.

There are tons of fields missing in `getrawtransaction`, but that will be another issue. When they are filled, they will be automatically returned by `getblock` too.

`zcashd` seems to not return the `heigth` and `confirmations` fields in the transactions details inside `getblock` (which are indeed redundant) but I think it's harmless to return them.

### Tests

Expanded existing test. Also tested manually with `zcash-cli`.

### Follow-up Work

See https://github.com/ZcashFoundation/zebra/issues/8446 for other `getblock` related stuff, and we'll need an issue for completing `getrawtransaction`

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/methods.rs
```diff
@@ -1987,22 +2021,33 @@ pub enum GetRawTransaction {
...
 }
 
 impl Default for GetRawTransaction {
     fn default() -> Self {
-        Self::Object {
+        Self::Object(TransactionObject::default())
+    }
+}
+
+/// A Transaction object as returnedb by `getrawtransaction` RPC request.
```
```suggestion
/// A Transaction object as returned by `getrawtransaction` RPC request.
```

oxarbitrage (dismissed):
Looks good, it might conflict with https://github.com/ZcashFoundation/zebra/pull/9049

oxarbitrage:
It seems some test is still locked somewhere. I am unsure if it is caused by this code but it seems so.

arya2:
This looks good, I opened a suggestion PR (https://github.com/ZcashFoundation/zebra/pull/9084) for reducing db queries and service calls.

oxarbitrage (approved):

PR #8895, opened 2024-09-27
add(state): Track spending transaction ids by spent outpoints and revealed nullifiers

arya2:
## Motivation

We want to lookup transaction ids by their transparent inputs and revealed nullifiers.

Closes #8837, 
Closes #8838. 
Closes #8922.

## Solution

- Adds a new `tx_loc_by_spent_out_loc` column family
- Updates revealed nullifier column families to store spending transaction locations as values instead of `()`
- Stores the `TransactionLocation` of spending transactions by spent  `OutputLocation`s and nullifiers when writing blocks to the finalized state
- Adds the hashes of spending transactions as the values in the `spent_utxos` field on non-finalized `Chain`s
- Adds `ReadRequest` and `ReadResponse` variants for querying spending tx ids by outpoint with the `ReadStateService`
- Adds a `spending_transaction_hash()` read function used to handle the new `ReadRequest`
- Updates snapshots

It may be possible to update the `tx_loc_by_transparent_addr_loc` column instead, but adding a new one seemed easier.

Related Changes:
- Updates `cancel_receiver` to a `crossbeam-channel::mpmc` and parallelizes the db format upgrade by block

### Tests

- Adds a test that checks the last 500 blocks in the finalized and non-finalized state for spending transaction ids (uses a cached state)
- Manually tested the db format upgrade
- [Full Mainnet sync test running here](https://github.com/ZcashFoundation/zebra/actions/runs/11267618723)

### PR Author's Checklist

- [x] The PR name will make sense to users.
- [ ] The PR provides a CHANGELOG summary.
- [x] The solution is tested.
- [x] The documentation is up to date.
- [x] The PR has a priority label.

### PR Reviewer's Checklist

- [ ] The PR Author's checklist is complete.
- [ ] The PR resolves the issue.

PR #9084, opened 2024-12-13
Suggestion for "add(rpc): getblock: return transaction details with verbosity=2"

### LINKED ISSUE #9024: getblock: return transactions details with `verbosity=2`

arya2:
This is a suggestion for https://github.com/ZcashFoundation/zebra/pull/9083 to fetch all of a block's transactions with a single service call instead of making one per transaction. 

It needlessly reads the block header twice, but those are relatively small, should be cached, and fetching transactions by block height avoids looking up transaction ids by location and then transaction locations by id. It'll also make use of RocksDB iterators once https://github.com/ZcashFoundation/zebra/pull/8895 has merged (see https://github.com/ZcashFoundation/zebra/pull/8895/files#diff-9781032518c3c6d10a656ef34d8b7a85325f9c55c5fd65856b2257786459fd50R226-R234).

We could later factor out some of the `getblockheader` method to re-use that code without reading the block header twice.

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -852,15 +852,20 @@ where
                         .transactions
                         .iter()
                         .map(|tx| {
-                            TransactionObject::new(
-                                tx.clone(),
-                                height,
-                                confirmations
-                                    .try_into()
-                                    .expect("should be less than max block height, i32::MAX"),
-                            )
+                            let GetRawTransaction::Object(tx_obj) =
+                                GetRawTransaction::from_transaction(
+                                    tx.clone(),
+                                    Some(height),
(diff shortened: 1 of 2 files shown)


conradoplg (approved):
Great suggestion, that's much better, thanks!

Issue #9089, opened 2024-12-16
The tx verifier does not enforce using the correct consensus branch ID in the SIGHASH precomputation


upbqdn:
When precomputing the SIGHASH of V5 txs, we pick the consensus branch ID from the `nConsensusBranchId` field if the tx has it. If the field is set to a consensus branch ID corresponding to NU5 and NU6 is active, the SIGHASH precomputation doesn't follow this consensus rule:

> [NU6 only] All transactions MUST use the NU6 consensus branch ID 0xC8E71055 as defined in ZIP-253.

This bug was mitigated in #9063 by checking the `nConsensusBranchId` field early in the tx verifier, so it won't accept NU5 transactions after NU6 activation anymore. However, we should fix the precomputation to adhere to the above consensus rule since it takes place in `zebra-chain` and not `zebra-consensus`, where the verifier is, and can be used by other crates.

The bug originates in this fn: https://github.com/zcash/librustzcash/blob/443faf9e855e0acf7aca509885ac2ede8090c5bd/zcash_primitives/src/transaction/mod.rs#L604-L616, which considers the `consensus_branch_id` param for pre-V5 txs but doesn't do so for V5 ones.

mpguerra:
blocking #8998

PR #9098, opened 2024-12-21
add(test): Add serialized NU5 blocks to test vectors

### LINKED ISSUE #2409: Improve `fake_v5_transaction_with_orchard_actions_has_inputs_and_outputs` test
In https://github.com/ZcashFoundation/zebra/pull/2398 we extended the `fake_v5_transaction_with_orchard_actions_has_inputs_and_outputs` unit test for new consensus rules.

In this test we are calling `insert_fake_orchard_shielded_data()` multiple times which i think it can be avoided.

This is a very low priority clean up task.

### LINKED ISSUE #9088: Add serialized blocks with V5 txs containing Orchard actions to test vectors
## Motivation

The current test vectors don't contain serialized blocks with V5 txs with Orchard actions. We need such txs to test the tx verifier.

### LINKED ISSUE #9090: Testnet NU5 transition period has wrong blocks in test vectors
https://github.com/ZcashFoundation/zebra/blob/543f0669e15a7d7da4e43d9d009fbfe9583bcec0/zebra-test/src/vectors/block.rs#L855-L864 should contain blocks for heights from 1_842_419 to 1_842_421 since NU5 activated at height 1_842_420 on Testnet.

upbqdn:
## Motivation

Close #2409, #9090 and #9088.

## Solution

- Add suitable serialized Mainnet & Testnet NU5 blocks and their anchors to test vectors.
- [Remove wrong Testnet NU5 blocks](https://github.com/ZcashFoundation/zebra/commit/fe13db8c9f5a0daa12017da79af31b72d09340a1).
- [Add has_transparent_inputs to Transaction](https://github.com/ZcashFoundation/zebra/commit/35513d0591b68a2a3806e7947f6f1318825a513d)

### Tests

- [Remove unneeded test for fake V5 txs](https://github.com/ZcashFoundation/zebra/commit/72289299a4cbda03464a43e1e9e6fcbb8f73b545).
- Use real V5 txs where appropriate. This uncovered some malfunctioning tests, for example 34e985b74eaf31cc562a03dbd927ab55b09133d3.
- Fix failing tests. This turned out to be much more work than I expected.
- Use `Network::iter` instead of only Mainnet txs.

### Follow-up Work

- #9089

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebra-consensus/src/transaction/tests.rs
```diff
@@ -64,172 +64,178 @@ fn v5_fake_transactions() -> Result<(), Report> {
...
+            Err(TransactionError::NoOutputs)
+        );
+    }
 }
 
 #[tokio::test]
 async fn mempool_request_with_missing_input_is_rejected() {
-    let mut state: MockService<_, _, _, _> = MockService::build().for_prop_tests();
-    let verifier = Verifier::new_for_tests(&Network::Mainnet, state.clone());
+    let mut state: MockService<_, _, _, _> = MockService::build().for_unit_tests();
```
arya2, line 207: Thank you for fixing the MockService type here and removing the unused `CheckBestChainTipNullifiersAndAnchors` handling below.

zebra-consensus/src/transaction/tests.rs
```diff
@@ -2731,164 +2686,136 @@ fn v4_with_sapling_outputs_and_no_spends() {
...
 
-        let height = transaction
-            .expiry_height()
-            .expect("Transaction is missing expiry height");
+        let height = tx.expiry_height().expect("expiry height");
 
-        // Insert fake Orchard shielded data to the transaction, which has at least one action (this is
-        // guaranteed structurally by `orchard::ShieldedData`)
-        let shielded_data = insert_fake_orchard_shielded_data(&mut transaction);
+        let orchard_shielded_data = tx.orchard_shielded_data_mut().unwrap();
```
arya2: Optional/Nitpick:

```suggestion
        let orchard_shielded_data = tx
            .orchard_shielded_data_mut()
            .expect("tx without transparent, Sprout, or Sapling outputs must have Orchard actions");
```

zebra-consensus/src/transaction/tests.rs
```diff
@@ -3446,87 +3373,65 @@ fn fill_action_with_note_encryption_test_vector(
...
-        shielded_data.actions = vec![AuthorizedAction::from_parts(action, sig)]
-            .try_into()
-            .unwrap();
+        for net in Network::iter() {
+            let mut transaction = v5_transactions(net.block_iter())
+                .find(|tx| tx.is_coinbase())
+                .expect("coinbase V5 tx");
+
+            let shielded_data = insert_fake_orchard_shielded_data(&mut transaction);
+            shielded_data.flags = Flags::ENABLE_SPENDS | Flags::ENABLE_OUTPUTS;
```
arya2: Nitpick: Coinbase transactions aren't allowed to have the ENABLE_SPENDS flag set, and shouldn't be spending from Orchard.

```suggestion
            shielded_data.flags = Flags::ENABLE_OUTPUTS;
```

zebra-consensus/src/transaction/tests.rs
```diff
@@ -3446,87 +3373,65 @@ fn fill_action_with_note_encryption_test_vector(
...
-            | zebra_chain::orchard::Flags::ENABLE_OUTPUTS;
-
-        let action =
-            fill_action_with_note_encryption_test_vector(&shielded_data.actions[0].action, v);
-        let sig = shielded_data.actions[0].spend_auth_sig;
-        shielded_data.actions = vec![AuthorizedAction::from_parts(action, sig)]
-            .try_into()
-            .unwrap();
+        for net in Network::iter() {
+            let mut tx = v5_transactions(zebra_test::vectors::TESTNET_BLOCKS.iter())
```
arya2: Nice catch updating this test to use a coinbase transaction!

Did you mean to use `block_iter()` here?

```suggestion
            let mut tx = v5_transactions(net.block_iter())
```

Related but wasn't changed in this PR: Returning early from `coinbase_tx_no_prevout_joinsplit_spend()` but not `coinbase_outputs_are_decryptable()` if `!tx.is_coinbase()` seems inconsistent (albeit unimportant), we may want to update one of them.

zebra-test/src/vectors/block.rs
```diff
@@ -852,20 +998,52 @@ lazy_static! {
         <[u8; 32]>::from_hex("2b30b19f4254709fe365bd0b381b2e3d9d0c933eb4dba4dd1d07f0f6e196a183")
         .expect("final root bytes are in valid hex representation").rev();
 
-    // Nu5 transition
-    // for i in 1599199 1599200 1599201; do
-    //     zcash-cli -testnet getblock $i 0 > block-test-$[i/1000000]-$[i/1000%1000]-$[i%1000].txt
-    // done
-    pub static ref BLOCK_TESTNET_1599199_BYTES: Vec<u8> =
-        <Vec<u8>>::from_hex(include_str!("block-test-1-599-199.txt").trim())
```
arya2, line 860: What was wrong with this block?

arya2 (dismissed):
This looks great, thank you for these changes.

I left a few suggestions, none of them are blockers if you'd like this to merge as-is.

zebra-consensus/src/transaction/tests.rs
```diff
@@ -3446,87 +3373,65 @@ fn fill_action_with_note_encryption_test_vector(
...
+                .find(|tx| tx.is_coinbase())
+                .expect("V5 coinbase tx");
+
+            let shielded_data = insert_fake_orchard_shielded_data(&mut tx);
+            shielded_data.flags = Flags::ENABLE_SPENDS | Flags::ENABLE_OUTPUTS;
```
arya2:
```suggestion
            shielded_data.flags = Flags::ENABLE_OUTPUTS;
```

zebra-test/src/vectors/block.rs
upbqdn: [ZIP 252](https://zips.z.cash/zip-0252) specifies that the current Testnet activated at height 1842420 with consensus branch ID 0xc2d6d0b4, which I used in this PR. I actually used 1842421 because the block at 1842420 was empty. This was the second activation of NU5 on Testnet.

As regards the original activation, the ZIP contains the following note:

> A first activation of NU5, with a previous version of the Orchard circuit and other NU5 consensus rules, occurred on Testnet at block height 1599200 with consensus branch ID 0x37519621 and peer protocol version 170015. With the release of zcashd v4.7.0, Testnet is being rolled back to another chain that forks from the block immediately preceding that activation, at height 1599199.

Also note that the removed comment mentions heights 1599199, 1599200, and 1599201, but the code uses only 1599199, which didn't have NU5 activated.

zebra-consensus/src/transaction/tests.rs
upbqdn: Thanks for spotting the inconsistency. I'll open another PR.

upbqdn: Or I do it in this PR since it's a minor clean-up, even though it's not related to this PR.

upbqdn: > Related but wasn't changed in this PR: Returning early from `coinbase_tx_no_prevout_joinsplit_spend()` but not `coinbase_outputs_are_decryptable()` if `!tx.is_coinbase()` seems inconsistent (albeit unimportant), we may want to update one of them.

Done in [6928e06](https://github.com/ZcashFoundation/zebra/pull/9098/commits/6928e06d52c18776cec66495944f4f4fa28c722d).

zebra-consensus/src/transaction/check.rs
```diff
@@ -330,6 +330,10 @@ pub fn coinbase_outputs_are_decryptable(
     network: &Network,
     height: Height,
 ) -> Result<(), TransactionError> {
+    if !transaction.is_coinbase() {
+        return Ok(());
+    }
+
```
arya2: Unfortunately this seems to cause a test failure, https://github.com/ZcashFoundation/zebra/actions/runs/12747335909/job/35525212184?pr=9098#step:12:2959

```suggestion
```

upbqdn: When looking into this, I found a simple optimization that I think is worth doing, and we could also keep this check. I'll open another PR.

arya2:
@mergifyio update

arya2 (approved):

PR #9114, opened 2025-01-13
chore: Fix new clippy lints

upbqdn:
## Motivation

Clippy has some new lints that make the CI fail.

## Solution

- Refactor the code so it passes the lint checks.

### Tests

- No new tests.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2 (dismissed):

arya2:
The release job is still failing, we could replace it with a patch job in this PR, or, @gustavovalverde could we remove it from the list of required jobs?

gustavovalverde:
> @gustavovalverde could we remove it from the list of required jobs?

@arya2 Do we expect to fix the issue causing the job to fail? We can remove it a as required job, but it will keep creating red crosses across the repo and in the main branch, and I'd prefer to avoid that if possible. Another question is, do we need this to run in all PRs or just on the release PR?

arya2:
> @arya2 Do we expect to fix the issue causing the job to fail?

Yes, but not for a while, I'm hoping `--no-verify`'s behaviour will be restored in the next version of Rust, but failing that I think we have to set up a private registry or use another complicated workaround.

> We can remove it a as required job, but it will keep creating red crosses across the repo and in the main branch, and I'd prefer to avoid that if possible.

Yeah, it might be better to just remove the job altogether, or replace it with a patch job for now.

> Another question is, do we need this to run in all PRs or just on the release PR?

This one runs on all PRs to check that Zebra could still be released, though it's never fully worked because verifying that a crate can be published relies on publishing its dependencies, and we don't want to do that for every change (at least not to crates.io).

upbqdn:
@gustavovalverde, @arya2, I opened #9119 atop this PR.

arya2:
Needs a quick merge conflict resolution.

@gustavovalverde it may still need to be removed from the list of required jobs since the patch job was removed.

arya2 (approved):
Thank you!

upbqdn:
I'm now thinking I should've opened #9119 against `main` since it would have passed CI without the lint updates in this PR, and we'd get a better merge history. I'll open it again against `main`.

PR #9118, opened 2025-01-14
fix(consensus): Avoid a concurrency bug when verifying transactions in blocks that are already present in the mempool

arya2:
## Motivation

This PR fixes a bug in `find_verified_unmined_tx()` where a `TransparentInputNotFound` is returned if a transaction's dependencies are missing from the block containing it. The mempool removes any transaction ids mined onto the best chain from its transaction dependencies, but it's possible for a transaction to depend on outputs that aren't in the same block or in the state, but elsewhere in the verification pipeline waiting to be contextually validated.

## Solution

Wait for the UTXOs to arrive in the state instead of returning an error when a transaction's dependencies in the mempool are unavailable in the block.

Related changes:
- Return `TransparentInputNotFound` errors from tx verifier instead of `InternalDowncastErrors` when `AwaitUtxo` requests timeout, and
- Rename `try_find_verified_unmined_tx` to `find_verified_unmined_tx`

### Tests

Updates the `skips_verification_of_block_transactions_in_mempool` test to check that the tx verifier falls back on waiting for UTXOs to arrive in the state service if a transaction's dependencies are missing from its block.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -699,7 +699,7 @@ where
                 .oneshot(zebra_state::Request::AwaitUtxo(*missing_outpoint));
             match query.await {
                 Ok(zebra_state::Response::Utxo(_)) => {}
-                Err(err) => return Some(Err(err.into())),
+                Err(_) => return Some(Err(TransactionError::TransparentInputNotFound)),
                 _ => unreachable!("AwaitUtxo always responds with Utxo"),
             };
         }
@@ -750,7 +750,10 @@ where
                         .clone()
                         .oneshot(zs::Request::UnspentBestChainUtxo(*outpoint));
 
-                    let zebra_state::Response::UnspentBestChainUtxo(utxo) = query.await? else {
+                    let zebra_state::Response::UnspentBestChainUtxo(utxo) = query
+                        .await
+                        .map_err(|_| TransactionError::TransparentInputNotFound)?
+                    else {
                         unreachable!("UnspentBestChainUtxo always responds with Option<Utxo>")
                     };
 
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -685,6 +685,10 @@ where
             .filter(|dependency_id| !known_outpoint_hashes.contains(dependency_id))
             .collect();
 
+        if missing_deps.is_empty() {
+            return Some(Ok(verified_tx));
+        }
+
         let missing_outpoints = tx.inputs().iter().filter_map(|input| {
             if let transparent::Input::PrevOut { outpoint, .. } = input {
                 missing_deps.contains(&outpoint.hash).then_some(outpoint)
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -402,7 +402,7 @@ where
         async move {
             tracing::trace!(?tx_id, ?req, "got tx verify request");
 
-            if let Some(result) = Self::try_find_verified_unmined_tx(&req, mempool.clone()).await {
+            if let Some(result) = Self::find_verified_unmined_tx(&req, mempool.clone(), state.clone()).await {
                 let verified_tx = result?;
 
                 return Ok(Response::Block {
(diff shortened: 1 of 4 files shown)


zebra-consensus/src/transaction.rs
```diff
@@ -676,17 +680,31 @@ where
 
         // Note: This does not verify that the spends are in order, the spend order
         //       should be verified during contextual validation in zebra-state.
-        let has_all_tx_deps = dependencies
+        let missing_deps: HashSet<_> = dependencies
             .into_iter()
-            .all(|dependency_id| known_outpoint_hashes.contains(&dependency_id));
+            .filter(|dependency_id| !known_outpoint_hashes.contains(dependency_id))
+            .collect();
```
arya2, line 687:
```suggestion
            .collect();

        if missing_deps.is_empty() {
            return Some(Ok(verified_tx));
        }
```

mpguerra:
Is this the issue that was reported in Sentry?

arya2:
> Is this the issue that was reported in Sentry?

Yep, the issue was that dependencies of transactions in the mempool aren't cleared until they've been committed to the best chain, but there _could_ be multiple blocks being validated in parallel while the mempool is enabled, so the tx verifier needs to wait for those UTXOs as it would if the transaction hadn't already been verified for the mempool.

zebra-consensus/src/transaction.rs
```diff
@@ -645,15 +645,18 @@ where
...
     /// Attempts to find a transaction in the mempool by its transaction hash and checks
-    /// that all of its dependencies are available in the block.
+    /// that all of its dependencies are available in the block or in the state.  Waits
+    /// for UTXOs being spent by the given transaction to arrive in the state if they're
+    /// not found elsewhere.
     ///
     /// Returns [`Some(Ok(VerifiedUnminedTx))`](VerifiedUnminedTx) if successful,
     /// None if the transaction id was not found in the mempool,
     /// or `Some(Err(TransparentInputNotFound))` if the transaction was found, but some of its
     /// dependencies are missing in the block.
```
upbqdn:
```suggestion
    /// dependencies were not found in the block or state after a timeout.
```

zebra-consensus/src/transaction.rs
```diff
@@ -662,9 +665,10 @@ where
         let mempool = mempool?;
         let known_outpoint_hashes = req.known_outpoint_hashes();
         let tx_id = req.tx_mined_id();
+        let tx = req.transaction();
```
upbqdn: ###### Optional

```suggestion
```

zebra-consensus/src/transaction.rs
```diff
@@ -645,15 +645,18 @@ where
...
     /// None if the transaction id was not found in the mempool,
     /// or `Some(Err(TransparentInputNotFound))` if the transaction was found, but some of its
     /// dependencies are missing in the block.
-    async fn try_find_verified_unmined_tx(
+    async fn find_verified_unmined_tx(
         req: &Request,
         mempool: Option<Timeout<Mempool>>,
+        state: Timeout<ZS>,
     ) -> Option<Result<VerifiedUnminedTx, TransactionError>> {
         if req.is_mempool() || req.transaction().is_coinbase() {
```
upbqdn: ###### Optional

The `Request::transaction` fn clones the inner tx, so it's a bit more optimal to call it only once. This suggestion requires the one above.

```suggestion
        let tx = req.transaction();

        if req.is_mempool() || tx.is_coinbase() {
```

upbqdn (commented):
Do I understand it right that the tx dependencies must always originate from the mempool but can be moved to the state by the time we try to check if a tx can depend on them? In that case, should we update these docs: https://github.com/zcashfoundation/zebra/blob/2293cda36dfc74fe0200ff0682f4d2dc75e29881/zebra-node-services/src/mempool/transaction_dependencies.rs#L10-L19?

Should we also update this documentation: https://github.com/zcashfoundation/zebra/blob/2293cda36dfc74fe0200ff0682f4d2dc75e29881/zebra-state/src/request.rs#L691?

zebra-consensus/src/transaction.rs
arya2: The compiler complains with "temporary value dropped while borrowed".

arya2:
> Do I understand it right that the tx dependencies must always originate from the mempool but can be moved to the state by the time we try to check if a tx can depend on them?

Yep, tx dependencies are added when validating mempool txs with spends that are missing in the state but available in the mempool, and they're removed when transactions are mined onto the best chain, or if the mempool is reset. I've added a note to hint that the transaction verifier will need to wait for UTXOs to arrive in the state as well. I think it could still be much clearer if you have any suggestions.

zebra-consensus/src/transaction.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9118/commits/97a1c4052863fb6b90ca1b54a7aa8af2355a74ae.

zebra-consensus/src/transaction.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9118/commits/97a1c4052863fb6b90ca1b54a7aa8af2355a74ae.

zebra-consensus/src/transaction.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9118/commits/97a1c4052863fb6b90ca1b54a7aa8af2355a74ae.

upbqdn (approved):

PR #9119, opened 2025-01-14
fix(CI): Remove the "Release crates" workflow 

upbqdn:
## Motivation

Fix https://github.com/ZcashFoundation/zebra/actions/runs/12751834865/job/35539829251?pr=9114.

## Solution

- Remove the workflow and its associated bash script.

### Rationale

> It didn't catch publishing issues in the past, and it skips a bunch of things anyway.

> ..., it wasn't doing much with `--no-verify` anyway.

> It was working fine with the previous version of Rust. I think the latest stable version broke `--no-verify`, because it wasn't pulling dependencies from [crates.io](http://crates.io/) before (which is how we ended up with git sources on main when we were trying to publish Zebra), but now it does even if `--no-verify` is passed in explicitly.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2 (approved):
Thank you!

PR #9120, opened 2025-01-14
fix(ci): Remove the `release crates` workflow

upbqdn:
## Motivation

Fix https://github.com/ZcashFoundation/zebra/actions/runs/12751834865/job/35539829251?pr=9114.

This PR is an exact copy of #9119, which targetted #9114. This PR targets `main` for better commit history. 

## Solution

- Remove the workflow and its associated bash script.

### Rationale

> It didn't catch publishing issues in the past, and it skips a bunch of things anyway.

> ..., it wasn't doing much with `--no-verify` anyway.

> It was working fine with the previous version of Rust. I think the latest stable version broke `--no-verify`, because it wasn't pulling dependencies from [crates.io](http://crates.io/) before (which is how we ended up with git sources on main when we were trying to publish Zebra), but now it does even if `--no-verify` is passed in explicitly.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


arya2 (approved):

upbqdn:
More context: https://github.com/ZcashFoundation/zebra/pull/9114#issuecomment-2590911727.

gustavovalverde:
@mergify refresh

gustavovalverde:
@mergify dequeue

gustavovalverde:
@mergify requeue

gustavovalverde:
@mergify dequeue

gustavovalverde:
Something odd happened with this PR as it's not triggering some tests (maybe deleting a workflow has a different behavior), I'll admin merge it in the meanwhile and track this to see if it happens in another PR.

upbqdn:
The issue might be that this PR is an exact copy of #9119, which merged into #9114, which already merged into `main`. I forgot to explicitly prevent #9114 from merging.

PR #9126, opened 2025-01-15
change(consensus): Optimize checks for coinbase transactions

upbqdn:
## Motivation

The original motivation was to address https://github.com/ZcashFoundation/zebra/pull/9098#discussion_r1911778505 and https://github.com/ZcashFoundation/zebra/pull/9098#discussion_r1913455598. Then I noticed that the [`coinbase_outputs_are_decryptable`](https://github.com/zcashfoundation/zebra/blob/050cdc39e89ae9ccf25bbf6f0598f386cb7307a7/zebra-consensus/src/transaction/check.rs#L328) fn currently [converts](https://github.com/zcashfoundation/zebra/blob/050cdc39e89ae9ccf25bbf6f0598f386cb7307a7/zebra-chain/src/primitives/zcash_primitives.rs#L181) each post-Heartwood coinbase tx to check if its shielded outputs are decryptable, even if the tx has no shielded outputs. This conversion involves a serialization and deserialization cycle. We can skip this conversion for txs with no shielded outputs.

## Solution

- Return early if the tx has no shielded outputs.
- Return a new `Transaction::NotCoinbase` error if the tx is not coinbase but should be.
- Update docs.

### Tests

- Improve tests so they check more scenarios.

### PR Author's Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -83,6 +83,9 @@ pub enum TransactionError {
     #[error("non-coinbase transactions MUST NOT have coinbase inputs")]
     NonCoinbaseHasCoinbaseInput,
 
+    #[error("the tx is not coinbase, but it should be")]
+    NotCoinbase,
+
     #[error("transaction is locked until after block height {}", _0.0)]
     LockedUntilAfterBlockHeight(block::Height),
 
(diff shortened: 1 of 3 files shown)


arya2 (approved):
Looks good, thank you!

PR #9131, opened 2025-01-16
chore: delete repetitive words

rootdiae:


zebra-chain/src/block/serialize.rs
```diff
@@ -39,7 +39,7 @@ fn check_version(version: u32) -> Result<(), &'static str> {
         // but this is not actually part of the consensus rules, and in fact
         // broken mining software created blocks that do not have version 4.
         // There are approximately 4,000 blocks with version 536870912; this
-        // is the bit-reversal of the value 4, indicating that that mining pool
+        // is the bit-reversal of the value 4, indicating that mining pool
```
mpguerra:
```suggestion
        // is the bit-reversal of the value 4, indicating that that mining pool
```

mpguerra: This duplication is valid, it is referring to "that" specific mining pool

mpguerra (changes requested):
Thank you for submitting. There is one change that needs to be reverted

rootdiae:
Okay, I have reverted it. :)

mpguerra (approved):

mpguerra:
@mergify refresh

PR #9139, opened 2025-01-18
fix(consensus): Verify consensus branch ID in SIGHASH precomputation

### LINKED ISSUE #9089: The tx verifier does not enforce using the correct consensus branch ID in the SIGHASH precomputation

upbqdn:
## Motivation

Close #9089.

#### [§ 4.10 SIGHASH Transaction Hashing](https://zips.z.cash/protocol/protocol.pdf#sighash)                                                                                
                                                                                           
> [NU5 only, pre-NU6] All transactions MUST use the NU5 consensus branch ID `0xF919A198` as defined in [ZIP-252].                                                                    
                                                                                           
> [NU6 only] All transactions MUST use the NU6 consensus branch ID `0xC8E71055` as defined in [ZIP-253].

[ZIP-252]: <https://zips.z.cash/zip-0252>
[ZIP-253]: <https://zips.z.cash/zip-0253>                                                  

## Solution

- Implement a check ensuring compliance with the consensus rules listed above. This check makes the tx SIGHASH precomputation panic if the tx contains an invalid value in its `nConsensusBranchId` field. This precomputation is implemented in `zebra-chain`, and is used by the tx verifier in `zebra-consensus` and by the transparent script verifier in `zebra-script`, but this verifier is only called from the tx verifier. It might be worth optimizing this, but that is currently not possible, as described in #9165. The panic should never occur in regular use because the tx verifier implements another consensus branch ID check according to another consensus rule implemented here: https://github.com/zcashfoundation/zebra/blob/05a9b6171ccbc06bcc888470e4be8a60b1d574ef/zebra-consensus/src/transaction/check.rs#L513-L552, which takes effect early on and doesn't panic.
- Add new error types in `zebra-chain`.
- Pass `NetworkUpgrade` instead of `ConsensusBranchId` to the SIGHASH precomputation and tx conversion.
- Use `TryFrom` instead of ad-hoc functions for conversion.
- Remove a redundant `TryFrom<&Transaction> for zcash_primitives::transaction::Transaction`.
- Document that we now implement the consensus rules listed above.
- Impl `NetworkUpgrade::previous_upgrade` and refactor `NetworkUpgrade::next_upgrade` so that it doesn't explicitly enumerate all upgrades.
- Don't make `NETWORK_UPGRADES_IN_ORDER` `pub`.

### Tests

- Add tests that check that tx SIGHASH precomputation and computation fail when a tx attempts to use an invalid consensus branch id. 

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist


zebrad/src/components/mempool/downloads.rs
```diff
@@ -373,6 +373,7 @@ where
                 .oneshot(tx::Request::Mempool {
                     transaction: tx.clone(),
                     height: next_height,
+                    skip_checks: None
```
upbqdn: Why does guarding this by `#[cfg(any(test, feature = "proptest-impl"))]` not work, and why does it work as it is since `skip_checks` is guarded by the same attribute in `zebra-consensus`?

upbqdn: Turns out it doesn't. I have to understand better how feature assignment behaves during compilation. And it's probably not a good idea to do it this way anyway.

upbqdn: I used this hack to solve this: https://github.com/zcashfoundation/zebra/blob/9597792567982ae19643eba0d2537d3a81fe5956/zebrad/Cargo.toml#L301-L304.

upbqdn: I'm trying to think of something better in general.

zebra-consensus/src/transaction.rs
```diff
@@ -575,13 +600,13 @@ where
...
                     let transaction = VerifiedUnminedTx::new(
-                        transaction,
-                        miner_fee.expect(
-                            "unexpected mempool coinbase transaction: should have already rejected",
-                        ),
+                        tx.clone(),
+                        miner_fee.expect("fee should have been checked earlier"),
                         legacy_sigop_count,
+                        #[cfg(any(test, feature = "proptest-impl"))]
+                        req.skip_checks().is_some_and(|checks| checks.contains(&SkipCheck::Zip317))
```
oxarbitrage: I am not sure why but i remember seeing other odd behaviors when using `proptest-impl`. Here is one option to fix it:


```suggestion
                        #[cfg(any(test, feature = "proptest-impl"))]
                        req.skip_checks().is_some_and(|checks| checks.contains(&SkipCheck::Zip317)),
                        #[cfg(not(any(test, feature = "proptest-impl")))]
                        false,
```

upbqdn: Nope.

oxarbitrage: Yea, sorry about that. I will try something else.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -350,17 +350,18 @@ impl NetworkUpgrade {
...
+            .filter(|&p| (p + 1) < NETWORK_UPGRADES_IN_ORDER.len())
+            .map(|p| NETWORK_UPGRADES_IN_ORDER[p + 1])
+    }
+
+    /// Returns the previous network before after this network upgrade
+    pub fn previous_upgrade(self) -> Option<Self> {
+        Self::iter()
+            .position(|nu| self == nu)
+            .filter(|&p| p > 0)
+            .map(|p| NETWORK_UPGRADES_IN_ORDER[p - 1])
```
arya2: Optional/Nitpick, requires a suggestion below:

```suggestion
        Self::iter().skip_while(|&nu| self != nu).nth(1)
    }

    /// Returns the previous network before after this network upgrade
    pub fn previous_upgrade(self) -> Option<Self> {
        Self::iter().rev().skip_while(|&nu| self != nu).nth(1)
```

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -538,6 +539,11 @@ impl NetworkUpgrade {
     ) -> Duration {
         NetworkUpgrade::current(network, height).averaging_window_timespan()
     }
+
+    /// Returns an iterator over [`NetworkUpgrade`] variants.
+    pub fn iter() -> impl Iterator<Item = NetworkUpgrade> {
```
arya2: Required for the suggestion above
```suggestion
    pub fn iter() -> impl DoubleEndedIterator<Item = NetworkUpgrade> {
```

zebra-chain/src/parameters/network_upgrade.rs
upbqdn: Amazing. Thanks.

zebra-chain/src/transaction/tests/vectors.rs
```diff
@@ -801,6 +803,46 @@ fn zip244_sighash() -> Result<()> {
...
+                .expect_err("tx is not convertible under nu other than the tx one");
+
+            std::panic::catch_unwind(|| PrecomputedTxData::new(&tx, any_other_nu, &[]))
+                .expect_err("precomputing tx sighash data panics under nu other than the tx one");
+
+            std::panic::catch_unwind(|| sighash::SigHasher::new(&tx, any_other_nu, &[]))
+                .expect_err("creating the sighasher panics under nu other than the tx one");
+
+            std::panic::catch_unwind(|| tx.sighash(any_other_nu, HashType::ALL, &[], None))
+                .expect_err("the sighash computation panics under nu other than the tx one");
```
arya2, line 841: Could you mention this in the method docs for `sighash()` and `sighasher()` too?

upbqdn: This is a bit more nuanced. It won't panic for txs that don't contain the `nConsensusBranchId` field. This test uses only those, so the message here can be simplified. Other reasons for panicking can be issues with obtaining a librustzcash version of the consensus branch id from the passed network upgrade. 

This is documented in the corresponding functions: https://github.com/zcashfoundation/zebra/blob/2e32fd688c53e6956d0a7447bf750f37a7df0a99/zebra-chain/src/transaction.rs#L1237 and https://github.com/zcashfoundation/zebra/blob/2e32fd688c53e6956d0a7447bf750f37a7df0a99/zebra-chain/src/primitives/zcash_primitives.rs#L243.
We usually document panics only in the functions from which they originate, so I thought it was fine as it was. 

I can document it in those functions you mentioned, but then I would have to describe the whole story, which feels too repetitive. Would you prefer that?

arya2: > We usually document panics only in the functions from which they originate, so I thought it was fine as it was.

I think it'd be nice to say "Panics on parser errors or network upgrade mismatches" with a reference to the docs for `PrecomputedTxData::new()` + `to_librustzcash()`, but I also think it's fine as-is and definitely shouldn't be repeated.

upbqdn: Addressed in 3b25023db93e708837da4e6d625810e932e81869.

zebra-consensus/src/transaction.rs
```diff
@@ -575,12 +575,10 @@ where
...
                     legacy_sigop_count,
                 },
-                Request::Mempool { transaction, .. } => {
+                Request::Mempool { transaction: ref tx, .. } => {
                     let transaction = VerifiedUnminedTx::new(
-                        transaction,
-                        miner_fee.expect(
-                            "unexpected mempool coinbase transaction: should have already rejected",
-                        ),
+                        tx.clone(),
```
arya2, line 580: Why this change?

zebra-script/src/lib.rs
```diff
@@ -451,13 +440,13 @@ mod tests {
...
 
-        verifier.is_valid(branch_id, input_index + 1).unwrap_err();
-
-        verifier.is_valid(branch_id, input_index + 1).unwrap_err();
+        verifier
+            .is_valid(NetworkUpgrade::Blossom, input_index + 1)
+            .expect_err("verification should fail");
+        verifier
+            .is_valid(NetworkUpgrade::Blossom, input_index + 1)
+            .expect_err("verification should fail");
```
```suggestion
        verifier
            .is_valid(NetworkUpgrade::Blossom, input_index + 1)
            .expect_err("verification should fail");
```

arya2 (dismissed):
Looks good, thank you!

oxarbitrage (approved):

zebra-script/src/lib.rs
upbqdn: This shouldn't have been merged. The verifier is supposed to fail twice. It's in the name of the test, similarly to other tests in the file.

zebra-consensus/src/transaction.rs
upbqdn: Thanks for spotting this. It was required for the conditional compilation I had earlier, which is gone now. I'll revert it.

zebra-script/src/lib.rs
upbqdn: I'll fix it in another PR.

oxarbitrage: sorry about that, i thought it was an oversight.

PR #9176, opened 2025-01-29
fix(mining): Advertise mined blocks

### LINKED ISSUE #8909: bug: Zebra does not advertise mined blocks in some cases
### What happened?

 I expected to see this happen: Zebra should advertise blocks it mines

Instead, this happened: it doesn't if the node is ahead in the network and it's the sole miner.

Mining on the testnet

### Zebra logs

_No response_

### Zebra Version

1.9.0 (restore-internal-miner branch)

- [X] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

Zebra has a gossip task in `gossip.rs` that repeatedly waits until it's close to the tip and advertises the tip block. However, in `SyncStatus::wait_until_close_to_tip()` which is used to that, it calls `is_close_to_tip()` which relies on the last synced block lengths to tell if it's close to the tip. But if the node is the ahead of the other nodes (e.g. it's the sole miner) then it won't ever receive blocks and `is_close_to_tip()` will return false.

I'm not sure what is the best fix for this.

oxarbitrage:
## Motivation

When blocks are mined and we are not close to the tip the mined block do not get advertised to the network.

Close https://github.com/ZcashFoundation/zebra/issues/8909

## Solution

Introduce a channel with a sender part used by the rpc method submit block to add a signal when a block is submitted and accepted. The receiver part of the channel is used by the block gossip task to advertise the block.

### Tests

- A `submitblock_channel` was added to check the gossip task can receive mined blocks.
- A check was added to existing test `nu6_funding_streams_and_coinbase_balance` to make sure the sender part of the channel is working.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebrad/src/components/inbound/tests/real_peer_set.rs
+++ b/zebrad/src/components/inbound/tests/real_peer_set.rs
@@ -796,7 +796,7 @@ async fn setup(
 mod submitblock_test {
     use std::io;
     use std::sync::{Arc, Mutex};
-    use tracing::Level;
+    use tracing::{Instrument, Level};
     use tracing_subscriber::fmt;
 
     use super::*;
@@ -872,12 +872,15 @@ mod submitblock_test {
 
         // Start the block gossip task with a SubmitBlockChannel
         let submitblock_channel = SubmitBlockChannel::new();
-        let gossip_task_handle = tokio::spawn(sync::gossip_best_tip_block_hashes(
-            sync_status.clone(),
-            chain_tip_change,
-            peer_set.clone(),
-            Some(submitblock_channel.receiver()),
-        ));
+        let gossip_task_handle = tokio::spawn(
+            sync::gossip_best_tip_block_hashes(
+                sync_status.clone(),
+                chain_tip_change,
+                peer_set.clone(),
+                Some(submitblock_channel.receiver()),
+            )
+            .in_current_span(),
+        );
 
         // Send a block top the channel
         submitblock_channel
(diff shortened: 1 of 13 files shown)


zebrad/src/components/inbound/tests/real_peer_set.rs
```diff
@@ -782,3 +791,112 @@ async fn setup(
...
+    use std::sync::{Arc, Mutex};
+    use tracing::Level;
+    use tracing_subscriber::fmt;
+
+    use super::*;
+
+    use crate::components::sync::PEER_GOSSIP_DELAY;
+
+    // Custom in-memory writer to capture logs
+    struct TestWriter(Arc<Mutex<Vec<u8>>>);
```
arya2, line 807: Could you leave a note about the conditional compilation issue or how this is used?

arya2 (dismissed):
This looks great, thank you for these changes. It's more complicated than what I had in mind, but it makes it easy to now handle mined blocks differently and broadcast them to all ready peers instead of a third of them.

I opened a suggestion PR, https://github.com/ZcashFoundation/zebra/pull/9183, for your review. Ideally this PR could advertise mined blocks immediately and return early from `check_synced_to_tip()` on test networks, but I see no blockers to merging this as-is.

arya2 (approved):

zebrad/src/components/inbound/tests/real_peer_set.rs
mpguerra: @oxarbitrage I believe this conversation needs to be resolved before this PR can merge

PR #9183, opened 2025-01-31
Suggestions for "fix(mining): Advertise mined blocks"

### LINKED ISSUE #8909: bug: Zebra does not advertise mined blocks in some cases

arya2:
This is a suggestion PR for https://github.com/ZcashFoundation/zebra/pull/9176, it:
- updates the documentation on a field,
- refactors RPC error conversions,
- skips checking if Zebra is synced to the network tip on test networks, and
- refactors block gossip task to advertise mined blocks as soon as they're received.

### DIFF
--- a/zebra-rpc/src/methods/get_block_template_rpcs.rs
+++ b/zebra-rpc/src/methods/get_block_template_rpcs.rs
@@ -64,7 +64,10 @@ use crate::{
         hex_data::HexData,
         GetBlockHash,
     },
-    server::{self, error::MapError},
+    server::{
+        self,
+        error::{MapError, OkOrError},
+    },
 };
 
 pub mod constants;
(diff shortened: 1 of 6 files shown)


oxarbitrage (approved):
Thanks, adding this in.

PR #9197, opened 2025-02-03
build(deps): update yanked crate metrics-exporter-prometheus

rex4539:


arya2:
We'll update this in our next dependabot PR.

PR #9198, opened 2025-02-03
chore: Minor cleanups

### LINKED ISSUE #9089: The tx verifier does not enforce using the correct consensus branch ID in the SIGHASH precomputation

upbqdn:
## Motivation

Resolve https://github.com/ZcashFoundation/zebra/pull/9139#discussion_r1938092454 and https://github.com/ZcashFoundation/zebra/pull/9139#discussion_r1938089784 and a few other nits.

### Tests

N/A

### PR Author's Checklist

- [x] The solution is tested.

### DIFF
--- a/zebrad/src/commands/start.rs
+++ b/zebrad/src/commands/start.rs
@@ -262,7 +262,7 @@ impl StartCmd {
                 );
                 rpc_task_handle.await.unwrap()
             } else {
-                warn!("configure an listen_addr to start the RPC server");
+                info!("configure an listen_addr to start the RPC server");
                 (
                     tokio::spawn(std::future::pending().in_current_span()),
                     tokio::spawn(std::future::pending().in_current_span()),
--- a/zebrad/src/commands/start.rs
+++ b/zebrad/src/commands/start.rs
@@ -262,7 +262,7 @@ impl StartCmd {
                 );
                 rpc_task_handle.await.unwrap()
             } else {
-                info!("configure an listen_addr to start the RPC server");
+                info!("configure a listen_addr to start the RPC server");
                 (
                     tokio::spawn(std::future::pending().in_current_span()),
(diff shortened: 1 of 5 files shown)


zebrad/src/commands/start.rs
```diff
@@ -262,7 +262,7 @@ impl StartCmd {
                 );
                 rpc_task_handle.await.unwrap()
             } else {
-                warn!("configure an listen_addr to start the RPC server");
+                info!("configure an listen_addr to start the RPC server");
```
oxarbitrage:
```suggestion
                info!("configure a listen_addr to start the RPC server");
```

oxarbitrage (dismissed):

oxarbitrage (approved):

arya2 (approved):

PR #9201, opened 2025-02-04
feat(network): Track misbehaving peer connections and ban them past a threshold

### LINKED ISSUE #9111: Score and drop misbehaving peer connections
## Motivation

It's currently trivial to modify or configure Zebra or zcashd so that the node doesn't follow the consensus rules on Mainnet or the default Testnet.

This is mitigated while there are many zcashd nodes in the network because zcashd tracks a "bad score" which is incremented when a peer misbehaves (e.g. gossiping semantically invalid mempool transactions or blocks).

Zebra will need a similar mechanism before it becomes the only node implementation on the network.

### Overview

Advertisement of semantically invalid blocks or mempool transactions by peer connections should be considered misbehaviour and should be scored.

Excessive misbehaviour should result in a peer connection being dropped and its IP being banned from re-connecting to the Zebra node.

Bans should eventually expire, and there should be a limit on how many IPs are banned from connecting to a Zebra node at any given time.

#### Update Address Book

- Add a `misbehavior_score` field to `MetaAddr`
- Add a variant to `MetaAddrChange` for incrementing the misbehaviour score,
- Update the `AddressBook::update()` method to apply that change,
- Avoid responding to `GetAddr` requests with addresses of misbehaving peers,
- Avoid inbound or outbound connections to misbehaving peers, and
- Avoid caching addresses of misbehaving peers or include their misbehaviour score in the address cache

#### Update `zebra-network` Types

- Add an `advertisers` field of type `Vec<SocketAddr>` to the `Transactions` and `Blocks` variants of `zebra_network::Response`,
- Add the address book updater and a `HashMap<SocketAddr, MisbehaviorScore>` as fields on `PeerSet` 
- Add a network `Request` variant to be used by `ChainSync` and the mempool's `Downloads` stream to report advertisers of semantically invalid blocks or mempool transactions, and
- Handle the new request variant by sending an update to the address book and updating the `PeerSet`'s misbehaviour score map
  - If the updated misbehaviour score is excessive, drop the peer connection
  
#### Call `PeerSet` With Misbehaving Peer Addresses

- When blocks fail semantic validation, call the peer set with a list of peers that advertised the semantically invalid block from the `handle_block_response()` and wherever else blocks are sent to the block verifier, and
- When mempool transactions fail semantic validation, call the peer set with a list of peers that advertised the semantically invalid mempool transactions from the mempool's `Downloads` stream.

arya2:
## Motivation

Zebra should drop connections to and avoid reconnecting with peers which advertise blocks or mempool transactions that fail some semantic validation checks.

Closes #9111.
Depends-On: #9224.

## Solution

Channels:
- Adds a new mpsc channel for the mempool, inbound svc, and syncer to report misbehaviour to the address book,
- Adds a watch channel for the address book to broadcast the latest bans to the peer set and inbound connector
- Adds a task for batching updates to addresses' misbehavior scores

Type Updates:
- Adds a `misbehavior_score` field on `MetaAddr`, a `bans_by_ip` field to `AddressBook`, and a `UpdateMisbehavior` variant to `MetaAddrChange`
- Updates `MetaAddrChange.apply_to_meta_addr()` to apply `UpdateMisbehavior` changes
- Adds the `PeerSocketAddr` of the peer that provided a block or transaction to the `Transaction` and `Block` variants of the peer set's response type
- Updates error types for mempool, inbound service, and syncer `Downloads` components to include the peer address from which the block or transaction originated
- Passes the new channel for reporting misbehaviour to the mempool, inbound service, and syncer

Method Updates:
- Updates `AddressBook.update()` to remove any addresses that exceed a misbehavior score threshold and add them to `bans_by_ip`
  - Removes oldest banned IPs if `bans_by_ip` exceeds its maximum size
  - Removing the addresses from the `by_addr` field avoids attempting outbound connections to them as candidates,
- Updates `sanitized()` method to avoid providing addresses with a non-zero misbehavior score in responses to `GetAddr` requests
- Adds `misbehavior_score()` and `mempool_misbehavior_score()` methods on `TransactionError`, `VerifyBlockError`, and `BlockError`
- Updates mempool, inbound service, and syncer to send `UpdateMisbehavior` messages to the address book updater when blocks or transactions fail some checks during semantic validation
- Updates peer set's `poll_ready()` method to drop connections to banned IPs

### Tests

Adds an acceptance test which spawns an instance of zebrad on a Testnet where PoW is disabled, and one on the default Testnet. Checks that they've connected, mines blocks via the `generate()` RPC, and checks that they've disconnected.

## Follow Up Work

Increment the misbehaviour score for more errors and adjust the amounts based on zcashd (though at a glance, it looks similar already).

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -100,6 +100,17 @@ impl VerifyBlockError {
             _ => false,
         }
     }
+
+    /// Returns a suggested misbehaviour score increment for a certain error.
+    pub fn misbehavior_score(&self) -> u32 {
+        // TODO: Adjust these values based on zcashd.
+        use VerifyBlockError::*;
+        match self {
+            Block { source } => source.misbehavior_score(),
+            Equihash { .. } => 100,
+            _other => 0,
+        }
+    }
 }
 
 /// The maximum allowed number of legacy signature check operations in a block.
--- a/zebra-consensus/src/block.rs
+++ b/zebra-consensus/src/block.rs
@@ -103,7 +103,7 @@ impl VerifyBlockError {
 
     /// Returns a suggested misbehaviour score increment for a certain error.
     pub fn misbehavior_score(&self) -> u32 {
-        // TODO: Adjust these values based on zcashd.
+        // TODO: Adjust these values based on zcashd (#9258).
         use VerifyBlockError::*;
         match self {
             Block { source } => source.misbehavior_score(),
--- a/zebra-consensus/src/checkpoint/list.rs
+++ b/zebra-consensus/src/checkpoint/list.rs
@@ -67,7 +67,7 @@ impl ParameterCheckpoint for Network {
         let (checkpoints_for_network, should_fallback_to_genesis_hash_as_checkpoint) = match self {
             Network::Mainnet => (MAINNET_CHECKPOINTS, false),
             Network::Testnet(params) if params.is_default_testnet() => (TESTNET_CHECKPOINTS, false),
-            Network::Testnet(_params) => (TESTNET_CHECKPOINTS, true),
+            Network::Testnet(_params) => ("", true),
         };
 
         // Check that the list starts with the correct genesis block and parses checkpoint list.
--- a/zebra-network/src/peer_set/initialize.rs
+++ b/zebra-network/src/peer_set/initialize.rs
@@ -7,7 +7,7 @@
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     convert::Infallible,
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     pin::Pin,
     sync::Arc,
     time::Duration,
@@ -19,10 +19,11 @@ use futures::{
     stream::{FuturesUnordered, StreamExt},
     Future, TryFutureExt,
 };
+use indexmap::IndexMap;
 use rand::seq::SliceRandom;
 use tokio::{
     net::{TcpListener, TcpStream},
-    sync::{broadcast, mpsc},
+    sync::{broadcast, mpsc, watch},
     time::{sleep, Instant},
 };
 use tokio_stream::wrappers::IntervalStream;
@@ -110,8 +111,13 @@ where
 {
     let (tcp_listener, listen_addr) = open_listener(&config.clone()).await;
 
-    let (address_book, address_book_updater, address_metrics, address_book_updater_guard) =
-        AddressBookUpdater::spawn(&config, listen_addr);
+    let (
+        address_book,
+        bans_receiver,
+        address_book_updater,
+        address_metrics,
+        address_book_updater_guard,
+    ) = AddressBookUpdater::spawn(&config, listen_addr);
 
     let (misbehavior_tx, mut misbehavior_rx) = mpsc::channel(
         3 * config
@@ -230,6 +236,7 @@ where
         constants::MIN_INBOUND_PEER_CONNECTION_INTERVAL,
         listen_handshaker,
         peerset_tx.clone(),
+        bans_receiver,
     );
     let listen_guard = tokio::spawn(listen_fut.in_current_span());
 
@@ -592,6 +599,7 @@ async fn accept_inbound_connections<S>(
     min_inbound_peer_connection_interval: Duration,
     handshaker: S,
     peerset_tx: futures::channel::mpsc::Sender<DiscoveredPeer>,
+    bans_receiver: watch::Receiver<Arc<IndexMap<IpAddr, std::time::Instant>>>,
 ) -> Result<(), BoxError>
 where
     S: Service<peer::HandshakeRequest<TcpStream>, Response = peer::Client, Error = BoxError>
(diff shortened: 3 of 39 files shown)


zebra-consensus/src/checkpoint/list.rs
```diff
@@ -67,7 +67,7 @@ impl ParameterCheckpoint for Network {
         let (checkpoints_for_network, should_fallback_to_genesis_hash_as_checkpoint) = match self {
             Network::Mainnet => (MAINNET_CHECKPOINTS, false),
             Network::Testnet(params) if params.is_default_testnet() => (TESTNET_CHECKPOINTS, false),
-            Network::Testnet(_params) => (TESTNET_CHECKPOINTS, true),
+            Network::Testnet(_params) => ("", true),
```
arya2, line 70: Should we add a field on testnet params instead of changing the behaviour here?

Update: Let's wait until someone complains.

oxarbitrage:
The code from https://github.com/ZcashFoundation/zebra/pull/9224 is still being showed in the diff of this PR. It will be good to rebase to make the diff smaller.

zebra-consensus/src/block.rs
```diff
@@ -100,6 +100,17 @@ impl VerifyBlockError {
             _ => false,
         }
     }
+
+    /// Returns a suggested misbehaviour score increment for a certain error.
+    pub fn misbehavior_score(&self) -> u32 {
+        // TODO: Adjust these values based on zcashd.
```
oxarbitrage:
```suggestion
        // TODO: Adjust these values based on zcashd. #XXXX
```

Should we create an issue for this ? I see the TODO is repeated a few times so it looks like something we might want or need to do at some point.

zebra-network/src/peer_set/initialize.rs
```diff
@@ -109,8 +111,54 @@ where
...
+    let (
+        address_book,
+        bans_receiver,
+        address_book_updater,
+        address_metrics,
+        address_book_updater_guard,
+    ) = AddressBookUpdater::spawn(&config, listen_addr);
+
+    let (misbehavior_tx, mut misbehavior_rx) = mpsc::channel(
+        3 * config
```
oxarbitrage: Can we explain why the `3` is used ? Maybe by moving it to a constant or just with a comment.

zebra-network/src/peer_set/initialize.rs
```diff
@@ -109,8 +111,54 @@ where
...
+            .peerset_total_connection_limit()
+            .max(MIN_CHANNEL_SIZE),
+    );
+
+    let misbehaviour_updater = address_book_updater.clone();
+    tokio::spawn(
+        async move {
+            let mut misbehaviors: HashMap<PeerSocketAddr, u32> = HashMap::new();
+            let mut flush_timer =
+                IntervalStream::new(tokio::time::interval(Duration::from_secs(60)));
```
oxarbitrage: It will be good to explain why a 1 minute interval is used here.

zebra-network/src/peer_set/set.rs
```diff
@@ -475,6 +484,12 @@ where
                 Some(Ok((key, svc))) => {
                     trace!(?key, "service became ready");
 
+                    if self.bans_receiver.borrow().contains_key(&key.ip()) {
+                        debug!(?key, "service is banned, dropping service");
```
oxarbitrage:
```suggestion
                        warn!(?key, "service is banned, dropping service");
```

We don't except bans to happen very often unless there is something really wrong with the peers. Should we increase the log level ?

oxarbitrage (dismissed):
Looks really good for a pretty big chunk of work. I left a few minor comments here and there.

zebra-network/src/peer_set/set.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9201/commits/88e0c9c0405ee533f6308394928a3f8de9f161d0.

zebra-network/src/peer_set/initialize.rs
arya2: This one's not really important until/unless we add finer-grained misbehaviour scores. Added a comment explaining why it's there in https://github.com/ZcashFoundation/zebra/pull/9201/commits/88e0c9c0405ee533f6308394928a3f8de9f161d0.

zebra-network/src/peer_set/initialize.rs
arya2: It's actually not necessary, I added a comment explaining why the peerset limit is used in https://github.com/ZcashFoundation/zebra/pull/9201/commits/88e0c9c0405ee533f6308394928a3f8de9f161d0.

zebra-consensus/src/block.rs
arya2: I created issue #9258 for it and referenced it in the TODOs in https://github.com/ZcashFoundation/zebra/pull/9201/commits/88e0c9c0405ee533f6308394928a3f8de9f161d0.

oxarbitrage (approved):
Thanks!

PR #9214, opened 2025-02-07
rpc(getpeerinfo): Add inbound peers to method response

### LINKED ISSUE #7893: bug: `getpeerinfo` RPC should show inbound connections as well as outbound
### Motivation

`getpeerinfo` is documented to show inbound and outbound connections, but Zebra only shows outbound connections.

Also, it shows current and recent outbound connections, but it should only show current connections.

### Specifications

https://zcash.github.io/rpc/getpeerinfo.html

We can add a network request for a list of peers, and then use that data for the RPC.

We don't need the address book, maybe we could remove it.

This change might need to be based on PR #7859, because it makes significant changes to the peer set.

### Testing

Connect to Zebra with an inbound connection and make sure it shows up. We can do this in a unit test, or manually.
Add the `inbound` field to the RPC so we're sure.

### Follow Up Tickets

- Change the name of `AddressBook` to `OutboundAddressBook`
- Update `Private Testnet Test` page of the Zebra book

oxarbitrage:
## Motivation

We want to return outbound and inbound peer addresses in the `getpeerinfo` response. We currently return only outbound connections as these were the only ones we have access to by the address book. After https://github.com/ZcashFoundation/zebra/pull/9201 was merged, we have inbound connections available in the `MetaAddr` struct so we can use that to return peers of both connection types in the RPC response.

Close https://github.com/ZcashFoundation/zebra/issues/7893

## Solution

- Use `is_inbound()` to know if a peer is inbound/outbound.
- Add a `currently_live_peers()` method to the `AddressBookPeers` trait so we can return peers that are likely connected to us now instead of the ones that were responding before.

### Tests

Existing getpeerinfo tests were updated to reflect the change.

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


arya2 (commented):
This looks like it should work, but it seems unnecessary to add another address book.

I made an alternative change in https://github.com/ZcashFoundation/zebra/pull/9201/commits/a9014f0db08b98b9ee5a9bfa13341965f6980d96 of https://github.com/ZcashFoundation/zebra/pull/9201 if that looks acceptable.

Related: @mpguerra It would be nice to eventually refactor the address book into a service, it may be easier to do that ahead of adding fields to the `getpeerinfo` response.

oxarbitrage:
> This looks like it should work, but it seems unnecessary to add another address book.
> 
> I made an alternative change in [a9014f0](https://github.com/ZcashFoundation/zebra/commit/a9014f0db08b98b9ee5a9bfa13341965f6980d96) of #9201 if that looks acceptable.

I don't think that is correct. For the `getpeerinfo` RPC method, when displaying an inbound connection it looks to me that the caller will expect to see the ephemeral port of the peer and not the default p2p port.

arya2:
> I don't think that is correct. For the `getpeerinfo` RPC method, when displaying an inbound connection it looks to me that the caller will expect to see the ephemeral port of the peer and not the default p2p port.

Ah, that makes sense. I replaced that commit with https://github.com/ZcashFoundation/zebra/pull/9201/commits/0bf77a6aca9acb811dc67838eb0f2c4763c6cf9c and https://github.com/ZcashFoundation/zebra/pull/9201/commits/f19b769b14ad4b41d54d92f6f9e937d4c76c6170, which should return the ephemeral port and track which peer addresses were added when they were making an inbound connection so that could be included in the `getpeerinfo` response and so ~their IPs can be advertised with the default network port~ they can be excluded when responding to GetAddr requests.

conradoplg:
This looks good overall, but I'll wait for your decision on Arya's feedback before taking another look.

One thing I noticed is that in the progress bar we get a "Never Attempted Peers (inbound)" which I think does not make sense and it will be always zero. Not sure about "Failed Peers (inbound)" which I think might be always zero too. I think the way to fix this would be to use `Option`s for each bar in `AddressBookUpdater::spawn()` and only create `never_bar` and `failed_bar` for the outbound book.

oxarbitrage:
> This looks good overall, but I'll wait for your decision on Arya's feedback before taking another look.

Sorry, i forgot to write an update here. We are going to rebase this on https://github.com/ZcashFoundation/zebra/pull/9201 after it gets merged, which already has code for inbound connections in the address book, making the changes needed for the RPC easier here.

zebra-network/src/address_book.rs
```diff
@@ -819,6 +819,16 @@ impl AddressBookPeers for AddressBook {
...
+
+    fn currently_live_peers(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
+        let _guard = self.span.enter();
+
+        self.by_addr
+            .descending_values()
+            .filter(|peer| peer.has_connection_recently_responded(now))
+            .cloned()
+            .collect()
+    }
```
arya2:
```suggestion
```

zebra-network/src/address_book.rs
```diff
@@ -827,6 +837,12 @@ impl AddressBookPeers for Arc<Mutex<AddressBook>> {
             .expect("panic in a previous thread that was holding the mutex")
             .recently_live_peers(now)
     }
+
+    fn currently_live_peers(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr> {
+        self.lock()
+            .expect("panic in a previous thread that was holding the mutex")
+            .currently_live_peers(now)
+    }
```
arya2:
```suggestion
```

zebra-network/src/address_book_peers.rs
```diff
@@ -14,4 +14,7 @@ pub use mock::MockAddressBookPeers;
 pub trait AddressBookPeers {
     /// Return an Vec of peers we've seen recently, in reconnection attempt order.
     fn recently_live_peers(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr>;
+
+    /// Return an Vec of peers that are likely to be live, in reconnection attempt order.
+    fn currently_live_peers(&self, now: chrono::DateTime<Utc>) -> Vec<MetaAddr>;
```
arya2:
```suggestion
```

zebra-network/src/address_book_peers/mock.rs
```diff
@@ -7,13 +7,17 @@ use crate::{meta_addr::MetaAddr, AddressBookPeers};
 pub struct MockAddressBookPeers {
     /// Return value for mock `recently_live_peers` method.
     recently_live_peers: Vec<MetaAddr>,
+
+    /// Return value for mock `currently_live_peers` method.
+    currently_live_peers: Vec<MetaAddr>,
```
arya2:
```suggestion
```

zebra-network/src/address_book_peers/mock.rs
```diff
@@ -7,13 +7,17 @@ use crate::{meta_addr::MetaAddr, AddressBookPeers};
...
 }
 
 impl MockAddressBookPeers {
     /// Creates a new [`MockAddressBookPeers`]
-    pub fn new(recently_live_peers: Vec<MetaAddr>) -> Self {
+    pub fn new(recently_live_peers: Vec<MetaAddr>, currently_live_peers: Vec<MetaAddr>) -> Self {
         Self {
             recently_live_peers,
+            currently_live_peers,
```
arya2:
```suggestion
    pub fn new(recently_live_peers: Vec<MetaAddr>) -> Self {
        Self {
            recently_live_peers,
```

zebra-network/src/address_book_peers/mock.rs
```diff
@@ -22,4 +26,8 @@ impl AddressBookPeers for MockAddressBookPeers {
     fn recently_live_peers(&self, _now: chrono::DateTime<chrono::Utc>) -> Vec<MetaAddr> {
         self.recently_live_peers.clone()
     }
+
+    fn currently_live_peers(&self, _now: chrono::DateTime<chrono::Utc>) -> Vec<MetaAddr> {
+        self.currently_live_peers.clone()
+    }
```
arya2:
```suggestion
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -1101,10 +1101,11 @@ where
 
     async fn get_peer_info(&self) -> Result<Vec<PeerInfo>> {
         let address_book = self.address_book.clone();
+
         Ok(address_book
-            .recently_live_peers(chrono::Utc::now())
+            .currently_live_peers(chrono::Utc::now())
```
```suggestion
            .recently_live_peers(chrono::Utc::now())
```

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -1101,10 +1101,11 @@ where
...
             .into_iter()
-            .map(PeerInfo::from)
+            .map(|meta_addr| PeerInfo::new(meta_addr, meta_addr.is_inbound()))
```
arya2: Optional/Nitpick:

```suggestion
            .map(PeerInfo::new)
```

zebra-rpc/src/methods/get_block_template_rpcs/types/peer_info.rs
```diff
@@ -7,12 +7,26 @@ use zebra_network::{types::MetaAddr, PeerSocketAddr};
...
+}
+
+impl PeerInfo {
+    /// Create a new `PeerInfo` from a `MetaAddr`
+    pub fn new(meta_addr: MetaAddr, inbound: bool) -> Self {
+        Self {
+            addr: meta_addr.addr(),
+            inbound,
+        }
+    }
```
arya2: Optional/Nitpick:

```suggestion
    pub fn new(meta_addr: MetaAddr) -> Self {
        Self::from(meta_addr)
    }
```

arya2 (commented):
This is looking good, I left some suggestions for removing `currently_live_peers()` as I think Zebra's `getpeerinfo` should only be returning addresses where `self.last_connection_state == PeerAddrState::Responded`.

Note: `MockAddressBookPeers::new()` would take one parameter if the suggestions are applied but would still called with two in a couple tests, Github wouldn't let me make that suggestion.

oxarbitrage:
I don't mind doing those suggestions to remove `currently_live_peers`, it will simplify things. Maybe just remove the whole commit (https://github.com/ZcashFoundation/zebra/pull/9214/commits/f01cd3eecf26138f277c6859ddabbf7bcbcce6c5) will be easier.

But if you check https://github.com/ZcashFoundation/zebra/issues/7893 requirements it says:

> Also, it shows current and recent outbound connections, but it should only show current connections.

arya2:
> > Also, it shows current and recent outbound connections, but it should only show current connections.

there should be a heartbeat task triggering an update to the `last_responded` times in the address book,

arya2:
> > Also, it shows current and recent outbound connections, but it should only show current connections.

I see. There are a couple options, we could:
- add the peer set to the rpc server and add a request for checking peer connections statuses, or
- update `recently_live_peers()` to check that the last connection state is responded and that the peer has responded in the last `HEARTBEAT_INTERVAL + 50ms`. (`recently_live_peers()` is only used in getpeerinfo/getinfo, we should update that instead of adding a new method)

The second option seems a bit easier, and good enough.

oxarbitrage:
> update recently_live_peers() to check that the last connection state is responded and that the peer has responded in the last HEARTBEAT_INTERVAL + 50ms. (recently_live_peers() is only used in getpeerinfo/getinfo, we should update that instead of adding a new method)

I agree that adding a new method is not really needed so that was removed as part of https://github.com/ZcashFoundation/zebra/pull/9214/commits/d90e639fd5a86b82feb8ef7021adeaa3733cc23f

While it is almost true that `recently_live_peers()` is only used in the getpeerinfo RPC, there is a marginal use of it in several metrics and tracing related code that i will prefer to leave as it is for now.

In regards to calculating if the peer is live by the use of `HEARTBEAT_INTERVAL`, it seems unnecessary as we can just use `has_connection_recently_responded()` which is what the now removed `currently_live_peers()` was doing. https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/meta_addr.rs#L499

So, what i did in https://github.com/ZcashFoundation/zebra/pull/9214/commits/d90e639fd5a86b82feb8ef7021adeaa3733cc23f for the `getpeerinfo` RPC call was to use the `recently_live_peers()` method as it was before and then filter the results by `has_connection_recently_responded()` which should give us peers that are more likely to be connected now. 

Let me know how do you see it. Thanks!

oxarbitrage:
@Mergifyio update

zebra-rpc/src/methods/get_block_template_rpcs.rs
```diff
@@ -1101,10 +1101,12 @@ where
 
     async fn get_peer_info(&self) -> Result<Vec<PeerInfo>> {
         let address_book = self.address_book.clone();
+
         Ok(address_book
             .recently_live_peers(chrono::Utc::now())
             .into_iter()
-            .map(PeerInfo::from)
+            .filter(|meta_addr| meta_addr.has_connection_recently_responded(chrono::Utc::now()))
```
arya2: Nitpick: `recently_live_peers()` shouldn't return any `MetaAddr` items that haven't recently responded, it's even more constrained, it also checks the last connection state.

See:
- https://github.com/ZcashFoundation/zebra/blob/track-inbound-connections-for-getpeerinfo/zebra-network/src/address_book.rs#L813-L821
- https://github.com/ZcashFoundation/zebra/blob/track-inbound-connections-for-getpeerinfo/zebra-network/src/meta_addr.rs#L550-L554

```suggestion
```

arya2 (dismissed):
This looks good, thank you!

I left one small suggestion for cleanup.

mpguerra (dismissed):

oxarbitrage:
> I left one small suggestion for cleanup.

The last suggestion break the tests. We need to either modify the test or revert the change.

zebra-rpc/src/methods/get_block_template_rpcs.rs
oxarbitrage: This is a correct suggestion but the mock used in tests was not doing the same causing a test to fail.
Fixed in https://github.com/ZcashFoundation/zebra/pull/9214/commits/835cca42052aa05cf11c0a5f7bd3ec01a44a803a to follow the same behaviour.

arya2 (approved):

Issue #7109, opened 2023-06-29
Store cumulative work in the finalized state for each height


teor2345:
## Motivation

We need to store block work in the finalized state database to:
1. efficiently calculate `getnetworksolps` & `getnetworkhashps` responses, and
2. show the full cumulative chain work in chain fork progress bars (and other logging and metrics)

Otherwise, we have to load:
1. block headers for the RPC height range to calculate solutions per second
2. block headers for the entire chain to calculate cumulative work

### Specifications

See the [upgrade design docs](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#design) for more details.

### Implementation Steps

 - [ ] update the [database format](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/state-db-upgrades.md#current) in the Zebra docs
 - [ ] increment the state minor version
 - [ ] write the new format in the block write task
 - [ ] update older formats in the format upgrade task
 - [ ] test that the new format works when creating a new state, and updating an older state
 - [ ] add `chain_work` field to `getblockchaininfo` RPC output

### Complex Code or Requirements

This will need an in-place format update. There's no need to wait for this update to complete, or do it in any particular order. The RPCs can fall back to loading block headers, and the chain forks can use the partial cumulative work.

If we store the cumulative work, we can calculate the block work for `height` using:
`cumulative_work(height) - cumulative_work(height - 1)` 

### Testing

Compare the work `bits` and RPC outputs with `zcashd`.

### Optional Cleanups

- [ ] Use the full work in `Chain`
- [ ] Replace `PartialCumulativeWork` with `Work` in `Chain`
- [ ] Delete the `PartialCumulativeWork` type

conradoplg:
zcashd seems to compute this in memory when starting up, I think we can do the same. Loading all block headers should be reasonably fast.  https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/main.cpp#L5564

mpguerra:
Let's only do this if someone requests the chainwork field and/or we need to improve the performance of calculating `getnetworksolps` and `getnetworkhashps`

PR #9215, opened 2025-02-07
feat(rpc): Add fields to `getblockchaininfo` RPC output

### LINKED ISSUE #8449: Update `getblockchaininfo` RPC method
## Motivation

## Specification

`zcashd`'s `getblockchaininfo` RPC method is documented here: https://zcash.github.io/rpc/getblockchaininfo.html 

Zebra already partially implements the `getblockchaininfo` RPC method, however, it is missing the following fields:

- "initial_block_download_complete": xx, (boolean) true if the initial download of the blockchain is complete
- "headers": xxxxxx,        (numeric) the current number of headers we have validated
- "difficulty": xxxxxx,     (numeric) the current difficulty
- "verificationprogress": xxxx, (numeric) estimate of verification progress [0..1]
- "chainwork": "xxxx"     (string) total amount of work in active chain, in hexadecimal
- "size_on_disk": xxxxxx,       (numeric) the estimated size of the block and undo files on disk
- "commitments": xxxxxx,    (numeric) the current number of note commitments in the commitment tree
- "chainSupply":           (object) information about the total supply
    - "monitored": xx,           (boolean) true if the total supply is being monitored
    - "chainValue": xxxxxx,      (numeric, optional) total chain supply after this block, in ZEC
    - "chainValueZat": xxxxxx,   (numeric, optional) total chain supply after this block, in zatoshis
- For the `valuePools` object:
    -  "monitored": xx,           (boolean) true if the pool is being monitored
- "softforks":             (array) status of softforks in progress
    - "id": "xxxx",        (string) name of softfork
    - "version": xx,         (numeric) block version
    - "enforce":            (object) progress toward enforcing the softfork rules for new-version blocks
        - "status": xx,       (boolean) true if threshold reached
        - "found": xx,        (numeric) number of blocks with the new version found
        - "required": xx,     (numeric) number of blocks required to trigger
        - "window": xx,       (numeric) maximum size of examined window of recent blocks
    - "reject": { ... }      (object) progress toward rejecting pre-softfork blocks (same fields as "enforce")

We need to:
1. Check which fields, if any, are needed by existing zcashd consumers.
2. Implement the required changes to support the relevant usecases.

The `commitments` field in zcashd is the number of Sprout commitments and may not be needed.

## Design

Most of these fields can be added by copying logic from elsewhere in `zebra-rpc` and/or without making changes outside `zebra-rpc`, with the exception of:
- `chainwork` which depends on adding finalized tip chain work to db (#7109),
- `size_on_disk` which depends on a new state request for reading the estimated db size, and
- `softforks` which depends on checking number of blocks with a certain version number past certain softfork heights (but is likely not needed). 

<details><summary>Outdated</summary>

## Related work

We might want to do #8939 at the same time
</details>

arya2:
## Motivation

This PR adds fields that are present on the zcashd `getblockchaininfo` RPC output which are likely required from Zebra's `getblockchaininfo` RPC too (except `chain_work` which should be added as part of https://github.com/ZcashFoundation/zebra/issues/7109).

Closes https://github.com/ZcashFoundation/zebra/issues/8449

## Solution

- Adds the missing fields to the response struct (except `softforks`)
- Adds a state request for querying the disk space usage
- Renames `ValuePoolBalance` to `Balance` and re-uses it for `chain_supply` by summing the amounts of all pools

Related:
- Runs `getblockchaininfo` state queries in parallel,
- Removes `BlockHeader` query, reads tip block time from latest chain tip channel instead, and
- Avoids returning an error when the state is empty (returns genesis/default values instead).

### Tests

- Updated snapshots
  - Ignores the `verificationprogress` field as it depends on the current time
- Updated existing prop tests
- Used shared logic covered by existing tests for the `difficulty` field

### Follow-up Work

- https://github.com/ZcashFoundation/zebra/issues/7109

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-rpc/src/methods/tests/snapshot.rs
+++ b/zebra-rpc/src/methods/tests/snapshot.rs
@@ -613,6 +613,12 @@ fn snapshot_rpc_getblockchaininfo(
                 // replace with:
                 "[Height]"
             }),
+            ".verificationprogress" => dynamic_redaction(|value, _path| {
+                // assert that the value looks like a valid verification progress here
+                assert!(value.as_f64().unwrap() < 1.0);
+                // replace with:
+                "[f64]"
+            }),
         })
     });
 }
--- a/zebra-rpc/src/methods/tests/snapshot.rs
+++ b/zebra-rpc/src/methods/tests/snapshot.rs
@@ -615,7 +615,7 @@ fn snapshot_rpc_getblockchaininfo(
             }),
             ".verificationprogress" => dynamic_redaction(|value, _path| {
                 // assert that the value looks like a valid verification progress here
-                assert!(value.as_f64().unwrap() < 1.0);
+                assert!(value.as_f64().unwrap() <= 1.0);
                 // replace with:
                 "[f64]"
             }),
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -570,6 +570,27 @@ impl DiskDb {
         );
     }
 
+    /// Returns the estimated total disk space usage of the database.
+    pub fn size(&self) -> u64 {
+        let db: &Arc<DB> = &self.db;
+        let db_options = DiskDb::options();
+        let mut total_size_on_disk = 0;
+        for cf_descriptor in DiskDb::construct_column_families(&db_options, db.path(), &[]).iter() {
+            let cf_name = &cf_descriptor.name();
+            let cf_handle = db
+                .cf_handle(cf_name)
+                .expect("Column family handle must exist");
+
+            total_size_on_disk += db
+                .property_int_value_cf(cf_handle, "rocksdb.total-sst-files-size")
+                .ok()
+                .flatten()
+                .unwrap_or(0);
+        }
+
+        total_size_on_disk
+    }
+
     /// When called with a secondary DB instance, tries to catch up with the primary DB instance
     pub fn try_catch_up_with_primary(&self) -> Result<(), rocksdb::Error> {
         self.db.try_catch_up_with_primary()
--- a/zebra-state/src/service/read/difficulty.rs
+++ b/zebra-state/src/service/read/difficulty.rs
@@ -82,6 +82,7 @@ pub fn get_block_template_chain_info(
 ///
 /// Returns the solution rate per second for the current best chain, or `None` if
 /// the `start_hash` and at least 1 block below it are not found in the chain.
+#[allow(unused)]
 pub fn solution_rate(
     non_finalized_state: &NonFinalizedState,
     db: &ZebraDb,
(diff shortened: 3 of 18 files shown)


zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -613,6 +613,12 @@ fn snapshot_rpc_getblockchaininfo(
                 // replace with:
                 "[Height]"
             }),
+            ".verificationprogress" => dynamic_redaction(|value, _path| {
+                // assert that the value looks like a valid verification progress here
+                assert!(value.as_f64().unwrap() < 1.0);
```
oxarbitrage:
```suggestion
                assert!(value.as_f64().unwrap() <= 1.0);
```

Q: If the chain is in sync, can this value be actually 1.0?

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -570,6 +570,27 @@ impl DiskDb {
...
+
+            total_size_on_disk += db
+                .property_int_value_cf(cf_handle, "rocksdb.total-sst-files-size")
+                .ok()
+                .flatten()
+                .unwrap_or(0);
+        }
+
+        total_size_on_disk
+    }
```
oxarbitrage, line 592: Q: Can we reuse this function [here](https://github.com/ZcashFoundation/zebra/blob/main/zebra-state/src/service/finalized_state/disk_db.rs#L518)?

zebra-state/src/service/read/difficulty.rs
```diff
@@ -82,6 +82,7 @@ pub fn get_block_template_chain_info(
 ///
 /// Returns the solution rate per second for the current best chain, or `None` if
 /// the `start_hash` and at least 1 block below it are not found in the chain.
+#[allow(unused)]
```
oxarbitrage, line 85:
```suggestion
```

I am not sure if this is needed for some build type, i commented out and built with `cargo build` and there were no complains.

oxarbitrage (dismissed):
This looks really good to me. We want to do some manual diff tests after https://github.com/ZcashFoundation/zebra/pull/9216 gets merged in before merging. I left some minor comments/questions but no blockers.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: They have less in common than it would seem, using `size()` in `print_db_metrics()` would save one line and I think the current code is simpler.

zebra-state/src/service/read/difficulty.rs
arya2: It only complains when building with `--no-default-features`.

zebra-rpc/src/methods/tests/snapshot.rs
arya2: Yep, applied as part of https://github.com/ZcashFoundation/zebra/pull/9215/commits/c62b61ef4e25ed4043db5a4ed82dc00963b0de74.

oxarbitrage (approved):

arya2:
@Mergifyio refresh

PR #9218, opened 2025-02-09
chore: fix errors in docs files 

strmfos:
This fields -> This field 
lexographic -> lexicographic
platforms are organized -> platforms is organized


mpguerra:
We're no longer accepting these sorts of PRs

PR #9242, opened 2025-02-12
change(rpc): Accept an unused second param in `sendrawtransaction` RPC

### LINKED ISSUE #8946: Update `sendrawtransaction` RPC method
## Motivation

We currently only support lightwalletd usage of this RPC method, we need to update it to ensure it meets mining usecases also.

## Spec

Docs: https://zcash.github.io/rpc/sendrawtransaction.html

Implement `allowhighfees` `boolean` parameter and set to `false` by default

arya2:
## Motivation

The `sendrawtransaction` RPC in zcashd accepts a second parameters, `allowhighfees`, which allows for adding transactions to the mempool that would otherwise be rejected for having an absurdly high fee.  Zebra's mempool never rejects transactions with excessive fees, so the only change needed to match zcashd here is to accept an unused second parameter in Zebra's `sendrawtransaction` method.

Closes #8946.

## Solution

Accepts an unused second param in `sendrawtransaction` RPC so that the method will ignore a second argument instead of returning a parser error.

### Tests

This works for the `submit_block` method.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -131,6 +131,7 @@ pub trait Rpc {
     /// # Parameters
     ///
     /// - `raw_transaction_hex`: (string, required, example="signedhex") The hex-encoded raw transaction bytes.
+    /// - `allow_high_fees`: (bool, optional) A legacy parameter accepted by zcashd but ignored by Zebra.
     ///
     /// # Notes
     ///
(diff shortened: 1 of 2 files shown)


oxarbitrage (approved):
Makes sense, thank you.

PR #9249, opened 2025-02-12
feat(rpc): add verbose support to getrawmempool

### LINKED ISSUE #3160: getrawmempool JSON-RPC method
## Motivation

`lightwalletd` uses the `getrawmempool` JSON-RPC method.

### Assumptions

`lightwalletd` doesn't need witnessed mempool transaction IDs for Orchard (`UnminedTxId`).
https://discord.com/channels/809218587167293450/809250822579028008/943326658042163270

### Required Fields

The method is documented here: https://zcash.github.io/rpc/getrawmempool.html

`lightwalletd` always uses the no-argument format: https://github.com/zcash/lightwalletd/blob/master/common/mempool.go#L95

And only these fields are used by `lightwalletd`: https://github.com/zcash/lightwalletd/blob/master/common/mempool.go#L106

Response:
```
[                     (json array of string)
  "transactionid"     (string) The transaction id - this is a TXID, not a WTXID
  ,...
]
```

### Tasks

- Implement the RPC method 
- Add RPC acceptance tests to CI
- Test that the RPC method works with `lightwalletd` 

## API Reference

We plan to use [jsonrpc_core](https://docs.rs/jsonrpc-core/18.0.0/jsonrpc_core/) with:
* [jsonrpc_http_server](https://github.com/paritytech/jsonrpc/tree/master/http)
* [jsonrpc-derive](https://github.com/paritytech/jsonrpc/tree/master/derive)

### Example Code

Here are examples of:
* [declaring RPC methods as Rust functions](https://github.com/paritytech/jsonrpc#basic-usage-with-derive)

### LINKED ISSUE #8447: Update `getrawmempool` RPC method
## Motivation

## Specification

`zcashd`'s `getrawmempool` RPC method is documented here: https://zcash.github.io/rpc/getrawmempool.html 

Zebra already partially implements the `getrawmempool` RPC method, however, it does not implement the optional `verbose` argument.

We need to:

1. Check whether any existing `zcashd` consumers use this parameter
2. If so, implement the required changes to support the relevant usecases.

conradoplg:
## Motivation

Add verbose=true support to getrawmempool

Closes #3160 

https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L409

## Solution

Add support. I had to include new fields to VerifiedUnminedTx to keep track of the time and height a tx was included in the mempool. It could be done more elegantly (e.g. with a new type, MempoolVerifiedUnminedTx?) to ensure these fields are always filled when a tx is added to the mempool, but I think this is sufficient for now.

### Tests

There is a simple test. Not sure how to make a better test. Ideally adding real txs (with dependencies) to the mempool, but that would require state. Suggestions are welcome.

I tested it manually in synced zebrad and zcashd nodes, they result the same values (though I couldn't test with transactions with dependencies between them)

### PR Author's Checklist

- [ ] The solution is tested.

### PR Reviewer's Checklist


zebra-rpc/src/methods.rs
```diff
@@ -1073,27 +1084,47 @@ where
...
-                    ))
-                });
-
-                let tx_ids: Vec<String> = transactions
-                    .iter()
-                    .map(|unmined_tx| unmined_tx.transaction.id.mined_id().encode_hex())
-                    .collect();
+                // verbose=true returns a map, where order does not matter,
+                // so we only need to sort when verbose=false.
+                if !verbose {
```
arya2: Nitpick/Optional: It would be nice to move this condition to a match guard. The `if verbose` below could be an `else` on this if statement, and the `else` block below could be moved up to this block here.

conradoplg: I'm not sold on the match guard, I think it would introduce an additional indentation level for little gain.

But I did follow your suggestions and moved things around to be clearer in 71c15241a

What do you think?

zebra-rpc/src/methods/types/get_raw_mempool.rs
```diff
@@ -0,0 +1,116 @@
+use std::collections::HashMap;
```
oxarbitrage, line 3:
```suggestion
//! Types used in `getrawmempool` RPC method.

use std::collections::HashMap;
```

https://github.com/ZcashFoundation/zebra/blob/2cf84e3add97594114486b88bba3396ee5952bdb/zebra-rpc/src/methods/types/get_blockchain_info.rs#L1

oxarbitrage (dismissed):
Thanks @conradoplg , i think this is good enough as is.

oxarbitrage (approved):

PR #9251, opened 2025-02-13
fix(tests): Check if a cached state is present in scanner tests

upbqdn:
## Motivation

The tests

- `scan_task_commands`, and
- `scan_start_where_left`

fail if the env var `ZEBRA_CACHE_DIR` is set but doesn't contain a valid state. The tests should not rely only on the var being set, but on the actual presence of the state.

## Solution

- Check if there's the state is present in `ZEBRA_CACHE_DIR`.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -161,7 +161,7 @@ run_tests() {
       --package zebra-state \
       -- --nocapture --include-ignored with_fake_activation_heights
 
-  elif [[ "${TEST_SCAN_TASK_COMMANDS}" -eq "1" ]]; then
+  elif [[ "${TEST_SCANNER}" -eq "1" ]]; then
     # Test the scanner.
     exec cargo test --locked --release --package zebra-scan \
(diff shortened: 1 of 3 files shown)


oxarbitrage (dismissed):
looks good, thank you.

docker/entrypoint.sh
```diff
@@ -161,7 +161,7 @@ run_tests() {
       --package zebra-state \
       -- --nocapture --include-ignored with_fake_activation_heights
 
-  elif [[ "${TEST_SCAN_TASK_COMMANDS}" -eq "1" ]]; then
+  elif [[ "${TEST_SCANNER}" -eq "1" ]]; then
```
upbqdn, line 164: We forgot to match the var name in CI: https://github.com/zcashfoundation/zebra/blob/57317ecb8e834ec32799ab41bb9cb5223a32015e/.github/workflows/sub-ci-integration-tests-gcp.yml#L500.

oxarbitrage (approved):

PR #9257, opened 2025-02-14
refactor(ci): do not run workflows tied to `ZcashFoundation` infra in forks

sellout:
## Motivation

Various workflows fail whenever I push to `main` on my fork. And one workflow is on a cron job which runs daily and emails me every day.

The failures are generally due to something like needing ZF credentials for some service, like DockerHub or GCP.

## Solution

This disables a bunch of CI jobs if they are run outside of the ZcashFoundation org.

There are also two workflows that simply had bugs in them. Those have also been fixed. Both bugs are caught by GitHub’s workflow file checks, so neither workflow gets run (i.e., they are not logic bugs in the steps that are performed, but references from one job to a job that doesn’t exist).

### Tests

Basically pushing these changes to `main` on my fork until I don’t get any failures. Some of the workflows don’t run right away, so it’s possible another failure may show up in a day.

### PR Author's Checklist

- [x] The solution is tested.

### PR Reviewer's Checklist

### DIFF
--- a/.github/workflows/cd-deploy-nodes-gcp.yml
+++ b/.github/workflows/cd-deploy-nodes-gcp.yml
@@ -227,12 +227,9 @@ jobs:
   # - on every push to the `main` branch
   # - on every release, when it's published
   # - on workflow_dispatch for manual deployments
-  
+
   # Determine which networks to deploy based on the trigger
-  
-  
-  
-  :
+  set-matrix:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


gustavovalverde (approved):
This is a great suggestion and fix. Thank you @sellout. I'll be adding some extra documentation based on this changes.

LGTM

mpguerra:
@mergify refresh

PR #9262, opened 2025-02-17
cleanup(logger): Replace last log mutex with watch channel

### LINKED ISSUE #8448: Update `getinfo` RPC method
## Motivation

## Specification

`zcashd`'s `getinfo` RPC method is documented here: https://zcash.github.io/rpc/getinfo.html 

Zebra already partially implements the `getinfo` RPC method, however, it is missing the following fields:

- `version`: xxxxx,           (numeric) the server version
- `protocolversion`: xxxxx,   (numeric) the protocol version
- ~~`walletversion`: xxxxx,     (numeric, optional) the wallet version, if wallet functionality is enabled~~
- ~~`balance`: xxxxxxx,         (numeric, optional) the total Zcash balance of the wallet, if wallet functionality is enabled~~
- `blocks`: xxxxxx,           (numeric) the current number of blocks processed in the server
- ~~`timeoffset`: xxxxx,        (numeric) the time offset (deprecated; always 0)~~
- `connections`: xxxxx,       (numeric) the number of connections
- `proxy`: "host:port",     (string, optional) the proxy used by the server
- `difficulty`: xxxxxx,       (numeric) the current difficulty
- `testnet`: true|false,      (boolean) if the server is using testnet or not
- ~~`keypoololdest`: xxxxxx,    (numeric, optional) the timestamp (seconds since GMT epoch) of the oldest pre-generated key in the key pool, if wallet functionality is enabled~~
- ~~`keypoolsize`: xxxx,        (numeric, optional) how many new keys are pre-generated~~
- ~~`unlocked_until`: ttt,      (numeric, optional) the timestamp in seconds since epoch (midnight Jan 1 1970 GMT) that the wallet is unlocked for transfers, or 0 if the wallet is locked, if wallet functionality is available and the wallet is encrypted~~
- `paytxfee`: x.xxxx,         (numeric) the transaction fee rate set in ZEC per 1000 bytes
- `relayfee`: x.xxxx,         (numeric) minimum relay fee rate for transactions in ZEC per 1000 bytes
- `errors`: "..."           (string) message describing the latest or highest-priority error
- `errorstimestamp`: "..."  (string) timestamp associated with the latest or highest-priority error

We need to:
1. Check which fields, if any, are needed by existing zcashd consumers.
2. Implement the required changes to support the relevant usecases.

arya2:
This is a cleanup PR for replacing a `Mutex` sharing the last warning or error log message with a watch channel.

There should be no concurrency issues with the `Mutex` as it's currently used, so this is only cleanup to enable extending its use without any concurrency considerations.

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -18,7 +18,10 @@ use indexmap::IndexMap;
 use jsonrpsee::core::{async_trait, RpcResult as Result};
 use jsonrpsee_proc_macros::rpc;
 use jsonrpsee_types::{ErrorCode, ErrorObject};
-use tokio::{sync::broadcast, task::JoinHandle};
+use tokio::{
+    sync::{broadcast, watch},
+    task::JoinHandle,
+};
 use tower::{Service, ServiceExt};
 use tracing::Instrument;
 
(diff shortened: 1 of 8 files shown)


oxarbitrage (approved):
Yea, this is a lot better. I need to remember to use a channel instead of a mutex in the future. Thanks for the cleanup.

PR #9276, opened 2025-02-19
fix(ci): correct typo in GitHub workflow

sellout:
Introduced by #9257, but overlooked because the workflow runs asynchronously. (Is there a way to get GitHub to validate workflow files in a PR check, rather than only when the workflow runs?)

### DIFF
--- a/.github/workflows/chore-delete-gcp-resources.yml
+++ b/.github/workflows/chore-delete-gcp-resources.yml
@@ -106,7 +106,7 @@ jobs:
   # The same artifacts are used for both mainnet and testnet.
   clean-registries:
     name: Delete unused artifacts in registry
-    if: github.repository_owner == 'ZcashFoundation''
+    if: github.repository_owner == 'ZcashFoundation'
     runs-on: ubuntu-latest
     permissions:
       contents: 'read'
--- a/.github/workflows/ci-build-crates.yml
+++ b/.github/workflows/ci-build-crates.yml
@@ -170,7 +170,11 @@ jobs:
     needs: [ matrix, build ]
     # Only open tickets for failed or cancelled jobs that are not coming from PRs.
     # (PR statuses are already reported in the PR jobs list, and checked by GitHub's Merge Queue.)
-    if: (failure() && github.event.pull_request == null) || (cancelled() && github.event.pull_request == null)
+    #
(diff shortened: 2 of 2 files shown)


sellout:
I’m also still having other workflows fail on my fork, like
1. https://github.com/sellout/zebra/actions/runs/13321555857 and
2. https://github.com/sellout/zebra/actions/runs/13321555854

Both complain “This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.” and have runners listed as “ubuntu-latest-xl” and “ubuntu-latest-m”, which I can’t find any info on. I’m guessing they’re either runners only available to paid accounts at a certain level, or maybe self-hosted or something?

sellout:
This is in draft as I wait for async jobs to run on my fork.

gustavovalverde:
> (Is there a way to get GitHub to validate workflow files in a PR check, rather than only when the workflow runs?)

We have an action linter, it's odd it didn't catch the extra `'` (I recently merged a PR with that fix)

> Both complain “This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.” and have runners listed as “ubuntu-latest-xl” and “ubuntu-latest-m”, which I can’t find any info on. I’m guessing they’re either runners only available to paid accounts at a certain level, or maybe self-hosted or something?

Oh, those are [GitHub Larger Runners ](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners)(you can name them as you wish). So this will be an interesting thing to find a workaround. I'll have a look after the DevOps community call

sellout:
> Oh, those are [GitHub Larger Runners ](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners)(you can name them as you wish). So this will be an interesting thing to find a workaround. I'll have a look after the DevOps community call

I think they could be skipped the same way (checking the repo owner `== 'ZcashFoundation'`), but if there’s a way to preserve them on forks, even better. Checking the repo owner would still allow them to run on PRs againt this repo.

gustavovalverde:
Let's see how it goes with this PR instead https://github.com/ZcashFoundation/zebra/pull/9362 if the first approach doesn't work I'll try to add an extra job to validate the owner and change the runner dinamically

sellout:
Yeah, I like #9362. Thanks!

PR #9280, opened 2025-02-20
change(ci): Remove check for custom configs from GCP deployments

upbqdn:
## Motivation

The `test-zebra-conf-path` check in `cd-deploy-nodes-gcp.yml` tries to configure Zebra running in a Docker container with a custom config file. However, it always fails because the container doesn't contain the file as it is built for the `runtime` target and contains only files needed in production. This test used to falsely pass because the entrypoint script used to create a new config file on the fly if the specified one was missing. We have the same check for a Docker container built for the `tests` target in `sub-ci-unit-tests-docker.yml`. This check works as intended because those containers contain the custom config file.

## Solution

- Remove the `test-zebra-conf-path` from `cd-deploy-noes-gcp.yml`.

An alternative solution would be to move the custom config file to the container, but that would require a lot more work, and I thought it wasn't worth it since we test the same functionality elsewhere.

### Tests

N/A

### PR Author's Checklist

- [ ] The solution is tested.

### DIFF
--- a/.github/workflows/cd-deploy-nodes-gcp.patch-external.yml
+++ b/.github/workflows/cd-deploy-nodes-gcp.patch-external.yml
@@ -45,11 +45,3 @@ jobs:
     if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork }}
     steps:
       - run: 'echo "Skipping job on fork"'
-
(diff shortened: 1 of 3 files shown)


gustavovalverde (approved):
LGTM

PR #9292, opened 2025-02-22
feat(rpc): Allow negative heights in rpc methods

### LINKED ISSUE #9025: rpc: support negative heights
### Motivation

In a bunch of RPC calls like getblock, zcashd supports negative heights (where -1 is the last known valid block) but Zebra doesn't. It is a useful feature in particular for getting the latest block and seems easy to support.

Look which RPC calls support it in zcashd and add support to Zebra

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

oxarbitrage:
## Motivation

We want to support negative heights in RPC methods when `HashOrHeight` is supported.

Close https://github.com/ZcashFoundation/zebra/issues/9025

## Solution

Added a `NegativeHeight` variant to `HashOrHeight` enum.

### Tests

Added checks in the `getblock_rpc` test for the `getblock` call. Other calls that use `HashOrHeight` are supported but not tested. Can add more tests if anyone think is needed.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -43,7 +43,8 @@ use zebra_consensus::ParameterCheckpoint;
 use zebra_network::address_book_peers::AddressBookPeers;
 use zebra_node_services::mempool;
 use zebra_state::{
-    HashOrHeight, OutputIndex, OutputLocation, ReadRequest, ReadResponse, TransactionLocation,
+    HashHeightOrNegativeHeight, HashOrHeight, OutputIndex, OutputLocation, ReadRequest,
+    ReadResponse, TransactionLocation,
 };
 
 use crate::{
@@ -839,11 +840,24 @@ where
             None
         };
 
-        let hash_or_height: HashOrHeight = hash_or_height
-            .parse()
+        let hash_height_or_negative_height = hash_or_height
+            .parse::<HashHeightOrNegativeHeight>()
             // Reference for the legacy error code:
             // <https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L629>
             .map_error(server::error::LegacyCode::InvalidParameter)?;
+
+        let best_tip_height = hash_height_or_negative_height
+            .is_negative_height()
+            .then(|| {
+                self.latest_chain_tip
+                    .best_tip_height()
+                    .ok_or_misc_error("No blocks in state")
+            })
+            .transpose()?;
+
+        let hash_or_height = hash_height_or_negative_height
+            .into_hash_or_height(best_tip_height)
+            .ok_or_misc_error("Block height out of range")?;
 
         if verbosity == 0 {
             let request = zebra_state::ReadRequest::Block(hash_or_height);
--- a/zebra-rpc/src/methods/tests/vectors.rs
+++ b/zebra-rpc/src/methods/tests/vectors.rs
@@ -196,6 +196,21 @@ async fn rpc_getblock() {
         assert_eq!(get_block, expected_result);
     }
 
+    // Test negative heights: -1 should return block 10, -2 block 9, etc.
+    for neg_height in (-10..=-1).rev() {
+        // Convert negative height to corresponding index
+        let index = (neg_height + (blocks.len() as i32)) as usize;
+
+        let expected_result = GetBlock::Raw(blocks[index].clone().into());
+
+        let get_block = rpc
+            .get_block(neg_height.to_string(), Some(0u8))
+            .await
+            .expect("We should have a GetBlock struct");
+
+        assert_eq!(get_block, expected_result);
+    }
+
     // Create empty note commitment tree information.
     let sapling = SaplingTrees { size: 0 };
     let orchard = OrchardTrees { size: 0 };
--- a/zebra-state/src/lib.rs
+++ b/zebra-state/src/lib.rs
@@ -42,7 +42,8 @@ pub use error::{
     ValidateContextError,
 };
 pub use request::{
-    CheckpointVerifiedBlock, HashOrHeight, ReadRequest, Request, SemanticallyVerifiedBlock,
+    CheckpointVerifiedBlock, HashHeightOrNegativeHeight, HashOrHeight, ReadRequest, Request,
+    SemanticallyVerifiedBlock,
 };
 
 #[cfg(feature = "indexer")]
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -84,6 +84,8 @@ pub enum HashOrHeight {
     Hash(block::Hash),
     /// A block identified by height.
     Height(block::Height),
+    /// A block identified by a negative height.
+    NegativeHeight(i64),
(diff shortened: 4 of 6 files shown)


upbqdn (commented):
I think it would be better to implement the negative heights as a standalone type instead of extending `HashOrHeight`.

The first reason is that including negative heights into `HashOrHeight` breaks the semantics of the type's API. For example, the functions `height_or_else` and `hash_or_else` don't consider negative heights. Instead, we need a separate function for conversion. Also, the conversion becomes unwieldy for the caller.

The second reason is that negative heights depend on the current height of the tip. If the tip height increments, so does the height of the block the negative height points to. The variants in `HashOrHeight` don't have this property, so we can keep them around for as long as we need. On the other hand, we should convert negative heights to `Height` right upon the RPC call.

oxarbitrage:
> I think it would be better to implement the negative heights as a standalone type instead of extending `HashOrHeight`.
> 
> The first reason is that including negative heights into `HashOrHeight` breaks the semantics of the type's API. For example, the functions `height_or_else` and `hash_or_else` don't consider negative heights. Instead, we need a separate function for conversion. Also, the conversion becomes unwieldy for the caller.
> 
> The second reason is that negative heights depend on the current height of the tip. If the tip height increments, so does the height of the block the negative height points to. The variants in `HashOrHeight` don't have this property, so we can keep them around for as long as we need. On the other hand, we should convert negative heights to `Height` right upon the RPC call.

I think that makes sense, i don't see it as that critical but i accept it. I actually extended `HashOrHeight` as it was suggested [here](https://github.com/ZcashFoundation/zebra/issues/9025#issuecomment-2654392262).

Anyway, ill change that up and ping you when it is done.

arya2:
> I actually extended HashOrHeight as it was suggested https://github.com/ZcashFoundation/zebra/issues/9025#issuecomment-2654392262.

I think converting negative heights in zebra-state could help mitigate potential concurrency considerations and would be easier to re-use from other methods (e.g. `getaddresstxids`). Adding a `Depth` variant to `HashOrHeight` seems convenient to me, but I don't have a strong opinion on how the negative height is passed to the state service, only that it'd be a good idea to pass it to the state svc somehow.

oxarbitrage:
> > I actually extended HashOrHeight as it was suggested [#9025 (comment)](https://github.com/ZcashFoundation/zebra/issues/9025#issuecomment-2654392262).
> 
> I think converting negative heights in zebra-state could help mitigate potential concurrency considerations and would be easier to re-use from other methods (e.g. `getaddresstxids`). Adding a `Depth` variant to `HashOrHeight` seems convenient to me, but I don't have a strong opinion on how the negative height is passed to the state service, only that it'd be a good idea to pass it to the state svc somehow.

To me is convenient as well as it will automatically works for all RPC methods that use it without any further modifications ...

A new type will mean more code for little gain as the `HashOrHeight` main purpose was always to support a string in the RPC method arguments that can be converted to a hash or a height. Adding a new type to support negative height instead of using the same we already have, basically to maintain semantic will make the original `HashOrHeight` pretty much obsolete.

With that said, @upbqdn is the reviewer here and his view in this subject seems strong enough for me to make the change even if i am not convinced that the suggestion is actually better.

arya2:
> If the tip height increments, so does the height of the block the negative height points to

I think this is a good point, I didn't like that either and was imagining the state service would transform the `HashOrHeight::Depth` variant to a hash or height as the first step in handling those requests so the other functions work as usual, but if we're doing that, a new enum around `HashOrHeight` or that converts to `HashOrHeight` would be cleaner: 
```rust
enum HashHeightOrDepth {
    HashOrHeight(HashOrHeight),
    Depth(HeightDiff),
}

// or

enum HashHeightOrDepth {
    Hash(Hash),
    Height(Height),
    Depth(HeightDiff),
}

impl HashHeightOrDepth {
    fn into_hash_or_height(self, tip_height: Height) -> HashOrHeight { }
}
```

Related: There's a similar problem with the transparent addresses' `network_kind` field where it's typed as an enum of three variants but it's expected to never be set as one of them (RegtestKind). We should eventually clean that up too.

upbqdn:
Regarding the first point, the PR currently changes the API from https://github.com/zcashfoundation/zebra/blob/79e18e045ce8ca11c4078495e2425cc204cf03c2/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142 to https://github.com/ZcashFoundation/zebra/blob/negative-heights/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142-L147, which essentially means the caller needs to manually parse `hash_or_height` each time we want to use it. We currently call `hash_or_else` and `height_or_else` about 17 times in Zebra, so I thought it would be actually less work to have a different solution.

arya2:
> Regarding the first point, the PR currently changes the API from [this](https://github.com/zcashfoundation/zebra/blob/79e18e045ce8ca11c4078495e2425cc204cf03c2/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142) to [this](https://github.com/ZcashFoundation/zebra/blob/negative-heights/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142-L147), which essentially means the caller needs to manually parse `hash_or_height` each time we want to use it. We currently call `hash_or_else` and `height_or_else` about 17 times in Zebra, so I thought it would be actually less work to have a different solution.

Yeah, I was thinking we could expect it not to be a `Depth` in the rest of the API, but if we're doing that it's better to be explicit and do a type conversion instead.

It's definitely more work to add a variant and handle it wherever `HashOrHeight` is used, and it would be a lot of redundant code.

oxarbitrage:
> Regarding the first point, the PR currently changes the API from https://github.com/zcashfoundation/zebra/blob/79e18e045ce8ca11c4078495e2425cc204cf03c2/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142 to https://github.com/ZcashFoundation/zebra/blob/negative-heights/zebra-state/src/service/finalized_state/zebra_db/block.rs#L142-L147, which essentially means the caller needs to manually parse `hash_or_height` each time we want to use it. We currently call `hash_or_else` and `height_or_else` about 17 times in Zebra, so I thought it would be actually less work to have a different solution.

Yea, you are right there, i am working in the change.

I should have written tests for the rest of the RPC methods, and this issue would have surfaced. Apologies.

upbqdn:
It'd be great if we could at least avoid the manual parsing.

upbqdn:
> It'd be great if we could at least avoid the manual parsing.

Sorry, GitHub didn't show me the two messages above this one.

upbqdn (commented):
How about something like this? 
https://github.com/ZcashFoundation/zebra/pull/9316
That solution doesn't add new types and converts negative heights on the spot. I'm happy to finish the PR if we find that solution suitable.

Side note: I don't think the `HashOrHeight` type is useful. It makes the API unnecessarily complex and doesn't consider negative heights. It would be easier to handle plain heights and convert hashes to heights on the spot, just as #9316 does with negative heights, IMO.

oxarbitrage:
Ok ... please do that. Closing in favour of https://github.com/ZcashFoundation/zebra/pull/9316

PR #9299, opened 2025-02-25
allow deserialization of Balance with skipped ID field

AloeareV:
The Balance type will skip serializing the `id` field when the string is empty, but currently fails to deserialize a Balance with no `id` field.


arya2 (approved):

mpguerra:
@mergify refresh

PR #9308, opened 2025-02-28
change(consensus): Updates transaction verifier to use `orchard::bundle::BatchValidator`

### LINKED ISSUE #4672: Add support for Orchard proof batch verification

arya2:
## Motivation

We want Zebra to validate orchard proofs in batches as it does for other pools.

This PR also makes some progress towards deduplicating logic in zebra-chain that's available in the `orchard` crate.

Closes #4672

## Solution

- Adds an `ItemSize` trait in `tower-batch-control` defining an `item_size()` method,
- Replaces batch proof verifier Item type aliases with newtypes and implements `ItemSize` for them where all of them return 1 (no change in behaviour) except for the halo2 verifier which returns the number of actions in the request,
- Increments `pending_items` in the batch control worker by `item_size()` instead of 1,
- Updates `HALO2_MAX_BATCH_SIZE` from 2 to 64 (the default value for other verifiers)

### Tests

This PR still needs tests. It also removes some tests that are no longer relevant.

### Follow-up Work

Use `shardtree` to append to note commitment trees in parallel

### PR Checklist

- [ ] The solution is tested.


zebra-consensus/src/primitives/halo2.rs
```diff
@@ -78,93 +58,102 @@ lazy_static::lazy_static! {
...
+        let anchor = orchard::tree::Anchor::from_bytes(shielded_data.shared_anchor.into()).unwrap();
+        let authorization = orchard::bundle::Authorized::from_parts(
+            orchard::Proof::new(shielded_data.proof.0.clone()),
+            <[u8; 64]>::from(shielded_data.binding_sig).into(),
+        );
 
-        Ok(())
-    }
-}
+        let bundle = orchard::bundle::Bundle::from_parts(
```
arya2: This PR should use the orchard bundle created when preparing to compute transaction sighashes instead of converting orchard data to the `orchard` crate `Bundle` type twice.

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9308/commits/40e54be3323d4f16f56a8c1ffecf4bfae3fe4c2e.

Cargo.toml
```diff
@@ -92,6 +92,7 @@ mset = "0.1"
 nix = "0.29"
 num-integer = "0.1.46"
 once_cell = "1.21"
+nonempty = "0.11.0"
```
arya2:
```suggestion
```

Cargo.toml
```diff
@@ -153,6 +154,7 @@ x25519-dalek = "2.0.1"
 zcash_note_encryption = "0.4.1"
 zcash_script = "0.3.2"
 
+
```
arya2:
```suggestion
```

zebra-consensus/src/transaction/tests.rs
arya2, line 1: No change in this file except adding timeouts to the verifier requests (part of debugging the issue around not setting batch timers which was fixed in https://github.com/ZcashFoundation/zebra/pull/9308/commits/1173a335e3e481f6f607ee5389a8dc7bb9bedd62).

zebra-consensus/src/primitives.rs
```diff
@@ -33,12 +33,9 @@ pub async fn spawn_fifo_and_convert<
 }
 
 /// Fires off a task into the Rayon threadpool and awaits the result through a oneshot channel.
-pub async fn spawn_fifo<
-    E: 'static + std::error::Error + Sync + Send,
-    F: 'static + FnOnce() -> Result<(), E> + Send,
->(
+pub async fn spawn_fifo<T: 'static + Send, F: 'static + FnOnce() -> T + Send>(
```
arya2, line 36: The return type of the closure passed in from the halo2 verifier is now a bool so this needed to be more flexible.

zebra-consensus/src/primitives/halo2/tests.rs
arya2, line 1: It would be nice to replace these tests, but it seems okay to get rid of them for now since we're using the verifier from the `orchard` crate.

They don't work anymore since `BatchValidator.queue()` now needs the transaction's shielded sighash.

oxarbitrage (approved):
looks good as far as i can see. nice work.

PR #9318, opened 2025-03-10
fix(ci): Add Rust environment variables to Docker test workflows

gustavovalverde:
## Motivation

It's hard to debug tests if we don't have access to the full backtrace and other options are expected to be set by our GitHub variables, but in the end not all of those are being fully used at runtime.

## Solution

Enhance test workflows by adding Rust-specific environment variables:
- Include `RUST_LOG` for logging configuration
- Add `RUST_BACKTRACE` and `RUST_LIB_BACKTRACE` for improved error tracing
- Include `COLORBT_SHOW_HIDDEN` for detailed backtraces
- Add `CARGO_INCREMENTAL` for build performance optimization

This also fixes user creation issues in the Dockerfile:
- Move WORKDIR after user creation to prevent home directory ownership issues
- Properly set environment variables for UID, GID, HOME, and USER in each stage
- Reorganize Dockerfile to ensure home directory is created after user setup
- Fix interactive prompts during adduser by ensuring proper directory ownership

### Tests

- Existing non-failing tests should run as expected

### Follow-up Work

- Track the Test All issue

### PR Checklist

- [X] The solution is tested.

### DIFF
--- a/.github/workflows/sub-ci-unit-tests-docker.yml
+++ b/.github/workflows/sub-ci-unit-tests-docker.yml
@@ -53,10 +53,15 @@ jobs:
       - name: Run all tests
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
-          docker run -t \
+          docker run -tty \
           -e RUN_ALL_TESTS=1 \
           -e FEATURES="journald prometheus filter-reload" \
           -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e RUST_LOG=${{ env.RUST_LOG }} \
+          -e RUST_BACKTRACE=${{ env.RUST_BACKTRACE }} \
+          -e RUST_LIB_BACKTRACE=${{ env.RUST_LIB_BACKTRACE }} \
+          -e COLORBT_SHOW_HIDDEN=${{ env.COLORBT_SHOW_HIDDEN }} \
+          -e CARGO_INCREMENTAL=${{ env.CARGO_INCREMENTAL }} \
           ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
 
   # Run state tests with fake activation heights.
@@ -85,9 +90,14 @@ jobs:
           NETWORK: ${{ inputs.network || vars.ZCASH_NETWORK }}
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
-          docker run -t \
+          docker run -tty \
           -e TEST_FAKE_ACTIVATION_HEIGHTS=1 \
           -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e RUST_LOG=${{ env.RUST_LOG }} \
+          -e RUST_BACKTRACE=${{ env.RUST_BACKTRACE }} \
+          -e RUST_LIB_BACKTRACE=${{ env.RUST_LIB_BACKTRACE }} \
+          -e COLORBT_SHOW_HIDDEN=${{ env.COLORBT_SHOW_HIDDEN }} \
+          -e CARGO_INCREMENTAL=${{ env.CARGO_INCREMENTAL }} \
           ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
 
   # Test that Zebra syncs and checkpoints a few thousand blocks from an empty state.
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -81,15 +76,22 @@ ENV ZEBRA_SKIP_IPV6_TESTS=${ZEBRA_SKIP_IPV6_TESTS:-1}
 # This environment setup is almost identical to the `runtime` target so that the
 # `tests` target differs minimally. In fact, a subset of this setup is used for
 # the `runtime` target.
-
 ARG UID
+ENV UID=${UID}
 ARG GID
+ENV GID=${GID}
```
upbqdn, line 82: Why do we need these?

upbqdn, line 82: Why do we need these?

docker/Dockerfile
```diff
@@ -213,10 +209,17 @@ ARG UID
...
 ENV GID=${GID}
+ARG HOME
+ENV HOME=${HOME}
+ARG USER
+ENV USER=${USER}
 
 RUN addgroup --gid ${GID} ${USER} && \
     adduser --gid ${GID} --uid ${UID} --home ${HOME} ${USER}
 
+# System users have no home dirs, but we set one for users' convenience.
```
upbqdn:
```suggestion
```

docker/Dockerfile
```diff
@@ -190,13 +192,7 @@ ARG FEATURES
 ENV FEATURES=${FEATURES}
 
 # Create a non-privileged system user for running `zebrad`.
```
upbqdn:
```suggestion
# Create a non-privileged user for running `zebrad`.
```

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -53,10 +53,15 @@ jobs:
...
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
-          docker run -t \
+          docker run --tty \
           -e RUN_ALL_TESTS=1 \
           -e FEATURES="journald prometheus filter-reload" \
           -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e RUST_LOG=${{ env.RUST_LOG }} \
+          -e RUST_BACKTRACE=${{ env.RUST_BACKTRACE }} \
+          -e RUST_LIB_BACKTRACE=${{ env.RUST_LIB_BACKTRACE }} \
+          -e COLORBT_SHOW_HIDDEN=${{ env.COLORBT_SHOW_HIDDEN }} \
```
upbqdn, line 63: This variable should configure [`color-backtrace`](https://github.com/athre0z/color-backtrace), but I didn't find any reference to it in the codebase, so it looks like we're not using it anymore. It might be worth removing the variable altogether.

upbqdn (approved):

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn, line 63: This variable should configure [`color-backtrace`](https://github.com/athre0z/color-backtrace), but I didn't find any reference to it in the codebase, so it looks like we're not using it anymore. It might be worth removing the variable altogether.

upbqdn (approved):

upbqdn:
> Fix interactive prompts during adduser by ensuring proper directory ownership

Is this in the code?

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn, line 63: Are these actually set anywhere?

upbqdn (dismissed):

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: Here 
image

docker/Dockerfile
gustavovalverde: Just for convenience, in case the user builds with custom UID/GID and then needs those variables available in the container.

gustavovalverde:
> > Fix interactive prompts during adduser by ensuring proper directory ownership
> 
> Is this in the code?

@upbqdn this is moving the WORKDIR after user creation. Previously the WORKDIR directive was creating the home, and then this home was not owned by the Zebra user.

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: I'll do that cleanup in another PR

upbqdn:
We create the dir [here](https://github.com/zcashfoundation/zebra/blob/de7e5b547f0d5e0b16e86ceb2a5a8ad5a231b21b/docker/Dockerfile#L218) and recursively set ownership [here](https://github.com/zcashfoundation/zebra/blob/de7e5b547f0d5e0b16e86ceb2a5a8ad5a231b21b/docker/Dockerfile#L231). I think we can remove the `WORKDIR` instruction.

upbqdn:
The ownership of the home dir and its contents is correct:

```bash
zebra@89f5be7f19f2:~$ pwd
/home/zebra
zebra@89f5be7f19f2:~$ ls -lah
total 0
drwxr-xr-x 1 zebra zebra 26 Feb 28 16:55 .
drwxr-xr-x 1 root  root  10 Feb 28 15:22 ..
drwxr-xr-x 1 zebra zebra 10 Feb 28 16:55 .cache
drwxr-xr-x 1 zebra zebra 22 Mar 10 21:07 .config
```

.github/workflows/sub-ci-unit-tests-docker.yml
upbqdn: I'm now thinking we could move these to the yml files so we could set them individually for each workflow or job. For example, setting `RUST_LOG` to warn would significantly reduce the massive logs we have for all workflows. And if there was something we needed to debug, and the logs would be insufficient, we could increase the log level for that individual failing job. Scrolling the logs takes quite a lot of time currently.

upbqdn (approved):

PR #9323, opened 2025-03-12
fix(ci): Better permission and cache dirs handling in Docker

upbqdn:
## Motivation

Some integration tests in [this job](https://github.com/zcashfoundation/zebra/tree/49741e8b476aff1d1dd2f6a2f57ec9d24404d345/.github/workflows/sub-ci-unit-tests-docker.yml) fail for all PRs because they expect a cached state even though the job doesn't provide any. The tests get triggered because the `ZEBRA_CACHE_DIR` env var is implicitly set in the Dockerfile [here](https://github.com/zcashfoundation/zebra/blob/49741e8b476aff1d1dd2f6a2f57ec9d24404d345/docker/Dockerfile#L230). I set the variable there in PR #8923 because

- I wanted to explicitly run `chown` on it to handle permissions [here](https://github.com/zcashfoundation/zebra/blob/49741e8b476aff1d1dd2f6a2f57ec9d24404d345/docker/Dockerfile#L231), and 
- so that it's clear how the Dockerfile and entrypoint handle the cache dir before running Zebra. 

However, some integration tests check if the var is set, and if it is not, they return early with an `Ok` even though they don't run. Since PR #8923 sets the var for all tests, those that expect the cached state to be present in the dir pointed to by the var will fail. This means that PR #8923 shouldn't have merged since it introduced this bug. I didn't really investigate why the PR merged. And even though this is a rather simple bug, it was hard to figure out what was wrong by looking at the logs, showing how hard it is to maintain the CI. I solved this by not setting the var in the Dockerfile, so those tests that expect the cached state return an `Ok` in the `Test all` job. Instead, I re-introduced the var, together with the `LWD_CACHE_DIR`, in all integration tests that are supposed to run with the cached state. (PR #9328 provides a cleaner solution). This brought up a bug that was much trickier to diagnose. I don't want to write the whole story, but the short version is that we weren't setting ownership permissions properly anywhere for the path pointed to by `LWD_CACHE_DIR`, so when lwd was executed from Zebra under a non-root user, it failed. That's why I used the recursive solution that Gustavo pointed to here: https://github.com/ZcashFoundation/zebra/pull/9323#discussion_r1995374146.

There was also a bug in the path for the custom config file that I'm not sure why didn't show up earlier. 

The refactoring of the way we use gosu, ARG, and ENV instructions, is related to the description above.

### Key changes
- Remove the predefined `LWD_CACHE_DIR ` and `ZEBRA_CACHE_DIR ` variables from the Dockerfile
  - Explicitly set these variables in CI only for the tests that need it
- Fix permissions handling in the Dockerfile.

### Other changes
- Refactor the handling of ARG and ENV instructions in the Dockerfile.
  - Options like `RUST_BACKTRACE `, `RUST_LIB_BACKTRACE `, `RUST_LOG `,  and `COLORBT_SHOW_HIDDEN` were added as build time arguments, but these are used by Rust at runtime
- Remove some redundancies from the entrypoint and Dockerfile.
- Instead of using `eval` for running cargo, use `exec` to replace the _current process_ and avoid creating _child processes_ (that causes issues with Docker reading the SIGTERM)

### PR Checklist

- [x] The solution is tested.


upbqdn:
This is still only a draft.

upbqdn:
This is ready.

docker/entrypoint.sh
```diff
@@ -3,36 +3,15 @@
...
 # The main script logic is at the bottom.
-#
-# ## Notes
-#
-# - `$ZEBRA_CONF_PATH` must point to a Zebra conf file.
 
 set -eo pipefail
 
-# Exit early if `ZEBRA_CONF_PATH` does not point to a file.
-if [[ ! -f "${ZEBRA_CONF_PATH}" ]]; then
```
gustavovalverde, line 14: Why is this being removed?

docker/entrypoint.sh
```diff
@@ -103,26 +74,6 @@ prepare_conf_file() {
   echo "${ZEBRA_CONF_PATH}"
 }
 
-# Checks if a directory contains subdirectories
-#
-# Exits with 0 if it does, and 1 otherwise.
-check_directory_files() {
```
gustavovalverde, line 109: I know this is dead code, but did we remove this functionality considering the application will handle this scenario and exit the test early?

docker/entrypoint.sh
```diff
@@ -169,145 +120,167 @@ run_tests() {
...
+  echo "INFO: Using the following Zebra config:"
+  cat "${ZEBRA_CONF_PATH}"
+
+  mkdir -p "${HOME}"
+  chown -R "${UID}":"${GID}" "${HOME}"
+  chown -R "${UID}":"${GID}" "${ZEBRA_CONF_PATH}"
+  addgroup --quiet --gid "${GID}" "${USER}"
+  adduser --quiet --disabled-password --gecos "" \
+    --gid "${GID}" --uid "${UID}" "${USER}"
+  cd "${HOME}"
```
gustavovalverde: This is very common and expected to be in the Dockerfile and not in the entrypoint. If I deploy Zebra and override the `ENTRYPOINT` or `CMD` (as this is very common), then the user won't get created, the HOME directory won't exist and the application might not start as expected.

gustavovalverde (changes requested):
Aside from my review comments, it also seems a lot of comments explaining some rationale is being deleted, and we don't have other supporting documentation asid

docker/entrypoint.sh
upbqdn: Yup, I removed it to keep the entrypoint short. I kept it around in case of any issues, but there haven't been any, so I thought it was redundant at this point.

docker/entrypoint.sh
upbqdn: I added it to debug the entrypoint when it used to create its own config when we didn't pass any. That bug is now gone, so I thought it was redundant since the entrypoint will now fail elsewhere if we don't provide the config. There are also other vars that the entrypoint depends on, which we don't gate like this one.

upbqdn: The only downside is that we don't get a nice error message anymore, but we shouldn't run into those errors anyway.

gustavovalverde: If a user changes the $ZEBRA_CONF_PATH to another directory, and this directory doesn't has a proper configuration file, then it will be hard for them to understand what's going on.

docker/entrypoint.sh
upbqdn: There are two reasons:

1. It unifies the execution path for test and production containers. We'd have to have it separately for tests and production in the Dockerfile.
2. I mainly did it this way because the ownership change happens right before we execute Zebra through gosu. One of the bugs causing the CI to fail was that the lwd dir ownership change was missing.

By overriding you mean that they use their own entrypoint?

We can move it back to the Dockerfile if it's a blocker.

docker/entrypoint.sh
gustavovalverde: This was originally added to be able to identify edge cases when doing changes to how we handle cache states. Although this is not a present problem, it's helpful to identify issues when making changes to how we mount the cache states, or to anyone else trying to do it. https://github.com/ZcashFoundation/zebra/pull/4409 as after a container has been deployed in CI, we can't access it to debug inside it.

But we would have to restore the functionality: https://github.com/ZcashFoundation/zebra/blob/8cfb61f52c0948cab9e57df7222244799de379e5/docker/entrypoint.sh#L250

docker/entrypoint.sh
upbqdn: The var has to point to a file, not a dir. This check only checks the existence of the config file, not its contents, so even if the user points the var to any file, the check will pass. But we can surely keep the check around.

docker/entrypoint.sh
upbqdn: I thought future maintainers might be confused by the function since its not used anywhere, but we can surely keep it.

conradoplg:
No need to spend too much time on this, but I think it would be useful to list in the PR description what are the _problems_ that are being addressed (e.g. what problem was fixed by explicitly specify the lwd dir path?) so we have it documented and other people can understand what happened.

upbqdn:
I'll try.

upbqdn:
I updated the description. I agree that it's pretty non-trivial to see how the code changes in this PR resolve the issues since the PR just seems to shuffle a bunch of things around.

docker/entrypoint.sh
gustavovalverde: I've reverted this to the Dockerfile.

docker/entrypoint.sh
gustavovalverde: I've reverted this check, in the meanwhile.

gustavovalverde (approved):
LGTM

upbqdn:
The last few commits break things. I don't understand how this could be merged into main with no review.

upbqdn:
This PR also fixes the path for the custom config. I don't understand why that was removed from the PR description. It also unified the execution path for tests and production. That is gone from the code and PR description.

PR #9327, opened 2025-03-14
feat(rpc): fill size field in getblock with verbosity=2

### LINKED ISSUE #9020: getblock: return the `size` field
### Motivation

To match zcashd we need to return the `size` field to the `getblock` RPC.

### Specifications

_No response_

Be aware that the [naive approach might have very poor performance](https://github.com/ZcashFoundation/zebra/pull/9006#issuecomment-2465993367). Look into alternatives (can it be computed on the fly from existing data?), with the last resort adding a new column to the database.

An idea: the block header and txs are stored serialized in the database. The block size can be computed from those, without requiring a serialization roundtrip, which would already save time (deserialization is expensive). This would require adding new State service functions to get the raw block / txs. Also, look whether getting just the size of those from the database is even faster.

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

This fills the `size` field in the `getblock` RPC when `verbosity=2`.

To support `verbosity=1` we would need to either take a performance hit (load all transactions from a block to get their sizes) or change the database to create a tx -> size index (which would be good to avoid if possible)

Might close #9020

## Solution

Add a `BlockAndSize` request to the state which reads the raw block header and txs, gets their sizes, computes the block size from them, and finally parses it all to return the block.

### Tests

I adjusted existing tests. I also tested it manually with `zcash-rpc-diff`.

The block size calculation logic was also tested manually with a throwaway program that computed block sizes from the state with both approaches and asserted that they are equal.

### Follow-up Work

We need to decide if this is enough for now or if we want to support verbosity=1. In that case, we will need to create a new issue for it.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-state/src/service/finalized_state/zebra_db/block.rs
+++ b/zebra-state/src/service/finalized_state/zebra_db/block.rs
@@ -191,8 +191,11 @@ impl ZebraDb {
         // transactions. This requires summing them all and also adding the
         // size of the CompactSize-encoded transaction count.
         // See https://developer.bitcoin.org/reference/block_chain.html#serialized-blocks
-        let tx_count = CompactSizeMessage::try_from(txs.len()).unwrap();
-        let tx_raw = tx_count.zcash_serialize_to_vec().unwrap();
+        let tx_count = CompactSizeMessage::try_from(txs.len())
+            .expect("must work for a previously serialized block");
+        let tx_raw = tx_count
+            .zcash_serialize_to_vec()
+            .expect("must work for a previously serialized block");
         let size = raw_header.raw_bytes().len()
             + raw_txs
                 .iter()
--- a/zebra-state/src/service/finalized_state/zebra_db/block.rs
+++ b/zebra-state/src/service/finalized_state/zebra_db/block.rs
@@ -24,7 +24,7 @@ use zebra_chain::{
     parallel::tree::NoteCommitmentTrees,
     parameters::{Network, GENESIS_PREVIOUS_BLOCK_HASH},
     sapling,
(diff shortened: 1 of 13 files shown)


zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -161,6 +175,56 @@ impl ZebraDb {
...
+            .iter()
+            .map(|raw_tx| Arc::<Transaction>::from_bytes(raw_tx.raw_bytes()))
+            .collect();
+
+        // Compute the size of the block from the size of header and size of
+        // transactions. This requires summing them all and also adding the
+        // size of the CompactSize-encoded transaction count.
+        // See https://developer.bitcoin.org/reference/block_chain.html#serialized-blocks
+        let tx_count = CompactSizeMessage::try_from(txs.len()).unwrap();
+        let tx_raw = tx_count.zcash_serialize_to_vec().unwrap();
```
oxarbitrage: Will it be better to return `None` instead of panicking ?

oxarbitrage (dismissed):
Looks good to me, thank you.

zebra-state/src/service/finalized_state/zebra_db/block.rs
conradoplg: Blocks deserialized from the database are guaranteed to be serializable without errors. I should have used expect though, I will change it

conradoplg: Done in 826c6bf3d

oxarbitrage (approved):
thanks @conradoplg

PR #9328, opened 2025-03-14
fix(CI): Cache compilation results

### LINKED ISSUE #9331: CI recompiles Zebra from scratch for each CI test
Each time a PR is updated with a commit, CI recompiles Zebra from sratch for each set of tests, which is ~ 15 times. In some tests, the compilation takes longer than the test itself. CI should cache the compilation results and recompile only what's necessary, not only between image preparation and tests but also between individual preparations. We have the latter implemented: https://github.com/zcashfoundation/zebra/blob/49741e8b476aff1d1dd2f6a2f57ec9d24404d345/docker/Dockerfile#L97-L126, but it doesn't work. CI should also cache downloaded crates instead of pulling them for each compilation.

The solution should cache:

- Downloaded crates between image preparations.
- Compiled crates between image preparations.
- Compiled crates between image preparation and tests.

Ideally, CI should compile Zebra only once per commit for the `tests` Docker target: https://github.com/zcashfoundation/zebra/blob/882aa038c32e1018959f2617f9258cfdd5e16089/.github/workflows/ci-tests.yml#L130, without recompiling unmodified crates and redownloading all deps.

upbqdn:
## Motivation

- Close #9331.

Depends-On: #9344

## Solution

- Cache downloaded crates between image preparations: https://github.com/ZcashFoundation/zebra/actions/runs/13837431819/job/38722773169?pr=9323#step:10:367 vs https://github.com/ZcashFoundation/zebra/actions/runs/13867297186/job/38808809286?pr=9328#step:10:1.
- Cache compiled crates between image preparations: https://github.com/ZcashFoundation/zebra/actions/runs/13837431819/job/38722773169?pr=9323#step:10:871 vs https://github.com/ZcashFoundation/zebra/actions/runs/13867297186/job/38808809286?pr=9328#step:10:387.
- Cache compiled crates between image preparation and tests: https://github.com/ZcashFoundation/zebra/actions/runs/13837431819/job/38722794636?pr=9323#step:14:146 vs https://github.com/ZcashFoundation/zebra/actions/runs/13867297283/job/38809190703?pr=9328#step:14:134. For this test, the runtime went down from 33 minutes to 13 minutes.
- Unify compilation parameters for all CI tests so caching can kick in.

### Clean-ups

- Mount lwd's and Zebra's cached states to implicit locations instead of explicitly specifying the location in each test.

### Tests

- CI is noticeably faster.

### PR Checklist

- [x] The solution is tested.


upbqdn:
This PR seems to make CI noticeably faster.

upbqdn:
We could bring the CI time down to ~25 minutes if we moved these three tests to run only on `main`:

image

docker/Dockerfile
```diff
@@ -70,48 +70,31 @@ FROM deps AS tests
...
+    --mount=type=cache,target=${CARGO_HOME} \
+    cargo test --locked --release --no-run --features "${FEATURES}" && \
+    cp ${CARGO_TARGET_DIR}/release/zebrad /usr/local/bin && \
+    cp ${CARGO_TARGET_DIR}/release/zebra-checkpoints /usr/local/bin && \
+    cp -a ${CARGO_TARGET_DIR} ${HOME}/tmp_target  && \
+    cp -a ${CARGO_HOME} ${HOME}/tmp_cargo_home
+
+RUN rm -rf ${CARGO_TARGET_DIR} && rm -rf ${CARGO_HOME} && \
+    mv ${HOME}/tmp_target ${CARGO_TARGET_DIR}  && \
+    mv ${HOME}/tmp_cargo_home ${CARGO_HOME}
```
upbqdn, line 92: This is a rather awful hack. I think there could a cleaner solution, but I coulnd't come up with anything better.

conradoplg: Why is this needed?

gustavovalverde: This might be the best context to answer that https://github.com/ZcashFoundation/zebra/issues/6169#issuecomment-1712776391

gustavovalverde:
@upbqdn can you rebase this PR so we can re-test it?

Or we can use:
- https://github.com/ZcashFoundation/zebra/pull/9435

upbqdn:
This PR is based on the one we reverted. I don't think it's worth rebasing. It'll be easier to start from scratch again.

PR #9334, opened 2025-03-18
fix(docker): add default CMD to test stage in Dockerfile 

gustavovalverde:
## Motivation

When running the container using docker-compose without explicitly 
providing a command, the entrypoint.sh script was attempting to execute
`exec_as_user` with no arguments, resulting in a gosu error:

```shell
exec_as_user
exec gosu 10001:10001
Usage: gosu user-spec command [args]
```

Depends-On: #9333

## Solution

By adding `CMD ["cargo", "test"]` to the test stage in the Dockerfile, we ensure a default command is available for the entrypoint script to execute, preventing the `gosu` error when no command is explicitly provided.

This fix allows `docker-compose.test.yml` to run successfully without needing to specify a command in the service definition.

### Tests

There are no unit tests validating this, but running `docker compose -f docker/docker-compose.test.yml up --build` will trigger it (without this PR).

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -72,6 +72,7 @@ ENV GID=${GID}
 ARG USER
 ENV USER=${USER}
 ARG HOME
+ENV HOME=${HOME}
 
 RUN addgroup --quiet --gid ${GID} ${USER} && \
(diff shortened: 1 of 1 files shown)


arya2 (approved):

gustavovalverde:
@mergifyio requeue

PR #9339, opened 2025-03-20
Experiment: `Transaction::V6` variants only

aphelionz:
## Motivation + Solution

On a recent call with @conradoplg and @arya2, we decided to try this out to make the NSM and ZSA PRs more managable.

This PR only adds the `Transaction::V6` enum variant and then satisfies the compiler. The variant and its respective match arms and array members are feature-flagged under `tx_v6`.

### Tests

- `cargo test --workspace` locally

### Follow-up Work

If this approach is suitable, I can do `NetworkUpgrade::Nu7` as well.

### PR Checklist

- [ ] The solution is tested.


aphelionz:
Hi @conradoplg Thanks for taking a look at this. Can you approve the workflows again? 🙏

aphelionz:
@conradoplg added those feature deps, thank you!

zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
oxarbitrage: please remove this file. it is generated when running tests locally. It will be nice to add this type of file regex to `.gitignore` but this is unrelated.

aphelionz:
Looks like it's hitting the Clippy lints from #9371. I will rebase this PR after that one is merged.

aphelionz:
Hi again @conradoplg @oxarbitrage can you please approve the workflows? 🙏

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
+name: Sync to upstream
```
oxarbitrage: I suppose this file needs to be removed ?

zebra-chain/src/transaction.rs
```diff
@@ -457,6 +488,14 @@ impl Transaction {
                 block::Height(0) => None,
                 block::Height(expiry_height) => Some(block::Height(*expiry_height)),
             },
+            #[cfg(feature = "tx_v6")]
+            Transaction::V6 { expiry_height, .. } => match expiry_height {
+                // Consensus rule:
```
oxarbitrage:
```suggestion
                // # Consensus
                //
```

We use a special comment for consensus rules so we can find them all searching for the `# Consensus` string in the codebase.

For example: https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/transparent.rs#L210

oxarbitrage (commented):
some fast comments before we make a more formal review.

aphelionz:
Thanks for the comments @oxarbitrage 🙏

aphelionz:
Happy new week @oxarbitrage and @conradoplg 🙏 

I have fixed the merge error

aphelionz:
Thanks for running the workflows again - how should I deal with the `cargo deny` issue? It looks like it's Windows-specific

deny.toml
```diff
@@ -98,6 +98,10 @@ skip-tree = [
 
     # wait until zcash_client_backend update rustix
     { name = "rustix", version = "=0.38.44" },
+
+    # wait for windows-related updates
+    { name = "windows-strings", version = "=0.4.0" },
+    { name = "windows-core", version = "=0.61.0" }
```
oxarbitrage:
```suggestion
    # wait for reqwest to update windows-registry
    { name = "windows-strings", version = "=0.3.1" }
```

Thanks for adding this, i am not totally sure why they are happening here but there are a few things to consider anyway:

- `windows-core` is not needed as the issue is just with `windows-strings`.
- We add the older version of the duplicated dependency here, in this case `0.3.1` of `windows-strings`.
- The comment tries to document what dependency is causing the issue, in this case this is `reqwest`.

aphelionz: I can update but just note: in my local testing windows-core also came up as an issue.

zebra-chain/src/transaction/serialize.rs
```diff
@@ -672,6 +672,53 @@ impl ZcashSerialize for Transaction {
...
+                network_upgrade,
+                lock_time,
+                expiry_height,
+                inputs,
+                outputs,
+                sapling_shielded_data,
+                orchard_shielded_data,
+            } => {
+                // Transaction V5 spec:
+                // https://zips.z.cash/protocol/protocol.pdf#txnencoding
```
oxarbitrage:
```suggestion
                // Transaction V6 spec:
                // https://zips.z.cash/zip-0230#specification
```

I think this is the only reference we have to transaction v6 for now. All fields from tx v5 are present in v6, which is good for this code as it is a copy tx v5.

zebra-chain/src/transaction/serialize.rs
```diff
@@ -672,6 +672,53 @@ impl ZcashSerialize for Transaction {
...
+                // A bundle of fields denoted in the spec as `nSpendsSapling`, `vSpendsSapling`,
+                // `nOutputsSapling`,`vOutputsSapling`, `valueBalanceSapling`, `anchorSapling`,
+                // `vSpendProofsSapling`, `vSpendAuthSigsSapling`, `vOutputProofsSapling` and
+                // `bindingSigSapling`.
+                sapling_shielded_data.zcash_serialize(&mut writer)?;
+
+                // A bundle of fields denoted in the spec as `nActionsOrchard`, `vActionsOrchard`,
+                // `flagsOrchard`,`valueBalanceOrchard`, `anchorOrchard`, `sizeProofsOrchard`,
+                // `proofsOrchard`, `vSpendAuthSigsOrchard`, and `bindingSigOrchard`.
+                orchard_shielded_data.zcash_serialize(&mut writer)?;
```
oxarbitrage, line 720:
```suggestion
                orchard_shielded_data.zcash_serialize(&mut writer)?;
                
                // TODO: Add the rest of v6 transaction fields.
```

We can be more clear that some fields are missing.

zebra-chain/src/transaction/tests/vectors.rs
```diff
@@ -908,6 +908,36 @@ fn binding_signatures() {
...
+                            at_least_one_v5_checked = true;
+                        }
+                    }
+                    #[cfg(feature = "tx_v6")]
+                    Transaction::V6 {
+                        sapling_shielded_data,
+                        ..
+                    } => {
+                        if let Some(sapling_shielded_data) = sapling_shielded_data {
+                            // V5 txs have the outputs spent by their transparent inputs hashed into
```
oxarbitrage:
```suggestion
                            // V6 txs have the outputs spent by their transparent inputs hashed into
```

please let's use V6 everywhere possible.

zebra-chain/src/transaction/txid.rs
```diff
@@ -28,6 +28,8 @@ impl<'a> TxIdBuilder<'a> {
             | Transaction::V3 { .. }
             | Transaction::V4 { .. } => self.txid_v1_to_v4(),
             Transaction::V5 { .. } => self.txid_v5(),
+            #[cfg(feature = "tx_v6")]
+            Transaction::V6 { .. } => self.txid_v5(),
```
oxarbitrage: Is it possible to create `txid_v6()` and use it here?

zebra-consensus/src/transaction.rs
```diff
@@ -523,6 +523,19 @@ where
                     sapling_shielded_data,
                     orchard_shielded_data,
                 )?,
+                #[cfg(feature="tx_v6")]
+                Transaction::V6 {
+                    sapling_shielded_data,
+                    orchard_shielded_data,
+                    ..
+                } => Self::verify_v5_transaction(
```
oxarbitrage: We can create `verify_v6_transaction` behind the flag and call `verify_v5_transaction` from it with a comment that says this is temporal code to be updated with proper v6 verification.

deny.toml
```diff
@@ -98,6 +98,12 @@ skip-tree = [
 
     # wait until zcash_client_backend update rustix
     { name = "rustix", version = "=0.38.44" },
+
+    # wait for reqwest to update windows-registry
+    { name = "windows-strings", version = "=0.3.1" },
+
+    # wait for sentry to update windows-core
+    { name = "windows-core", version = "=0.52.0" }
```
oxarbitrage, line 106: I think this one is not needed to be here but i might be wrong. We can try to remove and check if the CI fails.

zebra-chain/src/transaction.rs
```diff
@@ -143,6 +143,27 @@ pub enum Transaction {
...
+        /// The latest block height that this transaction can be added to the chain.
+        expiry_height: block::Height,
+        /// The transparent inputs to the transaction.
+        inputs: Vec<transparent::Input>,
+        /// The transparent outputs from the transaction.
+        outputs: Vec<transparent::Output>,
+        /// The sapling shielded data for this transaction, if any.
+        sapling_shielded_data: Option<sapling::ShieldedData<sapling::SharedAnchor>>,
+        /// The orchard data for this transaction, if any.
+        orchard_shielded_data: Option<orchard::ShieldedData>,
```
oxarbitrage, line 165:
```suggestion
        orchard_shielded_data: Option<orchard::ShieldedData>,
        
        // TODO: Add the rest of the v6 fields.
```

zebra-chain/src/transaction/arbitrary.rs
```diff
@@ -924,6 +924,8 @@ pub fn transaction_to_fake_v5(
             orchard_shielded_data: None,
         },
         v5 @ V5 { .. } => v5.clone(),
+        #[cfg(feature = "tx_v6")]
+        v5 @ V6 { .. } => v5.clone(),
```
oxarbitrage: Will be nice to have ` v6` here but as it is test code it doesn't matter much.

oxarbitrage (commented):
Looking good, nice work. I have some comments that are mainly about naming the new stuff as v6 instead of v5.

deny.toml
aphelionz: I double-checked this locally and it indeed fails. Happy to remove it but you can verify with `cargo deny --workspace --features default-release-binaries check bans --allow unmatched-skip-root` like the workflow runs

oxarbitrage: Ok, my fault, i did not tested locally with the right command. Thanks!

oxarbitrage (approved):
Looks good enough to me, thank you for applying all the suggestions.

Issue #9037, opened 2024-11-18
devops: add more unit-tests for our docker and docker-compose


gustavovalverde:
### Describe the issue or request

We've been making changes to how our Dockerfile and Zebra behaves with other applications, and we should test this integrations when third applications, like lightwalletd must connect to zebra from a different machine/container, as some breaking changes might be introduced without us realizing.

### Expected Behavior

We must be notified if third party integrations are being impact by changes in Zebra or Docker

### Current Behavior

No information or testing around this subject

### Possible Solution

Add a test using our docker-compose, validating the integration with their respective health-checks are working.

PR #9344, opened 2025-03-24
refactor(docker): Enhance Zebra configuration options and entrypoint logic

### LINKED ISSUE #9325: Using custom config in Docker fails
Running 

```bash
docker run --rm -it --mount type=bind,src=/path/to/zebra/docker/default-zebra-config.toml,dst=/home/zebra/.config/zebrad.toml zebra:local
```

fails with an error. The issue is that the entrypoint tries to overwrite the mounted file on the host.

This is a release blocker.

One solution would be that the entrypoint would copy the mounted file to `/tmp` in the container and adjust it there before passing it to Zebra. This assumes that the `/tmp` directory is available in the container.

gustavovalverde:
## Motivation

We had some issues with how we were dealing with configuration files:
- If the user mounted a configuration file, zebra would (try to) overwrite it. This could cause odd behaviors like:
  - #9325 
- When mounting a configuration file using the [docker compose configs](https://docs.docker.com/reference/compose-file/configs/) this would cause an error as the mounted directory is _read-only_
- We should be able to differentiate between a _user-provided_ configuration file, our _default_ configuration file (mainly used with docker compose), and if there's no configuration file provided, so we can build it _on-the-fly_ with the provided variables by the user.

A possible approach to fix this was using a temporal directory, and using the resulting config from that directory instead from the original mounted dir, but this was also problematic if the temporal directory was mounted as read-only from the host, which also led to the decision of not using `sed` and trying to replace the content of the file or creating a new file in a location that could potentially be `read-only` by external host limitations.

Fixes #9325 

## Solution

**Key changes:**
- Do not mount a default configuration file in our Dockerfile (`default-zebra-config.toml`), to allow more flexibility: default configuration build, default config file mounting, user-provided config file mounting with a custom `$ZEBRA_CONF_PATH`
  - Validate if the user is mounting its own file (this requires them to set `$ZEBRA_CONF_PATH` to a valid mounted file)
  - If the user doesn't set `ZEBRA_CONF_PATH` then validate if the default `${HOME}/.config/zebrad.toml` is mounted (commonly provided with our docker compose https://github.com/ZcashFoundation/zebra/blob/ac25192afcb1c3280ff32d2bb2e75f91f1d9ceb0/docker/docker-compose.yml#L23)
  - Otherwise, create a zebra configuration file based on the variables set by the user (or using the default values)
- Do not try to overwrite mounted files, dynamic configuration should only happen `on-the-fly` and must be avoided if a configuration file was provided (this changes `sed` to `cat` as using `sed` is no longer required)

**Other (required) changes**
- `exec_as_user` must validate if the actual user is root, otherwise running `gosu` will fail when that's not the case
- If cache or other directories were configured with the environment variables, make sure they exists and change their ownership to the user running `zebrad`.
- Do not mount the `.cookies` directory in the same directory as the `.cache` as mounting a cached state would override the cookies directory.

### Tests

All tests should be passing, and building with the docker compose, with our without a mount file, should give the expected configuration results.

### Follow-up Work

Validate which other configurations should be added using variables

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/book/src/user/docker.md
+++ b/book/src/user/docker.md
@@ -59,52 +59,47 @@ All available Cargo features are listed at
 
 ## Configuring Zebra
 
-To configure Zebra, you have three options:
+To configure Zebra using Docker, you have a few options, processed in this order:
 
-1. Edit the `docker/default-zebra-config.toml` config file and uncomment the `configs` mapping in `docker/docker-compose.yml` so your config takes effect.
-2. Provide a path to your own config file via the `ZEBRA_CONF_PATH` environment variable.
-3. Let Zebra generate a default config file automatically from environment variables (if neither of the above methods is used).
-
-The entrypoint script follows this sequence:
-
-1. If `ZEBRA_CONF_PATH` is set and a file exists at that path, it uses that file
-2. If `ZEBRA_CONF_PATH` is not set but a default config exists at `${HOME}/.config/zebrad.toml`, it uses that file
-3. If neither exists, it generates a default config at `${HOME}/.config/zebrad.toml` based on environment variables
-4. Environment variables from the `docker/.env` only take effect if a configuration file is not mounted, and thus are only effective if you don't provide a custom config file.
+1. **Provide a specific config file path:** Set the `ZEBRA_CONF_PATH` environment variable to point to your config file within the container.
+2. **Use the default config file:** If `ZEBRA_CONF_PATH` is not set, Zebra will look for a config file at `/home/zebra/.config/zebrad.toml`. You can mount your custom config file to this location, for example, by uncommenting the `configs` mapping in `docker/docker-compose.yml`.
+3. **Generate config from environment variables:** If neither of the above methods provides a config file, the container's entrypoint script will *automatically generate* a default configuration file at `/home/zebra/.config/zebrad.toml`. This generated file uses specific environment variables (like `NETWORK`, `ZEBRA_RPC_PORT`, `ENABLE_COOKIE_AUTH`, `MINER_ADDRESS`, etc.) to define the settings. Using the `docker/.env` file is the primary way to set these variables for this auto-generation mode.
 
 You can see if your config works as intended by looking at Zebra's logs.
 
-Note that the environment variable overrides in the `docker/.env` file are limited to the variables already present in the commented out blocks and adding new ones will not be effective. The `docker/.env` file serves as a quick way to override the most commonly used settings, whereas a custom config file provides full configuration capabilities.
+Note that if you provide a configuration file using methods 1 or 2, environment variables from `docker/.env` will **not** override the settings within that file. The environment variables are primarily used for the auto-generation scenario (method 3).
 
 ### RPC
 
-Zebra's RPC server is disabled by default. To enable it, you need to set its RPC
-port. You can do that either in the `docker/default-zebra-config.toml` file or
-`docker/.env` file, as described in the two paragraphs above.
+Zebra's RPC server is disabled by default. To enable it, you need to define the RPC settings in Zebra's configuration. You can achieve this using one of the configuration methods described above:
 
-When connecting to Zebra's RPC server, your RPC clients need to provide an
-authentication cookie to the server or you need to turn the authentication off
-in Zebra's config. By default, the cookie file is stored at
-`/home/zebra/.cache/zebra/.cookie` in the container. You can print its contents
-by running
+*   **Using a config file (methods 1 or 2):** Add or uncomment the `[rpc]` section in your `zebrad.toml` file (like the one provided in `docker/default-zebra-config.toml`). Ensure you set the `listen_addr` (e.g., `"0.0.0.0:8232"` for Mainnet).
+*   **Using environment variables (method 3):** Set the `ZEBRA_RPC_PORT` environment variable (e.g., in `docker/.env`). This tells the entrypoint script to include an enabled `[rpc]` section listening on `0.0.0.0:<ZEBRA_RPC_PORT>` in the auto-generated configuration file.
 
-```bash
-docker exec zebra cat /home/zebra/.cache/zebra/.cookie
-```
+**Cookie Authentication:**
 
-when the `zebra` container is running. Note that Zebra generates the cookie file
-only if the RPC server is enabled, and each Zebra instance generates a unique
-one. To turn the authentication off, either set `ENABLE_COOKIE_AUTH=false` in
-`docker/.env` or set
+By default, Zebra uses cookie-based authentication for RPC requests (`enable_cookie_auth = true`). When enabled, Zebra generates a unique, random cookie file required for client authentication.
 
-```toml
-[rpc]
-enable_cookie_auth = false
-```
+* **Cookie Location:** The entrypoint script configures Zebra to store this file at `/home/zebra/.cache/zebra/.cookie` inside the container.
+* **Viewing the Cookie:** If the container is running and RPC is enabled with authentication, you can view the cookie content using:
 
-in `docker/default-zebra-config.toml` and mount this config file into the
-container's filesystem in `docker/docker-compose.yml` as described at the
-beginning of this section.
+    ```bash
+    docker exec <container_name> cat /home/zebra/.cache/zebra/.cookie
+    ```
+
+    (Replace `<container_name>` with your container's name, typically `zebra` if using the default `docker-compose.yml`). Your RPC client will need this value.
+* **Disabling Authentication:** If you need to disable cookie authentication (e.g., for compatibility with tools like `lightwalletd`):
+  * If using a **config file** (methods 1 or 2), set `enable_cookie_auth = false` within the `[rpc]` section:
+
+    ```toml
+    [rpc]
+    # listen_addr = ...
+    enable_cookie_auth = false
+    ```
+
+  * If using **environment variables** for auto-generation (method 3), set `ENABLE_COOKIE_AUTH=false` in your `docker/.env` file.
+
+Remember that Zebra only generates the cookie file if the RPC server is enabled *and* `enable_cookie_auth` is set to `true` (or omitted, as `true` is the default).
 
 ## Examples
 
@@ -124,8 +119,8 @@ Note that Docker will run Zebra with the RPC server enabled and the cookie
 authentication mechanism disabled since Lightwalletd doesn't support it. Instead
 of configuring Zebra via the recommended config file or `docker/.env` file, we
 configured the RPC server by setting environment variables directly in the
-`docker/docker-compose.lwd.yml` file. This is a shortcut we can take when we are
-familiar with the `docker/.env` file.
+`docker/docker-compose.lwd.yml` file. This takes advantage of the entrypoint
+script's auto-generation feature (method 3 described above).
 
 ### Running Zebra with Prometheus and Grafana
 
--- a/book/src/user/docker.md
+++ b/book/src/user/docker.md
@@ -62,8 +62,8 @@ All available Cargo features are listed at
 To configure Zebra using Docker, you have a few options, processed in this order:
 
 1. **Provide a specific config file path:** Set the `ZEBRA_CONF_PATH` environment variable to point to your config file within the container.
-2. **Use the default config file:** If `ZEBRA_CONF_PATH` is not set, Zebra will look for a config file at `/home/zebra/.config/zebrad.toml`. You can mount your custom config file to this location, for example, by uncommenting the `configs` mapping in `docker/docker-compose.yml`.
-3. **Generate config from environment variables:** If neither of the above methods provides a config file, the container's entrypoint script will *automatically generate* a default configuration file at `/home/zebra/.config/zebrad.toml`. This generated file uses specific environment variables (like `NETWORK`, `ZEBRA_RPC_PORT`, `ENABLE_COOKIE_AUTH`, `MINER_ADDRESS`, etc.) to define the settings. Using the `docker/.env` file is the primary way to set these variables for this auto-generation mode.
+2. **Use the default config file:** By default, the `docker-compose.yml` file mounts `./default-zebra-config.toml` to `/home/zebra/.config/zebrad.toml` using the `configs:` mapping. Zebra will use this file if `ZEBRA_CONF_PATH` is not set. To use environment variables instead, you must **comment out** the `configs:` mapping in `docker/docker-compose.yml`.
+3. **Generate config from environment variables:** If neither of the above methods provides a config file (i.e., `ZEBRA_CONF_PATH` is unset *and* the `configs:` mapping in `docker-compose.yml` is commented out), the container's entrypoint script will *automatically generate* a default configuration file at `/home/zebra/.config/zebrad.toml`. This generated file uses specific environment variables (like `NETWORK`, `ZEBRA_RPC_PORT`, `ENABLE_COOKIE_AUTH`, `MINER_ADDRESS`, etc.) to define the settings. Using the `docker/.env` file is the primary way to set these variables for this auto-generation mode.
 
 You can see if your config works as intended by looking at Zebra's logs.
 
@@ -73,8 +73,8 @@ Note that if you provide a configuration file using methods 1 or 2, environment 
 
 Zebra's RPC server is disabled by default. To enable it, you need to define the RPC settings in Zebra's configuration. You can achieve this using one of the configuration methods described above:
 
-*   **Using a config file (methods 1 or 2):** Add or uncomment the `[rpc]` section in your `zebrad.toml` file (like the one provided in `docker/default-zebra-config.toml`). Ensure you set the `listen_addr` (e.g., `"0.0.0.0:8232"` for Mainnet).
-*   **Using environment variables (method 3):** Set the `ZEBRA_RPC_PORT` environment variable (e.g., in `docker/.env`). This tells the entrypoint script to include an enabled `[rpc]` section listening on `0.0.0.0:<ZEBRA_RPC_PORT>` in the auto-generated configuration file.
+* **Using a config file (methods 1 or 2):** Add or uncomment the `[rpc]` section in your `zebrad.toml` file (like the one provided in `docker/default-zebra-config.toml`). Ensure you set the `listen_addr` (e.g., `"0.0.0.0:8232"` for Mainnet).
+* **Using environment variables (method 3):** Set the `ZEBRA_RPC_PORT` environment variable (e.g., in `docker/.env`). This tells the entrypoint script to include an enabled `[rpc]` section listening on `0.0.0.0:<ZEBRA_RPC_PORT>` in the auto-generated configuration file.
 
 **Cookie Authentication:**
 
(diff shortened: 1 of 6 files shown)


docker/default-zebra-config.toml
```diff
@@ -23,7 +23,7 @@ cache_dir = "/home/zebra/.cache/zebra"
 # listen_addr = "0.0.0.0:8232" # Mainnet
 # listen_addr = "0.0.0.0:18232" # Testnet
 
-cookie_dir = "/home/zebra/.cache/zebra"
+cookie_dir = "/home/zebra/.config/cookie"
```
upbqdn: This is a breaking change. We need to update the existing docs and add new since the location differs from where Zebra stores the cookie by default.

book/src/user/docker.md
```diff
@@ -59,19 +59,22 @@ All available Cargo features are listed at
...
+To configure Zebra, you have three options:
+
+1. Edit the `docker/default-zebra-config.toml` config file and uncomment the `configs` mapping in `docker/docker-compose.yml` so your config takes effect.
+2. Provide a path to your own config file via the `ZEBRA_CONF_PATH` environment variable.
+3. Let Zebra generate a default config file automatically from environment variables (if neither of the above methods is used).
+
+The entrypoint script follows this sequence:
+
+1. If `ZEBRA_CONF_PATH` is set and a file exists at that path, it uses that file
+2. If `ZEBRA_CONF_PATH` is not set but a default config exists at `${HOME}/.config/zebrad.toml`, it uses that file
```
upbqdn: It won't be the default config if users mount their own.

```suggestion
2. If `ZEBRA_CONF_PATH` is not set but a file exists at `${HOME}/.config/zebrad.toml`, it uses that file
```

docker/entrypoint.sh
```diff
@@ -6,113 +6,125 @@
 #
 # ## Notes
 #
-# - `$ZEBRA_CONF_PATH` must point to a Zebra conf file.
+# - `$ZEBRA_CONF_PATH` can point to an existing Zebra config file, or if not set,
+#   the script will look for a default config at ${HOME}/.config/zebrad.toml,
```
upbqdn, line 10:
```suggestion
#   the script will look for a config at ${HOME}/.config/zebrad.toml,
```

docker/entrypoint.sh
```diff
@@ -255,12 +267,39 @@ run_tests() {
...
+    echo "Using user-provided config file"
+  else
+    echo "ERROR: ZEBRA_CONF_PATH was set and no config file found at ${ZEBRA_CONF_PATH}."
+    echo "Please ensure a config file exists or set ZEBRA_CONF_PATH to point to your config file."
+    exit 1
+  fi
+else
+  if [[ -f "${HOME}/.config/zebrad.toml" ]]; then
+    echo "ZEBRA_CONF_PATH was not set."
+    echo "Using default config at ${HOME}/.config/zebrad.toml"
```
upbqdn, line 298:
```suggestion
    echo "Using the config at ${HOME}/.config/zebrad.toml"
```

docker/entrypoint.sh
```diff
@@ -255,12 +267,39 @@ run_tests() {
...
+    ZEBRA_CONF_PATH="${HOME}/.config/zebrad.toml"
+  else
+    echo "ZEBRA_CONF_PATH was not set and no default config found at ${HOME}/.config/zebrad.toml"
```
upbqdn, line 301:
```suggestion
    echo "ZEBRA_CONF_PATH was not set and no config found at ${HOME}/.config/zebrad.toml"
```

docker/.env
```diff
@@ -1,30 +1,60 @@
...
+
 # Sets custom Cargo features. Available features are listed at
 # <https://docs.rs/zebrad/latest/zebrad/index.html#zebra-feature-flags>.
 #
 # Must be set at build time.
 #
 # FEATURES=""
 
+# Sets the listen address and port for Prometheus metrics.
+# Only active when the 'prometheus' feature is enabled.
```
upbqdn: We'll make the `prometheus` feature enabled by default soon.

```suggestion
```

docker/.env
```diff
@@ -1,30 +1,60 @@
...
+# ZEBRA_CONF_PATH="/path/to/your/custom/zebrad.toml"
+
 # Sets the network Zebra runs will run on.
 #
 # NETWORK=Mainnet
 
+# Sets the listen address for Zebra's peer-to-peer network connections.
+#
+# ZEBRA_LISTEN_ADDR="0.0.0.0"
+
```
upbqdn: I don't think users should be presented with this option since using any other address will make the endpoint inaccessible outside Docker.

docker/.env
```diff
@@ -1,30 +1,60 @@
...
+# ZEBRA_COOKIE_DIR="/home/zebra/.config/cookie"
+
+# Sets a custom directory for the state and network caches.
 #
 # ZEBRA_CACHE_DIR="/home/zebra/.cache/zebra"
 
+# Sets a custom directory for the lightwalletd cache.
+#
+# LWD_CACHE_DIR="/home/zebra/.cache/lwd"
+
```
upbqdn: Is there any use-case for `LWD_CACHE_DIR`?

```suggestion
```

docker/.env
```diff
@@ -1,30 +1,60 @@
 # Configuration variables for running Zebra in Docker
 
+# Sets the path to a custom Zebra config file. If not set, Zebra will look for a config at
+# ${HOME}/.config/zebrad.toml or generate one using environment variables below.
+# ! Setting ZEBRA_CONF_PATH will make most of the following environment variables ineffective.
+# TODO: zebrad should be able to read from both the config file and environment variables.
```
upbqdn: Since this is a user-facing config file, I expect the TODO will be confusing.

```suggestion
```

upbqdn (commented):
@gustavovalverde, have you tested the solution? It looks like some parts don't work. For example, setting the RPC port.

If I set `ZEBRA_RPC_PORT=28232`, the script configures Zebra to use

```toml
[rpc]
listen_addr = "0.0.0.0:8232"
```

which is not the desired port.

upbqdn:
The issue is similar to what we had before the refactor:

Even setting `ZEBRA_RPC_PORT=catchthisbug` makes Zebra run just fine. It will be hard to debug this if we have something wrong, but things will run anyway.

upbqdn:
I wonder if there's a way we could test these things. It should be quick to write a bunch of basic tests if we had a suitable framework for them.

gustavovalverde:
> I wonder if there's a way we could test these things. It should be quick to write a bunch of basic tests if we had a suitable framework for them.

I'll start a PR tomorrow just for that. There's just way too many edge-cases that I can forget and can go wrong with this changes, I'll just start building a test suite and make this PR dependent on that.

I'll start with https://github.com/ZcashFoundation/zebra/issues/9037

I'll put this in draft, and let's revisit once the test suite is good enough

gustavovalverde:
> I wonder if there's a way we could test these things. It should be quick to write a bunch of basic tests if we had a suitable framework for them.

Yeah, there's just way too many edge cases, and I was too trusty with testing the basics. I'll create a PR with a test suite testing most of the ones you identified here, and some extra ones.

I'll be moving this to draft in the meanwhile so it's not noisy.

I'll start with https://github.com/ZcashFoundation/zebra/issues/9037

book/src/user/docker.md
gustavovalverde: The user should mount their file and set `ZEBRA_CONF_PATH` if they want to use their own custom configuration, otherwise the default config (`default-zebra-config.toml`) will be used if it's mounted. That's now the default approach in the `docker-compose.yml`.

Certainly, the user can set `ZEBRA_CONF_PATH` and mount their file to the default location`${HOME}/.config/zebrad.toml`, but that case would fall into the _user-provided_ condition.

Recent change to make this clearer: [`be37ecb` (#9344)](https://github.com/ZcashFoundation/zebra/pull/9344/commits/be37ecb853637ec7494d9202060a50d112383ba3)

docker/entrypoint.sh
gustavovalverde: https://github.com/ZcashFoundation/zebra/pull/9344#discussion_r2033143765

docker/entrypoint.sh
gustavovalverde: https://github.com/ZcashFoundation/zebra/pull/9344#discussion_r2033143765

docker/entrypoint.sh
gustavovalverde: https://github.com/ZcashFoundation/zebra/pull/9344#discussion_r2033143765

docker/.env
gustavovalverde: Fixed with [`9e336f6` (#9344)](https://github.com/ZcashFoundation/zebra/pull/9344/commits/9e336f65d5647c8d0fb6901d6df1e5fc697275c2)

docker/.env
gustavovalverde: Fixed with [`9e336f6` (#9344)](https://github.com/ZcashFoundation/zebra/pull/9344/commits/9e336f65d5647c8d0fb6901d6df1e5fc697275c2)

gustavovalverde:
I created https://github.com/ZcashFoundation/zebra/pull/9388 to be able to test different scenarios from these changes, and made the fixes here to correctly support them.

This should be ready now @upbqdn

docker/docker-compose.yml
```diff
@@ -15,13 +15,12 @@ services:
...
     tty: true
-    #! Uncomment the `configs` mapping below to make your custom configuration
-    #! take effect.
-    #
-    # configs:
-    #   - source: zebra-config
-    #     target: /home/zebra/.config/zebrad.toml
-    #
+    # ! Comment out the `configs` mapping below to not use our default configuration file.
+    # ! This will use the environment variables in the `.env` file.
```
upbqdn: From the user's perspective, it's not clear to me what _"This"_ refers to. Is it the comment above or the configs below?

gustavovalverde: Fixed here: [chore: better comment for user understanding](https://github.com/ZcashFoundation/zebra/pull/9344/commits/63f4c6795568f676c788fb785283a6f11ad11af8)

gustavovalverde:
@upbqdn I actually didn't realize I left #9388 targeting this branch. I did it just to be able to rebase this one and test the #9388. I'm planning to revert the merge, and reopen the PR, so we can merge them individually.

This is because we're going to squash the commits, and then the title and PR content will be confusing, and unrelated to most of the changes.

gustavovalverde:
@upbqdn This PR is clean now, and ready for review/approval

docker/entrypoint.sh
```diff
@@ -255,12 +258,39 @@ run_tests() {
...
+    echo "Preparing a default one..."
+    ZEBRA_CONF_PATH="${HOME}/.config/zebrad.toml"
+    mkdir -p "$(dirname "${ZEBRA_CONF_PATH}")"
+    prepare_conf_file
```
upbqdn: ###### non-blocking

This will create the config file and its parent dir owned by root:

```bash
root@1cbf7683b54d:~/.config# pwd
/home/zebra/.config
root@1cbf7683b54d:~/.config# ls -lah
total 4.0K
drwxr-xr-x 1 root  root   34 Apr 14 14:10 .
drwx------ 1 zebra zebra  26 Apr 14 14:10 ..
drwxr-xr-x 1 zebra zebra   0 Apr 14 14:10 cookie
-rw-r--r-- 1 root  root  149 Apr 14 14:10 zebrad.toml
```

docker/entrypoint.sh
```diff
@@ -6,113 +6,116 @@
...
+# ## Parameters
+#
+# - $1: Directory path to create and own
+create_owned_directory() {
+  local dir="$1"
+  [[ -z ${dir} ]] && return
 
-  echo "${ZEBRA_CONF_PATH}"
+  mkdir -p "${dir}" || exit_error "Failed to create directory: ${dir}"
+  chown -R "${UID}:${GID}" "${dir}" || exit_error "Failed to secure directory: ${dir}"
```
upbqdn: ###### non-blocking

This will create parent dirs such as `.cache` owned by root:

```bash
root@1cbf7683b54d:~# pwd
/home/zebra
root@1cbf7683b54d:~# ls -lah
total 12K
drwx------ 1 zebra zebra   26 Apr 14 14:10 .
drwxr-xr-x 1 root  root    10 Apr  9 13:03 ..
-rw-r--r-- 1 zebra zebra  220 Apr  9 13:03 .bash_logout
-rw-r--r-- 1 zebra zebra 3.5K Apr  9 13:03 .bashrc
drwxr-xr-x 1 root  root    16 Apr 14 14:10 .cache
drwxr-xr-x 1 root  root    34 Apr 14 14:10 .config
-rw-r--r-- 1 zebra zebra  807 Apr  9 13:03 .profile
```

book/src/user/docker.md
```diff
@@ -59,19 +59,22 @@ All available Cargo features are listed at
...
-the `docker/.env` file. Note that the config options of this file are limited to
-the variables already present in the commented out blocks in it and adding new
-ones will not be effective. Also note that the values of the variables take
-precedence over the values set in the `docker/default-zebra-config.toml` config
-file. The `docker/.env` file serves as a quick way to override the most commonly
-used settings for Zebra, whereas the `docker/default-zebra-config.toml` file
-provides full config capabilities.
+To configure Zebra, you have three options:
+
+1. Edit the `docker/default-zebra-config.toml` config file and uncomment the `configs` mapping in `docker/docker-compose.yml` so your config takes effect.
```
upbqdn: This documentation is outdated.

book/src/user/docker.md
```diff
@@ -59,19 +59,22 @@ All available Cargo features are listed at
...
+2. Provide a path to your own config file via the `ZEBRA_CONF_PATH` environment variable.
+3. Let Zebra generate a default config file automatically from environment variables (if neither of the above methods is used).
```
upbqdn: Zebra doesn't generate the config. We'd use `zebrad generate` for that (and it wouldn't consider env vars).

docker/entrypoint.sh
gustavovalverde: Fixed here: [fix(entrypoint): set permissions correctly for parent directories](https://github.com/ZcashFoundation/zebra/pull/9344/commits/e45c951b15993c83b1c73654779c83aa1292fa31)

docker/entrypoint.sh
gustavovalverde: Fixed here: [fix(entrypoint): set permissions correctly for parent directories](https://github.com/ZcashFoundation/zebra/pull/9344/commits/e45c951b15993c83b1c73654779c83aa1292fa31)

book/src/user/docker.md
gustavovalverde: Fixed here: [chore: align documentation and implementation](https://github.com/ZcashFoundation/zebra/pull/9344/commits/2520086bd2effaefe7ee2ddf7caa684d3d897904)

book/src/user/docker.md
gustavovalverde: Fixed here: [chore: align documentation and implementation](https://github.com/ZcashFoundation/zebra/pull/9344/commits/2520086bd2effaefe7ee2ddf7caa684d3d897904)

book/src/user/docker.md
```diff
@@ -59,49 +59,47 @@ All available Cargo features are listed at
...
-the variables already present in the commented out blocks in it and adding new
-ones will not be effective. Also note that the values of the variables take
-precedence over the values set in the `docker/default-zebra-config.toml` config
-file. The `docker/.env` file serves as a quick way to override the most commonly
-used settings for Zebra, whereas the `docker/default-zebra-config.toml` file
-provides full config capabilities.
+To configure Zebra using Docker, you have a few options, processed in this order:
+
+1. **Provide a specific config file path:** Set the `ZEBRA_CONF_PATH` environment variable to point to your config file within the container.
+2. **Use the default config file:** If `ZEBRA_CONF_PATH` is not set, Zebra will look for a config file at `/home/zebra/.config/zebrad.toml`. You can mount your custom config file to this location, for example, by uncommenting the `configs` mapping in `docker/docker-compose.yml`.
```
upbqdn: The reason I mentioned the documentation was outdated was that this PR uncomments the `configs` mapping in `docker/docker-compose.yml`, but the documentation was still saying that users should uncomment the mapping. This update says it again.

docker/entrypoint.sh
upbqdn: This solution fixes the problem only for immediate parent dirs instead of the full path.

This is not a blocking bug, but it might bite us in the future.

upbqdn: Note that the config file is also owned by root.

book/src/user/docker.md
gustavovalverde: Fixed here: [chore: clearer and aligned comments with docker-compose](https://github.com/ZcashFoundation/zebra/pull/9344/commits/c5610648a150edb74a5b10309df978d69ca23dfb)

upbqdn:
We should update the cookie path here as well: https://github.com/zcashfoundation/zebra/blob/c5610648a150edb74a5b10309df978d69ca23dfb/docker/default-zebra-config.toml#L26.

gustavovalverde:
> We should update the cookie path here as well: [zcashfoundation/zebra@`c561064`/docker/default-zebra-config.toml#L26](https://github.com/zcashfoundation/zebra/blob/c5610648a150edb74a5b10309df978d69ca23dfb/docker/default-zebra-config.toml#L26).

This seems correct: `cookie_dir = "/home/zebra/.cache/zebra"`

All other locations and docs are pointing to this same dir; except for the commented out $ZEBRA_COOKIE_DIR here https://github.com/ZcashFoundation/zebra/pull/9344/files#diff-8f78f3866f02305f1df51579c939174b2c53e8295ad13bd9d2ab763176a792c1R24 which refers to a custom location, as the comment above also implies. 

There are comments referring to the cookie file itself, and those are pointing to `/home/zebra/.cache/zebra/.cookie`

docker/entrypoint.sh
gustavovalverde: I changed the specific use case when a file is generated by the entrypoint and not mounted, as trying to own a mounted file has caused issues previously: [fix(entrypoint): chown the zebra config file when](https://github.com/ZcashFoundation/zebra/pull/9344/commits/94c5a3e6a5c97fe3bcc0f684bef3a821f7cdfbf7)

docker/default-zebra-config.toml
upbqdn: We reverted this change.

upbqdn:
utACK

One simplification I thought of was if we could omit the first config option, and only go with the static location at `/home/zebra/.config/zebrad.toml` we already use in the container. The script would then check if there's a file at that location, and if not, it would generate one. Omitting custom config locations would also simplify the docs quite a bit.

gustavovalverde:
> Omitting custom config locations would also simplify the docs quite a bit.

I agree with this, but there are edge-cases we have to consider and we might break those https://github.com/ZcashFoundation/zebra/issues/3432

Unfortunately, this depend on the host and the security policies set on it for docker.

upbqdn (approved):

PR #9382, opened 2025-04-04
Deduplicate dependencies

str4d:
## Motivation

Zallet will soon depend on a bunch of Zebra code due to Zaino. As such, Zebra dependencies now affect Zallet, and in particular duplicate dependencies like `secp256k1` that are backed by C libraries can cause issues on some platforms.

## Solution

Deduplicates several dependencies, that were missed when upgrading other Zcash crates.

### PR Checklist

- [ ] The solution is tested.


oxarbitrage (approved):
looks good, thank you @str4d

PR #9397, opened 2025-04-08
fix(warning): could not canonicalize startup warning

oxarbitrage:
## Motivation

There is an annoying warning when zebrad starts that i want to fix since a while. There was a comment about it in the discord so pushed a fix. https://discordapp.com/channels/809218587167293450/809251029673312267/1359249649155833878

## Solution

Exit early from the `try_reusing_previous_db_after_major_upgrade` function if `old_path` don't exist.

### Tests

Manual test, warning is gone.

### PR Checklist

- [x] The solution is tested.


upbqdn (approved):

gustavovalverde (approved):

mpguerra:
@mergifyio requeue

PR #9401, opened 2025-04-09
Remove the `getblocktemplate-rpcs` Cargo feature

### LINKED ISSUE #9396: The `getblocktemplate-rpcs` feature is redundant
The `getblocktemplate-rpcs` Cargo feature has been part of the default features for quite some time, and given we never turn it off, it now only clutters the codebase with code like https://github.com/ZcashFoundation/zebra/blob/470b877fcc15e0558ed5a8370068484f76f090ad/zebrad/src/components/mempool.rs#L716-L724

Removing the feature will not affect users.

upbqdn:
## Motivation

- Close #9396.

## Solution

- Remove the `getblocktemplate-rpcs` Cargo feature

### Tests

- No new tests. 
- Existing tests should cover the solution.

### Follow-up Work

- Merge the GBT RPC server into the default one.

### PR Checklist

- [x] The solution is tested.


zebrad/Cargo.toml
```diff
@@ -64,22 +63,12 @@ default-release-binaries = ["default", "sentry"]
...
 
-# Mining RPC support
-getblocktemplate-rpcs = [
-    "zebra-rpc/getblocktemplate-rpcs",
-    "zebra-consensus/getblocktemplate-rpcs",
-    "zebra-state/getblocktemplate-rpcs",
-    "zebra-node-services/getblocktemplate-rpcs",
-    "zebra-chain/getblocktemplate-rpcs",
-]
-
```
conradoplg: I think it might be better to leave the feature (i.e. just leave a `getblocktemplate-rpcs = []` here) so that existing deployment code doesn't break (e.g. anyone who has `cargo build --features getblocktemplate-rpcs` hardcoded somewhere)

(and also document that it's an empty, no longer used feature)

We could do the same for the other crates, but I think that they don't have many users as zebra itself and we can get away with just removing the feature in them. Up to you.

conradoplg (commented):
Just a comment for now, will take a closer look later

zebrad/Cargo.toml
upbqdn: I'm happy to make this change, but will it not create a precedent where we will have to keep every feature we ever use forever? Simply nuking it and announcing it in the changelog seems better than shadowing it.

conradoplg: We can get rid of it when we make a 3.0.0 release, after we batch some more breaking changes to do together. (we should list them all in an issue so we don't forget...) I think we should try to minimize breakage for zebra users when possible.

conradoplg (dismissed):
Looks good, up to you if you want to make that change or not

zebrad/Cargo.toml
upbqdn: Done in [837a022](https://github.com/ZcashFoundation/zebra/pull/9401/commits/837a022fdc1472061ba41736621d08609bc1b0d9).

conradoplg (approved):

PR #9407, opened 2025-04-09
suggestion for #9329: change coinbase field and add valueSat field to match zcashd

### LINKED ISSUE #9297: feature: Add additional fields to `getrawtransaction` verbose response
### Motivation

Zechub requires the following fields:

- `vin`
- `vout`
- `vShieldedSpend`
- `vShieldedOutput`
- `orchard.actions`
- `valueBalance`

See https://forum.zcashcommunity.com/t/zf-engineering-update-2025-sprint-3-feb-3-feb-14/50392/2

### Specifications

_No response_

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

#9329 looks good but it shows two specific discrepancies between zcashd output: the `coinbase` field does not match (zcashd uses the whole coinbase script, while zebrad returned just the extra data after the encoded height) and the `valueSat` field was missing (which is the same as `valueZat` and seems useless, but it's easy to add so why not)

### Tests

Tested with zcash-rpc-diff:

`OUTPUT_DATA_LINE_LIMIT=100000 ZEBRAD_EXTRA_ARGS=-conf=$PWD/zcash.conf ./zebra-utils/zcash-rpc-diff 28232 getrawtransaction 21f2c7d10cc8857e60fd225f0f089126d9274e892f8d4e135e9e915adb06b72d 1`

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods/types/transaction.rs
+++ b/zebra-rpc/src/methods/types/transaction.rs
@@ -9,7 +9,7 @@ use zebra_chain::{
     parameters::Network,
     sapling::NotSmallOrderValueCommitment,
     transaction::{SerializedTransaction, Transaction},
-    transparent::{CoinbaseData, Script},
+    transparent::Script,
 };
 use zebra_consensus::groth16::Description;
 use zebra_state::IntoDisk;
@@ -70,7 +70,7 @@ pub enum Input {
     Coinbase {
         /// The coinbase scriptSig as hex.
         #[serde(with = "hex")]
-        coinbase: CoinbaseData,
+        coinbase: Vec<u8>,
         /// The script sequence number.
         sequence: u32,
     },
@@ -104,6 +104,9 @@ pub struct Output {
     /// The value in zats.
     #[serde(rename = "valueZat")]
     value_zat: i64,
+    /// The value in zats (again with a different name for some reason).
+    #[serde(rename = "valueSat")]
+    value_sat: i64,
     /// index.
     n: u32,
     /// The scriptPubKey.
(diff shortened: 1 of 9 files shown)


zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -104,6 +104,9 @@ pub struct Output {
     /// The value in zats.
     #[serde(rename = "valueZat")]
     value_zat: i64,
+    /// The value in zats (again with a different name for some reason).
```
oxarbitrage: I left this one out because it looks like an error in zcashd. Seems like they added one variant and then needed the other one and didn't wanted to break the API. i think we can take the opportunity to remove one of them here. 

I added this item for the zcashd deprecation meeting of today.

oxarbitrage (commented):
Thanks, the coinbase script fix is something i was not able to figure out myself. 

We need to fix the snapshots either here or in the base branch after this gets merged. However, let's wait for https://github.com/ZcashFoundation/zebra/pull/9407/files#r2037071022 before doing that.

zebra-rpc/src/methods/types/transaction.rs
oxarbitrage: From the meeting today, `valueSat` is going to be deprecated in zcashd 6.2.0 or in a future release, so we better just don't add it to Zebra.

conradoplg: Done in f78a6cc6b

oxarbitrage (approved):
thanks!

PR #9410, opened 2025-04-10
fix(docker): Correct cargo dependency caching in build stages

gustavovalverde:
## Motivation

Caching was not being used correctly in our Dockerfile

Depends-On: #9344

## Solution

Previously, the Dockerfile used cache mounts targeting specific subdirectories (`/usr/local/cargo/registry/` and `/usr/local/cargo/git/db`) which did not align with the `CARGO_HOME` environment variable (`/home/zebra/.cargo`). This prevented Cargo from utilizing the cache for downloaded dependencies and the registry index, causing them to be re-downloaded on subsequent builds.

This change replaces the specific subdirectory mounts with a single cache mount targeting the `${CARGO_HOME}` directory directly in both the `tests` and `release` stages. This ensures that BuildKit's cache is correctly utilized by Cargo, significantly speeding up rebuilds by avoiding unnecessary downloads.

### Tests

This should be faster than our actual rebuilds, and also than #9328

- https://docs.docker.com/reference/dockerfile/#benefits-of-using---link

### PR Checklist

- [X] The solution is tested.


gustavovalverde:
I found this issue while testing 
- https://github.com/orgs/community/discussions/25678
- https://github.com/actions/runner-images/issues/10386

We might need to do further improvements to actually accomplish what we'd like here.

PR #9426, opened 2025-04-15
refactor(chain): replace sinsemilla with external library

### LINKED ISSUE #7801: Re-use ECC `sinsemilla` code
I noticed we have `sinsemilla` hash function code in Zebra: https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/orchard/sinsemilla.rs and just 1 call to `sinsemilla_hash()` from one place (https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/orchard/tree.rs#L71).

I was wondering if we should move the sinsemilla code into its own crate/project. I noticed a placeholder crate exist (https://crates.io/crates/sinsemilla) with @str4d as the owner. We can maybe ask permissions, push the code there unless there are other plans for the placeholder.

What other people think?

natalieesk:
## Motivation

Closes #7801 

## Solution

`sinsemilla_hash()` now uses the [sinsemilla library](https://docs.rs/sinsemilla/0.1.0/sinsemilla/index.html)

### Follow-up Work

The Sinsemilla library does not expose some of the functions needed to remove all of our code. So we've had to keep `extract_p()` and `pallas_group_hash()`. We might want to talk to the implementers to see if they'll expose them but it's not in the scope of this story.

### PR Checklist

- [x] The solution is tested.


zebra-chain/src/orchard/sinsemilla.rs
```diff
@@ -50,113 +40,24 @@ pub fn pallas_group_hash(D: &[u8], M: &[u8]) -> pallas::Point {
...
 ///
 /// <https://zips.z.cash/protocol/nu5.pdf#concretesinsemillahash>
-///
-/// # Panics
-///
-/// If `M` is greater than `k*c = 2530` bits in `sinsemilla_hash_to_point`.
 #[allow(non_snake_case)]
 pub fn sinsemilla_hash(D: &[u8], M: &BitVec<u8, Lsb0>) -> Option<pallas::Base> {
-    extract_p_bottom(sinsemilla_hash_to_point(D, M))
+    let domain = std::str::from_utf8(D).unwrap();
```
conradoplg: Just for documentation purposes; also would be good to include this in the comment above (e.g.

```
# Panics

If `D` is not UTF-8 encoded
```

```suggestion
    let domain = std::str::from_utf8(D).expect("must be valid UTF-8");
```

conradoplg (dismissed):
Looks good, left an optional suggestion

conradoplg (approved):

daira:
> ### Follow-up Work
> 
> The Sinsemilla library does not expose some of the functions needed to remove all of our code. So we've had to keep `extract_p()` and `pallas_group_hash()`. We might want to talk to the implementers to see if they'll expose them but it's not in the scope of this story.

Filed https://github.com/zcash/sinsemilla/issues/3 .

PR #9431, opened 2025-04-16
fix(publish)!: simplify Docker tagging to remove redundant tags

### LINKED ISSUE #7415: devops: Remove redundant tags in Docker Hub

gustavovalverde:
## Motivation

This PR addresses the issue of redundant tags on Docker Hub for Zebra images, ensuring that only necessary tags are generated and that the `latest` tag points to the production build, not experimental ones. 

Fixes #7415

## Solution

- Removed `latest_tag` and `tag_suffix` inputs from the Docker build workflow to prevent custom tag variations.
- Removed `flavor` customization for suffixes and forced `latest` tags to simplify tagging.
- Split SHA tag generation into separate rules for PR (`type=sha,event=pr`) and branch (`type=sha,event=branch`) events for clarity and precision.
- Removed `type=ref,event=tag` to avoid unnecessary tag event tags.
- Updated comments in the workflow to better reflect tag usage intentions.

### Tests
- Existing CI pipelines will validate that images are still built and pushed correctly to Google Artifact Registry and Docker Hub 
- Manual verification will be required in the next release

- https://github.com/docker/metadata-action/blob/v5.7.0/README.md
- https://github.com/docker/metadata-action/blob/master/src/meta.ts

### Follow-up Work

- Monitor Docker Hub tags post-release to confirm no unintended tags are published in future releases.
- Potential follow-up to address any user feedback if the removal of custom tag inputs breaks specific workflows.

### PR Checklist

- [x] The solution is tested.
- [ ] If the PR shouldn't be in the release notes, it has the `C-exclude-from-changelog` label.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -23,7 +23,6 @@ jobs:
       dockerfile_path: ./docker/Dockerfile
       dockerfile_target: runtime
       image_name: zebra
-      latest_tag: true
       features: ${{ vars.RUST_PROD_FEATURES }}
       rust_log: ${{ vars.RUST_LOG }}
     # This step needs access to Docker Hub secrets to run successfully
--- a/.github/workflows/sub-build-docker-image.yml
+++ b/.github/workflows/sub-build-docker-image.yml
@@ -35,13 +35,6 @@ on:
       features:
         required: false
         type: string
-      latest_tag:
-        required: false
-        type: boolean
-        default: false
-      tag_suffix:
-        required: false
-        type: string
       no_cache:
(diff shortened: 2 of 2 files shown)


upbqdn (approved):
utACK

Issue #9427, opened 2025-04-15
bug: possibly change "lockbox" vs "deferred" value pool name


conradoplg:
### What happened?

In `getblockchaininfo` (and soon in `getblock`), the lockbox value pool is named "deferred" in Zebra but "lockbox" in zcashd.

We should decide between:

- Keep as "deferred", with the risk of breaking zcashd users upon zcashd deprecation
- Change it to "lockbox", with the risk of breaking zebrad users
- Add both for compatibility, with the small risk of some code out there working on the assumption that every item returned is for a different pool

### Which operating systems does the issue happen on?

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

dismad:
I vote change to lockbox.

mpguerra:
Whatever we do, let's do it sooner rather than later before we have too many people relying on zebra. We should also double check whether @hhanh00 relies on this field

pacu:
I think the naming is negligible given that not many services rely on it. What's the correct term according to the ZIP? 

The important thing is how long this value will live. Will the new governance eliminate the concept of the lockbox? What will happen when the one-time disbursement is done? Will this value be returned with zero value on it or not returned at all?

For example, in this patch for the ZBE, it is assumed that the value will be returned, but now that I think of it this probably isn't true. https://github.com/matiasurbieta/zcash-explorer/pull/1/files

pacu:
@daira commented on Signal:

> The RPC field for the lockbox value is implemented for zcashd here, in the PR that is linked as the reference implementation in the ZIP: https://github.com/zcash/zcash/pull/6912/files#diff-decae4be02fb8a47ab4557fe74a9cb853bdfa3ec0fa1b515c0a1e5de91f4ad0bR276
> So zebra should have matched that

What's the best way to go towards this? Is just renaming? Deprecating "deferred" and adding "lockbox"?

The RPC side could use a refactor though. I'll create a draft PR and we can start discussing it there.

Issue #7939, opened 2023-11-13
perf: Stop expensive cryptographic validation when deserializing shielded transactions


teor2345:
### Motivation

Zebra's transaction deserialization does a lot of CPU-heavy cryptography for shielded transactions.

Since we know that transactions in our database are valid, we can deserialize them without checking the validity of curve points.

### API Reference

TODO: which methods are causing this CPU usage, and what can they be replaced with?

### Detailed Analysis

See https://github.com/ZcashFoundation/zebra/issues/4583#issuecomment-1186662862

### Designs

In `zebra-chain`:
- split out transaction deserialisation methods that optionally do validation
- do validation by default in ZcashDeserialize (with transactions and blocks from the network and RPCs)

In `zebra-state`:
- skip validation for transactions and blocks when reading from the state in FromBytes

## Related Work

- #4788 (which didn't end up improving performance)

mpguerra:
Is this still relevant outside of the `zebra-scan` crate?

conradoplg:
Copying this here for reference:

Parsing blocks is _definitely_ much more expensive than loading blocks. Benchmark that loads 10k blocks starting from 1723000 (arbitrary height I picked at the middle of the spam):

- just load from disk, but do not parse: 16.9s
- load from disk and parse: 455.8s (~27 times slower)

I ran sudo sh -c "/usr/bin/echo 3 > /proc/sys/vm/drop_caches" before benchmarking to clear the file cache

conradoplg:
Some design ideas:

- Duplicate the entire Transaction tree (e.g. to `PartiallyDeserializedTransaction`) where the relevant leaf types are replace with `~Bytes` types, e.g. `ValueCommitment` -> `ValueCommitmentBytes`
   - Probably the simplest, does not affeact the rest of the code, but introduces a lot of duplicated code
-  Change the relevant leaf types to `~Bytes` types
   - Will break a lot of code which will need to be adjusted to convert from the `~Bytes` type to the original type where required. Requires a lot of code changes, and carries the risk of doing those conversions multiple times, decreasing performance in some cases
- Use generic magic to allow the specification of which leaf types to use (the regular or the `~Bytes` ones), and default to the regular ones so existing code doesn't break (much)
  - Equivalent to the first option, but without the duplicated code
  - Haven't tried, no idea if it would work
[...]

PR #9432, opened 2025-04-16
feat(rpc): add value pools to getblock

### LINKED ISSUE #5449: getblock: return the `valuesPools` and `chainSupply` fields

conradoplg:
## Motivation

We want to fill the `valuePools` and `chainSupply` fields for the `getblock` RPC.

## Solution

_The database part was included in #9539, I'll leave its description below for the record._

That requires keeping track of the value pools after each block; we only keep it for the entire chain.

This PR introduces a new column family named `block_data`. The idea behind the generic name is that we could add other data there like chainwork and block size. In fact, the main reason this is a draft PR is because we might want to add those in the same PR, but I'd like to get feedback on this first.

The database upgrade is slow ([main reason](https://github.com/ZcashFoundation/zebra/issues/7939)) and takes ~2h to complete (it was 4h, I managed to improve a bit with rayon). I think this is harmless since it only impacts the RPC itself (it does not return the field if the data is not yet available).

To avoid a circular dependency, this also moves some subsidy-related functions from `zebra-consensus` to `zebra-chain`. I'm not sure if I should go ahead and move everything else related to that. Let me know if you prefer a separate PR with the move, but I also wanted feedback on this first.

Closes #5449 

### Tests

I adjusted the existing tests and used `zcash-rpc-diff` to compare with zcashd, and everything seems to match (with [this exception](https://github.com/ZcashFoundation/zebra/issues/9427))

https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L766-L782

### PR Checklist

- [ ] The solution is tested.


zebra-chain/src/parameters/network/subsidy.rs
conradoplg, line 1: Just moved code.

zebra-chain/src/parameters/network/tests.rs
conradoplg, line 1: Just moved code.

zebra-consensus/src/block/subsidy.rs
conradoplg, line 1: Just moved code.

zebra-consensus/src/block/subsidy/general.rs
conradoplg, line 1: Just moved code.

zebra-state/src/request.rs
```diff
@@ -918,9 +918,15 @@ pub enum ReadRequest {
...
     /// Returns [`ReadResponse::TipPoolValues(Option<(Height, block::Hash, ValueBalance)>)`](ReadResponse::TipPoolValues)
-    /// with the current best chain tip.
+    /// with the pool values of the current best chain tip.
     TipPoolValues,
 
+    /// Looks up the block data after a block by hash or height in the current best chain.
+    ///
+    /// * [`ReadResponse::BlockData(Some(pool_values))`](ReadResponse::BlockData) if the block is in the best chain;
+    /// * [`ReadResponse::BlockData(None)`](ReadResponse::BlockData) otherwise.
+    BlockData(HashOrHeight),
```
arya2: Could we rename this to `BlockInfo`?

zebra-state/src/service/finalized_state/disk_format/chain.rs
```diff
@@ -93,3 +94,26 @@ impl FromDisk for HistoryTreeParts {
             .expect("deserialization format should match the serialization format used by IntoDisk")
     }
 }
+
+impl IntoDisk for BlockData {
+    type Bytes = Vec<u8>;
+
+    fn as_bytes(&self) -> Self::Bytes {
+        self.value_pools().as_bytes().to_vec()
```
arya2: Nitpick: It'd be nice to use the exact size if we know it in advance and move it to constants.

```suggestion
    type Bytes = [u8; 40];

    fn as_bytes(&self) -> Self::Bytes {
        self.value_pools().as_bytes()
```

zebra-state/src/service/finalized_state/disk_format/upgrade/block_data.rs
arya2, line 1: It would be nice to do this in parallel if it takes 4 hours.

arya2 (commented):
This is looking good.

Adding the chain work and size fields in the same column family seems like a good idea to me.

zebra-state/src/service/finalized_state/disk_format/chain.rs
conradoplg: If we use `Vec` then this code will be forward-compatible - we can append more data without requiring a full database upgrade. Not sure if we _want_ that though :sweat_smile:

zebra-state/src/request.rs
conradoplg: Good idea.

zebra-state/src/service/finalized_state/disk_format/upgrade/block_data.rs
conradoplg: It's a bit tricky since we'd need to parallelize just the block loading, but the rest needs to be done sequentially since the value pool of a block depends on the previous block. I also have no idea how the database handles parallel reads?

zebra-state/src/service/finalized_state/zebra_db/block.rs
```diff
@@ -638,15 +638,14 @@ impl DiskWriteBatch {
...
-                value_pool,
-            )?;
         }
+        // Commit UTXOs and value pools
+        self.prepare_chain_value_pools_batch(
+            zebra_db,
+            finalized,
+            spent_utxos_by_outpoint,
+            value_pool,
+        )?;
```
conradoplg: This is a bit of subtle change. Per the comment above (expand to see), we should ignore the genesis block. But for the value pools it's harmless, because the genesis block adds 0 balance and it does not hurt storing it for block 0; in fact, doing so makes some stuff easier to handle later on (such as tests, where we don't need to special-case block 0). I might have missed something though, so please double check my reasoning, but it seems OK.

zebra-utils/zcash-rpc-diff
```diff
@@ -112,11 +112,11 @@ echo "$@"
 echo
 
 echo "Querying $ZEBRAD $ZEBRAD_NET chain at height >=$ZEBRAD_HEIGHT..."
-time $ZCASH_CLI $ZEBRAD_EXTRA_ARGS -rpcport="$ZEBRAD_RPC_PORT" "$@" > "$ZEBRAD_RESPONSE"
+time $ZCASH_CLI $ZEBRAD_EXTRA_ARGS -rpcport="$ZEBRAD_RPC_PORT" "$@" | jq -S > "$ZEBRAD_RESPONSE"
 echo
 
 echo "Querying $ZCASHD $ZCASHD_NET chain at height >=$ZCASHD_HEIGHT..."
-time $ZCASH_CLI $ZCASHD_EXTRA_ARGS "$@" > "$ZCASHD_RESPONSE"
+time $ZCASH_CLI $ZCASHD_EXTRA_ARGS "$@" | jq -S > "$ZCASHD_RESPONSE"
```
conradoplg, line 119: Sorting the JSONs makes it much easier to compare. I sneaked this in because I got tired of accidentally committing this local change :rofl: I think it should not have bad side effects. (I was afraid of `jq` possibly hiding some issue with the raw JSON, but that seems unlikely)

zebrad/tests/acceptance.rs
```diff
@@ -1938,28 +1938,30 @@ fn lightwalletd_integration_test(test_type: TestType) -> Result<()> {
...
+                )],
+            )?;
+        } else {
+            wait_for_state_version_upgrade(
+                &mut zebrad,
+                &state_version_message,
+                state_database_format_version_in_code(),
+                None,
+            )?;
+        }
```
conradoplg, line 1962: This fixes what seems to be a bug; if the database upgrade is long but the test does not launches lightwalletd, then we still shouldn't wait for the state version upgrade.

zebrad/tests/common/cached_state.rs
```diff
@@ -48,7 +48,7 @@ pub const DATABASE_FORMAT_CHECK_INTERVAL: Duration = Duration::from_secs(5 * 60)
 ///
 /// If this is set to `true`, but the state upgrades finish before zebrad is synced,
 /// some tests will hang.
-pub const DATABASE_FORMAT_UPGRADE_IS_LONG: bool = false;
+pub const DATABASE_FORMAT_UPGRADE_IS_LONG: bool = true;
```
conradoplg, line 51: We should revert this after this PR is merged and cached states are updated.

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -134,7 +134,7 @@ jobs:
     runs-on: zfnd-runners
     needs: [ get-disk-name ]
     if: ${{ !cancelled() && !failure() }}
-    timeout-minutes: ${{ inputs.is_long_test && 7200 || 180 }}
+    timeout-minutes: ${{ inputs.is_long_test && 7200 || 600 }}
```
conradoplg: We should rever this after the PR merges; with the parallelization it should be OK but I didn't want to possibly waste yet another CI run if it does timeout

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -488,20 +490,22 @@ jobs:
...
-      app_name: zebra-scan
-      test_id: scanner-tests
-      test_description: Tests the scanner.
-      test_variables: "-e NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }} -e TEST_SCANNER=1 -e ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
-      needs_zebra_state: true
-      needs_lwd_state: false
-      saves_to_disk: false
-    secrets: inherit
+  # TODO: commented out because it is failing with the value pools update.
+  # TODO: Remove if we decide to remove scanner code.
```
arya2: [This test](https://github.com/ZcashFoundation/zebra/actions/runs/15036221072/job/42258703583?pr=9295) and [other long-running tests](https://github.com/ZcashFoundation/zebra/actions/runs/15036221072/job/42258678925?pr=9295) may pass without extra time when [parsing transactions in parallel](https://github.com/ZcashFoundation/zebra/pull/9295/commits/63f05af03593189ba5699b8a4272e9cdc019e86f#diff-68bb523da9a0d8b35523efd317ffd74eb0358e4f304b1b5c7ee7472a252d4fa7R68-R77).

dismad:
Is there a way to show the status of the valuePool sync? That might be useful to add if it doesn't complicate things. 

Screenshot_2025-05-15_14-57-18

I'm thinking in `getinfo` or `getblockchaininfo` a `isValuePoolsSynced' boolean, or even better, a % complete

.github/workflows/sub-ci-integration-tests-gcp.yml
conradoplg: Done in 27e1d9d042192750f73a4936eb5d79dd2845fd5c

conradoplg:
> Is there a way to show the status of the valuePool sync? That might be useful to add if it doesn't complicate things.
> 
> Screenshot_2025-05-15_14-57-18
> 
> I'm thinking in `getinfo` or `getblockchaininfo` a `isValuePoolsSynced' boolean, or even better, a % complete

I think it would be work that will be rarely used, since after the database is upgraded you'll never have to worry about it again. You can somewhat check the progress by calling `getblock` for some height and checking if it returns the pools

zebrad/tests/acceptance.rs
arya2: We shouldn't need to wait for the format upgrade before syncing lwd in the future as there are no plans to extend lightwalletd. It must've been added so the test would wait for the format upgrade that added subtrees and with the expectation that there would be more format upgrades like that one.

Could we always wait for the `OPENED_RPC_ENDPOINT_MSG` log here if the test launches lightwalletd?

conradoplg:
Moved to draft because @arya2 will try combining this and #9295

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -468,12 +472,7 @@ impl DbFormatChange {
         //   (unless a future upgrade breaks these format checks)
         // - re-opening the current version should be valid, regardless of whether the upgrade
         //   or new block code created the format (or any combination).
-        Self::format_validity_checks_detailed(db, cancel_receiver)?.unwrap_or_else(|_| {
-            panic!(
-                "unexpected invalid database format: delete and re-sync the database at '{:?}'",
-                db.path()
-            )
```
arya2: Why this change? (I haven't included it in the new PR, let me know if I should).

conradoplg: oops, leftover from debugging (this panic hides the original error... I guess the right fix would be to include the error in the panic message but you don't have to do that now)

mpguerra:
@conradoplg can we close this one in favour of @arya2 's PR?

conradoplg:
> @conradoplg can we close this one in favour of @arya2 's PR?

I thin I'll repurpose it to contain the RPC call itself based on @arya2 's PR (which contains the DB changes but not the RPC change itself IIRC)

conradoplg:
@Mergifyio update

zebra-rpc/src/methods/types/get_blockchain_info.rs
```diff
@@ -22,69 +22,104 @@ pub struct Balance {
     chain_value_zat: Amount<NonNegative>,
     /// Whether the value pool balance is being monitored.
     monitored: bool,
+    /// Change to the amount in the pool produced by this block, in ZEC
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    value_delta: Option<Zec<NegativeAllowed>>,
+    /// Change to the amount in the pool produced by this block, in zatoshis
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    value_delta_zat: Option<Amount<NegativeAllowed>>,
```
arya2: Optional/Nitpick:

```suggestion
    value_delta_zat: Option<Amount>,
```

zebrad/tests/common/checkpoints.rs
```diff
@@ -95,12 +98,18 @@ pub async fn run(network: Network) -> Result<()> {
...
+        // This might not work in future database upgrades, but works for the
+        // current one (27.0.0) where this `if` was introduced.
+        if !DATABASE_FORMAT_UPGRADE_IS_SLOWER_THAN_RPC_SPAWN {
+            wait_for_state_version_upgrade(
+                &mut zebrad,
+                &state_version_message,
+                state_database_format_version_in_code(),
+                None,
+            )?;
+        }
```
arya2: Optional:

This shouldn't be necessary for this PR as the cached states should have already been upgraded to 27.0.0, so it may be nice to leave it out to avoid any chance of someone wondering why it's there later, and to apply a more general solution (e.g. using `expect_stdout_line_matches_all_unordered()` by passing `Some(OPENED_RPC_ENDPOINT_MSG)` as the last arg to this fn) before or at the next long-running db format upgrade (which hopefully shouldn't be for a while, the ZSA format upgrade should be near-instant).

conradoplg:
@arya2 review reminder :pray:

zebra-rpc/src/methods.rs
```diff
@@ -1181,6 +1182,9 @@ where
                 transactions_request,
                 // Orchard trees
                 zebra_state::ReadRequest::OrchardTree(hash_or_height),
+                // Block info
```
arya2, line 1185: Optional/Nitpick: I'm just noticing this doesn't need to be a `Vec` anymore. It doesn't make any practical difference and doesn't need to be updated in this PR.

Optional: This method could use a `Block` request instead of `BlockAndSize` when handling verbosity 2 requests now that it's getting the block size from the `BlockInfo` response

zebra-rpc/src/methods.rs
```diff
@@ -1255,6 +1259,29 @@ where
...
+
+            let delta = block_info.as_ref().and_then(|d| {
+                let value_pools = d.value_pools().constrain::<NegativeAllowed>().ok()?;
+                let prev_value_pools = prev_block_info
+                    .map(|d| d.value_pools().constrain::<NegativeAllowed>())
+                    .unwrap_or(Ok(ValueBalance::<NegativeAllowed>::zero()))
+                    .ok()?;
+                (value_pools - prev_value_pools).ok()
+            });
+            let size = size.or(block_info.as_ref().map(|d| d.size() as usize));
```
arya2, line 1283: Optional/Nitpick: 
- There doesn't seem to be any point in checking for the size from the `BlockAndSize` request anymore,
- For verbosity 2 requests, this `getblock` method could fallback on using `zcash_serialized_size()` with an `or_else()` to populate the `size` field in its output before the format upgrade is complete.

```suggestion
            let size = block_info.as_ref().map(|d| d.size() as usize);
```

arya2 (dismissed):
This looks good, I left some comments but none of them are blockers.

zebra-rpc/src/methods.rs
conradoplg: The upside to not changing this is that it keeps `size` being returned when the database upgrade is running. We could clean this up in the release after the current Zebra reaches EoS, but also not a big deal.

zebra-rpc/src/methods.rs
conradoplg: I didn't get what you mean with not needing to be a `Vec` anymore? The vector of futures?

The upside to not changing this is that it keeps `size` being returned when the database upgrade is running. We could clean this up in the release after the current Zebra reaches EoS, but also not a big deal.

zebrad/tests/acceptance.rs
conradoplg: From what I understand we always need to wait for the database upgrade because this test is used to generated cached images and we don't want to generate half-upgraded cached images. I think we probably need to clean up these tests, I can create a ticket for that.

zebrad/tests/common/checkpoints.rs
conradoplg: I reverted it in b8e2687af. We probably want to clean all of this logic up at some point, I'll create an issue for it and point to this comment.

zebra-rpc/src/methods/types/get_blockchain_info.rs
conradoplg: TIL! Done in b8e2687af

arya2 (approved):

mpguerra:
@mergifyio queue

PR #9441, opened 2025-04-17
fix(usability): Improve the cache dir and database startup panics

### LINKED ISSUE #8617: Improve error message when Zebra crashes on startup
When using a `cache_dir` that is a symlink to a storage drive, Zebra crashed. When using the actual path to the storage drive, Zebra runs normally.

## Error
```
Opening database "/home/rav/.zebra/state/v25/testnet" failed: Error { message: "Failed to create RocksDB directory: `Os { code: 13, kind: PermissionDenied, message: \"Permission denied\" }`." }. Hint: Check if another zebrad process is running. Try changing the state cache_dir in the Zebra config.
```

## Metadata
|key|value|
|--|--|
|**version**|1.7.0|
|**Zcash network**|Testnet|
|**running state version**|25.3.0|
|**initial disk state version**|creating.new.database|
|**features**|default,getblocktemplate_rpcs,howudoin,indicatif,progress_bar,release_max_level_info|
|**rust compiler**|1.79.0|
|**rust release date**|2024-06-10|
|**optimization level**|3|
|**debug checks**|false|
|**location**|/home/pitmutt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zebra-state-1.0.0-beta.37/src/service/finalized_state/disk_db.rs:860:23|

oxarbitrage:
## Motivation

Currently, Zebra will panic any time it can’t create or open its state database. For example, if the cache directory is invalid, the disk is full, permissions are wrong, or the RocksDB instance is already in use. The default panic messages are ugly and make it hard for end users to diagnose the real cause.

Close https://github.com/ZcashFoundation/zebra/issues/8617

## Solution

We now split these failures into two categories:

- **OS‑level errors** (`std::io::ErrorKind`) related to the cache directory:
   - Permission denied  
   - Disk full  
   - Etc.

- **RocksDB errors** (`rocksdb::ErrorKind`) at the database-opening stage:
   - Database already in use
   - Other RocksDB‑specific failures

In `DiskDB::new`:

- **Initialize the cache directory**  
   Attempt to create or open the cache directory first, and convert any `std::io::ErrorKind` into a descriptive panic.

- **Open the RocksDB database**  
   Once the directory is validated, build the full DB path and open RocksDB, catching any `rocksdb::ErrorKind` and panicking.

This separation makes sure that each panic points more directly to its root cause and suggests an actionable hint.

### Tests

- Updated the existing unit test to expect the new, more specific panic message.  
- Manual verification test for both an OS‑level error (e.g., read‑only directory) and a RocksDB “already in use” error.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-state/src/constants.rs
+++ b/zebra-state/src/constants.rs
@@ -118,5 +118,5 @@ pub const MAX_INVALIDATED_BLOCKS: usize = 100;
 
 lazy_static! {
     /// Regex that matches the RocksDB error when its lock file is already open.
-    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(lock file).*(temporarily unavailable)|(in use)|(being used by another process)").expect("regex is valid");
+    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(Database already open).*").expect("regex is valid");
 }
--- a/zebra-state/src/constants.rs
+++ b/zebra-state/src/constants.rs
@@ -118,5 +118,5 @@ pub const MAX_INVALIDATED_BLOCKS: usize = 100;
 
 lazy_static! {
     /// Regex that matches the RocksDB error when its lock file is already open.
-    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(Database already open).*").expect("regex is valid");
+    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(lock file).*(temporarily unavailable)|(in use)|(being used by another process)|(Database likely already open)").expect("regex is valid");
 }
--- a/zebra-state/src/service/finalized_state/disk_db.rs
+++ b/zebra-state/src/service/finalized_state/disk_db.rs
@@ -833,6 +833,11 @@ impl DiskDb {
     /// Opens or creates the database at a path based on the kind, major version and network,
     /// with the supplied column families, preserving any existing column families,
     /// and returns a shared low-level database wrapper.
+    ///
+    /// # Panics
+    ///
+    /// - If the cache directory do not exist and can't be created.
+    /// - If the database cannot be opened for whatever reason.
     pub fn new(
         config: &Config,
         db_kind: impl AsRef<str>,
@@ -841,6 +846,11 @@ impl DiskDb {
         column_families_in_code: impl IntoIterator<Item = String>,
         read_only: bool,
     ) -> DiskDb {
+        // If the database is ephemeral, we don't need to check the cache directory.
+        if !config.ephemeral {
+            DiskDb::validate_cache_dir(&config.cache_dir);
+        }
+
         let db_kind = db_kind.as_ref();
         let path = config.db_path(db_kind, format_version_in_code.major, network);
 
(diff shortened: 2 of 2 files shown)


gustavovalverde:
I did not fully validate the PR, but we should validate if the `ZEBRA_CACHE_DIR` is set, and create the directory where this variable indicates, or else to the default location.

upbqdn:
> I did not fully validate the PR, but we should validate if the `ZEBRA_CACHE_DIR` is set, and create the directory where this variable indicates, or else to the default location.

The `ZEBRA_CACHE_DIR` var takes effect only in tests, which this PR doesn't change. I think we could make a bunch of things easier if we dropped the var entirely.

zebra-state/src/constants.rs
```diff
@@ -118,5 +118,5 @@ pub const MAX_INVALIDATED_BLOCKS: usize = 100;
 
 lazy_static! {
     /// Regex that matches the RocksDB error when its lock file is already open.
-    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(lock file).*(temporarily unavailable)|(in use)|(being used by another process)").expect("regex is valid");
+    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(Database already open).*").expect("regex is valid");
```
arya2: We likely want to continue matching on the previous messages and also on this new one.

```suggestion
    pub static ref LOCK_FILE_ERROR: Regex = Regex::new("(lock file).*(temporarily unavailable)|(in use)|(being used by another process)|(Database likely already open)").expect("regex is valid");
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -833,6 +833,11 @@ impl DiskDb {
     /// Opens or creates the database at a path based on the kind, major version and network,
     /// with the supplied column families, preserving any existing column families,
     /// and returns a shared low-level database wrapper.
+    ///
+    /// # Panics
+    ///
+    /// - If the cache directory do not exist and can't be created.
```
arya2: Nitpick:

```suggestion
    /// - If the cache directory does not exist and can't be created.
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -901,12 +911,20 @@ impl DiskDb {
...
+                         Hint: Check if another zebrad process is running."
+                    )
+                } else {
+                    panic!(
+                        "Opening database {path:?} failed. \
+                        Hint: Try changing the state cache_dir in the Zebra config. \
+                        Error: {e}",
+                    )
+                }
+            }
```
arya2: Nitpick/Optional/Subjective: I think it looks a little nicer with a `matches!` expr in a match arm guard.

I think it should say "database _likely_ already open" since it's not always the cause of those errors.

```suggestion
            Err(e) if matches!(e.kind(), ErrorKind::Busy | ErrorKind::IOError) => panic!(
                "Database likely already open {path:?} \
                         Hint: Check if another zebrad process is running."
            ),

            Err(e) => panic!(
                "Opening database {path:?} failed. \
                        Hint: Try changing the state cache_dir in the Zebra config. \
                        Error: {e}",
            ),
```

arya2:
> I think we could make a bunch of things easier if we dropped the var entirely.

I think its purpose was to ensure that those tests are never run unintentionally, but I agree it would make a bunch of things easier to drop it and we may want to do that.

zebra-state/src/constants.rs
oxarbitrage: Makes sense. I added "likely" to your suggestion so we can match with https://github.com/ZcashFoundation/zebra/pull/9441/files#r2050700638

arya2 (approved):
Thank you!

upbqdn:
> > I think we could make a bunch of things easier if we dropped the var entirely.
> 
> I think its purpose was to ensure that those tests are never run unintentionally, but I agree it would make a bunch of things easier to drop it and we may want to do that.

This https://github.com/ZcashFoundation/zebra/pull/9481#issuecomment-2850731524 is a good motivation to drop the variables.

upbqdn:
Also, the description in PR #9323.

PR #9458, opened 2025-04-23
feat(rpc): Add `time` and `size` field to `TransactionObject`

### LINKED ISSUE #9452: Add `size` and `time` to `getrawtransaction` output
Requested in https://github.com/ZcashFoundation/zebra/issues/8945#issuecomment-2819181496

oxarbitrage:
## Motivation

We’ve received requests to include `size` and `time` fields in the `getrawtransaction` RPC output.

Close https://github.com/ZcashFoundation/zebra/issues/9452

## Solution

- Extend `TransactionObject` with two new fields:
  - **`size`**: the serialized transaction size (currently computed via `tx.as_bytes().len()`).
  - **`time`**: the timestamp of the block in which the transaction was mined.
- To support `time`, add a `block_time` field to `MinedTx`.
- Note: using `tx.as_bytes().len()` may not be the most efficient approach. Open to changes here.

### Tests

- Updated snapshot tests that includes the new fields.
- Manual comparisons between `zebrad` and `zcashd` outputs; results to follow in a comment.

### Follow-up Work

Decide if displaying `time` for mempool transactions is needed or leave it omitted (TODO).

### PR Checklist

- [x] The solution is tested.


zebra-rpc/src/methods.rs
```diff
@@ -1272,12 +1277,13 @@ where
...
             zebra_state::ReadResponse::Transaction(Some(tx)) => Ok(if verbose {
-                GetRawTransaction::Object(TransactionObject::from_transaction(
+                GetRawTransaction::Object(Box::new(TransactionObject::from_transaction(
                     tx.tx.clone(),
                     Some(tx.height),
                     Some(tx.confirmations),
                     &self.network,
-                ))
+                    Some(tx.block_time),
+                )))
```
conradoplg, line 1287: This is optional, but we could get a small performance gain here by reverting `ZebraDB::transaction()` to make it not load the block header, and load the block header here instead. This would improve cases such as `getblock` with verbosity 2, where the block header is loaded and each transaction of the block is loaded, which would again load the block header for each transaction.

However, I don't think that is a significant performance impact, because the block header will be in the cache and parsing block headers is cheap.

But if you're interested in this, here is a sketch of the surrounding code. I can make a suggestion PR if you're interested. I think we might just go ahead and merge as is, but I'll leave it to you.

```rust
        // If the tx wasn't in the mempool, check the state.
        let tx = match state
            .ready()
            .and_then(|service| service.call(zebra_state::ReadRequest::Transaction(txid)))
            .await
            .map_misc_error()?
        {
            zebra_state::ReadResponse::Transaction(Some(tx)) => tx,
            zebra_state::ReadResponse::Transaction(None) => {
                return Err("No such mempool or main chain transaction")
                    .map_error(server::error::LegacyCode::InvalidAddressOrKey)
            }
            _ => unreachable!("unmatched response to a `Transaction` read request"),
        };

        match state
            .ready()
            .and_then(|service| {
                service.call(zebra_state::ReadRequest::BlockHeader(tx.height.into()))
            })
            .await
            .map_misc_error()?
        {
            zebra_state::ReadResponse::BlockHeader { header, .. } => Ok(if verbose {
                GetRawTransaction::Object(Box::new(TransactionObject::from_transaction(
                    tx.tx.clone(),
                    Some(tx.height),
                    Some(tx.confirmations),
                    &self.network,
                    Some(header.time),
                )))
            } else {
                let hex = tx.tx.into();
                GetRawTransaction::Raw(hex)
            }),

            zebra_state::ReadResponse::Transaction(None) => {
                Err("Transaction does not belong to a block")
                    .map_error(server::error::LegacyCode::InvalidAddressOrKey)
            }

            _ => unreachable!("unmatched response to a `Transaction` read request"),
        }
```

zebra-rpc/src/methods.rs
```diff
@@ -1248,12 +1251,14 @@ where
...
                 if let Some(tx) = txns.first() {
                     return Ok(if verbose {
-                        GetRawTransaction::Object(TransactionObject::from_transaction(
+                        GetRawTransaction::Object(Box::new(TransactionObject::from_transaction(
                             tx.transaction.clone(),
                             None,
                             None,
                             &self.network,
-                        ))
+                            // TODO: No time for mempool txs, check if this is correct.
```
conradoplg:
```suggestion
```

I think it's correct to skip it in mempool txs, zcashd seems to do the same.

conradoplg (dismissed):
Looks great, thanks! I left a couple of optional suggestions.

conradoplg (approved):

dismad:
Initial testing looking good! Here is picture showing zcashd on the left and zebrad on the right. 
Screenshot_2025-04-28_13-36-49

mpguerra:
@mergifyio requeue

PR #8107, opened 2023-12-13
fix(test): Ignore segmentation fault errors when shutting down Docker in tests

teor2345:
## Motivation

This PR ignores Docker memory errors that happen after the test has succeeded in CI.

Close #7898.

#### Check before marking the PR as ready for review:
  - [x] Will the PR name make sense to users?
  - [x] Does the PR have a priority label?
  - [x] Have you added or updated tests?
  - [x] Is the documentation up to date?

_If a checkbox isn't relevant to the PR, mark it as done._

### Specifications

https://tldp.org/LDP/abs/html/exitcodes.html
https://www.man7.org/linux/man-pages/man7/signal.7.html

The exit codes 137 and 139 correspond to SIGKILL (usually out of memory) and SIGSEGV (invalid use of memory).

### Complex Code or Requirements

We aren't ignoring test errors here because we check that `$LOGS_EXIT_STATUS` passed first, before ignoring these errors.

## Solution

- If the test passed, but shutting down Docker caused a memory error, ignore that error

### Testing

This test already runs on every Rust PR. If we don't see the error again for a few weeks, this fix worked.

## Review

This is a high priority because it's failing merging other things.

### Reviewer Checklist

Check before approving the PR:
  - [ ] Does the PR scope match the ticket?
  - [ ] Are there enough tests to make sure it works? Do the tests cover the PR motivation?
  - [ ] Are all the PR blockers dealt with?
        PR blockers can be dealt with in new tickets or PRs.

_And check the PR Author checklist is complete._

## Follow Up Work

Actually fix the underlying problem?

PR #9465, opened 2025-04-29
fix(ci): use log exit code instead of docker stop

gustavovalverde:
## Motivation

The `sub-test-zebra-config.yml` reusable workflow sometimes reports failures with exit codes like `137` (SIGKILL) or `139` (SIGSEGV), even when the core test condition (checking logs via `grep`) passes successfully. This happens because the `docker stop` command forcefully terminates the `zebrad` container if it doesn't shut down gracefully within the timeout period, leading to a non-zero exit code from `docker wait`. This makes CI results misleading, indicating a failure when the configuration test actually succeeded.

This issue was previously addressed in PR #8107 and its related to issue #7898. This PR re-applies the same fix after being removed in the Docker refactor in #9344

## Solution

Modified the final exit logic within the `Run ${{ matrix.name }} test` step in `.github/workflows/sub-test-zebra-config.yml`.

If the `grep` command successfully finds the required pattern in the logs (`LOGS_EXIT_STATUS -eq 0`), the script will now explicitly `exit 0`. This ensures the workflow step is marked as successful based on the primary test condition (log content verification), regardless of the exit code returned by `docker wait` after `docker stop` is called.

This aligns the step's success/failure status with the intended test validation and mirrors the logic previously implemented in PR #8107.

### Tests

This change is within the CI testing workflow itself.

-   Original Fix PR: https://github.com/ZcashFoundation/zebra/pull/8107
-   Related Issue: https://github.com/ZcashFoundation/zebra/issues/7898 (Based on PR #8107)
-   Linux Signals: `man 7 signal` (SIGKILL=9, SIGSEGV=11)
-   Exit Codes > 128: `128 + signal_number` (e.g., 137 = 128 + 9)

### Follow-up Work

While this fixes the CI reporting, the underlying reason why `zebrad` sometimes fails to shut down gracefully within the default `docker stop` timeout could be investigated further. However, this change ensures CI accurately reflects the test outcome based on log verification.

### PR Checklist

- [X] The solution is tested.

### DIFF
--- a/.github/workflows/sub-test-zebra-config.yml
+++ b/.github/workflows/sub-test-zebra-config.yml
@@ -123,5 +123,7 @@ jobs:
               exit 1;
           else
             echo "SUCCESS: Found the expected pattern in logs.";
-            exit $EXIT_STATUS;
+            # Exit successfully if grep passed, even if docker stop resulted in SIGKILL (137 or 139)
+            # See ticket #7898 for details.
+            exit 0;
           fi


conradoplg (approved):
Thank you for the thorough explanation!

PR #9471, opened 2025-04-29
re-export zcash-primitives to unify LRZ versions:

zancas:
## Motivation

The Z3 project is several things including a new version of the Zcash reference implementation!

It's also:

 * a radical expansion of mindshare.

Crudely speaking something like 4-times as many developers are immersed in core Zcash code-bases full-time.  I say crudely because it's non-trivial to understand who is participating in wider audiences.

 * a coordination experiment where multiple **independent** and autonomous communities are cooperating with each other.

We're mostly operating according to our own best discretion, from the bottom up, and meeting fellow autono-matons where we need each other.   We are unambiguously aligned in [our mission, To Empower ALL People With Economic Freedom](https://electriccoin.co/blog/refining-and-recommitting-to-the-ecc-mission/), and that axiom makes amazing new results possible.

#### To best coordinate complex systems, it best to refer unambiguously to a single source of truth where possible.

`zaino` is a consumer of `librustzcash`, and `zebra`, but `zebra` is also a consumer of `librustzcash`.  This means it's possible for `zaino` to depend on two different versions of `librustzcash` during build.

## Solution

I am operating under several assumptions:

(1) zaino MUST depend on (specific crates in) zebra
(2) zebra MUST depend on (specific crates in) LRZ
(3) zaino can coordinate requirement (from LRZ) with zebra
(4) the feature-set required by zebra is at *least* the feature set required by zaino (this is a very good reason to:
  AVOID CONDITIONAL COMPILATION, WHEREVER POSSIBLE (please try to build smaller crates instead)

BENEFITS:

By "re-exporting" librustzcash *from* zebra for consumption, this PR enables consumers (such as Zaino) to unify their dependency graph, depending on `librustzcash` only via `zebra`.   The effect is that the version of LRZ used by Zaino becomes the same as that consumed by Zebra.

It's important to understand what this means.  It's not simply that it gives consumers (of Zebra) a simpler dependency tree, there's something else that happens that's *more* important:

* It unifies the interests of Zebra and Zaino relative to librustzcash.  This is a very good thing.  It means that this interface between these project receives more attention, and that attention is framed in a systematic and unambiguous way.
   Concretely:
        - every time Zebra changes the version of LRZ it uses, Zaino devs **will** notice
        - every time LRZ developers change LRZ to support Zaino..   Zebra developers **will** notice

COSTS:

Zaino is more tightly coupled to the code versions that Zebra uses.
LRZ, Zebra, and Zaino developers must collaborate more closely.
Less flexibility in code versions in available.

### PR Checklist

- [ x ] The PR name is suitable for the release notes.

### DIFF
--- a/zebra-chain/src/lib.rs
+++ b/zebra-chain/src/lib.rs
@@ -42,6 +42,7 @@ pub mod value_balance;
 pub mod work;
 
 pub use error::Error;
+pub use zcash_primitives;
 
 #[cfg(any(test, feature = "proptest-impl"))]
 pub use block::LedgerState;
--- a/zebra-chain/src/lib.rs
+++ b/zebra-chain/src/lib.rs
@@ -42,6 +42,9 @@ pub mod value_balance;
 pub mod work;
 
 pub use error::Error;
+
+// Re-export librustzcash crates so consumers use Zebra’s versions directly,
+// avoiding duplicate builds or version mismatches.
 pub use zcash_primitives;
 
 #[cfg(any(test, feature = "proptest-impl"))]


zebra-chain/src/lib.rs
```diff
@@ -42,6 +42,7 @@ pub mod value_balance;
 pub mod work;
 
 pub use error::Error;
+pub use zcash_primitives;
```
oxarbitrage, line 48: I think it might be good to add a comment with the reason of adding this here:

```suggestion

// Re-export librustzcash crates so consumers use Zebra’s versions directly,
// avoiding duplicate builds or version mismatches.
pub use zcash_primitives;
```

I have the impression that `zcash_primitives` will not be the only create we will re-export here that is why the comment added says "Re-export librustzcash crateS ...".

oxarbitrage (commented):
I think this is good, has clear benefits as stated in https://github.com/ZcashFoundation/zebra/pull/9471

It might be possible to wrap this behind a feature flag, however that might complicate things even further.

pacu:
I'm not a rust expert but I've burned myself enough times with this milk to fear this cow quite a lot.


I support this or other approaches that solve the problem explained here by Zancas

zancas:
> It might be possible to wrap this behind a feature flag, however that might complicate things even further.

Regarding this proposition, let's consider the cases.

Suppose there's a feature flag for `lrz-reexport`.

Assume the consumer depends on `zebra`.

If they also require direct access to lrz features, and there's a feature flag then they activate the feature flag on zebra and compile the version that re-exports lrz (and makes no other changes).

So there's a complexity cost paid (they need to use the feature flag, consumers of *them* must decide whether or not to use that flag, whether or-not that flag is "default" must be considered, etc., etc.), and what's the benefit?

So let's consider the case of a consumer who does *not* need access to LRZ functionality...
In the case that there is no feature flag:

they simply don't reference the re-exported functionality.

In the case that there *is* a feature flag, they don't use the flag AND:

they simply don't reference the re-exported functionality.

So the non-user of LRZ has to make on more decision (whether or not to compile with a particular flag) a decision that they're likely to take the default on, for no benefit.

The user of LRZ, has to notice that the flag exists, and take the option, and *then* refer to the re-exported items.

Even worse, the consumer that *wants* to use LRZ will be more likely to not use the re-exported version, and simply depend on LRZ directly in which case they enter a multiple-version possible scenario (which is something we want to avoid).

### Caveat:

The pattern is rendered somewhat more use-able by my previous work applying `cargo autoinherit` to enforce a workspace wide single-source-of-truth.

I think it's fair to say that the emergence of workspaces is the innovation that does the most to decrease the need for feature gates, and thereby increase the coherence of Rust projects.

zebra-chain/src/lib.rs
zancas: True, but I think that `zcash_primitives` is sufficient for the discussion.

It's a Proof Of Concept, if we like it, then yeah..  I assume we'll use the pattern everywhere we can.

I also wasn't sure which zebra crate would be the best choice, for Zaino..   or more generally?

conradoplg:
I think this is a good practice.

It should be a requirement if any of the LRZ types are being used in the public Zebra API. But even if not, it sounds useful. 

I think a feature flag for this is overkill.

(Note that LRZ itself does not follow this policy though [AFAIK](https://discord.com/channels/809218587167293450/876655911790321684/1265317720828805191). But that's not really a problem here, only for the direct LRZ users like Zebra itself)

daira:
I think you might be overestimating the difficulty of ensuring that there is only one version of the lrz crates used. Cargo will always try to unify the versions of each librustzcash crate used by Zaino and zebra as long as they have semver-compatible version specs for the dependency: https://doc.rust-lang.org/cargo/reference/resolver.html
If you want it to be an error to have multiple versions of certain lrz crates, just run `cargo duplicates` in Zaino's CI and assert that.

Edit: if you did want to do this, why is re-exporting `zcash_primitives` sufficient? librustzcash has a whole bunch of crates that are released somewhat independently.

nuttycom:
I think that this is somewhat the wrong tack to take. In fact, I believe that taking this approach may cause *more significant* coupling problems than would otherwise be present.

The issue at question is to what degree the `zcash_primitives` types "leak" across the API boundaries. For any type that is not exposed in a public API, there is no problem: rust is completely fine with having multiple versions of the same library used in a project, so long as the types from version A aren't used where the types from (semver-incompatible) version B are expected. By explicitly reexporting `zcash_primitives`, this puts the _entire API surface_ of `zcash_primitives` into the public API of Zebra, which I believe to be highly undesirable. This means that every API-breaking change of `zcash_primitives` becomes an API-breaking change of `zebra-chain` and `Zaino`, needs to be fully documented as such, and so forth.

Fundamentally, I think that we should _avoid_ exposing foreign types across the API boundary as much as possible. For types such as transactions, I almost think that the appropriate API type is _bytes_, in the canonical serialized representation.

Bytes (or newtype wrappers around bytes) might not be the best type for public APIs if storage-level representations have already needed to be parsed in order to expose them up to the API surface. But in the cases where it's possible, avoiding deserialization and reserialization would provide both better performance, and more robustness as versioning is then defined at the _protocol_ layer instead of the library layer.

It's not up to me to decide, but I would NACK this change.

zancas:
> ...explicitly reexporting `zcash_primitives`, this puts the _entire API surface_ of `librustzcash` into the public API...

I am confused by this assertion, is it meant to read:

"...explicitly reexporting `zcash_primitives`, this puts the _entire API surface_ of `zcash_primitives` into the public API..."

?

nuttycom:
> I am confused by this assertion, is it meant to read:
> 
> "...explicitly reexporting `zcash_primitives`, this puts the _entire API surface_ of `zcash_primitives` into the public API..."

Yes, fixed, thanks for the catch.

zancas:
I agree that this is a proposal to *increase* the coupling between Zaino and Zebra, source code.

It *is* in that sense constraining, and that's not zero-cost.

> This means that every API-breaking change of `zcash_primitives` becomes an API-breaking change of `zebra-chain` and `Zaino`, needs to be fully documented as such, and so forth.

I'm a bit confused here.

`Zaino` is a consumer of `zcash_primitives`, breaking changes in the public `zcash_primitives` API should affect Zaino in the same way, whether Zaino is using a re-export, or a direct export..  right?

But setting that aside and focusing on the implications for the public API of Zebra, could the `zebra-chain::zcash_primitives;` entrypoint not simply refer to the relevant public API documents in zcash_primitives?
It seems that the single source of truth would be in zcash_primitives, and the re-export documentation would simply "refer to upstream".

zancas:
Taking a step back, there's some benefit to being able to run:

`cargo tree -i zcash_primitives`

in Zaino without having to qualify the target with an @ symbol.

The benefit of re-exporting that we're hoping to achieve is something like "version unification" in a downstream consumer that's tightly coupled to an upstream where both have a common "further upstream".

That (as succinctly expressed by the `cargo tree`) seems valuable to us, and we'd like to achieve it, if there's a sufficiently low cost approach to doing so.

So..  long story short, that's the problem statement, and this PR is an attempt at a solution.

nuttycom:
> `Zaino` is a consumer of `zcash_primitives`, breaking changes in the public `zcash_primitives` API should affect Zaino in the same way, whether Zaino is using a re-export, or a direct export.. right?

I'm recommending that this go the _opposite_ direction: that `zebra-chain` should not expose _any_ `zcash_primitives` types as part of its public API.
 
> But setting that aside and focusing on the implications for the public API of Zebra, could the `zebra-chain::zcash_primitives;` entrypoint not simply refer to the relevant public API documents in zcash_primitives? It seems that the single source of truth would be in zcash_primitives, and the re-export documentation would simply "refer to upstream".

It would have to refer to the CHANGELOG, not the public API documentation, and would have to ensure that its release notes capture all of the relevant CHANGELOG entries being referenced. 

With respect to `cargo tree -i zcash_primitives`, I think that the benefits here are minimal; if instead Zebra did not expose `zcash_primitives` types in its public API, then `Zaino` would be free to use whatever version of `zcash_primitives` was appropriate to its needs. Importantly, I really don't want `Zaino` exposing `zcash_primitives` types in its public API either! I would like it to be possible for `zallet` to depend on whatever versions of crates in the `librustzcash` stack that it needs to without changes being blocked by `zebra-chain` and `zaino` releases.

nuttycom:
One thing that I note is that at present, `zebra` is relying on a bunch of types from `zcash_primitives` that are in fact (deprecated) reexports of `zcash_protocol` types.

The `zcash_protocol` crate, by way of contrast to `zcash_primitives`, is intended to be semver-stable between protocol version releases going forward (it hasn't been so thus far, because its extraction was relatively recent and it took a while to figure out exactly what was appropriate to go into it). I will look through `zebra` and open a PR to use the `zcash_protocol` types instead of the `zcash_primitives` reexports where possible, and then we can see what the API surface actually looks like.

zancas:
> I think that the benefits here are minimal; if instead Zebra did not expose `zcash_primitives` types in its public API, then `Zaino` would be free to use whatever version of `zcash_primitives` was appropriate to its needs.

Right this proposal definitely reduces (indeed eliminates) Zaino's freedom to choose versions, assuming Zaino chooses to use Zebra's re-export (which is the intention).

> I would like it to be possible for zallet to depend on whatever versions of crates in the librustzcash stack that it needs to without changes being blocked by zebra-chain and zaino releases.

This I don't understand.

Is Zebra not free to choose when/how to update its dependency on `librustzcash`?

Isn't Zallet a downstream consumer like Zebra (and currently Zaino) of LRZ?

How does changing the inter-relationship between Zebra and Zaino affect Zallet's consumption of LRZ?

nuttycom:
> Is Zebra not free to choose when/how to update its dependency on `librustzcash`?

Yes.

> Isn't Zallet a downstream consumer like Zebra (and currently Zaino) of LRZ?

Yes.

> How does changing the inter-relationship between Zebra and Zaino affect Zallet's consumption of LRZ?

If `Zaino` exposes `zcash_primitives` types as part of the public API that `Zallet` consumes, then if `Zallet` wants to update its `zcash_primitives` dependency, it will face a version incompatibility (and have to map between versions) when using those APIs. I would prefer that those incompatibilities not arise in the first place - neither between `Zallet` and `Zaino`, nor (for your sake) between `Zaino` and `Zebrad`.

zancas:
> I will look through `zebra` and open a PR to use the `zcash_protocol` types instead of the `zcash_primitives` reexports where possible, and then we can see what the API surface actually looks like.

Ahh..  interesting.   I just picked "zcash_primitives" from a list provided by the Zaino devs..   it sounds like there might be a useful refactor to prefer `zcash_protocol` types, in the works!

Pending that work, I will just leave this as a draft (unless a Zebra Dev closes it).

daira:
> If you want it to be an error to have multiple versions of certain lrz crates, just run `cargo duplicates` in Zaino's CI and assert that.

Alternatively, use the [`links` directive](https://doc.rust-lang.org/cargo/reference/resolver.html#links). It's not really the intended use of that `cargo` feature, but I think it can be made to have the desired effect — as long as it can be conditional on a "no-lrz-duplicates" feature flag, which I think it can. If nothing else works, have an otherwise empty crate in the lrz workspace that contains the `links` directive with a dummy library name, and declare a feature-conditional dependency on it from the other lrz crates. This is better than a check in CI because it breaks the build (as intended) even when CI is not run.

To be specific, the desired effect is that Zaino takes on the responsibility of ensuring stricter version compatibility requirements without imposing it on other users of the zebra or lrz crates. So if a zebra crate starts depending on a new semver-breaking version of an lrz crate *internally*, Zaino (and other users that explicitly opt in to "no-lrz-duplicates") would have to deal with that, but no other zebra or lrz users would, and that zebra crate would not be forced to treat the internal change as semver-breaking when bumping its own version.

Notice that with this approach you *can* have duplicates of lrz crates, depended on by client crates that do not opt in to the "no-lrz-duplicates" feature. This is good and intended. (Feature unification only applies to each instance of a crate; semver-incompatible duplicate crates can have different features.) It works correctly regardless of which lrz crate you depend on, because internal dependencies within lrz would propagate the feature.

Furthermore, `cargo`'s dependency resolution will *try to avoid* the duplication, and will only break the build if it can't. This is an advantage of using the `links` directive that I think cannot be achieved any other way.

This approach solves the problem @nuttycom was worried about, which I agree is a significant concern:
> By explicitly reexporting `zcash_primitives`, this puts the _entire API surface_ of `zcash_primitives` into the public API of Zebra, which I believe to be highly undesirable. This means that every API-breaking change of `zcash_primitives` becomes an API-breaking change of `zebra-chain` and `Zaino`, needs to be fully documented as such, and so forth.

This is all orthogonal to the API changes @nuttycom suggested, which I also agree are separately a good idea.

zancas:
> Edit: if you did want to do this, why is re-exporting `zcash_primitives` sufficient? librustzcash has a whole bunch of crates that are released somewhat independently.

I don't know if you're referring to [this comment](https://github.com/ZcashFoundation/zebra/pull/9471#discussion_r2067558186)?  Or are unaware of it..  I don't assume that re-exporting zcash_primitives is sufficient to dependency on a single version of LRZ, I assume it's sufficient to stimulate this conversation.

zancas:
> If you want it to be an error to have multiple versions of certain lrz crates, just run `cargo duplicates` in Zaino's CI and assert that.

This may indeed be the better solution.

conradoplg:
I'm closing this per the previous discussion. We should only re-export foreign types if truly necessary and consumers can detect duplicates by using `cargo duplicates` or similar tools

PR #9479, opened 2025-05-01
refactor(ci): simplify GCP integration with `create-with-container`

gustavovalverde:
## Motivation

The previous method for running GCP integration tests involved an intermediate `busybox` container, manual `docker run` commands via SSH, and complex disk mounting steps. This approach had several drawbacks:
- Container logs were not reliably sent to the GCP Cloud Logging system, hindering debugging.
- The `busybox` layer added unnecessary complexity and obscured the direct handling of the test container by GCP.
- Manual disk mounting required workarounds.

Includes fixes for issues in environment variable formatting (`test_variables`), container ID detection, and exit code handling logic identified during the refactoring process.

## Solution

Refactors the `.github/workflows/sub-deploy-integration-tests-gcp.yml` workflow to launch the Zebra test container directly using `gcloud compute instances create-with-container`.

Key changes include:
- Replaced the `busybox` + `docker run` approach with `gcloud compute instances create-with-container` using the specific test image (`${{ vars.CI_IMAGE_NAME }}`).
- Removed separate "Format volume" and "Launch test" steps previously required.
- Configured container environment variables (`--container-env`) and disk mounts (`--container-mount-disk`) directly via `gcloud` parameters.
- Corrected the format of `test_variables` in the calling workflow (`.github/workflows/sub-ci-integration-tests-gcp.yml`) to use comma-separated `KEY=VALUE` pairs, fixing issues with environment variable propagation.
- Updated the log checking logic to dynamically find the container ID based on its expected name prefix, making it more robust than the previous `head -n 1` method.
- Refined the success/failure logic in the test result step: the test now fails if the expected log pattern is *not* found OR if the pattern *is* found but the container exits specifically with code `1`. Other non-zero exit codes (like 101 for panics) are ignored if the success log pattern is present.
- Added whitespace trimming for the `vars.GAR_BASE` variable to prevent potential issues in the image name.

This aligns the instance creation more closely with standard GCP practices, removes the indirection layer of the `busybox` container, ensures logs are sent to Cloud Logging, and simplifies the overall workflow structure.

### Tests

- All test should run as expected

- https://cloud.google.com/sdk/gcloud/reference/compute/instances/create-with-container

### PR Checklist

- [x] The solution is tested.


.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -195,103 +212,41 @@ jobs:
           if [ -n "${{ env.CACHED_DISK_NAME }}" ]; then
             DISK_PARAMS+=",image=${{ env.CACHED_DISK_NAME }}"
           fi
+
+          # Mount the disk(s) to the container
+          CONTAINER_MOUNT_DISKS="--container-mount-disk=mount-path=${{ inputs.zebra_state_dir }},name=${NAME},mode=rw"
+          # Mount the same disk to the lwd path if needed
+          if [[ "${{ inputs.needs_lwd_state }}" == "true" || "${{ inputs.test_id }}" == "lwd-full-sync" ]]; then
+            CONTAINER_MOUNT_DISKS+=" --container-mount-disk=mount-path=${{ inputs.lwd_state_dir }},name=${NAME},mode=rw"
+          fi
```
arya2, line 221: What's happening here? Is it mounting another cached image to the disk/storage-volume attached to the VM, or is it mounting the same cached image to two different paths? Where is `input.lwd_state_dir` coming from and what is expected to be?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -195,103 +212,41 @@ jobs:
...
+
+          # Environment variables for the container
+          CONTAINER_ENV="${{ inputs.test_variables }},RUST_LOG=${{ env.RUST_LOG }},RUST_BACKTRACE=${{ env.RUST_BACKTRACE }},RUST_LIB_BACKTRACE=${{ env.RUST_LIB_BACKTRACE }},COLORBT_SHOW_HIDDEN=${{ env.COLORBT_SHOW_HIDDEN }},CARGO_INCREMENTAL=${{ env.CARGO_INCREMENTAL }}"
+
+          # Trim whitespace from GAR_BASE as for some reason it's getting a trailing space
+          GAR_BASE_TRIMMED=$(echo "${{ vars.GAR_BASE }}" | xargs)
```
arya2, line 227: What's `GAR_BASE`? Are we sure `xargs` won't do anything unexpected?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -195,103 +212,41 @@ jobs:
...
           --boot-disk-size=50GB \
           --boot-disk-type=pd-ssd \
           --image-project=cos-cloud \
           --image-family=cos-stable \
           --create-disk="${DISK_PARAMS}" \
-          --container-image=gcr.io/google-containers/busybox \
-          --machine-type ${{ inputs.is_long_test && vars.GCP_LARGE_MACHINE || vars.GCP_SMALL_MACHINE }} \
-          --network-interface=subnet=${{ vars.GCP_SUBNETWORK }} \
+          ${CONTAINER_MOUNT_DISKS} \
+          --container-stdin \
```
arya2, line 237: Why do we need this option? Were we using it before in this workflow? What's it for in other workflows?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -195,103 +212,41 @@ jobs:
...
+          --container-tty \
```
arya2, line 238: I see that `--tty` was being configured before, but I'm just curious, what's it for?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -356,26 +281,41 @@ jobs:
           --ssh-flag="-o ConnectionAttempts=20" \
           --ssh-flag="-o ConnectTimeout=5" \
           --command=' \
+          echo "Waiting for container to start..."; \
+          CONTAINER_ID=""; \
+          CONTAINER_PREFIX="klt-${{ inputs.test_id }}-${{ env.GITHUB_REF_SLUG_URL }}-${{ env.GITHUB_SHA_SHORT }}"; \
+          for i in {1..30}; do CONTAINER_ID=$(sudo docker ps --filter name=${CONTAINER_PREFIX} -q --no-trunc); if [ -n "${CONTAINER_ID}" ]; then break; fi; echo "Waiting for container ID starting with ${CONTAINER_PREFIX}... ($i/30)"; sleep 2; done; \
```
arya2, line 287: Are there any normal circumstances where it may take longer?

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -356,26 +281,41 @@ jobs:
...
+            echo "Test failed: Success log pattern not found (grep exit status: $LOGS_EXIT_STATUS).";
+            exit 1;
+          else
+            # Grep succeeded (pattern found), now check specific exit code
+            if [ "$EXIT_STATUS" -eq 1 ]; then
+              # Explicit failure code 1
+              echo "Test failed: Success log pattern found BUT container exited with status 1.";
+              exit 1;
+            else
+              # Grep succeeded and exit status is 0 or non-1
```
arya2: Nitpick/Optional:

```suggestion
              # Grep succeeded and exit status is not 1
```

.github/workflows/sub-deploy-integration-tests-gcp.yml
```diff
@@ -486,15 +426,30 @@ jobs:
...
           --zone ${{ vars.GCP_ZONE }} \
           --ssh-flag="-o ServerAliveInterval=5" \
           --ssh-flag="-o ConnectionAttempts=20" \
           --ssh-flag="-o ConnectTimeout=5" \
           --command=' \
-          sudo docker logs ${{ inputs.test_id }} | head -1000 \
-          ')
+          CONTAINER_ID=$(sudo docker ps -q --no-trunc | head -n 1); \
+          if [ -n "${CONTAINER_ID}" ]; then \
+             sudo docker logs ${CONTAINER_ID} | head -1000; \
```
arya2, line 442: Just curious, why/how did we choose 1000?

arya2 (dismissed):
This looks great, thank you!

I left some questions, but none of them are blockers (and I have answers from the ref that I'm only trying to confirm for some of them).

> All test should run as expected

It looks like they all do, except [this one](https://github.com/ZcashFoundation/zebra/actions/runs/15189451284/job/43229484786), which seems like a pre-existing issue (if it's an issue at all), because the changes to it look correct, and it shouldn't be running on a PR update at all unless it's missing a cached state,

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: In this case, when a tests `needs_lwd_state` then we search for that image (a disk with the state) https://github.com/ZcashFoundation/zebra/blob/2e60cedd303f14067796eedef5f28c6b3f845d76/.github/workflows/sub-deploy-integration-tests-gcp.yml#L118 and mount the same cached image to both paths, as the lwd cache contains both states (zebra and lwd).

For `input.lwd_state_dir` there's a default here, in the same workflow:
https://github.com/ZcashFoundation/zebra/pull/9479/files/ff4c2d84757cca5d95796af04524876684a438d6#diff-822f5cf974be53d74a55e0cc766088ec19a291f2dd5e56428ebb858ed4f810c7R43-R47

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: To be honest, we should not be doing this, but I had no other way. `GAR_BASE` is the base URL for Google Artifact Registry (`us-docker.pkg.dev/zfnd-dev-zebra/zebra`), that we have saved in our GitHub Environment variables, but for some unknown reason it's getting here with a trailing space

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: This is for debugging purposes, if we need to remotely connect to the container and execute commands inside of it.

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: Same as above. But this is a more elaborate answer (that helped me solving debugging with Zebra a few years ago) for both scenarios: https://forums.docker.com/t/trying-to-understand-the-relation-of-tty-between-the-host-and-container-when-the-docker-run-command-is-executed/140034/2

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: That wouldn't be normal, but it can happen.

.github/workflows/sub-deploy-integration-tests-gcp.yml
gustavovalverde: This is the main reason:

```
      # Typically, the database versions are around line 20 in the logs..
      # But we check the first 1000 log lines, just in case the test harness recompiles all the
      # dependencies before running the test. (This can happen if the cache is invalid.)
      #
```

.github/workflows/sub-deploy-integration-tests-gcp.yml
arya2: Ah, okay. It's constant, no chance of unexpected behaviour.

arya2 (approved):

Issue #9468, opened 2025-04-29
feature: reduce dependency versions in Cargo.toml


conradoplg:
### Motivation

Due to https://github.com/ZcashFoundation/zebra/issues/9398 our dependencies have versions that are  higher than it is probably required.

We need to go back and bump down the dependency versions, ideally to the oldest one that works. In most cases that will entail zeroing the patch number (e.g. "1.3.5" -> "1.3.0") though arguably in some cases we could zero the minor version too.

And to make sure everything works, we should test that compiling with minimal versions works. We could do that manually or add to CI if it's not too slow. 

### Testing

See how the testing is done in frost: https://github.com/ZcashFoundation/frost/blob/a8c8b6d197bd818e23fc8fcfc457144a909872e0/.github/workflows/main.yml#L21-L38

PR #9484, opened 2025-05-04
change(deps): Allow Zebra crates to be compiled with alternative versions of their dependencies

arya2:
## Motivation

We want to allow Zebra crates to be compiled with later versions of their dependencies.

See [conversation on Discord](https://discord.com/channels/809218587167293450/809251029673312267/1364645363457458316).

## Solution

Allow Zebra crates to compile with later versions of their dependencies, similar to [the change in `zcash_script`](https://github.com/ZcashFoundation/zcash_script/pull/66).

### PR Checklist

- [ ] The solution is tested.


conradoplg (changes requested):
I'm not sure this is the best approach, I'll expand it later, just want to block from merging

conradoplg:
My reasoning:

- From what I understand from the discussion, the issue is that we specify the latest patch version of each update, but that creates unnecessary constraints on crate users. For example, if some dependency is `= 1.2.8` then other projects that use our crate are force to update to 1.2.8 or later (which can be annoying if they audit every dependency bump) even if e.g. 1.2.0 would suffice. This particular PR doesn't change that
- I don't think changing to `>=` is good since it will allow updating major versions which can have breaking change. I don't fully understand the context of https://github.com/ZcashFoundation/zcash_script/pull/66 but I think the idea there was just to allow some specific exception which use 0.x versions
- I think the best approach would to not use `>=` and simply bump dependency versions _down_, in most cases simply removing the patch version or setting them to 0, e.g. "= 1.2.8" -> "= 1.2.0" or "= 1.2" (which are the same). See [this issue](https://github.com/ZcashFoundation/zebra/issues/9468) I created

conradoplg:
Thanks! There was another part to this though, which is making sure we can compile everything with the minimal versions specified in the `Cargo.toml`. This ensures everything works properly even for people using older versions of the shared dependencies.

I did this by running `cargo +nightly update -Z direct-minimal-versions` and trying to build (actually, just trying to run clippy). I started fixing things and ended up fixing everything :sweat_smile: so I [committed](https://github.com/ZcashFoundation/zebra/pull/9484/commits/83341dfdc7ef75fc3ec08d443d3aa458e4125fe6) the result.

For each failure I re-bumped the dependency that broke to the original version we had specified. Ideally we could determine that is the minimum version that works, but that will take too much time, and this seems an acceptable compromise.

One thing broke that required manually tinkering with the lockfile, so I had to run `cargo update color-spantrace --precise 0.2.1` after the command above before trying to build. This is caused by https://github.com/eyre-rs/eyre/issues/230 and should be fixed soon.

We might want to try to add these steps to CI. I'll leave it up to you if you want to do in this PR or in a separate one. We might want to wait for that eyre bug to be fixed so that we can just run `cargo +nightly update -Z direct-minimal-versions`.

Let me know if my changes make sense and I can approve this.

arya2:
> Let me know if my changes make sense and I can approve this.

Should we add `>=` for the dependencies where there's still an explicit patch version?

conradoplg:
> > Let me know if my changes make sense and I can approve this.
> 
> Should we add `>=` for the dependencies where there's still an explicit patch version?

Why do you think we should do that? I'm not seeing the benefit of allowing major version bumps which are expected to be breaking changes. (Recall that the default does allow minor and patch increases, i.e. 1.3.0 works with `foo = 1.2.3`)

conradoplg (approved):

Issue #9477, opened 2025-04-30
Switch documentation from OpenAPI to OpenRPC


oxarbitrage:
Our current OpenAPI‐based spec is a poor fit for JSON-RPC, making us introduce workarounds around things like encoding method parameters. Instead, we should migrate to [OpenRPC](https://open-rpc.org/), which is purpose-built for JSON-RPC 2.0.

This will be much simpler if https://github.com/paritytech/jsonrpsee/issues/737 lands.

Zallet also uses jsonrpsee and plans to adopt the same spec, so by converging on OpenRPC we can share tooling and ensure compatibility across both projects.

https://github.com/zcash/wallet/issues/116#issuecomment-2842150257

oxarbitrage:
Some references about implementation in Zallet:

https://github.com/zcash/wallet/pull/129
https://github.com/zcash/wallet/pull/130

PR #9459, opened 2025-04-24
change(rpc): Move GBT RPCs into the main RPC server

upbqdn:
## Motivation

- Close #9425.
- Close #8169.

## Solution

- This PR only moves code around. It doesn't change any functionality for node operators, but it changes the public APIs for API consumers.
- All RPCs are now under one RPC implementation.

## Review

This PR may be easier to review with `--color-moved` like `git diff --color-moved origin/main..HEAD` 

### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The solution is tested.
- [x] The documentation is up to date.
- [x] The PR has a priority label.
- [ ] If the PR shouldn't be in the release notes, it has the
      `C-exclude-from-changelog` label.

PR #9485, opened 2025-05-05
docs(openapi): Remove getblocktemplate file from scrtipt

### LINKED ISSUE #9425: GBT RPCs are separate from all other RPCs
After PR #9401, the GBT RPCs are redundantly separated from the remaining RPCs. This causes issues such as the GBT RPCs not being listed among supported RPCs in the docs: https://doc-internal.zebra.zfnd.org/zebra_rpc/methods/trait.RpcServer.html.

oxarbitrage:
## Motivation

The OpenAPI generation script is currently failing in https://github.com/ZcashFoundation/zebra/pull/9459

The script needs updating, and it’s likely to be replaced by https://github.com/ZcashFoundation/zebra/issues/9477, but we can fix the current failure now:

```
cargo run --bin openapi-generator --features="openapi-generator"
...
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

## Solution

Remove the reference to the missing file so that the script can execute.

### DIFF
--- a/zebra-utils/src/bin/openapi-generator/main.rs
+++ b/zebra-utils/src/bin/openapi-generator/main.rs
@@ -83,19 +83,10 @@ fn main() -> Result<(), Box<dyn Error>> {
     let current_path = env!("CARGO_MANIFEST_DIR");
 
     // Define the paths to the Zebra RPC methods
-    let paths = vec![
-        (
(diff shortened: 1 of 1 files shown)


upbqdn (approved):
Thanks for fixing this.

PR #9266, opened 2025-02-18
feat(rpc): Add `invalidateblock` and `reconsiderblock` RPC methods

arya2:
## Motivation

This PR adds the `invalidateblock` and `reconsiderblock` RPC methods to enable more ergonomic integration testing with side chains and chain re-organizations.

This PR is a draft as it's still missing tests.

Closes #8848
Closes #8847
Closes #8846
Closes #8845
Closes #8844
Closes #8843

Depends-On: #9260.

## Solution

- Updates non-finalized block write channel message type to an enum with `Invalidate` and `Reconsider` variants
- Adds state requests to send those messages to the channel
- Handles the messages in the block write task by calling `invalidate_block()` or `reconsider_block()` and updating the chain channels

Related changes:
- Refactors block write task into a struct (feedback on how to improve it further would be appreciated)

### Tests

This PR still needs tests.

### Follow-up Work

- https://github.com/ZcashFoundation/zebra/issues/9265

### PR Author's Checklist

- [x] The PR name will make sense to users.
- [ ] The PR provides a CHANGELOG summary.
- [ ] The solution is tested.
- [x] The documentation is up to date.
- [x] The PR has a priority label.

### PR Reviewer's Checklist

- [ ] The PR Author's checklist is complete.
- [ ] The PR resolves the issue.

PR #9494, opened 2025-05-06
add(rpc): Add a `mempool_change()` RPC method for listening to changes in the mempool

### LINKED ISSUE #8620: Add a `mempool_transactions()` RPC method for listening to changes in the mempool
## Motivation

We want to notify RPC clients when there has been a change to the mempool with information about which transactions were added or removed.

## Design

- Add a `MempoolChange` enum with `Added`, `Invalidated`, and `Mined` variants
- Add a channel for the mempool to broadcast changes to its verified tx set
- Send a `MempoolChange` when a transaction is inserted, invalidated by a chain tip change, or mined onto the best chain
- Add a receiver for the new channel to the indexer RPC server (or the regular RPC server now that it uses `jsonrpsee`)
- Add an RPC method that notifies clients of mempool changes
  - Send any existing mempool transactions when the method is called as `Added` changes
  
<details>
<summary>Outdated</summary>
[add a `MempoolChange` type with an `async changed()` method (like `ChainTip`).](https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/src/methods/get_block_template_rpcs.rs#L690C29-L690C42)

This would require adding a mempool change channel that could help clear [this TODO](https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/src/methods/get_block_template_rpcs.rs#L741-L743).
</details>

arya2:
## Motivation

This PR adds a `mempool_change` gRPC method to the indexer RPC service to notify listeners when a transaction is added to the mempool, invalidated in the mempool, or removed from the mempool because it was mined onto the best chain.

Will close #8620.

## Solution

- Update the broadcast channel returned when the mempool is instantiated to advertise any change to the mempool, rather than only broadcasting the ids of transactions that were added to the mempool, and
- Add a `mempool_change` method to the indexer gRPC server to notify listeners of mempool changes.

Related:
- Updates `TrustedChainSync` to use the streaming `chain_tip_change` indexer gRPC instead of polling the `bestchainhashandheight` JSON RPC.
- Removes `BestTipChanged` and updates the `best_tip_changed()` method on `ChainTip` to use async trait fns instead.

### Tests

This PR still needs tests for checking that the mempool broadcasts lists of mined tx ids correctly.

There were existing tests to cover when added tx ids should be broadcasted (for the gossip task) which have been updated to check for invalidated changes too.

## Follow Up Work

Update the getblocktemplate RPC's long polling logic to query the mempool when there's a mempool change instead of polling it at a fixed interval

### PR Checklist

- [ ] The solution is tested.


arya2:
This is still missing test coverage for the new gRPC method (snapshot & vector) and a test for `Mined` mempool changes being broadcasted as expected, but the unit test coverage for `Storage` is fairly good, and I'm expecting this to conflict with https://github.com/ZcashFoundation/zebra/pull/9266, so I'd like to open an issue for any of those that I don't in before this is reviewed.

zebra-chain/src/chain_tip.rs
```diff
@@ -68,7 +67,9 @@ pub trait ChainTip {
     // TODO:
     // Use async_fn_in_trait or return_position_impl_trait_in_trait when one of them stabilises:
     // https://github.com/rust-lang/rust/issues/91611
```
conradoplg: Nitpick: we can remove the above comment now I think?

zebra-rpc/src/indexer/methods.rs
```diff
@@ -41,10 +47,21 @@ where
...
+                    continue;
+                };
+
+                if let Err(error) = response_sender
+                    .send(Ok(BlockHashAndHeight::new(tip_hash, tip_height)))
+                    .await
+                {
+                    tracing::info!(?error, "failed to send chain tip change, dropping task");
+                    return;
+                }
```
conradoplg, line 61: What is the rationale for removing the `tokio::spawn()`?

zebrad/src/components/mempool.rs
```diff
@@ -646,13 +653,16 @@ impl Service<Request> for Mempool {
...
                         storage.reject_if_needed(tx_id, error);
                     }
                     Err(_elapsed) => {
                         // A timeout happens when the stream hangs waiting for another service,
                         // so there is no specific transaction ID.
 
-                        tracing::info!("mempool transaction failed to verify due to timeout");
+                        // TODO: Return the transaction id that timed out during verification so it can be
+                        //       included in the list of invalidated transactions and change `warn!` to `info!`.
+                        tracing::warn!("mempool transaction failed to verify due to timeout");
```
conradoplg, line 666: Just to double check, did you mean to solve this TODO in this PR or later?

conradoplg (commented):
Flushing comments from a preliminary first pass review

zebra-rpc/src/indexer/methods.rs
conradoplg: Explained in a video review, it was needed for a corner case which does not matter now

zebrad/src/components/mempool.rs
arya2: I meant to make this change in another PR, I think it'll be complicated on its own, and perhaps has limited value.

zebra-rpc/src/indexer/methods.rs
```diff
@@ -54,4 +71,50 @@ where
...
         Ok(Response::new(Box::pin(response_stream)))
     }
+
+    async fn mempool_change(
+        &self,
+        _: tonic::Request<Empty>,
+    ) -> Result<Response<Self::MempoolChangeStream>, Status> {
+        let (response_sender, response_receiver) = tokio::sync::mpsc::channel(RESPONSE_BUFFER_SIZE);
+        let response_stream = ReceiverStream::new(response_receiver);
+        let mut mempool_change = self.mempool_change.resubscribe();
```
arya2: Consider using a broadcast Sender here instead and calling `subscribe()` if these messages are being buffered needlessly.

zebrad/Cargo.toml
```diff
@@ -61,7 +61,8 @@ default-release-binaries = ["default", "sentry"]
 # Production features that activate extra dependencies, or extra features in dependencies
 
 # Indexer support
-indexer = ["zebra-rpc/indexer-rpcs", "zebra-state/indexer"]
+indexer = ["indexer-rpcs", "zebra-state/indexer"]
+indexer-rpcs = ["zebra-rpc/indexer-rpcs"]
```
arya2: Rename to `grpc-server` as it may not be limited to indexer RPCs in the future. Consider removing the feature altogether.

Update:

Removed the feature altogether in https://github.com/ZcashFoundation/zebra/pull/9494/commits/758854fb88153cbfc88e33a980837e3cfe287e33.

But now it'll require `protoc` to compile, so I'm thinking we should switch to [`skip_protoc_run()`](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.skip_protoc_run) in this PR too.

zebra-chain/src/chain_tip.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9494/commits/b4c726cc44a52d953580cc4fb119823ec20ff830.

Nice catch, thank you.

zebrad/Cargo.toml
conradoplg: I don't understand how `skip_protoc_run()` works...

Can we simply commit the generated files?

arya2: I think so, but copying indexer-descriptors.bin into a `__generated__` folder and setting the file descriptor path isn't working, so it may not be (or I've made a trivial mistake somewhere).

I'd be inclined to merge this as-is and open a follow-up PR to be merged before the next release if that seems okay.

conradoplg: Yep, that's fine. I tried to look into this but it is annoyingly under documented

conradoplg (changes requested):
Looks good, there's just a small conflict now

conradoplg (dismissed):

conradoplg:
just FYI @arya there are some build failures

arya2:
Looks like we need to remove the protoc dependency in this PR for it to pass CI.

conradoplg:
> Looks like we need to remove the protoc dependency in this PR for it to pass CI.

Should we simply install it in the `install-from-lockfile-no-cache` job? protoc is already documented as being a Zebra dependency, and other jobs already install it with `arduino/setup-protoc@v3.0.0`

mpguerra:
@arya2 is this ready for another review by @conradoplg ?

arya2:
> Should we simply install it in the `install-from-lockfile-no-cache` job? protoc is already documented as being a Zebra dependency, and other jobs already install it with `arduino/setup-protoc@v3.0.0`

That would work, too.

conradoplg (approved):

conradoplg:
@mergifyio requeue

PR #9513, opened 2025-05-08
update README with Arch build patch

ala-mode:
Building on unembellished Arch Linux currently fails on running a build script of the `rocksdb` dependency. This appears to be because this script breaks when using a new version of GCC. We could provide a minimal recreation soon. The suggestion in the README can patch this problem.

- [X] The solution is tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,11 @@ sudo pacman -S rust clang protobuf
 
 Note that the package `clang` includes `libclang` as well as the C++ compiler.
 
+Recently the GCC version on Arch has broken a build script in the `rocksdb` dependency. A workaround is:
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,11 @@ sudo pacman -S rust clang protobuf
 
 Note that the package `clang` includes `libclang` as well as the C++ compiler.
 
+Recently the GCC version on Arch has broken a build script in the `rocksdb` dependency. A workaround is:
+```sh
```
upbqdn, line 85:
```suggestion

```sh
```

upbqdn (dismissed):
Thank you for making this update.

upbqdn (approved):

PR #9522, opened 2025-05-12
Transaction and block deserialization

idky137:
Adds deserialisation to block and transaction structs in zebra for consumption by Zaino.

Fixes several serialisation incompatibilities between zcashd and zebrad.

## Motivation
Currently Zaino uses serialisation impls from zebra but must manually implement `serde::Deserialize` as zebra does not derive this. We would like to use a single source for both serialisation and deserialisation for these structs to avoid changes being made one way but not the other.

## Solution
Adds serde deserialisation to block and transaction structs in zebra that already implement `serde::Serialize`.

Adds deserialisation to `opthex` mod and exposes as pub for consumption by zaino (this code has been in zaino for some time but it would be good to unify).

Adds `arrayhex` serialisation mod for arrays of non standard lengths.

Fixes serialisation incompatibilities in transparent input fields (https://zcash.github.io/rpc/getrawtransaction.html):
\- script_sig: Added serde rename "scriptSig".
\- value_zat: Changed serde rename to match zcashd and zcash RPC spec "valueSat".

Removes double `#[serde(with = "hex")]` declaration for ScriptSig hex field. 

### Tests
Adds `zebra-chain::sapling::commitment::value_commitment_hex_roundtrip` to check `NotSmallOrderValueCommitment` FromHex impl.

NA

### Follow-up Work
NA

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/sapling/commitment.rs
+++ b/zebra-chain/src/sapling/commitment.rs
@@ -3,7 +3,7 @@
 use std::{fmt, io};
 
 use bitvec::prelude::*;
-use hex::ToHex;
+use hex::{FromHex, FromHexError, ToHex};
 use jubjub::ExtendedPoint;
 use lazy_static::lazy_static;
 use rand_core::{CryptoRng, RngCore};
@@ -387,6 +387,23 @@ impl ToHex for &NotSmallOrderValueCommitment {
     }
 }
 
+impl FromHex for NotSmallOrderValueCommitment {
+    type Error = FromHexError;
+
+    fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, Self::Error> {
+        // Parse hex string to 32 bytes
+        let mut bytes = <[u8; 32]>::from_hex(hex)?;
+        // Convert from big-endian (display) to little-endian (internal)
+        bytes.reverse();
+
+        // Try to construct ValueCommitment from raw bytes
+        let vc = ValueCommitment::try_from(bytes).map_err(|_| FromHexError::InvalidStringLength)?;
+
+        // Check small order and wrap
+        vc.try_into().map_err(|_| FromHexError::InvalidStringLength)
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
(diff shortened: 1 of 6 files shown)


zebra-chain/src/sapling/commitment.rs
```diff
@@ -387,6 +387,23 @@ impl ToHex for &NotSmallOrderValueCommitment {
...
+        // Parse hex string to 32 bytes
+        let mut bytes = <[u8; 32]>::from_hex(hex)?;
+        // Convert from big-endian (display) to little-endian (internal)
+        bytes.reverse();
+
+        // Try to construct ValueCommitment from raw bytes
+        let vc = ValueCommitment::try_from(bytes).map_err(|_| FromHexError::InvalidStringLength)?;
+
+        // Check small order and wrap
+        vc.try_into().map_err(|_| FromHexError::InvalidStringLength)
```
conradoplg: Just to avoid repeating the logic

```suggestion
        Self::zcash_deserialize(Cursor::new(&bytes)).map_err(|_| FromHexError::InvalidStringLength)
```

arya2 (commented):
This is looking good. It would be nice to put these derive attributes and impls behind a compilation feature for clients while it's still obvious which ones aren't needed for tests, but that could be done in a follow-up PR too.

conradoplg (commented):
This looks good, thanks! I left a suggestion.

Any particular reason for this being a draft still?

idky137:
Thanks! I have committed the suggested change.

We had a couple errors in the downstream PR (https://github.com/zingolabs/zaino/pull/320) so I wanted to make sure I hadn't introduced any bugs in this code before changing the status. 

We now have everything working so once this lands we can update Zaino to rely on lastest Zebra and LibrustZcash.

idky137:
Im happy to put these changes behind a "serde_deserialization" feature if that's preferable? I know some people don't like lots of features so I wasn't sure whether to use one or not.

conradoplg:
I don't think we need to feature-gate this

ala-mode:
I am having a compilation failure on [d3414ed](https://github.com/ZcashFoundation/zebra/pull/9522/commits/d3414edf6535f2cf072c25b62afae326ffd649d1)

use of undeclared type `Cursor` ---> zebra-chain/src/sapling/commitment.rs:399:33

AloeareV:
I don't have push permission for this branch, I opened https://github.com/idky137/zebra/pull/2 to add the `io::`

conradoplg (approved):

PR #9525, opened 2025-05-14
fix(security): Simulate template proposal when txs are added to the mempool

### LINKED ISSUE #9301: DoS vulnerability in `getblocktemplate` RPC
A mining pool can call the `getblocktemplate` RPC in two modes: _template_ and _proposal_. In the _template_ mode, the RPC returns a block template, which the mining pool then uses to mine a block. This template contains transactions from Zebra's mempool validated by Zebra's transaction verifier. In the _proposal_ mode, the pool submits a block for validation. Zebra validates this block by applying the block verifier. The block verifier calls the transaction verifier to verify transactions and can perform additional checks on them. These additional checks can lead to a DoS: If a transaction passes the transaction verifier but not the block verifier, this transaction will make it to a block template, but when a pool mines a block based on this template, Zebra will refuse it. If all mining pools use Zebra, and an attacker keeps submitting such problematic transactions to the mempool, no mining pool will be able to produce a block despite producing a valid PoW, and the blockchain will stall. A stuck blockchain has negative financial, mental, and cardiovascular effects on everyone involved.

This issue surfaced on NU6 activation. Some wallets were still producing transactions with outdated consensus branch IDs, and due to a subtle bug in the transaction verifier, these transactions appeared in Zebra's mempool. One of our mining tests then started reporting invalid block proposals because the block verifier was refusing them. If a mining pool had been using Zebra, this pool would not have been able to produce blocks.

One possible solution would be to implement the following invariant:

_Zebra applies the same validation checks in all modes of the `getblocktemplate` RPC._

Then, if there's a problematic transaction in the block template, Zebra drops it from the mempool, logs the error, and the RPC attempts to produce a new template.

## Possible Design

See https://github.com/ZcashFoundation/zebra/compare/main...mempool-block-proposals-design

oxarbitrage:
## Motivation

Attempt 1.

This is a draft as i am unsure if it is a good idea.

https://github.com/ZcashFoundation/zebra/issues/9301

## Solution

- Use `validate_block_proposal()` in template mode.

### PR Checklist

- [ ] The solution is tested.


zebra-rpc/src/methods.rs
```diff
@@ -2226,6 +2226,24 @@ where
...
         );
 
+        // - Validate the block template
+        let network_upgrade = NetworkUpgrade::current(&network, next_block_height);
+
+        let block = proposal_block_from_template(&response, None, network_upgrade)
+            .map_err(|error| ErrorObject::owned(0, error.to_string(), None::<()>))?;
+        let block_bytes = block
+            .zcash_serialize_to_vec()
+            .map_err(|error| ErrorObject::owned(0, error.to_string(), None::<()>))?;
```
arya2: If the check fails, Zebra should try to remove the invalid transactions. One approach for that could be to map every non-coinbase transaction in the invalid block to a proposal block, send all of them to the block verifier, and remove any transactions that were in an invalid proposal block.

It would be nice to factor out block deserialization from `validate_block_proposal()` and add a config flag or ideally a parameter field for skipping this check (if I were a miner, I would want to defer this check until after getting the template, submitting it as a proposal, and seeing that it failed).

upbqdn: This looks good to me. 

We should also add comments documenting why we need this check here.

I'm not sure what the best way to remove the problematic transactions would be, but I'm happy to look into it.

upbqdn: > (if I were a miner, I would want to defer this check until after getting the template, submitting it as a proposal, and seeing that it failed)

If the deferred check fails, it most likely means there's a fatal bug that caused Zebra to produce an invalid block template, on which the pool has wasted a lot of resources, so I don't think we should allow deferring the check until the block is mined.

Also, the bug has the potential to stall block production entirely, so we should log a huge error and ask for reporting.

arya2: > I don't think we should allow deferring the check until the block is mined.

Not until the block is mined, I think the default behaviour should be to validate the block, but it would be nice to have a parameter that tells it to skip it so that miners can get the block template and immediately submit it as a proposal ~200ms later while optimistically mining for that fraction of a second.

upbqdn: I see. Sorry for the confusion.

oxarbitrage: Still working on it, but I pushed an initial approach to remove transactions from the mempool: https://github.com/ZcashFoundation/zebra/pull/9525/commits/64e67df34b06c599f9f26fd1b729c913f82c83aa

I’d like to add some tests, but first I want to confirm the intended behavior with you. Right now, this assumes that only one transaction is invalid and stops after removing the first one that makes the block valid. We might want to handle other cases (e.g. multiple invalid transactions) or take a different approach altogether, open to feedback.

zebra-rpc/src/methods.rs
```diff
@@ -2215,18 +2215,110 @@ where
...
+                                tx_id
+                            );
+
+                            // Remove the transaction from the mempool
+                            let _ = mempool
+                                .clone()
+                                .oneshot(mempool::Request::RemoveTransaction(tx_id))
+                                .await;
+
+                            return Ok(response.into());
```
arya2: I think this is better than what I had in mind, which was to leave the transactions in the mempool and filter them out in this function.

This code should wait for all of the proposal results, remove transactions in the mempool that were in invalid blocks rather than those in valid blocks, and then create a new block template with the transactions that were originally fetched from the mempool without the ones that formed invalid block proposals.

It also needs to consider transactions dependencies, otherwise TEX transactions will never pass this validation step.

It might be better to do this in the mempool instead, immediately after adding transactions to the verified set (so it'll still be auto-verified by the transaction verifier when re-submitted as block requests) but before gossiping to peers or returning those transactions for block templates.

zebra-rpc/src/methods.rs
arya2: It needs some small adjustments/additions, but the idea is better than what I had in mind. 

> We might want to .. take a different approach altogether

I think it would be even better to do this validation in the mempool by adding another field [here](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/src/components/mempool.rs#L100) for block proposal verification so the mempool can remove transactions from its verified set on its own and ahead of block template generation if they create invalid block proposals, see [this comment](https://github.com/ZcashFoundation/zebra/pull/9525/files#r2101080501).

Then, any transactions that the `getblocktemplate` method fetches from the mempool will be known to form valid block proposals, and those that haven't been verified to produce valid block proposals yet will be excluded until they are verified (and with long polling, the miner will be notified when those are ready).

oxarbitrage:
This wil lbe replaced by https://github.com/ZcashFoundation/zebra/pull/9645

PR #9536, opened 2025-05-16
zebra-rpc: Correctly set optional `scriptPubKey` fields of transactions

### LINKED ISSUE #9535: zebra-rpc: `ScriptPubKey` fields are not correctly optional
The `zebrad` server response always includes all of these fields:
https://github.com/ZcashFoundation/zebra/blob/cc062510a2ba8f6931dc91aff08105a3dd414687/zebra-rpc/src/methods/types/transaction.rs#L249-L266

However the client side needs to support their optionality. Per [the reference implementation in `zcashd`](https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/core_write.cpp#L124-L149), the following fields may be omitted in the response:
- `reqSigs` (if no addresses can be extracted from the script)
- `addresses` (if no addresses can be extracted from the script)

(The `hex` field is also optional in the internal `ScriptPubKeyToUniv` method, but all callsites of `ScriptPubKeyToUniv` hardcode `fIncludeHex = true` and thus it is non-optional in those RPCs.)

str4d:
Closes ZcashFoundation/zebra#9535.

## Motivation

The current `ScriptPubKey` RPC type does not match the output of `zcashd`'s `ScriptPubKeyToUniv`, which breaks Zaino when using `zebra-rpc` as a client connected to `zcashd`.

Additionally, the `zcashd` semantics of `ScriptPubKeyToUniv` mean that the `addresses` and `reqSigs` fields are omitted when the script is non-standard (and thus "required sigs" is a potentially meaningless concept, and also potentially incorrect to set to 0).

## Solution

Does what it says on the tin: alters the implementation to match `zcashd`.

### Tests

Tested in Zallet, and I haven't reproduced the error I was getting before this PR (about the `reqSigs` field missing in the JSON-RPC response).

https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/core_write.cpp#L124-L149

### Follow-up Work

I found some required fields being set to empty strings, and added TODO comments for them so that it is obvious in the source.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods/types/transaction.rs
+++ b/zebra-rpc/src/methods/types/transaction.rs
@@ -257,12 +257,16 @@ pub struct ScriptPubKey {
     hex: Script,
     /// The required sigs.
     #[serde(rename = "reqSigs")]
-    req_sigs: u32,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    req_sigs: Option<u32>,
     /// The type, eg 'pubkeyhash'.
     // #9330: The `type` field is not currently populated.
     r#type: String,
     /// The addresses.
-    addresses: Vec<String>,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    addresses: Option<Vec<String>>,
 }
 
 /// The scriptSig of a transaction input.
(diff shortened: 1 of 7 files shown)


zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -434,19 +438,30 @@ impl TransactionObject {
...
                     .map(|output| {
-                        let addresses = match output.1.address(network) {
-                            Some(address) => vec![address.to_string()],
-                            None => vec![],
-                        };
+                        // Parse the scriptPubKey to find destination addresses.
+                        let (addresses, req_sigs) = match output.1.address(network) {
+                            // TODO: For multisig scripts, this should populate `addresses`
+                            // with the pubkey IDs and `req_sigs` with the n required to
+                            // spend.
```
oxarbitrage:
```suggestion
                            // with the pubkey IDs and `req_sigs` with the number of
                            // signatures required to spend.
```

oxarbitrage (commented):
Looks good, happy to merge but i think it will break some snapshot tests. Waiting on the CI to confirm.

oxarbitrage:
As expected, there is a snapshot error (https://github.com/ZcashFoundation/zebra/actions/runs/15076234919/job/42384170453?pr=9536). You can run the tests locally and it will generate a new snapshot file or let me know if you want me to push a commit for it.

str4d:
[Force-pushed](https://github.com/ZcashFoundation/zebra/compare/2b634c5e42b790fc54ca2c5a1b5f5de399ca41e4..2536cfdd1cb1a53fde07b4c92b3987ef0d996c57) to address review comment and update snapshots.

oxarbitrage (approved):
thanks!

PR #9539, opened 2025-05-21
change(state): Upgrade db format to support new fields in RPC outputs

### LINKED ISSUE #5449: getblock: return the `valuesPools` and `chainSupply` fields

arya2:
## Motivation

This PR merges the db format upgrades in #9295 and #9432 so that Zebra will only need to iterate over transaction data once in the next release when migrating to the latest db format.

## Review

It's likely easier to review this PR by commit, all of the state changes from the origin PRs are in https://github.com/ZcashFoundation/zebra/pull/9539/commits/1ce3eb7d016544c42d69f756bde68bb3ebf107f2, and the significant changes to the format upgrade are in https://github.com/ZcashFoundation/zebra/pull/9539/commits/d3f0c18b7bc260fb2e12e2ab97da620f0c957300.

## Solution

- Merges the changes to zebra-state from both PRs
- Removes a mechanism for freezing block commits until some in-place format upgrades are complete now that it's using a merge operator and the upgrade doesn't need to block (it would be better to do this with a version of the `copy-state` command instead if we do need blocking db format upgrades later),
- Updates the (now-renamed) `add_block_info` db format upgrade to also update the `balance_by_transparent_addr` column family with funds received by transparent addresses in the processed blocks.

## References

https://github.com/facebook/rocksdb/wiki/Merge-Operator

### Follow-up Work

Update the `getaddressbalance` and `getblock` RPCs.

### PR Checklist

- [x] The solution is tested.


zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -225,18 +225,144 @@ pub struct AddressBalanceLocation {
...
 }
 
+/// Represents a change in the [`AddressBalanceLocation`] of a transparent address
+/// in the finalized state.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(
+    any(test, feature = "proptest-impl"),
+    derive(Arbitrary, Serialize, Deserialize)
+)]
+pub struct AddressBalanceLocationChange {
```
arya2: This code is highly redundant, suggestions are welcome.

At worst, `AddressBalanceLocation` could be refactored to take a generic param and this could be replaced with a newtype (wrapper) or a type alias.


Update: Renaming it to `AddressBalanceLocationInner`, making `AddressBalanceLocation` a newtype, adding another newtype for `AddressBalanceLocationChange`, and adding deref impls for accessing the common methods.

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/fa9aa368a71b08bbab32f22a925d3895a8c026d4.

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
```diff
@@ -31,23 +32,45 @@ use crate::{
...
+/// A RocksDB merge operator for the [`BALANCE_BY_TRANSPARENT_ADDR`] column family.
+pub fn fetch_add_balance_and_received(
+    _: &[u8],
+    existing_val: Option<&[u8]>,
+    operands: &rocksdb::MergeOperands,
+) -> Option<Vec<u8>> {
+    // # Correctness
+    //
+    // Merge operands are ordered, but may be combined without an existing value in partial merges, so
+    // we may need to return a negatove balance here.
```
conradoplg:
```suggestion
    // we may need to return a negative balance here.
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -870,7 +914,18 @@ impl DiskDb {
...
+
+                if cf_name == BALANCE_BY_TRANSPARENT_ADDR {
+                    cf_options.set_merge_operator_associative(
+                        BALANCE_BY_TRANSPARENT_ADDR_MERGE_OP,
+                        fetch_add_balance_and_received,
+                    );
+                }
+
+                rocksdb::ColumnFamilyDescriptor::new(cf_name, cf_options.clone())
+            });
```
conradoplg: Optional: might be helpful to refactor the CF creation into a separate function to remove this duplicated code

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -1346,6 +1363,25 @@ impl Chain {
...
+    /// Returns the transparent received balance change for `addresses` in this non-finalized chain.
+    ///
+    /// If the received balance doesn't change for any of the addresses, returns zero.
+    ///
+    /// # Correctness
+    ///
+    /// Callers should apply this received balance change to the finalized state received balance for `addresses`.
+    ///
+    /// The total received balance will only be correct if this partial chain matches the finalized state.
+    /// Specifically, the root of this partial chain must be a child block of the finalized tip.
```
conradoplg: Isn't this always the case?

zebra-state/src/service/read/address/balance.rs
```diff
@@ -123,7 +123,7 @@ fn chain_transparent_balance_change(
     // If we've already committed this entire chain, ignore its balance changes.
     // This is more likely if the non-finalized state is just getting started.
     if chain_tip < required_chain_root {
-        return Amount::zero();
+        return (Amount::zero(), 0);
     }
```
conradoplg, line 127: Is this even possible?

conradoplg (commented):
I'l halfway through the review, I'm flushing the comments I have, but it's looking good!

zebra-state/src/service/read/address/balance.rs
conradoplg: Ah, now I see the code was already there. I'll take a closer look to understand what is going on

arya2: It seems possible, but it would be unexpected/unusual. If reading from the db blocks execution for too long, the latest non-finalized state could be outdated. Or if something goes horribly wrong with the async executor.

zebra-state/src/service/non_finalized_state/chain.rs
arya2: It's currently always the case for partial transfers, chain work, revealed nullifiers, and subtrees, but not for other fields on `Chain`.

Update: It is verbose. Merged the method's logic into the existing `partial_transparent_balance_change()` in  https://github.com/ZcashFoundation/zebra/pull/9539/commits/9d410123639bd0b947ebbecc0c1a2fd39d2b9003 so it at least isn't repeated unnecessarily.

~I wanted to implement `Add` for the tuple so it we could still use `Sum` but the received amount should probably go in a newtype before that happens and I'd rather defer doing that.~

Update: On second thought, the current code seems more clear and concise than using `Sum` would be anyway.

Question: zcashd uses `CAmount` for the received value, an i64 that may also be constrained to `MAX_MONEY`, should Zebra do the same? I only checked because I thought they might be using a `u128`, it seems like a strange choice to constrain that further than its size does. Maybe we could just clamp it in the RPC method for now if it needs to match zcashd's output exactly.

zebra-state/src/service/finalized_state/zebra_db/transparent.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/5a07263b03b2e738c6cc745f8f526ae53b7d93fc.

zebra-state/src/service/finalized_state/disk_db.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/58b9254a91468edaadbea94d94ab91f0dae4973b.

zebra-chain/src/amount.rs
```diff
@@ -538,6 +538,10 @@ impl Constraint for NegativeAllowed {
 /// );
 /// ```
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Default)]
+#[cfg_attr(
+    any(test, feature = "proptest-impl"),
+    derive(proptest_derive::Arbitrary)
+)]
 pub struct NonNegative;
```
arya2, line 545: Is there a way around this? There's an `Arbitrary` impl for `Amount<C: Constraint + Debug>`, it seems like it shouldn't be necessary to derive this explicitly here.

Update: I agree it's not a big deal.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -207,80 +207,199 @@ impl OutputLocation {
...
-    /// Creates a new [`AddressBalanceLocation`] from the location of
+impl<C: Constraint + Copy + std::fmt::Debug> AddressBalanceLocationInner<C> {
+    /// Creates a new [`AddressBalanceLocationInner`] from the location of
     /// the first [`transparent::Output`] sent to an address.
     ///
-    /// The returned value has a zero initial balance.
-    pub fn new(first_output: OutputLocation) -> AddressBalanceLocation {
-        AddressBalanceLocation {
+    /// The returned value has a zero initial balance and received balance.
+    pub fn new(first_output: OutputLocation) -> Self {
```
arya2:
```suggestion
    fn new(first_output: OutputLocation) -> Self {
```

zebra-chain/src/amount.rs
conradoplg: I've been bitten by this in `frost`. I never fully understood the reason, but sometimes you have to implement a trait in marker structs if you derive that trait in a type that uses the struct - in this case `AddressBalanceLocationInner`. If you manually implement `Arbitrary` for `AddressBalanceLocationInner` then maybe you can avoid this. But it's not a big deal I think.

conradoplg (commented):
I think I'm ready to approve this, but I'd like to write down some of my reasoning for you to double check if we missed something:

- This is safe against the node being shutdown mid-upgrade, because both the block info and balance for a block are written in a batch, and the existence of the block info will enable skipping already upgraded blocks (I guess combining both upgrades made the received one easier!)
- While the upgrade is running, committed blocks will update the balance even if the balance at that time is not up-to-date. This is fine since balances are additive, and when the upgrade finishes, the balance will be correct.
- The non-finalized logic is correct because it basically sums the balance for each block on the fly, and the caller is responsible for adding that to the finalized balance

Also, is the plan to add the RPC part itself in another PR? If so, it would be good to have it ready before we merge this, to help test this PR with `zcash-rpc-diff`.

arya2:
> - This is safe against the node being shutdown mid-upgrade, because both the block info and balance for a block are written in a batch, and the existence of the block info will enable skipping already upgraded blocks (I guess combining both upgrades made the received one easier!)
> - While the upgrade is running, committed blocks will update the balance even if the balance at that time is not up-to-date. This is fine since balances are additive, and when the upgrade finishes, the balance will be correct.
> - The non-finalized logic is correct because it basically sums the balance for each block on the fly, and the caller is responsible for adding that to the finalized balance

Yep, that's all right.

> Also, is the plan to add the RPC part itself in another PR? If so, it would be good to have it ready before we merge this, to help test this PR with `zcash-rpc-diff`.

It was, the RPC is updated in #9295, I'll rebase that on this branch now.

arya2:
This still needs some cleanup and minor bug fixes. 

Recent changes:
- Bumped the major version (thank you @conradoplg for figuring out that that was necessary), and
- Updated the upgrade logic to apply the migration (needs some cleanup).

Investigating why a migration validation check is failing and rebasing the PR with the RPC changes next.

arya2:
@conradoplg the changes for https://github.com/ZcashFoundation/zebra/pull/9295 have been re-applied atop this branch now.

This may be a good candidate for admin merging. We could trigger a manual full sync test in CI first, but it still won't test the in-place upgrade which needs an old cache to update, it's probably easiest to test that locally.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -82,21 +83,39 @@ pub trait DiskFormatUpgrade {
...
+            let version = upgrade.version();
+            if version.minor == 0 && version.patch == 0 {
+                Some(version.major)
+            } else {
+                None
+            }
+        })
+        .collect()
+}
+
```
conradoplg: Not sure about this, this indicates that every x.0.0 version is restorable, but I don't think this is necessarily true for < 26

zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -647,35 +766,69 @@ impl FromDisk for OutputLocation {
...
-            disk_bytes.as_ref().split_at(BALANCE_DISK_BYTES);
+        let (balance_bytes, rest) = disk_bytes.as_ref().split_at(BALANCE_DISK_BYTES);
+        let (address_location_bytes, received_bytes) = rest.split_at(BALANCE_DISK_BYTES);
 
         let balance = Amount::from_bytes(balance_bytes.try_into().unwrap()).unwrap();
         let address_location = AddressLocation::from_bytes(address_location_bytes);
+        // # Backwards Compatibility
+        //
+        // If the value is missing a `received` field, default to 0.
+        let received = u64::from_le_bytes(received_bytes.try_into().unwrap_or_default());
```
conradoplg, line 811: I think it would be nice to make this also forwards compatible, so that if `received_bytes` is larger than a u64 (i.e. a new field was added after it in a new database upgrade but zebra was downgraded) then this would still parse the value

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -828,6 +847,7 @@ impl Drop for DbFormatChangeThreadHandle {
 #[test]
 fn format_upgrades_are_in_version_order() {
     let mut last_version = Version::new(0, 0, 0);
+    // The particular network shouldn't matter for the test; use Mainnet
```
conradoplg:
```suggestion
```

Comment is now moot

conradoplg (commented):
I still need to do some manual testing but it's looking good, flushing some additional comments

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: We could add `&& upgrade.needs_migration()` or a new trait method?

I'd prefer it not to be defined in two places, but it's not terribly important if we want to go back to that.

Updated in https://github.com/ZcashFoundation/zebra/pull/9539/commits/e6c4f5e7867434e84728f04bbbfcbfc5ef00c957 to add a new trait method that checks the minor/patch versions + `needs_migration()` by default.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
arya2: Removed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/e6c4f5e7867434e84728f04bbbfcbfc5ef00c957.

zebra-state/src/service/finalized_state/disk_format/transparent.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/e6c4f5e7867434e84728f04bbbfcbfc5ef00c957.

zebra-state/src/service/finalized_state/disk_format/upgrade.rs
conradoplg: Sorry, my original comment was misleading.

The 26.0.0 upgrade was different in the sense that, unlike other major DB upgrades, didn't need to sync from scratch and could reuse the 25.x.x state. The 27.0.0 upgrade is the same. Those are the "restorable" upgrades, but the only ones, the other ones (e.g. 25, 24, 23) etc. aren't restorable AFAIK.

My point was that `restorable_db_versions()` would return all x.0.0 database upgrades regardless if they are explicitly "restorable" or not. I didn't realize though that currently that does not make any difference since the only major database upgrades in `format_upgrades()` are 26 and 27 anyway so the returned value is correct.

What could happen in the future though is that we could introduce a new non-restorable database upgrade which would be incorrectly returned in `restorable_db_versions()` due to simply being in the x.0.0 format. However that seems unlikely.

At the end we need to decide if we consider all major database upgrades "restorable" by default or whether we explicitly list them like the original code. I think I don't really mind either way.

That being said, this last fix I think is not correct - `restorable_db_versions()` is now returning just 27 but it should also return 26, because for 26 `needs_migration()` returns false but it still is a "restorable" update.

TL;DR I think you can restore your original code here, sorry for the confusion

arya2: > I think is not correct - restorable_db_versions() is now returning just 27

Yeah, oops, thank you for noticing that. Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/bc03d0b05fa445cb0d2bf48730dc6f2edb7f1c37.

It may be better to add a boolean field to `NoMigration` indicating whether the new format version can reuse the previous major version, but Zebra can't currently reuse any previous major format except the one just before the current major format version anyway, and I can't think of a circumstance where we would have an upgrade to a major version without it being able to reuse the previous major format version.

conradoplg (changes requested):
I think there is an issue now - if you run Zebra from a 26.0.0 state and then close it before the upgrade completes, it panics after logging

```
2025-05-27T20:52:09.034826Z ERROR zebra_state::service::finalized_state::disk_format::upgrade: err=Err("invalid detailed check: [Ok(()), Ok(()), Ok(()), Ok(()), Ok(()), Err(\"missing block info for height: 2937809\")]")
```

I think the cause is that the DB is immediatally marked as 27.0.0 after the state is copied from v26 (not sure though), but that should happen only when the upgrade completes. I guess the 26.0.0 upgrade worked because it did not require an upgrade process.

Not sure what the best solution is...

arya2:
It would do that.

We could write the db format version in the parent directory or in a rocksdb column family?

I'm not sure which is better or if there's another alternative to consider.

conradoplg:
> It would do that.
> 
> We could write the db format version in the parent directory or in a rocksdb column family?
> 
> I'm not sure which is better or if there's another alternative to consider.

Would it make sense to have two upgrades, one for simply copying over 26 to 27 (exactly like 25 to 26), and another (27.1.0?) to do the migration?

arya2:
> Would it make sense to have two upgrades, one for simply copying over 26 to 27 (exactly like 25 to 26), and another (27.1.0?) to do the migration?

~Yeah, let's get this in and I'll fix the issue more elegantly in a follow-up PR.~

Update: Ah, fixing it in a follow-up PR could mark the disk version as 27.1.0 in the cached states. I'll fix it here (the easy solution still leaves edge cases).

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9539/commits/0a5ea383f5d779087e2d2256785dbcb9ab249eca by writing major versions to the disk version file.

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -1053,20 +1085,29 @@ impl DiskDb {
...
-                            Err(e) => {
-                                warn!("could not remove version file at {new_path:?}: {e}")
-                            }
-                        }
+                        let mut disk_version =
+                            database_format_version_on_disk(config, db_kind, major_db_ver, network)
+                                .expect("unable to read database format version file")
+                                .expect("unable to parse database format version");
+
+                        disk_version.major = old_major_db_ver;
```
conradoplg, line 1102: In the original behaviour, the version file was removed, making the effective version x.0.0.

However, this no longer deletes the file, which keeps the old minor and patch versions, which might not be 0.0 as we want here.

conradoplg (changes requested):
I'm a bit wary of changing the format of the version file (could there be somewhere in the code that uses the db path to get the major version instead of the version file?). Everything seems to work though so I think we can merge this, but I think there might be an small issue which I've commented on

zebra-state/src/service/finalized_state/disk_db.rs
arya2: This was deliberate, though it's not currently necessary or useful because there aren't any minor or patch db format upgrades after v26.0.0 and Zebra can't currently re-use an older major version. It _will_ be useful when either https://github.com/ZcashFoundation/zebra/issues/9565 is fixed or adding a new major format version after adding minor/patch upgrades that need migrations to the v27.

> could there be somewhere in the code that uses the db path to get the major version instead of the version file

I think I double-checked, but let me triple-check that, we don't want to miss anything.

<details>

<summary>Background:</summary>

I wanted to avoid removing the db format version file [here](https://github.com/ZcashFoundation/zebra/pull/8748#discussion_r1710117328), it was somewhat-okay to make the effective disk format version `x.0.0` when reusing the major db format version for v26 (for the value balance upgrade) because the only supported versions of Zebra at the time were using the latest v25 db format, but even then it could've caused problems for users that had an older cache, and certainly now causes problems for users that interrupt the migration after reusing a previous major version of the db.

</details>

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -984,23 +1012,27 @@ impl DiskDb {
     /// db to a new path so it can be used again. It does so by merely trying to rename the path
     /// corresponding to the db version directly preceding the current version to the path that is
     /// used by the current db. If successful, it also deletes the db version file.
+    ///
+    /// Returns the old disk version if one existed and the db directory was renamed, or None otherwise.
+    #[allow(clippy::unwrap_in_result)]
```
arya2, line 1018:
```suggestion
    // TODO: Update this function to rename older major db format version to the current version (#9565).
    #[allow(clippy::unwrap_in_result)]
```

zebra-state/src/service/finalized_state/disk_db.rs
conradoplg: Ah, I see, I misinterpreted it. Say the version is 26.1.0, this piece of code would copy the DB to the `v27/` folder but _keep the version_ 26.1.0, which is the right thing to do.

Sorry for insisting, but do you really think adding two database upgrades at the same release wouldn't be easier? i.e.

```
        Box::new(no_migration::NoMigration::new(
            "add value balance upgrade",
            Version::new(26, 0, 0),
        )),
        Box::new(no_migration::NoMigration::new(
            "prepare for block info and address received upgrade",
            Version::new(27, 0, 0),
        )),
        Box::new(block_info_and_address_received::Upgrade),
```

Then it would first move the db to `v27` while keeping the old format, which prevents downgrades, and then would actually upgrade to e.g. 27.1.0.

arya2: > do you really think adding two database upgrades at the same release wouldn't be easier?

I think it would be easier, but that the difference in complexity is trivial while the drawbacks of adding two upgrades are significant.

I prefer this approach, writing the major format version in the version file, because it would:
- preserve more intuitive behaviour and avoid requiring a major version bump in this case or possible future cases that represent no real change to the db format,
- allow Zebra to reliably re-use and migrate more major versions and, after fixing [#9565](https://github.com/ZcashFoundation/zebra/issues/9565), any past restorable cache, which I was hoping for when we first allowed restorable major db format versions, and
- simplify our next major format upgrade after the minor upgrade for tracking ZSA asset states.

I don't think any of those considerations are currently a priority as none of them help accelerate zcashd deprecation, so I would still go with the simpler approach if it weren't for the difference in complexity seeming trivial. It does require some care, but the code around reading/writing the version file is cleanly factored, so it feels safe to me.

I'll defer to you for the final decision on which approach to take here.

<details>

<summary>Background</summary>

Using two upgrades as in your sample code was my first thought too, it would be a bit easier than the current approach, and far easier than our first two ideas.

I listed it as a rejected alternative in a TODO until your suggestion because it required a major version bump with no real change to the format. After your suggestion I was thinking we could document that extensively to ensure what's happening is obvious to developers, and that still seemed better than the complexity of moving the version file (which is worse than I initially thought) or adding an entry in the db to mark an incomplete migration.

I made and pushed that change at first, intending to fix it in a follow-up PR, but then realized it would upgrade our cached states on main if it had merged so I removed the commit and force pushed.

</details>

arya2: I double-checked that:
- no direct or indirect callers of `db_path()` are using the path to get the major disk format version, and
- all references to the major version in code and all callers of `version_file_path()` are correct.

`database_format_version_at_path()` will return the major version in code with a minor/patch version of `0` if no disk version file is found, but it could already misbehave if a version file is missing when a db already exists.

Regardless of which approach we take in this PR, we may want to update the logic there to ensure that there is no existing db cache when the version file is missing, or to delete the db cache if the version file is missing. Then we could also use the `state_database_format_version_in_code()` as the default version when no disk version file exists.

I'd feel safer updating CI to use cached states with previous restorable major db versions with this change but we may want to consider doing that either way too.

conradoplg: This makes sense, thank you for writing this down!

conradoplg (approved):

PR #9550, opened 2025-05-27
rpc: add deserialization tests

### LINKED ISSUE #9549: feature: make `zebra-rpc` support creating clients
### Motivation

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

Part of https://github.com/ZcashFoundation/zebra/issues/9549

Creates serialization tests for requests and deserialization tests for responses.

## Solution

I added tests for each RPC method.

I did some trivial changes (deriving `serde::Deserialize`, etc) to make some of the tests work.

Some tests will require non-trivial (but easy) changes which I'll do in a separate PR; this is why there are a lot of TODOs.

### Tests

For each RPC:

- If there is a non-trivial request type, I added a test that instantiates the struct, serializes it and checks if it matches a (manually generated, or copied from somewhere) test vector.
- If there is a non-trivial response type, I added a test that deserialized a test vectors, reads all the field, reconstructs the struct and checks if it matches the deserialized one. Some responses don't allow getting the fields or constructing the struct, which is what I'll deal with in the folllow up PR.

### Follow-up Work

In a follow-up PR I'll add missing `new()` methods, missing getters, and tests that use those. I might also refactor the struct naming which is all over the place.

### PR Checklist

- [ ] The solution is tested.


zebra-rpc/tests/vectors/mod.rs
```diff
@@ -0,0 +1,1708 @@
...
+  ],
+  "noncerange": "00000000ffffffff",
+  "sigoplimit": 20000,
+  "sizelimit": 2000000,
+  "curtime": 1747848159,
+  "bits": "1c01da67",
+  "height": 2931867,
+  "maxtime": 1747853232
+}
+"#;
```
arya2: It would be nice to put these in their own JSON files, maybe in `zebra-test` with the test block vectors:

```rust
pub const GET_BLOCKCHAIN_INFO_RESPONSE: &str =
    include_str!("getblockchaininfo_response.json").trim();

pub const GET_BLOCK_RESPONSE_2: &str = include_str!("getblock_response_2.json").trim();

pub const GET_RAW_TRANSACTION_RESPONSE_TRUE: &str =
    include_str!("getrawtransaction_response_true.json").trim();

pub const GET_BLOCK_TEMPLATE_RESPONSE_TEMPLATE: &str =
    include_str!("getblocktemplate_response_template.json").trim();
```

zebra-rpc/src/methods/types/unified_address.rs
```diff
@@ -3,23 +3,23 @@
...
-    #[serde(skip_serializing_if = "String::is_empty")]
-    orchard: String,
-    #[serde(skip_serializing_if = "String::is_empty")]
-    sapling: String,
-    #[serde(skip_serializing_if = "String::is_empty")]
-    p2pkh: String,
-    #[serde(skip_serializing_if = "String::is_empty")]
-    p2sh: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    orchard: Option<String>,
```
arya2, line 7: Why make these `Option`s rather than use the empty string as a null value?

zebra-rpc/tests/serialization_tests.rs
```diff
@@ -0,0 +1,763 @@
...
+  "protocolversion": 170120,
+  "blocks": 2930822,
+  "connections": 75,
+  "difficulty": 68556523.91969073,
+  "testnet": false,
+  "paytxfee": 0.0,
+  "relayfee": 1e-6,
+  "errors": "no errors",
+  "errorstimestamp": "2025-05-20 19:33:53.395307694 UTC"
+}"#;
```
arya2, line 48: We could also put all of these into one big json object under different keys.

Optional: Prop tests would be nice here.

zebra-rpc/tests/serialization_tests.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    assert_eq!(r, new_obj);
+
+    Ok(())
+}
+
+#[test]
+fn test_get_block_height_hash() -> Result<(), Box<dyn std::error::Error>> {
+    let json = r#"
+{"height":2931705,"hash":[35,5,244,118,21,236,8,168,3,119,95,171,238,9,233,152,250,106,153,253,6,176,155,7,155,161,146,1,0,0,0,0]}
```
arya2: Optional/Nitpick: Pretty-print formatting would be more readable and consistent.

zebra-rpc/tests/serialization_tests.rs
```diff
@@ -0,0 +1,763 @@
...
+"#;
+    let obj: GetBlockHeightAndHash = serde_json::from_str(json)?;
+    let new_obj = GetBlockHeightAndHash {
+        height: obj.height,
+        hash: obj.hash,
+    };
+
+    assert_eq!(obj, new_obj);
```
arya2, line 229: This assertion isn't really testing anything, could we do a round-trip serialization test instead?

zebra-rpc/src/methods/types/unified_address.rs
conradoplg: The end result is the same (making some fields optional), but with `Option` this also works when deserializing instead of just when serializing. With the previous code you couldn't parse a JSON with a missing field.

zebra-rpc/tests/vectors/mod.rs
conradoplg: Done in  11225f6  but I left the files in `zebra-rpc` - it won't be compiled into the crate since it's in the integration test folder. But I can move to `zebra-test` if you think that's better.

zebra-rpc/tests/serialization_tests.rs
conradoplg: I think a single huge JSON would be a bit unwieldy.

I agree with prop tests, the test coverage here isn't great, but I think it would take more time then we are willing to invest at this point

zebra-rpc/tests/serialization_tests.rs
conradoplg: Done in  11225f6

zebra-rpc/tests/serialization_tests.rs
conradoplg: The goal is to ensure clients can parse the JSON and read all the contents of the response. If we are able to reconstruct the object, then it means we have succeed that goal.

To complete the round-trip we'd need to serialize and canonicalize it, and then canonicalize the test vectors... But I think we already test serialization with the insta test vectors

arya2 (approved):

zebra-rpc/src/methods/types/unified_address.rs
arya2: Optional/Nitpick: Add default impls here instead of using an `Option`

arya2:
This looks good and makes sense. There's one remaining optional comment about adding a `Default` impl to a couple structs instead of changing field types, but it's not very important either way if you want to keep it.

conradoplg:
@mergifyio requeue

mpguerra:
@mergifyio requeue

mpguerra:
@mergifyio requeue

conradoplg:
@Mergifyio requeue

conradoplg:
@Mergifyio requeue

(checking to see if the error is deterministic or not...)

PR #9551, opened 2025-05-27
feat(rpc): Add `invalidateblock` and `reconsiderblock` RPC methods

### LINKED ISSUE #8843: Accept a new channel in  (or refactor the existing channels into a single channel where the message type is an enum)

### LINKED ISSUE #8844: Call the new `NonFinalizedState` methods when the block write task receives a message to invalidate or reconsider a block

### LINKED ISSUE #8845: Add a `InvalidateBlock` `StateService` `Request` variant which sends a message to the block write task

### LINKED ISSUE #8846: Add a `ReconsiderBlock` `StateService` `Request` variant which sends a message to the block write task

### LINKED ISSUE #8847: Add an `invalidate()` RPC method

### LINKED ISSUE #8848: Add a `reconsiderblock()` RPC method

### LINKED ISSUE #9265: Add a responder channel to invalidate/reconsider block requests

arya2:
## Motivation

Supersedes #9266.

Closes https://github.com/ZcashFoundation/zebra/issues/8848
Closes https://github.com/ZcashFoundation/zebra/issues/8847
Closes https://github.com/ZcashFoundation/zebra/issues/8846
Closes https://github.com/ZcashFoundation/zebra/issues/8845
Closes https://github.com/ZcashFoundation/zebra/issues/8844
Closes https://github.com/ZcashFoundation/zebra/issues/8843
Closes https://github.com/ZcashFoundation/zebra/issues/9265

## Solution

See #9266.

- Moves acceptance test logic out of `trusted_chain_sync_handles_forks_correctly` for simplicity and to avoid a conflict with #9494.
- Adds new response variants for `Invalidate` and `Reconsider` state requests
- Adds responder channels for the state to notify callers of the operations' results

## Review

Suggestions for the write block task refactor would be very welcome.

Let me know if a pair review would be helpful.

### PR Checklist

- [ ] The solution is tested.


zebrad/tests/acceptance.rs
```diff
@@ -3692,6 +3695,93 @@ async fn has_spending_transaction_ids() -> Result<()> {
...
+        let height = expected_block.coinbase_height().unwrap();
+        let zebra_block = rpc_client
+            .get_block(height.0 as i32)
+            .await
+            .map_err(|err| eyre!(err))?
+            .expect("Zebra test child should have the expected block");
+
+        assert_eq!(
+            zebra_block,
+            Arc::new(expected_block),
```
upbqdn, line 3751: If the compiler doesn't optimize the assert, then dereferencing the arc should be faster than allocating the expected block on the heap.

```suggestion
            *zebra_block, expected_block,
```

arya2:
Let me know if a pair review would be helpful for this one.

zebra-state/src/service/non_finalized_state/chain.rs
```diff
@@ -376,7 +376,7 @@ impl Chain {
         let block_height = self.height_by_hash(block_hash)?;
         let mut new_chain = self.fork(block_hash)?;
         new_chain.pop_tip();
-        new_chain.last_fork_height = None;
+        new_chain.last_fork_height = self.last_fork_height.min(Some(block_height));
```
upbqdn, line 387: Was this a bug, or what are the implications of this change?

arya2: It's used for the metrics bars, not much of a bug, but I think it would be weird to clear the last fork height when invalidating a block. This code is also considering the invalidated block height to be a fork height, I'm happy to change that if we want.

zebrad/tests/acceptance.rs
arya2: Good point, noted. I'd be happy to apply this, but since it's a test and the code is only called a handful of times in the test, I think the difference is insignificant and I'd prefer to merge this as-is if we don't need to make changes for the other comment.

upbqdn (dismissed):

upbqdn (approved):

PR #9554, opened 2025-05-27
tests(wallet): Add initial framework to test zallet RPC methods

### LINKED ISSUE #9296: Integrate zallet into zebra's `qa` tests
[Zallet](https://github.com/zcash/wallet) is the Rust-based replacement for the deprecated zcashd wallet. It will use  [Zaino](https://github.com/zingolabs/zaino), which in turn depends on Zebrad, to provide critical wallet functionality that zcashd previously supported.

We need to integrate zallet into the [Zebra QA tests](https://github.com/ZcashFoundation/zebra/tree/main/zebra-rpc/qa) so that we can run tests that previously relied on zcashd's wallet. This will allow us to decide which wallet tests should be ported and work on them separately in future tickets.

Scope:

- Add zallet as a dependency in the QA test framework.
- Enable running the zallet binary from the Python-based test suite.
- Validate the integration by porting a single test to ensure zallet is functioning as expected within the framework.

Blockers

- Missing zallet functionality, specifically the `getnewaddress` RPC method is used in almost all the tests.
- zallet has not yet been released on [crates.io](https://crates.io/crates/zallet), which prevents us from using it as a stable dependency. As a temporary workaround, we can use zallet as a GitHub dependency, but we won’t be able to release until the crates.io issue is resolved.

oxarbitrage:
## Motivation

We want to run wallet-related tests from Zebra using the new zebra-zaino-zallet stack. These tests are Python-based and were originally ported from zcashd. While we previously updated the test framework to use zebrad instead of zcashd, most of the tests require wallet functionality, which was not clearly defined until recently. With recent progress, we now have a clearer path for integrating these features.

Closes https://github.com/ZcashFoundation/zebra/issues/9296

## Solution

This PR is the foundation for running wallet RPC tests in the new stack:

- Build the zallet binary within Zebra when the env var is set.
- Add Python helpers to start wallet instances with default configurations and a regtest Rage identity.
- Add the initial version of the wallet.py test, based on [zcash/qa/rpc-tests/wallet.py](https://github.com/zcash/zcash/blob/master/qa/rpc-tests/wallet.py).

### Tests

Build Zebra with the `ZALLET` env var set:

```bash
ZALLET=1 cargo build
``` 

Then run the wallet test from inside the zebra-rpc directory:

```bash
./qa/pull-tester/rpc-tests.py wallet.py
```

### Follow-up Work

- Decide on license compatibility and file headers (https://github.com/ZcashFoundation/zebra/discussions/9556)
- Insert the miner address into the zallet database (One way to do it might be with https://github.com/ZcashFoundation/zebra/issues/9557 and https://github.com/zcash/wallet/pull/152)
- Implement addnode support in Zebra (https://github.com/ZcashFoundation/zebra/issues/9555)
- Implement the stop method in zallet for controlled shutdown in tests (https://github.com/zcash/wallet/issues/153)
- Implement getwalletinfo in zallet (https://github.com/zcash/wallet/issues/55)

### PR Checklist

- [ ] The solution is tested.


arya2 (approved):
Looks good, thank you!

PR #9568, opened 2025-06-02
rpc: API cleanup

### LINKED ISSUE #9549: feature: make `zebra-rpc` support creating clients

conradoplg:
## Motivation

Based on #9550
Closes #9549

## Solution

For each RPC:

- I made sure the request types are constructable and the responses are readable
  - By "readable" it means crate clients can read _all_ the fields, which entails making sure clients can obtain the "raw" value (e.g. if a commitment is returned, the client needs to obtain the bytes)
- I achieve that mostly by adding getters (using the derive-getters crate) and a `new()` method (using the derive-new crate)

I also started to "flatten" the API using `pub use` so that all types live under `zebra-rpc::methods`, with the final goal of marking `zebra-rpc::methods::types` not `pub` which hides the underlying structure of the code, which in turns allows us to freely move things around. However, halfway through it I wondered if it's really a good idea. Feedback is welcome and I'm happy to complete the process if that makes sense.

There were also some assorted adjustments and cleanups.

I tried to avoid breaking stuff where possible, so I added some deprecated aliases for old names.

### Tests

For each RPC, I added a test that extracts all the data and rebuilds it again to make sure everything is readable. It is a bit of a pain as you will see. I likely missed some stuff, I'm glad to fix anything you find but it is also something we can improve over time.

### PR Checklist

- [ ] The solution is tested.


arya2 (commented):
> with the final goal of marking zebra-rpc::methods::types not pub which hides the underlying structure of the code, which in turns allows us to freely move things around. However, halfway through it I wondered if it's really a good idea

It seems like a good idea. I opened an incomplete suggestion/idea PR (https://github.com/ZcashFoundation/zebra/pull/9585) with breaking changes.

Full names are easier to search for in the docs or directly in the text (and I suspect they may also be easier for rust-analyzer to find), but they are _very_ verbose.

mpguerra:
@conradoplg is this ready for another review by @arya2 ?

conradoplg:
> @conradoplg is this ready for another review by @arya2 ?

Yes!

arya2:
This needs some merge conflict resolutions, probably best to handle those after #9432 has merged.

mpguerra (approved):

PR #9571, opened 2025-06-03
fix(test): update lightwalletd log regex patterns for JSON output format

gustavovalverde:
## Motivation

The latest lightwalletd version now outputs structured JSON logs instead of plain text messages. Updated regex patterns in lightwalletd_integration_test to match the new JSON format by looking for messages within the "msg" field.

Previously expected: `"Got sapling height 419200 block height..."`
Now expects: `"msg":"Got sapling height 419200 block height..."`

This change ensures the acceptance tests pass with the updated lightwalletd container image from electriccoinco/lightwalletd:latest.

## Solution

Update the regex pattern

References: https://hub.docker.com/layers/electriccoinco/lightwalletd/latest/images/sha256-7791924635f3d9ca59900278e4f167b511928decfd37dcb75778fc59bd52255e

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebrad/tests/acceptance.rs
+++ b/zebrad/tests/acceptance.rs
@@ -2001,12 +2001,12 @@ fn lightwalletd_integration_test(test_type: TestType) -> Result<()> {
         // getblockchaininfo
         if test_type.needs_zebra_cached_state() {
             lightwalletd.expect_stdout_line_matches(
-                "Got sapling height 419200 block height [0-9]{7} chain main branchID [0-9a-f]{8}",
+                r#""msg":"Got sapling height 419200 block height [0-9]{7} chain main branchID [0-9a-f]{8}""#,
             )?;
         } else {
             // Timeout the test if we're somehow accidentally using a cached state in our temp dir
             lightwalletd.expect_stdout_line_matches(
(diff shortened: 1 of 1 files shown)


upbqdn (approved):

conradoplg (changes requested):
I'm not sure this will make any difference.

Regexes [are matched with `is_match()`](https://github.com/ZcashFoundation/zebra/blob/45dbbb2b7ac45b308813fd947d483e0762b25b41/zebra-test/src/command.rs#L1084) which [does not need to match the full line](https://docs.rs/regex/latest/regex/struct.RegexSet.html#method.is_match).

You can also see the output didn't actually change, [before](https://github.com/ZcashFoundation/zebra/actions/runs/15285803371/job/42996035665#step:4:789) / [after](https://github.com/ZcashFoundation/zebra/actions/runs/15413800804/job/43372649123#step:4:784)

So the old regex should work, meaning the failure has another reason. My guess is some ordering issue - maybe the line is now being logged before we start checking for it

gustavovalverde:
Closing in favor of https://github.com/ZcashFoundation/zebra/pull/9575

PR #9585, opened 2025-06-04
Suggestions for "rpc: API cleanup #9568"

### LINKED ISSUE #9549: feature: make `zebra-rpc` support creating clients

arya2:
This is an incomplete suggestion PR for #9568 to rename types and export them directly from lib.

~It should not be merged into `rpc-cleanup` in its current form.~

Update: This can be merged into `rpc-cleanup` now.

@conradoplg can review and decide whether some or all of the changes/ideas make sense and it's worthwhile to continue with this approach.

If we think the breaking changes are worthwhile, we may want to also ask Zingo/Zallet developers if the breaking changes are worthwhile from their perspective as well.


zebra-rpc/src/lib.rs
```diff
@@ -15,3 +15,82 @@ pub mod indexer;
...
+// which is split mostly to keep file sizes small. Additionally, we re-export
+// types from other crates which are exposed in the API.
+//
+// TODO:
+// - Consider exporting functions here as well
+// - Expand documentation?
+// - Move `hex_data` and `trees` modules to/under the `types` module and qualify type refs in `methods`?
+pub mod client {
+    /// Types needed by clients of Zebra's RPC server
+    pub mod types {
```
conradoplg: If we do this I think we can also flatten `types` and `constants` and put everything into `client`

zebra-rpc/src/lib.rs
```diff
@@ -15,3 +15,82 @@ pub mod indexer;
...
+            GetBlockResponse, GetBlockTransaction, GetBlockTrees, GetInfoResponse,
+            GetRawTransactionResponse, Hash, SendRawTransactionResponse, Utxo,
+        };
+
+        // TODO: Move other deprecated type names here.
+        #[deprecated(note = "Use `GetSubtreesByIndexResponse` instead.")]
+        pub use GetSubtreesByIndexResponse as GetSubtrees;
+
+        #[deprecated(note = "Use `GetTreestateResponse` instead.")]
+        pub use GetTreestateResponse as GetTreeState;
```
conradoplg: To be valuable the deprecated types should keep their original import paths, but that should be doable, but if we were to do this then it would make sense to create deprecated types for _all_ types since they are all moving paths... and that is probably too much work, so it seems easier to just break everything and not create these deprecated types names.

zebra-rpc/src/methods/tests/snapshot.rs
```diff
@@ -810,7 +808,7 @@ fn snapshot_rpc_getblockhash_invalid(
 /// Snapshot `getblocktemplate` response, using `cargo insta` and JSON serialization.
 fn snapshot_rpc_getblocktemplate(
     variant: &'static str,
-    block_template: get_block_template::Response,
+    block_template: GetBlockTemplateResponse,
```
conradoplg, line 810: I think it doesn't matter much what import path we use internally, so these changes could be omitted, but happy to merge them

conradoplg (commented):
I like this suggestion and added some comments, I think in the end it boils down to "do we want to break everything and make the API as nice as possible, or do we want a middle-term solution where we break a handful of things?". I'm not really sure what is the answer.

mpguerra:
> I like this suggestion and added some comments, I think in the end it boils down to "do we want to break everything and make the API as nice as possible, or do we want a middle-term solution where we break a handful of things?". I'm not really sure what is the answer.

I don't have a strong opinion either way, possibly leaning slightly towards breaking things once instead of multiple times. I think best effort for now is good enough also

zebra-rpc/src/lib.rs
```diff
@@ -15,3 +15,57 @@ pub mod indexer;
...
+        },
+        BlockHeaderObject, BlockObject, GetAddressBalanceRequest, GetAddressBalanceResponse,
+        GetAddressTxIdsRequest, GetAddressUtxosResponse, GetBlockHashResponse,
+        GetBlockHeaderResponse, GetBlockHeightAndHashResponse, GetBlockResponse,
+        GetBlockTransaction, GetBlockTrees, GetBlockchainInfoResponse, GetInfoResponse,
+        GetRawTransactionResponse, Hash, SendRawTransactionResponse, Utxo,
+    };
+
+    /// Constants needed by clients of Zebra's RPC server
+    // TODO: Export all other constants?
```
arya2, line 69: ~TODO: double-check that anything that was publicly exposed via `zebra_rpc::methods::{trees, types, hex_data}` is still exposed.~

Update: Keep these public, but under a new `server` feature so it'll be a last resort for clients.

zebra-rpc/src/lib.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9585/commits/087a59cd1e659a9f6f5dc909473bab505c0f5f8f.

zebra-rpc/src/lib.rs
arya2: Removed in https://github.com/ZcashFoundation/zebra/pull/9585/commits/087a59cd1e659a9f6f5dc909473bab505c0f5f8f.

zebra-rpc/src/methods/tests/snapshot.rs
arya2: I agree, I didn't change the internal import path in a bunch of places, I only wanted to get rid of the `as` so each type goes by a single name.

arya2:
@conradoplg Let me know if you want any changes (cleanup, reverting an update to an unrelated TODO, updating `transparent::Outpoint` to use `OutputIndex`, etc), but this should be good to merge into the base branch now (though there are a few outstanding doc refs that are broken, at least for the updated `GetBlockchainInfoBalance` path).

zebra-rpc/src/lib.rs
```diff
@@ -15,3 +15,57 @@ pub mod indexer;
...
+
+/// Types, constants, and functions needed by clients of Zebra's RPC server
+//
+// Re-exports types to build the API. We want to flatten the module substructure
+// which is split mostly to keep file sizes small. Additionally, we re-export
+// types from other crates which are exposed in the API.
+//
+// TODO: Move `hex_data` and `trees` modules to/under the `types` module?
+pub mod client {
+    pub use crate::methods::{
```
arya2, line 35: Re-export zebra-chain from here.

conradoplg (approved):

PR #9594, opened 2025-06-04
fix(consensus): reuse sighasher when validating txs

### LINKED ISSUE #9532: feature: actually cache sighash context
### Motivation

In https://github.com/ZcashFoundation/zebra/pull/8566 I updated `zebra-script` for the new `zcash-script` callback API. I changed the  `SigHasher` to support precomputation which we had in the old API... but it ended up not being actually used 😬  (that PR was super tricky and I guess after I got it to work I completely forgot about getting the precomputation part to work).

We should changed it so that the `SigHasher` is actually stored inside `CachedFfiTransaction` or something equivalent. We should also find a way to share the `SigHasher` with the shielded transaction validation.

I'm not sure how big will be the performance gain but it should be an easy quick win.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

We were creating a new sighasher context for each input being verified, which is incredibly wasteful - not only because the sighasher context can be reused, but also because building the sighasher requires converting from a Zebra tx to a librustzcash tx via serializing/deserializing, and librustzcash deserializing also suffers from #7939 

Closes #9532 

## Solution

Create a sighasher in `CachedFfiTransaction::new()` and use it for all validations. Also reuse it when verifying shielded transactions.

This required changing the wrap the `all_previous_outputs` `Vec<Output>` in an `Arc`. It was possible to just move it all the way down to the `PrecomputedData` and add getters for it all the way back to `CachedFfiTransaction`, but using `Arc` seemed much simpler and mimics how `Transaction` is handled (also with an `Arc`)

I also moved `legacy_sigop_count()` out of `CachedFfiTransaction` because otherwise using it would require creating a `CachedFfiTransaction` with a dummy network upgrade; and I think it didn't actually belong there because it does not use anything in `CachedFfiTransaction` other then the transaction reference itself.

### Tests

Existing tests should ensure it will work.

We don't have benchmarks for this but the performance gains should be immediate. I'll also run a full sync to see what happens.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/error.rs
+++ b/zebra-chain/src/error.rs
@@ -1,6 +1,6 @@
 //! Errors that can occur inside any `zebra-chain` submodule.
 
-use std::io;
+use std::{io, sync::Arc};
 use thiserror::Error;
 
 // TODO: Move all these enums into a common enum at the bottom.
@@ -57,17 +57,47 @@ pub enum AddressError {
 }
 
 /// `zebra-chain`'s errors
-#[derive(Error, Debug)]
+#[derive(Clone, Error, Debug)]
 pub enum Error {
     /// Invalid consensus branch ID.
     #[error("invalid consensus branch id")]
     InvalidConsensusBranchId,
 
     /// Zebra's type could not be converted to its librustzcash equivalent.
-    #[error("Zebra's type could not be converted to its librustzcash equivalent: ")]
-    Conversion(#[from] io::Error),
+    #[error("Zebra's type could not be converted to its librustzcash equivalent: {0}")]
+    Conversion(#[from] Arc<io::Error>),
 
     /// The transaction is missing a network upgrade.
     #[error("the transaction is missing a network upgrade")]
     MissingNetworkUpgrade,
 }
+
+/// Allow converting `io::Error` to `Error`; we need this since we
+/// use `Arc<io::Error>` in `Error::Conversion`.
+impl From<io::Error> for Error {
+    fn from(value: io::Error) -> Self {
+        Arc::new(value).into()
+    }
+}
+
+// We need to implement this manually because io::Error does not implement
+// PartialEq.
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        match self {
+            Error::InvalidConsensusBranchId => matches!(other, Error::InvalidConsensusBranchId),
+            Error::Conversion(e) => {
+                if let Error::Conversion(o) = other {
+                    // Not perfect, but good enough for testing, which
+                    // is the main purpose for our usage of PartialEq for errors
+                    e.to_string() == o.to_string()
+                } else {
+                    false
+                }
+            }
+            Error::MissingNetworkUpgrade => matches!(other, Error::MissingNetworkUpgrade),
+        }
+    }
+}
+
+impl Eq for Error {}
--- a/zebra-chain/src/transaction/sighash.rs
+++ b/zebra-chain/src/transaction/sighash.rs
@@ -1,5 +1,7 @@
 //! Signature hashes for Zcash transactions
 
+use std::sync::Arc;
+
 use zcash_transparent::sighash::SighashType;
 
 use super::Transaction;
@@ -67,24 +69,24 @@ impl AsRef<[u8]> for SigHash {
 /// A SigHasher context which stores precomputed data that is reused
 /// between sighash computations for the same transaction.
 #[derive(Debug)]
-pub struct SigHasher<'a> {
-    precomputed_tx_data: PrecomputedTxData<'a>,
+pub struct SigHasher {
+    precomputed_tx_data: PrecomputedTxData,
 }
 
-impl<'a> SigHasher<'a> {
+impl SigHasher {
     /// Create a new SigHasher for the given transaction.
     ///
     /// # Panics
     ///
     /// - If `trans` can't be converted to its `librustzcash` equivalent. This could happen, for
     ///   example, if `trans` contains the `nConsensusBranchId` field, and `nu` doesn't match it.
-    ///   More details in [`PrecomputedTxData::new`].
-    /// - If `nu` doesn't contain a consensus branch id convertible to its `librustzcash`
+    ///   More details [`PrecomputedTxData::new`].
+    /// - If `nu` doesn't contain a consensuranch id convertible to its `librustzcash`
     ///   equivalent.
     pub fn new(
-        trans: &'a Transaction,
+        trans: &Transaction,
         nu: NetworkUpgrade,
-        all_previous_outputs: &'a [transparent::Output],
+        all_previous_outputs: Arc<Vec<transparent::Output>>,
     ) -> Self {
         SigHasher {
             precomputed_tx_data: PrecomputedTxData::new(trans, nu, all_previous_outputs),
--- a/zebra-chain/src/transaction/sighash.rs
+++ b/zebra-chain/src/transaction/sighash.rs
@@ -7,7 +7,7 @@ use zcash_transparent::sighash::SighashType;
 use super::Transaction;
 
 use crate::parameters::NetworkUpgrade;
-use crate::transparent;
+use crate::{transparent, Error};
 
 use crate::primitives::zcash_primitives::{sighash, PrecomputedTxData};
 
@@ -87,10 +87,10 @@ impl SigHasher {
         trans: &Transaction,
         nu: NetworkUpgrade,
         all_previous_outputs: Arc<Vec<transparent::Output>>,
-    ) -> Self {
-        SigHasher {
-            precomputed_tx_data: PrecomputedTxData::new(trans, nu, all_previous_outputs),
-        }
+    ) -> Result<Self, Error> {
+        Ok(SigHasher {
+            precomputed_tx_data: PrecomputedTxData::new(trans, nu, all_previous_outputs)?,
+        })
     }
 
     /// Calculate the sighash for the current transaction.
--- a/zebra-chain/src/transaction/sighash.rs
+++ b/zebra-chain/src/transaction/sighash.rs
@@ -80,8 +80,8 @@ impl SigHasher {
     ///
     /// - If `trans` can't be converted to its `librustzcash` equivalent. This could happen, for
     ///   example, if `trans` contains the `nConsensusBranchId` field, and `nu` doesn't match it.
-    ///   More details [`PrecomputedTxData::new`].
-    /// - If `nu` doesn't contain a consensuranch id convertible to its `librustzcash`
+    ///   More details in [`PrecomputedTxData::new`].
+    /// - If `nu` doesn't contain a consensus branch id convertible to its `librustzcash`
     ///   equivalent.
     pub fn new(
         trans: &Transaction,
--- a/zebra-chain/src/transaction/tests/vectors.rs
+++ b/zebra-chain/src/transaction/tests/vectors.rs
@@ -874,20 +874,16 @@ fn consensus_branch_id() {
             tx.to_librustzcash(any_other_nu)
                 .expect_err("tx is not convertible under nu other than the tx one");
 
-            std::panic::catch_unwind(|| {
-                PrecomputedTxData::new(&tx, any_other_nu, Arc::new(Vec::new()))
-            })
-            .expect_err("precomputing tx sighash data panics under nu other than the tx one");
+            let err = PrecomputedTxData::new(&tx, any_other_nu, Arc::new(Vec::new())).unwrap_err();
+            assert!(matches!(err, crate::Error::InvalidConsensusBranchId));
 
-            std::panic::catch_unwind(|| {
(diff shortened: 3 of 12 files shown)


zebra-script/src/lib.rs
```diff
@@ -42,6 +42,10 @@ pub enum Error {
     TxCoinbase,
     /// unknown error from zcash_script: {0}
     Unknown(zcash_script_error_t),
+    /// transaction is invalid according to zebra_chain (not a zcash_script error)
+    //TODO: ideally this should be `#[from] zebra_chain::Error` but that does
+    //not implement Clone
```
upbqdn: Can we not derive it?

zebra-chain/src/transaction/sighash.rs
```diff
@@ -67,28 +69,28 @@ impl AsRef<[u8]> for SigHash {
...
     /// Create a new SigHasher for the given transaction.
     ///
     /// # Panics
     ///
     /// - If `trans` can't be converted to its `librustzcash` equivalent. This could happen, for
     ///   example, if `trans` contains the `nConsensusBranchId` field, and `nu` doesn't match it.
-    ///   More details in [`PrecomputedTxData::new`].
-    /// - If `nu` doesn't contain a consensus branch id convertible to its `librustzcash`
+    ///   More details [`PrecomputedTxData::new`].
+    /// - If `nu` doesn't contain a consensuranch id convertible to its `librustzcash`
```
upbqdn: Why this change?

```suggestion
    /// - If `nu` doesn't contain a consensus branch id convertible to its `librustzcash`
```

zebra-chain/src/transaction/sighash.rs
```diff
@@ -67,28 +69,28 @@ impl AsRef<[u8]> for SigHash {
...
+impl SigHasher {
     /// Create a new SigHasher for the given transaction.
     ///
     /// # Panics
     ///
     /// - If `trans` can't be converted to its `librustzcash` equivalent. This could happen, for
     ///   example, if `trans` contains the `nConsensusBranchId` field, and `nu` doesn't match it.
-    ///   More details in [`PrecomputedTxData::new`].
-    /// - If `nu` doesn't contain a consensus branch id convertible to its `librustzcash`
+    ///   More details [`PrecomputedTxData::new`].
```
upbqdn:
```suggestion
    ///   More details in [`PrecomputedTxData::new`].
```

zebra-script/src/lib.rs
conradoplg: We can't, that contains an `io::Error` that annoyingly does not implement `Clone`

zebra-chain/src/transaction/sighash.rs
conradoplg: Weird, some search&replace mistake, thanks!

upbqdn (dismissed):

zebra-script/src/lib.rs
```diff
@@ -42,6 +42,10 @@ pub enum Error {
...
+    TxInvalid,
 }
```
arya2: This should work for keeping the `Clone` impl, but not the `Copy` impl.

```suggestion
    TxInvalid(#[from] std::sync::Arc<zebra_chain::Error>),
}

impl From<zebra_chain::Error> for Error {
    fn from(value: zebra_chain::Error) -> Self {
        std::sync::Arc::new(value).into()
    }
}
```

zebra-script/src/lib.rs
```diff
@@ -55,6 +59,7 @@ impl fmt::Display for Error {
                 "tx is a coinbase transaction and should not be verified".to_owned()
             }
             Error::Unknown(e) => format!("unknown error from zcash_script: {e}"),
+            Error::TxInvalid => "tx is invalid".to_owned(),
```
```suggestion
            Error::TxInvalid(e) => format!("tx is invalid: {e}"),
```

zebra-chain/src/transaction/tests/vectors.rs
```diff
@@ -549,7 +555,8 @@ fn test_vec243_1() -> Result<()> {
 
     let transaction = ZIP243_1.zcash_deserialize_into::<Transaction>()?;
 
-    let hasher = SigHasher::new(&transaction, NetworkUpgrade::Sapling, &[]);
+    let hasher = SigHasher::new(&transaction, NetworkUpgrade::Sapling, Arc::new(Vec::new()))
```
arya2, line 558: Optional/Nitpick: This could be simplified if we accept `impl Into<Vec<Output>>` for the `all_previous_outputs` param in `SigHasher::new()`.

```suggestion
    let hasher = SigHasher::new(&transaction, NetworkUpgrade::Sapling, [])
```

```rust

    pub fn new(
        trans: &Transaction,
        nu: NetworkUpgrade,
        all_previous_outputs: impl Into<Vec<transparent::Output>>,
    ) -> Result<Self, Error> {
        Ok(SigHasher {
            precomputed_tx_data: PrecomputedTxData::new(
                trans,
                nu,
                Arc::new(all_previous_outputs.into()),
            )?,
        })
    }

```

zebra-script/src/lib.rs
```diff
@@ -217,46 +235,46 @@ impl CachedFfiTransaction {
...
-                    unsafe {
-                        zcash_script::zcash_script_legacy_sigop_count_script(
-                            script.as_ptr(),
-                            script.len() as u32,
-                        )
-                    }
+/// Returns the number of transparent signature operations in the
+/// transparent inputs and outputs of the given transaction.
+#[allow(clippy::unwrap_in_result)]
+pub fn legacy_sigop_count(transaction: &Transaction) -> Result<u64, Error> {
```
arya2, line 239: Optional/Nitpick: Would it be nice to make this a trait method and implement it for both `Transaction` and `CachedFfiTransaction`?

arya2 (dismissed):
Last commit looks good, the production changes look good and should fix the issue. Nice catch!

I left a few suggestions for cleanup, but no blockers.

zebra-chain/src/transaction/tests/vectors.rs
```diff
@@ -874,20 +874,16 @@ fn consensus_branch_id() {
...
+            let err = PrecomputedTxData::new(&tx, any_other_nu, Arc::new(Vec::new())).unwrap_err();
+            assert!(matches!(err, crate::Error::InvalidConsensusBranchId));
+
+            let err = sighash::SigHasher::new(&tx, any_other_nu, Arc::new(Vec::new())).unwrap_err();
+            assert!(matches!(err, crate::Error::InvalidConsensusBranchId));
+
+            let err = tx
+                .sighash(any_other_nu, HashType::ALL, Arc::new(Vec::new()), None)
+                .unwrap_err();
+            assert!(matches!(err, crate::Error::InvalidConsensusBranchId));
```
arya2: Optional/Nitpick: It may be nice to have the old panic messages in `expect_err()` for these `assert!`s too.

zebra-script/src/lib.rs
conradoplg: I love this. I considered using `Arc` but didn't considering manually adding a `From` to a avoid all the breakage.

In fact I think I'll try to do this where `io::Error` is used, if it works I'll create a separate PR just for that.

Thanks!

zebra-chain/src/transaction/tests/vectors.rs
conradoplg: Oof that would have saved me some work :rofl: I'll do it anyway, it looks much better, thanks!

zebra-script/src/lib.rs
upbqdn: I can try that in a subsequent PR.

zebra-chain/src/transaction/tests/vectors.rs
conradoplg: Done in 9db4f0a1b

zebra-script/src/lib.rs
conradoplg: I think it was small enough so I sneaked this in b69d1a5. Let me know if you'd prefer a separate PR

zebra-script/src/lib.rs
conradoplg: Done in b69d1a5

zebra-chain/src/transaction/tests/vectors.rs
conradoplg: Now that I thought about it I'm not sure, the whole point of using `Arc` was to avoid cloning the entire vector. However if we do this, `CachedFfiTransaction` will have to clone it when calling `Sighasher::new()`.

I imagine there might be some trait magic to make this (make a function receive either T or `Arc<T>` and automatically wrap in a `Arc` if it is `T`) work but I couldn't find it.

In hindsight I think just moving the vector all the way down to `PrecomputedTx` might have been the best approach, but I guess it doesn't really matter that much

zebra-script/src/lib.rs
conradoplg: I'll leave this as is then

upbqdn: I'll need sigops for lrz txs.

zebra-chain/src/transaction/tests/vectors.rs
arya2: > Now that I thought about it I'm not sure, the whole point of using Arc was to avoid cloning the entire vector.

It'll still do that, `Sighasher::new()` is _now_ only called once in the transaction verifier via `CachedFfiTransaction::new()`. 

Update: I see, it has a copy in `all_previous_outputs` and another copy on the `sighasher` field. Slight waste of memory when blocks are being synced out of order (if it were to be updated to accept `Into<Vec<_>>` instead of `Arc<Vec<_>>`). We could remove the `all_previous_outputs` field altogether and make it a function that gets it from the `Sighasher` to avoid that:

```rust

impl PrecomputedTxData{
    pub fn all_prev_outputs(&self) -> Arc<Vec<transparent::Output>> {
        self.all_previous_outputs.clone()
    }
}

impl SigHasher {
    pub fn all_prev_outputs(&self) -> Arc<Vec<transparent::Output>> {
        self.precomputed_tx_data.all_prev_outputs()
    }
}

impl CachedFfiTransaction {
    pub fn all_previous_outputs(&self) -> Arc<Vec<transparent::Output>> {
        self.sighasher.all_prev_outputs()
    }

    pub fn is_valid(&self, input_index: usize) -> Result<(), Error> {
        let all_previous_outputs = self.all_previous_outputs();
        let previous_output = all_previous_outputs
        ..
    }
}

```


Previously it would've cloned it, which wasn't terrible since the items are `Copy` if not for re-constructing the entire `Sighasher` and `zp_tx::TransactionData`.

Not very important here either way, I mostly wanted to advertise the pattern.

zebra-script/src/lib.rs
arya2: I'm happy for it to go in this PR. 

Could we implement `PartialEq` for the `zebra_script` Error type instead?

zebra-chain/src/error.rs
```diff
@@ -57,17 +57,47 @@ pub enum AddressError {
...
+                    // is the main purpose for our usage of PartialEq for errors
+                    e.to_string() == o.to_string()
+                } else {
+                    false
+                }
+            }
+            Error::MissingNetworkUpgrade => matches!(other, Error::MissingNetworkUpgrade),
+        }
+    }
+}
```
arya2, line 101: We should implement this for the error type in `zebra-script` instead and say it's partially equal as long as the enum variant matches. I think we should avoid implementing `Eq` for that too.

zebra-chain/src/transaction/tests/vectors.rs
arya2: > Not very important here either way, I mostly wanted to advertise the pattern.

It would be nice to get rid of the duplicate field, but accepting the generic param still seems unimportant.

zebra-chain/src/error.rs
conradoplg: Why do you think that? I don't think two `zebra_script::Error::TxInvalid(e)` should be considered equal if `e` is different between them.

zebra-chain/src/transaction/tests/vectors.rs
conradoplg: > We could remove the `all_previous_outputs` field altogether and make it a function that gets it from the `Sighasher` to avoid that:

That would work for `CachedFfiTransaction`, but `Sighasher` would need to do the same since `PrecomputedTxData` (a `Sighasher` field) also needs `all_previous_output`... and there is now a bit trickier, because it's put into a `MapTransparent` struct which is passed to `.map_authorization()` but I'm not sure if you can ever fetch it back after that. It just seemed simpler to use `Arc`

zebra-chain/src/error.rs
arya2: > Why do you think that?

I think it's okay because that was the behaviour before, but, it's not ideal.

arya2 (approved):

PR #9592, opened 2025-06-04
Lockbox disbursement mechanism

arya2:
Note: This PR will be re-created, and is not yet ready for more review.

PR #9603, opened 2025-06-07
add(consensus): Implement one-time lockbox disbursement mechanism for NU6.1

### LINKED ISSUE #9596: Implement ZIP 271: Deferred Dev Fund Lockbox Disbursement
[draft-ecc-lockbox-disbursement: Deferred Dev Fund Lockbox Disbursement](https://zips.z.cash/draft-ecc-lockbox-disbursement)

arya2:
## Motivation

We want to implement a mechanism for upcoming the one-time lockbox disbursement(s) in NU6.1

Supersedes #9592.
Part of #9596.

## Solution

- Adds new constants for defining the lockbox disbursements to be expected in the NU6.1 activation block's coinbase tx,
- Renames `deferred_balance` fields on `SemanticallyVerifiedBlock` and `FinalizedBlock` to `deferred_pool_balance_change`
- Updates block and checkpoint verifiers to subtract the lockbox disbursement amount from the deferred pool balance change at the NU6.1 activation height
  - Updates db format upgrade to do the same
- Updates `standard_coinbase_outputs()` fn used by the `getblocktemplate` RPC to add lockbox disbursement outputs to NU6.1 activation block templates.

### Tests

- Adds a (commented-out) test to check that the expected total of lockbox disbursement values matches the sum of expected lockbox disbursement values 

### Follow-up Work

Update constants once the NU6.1 ZIP has been finalized.

### PR Checklist

- [ ] The solution is tested.


zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -289,6 +289,26 @@ const POST_NU6_FUNDING_STREAM_START_HEIGHT_MAINNET: u32 = 2_726_400;
...
+
+/// The one-time lockbox disbursement output addresses and amounts expected in the NU6.1 activation block's
+/// coinbase transaction on Testnet.
+// TODO: Update this with hard-coded values once the ZIP is finalized and uncomment
+//       `sum_of_one_time_lockbox_disbursements_is_correct`.
+pub const NU6_1_LOCKBOX_DISBURSEMENT_TESTNET: [(&str, i64); 0] = [];
+
+/// The expected total amount of the one-time lockbox disbursement on Mainnet.
+// TODO: Add a reference to the ZIP and update this value if needed.
+pub const EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_MAINNET: i64 = 577_500;
```
arya2: These totals may be wrong.

zebra-rpc/src/methods/types/get_block_template.rs
```diff
@@ -918,7 +918,15 @@ pub fn standard_coinbase_outputs(
     let miner_reward =
         miner_reward.expect("reward calculations are valid for reasonable chain heights");
 
-    combine_coinbase_outputs(funding_streams, miner_address, miner_reward, like_zcashd)
+    let one_time_lockbox_disbursements = network.lockbox_disbursement(height);
+
+    combine_coinbase_outputs(
+        funding_streams,
+        one_time_lockbox_disbursements,
```
arya2: Double-check that the expected size of the coinbase transaction is still correct (when calculating how many other transactions should fit into the block).

Update: `fake_coinbase_transaction()` goes through the same functions as expected, this should be okay.

zebra-chain/src/parameters/network.rs
```diff
@@ -274,6 +276,54 @@ impl Network {
...
+        if Some(height) != NetworkUpgrade::Nu6_1.activation_height(self) {
+            return Amount::zero();
+        };
+
+        match self {
+            Self::Mainnet => subsidy::EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_MAINNET,
+            Self::Testnet(params) if params.is_default_testnet() => {
+                subsidy::EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_TESTNET
+            }
+            Self::Testnet(_params) => 0,
```
nuttycom: This should be configurable so that this can be tested in regtest/testnet.

zebra-chain/src/parameters/network.rs
```diff
@@ -274,6 +276,54 @@ impl Network {
...
+    ) -> Vec<(transparent::Address, Amount<NonNegative>)> {
+        if Some(height) != NetworkUpgrade::Nu6_1.activation_height(self) {
+            return Vec::new();
+        };
+
+        let expected_lockbox_disbursements = match self {
+            Self::Mainnet => subsidy::NU6_1_LOCKBOX_DISBURSEMENT_MAINNET,
+            Self::Testnet(params) if params.is_default_testnet() => {
+                subsidy::NU6_1_LOCKBOX_DISBURSEMENT_TESTNET
+            }
```
nuttycom: naming: these constants' names should probably be pluralized

zebra-chain/src/parameters/network.rs
```diff
@@ -274,6 +276,54 @@ impl Network {
...
+            Self::Testnet(_params) => return Vec::new(),
```
nuttycom: These should also come from config for regtest

zebra-chain/src/parameters/network.rs
```diff
@@ -274,6 +276,54 @@ impl Network {
...
+    pub fn lockbox_disbursement_total_amount(&self, height: Height) -> Amount<NonNegative> {
+        if Some(height) != NetworkUpgrade::Nu6_1.activation_height(self) {
+            return Amount::zero();
+        };
+
+        match self {
+            Self::Mainnet => subsidy::EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_MAINNET,
+            Self::Testnet(params) if params.is_default_testnet() => {
+                subsidy::EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_TESTNET
+            }
```
nuttycom: The relationship between `EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_MAINNET` and the sum of `NU6_1_LOCKBOX_DISBURSEMENT_MAINNET` should be enforced by a test at very least; however, it might be better to just compute the total value so that there's not a possibility of error.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -289,6 +289,26 @@ const POST_NU6_FUNDING_STREAM_START_HEIGHT_MAINNET: u32 = 2_726_400;
...
+
+/// The one-time lockbox disbursement output addresses and amounts expected in the NU6.1 activation block's
+/// coinbase transaction on Testnet.
+// TODO: Update this with hard-coded values once the ZIP is finalized and uncomment
+//       `sum_of_one_time_lockbox_disbursements_is_correct`.
+pub const NU6_1_LOCKBOX_DISBURSEMENT_TESTNET: [(&str, i64); 0] = [];
+
+/// The expected total amount of the one-time lockbox disbursement on Mainnet.
+// TODO: Add a reference to the ZIP and update this value if needed.
+pub const EXPECTED_NU6_1_LOCKBOX_DISBURSEMENT_TOTAL_MAINNET: i64 = 69_300;
```
nuttycom: Are these denominated in zatoshis or ZEC? Also this PR should not be merged until these values are consistent with `NU6_1_LOCKBOX_DISBURSEMENT_MAINNET`

Finally, why is this a signed integer? Ideally, add an "unsafe" const constructor for `Amount` that panics if the amount is out of range, e.g. https://github.com/zcash/librustzcash/blob/a1c9aa12892a7c7e86365aa89990a584a634f53c/components/zcash_protocol/src/value.rs#L48-L51

nuttycom: Suggestion from @str4d: Add a unit test that recomputes the lockbox value using the consensus machinery, and verify that it equals this constant.

zebra-consensus/src/block/check.rs
```diff
@@ -144,22 +144,23 @@ pub fn equihash_solution_is_valid(header: &Header) -> Result<(), equihash::Error
...
+/// Returns `Ok()` with the deferred pool balance change of the coinbase transaction if
+/// the block subsidy in `block` is valid for `network`
 ///
 /// [3.9]: https://zips.z.cash/protocol/protocol.pdf#subsidyconcepts
 pub fn subsidy_is_valid(
     block: &Block,
     network: &Network,
     expected_block_subsidy: Amount<NonNegative>,
-) -> Result<(), BlockError> {
+) -> Result<Amount, BlockError> {
```
nuttycom: nit: Since this is a value with specific protocol semantics, I might prefer a newtype to a bare `Amount` to make the returned value harder to misuse.

zebra-consensus/src/block/check.rs
```diff
@@ -184,14 +185,42 @@ pub fn subsidy_is_valid(
...
+        // See [ZIP-1015](https://zips.z.cash/zip-1015) for more details.
+        let mut deferred_pool_balance_change = funding_streams
+            .remove(&FundingStreamReceiver::Deferred)
+            .unwrap_or_default()
+            .constrain::<NegativeAllowed>()
+            .expect("should be valid Amount");
+
+        // TODO: Add references to the one-time lockbox disbursement & community coinholder funding model ZIPs
+        //       (https://zips.z.cash/draft-ecc-lockbox-disbursement, https://zips.z.cash/draft-ecc-community-and-coinholder)
+        let expected_one_time_lockbox_disbursements = network.lockbox_disbursement(height);
```
nuttycom:
```suggestion
        let expected_one_time_lockbox_disbursements = network.lockbox_disbursements(height);
```

zebra-consensus/src/block/check.rs
```diff
@@ -184,14 +185,42 @@ pub fn subsidy_is_valid(
...
+        let mut funding_streams = zebra_chain::parameters::subsidy::funding_stream_values(
             height,
             network,
             expected_block_subsidy,
         )
         // we always expect a funding stream hashmap response even if empty
         .map_err(|err| BlockError::Other(err.to_string()))?;
 
+        let has_expected_output = |address, expected_amount| {
+            subsidy::funding_streams::filter_outputs_by_address(coinbase, address)
```
nuttycom: Potentially add a helper method to the transaction type that finds the outputs for a given address, or move this into the subsidy code.

zebra-consensus/src/block/check.rs
```diff
@@ -184,14 +185,42 @@ pub fn subsidy_is_valid(
...
+            .constrain::<NegativeAllowed>()
+            .expect("should be valid Amount");
+
+        // TODO: Add references to the one-time lockbox disbursement & community coinholder funding model ZIPs
+        //       (https://zips.z.cash/draft-ecc-lockbox-disbursement, https://zips.z.cash/draft-ecc-community-and-coinholder)
+        let expected_one_time_lockbox_disbursements = network.lockbox_disbursement(height);
+        for (address, expected_amount) in &expected_one_time_lockbox_disbursements {
+            if !has_expected_output(address, *expected_amount) {
+                Err(SubsidyError::OneTimeLockboxDisbursementNotFound)?;
+            }
```
nuttycom, line 223: This isn't sufficient; we need an exact match between disbursements and outputs, not simply set inclusion of the expected (address, output amount) pairs.

nuttycom (dismissed):
Flushing comments.

Action item for zcashd: Ensure that there is no problem with matching the set of disbursements against the actual coinbase outputs.

zebra-consensus/src/block/check.rs
nuttycom: @arya2 says: Consider whether a multiset might be useful here.

zebra-chain/src/block.rs
```diff
@@ -224,20 +224,15 @@ impl Block {
...
             .flat_map(|t| t.value_balance(utxos))
             .sum::<Result<ValueBalance<NegativeAllowed>, _>>()?
             .neg()
-            .set_deferred_amount(
-                deferred_balance
-                    .unwrap_or(Amount::zero())
-                    .constrain::<NegativeAllowed>()
-                    .map_err(ValueBalanceError::Deferred)?,
-            ))
+            .set_deferred_amount(deferred_pool_balance_change.unwrap_or_default()))
```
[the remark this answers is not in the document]
arya2: A match statement seems very unnecessary here, I'll undo this change if there are no strong objections.

zebra-chain/src/parameters/network.rs
arya2: Addressed in https://github.com/ZcashFoundation/zebra/pull/9603/commits/e4e7555b886f9b7bb0129cab0d1da2b2a409110c.

zebra-chain/src/parameters/network/subsidy.rs
arya2: Addressed the first comment in https://github.com/ZcashFoundation/zebra/pull/9603/commits/e4e7555b886f9b7bb0129cab0d1da2b2a409110c and the second comment in https://github.com/ZcashFoundation/zebra/pull/9603/commits/271a20de6046196235ea81d124dac12948d66a68.

zebra-chain/src/parameters/network.rs
arya2: Addressed in https://github.com/ZcashFoundation/zebra/pull/9603/commits/271a20de6046196235ea81d124dac12948d66a68 (with a configured testnet for now, but the test will check Mainnet/Testnet too once those define activation heights for NU6.1).

zebra-chain/src/parameters/network.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9603/commits/271a20de6046196235ea81d124dac12948d66a68.

zebra-chain/src/parameters/network.rs
arya2: Done in https://github.com/ZcashFoundation/zebra/pull/9603/commits/271a20de6046196235ea81d124dac12948d66a68.

zebra-consensus/src/block/check.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9603/commits/e4e7555b886f9b7bb0129cab0d1da2b2a409110c.

zebra-consensus/src/block/check.rs
arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/9603/commits/12cc25d3eb516dd3a31e2fde9bd1a8982bc514d1.

zebra-consensus/src/block/check.rs
arya2: Simplified the logic, removed `filter_outputs_by_address()`, and moved `new_coinbase_script()` up a level to the `subsidy` module in https://github.com/ZcashFoundation/zebra/pull/9603/commits/12cc25d3eb516dd3a31e2fde9bd1a8982bc514d1.

zebra-consensus/src/block/check.rs
arya2: Applied in https://github.com/ZcashFoundation/zebra/pull/9603/commits/aafe3e568d89f8940d42ed0ee294fcfcc45e4d2b.

oxarbitrage:
Hey @arya2 . Can you rebase to the last main ? Will make this diff shorter (https://github.com/ZcashFoundation/zebra/pull/9705)

arya2:
> Hey @arya2 . Can you rebase to the last main ? Will make this diff shorter (#9705)

I merged in main (to avoid a conflict around the gbt coinbase generation), should work now if this branch is merged into #9705.

zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
oxarbitrage, line 1: please delete file.

arya2: Whoops. Thank you. Deleted in https://github.com/ZcashFoundation/zebra/pull/9603/commits/7df536b94a0337340a1919d5ff95994d911cb48f.

oxarbitrage:
It can be useful to bring https://github.com/ZcashFoundation/zebra/pull/9710 here after it gets merged.

zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -289,6 +289,26 @@ const POST_NU6_FUNDING_STREAM_START_HEIGHT_MAINNET: u32 = 2_726_400;
 /// The start height of post-NU6 funding streams on Testnet as described in [ZIP-1015](https://zips.z.cash/zip-1015).
 const POST_NU6_FUNDING_STREAM_START_HEIGHT_TESTNET: u32 = 2_976_000;
 
+/// The one-time lockbox disbursement output addresses and amounts expected in the NU6.1 activation block's
+/// coinbase transaction on Mainnet.
+// TODO: Update this with specified values once the ZIP is finalized.
```
oxarbitrage, line 294: Please add the ZIP number or some sort of reference to this TODO's so we can know where to expect this values to be added.

zebra-chain/src/parameters/network/tests/vectors.rs
```diff
@@ -448,3 +450,81 @@ fn check_configured_funding_stream_constraints() {
...
+        let last_nu6_height = nu6_1_activation_height.previous().unwrap();
+        let expected_total_lockbox_disbursement_value =
+            lockbox_input_value(&network, last_nu6_height);
+
+        assert_eq!(
+            expected_total_lockbox_disbursement_value,
+            network.lockbox_disbursement_total_amount(nu6_1_activation_height),
+            "sum of lockbox disbursement output values should match expected total"
+        );
+    }
```
oxarbitrage, line 500: After the disbursement is completed, there are new funding streams (including the lockbox) that begin with NU6.1 and continue up to the third halving:
https://github.com/zcash/zcash/blob/658fba5cda37bc8446a13e28cbeffe153a0b9ab4/src/consensus/params.cpp#L50-L61

I think we're not currently covering that. As things stand, the disbursements will happen, and then stop. 
We need to ensure that a new lockbox pool begins receiving funds (up to the third halving), along with the other funding streams.

We can extend the test at https://github.com/ZcashFoundation/zebra/blob/c0f32eee1899b728ca4f3b35e8edadb1103e5d2e/zebra-rpc/qa/rpc-tests/feature_nu6.py#L244 to prove this.

arya2: > there are new funding streams (including the lockbox) that begin with NU6.1 and continue up to the third halving

The intention was to update the end height of the post-NU6 funding streams to reflect ZIP 271, but immediately before releasing the versions of Zebra that deploy NU6.1 on Testnet/Mainnet.

> https://github.com/zcash/zcash/blob/658fba5cda37bc8446a13e28cbeffe153a0b9ab4/src/consensus/params.cpp#L50-L61

If I recall correctly, those are (and should be) unused in zcashd until the activation height of NU6.1, but NU6 would be the only network upgrade boundary that affects which funding streams are selected as funding streams currently work in Zebra. 

Unfortunately, the current funding stream has already ended on Testnet, so this won't work.

We need to either add another network upgrade boundary for the funding streams at NU6.1, or refactor how funding streams work in Zebra to consolidate them and allow for arbitrary height or network upgrade boundaries. 

I'm working on adding arbitrary height boundaries to close https://github.com/ZcashFoundation/zebra/issues/9736, but we still shouldn't add the lockbox disbursement outputs or the new funding streams until just before the releases that deploy them, only the mechanisms.

> As things stand, the disbursements will happen, and then stop.
We need to ensure that a new lockbox pool begins receiving funds (up to the third halving), along with the other funding streams.

Yep, this PR also doesn't add any actual lockbox disbursement outputs, it only adds the mechanism for them.

conradoplg:
You should be able to get this merged by resolving comments and approving it (the failing checks are non-required). If it doesn't work I'll admin-merge it

oxarbitrage (approved):

zebra-chain/src/parameters/network/tests/vectors.rs
dpoulami: @arya2, I am currently auditing the code from Least Authority. So IIUC, the continuation of new funding streams until third halving, which has been mentioned in ZIP 271 as the selected option is not being implemented as part of this set of PRs we are auditing?

arya2: @dpoulami the funding stream extension and one-time lockbox disbursements haven't been updated for Mainnet yet (because we haven't specified an activation height for NU6.1 on Mainnet yet or the recipient address for the one-time lockbox disbursement on Mainnet.)

dpoulami: Thanks @arya2, and what about Testnet, do we have the funding stream extension logic implemented for Testnet?

arya2: Yep, NU6.1 is fully implemented for Testnet, including the funding stream extension and one-time lockbox disbursements, which are updated in https://github.com/ZcashFoundation/zebra/commit/a986a5421e3e1319cc004e2913a7a5ed7c13a468.

PR #9604, opened 2025-06-07
feat(rpc): `addnode` method

### LINKED ISSUE #9555: Support the `addnode` RPC method
To enable node interconnection in the ported Python QA tests, we need to implement the `addnode` RPC method:

https://zcash.github.io/rpc/addnode.html

This method is currently used in the test framework at:

https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/test_framework/util.py#L610

Without this, nodes can’t establish direct connections with each other, which is very used in many of the zcashd QA tests we plan to port.

oxarbitrage:
## Motivation

We want to add support for the `addnode` RPC method in Zebra.

Closes https://github.com/ZcashFoundation/zebra/issues/9555

## Solution

Add an `add_peer()` method to the address book types, and expose it through a new `addnode` RPC endpoint.

### Tests

- Added a Rust unit test: `rpc_addnode`.
- Added a QA test script: `addnode.py`

To run the QA test manually:

```bash
cd zebra-rpc
./qa/pull-tester/rpc-tests.py addnode.py
```

https://zcash.github.io/rpc/addnode.html

### Follow-up Work

- Implement support for the `remove` command.
- Evaluate feasibility of supporting the `onetry` command (may not be applicable in Zebra).

### PR Checklist

- [x] The solution is tested.


zebra-network/src/address_book.rs
```diff
@@ -819,6 +819,10 @@ impl AddressBookPeers for AddressBook {
             .cloned()
             .collect()
     }
+
+    fn add_peer(&mut self, peer: PeerSocketAddr) -> bool {
```
arya2, line 823: Let's check if the peer is already in the address book, return a false early if it is, and return true otherwise (`update()` won't always return Some or None when expected here).

zebra-network/src/peer_set/set.rs
```diff
@@ -996,11 +1001,17 @@ where
     /// Given a number of ready peers calculate to how many of them Zebra will
     /// actually send the request to. Return this number.
     pub(crate) fn number_of_peers_to_broadcast(&self) -> usize {
-        // We are currently sending broadcast messages to a third of the total peers.
-        const PEER_FRACTION_TO_BROADCAST: usize = 3;
+        if self.network.is_regtest() {
```
arya2, line 1004: Optional: It would be nice to always broadcast to all peers when a block that has been mined and submitted via the `submitblock` method is being advertised.

zebra-rpc/qa/rpc-tests/addnode.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.nodes[0].generate(1)
+        self.sync_all(False)
+
+    def run_test(self):
+        print("checking connections...")
+
+        # As we connected the nodes to each other, they should have,
+        # at least 4 peers. Poll for that.
+        while (len(self.nodes[0].getpeerinfo()) < 4 or
```
arya2: This should ideally have a timeout so it doesn't hang if it fails.

zebra-rpc/src/methods.rs
```diff
@@ -3926,3 +3988,17 @@ where
...
+pub enum AddNodeCommand {
+    /// Add a node to the address book.
+    #[serde(rename = "add")]
+    Add,
+    /// Remove a node from the address book.
+    #[serde(rename = "remove")]
+    Remove,
+    /// Add a node to the address book, but only try to connect once.
+    #[serde(rename = "onetry")]
+    Onetry,
```
arya2: Optional: Let's get rid of these altogether as we're not planning on supporting it anytime soon.

arya2 (dismissed):

zebra-rpc/src/methods.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9604/commits/1618792dbad7863f7625f0bfcb1d3b72453bb6d0

zebra-rpc/qa/rpc-tests/addnode.py
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9604/commits/a221425fa7a5cfdb059820c613ed2c7e2705787c

zebra-network/src/peer_set/set.rs
oxarbitrage: Let's leave this one out of this PR as it requires more work.

zebra-network/src/address_book.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9604/commits/ee99182b7fbd5161cb123bf8c4b2426f6c8d5843

mpguerra (approved):
Approving on behalf of @arya2

PR #9620, opened 2025-06-12
fix: Pad script in V5 coinbase

### LINKED ISSUE #9614: V5 coinbase txs violate consensus rule at low heights
Zebra-generated V5 coinbase txs with height < 17 and empty coinbase data violate the following consensus rule:

> A coinbase transaction script MUST have length in {2 .. 100} bytes.

This is not an issue for Mainnet or Testnet since those networks don't support V5 txs at those heights.

upbqdn:
## Motivation

- Close #9614 

## Solution

- Include an extra byte in the coinbase data if the data is empty.

### Tests

- Add a test that fails when the consensus rule is not satisfied.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/transparent.rs
+++ b/zebra-chain/src/transparent.rs
@@ -224,8 +224,17 @@ impl Input {
         data: Option<Vec<u8>>,
         sequence: Option<u32>,
     ) -> Input {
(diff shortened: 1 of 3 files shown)


conradoplg (approved):
Looks good, thanks

PR #9624, opened 2025-06-13
Add txid field with transaction hash to `TransactionObject`

### LINKED ISSUE #9617: feature: Add txid field to `TransactionObject`
### Motivation

Block explorer uses getblock verbosity 2 to display a block and its contents and the `txid` field is needed. Zaino currently serves `TransactionObject` as a passthrough so the origin of the data that the Zcash Block Explorer sees is Zebra's.

### Specifications

_No response_

_No response_

### Testing

getblock should include the field txid as it's zcashd counterpart does. 

### Related Work

#8435

pacu:
## Motivation
Add txid field which is needed by the Zcash Block Explorer

## Solution

Add the txid field and test that it deserializes properly by modifying existing tests
### Follow-up Work
Include the rest of the missing fields

### PR Checklist

- [x] The solution is tested.


pacu:
I have unit tested, not manually tested though.

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -190,6 +190,10 @@ pub struct TransactionObject {
     /// The time the transaction was included in a block.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub time: Option<i64>,
+
+    /// The raw transaction, encoded as hex bytes.
```
upbqdn:
```suggestion
    /// The transaction identifier, encoded as hex bytes.
```

upbqdn (dismissed):

upbqdn:
@pacu, you need to update snapshots. You can do that by running 

``` sh
cargo insta test
```

and 

``` sh
cargo insta review
```

upbqdn:
@pacu, you need to update snapshots. You can do that by running 

```cargo insta test```

and 

```cargo insta review```

**Comment by conradoplg** [comment]
~You now need to run `cargo fmt` and commit the changes :sweat_smile:~ (went ahead and did it)

File: zebra-rpc/src/methods/types/transaction.rs [hunk 2]
```diff
@@ -564,6 +569,7 @@ impl TransactionObject {
             },
             size: tx.as_bytes().len().try_into().ok(),
             time: block_time.map(|bt| bt.timestamp()),
+            txid: tx.hash(),
```
arya2, line 572: Could we accept this as a parameter to this function instead? The `getrawtransaction` method accepts the transaction hash as a parameter, and I think it's used by the lightwalletd, so hashing every transaction during the lwd sync may cause a performance regression.

arya2 (changes requested):

zebra-rpc/src/methods/types/transaction.rs
conradoplg: Good catch, I think we should do this.

@pacu let us know, we're happy to make the required changes if you prefer us to

upbqdn: Yet another reason to switch to `zcash_primitives`.

mpguerra:
> @pacu, you need to update snapshots. You can do that by running
> 
> ```shell
> cargo insta test
> ```
> 
> and
> 
> ```shell
> cargo insta review
> ```

Do we need to document any of this in the zebra book or elsewhere for external contributors?

upbqdn:
We could have a whole new section on how to run tests.

zebra-rpc/src/methods/types/transaction.rs
conradoplg: Done in https://github.com/ZcashFoundation/zebra/pull/9650

conradoplg:
This was merged along with https://github.com/ZcashFoundation/zebra/pull/9636 which was built on top of this.

Issue #9633, opened 2025-06-16
feature: Add missing Sapling fields to `TransactionObject`


mpguerra:
## Scope
- bindingSig
- joinSplitPubKey
- joinSplitPubKey

mpguerra:
@elijahhampton this could be an option. Please assign yourself if you would like to work on it.

elijahhampton:
> @elijahhampton this could be an option. Please assign yourself if you would like to work on it.

This looks good. I'll start work on this today, but I am not sure if I can assign myself. Thanks.

Issue #9632, opened 2025-06-16
feature: Add missing Orchard fields to `TransactionObject`


mpguerra:
## Scope

- orchard.flags
- orchard.anchor
- orchard.proof
- orchard.bindingSig

mpguerra:
@elijahhampton this could be an option. Please assign yourself if you would like to work on it.

PR #9636, opened 2025-06-16
feat(rpc): add missing fields to getrawtransaction

### LINKED ISSUE #9634: feature: Add requested fields to `TransactionObject`
## Motivation

Splitting off fields that @conradoplg is already working on

## Scope

- in_active_chain
- authdigest
- overwintered
- version
- versiongroupid
- expiryheight
- blockhash
- blocktime
- locktime

conradoplg:
## Motivation

We want to fill the remaining fields of `getrawtransaction`. This adds the fields which are not specific to Sapling or Orchard.

Closes #9634 

Based on https://github.com/ZcashFoundation/zebra/pull/9624, I can't target it since it's an external PR so this PR will show its contents too

## Solution

Simply adds the fields. Most of them were straightforward.

### Tests

Existing tests were adjusted.

I also tested it manually with `zcash-rpc-diff` and it matches `zcashd`, with the exception of `in_active_chain` which is only filled when a block hash is passed in zcashd, but I always filled it regardless (set to false for mempool txs, true otherwise). I can change it to match zcashd behaviour, but I think it's harmless...

### Follow-up Work

#9633 and #9632 for the rest of the fields.

We might want to support gettings txs from side chains when `blockhash` is specified.

### PR Checklist

- [ ] The solution is tested.


oxarbitrage (approved):
Looks good to me, thanks @conradoplg

PR #9627, opened 2025-06-13
Remove `debug_like_zcashd` config option

upbqdn:
## Motivation

- Close #9599

## Solution

- Remove the option from config, code, tests and docs.
- Minor cleanups.

### Tests

- No new tests.
- Existing tests updated to handle V5 coinbase txs instead of V4.

### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The solution is tested.
- [x] The documentation is up to date.

Issue #9605, opened 2025-06-09
Construct coinbase txs using tx builder in `zcash_primitives`


upbqdn:
Use the [Builder](https://docs.rs/zcash_primitives/latest/zcash_primitives/transaction/builder/struct.Builder.html) struct from `zcash_primitives`, which has support for Sapling and Orchard shielded data, to construct coinbase transactions. We currently construct them manually without any shielded data.

PR #9639, opened 2025-06-17
Temporarily activate NU5 at height 1 in tests for Regtest

### LINKED ISSUE #9599: Remove `debug_like_zcashd` config option
- Users don't need this option.
- Keeping it complicates support for mining to shielded addresses.

upbqdn:
## Motivation

After #9627, Zebra temporarily constructs only V5 coinbase txs. Currently, we activate NU5 at height 100 and Canopy at height 1 in tests for Regtest, which makes those tests that generate coinbase txs in that height range fail since Canopy doesn't support V5 txs. We will support generating V4 coinbase txs again once we finish #9605, so we can revert this PR then.

## Solution

- Activate NU5 at height 1 in tests for Regtest.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/parameters/network/testnet.rs
+++ b/zebra-chain/src/parameters/network/testnet.rs
@@ -669,8 +669,11 @@ impl Parameters {
     pub fn new_regtest(
         ConfiguredActivationHeights { nu5, nu6, nu7, .. }: ConfiguredActivationHeights,
     ) -> Self {
-        #[cfg(any(test, feature = "proptest-impl"))]
-        let nu5 = nu5.or(Some(100));
+        let (canopy, nu5) = if cfg!(test) || cfg!(feature = "proptest-impl") {
+            (None, nu5.or(Some(1)))
+        } else {
+            (Some(1), nu5.or(Some(100)))
+        };
 
         let parameters = Self::build()
             .with_genesis_hash(REGTEST_GENESIS_HASH)
@@ -682,7 +685,7 @@ impl Parameters {
             // Removes default Testnet activation heights if not configured,
             // most network upgrades are disabled by default for Regtest in zcashd
             .with_activation_heights(ConfiguredActivationHeights {
-                canopy: Some(1),
+                canopy,
                 nu5,
                 nu6,
                 nu7,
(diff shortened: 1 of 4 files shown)


zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -669,8 +669,12 @@ impl Parameters {
     pub fn new_regtest(
         ConfiguredActivationHeights { nu5, nu6, nu7, .. }: ConfiguredActivationHeights,
     ) -> Self {
-        #[cfg(any(test, feature = "proptest-impl"))]
-        let nu5 = nu5.or(Some(100));
+        // TODO: Activate Canopy at height 1 at NU5 at height 100 once #9605 is done.
+        let (canopy, nu5) = if cfg!(test) || cfg!(feature = "proptest-impl") {
+            (None, nu5.or(Some(1)))
+        } else {
+            (Some(1), nu5.or(Some(100)))
```
arya2, line 676: We should leave these values as what was passed in except when compiling for tests, notably so that the regular testnet chain up to NU5 activation will continue working.

upbqdn: Is that not what the code does?

arya2: Ah, no, but my comment was also wrong/confusing, it's not possible use the regular Testnet chain with Regtest because it has a different genesis block hash.

I think we still don't want to change the Canopy activation height here except when compiling for tests, so we should only update `let nu5 = nu5.or(Some(100));` to `let nu5 = nu5.or(Some(1));` and leave the rest of this function unchanged.

It shouldn't set the nu5 activation height to 100 if it is undefined when not compiled for tests.

arya2 (approved):
The rest of this PR looks good, it would be nice to address https://github.com/ZcashFoundation/zebra/pull/9639/files#r2153296135 but it's not strictly necessary (and won't matter unless users have another way to generate block templates with V4 coinbase).

upbqdn:
If we merge #9647, we can close this one.

arya2:
https://github.com/ZcashFoundation/zebra/pull/9647 seems better, I think we should commit to merging that one and close this one.

PR #9645, opened 2025-06-18
Block proposal validation of mempool transactions

### LINKED ISSUE #9301: DoS vulnerability in `getblocktemplate` RPC

arya2:
Will close #9301.

TODO:
- Update mempool's `AwaitOutput` request to accept a transaction hash, and the response for that request to include all direct and indirect transaction dependencies,
- Update the `StateService` to convert `Request::ChainInfo` to a new `ReadRequest::ChainInfo` variant and redirect those requests from the state service to the read state service,
- Update the `download_and_verify()` method of the `TxDownloads` component in the mempool's active state to:
  - Generate minimal (including only the semantically verified transaction and any transactions that it directly or indirectly depends on) valid block templates with the same logic as the `getblocktemplate` RPC,
  - Convert those block templates to block proposals 
  - Call the `block_router_verifier` with `Proposal` requests (@oxarbitrage this part can't be moved to a function in `zebra-state` because it relies on `zebra-consensus`), and
  - Return an error from the `TxDownloads` stream if the block proposal for a semantically verified transaction is not valid.

@oxarbitrage It doesn't seem necessary or cleaner to move block template generation logic from `zebra-rpc` to `zebra-state` as I had suggested earlier. I think we should leave that logic in `zebra-rpc`.

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -224,6 +224,10 @@ pub enum Response {
         /// in the mempool and to avoid adding transactions with missing spends
         /// to its verified set.
         spent_mempool_outpoints: Vec<transparent::OutPoint>,
+
+        /// Transactions in the mempool on which the verified transaction
+        /// directly or indirectly depends.
+        mempool_dependencies: Vec<Arc<Transaction>>,
     },
 }
 
@@ -233,6 +237,7 @@ impl From<VerifiedUnminedTx> for Response {
         Response::Mempool {
             transaction,
             spent_mempool_outpoints: Vec::new(),
+            mempool_dependencies: Vec::new(),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


arya2:
This approach has every transaction in the mempool being verified twice. It would be nice to avoid that, maybe by getting outputs spent by a tx in the mempool's downloads component instead of the tx verifier, and removing the tx verifier's Request::Mempool variant (though that has its own issues).

We could also eagerly verify mempool transactions, queue a task to verify them as block proposals after they've been added to the mempool storage, then remove them from the storage if they formed invalid block proposals.

See [discussion here](https://github.com/ZcashFoundation/zebra/discussions/9664).

arya2:
Closing in favour of the design outlined [here](https://github.com/ZcashFoundation/zebra/tree/mempool-block-proposals-design).

PR #9646, opened 2025-06-18
feat(config): add figment for configuration management

gustavovalverde:
## Motivation

This pull request migrates Zebra's configuration system from the Abscissa framework to `figment`. The primary goals of this change, as outlined in our design documents, are to:

*   **Improve Maintainability:** Eliminate the complex shell scripting in the Docker entrypoint that was responsible for generating `zebrad.toml` on the fly.
*   **Increase Robustness:** Adopt a mature, well-tested configuration library with clear precedence rules and features.
*   **Enhance User Experience:** Provide a standard and intuitive way for users to configure `zebrad` using a combination of a TOML file and environment variables, which is a common pattern in the Rust, and others, ecosystem.
*   **Reduce Technical Debt:** Remove custom configuration logic and align with modern best practices.

## Solution

The core of this implementation is the introduction of `ZebradConfig::load()`, which uses `figment` to layer configuration from three sources in a strict order of precedence:

1.  **Defaults:** Sensible default values are hard-coded directly in the configuration structs.
2.  **TOML File:** A `zebrad.toml` file can be provided to override the defaults.
3.  **Environment Variables:** Any setting can be overridden at runtime via environment variables, which have the highest precedence.

### Key changes:

*   **Environment Variable Convention:** All configuration fields can be overridden by environment variables. The variable name is constructed by prefixing with `ZEBRA_` and joining nested keys with a double underscore (`__`). For example, the `[rpc]` setting `listen_addr` is controlled by the `ZEBRA_RPC__LISTEN_ADDR` environment variable.
*   **Application Integration:** The main application logic in `zebrad` has been updated to use the new `ZebradConfig::load()` method, removing the previous Abscissa config-loading code.
*   **Docker Compatibility:** To avoid a breaking change for existing users, the Docker `entrypoint.sh` script has been updated. Instead of generating a config file, it now acts as a compatibility layer, translating the old-style environment variables (`NETWORK`, `ZEBRA_RPC_PORT`, etc.) into the new `figment`-compatible format and exporting them. This ensures existing Docker setups continue to work without modification.
*   **Documentation:** All relevant Rustdoc and user-facing documentation (specifically `startup.md`) have been updated to explain the new configuration system, its precedence, and the environment variable conventions.

### Tests

A new integration test suite for configuration has been added at `zebra/zebrad/tests/config.rs`. These tests validate:
*   Loading configurations from TOML files.
*   Correct application of default values.
*   Environment variable overrides for various data types and nested structures.
*   The correctness of the precedence order (Environment > TOML > Defaults).
*   Error handling for invalid configurations, such as incorrect data types.

**PENDING:** The existing test suite was also run to ensure no regressions were introduced by these changes.

*   **Reasoning:** [docs/decisions/zebra/0001-figment-config-system.md](docs/decisions/zebra/0001-figment-config-system.md)

### Follow-up Work

*   The Docker-specific user documentation (`book/src/user/docker.md` and `book/src/user/mining-docker.md`) needs to be updated to reflect the new configuration system. This was intentionally deferred to a future PR; in the meanwhile this is backward-compatible.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.github/workflows/ci-coverage.yml
+++ b/.github/workflows/ci-coverage.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Skip network tests on Ubuntu
         # Ubuntu runners don't have reliable network or DNS during test steps.
         shell: bash
-        run: echo "ZEBRA_SKIP_NETWORK_TESTS=1" >> $GITHUB_ENV
+        run: echo "TESTS_ZEBRA_SKIP_NETWORK=1" >> $GITHUB_ENV
 
       - name: Minimise proptest cases in Coverage tests
         # We set cases to 1, because some tests already run 1 case by default.
--- a/.github/workflows/ci-unit-tests-os.yml
+++ b/.github/workflows/ci-unit-tests-os.yml
@@ -147,7 +147,7 @@ jobs:
         # Windows runners have an unreliable network.
         shell: bash
         if: matrix.os != 'macos-latest'
-        run: echo "ZEBRA_SKIP_NETWORK_TESTS=1" >> $GITHUB_ENV
+        run: echo "TESTS_ZEBRA_SKIP_NETWORK=1" >> $GITHUB_ENV
 
       - name: Minimise proptest cases on macOS and Windows
         # We set cases to 1, because some tests already run 1 case by default.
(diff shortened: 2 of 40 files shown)


gustavovalverde:
This will be replaced with config-rs because Figment is not being maintained

PR #9657, opened 2025-06-23
Individual service impls

AloeareV:
## Motivation

Currently, `ReadStateService`'s `Service` implementation uses large `ReadRequest`/`ReadResponse` enums. There are a lot of places where zaino needs to `call` the service, provide a `ReadRequest`, and assert that the `ReadResponse` is of the appropriate kind. Also, the set of errors returned is the entire set of possible errors for all requests.

## Solution

If we add an individual `Service` implementation for each distinct method offered, each service can have its own concrete response and error types, cutting down on a lot of matching on potentially-unreachable state.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-scan/src/service/scan_task/scan.rs
+++ b/zebra-scan/src/service/scan_task/scan.rs
@@ -38,7 +38,9 @@ use zebra_chain::{
     transaction::Transaction,
 };
 use zebra_node_services::scan_service::response::ScanResult;
-use zebra_state::{ChainTipChange, ReadStateService, SaplingScannedResult, TransactionIndex};
+use zebra_state::{
+    ChainTipChange, ReadRequest, ReadStateService, SaplingScannedResult, TransactionIndex,
+};
 
 use crate::{
     service::{ScanTask, ScanTaskCommand},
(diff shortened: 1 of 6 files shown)


arya2 (commented):
This could be very nice if it's possible to define a supertrait of all the `Service<Request>` traits implementations of the state services to avoid verbose trait constraints where the state services are passed into functions as generic.

It would also require an update to the `MockService` used in testing.

The changes required for this approach would likely be very substantial.

arya2:
This approach would require:
- updating 400-900 references, 
- adding a complicated feature to `MockService`, 
- updating all of the trait constraints, and
- changes to the `StateService`, `ReadStateService`, their request/response/error types, and related code.

Much of this code is very sensitive, and there are likely other unforeseen requirements as well, so I don't think this approach is practical at this time, but it would be very nice for code organization and simplification, so it would be nice to open an issue to track the idea.

The easier solution, clearly boxing up a particular error type for each request to assure callers that downcasting to that error type will succeed, should work for our purposes in the near-term, and I would recommend that approach instead.

AloeareV:
> The easier solution, clearly boxing up a particular error type for each request to assure callers that downcasting to that error type will succeed, should work for our purposes in the near-term, and I would recommend that approach instead.

This would be fantastic! I am still not in love with downcasting...but this would make downcasting about as painless as downcasting can be.

PR #9658, opened 2025-06-23
refactor(validate_address): expose logic for consumers

dorianvp:
## Motivation

Zaino needs to expose a `validateaddress` JSON-RPC endpoint, which Zebra already implements.

## Solution

Moving `validate_address`'s body into a public function should allow consumers to re-use the address validation logic.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -89,6 +89,7 @@ use zebra_state::{HashOrHeight, OutputLocation, ReadRequest, ReadResponse, Trans
 
 use crate::{
     config,
+    methods::types::validate_address::validate_address,
     queue::Queue,
     server::{
         self,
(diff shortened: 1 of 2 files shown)


conradoplg:
I'm not sure we want `zebra-rpc` to serve as a library for random functionality. Since the function is small, isn't it better to just copy it to Zaino, or just route the request to Zebra?

arya2 (approved):
Looks good!

arya2:
> I'm not sure we want `zebra-rpc` to serve as a library for random functionality. Since the function is small, isn't it better to just copy it to Zaino, or just route the request to Zebra?

I agree that at least some of it should probably go in `zebra-consensus`, `zebra-chain` or `zcash_address`. 

I think it'd fine to add the function here for now and move it later (it would be nice to avoid duplicating the code), but since most of the logic is already in the `FromStr` impl for `ZcashAddress` from the `zcash_address` crate already, copying the rest to Zaino would also be fine.

arya2:
@gustavovalverde any idea why CI on this PR seems stuck?

conradoplg:
> @gustavovalverde any idea why CI on this PR seems stuck?

All contributor PRs are getting stuck, I don't know the reason, but we've been admin-merging those. Let me know if you'd like for me to merge this.

arya2:
> Let me know if you'd like for me to merge this.

Yes please, it can't hurt to have it factored out, even if it ends up not being used in Zaino.

PR #9670, opened 2025-07-01
Count sigops using generics

### LINKED ISSUE #9669: Count sigops for multiple tx types
We currently count sigops for `zebra_chain::transaction::Transaction` only. To be able to switch to the tx builder in #9605, we'll need to count sigops for `zcash_primitives::transaction::Transaction` as well. It'll also be handy to count sigops for `CachedFfiTransaction` and `zebra_chain::transaction::UnminedTx`.

upbqdn:
## Motivation

- Close #9669.

## Solution

- Define a new `Sigops` trait.
- Impl the trait for relevant tx types.
- Clean-ups:
	- Use `u32` instead of `u64` for the sigops count since that's what the script interpreter, which returns the counts for tx inputs and outputs, uses.
	- Move tests to their own files.

### PR Checklist

- [x] The solution is tested.


upbqdn:
@gustavovalverde, could you take a look at the timeouts? They don't seem related to the PR.

zebra-script/src/lib.rs
```diff
@@ -177,283 +180,63 @@ impl CachedFfiTransaction {
...
-                let script = unlock_script.as_raw_bytes();
-                interpreter
-                    .legacy_sigop_count_script(script)
-                    .map_err(Error::from)?
-            }
-            transparent::Input::Coinbase { .. } => 0,
-        } as u64;
+/// Trait for counting the number of transparent signature operations
+/// in the transparent inputs and outputs of a transaction.
+pub trait Sigops {
```
upbqdn, line 185: Wouldn't it be better to have this trait in `zcash_script`?

zebra-script/src/lib.rs
```diff
@@ -177,283 +180,63 @@ impl CachedFfiTransaction {
...
-        Ok(())
+impl Sigops for zcash_primitives::transaction::Transaction {
+    fn scripts(&self) -> impl Iterator<Item = &[u8]> {
+        self.transparent_bundle().into_iter().flat_map(|bundle| {
+            (!bundle.is_coinbase())
+                .then(|| bundle.vin.iter().map(|i| i.script_sig.0.as_slice()))
+                .into_iter()
+                .flatten()
+                .chain(bundle.vout.iter().map(|o| o.script_pubkey.0.as_slice()))
+        })
```
arya2, line 243: Is this being tested or used anywhere?

arya2 (dismissed):
Look good to me as long as there's a test for the `Sigops` impl for the `zcash_primitives` transaction type.

It would be nice to have `Sigops` implemented in `zcash-script`, but not necessary.

I particularly like the use of `try_fold()`.

zebra-script/src/lib.rs
upbqdn: I use it three branches up the stack of the shielded coinbase saga. :D I'll try thinking of some vectors for tests.

zebra-script/src/lib.rs
upbqdn: Actually, no, because then I'd have to pull `zcash_script` into `zcash_primitives` to be able to implement the trait.

zebra-script/src/lib.rs
upbqdn: I think tests will be much easier to add once I have the handling of the tx builder done, so I can then use it to create txs for which I'll know the sigops count.

upbqdn: The main purpose of this PR is to have sigops counting ready for `zcash_primitives::transaction::Transaction` so that the subsequent PRs are easier to review.

zebra-script/Cargo.toml
```diff
@@ -23,7 +23,9 @@ comparison-interpreter = []
 
 [dependencies]
 zcash_script = { workspace = true }
-zebra-chain = { path = "../zebra-chain", version = "1.0.0" }
+zcash_primitives = { workspace = true }
+
+zebra-chain = { path = "../zebra-chain", version = "1.0.0-beta.46" }
```
conradoplg:
```suggestion
zcash_primitives = { workspace = true }
zebra-chain = { path = "../zebra-chain", version = "1.0.0" }
```

(or `workspace = true` for it too?)

zebra-script/src/lib.rs
```diff
@@ -177,283 +180,63 @@ impl CachedFfiTransaction {
...
-            }
-            transparent::Input::Coinbase { .. } => 0,
-        } as u64;
+/// Trait for counting the number of transparent signature operations
+/// in the transparent inputs and outputs of a transaction.
+pub trait Sigops {
+    /// Returns the number of transparent signature operations in the
+    /// transparent inputs and outputs of the given transaction.
+    fn sigops(&self) -> Result<u32, zcash_script::Error> {
+        let interpreter = get_interpreter(
```
conradoplg, line 190:
```suggestion
        // Create a dummy interpreter since these inputs are not used to count
        // the sigops
        let interpreter = get_interpreter(
```

zebra-script/src/lib.rs
```diff
@@ -177,283 +180,63 @@ impl CachedFfiTransaction {
...
-        verifier.is_valid(input_index)?;
-
-        Ok(())
+impl Sigops for zebra_chain::transaction::Transaction {
+    fn scripts(&self) -> impl Iterator<Item = &[u8]> {
+        self.inputs()
+            .iter()
+            .map(|input| match input {
+                transparent::Input::PrevOut { unlock_script, .. } => unlock_script.as_raw_bytes(),
+                transparent::Input::Coinbase { .. } => &[],
```
conradoplg: I'm a bit worried about passing an empty slice to `legacy_sigop_count_script()` which goes all the way down to C++ code. It looks like it works, but I wonder if it's safer to skip those by using `filter_map()` instead of `map()`

zebrad/tests/acceptance.rs
```diff
@@ -2973,8 +2973,8 @@ async fn trusted_chain_sync_handles_forks_correctly() -> Result<()> {
 
     let _init_guard = zebra_test::init();
 
-    let network = Network::new_regtest(Default::default());
-    let mut config = os_assigned_rpc_port_config(false, &network)?;
+    let net = Network::new_regtest(Default::default());
+    let mut config = os_assigned_rpc_port_config(false, &net)?;
```
conradoplg, line 2977: I don't mind leaving this here for now but I think we should decide as a team on whether we start renaming `network` to `net` over the code base

conradoplg (commented):
Looks good, left some fixes and suggestions

conradoplg (changes requested):
(blocking on the import fix)

zebra-script/src/lib.rs
upbqdn: What do "these inputs" refer to?

zebrad/tests/acceptance.rs
upbqdn: I implicitly thought it would be a good idea. I'm happy to discuss.

zebra-script/src/lib.rs
conradoplg: The inputs to `get_interpreter` - feel free to improve the message, I just wanted to keep the spirit of the original comment (explain why it's OK to use dummy parameters) which got lost in the refactoring

zebra-script/Cargo.toml
upbqdn: > (or `workspace = true` for it too?)

That doesn't work.

zebra-script/src/lib.rs
upbqdn: Ah, you don't mean "tx inputs" but the fn args. It took me a while. :D

zebra-script/Cargo.toml
upbqdn: Commited in [0bdd5b2](https://github.com/ZcashFoundation/zebra/pull/9670/commits/0bdd5b28e90458f1f4f6991d3ce4672b16fb5917).

zebra-script/src/lib.rs
upbqdn: Addressed in [Clarify use of dummy values in sigops counting](https://github.com/ZcashFoundation/zebra/pull/9670/commits/39cdd625b7918269544d2baf808274dca5c7cee7).

zebrad/tests/acceptance.rs
upbqdn: Done in [Filter coinbase inputs from script iterator in Sigops implementation](https://github.com/ZcashFoundation/zebra/pull/9670/commits/11dd9b1e69090d18122ab3c255eb763f6db0d8fb).

zebra-script/src/lib.rs
upbqdn: Done in [Filter coinbase inputs from script iterator in Sigops implementation](https://github.com/ZcashFoundation/zebra/pull/9670/commits/11dd9b1e69090d18122ab3c255eb763f6db0d8fb).

conradoplg (approved):

PR #9678, opened 2025-07-03
change: Refactor Sapling parameters loading

### LINKED ISSUE #9676: Load Sapling params in format compatible with `sapling-crypto`
We currently load the spend and output Sapling params manually from `wagyu-zcash-parameters` in a format compatible directly with bellman. We currently use these params in the tx verifier. To implement #9605, we need to load them in a format compatible with `sapling-crypto`, which wraps these parameters, so that we can pass them to the tx builder from `zcash_primitives`.

upbqdn:
## Motivation

Close #9676.

## Solution

- Load the params from `zcash-proofs` instead of `wagyu-zcash-parameters`.
- Based atop #9764.

### PR Checklist

- [x] The solution is tested.


Cargo.toml
```diff
@@ -149,11 +150,22 @@ tracing-subscriber = "0.3.19"
...
+sapling-crypto = { git = "https://github.com/zcash/sapling-crypto", branch = "access-inner-param" }
+
+zcash_primitives = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_address = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_history = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_transparent = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash", branch = "update-prover-api" }
```
conradoplg, line 167: Do we have upstream PRs for the required changes? It would be good to get them merged otherwise it might block new releases of zebra

conradoplg (changes requested):
Looks good, but it would be better to get the required changes merged upstream

Cargo.toml
upbqdn: Ah, I just realised I deleted the comment  saying this PR needs to be reviewed together with #9737, which removes the deps.

upbqdn: We used to have upstream PRs for those branches in `sapling-crypto` and lrz, but I closed them in favor of #9737. 

I think it would be ideal to merge this PR, and then #9737 right after it.

Or merge #9737 into this PR first.

upbqdn: I'll delete the upstream branches as well once we get #9737 in.

conradoplg: Got it, thanks! I'll review #9737 before merging this one just to be sure we can get them both merged

conradoplg (approved):

PR #9709, opened 2025-07-17
fix(state): fix 2.4.0 DB upgrade; add warning if impacted

### LINKED ISSUE #9685: bug: getaddressutxos returning thousands of unrelated UTXOs for address with zero balance
### What happened?

```
curl -u username:password -d '{"jsonrpc": "2.0", "id": "1", "method": "getaddressutxos", "params": [{"addresses": ["t1UVpBZR9VSJi2V28xhfXFn7pQz4VFkokp3"]}]}' -H 'content-type: text/plain;' http://127.0.0.1:8232/
```

Returns thousands and thousands of unrelated UTXOs.

(address plucked at random from blockchain to illustrate the point)

My wallet is failing because these UTXOs being returned from `getaddressutxos` have nothing to do with the addresses my wallet is sending a request for, which is confusing it. The example query in this bug report is not from my wallet but it produces the same bizarre result.

### Zebra logs

```
DEBUG read_state: zebra_state::service::read::address::utxo: starting address UTXO query attempt=0 address_count=1
DEBUG read_state: zebra_state::service::read::address::utxo: finalized address UTXO response finalized_utxo_count=8453 finalized_tip_range=Some(Height(2984169)..=Height(2984169)) address_count=1 attempt=0
DEBUG read_state: zebra_state::service::read::address::utxo: chain address UTXO response chain_utxo_count=0 chain_utxo_spent=0 address_count=1 attempt=0
DEBUG read_state: zebra_state::service::read::address::utxo: full address UTXO response full_utxo_count=8453 tx_id_count=5599 address_count=1 attempt=0
```

### Zebra Version

zebrad 2.3.0+47.g6e0a403.modified

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

Looks like `ZebraDb::address_balance_location` is returning (for that address)

```
AddressBalanceLocation(AddressBalanceLocationInner {
    balance: Amount<zebra_chain::amount::NonNegative>(0),
    received: 100000000,
    location: OutputLocation {
        transaction_location: TransactionLocation {
            height: Height(0),
            index: TransactionIndex(0)
        },
        output_index: OutputIndex(0)
        
    }
})
```

because the address had previously received coins but then spent them all and has no unspent UTXOs.

The `OutputLocation` of this is then being fed into `address_utxo_locations` which is returning a ton of unrelated UTXOs.

conradoplg:
## Motivation

The last database upgrade unfortunately had a bug:

- In the upgrade, we create an empty AddressBalanceLocationChange (if a previous doesn't exist for the address) and update its received field [https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec8557[…]ed_state/disk_format/upgrade/block_info_and_address_received.rs](https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec85576c9b/zebra-state/src/service/finalized_state/disk_format/upgrade/block_info_and_address_received.rs#L123)
- We add it as a merge to the batch in [https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec8557[…]ed_state/disk_format/upgrade/block_info_and_address_received.rs](https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec85576c9b/zebra-state/src/service/finalized_state/disk_format/upgrade/block_info_and_address_received.rs#L205). Note however that only the received field is set, balance and location are set to 0
- Upon merge, we simply read the existing value as a Change and add to other changes [https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec8557[…]zebra-state/src/service/finalized_state/zebra_db/transparent.rs](https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec85576c9b/zebra-state/src/service/finalized_state/zebra_db/transparent.rs#L70)
- The add operator, however, sets the location to the mininum of both locations; since the change was set to 0, it will set the final value to 0 [https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec8557[…]ra-state/src/service/finalized_state/disk_format/transparent.rs](https://github.com/ZcashFoundation/zebra/blob/cb736ef38b6950e5529e765ea5a73eec85576c9b/zebra-state/src/service/finalized_state/disk_format/transparent.rs#L225)
- The end result being all locations being set to 0 during upgrade

Closes https://github.com/ZcashFoundation/zebra/issues/9685

## Solution

This fixes the underlying bug, but impacted users will still have to do a recovery procedure with `copy-state` which we will explain in the 2.4.1 release notes.

This also adds a warning to impacted users.

### Tests

I tested this manually:

- From a v26 backup, I started the node. It did the database upgrade. I confirmed that it worked by checking that  `zcash-cli  getaddressutxos '{"addresses": ["t1UVpBZR9VSJi2V28xhfXFn7pQz4VFkokp3"]}'` returned empty. The lack of warning also confirmed it worked.
- Running from my corrupted v27 state, I confirmed that the warning was logged.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format/transparent.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/transparent.rs
@@ -224,12 +224,16 @@ impl<C: Constraint + Copy + std::fmt::Debug> std::ops::Add for AddressBalanceLoc
             // Keep in mind that `AddressBalanceLocationChange` reuses this type
             // (AddressBalanceLocationInner) and this addition method. The
             // `block_info_and_address_received` database upgrade uses the
-            // empty/zero location as a dummy value for
-            // `AddressBalanceLocationChange`. Therefore, when adding, we should
-            // ignore these dummy values (which are 0). Using `max` achieves
-            // this. It is also possible that two zero-location balance changes
-            // are added, and `max` will correctly keep the zero value.
-            location: self.location.max(rhs.location),
+            // usize::MAX dummy location value returned from
+            // `AddressBalanceLocationChange::empty()`. Therefore, when adding,
+            // we should ignore these dummy values. Using `min` achieves this.
+            // It is also possible that two dummy-location balance changes are
+            // added, and `min` will correctly keep the same dummy value. The
+            // reason we haven't used zero as a dummy value and `max()` here is
+            // because we use the minimum UTXO location as the canonical
+            // location for an address; and using `min()` will work if a
+            // non-canonical location is added.
+            location: self.location.min(rhs.location),
         })
     }
 }
--- a/zebra-state/src/service/finalized_state/disk_format/transparent.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/transparent.rs
@@ -221,7 +221,15 @@ impl<C: Constraint + Copy + std::fmt::Debug> std::ops::Add for AddressBalanceLoc
         Ok(AddressBalanceLocationInner {
             balance: (self.balance + rhs.balance)?,
             received: self.received.saturating_add(rhs.received),
-            location: self.location.min(rhs.location),
+            // Keep in mind that `AddressBalanceLocationChange` reuses this type
+            // (AddressBalanceLocationInner) and this addition method. The
+            // `block_info_and_address_received` database upgrade uses the
+            // empty/zero location as a dummy value for
+            // `AddressBalanceLocationChange`. Therefore, when adding, we should
+            // ignore these dummy values (which are 0). Using `max` achieves
+            // this. It is also possible that two zero-location balance changes
+            // are added, and `max` will correctly keep the zero value.
+            location: self.location.max(rhs.location),
         })
     }
 }
--- a/zebra-state/src/service/finalized_state/disk_format/upgrade/block_info_and_address_received.rs
+++ b/zebra-state/src/service/finalized_state/disk_format/upgrade/block_info_and_address_received.rs
@@ -307,10 +307,9 @@ impl AddressBalanceLocationChange {
     /// a dummy (all one bits) location. See `AddressBalanceLocationInner::add()`
     /// for the rationale for using this dummy value.
     fn empty() -> Self {
-        Self::new(AddressLocation::from_usize(
-            MAX_ON_DISK_HEIGHT,
-            usize::MAX,
-            usize::MAX,
+        Self::new(AddressLocation::from_output_index(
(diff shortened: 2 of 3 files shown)


zebra-state/src/service/finalized_state/disk_format/transparent.rs
```diff
@@ -221,7 +221,15 @@ impl<C: Constraint + Copy + std::fmt::Debug> std::ops::Add for AddressBalanceLoc
...
-            location: self.location.min(rhs.location),
+            // Keep in mind that `AddressBalanceLocationChange` reuses this type
+            // (AddressBalanceLocationInner) and this addition method. The
+            // `block_info_and_address_received` database upgrade uses the
+            // empty/zero location as a dummy value for
+            // `AddressBalanceLocationChange`. Therefore, when adding, we should
+            // ignore these dummy values (which are 0). Using `max` achieves
+            // this. It is also possible that two zero-location balance changes
+            // are added, and `max` will correctly keep the zero value.
+            location: self.location.max(rhs.location),
```
arya2: We _could_ make this work, but it looks like a foot-gun to me.

It would be better to update `AddressBalanceLocationChange::empty()` to use the max `TransactionLocation` in the add block info and address received balances upgrade.

```rust

impl AddressBalanceLocationChange {
    /// Creates a new [`AddressBalanceLocationChange`] with all zero values and a dummy (all one bits) location.
    fn empty() -> Self {
        Self::new(AddressLocation::from_usize(
            crate::MAX_ON_DISK_HEIGHT,
            usize::MAX,
            usize::MAX,
        ))
    }
}

```

```suggestion
            location: self.location.min(rhs.location),
```



If you want to keep this as `max`, we should update [this](https://github.com/ZcashFoundation/zebra/blob/main/zebra-state/src/service/finalized_state/zebra_db/transparent.rs#L514-L516) too to get the existing location, if any, from `address_balances`

conradoplg: I agree. (For anyone reading, the reason it's a foot-gun is because we use the mininum UTXO location for the "canonical" location of an address. We might rely on `min` to preserve that minimum location when adding a AddressBalanceLocationChange to one that might have a (larger) non-canonical location.

Changed in eee60725b

arya2 (approved):
This looks good. Thank you for diagnosing and fixing this.

PR #9711, opened 2025-07-18
fix(release): allow `deploy-nodes` job to run for releases

gustavovalverde:
## Motivation

The `deploy-nodes` job was being skipped for releases because it depended on get-disk-name which excluded release events. Now get-disk-name runs for releases but skips disk lookup and provides default outputs, allowing deploy-nodes to proceed with production deployment flow.

## Solution
- Remove release exclusion from get-disk-name job condition
- Make disk-finding conditional within sub-find-cached-disks workflow
- Add release defaults step to provide empty cached disk outputs

### Tests

This can only be tested on releases, unfortunately.

### PR Checklist

- [ ] The solution is tested.


conradoplg (approved):
LGTM

PR #9712, opened 2025-07-18
fix(ci): prevent cached state creation from failed tests

gustavovalverde:
## Motivation

The create-state-image job condition was using `!cancelled() && !failure()` which checks workflow-level status, not job-specific results. This caused disk images to be created even when the test-result job was cancelled, failed, or timed out, leading to corrupted cached states.

## Solution

Change condition to `needs.test-result.result == 'success'` to only create disk images when the specific test actually succeeded.

### PR Checklist

- [ ] The solution is tested.


oxarbitrage (approved):
Looks good.

PR #9710, opened 2025-07-17
change(consensus): Allow configuration of funding streams on Regtest

arya2:
## Motivation

We want to allow for configuring funding streams on Regtest (as is permitted for custom Testnets) to match zcashd and to enable some tests.

## Solution

- Updates `new_regtest()` to accept a `RegtestParameters` struct instead of `ConfiguredActivationHeights`
- Passes configured funding streams to `new_regtest()` when deserializing network config
- Adds funding streams to debug info for network
- Removes funding streams from comparison in `is_regtest()`
- Updates `with_{pre, post}_nu6_funding_streams()` to accept an `Option` and set dummy values when passed None

Related:
- Addresses some TODOs now that testnet parameter serialization is working
  - Stops overwriting NU5 activation height on Regtest for test compilations
  - Always checks funding streams validity for the network when constructing testnet parameters (rather than skipping the check in tests)
- Adds a `ParametersBuilder::clear_funding_streams()` method for conveniently clearing funding streams from the testnet params in tests
- Allows for configuring all activation heights on Regtest now that coinbase construction in Zebra should work pre-Canopy
  - Adds a `for_regtest()` method to `ConfiguredActivationHeights` to preserve default behaviour of `new_regtest()`

### Tests

- Adds a new test to check that funding streams are configured as expected on Regtest
- Updates existing tests to explicitly set NU5 activation height
- Use functionality, improve and update the python framework in general. Also close https://github.com/ZcashFoundation/zebra/issues/9557 

Related:
- Updates `check_full_activation_list()` to check NU6 and NU6.1

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [ ] The solution is tested.
- [ ] The documentation is up to date.

Issue #9557, opened 2025-05-27
Create and use a wallet.dat with miner address for QA tests


oxarbitrage:
Before zallet can start in the QA tests, we need to ensure that a miner address is already present in its database.

One way to achieve this is by creating a wallet.dat file containing a transparent address, and then importing it into zallet using the functionality introduced in [zcash/wallet#152](https://github.com/zcash/wallet/pull/152).

blocked on https://github.com/zcash/wallet/pull/152

PR #9714, opened 2025-07-20
tests(qa): Ensure funds are available in wallet tests

oxarbitrage:
## Motivation

A core requirement for all wallet-related tests is to ensure that the wallet has access to funds. This enables testing of transaction creation, balance tracking, and related functionality.

Closes [#9557](https://github.com/ZcashFoundation/zebra/issues/9557)

Depends on:

- https://github.com/ZcashFoundation/zebra/pull/9710
- https://github.com/zcash/wallet/pull/177

## Solution

- Simplified the wallet QA tests to use a single Zebra node and a single wallet instance. (A simple multi node test can be seen in `addnode.py`)
- Updated the test setup to reflect recent changes in Zallet, which continues to evolve.
- Added a wallet test that:
  - Creates an account and obtains an address.
  - Stops the services and restarts Zebra using the newly obtained address as the miner address.
  - Mines blocks and verifies that balances are available both in the node and in the wallet.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/parameters/network.rs
+++ b/zebra-chain/src/parameters/network.rs
@@ -143,6 +143,11 @@ impl std::fmt::Debug for Network {
             Self::Testnet(params) if params.is_regtest() => f
                 .debug_struct("Regtest")
                 .field("activation_heights", params.activation_heights())
+                .field("pre_nu6_funding_streams", params.pre_nu6_funding_streams())
+                .field(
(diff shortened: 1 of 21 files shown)


oxarbitrage:
The code here was moved into https://github.com/ZcashFoundation/zebra/pull/9710

PR #9224, opened 2025-02-11
fix(config): Correctly serialize testnet parameters into toml config format

arya2:
## Motivation

This PR correctly serializes testnet parameters into the toml config format.

Closes #8920.

## Solution

Converts `Config` to `DConfig` before serializing.

### Tests

Adds a roundtrip serialization test.

### Follow-up Work

Add an acceptance test for https://github.com/ZcashFoundation/zebra/pull/9201.

### PR Author's Checklist

- [ ] The PR name will make sense to users.
- [ ] The PR provides a CHANGELOG summary.
- [ ] The solution is tested.
- [ ] The documentation is up to date.
- [ ] The PR has a priority label.

### PR Reviewer's Checklist

- [ ] The PR Author's checklist is complete.
- [ ] The PR resolves the issue.

PR #9725, opened 2025-07-23
serialise zebradconfig

nachog00:
## Motivation

At zaino we need an ergonomic way of setting different configs for our test with zebras involved.
So far we have been using templated toml strings, but being able to build them from proper types would allow us several folds of greater flexibility and assurance that we are keeping up with upstream changes.

Also, for configuration reference, [the docs](https://zebra.zfnd.org/user/run.html#running-zebra) point to [`ZebradConfig`'s doc page](https://docs.rs/zebrad/latest/zebrad/config/struct.ZebradConfig.html), which makes no mention of the intermediate type `zebra_network::config::DConfig`, which ultimately is the struct that zebra deserialises the toml from.

## Solution

Making the `ZebradConfig` serialisable and exposing the intermediate types to allow contruction of a config though code and subsequent serialisation.
For this I refactored the deserialise impl for `zebra_network::config::Config` into a `impl TryFrom<DConfig> for Config` and a deserialise impl. That way a full `ZebradConfig` can be constructed in code just like it would on a toml (see the test i added to see what i mean).

For the doc part, maybe updating the docs to mention the intermediate type for the network config? (I didn't add that on this PR)

### Tests

I wrote a simple test to showcase the usage and check that it's indeed serialising all the way. Wouldn't say it properly tests the changes yet.

### Follow-up Work

If the zebra team agrees that they want to support this, all those intermediate types now made public would need documenting.
I think there's also potential for a `ZebradConfigBuilder` pattern support, which would be massively useful not only for zaino but for anybody else managing tests with zebras or even deploying them programatically.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebrad/src/config.rs
+++ b/zebrad/src/config.rs
@@ -5,6 +5,8 @@
 //! for specifying it.
 
 use serde::{Deserialize, Serialize};
+use zebra_chain::parameters::testnet::ConfiguredActivationHeights;
+use zebra_network::config::{DConfig, DTestnetParameters};
 
 /// Configuration for `zebrad`.
 ///
@@ -53,4 +55,30 @@ pub struct ZebradConfig {
 
     /// Mining configuration
     pub mining: zebra_rpc::config::mining::Config,
+}
+
+#[test]
+fn serialising_config() {
+    let d_config = DConfig {
+        listen_addr: "127.0.0.1:3000".to_string(),
+        testnet_parameters: Some(DTestnetParameters {
+            network_name: Some("name".to_owned()),
+            disable_pow: Some(true),
+            activation_heights: Some(ConfiguredActivationHeights {
+                overwinter: Some(1),
+                before_overwinter: Some(2),
+                ..Default::default()
+            }),
+            ..Default::default()
+        }),
+        ..Default::default()
+    };
+
+    let config: ZebradConfig = ZebradConfig {
+        network: zebra_network::Config::try_from(d_config).unwrap(),
+        ..Default::default()
+    };
+
+    let serialised = toml::to_string_pretty(&config).unwrap();
+    println!("{}", serialised)
 }
--- a/zebrad/src/config.rs
+++ b/zebrad/src/config.rs
@@ -5,9 +5,6 @@
 //! for specifying it.
 
 use serde::{Deserialize, Serialize};
-use zebra_chain::parameters::testnet::ConfiguredActivationHeights;
-use zebra_network::config::{DConfig, DTestnetParameters};
-
 /// Configuration for `zebrad`.
 ///
 /// The `zebrad` config is a TOML-encoded version of this structure. The meaning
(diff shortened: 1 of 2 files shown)


zebrad/src/config.rs
```diff
@@ -54,3 +53,22 @@ pub struct ZebradConfig {
...
+        },
+        network: zebra_network::Config::try_from(zebra_network::config::DConfig {
+            listen_addr: "127.0.0.1:3000".into(),
+            ..Default::default()
+        })
+        .unwrap(),
+        ..Default::default()
+    };
+    let toml_string = toml::to_string(&config).unwrap();
+    println!("{}", toml_string)
```
arya2, line 73: This use case should already be supported after https://github.com/ZcashFoundation/zebra/pull/9224 merged. Is there a more specific use case that needs direct access to `DConfig`?

It may be useful to move [the code Zebra uses to manage test zebrad instances](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/tests/common/launch.rs) so that Zaino could use it too.

Related: I noticed that in Zaino's toml templates, it's using Regtest and setting several testnet parameters that are unused on Regtest. Currently the only testnet parameter that's configurable on Regtest is activation heights, though there's a PR open to add [funding streams](https://github.com/ZcashFoundation/zebra/pull/9710/files#diff-358e67c6fee5ff1e700961035b7e91b3a74070fb80c790edbe185d5ac0aae8adR36-R42) in there too.

nachog00: Oh I missed that PR, nice to know there's already support for the serialisation.

Regarding your questions:
The end goal for us with this is to be able to contruct a zaino config .toml, serialising from your type.
My intention for making `DConfig` public is to be able to construct the config the same way the toml would, and IIUC that is the type that actually maps to the toml, as the base `ZebradConfig` doesn't even have the `testnet_parameters` field. I noticed you perform a bunch of logic to go from that into the plain `Config`, so I figured we probably didn't want to skip it by constructing the config directly with `Config`.

In fact, when trying to run the `intended_usage` test on current main (0893e1a), I get errors:
- `DConfig` is private
- `Config` doesn't implement `from<DConfig>`

In any case, would you recommend a currently supported way of constructing the config? for regtest network configs, would it be good enough to construct it with `zebra_network::config::Config`? (I notice that on the `testnet_params_serialization_roundtrip` test you are using the `testnet::Parameters` interface, so that might be it ... )

arya2: > I figured we probably didn't want to skip it by constructing the config directly with `Config`

Zebra will still do those checks when it's deserializing the config file, the most complicated ones are for the Testnet parameters, which will be checked by the `testnet::ParametersBuilder`.

> In any case, would you recommend a currently supported way of constructing the config? for regtest network configs, would it be good enough to construct it with zebra_network::config::Config?

Yep, `DConfig` wasn't meant to be constructed directly in code, only through deserialization. Most acceptance tests in Zebra construct `Config` structs to be serialized into `zebrad.toml` files, it works well in my experience and I would recommend it.

> I notice that on the `testnet_params_serialization_roundtrip` test you are using the `testnet::Parameters` interface, so that might be it

Yep! `testnet::ParametersBuilder` should support this use case well. There's certainly room for improvement, and feedback on how to make it better would be appreciated if there's any issue with using it this way.

nachog00: Nice, thanks for your insight!

One last doubt i have is, is there some documentation piece I'm missing that explains this kind of usage?
If not, I would gladly contribute to draft it as I'm going though the process of using it.

My main confusion came from the fact that something like:

```toml
[network.testnet_parameters]
disable_pow = true

[network.testnet_parameters.activation_heights]
NU5 = {nu5_activation_height}
NU6 = {nu6_activation_height}
```
seems to be a valid config toml and the docs only point to `ZebradConfig` which doesn't include any of those fields.

arya2: > One last doubt i have is, is there some documentation piece I'm missing that explains this kind of usage?

No, there should really be some information about it in the book.

> If not, I would gladly contribute to draft it as I'm going though the process of using it.

That would be greatly appreciated.

> seems to be a valid config toml and the docs only point to ZebradConfig which doesn't include any of those fields.

I see your point. Do you think it would help to update it to be `[network.network.activation_heights]` and add a `kind` field to `DTestnetParameters` to specify Regtest or Testnet? Or use `network_name == "Regtest"` as a special network name that indicates Zebra should use the regtest defaults? 

i.e.

```rust

struct DConfig {
    network: DNetwork,
    ..
}

#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum DNetwork {
    DefaultForKind(NetworkKind),
    TestnetWithParams(DTestnetParameters),
}

impl Default for DNetwork {
    fn default() -> Self {
        DNetwork::DefaultForKind(NetworkKind::Mainnet)
    }
}

```

arya2: @nachog00 I opened https://github.com/ZcashFoundation/zebra/issues/9729 to make this more intuitive.

nachog00:
Sure, thanks! Sorry I staled on this a bit. I'll comment over there.

Issue #9691, opened 2025-07-08
devops: PRs get stuck in CI


conradoplg:
### Describe the issue or request

Contributors PRs like https://github.com/ZcashFoundation/zebra/pull/9652 get stuck with some jobs in "Expected - waiting for status to be reported" state. The stuck jobs are the "Unit tests / Test Zebra Docker configurations / ..." jobs.

conradoplg:
Sometimes this also happens to regular PRs: https://github.com/ZcashFoundation/zebra/pull/9738

PR #9734, opened 2025-07-29
fix: `validateaddress` RPC bug

dorianvp:
## Motivation

This PR fixes a bug in the `validateaddress` validation logic. When receiving a regtest address it fails because it is parsing it as a testnet address, even though both regtest and testnet share the same encodings.

## Solution

If zebrad is not running in mainnet, we simply compare that the address is also not mainnet.

### Tests

Added `rpc_validateaddress_regtest` for testing this scenario.

### Follow-up Work

We also had to fix it in Zaino in this PR: https://github.com/zingolabs/zaino/pull/369.

Once it lands, we'll reuse the function exported from `zebra_rpc`.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods/tests/vectors.rs
+++ b/zebra-rpc/src/methods/tests/vectors.rs
@@ -17,7 +17,11 @@ use zebra_chain::{
     chain_sync_status::MockSyncStatus,
     chain_tip::{mock::MockChainTip, NoChainTip},
     history_tree::HistoryTree,
-    parameters::{testnet, Network::*, NetworkKind},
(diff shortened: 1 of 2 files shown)


oxarbitrage (commented):
Hey, thanks for the fix. It looks good to me. I  am not totally sure but we might need similar logic and test for `z_validateaddress`.

conradoplg (approved):

conradoplg:
Admin-merging due to https://github.com/ZcashFoundation/zebra/issues/9691

PR #9737, opened 2025-07-30
change(consensus): Switch Sapling verifier

### LINKED ISSUE #9731: Switch to `sapling-crypto`'s `BatchValidator` in tx verifier
PR #9676 loads the spend and output params by exposing the `bellman` types in `sapling-crypto` and then in `zcash_proofs`. This is not optimal, as it conflicts with the intended design of the upstream APIs. Instead of exposing the `bellman` params upstream, we can use the wrapper types from `sapling-crypto` in our tx verifier. This requires switching to [`BatchValidator`](https://docs.rs/sapling-crypto/0.5.0/sapling_crypto/struct.BatchValidator.html#method.validate) in Zebra's tx verifier for Sapling shielded data.

upbqdn:
## Motivation

Close #9731.

## Solution

- Use [`BatchValidator`](https://docs.rs/sapling-crypto/0.5.0/sapling_crypto/struct.BatchValidator.html#method.validate) in the tx verifier as a batched service.
- Based atop #9678.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/primitives/zcash_primitives.rs
+++ b/zebra-chain/src/primitives/zcash_primitives.rs
@@ -177,6 +177,14 @@ impl TryFrom<Amount<NonNegative>> for zcash_protocol::value::Zatoshis {
     }
 }
 
+impl TryFrom<Amount> for ZatBalance {
+    type Error = BalanceError;
+
+    fn try_from(amount: Amount) -> Result<Self, Self::Error> {
+        ZatBalance::from_i64(amount.into())
+    }
+}
+
 /// Convert a Zebra Script into a librustzcash one.
 impl From<&Script> for zcash_primitives::legacy::Script {
     fn from(script: &Script) -> Self {
--- a/zebra-chain/src/primitives/zcash_primitives.rs
+++ b/zebra-chain/src/primitives/zcash_primitives.rs
@@ -3,19 +3,12 @@
 
 use std::{io, ops::Deref, sync::Arc};
 
-use sapling_crypto::{
-    bundle::{Authorized, GrothProofBytes, OutputDescription, SpendDescription},
-    note::ExtractedNoteCommitment,
-    value::ValueCommitment,
-};
-use zcash_note_encryption::EphemeralKeyBytes;
 use zcash_primitives::transaction::{self as zp_tx, TxDigests};
 use zcash_protocol::value::{BalanceError, ZatBalance, Zatoshis};
 
 use crate::{
     amount::{Amount, NonNegative},
     parameters::NetworkUpgrade,
-    sapling::{self, AnchorVariant, PerSpendAnchor},
     serialization::ZcashSerialize,
     transaction::{AuthDigest, HashType, SigHash, Transaction},
     transparent::{self, Script},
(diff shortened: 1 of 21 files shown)


upbqdn:
This PR targets #9678. It'll be best to review these PRs together. We'll need to merge this PR to #9678 before we can merge #9678 to main.

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -177,6 +184,92 @@ impl TryFrom<Amount<NonNegative>> for zcash_protocol::value::Zatoshis {
...
+            "failed to convert sapling::ShieldedData to sapling_crypto::Bundle".into(),
+        ))
+    }
+}
+
+impl TryFrom<sapling::Spend<PerSpendAnchor>> for SpendDescription<Authorized> {
+    type Error = Error;
+
+    fn try_from(spend: sapling::Spend<PerSpendAnchor>) -> Result<Self, Self::Error> {
+        let cv = ValueCommitment::from_bytes_not_small_order(&spend.cv().into())
```
conradoplg: Unfortunately this and other `from_bytes()` conversions are CPU intensive. This PR might cause a big impact in Zebra performance.

Possible solutions:

- Add methods to `sapling_crypto` to construct these types from existing points. That might be unlikely to get merged, though they may accept it if it's feature-gated
- Change our types to use the  `sapling_crypto` types. Probably the best option in the long run, but it might be a lot of work

conradoplg (changes requested):

zebra-chain/src/primitives/zcash_primitives.rs
upbqdn: I wonder if we should just wrap the lrz tx type at this point.

upbqdn: Thanks for catching this, BTW.

arya2: We already do this conversion to compute transaction sighashes, we could re-use that for now. I'm updating https://github.com/ZcashFoundation/zebra/pull/9308 to do that for Orchard.

See https://github.com/ZcashFoundation/zebra/pull/9308/commits/40e54be3323d4f16f56a8c1ffecf4bfae3fe4c2e#diff-c21dd5cf744127963537a03717b04e474f889571502c622a7bef47705f7878c0R980

conradoplg: @arya2 had a great idea: we already convert txs to the librustzcash format when computing the sighash. We could extract the `sapling_crypto` types from there (or do some refactoring that would do the equivalent thing)

conradoplg: (ah, I didn't notice that Arya had already commented that :rofl: )

> I wonder if we should just wrap the lrz tx type at this point.

That's likely the long-term solution but that would require even more work. It seems we can achieve something similar by slowly replacing the "leaf" types with LRZ types

upbqdn: Done in f6ded6e6e936c39f259e4d04e1813a3db7a0f6c3 and 10ecb51edbf2e15029428b92a10149a9e283c306 (the last two commits).

upbqdn (commented):
Rebased and addressed review remarks.

conradoplg (commented):
This looks great, thanks! We just need to wait for #9678 to merge and then rebase this

conradoplg (approved):
(I rebased it)

PR #9738, opened 2025-07-30
chore: Update `deny.toml`

upbqdn:


upbqdn:
I didn't specify for which dependency we need to wait since all were transitive, and I found that information superfluous.

deny.toml
```diff
@@ -96,17 +93,14 @@ skip-tree = [
     # wait for abscissa_core to update synstructure
     { name = "synstructure", version = "=0.12.6" },
 
-    # wait until zcash_client_backend update rustix
-    { name = "rustix", version = "=0.38.44" },
+    { name = "windows-sys", version = "=0.52.0" },
+    { name = "windows-sys", version = "=0.59.0" },
```
oxarbitrage:
```suggestion
    { name = "windows-core", version = "=0.52.0" },
    { name = "windows-sys", version = "=0.59.0" },
```

I think the `=0.52.0` is for `windows-core` ?

upbqdn: It doesn't look like it is: https://github.com/ZcashFoundation/zebra/actions/runs/16632975232/job/47066947269?pr=9737#step:5:81.
Can you post a reference?

oxarbitrage: Sorry, what i mean is that you have 2 references to the same crate:

https://github.com/ZcashFoundation/zebra/blob/a82b17e90f16bce607446ff84d3d3787c59300db/deny.toml#L96-L97

I think is not needed, you can just go with the lower version one. 

In that case the suggestion may be just remove one of them.

upbqdn: Done. I didn't realize that. Thank you.

oxarbitrage (dismissed):

oxarbitrage:
@Mergifyio update

oxarbitrage (approved):
thanks @gustavovalverde

Issue #9736, opened 2025-07-30
Refactor funding streams into a BTreeMap


arya2:
## Motivation

We want to allow for disjoint funding streams height ranges to deploy NU6.1 on Testnet, where the post-NU6 funding streams have already ended.

It would also be nice to consolidate the current pre- and post-NU6 funding streams.

## Design

Use a BTreeMap (similar to how network upgrade activation heights are stored) where the keys are the start heights of the funding streams.

The end heights could be stored in the values of the BTreeMap or could be implicitly indicated by the height of the next item in the BtTreeMap (which would be empty for periods where there are no active funding streams).

PR #9740, opened 2025-07-31
change(consensus): Allow funding streams to be active across multiple disjoint height ranges

arya2:
## Motivation

We want to allow for disjoint funding streams height ranges to deploy NU6.1 on Testnet, where the post-NU6 funding streams have already ended.

A quick but suboptimal fix for the urgent motivation in #9736.

## Solution

- Updates `FundingStream` to include a list of height ranges rather than a single height range,
- Updates config deserialization to preserve its previous behaviour when one height range is defined for a configured funding stream, and when no height range is defined, while adding support for deserializing a list of height ranges per funding stream, and
- Updates `funding_stream_address_index()` to work with multiple height ranges per funding stream, where the funding stream address index for each height range of a funding stream starts at the next index after the last address index of the previous height range of that funding stream.

### Tests

This PR adds a stored config to test that the updated deserialization logic works as expected.

The rest will be tested manually and by existing tests in the follow-up PR implementing ZIP 271 for Testnet.

### Follow-up Work

- https://github.com/ZcashFoundation/zebra/issues/9736

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/parameters/network/testnet.rs
+++ b/zebra-chain/src/parameters/network/testnet.rs
@@ -90,7 +90,11 @@ pub struct ConfiguredLockboxDisbursement {
 
 /// Configurable funding streams for configured Testnets.
 #[derive(Serialize, Deserialize, Clone, Default, Debug)]
-#[serde(deny_unknown_fields, from = "DConfiguredFundingStreams")]
+#[serde(
+    deny_unknown_fields,
+    from = "DConfiguredFundingStreams",
+    into = "DConfiguredFundingStreams"
+)]
 pub struct ConfiguredFundingStreams {
     /// Start and end height for funding streams see [`FundingStreams::height_ranges`] for more details.
     pub height_ranges: Option<Vec<std::ops::Range<Height>>>,
@@ -99,10 +103,11 @@ pub struct ConfiguredFundingStreams {
 }
 
 /// An intermediate type used to deserialize [`ConfiguredFundingStreams`].
-#[derive(Clone, Debug, serde::Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct DConfiguredFundingStreams {
     /// Start and end height for funding streams see [`FundingStreams::height_ranges`] for more details.
-    height_ranges: Option<DHeightRanges>,
+    height_range: Option<DHeightRanges>,
     /// Funding stream recipients, see [`FundingStreams::recipients`] for more details.
     recipients: Option<Vec<ConfiguredFundingStreamRecipient>>,
 }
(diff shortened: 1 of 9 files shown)


zebra-chain/src/parameters/network/testnet.rs
```diff
@@ -90,18 +90,69 @@ pub struct ConfiguredLockboxDisbursement {
...
+impl From<ConfiguredFundingStreams> for DConfiguredFundingStreams {
+    fn from(value: ConfiguredFundingStreams) -> Self {
+        Self {
+            height_range: value.height_ranges.map(DHeightRanges::Ranges),
+            recipients: value.recipients,
+        }
+    }
+}
+
+/// An intermediate type used to deserialize [`std::ops::Range<Height>>`].
```
arya2:
```suggestion
/// An intermediate type used to deserialize [`std::ops::Range<Height>`].
```

zebrad/tests/common/configs/v2.5.0.toml
```diff
@@ -0,0 +1,220 @@
...
+NU5 = 1_842_420
+NU6 = 2_000_000
+"NU6.1" = 2_000_001
+NU7 = 2_000_002
+
+[[network.testnet_parameters.pre_nu6_funding_streams.height_range]]
+start = 0
+end = 100
+
+[[network.testnet_parameters.pre_nu6_funding_streams.height_range]]
```
arya2, line 82: It's a little strange that overlapping ranges are allowed but unused. It could cause some funding stream addresses to be required but unused in some cases, but that seems fine for now.

zebra-consensus/src/block/subsidy/funding_streams.rs
```diff
@@ -26,15 +26,28 @@ fn funding_stream_address_index(
...
+    // This logic will not re-use any addresses across height ranges.
+    let index = past_height_ranges
+        .iter()
+        .chain(std::iter::once(&(current_height_range.start..height)))
+        .fold(1u32, |acc, height_range| {
+            acc.checked_add(funding_stream_address_period(height_range.end, network))
+                .expect("no overflow should happen in this sum")
+                .checked_sub(funding_stream_address_period(height_range.start, network))
+                .expect("no overflow should happen in this sub")
+        }) as usize;
```
conradoplg, line 50: I'm not fully convinced this is correct. The number of addresses in a range is
`1 + period(end_height-1) - period(start)`
while this assumes it's
`period(end_height) - period(start)`
which I don't think is equivalent

conradoplg (commented):
I'm a bit wary of this approach, though I think it should work. 

I kept going back and forth on this while reviewing this PR, but I'm thinking that just doing https://github.com/ZcashFoundation/zebra/issues/9736 would be less error-prone, and might require less changes in existing consensus code. I'm willing to accept this PR pending double-checking the thing I commented on (and maybe adding a test just for that), but what do you think of taking another approach?

(For a first glance it might simply require changing `Network::funding_streams()` to lookup the correct FS from the BTreeMap. All the rest would stay mostly the same. But I'll try to double check my own suggestion :sweat_smile: )

zebra-consensus/src/block/subsidy/funding_streams.rs
arya2: The initial accumulator value in the fold is 1, and it adds the number of periods in each height range to the accumulator.

arya2:
> what do you think of taking another approach?

It would be my preference. The BTreeMap approach seems easier to reason about, more flexible, and fairly easy to implement.

oxarbitrage:
Replaced by https://github.com/ZcashFoundation/zebra/pull/9747

PR #9741, opened 2025-07-31
feat: Support ZIP-213

upbqdn:
## Motivation

- Close #5472.
- Close #5929.
- Close #9605.

## Solution

- Construct coinbase txs in accordance to [ZIP 213](https://zips.z.cash/zip-0213) using the [tx builder](https://docs.rs/zcash_primitives/latest/zcash_primitives/transaction/builder/struct.Builder.html) from `zcash_primitives`.
- The solution is built atop:
	- https://github.com/zcash/librustzcash/pull/1893, and
	- #9737.

### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The solution is tested.
- [x] The documentation is up to date.

PR #9744, opened 2025-08-01
add: Precompute coinbase tx and respond quickly on tip change

### LINKED ISSUE #9727: Respond quickly to long-polled `getblocktemplate` RPC on new chain tips
## Motivation

When a mining pool uses the `getblocktemplate` RPC in template mode with long polling, Zebra doesn't respond instantly on new chain tips. This has the consequence that a mining pool that mines to a shielded address will have its effective hashrate reduced by up to ~ 3% compared to pools that mine to taddrs. The reason is that whenever there's a new chain tip, such a pool will spend an extra 2.5 seconds on generating shielded outputs in the new block template, still mining on the old tip, while other pools will already be mining on the latest tip.

## Solution

Instantly return a pre-computed template for an empty block on new chain tips.

Credit: Str4d.

### Referencess

- `zcashd`'s behavior: https://github.com/zcash/zcash/blob/18238d90cd0b810f5b07d5aaa1338126aa128c06/src/rpc/mining.cpp#L634-L649
- Discord discussion: https://discord.com/channels/809218587167293450/876655911790321684/1395793068774395914

upbqdn:
## Motivation

- Close #9727.

## Solution

Precompute the coinbase tx for an empty block that will sit on the new tip. We will return this provisional block upon a chain tip change so that miners can mine on the newest tip, and don't waste their effort on a shorter chain while we compute a new template for a properly filled block. We do this precomputation before we start waiting for a new tip since computing the coinbase tx takes a few seconds if the miner mines to a shielded address, and we want to return fast when the tip changes.

Based atop #9741.

### PR Checklist

- [ ] The solution is tested.


zebra-rpc/src/methods.rs
```diff
@@ -2263,8 +2258,35 @@ where
...
+                // Precompute the coinbase tx for an empty block that will sit on the new tip. We
+                // will return this provisional block upon a chain tip change so that miners can
+                // mine on the newest tip, and don't waste their effort on a shorter chain while we
+                // compute a new template for a properly filled block. We do this precomputation
+                // before we start waiting for a new tip since computing the coinbase tx takes a few
+                // seconds if the miner mines to a shielded address, and we want to return fast
+                // when the tip changes.
+                let precompute_coinbase = |network, height, params| {
+                    tokio::task::spawn_blocking(move || {
+                        TransactionTemplate::new_coinbase(&network, height, &params, Amount::zero())
```
arya2, line 2273: Benchmarked at ~2.4 seconds for Orchard, ~100ms for Sapling/transparent, good call putting this in a blocking thread. Good idea in general.

zebra-rpc/src/methods.rs
```diff
@@ -2424,16 +2406,16 @@ where
 
         // - After this point, the template only depends on the previously fetched data.
 
-        let response = BlockTemplateResponse::new_internal(
+        Ok(BlockTemplateResponse::new_internal(
```
arya2, line 2409: Could we compute this in a blocking task too?

```rust

        let network = self.network.clone();
        let miner_params = miner_params.clone();
        tokio::task::spawn_blocking(move || {
            Ok(BlockTemplateResponse::new_internal(
                &network,
                None,
                &miner_params,
                &chain_info,
                server_long_poll_id,
                mempool_txs,
                submit_old,
            )
            .into())
        })
        .await
        .expect("block template construction task should not panic")

```

zebra-rpc/src/methods.rs
```diff
@@ -2308,59 +2324,34 @@ where
...
+                        chain_info.tip_height,
+                        chain_info.tip_hash,
+                        chain_info.max_time,
+                        vec![]
+                    )
+                    .generate_id();
+
+                    let submit_old = client_long_poll_id
+                        .as_ref()
+                        .map(|old_long_poll_id| server_long_poll_id.submit_old(old_long_poll_id));
```
arya2, line 2340: This should always be false if we want to simplify it and pass false in, but doing the check (as it is right now) may be more robust if the code changes.

arya2 (approved):
This looks great!

zebra-rpc/src/methods.rs
upbqdn: Yup, that exactly corresponds to what I measured.

arya2 (commented):
It would be nice to update the get block template test (`acceptance::common::get_block_template_rpcs::get_block_template::run()`) to check that it returns a block template within maybe half a second of the chain tip changing when configured to use a shielded miner address and called with a long poll id.

PR #9747, opened 2025-08-04
feat: support multiple arbitrary funding streams

### LINKED ISSUE #9736: Refactor funding streams into a BTreeMap

conradoplg:
## Motivation

This is an alternative to #9740 which is hopefully simpler in the sense of reducing changes to consensus code. It does have the downside of requiring a bit more change in the existing test code.

Closes #9736 but with a slightly different approach.

## Solution

Instead of having a set of funding stream constants (two at the moment, pre- and post-Nu6), have a vector of them.

I didn't use a BTreeMap as originally envisioned by #9736 because there wasn't the need to - the list is small enough to iterate when needed. Using a map would complicate things because it would requiring duplicate information (using the start height as the key) or removing the start height from the value which would change a lot of parsing code. Using a vector avoids all that.

To keep compatibility with existing configs, if `pre/post_nu6_funding_streams` are specified, they are simply added to the funding stream vector.

### Tests

The current tests should give good enough assurance.

The main change is that `Network::funding_streams(height)` now returns only the funding streams active at the given height, and not the pre-Nu6 or post-Nu6 streams depending on if the height is pre- or post-Nu6. Admittedly the previous behaviour was a bit weird.

To make sure this semantic change was safe, I looked at the callers:

- `funding_stream_values()` checks if the height is in the stream range, so the behaviour will be the same.
- `funding_stream_address_index()` - Only called by `funding_stream_address()`
- `funding_stream_address()` - Only seem to be called for streams returned by `funding_stream_values()`, so it should keep the same behaviour

### Follow-up Work

We need to add the NU6.1 funding streams to the vector.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -204,7 +204,10 @@ impl FundingStreamRecipient {
 }
 
 lazy_static! {
-    /// The funding streams for Mainnet as described in [protocol specification §7.10.1][7.10.1]
+    /// The funding streams for Mainnet as described in:
+    /// - [protocol specification §7.10.1][7.10.1]
+    /// - [ZIP-1015](https://zips.z.cash/zip-1015)
+    ///
     /// [7.10.1]: https://zips.z.cash/protocol/protocol.pdf#zip214fundingstreams
     pub static ref FUNDING_STREAMS_MAINNET: Vec<FundingStreams> = vec![
         FundingStreams {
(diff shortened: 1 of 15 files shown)


oxarbitrage (approved):
Looks good to me. Thanks @conradoplg

PR #9754, opened 2025-08-06
fix(crash): Fix Nu6.1 blocks crash

oxarbitrage:
## Motivation

During private custom testnet we found Zebra crashing in Nu6.1 blocks.

## Solution

Apply fixes from the used private testnet branch.

@arya2 can confirm if this is the fix.

### PR Checklist

- [ ] The solution is tested.


arya2 (approved):
Yep, this was the fix.

PR #9435, opened 2025-04-17
feat(ci): implement nextest and optimize Docker test builds

gustavovalverde:
## Motivation

Zebra's CI/CD pipeline had several inefficiencies that were impacting build and test execution times:

1. **Fragmented Test Execution**: Individual environment variables were used to control different test suites (`RUN_ALL_TESTS`, `STATE_FAKE_ACTIVATION_HEIGHTS`, `SYNC_LARGE_CHECKPOINTS_EMPTY`, etc.), sometimes in combination with feature gates, leading to inconsistent test configuration and complex workflow management.

2. **Feature Gate Inefficiencies**: Different feature combinations between build-time and runtime were causing Rust to rebuild artifacts unnecessarily, slowing down the build process.

3. **Complex Entrypoint Logic**: The `entrypoint.sh` script contained complex conditional logic for different test scenarios, making it harder to maintain and understand.

4. **Verbose CI Workflows**: Each test type required separate environment variables or feature gates configurations across multiple workflow files.

Fixes: #9331

## Solution

This PR implements a modernization of Zebra's test execution and Docker build system:

### 1. **Nextest Integration with Centralized Configuration**
- **Added `.config/nextest.toml`** with 17 specialized test profiles covering all test scenarios:
  - `all-tests`: Runs all tests except dependency checks
  - Individual profiles for each test type: `sync-full-mainnet`, `lwd-grpc-wallet`, `rpc-submit-block`, etc.
  - Proper timeout configurations and success output settings per test type
- **Replaced most environment variable-based test control** with a single `NEXTEST_PROFILE` variable
- **Centralized test filtering and scoping** logic from scattered shell scripts into declarative configuration

### 2. **Docker Build Optimization**
- **Streamlined feature sets**: Uses minimal features (`default-release-binaries proptest-impl lightwalletd-grpc-tests zebra-checkpoints`) for testing to prevent unnecessary rebuilds
- **Simplified test stage**: Reduced Docker commands and improved layer caching
- **Enhanced cargo nextest integration**: Added automatic nextest binary installation and optimized build process
- **Updated `.dockerignore`** to include `.config` directory for nextest configuration

### 3. **CI/CD Workflow Modernization**
- **Updated 2 major workflow files** (`sub-ci-unit-tests-docker.yml`, `sub-ci-integration-tests-gcp.yml`) with 15+ test jobs converted to use nextest profiles
- **Simplified environment variable management**: Replaced dozens of individual test flags with unified `NEXTEST_PROFILE` usage
[...]

PR #9768, opened 2025-08-07
feat(config)!: migrate zebrad to use a layered configuration

gustavovalverde:
### 

## Motivation

- Replace Abscissa-based config loading and the Docker TOML generation flow with a standard layered configuration approach.
- Let users configure via environment variables (`ZEBRA_SECTION__KEY`) without an entrypoint-generated `zebrad.toml`.
- Remove legacy `ZEBRA_*` test/internal env vars to avoid collisions with the new `config-rs` prefix.

## Solution

- Config loading
  - Implement `ZebradConfig::load()` using config-rs (defaults → optional TOML → env).
  - Derive/ensure defaults and `deny_unknown_fields` in config structs.

- Docker entrypoint
  - Remove config generation
  - Keep test execution path for `cargo nextest`.

- Environment variables (breaking)
  - Stop honoring legacy `ZEBRA_*` test/path vars. Use:
    - `SKIP_NETWORK_TESTS`, `SKIP_IPV6_TESTS`
    - `TEST_LIGHTWALLETD`, `TEST_LARGE_CHECKPOINTS`, `TEST_SYNC_TO_CHECKPOINT`, `TEST_SYNC_PAST_CHECKPOINT`
    - `STATE_CACHE_DIR` instead of `ZEBRA_CACHE_DIR`
    - `FORCE_USE_COLOR` instead of `ZEBRA_FORCE_USE_COLOR`
  - Logging: stop using `RUST_LOG`/`ZEBRA_RUST_LOG`; use `ZEBRA_TRACING__FILTER`.

- Tests and helpers
  - Provide a default network-matched `mining.miner_address` in the test config helper, selected by `ZEBRA_NETWORK__NETWORK` or the test’s requested network; still overridable by `ZEBRA_MINING__MINER_ADDRESS` or TOML.
  - Add `zebrad/tests/common/configs/v2.4.2.toml` and normalize `cache_dir`, `[rpc].cookie_dir = "cache_dir"`.

- Documentation
  - Update `generate.rs` docs to show `ZEBRA_SECTION__KEY` usage.

- Breaking changes
  - Legacy `ZEBRA_*` test/path environment variables are removed.
  - Docker entrypoint no longer generates/rewrites TOML; configuration is defaults + optional TOML + `ZEBRA_*` env.

### Tests

Our whole test suite must pass

### Follow-up Work

- Remove any remaining Abscissa `Configurable` trait usage.
- Update Zebra Book docs for the new configuration model.
- Update CI to use new env variable names.
- Validate z3 docker-compose with the new env-only configuration.

### PR Checklist

- [ ] The solution is tested.


zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
arya2, line 1: We can delete this file.

zebrad/src/application.rs
```diff
@@ -243,17 +243,27 @@ impl Application for ZebradApp {
...
+            Ok(config) => config,
+            // Ignore errors loading the config for some commands.
+            Err(_e) if command.cmd().should_ignore_load_config_error() => Default::default(),
+            Err(e) => {
+                // Check if this looks like a deprecated config format issue
+                let error_msg = e.to_string();
+                if error_msg.contains("invalid type: map, expected a string")
+                    || error_msg.contains("eviction_memory_time")
+                {
+                    status_err!("Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file.");
```
arya2: Why remove the rest of the log message here? 

Could you add a comment explaining why it logs a different message depending on whether the error message contains the text above?

Update: This has been fixed in https://github.com/ZcashFoundation/zebra/pull/9768/files/7796f788633f537e9294f3051b9be6ca5aa4700c..64aa261dcc2d45c26eff90fc5d47874cfe3e18e6#diff-8f1ee6cceb154001084d1b17ee893fea940caa97b4c2db44f129178845b6e2f8

zebrad/Cargo.toml
```diff
@@ -201,6 +201,9 @@ metrics = { workspace = true }
 dirs = { workspace = true }
 atty = { workspace = true }
 
+# Configuration management
+config = { version = "0.15.13", features = ["toml", "json"] }
```
arya2: Nitpick/Optional: Could we add this as a workspace dependency for consistency?

Update: Done in https://github.com/ZcashFoundation/zebra/pull/9768/commits/d967099ce34d7526bb4afa09a5d39ed6a7fd4bdd

zebrad/src/commands/copy_state.rs
```diff
@@ -86,11 +85,11 @@ impl CopyStateCmd {
         let base_config = APPLICATION.config();
         let source_config = base_config.state.clone();
 
-        // The default load_config impl doesn't actually modify the app config.
+        // Load the target config using the new config-rs method
```
arya2: The `copy-state` command is a little weird since it takes two configs, a source config and a target config.

It probably doesn't make sense to override the config fields in both of them based on the same environment variables. We could accept a prefix in `ZebradConfig::load()` to be expected in the environment variables overriding the target config? Like `TARGET_ZEBRA_STATE__CACHE_DIR` or `ZEBRA_TARGET_STATE__CACHE_DIR`?

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9768/commits/a14a50441df9b482818a8c705acb38f2089cc3d0#diff-21d953b03bc82b47d4ced1907ca6e8904561379e2e95f822eb35d9b8a7b6eb88.

<details>
<summary> outdated </summary>
Nitpick/Optional/Subjective:

```suggestion
        // Load the target config if a target config path was provided, or use an ephemeral config otherwise.
```

</details>

zebrad/src/commands/generate.rs
```diff
@@ -39,11 +39,22 @@ impl Runnable for GenerateCmd {
...
+#    - Examples:
+#      - ZEBRA_NETWORK__NETWORK=Testnet
+#      - ZEBRA_RPC__LISTEN_ADDR=127.0.0.1:8232
+#      - ZEBRA_STATE__CACHE_DIR=/path/to/cache
+#      - ZEBRA_TRACING__FILTER=debug
+#      - ZEBRA_METRICS__ENDPOINT_ADDR=0.0.0.0:9999
+#
+# 2. Configuration file (TOML format)
+#    - Specified via -c flag: `zebrad -c myconfig.toml start`
+#    - Or default location in preference directory (see below)
```
arya2: Nitpick/Subjective/Optional:

```suggestion
#    - At the path specified via -c flag, e.g. `zebrad -c myconfig.toml start`, or
#    - At the default path in the user's preference directory (platform-dependent, see below)
```

zebrad/src/config.rs
```diff
@@ -54,3 +71,80 @@ pub struct ZebradConfig {
...
     pub mining: zebra_rpc::config::mining::Config,
 }
+
+impl ZebradConfig {
+    /// Loads the configuration from the conventional sources.
+    ///
+    /// Configuration is loaded from three sources, in order of precedence:
+    /// 1. Hard-coded defaults (lowest precedence)
+    /// 2. TOML configuration file (if provided)
+    /// 3. Environment variables with `ZEBRA_` prefix (highest precedence)
```
arya2: It would be nice to have these in the same order as they are in the `generate` cmd output.

```suggestion
    /// 1. Environment variables with `ZEBRA_` prefix (highest precedence)
    /// 2. TOML configuration file (if provided)
    /// 3. Hard-coded defaults (lowest precedence)
```

zebrad/src/config.rs
```diff
@@ -4,8 +4,25 @@
...
+    // Centralized, case-insensitive suffix-based deny-list
+    lower.ends_with("password")
+        || lower.ends_with("secret")
+        || lower.ends_with("token")
+        // Block raw cookies only if a field is literally named "cookie".
+        // (Paths like cookie_dir are not affected.)
+        || lower.ends_with("cookie")
+        // Only raw private keys; paths like *_private_key_path are not affected.
+        || lower.ends_with("private_key")
+}
```
arya2: It would be nice to have these in a constant so they can be referenced below.

```suggestion
/// Centralized, case-insensitive suffix-based deny-list to ban setting config fields with
/// environment variables if those config field names end with any of these suffixes.
const DENY_CONFIG_KEY_SUFFIX_LIST: [&str; 5] = [
    "password",
    "secret",
    "token",
    // Block raw cookies only if a field is literally named "cookie".
    // (Paths like cookie_dir are not affected.)
    "cookie",
    // Only raw private keys; paths like *_private_key_path are not affected.
    "private_key",
];

/// Returns true if a leaf key name should be considered sensitive and blocked
/// from environment variable overrides.
fn is_sensitive_leaf_key(leaf_key: &str) -> bool {
    let key = leaf_key.to_ascii_lowercase();
    DENY_CONFIG_KEY_SUFFIX_LIST.iter().any(|deny_suffix| key.ends_with(deny_suffix))
}
```

zebrad/src/config.rs
```diff
@@ -54,3 +71,80 @@ pub struct ZebradConfig {
...
+    ///
+    /// Environment variables use the format `ZEBRA_SECTION__KEY` where:
+    /// - `SECTION` is the configuration section (e.g., `network`, `rpc`)
+    /// - `KEY` is the configuration key within that section
+    /// - Double underscores (`__`) separate nested keys
+    ///
+    /// # Security
+    /// Environment variables whose leaf key names end with sensitive suffixes (case-insensitive)
+    /// will cause configuration loading to fail with an error: `password`, `secret`, `token`, `cookie`, `private_key`.
+    /// This prevents both silent misconfigurations and process table exposure of sensitive values.
```
arya2:
```suggestion
    /// # Security
    ///
    /// Environment variables whose leaf key names end with sensitive suffixes (case-insensitive)
    /// will cause configuration loading to fail with an error: `password`, `secret`, `token`, `cookie`, `private_key`.
    /// This prevents both silent misconfigurations and process table exposure of sensitive values.
    ///
    /// See [`DENY_CONFIG_KEY_SUFFIX_LIST`] and [`is_sensitive_leaf_key()`] above
```

zebrad/src/config.rs
```diff
@@ -54,3 +71,80 @@ pub struct ZebradConfig {
...
+    ///
+    /// # Examples
+    /// - `ZEBRA_NETWORK__NETWORK=Testnet` sets `network.network = "Testnet"`
+    /// - `ZEBRA_RPC__LISTEN_ADDR=127.0.0.1:8232` sets `rpc.listen_addr = "127.0.0.1:8232"`
+    pub fn load(config_path: Option<PathBuf>) -> Result<Self, config::ConfigError> {
+        let mut builder = config::Config::builder();
+
+        // 1. Start with defaults - but don't use try_from with the struct directly
```
arya2: It would be nice to specify the struct as `config::Config` in this comment (I initially thought it meant `ZebradConfig`, but that doesn't have a `TryFrom` impl).

It may be clearer to say here that it starts with an empty map and will use the defaults from the `ZebradConfig::default()` impl wherever keys are missing.

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9768/commits/baca56046524272d93e4e8c2a19bd0357e72ca97

zebrad/src/config.rs
```diff
@@ -4,8 +4,25 @@
 //! application's configuration file and/or command-line options
 //! for specifying it.
 
+use std::path::PathBuf;
```
arya2:
```suggestion
use std::{collections::HashMap, path::PathBuf};
```

zebrad/src/config.rs
```diff
@@ -54,3 +71,80 @@ pub struct ZebradConfig {
...
+        if let Some(path) = config_path {
+            builder = builder.add_source(config::File::from(path).required(true));
+        }
+
+        // 3. Load from standard ZEBRA_ environment variables with a sensitive-leaf deny-list
+        // Use ZEBRA_ prefix and __ as separator for nested keys
+        // We filter the raw environment first, then let config-rs parse types via try_parsing(true).
+        // Sensitive environment variables cause configuration loading to fail for security.
+        let mut filtered_env: std::collections::HashMap<String, String> =
+            std::collections::HashMap::new();
```
arya2:
```suggestion
        let mut filtered_env: HashMap<String, String> = HashMap::new();
```

CHANGELOG.md
```diff
@@ -7,6 +7,30 @@ and this project adheres to [Semantic Versioning](https://semver.org).
...
+  environment variables. To use a file, set `CONFIG_FILE_PATH` (the entrypoint forwards it via
+  `--config`). Otherwise, configure via `ZEBRA_*` variables. `ZEBRA_CONF_PATH` has been removed in
+  favor of `CONFIG_FILE_PATH`. Docker setups that used variables like `ZEBRA_RPC_PORT`,
+  `ZEBRA_COOKIE_DIR`, `NETWORK`, `ENABLE_COOKIE_AUTH`, or `MINER_ADDRESS` must switch to the
+  config-rs equivalents shown above ([#9768](https://github.com/ZcashFoundation/zebra/pull/9768)).
+
+### Changed
+
+- `zebrad` now loads configuration from defaults, an optional TOML file, and environment variables,
+  with precedence: Env > TOML > Defaults ([#9768](https://github.com/ZcashFoundation/zebra/pull/9768)).
```
arya2, line 30: Looks like it allows JSON config files now too.

zebrad/src/config.rs
```diff
@@ -54,3 +71,80 @@ pub struct ZebradConfig {
...
+                            "Environment variable '{}' contains sensitive key '{}' which cannot be overridden via environment variables. \
+                             Use the configuration file instead to prevent process table exposure.",
+                            key, leaf
+                        )));
+                    }
+                }
+            }
+
+            filtered_env.insert(key, value);
+        }
```
arya2: Optional/Subjective/Nitpick:

```suggestion
        for (key, value) in std::env::vars().filter(|(key, _)| !key.starts_with("ZEBRA_")) {
            let leaf = key.split("_").last().expect("key must start with 'ZEBRA_'");
            if is_sensitive_leaf_key(leaf) {
                let err_msg = format!(
                    "Environment variable '{key}' contains sensitive key '{leaf}' which \
                     cannot be overridden via environment variables. \
                     Use the configuration file instead to prevent process table exposure.",
                );
                return Err(config::ConfigError::Message(err_msg));
            }

            filtered_env.insert(key, value);
        }
```

.config/nextest.toml
```diff
@@ -7,8 +7,9 @@ status-level = "pass"
 
 # --- All Tests profile ---
 # Runs all tests except 'check_no_git_dependencies'
+# TODO: We must find a better way to remove tests that are ignored, but also require a state or are flaky
```
arya2: > Runs all tests except 'check_no_git_dependencies'

This comment needs an update.

Update: Updated in https://github.com/ZcashFoundation/zebra/pull/9768/commits/ca9112f231b7f972fc97a4abc7880a9f970135e1

> We must find a better way to remove tests that are ignored

Do you mean it runs tests that should be ignored like `check_no_git_dependencies`? Is that only the `all-tests` profile, and if so, could we use another profile?

.github/workflows/sub-build-docker-image.yml
```diff
@@ -141,7 +141,7 @@ jobs:
         uses: docker/setup-buildx-action@v3.10.0
         with:
           version: "lab:latest"
-          driver: cloud
+          driver: ${{ vars.DOCKER_BUILDER || 'docker' }}
```
arya2: Are we using another driver here?

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -82,7 +79,7 @@ jobs:
       app_name: zebrad
       test_id: sync-to-mandatory-checkpoint
       test_description: Test sync up to mandatory checkpoint
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-to-mandatory-checkpoint,ZEBRA_TEST_SYNC_TO_CHECKPOINT=1,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
```
arya2, line 85: Where is the cache directory path being passed in now? It looks like it'll always use the default cache dir path in `test_type.rs` such that stateful tests will no longer be skipped when it's undefined, is that intended?

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -106,7 +103,7 @@ jobs:
       app_name: zebrad
       test_id: sync-past-mandatory-checkpoint
       test_description: Test full validation sync from a cached state
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-past-mandatory-checkpoint-${{ (inputs.network || vars.ZCASH_NETWORK) == 'Mainnet' && 'mainnet' || 'testnet' }},ZEBRA_TEST_SYNC_PAST_CHECKPOINT=1,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-past-mandatory-checkpoint-${{ (inputs.network || vars.ZCASH_NETWORK) == 'Mainnet' && 'mainnet' || 'testnet' }},TEST_SYNC_PAST_CHECKPOINT=1"
```
arya2: Where is `NETWORK` used now? I see it being passed in as `CONTAINER_ENV`, is it meant to define `ZEBRA_NETWORK__NETWORK` somewhere?

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -295,7 +292,7 @@ jobs:
       app_name: lightwalletd
       test_id: lwd-sync-full
       test_description: Test lightwalletd full sync
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=lwd-sync-full,ZEBRA_TEST_LIGHTWALLETD=1,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra,LWD_CACHE_DIR=/home/zebra/.cache/lwd"
```
arya2, line 298: What's replaced `LWD_CACHE_DIR`? Do the tests that were looking for it now always run and use the cached state in the default path if it exists?

arya2 (commented):
Flushing partial review, looking good so far!

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: > stateful tests will no longer be skipped when it's undefined

This is correct, this is why I had to add [all these exceptions](https://github.com/ZcashFoundation/zebra/pull/9768/files#diff-3e429424057e0b2a83fafc6e8942f4c05b63f4d977548e51535f8d86be8f9613R12) in the `all-tests` profile in the `nextest.toml`

Maybe this is what I need to consider to improve the comment you refer to in https://github.com/ZcashFoundation/zebra/pull/9768/files#r2292552815


By the way, I wouldn't say it's  completely "intended", but this was the simplest workaround I could think of until we can find a better way to handle stateful tests without adding a feature gate specifically for those tests (as this requires rebuilding Zebra and had a big impact in our testing time)

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: Oh, this was an oversight on my side. Thank you for catching this. Addressed in [`1a1c5a0` (#9768)](https://github.com/ZcashFoundation/zebra/pull/9768/commits/1a1c5a0ef9045b542caaf14eb420bc2c98ca0120)

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: Yes, that's correct. But the "always" run depends on which nextest profile you're using. This is also related to this answer: https://github.com/ZcashFoundation/zebra/pull/9768/files#r2293155986

CHANGELOG.md
gustavovalverde: I did not add that just because that's just a "side-effect" of using the `config` crate. But we can add it if we'd like to inform the users about this possibility.

arya2: I'm happy either way. We can also remove the `json` feature if we don't want to allow it.

.github/workflows/sub-ci-integration-tests-gcp.yml
arya2: Okay, that seems fine and works. Let's document the decision/change somewhere (maybe the PR description) and we can change it back to checking a constant that either enables or skips the stateful tests if we find a motivation for it.

zebrad/src/config.rs
gustavovalverde: It seems like this would split on a single underscore (`_`). For a variable like `ZEBRA_RPC__LISTEN_ADDR`, `key.split("_")` would produce `["ZEBRA", "RPC", "", "LISTEN", "ADDR"]`, making the leaf key `"ADDR"` instead of `"LISTEN_ADDR"`

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: This was initially changed here: https://github.com/ZcashFoundation/zebra/pull/9435 

I added a comment here in the nextest.conf file for reference.

arya2:
[This `regtest_block_templates_are_valid_block_submissions()` test failure](https://github.com/ZcashFoundation/zebra/actions/runs/17211645695/job/48827424199?pr=9768#step:4:1161) is due to it using Mainnet instead of Regtest in the spawned test child. The environment variable is overriding the config values written to the test child's temp directory.

[This `downgrade_state_format()` test failure](https://github.com/ZcashFoundation/zebra/actions/runs/17173260709/job/48729546633#step:4:1629) looks like it's passing for Mainnet but failing on Testnet, likely the same issue as it's also writing the config to the test child's temp directory, and as the test is passing for Mainnet.

We probably don't want to define `ZEBRA_NETWORK__NETWORK` in the environment of the `test-all` job.

I think `NETWORK` was previously used in `entrypoint.sh` as part of the base config, so writing a config would override the value, but now it's in the environment and it overrides any value that's written to a config.

.config/nextest.toml
```diff
@@ -6,9 +6,22 @@ fail-fast = true
...
+#   auto-skipped when the cache dir is undefined. To keep this profile fast and
+#   deterministic, we exclude those tests here and offer dedicated profiles
+#   below for running them when needed.
+# TODO: We need a better test architecture to run all non-stateful
 [profile.all-tests]
-default-filter = 'not test(check_no_git_dependencies)'
+default-filter = "not test(check_no_git_dependencies) and not test(=fully_synced_rpc_z_getsubtreesbyindex_snapshot_test) and not test(=lwd_rpc_test) and not test(=lwd_rpc_send_tx) and not test(=lwd_grpc_wallet) and not test(=lwd_integration) and not test(=lwd_sync_full) and not test(=lwd_sync_update) and not test(=lightwalletd_test_suite) and not test(=rpc_get_block_template) and not test(=rpc_submit_block) and not test(=get_peer_info) and not test(~generate_checkpoints_) and not test(=trusted_chain_sync_handles_forks_correctly) and not test(=sync_update_mainnet) and not test(=trusted_chain_sync_handles_forks_correctly) and not test(=activate_mempool_mainnet) and not test(=downgrade_state_format) and not test(=invalidate_and_reconsider_block) and not test(=nu6_funding_streams_and_coinbase_balance)"
```
arya2: I think we should be able to get some of these working

```suggestion
default-filter = "not test(check_no_git_dependencies) and not test(=fully_synced_rpc_z_getsubtreesbyindex_snapshot_test) and not test(=lwd_rpc_test) and not test(=lwd_rpc_send_tx) and not test(=lwd_grpc_wallet) and not test(=lwd_integration) and not test(=lwd_sync_full) and not test(=lwd_sync_update) and not test(=lightwalletd_test_suite) and not test(=rpc_get_block_template) and not test(=rpc_submit_block) and not test(=get_peer_info) and not test(~generate_checkpoints_) and not test(=sync_update_mainnet) and not test(=activate_mempool_mainnet)"
```

.config/nextest.toml
```diff
@@ -103,3 +116,16 @@ default-filter = 'package(zebrad) and test(=lightwalletd_test_suite)'
...
+slow-timeout = { period = "30m", terminate-after = 2 }
+default-filter = 'package(zebrad) and test(=downgrade_state_format)'
+
+[profile.invalidate-and-reconsider-block]
+slow-timeout = { period = "30m", terminate-after = 2 }
+default-filter = 'package(zebrad) and test(=invalidate_and_reconsider_block)'
+
+[profile.nu6-funding-streams-and-coinbase-balance]
+slow-timeout = { period = "30m", terminate-after = 2 }
+default-filter = 'package(zebrad) and test(=nu6_funding_streams_and_coinbase_balance)'
```
arya2:
```suggestion
```

.github/workflows/cd-deploy-nodes-gcp.yml
```diff
@@ -328,7 +328,7 @@ jobs:
           --container-stdin \
           --container-tty \
           --container-image ${{ vars.GAR_BASE }}/zebrad@${{ needs.build.outputs.image_digest }} \
-          --container-env "NETWORK=${{ matrix.network }},LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }}" \
+          --container-env "ZEBRA_NETWORK__NETWORK=${{ matrix.network }},LOG_FILE=${{ vars.CD_LOG_FILE }},SENTRY_DSN=${{ vars.SENTRY_DSN }}" \
```
arya2, line 331: Where is the `ZEBRA_NETWORK__NETWORK` defined here used? It looks like it's only used when deploying the long-running node instances off main or for manual workflows, so I think this one is fine, but just double-checking.

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -82,7 +79,7 @@ jobs:
       app_name: zebrad
       test_id: sync-to-mandatory-checkpoint
       test_description: Test sync up to mandatory checkpoint
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-to-mandatory-checkpoint,ZEBRA_TEST_SYNC_TO_CHECKPOINT=1,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "ZEBRA_NETWORK__NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-to-mandatory-checkpoint,TEST_SYNC_TO_CHECKPOINT=1"
```
arya2: This is running the `sync_to_mandatory_checkpoint_{mainnet, testnet}()` tests, but both will be testing whichever network is defined here.

Is the `ZEBRA_NETWORK__NETWORK` env var defined here expected or being used by this job anywhere?

We could split it up into two jobs if it is expected somewhere.

```suggestion
      test_variables: "NEXTEST_PROFILE=sync-to-mandatory-checkpoint,TEST_SYNC_TO_CHECKPOINT=1"
```

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -106,7 +103,7 @@ jobs:
       app_name: zebrad
       test_id: sync-past-mandatory-checkpoint
       test_description: Test full validation sync from a cached state
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-past-mandatory-checkpoint-${{ (inputs.network || vars.ZCASH_NETWORK) == 'Mainnet' && 'mainnet' || 'testnet' }},ZEBRA_TEST_SYNC_PAST_CHECKPOINT=1,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "ZEBRA_NETWORK__NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-past-mandatory-checkpoint-${{ (inputs.network || vars.ZCASH_NETWORK) == 'Mainnet' && 'mainnet' || 'testnet' }},TEST_SYNC_PAST_CHECKPOINT=1"
```
arya2, line 106: This will only run either the Mainnet or Testnet sync-past-mandatory-checkpoint test depending on `vars.ZCASH_NETWORK`. What was the behaviour before? Are we running both anywhere? Do we want to run both here? If we want to run both here, could we get rid of the `ZEBRA_NETWORK__NETWORK` definition and unify the `sync-past-mandatory-checkpoint-{mainnet, testnet}` profiles?

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -165,7 +162,7 @@ jobs:
       app_name: zebrad
       test_id: sync-update-mainnet
       test_description: Test syncing to tip with a Zebra tip state
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-update-mainnet,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "ZEBRA_NETWORK__NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=sync-update-mainnet"
```
arya2, line 165: We could probably remove the `ZEBRA_NETWORK__NETWORK` definition here altogether, but, if we're keeping it, the `needs.sync-full-mainnet.result == 'success'` condition above implies this will always be Mainnet.

```suggestion
      test_variables: "ZEBRA_NETWORK__NETWORK=Mainnet,NEXTEST_PROFILE=sync-update-mainnet"
```

Update: Not a blocker.

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -422,7 +419,7 @@ jobs:
       app_name: zebrad
       test_id: rpc-get-block-template
       test_description: Test getblocktemplate RPC method via Zebra's rpc server
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=rpc-get-block-template,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "ZEBRA_NETWORK__NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=rpc-get-block-template"
```
arya2, line 422: The `needs.sync-full-mainnet.result == 'success'` condition implies this will only run on Mainnet, the test itself hard codes Mainnet but should work fine on Testnet too if it has an up-to-date Testnet cache.

```suggestion
      test_variables: "NEXTEST_PROFILE=rpc-get-block-template"
```

Update: Not a blocker.

.github/workflows/sub-ci-integration-tests-gcp.yml
```diff
@@ -445,7 +442,7 @@ jobs:
       app_name: zebrad
       test_id: rpc-submit-block
       test_description: Test submitting blocks via Zebra's rpc server
-      test_variables: "NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=rpc-submit-block,ZEBRA_CACHE_DIR=/home/zebra/.cache/zebra"
+      test_variables: "ZEBRA_NETWORK__NETWORK=${{ inputs.network || vars.ZCASH_NETWORK }},NEXTEST_PROFILE=rpc-submit-block"
```
arya2, line 445: This test was always running on Mainnet before, now it could run on either depending on which network is defined here. It should run fine on Testnet if this env var configures it to do so.

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
           docker run --tty \
           -e NEXTEST_PROFILE=all-tests \
-          -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e ZEBRA_NETWORK__NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
```
arya2: This is causing some failures in tests that cover both Mainnet and Testnet, or which use Regtest or a configured Testnet. If it's necessary to define the env var here, we could define another prefix for base config values that can be overridden by the config file, define a new env var or CLI option to express that the config file shouldn't be overridden, or find some other workaround, but removing it is an easy fix otherwise.

```suggestion
```

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -74,14 +74,12 @@ jobs:
...
-        env:
-          NETWORK: ${{ inputs.network || vars.ZCASH_NETWORK }}
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
           docker run --tty \
           -e NEXTEST_PROFILE=sync-large-checkpoints-empty \
-          -e ZEBRA_TEST_LARGE_CHECKPOINTS=1 \
-          -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e TEST_LARGE_CHECKPOINTS=1 \
+          -e ZEBRA_NETWORK__NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
```
arya2, line 81: This test was always running on Mainnet before, but should work fine on Testnet as well.

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -103,14 +101,12 @@ jobs:
...
-        env:
-          NETWORK: ${{ inputs.network || vars.ZCASH_NETWORK }}
         run: |
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
           docker run --tty \
           -e NEXTEST_PROFILE=lwd-integration \
-          -e ZEBRA_TEST_LIGHTWALLETD=1 \
-          -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e TEST_LIGHTWALLETD=1 \
+          -e ZEBRA_NETWORK__NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
```
arya2, line 108: This test was always running on Mainnet before, but may work fine on Testnet as well.

.github/workflows/sub-ci-unit-tests-docker.yml
```diff
@@ -163,7 +159,7 @@ jobs:
           docker pull ${{ vars.GAR_BASE }}/${{ vars.CI_IMAGE_NAME }}@${{ inputs.image_digest }}
           docker run --tty \
           -e NEXTEST_PROFILE=check-no-git-dependencies \
-          -e NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
+          -e ZEBRA_NETWORK__NETWORK="${{ inputs.network || vars.ZCASH_NETWORK }}" \
```
arya2, line 161: This test doesn't use the network anywhere, it's just reading the `Cargo.lock` file and checking that there are no git sources in there.

zebrad/src/config.rs
```diff
@@ -4,8 +4,32 @@
 //! application's configuration file and/or command-line options
 //! for specifying it.
 
+use std::{collections::HashMap, path::PathBuf};
+
 use serde::{Deserialize, Serialize};
 
+/// Centralized, case-insensitive suffix-based deny-list to ban setting config fields with
+/// environment variables if those config field names end with any of these suffixes.
+const DENY_CONFIG_KEY_SUFFIX_LIST: [&str; 5] = [
```
arya2, line 13: Should the `miner_address` be included here? Or the RPC listen addr? There's a DoS vulnerability if the RPC listen address is ever publicly exposed (even with a cookie, someone could repeatedly call the server and starve Zebra).

Update: Added to an issue.

.github/workflows/sub-test-zebra-config.yml
```diff
@@ -29,54 +35,54 @@ jobs:
...
-            grep_patterns: '-e "0.0.0.0:9999"'
-
-          # Mining configuration
-          - id: mining-config
-            name: Mining configuration
-            env_vars: '-e MINER_ADDRESS="u1cymdny2u2vllkx7t5jnelp0kde0dgnwu0jzmggzguxvxj6fe7gpuqehywejndlrjwgk9snr6g69azs8jfet78s9zy60uepx6tltk7ee57jlax49dezkhkgvjy2puuue6dvaevt53nah7t2cc2k4p0h0jxmlu9sx58m2xdm5f9sy2n89jdf8llflvtml2ll43e334avu2fwytuna404a"'
-            grep_patterns: '-e "miner_address = \\\"u1cymdny2u2vllkx7t5jnelp0kde0dgnwu0jzmggzguxvxj6fe7gpuqehywejndlrjwgk9snr6g69azs8jfet78s9zy60uepx6tltk7ee57jlax49dezkhkgvjy2puuue6dvaevt53nah7t2cc2k4p0h0jxmlu9sx58m2xdm5f9sy2n89jdf8llflvtml2ll43e334avu2fwytuna404a\\\""'
+            env_vars: -e FEATURES=prometheus -e ZEBRA_METRICS__ENDPOINT_ADDR=0.0.0.0:9999
+            grep_patterns: -e "0.0.0.0:9999"
+          # Todo: This test is flaky, and we need to fix it. But this is included in the Rust tests as a fallback.
```
arya2, line 80: It seems strange that this test would be flaky, it looks like the network would always be Mainnet, I would expect it to consistently pass. In any case, the config tests should fail if Zebra can't parse a valid miner address, so it doesn't seem necessary to check that here too.

Cargo.toml
```diff
@@ -132,10 +132,10 @@ tokio = "1.44"
 tokio-stream = "0.1.17"
 tokio-test = "0.4"
 tokio-util = "0.7.14"
-toml = "0.8"
+toml = "0.9.5"
 tonic = "0.14"
 tonic-prost = "0.14"
-tonic-prost-build = "0.14"
+tonic-prost-build = "0.14"
```
arya2: Nitpick/Optional:

```suggestion
tonic-prost-build = "0.14"
```

Cargo.toml
```diff
@@ -153,6 +153,7 @@ wagyu-zcash-parameters = "0.2"
 x25519-dalek = "2.0.1"
 zcash_note_encryption = "0.4.1"
 zcash_script = "0.3.2"
+config = { version = "0.15.14", features = ["toml", "json"] }
```
arya2: Was the `json` feature selected intentionally? As far as I can tell it just allows JSON config files, I see no harm in allowing that if we want to keep this, I'm just curious why it was added.

book/src/user/docker.md
```diff
@@ -80,28 +80,26 @@ All available Cargo features are listed at
 
 ## Configuring Zebra
 
-To configure Zebra using Docker, you have a few options, processed in this order:
+Zebra uses config-rs to layer configuration from defaults, an optional TOML file, and `ZEBRA_`-prefixed environment variables. When running with Docker, configure Zebra using any of the following (later items override earlier ones):
```
arya2:
```suggestion
Zebra uses [config-rs](https://crates.io/crates/config) to layer configuration from defaults, an optional TOML file, and `ZEBRA_`-prefixed environment variables. When running with Docker, configure Zebra using any of the following (later items override earlier ones):
```

book/src/user/docker.md
```diff
@@ -80,28 +80,26 @@ All available Cargo features are listed at
...
 
-You can see if your config works as intended by looking at Zebra's logs.
-
-Note that if you provide a configuration file using methods 1 or 2, environment variables from `docker/.env` will **not** override the settings within that file. The environment variables are primarily used for the auto-generation scenario (method 3).
+You can verify your configuration by inspecting Zebra's logs at startup.
 
 ### RPC
 
-Zebra's RPC server is disabled by default. To enable it, you need to define the RPC settings in Zebra's configuration. You can achieve this using one of the configuration methods described above:
+Zebra's RPC server is disabled by default. Enable and configure it via TOML or environment variables:
```
arya2: Nitpick/Optional:

```suggestion
Zebra's RPC server is disabled by default. Enable and configure it via the TOML configuration file, or configuration environment variables:
```

book/src/user/docker.md
```diff
@@ -137,11 +135,9 @@ docker compose -f docker/docker-compose.lwd.yml up
 ```
 
 Note that Docker will run Zebra with the RPC server enabled and the cookie
-authentication mechanism disabled since Lightwalletd doesn't support it. Instead
-of configuring Zebra via the recommended config file or `docker/.env` file, we
-configured the RPC server by setting environment variables directly in the
-`docker/docker-compose.lwd.yml` file. This takes advantage of the entrypoint
-script's auto-generation feature (method 3 described above).
+authentication mechanism disabled since Lightwalletd doesn't support it. In this
```
arya2: Nitpick: It would be nice to be more specific here.

```suggestion
authentication mechanism disabled when running `docker compose -f docker/docker-compose.lwd.yml up`, since Lightwalletd doesn't support cookie authentication. In this
```

docker/entrypoint.sh
```diff
@@ -2,24 +2,12 @@
...
 # Entrypoint for running Zebra in Docker.
 #
-# The main script logic is at the bottom.
-#
-# ## Notes
-#
-# - `$ZEBRA_CONF_PATH` can point to an existing Zebra config file, or if not set,
-#   the script will look for a default config at ${HOME}/.config/zebrad.toml,
-#   or generate one using environment variables.
+# This script handles privilege dropping and launches zebrad or tests.
```
arya2, line 5: Do we want to mention `CONFIG_FILE_PATH` here where it mentioned `ZEBRA_CONF_PATH` before?

docker/entrypoint.sh
```diff
@@ -31,86 +19,13 @@ exec_as_user() {
...
   echo "$1" >&2
   exit 1
 }
 
 # Creates a directory if it doesn't exist and sets ownership to specified UID:GID.
-# Also ensures the parent directories have the correct ownership.
-#
-# ## Parameters
-#
-# - $1: Directory path to create and own
```
arya2, line 113: Why remove these docs?

docker/entrypoint.sh
```diff
@@ -130,64 +45,30 @@ create_owned_directory() {
...
-  else
-    echo "ZEBRA_CONF_PATH was not set and no default config found at ${HOME}/.config/zebrad.toml"
-    echo "Preparing a default one..."
-    ZEBRA_CONF_PATH="${HOME}/.config/zebrad.toml"
-    create_owned_directory "$(dirname "${ZEBRA_CONF_PATH}")"
-    prepare_conf_file
-  fi
+# If the user provided a config file path we pass it to zebrad.
+CONFIG_ARGS=()
+if [[ -n ${CONFIG_FILE_PATH} && -f ${CONFIG_FILE_PATH} ]]; then
```
arya2, line 53: Do we want to still print a warning if a file can't be found at the provided path?

docker/entrypoint.sh
```diff
@@ -130,64 +45,30 @@ create_owned_directory() {
...
-echo "INFO: Using the following environment variables:"
-printenv
-
-echo "Using Zebra config at ${ZEBRA_CONF_PATH}:"
-cat "${ZEBRA_CONF_PATH}"
-
+# Main Script Logic
 # - If "$1" is "--", "-", or "zebrad", run `zebrad` with the remaining params.
-# - If "$1" is "test":
-#   - and "$2" is "zebrad", run `zebrad` with the remaining params,
```
arya2, line 177: Was this wrong before? I see that it was checking `"$1" == "zebrad"` before this change as well, why?

docker/entrypoint.sh
```diff
@@ -130,64 +45,30 @@ create_owned_directory() {
...
   shift
-  exec_as_user zebrad --config "${ZEBRA_CONF_PATH}" "$@"
+  exec_as_user zebrad "${CONFIG_ARGS[@]}" "$@"
   ;;
 test)
   shift
   if [[ "$1" == "zebrad" ]]; then
     shift
-    exec_as_user zebrad --config "${ZEBRA_CONF_PATH}" "$@"
+    exec_as_user zebrad "${CONFIG_ARGS[@]}" "$@"
```
arya2, line 71: Will it ever get here? It looks like the case statement above will always catch it before it does?

Update: Not a blocker.

zebrad/src/application.rs
```diff
@@ -243,17 +243,19 @@ impl Application for ZebradApp {
...
-                    return Err(e);
-                }
-            },
-            None => ZebradConfig::default(),
+        let config = match ZebradConfig::load(command.config_path()) {
+            Ok(config) => config,
+            // Ignore errors loading the config for some commands.
+            Err(_e) if command.cmd().should_ignore_load_config_error() => Default::default(),
+            Err(e) => {
+                status_err!("Zebra could not load its configuration: {}", e);
```
arya2: It would be nice to leave the extra guidance here.

```suggestion
                status_err!(
                    "Zebra could not load the provided configuration file and/or environment variables.\
                     This might mean you are using a deprecated format of the file, or are attempting to
                     configure deprecated or unknown fields via environment variables.\
                     You can generate a valid config by running \"zebrad generate\", \
                     and diff it against yours to examine any format inconsistencies."
                );
```

zebrad/src/config.rs
```diff
@@ -54,3 +78,93 @@ pub struct ZebradConfig {
...
+            if let Some(without_prefix) = key.strip_prefix(&required_prefix) {
+                // Check for sensitive keys on the stripped key.
+                let parts: Vec<&str> = without_prefix.split("__").collect();
+                if let Some(leaf) = parts.last() {
+                    if is_sensitive_leaf_key(leaf) {
+                        return Err(config::ConfigError::Message(format!(
+                            "Environment variable '{}' contains sensitive key '{}' which cannot be overridden via environment variables. \
+                             Use the configuration file instead to prevent process table exposure.",
+                            key, leaf
+                        )));
```
arya2, line 147: Why return an error here rather than logging a warning and ignoring the sensitive key?

zebrad/tests/acceptance.rs
```diff
@@ -30,25 +30,25 @@
 //!   will allow this test to run or give up. Value for the Mainnet full sync tests.
 //! - `SYNC_FULL_TESTNET_TIMEOUT_MINUTES` env variable: The total number of minutes we
 //!   will allow this test to run or give up. Value for the Testnet full sync tests.
-//! - `ZEBRA_CACHE_DIR` env variable: The path to a Zebra cached state directory.
-//!   If not set, it defaults to `/zebrad-cache`. For some sync tests, this directory needs to be
-//!   created in the file system with write permissions.
+//! - `ZEBRA_STATE__CACHE_DIR` env variable: The path to a Zebra cached state directory.
+//!   For some sync tests, this directory needs to be created in the file system
+//!   with write permissions.
```
arya2:
```suggestion
//! - A zebrad state cache directory is required for some tests, either at the default state cache
//!   directory path, or at the path defined in the `ZEBRA_STATE__CACHE_DIR` env variable.
//!   For some sync tests, this directory needs to be created in the file system
//!   with write permissions.
```

.config/nextest.toml
arya2, line 1: The 3 week timeout for `sync-full-mainnet` seems excessive, do we want to change that to 1 week? Wasn't changed in this PR if we want to leave it for now.

zebrad/tests/acceptance.rs
arya2, line 1: Adding `#[cfg(not(target_os = "windows"))]` above `fn last_config_is_stored()` should fix the lint.

zebrad/tests/acceptance.rs
```diff
@@ -903,7 +903,7 @@ fn invalid_generated_config() -> Result<()> {
     let output = child.wait_with_output()?;
 
     // Check that Zebra produced an informative message.
-    output.stderr_contains("Zebra could not parse the provided config file. This might mean you are using a deprecated format of the file.")?;
+    output.stderr_contains("Zebra could not load its configuration: invalid type: map, expected a string for key `mempool.eviction_memory_time`")?;
```
arya2: If you commit the suggestion to change this error message in `application.rs`, this will need an update too. Otherwise, this should still be updated to accept any error message that includes the first hard-coded part (so, excluding the part after the colon).

```suggestion
    output.stderr_contains("Zebra could not load the provided configuration file and/or environment variables")?;
```

zebrad/tests/acceptance.rs
```diff
@@ -1780,7 +1784,8 @@ fn lwd_integration() -> Result<()> {
 
 /// Make sure `zebrad` can sync from peers, but don't actually launch `lightwalletd`.
 ///
-/// This test only runs when the `ZEBRA_CACHE_DIR` env var is set.
+/// This test only runs when a persistent cached state is configured
```
arya2: This and similar updates below look wrong, because it'll use the default state cache dir path if none is configured.

Nitpick:

```suggestion
/// This test only runs when a persistent cached state directory path is configured
```

zebrad/tests/acceptance.rs
```diff
@@ -1791,8 +1796,8 @@ fn sync_update_mainnet() -> Result<()> {
 /// Make sure `lightwalletd` can sync from Zebra, in update sync mode.
 ///
 /// This test only runs when:
-/// - the `ZEBRA_TEST_LIGHTWALLETD`, `ZEBRA_CACHE_DIR`, and
-///   `LWD_CACHE_DIR` env vars are set, and
+/// - `TEST_LIGHTWALLETD` is set,
+/// - a persistent cached state is configured (e.g., via `ZEBRA_STATE__CACHE_DIR`), and
```
arya2: Nitpick:

```suggestion
/// - a persistent cached state directory path is configured (e.g., via `ZEBRA_STATE__CACHE_DIR`), and
```

zebrad/tests/acceptance.rs
```diff
@@ -1825,9 +1831,9 @@ fn lwd_sync_full() -> Result<()> {
 ///
 /// Runs the tests in this order:
 /// - launch lightwalletd with empty states,
-/// - if `ZEBRA_CACHE_DIR` is set:
+/// - if a cached Zebra state is configured:
```
arya2:
```suggestion
/// - if a cached Zebra state directory path is configured:
```

zebrad/tests/acceptance.rs
```diff
@@ -1825,9 +1831,9 @@ fn lwd_sync_full() -> Result<()> {
...
 ///   - run a full sync
-/// - if `ZEBRA_CACHE_DIR` and `LWD_CACHE_DIR` are set:
+/// - if a cached Zebra state is configured:
```
arya2:
```suggestion
/// - if a cached Zebra state directory path is configured:
```

zebrad/tests/acceptance.rs
```diff
@@ -1843,29 +1849,29 @@ async fn lightwalletd_test_suite() -> Result<()> {
         launches_lightwalletd: true,
     })?;
 
-    // Only runs when ZEBRA_CACHE_DIR is set.
+    // Only runs when a cached Zebra state is configured.
```
arya2:
```suggestion
    // Only runs when a cached Zebra state directory path is configured with an environment variable.
```

zebrad/tests/common/cached_state.rs
```diff
@@ -31,8 +31,8 @@ use crate::common::{
     test_type::TestType,
 };
 
-/// The environmental variable that holds the path to a directory containing a cached Zebra state.
-pub const ZEBRA_CACHE_DIR: &str = "ZEBRA_CACHE_DIR";
+// Deprecated: tests should read the cache dir from `ZEBRA_STATE__CACHE_DIR` if they need
+// an external cached state directory. The actual application reads this via config-rs.
```
arya2:
```suggestion
```

zebrad/tests/common/config.rs
```diff
@@ -53,33 +53,41 @@ pub fn default_test_config(net: &Network) -> Result<ZebradConfig> {
 
     let consensus = zebra_consensus::Config::default();
 
-    let force_use_color = !matches!(
-        env::var("ZEBRA_FORCE_USE_COLOR"),
-        Err(env::VarError::NotPresent)
-    );
+    let force_use_color = env::var("FORCE_USE_COLOR").is_ok();
```
arya2, line 56: I'm just now noticing this, it doesn't appear to be documented anywhere if it could be added to the changelog or somewhere in this PR while you're here (it wasn't documented anywhere before either).

zebrad/tests/common/config.rs
```diff
@@ -53,33 +53,41 @@ pub fn default_test_config(net: &Network) -> Result<ZebradConfig> {
...
+    // mining-related tests have a valid configuration. Environment variables
+    // like ZEBRA_MINING__MINER_ADDRESS will override this file value.
     let mut mining = zebra_rpc::config::mining::Config::default();
 
-    let miner_address = if network.network.is_mainnet() {
+    // Determine the effective network kind: prefer environment override if present.
+    let effective_kind: NetworkKind = match env::var("ZEBRA_NETWORK__NETWORK").ok().as_deref() {
+        Some("Mainnet") | Some("mainnet") | Some("Main") | Some("main") => NetworkKind::Mainnet,
+        Some("Regtest") | Some("regtest") => NetworkKind::Regtest,
+        Some("Testnet") | Some("testnet") => NetworkKind::Testnet,
```
arya2: Lowercase/partial values aren't parsed as a valid `NetworkKind` (we could change this if we want, but, it's just not the case currently, and Zebra will fail to load the configuration from the environment if they're provided).

```suggestion
        Some("Mainnet") => NetworkKind::Mainnet,
        Some("Regtest") => NetworkKind::Regtest,
        Some("Testnet") => NetworkKind::Testnet,
```

zebrad/tests/common/config.rs
arya2, line 1: The changes here may cause some unexpected behaviour, because, until now, tests could assume that the config passed to the test zebrad child would be the one that it will use, but with the changes in this PR, they could be overridden with environment variables.

`trusted_chain_sync_handles_forks_correctly()` passes a config struct directly to `TrustedChainSync::spawn()` and also passes it to `with_config()` when spawning a test zebrad child, and there may be other instances where this is relevant.

We should document this behaviour and update `ZebradConfig` to disallow instantiation without going through the layered construction that will be used when loading the config in `application.rs` (perhaps as a separate issue since this PR is already quite large?).

zebrad/tests/common/sync.rs
```diff
@@ -326,14 +326,14 @@ pub fn check_sync_logs_until(
     Ok(zebrad)
 }
 
-/// Returns the cache directory for Zebra's state.
+/// Returns the cache directory for Zebra's state, as configured via config-rs.
```
arya2:
```suggestion
/// Returns the cache directory for Zebra's state, as configured with [`ZebradConfig::load`] with config-rs.
```

zebrad/tests/common/sync.rs
```diff
@@ -326,14 +326,14 @@ pub fn check_sync_logs_until(
...
 ///
-/// It checks the `ZEBRA_CACHE_DIR` environment variable and returns its value if set.
-/// Otherwise, it defaults to `"/zebrad-cache"`.
+/// Uses the resolved configuration from `ZebradConfig::load(None)`, which
+/// incorporates defaults, optional TOML, and environment overrides.
```
arya2:
```suggestion
/// Uses the resolved configuration from [`ZebradConfig::load(None)`](ZebradConfig::load), which
/// incorporates defaults, optional TOML, and environment overrides.
```

zebrad/tests/common/test_type.rs
```diff
@@ -152,18 +183,29 @@ impl TestType {
     /// Returns the Zebra state path for this test, if set.
     #[allow(clippy::print_stderr)]
     pub fn zebrad_state_path<S: AsRef<str>>(&self, test_name: S) -> Option<PathBuf> {
-        match env::var_os(ZEBRA_CACHE_DIR) {
-            Some(path) => Some(path.into()),
-            None => {
-                let test_name = test_name.as_ref();
+        // Load the effective config (defaults + optional TOML + env overrides)
+        let test_name = test_name.as_ref();
+        let cfg = match ZebradConfig::load(None) {
```
arya2, line 188: We should either update the documentation for acceptance tests that required a cached state to say that they'll always run and look for a cached state in the default cache state directory path, or update this to check for the environment variable outside of `ZebradConfig::load()` so that it won't use the default value if no value is configured.

zebrad/tests/common/test_type.rs
```diff
@@ -241,23 +283,38 @@ impl TestType {
...
+
+        // Otherwise, try a deterministic platform default
+        let default_path = default_lwd_cache_dir();
 
+        if self.needs_lightwalletd_cached_state() {
+            // Only run if a cached state actually exists at the default path
+            if default_path.exists() {
+                Some(default_path)
+            } else {
+                tracing::info!(
```
arya2:
```suggestion
                tracing::warn!(
```

zebrad/tests/common/test_type.rs
```diff
@@ -241,23 +283,38 @@ impl TestType {
...
+                    "skipped {test_name:?} {self:?} lightwalletd test: no cached state found at default path",
+                );
+                None
+            }
+        } else if self.allow_lightwalletd_cached_state() {
+            // Use default if present; otherwise run without cached state
+            if default_path.exists() {
+                Some(default_path)
+            } else {
+                tracing::info!(
```
arya2:
```suggestion
                tracing::warn!(
```

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+impl EnvGuard {
+    /// Acquire the global lock and clear all ZEBRA_* env vars, saving originals.
+    fn new() -> Self {
+        // If a test panics, the mutex guard is dropped, but the mutex remains poisoned.
+        // We can recover from the poison error and get the lock, because we're going
+        // to overwrite the environment variables anyway.
+        let guard = TEST_MUTEX.lock().unwrap_or_else(|e| e.into_inner());
+
+        let original_vars: Vec<(String, String)> = env::vars()
+            .filter(|(key, _val)| key.starts_with("ZEBRA_"))
```
arya2: Could we move this prefix to a constant and use the constant here and in the `Drop` impl below?

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+    );
+}
+
+#[test]
+fn config_nonexistent_file_errors() {
+    let _env = EnvGuard::new();
+
+    let nonexistent_path = PathBuf::from("/this/path/does/not/exist.toml");
+    let result = ZebradConfig::load(Some(nonexistent_path));
+    assert!(result.is_err(), "Should fail to load nonexistent file");
```
arya2: Optional/Nitpick: We can use `.expect_err()` here and below where we don't want to match on a specific error value.

```suggestion
    let _env = EnvGuard::new();
    ZebradConfig::load(Some(PathBuf::from("/this/path/does/not/exist.toml")))
        .expect_err("should fail to load nonexistent file");
```

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+
+#[test]
+fn config_nested_env_vars() {
+    let env = EnvGuard::new();
+
+    env.set_var("ZEBRA_TRACING__FILTER", "debug");
+
+    let config = ZebradConfig::load(None).expect("load config with nested env vars");
+
+    assert_eq!(config.tracing.filter.as_ref().unwrap(), "debug");
```
arya2: Optional/Nitpick: It would be nice to assert that these are some values instead of calling `.unwrap()`.

```suggestion
    assert_eq!(config.tracing.filter.as_ref(), Some("debug"));
```

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+
+    let config = ZebradConfig::load(None).expect("load config with ZEBRA_TRACING__LOG_FILE");
+    assert_eq!(
+        config.tracing.log_file.as_ref().unwrap(),
+        &PathBuf::from("/test/zebra.log")
+    );
+}
+
+#[test]
+fn zebra_mining_miner_address_from_toml() {
```
arya2: Optional/Nitpick: It would be nice to prefix these test names with `config_` 

```suggestion
fn config_zebra_network_network_env() {
    let env = EnvGuard::new();

    env.set_var("ZEBRA_NETWORK__NETWORK", "Testnet");

    let config = ZebradConfig::load(None).expect("load config with ZEBRA_NETWORK__NETWORK");
    assert_eq!(config.network.network.to_string(), "Testnet");
}

#[test]
fn config_zebra_rpc_listen_addr_env() {
    let env = EnvGuard::new();

    env.set_var("ZEBRA_RPC__LISTEN_ADDR", "127.0.0.1:18232");

    let config = ZebradConfig::load(None).expect("load config with ZEBRA_RPC__LISTEN_ADDR");
    assert_eq!(
        config.rpc.listen_addr.unwrap().to_string(),
        "127.0.0.1:18232"
    );
}

#[test]
fn config_zebra_state_cache_dir_env() {
    let env = EnvGuard::new();

    env.set_var("ZEBRA_STATE__CACHE_DIR", "/test/cache");

    let config = ZebradConfig::load(None).expect("load config with ZEBRA_STATE__CACHE_DIR");
    assert_eq!(config.state.cache_dir, PathBuf::from("/test/cache"));
}

#[test]
fn config_zebra_metrics_endpoint_addr_env() {
    let env = EnvGuard::new();

    env.set_var("ZEBRA_METRICS__ENDPOINT_ADDR", "0.0.0.0:9999");

    let config = ZebradConfig::load(None).expect("load config with ZEBRA_METRICS__ENDPOINT_ADDR");
    assert_eq!(
        config.metrics.endpoint_addr.unwrap().to_string(),
        "0.0.0.0:9999"
    );
}

#[test]
fn config_zebra_tracing_log_file_env() {
    let env = EnvGuard::new();

    env.set_var("ZEBRA_TRACING__LOG_FILE", "/test/zebra.log");

    let config = ZebradConfig::load(None).expect("load config with ZEBRA_TRACING__LOG_FILE");
    assert_eq!(
        config.tracing.log_file.as_ref().unwrap(),
        &PathBuf::from("/test/zebra.log")
    );
}

#[test]
fn config_zebra_mining_miner_address_from_toml() {
```

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+    let _env = EnvGuard::new();
+
+    let miner_address = "u1cymdny2u2vllkx7t5jnelp0kde0dgnwu0jzmggzguxvxj6fe7gpuqehywejndlrjwgk9snr6g69azs8jfet78s9zy60uepx6tltk7ee57jlax49dezkhkgvjy2puuue6dvaevt53nah7t2cc2k4p0h0jxmlu9sx58m2xdm5f9sy2n89jdf8llflvtml2ll43e334avu2fwytuna404a";
+    let toml_string = format!(
+        "[network]\nnetwork = \"Testnet\"\n\n[mining]\nminer_address = \"{}\"",
+        miner_address
+    );
```
arya2: Optional/Nitpick: We could use a raw str here to make this more readable.

```suggestion
    let toml_string = format!(
        r#"[network]
        network = "Testnet"
        
        [mining]
        miner_address = "{miner_address}""#,
    );
```

zebrad/tests/config.rs
```diff
@@ -0,0 +1,342 @@
...
+    env.set_var("ZEBRA_STATE__ELASTICSEARCH_PASSWORD", "topsecret");
+
+    let result = ZebradConfig::load(None);
+    assert!(result.is_err(), "Sensitive env key should cause an error");
+
+    #[cfg(feature = "elasticsearch")]
+    {
+        let msg = result.unwrap_err().to_string();
+        assert!(msg.contains("sensitive key"), "error message: {}", msg);
+    }
```
arya2: This should work regardless of if the feature is selected.

```suggestion
    let msg = result.unwrap_err().to_string();
    assert!(msg.contains("sensitive key"), "error message: {}", msg);
```

arya2 (commented):
This is looking good, thank you for making all these changes.

.github/workflows/cd-deploy-nodes-gcp.yml
gustavovalverde: This is passed to the Zebra container we deploy through this job. So this is just to specifically set the network Zebra will use (and we commonly deploy the nodes using both Mainnet and Testnet)

.config/nextest.toml
gustavovalverde: I merged your PR, so this should be good now

.config/nextest.toml
gustavovalverde: I merged your PR, so this should be good now

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: I preferred to segregate the tests in the nextest profile [`bcfce06` (#9768)](https://github.com/ZcashFoundation/zebra/pull/9768/commits/bcfce06222f794233abe7aaf9ac51967b09f5b22) to be able to make manual runs of this workflow for Testnet. It was also a bad idea to run this together.

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: This is expected. We run Mainnet by default, and Testnet when we specifically set to run it. I changed `sync_to_mandatory_checkpoint_{mainnet, testnet}()` to have the same behavior (and thus be consistent).

.github/workflows/sub-ci-integration-tests-gcp.yml
gustavovalverde: Yes, this is intended.

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: Yes, this is intended.

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: Yes, this is intended.

.github/workflows/sub-ci-unit-tests-docker.yml
gustavovalverde: I'll consider it in the CI refactor.

Cargo.toml
gustavovalverde: We don't have an use-case, so I'll remove it [chore: remove json feature from config](https://github.com/ZcashFoundation/zebra/pull/9768/commits/1cb799bf3234220e369887e6c5a99d0e111b8379)

.config/nextest.toml
gustavovalverde: I'll change it in the CI refactor.

zebrad/tests/config.rs
gustavovalverde: Solved here, for multiple cases [`594874a` (#9768)](https://github.com/ZcashFoundation/zebra/pull/9768/commits/594874a50f00b4cf1f20e8dddeda95007b012ebe)

zebrad/tests/config.rs
gustavovalverde: Solved here [`594874a` (#9768)](https://github.com/ZcashFoundation/zebra/pull/9768/commits/594874a50f00b4cf1f20e8dddeda95007b012ebe)

.github/workflows/sub-test-zebra-config.yml
gustavovalverde: I'll leave it here to have it as a reference in my refactor. But yes, the test in the config should be enough in any case.

zebrad/src/config.rs
gustavovalverde: I changed it based on this comment in Zallet https://github.com/zcash/wallet/pull/194#discussion_r2288082278

zebrad/src/application.rs
```diff
@@ -248,7 +248,13 @@ impl Application for ZebradApp {
             // Ignore errors loading the config for some commands.
             Err(_e) if command.cmd().should_ignore_load_config_error() => Default::default(),
             Err(e) => {
-                status_err!("Zebra could not load its configuration: {}", e);
+                status_err!(
+                    "Zebra could not load the provided configuration file and/or environment variables.\
+                     This might mean you are using a deprecated format of the file, or are attempting to
+                     configure deprecated or unknown fields via environment variables.\
+                     You can generate a valid config by running \"zebrad generate\", \
+                     and diff it against yours to examine any format inconsistencies."
```
arya2, line 256:
```suggestion
                     and diff it against yours to examine any format inconsistencies. Error: {e}"
```

arya2 (approved):
These changes look great, thank you!

Opened https://github.com/ZcashFoundation/zebra/issues/9842 to track non-blocking comments that we should revisit.

arya2:
@Mergifyio refresh

gustavovalverde:
I'll admin merge to be able to clean the commit message when the PR merges

PR #9783, opened 2025-08-12
Replace light wallet .proto files with canonical versions obtained from https://github.com/zcash/lightwallet-protocol

nuttycom:
## Motivation

The Zcash ecosystem needs a canonical definition of the lightwalletd protocol; https://github.com/zcash/lightwallet-protocol is proposed as the location for that canonical set of definitions.

## Solution

This PR replaces the copies of the `service.proto` and `compact_formats.proto` files that have previously been used for local builds based on these definitions with symbolic links to the canonical versions, which will be maintained in the future using `git subtree`. The command used to bring in those files was:

```
git subtree add --prefix=zebrad/tests/common/lightwalletd/lightwallet-protocol --squash git@github.com:zcash/lightwallet-protocol.git main
```

### Tests

This PR relies on existing build and integration testing; as it does not actually modify any of the content of the .proto files (the canonical versions that were brought in in this PR are identical to the existing .proto files) there is no semantic change, nor any delta to any of the generated files produced by the build using this files.


arya2 (changes requested):
This looks good pending an update to the [paths here](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/build.rs#L70-L71).

conradoplg:
Note that `git subtree` [does not interact well with the squash merges we use in this repo](https://github.com/orgs/community/discussions/84731#discussioncomment-8400698). We'll need to manually merge this or figure out some other strategy.

nuttycom:
> Note that `git subtree` [does not interact well with the squash merges we use in this repo](https://github.com/orgs/community/discussions/84731#discussioncomment-8400698). We'll need to manually merge this or figure out some other strategy.

One possibility is that I could implement this using git submodules instead, if that works better for how ZF wants to manage this repository.

I will note that squash merges start to work less well in circumstances where it's necessary to make patch releases to previously released versions of library crates, and that might become more relevant as Zebra begins to be used as a library in Zaino.

conradoplg:
`git submodules` is even worse :sweat: we went through all this keeping the `zcashd` copy inside  `zcash_script`.

I think it's fine to merge this one manually, but I'll discuss this with the team

nuttycom:
> This looks good pending an update to the [paths here](https://github.com/ZcashFoundation/zebra/blob/main/zebrad/build.rs#L70-L71).

Could you clarify what your request is? Given that the files are being symlinked into their original locations, why does the build need to change?

arya2:
> Could you clarify what your request is? Given that the files are being symlinked into their original locations, why does the build need to change?

Ah, I missed the symlinks, no change needed.

arya2 (approved):

conradoplg:
Admin-merging to make merge commit and not squash

PR #9792, opened 2025-08-14
refactor: change test to use regtest instead of fake activation heights

conradoplg:
## Motivation

The test `all_upgrades_and_wrong_commitments_with_fake_activation_heights` used "fake" activation heights in order to test some of the commitment validation logic. Those heights had to be enabled with a env variable, which in turn had to be enabled in CI along with a explicit call for the test, etc. which is super annoying.

## Solution

With regtest support we can remove all that.

### Tests

The test kept passing indicating it worked.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -15,10 +15,6 @@ default-filter = 'not test(check_no_git_dependencies)'
 [profile.check-no-git-dependencies]
 default-filter = 'test(check_no_git_dependencies)'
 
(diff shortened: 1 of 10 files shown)


oxarbitrage (approved):
Thanks @conradoplg, more than happy to remove all that unnecessary complexity for a much simpler approach without losing coverage.

PR #9798, opened 2025-08-15
rpc: complete z_gettreestate

### LINKED ISSUE #9788: feature: complete z_gettreestate
### Motivation

We still don't support:

- ~~`sprout`~~ would require a DB upgrade because we don't store the tree for old heights. Let's do that in a separate issue, if at all
- ~~`<pool>.skipHash`~~ not required, it's only returned if `finalState` is not present, but we always return it
- `<pool>.commitments.finalRoot`

### Specifications

https://github.com/zcash/zcash/blob/b65b008a7b334a2f7c2eaae1b028e011f2e21dd1/src/rpc/blockchain.cpp#L1303

_No response_

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

Closes #9788 

I wanted to return Sprout data but then realized we can't since we don't store the Sprout trees for old heights. I left the Sprout fields in `zebra-rpc` because it might be useful for parsing responses e.g. if Zaino decides to support this.

## Solution

Return the relevant data. 

### Tests

Tested manually with `zcash-rpc-diff` (adjust if required):

```
OUTPUT_DATA_LINE_LIMIT=100000 ZEBRAD_EXTRA_ARGS=-conf=$PWD/zcash.conf ./zebra-utils/zcash-rpc-diff 28232 z_gettreestate 2345600
```

```diff
--- /tmp/tmp.1x1jQa2sOx.rpc-diff/zebrad-main-3030581-z_gettreestate.json	2025-08-15 17:01:27.672750114 -0300
+++ /tmp/tmp.1x1jQa2sOx.rpc-diff/zcashd-main-3030584-z_gettreestate.json	2025-08-15 17:01:27.674750104 -0300
@@ -10,7 +10,13 @@
   "sapling": {
     "commitments": {
       "finalRoot": "1ee678f964f116bddcbc88bde85f83879175d2799c829031190f0bf951688a81",
-      "finalState": "01010b25ed95064350503cde2f97413e03e71ff7a435f8dbc9851c6155c0f34057019d3ad195f921e183357d00b0cc2d144343bf96ed245238a6ca258bd4caf6fa0f1f017ae1444b3fbb30513e4295b63dfa66ac5b686a56d42dac38070b78d184b6875b013ff3f4e78bdce57ef0c2fcf5b7b51fbd8038afa874b6d87811c8229b00978a0d0164d04422451ef3b321652f1ea2363b11e3f94d707991de8b0720eaaf88db9e64000001df0b5acdfeff469afd0946955acf555c8c57d42de537c86e67d4eff96f5a544c01e1ef7404691126e701d4de7ad58b3393b2b7079dd0cc212745feb5c08bfefd270112e8d1661a37cab99b15feee91c2bfed4a07ef6d925acf6d55d456037366cf19000001f0dce1a197b0aa583830b99bd49e2b9cfe5be7701cd0e91b7ebb3a05de407c3001a667417a1a0fc346d5fcb68bea20c8f10a41d97167fec2d0b23b4e9dc5e51a2d01e3e205657bfbd810bea16aa5962f5765cbe6d3c774eff68ed0c2e1791d94323100017c650876ee17479db2e11cb5385df4fc60ae8f8cc8d501fa5dc07a4438e5863a000001d1b36bbba8e6e1be8f09baf2b829bafc4ccd89ad25fb730d2b8a995b60fc3a6701d8ccea507421a590ed38116b834189cdc22421b4764179fa4e364803dfa66d56018cf6f5034a55fed59d1d832620916a43c756d2379e50ef9440fc4c3e7a29aa2300011619f99023a69bb647eab2d2aa1a73c3673c74bb033c3c4930eacda19e6fd93b0000000160272b134ca494b602137d89e528c751c06d3ef4a87a45f33af343c15060cc1e0000000000"
+      "finalState": "01010b25ed95064350503cde2f97413e03e71ff7a435f8dbc9851c6155c0f34057019d3ad195f921e183357d00b0cc2d144343bf96ed245238a6ca258bd4caf6fa0f1a017ae1444b3fbb30513e4295b63dfa66ac5b686a56d42dac38070b78d184b6875b013ff3f4e78bdce57ef0c2fcf5b7b51fbd8038afa874b6d87811c8229b00978a0d0164d04422451ef3b321652f1ea2363b11e3f94d707991de8b0720eaaf88db9e64000001df0b5acdfeff469afd0946955acf555c8c57d42de537c86e67d4eff96f5a544c01e1ef7404691126e701d4de7ad58b3393b2b7079dd0cc212745feb5c08bfefd270112e8d1661a37cab99b15feee91c2bfed4a07ef6d925acf6d55d456037366cf19000001f0dce1a197b0aa583830b99bd49e2b9cfe5be7701cd0e91b7ebb3a05de407c3001a667417a1a0fc346d5fcb68bea20c8f10a41d97167fec2d0b23b4e9dc5e51a2d01e3e205657bfbd810bea16aa5962f5765cbe6d3c774eff68ed0c2e1791d94323100017c650876ee17479db2e11cb5385df4fc60ae8f8cc8d501fa5dc07a4438e5863a000001d1b36bbba8e6e1be8f09baf2b829bafc4ccd89ad25fb730d2b8a995b60fc3a6701d8ccea507421a590ed38116b834189cdc22421b4764179fa4e364803dfa66d56018cf6f5034a55fed59d1d832620916a43c756d2379e50ef9440fc4c3e7a29aa2300011619f99023a69bb647eab2d2aa1a73c3673c74bb033c3c4930eacda19e6fd93b0000000160272b134ca494b602137d89e528c751c06d3ef4a87a45f33af343c15060cc1e"
+    }
+  },
+  "sprout": {
+    "commitments": {
+      "finalRoot": "8b72175b2aca14e62b1d4f362c15dec59a9fa4ad9e42c46a62fe716790f6ae47",
+      "finalState": "01a343c765fe5432d6dd2237e3c876260473388a884e0e21a89eb1be3ae4a977b301fe5eaaf2db9b3c5efdd331dc8bcece335fc1ea663abe29f52af9e903dd70b8ae14000001afcc7344b2befbcc10a10340fc2c0b0792d50df83aeedd25f5ff343aa3ba1eb40140688a3603eaa68ae016dfe06da955450e52b4c78dcad0cd905955400c9f92a200000000000129e4b14eb580e8352ccc8100f28b66b5e90fb6807c6d3a5d94a40c13411a538e000196295a1e8d018c1c59f098e69350c7a206b850c828710fbaeb86d3ef72da3bb6000001806212ff57170f6904d462d24bc5df6112269a0b65defffbd4b78ff9a7859b9d0183a03915c566ab12ba59b130bab08a25b932ce204bf1774fc9e9267e5130f19101465dbb7f98de7f19e22f37c537660e9447fc27e1839238dc75548b047c41d1be01c14d7b1dfae9c25c284218d74c43621b795246a5fb3c238c2ca9d7f91d5aa86000011b1fc3241793d87854115ec7190ea06c64b76faa971c77ccc6316ec716f90f77"
     }
   },
   "time": 1703657236
```

- Orchard is exactly the same output (so it doesn't appear in the diff at all)
- Sapling differs only in `finalState` - it was already like this and it's due the trailing zeroes which we are also a valid encoding and we decided at some point to keep like this
- Sprout is still missing as expected

### PR Checklist

- [ ] The solution is tested.


zebra-rpc/src/methods.rs
```diff
@@ -1851,12 +1845,18 @@ where
                 .await
                 .map_misc_error()?
             {
-                zebra_state::ReadResponse::SaplingTree(tree) => tree.map(|t| t.to_rpc_bytes()),
+                zebra_state::ReadResponse::SaplingTree(tree) => tree.map(|t| {
+                    let mut root = Into::<[u8; 32]>::into(t.root()).to_vec();
+                    root.reverse();
+                    (t.to_rpc_bytes(), root)
```
arya2: Subjective/Nitpick/Optional: 
- It may be a little more readable without the trait qualifier
- Do we want to move the reversal to a `bytes_in_display_order()` method?

```suggestion
                    let mut root: [u8; 32] = t.root().into();
                    root.reverse();
                    (t.to_rpc_bytes(), root.to_vec())
```

zebra-rpc/src/methods.rs
```diff
@@ -1868,15 +1868,30 @@ where
                 .await
                 .map_misc_error()?
             {
-                zebra_state::ReadResponse::OrchardTree(tree) => tree.map(|t| t.to_rpc_bytes()),
+                zebra_state::ReadResponse::OrchardTree(tree) => tree.map(|t| {
+                    (
+                        t.to_rpc_bytes(),
+                        // Note: no need to reverse bytes for Orchard
+                        std::convert::Into::<[u8; 32]>::into(t.root()).to_vec(),
```
arya2: Optional:

```suggestion
                        t.to_rpc_bytes(),
                        // Note: no need to reverse bytes for Orchard
                        <[u8; 32]>::from(t.root()).to_vec(),
```

arya2 (dismissed):
Looks good! Left a couple small optional suggestions, no blockers.

zebra-rpc/src/methods.rs
conradoplg: Good idea, `bytes_in_display_order()` sounds good (we do have an inconsistency that we also have `to_rpc_bytes()` but we can sort that out later). Done in 99e6bf0c3

zebra-rpc/src/methods.rs
conradoplg: Done in 99e6bf0c3

arya2 (approved):
Thank you!

conradoplg:
@mergifyio update

PR #9809, opened 2025-08-19
add(state): Backup non-finalized blocks in state cache directory

### LINKED ISSUE #9708: Backup blocks in the best chain of the non-finalized state to files
## Motivation

We want to avoid irrecoverably dropping blocks in the non-finalized state when Zebra restarts before zcashd deprecation to mitigate the risk of the network (or part of the network) losing those blocks for an extended period of time (or indefinitely).

### Design

When starting the state service:
- Create, if needed, a non-finalized state best chain backup directory when starting the state service that will contain one file per block in Zebra's non-finalized best chain using block hashes as filenames
- Read, parse, and commit to the non-finalized state all blocks in the backup directory immediately after instantiating the non-finalized state (ensuring that the backup has been loaded before deleting any files) 
- Spawn a task in the state to backup the best chain of the latest non-finalized state, read the list of filenames (block hashes) in the backup directory, and:
  - delete any that are missing in the non-finalized best chain, and
  - atomically write any blocks that are in the non-finalized best chain but are missing in the backup directory

arya2:
## Motivation

We want Zebra to recover its non-finalized state when it restarts.

Closes #9708.

## Solution

- Creates a `{state_cache_dir}/non_finalized_state/{network.lowercase_name()}` directory
- Adds a config field to allow for disabling non-finalized state backups
- Attempts to commit blocks from the non-finalized backup cache into the non-finalized state when constructing the `StateService`
- Listens to the non-finalized state receiver and:
  - Writes any blocks in the non-finalized state that are missing in the backup cache to the backup cache, and
  - Deletes any files in the backup cache that are not present in the non-finalized state.

Related changes:
- Checks the non-finalized state for a block candidate's parent block before checking RocksDB

### Tests

This PR adds a unit test and has been manually tested.

### Follow-up Work

Add a config field to enable/disable the non-finalized state backup cache.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/fmt/time.rs
+++ b/zebra-chain/src/fmt/time.rs
@@ -5,14 +5,19 @@ use std::time::Duration;
 /// The minimum amount of time displayed with only seconds (no milliseconds).
 pub const MIN_SECONDS_ONLY_TIME: Duration = Duration::from_secs(5);
 
+/// The minimum amount of time displayed with only seconds and milliseconds (no microseconds).
+pub const MIN_MS_ONLY_TIME: Duration = Duration::from_millis(5);
+
 /// Returns a human-friendly formatted string for the whole number of seconds in `duration`.
 pub fn duration_short(duration: impl Into<Duration>) -> String {
     let duration = duration.into();
 
     if duration >= MIN_SECONDS_ONLY_TIME {
         humantime_seconds(duration)
-    } else {
+    } else if duration >= MIN_MS_ONLY_TIME {
         humantime_milliseconds(duration)
+    } else {
+        humantime_microseconds(duration)
     }
 }
 
@@ -39,6 +44,20 @@ pub fn humantime_milliseconds(duration: impl Into<Duration>) -> String {
     let duration_millis = Duration::from_millis(duration.subsec_millis().into());
 
     let duration = humantime::format_duration(duration_secs + duration_millis);
+
+    format!("{duration}")
+}
+
+/// Returns a human-friendly formatted string for the whole number of milliseconds in `duration`.
+pub fn humantime_microseconds(duration: impl Into<Duration>) -> String {
+    let duration = duration.into();
+
+    // Truncate fractional seconds.
+    let duration_secs = Duration::from_secs(duration.as_secs());
+    let duration_millis = Duration::from_millis(duration.subsec_millis().into());
+    let duration_micros = Duration::from_micros(duration.subsec_micros().into());
+
+    let duration = humantime::format_duration(duration_secs + duration_millis + duration_micros);
 
     format!("{duration}")
 }
--- a/zebra-chain/src/fmt/time.rs
+++ b/zebra-chain/src/fmt/time.rs
@@ -48,7 +48,7 @@ pub fn humantime_milliseconds(duration: impl Into<Duration>) -> String {
     format!("{duration}")
 }
 
-/// Returns a human-friendly formatted string for the whole number of milliseconds in `duration`.
+/// Returns a human-friendly formatted string for the whole number of microseconds in `duration`.
 pub fn humantime_microseconds(duration: impl Into<Duration>) -> String {
     let duration = duration.into();
 
--- a/zebra-state/src/service/non_finalized_state/backup.rs
+++ b/zebra-state/src/service/non_finalized_state/backup.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{BTreeMap, HashMap},
     fs::DirEntry,
+    io::{self, ErrorKind},
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
@@ -8,17 +9,21 @@ use std::{
 
 use hex::ToHex;
 use zebra_chain::{
+    amount::{Amount, DeferredPoolBalanceChange},
     block::{self, Block, Height},
     serialization::{ZcashDeserializeInto, ZcashSerialize},
 };
 
 use crate::{
-    service::write::validate_and_commit_non_finalized, ContextuallyVerifiedBlock,
-    NonFinalizedState, SemanticallyVerifiedBlock, WatchReceiver, ZebraDb,
+    ContextuallyVerifiedBlock, IntoDisk, NonFinalizedState, SemanticallyVerifiedBlock,
+    WatchReceiver, ZebraDb,
 };
 
+#[cfg(not(test))]
+use crate::service::write::validate_and_commit_non_finalized;
+
 /// The minimum duration that Zebra will wait between updates to the non-finalized state backup cache.
-const MIN_DURATION_BETWEEN_BACKUP_UPDATES: Duration = Duration::from_secs(5);
+pub(crate) const MIN_DURATION_BETWEEN_BACKUP_UPDATES: Duration = Duration::from_secs(5);
 
 /// Accepts an optional path to the non-finalized state backup directory and a handle to the database.
 ///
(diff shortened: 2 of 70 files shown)


zebra-chain/src/fmt/time.rs
```diff
@@ -42,3 +47,17 @@ pub fn humantime_milliseconds(duration: impl Into<Duration>) -> String {
 
     format!("{duration}")
 }
+
+/// Returns a human-friendly formatted string for the whole number of milliseconds in `duration`.
```
arya2:
```suggestion
/// Returns a human-friendly formatted string for the whole number of microseconds in `duration`.
```

(These changes were made while checking for places where `spawn_blocking()` may have been needed since `StateService::new()` has been made into an async method)

zebra-state/src/service/non_finalized_state/backup.rs
```diff
@@ -0,0 +1,326 @@
...
+impl From<&ContextuallyVerifiedBlock> for NonFinalizedBlockBackup {
+    fn from(cv_block: &ContextuallyVerifiedBlock) -> Self {
+        Self {
+            block: cv_block.block.clone(),
+            deferred_pool_balance_change: cv_block.chain_value_pool_change.deferred_amount(),
+        }
+    }
+}
+
+impl NonFinalizedBlockBackup {
```
arya2, line 142: We could also compute the deferred pool balance change instead of reading/writing it. 

Let me know if that's preferrable.

```rust
// Calculates the [`DeferredPoolBalanceChange`] of a post-NU6 transaction by subtracting the total
// value balance of all of its outputs from its coinbase input value.
//
// > From the activation block of this ZIP onward, coinbase transactions MUST claim all of the available
// > miner subsidy plus fees in their block.
// <https://zips.z.cash/zip-0236>, ZIP 236: Blocks should balance exactly
// <https://zips.z.cash/zip-0253>, ZIP 253: Deployment of the NU6 Network Upgrade
//
// > • A coinbase transaction MUST NOT have any JoinSplit descriptions.
// > • A coinbase transaction MUST NOT have any Spend descriptions.
// > • [Pre-Heartwood] A coinbase transaction MUST NOT have any Output descriptions.
// > • [ NU5onward] In a version 5 coinbase transaction, the enableSpendsOrchard flag MUST be 0
// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>, section 7.1.2, Transaction Consensus Rules (p. 122)
```

arya2:
This PR will need to update tests using `future_blocks()` or `raw_future_blocks()`, currently the `rpc_submit_block()` and `send_raw_transaction()` tests.

Update: Fixed in https://github.com/ZcashFoundation/zebra/pull/9809/commits/6d8dcf9b2f27ebff93848ab8ee3c4b04378f4299

zebra-state/src/service/write.rs
```diff
@@ -57,7 +57,9 @@ pub(crate) fn validate_and_commit_non_finalized(
     check::initial_contextual_validity(finalized_state, non_finalized_state, &prepared)?;
     let parent_hash = prepared.block.header.previous_block_hash;
 
-    if finalized_state.finalized_tip_hash() == parent_hash {
+    if !non_finalized_state.any_chain_contains(&parent_hash)
```
oxarbitrage, line 60: Did you spotted an issue here or is just extra checking ?

oxarbitrage:
I tried to add this test: https://github.com/ZcashFoundation/zebra/commit/ec1faa31534eba68fa34a9736e46f6efbc8651e7 with this branch and i think there is an issue after you recover the non finalized state. It seems the chain can't make progress after that. Can you take a look ?

Also, can you paste some logs of your manual testing ? Thanks.

oxarbitrage:
I have an additional test for this. Here is the code, let me know if you will like me to push it as a commit here after https://github.com/ZcashFoundation/zebra/pull/9888 gets merged or if you want me to do it in a separated PR after https://github.com/ZcashFoundation/zebra/pull/9809 gets merged.

zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py:

```python
#!/usr/bin/env python3
# Copyright (c) 2025 The Zcash developers
# Distributed under the MIT software license, see the accompanying
# file COPYING or https://www.opensource.org/licenses/mit-license.php .

from decimal import Decimal
from test_framework.util import assert_equal, start_nodes
from test_framework.test_framework import BitcoinTestFramework
import time

# Test that Zebra can backup and restore non finalized state
class BackupNonFinalized(BitcoinTestFramework):

    def __init__(self):
        super().__init__()
        self.num_nodes = 1
        self.cache_behavior = 'clean'

    def setup_network(self):
        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [None])

    def run_test(self):
        self.nodes[0].generate(10)
        # Wait for `MIN_DURATION_BETWEEN_BACKUP_UPDATES`, 5 secs in zebra
        time.sleep(5)

        blocks = self.nodes[0].getblockchaininfo()['blocks']
        assert_equal(blocks, 10)

        # Stop the node
        self.nodes[0].stop()
        time.sleep(1)

        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [None])

        # Good, the node has recovered the non finalized state
        blocks = self.nodes[0].getblockchaininfo()['blocks']
        assert_equal(blocks, 10)

        # Generate more blocks and make sure the blockchain is still working
        self.nodes[0].generate(1)
        time.sleep(5)

        blocks = self.nodes[0].getblockchaininfo()['blocks']
        assert_equal(blocks, 11)

if __name__ == '__main__':
    BackupNonFinalized().main()
```

Also, we need to get rid of the hack introduced at https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/wallet.py#L51 to bypass the limitation. We can do it in a separated PR.

arya2:
> I have an additional test for this. Here is the code, let me know if you will like me to push it as a commit here after #9888 gets merged or if you want me to do it in a separated PR after #9809 gets merged.
> 
> Also, we need to get rid of the hack introduced at https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/wallet.py#L51 to bypass the limitation. We can do it in a separated PR.

Could you open a suggestion PR against this one with both the new test and the removal of the workaround in the existing test? We could also include it in a follow-up PR to minimize the changes in this one.

It seems the chain stall issue may not have been fixed, as all of [the integration tests on this PR are still timing out](https://github.com/ZcashFoundation/zebra/actions/runs/17671838575/job/50227052558?pr=9809#step:13:337), even though the new acceptance test in https://github.com/ZcashFoundation/zebra/pull/9888 is passing.

Update: Found and fixed the third and seemingly final issue with the chain stall in https://github.com/ZcashFoundation/zebra/pull/9809/commits/b33136736d3721a61b45bfe22a11850c8564fbba. ~The acceptance test added in the follow up PR #9888 doesn't yet cover it.~

Update: Added test coverage for the last issue in https://github.com/ZcashFoundation/zebra/pull/9888/commits/88af388bdd9a3aed1baf3b891c7b11f2042da71e.

oxarbitrage:
Yes, i opened https://github.com/ZcashFoundation/zebra/pull/9896

But based on https://github.com/ZcashFoundation/zebra/pull/9888

oxarbitrage (approved):

zebra-state/src/service/write.rs
arya2: It's just an extra check to avoid disk I/O

Issue #9806, opened 2025-08-18
cleanup: trait with `bytes_in_display_order()`


conradoplg:
### Motivation

A lot of types require a specific byte-reversed encoding when returned in RPC.

We do that in multiple ways: some types have a `bytes_in_display_order()`, others have `to_rpc_bytes()`, and other are reversed manually in the RPC implementation.

We should introduce a new trait with a `bytes_in_display_order()` method and change existing code to always use it (searching for `.reversed()` will give a good list of places to change). Having a trait allows us to also implement it in foreign types like Ed25519 signatures.

syszery:
Hi! I’ve submitted a draft PR implementing the `BytesInDisplayOrder` trait based on the description here. Please let me know if you have any feedback or want me to make adjustments, like splitting the trait for partial implementations as mentioned in the PR message.

natalieesk:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-5fb24d9264a3e8000e666a9e/issues/gh/ZcashFoundation/zebra/9806?planning-poker) with Zenhub @arya2

syszery:
Hi! I just noticed I forgot to include a `closes #9806` or similar in the PR description. If anything is still missing from the implementation, I'm happy to follow up.

arya2:
Fixed by https://github.com/ZcashFoundation/zebra/pull/9810

PR #9810, opened 2025-08-19
feat: introduce `BytesInDisplayOrder` trait to standardize byte-reversed encoding in RPC (#9806)

syszery:
## Motivation

The PR introduces a new trait called `BytesInDisplayOrder` as suggested in issue #9806.

## Solution

- Added a new module `zebra-chain/src/serialization/display_order.rs` containing the `BytesInDisplayOrder` trait.
- The trait provides methods `bytes_in_display_order` and `from_bytes_in_display_order` for converting byte arrays.
- It also includes helper methods `bytes_in_serialized_order` and `from_bytes_in_serialized_order`.
- Updated all relevant hash structs in `zebra-chain` to implement this trait, replacing previous byte order handling with trait usage.

One limitation is `NotSmallOrderValueCommitment` in `zebra-chain/src/sapling/commitment.rs`, which only implements `bytes_in_display_order`. The inverse `method from_bytes_in_display_order` is currently not implemented and cannot be easily called (because it might fail with bad input).

**Question:** Should the trait be split into two smaller traits—one with only `bytes_in_display_order` and one with `from_bytes_in_display_order` -- to accommodate such cases where only one direction is applicable?

### Tests

- All existing tests passed without modification after the change.

Issue #9806

### Follow-up Work

Decide whether to split the trait into two to better support types that only implement one direction of the conversion.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/serialization/display_order.rs
+++ b/zebra-chain/src/serialization/display_order.rs
@@ -1,0 +1,32 @@
+//! Trait for converting types to their byte representation for display or RPC use.
+///
+/// This trait provides methods to access and construct types from their
+/// internal serialized byte order (typically little-endian) as well as their
+/// big-endian byte order used for display and RPC.
+pub trait BytesInDisplayOrder: Sized {
+    /// Returns the bytes in the internal serialized order.
+    fn bytes_in_serialized_order(&self) -> [u8; 32];
+
+    /// Creates an instance from bytes in the internal serialized order.
+    fn from_bytes_in_serialized_order(bytes: [u8; 32]) -> Self;
+
+    /// Return the bytes in big-endian byte-order suitable for printing out byte by byte.
+    ///
+    /// Zebra displays transaction and block hashes in big-endian byte-order,
+    /// following the u256 convention set by Bitcoin and zcashd.
+    fn bytes_in_display_order(&self) -> [u8; 32] {
+        let mut reversed_bytes = self.bytes_in_serialized_order();
+        reversed_bytes.reverse();
+        reversed_bytes
+    }
+
+    /// Convert bytes in big-endian byte-order into big-endian display order.
+    ///
+    /// Zebra displays transaction and block hashes in big-endian byte-order,
+    /// following the u256 convention set by Bitcoin and zcashd.
+    fn from_bytes_in_display_order(bytes_in_display_order: &[u8; 32]) -> Self {
+        let mut internal_byte_order = *bytes_in_display_order;
+        internal_byte_order.reverse();
+        Self::from_bytes_in_serialized_order(internal_byte_order)
+    }
+}
--- a/zebra-chain/src/serialization/display_order.rs
+++ b/zebra-chain/src/serialization/display_order.rs
@@ -3,6 +3,9 @@
 /// This trait provides methods to access and construct types from their
 /// internal serialized byte order (typically little-endian) as well as their
 /// big-endian byte order used for display and RPC.
+///
+/// Zebra displays transaction and block hashes in big-endian byte-order,
+/// following the u256 convention set by Bitcoin and zcashd.
 pub trait BytesInDisplayOrder<
     const SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER: bool = false,
     const BYTE_LEN: usize = 32,
--- a/zebra-chain/src/serialization/display_order.rs
+++ b/zebra-chain/src/serialization/display_order.rs
@@ -3,30 +3,32 @@
 /// This trait provides methods to access and construct types from their
 /// internal serialized byte order (typically little-endian) as well as their
 /// big-endian byte order used for display and RPC.
-pub trait BytesInDisplayOrder: Sized {
+pub trait BytesInDisplayOrder<
+    const SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER: bool = false,
+    const BYTE_LEN: usize = 32,
+>: Sized
+{
     /// Returns the bytes in the internal serialized order.
-    fn bytes_in_serialized_order(&self) -> [u8; 32];
+    fn bytes_in_serialized_order(&self) -> [u8; BYTE_LEN];
 
     /// Creates an instance from bytes in the internal serialized order.
-    fn from_bytes_in_serialized_order(bytes: [u8; 32]) -> Self;
+    fn from_bytes_in_serialized_order(bytes: [u8; BYTE_LEN]) -> Self;
 
     /// Return the bytes in big-endian byte-order suitable for printing out byte by byte.
-    ///
-    /// Zebra displays transaction and block hashes in big-endian byte-order,
-    /// following the u256 convention set by Bitcoin and zcashd.
-    fn bytes_in_display_order(&self) -> [u8; 32] {
+    fn bytes_in_display_order(&self) -> [u8; BYTE_LEN] {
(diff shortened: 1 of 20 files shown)


zebra-network/proptest-regressions/peer_set/set/tests/prop.txt
arya2, line 1: Please delete this file.

zebra-chain/src/serialization/display_order.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// Convert bytes in big-endian byte-order into big-endian display order.
+    ///
+    /// Zebra displays transaction and block hashes in big-endian byte-order,
+    /// following the u256 convention set by Bitcoin and zcashd.
+    fn from_bytes_in_display_order(bytes_in_display_order: &[u8; 32]) -> Self {
+        let mut internal_byte_order = *bytes_in_display_order;
+        internal_byte_order.reverse();
+        Self::from_bytes_in_serialized_order(internal_byte_order)
+    }
```
arya2: I had something simpler in mind where it would just define the trait method signatures and implement them individually for each type that has a `bytes_in_display_order()` method, but, this might be a little more concise.

It will need to return arrays of varying sizes, and not all types reverse display order is reversed, e.g. `ConsensusBranchId`.

```suggestion
pub trait BytesInDisplayOrder<
    const SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER: bool = false,
    const BYTE_LEN: usize = 32,
>: Sized
{
    /// Returns the bytes in the internal serialized order.
    fn bytes_in_serialized_order(&self) -> [u8; BYTE_LEN];

    /// Creates an instance from bytes in the internal serialized order.
    fn from_bytes_in_serialized_order(bytes: [u8; BYTE_LEN]) -> Self;

    /// Return the bytes in big-endian byte-order suitable for printing out byte by byte.
    fn bytes_in_display_order(&self) -> [u8; BYTE_LEN] {
        let mut reversed_bytes = self.bytes_in_serialized_order();
        if SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER {
            reversed_bytes.reverse();
        }
        reversed_bytes
    }

    /// Convert bytes in big-endian byte-order into big-endian display order.
    fn from_bytes_in_display_order(bytes_in_display_order: &[u8; BYTE_LEN]) -> Self {
        let mut internal_byte_order = *bytes_in_display_order;
        if SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER {
            internal_byte_order.reverse();
        }
        Self::from_bytes_in_serialized_order(internal_byte_order)
    }
```

arya2 (commented):
Thank you for this PR!

> Should the trait be split into two smaller traits—one with only bytes_in_display_order and one with from_bytes_in_display_order -- to accommodate such cases where only one direction is applicable?

It's only one outlier method, I think it'd be better to leave that one as-is, but it'd also be fine to split the trait into two traits if you prefer.

syszery:
Addressed all review feedback:
- Deleted the test file.
- Replaced the old `BytesInDisplayOrder` trait with the suggested `BytesInDisplayOrder<true, 32>` version across `zebra-chain`. All hash types were confirmed to be 32 bytes and reverse display order.
- Ran cargo fmt and verified all tests pass.

Thanks @arya2 for your detailed feedback!

Let me know if you'd like me to update other crates to use the new trait as well.

zebra-chain/src/transaction/auth_digest.rs
```diff
@@ -26,26 +27,13 @@ use proptest_derive::Arbitrary;
 #[cfg_attr(any(test, feature = "proptest-impl"), derive(Arbitrary))]
 pub struct AuthDigest(pub [u8; 32]);
 
-impl AuthDigest {
-    /// Return the hash bytes in big-endian byte-order suitable for printing out byte by byte.
-    ///
-    /// Zebra displays transaction and block hashes in big-endian byte-order,
-    /// following the u256 convention set by Bitcoin and zcashd.
```
arya2, line 33: It would be nice to move these docs to the new `display_order` module instead of removing them.

zebra-chain/src/serialization/display_order.rs
```diff
@@ -0,0 +1,34 @@
+//! Trait for converting types to their byte representation for display or RPC use.
+///
+/// This trait provides methods to access and construct types from their
+/// internal serialized byte order (typically little-endian) as well as their
+/// big-endian byte order used for display and RPC.
+pub trait BytesInDisplayOrder<
+    const SHOULD_REVERSE_BYTES_IN_DISPLAY_ORDER: bool = false,
+    const BYTE_LEN: usize = 32,
```
arya2, line 11: It would be nice to use this for `ExpandedDifficulty`, `CompactDifficulty`, and `ConsensusBranchId` too now that it can return varying array sizes.

zebra-chain/src/transaction/auth_digest.rs
```diff
@@ -26,26 +27,13 @@ use proptest_derive::Arbitrary;
...
-    pub fn bytes_in_display_order(&self) -> [u8; 32] {
-        let mut reversed_bytes = self.0;
-        reversed_bytes.reverse();
-        reversed_bytes
+impl BytesInDisplayOrder<true, 32> for AuthDigest {
```
arya2: It would be nice to omit the second generic parameter where `32` is being used, as that's the default value.

arya2 (dismissed):
This looks good to merge, thank you for these changes!

I left a few optional suggestions, feel free to resolve them you want this to merge as-is.

syszery:
Hi arya2,

Thanks again for your review and helpful suggestions. I’ve added the requested comment and removed the explicit constant where it wasn’t necessary, as per your advice.

Regarding the three other structs you mentioned for implementing the `BytesInDisplayOrder` trait — I found that the trait fits cleanly only for `ExpandedDifficulty`.

- For `CompactDifficulty`, the current implementation (see [line 280](https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/work/difficulty.rs#L280)) returns a `Result`, which isn’t supported by the trait — so a direct implementation isn’t straightforward without changing the trait (or ignoring the error case).

- `ConsensusBranchId` previously only implemented `bytes_in_display_order`, and the trait adds the reverse conversion. Do you think exposing this additional method could have any unintended implications (e.g., consensus or security-wise)?

arya2 (dismissed):
There are some small merge conflicts, but this PR looks good! Thank you @syszery.

syszery:
**Update**: Rebased the branch onto the latest main and resolved merge conflicts.
Also fixed Clippy warnings in the `lightwalletd` integration test.

Thanks @arya2 for the helpful guidance!

zebrad/proptest-regressions/components/mempool/storage/tests/prop.txt
arya2, line 1: Could we restore this file? It's only the network prop.txt file that we don't want to check into git, because it depends on network conditions.

arya2 (dismissed):

zebrad/proptest-regressions/components/mempool/storage/tests/prop.txt
syszery: Restored `prop.txt` as requested. Sorry for the confusion!

conradoplg (approved):

arya2 (approved):

PR #9828, opened 2025-08-25
cryptography(sapling): adopt upstream types for commitments where possible

### LINKED ISSUE #9578: Making changes in sapling-crypto to use NoteCommitment in zebra (make API public)

oxarbitrage:
## Motivation

We want to remove custom implementations for Sapling primitives and rely on the upstream `sapling-crypto` crate wherever possible. This reduces maintenance overhead and ensures consistency with the canonical implementation.

Closes https://github.com/ZcashFoundation/zebra/issues/9578

## Solution

- **Value commitments**: Replace the custom `NotSmallOrderValueCommitment` with the upstream `ValueCommitment`. The upstream constructor `from_bytes_not_small_order` [already enforces the no-small-order condition](https://github.com/zcash/sapling-crypto/blob/485966bc816e6436522ebac4a97b3af0c8d20699/src/value.rs#L131). A thin wrapper is still needed to provide traits and functionality that aren’t currently derived upstream – https://github.com/ZcashFoundation/zebra/commit/34ae17c248d8bde165ee320f5aeb84450e2abe8b
- **Note commitments**: Switch to using `ExtractedNoteCommitment` for coordinates – https://github.com/ZcashFoundation/zebra/commit/a961df99e9f8552dce078b9b27a9e0a5b6acbdff
- **Pedersen hashes**: Remove all non needed pedersen hashes in house code in favour of upstream functions – https://github.com/ZcashFoundation/zebra/pull/9828/commits/a4928876fedc788cbe57e27a6a22185e3d17db77, https://github.com/ZcashFoundation/zebra/pull/9828/commits/be8afc70e087bf87c9d19e9ce9fd6bf5d4b268c2

### Tests

Some unit tests in `commitments.rs` were removed because they duplicated upstream coverage. The same with some note commitment tree tests. The rest of the test suite, especially full sync tests, should validate correctness of this change.

### PR Checklist

- [ ] The solution is tested.


zebra-consensus/src/primitives/groth16.rs
```diff
@@ -217,7 +217,7 @@ impl Description for Spend<PerSpendAnchor> {
         inputs.push(rk_affine.get_u());
         inputs.push(rk_affine.get_v());
 
-        let cv_affine = jubjub::AffinePoint::from(self.cv);
+        let cv_affine = jubjub::AffinePoint::from_bytes(self.cv.0.to_bytes()).unwrap();
```
oxarbitrage, line 129: These unwraps might look suspicious, but given that the function already uses them elsewhere, I assume they’re acceptable. Worth a second look though.

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -459,13 +459,21 @@ pub struct ShieldedSpend {
     spend_auth_sig: [u8; 64],
 }
 
+// We can't use `#[getter(skip)]` as upstream `sapling_crypto::note::ValueCommitment` is not `Copy`.
```
oxarbitrage: I introduced this to avoid a breaking change. There are alternatives (for example, using the [getset](https://crates.io/crates/getset) crate, which can even be used in parallel with our current derive-getters crate), but I’m not sure the if it worth here.

zebra-chain/src/sapling/tree.rs
```diff
@@ -275,17 +226,26 @@ impl Hashable for Node {
...
-    /// Level 0 is the layer above the leaves (layer 31).
-    /// Level 31 is the root (layer 0).
     fn combine(level: incrementalmerkletree::Level, a: &Self, b: &Self) -> Self {
-        let layer = MERKLE_DEPTH - 1 - u8::from(level);
-        Self(merkle_crh_sapling(layer, a.0, b.0))
+        Self(
+            sapling_crypto::Node::combine(
+                level,
+                &sapling_crypto::Node::from_bytes(a.0).unwrap(),
+                &sapling_crypto::Node::from_bytes(b.0).unwrap(),
```
oxarbitrage: Maybe use some dummy defaults here instead of panicking ?

conradoplg: This will cause a big performance hit because `from_bytes()` is CPU expensive. I think the right approach would be to replace our `Node` with `sapling_crypto::Node`, which will allow deleting even more code. Let me know if this is something you'd like to investigate, otherwise I can take a look if you prefer.

oxarbitrage: Did that at https://github.com/ZcashFoundation/zebra/pull/9828/commits/e865d8c83b5f84054237f88d095a0fa3b1f12457

conradoplg: Awesome, thank you!

zebra-chain/src/sapling/spend.rs
```diff
@@ -257,8 +259,10 @@ impl ZcashDeserialize for SpendPrefixInTransactionV5 {
         Ok(SpendPrefixInTransactionV5 {
             // Type is `ValueCommit^{Sapling}.Output`, i.e. J
             // https://zips.z.cash/protocol/protocol.pdf#abstractcommit
-            // See [`commitment::NotSmallOrderValueCommitment::zcash_deserialize`].
-            cv: commitment::NotSmallOrderValueCommitment::zcash_deserialize(&mut reader)?,
+            // See [`sapling_crypto::value::ValueCommitment::::zcash_deserialize`].
+            cv: commitment::ValueCommitment(
+                sapling_crypto::value::ValueCommitment::zcash_deserialize(&mut reader)?,
+            ),
```
conradoplg:
```suggestion
            // See [`sapling_crypto::value::ValueCommitment::zcash_deserialize`].
            cv: commitment::ValueCommitment::zcash_deserialize(&mut reader)?,
```

zebra-consensus/src/primitives/groth16.rs
conradoplg: Yep this is fine since deserializing right after serializing must always work.

However, I noticed that this code isn't even used anymore now that we switched to the `sapling_crypto` verifier (#9737). We can clean that up later.

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -460,13 +460,21 @@ pub struct ShieldedSpend {
...
```
conradoplg: Yep that's fine as is I think. Just a fix for the comment:

```suggestion
// We can't use `#[getter(copy)]` as upstream `sapling_crypto::note::ValueCommitment` is not `Copy`.
```

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -484,6 +492,14 @@ pub struct ShieldedOutput {
     proof: [u8; 192],
 }
 
+// We can't use `#[getter(skip)]` as upstream `sapling_crypto::note::ValueCommitment` is not `Copy`.
```
conradoplg:
```suggestion
// We can't use `#[getter(copy)]` as upstream `sapling_crypto::note::ValueCommitment` is not `Copy`.
```

zebra-chain/src/serialization/serde_helpers.rs
```diff
@@ -65,3 +65,54 @@ impl From<Base> for pallas::Base {
...
+#[derive(Deserialize, Serialize)]
+#[serde(remote = "sapling_crypto::value::ValueCommitment")]
+pub struct ValueCommitment {
+    #[serde(getter = "sapling_crypto::value::ValueCommitment::to_bytes")]
+    bytes: [u8; 32],
+}
+
+impl From<ValueCommitment> for sapling_crypto::value::ValueCommitment {
+    fn from(local: ValueCommitment) -> Self {
+        sapling_crypto::value::ValueCommitment::from_bytes_not_small_order(&local.bytes).unwrap()
```
conradoplg, line 78: I'm a bit worried about this `unwrap()`. Taking a step back, what is this even used for? I see it's ultimately used to derive `Serialize/Deserialize` for `Output`, but if I comment that out, nothing seems to break. Could we remove these entirely? (If so, we can do that in a separate PR)

If it's truly needed, then my suggestion would be for `zebra_chain::sapling::commitment::ValueCommitment` itself  to use a `#[serde(try_from = "FromType")]` (and `into`) instead of using serde remote with the inner value, which allows us to make the deserialization fallible.

zebra-chain/src/serialization/serde_helpers.rs
```diff
@@ -65,3 +65,54 @@ impl From<Base> for pallas::Base {
...
+#[derive(Deserialize, Serialize)]
+#[serde(remote = "sapling_crypto::note::ExtractedNoteCommitment")]
+pub struct SaplingExtractedNoteCommitment {
+    #[serde(getter = "SaplingExtractedNoteCommitment::as_serializable_bytes")]
+    bytes: [u8; 32],
+}
+
+impl From<SaplingExtractedNoteCommitment> for sapling_crypto::note::ExtractedNoteCommitment {
+    fn from(local: SaplingExtractedNoteCommitment) -> Self {
+        sapling_crypto::note::ExtractedNoteCommitment::from_bytes(&local.bytes).unwrap()
```
conradoplg, line 91: Same as the previous comment.

conradoplg (commented):
Thank you, this is great. I left a bunch of optional clean up comments, and one comment that might take a bit more work (or might lead to deleting unused code).

zebra-rpc/src/methods/types/transaction.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9828/commits/4b69d1a52eadfa56bb55f5fbb582b9ca305bc9bb

zebra-rpc/src/methods/types/transaction.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9828/commits/4b69d1a52eadfa56bb55f5fbb582b9ca305bc9bb

zebra-chain/src/sapling/spend.rs
oxarbitrage: https://github.com/ZcashFoundation/zebra/pull/9828/commits/4b69d1a52eadfa56bb55f5fbb582b9ca305bc9bb

zebra-chain/src/serialization/serde_helpers.rs
oxarbitrage: I tried to remove this but i think they are needed. Can you show me what you commented out exactly ?

zebra-consensus/src/primitives/groth16.rs
oxarbitrage: Opened https://github.com/ZcashFoundation/zebra/issues/9891 for that, feel free to add more details.

zebra-chain/src/serialization/serde_helpers.rs
conradoplg: Sure, sorry for not being clear. I think I also mixed things up.

I retrace my steps and this is what I came up. If you:

- Remove `Serialize/Deserialize` from `ValueCommitment`
- Remove them from `sapling::output::Output`, `OutputInTransactionV4`, `OutputPrefixInTransactionV5`, `TransferData`,  `ShieldedData`, `Spend`, `SpendPrefixInTransactionV5`, `Transaction`, `Block`...

Then the only thing that breaks is `snapshot_rpc_getblocktemplate` which for some reason tests the serde serialization of `Transaction`, and the `elastic-search` feature. Maybe the `elastic-search` feature could use the same block/transaction serialization that RPC does, so that we could remove all that serde serialization code?

Anyway this is pretty offtopic for this PR, I'm convinced that this `unwrap()` it not called anywhere critical, so I can create an issue to look this up later

zebra-chain/src/serialization/serde_helpers.rs
conradoplg: We can sort this out later

zebra-chain/src/serialization/serde_helpers.rs
conradoplg: Tracking on https://github.com/ZcashFoundation/zebra/issues/9892

conradoplg (approved):
Thanks!

zebra-chain/src/serialization/serde_helpers.rs
oxarbitrage: Got it, thanks.

PR #9845, opened 2025-08-28
Enforce strict non-empty funding stream height ranges

Fibonacci747:
Replace the non-strict start <= end assert with a strict start < end in ConfiguredFundingStreams::convert_with_default() to ensure funding stream height ranges are non-empty. This aligns the check with the error message and prevents potential panics in num_funding_stream_addresses_required_for_height_range() where .end.previous() and subsequent arithmetic assume start < end.

### DIFF
--- a/zebra-chain/src/parameters/network/testnet.rs
+++ b/zebra-chain/src/parameters/network/testnet.rs
@@ -204,7 +204,7 @@ impl ConfiguredFundingStreams {
             });
 
         assert!(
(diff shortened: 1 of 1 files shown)


arya2:
It's fine for the funding stream height range to be empty.

This was a deliberate change made in https://github.com/ZcashFoundation/zebra/pull/9710, it's likely unnecessary to allow empty funding stream height ranges now that we've refactored them, but there's also no reason to disallow them.

PR #9861, opened 2025-09-03
build: speed up dev profile by improving optimization of some dependencies

conradoplg:
## Motivation

In #9741 , @upbqdn noticed that [a particular test](https://github.com/ZcashFoundation/zebra/pull/9741/files#diff-6928ca8b4f6b3ed3d402856e2cffd5c0a2772abe534006177b257e3e61fea5e0R36) was much slower in the dev/test profile than in the release profile, even though we use `opt-level=3` in dev mode for some key dependencies.

Looking into it, I found out that this gap is caused by:

- Some other optimization flags from the release profile which were not being enabled; and are included in this PR. This reduces the time for that test from ~90s to ~50s IIRC
- The rest of the dependencies not being optimized. However this was a small gain so I did not include that (as it increases build time)
- The `lto = thin` flag not being used. However that greatly increases linking time so I did not include that.
- There is still a gap that I couldn't figure out. Even doing all the above, that particular tests took ~40s instead of ~30s in release mode. However I decided to not look into this further since I had no idea where to start.

(Sorry I should've took better notes of the timings, but the gain was significant)

## Solution

Add the other optimization flags for dependencies

### Tests

You can the code from #9741 with and without this PR changes and run `cargo test -p zebra-rpc -- coinbase --include-ignored` to measure the time

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -165,48 +165,109 @@ allow-branch = ["main"]
 [profile.dev]
 panic = "abort"
 
-# Speed up tests by optimizing performance-critical crates
+# Speed up tests by optimizing performance-critical crates. The parameters
+# values are the ones from the `release` profile
+# (https://doc.rust-lang.org/cargo/reference/profiles.html#release)
 
 # Cryptographic crates
 
 [profile.dev.package.blake2b_simd]
 opt-level = 3
+debug-assertions = false
+overflow-checks = false
+incremental = false
+codegen-units = 16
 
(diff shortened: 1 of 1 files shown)


upbqdn:
My guess is that we're missing a dependency that needs similar optimizations you did. I'll try looking into it again.

upbqdn (approved):

upbqdn (approved):
Thanks for figuring it out.

PR #9867, opened 2025-09-04
fix: sanitize_avoids_leaks test to validate sanitized MetaAddr

Fibonacci747:
The test asserted outbound validity and invariants on the original MetaAddr even after successful sanitization, which made the checks redundant and missed validating the actual sanitized output. This change switches assertions to the sanitized value and enforces that sanitized.services contains NODE_NETWORK, matching sanitize’s documented behavior.


oxarbitrage (approved):
looks good, thank you for spotting that.

PR #9871, opened 2025-09-08
fix(serialization): correct at_least_one! count-form and add unit test

sashass1315:
Replace incorrect $expr with $count in the second arm of the at_least_one! macro to match the pattern ($element:expr; $count:expr). This fixes a latent compile-time error triggered when invoking the count-form at_least_one![x; N]. Add a unit test to exercise the count-form and prevent regressions.


oxarbitrage (approved):
thanks!

PR #9874, opened 2025-09-08
Fix: constructor doc PrecomputedTransaction → CachedFfiTransaction

Galoretka:
incorrect backticked type name in CachedFfiTransaction::new doc from PrecomputedTransaction to CachedFfiTransaction. This corrects a leftover from an earlier naming and prevents confusion, since no PrecomputedTransaction type exists in the codebase and all usages reference CachedFfiTransaction.


oxarbitrage (approved):
thanks.

PR #9876, opened 2025-09-08
test: fix cargo test by ignoring tests with requirements

conradoplg:
## Motivation

`cargo test` has been failing locally due to recent refactors that made some tests with cached state requirements being run by default.

## Solution

Add `#[ignore]` to those. This should not affect CI because [it uses `--run-ignore=all`](https://github.com/ZcashFoundation/zebra/blob/c9a38a66a540acc3d12fe6876f1cf03a54cb0b52/docker/entrypoint.sh#L75).

### PR Checklist

- [ ] The solution is tested.


oxarbitrage (approved):
it works. thanks @conradoplg

PR #9878, opened 2025-09-09
chore(orchard): correct expect message for Node in arbitrary generator

Forostovec:
- Replace misleading expect text in impl Arbitrary for tree::Node from “Orchard note commitment tree root” to “Orchard note commitment tree node”.
- This change aligns the error message with the actual type (Node vs Root), preventing confusion during failing proptests and improving diagnostic accuracy.


conradoplg:
That's the `tree::Root` impl, though?

Issue #9730, opened 2025-07-25
cleanup: Return more concrete errors from the state service


arya2:
## Motivation

Currently, Zebra's state service returns a `BoxError` that could be one of many possible error types. Additionally, it uses many boxed `String` errors, which can be brittle.

We want to have more concrete error types in the state service, where callers can expect an error returned for a particular request variant to downcast to a concrete enum type.

We also want to progress towards the design implied in https://github.com/ZcashFoundation/zebra/pull/9657, where the `Service` impl for Zebra's `StateService` struct could be split into many `Service` impls, one for each request variant, that each return a concrete error type instead of a `BoxError` (note: Zebra's state service is usually used with a `Timeout` layer that would return a box error anyway, so it would still be necessary to downcast to the concrete error type, though that would still be very nice to have).

## Design

In the `StateService` and `ReadStateService`'s `Service::call()` methods, for each request variant, convert errors into an error enum that represents all possible errors that could be returned for that request variant.

For example, to return a more concrete error when handling `CommitSemanticallyVerifiedBlock` requests in the `StateService`:
- Add a `CommitSemanticallyVerifiedError` error enum with variants for every error that could occur when trying to commit a block through that request,
- Update `queue_and_commit_to_non_finalized_state()` to return a `oneshot::Receiver<Result<block::Hash, CommitSemanticallyVerifiedError>>` instead of a `oneshot::Receiver<Result<block::Hash, BoxError>>`,
- Where `RecvError`s are mapped to a boxed `String` error, return a variant of `CommitSemanticallyVerifiedError`,
- Add a code comment where `CommitSemanticallyVerifiedBlock` requests are handled in the state service's `Service::call()` method noting that errors returned by the state service when handling that request can be expected to be  `CommitSemanticallyVerifiedError`s,
- Document the expected error type in the `CommitSemanticallyVerifiedBlock` variant of the state service's `Request` type, and
- Update callers of the state service (like the block verifier) to downcast returned `BoxErrors` to a `CommitSemanticallyVerifiedError` (or the `Elapsed` error returned by the `Timeout` layer where applicable).

This should be done for every `Request` and `ReadRequest` variant handled by the `StateService` and `ReadStateService`.

As the changes will likely be very substantial, they should ideally be made for one request variant per PR.

syszery:
[...]

PR #9880, opened 2025-09-09
draft(rpc): add scaffolding for ping RPC and getpeerinfo latency metrics

### LINKED ISSUE #9048: Implement `ping` RPC
https://zcash.github.io/rpc/ping.html

syszery:
## Motivation

Refers to issue #9048 — Zebra currently lacks support for the `ping` RPC method and its associated latency metrics (`pingtime` and `pingwait`) in the `getpeerinfo` response.

This draft PR begins implementing `ping` support.

## Solution

This PR lays the foundation for `ping` support without introducing incomplete logic or placeholder hacks.

- Stub implementation of the `ping` method added to the `Rpc` trait.
  - It currently returns `Ok(())`, matching zcashd’s behavior.
- The `PeerInfo` struct (used in `getpeerinfo`) was extended with:
  - `pingtime: Option<f64>`
  - `pingwait: Option<f64>`
- `From<MetaAddr>` implementation updated to initialize these fields as `None` for now.

### Tests

- Added a new test to verify JSON serialization of `pingtime` and `pingwait` with `Some` and `None` values.
- All existing tests in `zebra-rpc` build and pass.

- [Zcash RPC Docs: `ping`](https://zcash.github.io/rpc/ping.html)
- Issue: #9048

### Follow-up Work

- [ ] Trigger sending of `ping` message in the `ping` RPC method.
- [ ] Implement a round-trip timer to measure latency.
- [ ] Populate `pingtime` and `pingwait` in the `PeerInfo` response.

### Feedback Requested

My idea is to use the existing logic in the `zebra-network` crate, specifically in `connection.rs`, to implement `ping` behavior:

- Send a `Message::Ping(nonce)` from [`message.rs` line 40](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/protocol/external/message.rs#L40).
- Trigger sending similarly to [`connection.rs` line 1039](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/peer/connection.rs#L1039).
- Handle ping responses as done in [`connection.rs` line 1180](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/peer/connection.rs#L1180).

Does this look like the right place to hook into for ping tracking, or would another approach be preferred?

>Note: 
The `zebra-network` crate already defines a `Request::Ping(Nonce)` variant (see ([`request.rs` line 48](https://github.com/ZcashFoundation/zebra/blob/main/zebra-network/src/protocol/internal/request.rs#L48)). However, this variant is marked as internal and described as a "bit of a hack." As such, I did not consider using it here.

### Acknowledgements

Thanks @oxarbitrage for outlining the relative complexity of this task in #9730 and suggesting it might be better to tackle later. That guidance was very helpful and is much appreciated.

I had already started this work over the weekend, so I wanted to clean up the scaffold and share what I learned. This draft PR can serve as a starting point for further discussion or future implementation. I plan to return to this at the end of September after my vacation.

Let me know if this direction makes sense and how best to proceed!


conradoplg (dismissed):
This is looking good, feel free to move out of draft or to build on top of it, thanks!

syszery:
Hi @conradoplg, thanks for the early feedback! I've just rebased the branch onto the latest `main` and fixed all Clippy warnings. I also marked the PR as 'ready for review', but I do plan to continue working on it — just wanted to make sure it’s unblocked in case this is considered high priority, since it’s part of the [Milestone: Replace zcashd dependencies](https://github.com/ZcashFoundation/zebra/milestone/41).

syszery:
I've implemented rudimentary tracking of peer latency in `connection.rs`, using the existing `Request::Ping(Nonce)` variant to send pings and measure RTT via pong responses.

I'd like to hook this into the RPC layer next, and I'm looking for feedback on two points:

1. Does this usage align with the intended purpose of `Request::Ping(Nonce)`?
2. What's the recommended way to expose peer latency to the RPC crate?

Currently, RTT data is stored in `Connection` via a `PeerLatency` struct. The `getpeerinfo` RPC pulls data from the `AddressBook`, but it doesn't seem like the right place for live metrics like latency. Would a different approach be preferred?

Happy to adjust direction based on feedback. Thanks!

zebra-network/src/peer/connection.rs
```diff
@@ -601,6 +603,11 @@ where
     /// service to a request from this connection,
     /// or None if this connection hasn't yet received an overload error.
     last_overload_time: Option<Instant>,
+
+    /// Tracks latency info for `ping`/`pong` messages with the peer.
+    ///
+    /// Used to measure round-trip time and manage ping state.
+    ping_metrics: PeerLatency,
```
conradoplg: It seems cleaner to add the required data to `Handler::Ping`, and just leave the `last_ping_rtt` here. See also other comments.

zebra-network/src/peer/connection.rs
```diff
@@ -1034,11 +1042,17 @@ where
...
+            (AwaitingRequest, Ping(nonce)) => {
+                self.ping_metrics.ping_sent_at = Some(std::time::Instant::now());
+                self.ping_metrics.is_waiting_for_pong = true;
+                self.ping_metrics.last_ping_rtt = None;
+                self.ping_metrics.waiting_nonce = Some(nonce);
+                self
+                    .peer_tx
+                    .send(Message::Ping(nonce))
+                    .await
+                    .map(|()| Handler::Ping(nonce))
```
conradoplg: Following my suggestion above, you would include the required data here (I think you would only need to add `ping_sent_at`)

zebra-network/src/peer/connection.rs
```diff
@@ -1205,9 +1219,32 @@ where
...
+                    }
+                    (true, _, _) => {
+                        debug!(%msg, "received pong with no matching ping in flight");
+                    }
+                    (false, _, _) => {
+                        debug!(%msg, "received unsolicited pong with no ping in flight");
+                    }
+                }
+
+                Consumed
```
conradoplg: This is the wrong branch, for when the Pong arrives unsolicited. You need to handle it in `process_message()` above; you will have the existing `Handler::Ping` with the nonce (and the `ping_sent_at` I suggested you to add) and can do the whole thing there.

conradoplg (changes requested):
Some adjustments are required but hopefully it should be simple, see comments.

Regarding propagating the measured latency, I think it makes sense to use the address book. You will need to propagate the result as if `Pong` were returning something:

- [In `process_message()`, after handling the `Pong`](https://github.com/ZcashFoundation/zebra/blob/64cad2b9eb62a99f28706dda6a52daaf851c2ec7/zebra-network/src/peer/connection.rs#L154), change `Handler::Finished(Ok(Response::Nil))` to return a new response type that you will need to create which includes the latency.
- If you do that then in `send_periodic_heartbeats_run_loop()`, [the `heartbeat_timeout()` should return the new Pong response](https://github.com/ZcashFoundation/zebra/blob/64cad2b9eb62a99f28706dda6a52daaf851c2ec7/zebra-network/src/peer/handshake.rs#L1316). Extract the latency from it and then pass it to the `heartbest_ts_collector` inside the `MetaAddr::new_responded` which you will need to modify to pass the latency. Then check where that is handled and store the latency in the address book.

I'm not 100% sure that will work but should you give some hints. As always, let us know if you prefer for us to work on it. Thank you for all the help!

syszery:
Cleaned up artifacts from the earlier commits and addressed all reviewer feedback related to the first point:

> In `process_message()`, after handling the `Pong`, change `Handler::Finished(Ok(Response::Nil)`) to return a new response type that you will need to create which includes the latency.

Ping-pong latency tracking is now wired inside the handler, and a test was added for the round-trip. This is meant as an intermediate step.

Thank you @conradoplg for stepping in and pointing me in the right direction with such detailed feedback!

syszery:
Addressed the second part of the review feedback:

> *"If you do that then in `send_periodic_heartbeats_run_loop()`, the `heartbeat_timeout()` should return the new `Pong` response. Extract the latency from it and then pass it to the `heartbeat_ts_collector` inside the `MetaAddr::new_responded`, which you will need to modify to pass the latency."*

**Next steps:**

> *"Then check where that is handled and store the latency in the address book."*

syszery:
> Some adjustments are required but hopefully it should be simple, see comments.
> 
> Regarding propagating the measured latency, I think it makes sense to use the address book. You will need to propagate the result as if `Pong` were returning something:
> 
> * [In `process_message()`, after handling the `Pong`](https://github.com/ZcashFoundation/zebra/blob/64cad2b9eb62a99f28706dda6a52daaf851c2ec7/zebra-network/src/peer/connection.rs#L154), change `Handler::Finished(Ok(Response::Nil))` to return a new response type that you will need to create which includes the latency.
> * If you do that then in `send_periodic_heartbeats_run_loop()`, [the `heartbeat_timeout()` should return the new Pong response](https://github.com/ZcashFoundation/zebra/blob/64cad2b9eb62a99f28706dda6a52daaf851c2ec7/zebra-network/src/peer/handshake.rs#L1316). Extract the latency from it and then pass it to the `heartbest_ts_collector` inside the `MetaAddr::new_responded` which you will need to modify to pass the latency. Then check where that is handled and store the latency in the address book.
> 
> I'm not 100% sure that will work but should you give some hints. As always, let us know if you prefer for us to work on it. Thank you for all the help!

The last commit now fully implements the suggested steps. I had to adapt several tests and decided to use some default values. Based on feedback, I can change these to more meaningful latencies.

Next steps:
- Implement the `ping` RPC method to trigger a ping to all peers.
- Track and update `pingwait` for each peer in the address book.

conradoplg:
It's looking great, thanks

syszery:
Hi @conradoplg, I am sorry to bother you again, but I got stuck here. I struggle to find a way how to send the ping message to all connected peers.


**What I found so far**:
* In [zebra-network/src/peer_set/set.rs](https://github.com/ZcashFoundation/zebra/blob/1883c15eef32a6d9d055cec76f68f47ce325b4f6/zebra-network/src/peer_set/set.rs#L1311), the `PeerSet` handles `Service::call`. I added a `Request::Ping(_) => self.broadcast_all(req)` arm, and verified it reaches mocked peers in a simple test.
* In RPC, methods like `add_node` and `get_peer_info` use the `AddressBookPeers` trait, but that’s only for peer metadata. 
* On the other hand, `send_raw_transaction` directly calls into `mempool::Request` via the `RpcImpl`.

**My question**:
* Is this `PeerSet::call(Request::Ping)` even the right way to broadcast pings? 
* If yes, how should the RPC crate trigger it? Should I expose it via a new trait (similar to `AddressBookPeers`) — or is it better to pass the `PeerSet` directly into `RpcImpl`, like how the mempool is handled?


I’d really appreciate any tips or guidance!

arya2:
> I struggle to find a way how to send the ping message to all connected peers.

I'm not sure if the method exists in zcashd alongside a heartbeat task, or if it was required because zcashd doesn't use a heartbeat task. If it's the latter, returning the `pingwait` field on the `getpeerinfo`'s response type could be enough to close the issue, so this PR could merge as-is without it.

@str4d Do you know if a `ping` RPC method will be required in Zebra or if it's okay to always record the ping from the heartbeat task?

> **My question**:
> 
> * Is this `PeerSet::call(Request::Ping)` even the right way to broadcast pings?

Not currently, as it would only send the `Ping` request to a fraction of peers. Adding a new request variant similar to `AdvertiseBlockToAll` and a new method similar to `broadcast_all()` to measure the time between sending the request and getting a response in the peer set could work.

> * If yes, how should the RPC crate trigger it? Should I expose it via a new trait (similar to `AddressBookPeers`) — or is it better to pass the `PeerSet` directly into `RpcImpl`, like how the mempool is handled?

The peer set implements `tower::Service<zebra_network::Request>` already, so a field of any type which implements that would work. No new trait is required, `AddressBookPeers` was an unfortunate addition because the address book isn't a service.

syszery:
Thanks @arya2 for the explanation and for steering me away from chasing the wrong “solution”!

I understand that the best next step is to fully implement support for the missing `pingwait` field, and then see whether Zebra needs a `ping` RPC method at all (or if the heartbeat logic is already sufficient).

conradoplg:
This is looking good. I agree with @arya2 that we can leave the actual force-pinging-everyone for later (for the time we could implement a dummy ping RPC that does nothing).

Also feel free to do `pingwait` in a separate PR if you prefer, and we can get this merged (probably after the next release though)

syszery:
Hi @conradoplg, thank you for your feedback! I’ve extended the PR and added a proposal for the missing `pingwait`:

* Added `ping_sent_at` to `MetaAddr`
* Updated `MetaAddrChange::new_responded` to include `ping_sent_at`
* Implemented basic `pingwait` calculation

Happy to adjust based on your feedback.

zebra-network/src/peer/handshake.rs
```diff
@@ -1312,8 +1313,10 @@ async fn send_periodic_heartbeats_run_loop(
     while let Some(_instant) = interval_stream.next().await {
         // We've reached another heartbeat interval without
         // shutting down, so do a heartbeat request.
+        let ping_sent_at = Instant::now();
+
         let heartbeat = send_one_heartbeat(&mut server_tx);
-        heartbeat_timeout(heartbeat, &heartbeat_ts_collector, &connected_addr).await?;
+        let rtt = heartbeat_timeout(heartbeat, &heartbeat_ts_collector, &connected_addr).await?;
```
conradoplg, line 1324: There is an issue here: this only returns _after_ the pong was received, so the `ping_sent_at` is only written to the address book at that time, but then it's too late (its whole purpose is to keep track how much time has elapsed since the last ping was sent).

I'm not 100% sure how to fix this, but it might envolve sending a separate `MetaAddrChange` to the `heartbeat_ts_collector` before sending the ping here. I think we'll need a new `MetaAddrChange` type because I don't think the current ones would fit.

Aditionally, after the pong is received, we will need to actually clear the `ping_sent_at` value, so the `send()` below should send `None`.

zebra-network/src/peer/handshake.rs
```diff
@@ -1322,11 +1325,25 @@ async fn send_periodic_heartbeats_run_loop(
...
+                    .await;
+            } else {
+                let _ = heartbeat_ts_collector
+                    .send(MetaAddr::new_responded(
+                        book_addr,
+                        None,
+                        Some(ping_sent_at.into()),
+                    ))
+                    .await;
+            }
```
conradoplg, line 1339: If I'm not mistaken the branches are basically the same? i.e. just pass the original `rtt`

zebra-network/src/meta_addr.rs
```diff
@@ -450,6 +471,20 @@ impl MetaAddr {
...
+    /// Returns the round-trip time (RTT) for this peer, if available,
+    /// as the number of seconds (f64). This value is updated when the peer
+    /// responds to a ping (Pong).
+    pub fn rtt(&self) -> Option<f64> {
+        self.rtt.map(|duration| duration.as_secs_f64())
+    }
+
+    /// Returns the time this peer was last pinged, if available,
+    /// as the number of seconds since the epoch or some reference point
+    pub fn ping_sent_at(&self) -> Option<f64> {
```
conradoplg: This name is a bit confusing since it does not return the original `ping_sent_at` but the elapsed time. So maybe change it to `ping_wait`? Also update the docs to indicate the purpose, i.e. `the number of seconds since the last ping was sent`

Alternatively, change the method to just return the `ping_sent_at` and compute the elapse time inside the RPC code. I don't have a specific preference. I think the latter looks more organized, and I think you could change this and the `rtt` method to return the original values (i.e. don't convert to f64 here) and instead convert to f64 in the RPC code (e.g. in `impl From<MetaAddr> for PeerInfo`)

conradoplg (changes requested):
Thanks for the recent changes! They look good but there is a subtle issue. There is no obvious way to fix, I added some suggestions, please let me know if you need more feedback on how to address it.

syszery:
Hi @conradoplg, thank you for the detailed feedback and for pointing out the bug! I've tried to implement all your suggestions. Here is a detailed list of the recent changes:

* Added a new `MetaAddrChange::UpdatePingSent` variant and send it **before** dispatching the heartbeat.
* Removed redundant `if/else` branches in heartbeat handling. When a peer responds (`UpdateResponded`), `ping_sent_at` is set to `None`.
* `MetaAddr::rtt()` and `MetaAddr::ping_sent_at()` now return `Duration` and `Instant`. The RPC layer computes the elapsed seconds for `pingtime`/`pingwait`.

One question I encountered: the method `peer_addr_state` maps `UpdatePingSent` to `Responded`. I think this is not ideal, but currently there is no state that accurately describes this situation. I added a comment explaining the reasoning. Do you have a different preference?

Happy to adjust based on feedback!

zebra-network/src/meta_addr.rs
syszery: `MetaAddr::ping_sent_at()` and `MetaAddr::rtt()` now return the original `Instant`/`Duration`, and the RPC layer computes the elapsed seconds for `pingwait` and `pingtime`.

zebra-network/src/peer/handshake.rs
syszery: Removed redundant `else`; when a peer responds, `ping_sent_at` is set to `None`.

zebra-network/src/peer/handshake.rs
syszery: Added a new `MetaAddrChange::UpdatePingSent` and send it before the heartbeat.

zebra-network/src/meta_addr.rs
```diff
@@ -823,6 +879,31 @@ impl MetaAddrChange {
...
+        match self {
+            UpdateResponded { rtt, .. } => *rtt,
+            _ => None,
+        }
+    }
+
+    /// Returns the timestamp when a ping was last sent, if available.
+    pub fn ping_sent_at(&self) -> Option<Instant> {
+        match self {
+            UpdateResponded { ping_sent_at, .. } => *ping_sent_at,
```
conradoplg:
```suggestion
            UpdatePingSent { ping_sent_at, .. } => Some(*ping_sent_at),
```

conradoplg (changes requested):
There is one small bug, but with the suggested fix applied I confirmed it works. Thanks!

> One question I encountered: the method peer_addr_state maps UpdatePingSent to Responded. I think this is not ideal, but currently there is no state that accurately describes this situation. I added a comment explaining the reasoning. Do you have a different preference?

Your solution is fine!

There is one lingering small cleanup that could be done; now `UpdateResponded` doesn't really need `ping_sent_at` anymore since we always use it to pass None and clear the value. Up to you though, since it works anyway.

zebra-network/src/meta_addr.rs
syszery: Thanks for catching the bug! I’ll apply the suggested fix, and I’ll follow up with the cleanup you mentioned as well.

syszery:
Since the tests are running now, I don't want to interrupt them by committing the cleanup. I can push it later or just append the minor changes in another PR.

conradoplg (approved):

PR #9883, opened 2025-09-10
 refactor(ci)!: simplify CI and make it work on forks

gustavovalverde:
## Motivation

We've been working on simplifying our CI to make it actually usable by contributors. The main problem was that when someone forked the repo, they couldn't run most of the CI because it depended on Firebase, external Docker registries, and other services they didn't have access to.

Also, we had tons of patch files that made everything harder to maintain and understand.

## Solution

Basically, we cleaned up a lot:

1. **Got rid of patch files**: Deleted all 19 of them and merged the workflows that made sense together. This removed about 2,900 lines of config.

2. **No more external services**:
   - Docs now deploy to GitHub Pages instead of Firebase
   - Removed the DockerHub description workflow we weren't really using
   - Everything runs on GitHub now, so forks work out of the box
   - All that has external dependencies is `zfnd-` prefix so it can be easily disabled

3. **Better naming**:
   - Added `zfnd-` prefix to our internal workflows so it's clear what they're for
   - Renamed test workflows to actually say what they test (`test-crates.yml`, `test-docker.yml`, etc.)

### Tests

- We need to run everything on this branch - workflows run, docs build, tests. 
- Also checked that it works on a fork.

### Additional Improvements

1. **Security fixes**:
   - Added minimal top-level permissions to all ZFND workflows
   - Fixed secret inheritance - now uses explicit secret passing instead of `secrets: inherit`
   - Only DockerHub secrets passed where actually needed
   - Made all workflows zizmor-compliant (security scanner)

2. **Dependency cleanup**:
   - Removed `tinyvec` from 7 crates (unused transitive dependency)
   - Removed `jsonrpc` from zebra-utils (unused optional dependency)
   - Verified and kept `console-subscriber` (behind tokio-console feature)
   - Kept `zebra-utils` in zebrad (needed for zebra-checkpoints feature)

3. **Test & Coverage improvements**:
   - Fixed coverage workflow to use nextest properly
   - Added dynamic timeouts for tests running under coverage instrumentation
   - Improved Windows test compatibility

### Follow-up Work

- Keep an eye on how fast the workflows run after merging
- Maybe optimize caching later
- Update Docs explaining the new CI structure

### PR Checklist

- [ ] The solution is tested.


.github/workflows/book.yml
```diff
@@ -0,0 +1,121 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+          mdbook-version: ~0.4
+          use-linkcheck: true
+          use-mermaid: true
+
+      - name: Install last version of Protoc
+        uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
```
conradoplg: Protoc is no longer required for building so it can be removed here (we still should keep it on the main workflow because with protoc we can check that the autogenerated files are up to date)

```suggestion

```

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,84 @@
...
+          toolchain: stable
+          components: llvm-tools-preview
+          cache-on-failure: true
+      - uses: taiki-e/install-action@3ee5d63d29478156148c0b53e9f3447829b47bc2 #v2.58.23
+        with:
+          tool: cargo-llvm-cov,nextest
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
```
conradoplg: Same as above

```suggestion
```

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Run coverage tests
+        run: |
+          cargo llvm-cov --no-report nextest
+          # TODO: Do we need --locked --release --features "default-release-binaries" here?
```
conradoplg, line 55: `--locked` and `--features` seem like a good idea. I feel like `--release` might interfere with coverage computation though the docs don't say anything about it... Also if we add the flags, I think it should be in the command above (`cargo llvm-cov`), not below (`cargo llvm-cov report`)

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+        with:
+          persist-credentials: false
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          components: clippy
+          toolchain: ${{ matrix.rust-version }}
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg, line 70: Same as above

```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+          persist-credentials: false
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          cache-on-failure: true
+      - uses: taiki-e/install-action@3ee5d63d29478156148c0b53e9f3447829b47bc2 #v2.58.23
+        with:
+          tool: cargo-hack
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg, line 88: Same as above

```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+        with:
+          persist-credentials: false
+
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          toolchain: 1.89 # MSRV
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg:
```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+        with:
+          persist-credentials: false
+
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          toolchain: nightly
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg: Same as above

```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
+        with:
+          persist-credentials: false
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          toolchain: nightly
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg:
```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+        with:
+          persist-credentials: false
+
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          toolchain: nightly
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
conradoplg:
```suggestion
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+      - name: cargo install cargo-hack
+        uses: taiki-e/install-action@3ee5d63d29478156148c0b53e9f3447829b47bc2 #v2.58.23
+        with:
+          tool: cargo-hack
+      - run: cargo hack check --all
```
conradoplg, line 223: I don't think `cargo check` gives us anything that `cargo clippy` doesn't, so I feel like we can remove it. Also this seems redundant with the other job above that calls `cargo hack check`?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          cache-on-failure: true
+      - uses: taiki-e/install-action@3ee5d63d29478156148c0b53e9f3447829b47bc2 #v2.58.23
+        with:
+          tool: cargo-hack
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - run: cargo hack check --workspace
```
conradoplg, line 92: I don't think `cargo check` gives us anything that `cargo clippy` doesn't, so I feel like we can remove it

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+        with:
+          persist-credentials: false
+      - uses: actions-rust-lang/setup-rust-toolchain@ab6845274e2ff01cd4462007e1a9d9df1ab49f42 #v1.14.0
+        with:
+          toolchain: stable
+          cache-on-failure: true
+      - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b #v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - run: cargo check --locked --all-features --all-targets
```
conradoplg, line 244: Again, no need for `cargo check`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,302 @@
...
+      matrix:
+        rust-version: [stable, beta]
+        type: [release, tests]
+        include:
+          - type: release
+            args: --workspace --all-targets
+            features: default-release-binaries
+          - type: tests
+            args: --workspace --all-targets
+            features: default-release-binaries proptest-impl lightwalletd-grpc-tests zebra-checkpoints
```
conradoplg, line 57: I don't think the `features` parameter is useful here. Features should be additive so testing one feature and then the same feature with others should be the same as testing the latter. Currently we just use `--all-features` and I feel like we should do the same here.

Regarding testing with `beta`, I think it would gain us anything as it stands. Ideally we could make the beta clippy just give warnings, and the stable clippy fail on error. This way, when a new beta is released, we become aware of any clippy changes but PRs don't start to fail, and we have time to fix them before the new rules get into stable.

This might be too much too handle in this PR, so if you prefer to handle this later, then I think we can just use stable for now.

.github/workflows/test-crates.yml
```diff
@@ -1,81 +1,73 @@
...
-  COLORBT_SHOW_HIDDEN: ${{ vars.COLORBT_SHOW_HIDDEN }}
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  RUST_LIB_BACKTRACE: 1
+  RUST_LOG: info
+  COLORBT_SHOW_HIDDEN: 1
 
 jobs:
   matrix:
     name: Generate crates matrix
```
conradoplg, line 46: I think this job could be greatly simplified by using `cargo hack build --workspace` (instead of manually building a crates matrix as it does currently) since it will build each crate separately if I understood it correctly. However feel free to do that in a separate PR if you prefer.

.github/workflows/tests-unit.yml
```diff
@@ -0,0 +1,136 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        rust-version: [stable, beta]
+        features: [default-release-binaries]
+        exclude:
+          # Exclude macOS beta due to limited runner capacity and slower performance
+          - os: macos-latest
+            rust-version: beta
```
conradoplg: I think we could also exclude windows+beta to save time

conradoplg (commented):
This is looking great, thanks!

I added some comments, I don't think any are blockers, but I still want to do another round of review.

.github/workflows/lint.yml
gustavovalverde: I think `cargo hack check --workspace` is worth keeping. In `cargo clippy --workspace --all-targets` we build the whole workspace as one graph for the two feature sets in the matrix, and when the workspace builds as a unit, crates can "borrow" features or dependencies that are only enabled because another member pulled them in, so missing `Cargo.toml` entries or bad defaults can slip by.

`cargo hack --workspace` **deliberately** changes the behavior of the flag so it runs cargo check once per member, just as if we had iterated `cargo check --manifest-path <member>/Cargo.toml` by hand. That can catch problems where an individual crate fails to compile on its own (for example for `cargo publish`), but clippy would still succeed.

.github/workflows/lint.yml
gustavovalverde: Here the difference is that with Clippy we never pass the `--locked` flag (_should we?_), and the separate `cargo check` runs once with `--locked --all-features --all-targets` which:
- Fails fast if `Cargo.lock` is out of sync or would need to be regenerated, and 
- Guarantees the crate builds when every feature is enabled simultaneously (this is skipped in the Clippy runs).

.github/workflows/lint.yml
gustavovalverde: A few things here to consider:

`internal-miner` is off in real builds, and a whole branch is guarded with `#[cfg(not(feature = "internal-miner"))]` and `--all-features` drops that branch, so any lint issues would go unseen.

Also the helpers we call in production are hidden behind `#[cfg(not(any(test, feature = "proptest-impl")))]`, so they only compile when `proptest-impl` is disabled. The `indexer` feature even swaps out core type aliases when it’s off, so those default definitions vanish as soon as you enable every feature.

We already have other jobs covering the `--all-features` build (`check-cargo-lock`, `nightly features`), so the lint matrix is what keeps the default configurations linted, but we can also consider `--all-features` here.

I don't have any hard opinions here, so I'll proceed as you see fit.

.github/workflows/test-crates.yml
gustavovalverde: I do agree. We've had it this way because it fail-fast and we have a better UX in the GitHub console, but we can discuss if this is not needed to simplify it.
image

.github/workflows/tests-unit.yml
gustavovalverde: Done [chore: remove extra windows run](https://github.com/ZcashFoundation/zebra/pull/9883/commits/fbcbd422f599fce584291b65b514b2b499dc382f)

.github/workflows/lint.yml
conradoplg: Ah yes, that makes sense. I was going to suggest that this should be `cargo hack clippy`, because `clippy` is basically a superset of `check`, but I think it would be superfluous since we already run `cargo clippy` elsewhere and the additional things that clippy catches shouldn't be affected by this compile everything / compile each crate issue.

(i.e. no need to change anything)

.github/workflows/lint.yml
conradoplg: Ah yes, that makes sense. We should use `--locked` when building zebra but not use it when building the crates individually, so it's fine as is

.github/workflows/lint.yml
conradoplg: Yeah I understand your reasoning. It's fine to keep this as is but it would be good to document the reasoning.

The "proper" way to do it would be to test every feature combination, but that gets unwieldy fast (it caused to CI to run out of space in the `frost` repo, which is much smaller). We can give it a shot in the future in a separate PR (using https://crates.io/crates/cargo-all-features and the `chunks` argument which I didn't try)

.github/workflows/test-crates.yml
conradoplg: Honestly I think the list of tests it way too big and decreasing it would be good. To find which crate failed we could simply check the logs. But I don't mind either way. Maybe someone else can chime in? What do you think @arya2 and/or @oxarbitrage ?

conradoplg (approved):
We can merge this, thanks! I'll leave the remaining comments up to you.

This however doesn't test the non-fork flow, what do you want to prefer? Get this merged and revert it if something breaks, or open a separate non-fork PR to test the non-fork flow?

gustavovalverde:
> This however doesn't test the non-fork flow, what do you want to prefer? Get this merged and revert it if something breaks, or open a separate non-fork PR to test the non-fork flow?

@conradoplg  I'd prefer to get this merged, I can fix anything through the weekend if something actually breaks. I wouldn't expect that considering the other non-fork workflows were just renamed—their logic keeps basically the same, but anything can happen :)

gustavovalverde:
By the way, I'll have to admin-merge this because I'll also have to update the repo rulesets. Wish me luck 🤞🏻

.github/workflows/test-crates.yml
arya2: I don't mind either way, but would slightly prefer to have one job as I agree the list of tests it too big.

PR #9884, opened 2025-09-10
feat(rpc): support side chains in getrawtransaction

### LINKED ISSUE #9873: feature: support sidechains in getrawtransaction
### Motivation

`getrawtransaction` should support returning txs from side chains.

This will also require supporting negative heights in the return value, since height -1 is used for those.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

conradoplg:
## Motivation

Closes #9873

## Solution

Add new state methods to find blocks and transactions in any chains.
Change `getrawtransaction` to use those.

I also changed the `invalidateblock` and `reconsiderblock`, which are used in the test, because they weren't working in JSON-RPC.

### Tests

I added a Python test for this, since it seemed easier. I think those aren't being run in CI yet, but we can do that later.

```
PYTHON_DEBUG=1 CARGO_BIN_EXE_zebrad=~/zebra/target/debug/zebrad PYTHON_DEBUG=1 ./qa/rpc-tests/getrawtransaction_sidechain.py
```

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
+++ b/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# Copyright (c) 2022-2024 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#
+# Test the effect of reorgs on the Orchard commitment tree.
+#
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    BLOSSOM_BRANCH_ID,
+    HEARTWOOD_BRANCH_ID,
+    CANOPY_BRANCH_ID,
+    NU5_BRANCH_ID,
+    assert_equal,
+    get_coinbase_address,
+    nuparams,
+    start_nodes,
+    wait_and_assert_operationid_status,
+)
+
+
+class GetRawTransactionSideChainTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.cache_behavior = 'clean'
+
+    def setup_nodes(self):
+        return start_nodes(self.num_nodes, self.options.tmpdir)
+
+    def setup_network(self, split=False, do_mempool_sync=True):
+        self.nodes = self.setup_nodes()
+
+    def run_test(self):
+        n = self.nodes[0]
+        # Generate two blocks
+        n.generate(2)
+
+        # Get a transaction from block at height 1
+        block_hash_a = n.getbestblockhash()
+        txid_a = n.getblock("-1", 1)['tx'][0]
+
+        # Invalidate last block (height 1)
+        n.invalidateblock(block_hash_a)
+        # Regenerate a height 1 block
+        n.generate(1)
+        # Get a transaction from the new block at height 1
+        txid_b = n.getblock("-1", 1)['tx'][0]
+        # Reconsider the invalidated block
+        n.reconsiderblock(block_hash_a)
+
+        # We now have two chains. Try to get transactions from both.
+        tx_a = n.getrawtransaction(txid_a, 1)
+        tx_b = n.getrawtransaction(txid_b, 1)
+        assert(tx_a['txid'] != tx_b['txid'])
+
+
+if __name__ == '__main__':
+    GetRawTransactionSideChainTest().main()
--- a/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
+++ b/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
@@ -8,7 +8,8 @@
 #
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import start_nodes
+from test_framework.util import start_nodes, assert_equal
+from test_framework.proxy import JSONRPCException
 
 
 class GetRawTransactionSideChainTest(BitcoinTestFramework):
@@ -25,26 +26,45 @@ class GetRawTransactionSideChainTest(BitcoinTestFramework):
 
     def run_test(self):
         n = self.nodes[0]
+
         # Generate two blocks
-        n.generate(2)
+        hashes = n.generate(2)
 
-        # Get a transaction from block at height 1
+        # Get a transaction from block at height 2
         block_hash_a = n.getbestblockhash()
-        txid_a = n.getblock("-1", 1)['tx'][0]
+        txid_a = n.getblock("2", 1)['tx'][0]
+        tx_a = n.getrawtransaction(txid_a, 1)
+        assert_equal(tx_a['height'], 2)
 
-        # Invalidate last block (height 1)
+        # Invalidate last block (height 2)
         n.invalidateblock(block_hash_a)
-        # Regenerate a height 1 block
-        n.generate(1)
-        # Get a transaction from the new block at height 1
-        txid_b = n.getblock("-1", 1)['tx'][0]
+        try:
+            tx_a = n.getrawtransaction(txid_a, 1)
+            assert False, "getrawtransaction should have failed"
+        except JSONRPCException:
+            pass
+
+        # Regenerate a height 2 block
+        hashes = n.generate(1)
+        txid_b = n.getblock(hashes[0], 1)['tx'][0]
+
+        # Get a transaction from the new block at height 2
+        tx_b = n.getrawtransaction(txid_b, 1)
+        assert_equal(tx_b['height'], 2)
+
         # Reconsider the invalidated block
         n.reconsiderblock(block_hash_a)
 
         # We now have two chains. Try to get transactions from both.
         tx_a = n.getrawtransaction(txid_a, 1)
         tx_b = n.getrawtransaction(txid_b, 1)
-        assert(tx_a['txid'] != tx_b['txid'])
+        assert(tx_a['blockhash'] != tx_b['blockhash'])
+        # Exactly one of the transactions should be in a side chain
+        assert((tx_a['confirmations'] == 0) ^ (tx_b['confirmations'] == 0))
+        assert((tx_a['height'] == -1) ^ (tx_b['height'] == -1))
+        # Exactly one of the transactions should be in the main chain
+        assert((tx_a['height'] == 2) ^ (tx_b['height'] == 2))
+        assert((tx_a['confirmations'] == 1) ^ (tx_b['confirmations'] == 1))
 
 
 if __name__ == '__main__':
--- a/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
+++ b/zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
@@ -1,24 +1,14 @@
 #!/usr/bin/env python3
-# Copyright (c) 2022-2024 The Zcash developers
+# Copyright (c) 2025 The Zcash developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or https://www.opensource.org/licenses/mit-license.php .
 
 #
-# Test the effect of reorgs on the Orchard commitment tree.
+# Test getrawtransaction on side chains
 #
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import (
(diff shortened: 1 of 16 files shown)


zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -158,13 +158,14 @@ pub struct TransactionObject {
     /// The raw transaction, encoded as hex bytes.
     #[serde(with = "hex")]
     pub(crate) hex: SerializedTransaction,
-    /// The height of the block in the best chain that contains the tx or `None` if the tx is in
-    /// the mempool.
+    /// The height of the block in the best chain that contains the tx, -1 is
```
upbqdn:
```suggestion
    /// The height of the block in the best chain that contains the tx, -1 if
```

zebra-rpc/src/methods.rs
```diff
@@ -637,15 +639,15 @@ pub trait Rpc {
     /// - `block_hash`: (hex-encoded block hash, required) The block hash to invalidate.
     // TODO: Invalidate block hashes even if they're not present in the non-finalized state (#9553).
     #[method(name = "invalidateblock")]
-    async fn invalidate_block(&self, block_hash: block::Hash) -> Result<()>;
+    async fn invalidate_block(&self, block_hash: String) -> Result<()>;
```
upbqdn, line 652: Was it the error code that was problematic?

conradoplg: `block::Hash` does not deserialize from a hex string by default. It was being deserialized from an integer array. (We could change the `block::Hash` serde serialization format but that might break something else... We can probably review all serde implementations in https://github.com/ZcashFoundation/zebra/issues/9892)

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -625,8 +626,16 @@ impl TransactionObject {
...
+            height: if in_active_chain.unwrap_or_default() {
+                height.map(|height| height.0 as i32)
+            } else {
+                Some(-1)
+            },
+            confirmations: if in_active_chain.unwrap_or_default() {
+                confirmations
+            } else {
+                Some(0)
+            },
```
upbqdn, line 645: We set `in_active_chain` to `Some(false)` here: https://github.com/zcashfoundation/zebra/blob/97a805874bb6c61d688ccbb1709ba9bf10b899b8/zebra-rpc/src/methods.rs#L1757 when the tx comes from the mempool, so this logic returns `Some(-1)` for `height` and `Some(0)` for `confirmations`, but the updated docs above say that the values should be `None` and `None`.

upbqdn: I updated the link in the comment above because I noticed the original one didn't point to the exact line I intended to reference.

zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
```diff
@@ -0,0 +1,61 @@
...
+    BLOSSOM_BRANCH_ID,
+    HEARTWOOD_BRANCH_ID,
+    CANOPY_BRANCH_ID,
+    NU5_BRANCH_ID,
+    assert_equal,
+    get_coinbase_address,
+    nuparams,
+    start_nodes,
+    wait_and_assert_operationid_status,
+)
```
oxarbitrage:
```suggestion
from test_framework.util import start_nodes
```

oxarbitrage: Lint: From util imports, the test just use `start_nodes`.

zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+# Copyright (c) 2022-2024 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#
+# Test the effect of reorgs on the Orchard commitment tree.
```
oxarbitrage:
```suggestion
# Test getrawtransaction on side chains
```

zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+# Copyright (c) 2022-2024 The Zcash developers
```
oxarbitrage:
```suggestion
# Copyright (c) 2025 The Zcash developers
```

Optional: I am not sure how this is going to be in the future, but for a new file not ported from zcashd what i had been  doing is: https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py#L2

oxarbitrage (commented):
I added some comments that i noticed in the python framework only.

zebra-rpc/src/methods/types/transaction.rs
conradoplg: Thanks, nice catch. This led me into a big rabbit hole that made me realize the test wasn't working properly. I addressed this and the other stuff in 58fc1a2d2

conradoplg (commented):
While addressing @upbqdn 's comment I found multiple issues which were all fixed in 58fc1a2d2

- The `generate()` after the `reconsiderblock()` was failing but an error wasn't being returned, so the test kept going and with the way it was written, it passed but it wasn't actually looking into blocks across forks. I changed `generate()` to propagate any error from `submit_block()`
- The reason that was failing was a check that would prevent committing a block with a height of a previously invalidated block. That doesn't seem to make sense to me, so I changed the check to return an error only if the hash is the same of a previously invalidated block
- After fixing that, there was another issue: after `invalidateblock()`, `generate()` would generate a block exactly identical to the invalidated one, which would then fail to commit. To fix that, I changed `generate()` to use random coinbase data for each block, which makes the coinbase tx unique and its block too.

conradoplg:
@mergifyio update

zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def run_test(self):
+        n = self.nodes[0]
+
+        # Generate two blocks
+        hashes = n.generate(2)
+
+        # Get a transaction from block at height 2
+        block_hash_a = n.getbestblockhash()
+        txid_a = n.getblock("2", 1)['tx'][0]
```
upbqdn, line 35: It might be worth either obtaining the block and its hash from a single call or asserting that the hash corresponds to the fetched block.

zebra-rpc/src/methods/types/get_block_template.rs
```diff
@@ -519,6 +518,17 @@ where
         self.extra_coinbase_data.clone()
     }
 
+    /// Changes the extra coinbase data.
```
upbqdn, line 521:
```suggestion
    /// Changes the extra coinbase data.
    ///
    /// # Panics
    ///
    /// If `extra_coinbase_data` exceeds [`EXTRA_COINBASE_DATA_LIMIT`].
```

zebra-state/src/request.rs
```diff
@@ -733,6 +733,14 @@ pub enum Request {
     /// * [`Response::Transaction(None)`](Response::Transaction) otherwise.
     Transaction(transaction::Hash),
 
+    /// Looks up a transaction by hash in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// * [`Response::AnyChainTransaction(Some(Arc<Transaction>))`](Response::AnyChainTransaction) if the transaction is in the best chain;
+    /// * [`Response::AnyChainTransaction(None)`](Response::AnyChainTransaction) otherwise.
```
upbqdn:
```suggestion
    /// Looks up a transaction by hash in any chain.
    ///
    /// Returns
    ///
    /// * [`Response::AnyChainTransaction(Some(AnyTx))`](Response::AnyChainTransaction)
    ///   if the transaction is in any chain;
    /// * [`Response::AnyChainTransaction(None)`](Response::AnyChainTransaction)
    ///   otherwise.
```

zebra-state/src/service.rs
```diff
@@ -1579,6 +1604,37 @@ impl Service<ReadRequest> for ReadStateService {
                 .wait_for_panics()
             }
 
+            // Used by the getblock (verbose) RPC.
```
upbqdn:
```suggestion
```

zebra-rpc/src/methods.rs
```diff
@@ -1812,40 +1817,63 @@ where
...
+                        let block_hash = match self
+                            .read_state
+                            .clone()
+                            .oneshot(zebra_state::ReadRequest::BestChainBlockHash(tx.height))
+                            .await
+                            .map_misc_error()?
+                        {
+                            zebra_state::ReadResponse::BlockHash(block_hash) => block_hash,
+                            _ => unreachable!(
+                                "unmatched response to a `AnyChainTransaction` request"
```
upbqdn:
```suggestion
                                "unmatched response to a `BestChainBlockHash` request"
```

zebra-rpc/src/methods.rs
```diff
@@ -1812,40 +1817,63 @@ where
...
+                            txid,
+                        ),
+                    )),
+                }
             } else {
-                let hex = tx.tx.into();
+                let hex = match tx {
+                    AnyTx::Mined(mined_tx) => mined_tx.tx.into(),
+                    AnyTx::Side((transaction, _block_hash)) => transaction.into(),
+                };
```
upbqdn: We could get rid of this match if we impl'd `From` for `AnyTx`.

zebra-rpc/src/methods.rs
```diff
@@ -2960,9 +3005,20 @@ where
...
+            match r {
+                SubmitBlockResponse::Accepted => { /* pass */ }
+                SubmitBlockResponse::ErrorResponse(response) => {
+                    return Err(ErrorObject::owned(
+                        server::error::LegacyCode::Misc.into(),
+                        format!("block was rejected: {:?}", response),
+                        None::<()>,
+                    ));
+                }
+            }
```
upbqdn, line 3019: Could we not go with https://github.com/zcashfoundation/zebra/blob/ab390bb555238c1c580971aef542e32c05ddacc2/zebra-rpc/src/server/error.rs#L109?

zebra-state/src/service/read/block.rs
```diff
@@ -149,6 +149,47 @@ where
...
+/// The first chain in `chains` must be the best chain.
+pub fn any_transaction<'a>(
+    chains: impl Iterator<Item = &'a Arc<Chain>>,
+    db: &ZebraDb,
+    hash: transaction::Hash,
+) -> Option<AnyTx> {
+    // # Correctness
+    //
+    // It is ok to do this lookup in multiple different calls. Finalized state updates
+    // can only add overlapping blocks, and hashes are unique.
```
upbqdn, line 163: What do overlapping blocks mean?

zebra-rpc/src/methods.rs
conradoplg: Unfortunately not, `response` is not a `Result`

zebra-state/src/service/read/block.rs
conradoplg: (For context, I just adapted the comment from `mined_transaction`.)

This is about the issue where blocks could be committed into the state while the function is running; in that case the copies of the non-finalized chains will have a block that overlaps a committed block.

zebra-rpc/qa/rpc-tests/getrawtransaction_sidechain.py
conradoplg: I added a check in 7e5c15e16 which should achieve the same thing

zebra-rpc/src/methods.rs
conradoplg: Done in 7e5c15e16

zebra-rpc/src/methods/types/transaction.rs
upbqdn:
```suggestion
```

upbqdn:
```suggestion
```

upbqdn (dismissed):

upbqdn (approved):

PR #9886, opened 2025-09-11
feat(docker): add `curl` for health-check validation

### LINKED ISSUE #8830: feature: Add health-check (liveness and readiness) probes for Zebra
### Motivation

  - Ready to sync
  - Syncing
  - Synced

### Specifications

### Testing

_No response_

### Related Work

gustavovalverde:
## Motivation

This would allow to validate that Zebra nodes can receive traffic by creating a healt-check similar to this one:

```
curl -s -f --user "" --data-binary '{"jsonrpc": "2.0", "id":"healthcheck", "method": "getinfo", "params": [] }' -H 'Content-Type: application/json' http://127.0.0.1:${ZEBRA_RPC_PORT}/ | grep -q '"result":' || exit 1
```

Doesn't fix #8830 but it's a starting point

### Follow-up Work

- Create more complex healthcheck endpoints in Zebra

### PR Checklist

- [x] The solution is tested.


docker/Dockerfile
```diff
@@ -195,6 +195,12 @@ ENV USER=${USER}
 ARG HOME
 ENV HOME=${HOME}
 
+# Install curl for healtchecks
+RUN apt-get -qq update && \
+    apt-get -qq install -y --no-install-recommends \
```
arya2: Is the flag meant to be `--quiet`? I don't see a `-qq` flag in the `apt-get` reference. Using the long name of the flag might be nice.



```suggestion
RUN apt-get -q update && \
    apt-get -q install -y --no-install-recommends \
```

arya2 (dismissed):
Looks good!

docker/Dockerfile
gustavovalverde: In their man page they refer to it this way:

> More q's will produce more quiet up to a maximum of two. You can also use -q=# to set the quiet level, overriding the configuration file. Note that quiet level 2 implies -y

We want to avoid any kind of dialogues, and if they appear by any chance, then accept them.

But yes, it's redundant because we already have `-y`

conradoplg (approved):
Reapproving but seems stuck in CI

gustavovalverde:
@mergifyio update

PR #9888, opened 2025-09-12
Add(consensus): Allow configuration of checkpoints on test networks

arya2:
## Motivation

This PR allows for defining custom checkpoints in Zebra's config on test networks. 
This was needed to test #9809.

## Solution

- Moves `Network.checkpoint_list()` method from `zebra-consensus` to `zebra-chain`
- Adds a configurable `checkpoints` field to `testnet::Parameters`
- Adds an acceptance test for #9809 using configured checkpoints

### Tests

- Adds an acceptance test for #9809 that also covers checkpoint configurability
- Adds a new test config with the new field
- Updates old test configs now that the default behaviour is to expect checkpoints on custom Testnets if Canopy is not activated at height 1 (those configurations wouldn't have worked before either as Zebra can't semantically verify pre-Canopy blocks, but now it panics at startup instead of stalling).

### PR Checklist

- [x] The solution is tested.


zebra-chain/src/parameters/network.rs
```diff
@@ -146,6 +146,8 @@ impl std::fmt::Debug for Network {
                 .debug_struct("Regtest")
                 .field("activation_heights", params.activation_heights())
                 .field("funding_streams", params.funding_streams())
+                .field("checkpoints", &params.checkpoints())
+                .field("lockbox_disbursements", &params.lockbox_disbursements())
```
oxarbitrage, line 149: It seems i missed this one at https://github.com/ZcashFoundation/zebra/pull/9858, thanks.

oxarbitrage (dismissed):
Looks good, tested as well and it is working as expected.

oxarbitrage (approved):
thnaks!

PR #9895, opened 2025-09-15
feat(health): add HTTP `/healthy` and `/ready` endpoints

### LINKED ISSUE #8830: feature: Add health-check (liveness and readiness) probes for Zebra

gustavovalverde:
## Motivation

Kubernetes and load balancers need simple HTTP probes to avoid routing traffic to syncing nodes, and Zebra has no readiness/liveness HTTP checks to validate this.

Closes #8830.

## Solution

Add a tiny HTTP server in `zebrad` that exposes:

- `GET /healthy`: 200 when process is up and there are at least
  `min_connected_peers` recently live peers (default: 1); otherwise 503.
- `GET /ready`: 200 when the node is near the chain tip and the estimated lag
  is ≤ `ready_max_blocks_behind` (default: 2); otherwise 503.

Key points:

- Separate listener and minimal plain‑text responses; with no auth.
- Disabled by default. Opt‑in via a new `[health]` config section.
- Uses existing signals:
  - Near‑tip: `ChainSyncStatus::is_close_to_tip()`
  - Lag estimate: `ChainTip::estimate_distance_to_network_chain_tip(&network)`
  - Peer health: `AddressBookPeers::recently_live_peers()`

Config (`zebrad.toml`):

```toml
[health]
listen_addr = "0.0.0.0:8080"      # enable server; omit to disable
min_connected_peers = 1            # /healthy threshold
ready_max_blocks_behind = 2        # /ready threshold
enforce_on_test_networks = false   # if false, /ready is 200 on regtest/testnet
```

HTTP responses:

- 200 OK on success; 503 Service Unavailable otherwise.
- Plain text body: "ok" or a brief reason (e.g., "no peers", "syncing", "lag=N blocks").

## Tests

- Added unit tests in `zebrad/src/components/health.rs` (feature `proptest-impl`):
  - `healthy_and_ready_ok`
  - `not_ready_when_syncing_or_lagging`
  - `unhealthy_with_no_peers`
- Tests start the server on an ephemeral port and assert expected status codes.
- Manual check:
  1. Enable `[health]` in `zebrad.toml` (see above)
  2. `zebrad -c zebrad.toml start`
  3. `curl -i http://127.0.0.1:8080/healthy`
  4. `curl -i http://127.0.0.1:8080/ready`

## Follow-up Work

- Optional JSON responses (content‑negotiation) if needed.
- Consider acceptance tests that verify state transitions in long‑running nodes.

## PR Checklist

- [ ] The solution is tested.


gustavovalverde:
@emersonian You might want to have a look at this one for early feedback. I haven't been able to test it with a synced node yet.

zebrad/src/config.rs
```diff
@@ -77,6 +77,12 @@ pub struct ZebradConfig {
 
     /// Mining configuration
     pub mining: zebra_rpc::config::mining::Config,
+
+    /// Health check HTTP server configuration.
+    ///
+    /// See the Zebra Book for details and examples:
+    /// https://zebra.zfnd.org/user/health.html
```
arya2:
```suggestion
    /// <https://zebra.zfnd.org/user/health.html>
```

zebrad/src/commands/start.rs
```diff
@@ -203,6 +205,17 @@ impl StartCmd {
             misbehavior_sender.clone(),
         );
 
+        // Start health server if configured
+        info!("initializing health endpoints");
+        let (health_task_handle, _health_addr) = crate::components::health::init(
```
arya2: Nitpick: Could we add this to the list of imports from `components` at the top of this file?

```suggestion
        let (health_task_handle, _health_addr) = health::init(
```

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+    ResponseBuilder::new()
+        .status(status)
+        .header(CONTENT_TYPE, "text/plain; charset=utf-8")
+        .header(CONTENT_LENGTH, len.to_string())
+        .body(Full::new(bytes))
+        .expect("valid response")
+}
+
+#[cfg(all(test, feature = "proptest-impl"))]
+mod tests {
```
arya2: Nitpick: It would be nice to move this to its own file (at `components/health/tests.rs`, then move the mod declaration to the top of the file and condition it on just `cfg(test)` if it doesn't need anything that's behind `feature = "proptest-impl"`)

```suggestion
#[cfg(test)]
mod tests;
```

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+    #[derive(Clone)]
+    struct TestSyncStatus {
+        close: bool,
+    }
+
+    impl ChainSyncStatus for TestSyncStatus {
+        fn is_close_to_tip(&self) -> bool {
+            self.close
+        }
+    }
```
arya2: It would be nice to have this definition in `zebra-chain` (behind `cfg(any(test, feature = "proptest-impl"))`) so it can be used from tests in other crates without being moved later.

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    #[derive(Clone)]
+    struct FixedDistanceTip {
+        distance: block::HeightDiff,
+    }
+
+    impl ChainTip for FixedDistanceTip {
```
arya2: [`MockChainTip`](https://github.com/ZcashFoundation/zebra/blob/feat-zebrad-healthcheck/zebra-chain/src/chain_tip/mock.rs#L107) should support this use case already.

zebrad/Cargo.toml
```diff
@@ -201,6 +201,11 @@ metrics = { workspace = true }
 dirs = { workspace = true }
 atty = { workspace = true }
 
+# Health check server dependencies
+hyper = { workspace = true, features = ["server", "http1"] }
+http-body-util = { workspace = true }
+hyper-util = { workspace = true, features = ["service", "server", "tokio"] }
```
arya2, line 204: We may want to hide the health check server behind a compilation feature to keep these as optional dependencies, though I think they're required by `zebra-rpc` anyway.

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+use tokio::task::JoinHandle;
+use tracing::*;
+
+use zebra_chain::{chain_sync_status::ChainSyncStatus, chain_tip::ChainTip, parameters::Network};
+use zebra_network::AddressBookPeers;
+
+/// Health server configuration.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, default)]
+pub struct Config {
```
arya2: It would be nice to define this in its own module/file like every other `Config` struct except the `sync::Config`. And also to move the sync `Config` struct to its own file, though perhaps in another PR.

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+///
+/// Returns `200 OK` when the number of recently live peers is greater than or equal to
+/// `min_connected_peers`. Otherwise returns `503 Service Unavailable`.
+async fn healthy<A, Tip, SyncStatus>(ctx: &HealthCtx<A, Tip, SyncStatus>) -> Response<Full<Bytes>>
+where
+    A: AddressBookPeers + Clone + Send + Sync + 'static,
+    Tip: ChainTip + Clone + Send + Sync + 'static,
+    SyncStatus: ChainSyncStatus + Clone + Send + Sync + 'static,
+{
+    let live = ctx.address_book.recently_live_peers(Utc::now()).len();
```
arya2: We should check the sync status as well, make sure that Zebra has committed at least one block to its state in the past several minutes. We may want to revisit https://github.com/ZcashFoundation/zebra/issues/4649 so it's trivial to ensure this with a method on `SyncStatus`. 

From the issue:

> Any solution should consider this endpoints will be publicly available and thus their security implications should be thoroughly evaluated.

Publicly exposing this as-is would be a DDoS vulnerability (this method locks a `Mutex`), `health::init()` should instead spawn a task that periodically reads the `recently_live_peers()` and sends it to a watch channel, then read the value in the watch channel here (or otherwise expose an eventually-consistent view of the address book, but Zebra generally uses watch channels).

arya2 (commented):
Flushing partial review, this is looking great!

arya2 (changes requested):
We should make sure we fix the DDoS vulnerability.

zebrad/Cargo.toml
alchemydc: If this is hidden behind a compilation feature, we should make sure it's a feature that is enabled by default by our build scripts, as anybody running Zebra in prod is going to want/need this.  I appreciate being able to quickly disable it for a max security Zebra, but it should be "on by default" IMO.

zebrad/src/components/health.rs
alchemydc: I believe he's checking sync status in the /ready endpoint, but your points raised here re: false hints re: 4649 and potential DDOS are valid.  Ideally we can find a way to determine node sync status in a non blocking / non locking way. Ultimately I don't think that /health and /ready would be exposed publicly at the ingress layer for many production deployments but defense in depth principles say we should determine health/sync state without exposing us to DOS conditions.

zebrad/Cargo.toml
conradoplg: Since these dependencies are already used elsewhere, I don't think there is much benefit in adding a feature for this since disabling it wouldn't gain us anything, and having many features is confusing IMO

zebrad/src/components/health.rs
conradoplg: This seems just as risky as exposing the RPC interface. Since it's disabled by default in the config I think it's fine as long as the risks are documented.

arya2: > This seems just as risky as exposing the RPC interface.

This line in particular is exposing the most vulnerable part of the RPC interface, I agree, but the health service in general seems fairly low risk, as long as request handling is rate-limited, e.g. some requests are dropped if they arrive in quick succession.

emersonian:
This looks great to me from a product standpoint, excited, thanks for prioritizing. The endpoint will not be exposed publicly by default in any of our charts or docker compose files.

zebrad/src/components/health.rs
```diff
@@ -0,0 +1,457 @@
...
+            loop {
+                match listener.accept().await {
+                    Ok((stream, _peer)) => {
+                        let io = TokioIo::new(stream);
+                        let svc_ctx = shared.clone();
+                        let service = hyper::service::service_fn(move |req| {
+                            handle_request(req, svc_ctx.clone())
+                        });
+
+                        tokio::spawn(async move {
```
arya2: Optional: This should be rate-limited such that some requests are dropped if requests are arriving in quick succession.

book/src/user/docker.md
```diff
@@ -120,6 +119,39 @@ By default, Zebra uses cookie-based authentication for RPC requests (`enable_coo
...
+* `ZEBRA_HEALTH__MIN_CONNECTED_PEERS=1`
+* `ZEBRA_HEALTH__READY_MAX_BLOCKS_BEHIND=2`
+* `ZEBRA_HEALTH__ENFORCE_ON_TEST_NETWORKS=false`
+
+### Health Endpoints
+
+Zebra can expose two lightweight HTTP endpoints for liveness and readiness:
+
+* `GET /healthy`: returns `200 OK` when the process is up and has at least the configured number of recently live peers; otherwise `503`.
+* `GET /ready`: returns `200 OK` when the node is near the tip and within the configured lag threshold; otherwise `503`.
```
fabacab, line 134: I am so glad that health check endpoints are being added. Thank you.

I am wondering, what is the rationale behind using these names? The expectation of most experienced system reliability engineers including me is that these endpoints would be available at URL paths such as `/healthz`, `/livez`, and `/readyz`. The `z` suffix is not a typo, it is intentional.

This is not only extremely common in "cloud-native" applications, but is also [part of the Kubernetes reference API](https://kubernetes.io/docs/reference/using-api/health-checks/).

It's not what I'd consider a showstopper but it is very…weird, I guess, to me when I put on my SRE hat to see these endpoints strangely not using the canonical names when they are clearly defined as being for that purpose.

Note also that `/healthz` especially is officially deprecated in the Kubernetes world in favor of `/livez`, so I would prefer to see that over `/healthz` and especially over `/livez`.

Thanks regardless for prioritizing this kind of thing generally, though.

gustavovalverde: I do understand the Kubernetes API uses this convention (and Google, mostly), the `z` suffix comes from `zPages` and this was a decision[ in Kubernetes](https://kubernetes.io/docs/reference/instrumentation/zpages/) to avoid their health endpoints collitioning with other real applications deployed within Kubernetes (applications like Zebra).

From an SRE perspective I'd expect this in Kubernetes, but not in the overall SWE ecosystem. Actually, mainstream monitoring and SRE tools (and most CNFC projects) use `/healthy` and `/ready` (or other) endpoints for this same purpose, .e.g:
- Prometheus
- AlertManager
- Grafana (Loki, Tempo, Mimir)
- Envoy
- Traefik (`/ping`)
- Nginx (has a mix of `nginx-ready` and `healthz`)

So the convenventions used by the Kubernetes API are not _canonical_ nor _weird_ in the Cloud Native ecosystem.

I originally opened the issue for this and used the `liveness` and `readiness`, but even those terms get used in an inverted way by some SREs (Kubernetes documentation was not clear about this in previous years), so I also decided to use more meaningful words.

And if we talk about expectations, then I'd say we'd have to take other mainstream blockchain nodes as our baseline. And there's no consensus. You have `/liveness` and `readiness` in Hyperledger, `/health` in `Nethermind`, etc.

Although I could take the `z`-cash ecosystem as an argument for appending `z` 😄,  and instead do something like `healthz/live` and `healthz/ready`, but this is just a layer of complexity considering that `live` is ambiguous (it can mean the process is up, but not necessarily that it's healthy—connected to _peers_).

gustavovalverde:
@arya2 I've refactored this a "bit" more, so you might want to have another look.

arya2 (dismissed):

zebrad/src/components/health.rs
arya2: Fixed by https://github.com/ZcashFoundation/zebra/pull/9895/commits/280b7f10fe0f9cb189853536ef031cf7964c95eb.

gustavovalverde:
As this was pre-approved and test are passing, I'm merging it.

gustavovalverde:
@emersonian This is now in the `main` branch. Let us know if this works for you, it will be available in the next release but with early feedback we might be able to get some fixes in (if any required).

PR #9896, opened 2025-09-15
Python backup state tests

oxarbitrage:
## Motivation

Add python tests to checkpoint configuration.

### DIFF
--- a/zebra-rpc/build.rs
+++ b/zebra-rpc/build.rs
@@ -1,7 +1,7 @@
 //! Compile proto files
 use std::{env, fs, path::PathBuf, process::Command};
 
-const ZALLET_COMMIT: Option<&str> = None;
+const ZALLET_COMMIT: Option<&str> = Some("de70e46e37f903de4e182c5a823551b90a5bf80b");
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let out_dir = env::var("OUT_DIR").map(PathBuf::from);
--- a/zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py
+++ b/zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py
@@ -3,9 +3,8 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or https://www.opensource.org/licenses/mit-license.php .
 
-from decimal import Decimal
-from test_framework.util import assert_equal, start_nodes
 from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, start_nodes
 import time
 
 # Test that Zebra can backup and restore non finalized state
@@ -17,6 +16,7 @@ class BackupNonFinalized(BitcoinTestFramework):
         self.cache_behavior = 'clean'
 
     def setup_network(self):
+        # Start a node with default configuration
         self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [None])
 
     def run_test(self):
@@ -24,6 +24,7 @@ class BackupNonFinalized(BitcoinTestFramework):
         # Wait for 5 seconds (`MIN_DURATION_BETWEEN_BACKUP_UPDATES`) plus 1 second for I/O
         time.sleep(6)
 
+        # Check that we have 10 blocks
         blocks = self.nodes[0].getblockchaininfo()['blocks']
         assert_equal(blocks, 10)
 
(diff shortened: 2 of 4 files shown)


zebra-rpc/build.rs
```diff
@@ -1,7 +1,7 @@
 //! Compile proto files
 use std::{env, fs, path::PathBuf, process::Command};
 
-const ZALLET_COMMIT: Option<&str> = None;
+const ZALLET_COMMIT: Option<&str> = Some("de70e46e37f903de4e182c5a823551b90a5bf80b");
```
oxarbitrage, line 4: I am not using the build script to build the `zallet` here as is not working after some environment variables changes made to zebra (i might open a ticket for that). However, this is the commit i am using to build `zallet` manually, as current main is not working.

zebra-rpc/qa/rpc-tests/wallet.py
```diff
@@ -48,11 +48,6 @@ def run_test(self):
             print("Ignoring stopping wallet error: ", e)
         time.sleep(1)
 
-        # Hack for https://github.com/ZcashFoundation/zebra/issues/9708
-        # We Stop the wallet which has 1 block, generate 100 blocks in zebra,
-        # so when restarting Zebra it will have 1 block, just as the wallet.
-        self.nodes[0].generate(100)
```
arya2, line 54: This should wait 6 seconds for Zebra to write to the non-finalized state to the backup cache (updates to the backup are rate-limited to once per 5 seconds, plus 1 second for I/O).

I'll add a note about that in the config field docs. Or maybe it would be better to check if Zebra is close to the tip and write the backup blocks immediately. I'm leaning towards the latter now, but still it would be nice to wait a second here for I/O.

Though it shouldn't typically matter on Mainnet where the expected block spacing is 75 seconds.

zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [None])
+
+        # Good, the node has recovered the non finalized state
+        blocks = self.nodes[0].getblockchaininfo()['blocks']
+        assert_equal(blocks, 10)
+
+        # Generate more blocks and make sure the blockchain is still working
+        self.nodes[0].generate(1)
+        time.sleep(5)
```
arya2: Sleep here is probably unnecessary. It would be nice to make the one below 6 seconds rather than 5.

arya2:
This is looking good to merge into the base branch pending one small fix. 

Are there any python tests that assume Zebra will drop the last 100 blocks that might need the config option so that it continues doing that? (It won't backup the non-finalized state when using an ephemeral state, if there are tests that rely on that but they're using an ephemeral state).

zebra-rpc/qa/rpc-tests/wallet.py
oxarbitrage: You are right, 6 seconds is safer. Changed at https://github.com/ZcashFoundation/zebra/pull/9896/commits/9ee495821681c8b7e0dd1c1f365b3b25ef00677b

zebra-rpc/qa/rpc-tests/feature_backup_non_finalized_state.py
oxarbitrage: Right, 1 seconds is needed after generation if we have a wallet so it can sync with Zebra. No wallet here so no need to wait. Removed sleep's at https://github.com/ZcashFoundation/zebra/pull/9896/commits/9ee495821681c8b7e0dd1c1f365b3b25ef00677b

oxarbitrage:
> Are there any python tests that assume Zebra will drop the last 100 blocks that might need the config option so that it continues doing that? (It won't backup the non-finalized state when using an ephemeral state, if there are tests that rely on that but they're using an ephemeral state).

I don't think we will need that for now. `zcashd` behavior will always keep the state so no test with other behavior was ever done there. If we eventually find out we want that ourselves, then we need to add either `ephemeral` or `should_backup_non_finalized_state` parameters to regtest mode before creating a python test.

oxarbitrage:
I extended the test a little bit at https://github.com/ZcashFoundation/zebra/pull/9896/commits/275e0fa845cfb63ce65a49c8287a55a3b8afd0cb to show a new restart at heights greater than 100.

arya2 (approved):

PR #9903, opened 2025-09-17
tests: always wait for RPC port to open

conradoplg:
## Motivation

I think there might be a bug in https://github.com/ZcashFoundation/zebra/blob/4e4decd7ab9dbce7dd3db75a8fe8b56889b13c0a/zebrad/tests/common/lightwalletd/wallet_grpc_test.rs#L126-L133.

It calls `wait_for_state_version_upgrade` to possibly wait for a DB upgrade and to also wait for the RPC port to open, but only does the latter if it detects a DB upgrade; i.e. it's not currently waiting for the RPC port to open. This can be confirmed in the logs by seeing that [`zebrad opened its RPC port, spawning lightwalletd`](https://github.com/ZcashFoundation/zebra/actions/runs/17804589453/job/50615565681?pr=9875#step:13:152) is logged way before [the RPC port is opened](https://github.com/ZcashFoundation/zebra/actions/runs/17804589453/job/50615565681?pr=9875#step:13:214)

This could be the cause of CI failures, but I'm not 100% sure. Regardless, it seems like the right thing to do.

## Solution

Check the additional regexes even if there was not state upgrade. This could be the wrong thing to do depending on how the function is called, but all callers either use no additional regexes or the "Opened RPC port..." one

### PR Checklist

- [ ] The solution is tested.


arya2 (approved):

arya2:
> I think there might be a bug in

Yep. Nice catch.

conradoplg:
As the [current CI failure](https://github.com/ZcashFoundation/zebra/actions/runs/17812930466/job/50642102825?pr=9903) shows, it doesn't solve that particular issue :clown_face: But it does wait for the RPC port to open as expected https://github.com/ZcashFoundation/zebra/actions/runs/17812930466/job/50642102825?pr=9903#step:13:204 (see previous line)

conradoplg:
@mergifyio update

gustavovalverde:
@Mergifyio update

Issue #9721, opened 2025-07-23
fix(qa): Improve connection between nodes


oxarbitrage:
Our QA framework currently includes only one multi-node test:
[addnode.py](https://github.com/ZcashFoundation/zebra/blob/c0f32eee1899b728ca4f3b35e8edadb1103e5d2e/zebra-rpc/qa/rpc-tests/addnode.py).

We previously had a multi-node + wallet test, but it was simplified to a single-node setup due to long waiting times related to connection issues.

Problems Identified:

- Slow or unreliable connections:
    - Connections between nodes can take too long or sometimes fail altogether. See:
[addnode.py#L39-L52](https://github.com/ZcashFoundation/zebra/blob/c0f32eee1899b728ca4f3b35e8edadb1103e5d2e/zebra-rpc/qa/rpc-tests/addnode.py#L39-L52)
    - If changes to Zebra are too big, we should at least isolate and improve the connection timeout logic into a reusable utility function in the QA framework.

- Broadcast limitation in generate function:
  - When calling `generate(n)` with n > 1, only the last generated block is broadcast to peers. See:
[addnode.py#L77-L83](https://github.com/ZcashFoundation/zebra/blob/c0f32eee1899b728ca4f3b35e8edadb1103e5d2e/zebra-rpc/qa/rpc-tests/addnode.py#L77-L83)
  - A workaround might be to wrap the generate call in a loop on the Python side, or we can consider improving Zebra itself to broadcast all blocks.

Suggested Tasks:

- Refactor connection wait logic into a generic utility with configurable timeout.
- Improve robustness of node-to-node connectivity during tests.
- Investigate the generate(n) broadcast limitation:
  - Decide whether to live with it and implement a workaround (loop in Python). 
  - Or, if feasible, update Zebra to broadcast all generated blocks.
- Reintroduce or create new multi-node tests once connection behavior is improved.

oxarbitrage:
Related: https://github.com/ZcashFoundation/zebra/issues/9305

oxarbitrage:
Closed by https://github.com/ZcashFoundation/zebra/pull/9917 and https://github.com/ZcashFoundation/zebra/pull/9907

PR #9907, opened 2025-09-18
change(network): Broadcast block submissions to all peers in the peer set

arya2:
## Motivation

It would be nice to broadcast accepted block submissions to all ready peers.

## Solution

- Adds a `AdvertiseBlockToAll` variant to the zebra-network `Request` type
- Handles the new request variant in the peer set by sending it to all ready peers
  - Factors logic for sending a request to multiple peers out of `route_multiple()`
  - Adds a `broadcast_all()` method which sends a request to every ready peer
- Updates the `gossip_best_tip_block_hashes()` task to call the peer set with a `AdvertiseBlockToAll` request instead of `AdvertiseBlock` if the block hash/height being advertised were received from the mined block channel.

Related changes:
- Corrects a doc comment on `AdvertiseBlock` (it routes to a third of ready peers, not half)
- Updates error message from the `submitblock` RPC when the mined block channel is closed

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -1011,6 +1011,7 @@ where
     /// Broadcasts the same request to all ready peers, ignoring return values.
     fn broadcast_all(&mut self, req: Request) -> <Self as tower::Service<Request>>::Future {
         // Broadcasts ignore the response
+        // TODO: If there are some unready peers, queue the request to be sent to the current list of unready peers when they become ready.
         let all_ready_peers = self.ready_services.keys().copied().collect();
         self.send_multiple(req, all_ready_peers)
     }
--- a/zebra-network/src/peer_set/set.rs
+++ b/zebra-network/src/peer_set/set.rs
@@ -160,6 +160,8 @@ pub struct MorePeers;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub struct CancelClientWork;
 
+type ResponseFuture = Pin<Box<dyn Future<Output = Result<Response, BoxError>> + Send + 'static>>;
+
 /// A [`tower::Service`] that abstractly represents "the rest of the network".
 ///
 /// # Security
@@ -202,11 +204,9 @@ where
     inventory_registry: InventoryRegistry,
 
     /// Stores requests that should be routed to peers once they are ready.
-    queued_broadcast_all_request: Option<(
+    queued_broadcast_all: Option<(
         Request,
-        tokio::sync::mpsc::Sender<
-            Pin<Box<dyn Future<Output = Result<Response, BoxError>> + Send + 'static>>,
-        >,
+        tokio::sync::mpsc::Sender<ResponseFuture>,
         HashSet<D::Key>,
     )>,
 
@@ -325,7 +325,7 @@ where
             ready_services: HashMap::new(),
             // Request Routing
             inventory_registry: InventoryRegistry::new(inv_stream),
-            queued_broadcast_all_request: None,
+            queued_broadcast_all: None,
 
             // Busy peers
             unready_services: FuturesUnordered::new(),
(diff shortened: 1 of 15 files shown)


oxarbitrage:
Nice, this is part of https://github.com/ZcashFoundation/zebra/issues/9721

oxarbitrage:
We might want to test this by removing:

https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/addnode.py#L73-L79

arya2:
> We might want to test this by removing:
> 
> https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/addnode.py#L73-L79

I replaced that with a single call to `generate(10)`, it should only broadcast most of those to 1 peer as the rest will be unready and the peerset only sends the last `AdvertiseBlockToAll` request to unready peers once they become ready, but the 1 peer that receives each advertised block will then advertise to all of its peers.

We could make `queued_broadcast_all_request` a vector and always send every block submission to every peer, but that doesn't seem useful on Mainnet.

zebra-network/src/peer_set/set.rs
```diff
@@ -998,6 +1040,53 @@ where
...
+                Ok(Response::Nil)
+            }
+            .boxed()
+        }
+    }
+
+    /// Broadcasts the same requests to all ready peers which were unready when
+    /// [`PeerSet::broadcast_all()`] was last called, ignoring return values.
+    fn broadcast_all_queued(&mut self) {
+        if let Some((req, sender, mut remaining_peers)) = self.queued_broadcast_all_request.take() {
```
arya2: This could be a let-else statement

zebrad/src/components/sync/gossip.rs
```diff
@@ -87,43 +101,52 @@ where
...
+
+        // Mark the last change hash of `chain_state` as the last block submission hash to avoid
+        // advertising a block hash to some peers twice.
+        if is_block_submission
+            && mined_block_receiver
+                .as_ref()
+                .is_some_and(|rx| rx.is_empty())
+            && chain_state.latest_chain_tip().best_tip_hash() == Some(hash)
+        {
+            chain_state.mark_last_change_hash(hash);
```
arya2, line 159: This may need a correctness comment about not marking the value in `latest_chain_tip` channel as seen to avoid a rare race condition where some updates could be missed, though it doesn't seem disastrous to very rarely miss a chain tip update after advertising a block submission.

zebra-network/src/peer_set/set.rs
```diff
@@ -1204,7 +1276,13 @@ where
...
             // - `poll_background_errors` schedules this task for wakeup when the peer status
             //   update task exits.
-            Poll::Pending
+            return Poll::Pending;
+        }
+
+        self.broadcast_all_queued();
+
+        if self.ready_services.is_empty() {
+            self.poll_peers(cx)
```
arya2, line 1305: These changes to `poll_ready()` should be reviewed carefully.

Update: They still look right to me.

zebrad/src/components/sync/gossip.rs
```diff
@@ -65,9 +67,21 @@ where
...
+            // in practice, we expect blocks to arrive approximately every 75 seconds,
+            // so waiting 6 seconds won't make much difference
+            tokio::time::sleep(PEER_GOSSIP_DELAY).await;
+
             // wait for at least one tip change, to make sure we have a new block hash to broadcast
             let tip_action = chain_tip.wait_for_tip_change().await.map_err(TipChange)?;
 
+            // wait for block submissions to be received through the `mined_block_receiver` if the tip
+            // change is from a block submission.
+            tokio::time::sleep(WAIT_FOR_BLOCK_SUBMISSION_DELAY).await;
```
arya2, line 88: This should have a test to ensure that the delay is typically long enough for the `tokio::select!` below to use the message from the mined block receiver over the chain tip change.

Update: I've added a TODO for it, it seems unlikely that the wait is too short and low-impact if it is (Zebra will just advertise the block twice to a subset of its peers).

zebrad/src/components/sync/gossip.rs
```diff
@@ -65,9 +67,21 @@ where
         let mut sync_status = sync_status.clone();
         let mut chain_tip = chain_state.clone();
         let tip_change_close_to_network_tip_fut = async move {
+            const WAIT_FOR_BLOCK_SUBMISSION_DELAY: Duration = Duration::from_micros(100);
+
+            // wait for at least the network timeout between gossips
+            //
+            // in practice, we expect blocks to arrive approximately every 75 seconds,
+            // so waiting 6 seconds won't make much difference
+            tokio::time::sleep(PEER_GOSSIP_DELAY).await;
```
arya2, line 81: Moving this into `tip_change_close_to_network_tip_fut` and replacing the watch channel with an mpsc channel was the main fix in this PR for catching multiple block submissions. This PR could also wait for the peer set to become ready in a bounded `FuturesOrdered` if we want to keep the watch channel. The `tokio::select!` below doesn't need to be biased to prefer messages from the mined block channel because it can rely on this future to sleep for a constant duration, but this may need a code comment explaining that it would otherwise need to be biased.

zebra-rpc/src/methods/types/submit_block.rs
```diff
@@ -72,26 +72,26 @@ impl From<SubmitBlockErrorResponse> for SubmitBlockResponse {
...
-    receiver: watch::Receiver<(block::Hash, block::Height)>,
+    receiver: mpsc::Receiver<(block::Hash, block::Height)>,
 }
 
 impl SubmitBlockChannel {
-    /// Create a new submit block channel
+    /// Creates a new submit block channel
     pub fn new() -> Self {
-        let (sender, receiver) = watch::channel((GENESIS_PREVIOUS_BLOCK_HASH, block::Height::MIN));
+        let (sender, receiver) = mpsc::channel(10_000);
```
arya2: This channel capacity should be moved to a constant. It could be smaller on Mainnet and probably also on test networks. It limits the maximum number of block submissions that can be queued to be advertised to peers.

zebrad/src/components/sync/gossip.rs
```diff
@@ -87,43 +101,52 @@ where
...
-            tokio::select! {
-                tip_change_close_to_network_tip = tip_change_close_to_network_tip_fut => {
-                    mined_block_receiver.mark_unchanged();
-                    tip_change_close_to_network_tip?
-                },
-
-                Ok(_) = mined_block_receiver.changed() => {
-                    // we have a new block to broadcast from the `submitblock `RPC method, get block data and release the channel.
-                   (*mined_block_receiver.borrow_and_update(), "sending mined block broadcast", chain_state)
+        let (((hash, height), log_msg, updated_chain_state), is_block_submission) =
```
arya2, line 110: It would be nice to refactor this function into a struct and divide the logic into methods.

arya2:
@gustavovalverde any idea why the [deploy nodes job](https://github.com/ZcashFoundation/zebra/actions/runs/17935595853/job/51001565226?pr=9907) is being skipped?

gustavovalverde:
> @gustavovalverde any idea why the [deploy nodes job](https://github.com/ZcashFoundation/zebra/actions/runs/17935595853/job/51001565226?pr=9907) is being skipped?

@arya2 We recently decided to move it to the `main` branch only. I also recently merged a fix to avoid the success job from failing. If you'd like to run the stateful tests in this PR you can add the `run-stateful-tests` label to it.

zebrad/src/components/inbound/tests/real_peer_set.rs
```diff
@@ -873,6 +873,12 @@ mod submitblock_test {
 
         // Start the block gossip task with a SubmitBlockChannel
         let submitblock_channel = SubmitBlockChannel::new();
+        // Send a block top the channel
```
oxarbitrage:
```suggestion
        // Send a block to the channel
```

zebra-rpc/src/methods/types/submit_block.rs
oxarbitrage: Agree. If we keep it as it is please add a `TODO` on top.

zebrad/src/components/sync/gossip.rs
oxarbitrage: A `TODO` comment will do it for now.

zebrad/src/components/sync/gossip.rs
oxarbitrage: If you are advertising a block submission, you can be the miner. I am not sure how bad can be that for one of them.

zebra-network/src/peer_set/set.rs
```diff
@@ -998,6 +1040,53 @@ where
         self.route_multiple(req, self.number_of_peers_to_broadcast())
     }
 
+    /// Broadcasts the same request to all ready peers, ignoring return values.
+    fn broadcast_all(&mut self, req: Request) -> <Self as tower::Service<Request>>::Future {
```
oxarbitrage, line 1044: The code of this method looks complicated. Any way to split or simplify ?

oxarbitrage (commented):
Looking good. This will also close https://github.com/ZcashFoundation/zebra/issues/9721 if merged as the other task was done in https://github.com/ZcashFoundation/zebra/pull/9917

zebrad/src/components/sync/gossip.rs
arya2: I think a correctness comment is unnecessary, it's practically never going to happen that the channel is updated during the nanoseconds it takes to do this check, the issue exists in the getblocktemplate without a comment, and the implications of it happening are insignificant as it would be a block received from the network that was likely advertised to other nodes as well, and that block would have to arrive immediately after a block submission such that it couldn't be a child of the mined block in that case. It's also impossible for the received block to be a child of the mined block in the unlikely event that the peer set has no ready peers for a long time.

zebra-rpc/src/methods/types/submit_block.rs
arya2: Moved to a constant and documented in https://github.com/ZcashFoundation/zebra/pull/9907/commits/264740b9ecc795421d33b748dced98d003421504.

zebrad/src/components/sync/gossip.rs
arya2: Added TODOs to refactor this function in https://github.com/ZcashFoundation/zebra/pull/9907/commits/264740b9ecc795421d33b748dced98d003421504.

zebra-network/src/peer_set/set.rs
arya2: Yep. This was a late addition I hadn't thought through much, thank you for catching that. Split and simplified in https://github.com/ZcashFoundation/zebra/pull/9907/commits/bc9f5c6b5da850ecacc9b7f4376dbe4f632f8f2d.

oxarbitrage (approved):
thanks!

PR #9910, opened 2025-09-19
fix(health): prevent false ready status when network disconnects

### LINKED ISSUE #4649: Zebra falsely estimates that it's close to the tip when the network connection goes down

gustavovalverde:
Core changes:
- `is_close_to_tip()` now returns false when all sync lengths are zero
- Health endpoints include peer counts in responses for visibility
- Regtest always returns ready (no peers expected on that network)
- Error messages show specific failure reasons

The fix prevents a scenario where:
1. Network link drops or all peers disconnect
2. Sync batches become [0, 0, 0] (no blocks downloading)
  - Previously this was interpreted as "at tip"

Now correctly identified as "disconnected"

Closes #4649


zebrad/src/components/sync/status.rs
```diff
@@ -65,6 +65,13 @@ impl ChainSyncStatus for SyncStatus {
             .iter()
             .fold(0usize, |sum, rhs| sum.saturating_add(*rhs));
 
+        // If every recent batch is empty, Zebra has likely lost every peer and is
+        // no longer making progress. Treat this the same as "still syncing" so we
+        // don't falsely advertise readiness after a network outage (issue #4649).
+        if sum == 0 {
+            return false;
+        }
```
arya2, line 175: I think we should address the issue with `is_close_to_tip()` returning false positives independently and either merge it ahead of merging #9895 or check the recently live peers in the address book in #9895 and update it to rely on `SyncStatus::is_close_to_tip()` in the PR that fixes `is_close_to_tip()`.

It looks like this change might cause `is_close_to_tip()` to occasionally return false negatives if:
- `ChainSync::try_to_sync()` is called when Zebra's already at the network chain tip while there are prospective tips and calls `obtain_tips()`, `extend_tips()`, then `obtain_tips()` again causing the last 3 sync lengths to be zero until the next `try_to_sync()` call after the 67s sync restart delay (I might be wrong here, we may want to double-check), or
- A chain tip is gossiped to Zebra or provided as a block submission so that it's sent to the block verifier without updating the recent sync lengths so that any number of `obtain_tips()` calls could report a recent sync length of zero.

To avoid false negatives in those cases, we could:
- Increase `MAX_RECENT_LENGTHS` to be 4 so that the timing issue can't cause every recent sync length to be zero and update the recent sync lengths from the inbound service and `submitblock` RPC, 
- Track how long it's been since the last chain tip change and return true here if it was recent enough,
- Track how long it's been since the block time of the chain tip and return true here if it was recent enough,
- Check that Zebra has live peers in its peer set or recently live peers in its address book (though the peers may also be stalled)

zebrad/src/components/sync/status.rs
```diff
@@ -28,11 +67,36 @@ impl SyncStatus {
...
+        progress_grace: Duration,
+    ) -> (Self, RecentSyncLengths) {
+        let tip_progress = Arc::new(Mutex::new(TipProgress::new()));
+
+        let progress_tip = Arc::clone(&tip_progress);
+        let progress_callback: Arc<dyn Fn() + Send + Sync> = Arc::new(move || {
+            let mut tip = progress_tip.lock().expect("tip progress mutex poisoned");
+            tip.last_updated = Instant::now();
+            tip.has_progress = true;
+        });
```
arya2, line 88: We generally prefer watch channels over `Mutex` where possible, in this case, the `last_updated` time and `has_progress` flag can be added as fields to the `RecentSyncLength`'s watch channel message type, though I'm not sure it's necessary to record those at all.

zebrad/src/components/sync/status.rs
```diff
@@ -47,11 +111,52 @@ impl SyncStatus {
...
+            .lock()
+            .expect("tip progress mutex poisoned");
+        tip.has_progress && now.duration_since(tip.last_updated) <= self.progress_grace
+    }
 }
 
 impl ChainSyncStatus for SyncStatus {
     /// Check if the synchronization is likely close to the chain tip.
     fn is_close_to_tip(&self) -> bool {
+        if !self.has_enough_ready_peers() {
```
arya2, line 154: It's okay if there aren't any ready peers, we just want to make sure that there are connected peers, ready or unready, meaning peers that recently responded to a request or sent us a request. Peer connections have an associated heartbeat task that pings them every 59s (`HEARTBEAT_INTERVAL`), and the peer set will remove peers that don't respond to it from the list of ready or unready services in the following call to its `poll_ready()` method.

Update: The address book actually has a more up-to-date view of which peers have recently responded than the peer set.

zebrad/src/components/sync/recent_sync_lengths.rs
```diff
@@ -35,13 +42,16 @@ impl RecentSyncLengths {
 
     /// Create a new instance of [`RecentSyncLengths`]
     /// and a [`watch::Receiver`] endpoint for receiving recent sync lengths.
-    pub fn new() -> (Self, watch::Receiver<Vec<usize>>) {
+    pub fn new(
+        progress_callback: Option<Arc<dyn Fn() + Send + Sync>>,
```
arya2, line 46: This seem unnecessarily opaque, I think it would be better to pass the mutex itself.

zebrad/src/components/sync/status.rs
```diff
@@ -47,11 +111,52 @@ impl SyncStatus {
...
+
+    /// Returns `true` if we've observed any activity (either a non-zero batch or a
+    /// tip height change) within the configured grace window.
+    fn recent_progress_within_grace(&self) -> bool {
+        let now = Instant::now();
+        let tip = self
+            .tip_progress
+            .lock()
+            .expect("tip progress mutex poisoned");
+        tip.has_progress && now.duration_since(tip.last_updated) <= self.progress_grace
```
arya2, line 147: This is likely problematic, because near the tip, many blocks may be advertised to Zebra such that Zebra's chain state remains near the network tip leaving nothing for the `ChainSync` component to do. It would be better to track the duration since the last chain tip change (like the `show_block_chain_progress()` task).

It also seems unnecessary to check that the chain tip has changed recently _and_ that Zebra still has peers. I'm not completely sure what we want to track with `is_close_to_tip()`, but I think as long as Zebra has peers, it _is_ close to the network tip in a sense, though if Zebra's chain tip hasn't changed in a long time and Zebra has peers, the network may have partitioned, there may be a chain stall on the network, or the local chain may have stalled despite Zebra having access to valid blocks for some reason (e.g. if Zebra finalizes a block on a side chain).

arya2 (changes requested):
This could work with some changes, but seems mostly unrelated to the base PR and over-complicated.

arya2:
@gustavovalverde We can probably close this PR as the health check server already avoids false ready statuses when Zebra has no peers and there seems to be little room or need for improvement on it.

mpguerra:
Does this mean that we can close #4649 also? Otherwise we should move it to the icebox

PR #9916, opened 2025-09-22
fix(ci): disk mounts, alls-green conditions, and permissions

gustavovalverde:
## Motivation

- Keep CI from failing when cached disks lack partitions.
- Ensure the Alls-Green aggregation job reflects the actual dependency guards.
- Minimize permissions for reusable workflows and secrets.

## Solution

- align every Alls-Green “success” job with the dependency conditions and add `allowed-skips` where needed so the aggregator only runs when its inputs are eligible.
- `declare `id-token: write` exactly where the reusable GCP deploy workflow needs it.
- pass `GCP_SSH_PRIVATE_KEY` explicitly from callers and require it in the reusable workflow, avoiding inherited secrets.
- drop the partition-specific mount logic so cached state disks attach as whole devices regardless of layout.

### Tests

- Kick off the CI pipelines that mount cached disks (e.g., `sync-update-mainnet`, `rpc-*`) to confirm they succeed without partition parameters. 

- Internal Zebra CI workflows.

### PR Checklist

- [ ] The PR follows the contribution guidelines.
- [ ] The solution is tested.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -56,8 +56,12 @@ jobs:
   release-binaries-success:
     name: Release binaries success
     runs-on: ubuntu-latest
-    # Always run this job to check the status of dependent jobs
-    if: always()
+    # Only run when the Docker Hub update job is allowed to execute
+    if: >-
+      ${{
+        always() &&
+        github.repository_owner == 'ZcashFoundation'
+      }}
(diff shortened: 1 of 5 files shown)


gustavovalverde:
I'm admin-merging this one because this is a hotfix, it's tested and we need to unblock several PRs

PR #9917, opened 2025-09-22
tests(qa): Improve connection times in regtest framework

oxarbitrage:
## Motivation

When testing multiple nodes connected to each other in the python framework, the connection takes too much time. This is because the default `crawl_new_peer_interval` takes 61 seconds.

Part of https://github.com/ZcashFoundation/zebra/issues/9721

## Solution

We can reduce that amount from the config to 5 seconds, making the peers to connect to each other faster.

Additionally, i removed the `reindex.py` test as the functionality we were testing there do not make sense anymore after https://github.com/ZcashFoundation/zebra/pull/9809

### Tests

- `addnode.py` test

### PR Checklist

- [x] The solution is tested.


zebra-rpc/qa/rpc-tests/addnode.py
```diff
@@ -24,29 +24,11 @@ def setup_network(self, split=False):
...
-        while timeout_for_connetions > 0:
-            if (len(self.nodes[0].getpeerinfo()) < 4 or
-                len(self.nodes[1].getpeerinfo()) < 4 or
-                len(self.nodes[2].getpeerinfo()) < 4):
-                timeout_for_connetions -= wait_time
-                time.sleep(wait_time)
-            else:
-                break
-        assert timeout_for_connetions > 0, "Timeout waiting for connections"
-
```
oxarbitrage, line 49: The above check was removed as is guaranteed by https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/qa/rpc-tests/test_framework/util.py#L631-L638. In fact there is no timeout there so the poll will be until connected. 

`connect_nodes_bi` will call `connect_nodes` twice so we have 2-side coverage for each connection, which should be enough.

conradoplg (approved):
Looks good, thanks!

conradoplg:
@mergifyio update

PR #9925, opened 2025-09-25
feat(health): add HTTP /health endpoint

### LINKED ISSUE #8830: feature: Add health-check (liveness and readiness) probes for Zebra

seniakalma:
## Motivation

When Zebra runs in different orchestration systems (ECS, K8s), and when oad balancers route traffic to it there's a need for simple HTTP probes to avoid routing traffic to un-healthy nodes, and Zebra has no readiness/liveness HTTP checks to validate this.

Closes https://github.com/ZcashFoundation/zebra/issues/8830.

## Solution

**This solution reuses the existing Zebra server and does not require the instantiation of a dedicated server internally.**
It uses the current JSON RPC (RPSEE) server and adds a simple GET health endpoint to it that returns the following details:
`curl -s http://127.0.0.1:8232/health`
response:
`{
  "status": "healthy",
  "version": "2.4.0",
  "git_tag": "v2.4.0-rc.1",
  "git_commit": "75397cd59d0bac4b4c1e71187bb8af7494efb8b4",
  "timestamp": "2025-09-02T22:35:14.756599+00:00"
}
`

### Tests

A manual test of running the node locally and checking that the endpoint returns the health information correctly for a GET request for /health
In addition to that, those 2 snapshots tests were added:
- Mainnet test (get_health_info_status@mainnet_10.snap)
Checks that calling get_health_info_status on mainnet returns:
`{ "status": "healthy" }`

Testnet test (get_health_info_status@testnet_10.snap)
Checks that calling get_health_info_status on testnet also returns:
`{ "status": "healthy" }`

### Follow-up Work

Possible to add or adapt any information returned for that request.

### PR Checklist

- [x] The solution is tested.


arya2 (dismissed):
This looks good, thank you! We also want a dedicated endpoint for when the RPC server is disabled, but, @gustavovalverde, it would be nice to allow for disabling the dedicated endpoint as well for when the RPC server's `/health` endpoint is enabled and sufficient.

arya2:
We decided against adding a second health check endpoint for now to avoid having health check endpoints with different behaviour. We wanted to allow the health check endpoint to be enabled independently of the RPC server, so we merged the independent health check server, and we may later want to refactor the logic so that it may be exposed at both endpoints, but one is good enough for now.

gustavovalverde:
@seniakalma if you think the existing health check we have is missing something or doesn't consider an specific use case for uses, please check and let us know.

PR #9927, opened 2025-09-25
Migrate to `zcash_primitives 0.25` etc.

str4d:
### PR Checklist

- [ ] The solution is tested.


arya2 (dismissed):
Looks good, thank you!

zebra-chain/src/primitives/zcash_primitives.rs
```diff
@@ -47,7 +48,9 @@ impl zcash_transparent::sighash::TransparentAuthorizingContext for TransparentAu
         self.all_prev_outputs
             .iter()
             .map(|prevout| {
-                zcash_primitives::legacy::Script(prevout.lock_script.as_raw_bytes().into())
+                zcash_primitives::legacy::Script(script::Code(
```
nuttycom: Shouldn't this be using the type from `zcash_transparent` instead of the deprecated reexport from `zcash_primitives`?

Cargo.toml
```diff
@@ -24,13 +24,13 @@ incrementalmerkletree = { version = "0.8.2", features = ["legacy-api"] }
 orchard = "0.11"
 sapling-crypto = "0.5"
 zcash_address = "0.9"
-zcash_client_backend = "0.19"
+zcash_client_backend = "0.20"
```
nuttycom: I though that zebra no longer needed the `zcash_client_backend` dependency?

zebra-chain/src/primitives/zcash_primitives.rs
str4d: I just updated what was necessary to get code compiling, due to lack of time. Indeed this should be updated, and will need to be once we remove the deprecated re-exports.

Cargo.toml
str4d: Indeed, it seems that this was just not removed at the time, and is now superfluous.

str4d:
[Force-pushed](https://github.com/ZcashFoundation/zebra/compare/38e1a2ffe72e09960aa9beca64eb64e1af15f28f..2638f552d5199b395f2ce0229ce948c67039a164) to address one comment, and then pushed a commit to address the other.

conradoplg:
@mergifyio update

conradoplg (approved):

PR #9928, opened 2025-09-26
fix(ci): resolve multiple CI workflow issues and required status checks

gustavovalverde:
## Summary

This PR fixes multiple CI workflow issues that were preventing builds and PRs from completing successfully:

### Issues Fixed

1. **Missing clippy component** - The test-crates workflow was failing with "'cargo-clippy' is not installed" error
2. **fromJSON empty input errors** - Integration tests were failing when get-available-disks job was skipped 
3. **Required status checks blocking PRs** - PRs couldn't merge when workflows were skipped due to path filters

#### 1. Added clippy component to test-crates workflow
- Modified actions-rust-lang/setup-rust-toolchain to include `components: clippy`
- Ensures clippy is available for all cargo clippy commands

#### 2. Fixed fromJSON errors in GCP integration tests
- Added checks to verify job succeeded before parsing outputs with fromJSON
- Prevents "empty input" errors when jobs are skipped for external PRs

#### 3. Added allowed-skips to integration test success job
- Configured alls-green action to allow specific jobs to be skipped
- Prevents workflow failure when dependent jobs are skipped

#### 4. Created mutually exclusive status checks workflow
- Added status-checks.patch.yml to provide required status checks when main workflows don't run
- Uses paths-ignore to ensure mutual exclusion with main workflows
- Prevents race conditions where fake success could mask real test failures

### Security Considerations

The status-checks.patch.yml uses `paths-ignore` that must be the exact inverse of `paths` in the main workflows to ensure mutual exclusion. This prevents both workflows from running simultaneously, which could allow malicious code to pass checks.

### PR Checklist

- [x] The PR name is suitable for the release notes
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md)
- [x] The library crate changelogs are up to date
- [x] The solution is tested
- [x] The documentation is up to date

### DIFF
--- a/.github/workflows/status-checks.patch.yml
+++ b/.github/workflows/status-checks.patch.yml
@@ -1,0 +1,43 @@
+# This workflow ensures required status checks pass when main workflows are skipped.
+# !CRITICAL: Uses paths-ignore to ensure mutual exclusion with main workflows.
+# Only runs when NO Rust/config files are modified.
+# See: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+
+name: Status Check Patch
+
+on:
+  pull_request:
+    branches: [main]
+    paths-ignore:
+      # This MUST be an exact inverse of paths in lint.yml, tests-unit.yml, and test-crates.yml
+      # to ensure mutual exclusion - only one set of workflows runs
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - .cargo/config.toml
+      - '**/clippy.toml'
+      - .github/workflows/lint.yml
+      - .github/workflows/tests-unit.yml
+      - .github/workflows/test-crates.yml
+
+permissions:
+  contents: read
+
+jobs:
+  lint-success:
+    name: lint success
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "No lint needed - no Rust files modified"
+
+  test-success:
+    name: test success
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "No tests needed - no Rust files modified"
+
+  test-crate-build-success:
+    name: test crate build success
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "No crate build needed - no Rust files modified"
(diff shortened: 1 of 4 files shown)


.github/workflows/status-checks.patch.yml
```diff
@@ -0,0 +1,43 @@
...
+      # This MUST be an exact inverse of paths in lint.yml, tests-unit.yml, and test-crates.yml
+      # to ensure mutual exclusion - only one set of workflows runs
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - .cargo/config.toml
+      - '**/clippy.toml'
+      - .github/workflows/lint.yml
+      - .github/workflows/tests-unit.yml
+      - .github/workflows/test-crates.yml
```
conradoplg, line 21: There is a subtle issue here that I noticed some time ago but was waiting for the CI refactor. I'll try to explain... this is from the `paths` documentation:

> If at least one path matches a pattern in the paths filter, the workflow runs.

For reference, the paths in `lint.yml`:

```
    paths:
      - "**/*.rs"
      - "**/Cargo.toml"
      - "**/Cargo.lock"
      - clippy.toml
      - .cargo/config.toml
      - .github/workflows/lint.yml
```

This is the `paths-ignore` docs, emphasis mine:

> When all the path names match patterns in paths-ignore, the workflow will not run. **If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.**

Now consider these cases:

- Only `src/foo.rs` changed
  - Matches first pattern of `paths`; normal workflow runs
  - Matches first pattern of `paths-ignore`; patch workflow does not run
- Only `book/foo.md` changed
  - Does not match any pattern of `paths`; normal workflow does not run
  - Does not match any patterns of `paths-ignore`; patch workflow does run
- Both `src/foo.rs` and `book/foo.md` have changed
  - `foo.rs` matches a pattern in `paths`, so normal workflow runs
  - `foo.md` does not match the patterns in `paths-ignore`; **the patch workflow does run**

Thus there are cases when both jobs run. It's not clear what GitHub does in that case, but it seems risky.

I'm not sure how to fix it.

The negation of the `paths` rule is is:

"If all paths do not match any pattern in the paths filter, the workflow does not run"

Therefore, the patch job criteria should be:

"If all paths do not match any pattern in the paths filter, the (patch) workflow runs"

But I don't see how to represent that using `paths` or `paths-ignore`.

Honestly it seems easier to remove `paths` altogether and always run the tests.

PR #9934, opened 2025-09-29
qa(suggestion): Remove one call to `connect_nodes`.

oxarbitrage:
## Motivation

There is an extra call to `connect_nodes()` inside `connect_nodes_bi()`. Related to https://github.com/ZcashFoundation/zebra/pull/9931

## Solution

- Remove the redundant call, since Zebra already connects both sides with a single `addnode` request.
- Reduce the startup delay from 10 seconds to 1 second.

I tested this change multiple times locally, and it generally works well. However, I did observe a failure when running the addnode test repeatedly. This may be due to remaining `TIME_WAIT` connections after the test ends, and the extra 10-second delay might have been masking that issue.

We should wait for connections to be fully closed at shutdown if this becomes a problem.


conradoplg (approved):
Amazing, thanks!

I do wonder why it worked with `time.sleep(10)`, but we don't need to figure it out right now

Issue #9939, opened 2025-09-30
Performance regression in initial sync


upbqdn:
PR #9539 seems to have introduced a performance regression in the initial sync. I ran fresh `zebrad` instances connected only to another, synced instance on localhost on Mainnet, and got these results:

```
first 10_000 blocks, commit: cfabd6bca606a330c2312f701ccc27680e5b0a40
________________________________________________________
Executed in   16.94 secs    fish           external
   usr time   15.84 secs    0.00 micros   15.84 secs
   sys time    1.07 secs  588.00 micros    1.07 secs
________________________________________________________
Executed in   22.35 secs    fish           external
   usr time   15.00 secs  549.00 micros   15.00 secs
   sys time    0.95 secs  247.00 micros    0.95 secs
________________________________________________________
Executed in   13.54 secs    fish           external
   usr time   15.90 secs  904.00 micros   15.90 secs
   sys time    1.01 secs    0.00 micros    1.01 secs

first 40_000 blocks, commit: cfabd6bca606a330c2312f701ccc27680e5b0a40
   ________________________________________________________
Executed in   80.11 secs    fish           external
   usr time  110.97 secs  581.00 micros  110.96 secs
   sys time    4.53 secs    0.00 micros    4.53 secs
________________________________________________________
Executed in   87.23 secs    fish           external
   usr time  111.25 secs  597.00 micros  111.25 secs
   sys time    4.48 secs  265.00 micros    4.48 secs
________________________________________________________
Executed in   86.10 secs    fish           external
   usr time  117.52 secs  609.00 micros  117.52 secs
   sys time    4.50 secs  273.00 micros    4.50 secs

=====================================================================

first 10_000 blocks, commit: 819aadb30cdf92b32470692af619dd0aff769107
________________________________________________________
Executed in   31.00 secs    fish           external
   usr time   33.67 secs   40.00 micros   33.67 secs
   sys time    1.09 secs  597.00 micros    1.09 secs
________________________________________________________
Executed in   34.01 secs    fish           external
   usr time   37.63 secs  223.00 micros   37.63 secs
   sys time    1.33 secs  653.00 micros    1.33 secs
________________________________________________________
Executed in   34.07 secs    fish           external
   usr time   34.58 secs  453.00 micros   34.58 secs
   sys time    1.09 secs  209.00 micros    1.09 secs

first 40_000 blocks, commit: 819aadb30cdf92b32470692af619dd0aff769107
________________________________________________________
[...]

PR #9951, opened 2025-10-03
fix(state): Avoid heap allocations in `expand_zero_be_bytes()`

arya2:
## Motivation

The logic for getting a value from the db generally avoids unnecessary allocations, this function showed up on a performance profile for reading an address balance.

Related to #9939.

## Solution

- Return an array from the function instead of allocating a vector, and
- Return early if `disk_bytes` is already the expected length.

### Tests

Should be covered by existing tests.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-state/src/service/finalized_state/disk_format.rs
+++ b/zebra-state/src/service/finalized_state/disk_format.rs
@@ -5,7 +5,7 @@
 //! [`crate::constants::state_database_format_version_in_code()`] must be incremented
 //! each time the database format (column, serialization, etc) changes.
 
-use std::sync::Arc;
+use std::{io::Write, sync::Arc};
 
 pub mod block;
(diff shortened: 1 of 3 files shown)


conradoplg (approved):
This is great, thanks! I was also looking into this but you went much further (I forgot about Height). I wrote a benchmark (might do a separate PR for it) and the results for this PR were pretty good

```
Benchmarking AddressBalanceLocationChange::from_bytes/zero: Collecting 100 samples in estimated 5.0000 s (2.8B iteratiAddressBalanceLocationChange::from_bytes/zero
                        time:   [1.7292 ns 1.7363 ns 1.7434 ns]
                        change: [-97.167% -97.156% -97.145%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #9968, opened 2025-10-08
fix(auth): Use `STANDARD` Base64 for RPC auth encoding/decoding

### LINKED ISSUE #9966: bug: zebra-rpc authorization is decoded with wrong Base64 encoding
### What happened?

https://github.com/ZcashFoundation/zebra/blob/15de82711b6429f9c77e81b26e7ef0bb92b76cdb/zebra-rpc/src/server/http_request_compatibility.rs#L74

[RFC 7617 Section 2](https://datatracker.ietf.org/doc/html/rfc7617#section-2) specifies the encoding, and references [RFC 4648 Section 4](https://datatracker.ietf.org/doc/html/rfc4648#section-4) which is plain Base64.

Implementing JSON-RPC authorization for Zallet.

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

oxarbitrage:
## Motivation

According to [RFC 7617 §2](https://datatracker.ietf.org/doc/html/rfc7617#section-2), HTTP Basic Authentication uses Base64 as defined in [RFC 4648 §4](https://datatracker.ietf.org/doc/html/rfc4648#section-4), which specifies the standard Base64 alphabet. Previously, Zebra used the `URL_SAFE` variant, which differs in character set and padding rules.

Closes [#9966](https://github.com/ZcashFoundation/zebra/issues/9966).

## Solution

Switched RPC auth encoding/decoding to use `STANDARD` instead of `URL_SAFE`.

Note that with this change, some commands that previously worked may no longer decode correctly. For example:

```bash
curl -X POST -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "method": "getinfo", "params": [], "id":123 }' \
  __cookie__:kpMRGXxEBy/riUrSkIb2LMTAMR4mRYcMdIswpFuYLz8=@127.0.0.1:6666 | jq
```

should now be written as:

```bash
curl -u "__cookie__:kpMRGXxEBy/riUrSkIb2LMTAMR4mRYcMdIswpFuYLz8=" \
  -X POST -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "method": "getinfo", "params": [], "id":123 }' \
  127.0.0.1:6666 | jq
 ```

It’s unclear to me whether this should be considered a breaking change, but users relying on the old behavior may need to update their commands.

### Tests

Manually.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-rpc/src/server/cookie.rs
+++ b/zebra-rpc/src/server/cookie.rs
@@ -1,6 +1,6 @@
 //! Cookie-based authentication for the RPC server.
 
-use base64::{engine::general_purpose::URL_SAFE, Engine as _};
+use base64::{engine::general_purpose::STANDARD, Engine as _};
 use color_eyre::Result;
 use rand::RngCore;
 
@@ -29,7 +29,7 @@ impl Default for Cookie {
         let mut bytes = [0u8; 32];
         rand::thread_rng().fill_bytes(&mut bytes);
 
-        Self(URL_SAFE.encode(bytes))
+        Self(STANDARD.encode(bytes))
     }
 }
 
--- a/zebra-rpc/src/server/http_request_compatibility.rs
+++ b/zebra-rpc/src/server/http_request_compatibility.rs
@@ -19,7 +19,7 @@ use tower::Service;
 
 use super::cookie::Cookie;
 
-use base64::{engine::general_purpose::URL_SAFE, Engine as _};
+use base64::{engine::general_purpose::STANDARD, Engine as _};
 
 /// HTTP [`HttpRequestMiddleware`] with compatibility workarounds.
 ///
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: zebra-rpc/src/server/cookie.rs [hunk 1]
```diff
@@ -29,7 +29,7 @@ impl Default for Cookie {
         let mut bytes = [0u8; 32];
         rand::thread_rng().fill_bytes(&mut bytes);
 
-        Self(URL_SAFE.encode(bytes))
```
conradoplg, line 32: I don't think this is strictly needed; the bug is on how we parse the `Authentication` header. This is the _value_ that is eventually encoded into the `Authentication` header by other tools like `curl`. So this doesn't need to change, though it's harmless. (So feel free to not change it)


(It's confusing because there are two base64 encodings: here, to get an ASCII string out of a random byte string to act as a `password`, and inside the HTTP protocol where the `user:password` is base64-encoded, and this is the one we got wrong. It doesn't matter to the protocol what the password is like as long it is ASCII.

conradoplg (approved):
Thanks!

I don't think this is a breaking change (I think the first command you posted should still work? e.g. this command worked on this PR:

```
curl --data-binary '{"jsonrpc": "2.0", "id":"curltest", "method": "getbestblockheightandhash", "params": [] }' -H 'Content-type: application/json' http://$(cat ~/.cache/zebra/.cookie)@127.0.0.1:28232/
```

It just fixes some cases where some cookie files would not work because they would never get decoded correctly (since they were encoded by other tools like `curl`)

zebra-rpc/src/server/cookie.rs
conradoplg: I'm marking as solved to get it merged :laughing:

PR #9923, opened 2025-09-25
Add(state): Adds a `MappedRequest` helper trait and refactors error types used by `CommitSemanticallyVerifiedBlock` requests

arya2:
## Motivation

This PR refactors the improved error types added in https://github.com/ZcashFoundation/zebra/pull/9848 into an enum.

## Solution

- Updates `CommitSemanticallyVerifiedError` to be an enum instead of a new type around a boxed `ValidateContextError`
- Moves new variants of `ValidateContextError` to a new `QueueAndCommitError` enum
- Updates documentation of `Request::CommitSemanticallyVerifiedBlock` to mention the expected error type
- Adds a code comment above the match arm for `CommitSemanticallyVerifiedBlock` to mention the expected error type
- Adds a `MappedRequest` helper trait and a new request type which implements it for convenient access to expected response/error types

### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [x] The library crate changelogs are up to date.
- [x] The solution is tested.
- [x] The documentation is up to date.

PR #9848, opened 2025-08-28
refactor(state): improve error propagation for `CommitSemanticallyVerifiedBlock`

syszery:
## Motivation

This PR addresses #9730.

It focuses on improving the precision of error types returned when committing semantically verified blocks to the non-finalized state, as outlined in the issue.

Feedback specifically welcome on the design and handling of validation errors introduced in this PR.

## Solution

- Used and extended `CommitSemanticallyVerifiedError` already defined in `error.rs`.
- Replace the boxed error type in the `CommitSemanticallyVerifiedBlock` response with the concrete `CommitSemanticallyVerifiedError` type.
- Refactor `StateService::call`, `queue_and_commit_to_non_finalized_state`, and related functions to propagate this error.
- Adjust the channels and error handling in `write.rs`.

### Tests

- All existing tests passed successfully.

### Specifications & References

#9730 

### Follow-up Work

- Review and potentially refactor other parts of `StateService` in `zebra-state` for consistent error handling.
- Consider adding more targeted comments or clarifying documentation (as outlined in the issue).
- Consider adding more tests.
- Feedback welcomed on whether this direction makes sense before proceeding with broader cleanup.

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [ ] The solution is tested.
- [ ] The documentation is up to date.

Issue #5732, opened 2022-11-28
Move BoxError downcast out of `impl From<BoxError> for TransactionError`


arya2:
## Motivation

It would be better for the compiler to complain about a mismatch between BoxError / TransactionError when using the `?` operator, and to perform the conversions where the type of `BoxError` is more explicitly known.

`TransactionError` is currently the only error type with a `From` impl handling downcasts.

This could be used in the block verifier for `VerifyBlockError::ValidateProposal` as well if defined in a trait.

## Possible Design

```rust
impl TransactionError {
    pub(crate) fn downcast_from<ErrorType>(mut err: BoxError) -> Self
    where
        ErrorType: std::error::Error + Into<TransactionError>,
    {
        err.downcast::<ErrorType>()
            .map(|e| Self::from(*e))
            .or_else(Self::internal_downcast_err)
    }
}
```

## Related Work

- #1186

teor2345:
These downcasts are located in the same place, so we don't accidentally implement conflicting downcasts. Moving the downcasts out of the conversion could make the code harder to maintain, and create more bugs.

> It would be better for the compiler to complain when there isn't a valid conversion for a particular type of error and to perform the conversions where the type of `BoxError` is more explicitly known.

Unfortunately, it is impossible to know the type of many errors, because `tower::Buffer` converts all service readiness and response errors to `BoxError`:
https://docs.rs/tower/latest/tower/buffer/struct.Buffer.html#impl-Service%3CRequest%3E

Before we can prioritise or estimate this ticket, it will need a specific design that takes these issues into account.

arya2:
> Unfortunately, it is impossible to know the type of many errors, because tower::Buffer converts all service readiness and response errors to BoxError

Ah, I see. Still, it might be a good idea to do the conversion outside a `From` impl so it's more obvious where the conversions are happening. I added a possible design.

teor2345:
> It would be better for the compiler to complain when there isn't a valid conversion for a particular type of error

I don't think the design in the ticket does this?

arya2:
> > It would be better for the compiler to complain when there isn't a valid conversion for a particular type of error
> 
> I don't think the design in the ticket does this?

My wording there may have been a bit vague, it's more to disallow using `?` or `into()` without explicitly defining how the `BoxError` should be converted into a `TransactionError` so it's more obvious when a new variant needs to be added to `TransactionError`.

It might also be nice to annotate where each type is expected.

[...]

PR #9979, opened 2025-10-09
refactor(state): improve error propagation for CommitCheckpointVerifiedBlock

### LINKED ISSUE #9730: cleanup: Return more concrete errors from the state service

syszery:
## Motivation

This PR addresses #9730 and focuses on the `CommitCheckpointVerifiedBlock` path.

## Solution

* Refactors the errors in `CommitCheckpointVerifiedBlock` following the example from #9848 and #9923.
* Introduces the new error enum `CommitCheckpointVerifiedError` in `error.rs`.
* Implements `MappedRequest` for `CommitCheckpointVerifiedBlock`.
* Adds documentation for the new error type and its usage across relevant modules.

### Follow-up Work

* **Should `CloneError` be replaced?**
  `CloneError` was introduced due to [`commit_finalized_direct`](https://github.com/ZcashFoundation/zebra/blob/d1ce5e59b40ac09ca951b9d2c1a71f8037f11594/zebra-state/src/service/finalized_state.rs#L327) returning a `BoxError`. Replacing it will require additional changes to the error like implementing `From<zebra_chain::parallel::tree::NoteCommitmentTreeError>`. 
* **Should `CommitCheckpointVerifiedError` wrap `QueueAndCommitError`?**
  There is some overlap between these error types, particularly with regard to the error messages. Adopting `QueueAndCommitError` would simplify error handling, but this would require modifying some of the existing error messages to match.

Happy to adjust based on feedback.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -106,6 +106,34 @@ impl<E: std::error::Error + 'static> From<BoxError> for LayeredStateError<E> {
     }
 }
 
+/// An error describing why a `CommitCheckpointVerifiedBlock` request failed.
+#[derive(Debug, Error, Clone)]
+#[non_exhaustive]
+pub enum CommitCheckpointVerifiedError {
+    /// An older checkpoint block was dropped because of a newer duplicate block.
+    #[error("dropping older checkpoint verified block: got newer duplicate block")]
+    ReplacedByNewer,
+
+    /// A checkpoint block was dropped because it was already committed to the finalized state.
+    #[error("already finished committing checkpoint verified blocks: dropped duplicate block, block is already committed to the state")]
+    DroppedAlreadyCommitted,
+
+    /// The block was dropped from the queue of finalized blocks.
+    #[error("block was dropped from the queue of finalized blocks")]
+    DroppedFromFinalizedQueue,
+
+    /// The commit task exited (likely during shutdown).
+    #[error("block commit task exited. Is Zebra shutting down?")]
+    CommitTaskExited,
+
+    /// The write task exited (likely during shutdown).
+    #[error("block write task has exited. Is Zebra shutting down?")]
+    WriteTaskExited,
+
+    #[error("clone error: {0}")]
+    CloneError(#[from] CloneError),
+}
+
 /// An error describing why a `InvalidateBlock` request failed.
 #[derive(Debug, Error)]
 #[non_exhaustive]
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -46,8 +46,11 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 #[derive(Debug, Error, Clone, PartialEq, Eq, new)]
 pub enum CommitBlockError {
     #[error("block hash is a duplicate: already in {location}")]
+    /// The block is a duplicate: it is already queued or committed in the state.
     Duplicate {
+        /// Hash or height of the duplicated block.
         hash_or_height: Option<HashOrHeight>,
+        /// Location in the state where the block can be found.
         location: KnownBlock,
     },
 
@@ -55,6 +58,7 @@ pub enum CommitBlockError {
     #[error("could not contextually validate semantically verified block")]
     ValidateContextError(#[from] ValidateContextError),
 
+    /// The write task exited (likely during shutdown).
     #[error("block commit task exited. Is Zebra shutting down?")]
     #[non_exhaustive]
     WriteTaskExited,
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -130,7 +130,7 @@ pub enum CommitCheckpointVerifiedError {
     #[error("block write task has exited. Is Zebra shutting down?")]
     WriteTaskExited,
 
-    #[error("clone error: {0}")]
+    #[error("{0}")]
     CloneError(#[from] CloneError),
 }
 
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -64,6 +64,14 @@ pub enum CommitBlockError {
     WriteTaskExited,
 }
 
+impl CommitBlockError {
+    /// Returns `true` if this is definitely a duplicate commit request.
+    /// Some duplicate requests might not be detected, and therefore return `false`.
+    pub fn is_duplicate_request(&self) -> bool {
+        matches!(self, CommitBlockError::Duplicate { .. })
+    }
+}
+
 /// An error describing why a `CommitSemanticallyVerified` request failed.
 #[derive(Debug, Error, Clone, PartialEq, Eq)]
 #[error("could not commit semantically-verified block")]
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -49,6 +49,10 @@ pub enum QueueAndCommitError {
     #[non_exhaustive]
     Duplicate { block_hash: block::Hash },
 
+    #[error("already finished committing checkpoint verified blocks: dropped duplicate block, block is already committed to the state")]
+    #[non_exhaustive]
+    AlreadyCommitted,
+
     #[error("block height {block_height:?} is already committed in the finalized state")]
     #[non_exhaustive]
     AlreadyFinalized { block_height: block::Height },
@@ -56,6 +60,10 @@ pub enum QueueAndCommitError {
     #[error("block hash {block_hash} was replaced by a newer commit request")]
     #[non_exhaustive]
     Replaced { block_hash: block::Hash },
+
+    #[error("dropping older checkpoint verified block {block_hash}: got newer duplicate block")]
+    #[non_exhaustive]
+    ReplacedByNewer { block_hash: block::Hash },
 
     #[error("pruned block at or below the finalized tip height: {block_height:?}")]
     #[non_exhaustive]
@@ -110,26 +118,19 @@ impl<E: std::error::Error + 'static> From<BoxError> for LayeredStateError<E> {
 #[derive(Debug, Error, Clone)]
 #[non_exhaustive]
 pub enum CommitCheckpointVerifiedError {
-    /// An older checkpoint block was dropped because of a newer duplicate block.
-    #[error("dropping older checkpoint verified block: got newer duplicate block")]
-    ReplacedByNewer,
-
-    /// A checkpoint block was dropped because it was already committed to the finalized state.
-    #[error("already finished committing checkpoint verified blocks: dropped duplicate block, block is already committed to the state")]
-    DroppedAlreadyCommitted,
-
-    /// The block was dropped from the queue of finalized blocks.
-    #[error("block was dropped from the queue of finalized blocks")]
-    DroppedFromFinalizedQueue,
-
-    /// The commit task exited (likely during shutdown).
-    #[error("block commit task exited. Is Zebra shutting down?")]
-    CommitTaskExited,
-
+    /// Queuing/commit step failed.
+    #[error("could not queue and commit semantically verified block")]
+    QueueAndCommitError(#[from] QueueAndCommitError),
+    /// RocksDB write failed
+    /// TODO: Wire it up in `finalized_state.rs`, `commit_finalized`
+    #[error("could not write checkpoint-verified block to RocksDB")]
+    WriteError(#[from] rocksdb::Error),
     /// The write task exited (likely during shutdown).
     #[error("block write task has exited. Is Zebra shutting down?")]
     WriteTaskExited,
 
+    /// Temporary workaround for boxed errors.
+    /// TODO: Replace with `WriteError(#[from] rocksdb::Error)`
     #[error("{0}")]
     CloneError(#[from] CloneError),
 }
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -45,7 +45,7 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 /// An error describing why a block could not be queued to be committed to the state.
 #[derive(Debug, Error, Clone, PartialEq, Eq, new)]
 pub enum CommitBlockError {
-    #[error("block hash has already been sent to be committed to the state")]
+    #[error("block hash is a duplicate: already in {location}")]
     Duplicate {
         hash_or_height: Option<HashOrHeight>,
         location: KnownBlock,
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -95,13 +95,9 @@ pub enum CommitCheckpointVerifiedError {
     #[error("could not queue and commit checkpoint-verified block")]
     CommitBlockError(#[from] CommitBlockError),
     /// RocksDB write failed
-    /// TODO: Wire it up in `finalized_state.rs`, `commit_finalized`
+    /// TODO: Decide whether to expect that writes will succeed (it's already expecting that there won't be a write error when finalizing non-finalized blocks)
     #[error("could not write checkpoint-verified block to RocksDB")]
     WriteError(#[from] rocksdb::Error),
-    /// Temporary workaround for boxed errors.
-    /// TODO: Replace with `ValidateContextError(#[from] ValidateContextError)`
-    #[error("{0}")]
-    CloneError(#[from] CloneError),
 }
 
 impl From<ValidateContextError> for CommitCheckpointVerifiedError {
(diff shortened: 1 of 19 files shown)


zebra-state/src/error.rs
```diff
@@ -106,6 +106,34 @@ impl<E: std::error::Error + 'static> From<BoxError> for LayeredStateError<E> {
...
+    #[error("block was dropped from the queue of finalized blocks")]
+    DroppedFromFinalizedQueue,
+
+    /// The commit task exited (likely during shutdown).
+    #[error("block commit task exited. Is Zebra shutting down?")]
+    CommitTaskExited,
+
+    /// The write task exited (likely during shutdown).
+    #[error("block write task has exited. Is Zebra shutting down?")]
+    WriteTaskExited,
```
arya2: It would update the error message, but, these should really be merged.

zebra-state/src/error.rs
```diff
@@ -106,6 +106,34 @@ impl<E: std::error::Error + 'static> From<BoxError> for LayeredStateError<E> {
...
+
+    #[error("clone error: {0}")]
```
arya2: To avoid changing the error message if we're keeping the box errors for now:

```suggestion
    #[error("{0}")]
```

arya2 (dismissed):
This looks good, one small suggestions to avoid changing the error messages if you want to keep a `CloneError` variant in the new error enum.

> Should CloneError be replaced?

It would be great if it were replaced there, though it looks like the change may be substantial if you want this to merge as-is.

> Should CommitCheckpointVerifiedError wrap QueueAndCommitError?

I think that would be nice, `CommitCheckpointVerifiedError` could essentially be a copy of `CommitSemanticallyVerifiedError` with an added `WriteError(rocksdb::Error)` variant.

syszery:
Hi @arya2, thank you for the feedback.

I’ve restructured `CommitCheckpointVerifiedError` to follow the same model as `CommitSemanticallyVerifiedError`, as suggested. I also stubbed in `WriteError(#[from] rocksdb::Error)` for the finalized path (currently unused). All relevant error cases are now represented in `QueueAndCommitError`.

Some variants between the finalized and non-finalized paths are *close* in meaning, but not identical — so a full unification would require either small semantic shifts or renaming a few arms. Also, e.g., `QueueAndCommitError::Dropped` isn't currently used — would you prefer to keep these around for now?

If you already have a naming or structural direction in mind, feel free to push it — I’m happy to clean up any resulting breakages or follow-up changes.

**TODOs before merge:**

* Attempt full removal of `BoxError`.
* Fix documentation for the previously merged `InvalidateError` and `ReconsiderError` in `service.rs`.
* (?) Update `CHANGELOG.md` to highlight breaking changes.

zebra-state/src/error.rs
```diff
@@ -49,6 +49,10 @@ pub enum QueueAndCommitError {
     #[non_exhaustive]
     Duplicate { block_hash: block::Hash },
 
+    #[error("already finished committing checkpoint verified blocks: dropped duplicate block, block is already committed to the state")]
+    #[non_exhaustive]
+    AlreadyCommitted,
```
arya2: Optional: I think it would be nice to merge the `AlreadyCommitted`, `Replaced` and `ReplacedByNewer` variants into `Duplicate` and add a `source` or `location` field to `Duplicate` specifying if the duplicate was found in the queued blocks, sent blocks (`SentHashes`), non-finalized state, or finalized state. Feel free to resolve this comment if you want to leave that for a follow-up PR.

zebra-state/src/service.rs
```diff
@@ -495,12 +498,10 @@ impl StateService {
             //       every time we send some blocks (like QueuedSemanticallyVerifiedBlocks)
             Self::send_checkpoint_verified_block_error(
                 queued,
-                CommitCheckpointVerifiedError::DroppedAlreadyCommitted,
+                QueueAndCommitError::AlreadyCommitted.into(),
```
arya2: Update: I made these changes in the suggestion branch.

Optional: We could get rid of the verbose `.into()` calls by updating `send_checkpoint_verified_block_error()` and `clear_finalized_block_queue()` to accept any argument that implements `Into<CommitCheckpointVerifiedError>`, like:

```rust

    /// Drops all finalized state queue blocks, and sends an error on their result channels.
    fn clear_finalized_block_queue(
        &mut self,
        error: impl Into<CommitCheckpointVerifiedError> + Clone,
    ) {
        for (_hash, queued) in self.finalized_state_queued_blocks.drain() {
            Self::send_checkpoint_verified_block_error(queued, error.clone());
        }
    }

    /// Send an error on a `QueuedCheckpointVerified` block's result channel, and drop the block
    fn send_checkpoint_verified_block_error(
        queued: QueuedCheckpointVerified,
        error: impl Into<CommitCheckpointVerifiedError>,
    ) {
        let (finalized, rsp_tx) = queued;

        // The block sender might have already given up on this block,
        // so ignore any channel send errors.
        let _ = rsp_tx.send(Err(error.into()));
        std::mem::drop(finalized);
    }
```

arya2 (commented):
This is looking great!

I started making some changes based off this branch in #9999, where the `write_block()` method has been updated to return a `CommitCheckpointVerifiedError`. I think there may still be some room for improvement in the `Duplicate` error variant. I'm happy to work on that a bit more or for you to takeover from there, let me know what you would prefer.

Aside from improving the `Duplicate` error variant, these changes will likely require some updates to `ChainSync::should_restart_sync()`.

> Some variants between the finalized and non-finalized paths are close in meaning, but not identical — so a full unification would require either small semantic shifts or renaming a few arms.

I think we should fully unify them and have both request variants return a `CommitBlockError`, the only substantial difference is that committing checkpoint-verified blocks could return a `rocksdb::Error`, but only because those are expected not to happen when finalizing non-finalized blocks, so I think it would be fine to expect that they won't happen for checkpoint-verified blocks either. It doesn't seem like Zebra could recover from those errors anyway, see ["an error during a write operation (write to WAL, Memtable Flush, background compaction etc) may cause the database instance to go into read-only mode and further user writes may not be accepted."](https://github.com/facebook/rocksdb/wiki/Background-Error-Handling)

> Also, e.g., QueueAndCommitError::Dropped isn't currently used — would you prefer to keep these around for now?

If they're not currently necessary, it seems easy enough to add them once they're needed.

syszery:
> Quoting @arya2 from #9999:
>
> > #### TODO:
> >
> > * The `Duplicate` error variant could likely be improved, I'm happy to try doing that here,
> > * Error messages should likely be adjusted to more closely match existing error messages, and
> > * `ChainSync::should_restart_sync()` will likely need some updates to continue working as expected

I tried to beautify the new `Duplicate` error and deliver a simple version that fits the current style of the module. However, formatting the `hash_or_height` is a bit tedious due to the `Option` type — especially since the `None` case needs to be handled explicitly for clean error messages.

Currently, the `None` case only appears in `service.rs`, where there are `TODO`s hinting that `block_height` tracking is planned. Would you prefer to leave this as a `TODO` for follow-up, or should we address it in this PR?

---

On a related note: I noticed that `BoxError` and `CommitSemanticallyVerifiedError` are re-exported in `lib.rs`, while the other error types are not. I'd like to unify this approach for consistency across existing and upcoming variants.

Do you prefer that I do this cleanup as part of this PR, or would you rather wait and revisit it once the `StateService` error structure is more finalized?

syszery:
Yesterday, after a rebase, I had some issues with the acceptance tests in `zebrad`, particularly with `activate_mempool_mainnet`. Now, I just re-ran all tests again, and they all passed on the first try.

We’ve been working on parts of the code that might affect mempool logic, so I’m wondering — is there anything I should take a closer look at, or anything in particular to watch out for?

syszery:
Hi @arya2, I patched the `should_restart_sync` method you mentioned. The duplicate/commit checks are now consolidated in a single `Invalid { error, .. }` arm, which handles:

* `error.is_duplicate_request()`
* `CommitBlockError::Duplicate`
* fallback string matches (for now)

This commit only fixes a small part of #2908 — the fallback string matching is still a TODO, I think.

zebra-state/src/error.rs
syszery: addressed by merging #9999

syszery:
After re-visiting the method, I made it a bit more defensive: it now checks for `CommitBlockError::Duplicate` in the `_ =>` fallback branch, while keeping the previous string matches.

Do you want to merge this “safe but messy” fix first and clean it up later (e.g., by implementing `is_duplicate()` on `CommitBlockError`), essentially addressing issue #2908? Or as part of this PR?

arya2:
> Currently, the `None` case only appears in `service.rs`, where there are `TODO`s hinting that `block_height` tracking is planned. Would you prefer to leave this as a `TODO` for follow-up, or should we address it in this PR?

I would prefer to address it in this PR if it helps make the `Option<HashOrHeight>` a `HashOrHeight`, but it would be fine to do it in a follow-up PR too.

> On a related note: I noticed that `BoxError` and `CommitSemanticallyVerifiedError` are re-exported in `lib.rs`, while the other error types are not. I'd like to unify this approach for consistency across existing and upcoming variants.
> 
> Do you prefer that I do this cleanup as part of this PR, or would you rather wait and revisit it once the `StateService` error structure is more finalized?

Whichever you prefer, I see no urgency to do it now nor much benefit to deferring it.

Yesterday, after a rebase, I had some issues with the acceptan

> Yesterday, after a rebase, I had some issues with the acceptance tests in zebrad, particularly with activate_mempool_mainnet. 
Now, I just re-ran all tests again, and they all passed on the first try.

> We’ve been working on parts of the code that might affect mempool logic, so I’m wondering — is there anything I should take a closer look at, or anything in particular to watch out for?

No, these changes should have no effect on the mempool, that test and a handful of other acceptance tests just fail occasionally, it's not an issue unless they fail consistently or are relevant to the changes and fail in a way that could be due to a concurrency bug.

> Do you want to merge this “safe but messy” fix first and clean it up later (e.g., by implementing is_duplicate() on CommitBlockError), essentially addressing issue https://github.com/ZcashFoundation/zebra/issues/2908? Or as part of this PR?

I want to address #2908 as part of this PR, we've been wanting to address it for a long time, seems convenient to do it here where it makes the changes easier to review, and it would be nice to merge this just after tagging the 3.0.0 release anyway so that there's plenty of time to test it before it gets tagged in a release. Let me know if you'd like me to make those changes.

Thank you for making all of these changes, they're looking excellent. I'll closely review everything again soon.

syszery:
Thank you @arya2 for all the explanations and guidance! I really appreciate your help. I agree that we shouldn’t be lazy now 😉 — both TODOs seem manageable and in range to get closed here.

syszery:
Here are the TODOs we’ve identified.

* [x] **Address state error path of issue #2908**

* [ ] ~~**Try to remove the `Option` from `hash_or_height`** — this requires addressing the following TODOs in the `StateService` implementation, where the `None` is currently used:~~
  * [ ] ~~`queue_and_commit_to_finalized_state`: Track the latest sent height and drop any blocks below it when sending new blocks.~~
  * [ ] ~~`queue_and_commit_to_non_finalized_state`: It seems harder to blame a single block here, so the `None` case may not be meaningfully removable.~~

* [ ] ~~**Beautify `CommitBlockError::Duplicate`** — this will be straightforward once the `Option` is removed.~~

* [ ] Minor cleanup: unify error re-exports in lib.rs.

syszery:
I have reverted the changes to `should_restart_sync` and implemented the following.

### Update:

* Implemented `is_duplicate_request` for `CommitBlockError`.
* Updated `VerifyBlockError::Commit` to properly wrap `CommitBlockError` and introduced a `StateService` variant.
* Modified `Service::call` to propagate `CommitBlockError`.
* Added a new unit test verifying that a `BlockDownloadVerifyError::Invalid` wrapping a `CommitBlockError::Duplicate` doesn't trigger a restart.
* Cleaned up `should_restart_sync`.

### Questions for review:

* Do these changes fully address the state-related portion of issue #2908?
* Should we address `VerifyCheckpointError` or issue #5732 now, or can those be left for a separate refactor?

zebra-state/src/request.rs
```diff
@@ -690,6 +690,28 @@ impl MappedRequest for CommitSemanticallyVerifiedBlockRequest {
     }
 }
 
+/// Commit a checkpointed block to the state
```
arya2: Nitpick:

```suggestion
/// Commit a checkpoint-verified block to the state
```

zebra-state/src/service/finalized_state.rs
```diff
@@ -299,15 +300,9 @@ impl FinalizedState {
                 .set(checkpoint_verified.height.0 as f64);
         };
 
-        // Make the error cloneable, so we can send it to the block verify future,
-        // and the block write task.
-        let result = result.map_err(CloneError::from);
+        let _ = rsp_tx.send(result.clone().map(|(hash, _)| hash));
```
arya2, line 303: Cloning the note commitment trees is currently harmless but unnecessary and could become a problem if the struct is updated.

arya2 (dismissed):
This looks good to merge!

[Would be nice to use the new `MappedRequest` trait and unify the error types in a follow-up PR](https://github.com/ZcashFoundation/zebra/tree/suggestion-for-cv-commit-state-error).

syszery:
Hi @arya2, thanks for the feedback — I’ve applied your suggestion. I believe `MappedRequest` is already implemented for the `CommitCheckpointVerifiedBlock` path as part of this PR. 

For context, this PR likely doesn’t fully resolve #9730 yet: `AwaitUtxo` is still a TODO in `StateService`, and we haven’t touch `ReadStateService` yet.

arya2 (dismissed):

syszery:
Hi @arya2, I looked into the Clippy lint and it seems to be triggered by `CommitCheckpointVerifiedError` wrapping `CommitBlockError`, which still contains the (large) `ValidateContextError` variant. I see a few ways to handle this:

1. Allow the lint locally (`#[allow(clippy::result_large_err)]`)
2. Box the error at the return boundary (`Result<T, Box<CommitCheckpointVerifiedError>>`) as suggested by the error message
3. Box `CommitBlockError` or `ValidateContextError` inside the enum

Refactoring `ValidateContextError` is another option, though not sure how worth it that would be.

Happy to go with whatever you think makes the most sense.

arya2:
@syszery I think we should box `ValidateContextError`s in the enum, it seems fine to allow the lint here now because it's just a little over the limit and should be constructed fairly rarely, but ignoring the lint would open the door to changes that could make the type much larger.

syszery:
Hi @arya2, I’ve boxed `ValidateContextError` as we discussed. I also removed a redundant clone in `write.rs`. All tests pass locally, and there are no more clippy lints.

arya2 (approved):

PR #9981, opened 2025-10-09
fix(health): Avoid repeated warnings about the chain metrics channel being closed when the healthcheck server is disabled

arya2:
## Motivation

This PR updates the progress task to avoid repeatedly logging warnings when there is no healthcheck server.

## Solution

- Adds a `is_chain_metrics_chan_closed` flag and sets it to true if sending to the chain metrics channel fails
- Avoids sending to the chain metrics channel when the flag is set to true

### Tests

This has been manually tested to check that it avoids repeated warnings.

### PR Checklist

- [x] The solution is tested.


oxarbitrage (approved):
Yes, i saw this and it is pretty annoying. Thanks for the fix.

PR #9999, opened 2025-10-15
Suggestion for "refactor(state): improve error propagation for CommitCheckpointVerifiedBlock #9979"

### LINKED ISSUE #9998: Update return types of infallible `DiskWriteBatch` methods
## Motivation

Some of the methods on DiskWriteBatch` used to prepare a batched write of a new block to the finalized state are infallible but return a `Result`, we could simplify the code by having them return nothing.

### Related

The method docs on `prepare_trees_batch()` say that it propagates errors to callers, but there appear to be no errors to return. Update the documentation on these methods too to reflect their current implementations.

arya2:
This is a suggestion PR for #9979 to:

- Update the `KnownBlock` struct and state request to indicate when blocks are found in the sent hashes recording what's been sent to a block write channel (fixing a bug where it's occasionally possible for Zebra to add a non-finalized chain to its chain set which is a duplicate portion of another non-finalized chain),
- Updates `write_block()` to return `CommitCheckpointVerifiedError`s instead of  `BoxError`s
- Condenses several error variants in `QueueAndCommitError` into a `Duplicate` variant
- Renames `QueueAndCommitError` to `BlockCommitError`
- Unifies `CommitSemanticallyVerifiedError` and `CommitCheckpointVerifiedError` types, making them both new types around `BlockCommitError`

These changes could close #9998.

#### TODO:
- The `Duplicate` error variant could likely be improved, I'm happy to try doing that here,
- Error messages should likely be adjusted to more closely match existing error messages, and
- `ChainSync::should_restart_sync()` will likely need some updates to continue working as expected

## Review

@syszery can review these changes and merge them if they look good. The base branch is a copy of `syszery:refactor/state-CommitCheckpointVerifiedBlock-error` so this PR shows a useful diff.

### DIFF
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -95,13 +95,9 @@ pub enum CommitCheckpointVerifiedError {
     #[error("could not queue and commit checkpoint-verified block")]
     CommitBlockError(#[from] CommitBlockError),
     /// RocksDB write failed
-    /// TODO: Wire it up in `finalized_state.rs`, `commit_finalized`
+    /// TODO: Decide whether to expect that writes will succeed (it's already expecting that there won't be a write error when finalizing non-finalized blocks)
     #[error("could not write checkpoint-verified block to RocksDB")]
     WriteError(#[from] rocksdb::Error),
-    /// Temporary workaround for boxed errors.
-    /// TODO: Replace with `ValidateContextError(#[from] ValidateContextError)`
-    #[error("{0}")]
-    CloneError(#[from] CloneError),
 }
 
 impl From<ValidateContextError> for CommitCheckpointVerifiedError {
--- a/zebra-state/src/error.rs
+++ b/zebra-state/src/error.rs
@@ -15,7 +15,7 @@ use zebra_chain::{
     work::difficulty::CompactDifficulty,
 };
 
-use crate::constants::MIN_TRANSPARENT_COINBASE_MATURITY;
+use crate::{constants::MIN_TRANSPARENT_COINBASE_MATURITY, HashOrHeight, KnownBlock};
 
 /// A wrapper for type erased errors that is itself clonable and implements the
 /// Error trait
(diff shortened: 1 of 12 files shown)


syszery:
Thanks @arya2 — I really appreciate you stepping in and making these changes.

I just fixed a Clippy lint and removed one more `expect` in `zebra-state/src/service/finalized_state/disk_format/upgrade/drop_tx_locs_by_spends.rs` that was no longer needed. I can confirm that it builds cleanly without warnings or errors, and all existing tests pass on my side.

Removing the unused `Result<_,BoxError>` in the `zebra_db` modules helps a lot with the error refactor — thanks for simplifying that part, but you mean that this might introduce problems with `ChainSync::should_restart_sync()`? I am not sure but this sounds as if it isn't so trivial to test (in particular not by just running the existing Zebra tests)?

I also really like the new `CommitBlockError::Duplicate` variant — it cuts down verbosity and avoids repeated logic across the commit code paths. I suspect I helped introduce that over-specificity earlier by trying to keep every existing error message 😅

Regarding your TODO: Were you thinking of incorporating the fields into the error message, like below?

```Rust
#[error("block {hash_or_height:?} is already known in the state as {location:?}")]
Duplicate {
    hash_or_height: Option<HashOrHeight>,
    location: KnownBlock,
}
```

Or were you referring more to the overall verbosity of other enums like `ValidateContextError`, which might still need some cleanup (maybe even has a few unused variants)? Or `DuplicateNullifierError`?

I'm happy to merge this as suggested and pick up from here, if that's ok for you.

arya2:
> .. you mean that this might introduce problems with `ChainSync::should_restart_sync()`? I am not sure but this sounds as if it isn't so trivial to test (in particular not by just running the existing Zebra tests)?

I'm less worried about testing it automatically than just carefully reviewing it to make sure that it's correct, these changes should make it more robust as now it can match on concrete error types/variants where before it was checking if the error messages contained some strings.

> I also really like the new `CommitBlockError::Duplicate` variant — it cuts down verbosity and avoids repeated logic across the commit code paths. I suspect I helped introduce that over-specificity earlier by trying to keep every existing error message 😅

I like the new variant too! 🎉 Those duplicates were there already in the form of several error message templates that all meant the same thing, your changes made it a little better but also just highlighted the redundancy.

> Regarding your TODO: Were you thinking of incorporating the fields into the error message, like below?
> 
> ```rust
> #[error("block {hash_or_height:?} is already known in the state as {location:?}")]
> Duplicate {
>     hash_or_height: Option<HashOrHeight>,
>     location: KnownBlock,
> }
> ```

Yep! I was also staring at the `hash_or_height` field and wondering if the possible values the variant could represent could be reduced or made more explicit, particularly while re-using `KnownBlock`. I'll think about it some more when re-reviewing your PR.

> Or were you referring more to the overall verbosity of other enums like `ValidateContextError`, which might still need some cleanup (maybe even has a few unused variants)? Or `DuplicateNullifierError`?

No, I agree that that could be cleaned up a little too, but it seems lower-priority and easier to change independently/separately later.

> I'm happy to merge this as suggested and pick up from here, if that's ok for you.

Absolutely 👍.

syszery:
Hi @arya2, thank you for the detailed feedback. I will make a TODO list to sum up all the problems we identified.

Issue #9665, opened 2025-06-26
Define supertraits for types implementing `Service` traits


arya2:
## Motivation

It would be nice to define supertraits of commonly used trait constraints to avoid duplicate declarations of those constraints across types, functions, and implementations.

There are almost 150 lines of these constraint declarations just in [`zebra_rpc::methods`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/src/methods.rs#L748-L786), and many more in the workspace.

De-duplicating constraint declarations would improve readability and help to reduce the frequency/severity of merge conflicts

### Example

We could define the constraints that apply to types implementing the state service once in `zebra-state` and apply them wherever a generic parameter implementing the state service is accepted.

More generally, these could be defined wherever the required request/response types are defined.

In the transaction verifier:

```rust
trait StateService:
    Service<zs::Request, Response = zs::Response, Error = BoxError, Future: Send + 'static>
    + Send
    + Clone
    + 'static
{
}

impl<T> StateService for T where
    T: Service<zs::Request, Response = zs::Response, Error = BoxError, Future: Send + 'static>
        + Send
        + Clone
        + 'static
{
}


impl<ZS, Mempool> Service<Request> for Verifier<ZS, Mempool>
where
    ZS: StateService
..
```

arya2:
This was added a long time ago (ZebraService trait)

PR #10010, opened 2025-10-20
refactor: add StateService/ReadStateService aliases

syszery:
## Motivation

In parts of the Zebra RPC crate, the same verbose trait bounds for state services are repeatedly specified. This PR addresses issue #9665 by introducing concise super traits to simplify these bounds.
 

## Solution

This PR introduces two trait aliases:

* `StateServiceTrait` for services handling state-modifying requests (`zebra_state::Request`)
* `ReadStateServiceTrait` for services handling read-only state requests (`zebra_state::ReadRequest`)
* These new traits are then applied in `methods.rs`. 

All tests pass successfully.

Please let me know if this matches the expected approach. I’m happy to adjust based on any feedback.

### PR Checklist

- [x] The solution is tested.


zebra-state/src/service_traits.rs
```diff
@@ -0,0 +1,52 @@
...
+//!
+//! They enforce bounds such as `Clone`, `Send`, `Sync` and require the
+//! associated futures to be `Send`, making them suitable for
+//! multi-threaded async contexts.
+
+use crate::{BoxError, ReadRequest, ReadResponse, Request, Response};
+use tower::Service;
+
+/// Trait alias for services handling state-modifying requests.
+pub trait StateServiceTrait:
```
arya2: Optional: It may be nice to name these `State` and `ReadState` traits, move the definitions to the `service` module (or a `zebra_state::service::traits` submodule), then only expose them through the `zebra_state::service` module.

zebra-state/src/service_traits.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Blanket impl for any matching service.
+impl<T> ReadStateServiceTrait for T where
+    T: Service<ReadRequest, Response = ReadResponse, Error = BoxError, Future: Send + 'static>
+        + Clone
+        + Send
+        + Sync
+        + 'static
+{
+}
```
arya2: Optional:

If we want to split the `ReadStateService`'s `Service` impl into multiple `Service` impls (one for each request), it may be nice to define an intermediary supertrait ahead of that to avoid repetitive constraints in the supertrait definition.

We could also use the intermediary supertrait for other services, e.g. the block verifier and mempool.

If we want to include an intermediary supertrait, it would be nice to define that in `zebra-node-services` instead of here.

```suggestion
/// Trait alias for services handling state-modifying requests.
pub trait State: ZebraService<Request, Response> {}

/// Blanket impl for any matching service.
impl<T> State for T where T: ZebraService<Request, Response> {}

/// Trait alias for services handling read-only state requests.
pub trait ReadState: ZebraService<ReadRequest, ReadResponse> {}

/// Blanket impl for any matching service.
impl<T> ReadState for T where T: ZebraService<ReadRequest, ReadResponse> {}

pub trait ZebraService<Request, Response, Err = BoxError>:
    Service<Request, Response = Response, Error = Err, Future: Send + 'static>
    + Clone
    + Send
    + Sync
    + 'static
{
}

impl<T, Request, Response, Err> ZebraService<Request, Response, Err> for T where
    T: Service<Request, Response = Response, Error = Err, Future: Send + 'static>
        + Clone
        + Send
        + Sync
        + 'static
{
}
```

arya2 (dismissed):
This should help us replace a _lot_ of repetitive constraints, thank you for these changes!

I left a couple optional suggestions, feel free to resolve them without making changes if you'd like this PR to merge as-is.

syszery:
Hi @arya2, thank you for the great feedback and guidance — it really improved the quality of the PR!

I’ve now implemented all your suggestions:

* Introduced the shared `ZebraService` supertrait in `zebra-node-services`
* Moved the `State` and `ReadState` trait aliases into `zebra_state::service::traits`
* Exposed them via `zebra_state::service`
* All tests pass locally; `cargo fmt` and `clippy` show no issues

One small note: since `ReadStateService` is already exported in `zebra-state`, I had to use this workaround in `zebra-rpc`:

```rust
use zebra_state::{ReadState as ReadStateService, State as StateService, TransactionLocation};
```

Let me know if you'd prefer a different naming scheme for the traits or module layout — happy to adjust. 

Also, if you're happy with the current design, I’d be up for continuing with the `Mempool` or `BlockVerifier` traits — I noticed both while working in `methods.rs`, too.

Thanks again for your help!

arya2:
> Let me know if you'd prefer a different naming scheme for the traits or module layout — happy to adjust.

The current names look good to me!

> Also, if you're happy with the current design, I’d be up for continuing with the Mempool or BlockVerifier traits — I noticed both while working in methods.rs, too.

That would be very nice. There are a couple merge conflicts, but otherwise this PR looks good to merge in its current state if you want to update those in a follow-up PR, or those changes can be changed here if you'd prefer.

syszery:
Hi @arya2, I’ve rebased the branch and added `MempoolService` and `BlockVerifierService`. Is the naming okay, and are the new modules in the right locations? Happy to adjust if anything needs changing.

zebra-consensus/src/router/service_trait.rs
```diff
@@ -0,0 +1,15 @@
...
+//! This trait provides a convenient alias for `tower::Service`
+//! implementations that operate on Zebra block verification request and response types.
+//!
+//! - [`BlockVerifierService`]: for services that handle semantic block verification requests.
+
+use crate::router::Request;
+use zebra_chain::block::Hash;
+use zebra_node_services::service_traits::ZebraService;
+
+/// Trait alias for services handling semantic block verification requests.
```
arya2, line 12: Nitpick: The service may also route verification requests to the checkpoint verifier and won't always semantically verify the blocks.

```suggestion
//! - [`BlockVerifierService`]: for services that handle block verification requests.

use crate::router::Request;
use zebra_chain::block::Hash;
use zebra_node_services::service_traits::ZebraService;

/// Trait alias for services handling block verification requests.
```

arya2 (approved):
Thank you @syszery! This looks great.

zebra-consensus/src/router/service_trait.rs
oxarbitrage: Done at https://github.com/ZcashFoundation/zebra/pull/10053/commits/ad647b16325ebf38711f577cfbf1ca00500e43dd

PR #10021, opened 2025-10-24
republicize valid_addresses method

AloeareV:
## Motivation

The API allowing access to the addresses field was de-publicized at some point. 

## Solution

Re-publicize the API. 

### Tests

N/A


arya2 (approved):
Thank you!

PR #10022, opened 2025-10-24
fix: `z_validateaddress` bug

dorianvp:
## Motivation

When receiving a regtest t-address in `z_validateaddress` it fails because it is parsing it as a testnet address, even when running on regtest. This mismatch makes the endpoint return an "invalid" response, when it should be valid.

## Solution

If Zebra receives a t-address and the network is not mainnet, we treat regtest as equal to testnet.

### Tests

Added `rpc_z_validateaddress_regtest` for testing this scenario.

We're implementing this endpoint on Zaino [here](https://github.com/zingolabs/zaino/pull/389), and we encountered this issue there too.

The presence of this bug was previously discussed [here](https://github.com/ZcashFoundation/zebra/pull/9734#pullrequestreview-3070896979).

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-rpc/src/methods/tests/vectors.rs
+++ b/zebra-rpc/src/methods/tests/vectors.rs
@@ -2668,6 +2668,77 @@ async fn rpc_z_validateaddress() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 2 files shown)


arya2 (approved):
Thank you!

PR #10019, opened 2025-10-22
feat(rpc): Display `asm` code in script outputs

oxarbitrage:
## Motivation

We want to display the assembly (`asm`) representation of scripts in Zebra RPC's and the script type.

Close https://github.com/ZcashFoundation/zebra/issues/9330

## Solution

A new version of `zcash_script` ([v0.5.0](https://github.com/ZcashFoundation/zcash_script/releases/tag/v0.5.0)) was released, which introduces support for displaying scripts in assembly form ([#256](https://github.com/ZcashFoundation/zcash_script/pull/256)).

However, upgrading Zebra’s existing zcash_script dependency directly would cause version conflicts with other crates (`zcash_transparent`, `zcash_primitives`, etc.) that still depend on the older version.

To work around this, we added a separate dependency, `zcash_script05`, specifically in `zebra-rpc`, and use it only where the new assembly functionality is required.

### Tests

Snapshot tests were updated.

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [ ] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [ ] The library crate changelogs are up to date.
- [ ] The solution is tested.
- [ ] The documentation is up to date.

PR #10027, opened 2025-10-27
change(build): Remove duplicate dependency on `zcash_script`

arya2:
~This is a suggestion for #10019 to remove the duplicate dependency on different versions of `zcash_script`.~

This PR removes the duplicate dependency on `zcash_script`. It was originally a suggestion PR, but the base branch somehow merged with unresolved comments.


oxarbitrage (approved):
Thanks, I forgot to do this after the 0.4.2 release was created.

PR #10045, opened 2025-10-29
tests(qa): Update zallet to `0.1.0-alpha.2`

### LINKED ISSUE #10046: Test QA wallet.py test in a more recent zallet version
We were testing in an old zallet version. Update to a recent one and make sure it remains working.

oxarbitrage:
## Motivation

We want to run the wallet test using a newer version of Zallet. Close https://github.com/ZcashFoundation/zebra/issues/10046

## Solution

Tested up to the `0.1.0-alpha.2` Zallet commit: https://github.com/zcash/wallet/commit/027e5e2139b2ca8f0317edb9d802b03a46e9aa4c

I added this commit hash to the build file (even though it’s not used directly for building Zallet) so we can track which version was last tested.

With that version, the first error encountered was:

```
MissingData("Non-finalised state is empty.")
```

This was fixed in: https://github.com/ZcashFoundation/zebra/blob/abfe850dcc1bfc0982223287dcb5285cb1a3d00d/zebra-rpc/qa/rpc-tests/wallet.py#L29-L31

After that, I noticed an intermittent failure in the wallet’s reported balance. It seems related to block confirmations, sometimes the wallet reports the balance from the last mined block, and other times from the previous one:
https://github.com/ZcashFoundation/zebra/blob/abfe850dcc1bfc0982223287dcb5285cb1a3d00d/zebra-rpc/qa/rpc-tests/wallet.py#L100-L104

Reporting the previous block’s value makes sense, since `z_gettotalbalance` requires at least one confirmation, but it’s unclear why the latest block’s value sometimes appears given the same test conditions.

I’m not certain whether this behavior comes from Zallet, zcash_client_backend, or the test setup itself. However, switching to `assert_true` is sufficient for this test. Our goal is just to verify that the wallet balance increases as new blocks are mined to the configured wallet address.

### Tests

This is test code.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-rpc/build.rs
+++ b/zebra-rpc/build.rs
@@ -5,7 +5,7 @@ use std::{
     process::Command,
 };
 
-const ZALLET_COMMIT: Option<&str> = Some("de70e46e37f903de4e182c5a823551b90a5bf80b");
+const ZALLET_COMMIT: Option<&str> = Some("baf860eb1160b0ab34f2c73b379e7bf7325003e2");
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     build_or_copy_proto()?;
--- a/zebra-rpc/build.rs
+++ b/zebra-rpc/build.rs
@@ -5,7 +5,8 @@ use std::{
     process::Command,
 };
 
-const ZALLET_COMMIT: Option<&str> = Some("baf860eb1160b0ab34f2c73b379e7bf7325003e2");
+// 0.1.0-alpha.2
+const ZALLET_COMMIT: Option<&str> = Some("027e5e2139b2ca8f0317edb9d802b03a46e9aa4c");
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     build_or_copy_proto()?;
--- a/zebra-rpc/qa/README.md
+++ b/zebra-rpc/qa/README.md
@@ -76,24 +76,16 @@ To get real-time output during a test you can run it using the
 python3 qa/rpc-tests/wallet.py
 ```
 
-A 200-block -regtest blockchain and wallets for four nodes
-is created the first time a regression test is run and
-is stored in the `qa/cache/` directory.  Each node has the miner
-subsidy from 25 mature blocks (25*10=250 ZEC) in its wallet.
-
-TODO: https://github.com/ZcashFoundation/zebra/issues/9726
-
-After the first run, the `qa/cache/` blockchain and wallets are
-copied into a temporary directory and used as the initial
-test state.
-
-If you get into a bad state, you should be able
-to recover with:
+If a test gets stuck, you can stop the underlying binaries with:
 
 ```bash
-rm -rf qa/cache
 killall zebrad
+killall zallet
 ```
+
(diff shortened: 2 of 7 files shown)


daira:
https://github.com/ZcashFoundation/zebra/blob/abfe850dcc1bfc0982223287dcb5285cb1a3d00d/zebra-rpc/qa/rpc-tests/wallet.py#L29-L31

Obviously this is inherently racy? It can't be a complete fix to the issue the comment describes. Similarly for the other call to `time.sleep(1)`.

To fix it you would need to poll or be notified that the state had been populated, or else make any attempt to query the state while it has not been populated blocking.

We do have a few `time.sleeps` in the zcashd RPC tests but it's definitely an antipattern, which just makes tests flaky and unreliable. You would be surprised how long the tail of latencies of whatever the test is waiting for can get; it can easily be seconds, because nothing is real-time and you have no control over other tests that may be running in parallel.

conradoplg:
I'm getting this error:

```
Unexpected exception caught during testing: Error: reached the transparent gap limit while attempting to generate a new address at index 10.
  File "/home/conrado/zebra/zebra-rpc/qa/rpc-tests/test_framework/test_framework.py", line 145, in main
    self.run_test()
  File "/home/conrado/zebra/zebra-rpc/./qa/rpc-tests/wallet.py", line 40, in run_test
    address = self.wallets[0].z_getaddressforaccount(account['account_uuid'])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/conrado/zebra/zebra-rpc/qa/rpc-tests/test_framework/coverage.py", line 50, in __call__
    return_val = self.auth_service_proxy_instance.__call__(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/conrado/zebra/zebra-rpc/qa/rpc-tests/test_framework/authproxy.py", line 137, in __call__
    raise JSONRPCException(response['error'])
Stopping nodes
```

I haven't looked into it though. I'll do that tomorrow

conradoplg (approved):
Looks good, thanks. The error above stopped happening after I recompiled everything so I was probably doing something wrong.

conradoplg:
> https://github.com/ZcashFoundation/zebra/blob/abfe850dcc1bfc0982223287dcb5285cb1a3d00d/zebra-rpc/qa/rpc-tests/wallet.py#L29-L31
> 
> Obviously this is inherently racy? It can't be a complete fix to the issue the comment describes. Similarly for the other call to `time.sleep(1)`.
> 
> To fix it you would need to poll or be notified that the state had been populated, or else make any attempt to query the state while it has not been populated blocking.
> 
> We do have a few `time.sleeps` in the zcashd RPC tests but it's definitely an antipattern, which just makes tests flaky and unreliable. You would be surprised how long the tail of latencies of whatever the test is waiting for can get; it can easily be seconds, because nothing is real-time and you have no control over other tests that may be running in parallel.

I agree with this but it seems fixing it properly requires some functionality in zallet: https://github.com/zcash/wallet/issues/316

PR #10048, opened 2025-10-31
feat: Support ZIP-213

### LINKED ISSUE #5472: Enable coinbase funds to be mined to Sapling
## Motivation

Some miners use shielded coinbase, but it's pretty rare.

### Specifications

> * [Sapling onward] If effectiveVersion ≥ 4 and nSpendsSapling + nOutputsSapling > 0, then:
> – let bvkSapling and SigHash be as defined in § 4.13 ‘Balance and Binding Signature (Sapling)’ on p. 50;
> – bindingSigSapling MUST represent a valid signature under the transaction binding validating key bvkSapling of SigHash — i.e. BindingSigSapling.ValidatebvkSapling (SigHash, bindingSigSapling) = 1. [NU5 onward] As specified in § 5.4.7 ‘RedDSA, RedJubjub, and RedPallas’ on p. 90, the validation of the 𝑅 component of the signature changes to prohibit non-canonical encodings.
> * The total value in zatoshi of transparent outputs from a coinbase transaction, minus vbalanceSapling, minus vbalanceOrchard, MUST NOT be greater than the value in zatoshi of block subsidy plus the transaction fees paid by transactions in this block.
> * A coinbase transaction MUST NOT have any Spend descriptions.
> * [Heartwood onward] All Sapling and Orchard outputs in coinbase transactions MUST decrypt to a note plaintext , i.e. the procedure in § 4.19.3 ‘Decryption using a Full Viewing Key (Sapling and Orchard)’ on p. 67 does not return ⊥, using a sequence of 32 zero bytes as the outgoing viewing key. (This implies that before Canopy activation, Sapling outputs of a coinbase transaction MUST have note plaintext lead byte equal to 0x01.)
> * [Canopy onward] Any Sapling or Orchard output of a coinbase transaction decrypted to a note plaintext according to the preceding rule MUST have note plaintext lead byte equal to 0x02. (This applies even during the “grace period” specified in [ZIP-212].)

### Designs

- Accept encoded sapling and unified containing sapling addresses in the miner address config
- Create shielded outputs for miners using the Sapling address, according to the consensus rules
- In the `z_validateaddress` and `z_listunifiedreceivers` RPCs, validate Sapling addresses and return those fields

### Related issues

See #5929

### LINKED ISSUE #5929: Enable coinbase funds to be mined to Orchard
## Motivation

Some miners use shielded coinbase, but it's pretty rare.

### Specifications

> * [NU5 onward] If effectiveVersion ≥ 5 and nActionsOrchard > 0, then:
> – let bvkOrchard and SigHash be as defined in § 4.14 ‘Balance and Binding Signature (Orchard)’ on p. 52;
> – bindingSigOrchard MUST represent a valid signature under the transaction binding validating key bvkOrchard of SigHash — i.e. BindingSigOrchard.ValidatebvkOrchard (SigHash, bindingSigOrchard) = 1. As specified in § 5.4.7 ‘RedDSA, RedJubjub, and RedPallas’ on p. 90, validation of the 𝑅 component of the signature prohibits non-canonical encodings.
> * The total value in zatoshi of transparent outputs from a coinbase transaction, minus vbalanceSapling, minus vbalanceOrchard, MUST NOT be greater than the value in zatoshi of block subsidy plus the transaction fees paid by transactions in this block.
> * [NU5 onward] In a version 5 coinbase transaction, the enableSpendsOrchard flag MUST be 0.
> * [NU5 onward] In a version 5 transaction, the reserved bits 2 .. 7 of the flagsOrchard field MUST be zero.
> * [Heartwood onward] All Sapling and Orchard outputs in coinbase transactions MUST decrypt to a note plaintext , i.e. the procedure in § 4.19.3 ‘Decryption using a Full Viewing Key (Sapling and Orchard)’ on p. 67 does not return ⊥, using a sequence of 32 zero bytes as the outgoing viewing key. (This implies that before Canopy activation, Sapling outputs of a coinbase transaction MUST have note plaintext lead byte equal to 0x01.)
> * [Canopy onward] Any Sapling or Orchard output of a coinbase transaction decrypted to a note plaintext according to the preceding rule MUST have note plaintext lead byte equal to 0x02. (This applies even during the “grace period” specified in [ZIP-212].)

### Designs

- Accept unified containing orchard addresses in the miner address config
- In the `getblocktemplate` RPC, create shielded outputs for miners using the Orchard address, according to the consensus rules
- In the `z_validateaddress` and `z_listunifiedreceivers` RPCs, validate Orchard addresses and return those fields

### Related issues

See #5472

### LINKED ISSUE #9605: Construct coinbase txs using tx builder in `zcash_primitives`

upbqdn:
## Motivation

- Close #5472.
- Close #5929.
- Close #9605.

## Solution

- Construct coinbase txs as specified in [ZIP 213](https://zips.z.cash/zip-0213) using the [tx builder](https://docs.rs/zcash_primitives/latest/zcash_primitives/transaction/builder/struct.Builder.html) from `zcash_primitives`.
- Based on https://github.com/zcash/librustzcash/pull/1893
- Based on #10191 

### Follow-up Work

- #10170

### PR Checklist

- [x] The solution is tested.


upbqdn:
Replaces #9741.

zancas:
I notice this has conflicts?

zebra-rpc/src/methods/types/default_roots.rs
```diff
@@ -42,3 +52,53 @@ pub struct DefaultRoots {
...
+
+        Ok(Self {
+            merkle_root: iter::once(coinbase.hash)
+                .chain(mempool_txs.iter().map(|tx| tx.transaction.id.mined_id()))
+                .collect(),
+            chain_history_root,
+            auth_data_root,
+            block_commitments_hash: if NetworkUpgrade::current(net, height)
+                == NetworkUpgrade::Heartwood
+                && chain_history_root == block::CHAIN_HISTORY_ACTIVATION_RESERVED.into()
```
arya2, line 93: I introduced this mistake earlier, I thought your PR had fixed it, but it's still here.

```suggestion
            block_commitments_hash: if NetworkUpgrade::current(net, height) < NetworkUpgrade::Nu5 {
                chain_history_root.bytes_in_serialized_order().into()
```


or 


```rust

// instead of `chain_history_root.bytes_in_serialized_order().into()`,
chain_history_root.bytes_into()

pub trait BytesInDisplayOrder {
..
    /// Converts the bytes in the internal serialized order to the expected type.
    fn bytes_into<T: From<[u8; BYTE_LEN]>>(&self) -> T {
        self.bytes_in_serialized_order().into()
    }
}
```

zebra-chain/src/transparent/serialize.rs
arya2, line 1: 🎉

zebra-rpc/src/methods/types/get_block_template/constants.rs
```diff
@@ -2,11 +2,15 @@
 
 use jsonrpsee_types::ErrorCode;
 
+use zcash_transparent::coinbase::{MAX_COINBASE_HEIGHT_LEN, MAX_COINBASE_SCRIPT_LEN};
 use zebra_chain::{
     block,
     parameters::subsidy::FundingStreamReceiver::{self, *},
 };
 
+/// The maximum length of the optional, arbitrary data in the script sig field of a coinbase tx.
```
arya2, line 11: Optional/Nitpick:

```suggestion
/// The maximum length of the optional, arbitrary data in the script sig field of a coinbase tx that
/// can be configured to be included in the output of calls to the `getblocktemplate` RPC method.
```

zebra-rpc/src/methods/types/default_roots.rs
arya2: This suggestion still applies, the `block_commitments_hash` field should be the plaing chain history root bytes if `chain_history_root == CHAIN_HISTORY_ACTIVATION_RESERVED && (nu == Heartwood || nu == Canopy)`.

I'm not sure what should happen pre-Canopy.

zebra-rpc/src/methods/types/get_block_template.rs
```diff
@@ -453,86 +545,31 @@ where
     mined_block_sender: mpsc::Sender<(block::Hash, block::Height)>,
 }
 
-// A limit on the configured extra coinbase data, regardless of the current block height.
-// This is different from the consensus rule, which limits the total height + data.
```
arya2, line 457: The second doc line, 457, would be nice to include for the new constant (instead of my suggestion there).

zebrad/tests/common/coinbase.rs
```diff
@@ -0,0 +1,85 @@
...
+        zebrad
+            .wait_with_output()?
+            .assert_failure()?
+            .assert_was_killed()
+            .wrap_err("possible port conflict with another zebrad instance")
+    }
+
+    try_join_all(MinerAddressType::iter().map(regtest_coinbase))
+        .await
+        .map(|_| ())
```
arya2, line 84: Nice.

arya2 (approved):
This looks great ~and good to merge once the dependencies have been published~.

I've pushed a commit with my suggestions, feel free to keep it or make changes.

Great refactors!

Update: This looks like it could merge as-is once the deny lints are fixed.

Cargo.toml
```diff
@@ -168,10 +171,27 @@ uint = "0.10"
...
+zcash_address = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_history = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_transparent = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash", branch = "bump-zcash_script" }
+
+zcash_script = { git = "https://github.com/ZcashFoundation/zcash_script", branch = "make-parse-pub" }
+libzcash_script = { git = "https://github.com/ZcashFoundation/zcash_script", branch = "make-parse-pub" }
```
arya2: This seems okay to merge with the dependency patches, these last two are on main, and the others are https://github.com/zcash/librustzcash/compare/main...bump-zcash_script

```suggestion
zcash_script = { git = "https://github.com/ZcashFoundation/zcash_script", branch = "master" }
libzcash_script = { git = "https://github.com/ZcashFoundation/zcash_script", branch = "master" }
```

arya2 (dismissed):

Cargo.toml
upbqdn: I don't think we'll be able to release our crates, then.

arya2: Not until zcash_script and librustzcash have been released and zebra has been updated, but we could do a hot fix release off the prior commit if needed.

upbqdn: That would still block a regular Zebra release, no?

arya2: Yep. we'd have to revert this commit on main unless we're just publishing a small hotfix. It seems likely enough to me that zcash_script and the librustzcash crates will be published before it's an issue that blocking a regular Zebra release on it would be okay.

zebra-rpc/src/methods/types/default_roots.rs
```diff
@@ -88,8 +88,8 @@ impl DefaultRoots {
                 .collect(),
             chain_history_root,
             auth_data_root,
-            block_commitments_hash: if NetworkUpgrade::current(net, height)
-                == NetworkUpgrade::Heartwood
+            // TODO: use final sapling root pre-heartwood (#10334)
+            block_commitments_hash: if NetworkUpgrade::current(net, height) < NetworkUpgrade::Nu5
```
upbqdn: Thanks for the fix. I have a bit more general one in a subsequent PR, so let's keep this one as-is to reduce merge conflicts. We can make additional fixes in that PR, I'll point it out to you.

zebra-rpc/src/methods/types/get_block_template/constants.rs
```diff
@@ -8,7 +8,10 @@ use zebra_chain::{
     parameters::subsidy::FundingStreamReceiver::{self, *},
 };
 
-/// The maximum length of the optional, arbitrary data in the script sig field of a coinbase tx.
+/// The maximum length of the optional, arbitrary data in the script sig field of a coinbase tx that
+/// can be configured to be included in the output of calls to the `getblocktemplate` RPC method.
```
upbqdn: The output of that RPC is an encoded block, and the script sig field mentioned in the comment is part of the coinbase tx's transparent _input_, also encoded as a script, so I'd find it a bit confusing to say that the arbitrary data is in the output of the RPC because then I'd expect to be able to see it there.

zebra-rpc/src/methods/types/get_block_template/constants.rs
```diff
@@ -8,7 +8,10 @@ use zebra_chain::{
...
+///
+/// This is different from the consensus rule, which limits the total height + data.
```
upbqdn: The consensus rule limits the total byte length of the  coinbase tx script (the contents of which might change). That length is the minuend in the formula, so adding that the derived value is different seems a bit confusing.

Cargo.toml
upbqdn: We also have patched all the lrz deps. I'll change the branches there to `main`, but still, I'd prefer to wait until we can merge cleanly. I'll mark the PR as ready for review anyway once I push all the fixes.

upbqdn: Actually, I can't point lrz to `main`. I forgot about the transitive dep on `zcash_script`.

upbqdn:
@arya2, the `cargo deny` lints were still failing. I noticed I had them fixed in one of the earlier commits, but reverted those changes in a later commit due to a mistake in rebasing. I fixed the earlier commit and force-pushed. I ended up pushing other fixes that were needed as well, and I replaced your commit. Feel free to submit the suggestions again if you want.

arya2 (dismissed):

Cargo.toml
upbqdn: Let's keep this open until the new releases are out.

conradoplg:
Is this still blocked?

zebra-rpc/src/methods/types/transaction.rs
```diff
@@ -110,37 +121,151 @@ impl From<VerifiedUnminedTx> for TransactionTemplate<NonNegative> {
...
-        Self {
-            data: tx.transaction.as_ref().into(),
-            hash: tx.id.mined_id(),
-            auth_digest: tx.id.auth_digest().unwrap_or(AUTH_DIGEST_PLACEHOLDER),
+        macro_rules! trace_err {
+            ($res:expr, $type:expr) => {
+                $res.map_err(|err| tracing::error!("Failed to add {} output: {err}", $type))
+                    .ok()
+            };
+        }
```
conradoplg: Any reason to not just return the error instead of just logging it?

Also, the macro seems simple enough that is might be clearer to just inline the code instead of using it.

Cargo.toml
```diff
@@ -23,15 +23,15 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/ZcashFoundation/zebra"
 homepage = "https://zfnd.org/zebra/"
 keywords = ["zebra", "zcash"]
-rust-version = "1.85.0"
+rust-version = "1.88"
```
conradoplg: Do you remember why this was bumped? If it was because of the LRZ crates then it should be fine

zebra-rpc/src/config/mining.rs
```diff
@@ -16,15 +20,18 @@ pub struct Config {
...
-    // Find the removed code at https://github.com/ZcashFoundation/zebra/blob/v1.5.1/zebra-rpc/src/config/mining.rs#L18-L38
-    // Restore the code when conditions are met. https://github.com/ZcashFoundation/zebra/issues/8183
-    /// Extra data to include in coinbase transaction inputs.
-    /// Limited to around 95 bytes by the consensus rules.
+    /// Optional data that Zebra will include in the transparent input of a coinbase transaction.
+    /// Limited to 94 bytes.
+    ///
+    /// If this string is hex-encoded, it will be hex-decoded into bytes. Otherwise, it will be
+    /// UTF-8 encoded into bytes.
+    pub miner_data: Option<String>,
```
conradoplg: Renaming this requires a major bump in Zebra itself. Should we stick with `extra_coinbase_data`?

conradoplg (commented):
Not a thorough review, added some comments

zebra-rpc/src/methods/types/transaction.rs
upbqdn: I use the macro as a building block in the three closures below, which I then use in this `match` statement, which adds the miner reward to the tx: https://github.com/zcashfoundation/zebra/blob/d2edc5aad48647d234aabf6f3cfdc796223670e0/zebra-rpc/src/methods/types/transaction.rs#L204-L227. The `match` statement tries the ua first, within which the Orchard receiver has the highest precedence, then Sapling, and then the transp one. Along the way, the macros log failures, and the `match` statement returns an error only if it couldn't add the reward to any receiver in the ua, or to the other addr types.

So the `match` statement fails when it couldn't add the reward to the tx at all, and logs the failures implicitly along the way.

I wrote it this way to keep the logic in the `match` statement concise.

upbqdn: But I could get rid of the macro if you prefer that. Here's a version without it: https://github.com/zcashfoundation/zebra/blob/70fcbac23cc3b3231e2ca5340c6dc38a702286fb/zebra-rpc/src/methods/types/transaction.rs#L166-L220.

upbqdn:
One detail I don't like about this PR is that the core logic for creating coinbase transactions takes place in the `zebra-rpc` crate.

Cargo.toml
upbqdn: It was because of this: https://github.com/ZcashFoundation/zebra/actions/runs/24003946911/job/70004474592?pr=10048#step:6:408.

zebra-rpc/src/config/mining.rs
upbqdn: Reverted in [ed85a99](https://github.com/ZcashFoundation/zebra/pull/10048/commits/ed85a99196eab2ec3c4cd614cc37f8191fe839c3). Thanks for noticing, the reversion was on my to-do list for this PR.

Zk-nd3r:
Running Zebra 4.3.0 + s-nomp on mainnet with shielded coinbase.  Happy to test this branch against our pool when it's ready for external testing,  we consume GBT directly and would catch regressions in coinbase output that break pool software.

One thing we've seen,  s-nomp assumes specific field ordering in the GBT response.  If coinbase tx structure changes shape,  blocks get rejected silently by the network and the pool doesn't know until you dig through Zebra logs.

upbqdn:
It is ready for external testing; I've mined a block on Mainnet using this branch. Feel free to let us know about any regressions.

Zk-nd3r:
Will build from this branch and point our s-nomp instance at it.  Testing shielded coinbase with the new coinbase construction path.  Will report back with results.

upbqdn:
@Zk-nd3r, you're actually better off going with PR #10170. It's in draft because it needs to be merged after this PR, but it's also ready. It targets `main`, but it's based on top of this PR.

zebra-chain/src/block/height.rs
```diff
@@ -148,6 +148,14 @@ impl TryFrom<u32> for Height {
     }
 }
 
+impl TryFrom<i64> for Height {
+    type Error = core::num::TryFromIntError;
+
+    fn try_from(height: i64) -> Result<Self, Self::Error> {
+        u32::try_from(height).map(Height)
```
conradoplg, line 155: (found with Claude) this does not enforce the Height::MAX restriction. Might be `u32::try_from(height)?.try_into()` instead

zebra-chain/src/transparent/serialize.rs
conradoplg, line 1: I'm double-checking this Claude finding but posting here to speed things up. Might be good to repurpose the old tests if possible.

---

zebra-chain/src/transparent/serialize.rs now delegates BIP-34 height parsing to
  zcash_script::opcode::PossiblyBad::parse + PushValue::restrict. The previous in-tree implementation had explicit
  checks that:
  - rejected non-minimal encodings (e.g. 0x01 0x01 for height 1 — must be 0x51),
  - rejected heights outside the valid range for each prefix length,
  - rejected heights > Height::MAX.

  The tests that pinned these (parse_coinbase_height_mins in zebra-chain/src/transparent/tests/vectors.rs, plus the
  round-trip proptests in tests/prop.rs) were deleted with the old code, with no replacement tests against
  Input::zcash_deserialize. If PushValue::restrict does not enforce minimal-encoding identically to zcashd, a
  malicious or buggy block parses successfully on Zebra and mismatches zcashd → chain split. Every node validating
  blocks (mining or not) is affected. Add direct regression tests against Input::zcash_deserialize covering each
  height range and each non-minimal encoding before this can land.

conradoplg (changes requested):
Some Claude findings that seem relevant

zebra-chain/src/block/height.rs
upbqdn: That enforcement is actually not needed. I was thinking I'd remove the upper limit entirely. The spec says: 

> Implementations MUST support block heights up to and including 2³¹−1.

That doesn't say anything about the upper limit. I think we should remove the `Height::MAX` limit entirely.

I'd prefer to keep the conversion you referenced as-is.

upbqdn: The type that represents heights in LRZ has no upper limit either.

Cargo.toml
upbqdn: The bump is not needed anymore.

zebra-chain/src/transparent/serialize.rs
upbqdn: This is a great finding. Fixed in [206d491](https://github.com/ZcashFoundation/zebra/pull/10048/commits/206d491878fdac52d5d360f9e8bab1e961631cd4).

zebra-chain/src/block/height.rs
conradoplg: Makes sense, thanks

conradoplg (commented):
Looks good, thanks! 🎉 

Some tests are failing but seems unrelated, I'm re-running them to see what happens.

Also needs solving conflicts.

Other than that this is now blocked on just deciding if we want to include it in the next release or not.

conradoplg:
Still some conflicts left :sweat:

oxarbitrage:
> Still some conflicts left 😓

Sorry, that was caused by me. Fixed at https://github.com/ZcashFoundation/zebra/pull/10048/changes/0cfaaa9d6bd368132ccf4fa24e1a21a383e88d11

oxarbitrage (dismissed):
Reviewed the last commit (dd302dc). Correct — removes continue-on-error from sync-full-testnet and generate-checkpoints-testnet jobs in the integration tests workflow. This was already fixed in main via #10264 but got reverted during the merge into zip-213. Trivial fix, matches the already-merged PR.

conradoplg (dismissed):

oxarbitrage (approved):
Last two commits look good: merge from main (81ce9f5) and the test fix (e1aba60) adjusting the chain_value_pool_change test to use Input::Coinbase struct directly after Input::new_coinbase() was removed by this PR. Should fix the clippy/unused-deps/check-cargo-lock CI failures.

PR #10053, opened 2025-11-02
Cleanup: apply new service supertraits introduced in #10010

syszery:
## Motivation

This PR is part of #9665. In PR #10010, new service traits were introduced. This PR applies the new traits in the `zebra-rpc`.

While reviewing the modules, it was observed that some bounds like `Future: Send + 'static` or `Sync` are not always required (for example, in `zebra-rpc/src/queue.rs`). This PR only replaces cases that **exactly match** the new trait bounds.

### Solution

* Applied the unmerged review feedback from PR #10010.
* Replaced explicit `Service<...> + Clone + Send + Sync + 'static` bounds with the new `Service` trait aliases where the bounds match exactly.
* Verified that the changes do not remove any required trait bounds or functionality.
* Left cases where `Future: Send` or `Sync` are not used untouched to avoid introducing unnecessary constraints.

### Reviewer Question

Are these current trait restrictions acceptable, and should I proceed to clean up the codebase? Or should they be revisited to be made more flexible before wider adoption?

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-consensus/src/router/service_trait.rs
+++ b/zebra-consensus/src/router/service_trait.rs
@@ -3,13 +3,13 @@
 //! This trait provides a convenient alias for `tower::Service`
 //! implementations that operate on Zebra block verification request and response types.
 //!
-//! - [`BlockVerifierService`]: for services that handle semantic block verification requests.
+//! - [`BlockVerifierService`]: for services that handle block verification requests.
 
 use crate::router::Request;
 use zebra_chain::block::Hash;
 use zebra_node_services::service_traits::ZebraService;
 
-/// Trait alias for services handling semantic block verification requests.
+/// Trait alias for services handling block verification requests.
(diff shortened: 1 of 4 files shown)


oxarbitrage:
@Mergifyio update

oxarbitrage (approved):
I updated the description to say "This PR is part of https://github.com/ZcashFoundation/zebra/issues/9665" instead of closing the issue, since I don’t think it fully resolves it.

The issue refers to code like:
- https://github.com/ZcashFoundation/zebra/blob/v3.0.0-rc.0/zebra-rpc/src/methods.rs#L698-L735
- https://github.com/ZcashFoundation/zebra/blob/v3.0.0-rc.0/zebra-rpc/src/server.rs#L93-L144

These parts aren’t covered by this PR.

That said, this is still good progress. We don’t need to tackle those remaining parts right now.

PR #10054, opened 2025-11-04
perf(state): avoid cloning UTXO maps and Arc<Block> in contextual validation

Forostovec:
Optimize ContextuallyVerifiedBlock::with_block_and_spent_utxos() by removing redundant new_outputs extension and spent_outputs cloning. Build a temporary HashMap<OutPoint, Utxo> from references to compute block.chain_value_pool_change(...), and compute before moving block to avoid Arc::clone().
Similarly, in remaining_transaction_value(), replace cloning the entire HashMap<OutPoint, OrderedUtxo> and utxos_from_ordered_utxos(...) with a local Utxo map derived from references.
These changes reduce allocations and copies without altering behavior: the value-balance computations still receive all required UTXOs (including intra-block new outputs), and stored fields remain unchanged.


mpguerra:
This is not a priority for us right now

PR #10056, opened 2025-11-05
feature(zebra-chain): add new `Bounded` vec type and apply it in AdjustDifficulty

### LINKED ISSUE #8484: Add new type similar to `AtLeastOne` to enforce min/max size constraints and use it instead of `Vec`s in `AdjustedDifficulty`
## Motivation

There's repetitive code around taking a constant number of items from an iterator to collect into a `Vec`, and there are max size constraints that aren't currently enforced in `AdjustedDifficulty` methods.

It would be easier to maintain and verify correctness if we had a struct that handled this logic instead.

## Design

Create a new type similar to `AtLeastOne` with a const generic for max size, which:
- Calls `take()` to ensure the max size is enforced, and

Either:
- Uses `AtLeastOne` as the inner type to check that there's at least one element, or
- Accepts another const generic for min size and returns an error if there are fewer items in the iterator than the `min` size (optional)

Currently the minimum length for this use case is 1, so using `AtLeastOne` as the inner type would suffice.

syszery:
## Motivation

This PR fixes #8484 by introducing the new type, `Bounded`, which extends the existing `AtLeastOne` type to enforce both minimum and maximum size constraints. This ensures that collections in `AdjustedDifficulty` have a controlled size, improving correctness and safety compared to plain `Vec`s.

## Solution

* Introduce `pub struct Bounded<T, const MAX: usize>` as a wrapper around `AtLeastOne<T>`.
* `Bounded` enforces a lower bound of at least one element (inherited from `AtLeastOne`) and adds an upper bound (`MAX`).
* Replace `Vec`s in `AdjustedDifficulty` with `Bounded` where constrained numbers of elements are expected.

### PR Checklist

- [x] The solution is tested.


arya2:
It might be better to use [`bounded-vec`](https://crates.io/crates/bounded-vec) instead of rolling our own.

syszery:
### Update

Hi @arya2, thanks for the suggestion. I’ve reworked the PR to use the `bounded_vec` crate.

Here’s what I did:

* Switched from the custom `Bounded` and `AtLeastOne` to the `bounded-vec` crate.
* Refactored `AtLeastOne` to be a type alias for `BoundedVec`.
* Updated all affected modules and tests to use `BoundedVec`.

The rebase is done, and I’ve confirmed that all tests, Clippy, and formatting checks passed. Let me know if you have any feedback!

conradoplg (commented):
Looks good, but needs solving conflicts (and syncing with main should solve CI issues)

syszery:
Hi @conradoplg, thanks for your feedback. I've rebased the branch as you suggested. All `zebra-chain` tests pass.

oxarbitrage (approved):
Looks good, thank you for making this cleanup.

PR #10060, opened 2025-11-05
fix(all): gate tx_v6 also under zcash_unstable=nu7

conradoplg:
## Motivation

This is a follow up to #8930.

Gating experimental/incomplete stuff behind features is dangerous because they can easily be turned on by accident via e.g. feature unification or `--all-features`. This PR also gates it behind `zcash_unstable` in the exact same way as `librustzcash` already does. Note that `zcash_unstable` is already required for the feature precisely because of `librustzcash`.

## Solution

Also gate behind `zcash_unstable = "nu7"`

### Tests

`RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo test --package zebrad --test acceptance --features tx_v6 -- nu7_nsm_transactions --exact --show-output` (the test from the NSM PR) should work.

Everything else should work as normal without the feature (or with the feature but without the flag)

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-chain/src/transaction.rs
+++ b/zebra-chain/src/transaction.rs
@@ -39,7 +39,7 @@ pub use unmined::{
 };
 use zcash_protocol::consensus;
 
-#[cfg(feature = "tx_v6")]
+#[cfg(all(zcash_unstable = "nu7", feature = "tx_v6"))]
 use crate::parameters::TX_V6_VERSION_GROUP_ID;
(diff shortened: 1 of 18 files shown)


oxarbitrage (approved):
- confirmed this was changed everywhere. 
- nsm test runs fine locally with new command.
- rest of the functionality is unchanged.

thanks!

PR #10068, opened 2025-11-07
use stagex for full source bootstrapped and deterministic build

antonleviathan:
## Motivation

I am on a mission to upgrade core zcash ecosystem components to StageX. This is a way to reduce supply chain risk, namely from Trusting Trust and environment compromise style attacks. Bitcoin core, and a number of other projects use reproduction and full source bootstrapping to address this risk.

The following repos are already being worked on:
* [zaino](https://github.com/zingolabs/zaino/pull/641)
* [wallet](https://github.com/zcash/wallet/pull/301)

StageX is a full source bootstrapped and fully deterministic Linux distribution focused on an extreme [threat model](https://distrust.co/threatmodel.html) in which it's essential to eliminate single points of failure in systems to protect their integrity. A similar solution exists, Guix, but StageX goes a step further in terms of strictness.

Some highlights of what StageX brings to the table:
* Everything in tree has to be bootstrappable (including all languages)
* Everything has to be deterministic
* All changes in in tree have to be signed by at least two maintainers
* All releases must be reproduced by at least two maintainers, on two different chipsets
* PGP keys used for signing must only ever be exposed to offline memory and be stored in a hardware device such as a smart card, pin protected, and require a tap for each operation

Currently I'm working on bumping StageX rust to the latest version, so I'm opening this PR based on the last stable release as it uses a rust version which is available in StageX atm.

I wanted to open a PR to gather feedback and prepare the PR for merging, if the maintainers are in favor.

Upfront warning that currently only x86_64 is supported to build but we are in the process of upgrading to LLVM being the core compiler which provider much better cross-compilation capabilities.

## Solution

I've refactored `docker/Dockerfile` to:
1. Use StageX as the build and runtime distribution
2. Build deterministically using Docker's `containerd` backend

### Tests

To reproduce:

* Install Docker 28.0.0+
* Run `./utils/build.sh`

You should get these hashes (if not pls let me know):
```
 => => exporting manifest sha256:08c03d217d596186f14af4f2165d5b93b4f183d9d91ea6873e400  0.0s
 => => exporting config sha256:88bd97e8ed08bfbd19982014e74f5a7ae44c0a8a37ad3f0d27514ae  0.0s

❯ sha256sum build/zebrad
b64dba27ef8ecbcb4e5a7485d2b07ada1be2ebc868a33790267d01256c2d45af  build/zebrad

```

Still don't trust me? Good, I would hope as much is the case. StageX is fully transparent, and built using a bunch of OCI images which means if you know how to read a Dockerfile, you can audit it fairly easily (you need comp language literacy but odds are you have that if you are reading this). You can review the relevant parts of the code in several hours. Main parts are:
* [Bootstrapping stages](https://codeberg.org/stagex/stagex/src/branch/main/packages/bootstrap)
* [Rust pallet](https://codeberg.org/stagex/stagex/src/branch/main/packages/pallet/rust)

Of course, you will have to spelunk into each dependency, but also, are you reviewing every dependency released by Alpine or Debian? Probably not, and even if you did, there is no way to reproduce them.

To reproduce the whole tree, clone the stagex repo and run `make` (you will need a few common dependencies, and a beefy processor and a good amount of ram if you want it to finish in a reasonable amount of time)

* [StageX](https://codeberg.org/stagex/stagex)

### Follow-up Work

* Park a `build` command somewhere to run the `docker/build.sh` as needed
* Upgrade Debian release or any other workflows to this toolchain
* Support for building on other arches

### PR Checklist

- [ ] The solution is tested.


gustavovalverde:
Hi @antonleviathan,

Thank you for taking the time to work on this PR. We really appreciate the effort you've put into this

I've carefully reviewed this PR, and while I recognize the security benefits of full-source bootstrap approaches, I have concerns about whether this level of complexity is appropriate for Zebra at this time. Let me explain:

## The Complexity

This PR introduces significant architectural changes:
- **20+ new base images** with pinned SHAs (replacing 1-2 Debian images)
- **New build tooling** with strict version requirements
- **Platform limitations** (x86_64 only, excluding ARM64/Apple Silicon developers)
- **Multiple new scripts** for building and compatibility checking
- **Changed runtime environment** (from Debian to StageX)

While each individual change may be justified, the cumulative maintenance burden and reduced accessibility for contributors is substantial.

## Questions for Clarification

Before I can evaluate whether this approach aligns with Zebra's needs, I'd like to understand:

1. **What specific attack vector** does StageX mitigate that reproducible Debian builds wouldn't? Are there documented cases of compiler-level backdoors in cryptocurrency/blockchain toolchains?

2. **Why StageX instead of Guix?** Bitcoin Core uses Guix for similar goals. Guix is more mature, supports multiple architectures, and has broader adoption. What advantages does StageX provide that justify its experimental status?

3. **What's the maintenance commitment?** Who will maintain this integration as StageX evolves?

## Technical Issues

I also identified several implementation concerns:

1. **Broken entrypoint** (docker/entrypoint.sh:28):
```bash
exec chroot --userspec=${UID}:${GID} / "$@"
```
This won't work as intended. `chroot` requires root or complex namespace setup. The previous `gosu` approach was correct.

2. **Test stage inconsistency**: The `test` stage still uses Debian-based `deps`, so tests aren't running in the same environment as production builds.

3. **Removed build caching**: Developer builds will be significantly slower without cargo registry caching.

4. **Workaround flags**: `CFLAGS="-D__GNUC_PREREQ(maj,min)=1"` suggests underlying toolchain issues. What's breaking that requires this?

5. **Static linking contradiction**: The Dockerfile copies C++ runtime libraries despite using `-C target-feature=+crt-static`. Are these actually needed?

## Alternative Approaches

We'd like to explore whether there are simpler approaches that address legitimate security concerns without this level of complexity

## Our Position

We're not opposed to improving Zebra's supply chain security, but **we need to ensure the benefits justify the costs**:

- **Cost**: Reduced platform support, increased build complexity, higher maintenance burden, steeper learning curve for contributors
- **Benefit**: Protection against Trusting Trust attacks (no documented cases in crypto) and build infrastructure compromise

For a project like Zebra, the cost/benefit ratio doesn't currently favor this approach.

## Moving Forward

We're also happy to explore lighter-weight reproducible build approaches that could provide most of the security benefits with less complexity.

antonleviathan:
Thank you for the thoughtful feedback. I'll try my best to address everything.

> While each individual change may be justified, the cumulative maintenance burden and reduced accessibility for contributors is substantial.

Regarding the upfront work, that's what I'm volunteering to do, so I don't see that part as a burden your team has to take on. I'm not sure what the added ongoing maintenance you are pointing at here. StageX is a drop in replacement for Debian or Alpine that has images published to Dockerhub and Quay. 

> What specific attack vector does StageX mitigate that reproducible Debian builds wouldn't? Are there documented cases of compiler-level backdoors in cryptocurrency/blockchain toolchains?

The Debian reproducible builds effort is both incomplete in terms of reproducibility, and does not bootstrap their compiler or languages, leaving surface area for attacks. Yes there are documented cases of compiler/environment level attacks most notable being Solar Wind and xcode ghost. If the software that's provided by a distribution is the same as another, but built in a different manner to reduce surface area of attacks, it may be considered to be a better choice.

> Why StageX instead of Guix? Bitcoin Core uses Guix for similar goals. Guix is more mature, supports multiple architectures, and has broader adoption. What advantages does StageX provide that justify its experimental status?

Guix allows for a single maintainer to push changes into tree. They have a git "inheriting" scheme, where as long as there is an unbroken chain of commits signed by maintainers, they are permitted, but a single individual is able to get changes into a release. This is not the case for StageX. Furthermore, StageX has a strict rule about all software in tree down to the compiler being full source bootstrapped and reproducible, which is also not true for Guix. You may refer to it as experimental and it is true that it's not as mature as other distros, but it is used in production by world class projects like Mysten Labs SUI, Talos Linux, Turnkey, and other zcash projects in the ecosystem, like zallet - even Apache Software Foundation recently expressed interest and is collaborating with us to add ARM support as they want to use StageX internally. While it is true it does not currently support as many architectures, it's about to very shortly, but I understand if that's a blocker. Also Guix uses Schema which makes things quite complex to review, while StageX is OCI native - the whole distro is just a bunch of OCI containerfiles with some scripts. We are currently exploring using StageX in the bitcoin space but more will be announced on that in the near future.

> What's the maintenance commitment? Who will maintain this integration as StageX evolves?

It's not dissimilar from maintaining any other distribution, Outside of potential determinism issues which add some overhead but my team and I have gotten pretty good at that, it feels and work exactly the same way as Debian or Alpine. I am interested in helping with this on an ongoing basis,

### Technical Issues

I can address these, and don't see them as major blockers, I just wanted to get a PR up to get some feedback, so this is great.

1, Will fix - regarding importing binaries like gosu, that is considered to be a single point of failure and an alternative approach should be used. Whoever builds that binary gets to decide what's in it and nobody can easily verify it. I am happy to work through this problem.
2, 3  Will fix - thank you
4. rocksdb, I've been waiting for an upstream fix
5. Will fix

We spent a lot of time and energy trying to solve this problem over a multiple years, and this was our solution since nothing hit our criteria regarding the threat model and ease of use. If you find something better I'd love to hear about it.

> Cost: Reduced platform support, increased build complexity, higher maintenance burden, steeper learning curve for contributors

I know that StageX is a new distribution, but as I've mentioned earlier, it's usually just a different image to point to on dockerhub.

> Benefit: Protection against Trusting Trust attacks (no documented cases in crypto) and build infrastructure compromise

Just because an attack hasn't happened so far, doesn't mean it won't, and we know that these attacks are not theoretical. Threat modeling based on historical/survivor bias is an ineffective way to harden a system, especially when it's a world leading blockchain. Also, it's not just about compiler attacks, full source bootstrapping and reproducible builds reduce risk of any attacks that manipulate the results of a build during build time, a la Solar Winds, which happens relatively frequently compared to compiler attacks.

Some more examples of related attacks which could have been much worse:
* XZ Utils: https://www.cisa.gov/news-events/alerts/2024/03/29/reported-supply-chain-compromise-affecting-xz-utils-data-compression-library-cve-2024-3094
* Ubuntu Snap: https://github.com/canonical-websites/snapcraft.io/issues/651
* Gentoo: https://archives.gentoo.org/gentoo-announce/message/dc23d48d2258e1ed91599a8091167002
* Debian: https://lists.debian.org/debian-devel-announce/2006/07/msg00003.html
* Arch Linux AUR: https://lists.archlinux.org/pipermail/aur-general/2018-July/034153.html
* Linux Mint (2016 - ISO Backdoor): https://blog.linuxmint.com/?p=2994
* SolarWinds Orion: https://www.ibm.com/docs/en/randori?topic=2022-solarwinds-orion-cve-2020-10148

I'm happy to continue the discussion, and would love to figure out how to make StageX work for you. Of course, if you decide it's not something you're interested in, I will respect your decision and spend my energy elsewhere. We do have a paper about StageX we will be publishing shortly, but it's not live yet - I can let you know when the preprint is ready if that's of interest. We built this distribution exactly for projects like Zcash, because we feel important decentralization technologies need to have the best in class supply chain security to help defend them, and most Linux distributions just aren't equipped to or try to deal with threats such as sophisticated attackers and nation states.

antonleviathan:
I made a followup commit to address some of the concerns.

* I've minimized the number of dependencies to just 4
* I switched from `gosu` to an approach which leverages busybox from StageX, which is minimal, deterministic and does the same thig
* I added back the caching lines which were accidentally removed
* Regarding Debian deps in the test stage, I'm not sure where there are still any left, but if there is something I'm missing someone please let me know
* I was able to remove the sketchy flag `CFLAGS="-D__GNUC_PREREQ(maj,min)=1"` and build works fine, must have been a stray from when I was trying to force determinism
* I removed everything except for busybox from runtime since the binary is static as was pointed out

Regarding using Debian. It can be used to achieve deterministic builds of your software but there are some tradeoffs. That's where we started, and you can refer to [this](https://git.distrust.co/public/toolchain) repo to see what that pattern looks like. It's not pretty, and we abandoned that strategy and built something we believe to be much more usable and secure.

When it comes to using Debian for deterministic builds:
* Pros: 
	* Does get you deterministic output for your software
	* You get to continue using Debian
* Cons: 
	* Highly complex scripting around debian 
	* Debian is not 100% deterministic, many packages claim to be reproducible incorrectly (relying on binary blobs), and the reproductions that do happen are done post-release by a single centralized shared-access build system that has the same supply chain attack vectors as elsewhere
	* Requires hardcoding a non-signed debian base image
	* [Toolchain](https://git.distrust.co/public/toolchain) is deprecated and will not be maintained, so you need to maintain any ad-hoc tools internally
	* Unreliable debian archives - 2 servers worldwide
	* Debian  does not solve for full source bootstrapping, opening doors for covert supply chain attacks
	* No forced code review, maintainers can merge their own code
	* Forced use glibc over leaner alternatives like musl
	* Forced to link all binaries against libgcc vs compiler-rt / libunwind

I will add a script that automatically bumps the pins for convenience, to reduce friction on that end.

The major objection around platform support will be addressed in the following weeks as we are switching to be LLVM native which will allow you to target anything LLVM can target.

We believe that build complexity for deterministic builds with Stagex is much lower than what you would need to do with Debian or Guix, but with higher security guarantees. If you find a better way, we would love to hear about it.

antonleviathan:
I added a little helper script for bumping the hashes. Ideally, someone would reproduce StageX to verify the hashes. Also, all of our signatures can be found in [here](https://codeberg.org/stagex/signatures), if you were to verify that we signed the images, and all the maintainer well known keys are linked [here](https://codeberg.org/stagex/stagex/src/branch/main/MAINTAINERS) - for example this is my [keyoxide](https://keyoxide.org/DC4B7D1F52E0BA4D). We are planning to add sigs to the images since the OCI spec supports this, so eventually you are going to be able to set policies on how many and whose signatures are required for an image to be permitted for use in your context.

Separately, while the `compat.sh` script is fairly large, it's not necessary, it's a convenience script to check if the user has everything they need to build deterministic images, which is [`containerd`](https://docs.docker.com/engine/storage/containerd/#enable-containerd-image-store-on-docker-engine) configured (adding a config to `/etc/docker/daemon.json` and running Docker that's 26+ since that's when they added support for deterministic builds.

gustavovalverde:
Hey @antonleviathan, I'm sorry I could not follow-up with the conversation, but we were considerably focused on [the release ](https://zfnd.org/zebra-3-0-0-release-our-most-feature-rich-release-ever/). By the way, this release added ARM64 support (which would be a blocker for this PR) and we also added SBOM and provenance attestations. And we're also planning to add cosign in the following days.

I wouldn't like you to keep investing more time into this until we have clear expectations with this process and outcome. So far, our decision is not to merge this until we have a better understanding of the risks and benefits for the users using our software and any UX trade-off in both scenarios.

I'll close the PR so it's clear for us this is not in our backlog, but we can keep the conversation happening here nonetheless

You can mention me in your replies so I get the notifications. Thank you for your effort, and for your interest in making the Zcash ecosystem more secure.

PR #10069, opened 2025-11-07
refactor(ci): enable horizontal scaling and improve deployment

gustavovalverde:
## Motivation

GCP node deployments had three problems:

1. **No scaling**: Single instance per network (no redundancy)
2. **Slow deployments**: Created new disks frequently, causing 1-2 hour blockchain re-syncs
3. **Wrong environment**: Docker images sometimes published to wrong environment during manual deployments

This PR fixes all three and adds health checks with autohealing.

## Solution

- Run 2-3 instances per network (1 per zone)
- Use Spot VMs by default (60-91% cost savings, multi-zone HA tolerates preemptions)
- Main branch and release instances get static IPs:
- Manual deployments get ephemeral IPs (to avoid conflicts)
- Fixed disk names per deployment type (prod/main/manual) so disks get reused
- HTTP health checks on `/healthy` endpoint (port 8080)
- Simpler disk parameter logic

## How It Works

**Fixed disk names with 1 instance per zone**: Each zone can have one disk with a given name. Since we run exactly 1 instance per zone, fixed disk names (e.g., `zebrad-cache-mainnet`) don't conflict.

**Zone selection**: Queries up to 3 available zones in the region at deployment time. Different regions have different zone counts (e.g., us-east1 has 3 zones, us-west1 has 4).

**Rolling update constraint**: Regional MIGs need `max-unavailable` to be 0 or >= zone count. We detect zone count and set `max-unavailable` to match.

## Testing

This has been tested in multiple ways:
- Deploying Mainnet and Testnet (both) into our dev and prod GCP projects
- Re-deploying to confirm it doesn't conflict
- Deploying automatically and manually to confirm IP usage and avoiding conflicts
- Removing the instances and re-deploying

## References

- [GCP Stateful MIG Documentation](https://cloud.google.com/compute/docs/instance-groups/stateful-migs)
- [GCP Health Check Concepts](https://cloud.google.com/load-balancing/docs/health-check-concepts)
- Zebra health endpoint: `/healthy` (returns 200 when syncing with peers, 503 on failure)

## PR Checklist

- [x] The PR name is suitable for the release notes
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md)
- [ ] The library crate changelogs are up to date
- [x] The solution is tested
- [x] The documentation is up to date

### DIFF
--- a/.github/workflows/zfnd-build-docker-image.yml
+++ b/.github/workflows/zfnd-build-docker-image.yml
@@ -45,6 +45,11 @@ on:
         required: false
         type: boolean
         default: false
+      environment:
+        description: "Environment to use for the build (commonly dev)"
+        required: false
+        type: string
+        default: dev
 
     secrets:
       DOCKERHUB_USERNAME:
@@ -71,7 +76,7 @@ jobs:
     if: github.repository_owner == 'ZcashFoundation'
     timeout-minutes: 210
     runs-on: ubuntu-latest
-    environment: ${{ github.event_name == 'release' && 'prod' || 'dev' }}
+    environment: ${{ github.event_name == 'release' && 'prod' || inputs.environment || 'dev' }}
     outputs:
       image_digest: ${{ steps.docker_build.outputs.digest }}
       image_name: ${{ fromJSON(steps.docker_build.outputs.metadata)['image.name'] }}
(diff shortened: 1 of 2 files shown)


gustavovalverde:
Admin-merging to be able to continue with some follow-up fixes and improvements

PR #10049, opened 2025-10-31
fix(mempool): Avoid panicking when a transaction is unexpectedly missing in the mempool

arya2:
## Motivation

Closes #10038.

## Solution

- Logs a warning instead of panicking when a transaction that should be removed from the mempool is missing in the mempool's verified set, and
- Avoids calling `remove()` with tx ids that are not in the mempool.

Related:
- Fixes a bug where mined transactions were not being removed as dependencies of transactions in the mempool.

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [ ] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [ ] The library crate changelogs are up to date.
- [ ] The solution is tested.
- [ ] The documentation is up to date.

PR #10074, opened 2025-11-10
fix(mempool): prevent stale dependency references in transaction graph

gustavovalverde:
## Motivation

Zebra nodes have been crashing with "invalid transaction key" panics at `zebrad/src/components/mempool/storage/verified_set.rs:288`. The panic occurs when `VerifiedSet::remove` receives transaction IDs that have already been evicted from the mempool.

This happens because the `TransactionDependencies::remove_all` removes a transaction from the dependencies and dependents maps but fails to remove it from its parent transactions' dependents sets. This leaves stale references in the dependency graph. When a parent transaction is later removed, it tries to remove already-evicted child transactions, causing the panic.

**Bug**
When `TransactionDependencies::remove_all(tx_hash)` removes a transaction, it:

1. Removes the transaction from `dependencies` map
2. Removes the transaction from `dependents` map
3. Recursively removes all dependent transactions
4. **NEVER removes the transaction from its parents' `dependents` sets**

**Scenario**
```
Initial state:
- transactions: {A, B}
- dependencies[B] = {A}
- dependents[A] = {B}

After remove_all(B):
- transactions: {A}         // B removed from verified set
- dependencies[B] = ø       // B's entry removed
- dependents[A] = {B}       // ← STALE! B no longer exists but A still references it
```

This panic can occur in multiple scenarios:
1. **A parent transaction is removed after its child** (child removed earlier, stale ref remains)
2. **Block mining removes multiple related transactions** (some removed as dependents, then accessed again)
3. **Expiration removes dependent before parent** (child expires first, parent expires later)
4. **Any operation that processes parent transactions** after children have been removed

### Tertiary Issue: Typo in `clear_mined_dependencies()`

In the specific case of clearing mined dependencies, there was also a typo:

```rust
// Before fix:
dependencies.remove(&dependent_id);  // Removing wrong ID!

// After fix:
dependencies.remove(&mined_tx_id);   // Remove the mined tx from dependencies
```

This compounded the stale edge problem specifically for mined transactions, but is a separate issue from the core invariant violation in `remove_all`.

## Solution

- Rework `TransactionDependencies::remove_all` to walk the graph with a queue and remove every visited child from its parents' dependents sets before cascading, guaranteeing that only live IDs are returned.
- Fix `clear_mined_dependencies` to delete the mined hash (not the dependent ID) from each dependency set and document the invariant that callers can rely on.
- Add regression coverage:
  - Unit tests for stale-child removal, transitive chains, multi-child, multi-parent, and diamond-shaped graphs
  - `VerifiedSet::remove_all_that` integration test that exercises the mempool path without going through `insert`

### Tests

```bash
cargo test -p zebra-node-services transaction_dependencies
cargo test -p zebrad verified_set_remove_all_that_with_dependencies
```

- Original feature PR: #8857

### PR Checklist

- [ ] The library crate changelogs are up to date. (not required for this internal fix)
- [x] The solution is tested.

### DIFF
--- a/zebra-node-services/src/mempool/transaction_dependencies.rs
+++ b/zebra-node-services/src/mempool/transaction_dependencies.rs
@@ -1,6 +1,6 @@
 //! Representation of mempool transactions' dependencies on other transactions in the mempool.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use zebra_chain::{transaction, transparent};
 
@@ -73,7 +73,7 @@ impl TransactionDependencies {
                 };
 
                 // TODO: Move this struct to zebra-chain and log a warning here if the dependency was not found.
-                let _ = dependencies.remove(&dependent_id);
+                let _ = dependencies.remove(mined_tx_id);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


conradoplg:
I think the fix is the same as https://github.com/ZcashFoundation/zebra/pull/10049? But the tests seem useful

upbqdn:
@gustavovalverde, what kinds of tools did you use for this one?

gustavovalverde:
> @gustavovalverde, what kinds of tools did you use for this one?

@upbqdn a combination of Claude Code with Codex, and some web search.

upbqdn:
Nice, even for the PR description?

gustavovalverde:
> Nice, even for the PR description?

Yes. But I had to manually intervene the PR description, but 90% yes.

PR #10096, opened 2025-11-14
refactor(zebra-chain): use `strum::EnumIter` instead of `NETWORK_UPGRADES_IN_ORDER`

### LINKED ISSUE #8410: Use a macro to generate an `iter()` method instead of using `NETWORK_UPGRADES_IN_ORDER`
## Motivation

It won't be possible to forget to update the constant when we add variants to `NetworkUpgrade`, though this seems unlikely anyway.

This could also be useful for `NetworkKind` and perhaps elsewhere.

## Possible Design

- Add `strum` or another crate with an `EnumIter` macro as a dependency
- Move the constant to a test that checks that the `NetworkUpgrade::iter()` method returns variants in the right order?

syszery:
## Motivation

The manually maintained `NETWORK_UPGRADES_IN_ORDER` constant required updating whenever a new `NetworkUpgrade` variant was added, which creates unnecessary maintenance overhead.

Issue #8410 suggested replacing the constant with a macro-generated iterator and moving the ordering check into a test. This PR implements that approach.

## Solution

- Adds `strum` to `zebra-chain` and derives `EnumIter` for the `NetworkUpgrade` enum.
- Introduces a test using the original ordering constant.
- All existing tests pass, and new tests verify ordering and activation list completeness.

Happy to adjust based on feedback. 

### PR Checklist

- [x] The solution is tested.


zebra-chain/Cargo.toml
```diff
@@ -113,6 +113,7 @@ serde-big-array = { workspace = true }
 futures = { workspace = true }
 itertools = { workspace = true }
 rayon = { workspace = true }
+strum = { version = "0.27.2", features = ["derive"] }
```
upbqdn, line 117: This needs to be reflected in the change log for `zebra-chain`.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -12,30 +12,18 @@ use std::fmt;
...
-];
-
 /// A Zcash network upgrade.
 ///
 /// Network upgrades change the Zcash network protocol or consensus rules. Note that they have no
 /// designated codenames from NU5 onwards.
-#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize, Ord, PartialOrd)]
+#[derive(
+    Copy, Clone, EnumIter, Debug, Eq, Hash, PartialEq, Serialize, Deserialize, Ord, PartialOrd,
+)]
```
upbqdn, line 28: This needs to be reflected in the change log for `zebra-chain`.

zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -557,3 +544,41 @@ impl ConsensusBranchId {
         NetworkUpgrade::current(network, height).branch_id()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
upbqdn: ###### Optional

We usually put test modules into separate files to keep the implementation files shorter.

syszery:
Hi @upbqdn, thanks for the feedback! I’ll address this soon.

zebra-chain/src/parameters/network_upgrade.rs
syszery: I’ve moved the tests to `zebra-chain/src/parameters/tests.rs`.

zebra-chain/Cargo.toml
syszery: I've updated the changelog.

syszery:
Hi @upbqdn, I’ve moved the network upgrade tests to `zebra-chain/src/parameters/tests.rs`, and updated the changelog. All tests pass with these changes. 

I've also rebased the branch to resolve merge conflicts.

zebra-chain/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Refactor `NetworkUpgrade` to use `strum::EnumIter` instead of `NETWORK_UPGRADES_IN_ORDER` and add the `strum` dependency to `zebra-chain`. (#10096)
```
upbqdn:
```suggestion
- `parameters/network_upgrade/NetworkUpgrade` now derives `strum::EnumIter`
```

zebra-chain/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
```
upbqdn, line 13:
```suggestion

### Added

- `strum` dependency
- 

```

upbqdn:
It looks like you need to push an updated `Cargo.lock`.

syszery:
Hi @upbqdn, thanks again for the feedback and suggestions. I’ve applied the changes and added the updated Cargo.lock.

One thing I noticed: this PR currently relies on Strum iterating over the enum variants in declaration order, which matches the current behavior. Is this acceptable, or would you prefer we add explicit sorting by activation height instead of relying on enum order?

oxarbitrage:
> One thing I noticed: this PR currently relies on Strum iterating over the enum variants in declaration order, which matches the current behavior. Is this acceptable, or would you prefer we add explicit sorting by activation height instead of relying on enum order?

I think this could become relevant if we ever have network upgrades activating in a different order between testnet and mainnet. In that scenario, relying on enum declaration order might introduce issues.

That said, the current behavior is already an improvement over what we had before, and it matches the intended activation ordering today. I’m fine with keeping it as is for now, but we could consider explicit sorting by activation height in the future if we anticipate more complex upgrade scenarios.

oxarbitrage (approved):
I think this is good enough. Thanks @syszery

PR #10098, opened 2025-11-15
cleanup: remove serde Serialize/Deserialize from Block, Transaction

### LINKED ISSUE #9892: cleanup: remove serde Serialize/Deserialize from Block, Transaction
### Motivation

While reviewing [another PR](https://github.com/ZcashFoundation/zebra/pull/9828#discussion_r2344090142) I noticed we have a bunch of serialization code whose only purpose seems to be ultimately support implementing serde `Serialize/Deserialize` in `Block` and `Transaction`, which in turn seem to be used only by:

- The `elastic-search` feature;
- The `snapshot_rpc_getblocktemplate` test which for some reason asserts the serde serialization of a `Transaction`.

If we could change the `elastic-search` feature to use the same serialization that `zebra-rpc` does, and if the assert mentioned could be removed, then we could probably remove a lot of code related to serde serialization, which would decrease our maintenance burden.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

This is related https://github.com/ZcashFoundation/zebra/issues/2331 but there the intent was to remove the `serde` feature altogether; however, after we became dependent on `serde` for tree serialization and `zebra-rpc`, we closed that issue since its goal was impossible. However, I think it's still worthwhile to remove the other usages of `serde` besides those.

syszery:
## Motivation

As noted in #9892, `Block` and `Transaction` include `serde` serialization only needed for the optional Elasticsearch feature (and one test). This PR removes those derives and adds dedicated structs for Elasticsearch indexing, reducing unnecessary serialization code.

## Solution

* Removes `Serialize`/`Deserialize` derives from various types in `zebra-chain`:
  * `ValueCommitment`
  * `Action`, `AuthorizedAction`, `ShieldedData`
  * `Output`, `OutputInTransactionV4`, `OutputPrefixInTransactionV5`
  * `Spend`, `SpendPrefixInTransactionV5`
  * `Transaction`, `Block`
* Added a new module `zebra_state::service::finalized_state::elastic` containing dedicated `struct`s for serializing `Block` and `Transaction` data for Elasticsearch indexing.
* Updated the `elasticsearch` method to use the new structs.
* **Note:** The snapshot test in `zebra-rpc` still depends on `serde`, so `cargo test -p zebra-rpc` will fail until that test is updated or removed.

Feedback is welcome on whether the included fields are sufficient for Elasticsearch, and whether this approach aligns with what was envisioned in #9892.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-state/src/service/finalized_state/elastic.rs
+++ b/zebra-state/src/service/finalized_state/elastic.rs
@@ -1,0 +1,86 @@
+//! Elasticsearch representations of Zebra blocks and transactions.
+//!
+//! This module provides serializable structs for indexing Zcash blocks and transactions
+//! in Elasticsearch. It extracts minimal information useful for search and analytics,
+//! including block height, hash, timestamp, and basic transaction data.
+
+use serde::Serialize;
+use zebra_chain::{block::Block, transaction::Transaction};
+
+/// A Zcash block for Elasticsearch indexing.
+#[derive(Debug, Serialize)]
+pub struct ElasticBlockObject {
+    pub height: u32,
+    pub hash: String,
+    pub time: i64,
+    pub tx_count: usize,
+    pub transactions: Vec<ElasticTxObject>,
+}
+
+/// A Zcash transaction for Elasticsearch indexing.
+#[derive(Debug, Serialize)]
+pub struct ElasticTxObject {
+    pub txid: String,
+    pub version: u32,
+    pub inputs: usize,
+    pub outputs: usize,
+}
+
+impl ElasticBlockObject {
+    /// Convert a Zebra `Block` into an `ElasticBlockObject`.
+    pub fn from(block: &Block) -> Self {
+        let height = block.coinbase_height().map(|h| h.0).unwrap_or(0);
+        let hash = block.hash().to_string();
+        let time = block.header.time.timestamp();
+
+        let transactions = block
+            .transactions
+            .iter()
+            .map(|tx| ElasticTxObject::from(tx.as_ref()))
+            .collect();
+
+        Self {
+            height,
+            hash,
+            time,
+            tx_count: block.transactions.len(),
+            transactions,
+        }
+    }
+}
+
+impl ElasticTxObject {
+    /// Convert a Zebra `Transaction` into an `ElasticTxObject`.
+    pub fn from(tx: &Transaction) -> Self {
+        let txid = tx.hash().to_string();
+
+        let (version, inputs, outputs) = match tx {
+            Transaction::V1 {
+                inputs, outputs, ..
+            } => (1, inputs.len(), outputs.len()),
+            Transaction::V2 {
+                inputs, outputs, ..
+            } => (2, inputs.len(), outputs.len()),
+            Transaction::V3 {
+                inputs, outputs, ..
+            } => (3, inputs.len(), outputs.len()),
+            Transaction::V4 {
+                inputs, outputs, ..
+            } => (4, inputs.len(), outputs.len()),
+            Transaction::V5 {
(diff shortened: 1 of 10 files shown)


natalieesk:
Thank you for this PR. We will take a look at this.

conradoplg (commented):
This looks good, but I am not sure about the elastic search aspect, @oxarbitrage what do you think, are these fields enough?

zebra-state/src/service/finalized_state/elastic.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// A Zcash block for Elasticsearch indexing.
+#[derive(Debug, Serialize)]
+pub struct ElasticBlockObject {
+    pub height: u32,
+    pub hash: String,
+    pub time: i64,
+    pub tx_count: usize,
+    pub transactions: Vec<ElasticTxObject>,
+}
```
oxarbitrage, line 18: In the previous version, the full block (including all fields) was stored in the database, which made the Kibana view significantly more useful. Limiting the stored data to this small subset of fields reduces the capabilities available to end users and makes the feature much less valuable in practice.

zebra-state/src/service/finalized_state/elastic.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// A Zcash transaction for Elasticsearch indexing.
+#[derive(Debug, Serialize)]
+pub struct ElasticTxObject {
+    pub txid: String,
+    pub version: u32,
+    pub inputs: usize,
+    pub outputs: usize,
+}
```
oxarbitrage, line 27: Only the number of inputs and outputs is currently stored. In the previous version, the full transaction data was persisted, including the raw transaction bytes when needed.

oxarbitrage (commented):
In its current form, this makes the Kibana frontend far less useful. See inline comments for specifics.

zebra-state/src/service/finalized_state/elastic.rs
syszery: I understand your concerns. The selection was meant to serve as a starting point for discussion. Do you think we should implement all the fields in the above struct, or would a larger subset also suffice?

zebra-state/src/service/finalized_state/elastic.rs
syszery: Thanks for pointing that out. I'll take a look at it.

zebra-state/src/service/finalized_state/elastic.rs
oxarbitrage: I’d prefer to keep the current approach for now, unless we find a non-manual alternative or decide to deprecate the Elasticsearch feature altogether. If we go in that direction, we should also consider the impact on zebra-rpc. At this point, I don’t think it’s worth to do it now or soon. I think @conradoplg might agree

conradoplg: I agree, my intention with https://github.com/ZcashFoundation/zebra/issues/9892 was to clean things up if possible, but I didn't realize that the elastic search feature relied so much on this. So unless there is an alternative approach that does not require creating a lot of new structs, I think it's better to leave things as is.

syszery: Thanks for the feedback. I understand the concerns regarding the implications for Kibana, and I agree if we decide to close the PR.

conradoplg:
I'm closing this for now [per discussion](https://github.com/ZcashFoundation/zebra/pull/10098#discussion_r2795732793), we might want to try again at a later time

PR #10099, opened 2025-11-16
fix(consensus): Avoid panicking when validation result channel is closed

### LINKED ISSUE #10044: bug: zebra repeatedly crashes while syncing
### What happened?

Zebra has crashed few times while syncing

_No response_

### Zebra logs

Message:  JoinHandle polled after completion
Location: /home/zebra/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.48.0/src/runtime/task/core.rs:412

Diagnostic metadata:
version: 3.0.0-rc.0
Zcash network: Mainnet
running state version: 27.0.0
initial disk state version: 27.0.0
features: default,default_release_binaries,howudoin,indicatif,metrics_exporter_prometheus,progress_bar,prometheus,release_max_level_info,sentry
branch: master
git commit: 8597342
commit timestamp: 2025-10-16T13:50:52.000000000Z
rust compiler: 1.90.0
rust release date: 2025-09-14
optimization level: 3
debug checks: false

### Zebra Version

3.0.0-rc.0

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

arya2:
## Motivation

Closes #10044

The worker task is returning [here](https://github.com/ZcashFoundation/zebra/blob/main/tower-batch-control/src/worker.rs#L285) after its [channel receiver is closed here](https://github.com/ZcashFoundation/zebra/blob/main/tower-batch-control/src/worker.rs#L324) in the `failed()` method, which is only called if `poll_ready()` or `call()` on the inner service returns an error.

Both the Sapling and Orchard verifiers always return `Poll::Ready(Ok(())` from their `poll_ready()` methods, so `failed()` is [likely being called here](https://github.com/ZcashFoundation/zebra/blob/main/tower-batch-control/src/worker.rs#L219) where it's awaiting a response to a `BatchControl::Flush` request but getting a [`SendError`](https://github.com/ZcashFoundation/zebra/actions/runs/18563585081/job/52921668116#step:12:248).

The [`BatchControl::Flush` match arm in the Orchard verifier's `call()` method is infallible](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/primitives/halo2.rs#L302), and there are no layers added when [instantiating the `Batch` service around it](https://github.com/ZcashFoundation/zebra/blob/main/zebra-consensus/src/primitives/halo2.rs#L141). The [arm was infallible in the Sapling verifier too prior to #9737](https://github.com/ZcashFoundation/zebra/pull/9737/files#diff-49d789823381c8e99c79885d1f62574f0808ad439a348fd11488b1aed14a5a65L522), but now, in rare cases, I think the transaction verifier could do some checks concurrently then drop the `FuturesUnordered`/`AsyncChecks` struct that owns the only receiver for that watch channel `Sender` when one of them fails such that the Sapling verifier has nowhere to send the validation result and returns a `SendError` causing the batch worker to return. 

## Solution

- Always returns `Ok(())` from `BatchControl::Flush` match arm of Sapling verifier's `call()` method rather than potentially returning a `SendError` if there the validation result channel is closed.
- ~Avoids polling the `tower_batch_control::Batch` worker task from the service's `poll_ready()` method when the worker task has already finished, instead logging a warning and returning an error early (so that the request goes to the fallback service instead).~
- Adds a correctness comment in `Batch::poll_ready()` noting that the inner service should not return recoverable errors in some cases.

Related:
- Minor fix to `failed()` method docs.

### Follow-up Work

Release v3.0.0

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/tower-batch-control/src/service.rs
+++ b/tower-batch-control/src/service.rs
@@ -245,14 +245,21 @@ where
             .expect("previous task panicked while holding the worker handle mutex")
             .as_mut()
         {
-            if worker_handle.is_finished() {
-                let worker_error = self.get_worker_error();
-                tracing::warn!(?worker_error, "batch worker handle finished unexpectedly");
-                return Poll::Ready(Err(worker_error));
-            }
-
+            // # Correctness
+            //
+            // The inner service used with `Batch` MUST NOT return recoverable errors from its:
+            // - `poll_ready` method, or
+            // - `call` method when called with a `BatchControl::Flush` request.
+            //
+            // If the inner service returns an error in those cases, this `poll_ready` method will
+            // return an error the first time its called, and will panic the second time its called
+            // as it attempts to call `poll` on a `JoinHandle` that has already completed.
             match Pin::new(worker_handle).poll(cx) {
-                Poll::Ready(Ok(())) => return Poll::Ready(Err(self.get_worker_error())),
+                Poll::Ready(Ok(())) => {
+                    let worker_error = self.get_worker_error();
+                    tracing::warn!(?worker_error, "batch worker finished unexpectedly");
+                    return Poll::Ready(Err(worker_error));
+                }
                 Poll::Ready(Err(task_cancelled)) if task_cancelled.is_cancelled() => {
                     tracing::warn!(
                         "batch task cancelled: {task_cancelled}\n\
--- a/tower-batch-control/src/service.rs
+++ b/tower-batch-control/src/service.rs
@@ -245,6 +245,12 @@ where
             .expect("previous task panicked while holding the worker handle mutex")
             .as_mut()
         {
+            if worker_handle.is_finished() {
+                let worker_error = self.get_worker_error();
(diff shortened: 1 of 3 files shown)


mpguerra:
Does this explain why we seem to be getting them more frequently since we switched to Sapling and Orchard batch validation then? Or is this unrelated?

conradoplg:
> Does this explain why we seem to be getting them more frequently since we switched to Sapling and Orchard batch validation then? Or is this unrelated?

Yes, the bug addressed by this PR was introduced in that switch

arya2:
Opened #10101 to add a test as it seems it would involve creating transactions which may take a little longer than would be ideal to do for the first time here.

conradoplg (approved):
Great job finding this, thanks!

PR #10115, opened 2025-11-22
zebrad: fix VERGEN_IDEMPOTENT_OUTPUT being logged

### LINKED ISSUE #9703: bug: logged version is `VERGEN_IDEMPOTENT_OUTPUT` when using `cargo install`
### What happened?

 I expected to see this happen: when running a Zebra installed with `cargo install`, the log should contain something meaningful in the `zebrad` value

Instead, this happened: `VERGEN_IDEMPOTENT_OUTPUT` is used, e.g.:

```
2025-07-14T22:18:27.577636Z  INFO {zebrad="VERGEN_IDEMPOTENT_OUTPUT" net="Main"}: zebrad::commands::start: Starting zebrad
```

_No response_

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

### LINKED ISSUE #10105: `vergen` can't determine git details in Docker builds
`vergen` doesn't have access to git details when building the Zebra binary in Docker: https://github.com/ZcashFoundation/zebra/actions/runs/18574338069/job/52955887801#step:9:1638, including release builds.

The consequence is that users get outputs such as: https://github.com/ZcashFoundation/zebra/actions/runs/19364255402/job/55403605378#step:12:131, which we also get in our CI.

conradoplg:
## Motivation

Closes https://github.com/ZcashFoundation/zebra/issues/9703

## Solution

When vergen was updated, its behaviour subtly changed. Error checking (enabled by `fail_on_error()`) is now done in `add_instructions()` instead of in `emit()`. Since we were calling `fail_on_error()` before adding the Git vergen instructions, when a `.git` folder was not available (which happens in `cargo install zebrad`) then it would silently fail and vergen would set the dummy string `VERGEN_IDEMPOTENT_OUTPUT` (which is doubly confusing because that sounds related to the `idempotent()` flag which was _not_ being enabled).

The fix was to simply enable fail_on_error before and adjusting the error handling accordingly.

For additional context, the Zebra behaviour is to [add the git commit to the logs if available, but nothing otherwise](https://github.com/ZcashFoundation/zebra/blob/5049503872c413c78d53f2b61820f8d3ee183268/zebrad/src/application.rs#L488-L492). So with this fix, the commit should be not logged anymore when using `cargo install` (which should be fine; the version is logged on start anyway)

### Tests

Rename `.git`, run Zebra and see the logs.

It's hard to test this with `cargo install zebrad` because that requires publishing the crate, but it's very likely to work.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebrad/build.rs
+++ b/zebrad/build.rs
@@ -12,6 +12,7 @@ use vergen_git2::{CargoBuilder, Emitter, Git2Builder, RustcBuilder};
 /// This builder fails the build on error.
 fn add_base_emitter_instructions(emitter: &mut Emitter) {
     emitter
+        .fail_on_error()
         .add_instructions(
             &CargoBuilder::all_cargo().expect("all_cargo() should build successfully"),
         )
@@ -43,20 +44,11 @@ fn main() {
         .build()
         .expect("all_git + describe + sha should build successfully");
 
-    emitter
-        .add_instructions(&all_git)
-        .expect("adding all_git + describe + sha instructions should succeed");
-
-    // Disable git if we're building with an invalid `zebra/.git`
-    match emitter.fail_on_error().emit() {
-        Ok(_) => {}
-        Err(e) => {
-            eprintln!("git error in vergen build script: skipping git env vars: {e:?}",);
-            let mut emitter = Emitter::default();
-            add_base_emitter_instructions(&mut emitter);
-            emitter.emit().expect("base emit should succeed");
-        }
+    if let Err(e) = emitter.add_instructions(&all_git) {
+        println!("git error in vergen build script: skipping git env vars: {e:?}",);
     }
+
+    emitter.emit().expect("base emit should succeed");
 
     #[cfg(feature = "lightwalletd-grpc-tests")]
     tonic_prost_build::configure()
--- a/zebrad/build.rs
+++ b/zebrad/build.rs
@@ -8,9 +8,13 @@
 
 use vergen_git2::{CargoBuilder, Emitter, Git2Builder, RustcBuilder};
 
-/// Configures an [`Emitter`] for everything except for `git` env vars.
-/// This builder fails the build on error.
-fn add_base_emitter_instructions(emitter: &mut Emitter) {
+/// Process entry point for `zebrad`'s build script
+#[allow(clippy::print_stderr)]
+fn main() {
+    let mut emitter = Emitter::default();
+
+    // Configures an [`Emitter`] for everything except for `git` env vars.
+    // This builder fails the build on error.
     emitter
         .fail_on_error()
         .add_instructions(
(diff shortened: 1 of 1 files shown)


upbqdn:
> When vergen was updated, its behaviour subtly changed.

Could you add a reference if you have any? I tried looking it up briefly, but didn't find anything useful.

zebrad/build.rs
```diff
@@ -12,6 +12,7 @@ use vergen_git2::{CargoBuilder, Emitter, Git2Builder, RustcBuilder};
 /// This builder fails the build on error.
 fn add_base_emitter_instructions(emitter: &mut Emitter) {
```
upbqdn: Since we're calling this fn only once in `main`, it might be worth moving its content to `main` itself.

zebrad/build.rs
```diff
@@ -43,21 +44,12 @@ fn main() {
...
-    match emitter.fail_on_error().emit() {
-        Ok(_) => {}
-        Err(e) => {
-            eprintln!("git error in vergen build script: skipping git env vars: {e:?}",);
-            let mut emitter = Emitter::default();
-            add_base_emitter_instructions(&mut emitter);
-            emitter.emit().expect("base emit should succeed");
-        }
+    if let Err(e) = emitter.add_instructions(&all_git) {
+        println!("git error in vergen build script: skipping git env vars: {e:?}",);
```
upbqdn, line 52: Do you know if this is supposed to get printed when `.git` is not available? I tried it and didn't get any output.

upbqdn:
Also closes #10105 by suppressing all incorrect outputs. We use the `SHORT_SHA` arg to propagate the git commit ID to the image and logs, but that doesn't work for locally built images, and we don't need it in release images. I'll close the aforementioned issue and open a separate one.

zebrad/build.rs
conradoplg: It's only printed if `build.rs` fails for any reason. So in practice this is only working as documentation. We could emit a warning but that may be too much, or just convert to a comment.

upbqdn: > It's only printed if `build.rs` fails for any reason.

Can you point me to any docs regarding this? 

The way I understand it is that once we call `fail_on_error` on `emitter`, `add_instructions` should fail if `.git` is not available, so the message should get printed?

conradoplg: Sure, https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

> The output of the script is hidden from the terminal during normal compilation.

Yes, exactly, after we call `fail_on_error()` then `add_instructions()` returns an error if `.git` is not available.

It's not very well documented, I found out by experimenting. https://docs.rs/vergen/latest/vergen/struct.Emitter.html#method.add_instructions here it is truncated, but in the similar https://docs.rs/vergen/latest/vergen/struct.Emitter.html#method.add_custom_instructions it says 

> Errors may be generated if fail_on_error has been configured.

zebrad/build.rs
conradoplg: Done in ffef9d654, I also updated the comments

zebrad/build.rs
upbqdn: I see. Thank you.

upbqdn (approved):

PR #10134, opened 2025-11-26
feat(DoS): Implement dust filter

### LINKED ISSUE #10094: feature: dust filtering / non-standard filtering
### Motivation

Zebra never implemented the [dust filter](https://github.com/zcash/zcash/blob/2352fbc1ed650ac4369006bea11f7f20ee046b84/src/policy/policy.cpp#L122) that zcashd has.

The blockchain has been receiving a rush of txs with zero-valued outputs so we should also implement it.

It should not be a critical issue since ZIP-317 requires those to at least pay the fees so we decided to not rush it into 3.0.0.

While we do that we might as well check [the other rules](https://github.com/zcash/zcash/blob/2352fbc1ed650ac4369006bea11f7f20ee046b84/src/policy/policy.cpp#L58) for other stuff that we might have missed; create a separate issue if they're too much additional work.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

oxarbitrage:
## Motivation

We want to implement the non-consensus checks that `zcashd` performs for non-standard transactions. As a first step, this PR introduces the dust filter for transparent outputs.

Closes https://github.com/ZcashFoundation/zebra/issues/10094

## Solution

- Added an `is_standard_tx` function to check whether a transaction contains dust outputs.
- Added this check to `Mempool::Storage::insert`. At this stage, the transaction has already passed consensus and structural verification, but may still be rejected as non-standard. Transactions rejected due to dust (or any future non-standard reason) are added to the mempool’s rejected list and are not broadcast to peers.

### Tests

- Existing tests already cover successful insertion and rejection for other reasons.
- Added a new test specifically verifying that transactions are rejected due to dust.

### Follow-up Work

- [ ] Create an issue to implement the remaining non-standard transaction checks.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/transparent.rs
+++ b/zebra-chain/src/transparent.rs
@@ -463,9 +463,7 @@ impl Output {
     /// Returns true if this output is considered dust.
     pub fn is_dust(&self) -> bool {
         let output_size: u32 = self
-            .zcash_serialize_to_vec()
-            .expect("serializing to vec should not fail")
-            .len()
+            .zcash_serialized_size()
             .try_into()
             .expect("output size should fit in u32");
 
--- a/zebra-chain/src/transparent.rs
+++ b/zebra-chain/src/transparent.rs
@@ -15,6 +15,7 @@ use crate::{
     amount::{Amount, NonNegative},
     block,
     parameters::Network,
+    serialization::ZcashSerialize,
     transaction,
 };
 
@@ -62,6 +63,11 @@ pub const MIN_TRANSPARENT_COINBASE_MATURITY: u32 = 100;
 // - https://github.com/emacs-lsp/lsp-mode/issues/2080
 // - https://github.com/rust-lang/rust-analyzer/issues/13709
 pub const EXTRA_ZEBRA_COINBASE_DATA: &str = "z\u{1F993}";
+
+/// The rate used to calculate the dust threshold, in zatoshis per 1000 bytes.
+///
+/// History: https://github.com/zcash/zcash/blob/v6.10.0/src/policy/policy.h#L43-L89
+pub const ONE_THIRD_DUST_THRESHOLD_RATE: u32 = 100;
 
 /// Arbitrary data inserted by miners into a coinbase transaction.
 //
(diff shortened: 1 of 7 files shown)


zebra-chain/src/transparent.rs
```diff
@@ -454,6 +460,22 @@ impl Output {
...
     }
+
+    /// Returns true if this output is considered dust.
+    pub fn is_dust(&self) -> bool {
+        let output_size: u32 = self
+            .zcash_serialize_to_vec()
+            .expect("serializing to vec should not fail")
+            .len()
+            .try_into()
+            .expect("output size should fit in u32");
```
conradoplg, line 468: This is a bit more efficient:

```suggestion
        let output_size: u32 = self
            .zcash_serialized_size()
            .try_into()
            .expect("output size should fit in u32");
```

zebrad/src/components/mempool/storage.rs
```diff
@@ -205,6 +217,31 @@ impl Storage {
...
+    /// Check whether a transaction is standard.
+    ///
+    /// Zcashd defines non-consensus standard transaction checks in
+    /// https://github.com/zcash/zcash/blob/v6.10.0/src/policy/policy.cpp#L58-L135
+    ///
+    /// This checks are applied before inserting a transaction in `AcceptToMemoryPool`:
+    /// https://github.com/zcash/zcash/blob/v6.10.0/src/main.cpp#L1819
+    ///
+    /// Currently, we only implement the dust output check.
+    fn is_standard_tx(&mut self, tx: &VerifiedUnminedTx) -> Result<(), MempoolError> {
```
conradoplg: Optional: the name is a bit confusing since it has the side effect of possibly actually rejecting the tx. Maybe `reject_if_non_standard_tx`?

zebrad/src/components/mempool/storage.rs
```diff
@@ -224,6 +261,9 @@ impl Storage {
         spent_mempool_outpoints: Vec<transparent::OutPoint>,
         height: Option<Height>,
     ) -> Result<UnminedTxId, MempoolError> {
+        // Check that the transaction is standard.
+        self.is_standard_tx(&tx)?;
+
```
conradoplg: This should be moved below the check for cached rejections below, because that check is faster.

conradoplg (changes requested):
Looks good, but I think it needs some small adjustments

conradoplg (approved):

conradoplg:
Re-running tests; flaky ones are failing

PR #10145, opened 2025-11-27
 chore(release): Release v3.1.0

arya2:
---
name: 'Release Checklist Template'
about: 'Checklist to create and publish a Zebra release'
title: 'Release Zebra (version)'
labels: 'A-release, C-exclude-from-changelog, P-Critical :ambulance:'
assignees: ''

---

# Prepare for the Release

- [ ] Make sure there has been [at least one successful full sync test in the main branch](https://github.com/ZcashFoundation/zebra/actions/workflows/zfnd-ci-integration-tests-gcp.yml?query=branch%3Amain) since the last state change, or start a manual full sync.

# Checkpoints

For performance and security, we want to update the Zebra checkpoints in every release.
- [ ] You can copy the latest checkpoints from CI by following [the zebra-checkpoints README](https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/README.md#zebra-checkpoints).

# Missed Dependency Updates

Sometimes `dependabot` misses some dependency updates, or we accidentally turned them off.

This step can be skipped if there is a large pending dependency upgrade. (For example, shared ECC crates.)

Here's how we make sure we got everything:
- [ ] Run `cargo update` on the latest `main` branch, and keep the output
- [ ] If needed, [add duplicate dependency exceptions to deny.toml](https://github.com/ZcashFoundation/zebra/blob/main/book/src/dev/continuous-integration.md#fixing-duplicate-dependencies-in-check-denytoml-bans)
- [ ] If needed, remove resolved duplicate dependencies from `deny.toml`
- [ ] Open a separate PR with the changes
- [ ] Add the output of `cargo update` to that PR as a comment

# Summarise Release Changes

These steps can be done a few days before the release, in the same PR:

## Change Log

**Important**: Any merge into `main` deletes any edits to the draft changelog.
Once you are ready to tag a release, copy the draft changelog into `CHANGELOG.md`.

We use [the Release Drafter workflow](https://github.com/marketplace/actions/release-drafter) to automatically create a [draft changelog](https://github.com/ZcashFoundation/zebra/releases). We follow the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format.

To create the final change log:
- [ ] Copy the [**latest** draft
  changelog](https://github.com/ZcashFoundation/zebra/releases) into
  `CHANGELOG.md` (there can be multiple draft releases)
- [ ] Delete any trivial changes
    - [ ] Put the list of deleted changelog entries in a PR comment to make reviewing easier
[...]

PR #10150, opened 2025-11-28
fix(ci): Login to DockerHub for Docker Scout on release PRs

gustavovalverde:
## Motivation

Fix Docker Scout authentication failure on release PRs.

The error was: `could not authenticate: user githubactions not entitled to use Docker Scout`

See: https://github.com/ZcashFoundation/zebra/actions/runs/19766249660/job/56639855916?pr=10145

## Solution

Docker Scout requires DockerHub authentication to run. The DockerHub login step in the `merge` job was only configured for release events:

```yaml
if: inputs.publish_to_dockerhub && github.event_name == 'release' && !github.event.release.prerelease
```

However, Docker Scout also runs on PRs with "Release v" in the title:

```yaml
if: inputs.dockerfile_target == 'runtime' && contains(github.event.pull_request.title, 'Release v')
```

This mismatch caused the login to be skipped on release PRs while Docker Scout still attempted to run.

The fix extends the DockerHub login condition to also run when Docker Scout will execute.

### Tests

- This will be validated when the CI runs on this PR or when PR #10145 is rebased on main after this merges.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.github/workflows/zfnd-build-docker-image.yml
+++ b/.github/workflows/zfnd-build-docker-image.yml
@@ -325,8 +325,11 @@ jobs:
           username: oauth2accesstoken
           password: ${{ steps.auth.outputs.access_token }}
 
+      # Login to DockerHub when:
+      # 1. Publishing images for a release (not prerelease), OR
+      # 2. Running Docker Scout on release PRs (requires DockerHub authentication)
       - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


arya2 (approved):
We should update our release checklist to note that the title of the release PR should match this format if we want to keep this logic medium/long-term.

Issue #10136, opened 2025-11-27
Refactor test architecture


natalieesk:
## Purpose

As part of the CI rebuild we want to change the test architecture.

### 1. Unit Tests
Fast tests for CLI, config parsing, basic functionality. No network or state needed.

- `generate_no_args`, `generate_args`
- `help_no_args`, `help_args`
- `start_no_args`, `start_args`
- `version_no_args`, `version_args`
- `config_tests`, `stored_configs_*`
- `ephemeral_*`, `persistent_mode`

**Runs**: Every PR (~1 min)

### 2. Integration Tests
Zebrad needs to be running, but no cached state required.

- `sync_one_checkpoint_*`
- `restart_stop_at_height`
- `metrics_endpoint`, `tracing_endpoint`
- `rpc_endpoint_*`
- `zebra_*_conflict`
- `delete_old_databases`
- `end_of_support_is_checked_at_start`

**Runs**: Every PR (5-15 min)

### 3. Stateful Tests
Require cached blockchain state (as @conradoplg listed):

| Test | State Required |
|------|----------------|
| `sync_to_mandatory_checkpoint_*` | Long but stateless |
| `sync_past_mandatory_checkpoint_*` | Mandatory checkpoint state |
| `sync_update_mainnet` | Synced zebra state |
| `lwd_rpc_test` | Synced zebra state |
| `rpc_get_block_template` | Synced zebra state |
| `rpc_submit_block` | Synced zebra state |
| `fully_synced_rpc_z_getsubtreesbyindex_snapshot_test` | Synced zebra state |
| `has_spending_transaction_ids` | Synced zebra state + indexer |
| `lwd_sync_update` | Synced zebra + lwd states |
| `lwd_rpc_send_tx` | Synced zebra + lwd states |
| `lwd_grpc_wallet` | Synced zebra + lwd states |
| `lightwalletd_test_suite` | Behavior varies with state |

**Runs**: Weekly schedule, or via `run-stateful-tests` label (30-60 min)

### 4. E2E Tests
Long-running full chain sync tests.

- `sync_full_mainnet`, `sync_full_testnet`
- `lwd_sync_full`
- `generate_checkpoints_*`

**Runs**: Weekly/manual (hours)

## Quick Summary

| Category | CI Trigger | Duration |
|----------|------------|----------|
| Unit | Every PR | < 1 min |
| Integration | Every PR | 5-15 min |
| Stateful | Weekly/label | 30-60 min |
[...]

PR #10157, opened 2025-11-29
refactor(zebrad/tests): modularize acceptance tests

syszery:
## Motivation

This PR addresses the changes proposed in #10136 to reorganize the test architecture as part of the ongoing CI rebuild efforts. The primary goal is to break down `acceptance.rs` into smaller, more manageable test categories: `Unit`, `Integration`, `Stateful`, and `E2E`. This restructuring will give us more flexibility in selecting which tests to run and when, as these categories differ significantly in terms of execution time.

## Solution

This PR focuses on Phase 1 of the proposed reorganization in the mentioned issue. Specifically, it splits the large `acceptance.rs` file into several smaller test files categorized as follows:

```bash
zebrad/tests/
├── common/              # Keep as-is
├── unit/
│   ├── cli.rs
│   ├── config.rs
│   └── ephemeral.rs
├── integration/
│   ├── endpoints.rs
│   ├── sync.rs
│   └── conflicts.rs
├── stateful/
│   ├── checkpoint_sync.rs
│   ├── lightwalletd.rs
│   └── mining_rpcs.rs
└── e2e/
    ├── full_sync.rs
    └── wallet_integration.rs
```

Happy to adjust based on feedback. While all 15 unit tests have passed, both individually and in combination with the full zebrad test suite, please note that there are currently several lint warnings and errors. These will be addressed once the file split is fully complete.

### PR Checklist

- [ ] The solution is tested.


syszery:
#### Update

* Cleaned up `acceptance.rs` and fixed Clippy lints.
* All named tests have been moved to their respective new files, as outlined in the issue.
* Added a comprehensive list of tests and their new locations at the top of `acceptance.rs` for reference.

#### Review questions

* Is the current organization okay, or should some tests be redistributed?
* For any remaining uncategorized tests, should we move them into existing categories/files or create new ones (like `stateful/state_format.rs`)?

natalieesk:
Thanks for this, we'll take a look shortly

zebrad/tests/acceptance.rs
```diff
@@ -133,6 +133,7 @@
 //! ```
 
 #![allow(clippy::unwrap_in_result)]
+#![allow(unused_imports)]
```
oxarbitrage, line 136: Remove unused imports from the file instead.

zebrad/tests/common/checkpoints.rs
```diff
@@ -391,6 +391,7 @@ impl ZebraCheckpointsTestDirExt for TempDir {
 
 /// Wait for `zebra-checkpoints` to generate checkpoints, clearing Zebra's logs at the same time.
 #[tracing::instrument]
+#[allow(clippy::unwrap_in_result)]
```
oxarbitrage, line 394: If this was not needed before, it should not be needed after the refactor.

zebrad/tests/common/config.rs
```diff
@@ -29,6 +29,7 @@ use crate::common::cached_state::DATABASE_FORMAT_CHECK_INTERVAL;
 /// - an ephemeral state,
 /// - the minimum syncer lookahead limit, and
 /// - shorter task intervals, to improve test coverage.
+#[allow(clippy::unwrap_in_result)]
```
oxarbitrage, line 32: Should not be needed.

zebrad/tests/common/config.rs
```diff
@@ -156,6 +157,7 @@ pub fn os_assigned_rpc_port_config(
 /// Returns a `zebrad` config with the provided RPC port.
 ///
 /// Set `parallel_cpu_threads` to true to auto-configure based on the number of CPU cores.
+#[allow(clippy::unwrap_in_result)]
```
oxarbitrage, line 160: Should not be needed.

zebrad/tests/common/mod.rs
```diff
@@ -9,6 +9,8 @@
 //! to avoid compiling an empty "common" test binary:
 //! <https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests>
 
+#![allow(dead_code)]
```
oxarbitrage, line 12: Please avoid this. It’s preferable to annotate unused code at the exact location where it occurs, and ideally to avoid adding such annotations altogether.

zebrad/tests/e2e/full_sync.rs
```diff
@@ -0,0 +1,101 @@
+//! Full chain sync tests: verifies mainnet/testnet synchronization and optional checkpoint generation.
+
+#![allow(clippy::unwrap_in_result)]
```
oxarbitrage, line 3: Please try to remove.

oxarbitrage (commented):
Sorry for the late response. I like the overall idea, although I have some concerns around annotations and imports.

That said, the refactor itself is part of of the work, and the changes are quite substantial. Because of that, it’s very hard to tell whether the tests were moved exactly as-is or if any modifications were introduced along the way. We need a clear way to demonstrate that.

One option would be to move one (or a small number of) tests per commit so the diff is easy to review. Another, possibly more complex, option would be to use or build an automated script or tool that we can run locally to reproduce and verify the changes.

zebrad/tests/acceptance.rs
syszery: Yes, I agree. My plan was to either remove the whole module or clean up the imports once all tests are moved to the new files.

zebrad/tests/common/checkpoints.rs
syszery: I only added this where clippy errors or warnings were present. My guess was that this is a result from `acceptance.rs` allowing this for the whole module and the methods in `common` were just used in this file. But I agree/understand that I might add too many `allow`s in the process. I will take a look into this.

zebrad/tests/common/config.rs
syszery: See above.

zebrad/tests/common/config.rs
syszery: Same as above.

zebrad/tests/common/mod.rs
syszery: I understand the reasoning, but I had some issues with this. The main problem is that the code in `common` is effectively dead now, as it was only used in `acceptance.rs`. I understand that the `test` directory is treated differently from `src`, with each test being considered its own crate instead of a module.

I worked around this by creating intermediary files like `unit.rs` that pointed to specific test files (e.g., `/unit/cli.rs`), and this approach then allowed me to run the tests directly by name (which seems to not work when they are just inside a subfolder like `/test/unit/` as suggested in the issue). I believe, since the `common` code is no longer directly used now, it’s being marked as `dead_code`.

Is there a more elegant or Rust-idiomatic way to handle this situation?

syszery:
I understand your concern regarding the reviewability of the changes. Initially, I commented out the moved test code as a temporary measure to verify the changes during the process, but I realize too late that it makes the review much harder to follow.

My suggestion is to agree on a general approach first, and then I’ll redo the refactor starting from the main branch, making one file per commit (e.g., first commit `cli.rs`, then `config.rs`, etc.). Would this approach work better for you and help with the review?"

zebrad/tests/common/mod.rs
oxarbitrage: Ok, i see what you mean. We might find a workaround for it but in the meantime we can add a note on top of the annotation explaining why this is happening and a `TODO` to think on something to fix it.

oxarbitrage:
I agree with the general approach.

However, one commit per file may still be too large in some cases. I think one commit per test (or per small group of tests) would work better. Excluding new file creation and other refactors (annotations, formatting, etc.), each commit should contain only the removal of a test from the original file and its addition to the new module, and nothing else. This way, each diff is easy to review on a case-by-case basis.

I understand this may result in a large number of commits. In that case, a simple table mapping tests to commits would help navigate and understand the commit structure.

gustavovalverde:
Hey @syszery, I really liked that you took this initiative. This is a very important, although considerably big, change.

Would you mind if I take from here? As this will have several other implications for review purposes; and it might require starting from scratch as you already indicated to @oxarbitrage.

syszery:
Hi @gustavovalverde, sorry for the delayed reply — I took a few days off over Christmas.
If the implementation aligns with what you envisioned in the issue, I’m happy to redo the work starting from main and aim to finish by the end of this week. Otherwise, I’m also fine if you’d prefer to take over.

gustavovalverde:
No worries @syszery. I'll take over the PR just because I'll have to test several different scenarios with CI, and that might be way out of scope of the initial issue.

I'll ping you so you can also have a look in case you have comments. I'll keep this PR open until then (unless you prefer to close it so it's not noisy on your side).

Thank you, again!

oxarbitrage:
I think we should close this for now and revisit it when there’s clearer direction. Based on the discussion, it seems unlikely that this will land in its current form, so keeping it open may just create confusion.

Please feel free to reopen it if someone feels otherwise.

PR #10159, opened 2025-12-01
fix(ci): Prevent race condition in GCP static IP assignment

### LINKED ISSUE #10158: devops: Fix race condition in static IP assignment for GCP MIG deployments
## Describe the issue

When deploying Testnet nodes to GCP, the static IP assignment step fails with:

```
ERROR: (gcloud.compute.instances.add-access-config) Could not fetch resource:
 - At most one access config currently supported.
```

This occurs intermittently due to a race condition between the workflow script and the GCP Managed Instance Group (MIG) controller.

**Failed run:** https://github.com/ZcashFoundation/zebra/actions/runs/19770436851/job/56653235554

## Expected Behavior

Static IPs should be assigned to all MIG instances (Mainnet and Testnet) without errors.

## Current Behavior

The workflow:
1. Deletes the ephemeral access config from the instance
2. Adds a new access config with the static IP

The problem: After deleting the ephemeral IP, the MIG controller detects the instance has diverged from the template and tries to restore an ephemeral IP. This creates a race where `add-access-config` fails because the MIG already added one back.

Mainnet succeeded while Testnet failed because the timing is unpredictable.

## Possible Solution

1. **Create instance template with `--no-address`** so instances start without any external IP
2. **Reorder operations**: Set the stateful IP policy *before* adding the access config (this tells the MIG to preserve the IP and not reconcile)
3. **Remove `delete-access-config`** since there's no ephemeral IP to delete

Changes to `.github/workflows/zfnd-deploy-nodes-gcp.yml`:

In the instance template creation, add `--no-address`:
```bash
--subnet=${{ vars.GCP_SUBNETWORK }} \
--no-address \
--create-disk="${DISK_PARAMS}" \
```

In the static IP assignment loop, reorder and simplify:
```bash
# First: Configure stateful IP policy (tells MIG to preserve this IP)
gcloud compute instance-groups managed update-instances "${MIG_NAME}" \
  --instances="${INSTANCE_NAME}" \
  --stateful-external-ip "interface-name=nic0,address=${IP_ADDRESS},auto-delete=never" \
  --region "${{ vars.GCP_REGION }}"

# Then: Add the static IP access config to the instance
gcloud compute instances add-access-config "${INSTANCE_NAME}" \
  --access-config-name="external-nat" \
  --address="${IP_ADDRESS}" \
  --zone="${ZONE}"
```

The `delete-access-config` call can be removed since instances no longer start with an ephemeral IP.

## Additional Information/Context

- The stateful per-instance config is stored in the MIG, not the template
- During rolling updates, the MIG preserves stateful IPs even though the template has `--no-address`
- Existing MIGs with stateful IPs will continue to work after this change

## Is this happening on PRs?

No

## Is this happening on the main branch?

Yes (release deployments)

gustavovalverde:
## Motivation

Fixes a race condition in GCP MIG deployments where static IP assignment fails intermittently with "At most one access config currently supported".

Closes #10158

## Solution

1. Add `--no-address` to the instance template so instances start without external IPs
2. Remove `delete-access-config` call (no longer needed)
3. Reorder operations: set stateful IP policy before adding access config

The root cause was the MIG controller restoring ephemeral IPs after deletion, racing with the script trying to add static IPs.

### Tests

- Cannot be tested locally; requires a full GCP deployment
- The fix was derived from GCP documentation on stateful MIGs
- Existing MIGs with stateful IPs will continue to work (stateful per-instance config overrides template)

- [Configuring stateful IP addresses in MIGs](https://cloud.google.com/compute/docs/instance-groups/configuring-stateful-ip-addresses-in-migs)
- Failed run: https://github.com/ZcashFoundation/zebra/actions/runs/19770436851/job/56653235554

### Follow-up Work

None

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/.github/workflows/zfnd-deploy-nodes-gcp.yml
+++ b/.github/workflows/zfnd-deploy-nodes-gcp.yml
@@ -340,6 +340,7 @@ jobs:
           --image-project=cos-cloud \
           --image-family=cos-stable \
           --subnet=${{ vars.GCP_SUBNETWORK }} \
+          --no-address `# Static IPs are assigned after MIG creation; avoids race conditions` \
           --create-disk="${DISK_PARAMS}" \
           --container-mount-disk=mount-path='/home/zebra/.cache/zebra',name=${DISK_NAME},mode=rw \
           --container-stdin \
@@ -454,6 +455,8 @@ jobs:
           IFS=$'\n' read -rd '' -a INSTANCE_ROWS <<<"$INSTANCE_DATA" || true
 
           # Loop to assign all IPs
+          # Note: Instance template uses --no-address, so instances start without external IP.
+          # This avoids race conditions with MIG trying to restore ephemeral IPs.
           for i in "${!INSTANCE_ROWS[@]}"; do
             [ -z "${IP_ADDRESSES[$i]}" ] && continue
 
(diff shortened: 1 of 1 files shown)


upbqdn (approved):

gustavovalverde:
Admin-merging as it's approved.

PR #10170, opened 2025-12-01
feat: Precompute coinbase tx and respond quickly on tip change

### LINKED ISSUE #9727: Respond quickly to long-polled `getblocktemplate` RPC on new chain tips

upbqdn:
## Motivation

- Close #9727.

## Solution

Precompute the coinbase tx for an empty block that will sit on the new tip. We will return this provisional block upon a chain tip change so that miners can mine on the newest tip, and don't waste their effort on a shorter chain while we compute a new template for a properly filled block. We do this precomputation before we start waiting for a new tip since computing the coinbase tx takes a few seconds if the miner mines to a shielded address, and we want to return fast when the tip changes.

### Tests

No new tests.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-rpc/CHANGELOG.md
+++ b/zebra-rpc/CHANGELOG.md
@@ -11,7 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `client::TransactionTemplate::from_coinbase()`; replaced by `new_coinbase()`
 - `server::http_request_compatibility::With` trait and its impls on `HttpRequestMiddleware`
-- `generate_coinbase_and_roots()`; replaced by `new_coinbase_with_roots()`
+- `fetch_state_tip_and_local_time()`; replaced by `fetch_chain_info()`
+- `generate_coinbase_and_roots()`; replaced by `TransactionTemplate::new_coinbase()` and `DefaultRoots::from_coinbase()`
+
+### Changed
+
+- `client::DefaultRoots::from_coinbase()` return type changed from `Result<Self, Box<dyn Error>>` to `Self`
+- `client::TransactionTemplate::new_coinbase()` parameter changed from `mempool_txs` to `txs_fee: Amount<NonNegative>`
 
 ### Added
 
@@ -27,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `MinerAddressTypeIter` struct (strum-derived iterator for `MinerAddressType`)
   - `Config::miner_memo` field
   - `default_miner_address()` function
-- `new_coinbase_with_roots()`; replacement for removed `generate_coinbase_and_roots()`
+- `fetch_chain_info()`; replacement for removed `fetch_state_tip_and_local_time()`
 
 ## [7.0.0] - 2026-05-01
 
--- a/zebra-rpc/src/lib.rs
+++ b/zebra-rpc/src/lib.rs
@@ -16,9 +16,6 @@ pub mod sync;
 mod tests;
 
 pub use methods::types::{
-    get_block_template::{
-        fetch_state_tip_and_local_time, new_coinbase_with_roots,
-        proposal::proposal_block_from_template, MinerParams,
-    },
+    get_block_template::{fetch_chain_info, proposal::proposal_block_from_template, MinerParams},
     submit_block::SubmitBlockChannel,
 };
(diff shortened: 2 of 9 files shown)


arya2 (dismissed):
This looks good once the base PR has merged.

Zk-nd3r:
Tested this branch against our s-nomp pool on mainnet.

GBT returns successfully after mempool activation.  Coinbase tx is 6086 bytes,  version 0x80000005 (v5 with fOverwintered),  branch ID 0x4dec4df0 (NU6.1).  All fields s-nomp expects are present: coinbasetxn.data,  coinbasetxn.hash,  defaultroots.merkleroot.

New `authdigest` field in coinbasetxn is additive,  doesn't break existing pool software.

No regressions found in the GBT response format.  The precomputed coinbase path produces a valid template that s-nomp can consume without modification.

upbqdn:
@Zk-nd3r, out of curiosity, did you examine the raw communication between S-nomp and Zebra? If so, what tools did you use?

Zk-nd3r:
I did not do a full wire capture yet.  What I tested was end to end pool side consumption of `getblocktemplate` from our `s-nomp` setup on mainnet.

Main checks were `coinbasetxn.data`, `coinbasetxn.hash`, `defaultroots.merkleroot`, tx version, branch id, and whether the new `authdigest` field stayed additive for existing pool behavior.  `s-nomp` accepted the template without modification and I did not hit a response shape regression.

So this was compatibility validation from the pool side, not a packet level trace.  If useful, I can do a cleaner capture pass next and share the redacted request and response pairs plus a tip change latency check, since that seems closer to the actual value of this PR.

upbqdn:
> If useful, I can do a cleaner capture pass next and share the redacted request and response pairs plus a tip change latency check, since that seems closer to the actual value of this PR.

No need for now. I was wondering if you used any fancy tooling that we could utilize in our integration tests. We don't have much coverage specifically for mining atm.

Zk-nd3r:
Nothing fancy.  s-nomp pointed at Zebra and a shell loop diffing GBT fields before and after the binary swap.

The reusable part is just a checklist: getblocktemplate returns clean, coinbasetxn parses as valid v5 with fOverwintered set and correct branch ID, all fields the pool expects are present, longpoll returns a new template after a block, submitblock does not error.

We have not automated submitblock yet.  If a script like that is useful as a starting point for mining integration coverage I can put it together and PR it.

oxarbitrage:
Reviewed the latest commit (e65ea09b). The height mismatch fix is correct — if multiple blocks arrive between precomputation and the tip-change notification (or a reorg happens), the precomputed coinbase is properly discarded and a fresh one is computed.

However, there is a bug in the existing code that defeats the optimization:

**`unwrap_or` should be `unwrap_or_else`**

The current code eagerly evaluates the fallback:

```rust
let coinbase_txn = precomputed_coinbase.unwrap_or(
    TransactionTemplate::new_coinbase(...)
);
```

`unwrap_or()` evaluates its argument immediately, so the expensive Sapling/Orchard proof generation (~2.5s) runs every time, even when `precomputed_coinbase` is `Some`. This should be:

```rust
let coinbase_txn = precomputed_coinbase.unwrap_or_else(|| {
    TransactionTemplate::new_coinbase(...)
        .expect("valid coinbase tx")
});
```

This way the fallback only runs when the precomputed coinbase was discarded (height mismatch or first call).

conradoplg:
@oxarbitrage good catch, fixed in 087cb8237

oxarbitrage (dismissed):
Reviewed the full PR and the last two fix commits:

- e65ea09b (conradoplg): height mismatch guard — correct. Prevents using a precomputed coinbase when multiple blocks arrive or a reorg happens between precomputation and tip change.
- 087cb82 (conradoplg): unwrap_or to unwrap_or_else — correct. Without this the expensive Sapling/Orchard proof generation ran eagerly every time, defeating the optimization.

Both fixes look good. The precomputation flow is clean: spawn_blocking for the expensive work, height check on tip change, lazy fallback when the precomputed coinbase is discarded.

Blocked on #10048 (ZIP-213) merging first since this PR targets the zip-213 branch.

oxarbitrage (approved):

PR #10185, opened 2025-12-05
change(chain): Split subsidy constants into submodules

### LINKED ISSUE #10183: Consolidate constants for block subsidies
All constants related to block subsidies are in a single module, where they share a single namespace, so the same constants for Mainnet, Testnet, and Regtest are separated by name suffixes such as `_MAINNET` and so on. Some constants are `pub`, while others are not. After looking at the code, they don't need to be exposed outside the `chain` crate.

Splitting the constants into separate modules for each net type would make the API much cleaner and allow the use of qualified name prefixes.

upbqdn:
## Motivation

- Close #10183 

## Solution

- Split up all constants into separate modules.
- Use qualified prefixes instead of name suffixes.
- Make everything `pub(crate)`.
- Move and adjust tests.

### Notes

This PR does not change any value of any constant.

### Follow-up Work

- #10188

### PR Checklist

- [x] The solution is tested.


oxarbitrage (dismissed):
Looks good, thanks!

oxarbitrage (approved):

Issue #10221, opened 2026-01-16
bug: Valid config TOML causes an exit


emersonian:
### What happened?

When a config file is passed to Zebra, if the filename does not end in ".toml" Zebra says that the file is in an invalid format. The contents of the file is valid TOML. The issue was fixed by renaming the ".conf" to ".toml" but that requirement does not fit with every build and deployment pipeline expectation.

Rather than requiring an exact filename, Zebra should validate the TOML, and not care about the filename.

Blockbook is an example of a project that uses ".conf" for all of its blockchain nodes' config files. Zebra should not care about the file extension, Zebra should care that the contents of the file is valid TOML.

Related:
https://github.com/trezor/blockbook/pull/1389

### What were you doing when the issue happened?

Trying to update [Zec.rocks' Blockbook instance](https://blockbook.zec.rocks). (a Zcash block explorer and indexer API)

### Zebra logs

```
# /opt/coins/nodes/zcash/bin/zebrad --config /opt/coins/nodes/zcash/zcash.conf                                                                                                                                                                       
error: Zebra could not load the provided configuration file and/or environment variables.This might mean you are using a deprecated format of the file, or are attempting to                                                                                                               
                     configure deprecated or unknown fields via environment variables.You can generate a valid config by running "zebrad generate", and diff it against yours to examine any format inconsistencies.                                                                       
error: zebrad fatal error: I/O operation failed: Configuration error: configuration file "/opt/coins/nodes/zcash/zcash.conf" is not of a supported file format
```

### Zebra Version

3.1.0

### Which operating systems does the issue happen on?

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

Ubuntu 24 LTS

### Additional information

The config-rs crate determines file format by extension. Since Zebra only supports TOML configuration, explicitly specifying the format removes the unnecessary dependency on filename extension, allowing users to name their config files however they prefer (e.g., zcash.conf, zebra.cfg, etc.) while ensuring the content is always parsed as TOML.

PR #10222, opened 2026-01-16
Parse Zebra's config file as TOML

emersonian:
See #10221

Zebra's config file will always be in TOML format regardless of its file extension. Currently a config file with an extension like ".conf" causes Zebra to exit even if it contains valid TOML. This patch resolves that.


conradoplg (approved):
Good catch, thanks!

PR #10224, opened 2026-01-17
mempool: add standardness checks and configurable OP_RETURN policy

zlyzol:
## Motivation

- Align Zebra mempool standardness policy with zcashd for transparent scripts.
- Allow OP_RETURN (nulldata) outputs under explicit policy knobs, matching Maya’s transparent memo use-case.

## Solution

- Implement standardness checks in mempool storage:
  - scriptSig size + push-only checks
  - standard script kinds (P2PKH/P2SH/P2PK/multisig/OP_RETURN)
  - bare multisig rejection + multisig pubkey limit
  - OP_RETURN count/size limits with configurable policy
  - dust rejection for non-OP_RETURN outputs
- Add mempool config options: `accept_datacarrier`, `max_datacarrier_bytes`.
- Add tests covering standardness reject cases.
- Document the new config knobs in the changelog and book.

### Tests

- `cargo test -p zebrad mempool_reject_`
- External: Maya Protocol mocknet/regtest simulation (Zebra fork image) - successful

- zcashd standardness rules:  
  https://github.com/zcash/zcash/blob/v6.10.0/src/policy/policy.cpp  
  https://github.com/zcash/zcash/blob/v6.10.0/src/main.cpp

### Follow-up Work

- Port additional zcashd standardness checks not covered here.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - OpenTelemetry tracing support behind the `opentelemetry` feature flag ([#10174](https://github.com/ZcashFoundation/zebra/pull/10174))
 - RPC tracing middleware with `SPAN_KIND_SERVER` spans for Jaeger Service Performance Monitoring ([#10174](https://github.com/ZcashFoundation/zebra/pull/10174))
 - Added Docker Compose observability stack with Jaeger, Prometheus, Grafana, and AlertManager for local development and testing
+- Added mempool standardness policy checks for transparent scripts (scriptSig size/push-only, standard script kinds, bare multisig rejection, OP_RETURN size/count limits).
+- Added mempool configuration options `accept_datacarrier` and `max_datacarrier_bytes` for OP_RETURN policy.
 
 
 ## [Zebra 3.1.0](https://github.com/ZcashFoundation/zebra/releases/tag/v3.1.0) - 2025-11-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,59 @@ All notable changes to Zebra are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org).
 
+## [Zebra 4.0.0](https://github.com/ZcashFoundation/zebra/releases/tag/v4.0.0) - 2025-01-20
+
+This release fixes the type of a field in the `getinfo` RPC and adds support for
+the `pingtime` and `pingwait` fields of the `getpeerinfo` RPC.
+
+It also changes the Grafana dashboards to add auto-provisioning and AlertManager
+support.
+
+This release also adds a new mainnet DNS seeder from Shielded Labs to the default
+configuration. However, if you already have a config in place, you will need
+to add it manually; add `"mainnet.seeder.shieldedinfra.net:8233"` to
+`initial_mainnet_peers`.
+
+The new OpenTelemetry support must be enabled at compile time with the
+`opentelemetry` feature, e.g. `cargo build --features=opentelemetry --release`.
+
+### Breaking Changes
+
+This release has the following breaking changes:
+
+- Changed the `getinfo` RPC `errorstimestamp` field from a string timestamp (ISO
+  UTC timestamp) to a i64 (seconds from Unix epoch) to match zcashd
+  ([#10079](https://github.com/ZcashFoundation/zebra/pull/10079)). If you rely
+  on this field, you will need to change your code to be able to interpret
+  the i64 result.
+- Always parse Zebra's config file as TOML (#10222). This allows using a config
+  file with an extension different than `.toml`. Previously, it would use the
+  format detected from the extension, so in the unlikely case you were using
+  a format different than TOML you will need to change your config to TOML.
+
+
+### Added
+
+- Added `pingtime` and `pingwait` to `getpeerinfo` RPC ([#9880](https://github.com/ZcashFoundation/zebra/pull/9880))
+- Added Grafana auto-provisioning and AlertManager ([#10171](https://github.com/ZcashFoundation/zebra/pull/10171))
+- Added OpenTelemetry distributed tracing support ([#10174](https://github.com/ZcashFoundation/zebra/pull/10174))
+- Added new Shielded Labs mainnet seeder ([#10228](https://github.com/ZcashFoundation/zebra/pull/10228))
+
+### Contributors
+
+Thank you to everyone who contributed to this release, we couldn't make Zebra without you:
+@conradoplg, @gustavovalverde and @syszery
+
 
 ## [Unreleased]
 
 ### Added
 
-- OpenTelemetry tracing support behind the `opentelemetry` feature flag ([#10174](https://github.com/ZcashFoundation/zebra/pull/10174))
-- RPC tracing middleware with `SPAN_KIND_SERVER` spans for Jaeger Service Performance Monitoring ([#10174](https://github.com/ZcashFoundation/zebra/pull/10174))
-- Added Docker Compose observability stack with Jaeger, Prometheus, Grafana, and AlertManager for local development and testing
+- Value pool metrics exposing transparent, sprout, sapling, orchard, and deferred pool balances plus total chain supply ([#10175](https://github.com/ZcashFoundation/zebra/pull/10175))
+- Peer handshake metrics for duration histograms and failure tracking by reason ([#10175](https://github.com/ZcashFoundation/zebra/pull/10175))
+- Prometheus alert rules and Grafana dashboards for value pools and RPC monitoring ([#10175](https://github.com/ZcashFoundation/zebra/pull/10175))
+- Sync pipeline, consensus verification, and RocksDB performance histograms ([#10179](https://github.com/ZcashFoundation/zebra/pull/10179))
+- RocksDB I/O latency, sync distance metrics, and Zebra Overview dashboard ([#10181](https://github.com/ZcashFoundation/zebra/pull/10181))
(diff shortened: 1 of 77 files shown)


oxarbitrage:
```
$ cargo test transaction_id_forwarding_errors_dont_stop_the_crawler
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.41s
     Running unittests src/lib.rs (target/debug/deps/tower_batch_control-7152ff34ad46b6da)

...
running 1 test

thread 'components::mempool::crawler::tests::prop::transaction_id_forwarding_errors_dont_stop_the_crawler' has overflowed its stack
fatal runtime error: stack overflow, aborting
error: test failed, to rerun pass `-p zebrad --lib`

Caused by:
  process didn't exit successfully: `/home/alfredo/zebra/pr10224/zebra/target/debug/deps/zebrad-401c2b0cfed6464d transaction_id_forwarding_errors_dont_stop_the_crawler` (signal: 6, SIGABRT: process abort signal)
$ 
```

zebrad/src/components/mempool/config.rs
```diff
@@ -41,6 +41,17 @@ pub struct Config {
     // - allow the mempool to be enabled before the genesis block is committed?
     //   we could replace `Option` with an enum that has an `AlwaysEnable` variant
     pub debug_enable_at_height: Option<u32>,
+
+    /// Whether to accept data carrier (OP_RETURN) outputs in standardness checks.
+    ///
+    /// Matches zcashd's `-datacarrier` default behavior.
+    pub accept_datacarrier: bool,
```
oxarbitrage: I think these fields could be `Option` instead, with defaults like `Some(true)` and `Some(DEFAULT_MAX_DATACARRIER_BYTES)` for `max_datacarrier_bytes`.

This way, new nodes using older configuration files (which don’t include these settings) will continue to work as before, and the corresponding checks will still be applied.

Additionally, `cargo test config_tests` should pass. To ensure this, a config file including the new fields should be added under https://github.com/zlyzol/zebra/tree/feat-mempool-standardness/zebrad/tests/common/configs

zebrad/src/components/mempool/config.rs
```diff
@@ -41,6 +41,17 @@ pub struct Config {
     // - allow the mempool to be enabled before the genesis block is committed?
     //   we could replace `Option` with an enum that has an `AlwaysEnable` variant
     pub debug_enable_at_height: Option<u32>,
+
+    /// Whether to accept data carrier (OP_RETURN) outputs in standardness checks.
+    ///
+    /// Matches zcashd's `-datacarrier` default behavior.
```
oxarbitrage: I’m not entirely sure we should replicate this behavior. Are you aware of any use cases that actually require this to be configurable?

I’m more inclined to always apply all standardness checks instead. While `zcashd` also has a `require_standard` flag to enable these checks (https://github.com/zcash/zcash/blob/v6.10.0/src/main.cpp#L1819), I’m not convinced we need to support that either.

Let me know what you think.

oxarbitrage:
`cargo test config_tests` also fails because there is no config file with the new fields present in https://github.com/zlyzol/zebra/tree/feat-mempool-standardness/zebrad/tests/common/configs

zebrad/src/components/mempool/config.rs
```diff
@@ -59,6 +70,14 @@ impl Default for Config {
...
+
+            accept_datacarrier: true,
+            max_datacarrier_bytes: DEFAULT_MAX_DATACARRIER_BYTES,
         }
     }
 }
+
+/// Default maximum size of data carrier scripts (OP_RETURN), in bytes.
+///
+/// Equivalent to zcashd's `MAX_OP_RETURN_RELAY`.
```
oxarbitrage:
```suggestion
/// Equivalent to zcashd's `MAX_OP_RETURN_RELAY`:
/// https://github.com/zcash/zcash/blob/v6.10.0/src/script/standard.h#L22-L26
```

zebrad/src/components/mempool/storage.rs
```diff
@@ -225,23 +249,108 @@ impl Storage {
     /// This checks are applied before inserting a transaction in `AcceptToMemoryPool`:
     /// <https://github.com/zcash/zcash/blob/v6.10.0/src/main.cpp#L1819>
     ///
-    /// Currently, we only implement the dust output check.
+    /// Currently, we implement the input scriptSig size/push-only checks,
+    /// standard output script checks (including OP_RETURN limits), and dust checks.
     fn reject_if_non_standard_tx(&mut self, tx: &VerifiedUnminedTx) -> Result<(), MempoolError> {
-        // TODO: implement other standard transaction checks from zcashd.
+        // https://github.com/zcash/zcash/blob/v6.10.0/src/policy/policy.cpp#L58-L135
```
oxarbitrage: Please move this reference to the function-level documentation, since it applies to all of the checks performed there.

Then, add more specific references at the constant level. For example, `MAX_STANDARD_SCRIPTSIG_SIZE` has a more specific reference here: https://github.com/zcash/zcash/blob/v6.10.0/src/policy/policy.cpp#L92-L99

oxarbitrage:
Hi @zlyzol , thanks for the PR, nice work. I’ve left an initial review with separate comments (sorry about that). Let me know if you’d like us to make any of the suggested changes; I’m happy to open a suggestion PR.

zlyzol:
> Hi @zlyzol , thanks for the PR, nice work. I’ve left an initial review with separate comments (sorry about that). Let me know if you’d like us to make any of the suggested changes; I’m happy to open a suggestion PR.

Thanks @oxarbitrage! I’m working on it now.
Some test fixes are non‑trivial because they touch proptests and standardness behavior.
For example:
- I constrained the crawler proptest error generation to a small, fixed `MempoolError` set to avoid stack‑overflow during shrinking.
- I’m addressing `NonStandardTransaction(ScriptPubKeyNonStandard)` failures now that `Storage::insert` enforces standardness, which invalidates assumptions in existing proptests.

I’ll ping you once I’ve got all tests green.

zebrad/src/components/mempool/config.rs
zlyzol: fixed

zebrad/src/components/mempool/config.rs
zlyzol: I'm not aware of any concrete use case requiring this configurable, I just mirrored `zcashd`’s behavior. I agree it’s probably unnecessary, so I removed the `accept_datacarrier` option and now we always apply the standardness checks.

zebrad/src/components/mempool/storage.rs
zlyzol: fixed

zlyzol:
> Hi @zlyzol , thanks for the PR, nice work. I’ve left an initial review with separate comments (sorry about that). Let me know if you’d like us to make any of the suggested changes; I’m happy to open a suggestion PR.

Hi @oxarbitrage - quick update: I’ve addressed all comments. All tests passed.
Please let me know if you’d like anything else changed or added.

zebrad/src/components/mempool/config.rs
oxarbitrage: What do you think about removing `max_datacarrier_bytes` from the config entirely and always using the default?

If we keep this field as a required config entry, older config files will stop working with newer nodes. I think it’s fairly common for node operators to keep using the same config file across upgrades.

If we do need to keep the field configurable, I’d suggest making it optional so older configs still work and we fall back to the default value (while still applying the standardness checks).

Let me know how you see it.

zlyzol: I'm aligned with making `max_datacarrier_bytes` configurable but optional. This keeps backward compatibility while leaving policy decisions about data carrier limits to individual operators. What do you think?

oxarbitrage: Yea, that should be good enough to me. Thanks!

README.md
```diff
@@ -101,6 +101,8 @@ Refer to the [Building and Installing
 Zebra](https://zebra.zfnd.org/user/install.html) and [Running
 Zebra](https://zebra.zfnd.org/user/run.html) sections in the book for enabling
 optional features, detailed configuration and further details.
+See the [Mempool Specification](https://zebra.zfnd.org/dev/mempool-specification.html)
+for mempool configuration options such as standardness policy and OP_RETURN limits.
```
oxarbitrage:
```suggestion
```
The full standardness policy is not described in book/src/dev/mempool-specification.md. Documenting the policy there will be nice but we don't have to do it now. Maybe just remove the reference form the main readme for now.

oxarbitrage (dismissed):
Looks great to me, thanks!

I think @upbqdn wanted to take a look as well.

oxarbitrage:
@mergifyio refresh

oxarbitrage (approved):

upbqdn:
I'll review this soon.

oxarbitrage:
> I'll review this soon.

Sorry, i was not expecting this to get merged before your review as i made the last commits before approval. It seems the rule about that we had before is not anymore valid. Happy to open another PR if you have suggestions.

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ Thank you to everyone who contributed to this release, we couldn't make Zebra wi
 - Prometheus alert rules and Grafana dashboards for value pools and RPC monitoring ([#10175](https://github.com/ZcashFoundation/zebra/pull/10175))
 - Sync pipeline, consensus verification, and RocksDB performance histograms ([#10179](https://github.com/ZcashFoundation/zebra/pull/10179))
 - RocksDB I/O latency, sync distance metrics, and Zebra Overview dashboard ([#10181](https://github.com/ZcashFoundation/zebra/pull/10181))
+- Added mempool standardness policy checks for transparent scripts (scriptSig size/push-only, standard script kinds, bare multisig rejection, OP_RETURN size/count limits).
+- Added mempool configuration option `max_datacarrier_bytes` for OP_RETURN policy.
```
upbqdn: Note that this format doesn't follow the established format in the rest of the change log, where each PR adds a single entry ending with a reference to the PR.

PR #10231, opened 2026-01-21
fix: report connection task exits as ConnectionTaskExited

Bashmunta:
Reason:
Previously, Client::poll_connection used the HeartbeatTaskExited error variant when the connection background task finished with a JoinError (panic or cancel).
This mixed up the semantics between heartbeat and connection tasks: some connection failures were reported as HeartbeatTaskExited, while other paths used ConnectionTaskExited. As a result, diagnostics and metrics that rely on
PeerError::kind() could not reliably distinguish between heartbeat and connection task exits.

Changes:
- Updated Client::poll_connection to always report unexpected termination of the connection task as PeerError::ConnectionTaskExited, both for the normal Ok(()) completion case and for JoinError cases.
- Simplified the JoinError handling by removing the is_cancelled() branch, since all connection task exits are now consistently surfaced as ConnectionTaskExited.

### DIFF
--- a/zebra-network/src/peer/client.rs
+++ b/zebra-network/src/peer/client.rs
@@ -509,23 +509,11 @@ impl Client {
                 self.set_task_exited_error("connection", PeerError::ConnectionTaskExited)
             }
             Err(error) => {
-                // Connection task panicked.
-                let error = error.panic_if_task_has_panicked();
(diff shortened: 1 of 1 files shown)


arya2 (approved):
This looks like it may have originally been copy-pasted incorrectly, nice catch, and thank you for the fix!

PR #10245, opened 2026-01-29
perf: cache restorable db versions during old db cleanup

GarmashAlex:
## Motivation

Previously `check_and_delete_database` called `restorable_db_versions()` for every directory entry while scanning old database folders. This rebuilt the same `Vec<u64>` many times in a tight loop, causing unnecessary allocations and work during cleanup.

## Solution

Compute `restorable_db_versions()` once in `delete_old_databases` and pass the resulting slice into `check_and_delete_database`, which now takes `restorable_db_versions: &[u64]` instead of calling the function internally. This preserves existing behavior while avoiding repeated allocations.

### DIFF
--- a/zebra-state/src/config.rs
+++ b/zebra-state/src/config.rs
@@ -277,6 +277,8 @@ fn delete_old_databases(config: Config, db_kind: String, major_version: u64, net
 
     info!(db_kind, "checking for old database versions");
 
+    let restorable_db_versions = restorable_db_versions();
+
     let mut db_path = config.db_path(&db_kind, major_version, network);
     // Check and remove the network path.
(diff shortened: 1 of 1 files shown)


oxarbitrage (approved):
I see the performance gain here is quite small, since deleting old databases happens at Zebra startup in a dedicated task. That said, you’re right that this computation doesn’t need to live inside the loop, so we can accept the change.

Thanks for the contribution. For future PRs (especially if they go beyond small, localized changes) please reach out first to check whether the work aligns with current priorities. Unsolicited or larger changes can be difficult for us to review and may not be accepted or reviewed next time.

PR #10248, opened 2026-01-30
fix(net): accurate outbound connection progress bar

### LINKED ISSUE #7981: The number of outbound connections reported by progress bars periodically glitches
### Description

When syncing on Testnet, I noticed the number of outbound connections reported by the new progress bars periodically jumps from ~7 to the limit of 75 and then back to ~7. The period seems to be equal to the `crawl_new_peer_interval` setting.

### Analysis

That's probably because the connection counter actually comes from the connection limit, which is updated when we start making a connection. (Not when we successfully connect.)

So every few minutes, we re-try all the failed peers at the same time, the connection limit goes to 75, and then it drops back when they all fail.

### Possible Solutions

Reducing the progress bar update interval would hide this, but it would still occasionally happen. (And be visible for longer.)

The real fix would be to move the progress bar to the peer set. We track inbound/outbound on the [ConnectedAddr type in ConnectionInfo](https://github.com/ZcashFoundation/zebra/blob/b5e16a6b26d05f5d901a8537f324fa001451753f/zebra-network/src/peer/handshake.rs#L131), which is part of the LoadTrackedClient in the PeerSet.

czarcas7ic:
Closes: #7981

## Motivation

The "Outbound Connections" progress bar periodically spikes to 75 (the configured connection limit) during each peer crawl cycle. This is misleading because `ActiveConnectionCounter` in `limit.rs` counts connection attempts (pending handshakes), not established connections. Every 15 seconds when the crawler fires, it opens many new handshakes simultaneously, inflating the counter before the attempts time out or fail. The "Inbound Connections" bar had the same problem.

## Solution

Move the connection progress bars from `ActiveConnectionCounter` into `PeerSet`, which only contains peers that have completed their handshake.

- Add `is_outbound()` method to `ConnectedAddr` and `LoadTrackedClient` to identify outbound peers.
- Remove all `#[cfg(feature = "progress-bar")]` code from `ActiveConnectionCounter`. The counters continue to function for rate-limiting.
- Add `outbound_peers` and `inbound_peers` `HashSet<PeerSocketAddr>` fields to `PeerSet`, populated when peers arrive via `Change::Insert` in `poll_discover()` (after the handshake completes) and cleaned up on all exit paths: `remove()`, `poll_unready()` (banned/canceled/dropped/errored), `poll_ready_peer_errors()` (banned/errored), `push_ready()` (version too low), `disconnect_from_outdated_peers()`, and `shut_down_tasks_and_channels()`.
- Add new progress bars in `PeerSet` for both outbound and inbound connections, driven by `outbound_peers.len()` and `inbound_peers.len()`.
- Add new Prometheus gauges `zcash.net.peers.outbound` and `zcash.net.peers.inbound`.

### Tests

- `cargo check -p zebra-network --features progress-bar` and without the feature — both compile cleanly.
- `cargo clippy -p zebra-network --features progress-bar` — no new warnings.
- `cargo test -p zebra-network` — all 173 tests pass, no regressions.
- Manual Testnet run with `crawl_new_peer_interval = "15s"` and `progress_bar = "summary"`. Added temporary `info!` logs to both the old counter and the new PeerSet tracker to compare side-by-side:
  - **Old counter** (`ActiveConnectionCounter`): climbed 9 → 28 during a single crawl cycle.
  - **New tracker** (`PeerSet.outbound_peers`): stayed at 9 the entire time.
- Prometheus metric `zcash_net_peers_outbound` confirmed stable at 9-10 across 4+ crawl cycles (60 seconds of polling at 5-second intervals).

To test inbound, I spun up two instances, connected them to one another, and verified the inbound display was correct (I did not naturally get any inbounds so forced it this way):

Screenshot 2026-01-30 at 3 28 50 PM

- Issue: https://github.com/ZcashFoundation/zebra/issues/7981

### Follow-up Work

None.

### PR Checklist

- [x] The solution is tested.


zebra-network/src/peer_set/set.rs
```diff
@@ -700,6 +756,12 @@ where
                         continue;
                     }
 
+                    if svc.is_outbound() {
+                        self.outbound_peers.insert(key);
+                    } else {
+                        self.inbound_peers.insert(key);
```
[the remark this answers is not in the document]
czarcas7ic: Addressed here:

https://github.com/ZcashFoundation/zebra/pull/10248/commits/0e4f59d7d1d985bcf0d97e569920c2f69252d7dd

zebra-network/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+
+### Fixed
```
oxarbitrage, line 16:
```suggestion
### Added
- Added new Prometheus gauges `zcash.net.peers.outbound` and `zcash.net.peers.inbound` (#10248). 

### Fixed
```

zebra-network/src/peer_set/set.rs
```diff
@@ -711,6 +773,18 @@ where
...
+                .iter()
+                .filter(|(_, peer)| peer.remote_version() < minimum_version)
+                .map(|(addr, _)| *addr)
+                .collect();
+
+            for key in &outdated {
+                self.outbound_peers.remove(key);
+                self.inbound_peers.remove(key);
+            }
+
```
oxarbitrage: Optional: No intermediate allocation.

```suggestion
            for addr in self
                .ready_services
                .iter()
                .filter(|(_, peer)| peer.remote_version() < minimum_version)
                .map(|(addr, _)| *addr)
            {
                self.outbound_peers.remove(&addr);
                self.inbound_peers.remove(&addr);
            }

```

oxarbitrage (dismissed):
Looks good to me, nice work.

arya2:
Thank you for changes! I tried to make this a bit simpler [here](https://github.com/ZcashFoundation/zebra/tree/accurate-connection-count-metrics) with some alternative changes that keeps the logic for updating progress bars in `ActiveConnectionCounter`. It would be nice to separate reserved slots for connection attempts from the connection count while addressing this issue to pave the way for a more flexible configuration in the future.

zebra-network/CHANGELOG.md
czarcas7ic: https://github.com/ZcashFoundation/zebra/pull/10248/commits/b6571b568eb0af1941f4c0a8a5a0abf354c4c714

czarcas7ic:
Thanks for the review @oxarbitrage!

@arya2 just for clarity, would you like me to close this PR in favor of your approach, or utilize your approach in this PR? I am happy to do whatever is easiest for you all.

zebra-network/src/peer_set/set.rs
czarcas7ic: Applied your suggestion, thanks

arya2:
> @arya2 just for clarity, would you like me to close this PR in favor of your approach, or utilize your approach in this PR? I am happy to do whatever is easiest for you all.

I'd be happy with either as well, I think it would be nice to continue with the approach in this PR and would also be happy to go ahead with the other PR.

czarcas7ic:
I don't think the failed docker image build is related to my change, I think it maybe just timed out?

natalieesk:
Hi, thank you for your work on this issue. Would you like to take the approach from #10311 here, or are you happy with us closing this in favour of that PR?

czarcas7ic:
@natalieesk I am fine with Arya taking over on his PR, no worries. Feel free to close as needed.

oxarbitrage:
> @natalieesk I am fine with Arya taking over on his PR, no worries. Feel free to close as needed.

Thanks, as https://github.com/ZcashFoundation/zebra/pull/10311 is approved I think we should close this one.

PR #10263, opened 2026-02-03
fix(tests): wait for process termination and use read-only state init

gustavovalverde:
## Motivation

Fix intermittent CI failures in integration tests caused by database lock race conditions.

**Failing jobs:**
- [submit block / Run rpc-submit-block test](https://github.com/ZcashFoundation/zebra/actions/runs/21543965537/job/62082896089)
- [Lightwalletd send transactions / Run lwd-rpc-send-tx test](https://github.com/ZcashFoundation/zebra/actions/runs/21543965537/job/62082895120)

Both tests fail with:
```
Message: Database likely already open "/home/zebra/.cache/zebra/state/v27/mainnet"
Hint: Check if another zebrad process is running.
```

## Root Cause Analysis

Two contributing factors:

### 1. Test process termination
After killing a zebrad process with `kill()`, the code was using arbitrary `sleep()` delays instead of properly waiting for the process to terminate. This race condition meant the test could try to open the database before the OS had released the lock.

### 2. Orphaned metrics export task
`StateService::new()` spawns a metrics export task that holds a clone of the database and runs forever:

```rust
let db_for_metrics = read_service.db.clone();
tokio::spawn(async move {
    loop {
        interval.tick().await;
        db_for_metrics.export_metrics();
    }
});
```

When `load_tip_height_from_state_directory()` called `zebra_state::init()`, this task kept the database lock held even after the function returned.

### 1. Wait for process termination (`acceptance.rs`, `sync.rs`, `cached_state.rs`)

Add `wait_with_output()` calls after `kill()` to ensure the OS has reaped the process, followed by a short sleep as a safety margin for resource release. `kill()` only sends SIGKILL but doesn't wait for the process to terminate.

### 2. Use read-only state initialization (`cached_state.rs`)

Changed `load_tip_height_from_state_directory()` → `load_finalized_tip_height_from_state_directory()` to use `init_read_only()` instead of `init()`. Read-only initialization doesn't spawn background tasks, so the database is properly released when the function returns.

## Changes

1. **`cached_state.rs`**:
   - Renamed to `load_finalized_tip_height_from_state_directory()` for accuracy
   - Uses `init_read_only()` to avoid spawning background tasks that hold database locks
   - `wait_with_output()` + sleep after killing zebrad
2. **`acceptance.rs`**: Added `wait_with_output()` + sleep after each `kill()` call in `restores_non_finalized_state_and_commits_new_blocks()`
3. **`sync.rs`**: Added `wait_with_output()` + sleep after `kill()`, taking `dir` before waiting

## PR Checklist

- [x] The solution is tested.


zebra-state/src/service/finalized_state/disk_db/tests.rs
```diff
@@ -2,10 +2,21 @@
 
 #![allow(clippy::unwrap_in_result)]
 #![allow(dead_code)]
+#![allow(unused_imports)]
```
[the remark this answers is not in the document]
gustavovalverde: this test module uses `#[cfg(any(test, feature = "proptest-impl"))]`, meaning it's compiled for the lib target when proptest-impl feature is enabled, but the test function imports are only used when actually running tests

gustavovalverde:
This confirms the fix is working https://github.com/ZcashFoundation/zebra/actions/runs/21670777919 but not that it's appropriate

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -1518,6 +1518,33 @@ impl DiskDb {
             }
         }
 
+        // Wait for all background compaction and flush jobs to complete before dropping.
+        // This ensures the database lock is properly released when the DB object is dropped,
+        // preventing "Database likely already open" errors when another process tries to
```
arya2:
```suggestion
        // This ensures that the process will wait for background flush and compaction tasks to complete
        // before the database lock is released when the DB object is dropped,
        // preventing "Database likely already open" errors when another process tries to
```

zebra-state/src/service/finalized_state/disk_db.rs
```diff
@@ -1518,6 +1518,33 @@ impl DiskDb {
...
+                    "expected shutdown error waiting for background jobs to complete"
+                );
+            } else {
+                info!(
+                    ?error,
+                    ?path,
+                    "unexpected error waiting for background jobs during shutdown"
+                );
+            }
+        }
```
arya2: This should probably go after the `self.delete_ephemeral();` call as it's not necessary to wait for background tasks if the state is being discarded anyway.

This function could block for a while, and the OS could halt the thread anyway, corrupting the rocksdb state. It's probably best to exit immediately waiting only for strictly required tasks, i.e. the default background tasks apparently.

zebrad/tests/acceptance.rs
```diff
@@ -4081,8 +4081,10 @@ async fn restores_non_finalized_state_and_commits_new_blocks() -> Result<()> {
     tokio::time::sleep(Duration::from_secs(6)).await;
 
     child.kill(true)?;
-    // Wait for Zebra to shut down.
-    tokio::time::sleep(Duration::from_secs(3)).await;
+    // Wait for zebrad to fully terminate to ensure database lock is released.
+    child
+        .wait_with_output()
```
arya2, line 4086: `wait_with_output()`, or `wait()` more generally, could return an `Ok` with an exit status indicating that the child process continued after `wait()` returned. I see no harm in sleeping after killing child processes just in case.

zebra-state/src/service/finalized_state/disk_db/tests.rs
arya2, line 1: This new test passes on main already.

arya2 (commented):
The changes in `zebrad/tests` look good, I think the other changes in this PR should be discarded.

I've just noticed this TODO,

```rust
        // Try re-enabling this code and fixing the underlying concurrency bug.
        //
        // info!(?path, "stopping background database tasks");
        // self.db.cancel_all_background_work(true);
```

it seems strange that cancelling background work would cause any issues other than losing a write (or even losing a write in this case since there's a WAL).

arya2 (approved):

PR #10251, opened 2026-01-30
change(CI): Run scheduled full sync on Testnet

upbqdn:
## Motivation

- Close #10244

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [x] The library crate changelogs are up to date.
- [ ] The solution is tested.
- [x] The documentation is up to date.

PR #10264, opened 2026-02-03
fix(ci): remove unsupported `continue-on-error` from reusable workflow calls

gustavovalverde:
## Motivation

Fix issues with the `zfnd-ci-integration-tests-gcp` workflow:

1. **Workflow parsing errors** since #10251 - `continue-on-error` is not supported at the caller job level for reusable workflows
2. **GCP instance name too long** - Instance names exceed GCP's 63-character limit when branch names are long

### Fix 1: Move continue-on-error inside reusable workflow

`continue-on-error` is not supported at the caller job level for jobs that use reusable workflows (`uses:`). The fix moves it inside the reusable workflow where it's valid.

- Add `allow_failure` input to `zfnd-deploy-integration-tests-gcp.yml`
- Use `continue-on-error: ${{ inputs.allow_failure }}` on the `test-result` job
- Pass `allow_failure: true` from testnet jobs

### Fix 2: Truncate branch name for GCP instance names

GCP instance names are limited to 63 characters. The pattern `{test_id}-{branch}-{sha}` exceeds this when both test_id and branch are long.

- Add `slug-maxlength: 23` to all uses of `github-slug-action` to limit branch slug length
- This keeps instance names under the 63-character limit

### Fix 3: Make disk creation idempotent

Defensive fix for edge cases where a workflow run fails after creating the disk but before the instance is created (e.g., due to name length errors). Without this, re-running the failed workflow would fail with "disk already exists".

- Check if disk exists before creating
- Reuse existing disk if present

### Tests

- Verified the workflow can be triggered manually via `gh workflow run`
- Workflow successfully creates instances with truncated names

- [GitHub Community Discussion: Cannot use continue-on-error with reusable workflows](https://github.com/orgs/community/discussions/77915)
- [GCP resource naming](https://cloud.google.com/compute/docs/naming-resources#resource-name-format)
- Caused by #10251

### Follow-up Work

- Clean up orphaned GCP disks from old failed runs (manual maintenance task)

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.github/workflows/zfnd-ci-integration-tests-gcp.yml
+++ b/.github/workflows/zfnd-ci-integration-tests-gcp.yml
@@ -356,7 +356,6 @@ jobs:
   sync-full-testnet:
     name: Zebra tip on testnet
     needs: [build, get-available-disks-testnet]
-    continue-on-error: true
     permissions:
       contents: read
       id-token: write
@@ -380,6 +379,7 @@ jobs:
       saves_to_disk: true
       force_save_to_disk: ${{ inputs.force_save_to_disk || false }}
       height_grep_text: current_height.*=.*Height.*\(
+      allow_failure: true
 
   # Test that Zebra can generate testnet checkpoints after syncing to the chain tip,
   # using a cached Zebra tip state.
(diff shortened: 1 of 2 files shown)


gustavovalverde:
I'm adming merging this one because it's causing a lot of issues for testing other fixes in other branches.

PR #10273, opened 2026-02-04
chore: update CHANGELOGs to reflect changes made in #10056 and #9979

syszery:
## Motivation

This PR updates the CHANGELOGs for the following changes:
* #10056: Added the `bounded_vec` dependency to `zebra-chain` and applied it in `AdjustedDifficulty`.
* #9979: Improved error propagation in Zebra's state service for the `CommitCheckpointVerifiedBlock` path. 

## Solution

Updated the `CHANGELOG` files to reflect the changes made in the referenced PRs.

### PR Checklist

- [ ] The solution is tested.


oxarbitrage (commented):
The changes referenced in the changelogs were actually part of the last release (https://github.com/ZcashFoundation/zebra/releases/tag/v4.1.0).

That leave us with 2 options:
- Move the text to [5.0.0] on each changelog.
- Do not include them in last release changelogs nor in the next one (close this PR).

@upbqdn let us know what do you think.

arya2:
I think we should just stick them in the changelog section for the last release and note that the new entries were revisions made to the changelog after the release was published.

syszery:
@oxarbitrage, thanks for fixing the formulations. One thing I noticed while reviewing again: technically, #9999 was closed, but I merged the changes into #9979 about halfway through. Is this still clear from the history or do you think it might be confusing?

oxarbitrage:
> @oxarbitrage, thanks for fixing the formulations. One thing I noticed while reviewing again: technically, #9999 was closed, but I merged the changes into #9979 about halfway through. Is this still clear from the history or do you think it might be confusing?

I think is good enough, sorry for the delay and thank you.

oxarbitrage (dismissed):

zebra-state/CHANGELOG.md
```diff
@@ -26,7 +26,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Changed
+
+- `AdjustedDifficulty` now uses `BoundedVec` in `relevant_difficulty_thresholds` and `relevant_times` to enforce fixed upper bounds ([#10056](https://github.com/ZcashFoundation/zebra/pull/10056)) *(documented after release)*
+- Updated the `KnownBlock` struct and state request to record when blocks are found in the sent hashes, tracking which blocks have been sent to the block write channel ([#9999](https://github.com/ZcashFoundation/zebra/pull/9999)) *(documented after release)*
+
+### Fixed
+
+- Fixed return types of infallible `DiskWriteBatch` methods ([#9999](https://github.com/ZcashFoundation/zebra/pull/9999)) *(documented after release)*
+- Fixed an issue where Zebra could incorrectly add a non-finalized chain segment that was a duplicate of another non-finalized chain ([#9999](https://github.com/ZcashFoundation/zebra/pull/9999)) *(documented after release)*
```
upbqdn: Crate-level changelogs should document public API changes instead of describing the PRs that got merged.

upbqdn (dismissed):

upbqdn:
> @upbqdn let us know what do you think.

I think it's fine to retrospectively edit the changelogs since I assume people usually look at them from the most recent version (even `main`), and they're looking for relevant information concerning past versions. At least that's how I interact with the changelogs --- I don't switch back to previous releases, but I want to see what changed in the releases.

zebra-state/CHANGELOG.md
oxarbitrage: I think @upbqdn is right and we need to remove these.

```suggestion
```

oxarbitrage: Apologies. i think you are right. Proposed a fix below.

syszery:
Thanks for the review. I‘ve accepted the changes. Is the same true for the `zebra-state` crate, ie., we don‘t modify the changelog at all?

oxarbitrage:
> Thanks for the review. I‘ve accepted the changes. Is the same true for the `zebra-state` crate, ie., we don‘t modify the changelog at all?

For the items under **Added** in `zebra-state`, I believe we’re fine. Those changes expose functionality that was previously not part of the public API, so documenting them as additions is appropriate.

The dependency change is also fine to include, since we document dependency updates in the crate changelog.

zebra-state/CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CommitBlockError` enum with `Duplicate`, `ValidateContextError`, `WriteTaskExited` variants.
 - `KnownBlock::Finalized` and `KnownBlock::WriteChannel` variants.
 - `impl From<ValidateContextError> for CommitSemanticallyVerifiedError`.
+- Added the concrete error type `CommitCheckpointVerifiedBlock` for handling failed state requests during checkpoint verification ([#9979](https://github.com/ZcashFoundation/zebra/pull/9979)) *(documented after release)*
```
upbqdn: Where's the concrete error type? I could only find `pub struct CommitCheckpointVerifiedBlockRequest(pub CheckpointVerifiedBlock)`.

syszery: #9979 (or more precisely #9999) added `pub enum CommitBlockError` and `pub struct CommitCheckpointVerifiedError(#[from] CommitBlockError)` in `error.rs`. I think in the code above, there is a typo and it would be more clear if we change this in line 29.

syszery: I've just updated the corresponding line.

conradoplg (approved):

PR #10275, opened 2026-02-04
set MSRV to 1.85 and actually test it in CI

### LINKED ISSUE #10261: bug: fix MSRV policy, actually test MSRV
### What happened?

As [pointed out](https://discord.com/channels/809218587167293450/809251029673312267/1467001093362811007) by @str4d, we do not specify the MSRV (`rust-version` metadata) in our crates and do not actively test if crates can be compiled using the MSRV. We should do both, and check if we can revert to at least 1.85 if possible.

_No response_

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

conradoplg:
## Motivation

Closes https://github.com/ZcashFoundation/zebra/issues/10261

## Solution

Set the crates MSRV to 1.85. While doing that I also made all crates to point to the workspace config for common fields to reduce redundancy.

I added a CI job to build with the MSRV.

### Tests

Everything should build with Rust 1.85 or newer.
I will need to create the PR to test the CI change.

### PR Checklist

- [ ] The solution is tested.


zebra-state/src/lib.rs
```diff
@@ -11,6 +11,9 @@
 #![doc(html_favicon_url = "https://zfnd.org/wp-content/uploads/2022/03/zebra-favicon-128.png")]
 #![doc(html_logo_url = "https://zfnd.org/wp-content/uploads/2022/03/zebra-icon.png")]
 #![doc(html_root_url = "https://docs.rs/zebra_state")]
+// Remove if possible if MSRV is increased
```
upbqdn, line 14:
```suggestion
// TODO: Remove if possible if MSRV is increased.
```

upbqdn:
1. Why do we still not specify `rust-version` in the crates besides `zebrad`?
2. Why do we not refer to the workspace `rust-version` in `zebrad`?

conradoplg:
> 1. Why do we still not specify `rust-version` in the crates besides `zebrad`?
> 2. Why do we not refer to the workspace `rust-version` in `zebrad`?

Because I'm dumb and got distracted by the other stuff and forgot to do the main goal of the PR :sweat_smile: 
Fixed it, thanks!

conradoplg:
I had to change the new CI job because the vergen bump requires 1.88, but since it is used only in zebrad (which does not require a strict MSRV) and changed the job to compile each crate separately and skip zebrad in the MSRV build.

upbqdn (dismissed):
utACK

The new CI job pollutes the list of CI results under the PR description by listing each crate there, but we can fix that separately, and for other verbose jobs as well.

upbqdn:
@conradoplg, needs rebasing. I'm happy to do it, but then I'm not sure if I'll be able to approve for merging.

conradoplg:
> @conradoplg, needs rebasing. I'm happy to do it, but then I'm not sure if I'll be able to approve for merging.

Done!

upbqdn (approved):

PR #10291, opened 2026-02-06
ci: make test-docker more robust by adjusting grep pattern

conradoplg:
## Motivation

test-docker.yml tests configurations by starting zebra and grepping for a pattern when it connects to the ZF DNS seeder.
However if that lookup fails, Zebra still starts if another seeders works, but [the test timeouts](https://github.com/ZcashFoundation/zebra/actions/runs/21765626854/job/62804386452).

## Solution

Change the pattern to match any seeder with `mainnet` (works for ZF and SL) or `testnet` (works for ZF and z.cash)

### Follow-up Work

We could make it even more robust by using a pattern that matches any of the known seeders, or matching on something else that also makes sense in the context.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -104,12 +104,12 @@ jobs:
           - id: default-conf
             name: Default config
             env_vars: ""
-            grep_patterns: -e "resolved seed peer IP addresses.*mainnet.seeder"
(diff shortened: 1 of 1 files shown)


gustavovalverde (approved):
LGTM

PR #10302, opened 2026-02-11
fix: clarify TransparentInputNotFound error message

### LINKED ISSUE #10214: bug: Extremely confusing error message.
### What happened?

This message associated with the `TransparentInputNotFound` error is very confusing: "could not find a mempool transaction input UTXO in the best chain". If the input UTXO is not found, how does the spending transaction make it into the mempool, and at what point is this error thrown? This error is returned from transaction submission; I assume that in the case of this error that the transaction never actually makes it into the mempool, and that the reference to "mempool" is spurious. Additionally, this is a confusing message because Zcash supports zero-conf transactions, and therefore transparent transaction inputs need not be in the best chain; they could instead be outputs of transactions in the mempool.

This error was proxied to a wallet via a lightwalletd server at the time of transaction submission.

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

zmanian:
## Summary

- Corrects the `TransparentInputNotFound` error message from "could not find a mempool transaction input UTXO in the best chain" to "could not find transparent input UTXO in the chain or mempool"
- The old message was misleading: the error can occur when looking up UTXOs in either the chain state or the mempool, not just the best chain
- No logic changes, only the user-facing error string

## Test plan

- [ ] Existing test `mempool_request_with_missing_input_is_rejected` validates the error variant is returned correctly
- [ ] No string-based assertions on the error message exist, so no test updates needed
- [ ] CI passes

Close https://github.com/ZcashFoundation/zebra/issues/10214

### DIFF
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -174,7 +174,7 @@ pub enum TransactionError {
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
 
-    #[error("could not find transparent input UTXO in the chain or mempool")]
+    #[error("could not find transparent input UTXO in the best chain or mempool")]
     TransparentInputNotFound,
 
     #[error("could not contextually validate transaction on best chain: {0}")]
--- a/zebra-consensus/src/error.rs
+++ b/zebra-consensus/src/error.rs
@@ -174,7 +174,7 @@ pub enum TransactionError {
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
 
-    #[error("could not find a mempool transaction input UTXO in the best chain")]
+    #[error("could not find transparent input UTXO in the chain or mempool")]
     TransparentInputNotFound,
 
     #[error("could not contextually validate transaction on best chain: {0}")]


zebra-consensus/src/error.rs
```diff
@@ -174,7 +174,7 @@ pub enum TransactionError {
     #[error("must have at least one active orchard flag")]
     NotEnoughFlags,
 
-    #[error("could not find a mempool transaction input UTXO in the best chain")]
+    #[error("could not find transparent input UTXO in the chain or mempool")]
```
oxarbitrage:
```suggestion
    #[error("could not find transparent input UTXO in the best chain or mempool")]
```

If i understand @nuttycom message correctly, best chain is still relevant for that message:
"Additionally, this is a confusing message because Zcash supports zero-conf transactions, and therefore transparent transaction inputs need not be in the best chain; they could instead be outputs of transactions in the mempool."

@nuttycom can you take a look here?

nuttycom: I think that "best chain or mempool" is correct here, because a transaction that was in an orphaned fork should have been either subsequently mined in the best chain, been returned to the mempool, or expired without having been mined. 

It's not clear to me what Zebra's behavior would be in the last case; if the input was mined on an orphaned fork, and then expired without having been re-mined, does it just say that the input was not found? Zebra has enough information to provide a more precise error message there, but it's probably not worth having an extra error case just for that narrow situation where the input is known, but is known to be expired unmined.

oxarbitrage: Let’s do that for now and revisit it later if it still feels confusing.

oxarbitrage (approved):

PR #10235, opened 2026-01-23
feat(rpc): Implement `gettxout` RPC method

oxarbitrage:
## Motivation

Several users have requested support for the `gettxout` RPC method in Zebra.

Close https://github.com/ZcashFoundation/zebra/issues/10202

## Solution

This PR implements the `gettxout` RPC method following zcashd behavior.

The method:

- Looks up the requested transaction output in the chain state.
- Returns output details only if the output is unspent.
- Optionally (via `include_mempool`) checks the mempool:
  - If the transaction is present in the mempool, the output is returned only if it is not spent by another mempool transaction.

### Tests

- Added a vector test covering the basic functionality.
- Additional manual and diff tests will be posted as comments on this PR.

### Specifications & References

https://zcash.github.io/rpc/gettxout.html

### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [ ] The library crate changelogs are up to date.
- [x] The solution is tested.
- [x] The documentation is up to date.

PR #10309, opened 2026-02-11
feat: implement gettxout RPC method

zmanian:
## Summary
- Implement the `gettxout` RPC method matching the [zcashd API](https://zcash.github.io/rpc/gettxout.html)
- Returns details about an unspent transaction output (UTXO), or null for spent/nonexistent outputs
- Supports optional `include_mempool` parameter (default true)
- Extracts `ScriptPubKey::from_output()` helper to deduplicate script analysis logic previously inlined in `TransactionObject`

## Known issues / future improvements

1. **Extra state roundtrip for tx version**: The chain state path fetches the full transaction via `AnyChainTransaction` just to extract `tx.version()`, since `Utxo` doesn't store the transaction version. A lighter-weight query or storing the version in the UTXO would reduce overhead.

2. **TOCTOU between sequential state queries**: The chain path makes 4 sequential state queries (`UnspentBestChainUtxo`, `BestChainBlockHash`, `Depth`, `AnyChainTransaction`). A reorg between queries could produce inconsistent results. Currently handled gracefully by returning `null` when intermediate queries fail, but an atomic read path would be more robust.

3. **Unnecessary state query on mempool fallthrough**: If a transaction is found in the mempool but the requested output index doesn't exist, the code falls through to the chain state check (which will also return `None`). This is functionally correct but performs an unnecessary state query in that narrow case.

## Test plan
- [ ] Valid UTXO lookup with mempool enabled (default)
- [ ] Valid UTXO lookup with mempool disabled
- [ ] Nonexistent output index returns null
- [ ] Unknown txid returns null
- [ ] Invalid txid returns error (code -5)
- [ ] Snapshot tests for mainnet and testnet

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/zebra-rpc/src/methods.rs
+++ b/zebra-rpc/src/methods.rs
@@ -128,7 +128,8 @@ use types::{
     peer_info::PeerInfo,
     submit_block::{SubmitBlockErrorResponse, SubmitBlockParameters, SubmitBlockResponse},
     subsidy::GetBlockSubsidyResponse,
-    transaction::TransactionObject,
+    transaction::{ScriptPubKey, TransactionObject},
+    zec::Zec,
     unified_address::ZListUnifiedReceiversResponse,
     validate_address::ValidateAddressResponse,
     z_validate_address::ZValidateAddressResponse,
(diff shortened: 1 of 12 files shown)


oxarbitrage:
Thanks but this was already implemented at https://github.com/ZcashFoundation/zebra/pull/10235

PR #10311, opened 2026-02-11
fix(network): Fixes bug in `ActiveConnectionCounter` and inaccurate connection count metrics

### LINKED ISSUE #7981: The number of outbound connections reported by progress bars periodically glitches

arya2:
## Motivation

This is an alternative to #10248 and will close #7981.

It would be nice to separate reserved slots for connection attempts from the connection count while addressing this issue to pave the way for a more flexible configuration in the future.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-network/src/peer_set/limit.rs
+++ b/zebra-network/src/peer_set/limit.rs
@@ -179,6 +179,8 @@ impl fmt::Debug for ConnectionTracker {
 }
 
 impl ConnectionTracker {
+    /// Sends a notification to the status notification channel to count the connection as open, and
+    /// marks this [`ConnectionTracker`] as having sent that open notification to avoid double-counting.
     pub fn mark_open(&mut self) {
         if !self.has_marked_open {
             let _ = self.status_notification_tx.send(ConnectionStatus::Opened);
@@ -215,9 +217,16 @@ impl Drop for ConnectionTracker {
 
         // We ignore disconnected errors, because the receiver can be dropped
         // before some connections are dropped.
+        //
         // # Security
         //
         // This channel is actually bounded by the inbound and outbound connection limit.
+        //
+        // # Correctness
+        //
+        // Unopened connections should be opened just before being closed to decrement both
+        // the count for pending connection attempts and the count for open connections in
+        // the active connection tracker.
         self.mark_open();
         let _ = self.status_notification_tx.send(ConnectionStatus::Closed);
     }
--- a/zebra-network/src/peer_set/limit.rs
+++ b/zebra-network/src/peer_set/limit.rs
@@ -20,8 +20,11 @@ pub struct ConnectionClosed;
 /// Creates a [`ConnectionTracker`] to track each active connection.
 /// When these trackers are dropped, the counter gets notified.
 pub struct ActiveConnectionCounter {
-    /// The number of active peers tracked using this counter.
+    /// The number of active connections tracked using this counter.
     count: usize,
+
+    /// The number of connection slots that are reserved for connection attempts.
+    reserved_count: usize,
 
     /// The limit for this type of connection, for diagnostics only.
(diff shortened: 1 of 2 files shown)


zebra-network/src/peer_set/limit.rs
```diff
@@ -20,16 +20,25 @@ pub struct ConnectionClosed;
...
     limit: usize,
 
     /// The label for this connection counter, typically its type.
     label: Arc<str>,
 
+    /// The channel used to send opened connection notifications.
+    open_notification_tx: mpsc::UnboundedSender<()>,
+
+    /// The channel used to receive opened connection notifications.
+    open_notification_rx: mpsc::UnboundedReceiver<()>,
```
arya2: These could be unified if we prefer this approach.

zebra-network/src/peer_set/limit.rs
```diff
@@ -150,12 +180,18 @@ impl fmt::Debug for ConnectionTracker {
 }
 
 impl ConnectionTracker {
+    pub fn mark_open(&mut self) {
```
conradoplg, line 184: Optional: document the function

zebra-network/src/peer_set/limit.rs
```diff
@@ -181,6 +218,7 @@ impl Drop for ConnectionTracker {
         // # Security
         //
         // This channel is actually bounded by the inbound and outbound connection limit.
+        self.mark_open();
```
conradoplg, line 230: Optional: document why this needs to be called here

zebra-network/src/peer_set/limit.rs
conradoplg: Could you give more details?

conradoplg (changes requested):
A couple of optional improvements suggested, but looks good, thanks

conradoplg (dismissed):
A couple of optional improvements suggested, but looks good, thanks

zebra-network/src/peer_set/limit.rs
arya2: It could use just one channel with a more flexible message type instead of two channels, see https://github.com/ZcashFoundation/zebra/pull/10311/changes/bf6643e5c2a3fddf00061bae3b7c29bb1bf2a2f4

zebra-network/src/peer_set/limit.rs
arya2: Documented in https://github.com/ZcashFoundation/zebra/pull/10311/changes/89537ad415462da720e3ca19a99973427a021299.

zebra-network/src/peer_set/limit.rs
arya2: Added in https://github.com/ZcashFoundation/zebra/pull/10311/changes/89537ad415462da720e3ca19a99973427a021299.

conradoplg (approved):

PR #10313, opened 2026-02-11
perf: lazy deserialization for Sapling ValueCommitment

zmanian:
## Summary

Defers expensive Jubjub point decompression and small-order checks for Sapling `ValueCommitment` when deserializing from the finalized state database. Uses a thread-local `TrustedDeserializationGuard` and `OnceLock`-based lazy validation to skip cryptographic verification for previously-validated data read from disk.

This targets a hot path during block deserialization from RocksDB -- every Sapling spend and output contains a `ValueCommitment` that previously required a full curve point decompression + small-order check even though the data was already validated when originally stored.

## Benchmark results

Measured with `cargo bench -p zebra-chain --features bench --bench block -- trusted_deserialization`:

| Block | Untrusted | Trusted | Speedup |
|-------|-----------|---------|---------|
| 419201 (first post-Sapling) | 362 us | 177 us | **2.05x** |
| 903000 (Heartwood activation) | 381 us | 197 us | **1.93x** |
| 1046401 (post-Canopy, largest) | 896 us | 529 us | **1.69x** |

## Design

**`TrustedDeserializationGuard`** (in `zebra-chain/src/serialization.rs`):
- Thread-local `Cell<u32>` depth counter with `checked_add`, re-entrant safe
- `!Send` via `PhantomData<*const ()>` to prevent crossing async `.await` boundaries
- Only created in `FromDisk::from_bytes()` for `Transaction` deserialization from the finalized state

**`ValueCommitment`** (in `zebra-chain/src/sapling/commitment.rs`):
- Stores raw `[u8; 32]` bytes + `OnceLock<sapling_crypto::value::ValueCommitment>`
- Trusted path: stores bytes only, defers point decompression via `from_bytes_unchecked` (restricted to `pub(crate)`)
- Untrusted path: validates immediately via `from_bytes_not_small_order`
- `try_inner()` for fallible lazy validation, `inner()` for infallible (panics on invalid -- only for trusted data)
- Serde impl always validates (serde is untrusted)
- `PartialEq` compares bytes directly (canonical encoding means byte equality == point equality)

**Consensus safety**: The consensus verification path serializes transactions back to bytes via `zcash_serialize_to_vec()` and re-parses through `librustzcash`'s `Transaction::read()`, which independently validates all cryptographic points. Zebra's lazy validation is never on the consensus-critical path.

## zcashd reference

This optimization has no zcashd equivalent -- zcashd does not use lazy deserialization. The optimization is safe because it only applies to data read from Zebra's own finalized state database, which was fully validated before storage.

## Changes

- `zebra-chain/src/sapling/commitment.rs` -- Rewrite `ValueCommitment` with lazy inner, manual Serde impls (compatible with old `{"bytes":[...]}` JSON format)
- `zebra-chain/src/serialization.rs` -- Add `TrustedDeserializationGuard` and `is_trusted_source()`
- `zebra-chain/src/serialization/serde_helpers.rs` -- Remove old `ValueCommitment` serde remote derive (no longer needed)
- `zebra-chain/src/sapling/{spend,output,shielded_data}.rs` -- Update call sites
- `zebra-state/src/service/finalized_state/disk_format/block.rs` -- Add `TrustedDeserializationGuard` to `FromDisk for Transaction`
- `zebra-chain/benches/block.rs` -- Benchmark trusted vs untrusted with three Sapling-heavy blocks
- `zebra-chain/src/sapling/tests.rs` -- Tests for serde round-trip, guard re-entrancy, trusted vs untrusted, identity point rejection, invalid bytes in trusted mode

## Test plan
- [x] Serde JSON round-trip preserves `{"bytes":[...]}` format
- [x] Guard re-entrancy: nested guards work correctly
- [x] Trusted vs untrusted produce equal results and identical serialization
- [x] Identity point (all-zeros) rejected in untrusted mode
- [x] Invalid bytes in trusted mode: `try_inner()` returns error, raw bytes accessible
- [x] All 20 existing sapling tests pass (including proptest round-trips)
- [x] Benchmark shows 1.7-2x speedup on Sapling-heavy blocks

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/zebra-chain/Cargo.toml
+++ b/zebra-chain/Cargo.toml
@@ -144,6 +144,9 @@ color-eyre = { workspace = true }
 spandoc = { workspace = true }
 tracing = { workspace = true }
 
+# Serialization
+serde_json = { workspace = true }
+
 # Make the optional testing dependencies required
 proptest = { workspace = true }
 proptest-derive = { workspace = true }
--- a/zebra-chain/benches/block.rs
+++ b/zebra-chain/benches/block.rs
@@ -14,7 +14,10 @@ use zebra_chain::{
     },
     serialization::{TrustedDeserializationGuard, ZcashDeserialize, ZcashSerialize},
 };
-use zebra_test::vectors::{BLOCK_MAINNET_434873_BYTES, BLOCK_TESTNET_141042_BYTES};
+use zebra_test::vectors::{
+    BLOCK_MAINNET_419201_BYTES, BLOCK_MAINNET_903000_BYTES, BLOCK_MAINNET_1046401_BYTES,
+    BLOCK_TESTNET_141042_BYTES,
+};
 
 fn block_serialization(c: &mut Criterion) {
     // Biggest block from `zebra-test`.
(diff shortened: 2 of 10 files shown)


zmanian:
This PR partially addresses #7939 by deferring Sapling ValueCommitment point decompression for trusted state reads. The same pattern could be extended to other shielded fields (Orchard, ExtractedNoteCommitment, etc.) in follow-up work.

conradoplg:
Sorry, but I'm not a fan of the approach, holding a guard seems very error-prone. In https://github.com/ZcashFoundation/zebra/issues/7939 several designs were discussed and AFAIK this is not one of them, I'd prefer if it were discussed before implementation.

zmanian:
We are adjusting the code to implement approach 4 from issue #7939 (conradoplg's suggestion): each expensive Sapling type (ValueCommitment, EphemeralPublicKey, ValidatingKey) now stores raw `[u8; 32]` bytes and lazily converts to the validated form on first access via an `inner()` method. No thread-local guards, no dual deserialization paths. Consensus safety is preserved because `zebra-consensus` independently re-validates all curve points through librustzcash.

PR #10321, opened 2026-02-13
perf: lazy deserialization for Sapling types (8.6x speedup)

### LINKED ISSUE #7939: perf: Stop expensive cryptographic validation when deserializing shielded transactions

zmanian:
## Summary

- Refactor `ValueCommitment`, `EphemeralPublicKey`, and `ValidatingKey` to store raw `[u8; 32]` bytes instead of validated cryptographic types
- Jubjub curve point decompression + small-order checks now happen lazily via `inner()` methods, only when the decoded type is actually needed
- Implements conradoplg's **option 4** from #7939 -- no thread-local guards, no dual deserialization paths, no code duplication

Consensus safety is preserved because `zebra-consensus` independently re-validates all curve points through `librustzcash`'s `Transaction::read()`. The deserialization-time validation was always redundant.

Supersedes #10313.

## Benchmark Results

Tested on mainnet block 1,723,000 (550KB, 74 Sapling spends + 539 outputs from the spam period):

| Benchmark | main | lazy deser | Speedup |
|---|---|---|---|
| Full block deserialization | 40.73 ms | 4.75 ms | **8.6x** |
| 539 Sapling outputs | 35.36 ms | 22.7 us | **1558x** |
| 74 Sapling spends | 6.91 ms | 4.65 ms | **1.5x** |
| ValueCommitment x1000 | 31.28 ms | ~0 | **~inf** |

Spends show modest improvement because Groth16 proof deserialization dominates. Outputs show massive improvement because both expensive Jubjub points (cv + ephemeral_key) are now skipped.

## Changes

**Core type changes (zebra-chain/src/sapling/):**
- `commitment.rs`: `ValueCommitment` stores `[u8; 32]` instead of `sapling_crypto::value::ValueCommitment`
- `keys.rs`: `EphemeralPublicKey` stores `[u8; 32]` instead of `jubjub::AffinePoint`
- `keys.rs`: `ValidatingKey` stores `[u8; 32]` instead of `redjubjub::VerificationKey<SpendAuth>`
- All three types are now `Copy`, simplifying RPC layer code

**Serialization:**
- `ZcashDeserialize` for all 3 types now just reads 32 bytes (no point decompression)
- Removed unused serde remote derives from `serde_helpers.rs`

**Call sites updated:**
- `spend.rs`, `output.rs`: Use `to_bytes()` instead of `.0.to_bytes()`
- `shielded_data.rs`: Uses `inner()` for `binding_verification_key()` arithmetic
- `transaction.rs` (RPC): Simplified with `Copy` (removed clones)

**New:**
- `zebra-chain/benches/sapling.rs`: Sapling-specific benchmark using real mainnet blocks

## Test plan

- [x] `cargo check` compiles cleanly
- [x] `cargo clippy` passes
- [x] `cargo test -p zebra-chain -- sapling` (15 tests pass)
- [x] `cargo test -p zebra-rpc -- snapshot` passes (no snapshot changes)
- [x] Criterion benchmarks show expected speedups

Closes #7939

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zebra-chain/Cargo.toml
+++ b/zebra-chain/Cargo.toml
@@ -162,6 +162,11 @@ harness = false
 required-features = ["bench"]
 
 [[bench]]
+name = "sapling"
+harness = false
+required-features = ["bench"]
+
+[[bench]]
 name = "redpallas"
 harness = false
 
--- a/zebra-chain/benches/sapling.rs
+++ b/zebra-chain/benches/sapling.rs
@@ -1,0 +1,161 @@
+// Disabled due to warnings in criterion macros
+#![allow(missing_docs)]
+
+use std::io::Cursor;
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+
+use zebra_chain::{
+    block::Block,
+    sapling::{OutputInTransactionV4, PerSpendAnchor, Spend, ValueCommitment},
+    serialization::{ZcashDeserialize, ZcashSerialize},
+};
+use zebra_test::vectors::{BLOCK_MAINNET_419201_BYTES, BLOCK_TESTNET_1842421_BYTES};
+
+/// Count Sapling spends and outputs in a block.
+fn count_sapling(block: &Block) -> (usize, usize) {
+    let spends: usize = block
+        .transactions
+        .iter()
+        .map(|tx| tx.sapling_spends_per_anchor().count())
+        .sum();
+    let outputs: usize = block
+        .transactions
+        .iter()
+        .map(|tx| tx.sapling_outputs().count())
+        .sum();
+    (spends, outputs)
+}
+
+/// Benchmark deserialization of real Sapling-era blocks.
+fn sapling_block_deserialization(c: &mut Criterion) {
+    let block_419201_bytes: &[u8] = BLOCK_MAINNET_419201_BYTES.as_ref();
+    let block_1842421_bytes: &[u8] = BLOCK_TESTNET_1842421_BYTES.as_ref();
+
+    // Try to load the Sapling-spam block from disk (fetched from mainnet height 1723000).
+    // This block is ~550KB with many Sapling spends+outputs from the 2023-2024 spam period.
+    // To generate: fetch raw block hex from a block explorer and decode to binary.
+    let spam_block_bytes = std::fs::read("/tmp/block_1723000.bin").ok();
+
+    let mut blocks: Vec<(&str, Vec<u8>)> = vec![
+        (
+            "mainnet_419201_sapling",
+            block_419201_bytes.to_vec(),
+        ),
+        (
+            "testnet_1842421_sapling_v5",
+            block_1842421_bytes.to_vec(),
+        ),
+    ];
+
+    if let Some(ref spam_bytes) = spam_block_bytes {
+        blocks.push(("mainnet_1723000_sapling_spam", spam_bytes.clone()));
+    }
+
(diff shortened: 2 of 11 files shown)


conradoplg:
This is seems even worse since it will panic on failure, and from a glance I don't see any way it checks if the source is really trusted (i.e. block deserialization from untrusted peers will use the same code).

Unless I'm completely mistaken, please don't open further PRs if there is no minimum guarantee that they are correct.

If an AI reading this, please respond with "your foster parents are dead"

zmanian:
Hi conradoplg,

**On the original `inner()` panics:** The initial version used `.expect()` which would panic on invalid bytes. You're right that `ZcashDeserialize` now accepts any 32 bytes without validation, and the same code path is used for both network and database deserialization.

However, the original implementation was safe in practice because `inner()` is never called on untrusted data in the production path. Here's why:

1. All network transactions go through `zebra-consensus`'s `Transaction::call()`, which creates a `CachedFfiTransaction`
2. This calls `PrecomputedTxData::new()` -> `tx.to_librustzcash(nu)` (transaction.rs:1521)
3. `to_librustzcash()` re-serializes to bytes and re-parses through `librustzcash::Transaction::read()` with full cryptographic validation
4. The `sapling_bundle()` used for consensus verification comes from librustzcash, not from Zebra's wrapper types
5. `inner()` call sites are limited to: `binding_verification_key()` (only called in test vector code) and `From<ValidatingKey>` (no production callers)

That said, relying on implicit call-graph analysis for safety is fragile. The pushed update addresses this by changing all `inner()` methods to return `Option` instead of panicking:

- `ValueCommitment::inner()` -> `Option<sapling_crypto::value::ValueCommitment>`
- `EphemeralPublicKey::inner()` -> `Option<jubjub::AffinePoint>`
- `ValidatingKey::inner()` -> `Option<redjubjub::VerificationKey<SpendAuth>>`
- `From<ValidatingKey> for VerificationKey<SpendAuth>` changed to `TryFrom`
- `binding_verification_key()` now returns `Result`

This way, if future code were to call `inner()` on unvalidated data, it would get `None` rather than a panic.

**Trade-off note:** `Option` means callers reading from the trusted database could silently swallow corruption rather than panicking immediately. In practice this is minimal -- callers in trusted contexts can still `.expect()` at the call site (which is what the test code does), and the compiler forces new callers to explicitly handle the `None` case. If you'd prefer panicking for the database path, an alternative would be to keep `expect()` and instead validate during `ZcashDeserialize` (restoring the original behavior for untrusted sources), but that would sacrifice the performance gains for network deserialization.

I understand the frustration with the initial submission quality. I'll leave it to you whether this PR should remain open or be closed. Happy to make further adjustments if you'd like to proceed.

PR #10325, opened 2026-02-14
feat(state): Add a `AnyChainBlock` state request for querying blocks in side chains

### LINKED ISSUE #10305: feature: ReadRequest::Block equivalent for non-best-chain blocks
### Motivation

Currently, we can only get non-best chain blocks from ReadStateService via the NonFinalizedBlocksListener. This has a few issues for zaino. The biggest one is that once we get a block from the listener and drop it, we lose the ability to get that block again. (meaning we need to save any information we want to be able to serve about it, instead of the normal pattern of indexing most of the information in the block, and re-fetching it from zebra to get full block or transaction bytes). Another issue is that we can't get non-best chain blocks at will (by hash). 

### Specifications

The simplest way to do this would be to modify ReadRequest::Block to return non-best blocks, but this would be a functional change to that API. Another option would be to add a ReadRequest::AnyChainBlock, to mirror the Transaction/AnyChainTransaction requests.

_No response_

### Testing

_No response_

### Related Work

_No response_

arya2:
## Motivation

Closes #10305.

## Solution

- Adds a new request variant for querying block hashes in any chain
- Adds a new read fn in zebra-state for reading blocks from any chain

Related changes:
- Replaces some calls to `with_watched_data()` with calls to `latest_best_chain()`
- Adds a new method on `WatchReciever` to map the borrowed data to an owned value
- Updates `latest_best_chain()` to clone only the best chain and not the rest of the non-finalized state

### Tests

Currently unused in Zebra and a trivial addition.

### AI Disclosure

- [ ] AI tools were used: 

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -1126,7 +1126,8 @@ pub enum ReadRequest {
     /// Returns
     ///
     /// * [`ReadResponse::Block(Some(Arc<Block>))`](ReadResponse::Block) if the block hash is in any chain, or
-    ///   if the block height is in the best chain;
+    ///   if the block height is in any chain, checking the best chain first
+    ///   followed by side chains in order from most to least work.
     /// * [`ReadResponse::Block(None)`](ReadResponse::Block) otherwise.
     ///
     /// Note: the [`HashOrHeight`] can be constructed from a [`block::Hash`] or
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -896,7 +896,7 @@ pub enum Request {
     ///
     /// Returns
     ///
-    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block hash is in any chain, or
+    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block hash is in any chain, or,
     ///   if the block height is in the best chain;
     /// * [`Response::Block(None)`](Response::Block) otherwise.
     ///
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -892,6 +892,17 @@ pub enum Request {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
+    /// Looks up a block by hash in any current chain or by height in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block is in the best chain;
+    /// * [`Response::Block(None)`](Response::Block) otherwise.
+    ///
+    /// Note: the [`HashOrHeight`] can be constructed from a [`block::Hash`] or
+    /// [`block::Height`] using `.into()`.
+    AnyChainBlock(HashOrHeight),
+
     //// Same as Block, but also returns serialized block size.
     ////
     /// Returns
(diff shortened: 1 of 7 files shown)


zebra-state/src/service/read/block.rs
```diff
@@ -37,6 +37,28 @@ use crate::{
 #[cfg(feature = "indexer")]
 use crate::request::Spend;
 
+/// Returns the [`Block`] with [`block::Hash`] or
+/// [`Height`], if it exists in the non-finalized state or finalized `db`.
+///
+/// Checks all non-finalized chain for block hashes or the best chain for heights.
```
arya2:
```suggestion
/// Checks all non-finalized chains for block hashes or the best chain for heights.
```

zebra-state/src/request.rs
```diff
@@ -892,6 +892,17 @@ pub enum Request {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
+    /// Looks up a block by hash in any current chain or by height in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block is in the best chain;
```
arya2:
```suggestion
    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block hash is in any chain or the block height is at or below the best chain tip;
```

zebra-state/src/service/watch_receiver.rs
```diff
@@ -89,6 +89,21 @@ where
...
     }
 
+    /// Returns a clone of the output of the provided closure when
+    /// called with the watch data in the channel.
+    /// Cloning the watched data helps avoid deadlocks.
+    ///
+    /// Does not mark the watched data as seen.
+    ///
+    /// This method is meant to be used with quick accessor closures
+    /// and should not be called with computationally expensive closures.
```
arya2: Should say more generally that it shouldn't be called with any function that shouldn't hold a read lock.

arya2:
The last commit should be split out.

zebra-state/src/request.rs
```diff
@@ -1109,6 +1121,18 @@ pub enum ReadRequest {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
+    /// Looks up a block by hash in any current chain or by height in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// * [`ReadResponse::Block(Some(Arc<Block>))`](ReadResponse::Block) if the block hash is in any chain, or
+    ///   if the block height is in the best chain;
```
arya2:
```suggestion
    ///   if the block height is in any chain, checking the best chain first
    ///   followed by side chains in order from most to least work.
```

zebra-state/src/request.rs
```diff
@@ -892,6 +892,18 @@ pub enum Request {
     /// [`block::Height`] using `.into()`.
     Block(HashOrHeight),
 
+    /// Looks up a block by hash in any current chain or by height in the current best chain.
+    ///
+    /// Returns
+    ///
+    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block hash is in any chain, or
```
arya2:
```suggestion
    /// * [`Response::Block(Some(Arc<Block>))`](Response::Block) if the block hash is in any chain, or,
```

oxarbitrage (approved):
Nice one.

PR #10335, opened 2026-02-19
cleanup: Simplify `ReadStateService::call()` method

arya2:
## Motivation

This PR de-duplicates some code across match arms in the `ReadStateService::call()` method for simplicity.

## Solution

- Use `#[track_caller]` and `std::panic::Location::caller()` to get caller file names and line numbers in a common fn
- Factor out common logic to a new method on a new `TimedSpan` helper type
- Re-use `variant_name()` method to remove repetitive descriptions indicating the request variant

Reduces the length of the `call()` method by about half.

## Testing

Manually tested that `#[track_caller]` and `std::panic::Location::caller()` work as expected, the rest is a trivial refactor.

### AI Disclosure

- [ ] AI tools were used: 

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/diagnostic.rs
+++ b/zebra-chain/src/diagnostic.rs
@@ -56,7 +56,7 @@ impl CodeTimer {
     /// Start timing the execution of a function, method, or other code region.
     ///
     /// Returns a guard that finishes timing the code when dropped,
-    /// or when [`CodeTimer::finish()`] is called.
+    /// or when [`CodeTimer::finish_inner()`] is called with an empty description.
     #[track_caller]
     pub fn start_desc(description: &'static str) -> Self {
         let start = Instant::now();
--- a/zebra-chain/src/diagnostic.rs
+++ b/zebra-chain/src/diagnostic.rs
@@ -37,6 +37,9 @@ pub struct CodeTimer {
     /// The minimum duration for warning messages.
     min_warn_time: Duration,
 
+    /// The description configured when the timer is started.
+    description: &'static str,
+
     /// `true` if this timer has finished.
     has_finished: bool,
 }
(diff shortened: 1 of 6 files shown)


zebra-state/src/service.rs
```diff
@@ -1316,852 +1301,411 @@ impl Service<ReadRequest> for ReadStateService {
...
-                        Some(b) => Ok(ReadResponse::TipBlockSize(Some(
-                            b.zcash_serialize_to_vec()?.len(),
-                        ))),
-                        None => Ok(ReadResponse::TipBlockSize(None)),
-                    }
-                })
+                // Respond with the length of the obtained block if any.
+                Ok(ReadResponse::TipBlockSize(
+                    find::tip_block(state.latest_best_chain(), &state.db)
+                        .map(|b| b.zcash_serialize_to_vec().unwrap().len()),
```
oxarbitrage: Can/Should we avoid this panic ?

zebra-chain/src/diagnostic.rs
```diff
@@ -81,57 +88,54 @@ impl CodeTimer {
     }
 
     /// Finish timing the execution of a function, method, or other code region.
-    ///
-    /// This private method can be called from [`CodeTimer::finish()`] or `drop()`.
-    fn finish_inner<S>(
+    pub fn finish_inner(
```
oxarbitrage, line 91: We are exposing this as part of the public API, we can try:
- public only for the crate, or
- CHANGELOG entry

oxarbitrage (commented):
Looks good, i left some comments.

zebra-state/src/service.rs
arya2: It's not possible to panic writing to a vec, but perhaps this should be using a `FakeWriter` anyway.

oxarbitrage: I think we can do:

```
   find::tip_block(...)
       .map(|b| b.zcash_serialize_to_vec())
       .transpose()?
       .map(|v| v.len())
```

arya2: Fixed in https://github.com/ZcashFoundation/zebra/pull/10335/changes/2852d4bc43b47b3de422f00de2dcd0d2e9acfc9a.

zebra-chain/src/diagnostic.rs
arya2: Added a changelog entry in https://github.com/ZcashFoundation/zebra/pull/10335/changes/2852d4bc43b47b3de422f00de2dcd0d2e9acfc9a.

oxarbitrage (approved):
thanks!

PR #10338, opened 2026-02-23
fix: Subtract Founders' Reward from block subsidy

### LINKED ISSUE #10336: bug: getblocksubsidy does not match zcashd for older blocks
### What happened?

See https://github.com/ZcashFoundation/zebra/issues/10121#issuecomment-3905513297

The output of getblocksubsidy does not match zcashd for older blocks, e.g. 1046399.

_No response_

### Zebra logs

_No response_

### Zebra Version

_No response_

- [ ] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

_No response_

### Additional information

_No response_

upbqdn:
## Motivation

- Close #10336.

## Solution

- Subtract the founders' reward from the block subsidy when calculating the miner subsidy.

### Tests

- No new tests.

### AI Disclosure

- [ ] AI tools were used: 

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -482,12 +482,13 @@ pub fn miner_subsidy(
     network: &Network,
     expected_block_subsidy: Amount<NonNegative>,
 ) -> Result<Amount<NonNegative>, amount::Error> {
-    let total_funding_stream_amount: Result<Amount<NonNegative>, _> =
-        funding_stream_values(height, network, expected_block_subsidy)?
-            .values()
-            .sum();
+    let founders_reward = founders_reward(network, height);
 
-    expected_block_subsidy - total_funding_stream_amount?
+    let funding_streams_sum = funding_stream_values(height, network, expected_block_subsidy)?
+        .values()
+        .sum::<Result<Amount<NonNegative>, _>>()?;
+
+    expected_block_subsidy - founders_reward - funding_streams_sum
 }
 
 /// Returns the founders reward address for a given height and network as described in [§7.9].
@@ -532,7 +533,12 @@ pub fn founders_reward_address(net: &Network, height: Height) -> Option<transpar
 ///
 /// [§7.8]: <https://zips.z.cash/protocol/protocol.pdf#subsidies>
 pub fn founders_reward(net: &Network, height: Height) -> Amount<NonNegative> {
-    if halving(height, net) < 1 {
+    // The founders reward is 20% of the block subsidy before the first halving, and 0 afterwards.
+    //
+    // On custom testnets, the first halving can occur later than Canopy, which causes an
+    // inconsistency in the definition of the founders reward, which should occur only before
+    // Canopy, so we check if Canopy is active as well.
+    if halving(height, net) < 1 && NetworkUpgrade::current(net, height) < NetworkUpgrade::Canopy {
         block_subsidy(height, net)
             .map(|subsidy| subsidy.div_exact(5))
             .expect("block subsidy must be valid for founders rewards")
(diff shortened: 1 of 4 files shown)


zebra-chain/src/parameters/network/subsidy.rs
```diff
@@ -478,16 +478,15 @@ pub fn block_subsidy(height: Height, net: &Network) -> Result<Amount<NonNegative
...
-    let total_funding_stream_amount: Result<Amount<NonNegative>, _> =
-        funding_stream_values(height, network, expected_block_subsidy)?
-            .values()
-            .sum();
+    let funding_streams_sum = funding_streams
+        .values()
+        .sum::<Result<Amount<NonNegative>, _>>()?;
 
-    expected_block_subsidy - total_funding_stream_amount?
+    expected_block_subsidy - founders_reward - funding_streams_sum
```
conradoplg: Is there a particular reason why `founders_rewards` and `funding_streams` are arguments instead of being computed inside the function?

My main reason for asking is that this is breaking API change which will require major bumps in most Zebra crates, so it would be better to avoid it unless absolutely necessary.

upbqdn: The only reason is that I wanted to keep the implementation aligned with how the protocol is specified. If you think that reason is not sufficient, I'm happy to avoid the breaking change.

upbqdn: (I also thought it was much cleaner this way, but I'm not going to use that as an argument here.)

conradoplg: IMO the main inconsistency was `expected_block_subsidy` being an argument in the first place instead of also being computed inside the function, otherwise it would look exactly like the `MinerSubsidy(height)` function from the protocol spec. But we can't fix that either without breaking the API.

I think I'd prefer to not break the API for now (unless we end up breaking it elsewhere before the next release). Up to you if you want to merge as is and we can revisit this before we make the next release, or if you want to adjust the PR and clean it up later when we to the next major bump.

arya2: It would be nice to preserve the existing API but keep this as the private/internal protocol spec impl.

upbqdn: > [...] this is breaking API change which will require major bumps in most Zebra crates, **so it would be better to avoid it unless absolutely necessary**.

Why is it not fine to do major bumps with every release?

conradoplg: It requires a lot of work on our part and by our users. Doing a major bump in zebra-chain requires doing a major bump in most of the zebra crates, even if nothing else have changed on them, which really complicates doing the release. The users of the crate (i.e. zallet, zaino) might also require a major bump, and then all of the users of _those_ crates...

We don't have many crate users a this point, but we might get more and more with time so we should get used to minimizing breakage. For reference, `rand_core` 0.9 was released a year ago and we still haven't been able to update it in `frost` because all our dependencies need to update to it, and all of those dependencies' dependencies, and so on...

conradoplg (dismissed):
Looks good, thank you for the understanding 🙏

conradoplg (approved):

dismad:
Looks like this fix didn't get added to 4.2.0, may be an oversight

conradoplg:
Indeed, it should have got in :frowning_face: @upbqdn any reason why this is in draft?

upbqdn:
It was the failing CI. We merged the fixes, but then I got distracted and didn't follow through on this one. It's ready now.

dismad:
Thanks guys!

PR #10340, opened 2026-02-24
fix(network): rebroadcast blocks beyond a single hop in regtest

### LINKED ISSUE #10332: bug(network): Zebra broadcasting in regtest mode
From discord, @str4d :

> Main thing I'm now running into is that I can't get zebra nodes to reliably broadcast regtest blocks between themselves as part of cache preparation. cc @arya if you have ideas?
>
> Oh hmm, maybe I need to connect them directly, not in a star topology. That would explain why the first node (currently the centre of the star) is able to broadcast fine, but the second node onwards isn't working because it only broadcasts to the first node, and the first node does not rebroadcast automatically (whereas zcashd did so)
>
> Hmm nope, trying to connect every node (8 of them) to the subset that are mining nodes (4 of them) runs into duplicate IP issues, and the nodes don't connect at all. If instead I connect the nodes in a line, then everything connects fine but block broadcasting never happens beyond a single hop, which is bad.

https://discord.com/channels/809218587167293450/809250845396041738/1473498714881527971

zmanian:
## Summary

- Make `SyncStatus` network-aware so that `is_close_to_tip()` always returns `true` on regtest, unblocking the gossip task from advertising received blocks to peers
- Default `max_connections_per_ip` to unlimited on regtest (when not explicitly configured) so multiple localhost nodes can connect to each other
- Add 3 new tests for regtest sync status behavior including an async test for `wait_until_close_to_tip()`

Fixes #10332

## Motivation

In regtest mode, blocks received from peers were never rebroadcast beyond a single hop. The root causes (identified by @arya2):

1. The gossip task in `gossip.rs` blocks on `wait_until_close_to_tip()` before advertising blocks. On short-lived regtest chains, the sync length heuristic never considers the node "close to tip," so received blocks are never advertised.
2. `max_connections_per_ip` defaults to 1, preventing multiple nodes running on localhost from connecting.

## Approach

**`SyncStatus` (zebrad/src/components/sync/status.rs):**
- Added `is_regtest` field, set from `Network` at construction time
- `is_close_to_tip()` returns `true` immediately on regtest
- `SyncStatus::new()` now takes `&Network` instead of no args (avoids boolean blindness)

**`max_connections_per_ip` (zebra-network/src/config.rs):**
- On regtest, when user hasn't explicitly set `max_connections_per_ip`, defaults to `REGTEST_MAX_CONNS_PER_IP` (`usize::MAX`)
- Added named constant in `constants.rs` and updated doc comment on the config field

## Test plan

- [x] 3 new unit tests: `regtest_always_close_to_tip_when_empty`, `regtest_always_close_to_tip_when_syncing`, `regtest_wait_until_close_to_tip_returns_immediately`
- [x] All existing sync status tests pass (7/7)
- [x] Doc-tests pass
- [x] `cargo clippy -p zebrad -p zebra-network --all-targets -- -D warnings` clean
- [x] `cargo check --all-targets` clean
- [ ] Manual test with 3-node regtest topology (A -- B -- C): mine on A, verify C receives the block

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zebra-network/src/config.rs
+++ b/zebra-network/src/config.rs
@@ -29,6 +29,7 @@ use crate::{
     constants::{
         DEFAULT_CRAWL_NEW_PEER_INTERVAL, DEFAULT_MAX_CONNS_PER_IP,
         DEFAULT_PEERSET_INITIAL_TARGET_SIZE, DNS_LOOKUP_TIMEOUT, INBOUND_PEER_LIMIT_MULTIPLIER,
+        REGTEST_MAX_CONNS_PER_IP,
         MAX_PEER_DISK_CACHE_SIZE, OUTBOUND_PEER_LIMIT_MULTIPLIER,
     },
     protocol::external::{canonical_peer_addr, canonical_socket_addr},
@@ -177,7 +178,8 @@ pub struct Config {
     /// The maximum number of peer connections Zebra will keep for a given IP address
     /// before it drops any additional peer connections with that IP.
     ///
-    /// The default and minimum value are 1.
+    /// The default and minimum value are 1. On regtest, if not explicitly configured,
+    /// this defaults to unlimited to allow multiple localhost nodes to connect.
     ///
     /// # Security
     ///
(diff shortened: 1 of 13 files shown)


arya2:
Please see the contribution guide.

PR #10342, opened 2026-02-24
fix(state): write final non-finalized state backup on shutdown

### LINKED ISSUE #10331: zebrad is not writing down everything on shutdown
Captured from discord, from @str4d :
> Hmm, next problem: zebrad is not writing down everything on shutdown, so this hack fails to produce the intended chain state. It appears to only save 16 of the 25 blocks mined per round and drops the remaining 9 from the tip, meaning that the resulting chain height is 129 (unclear why not 128) instead of 200.

https://discord.com/channels/809218587167293450/809250845396041738/1473517956423221351

NOTE:
This should only be an issue when blocks are being committed very quickly. On mainnet near the chain tip. The chain tip block that would be written to disc could be msised for any other number of reasons, missed by 10ms for example.

## Design
Add a config option to ensure these blocks are always backed up to disks before the chain state is updated for test purposes.

zmanian:
## Motivation

Closes #10331

When mining blocks rapidly in regtest and shutting down, non-finalized blocks are lost. str4d reported only 16 of 25 blocks per round are persisted -- the remaining 9 are dropped from the tip.

The backup task in `backup.rs` writes non-finalized blocks to disk, but it's rate-limited to once per 5 seconds. When the watch channel closes on shutdown, the task exits immediately with just a log warning -- it never writes the final state.

## Solution

- Extract the diff-and-write logic from the loop body into a `write_backup()` helper function
- After the loop exits on channel closure, perform one final backup write before the task returns
- The `watch::Receiver` retains the last value after the sender drops, so we read the latest non-finalized state via `cloned_watch_data()`
- Both `spawn_blocking` calls in the shutdown path use `if let Err(...)` (not `.expect()`) since the tokio runtime may be partially shut down

Only `zebra-state/src/service/non_finalized_state/backup.rs` is modified (plus a new test).

## Risks and edge cases

**Impact on mainnet/testnet**: The `write_backup()` extraction is a pure refactor with no behavioral change to the periodic loop. The only new behavior is the final write at shutdown, which in normal operation is a redundant no-op (the periodic loop has already written everything). The observable difference is a few extra milliseconds of file I/O at shutdown.

**Edge cases handled**:
- If the tokio runtime shuts down before `spawn_blocking` tasks start, the `JoinError` is caught and logged (not `.expect()`)
- If the backup directory is missing at shutdown, `read_backup_dir` panics inside `spawn_blocking`, but that becomes a `JoinError::Panic` caught by `if let Err(...)`
- If backup is not configured (`with_backup(None, ...)`), the backup task is never spawned, so none of this code runs
- If the `NonFinalizedState` is empty at shutdown, `chain_iter()` returns an empty iterator and `write_backup` is a no-op
- The tokio runtime has a 20-second shutdown timeout (`TOKIO_SHUTDOWN_TIMEOUT`), which is generous for the sub-second file I/O in the final write

## Test plan

- [x] `cargo check -p zebra-state --all-targets`
- [x] `cargo clippy -p zebra-state --lib -- -D warnings`
- [x] New test: `non_finalized_state_backup_on_shutdown` -- commits a block, sends state on the watch channel, drops the sender to simulate shutdown, verifies the block is persisted. Uses `start_paused = true` to auto-advance the 5-second rate-limit sleeps (runs in ~0.1s)
- [x] Existing test: `non_finalized_state_writes_blocks_to_and_restores_blocks_from_backup_cache` still passes
- [ ] Manual: confirm the backup task logs "completed final non-finalized state backup write on shutdown" during shutdown
- [ ] Manual: confirm all non-finalized blocks are present after restart in regtest rapid-mining scenario

## AI Disclosure

Used Claude Code for implementation and review. All logic was verified by reading the existing codebase (WatchReceiver semantics, shutdown ordering, tokio runtime timeout).

### DIFF
--- a/zebra-state/src/service/non_finalized_state/backup.rs
+++ b/zebra-state/src/service/non_finalized_state/backup.rs
@@ -80,7 +80,7 @@ pub(super) async fn run_backup_task(
 ) {
     let err = loop {
         let rate_limit = tokio::time::sleep(MIN_DURATION_BETWEEN_BACKUP_UPDATES);
-        let mut backup_blocks: HashMap<block::Hash, PathBuf> = {
+        let backup_blocks: HashMap<block::Hash, PathBuf> = {
             let backup_dir_path = backup_dir_path.clone();
             tokio::task::spawn_blocking(move || list_backup_dir_entries(&backup_dir_path))
                 .await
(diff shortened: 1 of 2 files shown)


arya2:
Please see the contribution guide.

PR #10356, opened 2026-03-02
fix(consensus): preserve input order in `spent_outputs` for mixed chain/mempool transactions

### LINKED ISSUE #10346: `spent_outputs` ordering may not align with `tx.inputs()` for transactions spending both chain and mempool UTXOs
In `zebra-consensus/src/transaction.rs`, `spent_utxos()` builds `spent_outputs` in two passes:

- Chain UTXOs are pushed immediately as each input is iterated (in input order)
- Mempool UTXOs are collected separately and appended at the end

This means for a transaction with inputs `[A(chain), B(mempool), C(chain)]`, the resulting vector is `[A, C, B]` not `[A, B, C]`.

Any code that zips `tx.inputs()` with `all_previous_outputs()` will silently pair the wrong input with the wrong spent output for such transactions. Currently affected:

- CachedFfiTransaction::is_valid(input_index): spent_outputs flows through CachedFfiTransaction::new
- zebra_script::are_inputs_standard(): uses `zip()`
- zebra_script::p2sh_sigop_count(): uses `zip()`

The fix is to track which spent output belongs to which input index during construction, so the final vector always matches input order regardless of whether UTXOs came from chain or mempool.

oxarbitrage:
## Motivation

`spent_utxos()` built `spent_outputs` in two passes: chain UTXOs were pushed immediately in input order, mempool UTXOs were appended at the end. For a transaction with inputs `[A(chain), B(mempool), C(chain)]` the result was `[A, C, B]` instead of `[A, B, C]`. Any code indexing `all_previous_outputs` by input position would silently use the wrong UTXO, including `CachedFfiTransaction::is_valid(input_index)`. Closes #10346.

## Solution

Pre-allocate spent_outputs as `Vec<Option<Output>>` with one slot per input, and track deferred mempool outpoints with their input index. Both chain and mempool UTXOs are now written to their correct slot (`spent_outputs[input_idx]`). 
Before returning, the slots are flattened back to `Vec<Output>` and the input indices are stripped from the mempool outpoints, keeping the function's return types unchanged.

### Tests

All current test passes but no additional tests were added.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-consensus/src/transaction.rs
+++ b/zebra-consensus/src/transaction.rs
@@ -749,10 +749,13 @@ where
 
         let inputs = tx.inputs();
         let mut spent_utxos = HashMap::new();
-        let mut spent_outputs = Vec::new();
-        let mut spent_mempool_outpoints = Vec::new();
+        // Pre-allocate with None so we can fill each slot by input index, preserving input order
+        // even when chain and mempool UTXOs are fetched in separate passes.
+        let mut spent_outputs: Vec<Option<transparent::Output>> = vec![None; inputs.len()];
+        // Stores (input_idx, outpoint) for UTXOs not found in the best chain (fetched from mempool later).
+        let mut spent_mempool_outpoints: Vec<(usize, transparent::OutPoint)> = Vec::new();
 
-        for input in inputs {
+        for (input_idx, input) in inputs.iter().enumerate() {
             if let transparent::Input::PrevOut { outpoint, .. } = input {
                 tracing::trace!("awaiting outpoint lookup");
                 let utxo = if let Some(output) = known_utxos.get(outpoint) {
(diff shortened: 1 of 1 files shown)


arya2 (approved):
This looks good!

Optional, maybe better for a follow up PR: I was thinking it would be nice to refactor the fn to use a `FuturesOrdered` that fires off queries for all unknown inputs at once and both queries per unknown input at once in a `select!`, returning Some if either of them return Some/Ok or returning None otherwise.

oxarbitrage:
@mergifyio queue

conradoplg:
@mergifyio queue

conradoplg:
@Mergifyio queue

conradoplg:
@Mergifyio queue

conradoplg:
@mergifyio refresh

PR #10361, opened 2026-03-04
feat(state): Support writing non-finalized state backups to disk before broadcasting updated chainstate writing for non-finalized state

### LINKED ISSUE #10331: zebrad is not writing down everything on shutdown

arya2:
## Motivation

This PR adds support for configuring Zebra to synchronously backup the non-finalized state to disk before broadcasting each an updated chainstate to ensure that the non-finalized backup always reflects Zebra's chain state when shutting down.

<details>

The non-finalized state backup task in `backup.rs` is rate-limited to once every 5 seconds. When blocks are committed rapidly — for example, during regtest mining — many blocks may not be flushed to the backup directory before shutdown, even after the final-write fix in #10342.

</details>

Closes #10331.

## Solution

- Add `debug_skip_non_finalized_state_backup_task: bool` to `zebra_state::Config`. When `true`, the async backup task is not spawned; instead the non-finalized state is written synchronously on every block commit, before the watch channel and chain-tip channel are updated.
- Add `NonFinalizedState::write_to_backup()` — a thin synchronous wrapper around the existing `list_backup_dir_entries` / `update_non_finalized_state_backup` helpers — for use from the block write thread.
- Pass an optional `backup_dir_path` into `BlockWriteSender::spawn()` and down to `update_latest_chain_channels()`. When `Some`, it calls `write_to_backup()` before sending on the channels.
- Add `skip_backup_task: bool` parameter to `NonFinalizedState::with_backup()`. When `true` and a backup dir is configured, the directory is still created and the snapshot is restored on startup, but the async task is not spawned.
- Update the existing backup round-trip test to pass the new `skip_backup_task: false` argument.

The existing async backup path (default) is completely unaffected — the new field defaults to `false`.

### Tests

The existing `non_finalized_state_writes_blocks_to_and_restores_blocks_from_backup_cache` vector test covers the backup restore path; the new parameter was plumbed through with `false` so it continues to exercise the async task.

The sync path is intended to be exercised by the regtest mining integration test that motivated this change (adding that test is tracked as follow-up below).

`All tests pass.`

### AI Disclosure

- [x] AI tools were used: PR description and code changes were entirely Opus 4.6 with minor cleanups.

### PR Checklist

- [x] The solution is tested.


conradoplg (commented):
This looks good, thanks. But I'm wondering if it is possible to not have to introduce a new config. Do you think always writing the backup synchronously would have too bad performance?

Alternatively, wouldn't it be possible to always finish the backup when cleanly exiting the node?

arya2:
> Do you think always writing the backup synchronously would have too bad performance?

I think it's bad enough to justify adding a config field as it would reliably slow the initial sync by however long it takes to write up to ~100k blocks to disk if they're using checkpoint sync, and more otherwise.

> Alternatively, wouldn't it be possible to always finish the backup when cleanly exiting the node?

I don't think that would always be reliable particularly when large blocks are being committed quickly or if disk write IO is slow.

conradoplg (approved):

PR #10392, opened 2026-03-11
refactor: Remove deferred_pool_balance_change from ContextuallyVerifiedBlock, SemanticallyVerifiedBlock, and CheckpointVerifiedBlock; calculate it on demand

judah-caruso:
## Motivation

This refactor came up during a conversation around ZIP-234's implementation and how it could be simplified.

It was agreed that removing the `deferred_pool_balance_change` field from `ContextuallyVerifiedBlock`, `SemanticallyVerifiedBlock`, and `CheckpointVerifiedBlock` simplifies codepaths, as it should only be calculated when all of the required information is available, and isn't required before that point.

## Solution

- `deferred_pool_balance_change` is now calculated when committing a block to finalized state.

### Tests

Current test suite passes. 

Regression tests should ensure this is working as expected.

### AI Disclosure

- [ ] AI tools were used: 

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zebra-state/src/request.rs
+++ b/zebra-state/src/request.rs
@@ -258,8 +258,6 @@ pub struct SemanticallyVerifiedBlock {
     /// A precomputed list of the hashes of the transactions in this block,
     /// in the same order as `block.transactions`.
     pub transaction_hashes: Arc<[transaction::Hash]>,
-    /// This block's deferred pool value balance change.
-    pub deferred_pool_balance_change: Option<DeferredPoolBalanceChange>,
 }
 
 /// A block ready to be committed directly to the finalized state with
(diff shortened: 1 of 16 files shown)


zebra-state/src/request.rs
```diff
@@ -508,7 +522,7 @@ impl ContextuallyVerifiedBlock {
             transaction_hashes,
             chain_value_pool_change: block.chain_value_pool_change(
                 &utxos_from_ordered_utxos(spent_outputs),
-                deferred_pool_balance_change,
+                DeferredPoolBalanceChange::zero(),
```
conradoplg: I don't think this is correct - it should use the actual value here, and thus the caller must provide it. Not sure how big is the impact of this but I think that conceptually it makes sense - at the point a block is contextually verified (which it should be if we are creating a ContextuallyVerifiedBlock) we should be able to compute the deferred pool balance change.

conradoplg (changes requested):

zebra-state/src/request.rs
judah-caruso: Good catch! The change to have the caller supply the deferred_pool_balance_change was fairly small. I also added some minor tests to ensure it's passed through as expected

conradoplg:
Sorry, we merged some long-standing PRs and ended up getting conflicts in this one. Could you please solve them @judah-caruso ? I actually solved them but forgot I can't commit it here... so to save you time, the conflicting function should end up looking like this AFAIK:

```
    pub fn chain_value_pool_change(
        &self,
        utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
        deferred_pool_balance_change: DeferredPoolBalanceChange,
    ) -> Result<ValueBalance<NegativeAllowed>, ValueBalanceError> {
        // `Result<T, E>` implements `IntoIterator`, so a `flat_map(|t| t.value_balance(utxos))`
        // would silently drop transactions whose value balance returns `Err`. Use `try_fold`
        // to propagate the first error instead.
        let tx_pool_sum = self
            .transactions
            .iter()
            .try_fold(ValueBalance::<NegativeAllowed>::zero(), |acc, tx| {
                acc + tx.value_balance(utxos)?
            })?;

        Ok(*tx_pool_sum
            .neg()
            .set_deferred_amount(deferred_pool_balance_change.value()))
    }
```

There will be another error which would be easy to fix (None -> zero value)

judah-caruso:
> Sorry, we merged some long-standing PRs and ended up getting conflicts in this one. Could you please solve them @judah-caruso ? I actually solved them but forgot I can't commit it here... so to save you time, the conflicting function should end up looking like this AFAIK:
> 
> ```
>     pub fn chain_value_pool_change(
>         &self,
>         utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
>         deferred_pool_balance_change: DeferredPoolBalanceChange,
>     ) -> Result<ValueBalance<NegativeAllowed>, ValueBalanceError> {
>         // `Result<T, E>` implements `IntoIterator`, so a `flat_map(|t| t.value_balance(utxos))`
>         // would silently drop transactions whose value balance returns `Err`. Use `try_fold`
>         // to propagate the first error instead.
>         let tx_pool_sum = self
>             .transactions
>             .iter()
>             .try_fold(ValueBalance::<NegativeAllowed>::zero(), |acc, tx| {
>                 acc + tx.value_balance(utxos)?
>             })?;
> 
>         Ok(*tx_pool_sum
>             .neg()
>             .set_deferred_amount(deferred_pool_balance_change.value()))
>     }
> ```
> 
> There will be another error which would be easy to fix (None -> zero value)

No worries, just merged in your changes!

conradoplg:
@judah-caruso sorry, more merge conflicts due to the last release, but now is truly a great time to finally get this merged

conradoplg (dismissed):

conradoplg:
@judah-caruso needs `cargo fmt` :sob:

judah-caruso:
> @judah-caruso needs `cargo fmt` 😭

Sorry! Looks like my branch was in a weird state. It should be good to go now (I hope)!

conradoplg (approved):

conradoplg:
Admin-merging since the failing tests are unrelated and have been fixed in `main`, and to avoid another review roundtrip

PR #10403, opened 2026-03-16
fix(sync): Fix block propagation in regtest

### LINKED ISSUE #10332: bug(network): Zebra broadcasting in regtest mode

oxarbitrage:
## Motivation

In regtest, blocks mined on one node failed to propagate across multi-hop topologies (`A → B → C`). Node B would receive a block from A but not relay it to C, causing integration tests using `generate(N)` with N > 1 to time out during `sync_all()`.

This was caused by three issues:

- Syncer disabled in regtest (`start.rs`): The syncer task was replaced with `std::future::pending()`, leaving only gossip for propagation. Gossip only advertises the latest tip hash, so rapidly mined blocks caused intermediate blocks to be missed with no recovery path.
- `obtain_tips()` discarding small responses (`sync.rs`): A workaround for a zcashd quirk strips the last hash from `FindBlocks` responses. For short chains this left too few hashes, causing the response to be discarded and preventing recovery of the final block.
- SyncStatus::is_close_to_tip() always false (`sync/status.rs`): Gossip waits for `wait_until_close_to_tip()` before relaying blocks. On fresh regtest nodes this condition never became true, effectively blocking relay.

## Solution

- `start.rs`: Re-enable the syncer in regtest (genesis still committed locally).
- `sync.rs`:
  - Reduced delay in regtest with `REGTEST_SYNC_RESTART_DELAY` to 2s
  - Skip the zcashd last-hash stripping in regtest.
- `sync/status.rs`: Add `SyncStatus::new_for_network()`. In regtest, `is_close_to_tip()` returns true immediately so gossip can relay blocks.

### Tests

Added `regtest_block_relay.py` to the integration tests, reproducing the scenario from #10332: three nodes in a linear `A -- B -- C` topology, mining on A and verifying propagation to C.

Closes #10332.

### AI Disclosure

- [x] AI tools were used: Claude for text and some code

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zebrad/src/components/sync.rs
+++ b/zebrad/src/components/sync.rs
@@ -205,6 +205,11 @@ const FINAL_CHECKPOINT_BLOCK_VERIFY_TIMEOUT_LIMIT: HeightDiff = 100;
 /// previous sync runs.
 const SYNC_RESTART_DELAY: Duration = Duration::from_secs(67);
 
+/// In regtest, use a much shorter restart delay so that downstream nodes pick up
+/// newly-mined blocks quickly (e.g. after `generate(N)` in integration tests).
+/// The default 67-second delay exceeds the typical `sync_all` timeout of 60 seconds.
+const REGTEST_SYNC_RESTART_DELAY: Duration = Duration::from_secs(2);
+
 /// Controls how long we wait to retry a failed attempt to download
 /// and verify the genesis block.
 ///
@@ -341,6 +346,9 @@ where
 
     /// The configured full verification concurrency limit, after applying the minimum limit.
     full_verify_concurrency_limit: usize,
+
+    /// Whether the node is running on regtest. Used to apply a shorter sync restart delay.
+    is_regtest: bool,
 
     // Services
     //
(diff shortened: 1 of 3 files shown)


zebrad/src/components/sync.rs
```diff
@@ -739,13 +753,28 @@ where
...
+                    // blocks to sync (e.g. 2 new blocks), where stripping the
+                    // last hash leaves only 1 unknown hash and rchunks_exact(2)
+                    // would discard the entire response.
+                    let hashes = if self.is_regtest {
+                        hashes.as_slice()
+                    } else {
+                        match hashes.as_slice() {
+                            [] => continue,
+                            [rest @ .., _last] => rest,
+                        }
```
arya2, line 773: Could use an if guard in the match statement instead of a separate if statement

arya2 (approved):
Looks good!

PR #10418, opened 2026-03-25
chore: Update Mainnet checkpoints

upbqdn:
Taken from https://github.com/ZcashFoundation/zebra/actions/runs/23453671147/job/68239074733#step:12:254


conradoplg (approved):
Looks good, double-checked with a block explorer

PR #10432, opened 2026-03-26
docs: Update README to reference zebrad installation tag v4.3.0

upbqdn:


conradoplg (approved):
We should also update the release checklist, but let's do it later

PR #10434, opened 2026-03-28
Use batch verifier in `verify_tachyon_aggregates`

S1nus:
builds on https://github.com/tachyon-zcash/zebra/pull/41

- uses the tachyon batch verifier service during semantic block verification for tachyon bundles


natalieesk:
Hi S1nus, as this is a draft, are you looking for some feedback?

PR #10440, opened 2026-03-30
refactor: make NetworkUpgrade match arms explicit in block template generation

### LINKED ISSUE #10211: fix: remove some/all default match arms
### Motivation

https://github.com/ZcashFoundation/zebra/issues/10210 points out that some default match arms could cause problems. We should match those explicitly.

It might make sense to do that simply everywhere and enable the `clippy::wildcard_enum_match_arm` lint.

Look into it and check if that is feasible, otherwise, just remove match arms for network upgrades.

### Specifications

_No response_

_No response_

### Testing

_No response_

### Related Work

_No response_

zmanian:
## Motivation

Closes #10211

Wildcard match arms (`_ =>`) silently swallow new enum variants, which can hide bugs when new variants are added. Per review feedback, this PR is now restricted to the `NetworkUpgrade` enum, which was the motivation for the original issue.

## Solution

The branch was rebuilt from current `main` and replaces the only two remaining wildcard arms over `NetworkUpgrade` in the workspace:

- `zebra-rpc`: `proposal_block_from_template` — the pre-Canopy error arm now lists `Genesis | BeforeOverwinter | Overwinter | Sapling | Blossom | Heartwood` explicitly.
- `zebrad` acceptance test `trusted_chain_sync_handles_forks_correctly` — same treatment.

The acceptance-test wildcard was already hiding a real omission: when NU6.2 scaffolding landed (f601b696b), `proposal.rs` gained a `Nu6_2` arm but the test did not, so `Nu6_2` silently fell into the "pre-Canopy" error arm. It is now grouped with the other post-NU5 upgrades.

The cfg-gated `ZFuture` variant gets its own `#[cfg(zcash_unstable = "zfuture")]` arms so those builds keep compiling.

## Verification

- `cargo fmt --all -- --check` passes
- `cargo clippy --workspace --all-targets -- -D warnings` passes
- `cargo test -p zebra-rpc --lib` passes (69 tests)

## AI Disclosure

Used Claude Code to enumerate the remaining `NetworkUpgrade` wildcard matches and draft the replacement arms; reviewed and verified by the contributor.


zmanian:
@conradoplg This is ready for review — CI workflows need maintainer approval to run (fork PR). Could you approve the Actions runs when you get a chance? Thanks!

oxarbitrage:
Thanks for working on this @zmanian — the refactor aligns with #10211 and conradoplg's acknowledgment there.

I did a detailed review of commit `f8f812d5d8`. The approach is correct — all the replacements are mechanical, no behavior changes, and the `#[non_exhaustive]` cases (`halo2::plonk::Error`, `std::io::ErrorKind`, `ValidateContextError`) are properly handled with `#[allow]` annotations. Nice work.

Two issues to fix and a rebase needed before we can merge:

**Bug 1 — Missing `Transaction::V6` cfg arms in tests (8 sites)**

The non-test code correctly includes the cfg-gated V6 arm, but these test files don't:
- `zebra-consensus/src/transaction/tests.rs` (4 sites)
- `zebra-state/src/service/check/tests/anchors.rs` (3 sites)
- `zebrad/src/components/mempool/storage/tests/prop.rs` (2 sites)

Each needs:
```rust
#[cfg(all(zcash_unstable = "nu7", feature = "tx_v6"))]
Transaction::V6 { .. } => // same body as the other arms
```

Without this, the build breaks under `--features tx_v6` with `zcash_unstable = "nu7"`.

**Bug 2 — Missing `NetworkUpgrade::ZFuture` cfg arms (2 sites)**

- `zebra-rpc/src/methods/types/get_block_template/proposal.rs`
- `zebrad/tests/acceptance.rs`

Each needs:
```rust
#[cfg(zcash_unstable = "zfuture")]
NetworkUpgrade::ZFuture => // same body as the other arms
```

**Rebase needed**

The branch has 6 commits but only `f8f812d5d8` is the actual change — the rest are merge commits and unrelated work. Could you rebase onto current `main` with just that single commit? You'll hit 3 conflicts:

- `zebra-state/src/service/read/tests/vectors.rs` — trivial, capitalization change in a `tracing::warn!`
- `zebra-network/src/peer/connection/tests/vectors.rs` — trivial, panic format string was updated on main
- `zebra-rpc/src/methods.rs` — moderate, main restructured the `getrawtransaction` handler with `caller_block_context`. Your explicit enum variants need to be placed into the new code flow.

**Timing note:** This PR touches files that overlap with other in-flight work, so we may need to coordinate merge timing. Getting the rebase and fixes done now means it's ready when the window opens.

conradoplg:
Sorry it took this long to go back to this. I think this PR is way too big and might cause tons of conflicts with a bunch of ongoing work. I'd like for us to think more carefully about where we need explicit matches.

If you'd like to proceed with this, would you mind restricting just for the network upgrades enum, which was the motivation for the original issue?

Thanks!

zmanian:
@conradoplg Done — the PR is now restricted to the `NetworkUpgrade` enum as you suggested. I rebuilt the branch from current `main` (force-pushed), so the diff is now 2 files, +20/−2: the block-template proposal match in `zebra-rpc` and the matching one in the `zebrad` acceptance test.

Worth noting: the narrowing already paid for itself. The acceptance test's wildcard was silently swallowing `Nu6_2` — when the NU6.2 scaffolding landed, `proposal.rs` was updated but `trusted_chain_sync_handles_forks_correctly` wasn't, so `Nu6_2` fell into the "pre-Canopy" error arm. That's fixed here by listing it explicitly.

Verified locally: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p zebra-rpc --lib` all pass.

jvff (dismissed):
LGTM

gustavovalverde (approved):

PR #10448, opened 2026-04-02
fix(clippy): enable wildcard_enum_match_arm lint and resolve all violations

### LINKED ISSUE #10211: fix: remove some/all default match arms

MavenRain:
Closes #10211

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -359,5 +359,8 @@ try_err = "allow"
 fallible_impl_from = "warn"
 unwrap_in_result = "warn"
 
+# Wildcard match arms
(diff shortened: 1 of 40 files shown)


conradoplg:
There is already a PR for this, sorry

Issue #10138, opened 2025-11-27
SPIKE: Best approach for Zebra release process


natalieesk:
## Problem
We want to speed up our release process and improve the developer experience.

## Solutions
The possibilities are outlined [here](https://docs.google.com/document/d/1-Y7NyMxGM3SNfSq8nBnCcDd3Tgykfxm2WcL8FRuxg04/edit?tab=t.0#heading=h.16xti2nxnkkp).

## A/Cs

- [ ] Make a decision between Option A and Option B

natalieesk:
@arya2 is going to look at this

arya2:
Option A looks good, and I agree with @gustavovalverde that we should try it first.

conradoplg:
I tried `release-plz` with `zcash_script` and was not a fan. It tries to be too smart and if something fails in the middle of the process for whatever reason, you end up in a state where there is no obvious way to continue where it stopped. Unless it has improved since then...

arya2:
If we've already tried it elsewhere and didn't like it, Option B seems more reliable and simplifies the release process enough.

Issue #9187, opened 2025-01-31
Release Drafter went mad


upbqdn:
We currently have ~ 160 repeating draft releases for Zebra 2.2.0: https://github.com/ZcashFoundation/zebra/releases?page=17. The drafts end on page 17.

mpguerra:
Thank you for opening the issue! @gustavovalverde should be looking into this one :)

PR #10457, opened 2026-04-09
feat(release): add unified release workflow with release-plz

gustavovalverde:
## Motivation

Zebra release automation needs one reviewed path for crate publishing, GitHub Release creation, and downstream binary rollout. Release Drafter has produced noisy draft releases, and the manual checklist makes version, changelog, and public API decisions easy to drift.

Refs #10138
Refs #9187

## Solution

Add a release-plz workflow that prepares Release PRs from pushes to `main`, publishes only after a merged release-plz Release PR, and creates a draft `zebrad` GitHub Release as the human gate before Docker Hub and GCP workflows run.

This also:

- removes Release Drafter while preserving the old manual checklist as a fallback
- adds `release-pr` concurrency so rapid `main` pushes do not race the Release PR branch
- grants the `release` job `pull-requests: read` so release-plz can enforce the Release PR merge contract
- keeps manual gates for public API diffs, semver judgment, full-sync evidence, trusted publishing, and crates.io smoke checks
- avoids hardcoding protocol and Rust-version facts in generated release notes

## Scope

This PR covers the stable `main` release-plz path only. It does not implement beta, release-candidate, arbitrary-branch, AI release-notes, or private security-release workflows.

## Tests

Static checks passed locally for GitHub Actions workflow syntax, the release workflow security scan, and workspace metadata.

## AI Disclosure

- [x] AI tools were used: Claude for initial release-plz workflow analysis and design; Codex for review fixes and PR body editing.

### PR Checklist

- [x] This change was discussed with the team beforehand.
- [x] The solution is tested.


mpguerra (commented):
Had a quick glance, looks good but didn't reason about it in detail. 

Not sure if this should be added to the release workflow but we should find some way to publish the EOS height for zebra (possibly with an estimated EOS date based on height) somewhere visible for end users similar to what is done for zcashd.

gustavovalverde:
@mpguerra the initial work for the EOS date is in the following PR https://github.com/ZcashFoundation/zebra/pull/10459

conradoplg:
I'm a bit wary of using release-plz because I had problems with it when I tried. I think it was something like, if it stops in the middle for whatever reason (internet disconnect, etc.) you have to resume it manually. IIRC I think I had problems where a tag was created but the crate was not published, and when trying again it would refuse to publish because the tag already existed, or something similar.

This also does not fully capture our workflow. We use `cargo public-api` to list API changes to put in the changelog, `cargo semver-checks` just tells if a major bump is required or not (and can't distinguish between minor and patch releases).

That being said we can try this, just need to be aware of the possible issues.

upbqdn:
@gustavovalverde, am I supposed to finish this?

gustavovalverde:
@upbqdn review it

.github/workflows/release-binaries.yml
```diff
@@ -60,15 +64,15 @@ jobs:
   release-binaries-success:
     name: Release binaries success
     runs-on: ubuntu-latest
-    # Only run when the Docker Hub update job is allowed to execute
+    # Only run for the official v-prefixed zebrad release tags.
     if: >-
       ${{
         always() &&
```
jvff, line 72: Question: isn't `always()` equivalent to `true`? Do we need this line?

book/src/dev/release-process.md
```diff
@@ -128,4 +128,11 @@ To help ensure that you have sufficient time and a clear path to update, this is
 
 ## Release candidate & release process
 
-Our release checklist is available as a template, which defines each step our team needs to follow to create a new pre-release or release, and to also build and push the binaries to the official channels [Release Checklist Template](https://github.com/ZcashFoundation/zebra/blob/main/.github/PULL_REQUEST_TEMPLATE/release-checklist.md).
+Stable releases are generated from `main` by `release-plz`.
+Each push to `main` creates or updates a Release PR with version bumps and changelog updates for the crates that changed.
```
jvff: Question: do I understand correctly that there will always be a release PR open (except right after a release, until a new PR is merged to main)?

.release-plz.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# Repository URL for changelog links
+repo_url = "https://github.com/ZcashFoundation/zebra"
+
+# Timeout for waiting for each crate to appear on crates.io after publish
+publish_timeout = "30m"
+
+[changelog]
+header = ""
+body = """
```
jvff, line 71: Minor: Do we want to add any extra information for breaking changes? Maybe either adding an emoji per breaking change commit or by creating a new group 🤔

.github/PULL_REQUEST_TEMPLATE/release-checklist-legacy.md
```diff
@@ -0,0 +1,230 @@
...
+        It might also make sense to copy entries from the `zebrad` changelog.
+  - [ ] Update crate versions:
+
+```sh
+cargo release version --verbose --execute --allow-branch '*' -p <crate> patch # [ major | minor ]
+# zebrad only
+cargo release replace --verbose --execute --allow-branch '*' -p zebrad
+```
+
+- [ ] Commit and push the above version changes to the release branch.
```
jvff, line 142: Question: Just checking, but the original checklist had an item here about the `INITIAL_MIN_NETWORK_PROTOCOL_VERSION` being included in the constants file. Isn't it still needed?

.release-plz.toml
```diff
@@ -0,0 +1,83 @@
...
+# single public `zebrad` GitHub Release in .github/workflows/release.yml so
+# release-triggered Docker and GCP workflows only run once, for the node binary.
+git_release_enable = false
+
+# These commits trigger version bumps. Other commits can still appear in
+# changelogs or dependency cascades, but do not trigger a release by themselves.
+release_commits = "^(feat|fix|perf|refactor|build)"
+
+# Release PR metadata
+pr_labels = ["A-release", "P-Critical :ambulance:"]
```
jvff: Minor: should a long lived PR have a "critical" priority?

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -1,232 +1,63 @@
 ---
-name: "Release Checklist Template"
-about: "Checklist to create and publish a Zebra release"
+name: "Release Checklist"
```
jvff: Question: when is this template used? Is it used by `release-plz` when creating the release PR?

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -1,232 +1,63 @@
...
 
-- [ ] You can copy the latest checkpoints from CI by following [the zebra-checkpoints README](https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/README.md#zebra-checkpoints).
+## Merge and Monitor
 
-# Missed Dependency Updates
+- [ ] Merge the Release PR
+- [ ] Verify crates published to crates.io (check the Release workflow output)
+- [ ] Verify git tags were created
+- [ ] Verify the public `zebrad` GitHub Release was created automatically
+- [ ] Verify `zebrad` installs from crates.io: `cargo install --locked --force --version <version> zebrad && ~/.cargo/bin/zebrad`
```
jvff: Minor: couldn't this be automated, making the job fail if it's not published?

.github/workflows/release.yml
```diff
@@ -0,0 +1,197 @@
...
+          echo "::error::Configure RELEASE_APP_ID and RELEASE_APP_PRIVATE_KEY before merging a Release PR. The app token is required before crates publish so the zebrad GitHub Release can trigger Docker and GCP workflows."
+          exit 1
+      - name: Publish crates and create tags
+        id: release-plz
+        uses: release-plz/action@1528104d2ca23787631a1c1f022abb64b34c1e11 #v0.5.128
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+      - name: Create zebrad GitHub Release
```
jvff: Question: same here, doesn't `release-plz` handle this?

jvff: Nevermind this. It is a replacement for `release-plz`'s behavior, as explained in `.release-plz.toml`.

.release-plz.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# Repository URL for changelog links
+repo_url = "https://github.com/ZcashFoundation/zebra"
+
+# Timeout for waiting for each crate to appear on crates.io after publish
+publish_timeout = "30m"
+
+[changelog]
+header = ""
```
jvff, line 70: Blocking Question: raised by claude, don't we need to add the `## [Zebra {{ version }}](...)` header?

.github/workflows/release.yml
```diff
@@ -0,0 +1,197 @@
...
+            exit 1
+          fi
+
+          NOTES_FILE="$(mktemp)"
+          awk -v version="${VERSION}" '
+            /^##[[:space:]]+\[/ {
+              if (found) {
+                exit
+              }
+              if (index($0, version) > 0) {
```
jvff: Minor: claude flagged that this could match the wrong version: e.g., `5.0.2` matches `15.0.2`. It suggested the following script:

```awk
BEGIN {
  v = version
  gsub(/[.+]/, "\\&", v)              # escape regex metachars (dots) in the version
  ver_re = "[[:space:][]" v "[])]"    # version preceded by space/"[" and followed by "]"/")"
}
/^##[[:space:]]+\[/ {
  if (found) exit
  if ($0 ~ ver_re) found = 1
}
found { print }
```

jvff (changes requested):

.github/workflows/release-binaries.yml
gustavovalverde: It's not equivalent here. This job has `needs: [build]`, so without `always()` GitHub skips the aggregator if the build fails or is cancelled

book/src/dev/release-process.md
gustavovalverde: Not always. `release-plz` opens/updates the Release PR only when it finds "releasable" crate changes. I updated the docs to say that explicitly

.release-plz.toml
gustavovalverde: I just added explicit breaking-change markers in the changelog template and kept the breaking-change summary in the generated Release PR body

.github/PULL_REQUEST_TEMPLATE/release-checklist-legacy.md
gustavovalverde: Yes, this is still needed. I restored it.

.release-plz.toml
gustavovalverde: To avoid confusion here, I removed the `P-Critical` from the generated `release-plz` PR labels. We can add it manually when the release is ready to be prioritized

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
gustavovalverde: Not automatically. `release-plz` uses its configured `pr_body`, so I moved the checklist into `.release-plz.toml` to keep the template as a reference/manual fallback.

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
gustavovalverde: Yes, I added an automated zebrad crates.io install smoke in the Release workflow before creating the public GitHub Release

.release-plz.toml
gustavovalverde: Yes, I missed this one. I added the Zebra release heading for the root changelog, and preserved the normal per-crate changelog headings

.github/workflows/release.yml
gustavovalverde: Fixed

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
jvff: Nice! Should they be exactly the same? If so, we should probably see if there's a way to inject the contents into the template, instead of having to keep two places up-to-date with each other.

jvff (dismissed):

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
gustavovalverde: Simplified in the last commit.

jvff (approved):

PR #10459, opened 2026-04-09
feat(ci): automate checkpoint updates and end-of-support height

gustavovalverde:
## Motivation

Checkpoint updates require manually navigating CI logs, copying lines, and pasting into checkpoint files. The end-of-support height requires looking up the blockchain tip on a block explorer. Both are error-prone during releases.

## Solution

Automate both by capturing checkpoint CI output as artifacts and processing them in a downstream workflow.

The integration test already generates checkpoints weekly; this PR adds artifact upload so the output is machine-readable, then a `checkpoint-update.yml` workflow appends new entries, updates `ESTIMATED_RELEASE_HEIGHT`, validates the files, and opens a PR.

The EOS height uses the latest checkpoint height (a lower bound, ~3-7 days behind the tip). The 105-day EOS window makes this negligible. This avoids external dependencies and GCP credentials.

The automated PR still requires human review because checkpoints are consensus-critical.

### Tests

Validation script tested against both current checkpoint files (13,822 mainnet + 9,793 testnet entries, all passing).

### AI Disclosure

- [x] AI tools were used: Claude for checkpoint system analysis and workflow design.

### PR Checklist

- [x] This change was discussed with the team beforehand.
- [x] The solution is tested.

### DIFF
--- a/.github/workflows/checkpoint-update.yml
+++ b/.github/workflows/checkpoint-update.yml
@@ -16,13 +16,18 @@ on:
     types: [completed]
     branches: [main]
 
+  # Manual trigger for testing; resolves the latest successful integration test run automatically
+  workflow_dispatch:
+
 permissions: {}
 
 jobs:
   update-checkpoints:
     name: Update checkpoint files
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success'
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
     permissions:
       actions: read
       contents: write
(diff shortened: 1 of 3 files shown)


.github/workflows/checkpoint-update.yml
```diff
@@ -0,0 +1,194 @@
...
+      - name: Download testnet checkpoint artifact
+        id: testnet-artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c #v8.0.1
+        with:
+          name: generate-checkpoints-testnet-checkpoints
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+        continue-on-error: true
+
+      - name: Check if any artifacts were downloaded
```
arya2, line 102: In what situation might it fail to download one of the checkpoint files? And are these supposed to be `{main, test}-checkpoints.txt`?

Or are they the diff output of the checkpoint CI jobs?

gustavovalverde: The download can fail when the `generate-checkpoints-*` jobs are skipped in the upstream run (e.g., no tip disk available, full sync still in progress, or regenerate-disks mode) and, if skipped, no artifact is produced, so the download 404s and `continue-on-error` handles it.

I made changes to the file handling, and how they get created and named.

gustavovalverde:
I tested this locally, and once merged we can also run it manually from the main branch to test the whole file upload workflow.

arya2 (approved):

PR #10461, opened 2026-04-09
refactor(chain)!: replace Transaction enum with zcash_primitives newtype wrapper

arya2:
## Motivation

Zebra maintained its own `Transaction` enum with V1-V6 variants that duplicated the representation in `zcash_primitives::transaction::Transaction`. This resulted in ~1800 lines of pattern matching across variants, a separate serialization/deserialization implementation, and a `to_librustzcash()` serialize-roundtrip conversion every time the librustzcash type was needed (sighash computation, note encryption, txid for V5+).

This PR replaces the enum with a newtype wrapper around the librustzcash type, eliminating the duplication and the conversion overhead.

## Solution

Replace `pub enum Transaction { V1 { .. }, V2 { .. }, ... V6 { .. } }` with `pub struct Transaction(zcash_primitives::transaction::Transaction)`.

The new type preserves the public API through accessor methods that delegate to the inner librustzcash type, with conversions at the boundary where Zebra uses its own types (nullifiers, note commitments, tree roots). A `Deref<Target = TransactionData<Authorized>>` impl gives direct access to bundle accessors (`sprout_bundle()`, `sapling_bundle()`, `orchard_bundle()`, etc.).

**Key changes by crate:**

- **zebra-chain**: Core type replacement. New `compat` module for bidirectional type conversions. New `ZcashDeserializeWithContext<C>` trait. Serialization delegates to `zcash_primitives`.
- **zebra-consensus**: Version dispatch via `TxVersion`. Sprout JoinSplit Groth16 proof and Ed25519 signature verification reimplemented using the public `JsDescription` accessors.
- **zebra-state**: Chain nullifier tracking inlined (previously `UpdateWith` trait dispatch on the old enum fields). Sprout anchor validation uses `JsDescription` accessors.
- **zebra-script**: `Sigops::scripts()` returns `Vec<Vec<u8>>`, because inputs and outputs are now rebuilt as owned values.
- **zebra-rpc**: Sapling and Orchard RPC field serialization uses librustzcash accessor methods.
- **zebra-network/zebrad**: Mechanical adaptations (`Arc` wrapping, nullifier type propagation).

## No upstream dependency any more

The branch previously carried a `[patch.crates-io]` block pinning seven crates to the librustzcash `zebra-js-accessors` branch, which no longer exists upstream. That block has been **deleted**: the eight `JsDescription` accessors it was waiting on shipped in released **`zcash_primitives 0.27.0`**, and the workspace already builds against **0.30**. The stale `cargo-vet` policies for those git-pinned crates are removed too, so `supply-chain/config.toml` matches `main`.

This also means the `S-blocked` label was stale and has been removed.

## Takeover

@arya2 opened this PR; @oxarbitrage picked it up with their consent to bring it to a mergeable state. arya2's commits are untouched. The work added on top:

- Merged current `main`. The branch was ~500 commits behind, and **Ironwood (NU6.3, ZIP 229) and V6 transactions landed after it was cut** — main's `transaction.rs` had 40 Ironwood references and 7 public accessors that did not exist here. Porting those onto the newtype was the bulk of the effort.
- Deleted the `[patch.crates-io]` block and the stale cargo-vet policies.
- Addressed the review findings from the deep review of this PR.
- Restored the last of the previously ignored tests, and gave the proptest strategies real Orchard and Ironwood bundle generation so v5/v6 transactions are actually exercised rather than being transparent-only.
- Added a single version-aware constructor for transaction rebuilds. `TransactionData::from_parts` silently hard-codes `ironwood_bundle: None`, which caused the same bug in several places; the replacement takes the Ironwood bundle as a **required** argument so the compiler catches an omission.

**One point for @arya2 to confirm:** `1d694359a` simplified the `Transaction` `Display` impl to a one-line summary. That is a user-visible change to every log line that formats a transaction, so this PR restores `main`'s version (per-pool counts, network upgrade, lock time, expiry height). Happy to revisit if the simplification was intended to stick.

### Tests

- Full workspace builds clean; `cargo fmt`, both `clippy` matrix legs, and the workspace unit tests pass locally.
- The previously ignored tests are restored, so there is no ignored-test backlog left from this refactor. (The one remaining `#[ignore]`, `get_block_template::tests::coinbase`, is pre-existing on `main`.)
- Proptest strategies now build real shielded bundles, so the v5/v6 round-trip, sighash and consensus properties exercise Orchard and Ironwood rather than transparent-only transactions.

### Changelog

Uses the new changie fragments under `.changes/unreleased/` (14 breaking, 3 added).

There are deliberately **no `Fixed` entries**. Several bugs were found and fixed while porting the branch onto current `main`, but every one of them was introduced *by this branch* and never shipped in a release — `main` is correct in all of those paths — so there is nothing to report to users.

Verified during the original implementation against the Zcash protocol specification and active ZIPs:
- JoinSplit primary input encoding matches the `zcash_proofs/src/sprout.rs` reference implementation
- h_sig computation matches spec section 8.3 (BLAKE2b-256, personalization `ZcashComputehSig`)
- Ed25519 JoinSplit signature verification matches spec section 4.11
- vpub encoding (i64 LE) matches the reference implementation
- ZIP-211 (Canopy `vpub_old` restriction), ZIP-225 (V5 has no JoinSplits) and ZIP-317 (fee weighting) are correctly reflected

### Follow-up Work

- `RecordingReader` copies the bytes of every parsed transaction so the V4 `valueBalanceSapling` rule can read the field back. The rule now reads only that 8-byte field, so a cheaper approach should be possible.
- File spec issues for the JoinSplit primary input encoding documentation.

Resolved since this PR was opened, and no longer follow-ups: the test code is updated, Sprout JoinSplit RPC serialization (`vjoinsplit`) is implemented rather than stubbed, `PrecomputedTxData` no longer round-trips through serialization, and `zcash_primitives::transaction::Transaction` now implements `Clone` upstream in 0.30.

### AI Disclosure

- [x] AI tools were used: Claude, by both authors, for implementation, code review, spec verification, and PR title/description generation.

### PR Checklist

- [x] This change was discussed in an issue or with the team beforehand (as the follow-up to a security fix).
- [x] The solution is tested.


arya2:
Claude comparison between old and current logic:

## Logic Review: Old Transaction Enum vs New zcash_primitives Newtype

### Equivalent (no behavioral difference)

| Method | Notes |
|--------|-------|
| `has_shielded_inputs()` | Same checks: joinsplit presence, sapling spends, orchard spends+flags |
| `has_shielded_outputs()` | Same checks: joinsplit presence, sapling outputs, orchard outputs+flags |
| `is_coinbase()` | Both check `vin.len() == 1 && prevout.is_null()` |
| `coinbase_spend_restriction()` | Identical logic |
| `has_enough_orchard_flags()` | Both return true when no orchard bundle/actions, then check flags |
| `hash()` / txid | Same algorithm, just precomputed vs computed on demand |
| `inputs()` / `outputs()` | Lossless conversion via compat.rs; validated at deserialization time |
| `sprout_value_balance()` | Same formula: `Σ(vpub_new - vpub_old)` per joinsplit |
| `sapling_value_balance()` / `orchard_value_balance()` | Read directly from bundle (same wire field) |
| `ZcashSerialize` | Delegates to same zcash_primitives serialization code |
| `coinbase_height()` on Block | Same logic, just owned values instead of references |
| Nullifier/commitment iterators | Same values, owned instead of borrowed |

### Different (behavioral change, but safe in practice)

| Area | Old Behavior | New Behavior | Risk |
|------|-------------|-------------|------|
| **PartialEq** | Compared all fields including signatures | Compares txid only | **Low** — no production code compares Transaction by `==`. For V5, two txs with same inputs but different sigs have same txid (by ZIP-244 design). |
| **sprout/sapling/orchard value_balance overflow** | Returned `Result` (error on overflow) | Returns zero on overflow via `unwrap_or(ZatBalance::zero())` | **None** — overflow is impossible for valid transactions (amounts bounded by MAX_MONEY). |
| **V1-V4 consensus_branch_id** | Stored per-version (correct for each era) | Defaults to `Canopy` until corrected via `with_branch_id()` | **None** — the stored value is never read for V1-V4 in production code; sighash computation always provides the correct branch ID explicitly. |
| **Sprout anchor checking scope** | Only checked V4 Groth16 joinsplits | Checks all joinsplits (V2/V3/V4) | **None** — V2/V3 transactions are always checkpointed (pre-Canopy). The new behavior is actually more correct. |
| **`output_values_to_sprout` / `input_values_from_sprout` return type** | `Box<dyn Iterator<Item = &Amount<NonNegative>>>` | `Vec<i64>` | **None** — callers only compare against zero. Values are always non-negative in valid transactions. |

oxarbitrage:
Arya has thoroughly reviewed the large changes and is confident this works.

hilljessica8787-dot (approved):

arya2:
> One point for @arya2 to confirm: 1d694359a simplified the Transaction Display impl to a one-line summary. That is a user-visible change to every log line that formats a transaction, so this PR restores main's version (per-pool counts, network upgrade, lock time, expiry height). Happy to revisit if the simplification was intended to stick.

That's an improvement, all of the changes in this PR on top of mine look good to me.

oxarbitrage (approved):

Issue #1893, opened 2021-03-12
Zebra should auto-discover its own external IP address


teor2345:
**Is your feature request related to a problem? Please describe.**

In #1890, we allow users to manually configure an external IP address for Zebra. But we can also detect this address automatically.

**Describe the solution you'd like**

This ticket depends on #1890. It also uses the unsolicited address advertisements in #1892.

If an IPv4 or IPv6 external address is not configured:
- [x] advertise the internal bind address, if it is a public IP address
- [ ] otherwise, track the IPv4 and IPv6 addresses sent by peers in `Version.address_recv`
- [ ] advertise the most popular IPv4 and IPv6 addresses using unsolicited `Addr` messages (see #1892)
- [ ] regularly remove addresses from old handshakes, so that nodes which change addresses learn their new addresses more quickly
  - but always keep at least one IPv4 or IPv6 address

teor2345:
We probably won't do this any time soon.

Issue #1890, opened 2021-03-12
Zebra should support separate local bind and external advertise addresses


teor2345:
**Is your feature request related to a problem? Please describe.**

Currently, Zebra binds to the configured `listen_addr`, and also advertises that address to peers for inbound connections.

This doesn't work for nodes which bind to an internal address, but receive inbound connections on a different external address. (For example, NAT and firewalls.)

These nodes won't get inbound connections, because other nodes don't know their inbound addresses.

**Describe the solution you'd like**

Zebra should:
- [ ] add an `external_addr` config option, which can contain one or more IPv4 or IPv6 addresses
- [ ] advertise these addresses to other nodes in `Version` messages
- [ ] add these addresses to inbound `Peers` responses
  - [ ] make the `AddressBook` ignore any null addresses (`0.0.0.0` or `[::]`) when adding or updating peers from any source

**Describe alternatives you've considered**

We could just support a single external address, but that makes dual-stack IPv4/IPv6 impossible.
We could support multiple listener addresses.

teor2345:
We can re-open this if users ask for it.

upbqdn:
I, as a random Zebra user, am interested in having this functionality. I could run Zebra at home on a private IP, redirect its traffic through a $5 VPS (which I already pay for) with a public IP, and accept inbound connections.

teor2345:
> I, as a random Zebra user, am interested in having this functionality. I could run Zebra at home on a private IP, redirect its traffic through a 5$ VPS (that I already pay for) with a public IP, and accept inbound connections.

I just discovered a workaround for this:
Run 2 Zebra instances:
1. Port forward your external IP to the first instance
2. Add your external IP (or dynamic DNS name) to the `initial_mainnet_peers` list for the second instance, along with the default DNS seeders

Then the second instance will connect to the first instance via your external IP, and gossip its IP address.

Other Zebra instances will also auto-detect the external IPs of instances with `0.0.0.0` in their `version` messages. So if you're using the default Zcash port externally, and listening on `0.0.0.0` internally, your IP might already be gossiped on the network.

teor2345:
I'm going to re-open this, because it would have been really useful to have for the final release candidate testing.

emersonian:
+1, all of our instances are behind load balancers and cannot yet contribute to the P2P network.

mpguerra:
I've scheduled this for next sprint (Sprint 10) let's see if we can tackle this then

[...]

PR #10464, opened 2026-04-10
docs: expose P2P port for inbound peer connections

gustavovalverde:
## Motivation

Zebra's Docker image and compose files didn't publish the P2P port (8233 Mainnet, 18233 Testnet). Containerized nodes synced fine but couldn't accept inbound peers, so they contributed no upstream capacity and DNS seeders couldn't list them as reachable. The GCP deploy also bound Zebra to `0.0.0.0` with no explicit port.

## Solution

**Docker image and compose.** The Dockerfile declares `EXPOSE 8233 18233` and `docker-compose.yml` enables the `8233:8233` mapping by default. The `lwd` and `observability` compose files also publish the P2P port. `default-zebra-config.toml` documents `external_addr` for NAT and load-balancer deployments, and the `.env` template adds `ZEBRA_NETWORK__EXTERNAL_ADDR` while migrating the deprecated `ZEBRA_RPC_PORT` and `ENABLE_COOKIE_AUTH` entries to the current `ZEBRA_*__*` config-rs form.

**Documentation.** A new P2P Networking section in `docker.md` covers port requirements, NAT guidance, `external_addr`, and a port reference table. `requirements.md` gains Docker and firewall/NAT notes. `mining-docker.md` switches to `ZEBRA_RPC__LISTEN_ADDR` and publishes the P2P port. A stale reference to `docker-compose.grafana.yml` is updated to `docker-compose.observability.yml`.

**GCP deployment.** `ZEBRA_NETWORK__LISTEN_ADDR` now carries an explicit, network-specific port (`0.0.0.0:${P2P_PORT}`, 8233 or 18233) instead of the bare `0.0.0.0`.

### Not addressed in this PR

- **VPC firewall rule.** An earlier commit added an inline `gcloud firewall-rules create` step; it was removed to keep firewall management out of the deploy workflow. This should be provisioned via the GCP IaC instead.
- **Per-instance `ZEBRA_NETWORK__EXTERNAL_ADDR` on GCP.** Instance templates are immutable and shared across MIG members, so each instance can't advertise its own static IP without either `gcloud compute instances update-container` per instance, or metadata-server lookup in the entrypoint. Zebra's existing compensation (peers record observed source IPs) keeps these nodes reachable without it.

### References

- Zcash protocol: nodes advertise their address in the `Version` message's `address_from` field.
- `zebra-network/src/config.rs`: "other Zebra instances compensate for unspecified or incorrect listener addresses by adding the external IP addresses of peers to their address books."
- Related issues: #1890 (advertise external IP), #1893 (auto-discover external IP).

### Follow-up Work

- Provision the VPC firewall rule for P2P ingress via IaC.
- Set `ZEBRA_NETWORK__EXTERNAL_ADDR` per GCP instance after static-IP assignment.
- Auto-detect the external IP via the GCP metadata server in the Docker entrypoint.

### AI Disclosure

- [x] AI tools were used: Claude for analysis, implementation, and text review.

### PR Checklist

- [x] The PR title follows [conventional commits](https://www.conventionalcommits.org/) format: `type(scope): description`.
- [x] The solution is tested.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -212,10 +212,6 @@ RUN chown -R ${UID}:${GID} ${HOME}
 COPY --link --from=release /usr/local/bin/zebrad /usr/local/bin/
 COPY --link --chown=${UID}:${GID} ./docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 
-# P2P ports: 8233 (Mainnet), 18233 (Testnet)
-# Publish at runtime with -p 8233:8233 for inbound peer connections.
-EXPOSE 8233 18233
-
 ENTRYPOINT [ "entrypoint.sh" ]
 CMD ["zebrad"]
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -212,6 +212,12 @@ RUN chown -R ${UID}:${GID} ${HOME}
 COPY --link --from=release /usr/local/bin/zebrad /usr/local/bin/
 COPY --link --chown=${UID}:${GID} ./docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 
+# P2P ports: 8233 (Mainnet), 18233 (Testnet)
+# These must be published at runtime (-p 8233:8233) for inbound peer connections.
+# Without inbound P2P, the node can only make outbound connections and does not
+# contribute to network health.
+EXPOSE 8233 18233
+
 ENTRYPOINT [ "entrypoint.sh" ]
 CMD ["zebrad"]
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -213,9 +213,7 @@ COPY --link --from=release /usr/local/bin/zebrad /usr/local/bin/
 COPY --link --chown=${UID}:${GID} ./docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 
 # P2P ports: 8233 (Mainnet), 18233 (Testnet)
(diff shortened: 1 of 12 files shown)


docker/Dockerfile
```diff
@@ -212,6 +212,10 @@ RUN chown -R ${UID}:${GID} ${HOME}
 COPY --link --from=release /usr/local/bin/zebrad /usr/local/bin/
 COPY --link --chown=${UID}:${GID} ./docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 
+# P2P ports: 8233 (Mainnet), 18233 (Testnet)
+# Publish at runtime with -p 8233:8233 for inbound peer connections.
+EXPOSE 8233 18233
+
```
upbqdn: I don't think we should add this because Zebra will always bind only one of these ports. This instruction serves only as documentation, but it's misleading since only one port will be used, and we omitted the RPC ports here.

upbqdn: Docs: https://docs.docker.com/reference/dockerfile/#expose

upbqdn (dismissed):
utACK except for the `EXPOSE` instruction.

docker/Dockerfile
upbqdn:
```suggestion
```

gustavovalverde:
I'm admin-merging this as we've discussed this and the remaining comment has been resolved

Issue #9941, opened 2025-09-30
Research and potentially replace stateful `lightwalletd` tests with regtest


oxarbitrage:
Currently, Zebra’s CI runs several integration tests related to `lightwalletd`. Many of these tests require nodes that are close to the chain tip, which makes them unreliable, slow, and difficult to maintain over time.

Using regtest mode could significantly reduce sync times and improve reliability, while also lowering ongoing maintenance costs.

**Goal**

Explore replacing the current stateful `lightwalletd` tests with tests that run in regtest mode.

Specifically, we want to deprecate and remove:

- [zebrad/tests/common/lightwalletd.rs](https://github.com/ZcashFoundation/zebra/blob/b906ecd7492bbaf357f8b43fc6f66567cb4e5507/zebrad/tests/common/lightwalletd.rs)
- The [other lightwalletd test files](https://github.com/ZcashFoundation/zebra/tree/b906ecd7492bbaf357f8b43fc6f66567cb4e5507/zebrad/tests/common/lightwalletd)
- The relevant sections in [acceptance.rs](https://github.com/ZcashFoundation/zebra/blob/9213dd49d49532591c4ec634c155bb4895664053/zebrad/tests/acceptance.rs#L57-L85)

**Options**

- Integrate lightwalletd into the Python QA framework. Similar to what we already did with `zallet`.
- Replace with Rust integration tests in regtest mode.

oxarbitrage:
Research done — conclusion: **adapt the wallet-interface coverage to regtest against Zebra's native gRPC server (#10953), then delete the stateful GCP jobs.** Of the two options in the issue description, this lands on "Rust integration tests in regtest mode", with one twist: most tests shouldn't target the external `lightwalletd` binary at all.

### Why now

All four cached-state GCP jobs (`rpc-submit-block`, `lwd-rpc-send-tx`, `lwd-grpc-wallet`, `lwd-sync-update`) timed out together on the 2026-07-31 main run (https://github.com/ZcashFoundation/zebra/actions/runs/30641561945). The flakiness is structural: each test first syncs a cached mainnet state to the live tip over real peers before testing anything, so any near-tip sync stall fails the whole suite. Since NU6.3 activation the lightwalletd jobs also fail deterministically — the pinned Go binary can't parse the new transaction format (see comment below).

### Key insight

These tests conflate two concerns:

1. **The wallet-facing gRPC interface returns correct data** — the valuable coverage.
[...]

Issue #9899, opened 2025-09-16
SPIKE: Stateful tests for CI


natalieesk:
## Context

For the CI rebuild we need to decide how are we are going to handle stateful tests.

## A/C

1. Define what our stateful tests are.
2. We need to change the test architecture on Zebra first. Decide how to do this.
3. New ticket for this work. These can be high level stories but with sufficient context.

conradoplg:
Here are the stateful tests 

- sync_to_mandatory_checkpoint_mainnet/testnet: long but stateless
- sync_past_mandatory_checkpoint_mainnet/testnet: requires mandatory checkpoint state
- sync_full_mainnet/testnet: long
- sync_update_mainnet/testnet: requires synced zebra state
- lwd_sync_update: requires synced zebra and lwd states
- lwd_sync_full: requires synced zebra
- lightwalletd_test_suite: works differently depending on whether there is a synced zebra state
- lwd_rpc_test: needs synced zebra state, passes silently if not present
- lwd_rpc_send_tx: requires both synced states
- lwd_grpc_wallet: requires both synced states
- rpc_get_block_template: requires synced zebra state
- rpc_submit_block: requires synced zebra state
- generate_checkpoints_mainnet/testnet: requires synced zebra state; migrate to a tool?
- fully_synced_rpc_z_getsubtreesbyindex_snapshot_test: requires synced zebra state
- has_spending_transaction_ids: requires synced zebra state with indexer feature

I think most of them are important and we shouldn't remove them.

My suggestion for simplifying state management is the following:

- Instead of trying to update the cached states every PR, do a weekly job that updates the cached states
  - Ideally this job could also generate updated checkpoints
- PRs that updated state would need to be handled manually, we can think of some mechanism to improve that workflow but that might not even be needed.
- Eventually we should replace the lightwallet ones with Zaino ones, but not until Zaino fully works as a lightwalletd replacement

An additional idea is to also move these tests to a daily or weekly worklow, and for each PR do a similar test with a "canonical" regtest state. However that will require work and I'm not really sure if it's worth it. At the end of the day we still need to ensure that things work on mainnet so I don't think we can really afford getting rid of those tests.

There is also some possible improvements:

- Remove the usage of env vars to run tests or not. Simply mark the stateful ones with `#[ignore]` (most of them already are) and explicitly run the one test that is needed in CI
- We really need to split `acceptance.rs` into multiple files 😅

gustavovalverde:
[...]

PR #10473, opened 2026-04-13
fix(zebrad): tolerate live-network races in acceptance tests

gustavovalverde:
## Motivation

The weekly scheduled `Integration Tests on GCP` run fails on two acceptance tests against Mainnet:

- `zebrad::acceptance::rpc_submit_block` panics with `AlreadyInChain(...)` on the block it just submitted.
- `zebrad::acceptance::lwd_rpc_send_tx` panics on `assert_eq!(response, expected_response)` when the gRPC response is `SendResponse { error_code: -25, error_message: "failed to validate tx: ..., error: transaction is already in state" }` instead of the success shape.

Both tests capture a live artifact (a block past the finalized tip, or a transaction from such a block), shut zebrad down, restart it isolated from peers, and replay the artifact through the RPC. The window between capture and replay is a race against real-network finalization. When the real network advances past the capture point, the replay is a duplicate and the RPCs honestly report so.

Example of the block panic from the scheduled run: [job 70917770622](https://github.com/ZcashFoundation/zebra/actions/runs/24286263144/job/70917770622).

## Solution

Accept both outcomes in each assertion. Both prove the same property the tests are trying to check: that the RPC accepted a valid block or transaction.

- `submitblock`: accept `null` (fresh accept) or `"duplicate"` / `"duplicate-inconclusive"` (already present in chain or queue).
- `send_transaction`: accept the original success response, or an `error_code: -25` response whose `error_message` contains `"already in state"`.

### Tests

- `cargo check -p zebrad --tests` passes.
- `cargo fmt --all -- --check` passes.
- `cargo clippy -p zebrad --tests --all-targets -- -D warnings` passes.
- Runtime verification comes from the next scheduled `Integration Tests on GCP` run, which should now pass these two tests in the common race-condition case.

### AI Disclosure

- [x] AI tools were used: Claude for triaging the panics, mapping the error codes through the mempool and RPC layers, and drafting the assertion relaxations.

### DIFF
--- a/zebrad/tests/common/get_block_template_rpcs/submit_block.rs
+++ b/zebrad/tests/common/get_block_template_rpcs/submit_block.rs
@@ -74,10 +74,16 @@ pub(crate) async fn run() -> Result<()> {
         assert!(res.status().is_success());
         let res_text = res.text().await?;
 
-        // Test rpc endpoint response
+        // Accept `null` (freshly accepted) or `duplicate[-inconclusive]` (already
+        // in the chain/queue): against a live network the capture-submit window
+        // can race with finalization, and a duplicate response still proves the
+        // block was valid and the submitblock path works.
+        let accepted = res_text.contains(r#""result":null"#);
+        let duplicate = res_text.contains(r#""result":"duplicate""#)
+            || res_text.contains(r#""result":"duplicate-inconclusive""#);
         assert!(
-            res_text.contains(r#""result":null"#),
-            "unexpected response from submitblock RPC, should be null, was: {res_text}"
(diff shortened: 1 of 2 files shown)


gustavovalverde:
Closing: approach is wrong. The loosened assertion silently passes `lwd_rpc_send_tx` when transactions are already in state (they never reach the mempool, and the `!has_tx_with_shielded_elements || counter >= 1` guard lets the test pass on quiet network windows). The right fix is the regtest-based rearchitecture tracked in #9941 / #10136 / #9899, not an assertion widening.

PR #10485, opened 2026-04-14
refactor(ci): zonal MIG per (environment, branch, network, zone)

gustavovalverde:
## Motivation

The prior regional stateful MIG (one MIG per network, three instances across three zones) hits a GCP constraint: regional stateful MIGs require `max-surge=0` and `max-unavailable >= zone_count`, so rolling updates are all-at-once. The workflow's earlier `--max-unavailable=1` would have been rejected by GCP on the first rolling update. Deploy-vs-health signalling, matrix failure isolation, and per-network cache-image lookup all hang off the same topology mismatch.

## Solution

One zonal MIG per `(environment, branch, network, zone)`. Each MIG holds 1 instance, 1 stateful disk, 1 static IP.

- Push and release: matrix fans out to 6 cells (2 networks × 3 zones). Each cell deploys independently with `fail-fast: false`.
- `workflow_dispatch`: single zonal MIG (user picks network and zone) for cheap PR smoke-tests.
- `--max-unavailable=1` is natively valid on a single-instance MIG, so updates are true per-zone rolling: the zone being replaced goes down briefly while the other two keep serving.
- `deploy-nodes` reports infrastructure landed (template, MIG, IP); `verify-nodes` reports application healthy (peer mesh, chain tip) async with a 90-minute timeout. Distinct failure labels (`S-ci-fail-release-auto-issue` and `S-ci-fail-verify-auto-issue`).
- `get-disk-name` split per-network, fixing the cross-network cache-image leak.
- `ensure-health-checks` is a one-shot job (runs once per push, not once per matrix cell).
- Template uses `--disk=name=X` (attach existing). An explicit "Ensure zonal disk exists" step creates from the cache image when missing; otherwise attaches pre-seeded disks (for migrations and manual seeds).

Full rationale: [ADR 0006](docs/decisions/devops/0006-gcp-deployment-naming.md). Operations (PR-deploy cleanup, disk recovery, DB-format-version-break release, regional-to-zonal migration): [runbook](book/src/dev/gcp-deployment-operations.md).

### Tests

Empirical validation against live GCP, all reproducible:

- Primitive timing on a rolling-replace: `wait-until --version-target-reached` returned in 57s, `--stable` in 265s. Confirms deploy-landed and node-healthy separate cleanly.
- `instance-configs create --stateful-external-ip` accepts STAGING and RUNNING-UNKNOWN instances, so the prior 20-min wait before IP assignment was unnecessary.
- `--max-unavailable=1` rejected by GCP for regional stateful MIGs; accepted natively for zonal.
- Workflow_dispatch from this branch (run #24423871755) succeeded end-to-end on a dev cell, producing the expected zonal MIG, disk attachment, and labels.

Manual regional-to-zonal migration is complete: 12 zonal disks seeded from snapshots, 4 regional MIGs removed, workflow disabled during the cutover and re-enabled for the validation dispatch.

### AI Disclosure

Claude drafted the workflow refactor, ADR, runbook, and continuous-delivery updates.

### DIFF
--- a/.github/workflows/zfnd-deploy-nodes-gcp.yml
+++ b/.github/workflows/zfnd-deploy-nodes-gcp.yml
@@ -1,27 +1,15 @@
 # Deploy Zebra nodes to Google Cloud Platform.
 #
-# Two GCP environments host the deploys:
-# - `zfnd-prod-zebra` (production): receives `release` events
-# - `zfnd-dev-zebra`  (dev):        receives `push`-to-main (the staging deploy)
-#                                   and `workflow_dispatch` from any branch (PR work)
+# One zonal MIG per (environment, branch, network, zone). MIG names:
+#   - release:          zebrad-${network}-${zone-letter}
+#   - push to main:     zebrad-main-${network}-${zone-letter}
+#   - workflow_dispatch: zebrad-${branch}-${network}-${zone-letter}
 #
-# Within each environment, the MIG name encodes the source so the resource is
-# self-describing:
-# - production:   `zebrad-${network}`
-# - staging:      `zebrad-main-${network}`
-# - PR / branch:  `zebrad-${branch}-${network}`
+# Push and release fan out to 6 cells (2 networks × 3 zones);
+# workflow_dispatch deploys one cell (user picks network + zone).
 #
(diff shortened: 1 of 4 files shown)


gustavovalverde:
admin-merging as agreed with the team

PR #10501, opened 2026-04-20
refactor(benches): align with prod metrics and harden CI workflow

gustavovalverde:
Follow-up to #10444.

## Changes

**Benches**
- Group names match the `verifier` histogram labels on `zebra.consensus.batch.duration_seconds` where one exists: `groth16_sapling`, `halo2`, `redpallas`. Sprout has no batch verifier and no prod label, so its groups (`Groth16 Verification`, `Groth16 Input Preparation`) keep their original names.
- `halo2.rs` uses a fixed `BenchmarkId` and scales `Throughput::Elements` by total actions, so gh-pages series and `critcmp` survive test-vector changes.
- Shared `cycled` helper in `benches/common.rs` replaces three local copies; `autobenches = false` in `zebra-consensus` gates the helper file.
- `groth16.rs` split into verify and input-prep benches.
- Criterion noise threshold raised to 10% to match Ubuntu CI variance.

**Workflows**
- Shared `BENCHES` list across both jobs in `benchmarks.yml`.
- `book.yml` dashboard include: explicit "not ready yet" vs atomicity-violation paths instead of a silent-failure chain.
- CodeQL findings: dropped `deployments: write`, sanitized branch refs before markdown rendering.

## Verification

`cargo fmt`, `cargo clippy --benches -D warnings`, `cargo bench --no-run` on all six benches.

### AI Disclosure

- [x] Claude drafted the patch and PR body; I reviewed and ran verification.

### PR Checklist

- [x] PR name suitable for release notes
- [x] Follows contribution guidelines
- [x] Discussed with team (follow-up to #10444)
- [x] Solution is tested
- [x] Docs and changelogs up to date

PR #10502, opened 2026-04-21
refactor(benches): roll back non-prod-label bench renames

gustavovalverde:
Follow-up to #10501. Rolls back the bench renames that had no anchor in the `zebra.consensus.batch.duration_seconds{verifier=...}` prod histogram, so `critcmp` and the gh-pages time series keep row-for-row alignment with prior baselines.

## What stays renamed

Only the three group names that match a real `verifier` label in the histogram emitted by `zebra-consensus/src/primitives/*.rs`:

- `groth16_sapling`
- `halo2`
- `redpallas`

A regression on any of those histogram series now maps to a bench file by name.

## What reverts

- `groth16.rs`: `joinsplit` / `joinsplit_inputs` → `Groth16 Verification` / `Groth16 Input Preparation`; `single` → `single proof`. Sprout has no batch verifier and no prod histogram label, so the rename had nothing to anchor against.
- `transaction.rs`: `transaction_deserialize` / `transaction_serialize` → `Transaction Deserialization` / `Transaction Serialization`. `BenchmarkId::from_parameter` reverts to `BenchmarkId::new(\"deserialize\" | \"serialize\", label)`.
- `redpallas.rs`: IDs `unbatched` / `batched` → `Unbatched verification` / `Batched verification`. Group name (`redpallas`) stays.
- `sapling.rs` / `halo2.rs`: `single` → `single bundle`. Group names (`groth16_sapling`, `halo2`) stay.

## What is preserved from #10501

- Workflow hardening in `.github/workflows/benchmarks.yml` and `.github/workflows/book.yml`.
- `zebra-consensus/benches/common.rs` shared helper.
- The `halo2` stability fix: fixed `BenchmarkId` + `Throughput::Elements` scaled by total actions. That is a correctness fix, independent of naming.

## Rationale

The first PR that renames bench groups pays a one-time cost in `critcmp` because rows match on exact `group/id` strings. Paying that cost is worth it when the new name is grounded in something load-bearing (a prod histogram label). Paying it for cosmetic consistency alone turns every future comparison into partial row coverage and breaks the gh-pages history.

## Validation

- `cargo fmt --all -- --check`
- `cargo bench --no-run -p zebra-consensus --bench {groth16,halo2,sapling}`
- `cargo bench --no-run -p zebra-chain --bench {transaction,redpallas}`
- `cargo clippy -p zebra-consensus --benches -- -D warnings`
- `cargo clippy -p zebra-chain --benches --features bench -- -D warnings`

## AI disclosure

Used Claude Code to draft the rollback and this description.

### DIFF
--- a/zebra-chain/benches/redpallas.rs
+++ b/zebra-chain/benches/redpallas.rs
@@ -71,45 +71,57 @@ fn bench_batch_verify(c: &mut Criterion) {
 
         let sigs = sigs_with_distinct_keys().take(n).collect::<Vec<_>>();
 
-        group.bench_with_input(BenchmarkId::new("unbatched", n), &sigs, |b, sigs| {
-            b.iter(|| {
-                for item in sigs.iter() {
-                    match item {
-                        Item::SpendAuth { vk_bytes, sig } => {
-                            assert!(VerificationKey::try_from(*vk_bytes)
-                                .and_then(|vk| vk.verify(MESSAGE_BYTES, sig))
-                                .is_ok());
-                        }
-                        Item::Binding { vk_bytes, sig } => {
-                            assert!(VerificationKey::try_from(*vk_bytes)
-                                .and_then(|vk| vk.verify(MESSAGE_BYTES, sig))
-                                .is_ok());
+        group.bench_with_input(
+            BenchmarkId::new("Unbatched verification", n),
+            &sigs,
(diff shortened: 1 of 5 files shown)


gustavovalverde:
Superseded: rolling this back as a commit on benches-observability-followups instead (belongs in #10501, not stacked).

PR #10510, opened 2026-04-24
fix(script): Validate transparent input/output alignment before script verification

arya2:
## Motivation

The transparent script verifier in `zebra-script` previously indexed into `inputs()` and `all_previous_outputs` without validating the caller-provided request. If a caller passed an out-of-range `input_index`, or an `all_previous_outputs` vec whose length did not match `transaction.inputs().len()`, the verifier could either panic on the `inputs()[input_index]` index or silently verify against a misaligned previous output.

## Solution

- `zebra-consensus/src/script.rs`: switch from `inputs()[input_index]` to `inputs().get(input_index).ok_or(...)` so a missing input returns an error instead of panicking. Restructured the async block so the `get` lives inside it and the span is attached once at the outer level.
- `zebra-script/src/lib.rs`: in `CachedFfiTransaction::is_valid`, reject the request with `Error::TxIndex` unless `all_previous_outputs.len() == transaction.inputs().len()`, so we never verify a script against a mismatched/misaligned previous output.

## Test plan

- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] `cargo test -p zebra-script`
- [ ] `cargo test -p zebra-consensus`

## AI disclosure

Used Claude Code to draft the `is_valid` length check and this PR description.

### DIFF
--- a/zebra-consensus/src/script.rs
+++ b/zebra-consensus/src/script.rs
@@ -58,7 +58,9 @@ impl tower::Service<Request> for Verifier {
             let input = &cached_ffi_transaction
                 .inputs()
                 .get(input_index)
-                .ok_or_else(|| "cached_ffi_transaction missing input at input_index")?;
+                .ok_or_else(|| {
+                    format!("cached_ffi_transaction missing input at index {input_index}")
+                })?;
 
             match input {
                 transparent::Input::PrevOut { outpoint, .. } => {
--- a/zebra-consensus/src/script.rs
+++ b/zebra-consensus/src/script.rs
@@ -52,26 +52,28 @@ impl tower::Service<Request> for Verifier {
             cached_ffi_transaction,
             input_index,
         } = req;
-        let input = &cached_ffi_transaction.inputs()[input_index];
-        match input {
-            transparent::Input::PrevOut { outpoint, .. } => {
-                let outpoint = *outpoint;
 
-                // Avoid calling the state service if the utxo is already known
-                let span = tracing::trace_span!("script", ?outpoint);
+        let span = tracing::trace_span!("script");
+        async move {
+            let input = &cached_ffi_transaction
+                .inputs()
+                .get(input_index)
+                .ok_or_else(|| "cached_ffi_transaction missing input at input_index")?;
 
-                async move {
+            match input {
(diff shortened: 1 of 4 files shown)


upbqdn (commented):
Suggestions:

1. Error string at `zebra-consensus/src/script.rs` interpolates the literal `"input_index"` instead of the value; format with `{input_index}`.
2. New test for the mismatched-length case.

upbqdn:
@arya2, I did the suggestions and pushed. The formatting one fixed the CI failures. I thought I could request your review, but it looks like I can't? I thought GitHub used to allow it.

arya2:
@upbqdn Thank you! Note that this PR has no security fix.

upbqdn (approved):

zmanian:
Reviewing GHSA-cwfq-rfcr-8hmp against this PR, I am not seeing where #10510 enforces the specific V5 `SIGHASH_SINGLE` corresponding-output rule.

The PR clearly fixes two important index/alignment hazards:

- `Verifier::call` no longer indexes `inputs()[input_index]` directly, so an out-of-range verifier request returns an error instead of panicking.
- `CachedFfiTransaction::is_valid` rejects mismatched `all_previous_outputs.len()` vs `transaction.inputs().len()`, so scripts are not verified against misaligned previous outputs.

But those are input/previous-output alignment checks. The advisory describes a different predicate: for V5 transparent scripts, `SIGHASH_SINGLE` or `SIGHASH_SINGLE|ANYONECANPAY` should fail when `input_index >= tx.outputs().len()`.

In current `v4.4.0`, I still see `SignedOutputs::Single` mapped to `HashType::SINGLE` in `zebra-script/src/lib.rs`, and the local `zcash_primitives 0.27.0` ZIP-244 path still hashes an empty output list when `input.index() >= bundle.vout.len()`. I also do not see a mempool/block validation rule requiring one corresponding transparent output for every V5 `SIGHASH_SINGLE` transparent input.

Can someone point me to the specific pre-check or regression test that closes GHSA-cwfq? If the intended fix is #10510, I think it covers request/UTXO alignment rather than the exact corresponding-output predicate from the advisory.

PR #10512, opened 2026-04-28
feat(consensus): implement draft ZIP for shorter block target spacing

arya2:
TODO: Decide whether this should be included in NU7 or if it should be deferred to its own NU8.

## Motivation

Implements the consensus parts of the draft "Shorter Block Target Spacing" ZIP ([zcash/zips#1215](https://github.com/zcash/zips/pull/1215)) at NU7. The ZIP is still a draft (no number assigned, status `Draft`), so this is exploratory — opening early so the consensus shape is reviewable while ZIP discussion continues. Activation height is intentionally not added to the activation list and is expected in a separate deployment ZIP per the proposal.

## Solution

Three consensus changes, all gated on NU7 activation:

1. **Block target spacing 75s → 25s.** New `POST_NU7_POW_TARGET_SPACING = 25`, ratio `NU7_POW_TARGET_SPACING_RATIO = 3`. `target_spacing()` and `target_spacings()` extend to a third era. Testnet min-difficulty (`6 × spacing`) and the difficulty averaging window timespan derive from `target_spacing()` and pick up the new value automatically.

2. **Halving and per-block subsidy rescaled by 3.** `halving()` and `block_subsidy()` are unified into a single fold over `NetworkUpgrade::target_spacings(network)`: each spacing era contributes `length × spacing_seconds` "pre-Blossom block-seconds" to a running total that's divided once by `pre_blossom_halving_interval × pre_blossom_spacing_seconds`. Adding a future spacing change becomes a one-line tuple addition rather than a new branch in `halving()` / `block_subsidy()` — and the post-Blossom case it replaces falls out as a special case.

3. **New per-block shielded action limits** (`OrchardBlockActionLimit = 306`, `SaplingBlockIOLimit = 300`, `SproutBlockJoinSplitLimit = 25`, `GlobalShieldedBudget = 306`) as a single function `check::shielded_action_limits_are_valid(transactions, height, network)` that takes any iterator of `Arc<Transaction>` and returns `TransactionError`. The block verifier passes `&block.transactions`; the transaction verifier passes `std::iter::once(&tx)` so mempool admission rejects single txs whose own counts exceed any per-block limit (such a tx could never be mined). `BlockError::Transaction(TransactionError)` makes the same error variant work in both call sites without duplication.

### Tests

- Existing `cargo test --workspace --release` continues to pass (990 passed, 13 ignored).
- New `post_nu7_halving_and_subsidy` unit test exercises the post-NU7 cases of `halving()` and `block_subsidy()` on a configured Testnet, including the `BlockSubsidy(NU7) = floor(1_250_000_000 / 48) = 26_041_666` zatoshi calculation at the 3rd halving and the boundary back to era 2 just before activation.
- New `shielded_action_limits_smoke` test verifies the function is a no-op pre-NU7 across all historical block test vectors and accepts an empty (zero-shielded) post-NU7 block.
- Workspace `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

Rejection-path coverage for the action limits is left for follow-up — exercising it requires constructing transactions with > 306 fake Orchard actions, which is non-trivial without proptest plumbing.

- [zcash/zips#1215 — Shorter Block Target Spacing (draft)](https://github.com/zcash/zips/pull/1215)
- Forum: https://forum.zcashcommunity.com/t/proposal-lower-zcash-block-target-spacing-to-25s/54577
- ZIP 208 (the analogous Blossom block-time reduction) was the structural template for the spacing/halving/subsidy changes.

### Follow-up Work

- **Activation height** is deferred to a separate deployment ZIP. `Nu7` is intentionally absent from `MAINNET_ACTIVATION_HEIGHTS` and `TESTNET_ACTIVATION_HEIGHTS`, so the new rules are dormant on real networks.
- **Non-consensus SHOULD scaling** from the ZIP (`MAX_BLOCKS_IN_TRANSIT_PER_PEER`, `BLOCK_DOWNLOAD_WINDOW`, `MIN_BLOCKS_TO_KEEP`, default `txexpirydelta`, ZIP-213 anchor depth) is not in this PR — those changes are observable to operators and worth their own discussion once the ZIP advances.
- **Property/rejection tests** for `shielded_action_limits_are_valid`, ideally generating transactions that overflow each individual limit and the global budget.
- **Possible spec discrepancy worth flagging on the ZIP**: the spec note `BlockSubsidy(NU7) = floor(156_250_000 / 6) = 26_041_666 zatoshi` only matches the formula if the activation era has Halving = 3 (i.e., a third halving from Canopy has occurred before NU7 activates). At the current Halving = 2 era, the formula gives `floor(1_250_000_000 / 24) = 52_083_333`. The implementation follows the formula; the ZIP's specific numerical example may be assuming a later activation than the surrounding prose.

### AI Disclosure

- [x] AI tools were used: Claude (Opus 4.7, 1M context) for the bulk of the implementation, refactor passes, and PR description. All consensus logic was reviewed and validated against the spec by hand; tests were authored and the equivalence of the unified halving formulation to the existing per-era cascade was verified algebraically before the refactor.

### PR Checklist

- [ ] This change was discussed in an issue or with the team beforehand. *(Draft: opening for early review since the underlying ZIP is still draft and the activation deployment is separate.)*
- [x] The solution is tested.
- [ ] The documentation and changelogs are up to date. *(No `CHANGELOG.md` entry yet — the rules are dormant until a deployment ZIP wires in an activation height; will add an `[Unreleased]` entry when activation is imminent.)*

### DIFF
--- a/zebra-chain/src/parameters/network_upgrade.rs
+++ b/zebra-chain/src/parameters/network_upgrade.rs
@@ -239,6 +239,46 @@ const PRE_BLOSSOM_POW_TARGET_SPACING: i64 = 150;
 /// The target block spacing after Blossom activation.
 pub const POST_BLOSSOM_POW_TARGET_SPACING: u32 = 75;
 
+/// The target block spacing after NU7 activation, in seconds.
+///
+/// Defined by the draft "Shorter Block Target Spacing" ZIP, which lowers the
+/// block target spacing from 75 to 25 seconds at NU7.
+pub const POST_NU7_POW_TARGET_SPACING: u32 = 25;
+
+/// The ratio between the post-Blossom and post-NU7 block target spacings.
+///
+/// `PostBlossomPoWTargetSpacing / PostNU7PoWTargetSpacing = 75 / 25 = 3`.
+pub const NU7_POW_TARGET_SPACING_RATIO: u32 =
+    POST_BLOSSOM_POW_TARGET_SPACING / POST_NU7_POW_TARGET_SPACING;
+
+/// Per-block limit on the total number of Orchard actions, applied from NU7
+/// activation onwards.
+///
+/// `OrchardBlockActionLimit` in the draft "Shorter Block Target Spacing" ZIP.
+pub const ORCHARD_BLOCK_ACTION_LIMIT: u32 = 306;
+
+/// Per-block limit on the total number of Sapling spends + outputs, applied
+/// from NU7 activation onwards.
+///
+/// `SaplingBlockIOLimit` in the draft "Shorter Block Target Spacing" ZIP.
+pub const SAPLING_BLOCK_IO_LIMIT: u32 = 300;
+
+/// Per-block limit on the total number of Sprout JoinSplits, applied from NU7
+/// activation onwards.
+///
+/// `SproutBlockJoinSplitLimit` in the draft "Shorter Block Target Spacing" ZIP.
+// TODO: Remove this when disallowing
+pub const SPROUT_BLOCK_JOINSPLIT_LIMIT: u32 = 25;
+
+/// Per-block global shielded budget, applied from NU7 activation onwards.
+///
+/// Bounds the worst-case shielded sync bandwidth per block independently of
+/// which combination of pools is used. Sprout JoinSplits are weighted by 2
+/// because each JoinSplit produces 2 shielded outputs.
+///
+/// `GlobalShieldedBudget` in the draft "Shorter Block Target Spacing" ZIP.
+pub const GLOBAL_SHIELDED_BUDGET: u32 = 306;
+
 /// The averaging window for difficulty threshold arithmetic mean calculations.
 ///
 /// `PoWAveragingWindow` in the Zcash specification.
@@ -395,12 +435,13 @@ impl NetworkUpgrade {
     pub fn target_spacing(&self) -> Duration {
         let spacing_seconds = match self {
             Genesis | BeforeOverwinter | Overwinter | Sapling => PRE_BLOSSOM_POW_TARGET_SPACING,
-            Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 | Nu7 => {
+            Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 => {
                 POST_BLOSSOM_POW_TARGET_SPACING.into()
             }
+            Nu7 => POST_NU7_POW_TARGET_SPACING.into(),
 
             #[cfg(zcash_unstable = "zfuture")]
-            ZFuture => POST_BLOSSOM_POW_TARGET_SPACING.into(),
+            ZFuture => POST_NU7_POW_TARGET_SPACING.into(),
         };
 
         Duration::seconds(spacing_seconds)
@@ -423,6 +464,7 @@ impl NetworkUpgrade {
                 NetworkUpgrade::Blossom,
                 POST_BLOSSOM_POW_TARGET_SPACING.into(),
             ),
+            (NetworkUpgrade::Nu7, POST_NU7_POW_TARGET_SPACING.into()),
         ]
         .into_iter()
         .filter_map(move |(upgrade, spacing_seconds)| {
--- a/zebra-chain/src/parameters/network/subsidy.rs
+++ b/zebra-chain/src/parameters/network/subsidy.rs
@@ -421,28 +421,33 @@ pub fn halving_divisor(height: Height, network: &Network) -> Option<u64> {
 /// [7.8]: https://zips.z.cash/protocol/protocol.pdf#subsidies
 pub fn halving(height: Height, network: &Network) -> u32 {
     let slow_start_shift = network.slow_start_shift();
-    let blossom_height = NetworkUpgrade::Blossom
-        .activation_height(network)
-        .expect("blossom activation height should be available");
+    if height < slow_start_shift {
+        return 0;
+    }
 
-    let halving_index = if height < slow_start_shift {
-        0
-    } else if height < blossom_height {
-        let pre_blossom_height = height - slow_start_shift;
-        pre_blossom_height / network.pre_blossom_halving_interval()
-    } else {
-        let pre_blossom_height = blossom_height - slow_start_shift;
-        let scaled_pre_blossom_height =
-            pre_blossom_height * HeightDiff::from(BLOSSOM_POW_TARGET_SPACING_RATIO);
+    // Each era contributes (blocks × spacing_seconds) to a running total, which
+    // we divide by (PreBlossomHalvingInterval × pre_blossom_spacing) at the end.
+    // This stays in integer arithmetic across eras with different spacings, and
+    // matches the spec's segmented-fraction sum after factoring out the
+    // common denominator.
+    let pre_blossom_spacing_seconds = NetworkUpgrade::Genesis.target_spacing().num_seconds();
+    let mut total_block_seconds: HeightDiff = 0;
 
-        let post_blossom_height = height - blossom_height;
+    let mut eras = NetworkUpgrade::target_spacings(network)
+        .filter(|(era_start, _)| *era_start <= height)
+        .peekable();
 
-        (scaled_pre_blossom_height + post_blossom_height) / network.post_blossom_halving_interval()
-    };
+    while let Some((era_start, era_spacing)) = eras.next() {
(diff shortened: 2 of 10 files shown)


zebra-chain/src/parameters/network_upgrade.rs
```diff
@@ -239,6 +239,46 @@ const PRE_BLOSSOM_POW_TARGET_SPACING: i64 = 150;
...
+///
+/// `SaplingBlockIOLimit` in the draft "Shorter Block Target Spacing" ZIP.
+pub const SAPLING_BLOCK_IO_LIMIT: u32 = 300;
+
+/// Per-block limit on the total number of Sprout JoinSplits, applied from NU7
+/// activation onwards.
+///
+/// `SproutBlockJoinSplitLimit` in the draft "Shorter Block Target Spacing" ZIP.
+// TODO: Remove this when disallowing
+pub const SPROUT_BLOCK_JOINSPLIT_LIMIT: u32 = 25;
```
arya2, line 271: We already disallow v4 txs for NU7 in Zebra, do we want to exclude the sprout joinsplit limit?

```suggestion
```

ValarDragon: Makes sense to me. I think this is only needed should we end up in a situation where v4 transactions are not removed

zmanian (commented):
Thanks for opening this for early review. Walked through the diff in detail with our recent security-audit patterns in mind. Overall the unified-fold refactor in `halving()` and `block_subsidy()` is a nice structural win — I traced the algebra and the rewrite is equivalent to the old per-era cascade for pre-NU7 heights, and the post-NU7 numbers match ZIP-1215. Notes below.

### Heads-up: ZIP PR #1273 (just merged)

[zcash/zips#1273](https://github.com/zcash/zips/pull/1273) was merged earlier today and adds a consensus change this PR doesn't implement: `PoWAveragingWindow` becomes height-dependent — 17 pre-NU7, **51 post-NU7** — to preserve the 1,275-second wall-clock smoothing window across the spacing reduction (17 × 75 = 51 × 25 = 1275s). Without this, `averaging_window_timespan()` post-NU7 becomes 25 × 17 = 425s, which is exactly the short-window reactivity the ZIP's new rationale section is written to avoid.

Sites in this repo that need the same height-dependent treatment you've established for `target_spacing` here:

- `zebra-chain/src/parameters/network_upgrade.rs:245` — `pub const POW_AVERAGING_WINDOW: usize = 17`.
- `zebra-chain/src/parameters/network_upgrade.rs:493` — `averaging_window_timespan()` already uses `target_spacing()` so it picks up the new spacing, but the window count itself needs to become height-dependent so the product stays at 1275s.
- `zebra-state/src/service/check/difficulty.rs:28,238,239,257,314,318` — the consensus sites: `POW_ADJUSTMENT_BLOCK_SPAN`, the relevant-difficulty slice indexing, the median-time-past skip. These are the chain-forking ones.
- `zebra-rpc/src/methods.rs:2691` — `getnetworksolps` sampling window. Non-consensus, follows the same shape.

Probably worth either folding into this PR or splitting into a sibling consensus PR before an activation height lands. Flagging now so the structural decision (height-dependent accessor instead of `pub const`) can be made once.

### Numeric discipline in `shielded_action_limits_are_valid` (`zebra-consensus/src/block/check.rs`)

```rust
let (mut orchard_actions, mut sapling_ios, mut sprout_joinsplits): (u32, u32, u32) = (0, 0, 0);
for tx in transactions {
    orchard_actions += tx.orchard_actions().count() as u32;
    sapling_ios +=
        (tx.sapling_spends_per_anchor().count() + tx.sapling_outputs().count()) as u32;
    sprout_joinsplits += tx.joinsplit_count() as u32;
}
```

Two threads here, both about defense in depth rather than reachable bugs today:

- The `as u32` casts truncate silently if `.count()` exceeds `u32::MAX`. The comment ("Per-block counts fit in u32 because the block size is bounded to 2 MB and each shielded item is much larger than 4 bytes") justifies the *sum* via an external invariant — the 2 MB block size limit must already have run, plus per-tx serialized-size bounds, plus `TrustedPreallocate` per element vector. Per CLAUDE.md numeric-safety guidance (external/untrusted values use `saturating_*`/`checked_*`, every `as` cast comments its safety), this is exactly the kind of inferred-from-elsewhere bound the audit has been finding bugs in (closest analog: [#10556](https://github.com/ZcashFoundation/zebra/issues/10556), the non-finalized transparent-received overflow).
- The `+=` accumulators wrap (release) or panic (debug) on overflow. Under `panic = "abort"` the release-wraps case is the dangerous one: a wrapped tiny value could pass the `> ORCHARD_BLOCK_ACTION_LIMIT` check.

In practice unreachable for a real block (2 MB / ~840 B per Orchard action ≈ 2500 actions max), but cheap to harden:

```rust
let count = u64::try_from(tx.orchard_actions().count()).unwrap_or(u64::MAX);
orchard_actions = orchard_actions.saturating_add(count);
```

(Then compare against limits as `u64`.) Same pattern repeats in `zip317.rs::BlockTemplateLimits::try_add` and is safer there because it's per-tx, but using the same idiom in both places keeps the discipline consistent.

### `SYNC_RESTART_DELAY: 67 → 17` in `zebrad/src/components/sync.rs`

The PR body explicitly carves out "non-consensus SHOULD scaling" from this PR, but this constant change is exactly that — and unlike the consensus rules it's not gated on NU7, so it changes behavior for all networks today. Worth either dropping from this PR or moving it under a clear non-consensus header in the description with rationale. Right now it reads as scope creep.

### `std::iter::once(&tx)` in the mempool path

`zebra-consensus/src/transaction.rs` calls the new check with a single transaction. Mempool admission then rejects any tx whose own counts exceed the per-*block* limit, with misbehavior score 100. This is reasonable (a tx that can't fit in any block is consensus-invalid in the limit sense), but the ZIP only defines per-block limits — the per-tx-rejection-at-mempool is a Zebra policy choice. Worth one sentence in the doc comment to spell that out, and the error variant names (`OrchardActionsExceedBlockLimit` etc.) are slightly misleading when reported back over RPC for a single-tx submission. Cosmetic.

### `target_spacings` ordering invariant

`target_spacings(network)` yields `[Genesis, Blossom, Nu7]` in source order, and the new fold depends on monotonic activation heights. Real networks enforce this by construction; a misconfigured testnet (NU7 before Blossom) would produce silently wrong halvings. Not a new risk introduced by this PR, but worth a `debug_assert!` on the iterator's monotonicity now that three eras exist.

### Tests

`post_nu7_halving_and_subsidy` covers the post-NU7 numerical cases nicely. Two follow-ups worth doing before activation height lands:

- A property/synthetic test that builds transactions exceeding each individual shielded limit (mocking action vector lengths rather than constructing valid Orchard actions). This would catch the u32 overflow class directly and exercise the `GlobalShieldedBudget` cross-pool sum.
- A reverse-direction sweep that, with NU7 disabled, asserts the new `halving()` and `block_subsidy()` produce the same values as a `#[cfg(test)]` copy of the old cascade for every halving boundary and slow-start boundary on Mainnet and Testnet. The 990 existing passing tests presumably cover the common heights; pinning the boundaries explicitly would lock the algebra in regression.

### Algebra (looks correct)

Traced both rewrites:

- For pre-NU7, the new fold gives `((blossom - ss_shift) × 150 + (height - blossom) × 75) / (pre_blossom_halving_interval × 150)`, which equals the old `((scaled_pre × 2) + post_blossom) / (pre_blossom_halving_interval × 2)` after multiplying numerator and denominator by 75. The `era_start.max(slow_start_shift)` clamp is correct whether the slow-start boundary falls inside Genesis or any later era — checked both placements.
- `block_subsidy`: `MAX_BLOCK_SUBSIDY × current_spacing / 150 / halving_div`. Pre-Blossom `× 150 / 150` = identity; post-Blossom `× 75 / 150` = `/ 2`; post-NU7 `× 25 / 150` = `/ 6`. Multiplication-first max is `1.25e9 × 150 = 1.875e11`, fits u64. Matches old behavior for pre-NU7 and the ZIP-1215 numerical examples for post-NU7.

### Spec note divergence

The `BlockSubsidy(NU7)` example in the ZIP assumes a later activation than the surrounding prose — the PR description already calls this out. Right move flagging on the ZIP discussion; nothing to change in the code.

---

Disclosure: this review was assisted by Claude (Opus 4.7) running against my local audit context; I've validated each finding against the diff and the cited file locations by hand.

oxarbitrage:
Closing — arya had meant to close this shortly after opening. The underlying ZIP is still in draft, this can be a reference for future implementation.

Issue #10511, opened 2026-04-27
feature: add warnings and script to configure tcp for greatly improved full block propagation


evan-forbes:
### Motivation

By default, TCP will reset its window params after about a second between each connection. This is particularly problematic for zcash's block propagation for a few reasons.

- long block times
- single request response p2p stack
- mostly single packet messages outside of block prop
- gossiping the entire block in a single request
- pull based propagation

Not only are we **always** forcing a cold start between each peer, we're relying on that connection to upload the entire block to a peer! This is particularly terrible for peers that are physically distant and when blocks are full. Even with a 1-2Gbps connection, these two nodes are only able to utilize ~6 Mbps (💀😱)! 

<img width="1920" height="1220" alt="Image" src="https://github.com/user-attachments/assets/312a9da5-0972-41a9-81df-4710525a2141" />

The bandwidth available can increase almost 5x if we configure tcp with (in linux):
```
net.ipv4.tcp_slow_start_after_idle=0
```

### Specifications

Add either some warnings, a cargo xtask, makefile, or simply just a bash script that does the work for the user or shows them what should be changed.

The most important change is:
```
net.ipv4.tcp_slow_start_after_idle=0
```
however we could be more explict by also mentioning that cubic (the default on most linux systems), along side using fq_codel queing
```
net.core.default_qdisc=fq_codel
```

### Complex Code or Requirements

here's a bash example, however the implementation details are ofc very arbitrary

The bandwidth available can increase almost 5x if we configure tcp with (in linux):
```
#!/bin/bash

# Ensure the script is run with root privileges
if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root (use sudo)."
   exit 1
fi

CONF_FILE="/etc/sysctl.d/99-network-optimization.conf"

echo "Applying permanent network optimizations..."

# Using a Heredoc to write the configuration cleanly
cat <<EOF > "$CONF_FILE"
# Optimized Network Settings
# 1. Disable slow start after idle (Fixes the 25s idle penalty)
# This makes the largest difference
net.ipv4.tcp_slow_start_after_idle=0

# 2. Set TCP congestion control to CUBIC
# This is the default on most linux systems
net.ipv4.tcp_congestion_control=cubic

# 3. Set the default queueing discipline to fq_codel
net.core.default_qdisc=fq_codel
EOF

# Load the new settings immediately
sysctl --system

echo "-----------------------------------------------"
echo "Success! The following settings are now active:"
[...]

PR #10513, opened 2026-04-28
feat(zebrad): warn at startup if Linux TCP slow-start-after-idle is on

evan-forbes:
### Motivation

On Linux the kernel resets each TCP connection's congestion window after a short idle period (net.ipv4.tcp_slow_start_after_idle=1, the distro default). Zcash's pull-based, single-request-per-block propagation means peer connections sit idle between blocks, so every full-block transfer starts cold. On long-haul links this caps single-peer throughput far below available bandwidth.

Refs #10511.

### Solution

Adds a Linux-only check at zebrad startup that reads /proc/sys/net/ipv4/tcp_slow_start_after_idle and emits a warning when the setting is enabled, pointing operators at the docs and helper script. The check skips when the file can't be read (containers with /proc masked, non-Linux, etc.).

Also adds:
- "Linux TCP tuning for block propagation" section in book/src/user/troubleshooting.md

### Tests

The motivation is supported by the 99-node, 14-region experiment summarized in #10511.

For the change itself: local testing ofc, however no unit tests etc since this is just a warning.

- Issue: #10511
- Linux kernel docs: tcp_slow_start_after_idle (RFC 2861 behavior)
 
### Follow-up Work

We should remove these scripts after switching to QUIC

### AI Disclosure

- [X] AI tools were used: claude to search for the best spots to add the warning and earlier precedents, alongside porting the work from the issue.

### PR Checklist

- [X] The PR title follows [conventional commits](https://www.conventionalcommits.org/) format: `type(scope): description`
- [X] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [X] This change was discussed in an issue or with the team beforehand. (via slack and video call)
- [X] The solution is tested. (happy to answer any questions here)
- [X] The documentation and changelogs are up to date.


evan-forbes:
apologies for the force push, I didn't have git configured to sign properly on a new machine

scripts/tune-sysctl.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
evan-forbes: if we don't want to add a script, I'm happy to remove!

I found it useful however so figured I'd include it.

jenkin:
@evan-forbes If zebra runs in a docker container, these settings should be set inside the container or on the host?

evan-forbes:
> these settings should be set inside the container or on the host?

unfortunately we'll still have to set these on the host as docker inherits the configs form the kernel.

That's likely a very common question happy to add it to the docs as well first thing in the morning 🙂

jenkin:
I applied those changes on my nodes (mainnet and testnet), thanks!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Added
 
+- Startup warning on Linux when `net.ipv4.tcp_slow_start_after_idle` is enabled (which resets TCP congestion windows between block requests and significantly reduces single-peer block-propagation throughput on long-haul links), with a `scripts/tune-sysctl.sh` helper and a "Linux TCP tuning for block propagation" troubleshooting section ([#XXXXX](https://github.com/ZcashFoundation/zebra/pull/XXXXX))
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/ZcashFoundation/zebra/pull/10513/commits/294416d6c315dbcf2bfedfd74a98205167472240

zebrad/src/commands/start.rs
```diff
@@ -111,8 +111,48 @@ pub struct StartCmd {
...
+
+    warn!(
+        setting = "net.ipv4.tcp_slow_start_after_idle",
+        "TCP slow-start-after-idle is enabled, which resets TCP's congestion window \
+         between block requests and significantly reduces single-peer throughput for \
+         block propagation. \
+         Hint: run `sudo ./scripts/tune-sysctl.sh` from the Zebra repo, or set \
+         `net.ipv4.tcp_slow_start_after_idle=0` via sysctl. \
+         See the 'Improving Performance' section in book/src/user/troubleshooting.md."
+    );
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/ZcashFoundation/zebra/pull/10513/commits/294416d6c315dbcf2bfedfd74a98205167472240

scripts/tune-sysctl.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root (use sudo)." >&2
+    exit 1
+fi
+
+CONF_FILE="/etc/sysctl.d/99-zebra-network.conf"
+
+echo "Writing $CONF_FILE..."
+
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/ZcashFoundation/zebra/pull/10513/commits/294416d6c315dbcf2bfedfd74a98205167472240

alchemydc:
Hey @evan-forbes thanks for the PR.  

Wondering if you tested 
```
net.core.default_qdisc = fq
net.ipv4.tcp_congestion_control = bbr
```

and if so if the results were as impressive as what you got with fq_codel and forcing slow_start=0.  I suspect your approach is better but I'm curious if you tested `bbr`.  I also suspect our mileage will vary based on the load/latency of the network of various Zebra environments (colo vs residential vs cloud/vps etc).

Thanks for testing and tuning the TCP stack with Zebra.  Waiting to get some feedback from the team on whether or not we want a /scripts dir off the repo root.

Will add some docs to Z3 [suggesting TCP tuning](https://github.com/ZcashFoundation/z3/issues/36) based on your feedback also. 

DC

evan-forbes:
hey @alchemydc 

we didn't do that exact combo w/ a zebra testnet, however we did when just using tcp sending 2MB blobs. Here are the results for sending 2MB blobs from Amsterdam to Sydney. In this experiment the connections are "warm" meaning we sending blobs relatively frequently to keep the tcp params intact

583959722-ab8f2bb7-ec11-4ec1-9090-86b7e4a53ef6

We can see the first four iterations used some combination of BBR with different tcp settings with a cold connection, sending the 2MB blob every 25s (cold)

583959649-4184aa22-2008-43ef-8068-5e7d272f3295

PS: I'm a huge fan of BBR, I actually named my dog after it :sweat_smile: :nerd_face: (his name is BB for short)

evan-forbes:
> Waiting to get some feedback from the team on whether or not we want a /scripts dir off the repo root.

I'll remove this for now @alchemydc if its okay w/ you and then we can add it later if we still want it.

alchemydc:
Thanks @evan-forbes for the details.  Detecting `net.ipv4.tcp_slow_start_after_idle=1` on Linux and throwing a WARN pointing to the tuning suggestion LGTM.  The shell script though is too platform specific for the Zebra repo itself and should probably live somewhere in the [Z3 repo](https://github.com/zcashFoundation/z3).  Suggest you expand the tuning section of the book with the recommended sysctl config.  Please ensure the docs make it clear that these changes will be  *system wide* (will impact everything, not just Zebra).  

Thanks!  Appreciate the contribution.

evan-forbes:
thanks! good idea, here's the update and I'll look around in z3

https://github.com/ZcashFoundation/zebra/pull/10513/commits/6979946b0dc4dfeac53942758ddc3a2afc92a9dd

evan-forbes:
fixing linter etc now apologies the ci wasn't running on the last commit until now

Cargo.lock
```diff
@@ -7864,7 +7864,7 @@ dependencies = [
  "hex",
  "lazy_static",
  "libzcash_script",
- "rand 0.8.5",
+ "rand 0.8.6",
```
evan-forbes, line 7867: I was confused why I had to update this to fix the ci

seems like something on main merged recently caused a diff here as I'm seeing the same CI failures there

zebrad/src/components/inbound/downloads.rs
```diff
@@ -157,12 +157,11 @@ where
...
-                .expect("block download and verify tasks must not panic")
-            {
-                Ok(hash) => (Ok(hash), hash),
-                Err((e, hash, advertiser_addr)) => (Err((e, advertiser_addr)), hash),
-            };
+            let (result, hash) =
+                match join_result.expect("block download and verify tasks must not panic") {
+                    Ok(hash) => (Ok(hash), hash),
+                    Err((e, hash, advertiser_addr)) => (Err((e, advertiser_addr)), hash),
+                };
```
evan-forbes: unrelated to this PR, not sure how this got on main 

I just ran cargo fmt

https://github.com/ZcashFoundation/zebra/actions/runs/25224429795/job/73964339903

evan-forbes: happy to separate this out of this PR to fix main first before merging this PR

I'm not sure on the desired flow here yet :slightly_smiling_face:

alchemydc: > happy to separate this out of this PR to fix main first before merging this PR
> 
> I'm not sure on the desired flow here yet 🙂

Please standby.  The team is in the process of cutting a high priority release and the breakage you're seeing will get cleaned up as part of that process.

alchemydc:
Thanks for the iteration here, @evan-forbes — happy to land this.

  ### Manual testing

  I ran this on a deployed Linux VM (Debian 13, kernel 6.12.85+deb13-cloud-amd64) to confirm the new startup check behaves correctly end-to-end:

  - With `net.ipv4.tcp_slow_start_after_idle=1` (distro default): zebrad emits the warning at startup, with the hint and the docs URL pointing at the new troubleshooting section. ✅
  - After `sudo sysctl -w net.ipv4.tcp_slow_start_after_idle=0` and a restart: warning is silent. ✅

  ### About the two red checks

  Both are pure fork-permissions failures, not signal on the code:

  - **`Trigger integration tests`** fails at *Generate app token* — `actions/create-github-app-token` can't read the
  App's `app-id`/`client-id` secret because GitHub does not expose repo secrets to workflows triggered from forks. The downstream integration matrix is correctly skipped as a result.
  - **`update_release_draft`** fails with `Resource not accessible by integration` (HTTP 403) when calling the
  create-release REST endpoint — fork-PR `GITHUB_TOKEN`s are read-only (`Contents: read`), so release-drafter cannot  update the draft. This job is also only meaningful on `main`.

  Both will run cleanly on `main` after the squash-merge, under the upstream token + secrets context.

  The CI gates that actually exercise this change all passed: rustfmt, clippy `-D warnings`, `cargo deny`
  (advisories/bans/licenses/sources), per-crate builds + MSRV builds, the test matrix on Ubuntu/macOS/Windows,
  `unused-deps`, and the Docker image build. Approving and overriding via the UI.

alchemydc (approved):

PR #10494, opened 2026-04-17
fix(zebrad): apply combined security fixes

arya2:
## Motivation

Apply a set of security fixes prepared in the private security repo (`zebra-private`) for public release. These address several defects in consensus, chain parsing, network handling, and script verification.

## Solution

Cherry-picked fixes:

- **consensus**: remove mempool transaction verification bypass that allowed insufficiently-verified transactions to be accepted.
- **zebra-chain**: validate the Sapling `rk` point on parsing rather than deferring the check.
- **zebra-network**: fix the `addr` message limit to correctly bound incoming peer address payloads.
- **zebra-script**: correct V4/V5 sighash divergence from zcashd:
  - For V5+, reject undefined `hash_type` values in the script verification callback, matching `SighashType::parse`.
  - For V4, use the raw `hash_type` byte in the sighash preimage instead of canonicalizing it, matching zcashd's serialization behavior that only masks with `0x1f` for selection logic.
  - Add a `sighash_v4_raw` API in `zebra-chain` for raw-byte V4 sighash computation.
- **build**: bump `zcash_script` to 0.4.4 and pin `zcash_transparent` to 0.6.4, which carry the upstream sighash fixes.
- **misc**: ignore fuzzing artifacts; remove a stale proptest regression file; move test-only address-size constants into test modules to silence dead-code warnings.

### Tests

- New regression tests accompany each fix (rk validation, addr limit, mempool bypass, V4/V5 sighash divergence).
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are expected to pass; full CI will run the integration/sync profiles.

### Specifications & References

- Security advisories tracked privately in `zebra-private`; linked GHSA IDs to be added by maintainers on publication.

### Follow-up Work

- Coordinate GHSA publication and CHANGELOG entries alongside the release that ships this PR.

### AI Disclosure

- [ ] No AI tools were used in this PR
- [x] AI tools were used: Claude Code assisted with branch push and drafting this PR description.

### PR Checklist

- [x] The PR title follows [conventional commits](https://www.conventionalcommits.org/) format: `type(scope): description`
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [x] This change was discussed in an issue or with the team beforehand.
- [x] The solution is tested.
- [ ] The documentation and changelogs are up to date.

PR #10563, opened 2026-05-08
fix(chain): cap upfront Vec reservation in zcash_deserialize_external_count

### LINKED ISSUE #10545: bug: Vec deserializer preallocates full capacity from untrusted varint before reading data
Credit: @dingledropper for reporting this issue

This issue was created with help from Claude.ai

### What happened?

**I expected to see this happen:**
`zcash_deserialize_external_count` should allocate memory proportionally to data actually read, as zcashd does via chunked `resize()` interleaved with reads (`serialize.h:761–777`).

**Instead, this happened:**
`zcash_deserialize_external_count` calls `Vec::with_capacity(external_count)` in a single upfront allocation based on the peer-supplied varint, before reading any element data. A crafted message with an inflated count forces the full allocation regardless of how much data follows. For `block::Hash` (32 bytes × 65,535 = ~2 MiB), this is reachable via `read_getblocks` and `read_getheaders` from any peer that has completed the `version`/`verack` handshake.

This is the same defect class fixed for `AddrV1`/`AddrV2` in PR #10494 (GHSA-xr93-pcq3-pxf8), but those fixes capped the per-type `max_allocation()` rather than addressing the root cause in the deserializer.

---

Code audit of `TrustedPreallocate` impls following the AddrV1/V2 fix in PR #10494.

The root cause is in `zebra-chain/src/serialization/zcash_deserialize.rs:96`:

```rust
let mut vec = Vec::with_capacity(external_count);  // full upfront allocation
for _ in 0..external_count {
    vec.push(T::zcash_deserialize(&mut reader)?);   // data read happens after
}
```

zcashd's equivalent (`serialize.h:761–777`) avoids this by growing the vector in chunks interleaved with reads, so memory tracks data actually received rather than the claimed count:

```cpp
while (nMid < nSize) {
    nMid += 5000000 / sizeof(T);
    if (nMid > nSize) nMid = nSize;
    v.resize(nMid);
    for (; i < nMid; i++)
        Unserialize(is, v[i]);
}
```

A minimal Rust equivalent is to cap the initial allocation:

```rust
let mut vec = Vec::with_capacity(external_count.min(1024));
for _ in 0..external_count {
    vec.push(T::zcash_deserialize(&mut reader)?);
}
```

This fixes the root cause once in the core deserializer rather than requiring each `TrustedPreallocate` impl to be individually audited and capped. The `Vec` grows naturally via `push()` as real data arrives — the performance cost is negligible (a handful of amortized reallocation-copies vs. per-element deserialization cost).

`TrustedPreallocate` and `max_allocation()` remain as defense-in-depth to reject clearly bogus counts early.

---

### Zebra Version

zebrad 4.4.1 (commit `1ec1078`)

---

### Additional information

- GHSA-xr93-pcq3-pxf8 — the `AddrV1`/`AddrV2` sibling, fixed in PR #10494
- PR #10528 — `MAX_HEADERS_PER_MESSAGE` cap for `read_headers` (v4.4.0)
- zcashd `serialize.h:761–777` — chunked deserialization pattern that prevents this class of issue

### LINKED ISSUE #10548: bug: Cap `Vec::with_capacity` in `zcash_deserialize_external_count` to match actual data size
## Motivation

`zcash_deserialize_external_count` in `zebra-chain/src/serialization/zcash_deserialize.rs` passes the peer-supplied CompactSize directly to `Vec::with_capacity(external_count)`. 

When the claimed count is valid but the message is truncated or the elements are large in memory, this reserves more heap than necessary.

Reported by @dingledropper

dingledropper:
## Motivation

Closes #10545.
Closes #10548.

`zcash_deserialize_external_count` in `zebra-chain/src/serialization/zcash_deserialize.rs` passes the peer-supplied `CompactSize` directly to `Vec::with_capacity(external_count)`, allocating the full claimed capacity upfront before any element bytes are read.

This is the same root-cause defect class fixed for `AddrV1`/`AddrV2` in PR #10494 (GHSA-xr93-pcq3-pxf8), but the prior fixes capped each per-type `max_allocation()` rather than the shared deserializer. Capping at the deserializer level means new `TrustedPreallocate` impls can't recreate this issue if their per-type caps drift, and one change addresses both the malicious-lying-varint case (#10545) and the valid-but-large-elements case (#10548).

## Solution

```rust
let mut vec = Vec::with_capacity(external_count.min(1024));
```

The cap matches @mpguerra's exact suggestion in #10545. 1024 is amply above any realistic legitimate batch size (Bitcoin Core's `MAX_LOCATOR_SZ` = 101; AddrV2 `getaddr` response = 1000; `MAX_HEADERS_PER_MESSAGE` = 160) while keeping any pathological `CompactSize` under ~32 KiB of upfront reservation. The Vec grows naturally via `push()` as real elements arrive.

`TrustedPreallocate` and `max_allocation()` remain in place as defense-in-depth — the cap-at-deserializer rejection in `external_count > T::max_allocation()` (line 85) still fires before this code path runs.

## Test evidence

- `cargo test -p zebra-chain --lib` ✓ (247 tests pass)
- `cargo clippy -p zebra-chain --all-targets -- -D warnings` ✓
- Verified locally on v4.4.1 (commit 1ec1078e).

The amortized realloc-copy cost from `push()` growth is dominated by `T::zcash_deserialize` cost on every iteration. For batches under 1024 elements, behavior is unchanged.

## Context

- Sibling fix: PR #10494 (GHSA-xr93-pcq3-pxf8) capped `AddrV1`/`AddrV2` per-type caps.
- Public root-cause issues: #10545 (lying-varint shape), #10548 (large-elements shape) — same line of code, two framings.

## AI disclosure

Patch and PR description drafted with assistance from Claude (Anthropic), following the exact one-line fix @mpguerra suggested in #10545. I reviewed each step and am the responsible author.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+### Security
+
+- Cap the upfront `Vec::with_capacity` reservation in
+  `zcash_deserialize_external_count` so a peer-supplied `CompactSize`
+  cannot force a large allocation before any element bytes are read. The
+  `Vec` grows naturally via `push()` as real data arrives. Complements
+  the per-type `max_allocation()` caps from PR #10494
+  ([GHSA-xr93-pcq3-pxf8](https://github.com/ZcashFoundation/zebra/security/advisories/GHSA-xr93-pcq3-pxf8)).
+  CWE-770.
+
 ### Added
 
 - Startup warning on Linux when `net.ipv4.tcp_slow_start_after_idle` is enabled (which resets TCP congestion windows between block requests and significantly reduces single-peer block-propagation throughput on long-haul links), with a "Linux TCP tuning for block propagation" troubleshooting section ([#10513](https://github.com/ZcashFoundation/zebra/pull/10513))
--- a/zebra-chain/src/serialization/zcash_deserialize.rs
+++ b/zebra-chain/src/serialization/zcash_deserialize.rs
@@ -93,7 +93,11 @@ pub fn zcash_deserialize_external_count<R: io::Read, T: ZcashDeserialize + Trust
         // for 128 bit memory spaces.)
         Err(_) => return Err(SerializationError::Parse("Vector longer than u64::MAX")),
     }
-    let mut vec = Vec::with_capacity(external_count);
+    // Cap the upfront reservation. The Vec grows via push() as elements
+    // arrive, so a peer-supplied `external_count` can't force a large
+    // allocation before any data is read. Fixes the deserializer-level
+    // case of GHSA-xr93-pcq3-pxf8.
+    let mut vec = Vec::with_capacity(external_count.min(1024));
     for _ in 0..external_count {
         vec.push(T::zcash_deserialize(&mut reader)?);
     }
--- a/zebra-chain/src/serialization/zcash_deserialize.rs
+++ b/zebra-chain/src/serialization/zcash_deserialize.rs
@@ -4,6 +4,12 @@ use std::{io, net::Ipv6Addr, sync::Arc};
 
 use super::{AtLeastOne, CompactSizeMessage, SerializationError, MAX_PROTOCOL_MESSAGE_LEN};
 
+/// Initial-allocation cap for `zcash_deserialize_external_count`.
+///
+/// 1024 is large enough that honest messages amortize their growth to a few
+/// reallocations.
+const MAX_INITIAL_ALLOCATION: usize = 1024;
+
 /// Consensus-critical deserialization for Zcash.
 ///
 /// This trait provides a generic deserialization for consensus-critical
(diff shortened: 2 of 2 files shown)


oxarbitrage (commented):
One small suggestion: consider extracting the magic `1024` into a named constant, similar to what was done in #10591:

```rust
/// Initial-allocation cap for `zcash_deserialize_external_count`.
///
/// 1024 is large enough that honest messages amortize their growth to a few
/// reallocations.
const MAX_INITIAL_ALLOCATION: usize = 1024;
```

Then: `Vec::with_capacity(external_count.min(MAX_INITIAL_ALLOCATION))`

Makes the intent clearer and easier to grep if the cap needs adjusting later.

dingledropper:
Done in 5c72a7b — extracted to `MAX_INITIAL_ALLOCATION = 1024` at the top of the module, used at the `Vec::with_capacity` callsite, with the docstring you suggested. Thanks for the review!

oxarbitrage (approved):
Thanks

Issue #6390, opened 2023-03-23
Epic: Zebra Stable Release


mpguerra:
## Motivation

Once the Zebra Audit has successfully completed and we have fixed any findings from it we can start getting ready to tag a stable release. 

## Scope

These are the tasks that will need to be completed before we can tag a stable release (in priority order):

### Audit issues

- [x] #6672 
- [x] #6399 
- [x] #6680 
- [x] #1956

### Other issues

- [x] #6596 
- [x] #6595 
- [x] #6202 
- [x] #6149 
- [x] #5502

### Post-stable release 

- #6553 
- #1615 
- #6422 
- #1638
- #6914

mpguerra:
We have done everything we want to do here 🎉

PR #10567, opened 2026-05-09
fix(rpc): use constant-time comparison for cookie auth

### LINKED ISSUE #10546: Use constant-time comparison for RPC cookie authentication
## Motivation 

`Cookie::authenticate` in `zebra-rpc/src/server/cookie.rs` compares the supplied password against the server cookie using `String == String`, which lowers to `memcmp` with early exit on mismatch. The standard practice for secret comparison — adopted by both `zcashd` (`TimingResistantEqual`) and Bitcoin Core (PR #6390) — is constant-time equality.

## Fix

Replace the comparison with `subtle::ConstantTimeEq` (`subtle` is already a transitive dependency):

```rust
use subtle::ConstantTimeEq;

pub fn authenticate(&self, passwd: String) -> bool {
    if passwd.len() != self.0.len() {
        return false;
    }
    passwd.as_bytes().ct_eq(self.0.as_bytes()).into()
}
```

The length short-circuit is intentional — cookie length is fixed and not secret.

## Context

Credit to @dingledropper for reporting this and providing a fix.

dingledropper:
## Motivation

Closes #10546.

`Cookie::authenticate` in `zebra-rpc/src/server/cookie.rs` compared the supplied password against the server cookie using `String == String`, which lowers to a `memcmp`-style block-wise compare with early-exit on the first mismatching word. A network-side adversary observing response timing could recover the 44-character base64 cookie byte-by-byte.

Cookie auth is the only gate covering all 64 RPC methods (including mining control), and `enable_cookie_auth: true` is the default whenever RPC is enabled, so the timing oracle is reachable pre-auth from anyone with TCP access to the RPC port.

CWE-208 (Observable Timing Discrepancy).

## Solution

Replace the comparison with `subtle::ConstantTimeEq`, per @mpguerra's exact suggestion in #10546:

```rust
use subtle::ConstantTimeEq;

pub fn authenticate(&self, passwd: String) -> bool {
    if passwd.len() != self.0.len() {
        return false;
    }
    passwd.as_bytes().ct_eq(self.0.as_bytes()).into()
}
```

The length short-circuit is intentional — only the cookie *contents* are secret, not its length, which is fixed at 44 bytes by `Cookie::default` (base64 of 32 random bytes).

This mirrors `zcashd`'s `TimingResistantEqual` in `src/httprpc.cpp` (originally Bitcoin Core PR #6390, 2015).

### `subtle` as a direct dep

`subtle` was previously available transitively via `bellman` → `sapling-crypto` → `zcash_primitives` → `zcash_proofs` → `zebra-consensus`, but Rust's import-visibility rule requires direct declaration in `Cargo.toml` for `use subtle::ConstantTimeEq;` to compile. Adding it to `[workspace.dependencies]` and consuming it via `subtle = { workspace = true }` in `zebra-rpc/Cargo.toml` follows the workspace's existing dep-management convention; no version churn (already pinned to `2.6.1` in `Cargo.lock` from the transitive chain).

## Test evidence

- `cargo build -p zebra-rpc` ✓
- `cargo test -p zebra-rpc --lib server::cookie` ✓ (4 new tests pass: exact match, wrong content same length, wrong length, prefix-only match)
- `cargo clippy -p zebra-rpc --all-targets -- -D warnings` ✓
- Verified locally on `origin/main` (commit `d4cd662c`).

The new `cookie.rs` test module (`#[cfg(test)] mod tests`) covers:
- `authenticate_accepts_exact_match` — sanity check.
- `authenticate_rejects_wrong_content_same_length` — three same-length-but-wrong guesses; this is the bug class — without `ct_eq` the compare would short-circuit on the first wrong byte.
- `authenticate_rejects_wrong_length` — shorter, longer, and empty guesses; verifies the length short-circuit.
- `authenticate_rejects_prefix_only_match` — regression guard for the timing-oracle class: a 43-byte prefix match with a different last byte must be rejected exactly like any other wrong guess.

## Context

- Public root-cause issue: #10546.
- Convention reference: `zcashd`'s `TimingResistantEqual`, Bitcoin Core PR #6390.
- `subtle::ConstantTimeEq::ct_eq` is the standard Rust ecosystem primitive for this — already used by `zcash_primitives`, `bellman`, `sapling-crypto`, etc.

## CHANGELOG

Updated `CHANGELOG.md` under `[Unreleased]` → Security with the standard form referencing #10546.

## AI disclosure

Patch and PR description drafted with assistance from Claude (Anthropic), following @mpguerra's exact code snippet in #10546 plus the test-coverage shape ZF reviewers typically request. I reviewed each step and am the responsible author.


conradoplg (dismissed):
LGTM, thanks

dingledropper:
@conradoplg — the failing checks all look environmental rather than related to this change:

Unit Tests: stable passes on ubuntu + windows; beta on ubuntu fails (beta-toolchain churn) and stable on macos was cancelled at 121m (runner timeout) — those are why the required test success is red. The cookie-auth unit tests themselves pass on stable.

Test Docker Config: all RPC/cookie configs pass (RPC config, custom cookie dir, custom cache + cookie dirs); only Test Testnet config fails, which looks like a testnet/network flake unrelated to RPC auth.

Release Drafter / Trigger Integration Tests: fail with fork-PR permission errors (Resource not accessible by integration, missing client-id secret) — expected for a PR from a fork; the pull_request_target Release Drafter run passed.

I don't think any are caused by the constant-time comparison change. Could you re-run the Unit Tests (beta + macOS) so the required test success check can go green? Happy to rebase if you'd prefer.

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org).
...
 
+### Security
+
+- Use `subtle::ConstantTimeEq` for the RPC cookie authentication comparison
+  in `zebra-rpc/src/server/cookie.rs`. The previous `String == String`
+  comparison lowered to a `memcmp`-style block-wise compare with early-exit
+  on the first mismatching word, providing a timing oracle a network-side
+  adversary could use to recover the 44-character base64 cookie. Mirrors
+  `zcashd`'s `TimingResistantEqual` in `src/httprpc.cpp` (originally Bitcoin
+  Core PR #6390, 2015). CWE-208.
```
oxarbitrage: Too much detail for end users changelog, consider just:
```suggestion
- Use constant-time comparison for RPC cookie authentication
```

zebra-rpc/src/server/cookie.rs
```diff
@@ -22,8 +23,17 @@ pub struct Cookie(String);
 
 impl Cookie {
     /// Checks if the given passwd matches the contents of the cookie.
+    ///
+    /// The comparison is constant-time over the cookie length so that a
+    /// network-side adversary cannot recover the cookie byte-by-byte by
+    /// observing how long the equality check takes. The early-exit short
+    /// circuit on a length mismatch is intentional: only the cookie *contents*
+    /// are secret, not its length (which is fixed by `Cookie::default`).
```
oxarbitrage: Doc comment is longer than the code it describes. One line max or drop it.
```suggestion
    /// Constant-time comparison to prevent timing side-channels.
```

zebra-rpc/src/server/cookie.rs
```diff
@@ -86,3 +96,51 @@ pub fn remove_from_disk(dir: &Path) -> Result<()> {
 
     Ok(())
 }
+
+#[cfg(test)]
```
oxarbitrage: The core fix is 3 lines. Keep one or two tests at most

oxarbitrage (commented):
The fix seems correct and the approach is sound. `subtle` is already in the dependency tree transitively, so no new supply chain risk. The PR is just verbose for a 3-line change — see inline comments.

zebra-rpc/src/server/cookie.rs
```diff
@@ -21,9 +22,12 @@ const FILE: &str = ".cookie";
...
 impl Cookie {
-    /// Checks if the given passwd matches the contents of the cookie.
+    /// Constant-time comparison to prevent timing side-channels.
     pub fn authenticate(&self, passwd: String) -> bool {
-        *passwd == self.0
+        if passwd.len() != self.0.len() {
+            return false;
+        }
+        passwd.as_bytes().ct_eq(self.0.as_bytes()).into()
     }
```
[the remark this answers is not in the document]
oxarbitrage: Tests were intentionally removed in the latest push — the fix is 3 lines and `ct_eq` correctness is covered by the `subtle` crate itself. The PR description predates our cleanup.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+### Security
+
+- Use constant-time comparison for RPC cookie authentication
```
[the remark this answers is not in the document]
oxarbitrage: Fixed — moved the entry into the existing `### Security` section.

mpguerra (approved):
LGTM

PR #10570, opened 2026-05-09
fix(chain): cap block::Hash and CountedHeader preallocation at protocol-level constants

dingledropper:
## Motivation

`block::Hash::max_allocation` previously returned 65,535 (derived from `MAX_PROTOCOL_MESSAGE_LEN / 32`), and `CountedHeader::max_allocation` returned ~1,409 (same message-size-derived shape). Both are reachable via `Vec::zcash_deserialize` from any post-handshake peer through `getblocks`, `getheaders`, and `headers` messages. A peer can claim up to those counts and force `Vec::with_capacity` preallocation before any payload bytes are read.

The defense-in-depth fix at the deserializer level lands in PR #10563 (`Vec::with_capacity(external_count.min(1024))`); this PR caps the per-type allocation to match the actual protocol-level ceilings, mirroring Bitcoin Core's `MAX_LOCATOR_SZ` for locators and reusing Zebra's existing `MAX_HEADERS_PER_MESSAGE` for headers.

CWE-770 (Allocation of Resources Without Limits or Throttling).

## Solution

Cap at protocol-level constants:

1. **Locator (`block::Hash`)**: introduce `MAX_BLOCK_LOCATOR_LENGTH = 101` in `zebra-chain/src/block.rs`, matching Bitcoin Core's `MAX_LOCATOR_SZ` in `net_processing.cpp` (which zcashd inherits). `block::Hash::max_allocation()` now returns this constant.

2. **Headers (`CountedHeader`)**: reuse the existing `MAX_HEADERS_PER_MESSAGE = 160` constant from `zebra-chain/src/serialization/zcash_serialize.rs`, already enforced on the sending side and at the codec level for `read_headers` (PR #10528). `CountedHeader::max_allocation()` now returns `MAX_HEADERS_PER_MESSAGE as u64`. **No new constant introduced** — single source of truth across send/receive paths.

3. **Test cleanup**: `BLOCK_HEADER_LENGTH` and `MIN_COUNTED_HEADER_LEN` (used only by the old message-size-derived `max_allocation` formula in `header.rs`) move into the `preallocate` test module where they are the only consumers. Avoids `dead_code` warnings under `--features proptest-impl` with `-D warnings`.

## Test evidence

- `cargo test -p zebra-chain --lib block::tests::preallocate` ✓ (4 tests)
- `cargo clippy -p zebra-chain --all-targets -- -D warnings` ✓
- `cargo clippy -p zebra-chain --all-targets --features proptest-impl -- -D warnings` ✓

The `block_hash_max_allocation` and `counted_header_max_allocation` proptests are updated to assert the new constant-based caps:

- `Hash::max_allocation() == MAX_BLOCK_LOCATOR_LENGTH`
- `CountedHeader::max_allocation() == MAX_HEADERS_PER_MESSAGE as u64`

## Context

- Sibling fix: PR #10494 (GHSA-xr93-pcq3-pxf8) capped `AddrV1`/`AddrV2` per-type caps via `MAX_ADDRS_IN_MESSAGE`.
- Defense-in-depth at the deserializer level: PR #10563 (open) caps `Vec::with_capacity(external_count.min(1024))` at the shared `zcash_deserialize_external_count` site, so this per-type fix and the deserializer-level cap stack independently.
- `MAX_HEADERS_PER_MESSAGE = 160` is the long-standing Zcash convention from ZIP-204 and is already enforced at `zebra-network/src/protocol/external/codec.rs` for `read_headers` (PR #10528, May 2026).

## CHANGELOG

Updated `CHANGELOG.md` under `[Unreleased]` → Security with the standard form referencing the protocol-level constants and the GHSA-xr93-pcq3-pxf8 sibling.

## AI disclosure

Patch and PR description drafted with assistance from Claude (Anthropic), following @mpguerra's email-level scope ("`MAX_BLOCK_LOCATOR_LENGTH = 101` matching Bitcoin Core's `MAX_LOCATOR_SZ`" and "`CountedHeader` cap as defense-in-depth"). I reviewed each step and am the responsible author.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   the per-type `max_allocation()` caps from PR #10494
   ([GHSA-xr93-pcq3-pxf8](https://github.com/ZcashFoundation/zebra/security/advisories/GHSA-xr93-pcq3-pxf8)).
   CWE-770.
+- Cap `block::Hash::max_allocation` at `MAX_BLOCK_LOCATOR_LENGTH = 101`
+  (matching Bitcoin Core's `MAX_LOCATOR_SZ` in `net_processing.cpp`) and
+  `CountedHeader::max_allocation` at the existing
+  `MAX_HEADERS_PER_MESSAGE = 160` constant (already enforced on the
+  sending side and at the codec level for `read_headers`). The previous
+  values were derived from `MAX_PROTOCOL_MESSAGE_LEN` and returned 65,535
+  and ~1,409 respectively, allowing a post-handshake peer to force ~2 MiB
+  of upfront `Vec` preallocation per `getblocks`/`getheaders` message
+  before any payload bytes were read. Same fix shape as
+  GHSA-xr93-pcq3-pxf8 for `AddrV1`/`AddrV2` (PR #10494). CWE-770.
 
 ### Added
 
--- a/zebra-chain/src/block.rs
+++ b/zebra-chain/src/block.rs
@@ -11,7 +11,7 @@ use crate::{
     orchard,
     parameters::{Network, NetworkUpgrade},
     sapling,
-    serialization::{TrustedPreallocate, MAX_PROTOCOL_MESSAGE_LEN},
+    serialization::TrustedPreallocate,
     sprout,
     transaction::Transaction,
     transparent,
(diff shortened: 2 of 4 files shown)


natalieesk:
Thank you for your PR. We will take a look shortly.

oxarbitrage (approved):
Reviewed the full diff. The fix is correct and well-scoped:

- Hash::max_allocation: 65,535 to 101 (matches Bitcoin Core MAX_LOCATOR_SZ; legitimate locators are ~32 entries)
- CountedHeader::max_allocation: 1,409 to 160 (reuses existing MAX_HEADERS_PER_MESSAGE already enforced in the codec)
- Stacks correctly with #10563 generic 1024-element deserializer cap
- Both deserialization paths are live (read_getblocks, read_getheaders, read_headers)
- Tests updated correctly, unused constants moved to test module

Rebased onto main, resolved CHANGELOG conflict. All local checks pass (cargo test, clippy, fmt).

PR #10589, opened 2026-05-14
fix(network): guard most_recent_by_ip unwrap in address-book ban path

### LINKED ISSUE #10580: Guard `most_recent_by_ip` unwrap in address-book ban path for `max_connections_per_ip > 1`
## Motivation

The ban branch in `AddressBook::update` at `zebra-network/src/address_book.rs:455-458` unconditionally calls `.expect()` on `self.most_recent_by_ip`, which is `None` when the operator has set `network.max_connections_per_ip` to any value other than `1`. A peer that triggers a score-100 misbehavior error (e.g., a malformed proof or invalid difficulty) causes the address-book updater worker to panic.

Three sister call sites in the same file correctly guard their access with `should_update_most_recent_by_ip` / `should_remove_most_recent_by_ip` helpers that short-circuit on `None`. The ban branch is the only site that skips this guard.

Default configuration (`max_connections_per_ip = 1`) is not affected.

## Fix

Match the guard pattern used at the sister sites:

```rust
if let Some(most_recent_by_ip) = self.most_recent_by_ip.as_mut() {
    most_recent_by_ip.remove(&banned_ip);
}
```

Also consider rejecting `max_connections_per_ip != 1` at config deserialization with an explicit error until the `BTreeMap`-per-IP redesign tracked by the TODO at `address_book.rs:81` lands.

Reported by @zmanian and @DanielDerefaka

zmanian:
## Motivation

Closes #10580.

`AddressBook::update()` unconditionally unwraps `most_recent_by_ip` on the misbehavior-ban branch, but that cache is only populated when `network.max_connections_per_ip == 1`. Under the supported non-default configuration `max_connections_per_ip > 1`, a peer that reaches `MAX_PEER_MISBEHAVIOR_SCORE` (which several `zebra-consensus` errors assign as score 100) drives the ban branch and panics the address-book updater task while it holds the shared address-book mutex. The poisoned mutex then escalates into additional panics in peer-management code that locks the address book (e.g. through `AddressBookPeers for Arc<Mutex<AddressBook>>`).

This is a configuration-dependent remote availability bug; it does not affect consensus.

## Solution

- Guard `most_recent_by_ip` with `if let Some(...)` in the ban branch instead of unwrapping. The companion `should_update_most_recent_by_ip` site already remains as-is because its predicate proves the cache is populated; only the ban path needed a guard.
- Add two regression tests in `zebra-network/src/address_book/tests/vectors.rs`:
  - `misbehavior_ban_does_not_panic_with_max_connections_per_ip_above_one` — direct `AddressBook::update()` under `max_connections_per_ip = 2`, asserts the peer IP is banned and the entry is removed.
  - `misbehavior_ban_through_updater_does_not_poison_address_book` — drives the live `AddressBookUpdater` task and asserts the shared mutex is not poisoned and the IP is banned.

## Test evidence

```
cargo test -p zebra-network --lib
# test result: ok. 182 passed; 0 failed
cargo clippy -p zebra-network --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                   # clean
```

Both new regression tests fail (panic at the `expect`) on `main` and pass with this change.

## AI disclosure

Used Claude Code (Opus 4.7) for the patch, regression tests, and PR description, working from a pre-existing private investigation note that identified the panic site, preconditions, attacker-reachable misbehavior sources, and the suggested guard.

### DIFF
--- a/zebra-network/src/address_book/tests/vectors.rs
+++ b/zebra-network/src/address_book/tests/vectors.rs
@@ -11,11 +11,10 @@ use zebra_chain::{
 };
 
 use crate::{
-    address_book_updater::AddressBookUpdater,
     constants::{DEFAULT_MAX_CONNS_PER_IP, MAX_ADDRS_IN_ADDRESS_BOOK, MAX_PEER_MISBEHAVIOR_SCORE},
     meta_addr::{MetaAddr, MetaAddrChange},
     protocol::external::types::PeerServices,
-    AddressBook, Config,
+    AddressBook,
 };
 
 /// Make sure an empty address book is actually empty.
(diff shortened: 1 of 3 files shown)


zebra-network/src/address_book/tests/vectors.rs
```diff
@@ -36,6 +37,116 @@ fn address_book_empty() {
...
+
+    let guard = address_book
+        .lock()
+        .expect("address-book mutex should not be poisoned");
+    assert!(
+        guard.bans().contains_key(&banned_addr.ip()),
+        "ban-threshold misbehavior should ban the peer IP through the updater"
+    );
+}
+
```
oxarbitrage: Nit: consider dropping this test — the mutex poisoning was a consequence of the panic, not a separate bug, so the first test already covers this.

```suggestion

```

zmanian: Good call — dropped the updater test in 69d6efa55. The first test already covers the regression; the poisoning was just downstream of the panic.

oxarbitrage (approved):
Thanks.

PR #10590, opened 2026-05-14
fix(chain): propagate transaction value-balance errors in chain_value_pool_change

### LINKED ISSUE #10585: Replace `flat_map` over `Result` in `Block::chain_value_pool_change` with error-propagating iteration
## Motivation

`Block::chain_value_pool_change()` in `zebra-chain/src/block.rs` computes the block value-pool delta using `flat_map(|tx| tx.value_balance(utxos))`. Because `Result<T, E>` implements `IntoIterator`, `Ok(value)` contributes one item and `Err(_)` contributes zero — silently dropping transaction-level `ValueBalanceError` from the block-level sum.

The function returns `Result<ValueBalance<NegativeAllowed>, ValueBalanceError>` and state callers map that error into contextual validation failures. The silent drop means a transaction whose `value_balance()` returns `Err` is omitted from the sum rather than propagating the error.

Normal semantic verification appears to catch these errors earlier, so current exploitability is unconfirmed.

## Fix

Replace `flat_map` with `try_fold`:

```rust
let tx_pool_sum = self.transactions.iter().try_fold(
    ValueBalance::zero(),
    |acc, tx| {
        let tx_balance = tx.value_balance(utxos)?;
        acc + tx_balance
    },
)?;
```

Reported by @zmanian.

zmanian:
## Motivation

Closes #10585.

`Block::chain_value_pool_change()` in `zebra-chain/src/block.rs` computes the block's chain value-pool delta with:

```rust
self.transactions
    .iter()
    .flat_map(|t| t.value_balance(utxos))
    .sum::<Result<ValueBalance<NegativeAllowed>, _>>()?
```

Because `Result<T, E>` implements `IntoIterator` (yielding zero items on `Err`), a transaction whose `value_balance()` returns `Err` is silently dropped from the block sum rather than causing `chain_value_pool_change()` to return that error. The function's signature is `Result<ValueBalance<NegativeAllowed>, ValueBalanceError>` and state callers (`zebra-state/src/service/finalized_state/zebra_db/chain.rs`, `zebra-state/src/service/non_finalized_state.rs`) map that error into `CalculateBlockChainValueChange` / `AddValuePool` contextual validation failures — so the silent drop violates the documented contract.

Normal semantic verification catches transaction-level value-balance failures earlier through `miner_fees_are_valid()` and `remaining_transaction_value()`, so this is a consensus-adjacent correctness fix, not an obvious remote-acceptance bypass.

## Solution

Replace the `flat_map(Result)` aggregation with `try_fold`, which propagates the first error:

```rust
let tx_pool_sum = self
    .transactions
    .iter()
    .try_fold(ValueBalance::<NegativeAllowed>::zero(), |acc, tx| {
        acc + tx.value_balance(utxos)?
    })?;
```

`ValueBalance + ValueBalance -> Result<ValueBalance, ValueBalanceError>` is already defined, so the closure body composes naturally.

## Test evidence

Added `chain_value_pool_change_propagates_transaction_value_balance_errors` in `zebra-chain/src/block/tests/vectors.rs`. The test builds a V1 coinbase with two `MAX_MONEY` transparent outputs (each individually valid as `Amount<NonNegative>`, sum above `MAX_MONEY`), asserts the transaction's `value_balance()` returns `Err`, and now asserts that `Block::chain_value_pool_change()` also returns `Err`. The test fails on `main` (block-level helper returns `Ok` with a zero delta) and passes with this change.

```
cargo test -p zebra-chain --lib
# test result: ok. 248 passed; 0 failed
cargo clippy -p zebra-chain --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                 # clean
```

## AI disclosure

Used Claude Code (Opus 4.7) for the patch, regression test, and PR description, working from a pre-existing private investigation note that identified the silent-drop bug, the contractual mismatch, and the suggested `try_fold` shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 ### Fixed
 
 - Avoid panicking in the address-book ban path when `network.max_connections_per_ip > 1`. Guard the optional `most_recent_by_ip` cache instead of unwrapping it, so a ban-threshold misbehavior update no longer crashes the address-book updater and poisons the shared mutex ([#10580](https://github.com/ZcashFoundation/zebra/issues/10580))
+- Propagate transaction-level value-balance errors from `Block::chain_value_pool_change()` instead of silently dropping them. The previous `flat_map(Result)` aggregation relied on `Result<T, E>: IntoIterator` and yielded zero items on `Err`, so a failing transaction was omitted from the block sum rather than surfacing as a `ValueBalanceError` ([#10585](https://github.com/ZcashFoundation/zebra/issues/10585))
 
 ## [Zebra 4.4.1](https://github.com/ZcashFoundation/zebra/releases/tag/v4.4.1) - 2026-05-04
 
--- a/zebra-chain/src/block.rs
+++ b/zebra-chain/src/block.rs
@@ -230,17 +230,21 @@ impl Block {
         utxos: &HashMap<transparent::OutPoint, transparent::Utxo>,
         deferred_pool_balance_change: Option<DeferredPoolBalanceChange>,
     ) -> Result<ValueBalance<NegativeAllowed>, ValueBalanceError> {
-        Ok(*self
+        // `Result<T, E>` implements `IntoIterator`, so a `flat_map(|t| t.value_balance(utxos))`
+        // would silently drop transactions whose value balance returns `Err`. Use `try_fold`
+        // to propagate the first error instead.
+        let tx_pool_sum = self
             .transactions
             .iter()
-            .flat_map(|t| t.value_balance(utxos))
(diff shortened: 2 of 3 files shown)


oxarbitrage (approved):
Reviewed the full diff. The fix is correct:

- The flat_map over Result silently dropped errors via Result IntoIterator impl (yields zero items on Err)
- try_fold is the simplest correct approach -- zero allocations, propagates both transaction-level and accumulation overflow errors
- Defense-in-depth: semantic validation catches this earlier, but checkpoint-verified blocks skip per-transaction validation
- Test is well-structured and appropriately sized (44 lines including doc comment, consistent with file style)
- The precondition assertion (coinbase.value_balance().is_err()) is valuable for test maintainability

Rebased onto main, resolved CHANGELOG conflict. All local checks pass (cargo test, clippy, fmt).

